PR #1762, opened 2019-12-20
Changed implementation to deploy and reconcile etcd

georgekuruvillak:
**What this PR does / why we need it**:
This PR changes the reconciliation flow to involve [etcd-druid](https://github.com/gardener/etcd-druid) to perform etcd reconciliation. `etcd-druid` is deployed during seed-bootstrap phase. `etcd-druid` listens for etcd resources deployed/updated/deleted in the shoot namespaces. When an `etcd` resource is created, the `etcd-druid` adds finalizers to the dependant `secrets` such as the `tls` and the `object-store` secrets. It, then goes on to create the `statefulsets`, `services` and the `configmap` to `etcd` along with the `backup-restore` sidecar. The `ready` field in the `etcd` resource reflects the availablity of `etcd`. This is checked before `k8s-apiserver` is deployed by gardenlet. 

:rotating_light: :rotating_light: As this PR requires changes to the extensions repos(the mutating webhooks will interfere and cause erroneous update of sidecar container of etcd statefulset), it should be merged only after the respective PRs on the extension repos are merged. 
**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```action developer
With the introduction of the `etcd-druid` you have to make sure that the extension controllers that are deployed do no longer inject the backup-restore sidecar as this task is now handled by `gardener/etcd-druid` directly. The corresponding implementation in the `gardener/gardener-extensions` library has been adapted, see https://github.com/gardener/gardener-extensions/pull/603. You can also find an example adaptation for a specific provider extension here: https://github.com/gardener/gardener-extension-provider-aws/pull/42
```
```noteworthy operator
This release introduces `etcd-druid` to manage `Etcd` resources in shoot control planes.`etcd-druid` will be deployed during seed-bootstrap phase in the `garden` namespace. `etcd-druid` listens for `Etcd` resources deployed/updated/deleted in the shoot namespaces. When an `Etcd` resource is created, the `etcd-druid` adds finalizers to the dependant `secrets` such as the `tls` and the `object-store` secrets. It then goes on to create the `statefulsets`, `services` and the `configmap` to deploy `etcd` along with the `backup-restore` sidecar. The `ready` field in the `Etcd` resource reflects the availability of the related etcd `statefulsets`. This is checked before `kube-apiserver` is deployed by gardenlet.
```

PR #1869, opened 2020-01-28
Changed names of etcd-main and etcd-events contants.

georgekuruvillak:
`etcd-main` and `etcd-events` have the prefix `StatefulsetName` in the name for the constants. As we shall not be deploying Statefulsets directly, it no longer makes sense to keep this prefix.
Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/apis/core/v1alpha1/constants/types_constants.go
+++ b/pkg/apis/core/v1alpha1/constants/types_constants.go
@@ -97,12 +97,10 @@ const (
 	// StatefulSetNameAlertManager is a constant for the name of a Kubernetes stateful set object that contains
 	// the alertmanager pod.
 	StatefulSetNameAlertManager = "alertmanager"
-	// StatefulSetNameETCDMain is a constant for the name of a Kubernetes stateful set object that contains
-	// the etcd-main pod.
-	StatefulSetNameETCDMain = "etcd-main"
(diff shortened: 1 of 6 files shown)


danielfoehrKn (approved):

pkg/apis/core/v1alpha1/constants/types_constants.go
```diff
@@ -97,12 +97,10 @@ const (
 	// StatefulSetNameAlertManager is a constant for the name of a Kubernetes stateful set object that contains
 	// the alertmanager pod.
 	StatefulSetNameAlertManager = "alertmanager"
-	// StatefulSetNameETCDMain is a constant for the name of a Kubernetes stateful set object that contains
```
rfranzke, line 100: Can we keep those old constants and use it in other files where we still make assumptions that a `StatefulSet` is the result of this `Etcd` resource? E.g., https://github.com/gardener/gardener/pull/1869/files#diff-398162f3b552a38fdda6010a922a5998R497

rfranzke: OK, just discussed with @georgekuruvillak, the health checks or any direct usage of the `StatefulSet` will be removed as part of #1762. Instead, the care controller will fetch the health status information from the `status` of the `Etcd` custom resource.
/lgtm

rfranzke (approved):

PR #1870, opened 2020-01-29
Cleanup kubeconfig usage in Gardenlet Helm chart

rfranzke:
This PR cleans up the confusing usage of the `kubeconfig` value for the Gardenlet Helm chart.
Now, it is clear how the kubeconfig for the Garden cluster and how the kubeconfig for the seed cluster can be overwritten.

**Release note**:

```action operator
The `.gardenlet.kubeconfig` field in the Gardenlet Helm chart is no longer available. If you want to provide a kubeconfig for the garden cluster you have to use `.gardenlet.config.gardenClientConnection.kubeconfig` now. Also, you now have the possibility to overwrite the kubeconfig for the seed cluster by specifying `.gardenlet.config.seedClientConnection.kubeconfig`.
```


charts/gardener/gardenlet/values.yaml
```diff
@@ -38,11 +38,17 @@ global:
       # kubeconfigSecret:
```
timebertt: Can you add a comment here, stating that this is the secret that the generated kubeconfig from TLS bootstrapping will be stored in, please?
Could be something along the lines from the Config API doc: https://github.com/gardener/gardener/blob/5b69a02f511af5d085da8976f3aa5345c354c44d/pkg/gardenlet/apis/config/v1alpha1/types.go#L94-L96

timebertt (changes requested):
Generally, /lgtm
Only one minor remark to make it even easier to understand the different values fields.

charts/gardener/gardenlet/values.yaml
rfranzke: Done, thanks!

timebertt (approved):
Thanks, /approve

PR #1872, opened 2020-01-30
[GEP-8] Update SNI GEP to use PROXY protocol

mvladev:
This adds the benefit that TCP traffic is not wrapped in TCP.

Fixes # n/a

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/docs/proposals/08-shoot-apiserver-via-sni.md
+++ b/docs/proposals/08-shoot-apiserver-via-sni.md
@@ -206,136 +206,168 @@ Another issue occurs when the client tries to talk to the apiserver via the in-c
 
 To mitigate this problem an additional proxy must be deployed on every single Node. It does not terminate TLS and sends the traffic to the correct Shoot API Server. This is achieved by:
 
-- the apiserver master service reconciler is stopped (`--endpoint-reconciler-type=none`).
+- the apiserver master service reconciler is started and pointing to the `kube-apiserver`'s Cluster IP in the Seed cluster (e.g. `--advertise-address=10.1.2.3`).
 - the proxy runs in the host network of the `Node`.
 - the proxy has a sidecar container which:
-  - creates a dummy network interface and assigns the master service IP address to it.
-  - enables (with iptables) traffic to that IP address (kube-proxy rejects all traffic to `Services` without endpoints) and it removes connection tracking (conntrack) as the IP address is local to the `Node`.
-- the proxy needs to send multiple TLS streams over a single TCP connection (to reduce performance overhead).
-- the proxy listens on the master service IP address and it sends traffic to the correct API server.
+  - creates a dummy network interface and assigns the `10.1.2.3` to it.
+  - removes connection tracking (conntrack) if iptables/nftables is enabled as the IP address is local to the `Node`.
+- the proxy listens on the `10.1.2.3` and using the [PROXY protocol](http://www.haproxy.org/download/2.0/doc/proxy-protocol.txt) it sends the data stream to the Envoy ingress gateway (EIGW).
+- EIGW listens for PROXY protocol on a dedicated `8443` port. EIGW reads the destination IP + port from the PROXY protocol and forwards traffic to the correct upstream apiserver.
 
 The sidecar is a standalone component. It's possible to transparently change the proxy implementation without any modifications to the sidecar. The simplified flow looks like:
 
 ```text
-                                              +------------------+
-                                              |                  |
-                                              | Shoot API Server |
-                                              |                  |
-                                              | Cluster A        |
-                                              |                  |
-                                              +------------------+
-                                                        ^
-                                                        |
-                                                        |
-                                                        |
-                                                        |
-  +-----------------------------------------------------------------------+
-                                                        |   Shoot Cluster
-                                                        |
-                                                        |
-                                                        |
-                                                        |
-                                                        |
-                                                        |
-   +---------------------+                   +---------------------+
-   |                     |                   |                     |
-   |  Pod talking to     |                   |                     |
-   |  the kubernetes     |                   |       Proxy         |
-   |  service            +------------------>+  No TLS termination |
-   |                     |                   |                     |
-   |                     |                   |                     |
-   +---------------------+                   +---------------------+
++------------------+                    +----------------+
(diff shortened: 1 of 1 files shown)


docs/proposals/08-shoot-apiserver-via-sni.md
```diff
@@ -206,136 +206,102 @@ Another issue occurs when the client tries to talk to the apiserver via the in-c
...
-  - enables (with iptables) traffic to that IP address (kube-proxy rejects all traffic to `Services` without endpoints) and it removes connection tracking (conntrack) as the IP address is local to the `Node`.
-- the proxy needs to send multiple TLS streams over a single TCP connection (to reduce performance overhead).
-- the proxy listens on the master service IP address and it sends traffic to the correct API server.
+  - creates a dummy network interface and assigns the `10.1.2.3` to it.
+  - removes connection tracking (conntrack) as the IP address is local to the `Node`.
+- the proxy listens on the `10.1.2.3` and using the [PROXY protocol](http://www.haproxy.org/download/2.0/doc/proxy-protocol.txt) it sends the data stream to the Envoy ingress gateway (EIGW).
+- EIGW then sends the TCP stream to another Envoy proxy which listens for PROXY protocol. This proxy then reads the destination IP + port from the PROXY protocol and forwards traffic to that IP + port. As a malicious party could try to exploit and communicate with another component in the Seed cluster, ACL is **mandatory**. Two options are available:
```
zanetworker: I have missed why we need two envoy proxies, isn't it possible to customize EIGW to fetch the IP,Port and directly forward to the API-server?

mvladev: The problem is the security aspect. This was the main reason.

timuthy: > EIGW then sends the TCP stream to another Envoy proxy which listens for PROXY protocol. This proxy then reads the destination IP + port from the PROXY protocol and forwards traffic to that IP + port.

Does that mean that the `Kubernetes` service inside the shoot cluster has to have the same IP as the `Kube-Apiserver` service of that shoot in the seed?

docs/proposals/08-shoot-apiserver-via-sni.md
```diff
@@ -206,136 +206,102 @@ Another issue occurs when the client tries to talk to the apiserver via the in-c
...
-- the apiserver master service reconciler is stopped (`--endpoint-reconciler-type=none`).
+- the apiserver master service reconciler is started and pointing to the `kube-apiserver`'s Cluster IP (e.g. `--advertise-address=10.1.2.3`).
 - the proxy runs in the host network of the `Node`.
 - the proxy has a sidecar container which:
-  - creates a dummy network interface and assigns the master service IP address to it.
-  - enables (with iptables) traffic to that IP address (kube-proxy rejects all traffic to `Services` without endpoints) and it removes connection tracking (conntrack) as the IP address is local to the `Node`.
-- the proxy needs to send multiple TLS streams over a single TCP connection (to reduce performance overhead).
-- the proxy listens on the master service IP address and it sends traffic to the correct API server.
+  - creates a dummy network interface and assigns the `10.1.2.3` to it.
+  - removes connection tracking (conntrack) as the IP address is local to the `Node`.
```
timuthy:
```suggestion
  - removes connection tracking (conntrack) if iptables/nftables is enabled as the IP address is local to the `Node`.
```

docs/proposals/08-shoot-apiserver-via-sni.md
mvladev: No - the endpoint which is backing this service will have the ip of the kube-apiserver `Service` in the Seed cluster.

timuthy: Okay got it. Thanks for clarifying.

timuthy (dismissed):

mvladev:
I've updated the PR to use only one proxy (Envoy Ingress Gateway) in the Seed cluster.

vlerenc:
Thanks for the new proposal. You removed the VPN comment at the bottom. Why/what's your views on VPN (later) @mvladev ?

mvladev:
VPN is not needed for workload => apiserver connectivity as it's using TLS by default and Nodes/Pods already have egress access to the apiserver.

VPN change would be done in a separate GEP

rfranzke (approved):
Let's merge this change to the GEP for now. Thanks for keeping it up-to-date.

docs/proposals/08-shoot-apiserver-via-sni.md
```diff
@@ -206,136 +206,168 @@ Another issue occurs when the client tries to talk to the apiserver via the in-c
 
 To mitigate this problem an additional proxy must be deployed on every single Node. It does not terminate TLS and sends the traffic to the correct Shoot API Server. This is achieved by:
```
alban, line 207: To clarify: "on every single Node of the shoot cluster"

docs/proposals/08-shoot-apiserver-via-sni.md
```diff
@@ -206,136 +206,168 @@ Another issue occurs when the client tries to talk to the apiserver via the in-c
...
-   |  the kubernetes     |                   |       Proxy         |
-   |  service            +------------------>+  No TLS termination |
-   |                     |                   |                     |
-   |                     |                   |                     |
-   +---------------------+                   +---------------------+
++------------------+                    +----------------+
+| Shoot API Server |       TCP          |   Envoy IGW    |
+|                  +<-------------------+ PROXY listener |
+| Cluster A        |                    |     :8443      |
++------------------+                    +-+--------------+
```
alban, line 224: IIUC the Envoy IGW PROXY listener is exposed by another Network LB from the cloud provider. So each seed cluster would have 2 Network LB for all the shoots. Is it correct?

mvladev: Each seed would only need `1` Network LB - the LoadBalancer would listen on `443` and `8443` and forward the traffic to `Envoy IGW`.

Additional network LB(s) might be need if we reach some limit on the cloud provider (max amount of TCP connections) or just for redundancy. In this case, the LBs would again point to the same `Envoy IGW`

PR #1867, opened 2020-01-28
Keep the replica count same for kcm if the shoot is not transitioning from/to hibernated state

ggaurav10:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #1866

**Special notes for your reviewer**:

**Release note**:

```improvement operator

```

PR #1873, opened 2020-01-30
Changed dwd probe kubeconfigs to client certs.

amshuman-kr:
Earlier the shoot-wise kubeconfigs for dependency-watchdog probes were using [static tokens](https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/secrets.go#L560).
But these secrets were not rotated along with [`kubecfg`](https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/secrets.go#L687).
So, after rotation, the dependency-watchdog probes fail for that shoot.

1. Now client-certs are used for the probe kubeconfigs. 
1. I have also added the code to delete old probe kubeconfig secrets contain the tokens.

**Release note**:

```improvement operator
Changed dependency-watchdog probe shoot kubeconfigs to use client certs.
```

### DIFF
--- a/pkg/operation/botanist/secrets.go
+++ b/pkg/operation/botanist/secrets.go
@@ -552,13 +552,16 @@ func (b *Botanist) generateWantedSecrets(basicAuthAPIServer *secrets.BasicAuth, 
 	// Secret definition for dependency-watchdog-internal-probe
 	secretList = append(secretList, &secrets.ControlPlaneSecretConfig{
 		CertificateSecretConfig: &secrets.CertificateSecretConfig{
-			Name:      common.DependencyWatchdogInternalProbeSecretName,
+			Name: common.DependencyWatchdogInternalProbeSecretName,
+
+			CommonName:   common.DependencyWatchdogUserName,
+			Organization: nil,
+			DNSNames:     nil,
+			IPAddresses:  nil,
+
+			CertType:  secrets.ClientCert,
 			SigningCA: certificateAuthorities[v1beta1constants.SecretNameCACluster],
 		},
-
-		BasicAuth: basicAuthAPIServer,
-		Token:     kubecfgToken,
-
 		KubeConfigRequest: &secrets.KubeConfigRequest{
 			ClusterName:  b.Shoot.SeedNamespace,
 			APIServerURL: fmt.Sprintf("%s.%s.svc", v1beta1constants.DeploymentNameKubeAPIServer, b.Shoot.SeedNamespace),
@@ -568,13 +571,16 @@ func (b *Botanist) generateWantedSecrets(basicAuthAPIServer *secrets.BasicAuth, 
 	// Secret definition for dependency-watchdog-external-probe
 	secretList = append(secretList, &secrets.ControlPlaneSecretConfig{
 		CertificateSecretConfig: &secrets.CertificateSecretConfig{
-			Name:      common.DependencyWatchdogExternalProbeSecretName,
+			Name: common.DependencyWatchdogExternalProbeSecretName,
+
+			CommonName:   common.DependencyWatchdogUserName,
+			Organization: nil,
+			DNSNames:     nil,
+			IPAddresses:  nil,
+
+			CertType:  secrets.ClientCert,
 			SigningCA: certificateAuthorities[v1beta1constants.SecretNameCACluster],
 		},
-
-		BasicAuth: basicAuthAPIServer,
-		Token:     kubecfgToken,
-
 		KubeConfigRequest: &secrets.KubeConfigRequest{
 			ClusterName:  b.Shoot.SeedNamespace,
 			APIServerURL: b.Shoot.ComputeAPIServerURL(false, true, b.APIServerAddress),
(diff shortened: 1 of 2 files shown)


rfranzke:
Wow, didn't notice this. We should always use client certificates for our components.

amshuman-kr:
Update:
> Testing is pending. I will finish that tomorrow and remove the WIP prefix

Struggling with local setup. So, testing is taking longer than expected.

ggaurav10:
will it be better if dependency watchdog considers all error cases except `timeout` for internal and external probes as success cases, since its primary objective, if I understand correct, is to check the connectivity of other components to the kube-apsierver? This way dwd need not depend on any other controller to maintain different set of secrets for each shoot.

Alternatively dwd can just expect the error `UnAuthorised` as a success case.

Since with this change, shoot kubelets and dwd use different authentications, whether kubelets will be successfully authorised or not should not be in purview of dwd.
wdyt?

amshuman-kr:
> This way dwd need not depend on any other controller to maintain different set of secrets for each shoot.

This is not entirely true. Some secrets/configmaps per shoot will still be required for the cluster-local and the internal DNS (via loadbalancer).

> Since with this change, shoot kubelets and dwd use different authentications, whether kubelets will be successfully authorised or not should not be in purview of dwd.

This is true. But I am uncomfortable having a health check that accepts more than 200 OK as a success.

ggaurav10:
> This is not entirely true. Some secrets/configmaps per shoot will still be required for the cluster-local and the internal DNS (via loadbalancer).

How about we enhance dwd. It can be provided a configmap with a list of endpoints that it uses to connect to apiserver, or any other service for that matter.  Only `timeout` is considered as failed case (or may be list of errors provided in the config map)
This way dwd becomes more generic with lesser attack surface.
Just thinking out loud. May be we can take it as next enhancement.

amshuman-kr:
> It can be provided a configmap with a list of endpoints that it uses to connect to apiserver, or any other service for that matter.

Such a list would be dependent on the list of active shoots hosted on that seed cluster. It is not a good idea to aggregate any information related to shoots into something in the garden or extension namespaces. A single source of truth is the actual shoot control-plane namespaces and the resources inside them. I think it is better it remains that way.

> Only timeout is considered as failed case (or may be list of errors provided in the config map)

In general the expectation from `/healthz` endpoints is that `200 OK` is returned. Anything else looks hacky.

ggaurav10:
> Such a list would be dependent on the list of active shoots hosted on that seed cluster. It is not a good idea to aggregate any information related to shoots into something in the garden or extension namespaces

I meant the configmaps to be in each shoot control plane, and not in common space for all shoots. This gets rid of exposing another set of credentials to the seed clusters with direct access to apiservers which are not rotated, if there is a security breach.

If i understood correct, dwd is not using `/healthz` endpoint. Please correct me if I am wrong. Refer:
https://github.com/gardener/dependency-watchdog/blob/0.4.0/pkg/scaler/prober.go#L239

amshuman-kr:
> If i understood correct, dwd is not using `/healthz` endpoint. Please correct me if I am wrong. Refer:
> https://github.com/gardener/dependency-watchdog/blob/0.4.0/pkg/scaler/prober.go#L239

No, it is not. But I am talking about the general approach to health checks. If socket connection timeouts are all we care about (I don't that is true here) then we should be doing doing just socket connections (like [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#tcpsocketaction-v1-core).

I am not sure that is the right approach here. Besides, I am not exactly sure what we would gain by making that change. Per shoot configmaps/secrets will still be required.

ggaurav10:
> I am not sure that is the right approach here. Besides, I am not exactly sure what we would gain by making that change. Per shoot configmaps/secrets will still be required.

1. less attack surface of DWD - No need of secrets
1. More generic architecture - Can be provided a list of errors with list of endpoints in a confimap along with the dependency list (which is already provided)

These benefits come to my mind top of the head.
This is just a suggestion. We can take it up later also **if** we feel the need to enhance dwd. Just wanted to float the idea. :)

amshuman-kr:
> This is just a suggestion. We can take it up later also if we feel the need to enhance dwd. Just wanted to float the idea. :)

👍 During the first implementation, I did consider supporting all the variants of probes as supported by kubelet in [`livenessProbe` and `readinessProbe`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#probe-v1-core) but opted for the simpler implementation to begin with 😉. Adding socket connection probe certainly makes sense.

amshuman-kr:
> Struggling with local setup. So, testing is taking longer than expected.

Update: I am facing more local dev setup issues for testing.  However, #1867 is already merged (thanks, @ggaurav10!) and mitigates the problem of waking up clusters after rotation of kubeconfig.

amshuman-kr:
Finally, the local setup worked enough to test this. I have tested that the probes using the new certs work and that old toke-based certs are replaced for existing shoots.

This PR is now ready for review/merge.

pkg/operation/common/types.go
```diff
@@ -250,6 +250,9 @@ const (
 	// DependencyWatchdogInternalProbeSecretName is the name of the kubecfg secret with cluster IP access.
 	DependencyWatchdogInternalProbeSecretName = "dependency-watchdog-internal-probe"
 
+	// DependencyWatchdogUserName is the user name of the dependency-watchdog.
+	DependencyWatchdogUserName = "system:serviceaccount:dependency-watchdog"
```
```suggestion
	DependencyWatchdogUserName = "gardener.cloud:system:dependency-watchdog"
```

pkg/operation/botanist/secrets.go
```diff
@@ -724,6 +731,29 @@ func (b *Botanist) DeploySecrets(ctx context.Context) error {
...
+	// on rotation. The old secrets containing tokens must be removed so that new secrets without token can be generated.
+	for _, secretName := range []string{common.DependencyWatchdogInternalProbeSecretName, common.DependencyWatchdogExternalProbeSecretName} {
+		secret := &corev1.Secret{}
+		if err := b.K8sSeedClient.Client().Get(ctx, kutil.Key(b.Shoot.SeedNamespace, secretName), secret); err != nil {
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			return err
+		}
+		if restconfig, err := kubernetes.NewRESTConfigFromSecret(secret); err != nil {
```
rfranzke: Hm, can we save some code here and just check if the `token` key exists in `secret.Data`? There is no need to create a REST config. If `token` key exists it is enough to know that the secret shall be deleted.

pkg/client/kubernetes/client.go
```diff
@@ -76,6 +76,28 @@ func NewRuntimeClientForConfig(fns ...ConfigFunc) (client.Client, error) {
 	return client.New(conf.restConfig, conf.clientOptions)
 }
 
+// NewRESTConfigFromSecret creates a new REST config for a given secret.
+func NewRESTConfigFromSecret(secret *corev1.Secret) (*rest.Config, error) {
```
rfranzke: See below - these functions are probably not needed and can be removed again.

rfranzke (changes requested):

pkg/operation/botanist/secrets.go
amshuman-kr: Done

pkg/operation/common/types.go
amshuman-kr: Done

pkg/client/kubernetes/client.go
amshuman-kr: Removed.

amshuman-kr:
@rfranzke Thanks for the review! I have addressed the comments. Can you PTAL?

rfranzke (approved):
/lgtm

PR #1880, opened 2020-02-04
Introduce hvpa for gardener-apiserver

ggaurav10:
Introduce hvpa for gardener-apiserver

Fixes #

**Release note**:

```improvement operator
Gardener-apiserver is now scaled by HVPA
```

### DIFF
--- a/charts/gardener/controlplane/charts/runtime/templates/apiserver/hvpa.yaml
+++ b/charts/gardener/controlplane/charts/runtime/templates/apiserver/hvpa.yaml
@@ -1,0 +1,73 @@
+{{- if .Values.global.apiserver.hvpa.enabled }}
+{{- /* .Values.global.apiserver.replicaCount is of type 'float64', so let's cast it to string to have proper types for comparison */}}
+{{- if ne (.Values.global.apiserver.replicaCount | toString) "0" }}
+apiVersion: autoscaling.k8s.io/v1alpha1
+kind: Hvpa
+metadata:
+  name: gardner-apiserver-hvpa
+  namespace: garden
+spec:
+  replicas: 1
+{{- if .Values.global.apiserver.hvpa.scaleUpStabilization }}
+  scaleUpStabilization:
+{{ toYaml .Values.global.apiserver.hvpa.scaleUpStabilization | indent 4 }}
+{{- end }}
+{{- if .Values.global.apiserver.hvpa.scaleDownStabilization }}
+  scaleDownStabilization:
+{{ toYaml .Values.global.apiserver.hvpa.scaleDownStabilization | indent 4 }}
+{{- end }}
+  hpa:
+    selector:
+      matchLabels:
+        role: gardener-apiserver-hpa
+    template:
+      metadata:
+        labels:
+          role: gardener-apiserver-hpa
+      spec:
+        maxReplicas: {{ .Values.global.apiserver.hvpa.maxReplicas }}
+        minReplicas: {{ .Values.global.apiserver.hvpa.minReplicas }}
+        metrics:
+        - resource:
+            name: memory
+            targetAverageUtilization: {{ .Values.global.apiserver.hvpa.targetAverageUtilization }}
+          type: Resource
+        - resource:
+            name: cpu
+            targetAverageUtilization: {{ .Values.global.apiserver.hvpa.targetAverageUtilization }}
+          type: Resource
+    updatePolicy:
+      updateMode: "Auto"
+  vpa:
+    selector:
+      matchLabels:
+        role: gardener-apiserver-vpa
+    template:
+      metadata:
+        labels:
+          role: gardener-apiserver-vpa
+      spec:
+        resourcePolicy:
+          containerPolicies:
+            - containerName: gardener-apiserver
+              maxAllowed:
+                memory: 25G
+                cpu: "8"
+              minAllowed:
+                memory: 400M
+                cpu: 400m
+    updatePolicy:
+      updateMode: "Auto"
+  weightBasedScalingIntervals:
+    - vpaWeight: 0
+      startReplicaCount: {{ .Values.global.apiserver.hvpa.minReplicas }}
+      lastReplicaCount: {{ .Values.global.apiserver.hvpa.lastReplicaCountForHpa }}
+    - vpaWeight: 100
+      startReplicaCount: {{ .Values.global.apiserver.hvpa.maxReplicas }}
+      lastReplicaCount: {{ .Values.global.apiserver.hvpa.maxReplicas }}
+  targetRef:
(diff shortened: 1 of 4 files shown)


rfranzke (dismissed):
/lgtm

Can you elaborate on the pre-stop lifecycle hook?

ggaurav10:
The delay in preStop helps give the kube-proxies a chance to propagate the changes to the iptables in all the nodes as noted [here](https://github.com/kubernetes/kubernetes/issues/43576).

rfranzke:
Great, thanks Gaurav!

charts/gardener/controlplane/charts/runtime/templates/apiserver/hvpa.yaml
```diff
@@ -0,0 +1,73 @@
+{{- if .Values.global.apiserver.hvpa.enabled }}
```
timebertt:
```suggestion
{{- if and .Values.global.apiserver.enabled .Values.global.apiserver.hvpa.enabled }}
```
I think, it only makes sense to deploy the HVPA if the apiserver itself is enabled.

charts/gardener/controlplane/charts/runtime/templates/apiserver/hvpa.yaml
```diff
@@ -0,0 +1,73 @@
...
+{{- /* .Values.global.apiserver.replicaCount is of type 'float64', so let's cast it to string to have proper types for comparison */}}
+{{- if ne (.Values.global.apiserver.replicaCount | toString) "0" }}
+apiVersion: autoscaling.k8s.io/v1alpha1
+kind: Hvpa
+metadata:
+  name: gardner-apiserver-hvpa
+  namespace: garden
```
timebertt, line 8:
```suggestion
  namespace: garden
  labels:
    app: gardener
    role: apiserver
    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
    release: "{{ .Release.Name }}"
    heritage: "{{ .Release.Service }}"
```
Can you please add the default labels to the hvpa?

charts/gardener/controlplane/charts/runtime/templates/apiserver/hvpa.yaml
```diff
@@ -0,0 +1,73 @@
...
+      metadata:
+        labels:
+          role: gardener-apiserver-hpa
+      spec:
+        maxReplicas: {{ .Values.global.apiserver.hvpa.maxReplicas }}
+        minReplicas: {{ .Values.global.apiserver.hvpa.minReplicas }}
+        metrics:
+        - resource:
+            name: memory
+            targetAverageUtilization: {{ .Values.global.apiserver.hvpa.targetAverageUtilization }}
```
timebertt: I am not sure, but would it make sense to be able to configure the `targetAverageUtilization` separately for cpu and memory?

charts/gardener/controlplane/values.yaml
```diff
@@ -14,8 +14,8 @@ global:
         cpu: 100m
         memory: 100Mi
       limits:
-        cpu: 300m
-        memory: 256Mi
+        cpu: "8"
+        memory: 25G
```
timebertt, line 18: Why are the limits so high? Is this required for hvpa?
Is my assumption right, that the apiserver Pods will then have 8/25G limits initially right away?
In that case, the range `[requests,limits]` seems a bit too large IMO.

charts/gardener/controlplane/charts/runtime/templates/apiserver/hvpa.yaml
```diff
@@ -0,0 +1,73 @@
...
+                memory: 400M
+                cpu: 400m
+    updatePolicy:
+      updateMode: "Auto"
+  weightBasedScalingIntervals:
+    - vpaWeight: 0
+      startReplicaCount: {{ .Values.global.apiserver.hvpa.minReplicas }}
+      lastReplicaCount: {{ .Values.global.apiserver.hvpa.lastReplicaCountForHpa }}
+    - vpaWeight: 100
+      startReplicaCount: {{ .Values.global.apiserver.hvpa.maxReplicas }}
```
timebertt, line 74: Shouldn't this be something like: `{{ .Values.global.apiserver.hvpa.lastReplicaCountForHpa + 1 }}`?

timebertt (changes requested):

charts/gardener/controlplane/values.yaml
ggaurav10: The last release of HVPA controller doesn't scale `limits`, that's why it is kept very high.  In the next release of HVPA, there will be some `spec` changes along with scaling of `limits`. These values will be adopted accordingly when Gardener starts using the next release.

charts/gardener/controlplane/charts/runtime/templates/apiserver/hvpa.yaml
ggaurav10: Sure. Will add this to `vpa` chart too

charts/gardener/controlplane/charts/runtime/templates/apiserver/hvpa.yaml
ggaurav10: currently `lastReplicaCountForHpa` = `maxReplicas - 1`
I am not sure if there is a math operator so that this can be done in the chart itself.

rfranzke: See https://github.com/gardener/gardener/blob/master/charts/gardener/controlplane/charts/runtime/templates/apiserver/poddisruptionbudget.yaml#L14

ggaurav10: Thanks Rafael, this helps!

ggaurav10: Though this might be an issue if `maxReplicas` is 1 or equal to `minReplicas`. What do you suggest?

timebertt: I guess, in that case we would want to scale only vertically, right?
So we could just wrap the first interval in an if-block...

ggaurav10: right. Done changes. Thanks Tim.

charts/gardener/controlplane/values.yaml
timebertt: Got it. Thanks for explaining!

timebertt (approved):
/approve

rfranzke (approved):
Thanks @tim-ebert for your thorough review and thanks @ggaurav10 for addressing so quickly. /lgtm

PR #1884, opened 2020-02-04
fixing scheduler minimal distance

danielfoehrKn:

The Minimal distance Scheduler Strategy contains problems
- Seed Selector needs to be executed before the Strategy to filter the list of possible seed candidates
- Minimal Distance Strategy had a bug leading to wrong scheduling decisions, when the shoot specified region is shorter than the seed regions

Fixes #

**Release note**:

```improvement operator
Scheduler: Fixing MinimalDistance Strategy to correctly apply the seed selector from the CloudProfile and fixing a bug in the distance logic
```

### DIFF
--- a/pkg/scheduler/controller/shoot/scheduler_control.go
+++ b/pkg/scheduler/controller/shoot/scheduler_control.go
@@ -175,27 +175,25 @@ func determineSeed(shoot *gardencorev1beta1.Shoot, seedLister gardencorelisters.
 		return nil, err
 	}
 
-	return determineBestSeedCandidate(shoot, cloudProfile, shootList, seedList, strategy)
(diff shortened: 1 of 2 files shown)


ialidzhikov:
Could you please add a description about why we need this PR?

danielfoehrKn:
> Could you please add a description about why we need this PR?

Was still in the process of writing it - opened it too early :)

rfranzke:
Thanks for the PR! It looks good, I like the dedicated functions - it's way better readable now. I'll give it a try tomorrow and come back. 💯

rfranzke (approved):
/lgtm, works perfectly!

timebertt (approved):
/lgtm

PR #1890, opened 2020-02-05
RFC: simplify k8s go module definition

majst01:
As discussed today i want you to show how the go.mod dependencies to all kubernetes artefacts can be made more readable.

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -66,15 +66,15 @@ require (
 
 replace (
 	github.com/prometheus/client_golang => github.com/prometheus/client_golang v0.9.2
-	k8s.io/api => k8s.io/api v0.16.6
-	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.16.6
-	k8s.io/apimachinery => k8s.io/apimachinery v0.16.6
-	k8s.io/apiserver => k8s.io/apiserver v0.16.6
-	k8s.io/client-go => k8s.io/client-go v0.16.6
-	k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.16.6
-	k8s.io/code-generator => k8s.io/code-generator v0.16.6
-	k8s.io/component-base => k8s.io/component-base v0.16.6
+	k8s.io/api => k8s.io/api v0.16.0
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.16.0
+	k8s.io/apimachinery => k8s.io/apimachinery v0.16.0
+	k8s.io/apiserver => k8s.io/apiserver v0.16.0
+	k8s.io/client-go => k8s.io/client-go v0.16.0
+	k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.16.0
+	k8s.io/code-generator => k8s.io/code-generator v0.16.0
+	k8s.io/component-base => k8s.io/component-base v0.16.0
 	k8s.io/helm => k8s.io/helm v2.13.1+incompatible
-	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.16.6
-	k8s.io/kube-openapi => k8s.io/kube-openapi v0.16.6
+	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.16.0
+	k8s.io/kube-openapi => k8s.io/kube-openapi v0.16.0
 )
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by vpnachev** [comment]
Looks very nice!
@ialidzhikov wdyt?

**Comment by ialidzhikov** [comment]
Thanks for the suggestion.
The reason we use 1.16.0 is the controller-runtime. There is kubernetes-sigs/controller-runtime#732 and several similar PRs. I would prefer to stick to the k/k versions used by the current version of controller-runtime.

**Comment by majst01** [comment]
@ialidzhikov, ok i set all to 1.16.0 as before instead 1.16.6

File: go.mod [hunk 1]
```diff
@@ -46,35 +46,35 @@ require (
...
 	github.com/prometheus/client_golang => github.com/prometheus/client_golang v0.9.2
-	k8s.io/api => k8s.io/api v0.0.0-20190918155943-95b840bb6a1f // kubernetes-1.16.0
-	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.0.0-20190918161926-8f644eb6e783 // kubernetes-1.16.0
-	k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20190913080033-27d36303b655 // kubernetes-1.16.0
-	k8s.io/apiserver => k8s.io/apiserver v0.0.0-20190918160949-bfa5e2e684ad // kubernetes-1.16.0
-	k8s.io/client-go => k8s.io/client-go v0.0.0-20190918160344-1fbdaa4c8d90 // kubernetes-1.16.0
-	k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.0.0-20190918163108-da9fdfce26bb // kubernetes-1.16.0
-	k8s.io/code-generator => k8s.io/code-generator v0.0.0-20190912054826-cd179ad6a269 // kubernetes-1.16.0
-	k8s.io/component-base => k8s.io/component-base v0.0.0-20190918160511-547f6c5d7090 // kubernetes-1.16.0
+	k8s.io/api => k8s.io/api v0.16.0
```
ialidzhikov, line 69: k/k started to publish go modules friendly tags from `v0.16.4`. There is no such module:
```
$ go get k8s.io/api@v0.16.0
go: finding k8s.io/api v0.16.0
go: finding k8s.io/api v0.16.0
go get k8s.io/api@v0.16.0: k8s.io/api@v0.16.0: invalid version: unknown revision v0.16.0
```

majst01: compiles here, but `go mod tidy` does not work with 1.16.0. So if you need to stick with <=1.16.4 there is no way to go this way and we have to wait.

go.mod
```diff
@@ -46,35 +46,35 @@ require (
...
+	k8s.io/cluster-bootstrap v0.0.0
+	k8s.io/code-generator v0.0.0
+	k8s.io/component-base v0.0.0
 	k8s.io/helm v2.14.2+incompatible
-	k8s.io/klog v0.4.0
-	k8s.io/kube-aggregator v0.0.0-20191004104030-d9d5f0cc7532
-	k8s.io/kube-openapi v0.0.0-20191107075043-30be4d16710a
-	k8s.io/metrics v0.0.0-20191004105854-2e8cf7d0888c
-	k8s.io/utils v0.0.0-20190801114015-581e00157fb1
+	k8s.io/klog v1.0.0
```
timuthy, line 58: I think the `klog` version can also not be bumped if we rely on `v0.16.0` versions.

https://github.com/kubernetes/apimachinery/blob/27d36303b6556f377b4f34e64705fa9024a12b0c/go.mod#L34

go.mod
timuthy: At least we can remove the suffixes from the `require` block and stick to `v0.0.0`, can't we?

go.mod
majst01: We use klog v1.0.0 in all of our controllers with no probs, and i think there are no api breaking changes between v0.4.0 and v1.0.0:
https://github.com/kubernetes/klog/compare/v0.4.0...v1.0.0
the v1.0.0 tag was more of a cosmetic action of the k8s authors.

But nevertheless i raised this PR to show some of our findings to simplify the go.mod file and make it more readable/maintainable, because of the hard dependency to a older 1.16 version @ialidzhikov found, this can not be done now. 

Therefore i close this.

timuthy: Thanks anyway. Let's keep it in mind as soon as the controller-runtime uses a later version :)

majst01: Yes, i will remind us with a followup PR if you don´t mind.

PR #1898, opened 2020-02-07
Fix potential bug in secret releasing for binding deletion

rfranzke:
Previously, the `SecretBinding` controller was releasing the finalizer from the referenced `Secret` if the `SecretBinding` was not used by `Shoot`s in this namespace.
However, generally, there might be another `SecretBinding` in a different namespace that references the same `Secret`. Consequently, the finalizer may be only be removed from the `Secret` if there is no other `SecretBinding` references it.

**Release note**:

```improvement operator
A bug that may accidentally release the finalizer from `Secret`s when a referencing `SecretBinding` resource is deleted has been fixed.
```

### DIFF
--- a/pkg/controllermanager/controller/secretbinding/secretbinding_control.go
+++ b/pkg/controllermanager/controller/secretbinding/secretbinding_control.go
@@ -30,11 +30,11 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
 	kubecorev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func (c *Controller) secretBindingAdd(obj interface{}) {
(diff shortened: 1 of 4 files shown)


pkg/controllermanager/controller/secretbinding/secretbinding_control.go
```diff
@@ -135,19 +137,27 @@ func (c *defaultControl) ReconcileSecretBinding(obj *gardencorev1beta1.SecretBin
...
+			if err != nil {
+				secretBindingLogger.Error(err.Error())
+				return err
+			}
+
+			if mayReleaseSecret {
+				// Remove finalizer from referenced secret
+				secret, err := c.secretLister.Secrets(secretBinding.SecretRef.Namespace).Get(secretBinding.SecretRef.Name)
+				if err == nil {
+					secretFinalizers := sets.NewString(secret.Finalizers...)
```
ialidzhikov: Shouldn't we use .DeepCopy for things we get from listers?
Btw there is a RemoveFinalizer/RemoveGardenerFinalizer func. Does it makes sense to use it?

rfranzke: I didn't change this logic as part of this PR, but I can quickly refactor it.

ialidzhikov: Ah, yep, I guess we can do it with another PR as we would need to touch some other parts.

rfranzke: np, I'm doing it quickly, one sec.

ialidzhikov (dismissed):
/lgtm

pkg/controllermanager/controller/secretbinding/secretbinding_control.go
```diff
@@ -188,3 +191,19 @@ func (c *defaultControl) ReconcileSecretBinding(obj *gardencorev1beta1.SecretBin
 
 	return nil
 }
+
+// We may only release a secret if there is no other secretbinding that references it (maybe in a different namespace).
+func (c *defaultControl) mayReleaseSecret(namespace, name string) (bool, error) {
```
timebertt, line 196: Can you add a unit test for this?

timebertt (commented):
Generally, /lgtm
But a small unit test for this would be nice :)

pkg/controllermanager/controller/secretbinding/secretbinding_control.go
```diff
@@ -135,29 +136,31 @@ func (c *defaultControl) ReconcileSecretBinding(obj *gardencorev1beta1.SecretBin
...
+			if err != nil {
 				secretBindingLogger.Error(err.Error())
 				return err
 			}
 
+			if mayReleaseSecret {
+				// Remove finalizer from referenced secret
+				secret, err := c.secretLister.Secrets(secretBinding.SecretRef.Namespace).Get(secretBinding.SecretRef.Name)
+				if err == nil {
+					if err2 := controllerutils.RemoveFinalizer(ctx, c.k8sGardenClient.Client(), secret, gardencorev1beta1.ExternalGardenerName); err2 != nil {
```
ialidzhikov:
```suggestion
					if err2 := controllerutils.RemoveFinalizer(ctx, c.k8sGardenClient.Client(), secret.DeepCopy(), gardencorev1beta1.ExternalGardenerName); err2 != nil {
```

pkg/controllermanager/controller/secretbinding/secretbinding_control.go
rfranzke: done

ialidzhikov (approved):
/lgtm

PR #1899, opened 2020-02-07
Deploy gardenlet with highest possible priority

### LINKED ISSUE #1726: High priority for gardenlet deployment on seed
The gardenlet is a central and core component running on seeds. We should have some `PodPriority` to ensure that it is always ran, and potentially preempts other pods with a lower priority. I've seen cases in which the VPA scaled up the gardenlet but it could not be scheduled because no node had enough capacity. We had to wait until the cluster-autoscaler provisioned a new node which caused an undesired downtime.

rfranzke:
We should ensure that the gardenlet always runs and is never preempted in favor of other pods. This PR creates a `PriorityClass` for the gardenlet and deploys it with the highest possible prio (1 billion).

Fixes #1726

**Release note**:

```improvement operator
The gardenlet Helm chart now come with a `PriorityClass` for the gardenlet `Deployment` to ensure it always runs and is never preempted in favor of other pods.
```

### DIFF
--- a/charts/gardener/gardenlet/charts/runtime/templates/deployment.yaml
+++ b/charts/gardener/gardenlet/charts/runtime/templates/deployment.yaml
@@ -36,6 +36,7 @@ spec:
         release: "{{ .Release.Name }}"
         heritage: "{{ .Release.Service }}"
     spec:
+      priorityClassName: gardenlet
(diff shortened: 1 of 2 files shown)


timebertt (approved):
/approve

PR #1901, opened 2020-02-07
Validate minimum value for kubelet pod pids limit

### LINKED ISSUE #1536: Validate podPIDLimit
As user can configure the `podPIDLimit` of the kubelet we need to validate that number to ensure that all gardener components have enough PID's to work correctly.

Background: A user specified a limit of 10 which causes the kube-proxy to be unable to setup the iptables as it cannot spawn th iptables process. 

cc @zanetworker

rfranzke:
Validate minimum value for the pod pids limit field in the kubelet config to ensure at least the system processes can run smoothly. /cc @schrodit 

Fixes #1536

**Release note**:

```noteworthy user
The minimum value for the `podPidsLimit` field in the kubelet configuration for `Shoot` resources is now `100`. If you have specified a value less then `100` the gardener-apiserver will automatically migrate it for you, however, you should switch to the new minimum value as this auto-migration logic will be removed again with one of the next releases.
```

### DIFF
--- a/pkg/registry/core/shoot/strategy_test.go
+++ b/pkg/registry/core/shoot/strategy_test.go
@@ -15,7 +15,10 @@
 package shoot_test
 
 import (
+	"context"
+
 	"github.com/gardener/gardener/pkg/apis/core"
+	"github.com/gardener/gardener/pkg/apis/core/validation"
 	shootregistry "github.com/gardener/gardener/pkg/registry/core/shoot"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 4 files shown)


pkg/registry/core/shoot/strategy_test.go
```diff
@@ -25,6 +28,73 @@ import (
...
+							Kubelet: &core.KubeletConfig{
+								PodPIDsLimit: &tooSmallValue,
+							},
+						},
+					},
+				}
+
+				oldShoot := newShoot("foo")
+				oldShoot.Spec.Kubernetes.Kubelet = shoot.Spec.Kubernetes.Kubelet.DeepCopy()
+				oldShoot.Spec.Provider.Workers = []core.Worker{shoot.Spec.Provider.Workers[0]}
```
timebertt:
```suggestion
				oldShoot.Spec.Provider.Workers = []core.Worker{*shoot.Spec.Provider.Workers[0].DeepCopy()}
```
Can we use `DeepCopy()` to make sure the tested code is not operating on the exact same object?

pkg/registry/core/shoot/strategy_test.go
```diff
@@ -25,6 +28,73 @@ import (
...
+							Kubelet: &core.KubeletConfig{
+								PodPIDsLimit: &highEnoughValue,
+							},
+						},
+					},
+				}
+
+				oldShoot := newShoot("foo")
+				oldShoot.Spec.Kubernetes.Kubelet = shoot.Spec.Kubernetes.Kubelet.DeepCopy()
+				oldShoot.Spec.Provider.Workers = []core.Worker{shoot.Spec.Provider.Workers[0]}
```
timebertt:
```suggestion
				oldShoot.Spec.Provider.Workers = []core.Worker{*shoot.Spec.Provider.Workers[0].DeepCopy()}
```
Same here.

pkg/registry/core/shoot/strategy_test.go
```diff
@@ -25,6 +28,73 @@ import (
...
+
+				shootregistry.Strategy.PrepareForUpdate(context.TODO(), shoot, oldShoot)
+
+				Expect(*shoot.Spec.Kubernetes.Kubelet.PodPIDsLimit).To(Equal(validation.PodPIDsLimitMinimum))
+				Expect(*oldShoot.Spec.Kubernetes.Kubelet.PodPIDsLimit).To(Equal(validation.PodPIDsLimitMinimum))
+				Expect(*shoot.Spec.Provider.Workers[0].Kubernetes.Kubelet.PodPIDsLimit).To(Equal(*shoot.Spec.Provider.Workers[0].Kubernetes.Kubelet.PodPIDsLimit))
```
timebertt:
```suggestion
				Expect(*shoot.Spec.Provider.Workers[0].Kubernetes.Kubelet.PodPIDsLimit).To(Equal(*highEnoughValue))
```
I guess this check is the same as `0 == 0`... 😅

pkg/registry/core/shoot/strategy_test.go
```diff
@@ -25,6 +28,73 @@ import (
...
+				Expect(*oldShoot.Spec.Provider.Workers[0].Kubernetes.Kubelet.PodPIDsLimit).To(Equal(*oldShoot.Spec.Provider.Workers[0].Kubernetes.Kubelet.PodPIDsLimit))
```
timebertt:
```suggestion
				Expect(*oldShoot.Spec.Provider.Workers[0].Kubernetes.Kubelet.PodPIDsLimit).To(Equal(*highEnoughValue))
```
Same here.

timebertt (changes requested):
Thanks, just some minor nits.

pkg/registry/core/shoot/strategy_test.go
rfranzke: `Workers[0]` is of type `core.Worker`, hence, it doesn't have a `DeepCopy` function but is already a copy.

pkg/registry/core/shoot/strategy_test.go
rfranzke: Same here

pkg/registry/core/shoot/strategy_test.go
timebertt: Sorry, I don't quite get it.
`core.Worker` seems to have a `DeepCopy` function: https://github.com/gardener/gardener/blob/13d359fa934466fd8ed2452e8042ebd646d9260c/pkg/apis/core/zz_generated.deepcopy.go#L3257-L3265

Additionally when I debug the test, it shows me the same pointers for `shoot.Spec.Provider.Workers[0].Kubernetes` and `oldShoot.Spec.Provider.Workers[0].Kubernetes`, so I don't think the nested objects are copied.  
For the `core.Worker` objects, you are right, because they are just `struct` values and no pointers. But the copy of the `struct` value still contains the same pointer values for its fields like for example `.Kubernetes`.

rfranzke: You're right, it just copies the first level but not transitively. OK, fixed.

timebertt (approved):
Alright, thanks for fixing :) 
/approve

PR #1885, opened 2020-02-05
Remove deprecated label and annotation

ialidzhikov:
**What this PR does / why we need it**:
Remove deprecated label `"shoot.garden.sapcloud.io/hibernated"` and annotation `"shoot.garden.sapcloud.io/uid"`. A machinery needs to switch to the Cluster resource.
Remove the logic that duplicates the `"custom.shoot.sapcloud.io/"` prefixed annotations to the namespace in the Seed.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement operator
Deprecated annotation `shoot.garden.sapcloud.io/uid` and label `shoot.garden.sapcloud.io/hibernated` are no longer added to the Shoot namespace.
```
```improvement operator
Shoot annotations prefixed with `custom.shoot.sapcloud.io/` are no longer maintained on the Shoot namespace.
```

Issue #1649, opened 2019-11-22
Adapt labels and annotations to the new API group


ialidzhikov:
Currently the annotation to trigger reconciliation on crds such as `Infrastructure`, `Worker` is `gardener.cloud/operation=reconcile`. However the annotation for Shoot is `shoot.garden.sapcloud.io/operation=reconcile`. There are annotation/label keys such as `garden.sapcloud.io/createdBy`, `namespace.garden.sapcloud.io/project`, `shoot.garden.sapcloud.io/ignore`, etc. that could be switched to `<foo>.gardener.cloud/<bar>` format.

rfranzke:
According to @ialidzhikov most of this is done. We are now waiting a couple of releases before removing the final remaining artefacts, and then we can close this issue.
@ialidzhikov please correct me if I'm wrong and thanks for all your efforts!

rfranzke:
@ialidzhikov Could you provide an update of the status of this issue? :)

ialidzhikov:
> @ialidzhikov Could you provide an update of the status of this issue? :)

End-user related labels/annotations (API related ones) are adapted and cleaned up since longtime.

Wrt to the internal ones - I think the leftover items to tackle are:
- [x] controlplane NetworkPolicies still mainly use the deprecated `garden.sapcloud.io/role` label key in the selectors -> needs to be adapted to use `gardener.cloud/role` where it is possible and safe
  - https://github.com/gardener/gardener/pull/4572
  - https://github.com/gardener/gardener/pull/4607
- [x] controlplane Pods still maintain both `garden.sapcloud.io/role` and `gardener.cloud/role` -> end result we want is to have only  `gardener.cloud/role` (I am also leaving the etcd Pods out of scope as we don't want to trigger etcd restart just because of such change in the Pod labels).
  - https://github.com/gardener/gardener/pull/4768
  - https://github.com/gardener/gardener/pull/4783

PR #1906, opened 2020-02-10
Readd the shoot.garden.sapcloud.io/uid annotation to the Shoot namespace

ialidzhikov:
This annotation was removed with #1885. The backupentry controller still watches on namespaces and uses this annotation. See https://github.com/gardener/gardener-extensions/blob/a83364393e2f8c15c1f846e422aedd24856bbe80/pkg/controller/backupentry/mapper.go#L92

According to @swapnilgm , there might be a data race currently.

Fixes #

**Release note**:

```improvement operator
Readd the `shoot.garden.sapcloud.io/uid` annotation to the Shoot namespace as there is still machinery that relies on it.
```

### DIFF
--- a/pkg/apis/core/v1alpha1/constants/types_constants.go
+++ b/pkg/apis/core/v1alpha1/constants/types_constants.go
@@ -133,6 +133,10 @@ const (
 	// GardenRoleOptionalAddon is the value of the GardenRole key indicating type 'optional-addon'.
 	GardenRoleOptionalAddon = "optional-addon"
 
+	// DeprecatedShootUID is an annotation key for the shoot namespace in the seed cluster,
+	// which value will be the value of `shoot.status.uid`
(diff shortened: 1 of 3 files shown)


vpnachev (approved):
/approve

swapnilgm (approved):
Thanks for catching bug and providing quick fix. LGTM.
May be we can replace `shoot.garden.sapcloud.io` to `shoot.gardener.cloud` with nre release on both gardener and garndener-extensions.

ialidzhikov:
> May be we can replace `shoot.garden.sapcloud.io` to `shoot.gardener.cloud` with nre release on both gardener and garndener-extensions.

I think I will do it as part of https://github.com/gardener/gardener/issues/1649, or we can check whether the backupentry_controller can use the Cluster resource.

Issue #1631, opened 2019-11-19
Shoot Control Plane Migration


plkokanov:
This issue describes the necessary tasks for an initial implementation of the proposal for [Control Plane Migration](https://github.com/gardener/gardener/blob/master/docs/proposals/07-shoot-control-plane-migration.md).

The goal is to enable control plane migration between two working seeds.

The following things have to be done:
  - [x] ShootState resource added to gardener API Server #1634
  - [x] ShootState synchronization controller added to gardenlet #1731
  - [x] Extension Controllers extended to read and write state to their Custom Resource's 
`status.state` field gardener/gardener-extensions#476, gardener/gardener-extensions#528
  - [x] Extension Controllers support `gardener.cloud/operation=restore` to restore state from `status.state` and `gardener.cloud/operaton=migrate` to clean up resources
    - [x] Add Restore function to actuator interface in gardener-extensions 
    - [x] Implement Restore/Migrate function in extension controllers
      - [x] Infrastructure
      - [x] Worker
      - [x] Backupentry
      - [x] Network
      - [x] Controlplane
      - [x] OperatingSystemConfig
      - [x] Certificate extension controller
      - [x] DNS Controller 
      - [x] ContainerRuntime [#10](https://github.com/gardener/gardener-extension-runtime-gvisor/issues/10)
  - [x] Persist ETCD encryption data in ShootState 
  - [x] Extend ShootState and ShootState Sync Controller to be able to store secret data required by extension resources #2327 #2321 
  - [x] Add DNSOwners for DNSEntries during reconciliation and migrate/restore them during control plane migration
  - [x] Persist data for secrets generated by gardener in the ShootState
  - [x] Shoot reconciliation and deletion flows extended to support migration
  - [x] Enable modification of `Shoot.spec.seedName` field to trigger migration
  - [x] Integration tests
  - [x] Document what extensions controller writers need to do in order to support control plane migration
  - [x] Add a check if the __Destination Seed__'s provider is the same as the __Source Seed__
  - [x] Remove GardenerID and GardenerName checks (via Seed.Status.Gardener) when checking if the __Destination Seed__ is healthy as those values are defined for each gardenlet and will most likely be different. The version check should be kept though.
  - [x] Adapt worker configuration for the `Shooted Seed` which is created for CP Migration integration tests. The current configuration has too little workers and not all pods on the seed can be started.
  - [x] Improve updates of states of extension resources to the ShootState.
[...]

PR #1908, opened 2020-02-10
Adds shootstate admission plugin

plkokanov:
Adds an admission plugin for ShootState resources which will forbid deletion of the ShootState object if its corresponding Shoot object still exists

Part of #1631 

**Release note**:

```improvement operator
Adds admission plugin which forbids ShootState deletion if the corresponding Shoot still exist.
```

### DIFF
--- a/cmd/gardener-apiserver/app/gardener_apiserver.go
+++ b/cmd/gardener-apiserver/app/gardener_apiserver.go
@@ -27,6 +27,8 @@ import (
 	admissioninitializer "github.com/gardener/gardener/pkg/apiserver/admission/initializer"
 	"github.com/gardener/gardener/pkg/apiserver/storage"
 	gardencoreclientset "github.com/gardener/gardener/pkg/client/core/clientset/internalversion"
+	gardenversionedcoreclientset "github.com/gardener/gardener/pkg/client/core/clientset/versioned"
+	gardenexternalcoreinformers "github.com/gardener/gardener/pkg/client/core/informers/externalversions"
 	gardencoreinformers "github.com/gardener/gardener/pkg/client/core/informers/internalversion"
 	settingsclientset "github.com/gardener/gardener/pkg/client/settings/clientset/versioned"
 	settingsinformer "github.com/gardener/gardener/pkg/client/settings/informers/externalversions"
@@ -41,6 +43,7 @@ import (
 	openidconnectpreset "github.com/gardener/gardener/plugin/pkg/shoot/oidc/openidconnectpreset"
 	shootquotavalidator "github.com/gardener/gardener/plugin/pkg/shoot/quotavalidator"
 	shootvalidator "github.com/gardener/gardener/plugin/pkg/shoot/validator"
+	shootstatedeletionvalidator "github.com/gardener/gardener/plugin/pkg/shootstate/validator"
 
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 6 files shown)


danielfoehrKn (dismissed):
Looks good

cmd/gardener-apiserver/app/gardener_apiserver.go
```diff
@@ -41,6 +43,7 @@ import (
 	openidconnectpreset "github.com/gardener/gardener/plugin/pkg/shoot/oidc/openidconnectpreset"
 	shootquotavalidator "github.com/gardener/gardener/plugin/pkg/shoot/quotavalidator"
 	shootvalidator "github.com/gardener/gardener/plugin/pkg/shoot/validator"
+	shootstatedeletionvalidatior "github.com/gardener/gardener/plugin/pkg/shootstate"
```
rfranzke:
```suggestion
	shootstatedeletionvalidator "github.com/gardener/gardener/plugin/pkg/shootstate"
```

plugin/pkg/shootstate/admission.go
```diff
@@ -0,0 +1,177 @@
...
+	admissioninitializer "github.com/gardener/gardener/pkg/apiserver/admission/initializer"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apiserver/pkg/admission"
+)
+
+const (
+	// PluginName is the name of this admission plugin.
+	PluginName = "ShootStateDeletion"
```
rfranzke:
```suggestion
	PluginName = "ShootStateDeletionValidator"
```

according to your import in `cmd/gardener-apiserver`

plugin/pkg/shootstate/admission.go
```diff
@@ -0,0 +1,177 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: Can you move all the files into a subdirectory `./validator`, please?

plugin/pkg/shootstate/admission.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	return nil
+}
+
+func (d *ValidateShootStateDeletion) validateDelete(ctx context.Context, attrs admission.Attributes) error {
+	_, err := d.coreClient.Core().Shoots(attrs.GetNamespace()).Get(attrs.GetName(), metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
```
rfranzke: You should also handle other `err`s.

plugin/pkg/shootstate/admission_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	It("should do nothing because the resource is not ShootState", func() {
+		attrs := admission.NewAttributesRecord(nil, nil, core.Kind("Foo").WithVersion("version"), shoot.Namespace, shoot.Name, core.Resource("foos").WithVersion("version"), "", admission.Delete, &metav1.DeleteOptions{}, false, nil)
+
+		err := admissionHandler.Validate(context.TODO(), attrs, nil)
+
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("should forbid ShootState deletion because shoot object exist ", func() {
```
rfranzke:
```suggestion
	It("should forbid ShootState deletion because shoot object exists", func() {
```

rfranzke (changes requested):

plugin/pkg/shootstate/admission.go
plkokanov: I'll add a check for IsGone() all other errors should return forbidden IMHO.

rfranzke: Shouldn't they return something like "InternalServerError". Why is `410 GONE` a special case?

plkokanov: So you meant to add it so that the `forbidden` error message is improved? 
Anyway for the isGone comment: `410 GONE` is handled the same way as `404 not Found` -> the function returns nil

plkokanov: updated in the latest commit to make errors like Internal Server Error transparent and directly return them to the user

rfranzke:
Thanks, can you please squash the commits? Also, I'm still confused about the special handling for GONE. When is the API server sending this? Can we really `return nil` in this case?

plkokanov:
It's quite similar to the 404 error code. If it's according to the general 410 HTTP Error code it should be used to indicate that a resource existed, but no-longer does and there is no way to redirect to a place where the resource could be found, I guess it could generally be used with caches. Looked for it in k/k and k/apiserver and indeed it's used in watch caches. It's also deprecated in the current k/k master branch in favour of ResourceExpired.. 
I'll remove it as we don't need to check for it here.

rfranzke (approved):
/lgtm

PR #1909, opened 2020-02-10
Require issuerURL/clientID to be provided, if clientID/issuerURL is provided. 

### LINKED ISSUE #1878: Add validation about spec.kubernetes.kubeAPIServer.oidcConfig
1. Create Shoot with the following partial spec included:
```yaml
spec:
  kubernetes:
    kubeAPIServer:
      oidcConfig:
        groupsClaim: groups
        issuerURL: 'https://example.com'
        usernameClaim: email
```

2. Ensure that the kube-apiserver will fail to start with:
```
Error: oidc-issuer-url and oidc-client-id should be specified together
```

Basically the same Shoot will fail to delete as it will wait `kube-apiserver` to report readiness.

We could add validation to check whether `oidcConfig.clientID` is set. Probably the same applies also to the other fields.

swilen-iwanow:
Shoot validation requires OIDC fields issuerURL/clientID to be provided, if clientID/issuerURL is provided

Fixes #1878

**Release note**:

```improvement user
Shoot configuration for OIDC authentication (Kubernetes API Server), requires if one of the fields `clientID` and `issuerURl` is provided, the other must be provided also.
```

### DIFF
--- a/pkg/apis/core/validation/shoot.go
+++ b/pkg/apis/core/validation/shoot.go
@@ -17,6 +17,7 @@ package validation
 import (
 	"fmt"
 	"math"
+	"net/url"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


rfranzke:
We cannot transform optional fields into required fields (only vice versa).
Also, it's not what was explained in the issue #1878. 

> Error: oidc-issuer-url and oidc-client-id should be specified together

I think this is rather what we have to implement, isn't it?

ialidzhikov:
Hmm, I think @rfranzke is right. Currently the apiserver does not panic if you specify for example only
```yaml
        - --oidc-username-claim=email
        - --oidc-groups-claim=groups
```

omitting the required fields `--oidc-issuer-url` and `--oidc-client-id`. But currently this PR won't allow it.

So the check should be - if one of `--oidc-issuer-url`, `--oidc-client-id` is specified, check that the other one is also specified?

ialidzhikov:
Also for `--oidc-issuer-url` - `The URL of the OpenID issuer, only HTTPS scheme will be accepted.`
So specifying `--oidc-issuer-url='http://example.com'`, fails with `Error: invalid authentication config: parse 'http://example.com': first path segment in URL cannot contain colon`. Does it make sense to add validation for the https scheme?

swilen-iwanow:
> We cannot transform optional fields into required fields (only vice versa).
> Also, it's not what was explained in the issue #1878.
> 
> > Error: oidc-issuer-url and oidc-client-id should be specified together
> 
> I think this is rather what we have to implement, isn't it?

I went trough the k8s docs (https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server) which points them as a required fields, but no worries I will change it as @ialidzhikov suggests.

vpnachev:
Ok, with the current API we cannot make them required.
But we can do it with the `v1`.
@rfranzke @ialidzhikov - if you are ok for this change in the v1, then it will be nice @swilen-iwanow to add some TODOs in the `types_shoot.go` files or somewhere else, am I right?

mvladev:
I've added all the needed validation (with some extra) in https://github.com/gardener/gardener/blob/d7b2ff73c8280ed168a2526bee7c7aac9a0554d4/pkg/apis/settings/validation/validation.go#L50-L123

it can be used as a reference

pkg/apis/core/validation/shoot.go
```diff
@@ -414,34 +415,54 @@ func validateKubernetes(kubernetes core.Kubernetes, fldPath *field.Path) field.E
...
-				allErrs = append(allErrs, field.Invalid(oidcPath.Child("issuerURL"), *oidc.IssuerURL, "issuer url cannot be empty when key is provided"))
+				allErrs = append(allErrs, field.Invalid(oidcPath.Child("groupsPrefix"), *oidc.GroupsPrefix, "groupsPrefix cannot be empty when key is provided"))
 			}
 			if oidc.SigningAlgs != nil && len(oidc.SigningAlgs) == 0 {
-				allErrs = append(allErrs, field.Invalid(oidcPath.Child("signingAlgs"), oidc.SigningAlgs, "signings algs cannot be empty when key is provided"))
+				allErrs = append(allErrs, field.Invalid(oidcPath.Child("signingAlgs"), oidc.SigningAlgs, "signingAlgs cannot be empty when key is provided"))
 			}
 			if !geqKubernetes111 && oidc.RequiredClaims != nil {
-				allErrs = append(allErrs, field.Forbidden(oidcPath.Child("requiredClaims"), "required claims cannot be provided when version is not greater or equal 1.11"))
+				allErrs = append(allErrs, field.Forbidden(oidcPath.Child("requiredClaims"), "requiredClaims field cannot be provided when version is not greater or equal 1.11"))
```
ialidzhikov:
```suggestion
				allErrs = append(allErrs, field.Forbidden(oidcPath.Child("requiredClaims"), "requiredClaims cannot be provided when version is not greater or equal 1.11"))
```

pkg/apis/core/validation/shoot_test.go
```diff
@@ -958,6 +958,58 @@ var _ = Describe("Shoot Validation Tests", func() {
 				}))
 			})
 
+			DescribeTable("should forbird issuerURL to be empty string or nil, if cleintID exists ", func(errorListSize int, issuerURL *string) {
```
ialidzhikov:
```suggestion
			DescribeTable("should forbid issuerURL to be empty string or nil, if clientID exists ", func(errorListSize int, issuerURL *string) {
```

pkg/apis/core/validation/shoot_test.go
```diff
@@ -958,6 +958,58 @@ var _ = Describe("Shoot Validation Tests", func() {
...
+
+				errorList := ValidateShoot(shoot)
+				Expect(errorList).To(HaveLen(errorListSize))
+				Expect(*errorList[0]).To(MatchFields(IgnoreExtras, Fields{
+					"Type":  Equal(field.ErrorTypeInvalid),
+					"Field": Equal("spec.kubernetes.kubeAPIServer.oidcConfig.issuerURL"),
+				}))
+			},
+				Entry("should add error if clientID is set but issuerURL is nil ", 1, nil),
+				Entry("should add error if clientID is set but issuerURL is empty string", 2, makeStringPointer("")),
```
ialidzhikov:
```suggestion
				Entry("should add error if clientID is set but issuerURL is empty string", 2, pointer.StringPtr("")),
```

pkg/apis/core/validation/shoot_test.go
```diff
@@ -958,6 +958,58 @@ var _ = Describe("Shoot Validation Tests", func() {
...
+				}))
+			})
+
+			It("should not fail if both clientID and issuerURL are set", func() {
+				shoot.Spec.Kubernetes.KubeAPIServer.OIDCConfig.IssuerURL = makeStringPointer("https://issuer.com")
+				shoot.Spec.Kubernetes.KubeAPIServer.OIDCConfig.ClientID = makeStringPointer("someClientID")
+
+				errorList := ValidateShoot(shoot)
+
+				Expect(errorList).To(HaveLen(0))
```
ialidzhikov:
```suggestion
				Expect(errorList).To(BeEmpty())
```

pkg/apis/core/validation/shoot_test.go
```diff
@@ -958,6 +958,58 @@ var _ = Describe("Shoot Validation Tests", func() {
...
+			})
+
+			DescribeTable("should forbird cleintID to be empty string or nil, if issuerURL exists ", func(errorListSize int, clientID *string) {
```
ialidzhikov:
```suggestion
			DescribeTable("should forbid clientID to be empty string or nil, if issuerURL exists ", func(errorListSize int, clientID *string) {
```

pkg/apis/core/validation/shoot_test.go
```diff
@@ -958,6 +958,58 @@ var _ = Describe("Shoot Validation Tests", func() {
...
+
+				errorList := ValidateShoot(shoot)
+				Expect(errorList).To(HaveLen(errorListSize))
+				Expect(*errorList[0]).To(MatchFields(IgnoreExtras, Fields{
+					"Type":  Equal(field.ErrorTypeInvalid),
+					"Field": Equal("spec.kubernetes.kubeAPIServer.oidcConfig.clientID"),
+				}))
+			},
+				Entry("should add error if issuerURL is set but clientID is nil", 1, nil),
+				Entry("should add error if issuerURL is set but clientID is empty string ", 2, makeStringPointer("")),
```
ialidzhikov:
```suggestion
				Entry("should add error if issuerURL is set but clientID is empty string ", 2, pointer.StringPtr(("")),
```

ialidzhikov (commented):
Minor nits.

pkg/apis/core/validation/shoot.go
```diff
@@ -414,34 +415,54 @@ func validateKubernetes(kubernetes core.Kubernetes, fldPath *field.Path) field.E
...
+					allErrs = append(allErrs, field.Invalid(oidcPath.Child("clientID"), oidc.ClientID, "clientID must be set when issuerURL is provided"))
+				}
+			}
+
+			if fieldNilOrEmptyString(oidc.IssuerURL) {
+				if oidc.IssuerURL != nil {
+					allErrs = append(allErrs, field.Invalid(oidcPath.Child("issuerURL"), oidc.ClientID, "issuerURL cannot be empty when key is provided"))
+				}
+				if !fieldNilOrEmptyString(oidc.ClientID) {
+					allErrs = append(allErrs, field.Invalid(oidcPath.Child("issuerURL"), oidc.ClientID, "issuerURL must be set when clientID is provided"))
```
ialidzhikov:
```suggestion
					allErrs = append(allErrs, field.Invalid(oidcPath.Child("issuerURL"), oidc.IssuerURL, "issuerURL must be set when clientID is provided"))
```

pkg/apis/core/validation/shoot.go
```diff
@@ -414,34 +415,54 @@ func validateKubernetes(kubernetes core.Kubernetes, fldPath *field.Path) field.E
...
+					allErrs = append(allErrs, field.Invalid(oidcPath.Child("clientID"), oidc.ClientID, "clientID cannot be empty when key is provided"))
+				}
+				if !fieldNilOrEmptyString(oidc.IssuerURL) {
+					allErrs = append(allErrs, field.Invalid(oidcPath.Child("clientID"), oidc.ClientID, "clientID must be set when issuerURL is provided"))
+				}
+			}
+
+			if fieldNilOrEmptyString(oidc.IssuerURL) {
+				if oidc.IssuerURL != nil {
+					allErrs = append(allErrs, field.Invalid(oidcPath.Child("issuerURL"), oidc.ClientID, "issuerURL cannot be empty when key is provided"))
```
ialidzhikov:
```suggestion
					allErrs = append(allErrs, field.Invalid(oidcPath.Child("issuerURL"), oidc.IssuerURL, "issuerURL cannot be empty when key is provided"))
```

ialidzhikov (commented):
Nit

ialidzhikov (approved):
/lgtm

rfranzke (approved):

PR #1911, opened 2020-02-10
Set shoot conditions to unknown if gardenlet stops sending heartbeats

### LINKED ISSUE #1877: Shoot conditions should be set to "unknown" when gardenlet is not active anymore
The gardener-controller-manager should set the conditions' status to `Unknown` for shoots running on seeds whose gardenlet is not reporting a heartbeat anymore (similar to how it sets the `GardenletReady` condition on `Seed`s to `Unknown` after a certain threshold).

**Why is this needed**:
Today, when the Gardenlet is no longer active, the shoot conditions for all shoots assigned to the respective seed will not change anymore. Hence, it shows inaccurate data, false positive or false negative statuses. We should set it to `Unknown` to indicate that we can't judge at this point in time.

rfranzke:
The gardener-controller-manager now sets all shoot conditions and constraints to `Unknown` in case the responsible gardenlet is not sending heartbeats anymore.
Concretely, if the `GardenletReady` condition is `Unknown` for at least the configured `shootMonitorPeriod` then we will mark the conditions and constraints for all the shoots that belong to this seed as `Unknown`. The reason is that the gardenlet didn't send a heartbeat anymore, hence, it most likely didn't check the shoot status. This means that the current shoot status might not reflect the truth anymore. We are indicating this by marking it as `Unknown`.

Fixes #1877

**Release note**:

```noteworthy operator
The gardener-controller-manager configuration now supports a new `.controllers.seed.shootMonitorPeriod` field. This field controls the time after which the controller-manager will set the status to `Unknown` for all the shoot conditions and constraints in case the responsible gardenlet does not send heartbeats anymore.
```

### DIFF
--- a/charts/gardener/controlplane/charts/runtime/templates/controller-manager/configmap-componentconfig.yaml
+++ b/charts/gardener/controlplane/charts/runtime/templates/controller-manager/configmap-componentconfig.yaml
@@ -57,6 +57,9 @@ data:
         {{- if .Values.global.controller.config.controllers.seed.monitorPeriod }}
         monitorPeriod: {{ .Values.global.controller.config.controllers.seed.monitorPeriod }}
         {{- end }}
+        {{- if .Values.global.controller.config.controllers.seed.shootMonitorPeriod }}
+        shootMonitorPeriod: {{ .Values.global.controller.config.controllers.seed.shootMonitorPeriod }}
+        {{- end }}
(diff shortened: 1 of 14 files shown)


danielfoehrKn (dismissed):

timebertt (dismissed):
/lgtm

PR #1914, opened 2020-02-10
Remove check for worker pool zones

rfranzke:
Gardener cannot make assumptions about the `zones` of worker pools, so it shouldn't validate anything here. Instead, the provider extension is responsible.

PS: I anyways added tests for adding, removing, and swapping worker pools to g/g validation.

**Release note**:

```improvement user
A bug has been fixed that prevented worker pools from being added to existing shoot clusters.
```


timuthy (approved):
/lgtm

Issue #1351, opened 2019-08-19
Monitoring and logging configuration for extension controllers


rfranzke:
**What would you like to be added**:
While progressing with #308 we have moved out most of the provider-specific information of the Gardener code. However, there are still left-overs for monitoring and logging configuration of provider-specific components, e.g. for the cloud-controller-manager ([metric scrape config](https://github.com/gardener/gardener/blob/ff1c8d1a39528aee41d5ed7867f6b2b05011a8b2/charts/seed-monitoring/charts/core/charts/prometheus/templates/config.yaml#L318-L345), [alerting rules](https://github.com/gardener/gardener/blob/2d3f929213150ab88d25de684eb4573f0c7107c9/charts/seed-monitoring/charts/core/charts/prometheus/rules/cloud-controller-manager.rules.yaml), [dashboards](https://github.com/gardener/gardener/blob/bb4b16a385ec3d7c59df3ebcacfa6e704e793e29/charts/seed-monitoring/charts/grafana/dashboards/owners/kubernetes-control-plane-status-dashboard.json#L323), and [logging parse configuration](https://github.com/gardener/gardener/blob/2ea4dc7d7cb1797f96373fcf6073dda49ebbad6c/charts/seed-bootstrap/charts/fluentd-es/templates/fluent-bit-configmap.yaml#L85-L90)).
We should also move these parts to the provider-specific extension code, because only the extensions know these components and how they are scraped/displayed/... in the best way.

**Why is this needed**:
Part of #308, see motivation in [GEP-1](https://github.com/gardener/gardener/blob/master/docs/proposals/01-extensibility.md).

**Implementation proposal**:
The general idea is that the extension controllers, as part of the control loops that deploy provider-specific components, may write a `ConfigMap` (one or many) into the shoot namespace in the seed. This `ConfigMap` contains the respective information for Gardener's logging and monitoring stacks. It must be labelled in a special way such that Gardener finds them.
During the steps in Gardener's control loops that are deploying the logging/monitoring configuration it finds all such `ConfigMap`s, extends the default/non-provider-specific configuration, and finally deploys the logging/monitoring stacks.

#### Logging

In case of logging the parse information is always the same and not shoot-specific. Also, the configuration is part of `fluent-bit` which is deployed once per seed and scrapes all containers by default.
Here, the mentioned `ConfigMap` should be created in the `garden` namespace in the seed (instead of in any shoot namespace).
[...]

PR #1918, opened 2020-02-11
Expose kube-scheduler and cluster-autoscaler logs to end users

vpnachev:
With this PR the end users will be able to see the logs from `kube-scheduler` and `cluster-autoscaler` in the Kibana dashboard, just like they are already doing for kube-apiserver, kube-controller-manager and cloud-controller-manager.

Fixes #logging

**Release note**:

```improvement user
The Gardener users now have access to the logs from the `kube-scheduler` and `cluster-autoscaler` of their shoot clusters.
```

### DIFF
--- a/charts/seed-bootstrap/charts/fluentd-es/templates/fluent-bit-configmap.yaml
+++ b/charts/seed-bootstrap/charts/fluentd-es/templates/fluent-bit-configmap.yaml
@@ -252,7 +252,17 @@ data:
 
     [FILTER]
         Name record_modifier
+        Match *kube-scheduler*
+        Record type user
+
+    [FILTER]
+        Name record_modifier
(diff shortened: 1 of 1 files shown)


vpnachev:
@MartinWeindel @timuthy you might want to do the same for dns and certificate extension controllers.
If you think it is safe to expose the logs from these controllers to Gardener end users, you have to follow the concept defined in #1351. An example is the `provider-alicloud` extension https://github.com/gardener/gardener-extension-provider-alicloud/blob/dd98eb997b1027b3fe629b01ebf79a68505f4659/charts/gardener-extension-provider-alicloud/templates/configmap-logging.yaml#L24-L27

rfranzke (approved):

timebertt (approved):
/lgtm

PR #1919, opened 2020-02-11
Allow authenticated users to read `Seed` resources

rfranzke:
We should expose in which regions and with which configuration (networking, etc.) we have seed clusters to our end-users so that they can better define their shoot resources (e.g., disjunct networks).

/cc @vlerenc 

**Release note**:

```noteworthy operator
The `gardener.cloud:system:cloudprofiles` `ClusterRole` and `ClusterRoleBinding` resources have been renamed to `gardener.cloud:system:read-global-resources`. They are now allowing all authenticated users to read `CloudProfile` and `Seed` resources.
```
```noteworthy user
Authenticated users can now read/list/watch `Seed` resources.
```

### DIFF
--- a/charts/gardener/controlplane/charts/application/templates/rbac-user.yaml
+++ b/charts/gardener/controlplane/charts/application/templates/rbac-user.yaml
@@ -44,12 +44,12 @@ aggregationRule:
       gardener.cloud/role: project-member
 rules: []
 
-# Cluster role with cluster role binding allowing all authenticated users to read the cloudprofiles
+# Cluster role with cluster role binding allowing all authenticated users to read some global resources
 ---
 apiVersion: {{ include "rbacversion" . }}
 kind: ClusterRole
 metadata:
-  name: gardener.cloud:system:cloudprofiles
+  name: gardener.cloud:system:read-global-resources
   labels:
     app: gardener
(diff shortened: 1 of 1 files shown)


vlerenc:
Thanks @rfranzke for the very quick PR.

Some days ago, @vasu1124 suggested that we open up "seeds" in a broader way. I think, what he wants is that end users can start creating their own seeds and register them privately with their projects to be used only for their shoots. WDYT? That would turn them from cluster into namespaced resources. Not sure whether and when we would start thinking more about that or implement that, but when and if we do that, we will certainly introduce a new version, so whatever we do here in this PR will anyways look different in a future where we allow that (if we allow that), I guess.

vpnachev:
> That would turn them(Seeds) from cluster into namespaced resources.

If they are namespaced, I would like to share such seeds between multiple projects(namespaces) of mine.
Maybe something like `SeedBinding`(similar to SecretBinding) between the project with the shoots and the Seed itself can do the trick.

timebertt (approved):
/lgtm

vpnachev (approved):
I remember I had some concerns about this an year ago, but looking through a sample seed resource now I cannot recall what was the reason - maybe it was enhanced with the new API and it is no longer valid.

PR #1925, opened 2020-02-12
Validate worker zones only during creation and update

timuthy:
This PR makes Gardener only check configured availability zones for workers if the shoot is created or the `zones` portion is updated. Availability zones being removed from the `CloudProfile` shouldn't break once successfully reconciled shoots.

/cc @rfranzke 

**Release note**:

```improvement operator
The availability of configures worker `zones` is now only cross checked with the referring `CloudProfile` if a shoot is created or the worker `zones` in a shoot are changed. This avoids the breakage of existing shoots whose availability zones were removed from the `CloudProfile` afterwards.
```


timebertt (approved):
/lgtm

rfranzke (approved):
/lgtm

PR #1927, opened 2020-02-13
Fix check in SecretBinding controller for secret release

rfranzke:
The `mayReleaseSecret` function is called when a `SecretBinding` has a deletion timestamp. It lists all `SecretBinding`s in all namespaces, however, it was not excluding the `SecretBinding` it was called for. Hence, finalizer release for secrets could never succeed.

Thanks @Diaphteiros and @timuthy for reporting!

**Release note**:

```improvement user
The controller-manager does now correctly release the finalizer in`Secret`s if no referencing `SecretBinding`s exist anymore.
```


timuthy (approved):
/lgtm

timebertt (approved):
/lgtm

PR #1930, opened 2020-02-13
Make volume type optional for Worker resources

timuthy:
Certain extension controllers might not need the type information of a volume. Therefore, this field should be optional for `Worker` resources.

/cc @afritzler

**Release note**:

```improvement user
A bug has been fixed which caused a reconciliation failure for shoot clusters which specify `spec.provider.workers[].volume.size` without `spec.provider.workers[].volume.type`.
```


afritzler (approved):
lgtm

rfranzke (approved):

PR #1931, opened 2020-02-13
Increase sync period for central gardener-resource-manager

rfranzke:
This PR increases the sync period for the central seed grm to `1h` as it is not required to do it more often (only operators have access to the seed). For the per-shoot grm we don't change the sync period, however, we would like to get https://github.com/gardener/gardener-resource-manager/pull/33 in (with a separate PR once a release of g/grm is built) which drastically decreases the cache resync period. The motivation for both is to reduce unnecessary egress traffic.

* /cc @vlerenc
* Also related: https://github.com/gardener/gardener-resource-manager/pull/33 and a new g/grm release.

**Release note**:

```noteworthy user
NONE
```

### DIFF
--- a/charts/seed-bootstrap/templates/gardener-resource-manager/deployment.yaml
+++ b/charts/seed-bootstrap/templates/gardener-resource-manager/deployment.yaml
@@ -33,6 +33,7 @@ spec:
         - --max-concurrent-workers={{ .Values.gardenerResourceManager.concurrentSyncs }}
         - --resource-class={{ .Values.gardenerResourceManager.resourceClass }}
         - --sync-period={{ .Values.gardenerResourceManager.syncPeriod }}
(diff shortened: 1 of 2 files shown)


timuthy:
> For the per-shoot grm it is increased to `5m`

Sorry if I get something wrong but I don't see this change here?!

rfranzke:
Sorry, I didn't update the description. After talking with @vlerenc I looked a little deeper into the GRM for shoots. What actually causes unnecessary traffic is the resyncing of the cache (LIST calls against the seed API server every minute). With https://github.com/gardener/gardener-resource-manager/pull/33 this is mitigated. The cache will only be resynced every 24h, so the GRM can keep reconciling MRs every minute as the communication to the shoot API server is only cluster internal.
Will update the description, thanks.

timuthy (approved):
/lgtm

PR #1937, opened 2020-02-14
Allow sha256 image tags in imagevector

rfranzke:
The image vector now supports sha256 image tags. Along the way, I fixed a bug in the scoring and improved it so that equal matches lead to a higher score than `>=` or `<=` matches.

/cc @timuthy @ccwienk @AndreasBurger @RaphaelVogel @vlerenc 

**Release note**:

```improvement operator
The imagevector does now support `sha256` image tags.
```

### DIFF
--- a/charts/gardener/controlplane/charts/runtime/templates/apiserver/deployment.yaml
+++ b/charts/gardener/controlplane/charts/runtime/templates/apiserver/deployment.yaml
@@ -61,7 +61,7 @@ spec:
       {{- end }}
       containers:
       - name: gardener-apiserver
-        image: {{ required ".Values.global.apiserver.image.repository is required" .Values.global.apiserver.image.repository }}:{{ required ".Values.global.apiserver.image.tag is required" .Values.global.apiserver.image.tag }}
+        image: {{ include "utils-templates.image" .Values.global.apiserver.image }}
         imagePullPolicy: {{ .Values.global.apiserver.image.pullPolicy }}
         command:
         - /gardener-apiserver
--- a/pkg/operation/operation.go
+++ b/pkg/operation/operation.go
@@ -374,23 +374,19 @@ func (o *Operation) ShootVersion() string {
 	return o.Shoot.Info.Spec.Kubernetes.Version
 }
 
-func (o *Operation) injectImages(values map[string]interface{}, names []string, opts ...imagevector.FindOptionFunc) (map[string]interface{}, error) {
-	return chart.InjectImages(values, o.ImageVector, names, opts...)
-}
-
 // InjectSeedSeedImages injects images that shall run on the Seed and target the Seed's Kubernetes version.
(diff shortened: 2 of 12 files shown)


charts/gardener/controlplane/charts/runtime/templates/apiserver/deployment.yaml
```diff
@@ -61,7 +61,11 @@ spec:
       {{- end }}
       containers:
       - name: gardener-apiserver
+        {{- if hasPrefix .Values.global.apiserver.image.tag "sha256:" }}
+        image: {{ required ".Values.global.apiserver.image.repository is required" .Values.global.apiserver.image.repository }}@{{ required ".Values.global.apiserver.image.tag is required" .Values.global.apiserver.image.tag }}
+        {{- else }}
         image: {{ required ".Values.global.apiserver.image.repository is required" .Values.global.apiserver.image.repository }}:{{ required ".Values.global.apiserver.image.tag is required" .Values.global.apiserver.image.tag }}
+        {{- end }}
```
timuthy: This nicely fits to the utils charts as a template function. WDYT?

pkg/operation/operation.go
```diff
@@ -366,6 +366,9 @@ func (o *Operation) CleanShootTaskError(ctx context.Context, taskID string) {
 
 // SeedVersion is a shorthand for the kubernetes version of the K8sSeedClient.
 func (o *Operation) SeedVersion() string {
+	if o.Seed.Info.Status.KubernetesVersion != nil {
+		return *o.Seed.Info.Status.KubernetesVersion
```
timuthy: Did you add this for a specific reason / use case?
I don't see a clear benefit because now it depends on the reconciliation of the seed resource whereas `o.K8sSeedClient.Version()` always holds the version read from the seed's api server.

charts/gardener/controlplane/charts/runtime/templates/apiserver/deployment.yaml
timebertt: If I see it correctly, the only difference between the if/else branch is the `:`/`@` delimiter.
WDYT about adding a small helper template (in `utils-templates`?) that takes a image struct and prints either `repository:tag` or `repository@sha256:abcdef...`?

timebertt (dismissed):
/lgtm
Just a small suggestion from my side.

timuthy (changes requested):
Nice! Just a few comments. LGTM otherwise.

pkg/operation/operation.go
rfranzke: No specific reason. I thought I can get rid of some unneeded call for the version here, but it is anyway cached and stored at the client, so I reverted this change. Thanks for asking back.

timuthy (approved):

timebertt (approved):
/lgtm

PR #1945, opened 2020-02-17
Support Custom Pod Labels for Gardener runtime

vpnachev:
Support Custom Pod Annotations and Labels for Gardener runtime components.

A valid use case is to enable seamless integration with network policies in the hosting cluster.

Fixes #

**Release note**:

```improvement operator
The pods for the gardener components `apiserver`, `controller manager`, `scheduler` and `gardenlet` can be extended with custom annotations and labels via the gardener helm [chart](https://github.com/gardener/gardener/tree/master/charts/gardener). 
```


rfranzke (commented):
/lgtm, thank you. Though, while you're at it, can you do the same for annotations?

vpnachev:
Sure. Is there any particular use-case for them?

rfranzke:
I can't think of any, but they are similar to labels so it might make sense to expose them as well.

vpnachev:
done

rfranzke (approved):
/lgtm

Issue #1939, opened 2020-02-14
Seed bootstrapping fails


timebertt:
**What happened**:
#1863 introduced a bug which causes the Seed bootstrapping to fail for a new Seed.

Bootstrapping fails on apllying the newly introduced VPA `kube-state-metrics-vpa` (see [here](https://github.com/gardener/gardener/blob/master/charts/seed-bootstrap/charts/kube-state-metrics/templates/kube-state-metrics.yaml#L47-L58)):
```
time="2020-02-14T11:10:05+01:00" level=error msg="no matches for kind \"VerticalPodAutoscaler\" in version \"autoscaling.k8s.io/v1beta2\"" seed=local-seed
```
It looks like the issue here is the order in which we apply the rendered resources (VPA created before VPA CRD is created).

**What you expected to happen**:
Seed bootstrapping to succeed.

**How to reproduce it (as minimally and precisely as possible)**:
1. Spin up a cluster somewhere.
2. Register it as a seed.

**Anything else we need to know?**:

**Environment**:

- Gardener version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

rfranzke:
Duplicate of https://github.com/gardener/gardener/issues/1871?

ialidzhikov:
Wasn't this error message present only until the vpa components are started?

timebertt:
> Duplicate of #1871?

No, this is the actual Seed Bootstrapping that fails, not the installation of the ControllerInstallation's MR.


> Wasn't this error message present only until the vpa components are started?

The vpa components won't come up, because they are part of the same chart and the chart applier fails and returns before it can apply the vpa components...

rfranzke:
> The vpa components won't come up, because they are part of the same chart and the chart applier fails and returns before it can apply the vpa components...

In the next reconciliation the VPA CRD is installed and then also the VPA CRs should be able to get created.

The root cause is the same like in #1871, so I suggest closing one issue.

rfranzke:
After a deeper look it turns out that the root cause here is indeed completely different: Our Kubernetes applier returns an error if it cannot create/update a resource, causing the subsequent resources to not be considered at all. Hence, it will never succeed because the VPA CRD is coming after the VPA CR for the kube-state-metrics.

[...]

PR #1947, opened 2020-02-17
Fix manifest sorting in chart renderer

timebertt:
With this PR, the Chart Renderer splits up rendered chart template files into individual manifests before sorting them by Kind.
This is done to allow proper sorting of the manifests (for example: VPA CRD should be applied before VPA CRs). Previously the chart renderer only sorted rendered files as a whole and not individual manifests. This can cause issues when applying the rendered manifests in case there are multiple manifests of different kinds in a single file.

Ref #1939

I will open a follow-up PR for removing the early exit on error in the chart applier.

**Release note**:

```improvement developer
The Chart Renderer now splits up rendered template files into individual manifests to ensure proper sorting of the manifests.
```

### DIFF
--- a/pkg/chartrenderer/default.go
+++ b/pkg/chartrenderer/default.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/helm/pkg/engine"
 	"k8s.io/helm/pkg/manifest"
 	chartapi "k8s.io/helm/pkg/proto/hapi/chart"
+	"k8s.io/helm/pkg/releaseutil"
 	"k8s.io/helm/pkg/timeconv"
 )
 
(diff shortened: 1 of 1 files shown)


rfranzke (dismissed):
/lgtm

pkg/chartrenderer/default.go
```diff
@@ -141,14 +142,21 @@ func (r *chartRenderer) renderResources(ch *chartapi.Chart, values chartutil.Val
...
 		if strings.HasSuffix(k, notesFileSuffix) || strings.HasPrefix(path.Base(k), "_") {
 			delete(files, k)
 		}
 	}
 
-	manifests := manifest.SplitManifests(files)
+	manifestsMap := map[string]string{}
+	for origKey, file := range files {
+		for key, m := range releaseutil.SplitManifests(file) {
+			manifestsMap[origKey+"#"+key] = m
```
ialidzhikov: I thinks the controllerinstallation control is using this pkg and `#` in the manifest key breaks it:

```
time="2020-02-19T10:12:49+02:00" level=info msg="Error syncing ControllerInstallation provider-gcp-5pq58: Secret \"provider-gcp-5pq58\" is invalid: [data[provider-gcp_templates_vpa.yaml#manifest-0]: Invalid value: \"provider-gcp_templates_vpa.yaml#manifest-0\": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+'),
```

ialidzhikov (changes requested):
Generally it is a nice catch.

pkg/chartrenderer/default.go
timebertt: Thanks, nice catch.
I was not sure, which separator to pick...
WDYT: should we use `/` (will be replaced with `_` in ctrlinst controller) or should we use `_` directly?

ialidzhikov: Both of them looks good to me.

timebertt:
Fixed the map keys to use `_` as a separator.
ControllerInstallations work again now. The ctrlinst secret keys will look like this: `gardener-extension-provider-mock_templates_configmap.yaml_manifest-0`.

PTAL :)

ialidzhikov (approved):
/lgtm

PR #1948, opened 2020-02-17
Extension roles for project members

### LINKED ISSUE #1943: Support custom roles for project members
Today, a member in a `Project` can have exactly one role (`.spec.members[*].role`). This is very limiting and only supports roles that are "hard-coded" in the GCM's project controller. Let's extend this model slightly to a list of roles (`.spec.members[*].roles`; deprecate `.spec.members[*].role`) and allow to add "custom" roles (not known by GCM's project controller, but corresponding role bindings in the project namespaces are automatically reconciled by it).

**Why is this needed**:
Support out-of-tree extensions that want to extend the project member role model.
/cc @timuthy @tim-ebert @vlerenc @grolu @wyb1

rfranzke:
This PR adds a new list of roles to the `Project` resource (`.spec.members[*].roles` and deprecates the old `.spec.members[*].role` field). It also prepares the deprecation/removal of `.spec.owner` in favour of the new `owner` role.
Moreover, it enables adding extension roles not known by Gardener with the `extension:` prefix. The project controller will automatically create an aggregation `ClusterRole` and a `RoleBinding` which targets the members assigned to this extension role.

Fixes #1943

**Release note**:

```noteworthy developer
It is now possible to let the project controller manage extension roles for access to the Gardener-managed resources in the garden cluster. Please consult [this documentation](https://github.com/gardener/gardener/blob/master/docs/extensions/project-roles.md) for further information.
```


rfranzke:
@wyb1 @petersutter please check [the `Project` example](https://github.com/rfranzke/gardener/blob/feature/project-roles/example/05-project-dev.yaml) as well as [this documentation](https://github.com/rfranzke/gardener/blob/feature/project-roles/docs/extensions/custom-project-roles.md). Is it sufficiently described for you to understand what needs to be done?

pkg/apis/core/validation/project.go
```diff
@@ -61,8 +64,22 @@ func ValidateProjectUpdate(newProject, oldProject *core.Project) field.ErrorList
...
 	for i, member := range projectSpec.Members {
-		allErrs = append(allErrs, ValidateSubject(member.Subject, fldPath.Child("members").Index(i))...)
+		idxPath := fldPath.Child("members").Index(i)
+
+		allErrs = append(allErrs, ValidateProjectMember(member, idxPath)...)
+
+		for j, role := range member.Roles {
+			if role == core.ProjectMemberOwner {
+				if ownerFound {
+					allErrs = append(allErrs, field.Forbidden(idxPath.Child("roles").Index(j), role))
```
petersutter, line 78: is only one owner allowed? if yes, why? For compatibility reasons?

petersutter: > Is it sufficiently described for you to understand what needs to be done?

yes, thanks

petersutter:
> Is it sufficiently described for you to understand what needs to be done?

yes, thanks

mvladev:
Several issues:

- No validation checks for the size of the generated `ClusterRole` / `Binding` - `gardener.cloud:system:project-custom:` is already 37 characters which leave us with `27` (64 max name limit) 
- the `custom:` prefix sounds too magical - I would put it in a dedicated field? Or aren't all those roles custom??

rfranzke:
Thanks @mvladev, I'll try to make the name shorter.
Regarding the prefix: We decided to not change the API too much for now because the whole role/user concepts are not clear yet. Introducing `roles` in favour of `role` is a minimal change to support oot extensions without obstructing the API too much until the UAM concept is more clear.

pkg/apis/core/validation/project.go
rfranzke: Hm yes, mainly because it was like this before. We can remove the `.spec.owner` field in `v1` and then allow to have multiple owners. WDYT?

petersutter: Sounds good

docs/extensions/custom-project-roles.md
```diff
@@ -0,0 +1,35 @@
+# Custom project roles
+
+The `Project` resource allows to specify a list of roles for every member (`.spec.members[*].roles`).
+There are a few standard roles defined by Gardener itself:
+
+* `owner` (describes the owner/main contact of the project)
```
timebertt: Can you add a short sentence, that ATM only one owner can be specified?

docs/extensions/custom-project-roles.md
```diff
@@ -0,0 +1,35 @@
...
+    kind: User
+    name: alice.doe@example.com
+    roles:
+    - "admin"
+    - "owner"
+    - "custom:foo"
+```
+
+The project controller will, for every custom role, create a `ClusterRole` with name `name: gardener.cloud:custom:project:<projectName>:<roleName>`, i.e., for above example: `name: gardener.cloud:custom:project:dev:foo`.
+This `ClusterRole` aggregates other `ClusterRole`s that are labeled with `project.gardener.cloud/custom-role=foo` which might be created by extension controllers.
```
timebertt: Can you mention here, that extensions are also able to add permissions to gardener's default roles (`admin`/`viewer`) for their resources by adding the respective aggregation label (like `rbac.gardener.cloud/aggregate-to-project-member: "true"`)?

docs/extensions/custom-project-roles.md
```diff
@@ -0,0 +1,35 @@
...
+  name: dev
+spec:
+  members:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: User
+    name: alice.doe@example.com
+    roles:
+    - "admin"
+    - "owner"
+    - "custom:foo"
```
timebertt: As already discussed in person, I'm not a big fan of the prefix `custom:`. One could get the impression that these roles are something creating by the enduser.
I would be fine with something like: `extension:` or `add-on:`.
Still, I'm not really happy with those either. May have to give it some more thought.

pkg/apis/core/v1alpha1/conversions.go
```diff
@@ -193,3 +193,112 @@ func Convert_v1alpha1_ShootStatus_To_core_ShootStatus(in *ShootStatus, out *core
...
+	outer:
+		for i, member := range out.Members {
+			if member.Name == owner.Name && member.APIGroup == owner.APIGroup && member.Kind == owner.Kind {
+				mustAddOwnerRole := true
+				for _, role := range member.Roles {
+					if role == core.ProjectMemberOwner {
+						mustAddOwnerRole = false
+						break outer
+					}
+				}
```
timebertt: Why don't you use your helper function `ProjectMemberHasRole()` here?

pkg/apis/core/v1alpha1/conversions.go
```diff
@@ -193,3 +193,112 @@ func Convert_v1alpha1_ShootStatus_To_core_ShootStatus(in *ShootStatus, out *core
...
+	if err := autoConvert_core_ProjectSpec_To_v1alpha1_ProjectSpec(in, out, s); err != nil {
+		return err
+	}
+
+	if owner := out.Owner; owner != nil {
+	outer:
+		for i, member := range out.Members {
+			if member.Name == owner.Name && member.APIGroup == owner.APIGroup && member.Kind == owner.Kind {
+				mustAddOwnerRole := true
+				for _, role := range member.Roles {
```
timebertt, line 238: Same here

pkg/apis/core/v1alpha1/conversions.go
timebertt: Also, why are you stepping out of the outer loop?
This way the second loop for deleting the owner role from other members will not be executed...

pkg/apis/core/v1alpha1/conversions.go
```diff
@@ -193,3 +193,112 @@ func Convert_v1alpha1_ShootStatus_To_core_ShootStatus(in *ShootStatus, out *core
...
+					out.Members[i].Roles = append(out.Members[i].Roles[:j], out.Members[i].Roles[j+1:]...)
+				}
+			}
+			if member.Role != nil && *member.Role == ProjectMemberOwner {
+				if len(out.Members[i].Roles) > 0 {
+					out.Members[i].Role = &out.Members[i].Roles[0]
+				} else {
+					out.Members[i].Role = nil
+				}
+			}
```
timebertt: I'm not sure, if I understand the purpose of this snippet.
Why are you overriding `.member.role` if it is `== ProjectMemberOwner`?
Isn't the role copying already done by `Convert_core_ProjectMember_To_v1alpha1_ProjectMember`?

pkg/apis/core/v1alpha1/conversions_test.go
```diff
@@ -81,4 +89,348 @@ var _ = Describe("Conversion", func() {
...
+			)
+
+			BeforeEach(func() {
+				in = &ProjectMember{}
+				out = &core.ProjectMember{}
+			})
+
+			It("should do nothing because role not set", func() {
+				Expect(scheme.Convert(in, out, nil)).To(BeNil())
+				Expect(out).To(Equal(out))
```
timebertt:
```suggestion
				Expect(out).To(Equal(&core.ProjectMember{}))
```

pkg/apis/core/v1alpha1/conversions_test.go
```diff
@@ -81,4 +89,348 @@ var _ = Describe("Conversion", func() {
...
+			)
+
+			BeforeEach(func() {
+				in = &core.ProjectMember{}
+				out = &ProjectMember{}
+			})
+
+			It("should do nothing because roles are not set", func() {
+				Expect(scheme.Convert(in, out, nil)).To(BeNil())
+				Expect(out).To(Equal(out))
```
timebertt:
```suggestion
				Expect(out).To(Equal(&ProjectMember{}))
```

pkg/apis/core/v1alpha1/types_project.go
```diff
@@ -90,14 +91,23 @@ type ProjectMember struct {
 	// account that has a certain role.
 	rbacv1.Subject `json:",inline"`
 	// Role represents the role of this member.
-	Role string `json:"role"`
+	// Deprecated: Use roles instead.
```
timebertt: Can you specify here, what the behaviour of the `Role` field is (copying from and to `Roles`)?

pkg/apis/core/v1beta1/conversions.go
```diff
@@ -39,3 +40,112 @@ func addConversionFuncs(scheme *runtime.Scheme) error {
...
+	if err := autoConvert_v1beta1_ProjectSpec_To_core_ProjectSpec(in, out, s); err != nil {
+		return err
+	}
+
+	if owner := out.Owner; owner != nil {
+	outer:
+		for i, member := range out.Members {
+			if member.Name == owner.Name && member.APIGroup == owner.APIGroup && member.Kind == owner.Kind {
+				mustAddOwnerRole := true
+				for _, role := range member.Roles {
```
timebertt, line 54: Can you use `ProjectMemberHasRole()` for that?

pkg/apis/core/v1beta1/conversions.go
```diff
@@ -39,3 +40,112 @@ func addConversionFuncs(scheme *runtime.Scheme) error {
...
+	if err := autoConvert_core_ProjectSpec_To_v1beta1_ProjectSpec(in, out, s); err != nil {
+		return err
+	}
+
+	if owner := out.Owner; owner != nil {
+	outer:
+		for i, member := range out.Members {
+			if member.Name == owner.Name && member.APIGroup == owner.APIGroup && member.Kind == owner.Kind {
+				mustAddOwnerRole := true
+				for _, role := range member.Roles {
```
timebertt, line 85: Same here

pkg/apis/core/v1beta1/conversions.go
```diff
@@ -39,3 +40,112 @@ func addConversionFuncs(scheme *runtime.Scheme) error {
...
+func Convert_v1beta1_ProjectMember_To_core_ProjectMember(in *ProjectMember, out *core.ProjectMember, s conversion.Scope) error {
+	if err := autoConvert_v1beta1_ProjectMember_To_core_ProjectMember(in, out, s); err != nil {
+		return err
+	}
+
+	if in.Role == nil {
+		return nil
+	}
+
+	for _, role := range in.Roles {
```
timebertt, line 123: Same here

pkg/apis/core/v1alpha1/conversions.go
```diff
@@ -193,3 +193,112 @@ func Convert_v1alpha1_ShootStatus_To_core_ShootStatus(in *ShootStatus, out *core
...
+func Convert_v1alpha1_ProjectMember_To_core_ProjectMember(in *ProjectMember, out *core.ProjectMember, s conversion.Scope) error {
+	if err := autoConvert_v1alpha1_ProjectMember_To_core_ProjectMember(in, out, s); err != nil {
+		return err
+	}
+
+	if in.Role == nil {
+		return nil
+	}
+
+	for _, role := range in.Roles {
```
timebertt, line 276: Also, what about `ProjectMemberHasRole()` here?

pkg/apis/core/v1alpha1/defaults_test.go
```diff
@@ -0,0 +1,132 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
timebertt:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
See also all other new files...

pkg/apis/core/v1beta1/conversions_test.go
```diff
@@ -0,0 +1,385 @@
...
+			)
+
+			BeforeEach(func() {
+				in = &ProjectMember{}
+				out = &core.ProjectMember{}
+			})
+
+			It("should do nothing because role not set", func() {
+				Expect(scheme.Convert(in, out, nil)).To(BeNil())
+				Expect(out).To(Equal(out))
```
timebertt:
```suggestion
				Expect(out).To(Equal(&core.ProjectMember{}))
```

pkg/apis/core/v1beta1/conversions_test.go
```diff
@@ -0,0 +1,385 @@
...
+			)
+
+			BeforeEach(func() {
+				in = &core.ProjectMember{}
+				out = &ProjectMember{}
+			})
+
+			It("should do nothing because roles are not set", func() {
+				Expect(scheme.Convert(in, out, nil)).To(BeNil())
+				Expect(out).To(Equal(out))
```
timebertt:
```suggestion
				Expect(out).To(Equal(&ProjectMember{}))
```

pkg/apis/core/validation/project.go
```diff
@@ -61,8 +64,22 @@ func ValidateProjectUpdate(newProject, oldProject *core.Project) field.ErrorList
...
 
+	ownerFound := false
+
 	for i, member := range projectSpec.Members {
-		allErrs = append(allErrs, ValidateSubject(member.Subject, fldPath.Child("members").Index(i))...)
+		idxPath := fldPath.Child("members").Index(i)
+
+		allErrs = append(allErrs, ValidateProjectMember(member, idxPath)...)
+
+		for j, role := range member.Roles {
```
timebertt, line 75: What about using `ProjectMemberHasRole()` here for finding the owner?

pkg/apis/core/validation/project.go
```diff
@@ -114,6 +131,41 @@ func ValidateSubject(subject rbacv1.Subject, fldPath *field.Path) field.ErrorLis
 	return allErrs
 }
 
+var supportedRoles = sets.NewString(
+	core.ProjectMemberAdmin,
+	core.ProjectMemberViewer,
+	core.ProjectMemberOwner,
+)
+
+const customRoleMaxLength = 15
```
timebertt: Why is it now `15`? Any specific reason for that?

docs/extensions/custom-project-roles.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+The project controller will, for every custom role, create a `ClusterRole` with name `name: gardener.cloud:custom:project:<projectName>:<roleName>`, i.e., for above example: `name: gardener.cloud:custom:project:dev:foo`.
```
timebertt: Can you also add a sentence regarding the length restriction of custom role names?

pkg/controllermanager/controller/project/project_control_reconcile.go
```diff
@@ -120,31 +126,56 @@ func (c *defaultControl) reconcile(project *gardencorev1beta1.Project, projectLo
...
+				customRoleToSubjects[role] = append(customRoleToSubjects[role], member.Subject)
+				customRoles.Insert(strings.TrimPrefix(role, gardencorev1beta1.ProjectMemberCustomPrefix))
+			}
 		}
 	}
 
+	for _, name := range customRoles.List() {
+		customs = append(customs, map[string]interface{}{
+			"name":     name,
+			"subjects": customRoleToSubjects[name],
```
timebertt: This won't work, `subjects` will always be empty, because you are using the custom prefix in the keys in `customRoleToSubjects`, but in `customRoles` the names won't have the prefix. So `customRoleToSubjects[name]` will always be `nil`.

pkg/controllermanager/controller/project/project_control_reconcile.go
```diff
@@ -216,3 +247,48 @@ func (c *defaultControl) reconcileNamespaceForProject(project *gardencorev1beta1
 
 	return namespace, nil
 }
+
+// CustomProjectRole is a constant for a label value.
+const CustomProjectRole = "custom-project-role"
```
timebertt: Can we put this const into `pkg/apis/core/{v1alpha1,v1beta1}/constants/types_constants.go` or is there a reason, why it is placed here?

pkg/controllermanager/controller/project/project_control_reconcile.go
```diff
@@ -216,3 +247,48 @@ func (c *defaultControl) reconcileNamespaceForProject(project *gardencorev1beta1
...
+		}); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func getCustomRoleNameFromRBAC(name string) string {
+	if i := strings.LastIndex(name, ":"); i != -1 {
```
timebertt: This means, custom roles may not contain `:`, right?
I think we should rather do something like `strings.TrimPrefix(name, "gardener.cloud:custom:project:<project-name>:")`

pkg/controllermanager/controller/project/project_control_reconcile_test.go
```diff
@@ -0,0 +1,151 @@
...
+	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
+	Describe("#deleteStaleCustomRoles", func() {
+		var (
+			namespace           = "namespace"
+			prefix              = "name: gardener.cloud:custom:project:name:"
```
timebertt:
```suggestion
			prefix              = "gardener.cloud:custom:project:name:"
```

timebertt (changes requested):
Nice PR, Mr. Rafael!
Here are my comments from the first deeper look. Didn't fully test it yet, because of a small bug in the project controller.

Still, my biggest remark is the `custom:` prefix. Let's see, if we can find a better one.

grolu:
Does it make sense to have an owner role after all? Currently, the owner has no specific rights and we also have the createBy field which references a user that also has no special privileges. From my point of view, the semantic of the owner is more a person which is responsible and a point of contact and therefore highlighted in the dashboard.

rfranzke:
@grolu I guess it makes sense to have at least one main contact per project that can be reached out to in case of questions/issues/whatnot.

> we also have the createBy field which references a user that also has no special privileges.

This is not necessarily a human user, it could also be a technical user. It depends on how the RBAC privileges in the Garden cluster are designed.

> From my point of view, the semantic of the owner is more a person which is responsible and a point of contact and therefore highlighted in the dashboard.

Yes, that's exactly how it was before (`.spec.owner`) and how it is no (`owner`) role. Today, it has no special privileges, but tomorrow it maybe has - who knows. We have not yet worked out an overall role concept for the Garden cluster, hence, I would do minimal changes here now. We cannot get rid of the owner now anyways due to backwards compatibility reasons.

pkg/apis/core/v1alpha1/conversions.go
rfranzke: > Why don't you use your helper function ProjectMemberHasRole() here?

I'd have to import `pkg/apis/core/v1alpha1/helper` which itself imports `pkg/apis/core/v1alpha1`. The only way to use this function would be to move it to another package, though, I don't want to do this. Same reason for other comments.

> Also, why are you stepping out of the outer loop?
> This way the second loop for deleting the owner role from other members will not be executed...

Indeed, thanks, nice catch. fixed.

pkg/apis/core/v1alpha1/conversions.go
rfranzke: At this point in time the member is not the correct owner. However, if it has `.spec.role=owner` set (maybe because it was the owner before) then we have to remove it now. We replace it with the first role in the `.spec.roles` list.

pkg/apis/core/validation/project.go
rfranzke: Then I can't properly show which role is forbidden to be used when returning the error.

pkg/apis/core/validation/project.go
rfranzke: See https://github.com/gardener/gardener/pull/1948#issuecomment-587095890 - I changed it to 20 after renaming `custom` to `extension` and checking that we already validate the length of the project names to 10.

pkg/apis/core/validation/project_test.go
```diff
@@ -134,7 +134,71 @@ var _ = Describe("Project Validation Tests", func() {
...
+
+			errorList := ValidateProject(project)
+
+			Expect(errorList).To(ConsistOf(PointTo(MatchFields(IgnoreExtras, Fields{
+				"Type":  Equal(field.ErrorTypeTooLong),
+				"Field": Equal("spec.members[0].roles[0]"),
+			}))))
+		})
+
+		It("should prevent extension roles from being containing invalid characfters", func() {
```
ialidzhikov:
```suggestion
		It("should prevent extension roles from containing invalid characters", func() {
```

docs/extensions/project-roles.md
```diff
@@ -0,0 +1,39 @@
...
+  name: dev
+spec:
+  members:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: User
+    name: alice.doe@example.com
+    roles:
+    - "admin"
+    - "owner"
+    - "extension:foo"
```
ialidzhikov, line 28: Currently if I update the project spec from the example above to have the role `"extension:bar"` instead of the role `"extension:foo"` the ClusterRole with `gardener.cloud:extension:project:dev:foo`won't be deleted. Shouldn't we delete stale ClusterRoles which are no longer part of the roles.

pkg/apis/core/v1alpha1/conversions.go
```diff
@@ -193,3 +193,123 @@ func Convert_v1alpha1_ShootStatus_To_core_ShootStatus(in *ShootStatus, out *core
...
+
+	if owner := out.Owner; owner != nil {
+	outer:
+		for i, member := range out.Members {
+			if member.Name == owner.Name && member.APIGroup == owner.APIGroup && member.Kind == owner.Kind {
+				mustAddOwnerRole := true
+				for _, role := range member.Roles {
+					if role == core.ProjectMemberOwner {
+						mustAddOwnerRole = false
+						fmt.Println("jumping out")
```
ialidzhikov: I guess it is a left-over from debugging?

```suggestion

```

docs/extensions/project-roles.md
rfranzke: Why are you saying this? I cannot reproduce it. Can you provide steps? What I did:
1. Add `extension:foo` role to a user
1. `k get clusterrole | grep exten` shows `gardener.cloud:extension:project:dev:foo` only
1. Update `extension:foo` to `extension:bar`
1. `k get clusterrole | grep exten` shows `gardener.cloud:extension:project:dev:bar` only

ialidzhikov: Sorry, my mistake. You are right.

pkg/apis/core/helper/helpers.go
```diff
@@ -146,3 +146,13 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
 	}
 	return nil
 }
+
+// ProjectMemberHasRole returns true if the given project member has the given role.
+func ProjectMemberHasRole(member core.ProjectMember, role string) bool {
```
ialidzhikov: Are `ProjectMemberHasRole` funcs currently used?

pkg/apis/core/validation/project.go
```diff
@@ -114,6 +132,57 @@ func ValidateSubject(subject rbacv1.Subject, fldPath *field.Path) field.ErrorLis
...
+			// the extension role name will be used as part of a ClusterRole name
+			if errs := path.IsValidPathSegmentName(extensionRoleName); len(errs) > 0 {
+				var err string
+				for i, e := range errs {
+					err += e
+					if i < len(err)-1 {
+						err += ", "
+					}
+				}
+				allErrs = append(allErrs, field.Invalid(rolesPath, role, err))
```
ialidzhikov:
```suggestion
				allErrs = append(allErrs, field.Invalid(rolesPath, role, strings.Join(errs, ", ")))
```

ialidzhikov (commented):
Nits

pkg/apis/core/helper/helpers.go
rfranzke: Indeed, no, I've deleted them. Thanks.

ialidzhikov (dismissed):
/lgtm

pkg/apis/core/v1alpha1/conversions.go
```diff
@@ -193,3 +193,122 @@ func Convert_v1alpha1_ShootStatus_To_core_ShootStatus(in *ShootStatus, out *core
...
+func Convert_v1alpha1_ProjectSpec_To_core_ProjectSpec(in *ProjectSpec, out *core.ProjectSpec, s conversion.Scope) error {
+	if err := autoConvert_v1alpha1_ProjectSpec_To_core_ProjectSpec(in, out, s); err != nil {
+		return err
+	}
+
+	if owner := out.Owner; owner != nil {
+	outer:
+		for i, member := range out.Members {
+			if member.Name == owner.Name && member.APIGroup == owner.APIGroup && member.Kind == owner.Kind {
+				mustAddOwnerRole := true
```
timuthy: `mustAddOwnerRole` is not required since you early exit the loop in 210 in case the owner role was found.

pkg/apis/core/v1alpha1/conversions.go
```diff
@@ -193,3 +193,122 @@ func Convert_v1alpha1_ShootStatus_To_core_ShootStatus(in *ShootStatus, out *core
...
+func Convert_core_ProjectSpec_To_v1alpha1_ProjectSpec(in *core.ProjectSpec, out *ProjectSpec, s conversion.Scope) error {
+	if err := autoConvert_core_ProjectSpec_To_v1alpha1_ProjectSpec(in, out, s); err != nil {
+		return err
+	}
+
+	if owner := out.Owner; owner != nil {
+	outer:
+		for i, member := range out.Members {
+			if member.Name == owner.Name && member.APIGroup == owner.APIGroup && member.Kind == owner.Kind {
+				mustAddOwnerRole := true
```
timuthy: Same as previous comment.

pkg/apis/core/v1alpha1/types_project.go
```diff
@@ -61,6 +61,7 @@ type ProjectSpec struct {
 	// Owner is a subject representing a user name, an email address, or any other identifier of a user owning
 	// the project.
 	// +optional
+	// TODO: Remove this field in favor of the `owner` role in `v1`.
```
timuthy, line 67: Can you add a deprecation note as well?

timuthy (changes requested):
Some nits. LGTM otherwise.

timebertt:
I will also take another look this afternoon. Sorry for the delay.

pkg/apis/core/v1alpha1/conversions.go
timebertt: Ah, got it 👍

pkg/apis/core/v1beta1/conversions.go
```diff
@@ -39,3 +40,122 @@ func addConversionFuncs(scheme *runtime.Scheme) error {
...
 	return scheme.AddConversionFuncs()
 }
+
+func Convert_v1beta1_ProjectSpec_To_core_ProjectSpec(in *ProjectSpec, out *core.ProjectSpec, s conversion.Scope) error {
+	if err := autoConvert_v1beta1_ProjectSpec_To_core_ProjectSpec(in, out, s); err != nil {
+		return err
+	}
+
+	if owner := out.Owner; owner != nil {
+	outer:
```
timebertt, line 50: Same here regarding the `outer:` loop simplification

pkg/apis/core/v1beta1/conversions.go
```diff
@@ -39,3 +40,122 @@ func addConversionFuncs(scheme *runtime.Scheme) error {
...
+	return nil
+}
+
+func Convert_core_ProjectSpec_To_v1beta1_ProjectSpec(in *core.ProjectSpec, out *ProjectSpec, s conversion.Scope) error {
+	if err := autoConvert_core_ProjectSpec_To_v1beta1_ProjectSpec(in, out, s); err != nil {
+		return err
+	}
+
+	if owner := out.Owner; owner != nil {
+	outer:
```
timebertt, line 81: Same here regarding the `outer:` loop simplification

pkg/apis/core/validation/project.go
timebertt: Good point.

pkg/apis/core/v1alpha1/conversions.go
timebertt: > Also, why are you stepping out of the outer loop?
This way the second loop for deleting the owner role from other members will not be executed...

It seems to me, that you are still breaking out of the `outer:` loop. Is this by intention?

timebertt (changes requested):
Some nits left, LGTM otherwise

pkg/apis/core/v1alpha1/conversions.go
rfranzke: Indeed, nice catch, thank you. Removed.

pkg/apis/core/v1alpha1/types_project.go
rfranzke: The field is not deprecated yet, hence, I didn't add a deprecation note. Instead, I described that it will be removed in the next version so that people that are using the `owner` field now are aware.

pkg/apis/core/v1alpha1/conversions.go
rfranzke: It was by intention, yes. However, I simplified the members loop now to make it better readable.

timuthy (approved):
/lgtm

PR #1949, opened 2020-02-17
Fix early exit in chart applier

### LINKED ISSUE #1939: Seed bootstrapping fails

timebertt:
Previously, the chart applier did an early exit on error if one of the manifests could not be applied (for example because of a `KindMismatch`). Consequently all following manifests would not be applied, even if they were correct.
With this PR, the chart applier instead tries to apply all given manifests and returns a combined error for all failed manifests if there are any.

Fixes #1939

For example, when a new Seed is bootstrapped and one manifest cannot be applied (as described in #1939), the Seed condition will look similar to this:
```yaml
status:
  conditions:
  - lastTransitionTime: "2020-02-18T11:13:18Z"
    lastUpdateTime: "2020-02-18T11:13:18Z"
    message: 'failed to apply manifests: 1 error occurred: could not apply object
      of kind "VerticalPodAutoscaler" "garden/kube-state-metrics-vpa": no matches
      for kind "VerticalPodAutoscaler" in version "autoscaling.k8s.io/v1beta2"'
    reason: BootstrappingFailed
    status: "False"
    type: Bootstrapped
```
Though, the rest of the chart is still applied and the bootstrapping will succeed on the second try.

**Release note**:

```improvement operator
The Chart Applier now returns a combined error in case a rendered Manifest cannot be applied instead of doing an early exit on error.
```

### DIFF
--- a/pkg/client/kubernetes/apply.go
+++ b/pkg/client/kubernetes/apply.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"io"
 
+	utilerrors "github.com/gardener/gardener/pkg/utils/errors"
+	"github.com/hashicorp/go-multierror"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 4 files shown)


rfranzke (commented):
Nice, thank you!
/lgtm

rfranzke (dismissed):

pkg/client/kubernetes/apply.go
```diff
@@ -188,44 +188,64 @@ func (c *Applier) mergeObjects(newObj, oldObj *unstructured.Unstructured, mergeF
...
+			errList = append(errList, fmt.Errorf("could not read object: %+v", err))
+			continue
 		}
 		if obj == nil {
 			continue
 		}
 
 		if err := c.applyObject(ctx, obj, options); err != nil {
-			return err
+			errList = append(errList, fmt.Errorf("could not apply object of kind '%s' '%s/%s': %+v", obj.GetKind(), obj.GetNamespace(), obj.GetName(), err))
```
timuthy: Nit: Use `multierror.Append(errs, err)` instead of error slice.

timebertt: Nice suggestion, done :)
PTAL

pkg/client/kubernetes/apply.go
```diff
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/gardener/gardener/pkg/utils/errors"
```
rfranzke:
```suggestion
	utilerrors "github.com/gardener/gardener/pkg/utils/errors"
```
to not accidentally mix up with the core `errors` package

pkg/utils/errors/multierror.go
```diff
@@ -0,0 +1,41 @@
...
+// multierror.Error instances. The error string is prefixed with <prefix>, all errors are concatenated at the end.
+// This is similar to multierror.ListFormatFunc but does not use any escape sequences, which will look weird in
+// the status of Kubernetes objects or controller logs.
+func NewErrorFormatFuncWithPrefix(prefix string) multierror.ErrorFormatFunc {
+	return func(es []error) string {
+		if len(es) == 1 {
+			return fmt.Sprintf("%s: 1 error occurred: %s", prefix, es[0])
+		}
+
+		errStrings := make([]string, len(es))
```
rfranzke: What's the purpose/benefit of copying the errors over into `[]string` in order to iterate again over it when calling `strings.Join()`?
You are always iterating, can't you just construct the error message and return it?

Also, a small unit test could be added.

timebertt: Sorry, leftover.
Also added a small unit test, PTAL :)

rfranzke (approved):
/lgtm

timuthy (approved):

PR #1950, opened 2020-02-17
Clean up .spec.purpose defaulting from annotation

ialidzhikov:
- Clean up `.spec.purpose` defaulting from annotation
- Clean up deletion of legacy `gardener-external-admission-controller` resources during seed bootstrap.

Fixes #

**Release note**:

```improvement user
`gardener-apiserver` does no longer use the value of `garden.sapcloud.io/purpose` or `gardener.cloud/purpose` when `.spec.purpose` is not specified.
```


vpnachev (approved):
/aprove

PR #1952, opened 2020-02-19
Split kube-state-metrics charts into separate manifests

ialidzhikov:
According to https://helm.sh/docs/topics/chart_best_practices/templates/.:
> - Each resource definition should be in its own template file.

Fixes #

**Release note**:

```improvement operator
NONE
```


amshuman-kr (dismissed):
LGTM

But I copied this from the [seed-monitoring](https://github.com/gardener/gardener/blob/master/charts/seed-monitoring/charts/core/charts/kube-state-metrics-seed/templates/kube-state-metrics.yaml). So this needs to be fixed there as well.

ialidzhikov:
> LGTM
> 
> But I copied this from the [seed-monitoring](https://github.com/gardener/gardener/blob/master/charts/seed-monitoring/charts/core/charts/kube-state-metrics-seed/templates/kube-state-metrics.yaml). So this needs to be fixed there as well.

Adapted also in `seed-monitoring` chart.

rfranzke (approved):

PR #1959, opened 2020-02-19
Support multiple DNS providers

### LINKED ISSUE #1572: Support multiple DNS providers
With the new `core.gardener.cloud/v1alpha1.Shoot` API we are supporting to specify multiple DNS providers: https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml#L192-L205

However, our current implementation only deploys exactly one provider: https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/dns.go

We should fix our implementation to properly support the API.

timuthy:
This PR implements the support for multiple DNS providers.

A new (optional) field was added to the `DNSProvider` struct which identifies a given provider as the one managing the shoot domain. This handling is a best effort implementation to avoid removing such a provider by the end user.

```
dns:
  providers:    
  - secretName: custom-dns
    type: google-clouddns
    primary: true
```

Fixes #1572 

Some migration logic is necessary to fill out the newly introduced field. I'll remove this again in on of the upcoming versions.

**Release note**:

```noteworthy user
Gardener now supports the configuration of multiple DNS providers in the shoot.
```
```noteworthy operator
Gardener now supports the configuration of multiple DNS providers in the shoot. These providers can especially be used in combination with the `shoot-dns-service` (https://github.com/gardener/gardener-extension-shoot-dns-service) in your Garden environment.
```

### DIFF
--- a/hack/api-reference/core.md
+++ b/hack/api-reference/core.md
@@ -2920,6 +2920,18 @@ DNSIncludeExclude
 </tr>
 <tr>
 <td>
+<code>primary</code></br>
+<em>
+bool
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>Primary indicates that this DNSProvider is used for shoot related domains.</p>
+</td>
+</tr>
+<tr>
+<td>
 <code>secretName</code></br>
 <em>
 string
@@ -7296,5 +7308,5 @@ KubeletConfig
 <hr/>
 <p><em>
 Generated with <code>gen-crd-api-reference-docs</code>
-on git commit <code>2bdda9f61</code>.
+on git commit <code>8f4d7313c</code>.
 </em></p>
--- a/pkg/apis/core/types_shoot.go
+++ b/pkg/apis/core/types_shoot.go
@@ -176,6 +176,8 @@ type DNS struct {
 type DNSProvider struct {
 	// Domains contains information about which domains shall be included/excluded for this provider.
 	Domains *DNSIncludeExclude
+	// Primary indicates that this DNSProvider is used for shoot related domains.
+	Primary *bool
 	// SecretName is a name of a secret containing credentials for the stated domain and the
 	// provider. When not specified, the Gardener will use the cloud provider credentials referenced
 	// by the Shoot and try to find respective credentials there. Specifying this field may override
--- a/pkg/operation/botanist/dns.go
+++ b/pkg/operation/botanist/dns.go
@@ -21,12 +21,16 @@ import (
 	"strings"
 	"time"
 
+	"github.com/gardener/gardener/pkg/apis/core"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	"github.com/gardener/gardener/pkg/operation/common"
+	"github.com/gardener/gardener/pkg/utils/flow"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	"github.com/gardener/gardener/pkg/utils/retry"
 
 	dnsv1alpha1 "github.com/gardener/external-dns-management/pkg/apis/dns/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -36,26 +40,27 @@ import (
 var dnsChartPath = filepath.Join(common.ChartPath, "seed-dns")
 
 const (
-	// DNSPurposeInternal is a constant for a DNS record used for the internal domain name.
-	DNSPurposeInternal = "internal"
-	// DNSPurposeExternal is a constant for a DNS record used for the external domain name.
-	DNSPurposeExternal = "external"
+	// DNSInternalName is a constant for a DNS resources used for the internal domain name.
+	DNSInternalName = "internal"
+	// DNSExternalName is a constant for a DNS resources used for the external domain name.
+	DNSExternalName = "external"
+	// DNSProviderRolePrimary is a constant for a DNS providers used to manage shoot domains.
+	DNSProviderRolePrimary = "primary-dns-provider"
+	// DNSProviderRoleAdditional is a constant for additionally managed DNS providers.
+	DNSProviderRoleAdditional = "managed-dns-provider"
 )
 
 // DeployInternalDomainDNSRecord deploys the DNS record for the internal cluster domain.
 func (b *Botanist) DeployInternalDomainDNSRecord(ctx context.Context) error {
-	if err := b.deployDNSProvider(ctx, DNSPurposeInternal, b.Garden.InternalDomain.Provider, b.Garden.InternalDomain.SecretData, []string{b.Shoot.InternalClusterDomain}, nil, b.Garden.InternalDomain.IncludeZones, b.Garden.InternalDomain.ExcludeZones); err != nil {
+	if err := b.deployDNSProvider(ctx, DNSInternalName, DNSProviderRolePrimary, b.Garden.InternalDomain.Provider, b.Garden.InternalDomain.SecretData, []string{b.Shoot.InternalClusterDomain}, nil, b.Garden.InternalDomain.IncludeZones, b.Garden.InternalDomain.ExcludeZones); err != nil {
 		return err
 	}
-	return b.deployDNSEntry(ctx, DNSPurposeInternal, common.GetAPIServerDomain(b.Shoot.InternalClusterDomain), b.APIServerAddress)
+	return b.deployDNSEntry(ctx, DNSInternalName, common.GetAPIServerDomain(b.Shoot.InternalClusterDomain), b.APIServerAddress)
 }
 
 // DestroyInternalDomainDNSRecord destroys the DNS record for the internal cluster domain.
 func (b *Botanist) DestroyInternalDomainDNSRecord(ctx context.Context) error {
-	if err := b.deleteDNSEntry(ctx, DNSPurposeInternal); err != nil {
-		return err
-	}
-	return b.deleteDNSProvider(ctx, DNSPurposeInternal)
+	return b.deleteDNSEntry(ctx, DNSInternalName)
 }
 
 // DeployExternalDomainDNSRecord deploys the DNS record for the external cluster domain.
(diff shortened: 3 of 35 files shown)


timuthy:
FYI: I'm double checking why `pkg/client/core/openapi/zz_generated.openapi.go` includes so many changes.

hack/api-reference/core.md
```diff
@@ -7274,9 +7286,5 @@ KubeletConfig
 <hr/>
 <p><em>
 Generated with <code>gen-crd-api-reference-docs</code>
-<<<<<<< HEAD
```
mvladev: ~~merge conflict~~ 

disregard it

mvladev (commented):
just a a small comment

mvladev:
As a side note have a look at https://github.com/mvladev/gardener/commit/e7215351b8d3e60cdc2fee2b8af3ca0317b25c6c @timuthy 

It's still WIP, but it's based on #1969 and the entire DNS component is separated in own package and it's completely isolated from the botanist, making it extremely easy to test - now It has 100% test coverage on the deploy, destroy, wait and wait for deletion operations.

For each `DNSProvider` and `DNSEntry` which is needed a separate instance is created.

timuthy:
As discussed with @mvladev we'll merge this PR first and then he'll incorporate the changes done here.

pkg/operation/botanist/dns.go
```diff
@@ -64,25 +69,119 @@ func (b *Botanist) DeployExternalDomainDNSRecord(ctx context.Context) error {
...
+				b.Logger.Infof("Skipping deployment of DNS provider[%d] since it specifies type %q", i, core.DNSUnmanaged)
+				return nil
+			}
+			secretName := p.SecretName
+			if secretName == nil {
+				return fmt.Errorf("dns provider[%d] doesn't specify a secretName", i)
+			}
+			secret := &corev1.Secret{}
+			if err := b.K8sGardenClient.Client().Get(ctx, kutil.Key(b.Shoot.Info.Namespace, *secretName), secret); err != nil {
+				return fmt.Errorf("could not get dns provider secret \"%s\": %+v", *secretName, err)
```
ialidzhikov:
```suggestion
				return fmt.Errorf("could not get dns provider secret %q: %+v", *secretName, err)
```

pkg/operation/shoot/shoot.go
```diff
@@ -339,14 +324,24 @@ func ConstructExternalDomain(ctx context.Context, client client.Client, shoot *g
...
 
-	case len(shoot.Spec.DNS.Providers) > 0 && shoot.Spec.DNS.Providers[0].SecretName == nil:
-		externalDomain.SecretData = shootSecret.Data
-		externalDomain.Provider = *shoot.Spec.DNS.Providers[0].Type
-		if domains := shoot.Spec.DNS.Providers[0].Domains; domains != nil {
+	case primaryProvider != nil:
+		if primaryProvider.SecretName != nil {
+			secret := &corev1.Secret{}
+			if err := client.Get(ctx, kutil.Key(shoot.Namespace, *primaryProvider.SecretName), secret); err != nil {
+				return nil, fmt.Errorf("could not get dns provider secret \"%s\": %+v", *shoot.Spec.DNS.Providers[0].SecretName, err)
```
ialidzhikov:
```suggestion
				return nil, fmt.Errorf("could not get dns provider secret %q: %+v", *shoot.Spec.DNS.Providers[0].SecretName, err)
```

ialidzhikov (commented):
Nits. Could you please also rebase?

timuthy:
@ialidzhikov thanks a lot for you review. PTAL.

ialidzhikov (dismissed):
/lgtm

pkg/apis/core/types_shoot.go
```diff
@@ -176,6 +176,8 @@ type DNS struct {
 type DNSProvider struct {
 	// Domains contains information about which domains shall be included/excluded for this provider.
 	Domains *DNSIncludeExclude
+	// Primary indicates that this DNSProvider is used for shoot related domains.
+	Primary *bool
```
mvladev, line 180: Would it make more sense to have this a not a boolean, but rather something like `Class *DNSProviderClass` with `type DNSProviderClass string`, so later on, we can have multiple options, such as `Primary`, `FailOver` etc...

pkg/operation/botanist/dns.go
```diff
@@ -64,25 +69,119 @@ func (b *Botanist) DeployExternalDomainDNSRecord(ctx context.Context) error {
...
+				return nil
+			}
+			secretName := p.SecretName
+			if secretName == nil {
+				return fmt.Errorf("dns provider[%d] doesn't specify a secretName", i)
+			}
+			secret := &corev1.Secret{}
+			if err := b.K8sGardenClient.Client().Get(ctx, kutil.Key(b.Shoot.Info.Namespace, *secretName), secret); err != nil {
+				return fmt.Errorf("could not get dns provider secret %q: %+v", *secretName, err)
+			}
```
mvladev, line 123: Can this entire thing moved into https://github.com/gardener/gardener/blob/b375d42f6006d863a2f5cf1ec835a46c23174ce1/pkg/operation/garden/types.go#L21-L26

With this fetching this would be need only once during a reconciliation flow.

Also https://github.com/gardener/gardener/blob/8fb9d96f4ed16e1fd4d3069cfc20b0282d7ef2b5/pkg/operation/garden/garden.go#L41-L63

pkg/apis/core/types_shoot.go
timuthy: Can make sense, yes.
Although I would rather go with something like `Purpose`. Any thoughts @mvladev @rfranzke?

pkg/operation/botanist/dns.go
timuthy: I'm not sure why you propose to add it to the `Garden` struct since it only holds information related to the whole environment and nothing specific to the shoot (only secrets in the `Garden` namespace are read).

Imo, additional DNS providers don't have to be part of the `Operation` struct either because the data is only needed in the deploy step or no other func of `Operation` needs to access that data as well. I know that this will change with your isolation plans but would still prefer to leave it like this for the moment.

pkg/apis/core/types_shoot.go
mvladev: `Purpose` is fine as well.

zanetworker: @mvladev I guess what you are hinting could be added as a policy field in the future? Something link: 

```go
// DNS holds information about the provider, the hosted zone id and the domain.
type DNS struct {
	// Domain is the external available domain of the Shoot cluster. This domain will be written into the
	// kubeconfig that is handed out to end-users.
	Domain *string
	// Providers is a list of DNS providers that shall be enabled for this shoot cluster. Only relevant if
	// not a default domain is used.
	Providers []DNSProvider
        // Policy poilicy type e.g., active-passive, fail-over, ... 
        Policy PolicyType 
}
```

To me primary here looks okay.

zanetworker: @mvladev I guess what you are hinting could be added as a policy field in the future? Something link: 

```go
// DNS holds information about the provider, the hosted zone id and the domain.
type DNS struct {
	// Domain is the external available domain of the Shoot cluster. This domain will be written into the
	// kubeconfig that is handed out to end-users.
	Domain *string
	// Providers is a list of DNS providers that shall be enabled for this shoot cluster. Only relevant if
	// not a default domain is used.
	Providers []DNSProvider
        // Policy poilicy type e.g., active-passive, fail-over, ... 
        Policy PolicyType 
}
```

To me primary here looks okay.

zanetworker: @mvladev I guess what you are hinting could be added as a policy field in the future for DNS configuration? Something like: 

```go
// DNS holds information about the provider, the hosted zone id and the domain.
type DNS struct {
	// Domain is the external available domain of the Shoot cluster. This domain will be written into the
	// kubeconfig that is handed out to end-users.
	Domain *string
	// Providers is a list of DNS providers that shall be enabled for this shoot cluster. Only relevant if
	// not a default domain is used.
	Providers []DNSProvider
        // Policy poilicy type e.g., active-passive, fail-over, ... 
        Policy PolicyType 
}
```

To me primary here looks okay.

pkg/operation/botanist/dns.go
```diff
@@ -40,22 +44,23 @@ const (
...
+	// DNSProviderRolePrimary is a constant for a DNS providers used to manage shoot domains.
+	DNSProviderRolePrimary = "primary-dns-provider"
+	// DNSProviderRoleAdditional is a constant for additionally managed DNS providers.
+	DNSProviderRoleAdditional = "managed-dns-provider"
 )
 
 // DeployInternalDomainDNSRecord deploys the DNS record for the internal cluster domain.
 func (b *Botanist) DeployInternalDomainDNSRecord(ctx context.Context) error {
-	if err := b.deployDNSProvider(ctx, DNSPurposeInternal, b.Garden.InternalDomain.Provider, b.Garden.InternalDomain.SecretData, []string{b.Shoot.InternalClusterDomain}, nil, b.Garden.InternalDomain.IncludeZones, b.Garden.InternalDomain.ExcludeZones); err != nil {
+	if err := b.deployDNSProvider(ctx, DNSPurposeInternal, DNSProviderRolePrimary, b.Garden.InternalDomain.Provider, b.Garden.InternalDomain.SecretData, []string{b.Shoot.InternalClusterDomain}, nil, b.Garden.InternalDomain.IncludeZones, b.Garden.InternalDomain.ExcludeZones); err != nil {
```
zanetworker: I know its not part of this PR, but passing purpose to a name field looks a bit confusing. Could we do something link 

```go
const DNSInternalProviderName = DNSPurposeInternal 
``` 

and use that instead?

pkg/apis/core/types_shoot.go
timuthy: When I think about that @zanetworker then `primary` totally makes sense, i.e. it will always point out the provider that is currently responsible for the shoot domain. Even though this would rather be status than a spec.

timuthy: I reiterated over the proposal of using `Purpose *string` instead of `Primary *bool` and think sticking to `Primary` has the following benefits:

1. A shoot will always have only one primary provider, no matter if several other ones are available for fail over, etc.. We can introduce another field like `Policy` (as @zanetworker suggested) if we need to, but the `Primary` purpose will still be valid.
2. `Primary` is easy to use as opposed to an optional string field whose allowed values must be described in a documentation first.

timuthy:
Thanks for your reviews! PTAL.

rfranzke (approved):

PR #1962, opened 2020-02-20
fix ComputeClusterIP calculation

mvladev:

With the current implementation calling `ComputeClusterIP("192.168.64.16/28", 10)` would return `192.168.64.10` which is outside of the netmask.

```console
sipcalc 192.168.64.16/28

-[ipv4 : 192.168.64.16/28] - 0

[CIDR]
Host address            - 192.168.64.16
Host address (decimal)  - 3232251920
Host address (hex)      - C0A84010
Network address         - 192.168.64.16
Network mask            - 255.255.255.240
Network mask (bits)     - 28
Network mask (hex)      - FFFFFFF0
Broadcast address       - 192.168.64.31
Cisco wildcard          - 0.0.0.15
Addresses in network    - 16
Network range           - 192.168.64.16 - 192.168.64.31
Usable range            - 192.168.64.17 - 192.168.64.30
```

Fixes # n/a

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/addons.go
+++ b/pkg/operation/botanist/addons.go
@@ -183,11 +183,11 @@ func (b *Botanist) generateCoreAddonsChart() (*chartrenderer.RenderedChart, erro
 		vpnTLSAuthSecret = b.Secrets["vpn-seed-tlsauth"]
 		global           = map[string]interface{}{
 			"kubernetesVersion": b.Shoot.Info.Spec.Kubernetes.Version,
-			"podNetwork":        b.Shoot.GetPodNetwork(),
+			"podNetwork":        b.Shoot.Networks.Pods.String(),
 		}
 		coreDNSConfig = map[string]interface{}{
 			"service": map[string]interface{}{
-				"clusterDNS": common.ComputeClusterIP(b.Shoot.GetServiceNetwork(), 10),
+				"clusterDNS": b.Shoot.Networks.CoreDNS.String(),
 				// TODO: resolve conformance test issue before changing:
 				// https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/dns.go#L44
 				"domain": map[string]interface{}{
@@ -215,8 +215,8 @@ func (b *Botanist) generateCoreAddonsChart() (*chartrenderer.RenderedChart, erro
 			},
 		}
 		vpnShootConfig = map[string]interface{}{
-			"podNetwork":     b.Shoot.GetPodNetwork(),
-			"serviceNetwork": b.Shoot.GetServiceNetwork(),
+			"podNetwork":     b.Shoot.Networks.Pods.String(),
+			"serviceNetwork": b.Shoot.Networks.Services.String(),
 			"tlsAuth":        vpnTLSAuthSecret.Data["vpn.tlsauth"],
 			"podAnnotations": map[string]interface{}{
 				"checksum/secret-vpn-shoot": b.CheckSums["vpn-shoot"],
@@ -229,8 +229,8 @@ func (b *Botanist) generateCoreAddonsChart() (*chartrenderer.RenderedChart, erro
 			"provider":          b.Shoot.Info.Spec.Provider.Type,
 			"region":            b.Shoot.Info.Spec.Region,
 			"kubernetesVersion": b.Shoot.Info.Spec.Kubernetes.Version,
-			"podNetwork":        b.Shoot.GetPodNetwork(),
-			"serviceNetwork":    b.Shoot.GetServiceNetwork(),
+			"podNetwork":        b.Shoot.Networks.Pods.String(),
+			"serviceNetwork":    b.Shoot.Networks.Services.String(),
 			"maintenanceBegin":  b.Shoot.Info.Spec.Maintenance.TimeWindow.Begin,
 			"maintenanceEnd":    b.Shoot.Info.Spec.Maintenance.TimeWindow.End,
 		}
--- a/pkg/operation/common/utils.go
+++ b/pkg/operation/common/utils.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/big"
 	"net"
 	"regexp"
 	"sort"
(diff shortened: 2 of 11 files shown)


timuthy:
Since you mention the use case slipping under the the IP range, what about those hitting the upper boundary?
`ComputeClusterIP("192.168.64.16/28", 16)` --> 192.168.64.32 (for 192.168.64.16/28)

mvladev:
> ComputeClusterIP("192.168.64.16/28", 16) 

The method signature have to be changed to return error. I'll do it.

mvladev:
I've updated the PR. 

I'll slowly try to phase out all `strings` in favor of `net.IP` or `net.IPNet` where applicable.

pkg/operation/common/utils.go
```diff
@@ -63,13 +64,21 @@ func GetSecretKeysWithPrefix(kind string, m map[string]*corev1.Secret) []string
...
+// For example, <subnet> = 100.64.0.0/11 and <offset> = 10 the result would be 100.64.0.10
+func ComputeOffsetIP(subnet *net.IPNet, offset int64) (net.IP, error) {
+	if subnet == nil {
+		return nil, fmt.Errorf("subnet is nil")
+	}
+
+	base := big.NewInt(0).SetBytes(subnet.IP.To4())
+
+	ip := net.IP(big.NewInt(0).Add(base, big.NewInt(offset)).Bytes())
+	if !subnet.Contains(ip) {
```
timuthy, line 85: What do you think about to additionally check against the network and broadcast ip of the network?

pkg/operation/operation.go
```diff
@@ -116,13 +116,51 @@ func newOperation(
 	}
 
 	if shoot != nil {
-		shootObj, err := shootpkg.New(k8sGardenClient, k8sGardenCoreInformers, shoot, gardenObj.Project.Name, disableDNS, gardenObj.InternalDomain, gardenObj.DefaultDomains)
+
+		binding, err := k8sGardenCoreInformers.SecretBindings().Lister().SecretBindings(shoot.Namespace).Get(shoot.Spec.SecretBindingName)
```
timuthy: I think the responsibility of `shootpkg.New` was clearer before this change because the function almost calculated everything required for the shoot object. Now most of the complexity is outside while there is still a certain portion in `shootpkg.New`, i.e. I don't see a clear separation.
WDYT or what was the motivation?

timuthy (changes requested):
Nice! I just have two more concerns.

pkg/operation/operation.go
mvladev: I wanted to make it more testable, I'm slowly start to refactor big portions of the code and to make it more isolated from each other. 
With this change everything which the `Shoot` needs is passed and it doesn't actually need to perform any network calls to populate it's fields.

mvladev: For example `k8sGardenClient`, `k8sGardenCoreInformers` were only there to so the Shoot can fetch the `CloudProfile`, get extensions and so - this data can (and probably should) be passed from the top.

timuthy: > I wanted to make it more testable, I'm slowly start to refactor big portions of the code and to make it more isolated from each other.

That's nice and I understand that but on the other hand this seems like only shifting complexity (here from `shootpkg.New` to `operation.New`) at the moment. Actually, `shootpkg.New` is already testable but of course mock and lister preparation is needed for that.

Instead of a better testability, I rather see a separation of concern being broken here. 
Maybe it'd be more obvious to me if you had more refactoring in this area.

timuthy: \+ Testing `shootpkg.New` as it is today will enable a lower risk refactoring in the future.

mvladev: Yes, I'm moving complexity (or lets say dependency injection) to higher in the stack. With this PR, no mocks are needed are needed.

rfranzke: Instead of moving the stuff to the `operation.New`, WDYT about moving it to a separate function in the `shootpkg`? You could make `New` as you have it now (just pass the already read/computed objects), and in another function `NewXYZ` function you read/compute them and call `New` afterwards?

mvladev: > Instead of moving the stuff to the operation.New, WDYT about moving it to a separate function in the shootpkg?

Why? The `Shoot` has a specific dependencies and it should have only one way to construct it. Having multiple `NewXYZ` is pointless..

mvladev: Also, I haven't looked at it fully, but I have suspicion that higher in the stack `cloudprofile`, `seedName` and other similar resources are already available and simply not passed to here

rfranzke: > it should have only one way to construct it. Having multiple NewXYZ is pointless..

Can you elaborate? I don't get the rationale behind the "should" and "pointless" words.

What I'm proposing is just to bundle the read/compute stuff into  "helper function" (that is called `NewXYZ`) instead of doing it in the `operation.New` function. It will not only make the `operation.New` function easier, it will also help others that want to instantiate the `Shoot` object by enabling reuse. ?

timuthy: > Also, I haven't looked at it fully, but I have suspicion that higher in the stack cloudprofile, seedName and other similar resources are already available and simply not passed to here

+1 but this is why I think it'd make sense to include this change in a "broader" refactoring where we can see that it makes sense to move certain complexity/DIs to other places.

mvladev:
Updated the PR with broadcast check, IPv6 support. All other unrelated changes are removed.

timuthy (dismissed):
/lgtm

pkg/operation/botanist/addons.go
```diff
@@ -183,11 +183,11 @@ func (b *Botanist) generateCoreAddonsChart() (*chartrenderer.RenderedChart, erro
 		vpnTLSAuthSecret = b.Secrets["vpn-seed-tlsauth"]
 		global           = map[string]interface{}{
 			"kubernetesVersion": b.Shoot.Info.Spec.Kubernetes.Version,
-			"podNetwork":        b.Shoot.GetPodNetwork(),
+			"podNetwork":        b.Shoot.Networks.Pods.String(),
 		}
 		coreDNSConfig = map[string]interface{}{
 			"service": map[string]interface{}{
-				"clusterDNS": common.ComputeClusterIP(b.Shoot.GetServiceNetwork(), 10),
+				"clusterDNS": b.Shoot.Networks.APIServer.String(),
```
```suggestion
				"clusterDNS": b.Shoot.Networks.CoreDNS.String(),
```

rfranzke (changes requested):
Thanks, found one mistake, otherwise /lgtm

mvladev:
Updated the PR with correct CoreDNS entry (my cluster managed to boot even with this configuration...)

rfranzke (approved):
/lgtm

PR #1966, opened 2020-02-22
fix various lint problems

mvladev:

- capitalized return error strings
- `if someBool != true { ... }`
- `strings.EqualFold` usage
- dead code

etc

Fixes # n/a

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/cmd/gardener-scheduler/main.go
+++ b/cmd/gardener-scheduler/main.go
@@ -18,24 +18,23 @@ import (
 	"fmt"
 	"os"
 	"runtime"
-	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
 
 	"github.com/gardener/gardener/cmd/gardener-scheduler/app"
 	"github.com/gardener/gardener/cmd/utils"
 	"github.com/gardener/gardener/pkg/controllermanager/features"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 )
-
-func init() {
-	features.RegisterFeatureGates()
-}
 
 func main() {
 	if len(os.Getenv("GOMAXPROCS")) == 0 {
 		runtime.GOMAXPROCS(runtime.NumCPU())
 	}
 
+	features.RegisterFeatureGates()
+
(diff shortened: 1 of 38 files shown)


pkg/utils/timewindow_test.go
```diff
@@ -54,11 +49,14 @@ var _ = Describe("utils", func() {
 
 		Describe("#ParseMaintenanceTime", func() {
 			It("should return the correctly parsed maintenance time", func() {
+				const (
+					cetOffset = 2
+					hour      = 16
+					minute    = 5
+					second    = 54
+				)
 				var (
```
timuthy: Remove `var` block.

cmd/gardener-scheduler/main.go
```diff
@@ -18,24 +18,23 @@ import (
...
-func init() {
-	features.RegisterFeatureGates()
-}
-
 func main() {
 	if len(os.Getenv("GOMAXPROCS")) == 0 {
 		runtime.GOMAXPROCS(runtime.NumCPU())
 	}
 
+	features.RegisterFeatureGates()
```
timuthy, line 33: Why do you move this from `init()` to here and why only for the scheduler?

timuthy (changes requested):

pkg/utils/timewindow_test.go
mvladev: done

cmd/gardener-scheduler/main.go
mvladev: Why does it needs `init()` in the first place when it was in the `main` package?

> why only for the scheduler?

I was fixing a deprecation warning (package change). Didn't look into other components.

timuthy: I'm not opposing in general, just voting for doing it in a consistent way.
Do you mind adjusting the other places as well?

mvladev: Already done

timuthy (approved):

PR #1958, opened 2020-02-19
More metrics for seed workload.

amshuman-kr:
**What this PR does / why we need it**:
Some more aggregate metrics to monitor the seed workload with more focus on the health of the seed workload. Includes container restart counts, ready replicas for deployment/statefulsets, pod counts by phase and state etc.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
Follow up to #1863.

**Release note**:

```improvement operator
Some more aggregate metrics to monitor the seed workload with more focus on the health of the seed workload. Includes container restart counts, ready replicas for deployment/statefulsets, pod counts by phase and state etc.
```

PR #1956, opened 2020-02-19
Update hvpa-controller  to v0.2.1

ggaurav10:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement operator
Consider HPA scale out to be limited in case, overrideScaleUpStabilization is set in the status and hpa weight is 0 so that full VPA recommendation is immediately applied.
```

PR #1968, opened 2020-02-24
Disable memory metric for horizontal scaling of shoots' kube-apiserver in HVPA

ggaurav10:
This PR enabled memory metric for horizontal scaling only for seeds, and disables it for the shoot kube-apiservers.
Since kube-apiserver caches objects in the memory, this cache gets duplicated among its replicas. This throws off HPA's estimations of memory requirement for the kube-apiserver because of which we see multiple replicas of kube-apiserver even in shoots with no activity. These replicas have very low (baseline) CPU usage, however, high memory usage.
Currently the memory metric is disabled only for shoots. Depending on our observation of its efficacy and stability, we plan to make the same changes for shooted seeds also.

Fixes #

**Release note**:

```improvement operator
Disable memory metric for horizontal scaling of shoots' kube-apiserver in HVPA. This would help in scaling down kube-apiserver horizontally when the load is low since the affect of memory caching is eliminated. This helps in decreasing the cost of seeds.
```

### DIFF
--- a/charts/seed-controlplane/charts/kube-apiserver/values.yaml
+++ b/charts/seed-controlplane/charts/kube-apiserver/values.yaml
@@ -87,8 +87,11 @@ apiServerResources:
 
 maxReplicas: 1
 minReplicas: 1
-targetAverageUtilizationCpu: 80
-targetAverageUtilizationMemory: 80
+
+hpa:
+  targetAverageUtilizationCpu: 80
+  memoryMetricForHpaEnabled: false
+  targetAverageUtilizationMemory: 80
 
 auditConfig:
   auditPolicy: ""
--- a/charts/seed-controlplane/charts/kube-apiserver/templates/kube-apiserver-hpa.yaml
+++ b/charts/seed-controlplane/charts/kube-apiserver/templates/kube-apiserver-hpa.yaml
@@ -17,10 +17,10 @@ spec:
   - type: Resource
     resource:
       name: cpu
-      targetAverageUtilization: {{ .Values.targetAverageUtilizationCpu }}
+      targetAverageUtilization: {{ .Values.hpa.targetAverageUtilizationCpu }}
   - type: Resource
     resource:
       name: memory
-      targetAverageUtilization: {{ .Values.targetAverageUtilizationMemory }}
+      targetAverageUtilization: {{ .Values.hpa.targetAverageUtilizationMemory }}
 {{- end }}
 {{ end }}
--- a/charts/seed-controlplane/charts/kube-apiserver/templates/kube-apiserver-hvpa.yaml
+++ b/charts/seed-controlplane/charts/kube-apiserver/templates/kube-apiserver-hvpa.yaml
@@ -31,13 +31,15 @@ spec:
         maxReplicas: {{ .Values.maxReplicas }}
         minReplicas: {{ .Values.minReplicas }}
         metrics:
+{{- if .Values.hpa.memoryMetricForHpaEnabled }}
         - resource:
             name: memory
-            targetAverageUtilization: {{ .Values.targetAverageUtilizationMemory }}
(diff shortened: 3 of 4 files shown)


amshuman-kr (commented):
@ggaurav10 The PR looks good. Can you please add a bit more details about the memory cost/utilization improvement in the release notes as well?

Also, it would be good to release this change along with #1958.

amshuman-kr:
> Also, it would be good to release this change along with #1958.

Should we add a sentence about the possibility of a momentary OOMKill in case of low CPU and high memory utilisation in the `kube-apiserver` but should be managed by HVPA especially after #1956. Also it should be monitorable via the metrics introduced in #1958?

charts/seed-controlplane/charts/kube-apiserver/values.yaml
```diff
@@ -88,6 +88,7 @@ apiServerResources:
 maxReplicas: 1
 minReplicas: 1
 targetAverageUtilizationCpu: 80
+memoryMetricForHpaEnabled: false
```
ialidzhikov: Can we have the values organised for example such as:
```yaml
hvpa:
  enabled: false
  metrics:
    memory:
      enabled: false
      targetAverageUtilization: 80
    cpu:
      enabled: true
      targetAverageUtilization: 80
```

ggaurav10: when hvpa is not enabled, we delpoy HPA instead. The HPA template uses the same `values`. It might not be a good idea to have them under `hvpa` or duplicated for HPA as well. WDYT?

ialidzhikov: Wouldn't be cleaner to have them duplicated for `hvpa` and `hpa`?
You also currently introduce a value called `memoryMetricForHpaEnabled` which is only used in the hvpa template.

ggaurav10: `memoryMetricForHpaEnabled` works for HVPA since hvpa-controller observes any oomkill that might happen, and applies the VPA recommendations regardless of the vpaWeight provided in the hvpa spec. Since HPA alone does not have that ability, it might still be safer to scale horizontally using memory metric when using HPA alone.
With this change, we can expect some rare oomkills when using hvpa (as mentioned by @amshuman-kr's [comment](https://github.com/gardener/gardener/pull/1968#issuecomment-590205774) with a benefit of reduced costs of seeds.
If we see a stable behaviour, we plan to introduce this change for shooted seeds also, along with possibility of including this in only HPA based scaling.

amshuman-kr: Also, if we see the behaviour is not stable, the plan is to re-introduce memory-based scaling for HPA but with higher utilisation threshold (say, 120%) to still somehow optimise on cost still.

Eitherway is a imperfect solution to a problem which can only be truely solved when HVPA supports the scale sub-resource to [control HPA scaling](https://github.com/gardener/hvpa-controller/issues/7) so that we can scale vertically first and horizontally later.

ialidzhikov: I am only suggesting values restructuring (such as `hvpa.vpa.metrics.memory.enabled`) as at least for me checking values such as `targetAverageUtilizationCpu` or `memoryMetricForHpaEnabled` - I don't get in which template they are used.

ggaurav10: I understand your concern. i have added another [patch](https://github.com/gardener/gardener/pull/1968/commits/461352f126d3cdd93bd6da35576098965d489d78). Please let me know if this helpful.

rfranzke (dismissed):
/lgtm

timuthy:
@ggaurav10 can you squash the commits before we proceed with merging?

rfranzke (approved):

PR #1969, opened 2020-02-24
ChartApplier now takes interface{} as value

mvladev:

Reworked various parts of the public interface of `ChartApplier` to make it more composable. Extra methods `With...` are now removed and customization can be applied with implementing `ApplyOption` and`DeleteOption` interfaces.

Both `Delete` and `Apply` now can use `interface{}` instead of `map[string]interface{}` as values. With this change is now possible to pass a predefined struct such as

```go
type Foo struct {
	Baz string `json:"baz,omitempty"`
	Foo bool `json:"foo,omitempty"`
}
```

to it.

Fixes #

This entire change is also needed so we can start testing individual components e.g. for the DNS provider deployment/deletion

```go
type DNSProvider struct {
	Values         DNSProviderValues	
	ShootNamespace string
	*kubernetes.ChartApplier
}

type DNSProviderValues struct {
	Name       string            `json:"name,omitempty"`
	Purpose    string            `json:"purpose,omitempty"`
	Provider   string            `json:"provider,omitempty"`
	SecretData map[string][]byte `json:"secretData,omitempty"`
	Domains    IncludeExclude    `json:"domains,omitempty"`
	Zones      IncludeExclude    `json:"zones,omitempty"`
}
type IncludeExclude struct {
	Include []string `json:"include,omitempty"`
	Exclude []string `json:"exclude,omitempty"`
}

func (d *DNSProvider) Deploy(ctx context.Context) error {
	if err := d.Apply(ctx, filepath.Join(dnsChartPath, "provider"), d.Values.Name, kubernetes.Values(d.Values)); err != nil {
		return err
	}

	return waiter() // TODO...
}
```

With this, botanist can construct the `DNSProvider` when it starts working on a `Shoot`. 

**Release note**:

```improvement operator
NONE
```


mvladev:
bump @gardener/gardener-maintainers

pkg/client/kubernetes/apply.go
```diff
@@ -118,62 +117,58 @@ func (c *Applier) deleteObject(ctx context.Context, desired *unstructured.Unstru
...
-
-				for _, oldPort := range oldPorts {
-					op := oldPort.(map[string]interface{})
-					// np["port"] is of type float64 (due to Helm Tiller rendering) while op["port"] is of type int64.
-					// Equality can only be checked via their string representations.
-					if fmt.Sprintf("%v", np["port"]) == fmt.Sprintf("%v", op["port"]) {
-						if nodePort, ok := op["nodePort"]; ok {
-							np["nodePort"] = nodePort
-						}
+var DefaultApplierOptions = MergeFuncs{
```
zanetworker: Do you think it make sense to keep the name `DefaultApplierOptions` although it is 
 only restricted to MergeFunc? Maybe `DefaultMergeFuncs` would make more sense?

mvladev:
PR updated with rename `DefaultApplierOptions` -> `DefaultMergeFuncs`

mvladev:
rebased the PR to fix the merge conflicts

test/framework/gomega_utils.go
```diff
@@ -14,9 +14,50 @@
...
+// DeepDerivativeEqual is similar to DeepEqual except that unset fields in actual are
+// ignored (not compared). This allows us to focus on the fields that matter to
+// the semantic comparison.
+func DeepDerivativeEqual(expected interface{}) types.GomegaMatcher {
+	return &deepDerivaiveNMatcher{
+		expected: expected,
+	}
+}
+
+type deepDerivaiveNMatcher struct {
```
rfranzke:
```suggestion
type deepDerivativeNMatcher struct {
```

rfranzke (commented):
lgtm, nice PR - just a nit

Though, let's also wait with merge until v1.1.0 was cut.

mvladev:
fixed nit in ceb2aed69f54e6ab1bf8be1a12c23f6e1c837072

rfranzke (dismissed):
/lgtm

pkg/client/kubernetes/chartoptions.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (forceNamespace) MutateApplyOptions(opts *ApplyOptions) {
+	opts.ForceNamespace = true
+}
+
+func (forceNamespace) MutateDeleteOptions(opts *DeleteOptions) {
+	opts.ForceNamespace = true
+}
+
+// ValueOption contains value applier for Apply and Delete.
```
timebertt:
```suggestion
// ValueOption contains options for Apply and Delete.
```

At first I was a bit confused by the name `ValueOption`, because it does not only add chart values. Can you rename it to something like `ChartApplierOptions` (because it is primarily used in the ChartApplier)?

timebertt (commented):
/lgtm
Just a minor suggestion

pkg/client/kubernetes/chartoptions.go
mvladev: `ChartApplierOptions` is not good enough because the interface contains `Apply` and `Delete` options.

timebertt: Agreed, I was also not satisfied with `ChartApplierOptions`.
I just stumbled upon `ValueOption` because it might also configure other options than the values for a chart. But I have no strong opinion on changing the name.

rfranzke:
Can you rebase again?

mvladev:
Rebased to latest changes

timebertt (approved):

Issue #1454, opened 2019-09-24
Container Runtime Options


rfranzke:
**What would you like to be added**:
Today, Gardener has a hard requirement to Docker, but there are a couple of more container runtimes out there, e.g. `cri-o` or `gvisor`. We should investigate what is necessary to make the container runtime pluggable, and to allow end-users to configure which runtime shall be used on shoot level.

**Why is this needed**:
End-users are interested in using other container runtimes than Docker.

majst01:
Would love to help here

liorokman:
Seems to me that the biggest value would be to make Gardner-managed clusters use Containerd to launch pods by default. Both gVisor and Kata Containers support integrating into Kubernetes using Containerd shims.

Since the latest Docker CE packages come with Containerd, and CoreOS has Containerd preinstalled, the change would mainly require adding the relevant kubelet parameters (`--container-runtime=remote --runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock`). On CoreOS images the correct socket would be `/var/run/docker/libcontainerd/docker-containerd.sock` 

Once kubernetes is configured to use Containerd, actually adding additional runtimes like gvisor or kata containers would require adding the relevant binaries to the nodes (e.g. by providing a [torcx](https://github.com/coreos/torcx) package for CoreOS), configuring containerd correctly, and adding a RuntimeClass object into the Kubernetes control plane.

vpnachev:
Gardener is not working only with CoreOS, it also has support for Ubuntu and SUSE-JEOS. New operating systems might be added in the future, too. A clean contract from the Gardener is required w.r.t. container runtime, so that new operating systems controllers will be implemented without any change in the gardner code.

At the moment, the OS can be configured per worker pool, because of potential issues(missing support for container runtime X in OS Y) maybe the container runtime should be configurable per worker pool, WDYT?

majst01:
Configuration per worker pool is probably the most flexible solution, but probably kinda difficult to manage from a user perspective if there is no default in the cloudprofile.
Regarding container runtime, i would also prefer to add containerd as a second option for the start.

vpnachev:
> Configuration per worker pool is probably the most flexible solution, but probably kinda difficult to manage from a user perspective if there is no default in the cloudprofile.

OS type and version is optional in the shoot manifest, too. If the OS is omitted in the shoot manifest, it is defaulted to the first one available in the respective cloud profile.

[...]

PR #664, opened 2019-01-17
Introduce logrotate

ialidzhikov:
**What this PR does / why we need it**:
- Enables [logrotate](https://linux.die.net/man/8/logrotate) for container log rotation. 
- Configures [journal](https://www.freedesktop.org/software/systemd/man/journald.conf.html) to rotate the system components that do not run in container and log in journal - kubelet, the container runtime (docker). 

**Which issue(s) this PR fixes**:
Fixes #653 and #662.

**Special notes for your reviewer**:
See:
- https://kubernetes.io/docs/concepts/cluster-administration/logging/#logging-at-the-node-level
- https://docs.docker.com/config/containers/logging/json-file/

**Release note**:

```noteworthy operator
All shoot worker machines are now configured with log rotation for user containers and system components. Log files are kept for a maximum of 14 days.
```

Issue #653, opened 2019-01-15
Ensure Deletion of Logfiles older than Two weeks


marwinski:
In order to be DIPA compliant we need to ensure that logfiles older than 2 weeks are deleted.

### Implementation proposal

There are several ways on how the logging stack of Docker can be configured during cloud init. There may be an option to write to a new file every day and to rotate the log every day for up to fourteen days. In case there is no such option we may have to write a custom logger which provides this option.

### Definition of Done

There shall be a test that checkes that there are no logfiles older than 14 days. As staging is usually updated in shorter periods of time (where logs are being deleted) that test should be run on the canary landscape.

### Side conditions

We will be writing the logs to an external logging stack (EFK). This logging stack must be made compliant as well.

ialidzhikov:
I will take this one.

vlerenc:
Thanks for this ticket/taking it. Indeed we have two locations, we need to tweak the Docker log file handling and the ES log handling. I am not sure about Stackdriver for our Garden GKE clusters. I don't think we use that at all (@ThormaehlenFred, @mliepold), so there should be no issue with that (atm).

mliepold:
The container log files of our Garden GKE clusters are not delivered to Stackdriver at the moment.
However some AuditLog, cluster operation and GCE resource information is logged there by default.

vlerenc:
Is that a strict rule (14d) @marwinski? Sometimes, Gardener operators would like to have logs available for more than 14d in order to have enough time to follow issues and provide RCAs.

cc @ThormaehlenFred

JordanJordanov:
DIPA requirements state we need to keep logs for 14 days. 
Due to personal data in the logs (incl. IP addresses) DPP requirements state we need to delete the logs after a reasonable period of time, which we decided to be 14 days.

PR #1971, opened 2020-02-24
GEP 10 - ContainerD Enablement

nimrodoron:
Enablement of ContainerD according to GEP-10

Part of #1454

**Release note**:

```noteworthy user
The `core.gardener.cloud/v1beta1.Shoot` resource now has a new `.spec.provider.workers[].cri` optional property. This property contains a `name` field. The only valid value for this field at this time is: `containerd`. Setting this property with the relevant value will enable the kubelet to work with `containerd` as a CRI instead of the default Docker Shim in case the used operating system config extension supports it. Please note that this cannot be changed afterwards - only during cluster creation or when a new worker pool is added. Also, please double-check whether the corresponding extension controller for the operating system that you use is supporting `containerd`.
```

```noteworthy developer
The `extensions.gardener.cloud/v1alpha1.OperatingSystemConfig` resource has a new `.spec.criConfig` property. This property contains a new `name` field. This optional property will be set if the shoot's worker pool config contains CRI configuration with the current only valid value `containerd`. Each OS extensions should enable the `containerd` with the default configurations as mentioned in:
https://github.com/gardener/gardener/blob/master/docs/proposals/10-shoot-additional-container-runtimes.md
```


pkg/apis/extensions/v1alpha1/types_operatingsystemconfig.go
```diff
@@ -69,7 +69,8 @@ type OperatingSystemConfigList struct {
 type OperatingSystemConfigSpec struct {
 	// DefaultSpec is a structure containing common fields used by all extension resources.
 	DefaultSpec `json:",inline"`
-
+	// EnableContainerD indicate if kubelet should be started with ContainerD and not docker
```
rfranzke:
```suggestion
	// EnableContainerD indicate if kubelet should be started with ContainerD and not Docker.
	// +optional
```

pkg/apis/extensions/v1alpha1/types_operatingsystemconfig.go
```diff
@@ -69,7 +69,8 @@ type OperatingSystemConfigList struct {
...
+	EnableContainerD bool  `json:"enableContainerD,omitempty"`
```
rfranzke:
```suggestion
	EnableContainerD *bool  `json:"enableContainerD,omitempty"`
```

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -741,6 +741,8 @@ type Alerting struct {
 // Provider contains provider-specific information that are handed-over to the provider-specific
 // extension controller.
 type Provider struct {
+	// EnableContainerD indicate if kubelet should be started with ContainerD and not docker
+	EnableContainerD bool  `json:"enableContainerD,omitempty"`
```
rfranzke: Shouldn't this rather be part of the `Worker` struct so that one can enable/disable it per worker pool?
Also, WDYT about using a struct instead of just a bool? This would help in the future in case we decide to allow specifying more configuration options:

```golang
type ContainerD struct {
  // Enabled specifies whether containerd should be enabled or not.
  Enabled bool `json:"enabled"`
}
```

...

```golang
type Worker struct {
  ...
  // ContainerD contains configuration settings for containerd.
  // +optional
  ContainerD *ContainerD `json:"containerd,omitempty"`
}
```

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -744,6 +744,8 @@ type Alerting struct {
 // Provider contains provider-specific information that are handed-over to the provider-specific
 // extension controller.
 type Provider struct {
+	// EnableContainerD indicate if kubelet should be started with ContainerD and not docker
+	EnableContainerD bool  `json:"enableContainerD,omitempty"`
```
rfranzke: Same like for v1beta1 version

pkg/apis/core/types_shoot.go
```diff
@@ -613,6 +613,8 @@ type Alerting struct {
 // Provider contains provider-specific information that are handed-over to the provider-specific
 // extension controller.
 type Provider struct {
+	// EnableContainerD indicate if kubelet should be started with ContainerD and not docker
+	EnableContainerD bool
```
rfranzke: Same like for v1beta1, v1alpha1 version

charts/seed-operatingsystemconfig/original/templates/scripts/_health-monitor.sh
```diff
@@ -136,6 +148,8 @@
         echo "Start kubernetes health monitoring for $component"
         if [[ $component == "docker" ]]; then
```
rfranzke, line 149: What about the Docker monitoring script? Do we still need it even if we containerd is enabled?

charts/seed-operatingsystemconfig/original/templates/osc.yaml
```diff
@@ -32,3 +42,9 @@ spec:
 {{ include "kernel-config" . | indent 2 }}
 {{ include "health-monitor" . | indent 2 }}
 {{ include "gardener-user-script" . | indent 2 }}
+{{- if .Values.enableContainerD }}
+{{ include "init-containerd-config-script" . | indent 2 }}
+{{- end }}
+
```
rfranzke: Can we remove the unnecessary newlines at the end here? :)

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags
```diff
@@ -15,6 +15,9 @@
 {{- if semverCompare "< 1.14" .Values.kubernetes.version }}
 --feature-gates=SupportPodPidsLimit=true
 {{- end }}
+{{- if .Values.enableContainerD }}
+--feature-gates=CRIContainerLogRotation=true
```
rfranzke: Is this feature gate working for all Kubernetes version we support (1.10 - 1.17)?

charts/seed-operatingsystemconfig/downloader/templates/osc.yaml
```diff
@@ -14,6 +14,9 @@ metadata:
 spec:
   type: {{ required "type is required" .Values.type }}
   purpose: {{ required "purpose is required" .Values.purpose }}
+  {{- if .Values.enableContainerD }}
```
rfranzke: Can you add the new Helm chart values to the `values.yaml` so that one can see which values are possible?

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags
```diff
@@ -15,6 +15,9 @@
 {{- if semverCompare "< 1.14" .Values.kubernetes.version }}
 --feature-gates=SupportPodPidsLimit=true
 {{- end }}
+{{- if .Values.enableContainerD }}
```
rfranzke: Can you add the new Helm chart values to the `values.yaml` so that one can see which values are possible?

pkg/apis/core/validation/shoot.go
```diff
@@ -352,6 +353,14 @@ func validateKubernetesVersionUpdate(new, old string, fldPath *field.Path) field
 	return allErrs
 }
 
+func validateEnableContianerdField(new bool, old bool, fldPath *field.Path) field.ErrorList {
```
```suggestion
func validateContainerD(containerd *core.ContainerD, fldPath *field.Path) field.ErrorList {
```

pkg/controllermanager/controller/shoot/shoot_hibernation_types.go
```diff
@@ -103,6 +103,7 @@ func (h *hibernationJob) Run() {
 				return nil, fmt.Errorf("shoot %s/%s hibernation schedule changed mid-air", shoot.Namespace, shoot.Name)
 			}
 			shoot.Spec.Hibernation.Enabled = &h.enabled
+			shoot.Spec.Provider.EnableContainerD = true;
```
rfranzke: Hmm, I don't think we can do this yet. Yes, eventually we want all the clusters to use containerd, however, we should announce this properly with a deprecation period and let people switch themselves when the time fits for them.

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -177,6 +177,7 @@ func (c *defaultMaintenanceControl) Maintain(shootObj *gardencorev1beta1.Shoot,
 		if updatedKubernetesVersion != nil {
 			s.Spec.Kubernetes.Version = *updatedKubernetesVersion
 		}
+		s.Spec.Provider.EnableContainerD = true
```
rfranzke: Same like above

pkg/operation/shoot/types.go
```diff
@@ -43,6 +43,7 @@ type Shoot struct {
 	WantsAlertmanager      bool
 	IgnoreAlerts           bool
 	HibernationEnabled     bool
+	EnableContainerD	   bool
```
rfranzke: Do we really need this field? It's not very useful, is it?

rfranzke (changes requested):
Nice PR, thanks for your efforts, I'm having a few suggestions here and there.

rfranzke:
Also, please squash the commits. Maybe it makes sense to add some small document to `docs/usage` describing what you've implemented for end-users?

pkg/apis/core/v1beta1/types_shoot.go
nimrodoron: I first put it under worker but noticed that i'm changing it only by Shoot logic (Hibernation turns on, Kuberenetes version update etc) I didn't find a scenario to change it per Worker.
I'm not sure it is relevant for users to change it per worker because this change should be "transparent" to their cluster.
If you think it is does make sense to set it per Worker, maybe we can create a new struct which will already be used for the container runtime as well.
```
type Cri struct {
  // Enabled specifies whether containerd should be enabled or not.
  ContainerDEnabled bool `json:"enabled"`
  GVisorEnabled       bool `json:"enabled"`
}
```
WDYT?

pkg/apis/core/v1alpha1/types_shoot.go
nimrodoron: Please have a look on my pkg/apis/core/v1beta1/types_shoot.go comment.
Once we will decide i will update also for v1alpha1

pkg/apis/core/types_shoot.go
nimrodoron: Once we will decide i will also update here.

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags
nimrodoron: Do you mean adding the values:
--container-runtime=remote
--container-runtime-endpoint=/run/containerd/containerd.sock
?

pkg/controllermanager/controller/shoot/shoot_hibernation_types.go
nimrodoron: Ok, i will remove it but do you want to keep it in the GEP?

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
nimrodoron: Ok, i will remove it but do you want to keep it in the GEP?

pkg/operation/shoot/types.go
nimrodoron: I do use this field to render the OSC.
operatingsystemconfig.go line 92

pkg/apis/core/v1beta1/types_shoot.go
rfranzke: I don't have a scenario either, but that doesn't mean that no one exists. If there is no reason to not make it controllable per worker pool then we usually always prefer to give the end-users the maximum possible flexibility. Hence, I'd vote for moving it there.

> If you think it is does make sense to set it per Worker, maybe we can create a new struct which will already be used for the container runtime as well.

Hm, we should not "hard-code" the possible container runtimes in the API. It should be like you proposed in the GEP10, so a list of container runtimes plus proper provider configs that are not understandable by Gardener.

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags
rfranzke: No, I mean the `enableContainerD`, for example. It's a new value that you added for the Helm chart, and usually we document all possible values to the `values.yaml`.

pkg/controllermanager/controller/shoot/shoot_hibernation_types.go
rfranzke: Hm, I'll come back to this question. Gimme some time.

pkg/operation/shoot/types.go
rfranzke: I know. But usually the `Shoot` structs only contains fields that are "expensive" to re-compute. In this case you only read a value from the Shoot manifest, so there is no reason/need to store it as struct attribute.

pkg/apis/core/v1beta1/types_shoot.go
nimrodoron: True. it was here by mistake, this should be:
```
type Worker struct {

  // ContainerD contains configuration settings for containerd.
  // +optional
  Cri *Cri `json:"cri,omitempty"`
}

type Cri struct {
  // Enabled specifies whether containerd should be enabled or not.
  ContainerDEnabled bool `json:"enabled"`
 ContainerRuntime   []containerRuntime `json:"containerRuntime"`
}

type ContainerRuntime struct {
 type string  `json:"type"`
}
```

Also, I remove the auto change from false to true but keep the rule enforcement so changing the enablement of ContainerD from true to false won't be possible.
Is it ok?

rfranzke: Ah, ok, understood. Yes, this looks great, it makes sense 👍 thank you

charts/seed-operatingsystemconfig/downloader/templates/osc.yaml
nimrodoron: Ok i got you,
I will add it.

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags
nimrodoron: Ok i got you,
I will add it.

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags
nimrodoron: Yes,

Feature | Default | Stage | Since | Until
-- | -- | -- | -- | --
CRIContainerLogRotation | false | Alpha | 1.10 | 1.10
CRIContainerLogRotation | true | Beta | 1.11



CRIContainerLogRotation | false | Alpha | 1.10 | 1.10
CRIContainerLogRotation | true | Beta | 1.11

charts/seed-operatingsystemconfig/original/templates/osc.yaml
nimrodoron: Sure

charts/seed-operatingsystemconfig/original/templates/scripts/_health-monitor.sh
nimrodoron: For now, It is still possible to work with Docker. When you choose that docker monitor service (using the docker monitor script)

pkg/operation/shoot/types.go
nimrodoron: done

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -768,6 +768,8 @@ type Worker struct {
 	// CABundle is a certificate bundle which will be installed onto every machine of this worker pool.
 	// +optional
 	CABundle *string `json:"caBundle,omitempty"`
+	// Cri contains configurations of CRI support of every machine in the worker pool
+	Cri *Cri `json:"cri,omitempty"`
```
vpnachev:
```suggestion
	CRI *CRI `json:"cri,omitempty"`
```

also, change on the other places where used.

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -853,11 +855,19 @@ type Volume struct {
 	Encrypted *bool `json:"encrypted,omitempty"`
 }
 
+// Cri contains information about the Container Runtimes.
+type Cri struct {
+	// Enable Kubelet support for ContainerD
+	EnableContainerD bool `json:"enableContainerD"`
```
vpnachev: I do not like the explicit `enable` in the field/variable name.
Checking some fields of the pod spec how are they defined here https://github.com/kubernetes/kubernetes/blob/fe9073b8c14bbf6bbe42a2259f01d3d3a853a6a3/staging/src/k8s.io/api/core/v1/types.go#L2924-L2947

And I think a simple `containerd` is enough
```suggestion
	Containerd bool `json:"containerd"`
```

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -853,11 +855,19 @@ type Volume struct {
...
+}
+
 var (
 	// DefaultWorkerMaxSurge is the default value for Worker MaxSurge.
 	DefaultWorkerMaxSurge = intstr.FromInt(1)
 	// DefaultWorkerMaxUnavailable is the default value for Worker MaxUnavailable.
 	DefaultWorkerMaxUnavailable = intstr.FromInt(0)
+	// DefaultCri is the default value for worker Cri.
+	DefaultWorkerCri = Cri{EnableContainerD: true}
```
vpnachev:
```suggestion
	DefaultWorkerCRI = CRI{Containerd: true}
```

The default should be false, I think

pkg/apis/core/validation/shoot.go
```diff
@@ -168,6 +168,7 @@ func ValidateShootSpecUpdate(newSpec, oldSpec *core.ShootSpec, deletionTimestamp
 	allErrs = append(allErrs, validateKubernetesVersionUpdate(newSpec.Kubernetes.Version, oldSpec.Kubernetes.Version, fldPath.Child("kubernetes", "version"))...)
 	allErrs = append(allErrs, validateKubeProxyModeUpdate(newSpec.Kubernetes.KubeProxy, oldSpec.Kubernetes.KubeProxy, newSpec.Kubernetes.Version, fldPath.Child("kubernetes", "kubeProxy"))...)
 	allErrs = append(allErrs, validateKubeControllerManagerConfiguration(newSpec.Kubernetes.KubeControllerManager, oldSpec.Kubernetes.KubeControllerManager, fldPath.Child("kubernetes", "kubeControllerManager"))...)
+	allErrs = append(allErrs, ValidateProviderUpdate(&newSpec.Provider, &oldSpec.Provider, fldPath.Child("provider"))...)
 
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newSpec.Provider.Type, oldSpec.Provider.Type, fldPath.Child("provider", "type"))...)
```
vpnachev: Now this should be part of `func ValidateProviderUpdate`

pkg/apis/core/validation/shoot.go
```diff
@@ -352,6 +385,14 @@ func validateKubernetesVersionUpdate(new, old string, fldPath *field.Path) field
 	return allErrs
 }
 
+func validateCri(newCri *core.Cri, oldCri *core.Cri, fldPath *field.Path) field.ErrorList {
```
vpnachev:
```suggestion
func validateCRI(newCri *core.CRI, oldCRI *core.CRI, fldPath *field.Path) field.ErrorList {
```

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -312,6 +312,9 @@ func (b *Botanist) deployOperatingSystemConfigsForWorker(machineTypes []gardenco
 		"kubeletDataVolume": worker.KubeletDataVolumeName,
 	}
 
+	originalConfig["enableContainerD"] = worker.Cri != nil && worker.Cri.EnableContainerD
+	downloaderConfig["enableContainerD"] = worker.Cri != nil && worker.Cri.EnableContainerD
```
vpnachev: I think the downloader does not care about the container runtime, but I have to double check it.

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -850,11 +852,19 @@ type Volume struct {
...
+	EnableContainerD bool `json:"enableContainerD"`
+}
+
 var (
 	// DefaultWorkerMaxSurge is the default value for Worker MaxSurge.
 	DefaultWorkerMaxSurge = intstr.FromInt(1)
 	// DefaultWorkerMaxUnavailable is the default value for Worker MaxUnavailable.
 	DefaultWorkerMaxUnavailable = intstr.FromInt(0)
+	// DefaultCri is the default value for worker Cri.
+	DefaultWorkerCri = Cri{EnableContainerD: true}
```
vpnachev: The default should be `false`, otherwise this is change in the behavior.

pkg/apis/core/v1alpha1/defaults.go
```diff
@@ -163,6 +163,9 @@ func SetDefaults_Worker(obj *Worker) {
 	if obj.MaxUnavailable == nil {
 		obj.MaxUnavailable = &DefaultWorkerMaxUnavailable
 	}
+	if obj.Cri == nil {
+		obj.Cri = &DefaultWorkerCri
+	}
```
vpnachev: is this necessary?

pkg/apis/core/v1beta1/constants/types_constants.go
```diff
@@ -237,6 +237,10 @@ const (
 
 	// OperatingSystemConfigUnitNameKubeletService is a constant for a unit in the operating system config that contains the kubelet service.
 	OperatingSystemConfigUnitNameKubeletService = "kubelet.service"
+	// OperatingSystemConfigUnitNameKubeletService is a constant for a unit in the operating system config that contains the docker service.
```
vpnachev:
```suggestion
	// OperatingSystemConfigUnitNameDockerService is a constant for a unit in the operating system config that contains the docker service.
```

pkg/apis/core/v1beta1/constants/types_constants.go
```diff
@@ -237,6 +237,10 @@ const (
...
+	OperatingSystemConfigUnitNameDockerService = "docker.service"
+	// OperatingSystemConfigUnitNameKubeletService is a constant for a unit in the operating system config that contains the containerd service.
```
vpnachev:
```suggestion
	// OperatingSystemConfigUnitNameContainerDService is a constant for a unit in the operating system config that contains the containerd service.
```

pkg/apis/core/validation/shoot.go
```diff
@@ -352,6 +385,14 @@ func validateKubernetesVersionUpdate(new, old string, fldPath *field.Path) field
...
+	allErrs := field.ErrorList{}
+	if (newCri == nil || !newCri.EnableContainerD) && (oldCri != nil && oldCri.EnableContainerD) {
```
vpnachev: It looks like containerd cannot be disabled once enabled for a worker.
But the reverse is possible?

Please add some docs in the `types_shoot.go` files about this, at least.

pkg/apis/core/validation/shoot_test.go
```diff
@@ -790,6 +790,25 @@ var _ = Describe("Shoot Validation Tests", func() {
 
 				Expect(errorList).To(HaveLen(0))
 			})
+
+			It("should not allow changing the enablement of containerd in existing worker pool", func() {
```
vpnachev: if my assumption is right that containerd can be enabled, but not disabled then
```suggestion
			It("should not allow containerd to be disabled for existing worker pool", func() {
```

vpnachev (changes requested):
thank you!

pkg/apis/core/v1beta1/types_shoot.go
vpnachev: BTW, I've heard about users interacting with the docker socket on the nodes.
So the fact that we are not aware about some use-case does not mean that it does not exist.

And I am against changing the behavior in the current version of the API or forcefully migrating existing shoot clusters.

pkg/controllermanager/controller/shoot/shoot_hibernation_types.go
rfranzke: Hi Nimrod, yes, for now let's not change it on-the-fly. We should rather ask users to do it with a proper deprecation period.

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
rfranzke: Yes, I think it makes sense to remove it from the GEP

pkg/apis/core/v1alpha1/types_shoot.go
rfranzke: Instead of using a `bool`, can we rather use something like `Runtime string` that allows either `docker` or `containerd` and defaults to `docker` (even for new clusters for now)? I think API-wise this is cleaner instead of "hard-coding" containerd here.

vpnachev: I agree.

pkg/controllermanager/controller/shoot/shoot_hibernation_types.go
nimrodoron: Ok, i will remove the true defaults from the code and the GEP

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
nimrodoron: Ok

pkg/apis/core/v1alpha1/types_shoot.go
nimrodoron: I Prefer the section:
```
cri:
    name: containerd
    containerRuntimes:
    - type: gvisor
    - type: kata-containers
```

Docker is not a CRI (But there are others CRI options like CRI-O. So it is better to make it a string instead of hard-coding boolean but rather leave docker out of it.

pkg/apis/core/v1alpha1/types_shoot.go
nimrodoron: After the discussion i will remove all true defaults.
It will be a change triggered only by users and explicitly.
Also updating the GEP.

pkg/apis/core/v1alpha1/defaults.go
nimrodoron: After the discussion i will remove all true defaults.
It will be a change triggered only by users and explicitly.
Also updating the GEP.

pkg/apis/core/v1alpha1/types_shoot.go
rfranzke: +1, thanks for pointing it out, agreed

mvladev: I think that `name` would be better in the context of `ContainerRuntime`:
```yaml
cri:
  name: containerd
  containerRuntimes:
  - name: gvisor
  - name: kata-containers
```

rfranzke: No strong opinion on this, but in the end we are usually always calling the extension providers with `type`, e.g. `.spec.provider.type`, `spec.extensions[].type`. The only exception is the machine image _name_, but this wasn't done on purpose I guess.

nimrodoron: Lets keep it to the GEP and the next commit for containerRuntimes.
For now it is:
cri:
  name: containerd

pkg/apis/core/validation/shoot.go
nimrodoron: I changed it, it can be updated.

pkg/apis/extensions/v1alpha1/types_operatingsystemconfig.go
```diff
@@ -69,7 +69,9 @@ type OperatingSystemConfigList struct {
 type OperatingSystemConfigSpec struct {
 	// DefaultSpec is a structure containing common fields used by all extension resources.
 	DefaultSpec `json:",inline"`
-
+	// EnableContainerD indicate if kubelet should be started with ContainerD and not Docker.
+	// +optional
+	EnableContainerD *bool  `json:"enableContainerD,omitempty"`
```
mvladev: This is leftover?

nimrodoron: No, this is the indication in the OSC for the enablement of ContainerD.
OS extensions using this.

mvladev: Why it would be needed? The OSC will already contain `containerd.service` unit. Isn't this enough for the extension to handle it?

nimrodoron: In my opinion the explicit enablement of ContainerD is more readable and understood,
The ContainerD unit doesn't even contain a start command but just Drop In unit for an Pre execution init script.
More than that, ContainerD is enabled by default in most OS, Docker already using it even without the enablement of ContainerD in Gardener shoot.

pkg/apis/core/types_shoot.go
```diff
@@ -700,6 +702,19 @@ type Volume struct {
...
+type CRI struct {
+	// The name of the CRI library
+	Name string
+}
+
+// CRIName is a type alias for the CRI name string.
+type CRIName string
+
+const (
+	CRINameContainerD CRIName = "containerd"
```
rfranzke, line 720: Can we add a doc string for this? Also, I suggest to add `CRINameDocker` as second possible option for the CRI name.

pkg/apis/core/types_shoot.go
```diff
@@ -754,3 +769,5 @@ const (
 	// ShootPurposeInfrastructure is a constant for the infrastructure purpose.
 	ShootPurposeInfrastructure ShootPurpose = "infrastructure"
 )
+
```
rfranzke: Remove unneeded empty lines

pkg/apis/extensions/v1alpha1/types_operatingsystemconfig.go
rfranzke: I get your point @nimrodoron, but for the sake of of consistency I would probably also use the CRI name here. Concretely, maybe a `CRI struct` with (today) just one field `Name` is better from an API pov. Assuming we will add a third option next to docker and containerd tomorrow, we would have hard-coded containerd in this API which is not too nice.

pkg/apis/core/types_shoot.go
rfranzke: Also, in the `v1alpha1` and `v1beta1` version, can you default the `cri.name` field then to `docker` if its not set?

rfranzke (changes requested):

rfranzke:
Can you also add release notes for end-users (because the shoot contains new fields) and release notes for OS extension developers (because the `OSC` contains new fields)?

pkg/apis/core/types_shoot.go
nimrodoron: Docker shim doesn't implement the CRI.
Actually ContainerD is the CRI for Docker shim.
Maybe it is better to keep this struct empty if no CRI was added.
If CRI section was added currently only "containerd" value is possible.

rfranzke: OK, makes sense. Thank you.

nimrodoron:
> Can you also add release notes for end-users (because the shoot contains new fields) and release notes for OS extension developers (because the `OSC` contains new fields)?

Hi,
Can you point me to example of Release notes?
And where do i add them?

pkg/apis/extensions/v1alpha1/types_operatingsystemconfig.go
nimrodoron: Ok i added CRIConfig.

mvladev:
@nimrodoron I've updated your PR comment in the top with the template. You can use markdown syntax in it. The end result is merged in the release notes - e.g. https://github.com/gardener/gardener/releases/tag/v1.0.0

mvladev (dismissed):

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -306,10 +306,18 @@ func (b *Botanist) deployOperatingSystemConfigsForWorker(machineTypes []gardenco
 		}
 	}
 
+	CRIConfig := map[string]interface{}{
```
rfranzke:
```suggestion
	criConfig := map[string]interface{}{
```

rfranzke (changes requested):
Looks good now, thanks, just a few minors suggestions:
1. Run `make generate` and push again - I see changes here.
1. Can you add the new `cri` field to the example shoot manifest here: https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml#L29-L92?

rfranzke:
Let's wait with the merge until `v1.1.0` was cut (planned for tomorrow, cc @timuthy)

pkg/operation/botanist/operatingsystemconfig.go
nimrodoron: Done

nimrodoron:
> Looks good now, thanks, just a few minors suggestions:
> 
> 1. Run `make generate` and push again - I see changes here.
> 2. Can you add the new `cri` field to the example shoot manifest here: https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml#L29-L92?

Done

pkg/apis/core/validation/shoot.go
```diff
@@ -848,6 +851,10 @@ func ValidateWorker(worker core.Worker, fldPath *field.Path) field.ErrorList {
 		}
 	}
 
+	if (worker.CRI != nil) {
```
ialidzhikov:
```suggestion
	if worker.CRI != nil {
```

charts/seed-operatingsystemconfig/original/templates/osc.yaml
```diff
@@ -9,6 +9,10 @@ metadata:
 spec:
   type: {{ required ".osc.type is required" .Values.osc.type }}
   purpose: {{ required ".osc.purpose is required" .Values.osc.purpose }}
+  {{- if .Values.worker.cri.name }}
```
ialidzhikov:
```
$ make check

Error: render error in "original/templates/osc.yaml": template: original/templates/osc.yaml:12:16: executing "original/templates/osc.yaml" at <.Values.worker.cri.n...>: can't evaluate field name in type interface {}
make: *** [check] Error 1
```

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -306,10 +306,18 @@ func (b *Botanist) deployOperatingSystemConfigsForWorker(machineTypes []gardenco
 		}
 	}
 
+	criConfig := map[string]interface{}{
+		"enableContainerD": worker.CRI != nil && worker.CRI.Name == gardencorev1beta1.CRINameContainerD,
+	}
+	if (worker.CRI != nil) {
```
ialidzhikov:
```suggestion
	if worker.CRI != nil {
```

pkg/apis/core/validation/shoot.go
```diff
@@ -1156,3 +1163,13 @@ func IsNotMoreThan100Percent(intOrStringValue *intstr.IntOrString, fldPath *fiel
 
 	return allErrs
 }
+
+func ValidateCRI(CRI *core.CRI,  fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if (!avaliableWorkerCRINames.Has(string(CRI.Name))) {
```
ialidzhikov:
```suggestion
	if !avaliableWorkerCRINames.Has(string(CRI.Name)) {
```

pkg/client/core/openapi/zz_generated.openapi.go
```diff
@@ -23,10147 +23,64 @@ limitations under the License.
...
-	v1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
-	v1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	spec "github.com/go-openapi/spec"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	common "k8s.io/kube-openapi/pkg/common"
 )
 
 func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenAPIDefinition {
 	return map[string]common.OpenAPIDefinition{
-		"github.com/gardener/gardener/pkg/apis/core/v1alpha1.Addon":                                schema_pkg_apis_core_v1alpha1_Addon(ref),
```
ialidzhikov: From @tim-ebert comment in another PR:
> @timuthy figured out, that this is related to your installed bash version. Try regenerating with bash version 5.
> Seems that k8s.io/kube-openapi works only with bash version 5.

ialidzhikov (changes requested):

rfranzke:
Sorry, I totally forgot about it: As you change the `OperatingSystemConfig` CRD and extension controllers are expected to implement something when the containerd configuration gets set (https://github.com/gardener/gardener/pull/1971/files#diff-f0c5ef592d3ad798bdb23c30c05ba9c0R72), could you please add a small description to the docs (https://github.com/gardener/gardener/blob/master/docs/extensions/operatingsystemconfig.md) ?

nimrodoron:
> Sorry, I totally forgot about it: As you change the `OperatingSystemConfig` CRD and extension controllers are expected to implement something when the containerd configuration gets set (https://github.com/gardener/gardener/pull/1971/files#diff-f0c5ef592d3ad798bdb23c30c05ba9c0R72), could you please add a small description to the docs (https://github.com/gardener/gardener/blob/master/docs/extensions/operatingsystemconfig.md) ?

Hi, @rfranzke 
Regarding this, For SuSe, at the moment we still need to download ctr for the internet and for Ali-Cloud OS Extensions there is still no implementation (Should be similar to non Ali but i haven't checked yet).
If the OS Extensions won't fix the ContainerD and Docker configurations the Kubelet won't start and the Worker will fail.
What do you think is the recommended way for now? allow to set the CRI to ContainerD only in Ubuntu and CoreOS?
I will update also the docs accordingly.

pkg/client/core/openapi/zz_generated.openapi.go
nimrodoron: I re-generated with bash version 5

charts/seed-operatingsystemconfig/original/templates/osc.yaml
nimrodoron: Fixed the values for the test

rfranzke:
> Regarding this, For SuSe, at the moment we still need to download ctr for the internet

Yes, this should be changed as discussed in the PR/GEP.

> Ali-Cloud OS Extensions there is still no implementation (Should be similar to non Ali but i haven't checked yet).

@jia-jerry @minchaow @EmoinLanyu can you look into it and help out?

> If the OS Extensions won't fix the ContainerD and Docker configurations the Kubelet won't start and the Worker will fail.
What do you think is the recommended way for now? allow to set the CRI to ContainerD only in Ubuntu and CoreOS?

Gardener cannot decide for which extension types it is possible or not. As of today, the only option is to ship the OS extensions with a validation webhook (as proposed [here](https://github.com/gardener/gardener/blame/master/docs/proposals/10-shoot-additional-container-runtimes.md#L64)) that prevents setting the `CRI` fields if the machine image does not support it.

nimrodoron:
> here

Ok thanks, So the webhook will be done on each OS extension which won't support the ContainerD

docs/extensions/operatingsystemconfig.md
```diff
@@ -168,6 +168,34 @@ status:
 
 Once the `.status` indicates that the extension controller finished reconciling Gardener will continue with the next step of the shoot reconciliation flow.
 
+## CRI Support
+Gardener support specifying Container Runtime Interface (CRI) in the OperatingSystemConfig. The only CRI supported at the moment is: ContainerD.
```
rfranzke:
```suggestion
Gardener supports specifying Container Runtime Interface (CRI) configuration in the `OperatingSystemConfig` resource. The only CRI supported at the moment is `containerd`.
```

docs/extensions/operatingsystemconfig.md
```diff
@@ -168,6 +168,34 @@ status:
...
+  namespace: default
+spec:
+  type: <my-operating-system>
+  purpose: reconcile
+  reloadConfigFilePath: /var/lib/cloud-config-downloader/cloud-config
+  cri:
+    name: containerd
+...
+```
+If the cri section exist, the name property is mandatory. The only valid value at the moment is containerd.
```
rfranzke:
```suggestion
If the `.spec.cri` section exists then the `name` property is mandatory. The only valid value at the moment is `containerd`.
```

docs/extensions/operatingsystemconfig.md
```diff
@@ -168,6 +168,34 @@ status:
...
+When the cri is declared the Kubelet will be configured by Gardener to work with ContainerD. Gardener expects that ContainerD service is running on the OS with the default socket path: unix:///run/containerd/containerd.sock.
```
rfranzke:
```suggestion
When the `.spec.cri` field is declared the kubelet will be configured by Gardener to work with ContainerD. Gardener expects that ContainerD service is running on the OS with the default socket path: `unix:///run/containerd/containerd.sock`. 
```

docs/extensions/operatingsystemconfig.md
```diff
@@ -168,6 +168,34 @@ status:
...
+```
+If the cri section exist, the name property is mandatory. The only valid value at the moment is containerd.
+When the cri is declared the Kubelet will be configured by Gardener to work with ContainerD. Gardener expects that ContainerD service is running on the OS with the default socket path: unix:///run/containerd/containerd.sock. 
+
+Each OS extension must support the CRI configurations by:
+1. The operating system must have built-in  [ContainerD](https://containerd.io/) and the  [Client CLI](https://github.com/projectatomic/containerd/blob/master/docs/cli.md/)
+2. ContainerD service should be configure to work with the default configuration file in: /etc/containerd.config.toml (Created by Gardener).
+
+
+Or create a Webhook for the OperatingSystemConfig which doesn't allow cri configurations.
```
rfranzke:
```suggestion
If CRI configurations are not supported it is recommended create a validating webhook running in the garden cluster that prevents specifying the `.spec.providers.workers[].cri` section in the `Shoot` objects.
```

rfranzke (commented):
Looks great now, thanks @nimrodoron. A few minor suggestions for the docs.
Will test the PR locally as soon as possible.

charts/seed-operatingsystemconfig/original/values.yaml
```diff
@@ -16,6 +16,9 @@ kubernetes:
   version: 1.11.2
 worker:
   name: cpu-worker
+  cri:
+#   name: containerd
+    enableContainerD: false # true
```
timuthy: For me this seems to be a duplicated field because `name == containerd` implies that `enableContainerD: true`, right? Can you instead add a helper function to the chart?

rfranzke: Agreed, makes sense to remove it in favour of the name?

nimrodoron: I couldn't find any information about helper function to helm chart.
Do you have an example maybe?

nimrodoron: The reason i didn't check the string: containerd in the template is because the const is in the code.
"enableContainerD": worker.CRI != nil && worker.CRI.Name == gardencorev1beta1.CRINameContainerD,
What can be done is sending also the available CRI names as parameters to the template and compare the string in the template. 

This condition will remain the same:
  {{- if .Values.worker.cri.name }}

And this condition:
{{- if .Values.worker.cri.enableContainerD }}
will be changed for exmple to:
{{- if  eq .Values.worker.cri.name osc.cri.containerdName

I'm not sure is the best approach. For example enable or disable other CRI can be different conditions.

nimrodoron: > Agreed, makes sense to remove it in favour of the name?

It does make sense to keep the name. ContainerD is highly coupled with those conditions.

rfranzke: Yes, let's remove the `enableContainerD` in the API - what you are referring to with respect to the Helm chart is an implementation detail that should not influence how the API looks like.

nimrodoron: @rfranzke This is a value relevant only to render the helm chart.
In the OSC spec (API) there is no enableContainerD, only cri.name

rfranzke: Sorry, I missed this. Seeing it now. OK, understood. Fine for me. @timuthy ?

timuthy: > What can be done is sending also the available CRI names as parameters to the template and compare the string in the template.

I actually really like this idea :)

nimrodoron: Ok i changed it

rfranzke:
Thanks - can you please rebase to fix the conflicts? Will do a local test tomorrow

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -860,6 +863,19 @@ type Volume struct {
 	Encrypted *bool `json:"encrypted,omitempty"`
 }
 
+// CRI contains information about the Container Runtimes.
+type CRI struct {
+	// The name of the CRI library
+	Name CRIName `json:"name"`
```
rfranzke, line 869: Can we also already add an optional `providerConfig` field like for other extension types:

https://github.com/gardener/gardener/blob/master/pkg/apis/core/v1beta1/types_shoot.go#L251-L253

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -857,6 +860,19 @@ type Volume struct {
 	Encrypted *bool `json:"encrypted,omitempty"`
 }
 
+// CRI contains information about the Container Runtimes.
+type CRI struct {
+	// The name of the CRI library
+	Name CRIName `json:"name"`
```
rfranzke, line 866: Can we also already add an optional `providerConfig` field like for other extension types:

https://github.com/gardener/gardener/blob/master/pkg/apis/core/v1beta1/types_shoot.go#L251-L253

pkg/apis/core/types_shoot.go
```diff
@@ -705,6 +707,19 @@ type Volume struct {
 	Encrypted *bool
 }
 
+// CRI contains information about the Container Runtimes.
+type CRI struct {
+	// The name of the CRI library
+	Name CRIName
```
rfranzke, line 713: Can we also already add an optional `providerConfig` field like for other extension types:

https://github.com/gardener/gardener/blob/master/pkg/apis/core/v1beta1/types_shoot.go#L251-L253

charts/seed-operatingsystemconfig/original/templates/osc.yaml
```diff
@@ -19,11 +23,22 @@ spec:
 {{ include "update-ca-certs" . | indent 2 }}
 {{ include "docker-logrotate" . | indent 2 }}
```
rfranzke: What about these two units (docker-logrotate)? Are they still valid even when using containerd?

nimrodoron: Valid, because docker is still running. but maybe not that useful anymore

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags
```diff
@@ -16,6 +16,9 @@
 {{- if semverCompare "< 1.14" .Values.kubernetes.version }}
 --feature-gates=SupportPodPidsLimit=true
 {{- end }}
+{{- if .Values.worker.cri }}
+--feature-gates=CRIContainerLogRotation=true
```
ialidzhikov: Why the cri log rotation is enabled? Won't be a conflict between logrotate and cri log rotation such as the one described in https://github.com/gardener/gardener-extension-os-suse-jeos/issues/6?

charts/seed-operatingsystemconfig/original/templates/osc.yaml
rfranzke: Do we need a similar mechanism for containerd? How does the log rotation work here? cc @ialidzhikov 
I see we have the feature gate - is this enough?

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags
nimrodoron: Docker is not used as container runtime when the .Values.worker.cri exist.
According to what i saw, logrotate is responsible for the log rotation of Docker containers.

ialidzhikov: How we make sure with cri log rotation that logs older than 2 weeks are deleted? See https://github.com/gardener/gardener/issues/653.

nimrodoron: Well, need to be checked about 2 weeks
The log rotation can be controlled by:
--container-log-max-size and --container-log-max-files

charts/seed-operatingsystemconfig/original/templates/osc.yaml
nimrodoron: When moving to CRI, The Kubelet is responsible for rotating the containers logs.

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags
liorokman: The easy way to fix this is to not use the kubelet log rotation option, but rather use logrotated for containerd, similar to how this is done for Docker. See #664 - it's basically adding a logrotated configuration file instead of specifying these two options.

@ialidzhikov , @rfranzke : would it be OK to merge this PR and then handle the additional requirements for log rotation in a future PR?

rfranzke: @liorokman Your proposal sounds good, though, I'm afraid we have to do it in this PR. Once `containerd` is exposed in the `Shoot` manifest as option (which it will be with this PR) people expect similar experience wrt standards as before with Docker (#653). We already have some logrotate configuration, can this be enhanced to also work for the `containerd` case?

nimrodoron: Done

charts/seed-operatingsystemconfig/original/templates/osc.yaml
nimrodoron: logrotation of containerd was added instead of kubelet rotation.

rfranzke (approved):
/lgtm, tested with Ubuntu - works perfectly, thank you very much!

PR #1972, opened 2020-02-25
Add shootstate creation step in reconciliation flow

plkokanov:
- Adds a reference to the ShootState in the Operation struct so that steps which require the ShootState won't have to use an additional Get request.
- Adds a creation step for the ShootState at the beginning of the reconciliation
- Adds some convenience functions for working with a list of `ExtensionStateResources` (similar ones will be added for `GardenerResourceData` entries in following PRs)
- Changes ProviderConfig types with runtime.RawExtension and the spec for GardenerResourceData.
- Some fixes for the extension resources sync controller:
    - Now a sync is triggered even if the extension CR's state is nil and newly added
    - It will properly handle backupentry custom resources which are cluster scoped

Part of #1631 

**Release note**:

```improvement operator
The `ShootState` resource is now created at the beginning of the shoot reconciliation flow.
```
```improvement developer
Changed the specification of GardenerResourceData to use `runtime.RawExtension` instead of a `string` map to persist data required for deploying resources.
```
```improvement operator
Cluster scoped extension resources are now properly handled when their state has to be synced to the `ShootState`.
```


pkg/apis/core/v1alpha1/helper/shootstate_lister.go
```diff
@@ -0,0 +1,64 @@
...
+		if obj.Kind == kind && apiequality.Semantic.DeepEqual(obj.Name, name) && apiequality.Semantic.DeepEqual(obj.Purpose, purpose) {
+			*e = append((*e)[:i], (*e)[i+1:]...)
+		}
+	}
+}
+
+// Upsert either inserts or updates an already existing ExtensionResourceState with kind, name and purpose in the list
+func (e *ExtensionResourceStateList) Upsert(extesnsionResourceState *gardencorev1alpha1.ExtensionResourceState) {
+	for i, obj := range *e {
+		if obj.Kind == extesnsionResourceState.Kind && apiequality.Semantic.DeepEqual(obj.Name, extesnsionResourceState.Name) && apiequality.Semantic.DeepEqual(obj.Purpose, extesnsionResourceState.Purpose) {
```
danielfoehrKn:
```suggestion
		if obj.Kind == extensionResourceState.Kind && apiequality.Semantic.DeepEqual(obj.Name, extensionResourceState.Name) && apiequality.Semantic.DeepEqual(obj.Purpose, extensionResourceState.Purpose) {
```

danielfoehrKn:
```suggestion
		if obj.Kind == extensionResourceState.Kind && apiequality.Semantic.DeepEqual(obj.Name, extensionResourceState.Name) && apiequality.Semantic.DeepEqual(obj.Purpose, extensionResourceState.Purpose) {
```

pkg/apis/core/v1alpha1/helper/shootstate_lister.go
```diff
@@ -0,0 +1,64 @@
...
+
+package helper
+
+import (
+	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+)
+
+// GetExtensionResourceStateList returns a list of ExtensionResourceStates which can be modified.
+func GetExtensionResourceStateList(shootState *gardencorev1alpha1.ShootState) ExtensionResourceStateList {
```
rfranzke: I don't understand the purpose of this "helper" function - how does it help? 😕 😆 Should we remove it?

pkg/apis/core/v1alpha1/helper/shootstate_lister.go
```diff
@@ -0,0 +1,64 @@
...
+	return ExtensionResourceStateList(shootState.Spec.Extensions)
+}
+
+// SetExtensionResourceStateList sets the provided ExtensionResourceStateList to the ShootState.
+func SetExtensionResourceStateList(shootState *gardencorev1alpha1.ShootState, extensionsList ExtensionResourceStateList) {
```
rfranzke: Same here - why do we need such a function if can directly do what you do in line 29 instead of invoking it?

pkg/apis/core/v1alpha1/helper/shootstate_lister.go
```diff
@@ -0,0 +1,64 @@
...
+	shootState.Spec.Extensions = extensionsList
+}
+
+// ExtensionResourceStateList is a list of ExtensionResourceStates
+type ExtensionResourceStateList []gardencorev1alpha1.ExtensionResourceState
+
+// Get retrieves an ExtensionResourceState for given kind, name and purpose from a list of ExtensionResourceStates
+// If no ExtensionResourceStates can be found, nil is returned.
+func (e *ExtensionResourceStateList) Get(kind string, name, purpose *string) *gardencorev1alpha1.ExtensionResourceState {
+	for _, obj := range *e {
```
rfranzke:
```suggestion
	for _, obj := range e {
```
?

pkg/apis/core/v1alpha1/helper/shootstate_lister.go
```diff
@@ -0,0 +1,64 @@
...
+		if obj.Kind == kind && apiequality.Semantic.DeepEqual(obj.Name, name) && apiequality.Semantic.DeepEqual(obj.Purpose, purpose) {
+			return &obj
+		}
+	}
+	return nil
+}
+
+// Delete removes an ExtensionResourceState from the list by kind, name and purpose
+func (e *ExtensionResourceStateList) Delete(kind string, name, purpose *string) {
+	for i, obj := range *e {
```
rfranzke:
```suggestion
	for i, obj := range e {
```

?

pkg/apis/core/v1alpha1/helper/shootstate_lister.go
```diff
@@ -0,0 +1,64 @@
...
+		if obj.Kind == kind && apiequality.Semantic.DeepEqual(obj.Name, name) && apiequality.Semantic.DeepEqual(obj.Purpose, purpose) {
+			*e = append((*e)[:i], (*e)[i+1:]...)
+		}
+	}
+}
+
+// Upsert either inserts or updates an already existing ExtensionResourceState with kind, name and purpose in the list
+func (e *ExtensionResourceStateList) Upsert(extesnsionResourceState *gardencorev1alpha1.ExtensionResourceState) {
```
rfranzke:
```suggestion
func (e *ExtensionResourceStateList) Upsert(extensionResourceState *gardencorev1alpha1.ExtensionResourceState) {
```

pkg/gardenlet/controller/federatedseed/shootstate/common_sync.go
```diff
@@ -93,61 +88,65 @@ func (s *SyncController) createShootStateSyncReconcileFunc(ctx context.Context,
...
 
-	if currentExtensionState == nil {
-		shootState.Spec.Extensions = append(shootState.Spec.Extensions, gardencorev1alpha1.ExtensionResourceState{
-			Kind:    kind,
-			Name:    &name,
-			Purpose: purpose,
-			State:   gardencorev1alpha1.ProviderConfig{RawExtension: *extensionState},
-		})
+func fromRequest(req reconcile.Request) (projectNamespace, shootName string) {
+	// If the resource is cluster scoped then we need to get the projectName and shootName from the resource's name
```
rfranzke: Hmm, can you elaborate a bit here? The only resource where this is possible is the `Cluster` resource (which itself doesn't store any state that should be saved). Why do we need this?

pkg/gardenlet/controller/federatedseed/shootstate/common_sync.go
```diff
@@ -93,61 +88,65 @@ func (s *SyncController) createShootStateSyncReconcileFunc(ctx context.Context,
...
+	})
+	gardencorev1alpha1helper.SetExtensionResourceStateList(shootState, list)
+	return nil
+}
+
+func purposeToString(purpose *string) string {
+	if purpose == nil {
+		return "<nil>"
+	}
+	return string(*purpose)
```
rfranzke:
```suggestion
	return*purpose
```

pkg/operation/operation.go
```diff
@@ -430,6 +432,30 @@ func (o *Operation) SyncClusterResourceToSeed(ctx context.Context) error {
 	})
 }
 
+// EnsureShootStateExists creates the ShootState resource for the corresponding shoot and sets its ownerReferences to the Shoot.
+func (o *Operation) EnsureShootStateExists(ctx context.Context) error {
```
rfranzke, line 423: Can you write a small unit test for this function?

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -213,6 +213,9 @@ func (c *Controller) deleteShoot(shoot *gardencorev1beta1.Shoot, logger *logrus.
 		utilerrors.ToExecute("Check and sync Shoot with Seed", func() error {
 			return c.checkSeedAndSyncClusterResource(shoot, o)
 		}),
+		utilerrors.ToExecute("Ensure ShootState resource for Shoot", func() error {
```
rfranzke: What was the motivation to put this here instead of to the top of the files https://github.com/gardener/gardener/blob/master/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go and https://github.com/gardener/gardener/blob/master/pkg/gardenlet/controller/shoot/shoot_control_delete.go ? I think this rather belongs there.

rfranzke (changes requested):

pkg/gardenlet/controller/shoot/shoot_control.go
plkokanov: My thoughts were that the ShootState is a pretty important component and adding it here means that we won't have to add additional task dependency for every step that would require it in the flow graph.

Tbh I mostly did it because of `c.checkSeedAndSyncClusterResource()` which I was wondering why it was called here (it just does an additional check) and then the cluster resource is synced one more time at the start of the reconciliation flow graph

pkg/gardenlet/controller/federatedseed/shootstate/common_sync.go
plkokanov: The backupentry is also cluster scoped. Currently we also sync its state.

pkg/apis/core/v1alpha1/helper/shootstate_lister.go
plkokanov: `e` is a pointer to the slice `(*e)` is the actual slice, so we need that to be able to iterate over it. Go doesn't handle this automatically.

pkg/apis/core/v1alpha1/helper/shootstate_lister.go
plkokanov: Was wondering a lot whether I should just use explicit conversion :D in the calling code.

pkg/gardenlet/controller/shoot/shoot_control.go
rfranzke: The `Cluster` resource is special because it contains the `.status.lastOperation` of the `Shoot` which might be set to `Failed`. Extension controllers react on this and don't do anything in this case. The gardenlet does neither, i.e., it does not execute any flow, however, it has to sync the `Cluster` object to the seed so that the extensions get to know the status as well.

The `ShootState` belongs in the flow graph, I guess. Yes, it is important, but not "that important" that we require special handling like for the `Cluster` object.

pkg/apis/core/v1alpha1/helper/shootstate_lister.go
rfranzke: Hm, okay, I see. Thanks.

pkg/gardenlet/controller/federatedseed/shootstate/common_sync.go
rfranzke: Okay, thank you. Maybe add this to the comment so that people can get an idea why this is needed.

danielfoehrKn:
I noticed that the ShootState resource used is from core.gardener.cloud/v1alpha1 api version. Is there a reason why there is no  core.gardener.cloud/v1beta1 representation?

rfranzke:
It was not yet promoted to `v1beta1`. We should probably promote it soon now that it exists for some time, but not via this PR.

plkokanov:
Changed the way we retrieve the shoot name and namespace due to @timuthy linked issue and brought back what we had done initially - getting the shoot name and namespace from the cluster resource

pkg/gardenlet/controller/federatedseed/shootstate/shootstate_sync_control.go
```diff
@@ -172,3 +185,37 @@ func (sca *syncControllerArtifacts) shutdownQueues() {
...
+		return nil, fmt.Errorf("could not get cluster with name %s : %v", clusterName, err)
+	}
+
+	shootInternal := &core.Shoot{}
+	shoot := &gardencorev1beta1.Shoot{}
+
+	if cluster.Spec.Shoot.Raw == nil {
+		return nil, fmt.Errorf("cluster resource %s doesn't contain shoot resource in raw format", clusterName)
+	}
+	if _, _, err := s.Decode(cluster.Spec.Shoot.Raw, nil, shootInternal); err != nil {
```
rfranzke: I don't understand - why do you first decode into the internal version and directly after it convert to the `v1beta1`? Can't you directly decode into the `v1beta1`?

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -190,6 +190,11 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation, errorContext *er
 			Fn:   flow.TaskFn(botanist.SyncClusterResourceToSeed).RetryUntilTimeout(defaultInterval, defaultTimeout),
 		})
 
+		_ = g.Add(flow.Task{
```
rfranzke, line 193: Just to double check: Is it correct that no other task is depending on this one?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -105,6 +105,10 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 			Name: "Syncing shoot cluster information to seed",
 			Fn:   flow.TaskFn(botanist.SyncClusterResourceToSeed).RetryUntilTimeout(defaultInterval, defaultTimeout),
 		})
+		_ = g.Add(flow.Task{
```
rfranzke, line 108: Just to double check: Is it correct that no other task is depending on this one?

pkg/gardenlet/controller/shoot/shoot_control_delete.go
plkokanov: Yep, I will add the dependant tasks with following PRs. Same for below.

pkg/gardenlet/controller/federatedseed/shootstate/shootstate_sync_control.go
plkokanov: Copy pasta from the gardener-extenions repo.
I think the idea is if the Shoot is a different version than v1beta1 (e.g. v1beta2) then we won't have to immediately change the code here. But the conversion will only make sense if I extract this method in a utils package and make the gardener-extensions reuse it. Otherwise whoever is promoting the shoot version can update this code as well.

rfranzke: For the extensions it's different because they have no control over the version that is used for the shoot. For gardener, the same controller is writing the shoot in a specific version in the `Cluster` resource, hence, it can make assumptions which version to expect and doesn't need this.

rfranzke (dismissed):
/lgtm

pkg/apis/core/v1alpha1/helper/shootstate_lister.go
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
ialidzhikov: Wouldn't be better to name `shootstate_list.go` instead of `shootstate_lister.go`. I would rather consider it as client-go lister when I see lister suffix of the file. WDYT?

pkg/gardenlet/controller/federatedseed/shootstate/shootstate_sync_control.go
```diff
@@ -172,3 +184,31 @@ func (sca *syncControllerArtifacts) shutdownQueues() {
...
 }
+
+// ShootRetriever decodes raw Shoots into objects
+type ShootRetriever struct {
+	runtime.Decoder
+}
+
+func newShootRetriever() *ShootRetriever {
+	return &ShootRetriever{serializer.NewCodecFactory(gardenscheme).UniversalDecoder()}
+
```
ialidzhikov:
```suggestion
```

pkg/apis/core/v1alpha1/helper/shootstate_lister.go
plkokanov: makes sense

pkg/apis/core/v1alpha1/helper/shootstate_list.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+	}
+	return nil
+}
+
+// Delete removes an ExtensionResourceState from the list by kind, name and purpose
+func (e *ExtensionResourceStateList) Delete(kind string, name, purpose *string) {
+	for i, obj := range *e {
+		if obj.Kind == kind && apiequality.Semantic.DeepEqual(obj.Name, name) && apiequality.Semantic.DeepEqual(obj.Purpose, purpose) {
+			*e = append((*e)[:i], (*e)[i+1:]...)
```
ialidzhikov, line 40: This actually won't delete if slice element is duplicated in a row? If we want to remove only the first match, then we could early exit after the first removal?

Code snippet in smaller context:

```golang
package main

import "fmt"

func main() {
	s := []int{1, 2, 3, 3, 4, 5, 6, 7, 8, 9, 9, 10}

	shouldRemove := func(x int) bool { return x%3 == 0 }

	for i, x := range s {
		if shouldRemove(x) {
			s = append(s[:i], s[i+1:]...)
		}
	}

	fmt.Println(s)
        // Outputs: [1 2 3 4 5 7 8 9 10]
}
```

pkg/apis/core/v1alpha1/helper/shootstate_list_test.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
+	. "github.com/gardener/gardener/pkg/apis/core/v1alpha1/helper"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var _ = Describe("ShootState Data Lister", func() {
```
ialidzhikov:
```suggestion
var _ = Describe("ExtensionResourceStateList", func() {
```

pkg/apis/core/v1alpha1/helper/shootstate_list_test.go
```diff
@@ -0,0 +1,119 @@
...
+	It("should return a list of ExtensionStateResources correctly", func() {
+		list := ExtensionResourceStateList(shootState.Spec.Extensions)
+		Expect(len(list)).To(Equal(1))
+		Expect(list[0].Kind).To(Equal(extensionKind))
+		Expect(list[0].Name).To(Equal(extensionName))
+		Expect(list[0].Purpose).To(Equal(extensionPurpose))
+		Expect(list[0].State).To(Equal(extensionData))
+	})
+
+	It("should delete the extension resource state when it can be found", func() {
```
ialidzhikov: I would suggest also to group the tests for given func in a Context/Describe such as:
```golang
	Context("#Get", func() {
		It("should return the correct extension resource state", func() {})

		It("should return nil if extension resource state cannot be found", func() {})
	})

	Context("#Delete", func() {
		It("should delete the extension resource state when it can be found", func() {})

		It("should do nothing if extension resource state cannot be found", func() {})
	})

	Context("#Upsert", func() {
		It("should append new extension resource state to the list", func() {})

		It("should update an extension resource state in the list if it already exists", func() {})
	})
```

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -105,6 +105,10 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
...
+			Name: "Ensuring that ShootState exists",
+			Fn:   flow.TaskFn(botanist.EnsureShootStateExists).RetryUntilTimeout(defaultInterval, defaultTimeout),
```
ialidzhikov, line 110: Don't we need a dependency to the `syncClusterResourceToSeed` task as `createShootStateSyncReconcileFunc` will try to fetch the Cluster resource from the Seed?

plkokanov: The clusterResource is synced before that in [shoot_control.go: 330](https://github.com/gardener/gardener/blob/master/pkg/gardenlet/controller/shoot/shoot_control.go#L330). 
The function returned from `createShootStateSyncReconcileFunc` should only be executed if an extension resource already exists -> and that generally already requires the cluster resource to exist.

ialidzhikov: Ah, ok. Thank you!

pkg/apis/core/v1alpha1/helper/shootstate_list.go
plkokanov: Generally this list should only contain unique entries. The Upsert functionality takes care of that.
But you're right I don't see why not return after finding a match (no reason to loop the whole list)

ialidzhikov (approved):
/lgtm

plkokanov:
Thanks a lot for the reviews :)

rfranzke (approved):

PR #1978, opened 2020-02-26
Sync etcd encryption config from garden cluster

plkokanov:
This PR makes it so that if there is no etcd encryption secret in the Shoot cluster's control plane, but an encryption config has already been created it is synced from the Shoot's namespace in the garden cluster.

This is a quick temporary fix as it will later on be handled by the ShootState.

Fixes #

This might be necessary soon (@timuthy) and I am not sure when the feature to keep ETCD encryption config data in the ShootState will be merged.

**Release note**:

```improvement operator
Shoot's ETCD encryption configuration is synced from the Garden cluster if it does not exist in the Shoot's control plane
```

### DIFF
--- a/pkg/operation/botanist/etcdencryption.go
+++ b/pkg/operation/botanist/etcdencryption.go
@@ -27,6 +27,7 @@ import (
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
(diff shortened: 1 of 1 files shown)


pkg/operation/botanist/etcdencryption.go
```diff
@@ -41,18 +42,54 @@ import (
...
+
+	// Create the secret in the Shoot's seed namespace from the garden secret only if it doesn't exist to prevent overwriting.
+	secret := b.encryptionSecretFromGardenSecret(gardenSecret)
+	if err := b.K8sSeedClient.Client().Create(ctx, secret); err != nil && !apierrors.IsAlreadyExists(err) {
+		return err
+	}
+	return nil
+}
+
+func (b *Botanist) encryptionSecretFromGardenSecret(gardenSecret *corev1.Secret) *corev1.Secret {
```
timuthy: Can you make it a func without the Botanist pointer receiver and just pass the `SeedNamespace` additionally?

pkg/operation/botanist/etcdencryption.go
```diff
@@ -41,18 +42,54 @@ import (
 //
 // To mitigate data loss to a certain degree, the secret is also synced to the Garden cluster.
 func (b *Botanist) ApplyEncryptionConfiguration(ctx context.Context) error {
-	conf, err := b.createOrUpdateEncryptionConfiguration(ctx)
+	if err := b.syncEncryptionConfigurationFromGardenCluster(ctx); err != nil {
+		return err
+	}
+	conf, forcePlainTextSecrets, err := b.createOrUpdateEncryptionConfiguration(ctx)
```
timuthy: Instead of returning another type, can you just return the whole secret and pass it to `b.syncEncryptionConfigurationToGarden`?

timuthy (changes requested):
Thanks a lot for the fix. Looks good in general, just two concerns.

rfranzke (approved):

timuthy (approved):

PR #1980, opened 2020-02-27
Isolate dns entry

mvladev:

Isolate `DNSProvider` and `DNSEntry` lifecyle operations. Creation/deletetion/waiting/waitin for deletion is now a seprate package with complete isolation from Botanist. 

Fixes #

~~This is still WIP, because it's based on #1969 and it'll needs to be rebased on top of #1959.~~

**Release note**:

```improvement operator
NONE
```


rfranzke:
What about this PR?

mvladev:
`DNSProvider` and `DNSEntey` being separated into a separate  components (#2051) now with tests.

mvladev:
Hmm.. CI is running tests on draft PR @ccwienk

mvladev:
This PR is ready for review @timuthy @rfranzke

mvladev:
I found a bug in the reconciliation flow under specific conditions. I'll fix it and then update the PR.

mvladev:
The issue is fixed now.

charts/seed-dns/entry/templates/dnsentry.yaml
```diff
@@ -5,8 +5,8 @@ metadata:
   name: {{ .Values.name }}
   namespace: {{ .Release.Namespace }}
 spec:
-  {{- if .Values.ownerId }}
-  ownerId: {{ .Values.ownerId }}
+  {{- if .Values.ownerID }}
+  ownerID: {{ .Values.ownerID }}
```
ialidzhikov: The field name from external-dns-management side should be `ownerId`, right?
See https://github.com/gardener/external-dns-management/blob/0b4f9e1217d052f506056c0287db5f4347c9413f/pkg/apis/dns/v1alpha1/dnsentry.go#L45

```suggestion
  ownerId: {{ .Values.ownerID }}
```

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -426,8 +430,17 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation, errorContext *er
...
-			Fn:           flow.TaskFn(botanist.DestroyExternalDomainDNSRecord).DoIf(dnsEnabled),
+			Name: "Destroying external domain DNS record",
+			Fn: flow.TaskFn(func(ctx context.Context) error {
+				return botanist.ExternalDNS().Destroy(ctx)
+			}),
+			Dependencies: flow.NewTaskIDs(syncPointCleaned),
+		})
+		destroyInternalDomainDNSRecord = g.Add(flow.Task{
+			Name: "Destroying internal domain",
+			Fn: flow.TaskFn(func(ctx context.Context) error {
```
ialidzhikov: You can use directly the target fn (without the wrapper one) , right?

```suggestion
			Fn:           botanist.InternalDNS().Destroy,
```

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -426,8 +430,17 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation, errorContext *er
 			Dependencies: flow.NewTaskIDs(destroyInfrastructure),
 		})
 		destroyExternalDomainDNSRecord = g.Add(flow.Task{
-			Name:         "Destroying external domain DNS record",
-			Fn:           flow.TaskFn(botanist.DestroyExternalDomainDNSRecord).DoIf(dnsEnabled),
+			Name: "Destroying external domain DNS record",
+			Fn: flow.TaskFn(func(ctx context.Context) error {
```
ialidzhikov:
```suggestion
			Fn:           botanist.ExternalDNS().Destroy,
```

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -16,6 +16,8 @@ package shoot
 
 import (
 	"context"
+	"fmt"
+	"github.com/gardener/gardener/pkg/operation/botanist/component"
```
ialidzhikov: You could move below to the g/g import group.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -132,19 +132,18 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 			}(),
 		})
 		deployInternalDomainDNSRecord = g.Add(flow.Task{
-			Name:         "Deploying internal domain DNS record",
-			Fn:           flow.TaskFn(botanist.DeployInternalDomainDNSRecord).DoIf(dnsEnabled && managedInternalDNS && !o.Shoot.HibernationEnabled),
+			Name: "Deploying internal domain DNS record",
+			Fn: flow.TaskFn(func(ctx context.Context) error {
```
ialidzhikov:
```suggestion
			Fn:           flow.TaskFn(botanist.InternalDNS().Deploy).DoIf(!o.Shoot.HibernationEnabled),
```

pkg/operation/botanist/addons_test.go
```diff
@@ -0,0 +1,181 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/helm/pkg/engine"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	. "github.com/gardener/gardener/test/framework"
```
ialidzhikov: This pkg was initially added for integration tests and there are mainly integration test utilities. I wouldn't include it in g/g unit tests. I would rather place under g/g if there are some utilities which are used across multiple tests. cc @rfranzke

pkg/operation/botanist/dns.go
```diff
@@ -17,296 +17,316 @@ package botanist
...
+		b.Shoot.Components.DNS.ExternalEntry,
+		b.Shoot.Components.DNS.ExternalProvider,
+	)
 }
 
-// DeployExternalDomainDNSRecord deploys the DNS record for the external cluster domain.
-func (b *Botanist) DeployExternalDomainDNSRecord(ctx context.Context) error {
-	if b.Shoot.Info.Spec.DNS == nil || b.Shoot.Info.Spec.DNS.Domain == nil || b.Shoot.ExternalClusterDomain == nil || strings.HasSuffix(*b.Shoot.ExternalClusterDomain, ".nip.io") {
-		return nil
+// InternalDNS returns the external DNSProvider which deploys / destroys
```
ialidzhikov:
```suggestion
// InternalDNS returns the internal DNSProvider which deploys / destroys
```

pkg/operation/botanist/dns.go
```diff
@@ -17,296 +17,316 @@ package botanist
...
+		b.ChartsRootPath,
+		b.Logger,
+		seedClient,
+		nil,
+	))
+}
+
+// AdditionalDNSProviders returns a map containing a map of DNSProviders with their name as key.
+// Providers which are no longer needed / or in use, DeployWaiter which removes said DNSProvider.
+func (b *Botanist) AdditionalDNSProviders(gardenClient, seedClient client.Client) (map[string]component.DeployWaiter, error) {
```
ialidzhikov: You have a get call with `context.TODO()` inside this func. You can contextify this func.

```suggestion
func (b *Botanist) AdditionalDNSProviders(ctx context.Context, gardenClient, seedClient client.Client) (map[string]component.DeployWaiter, error) {
```

pkg/operation/botanist/dns/dnsentry.go
```diff
@@ -0,0 +1,133 @@
...
+
+package dns
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"time"
+
+	dnsv1alpha1 "github.com/gardener/external-dns-management/pkg/apis/dns/v1alpha1"
```
ialidzhikov, line 23: Can you organize the imports so they mach the convention - 1. golang pkgs, 2. gardener/gardener pkgs, 3. external ones to gardener/gardener.

ialidzhikov (changes requested):
I tried to play with this PR. The reconciliation of existing Shoot cluster stuck at `Initializing connection to Shoot` and my internal dnsentry got deleted.
```
$ k -n shoot--foo--bar get dnsentries
NAME       DNS                     TYPE          PROVIDER                   STATUS   AGE
external   api.bar.foo.[omitted]   aws-route53   shoot--foo--bar/external   Ready    36h
```

Can you check that?

Also can you rebase to origin master as there is #2088 - small improvement in the local setup.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
mvladev: Some of those DNSEntries can change their values therefore (by default they are not set ad only set once the APIServer address is correct) the function should be executed only when it's actually needed and not pass reference

pkg/operation/botanist/addons_test.go
mvladev: I've moved everything in `test/gomega`

pkg/operation/botanist/controlplane.go
```diff
@@ -292,13 +292,6 @@ func (b *Botanist) HibernateControlPlane(ctx context.Context) error {
 			return err
 		}
 
-		if err := flow.Parallel(
```
ialidzhikov, line 293: Recently https://github.com/gardener/gardener/pull/2100 was merged to delete the dnsentries for hibernated Shoots.
After hibernation with this PR, they are not deleted:
```
$ k -n shoot--foo--bar get dnsentries
external   api.foo.bar.example.com   aws-route53   shoot--foo--bar/external   Ready    74m
internal   api.foo.bar.example.com   aws-route53   shoot--foo--bar/internal   Ready    74m
```

Could you please adapt?

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -449,32 +458,32 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation, errorContext *er
...
 			Dependencies: flow.NewTaskIDs(deleteNamespace),
 		})
-
-		f = g.Compile()
 	)
+
+	for k, v := range botanist.Shoot.Components.DNS.Additional {
+		_ = g.Add(flow.Task{
+			Name:         fmt.Sprintf("Removing additional DNSProvider %q", k),
+			Fn:           flow.TaskFn(component.OpWaiter(v).Deploy).DoIf(wakeupRequired),
```
ialidzhikov: Is everything as expected here? The task name is `Removing additional DNSProvider %q` but the invoked func is `Deploy`, not `Destroy`.

ialidzhikov (changes requested):

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
ialidzhikov: Can you workaround it with
```suggestion
			Fn:           flow.TaskFn(botanist.DeployInternalDNS).DoIf(!o.Shoot.HibernationEnabled),
```

and
```golang
func (b *Botanist) DeployInternalDNS(ctx context.Context) error {
	return b.InternalDNS().Deploy(ctx)
}
```
?

pkg/operation/botanist/dns.go
```diff
@@ -17,295 +17,316 @@ package botanist
...
+		b.ChartsRootPath,
+		b.Logger,
+		seedClient,
+		nil,
+	))
+}
+
+// DefaultInternalDNSEntry returns DeployWaiter which removes the internal DNSEntry.
+func (b *Botanist) DefaultInternalDNSEntry(seedClient client.Client) component.DeployWaiter {
+	return component.OpDestroy(dns.NewDNSEntry(
```
ialidzhikov, line 200: Isn't the naming in `pkg/operation/botanist/component/deploy.go` too confusing? `OpDestroy` returns `DeployWaiter` which actually is created with `invert: true` and will do `Destroy`, instead of `Deploy`. Isn't `OperationWaiter` a better name?

pkg/operation/botanist/dns.go
```diff
@@ -17,295 +17,316 @@ package botanist
...
+		b.Shoot.SeedNamespace,
+		b.ChartApplierSeed,
+		b.ChartsRootPath,
+		b.Logger,
+		seedClient,
+		nil,
+	))
+}
+
+// AdditionalDNSProviders returns a map containing a map of DNSProviders with their name as key.
```
ialidzhikov:
```suggestion
// AdditionalDNSProviders returns a map containing DNSProviders where the key is the provider name.
```

pkg/operation/botanist/dns/dnsentry.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	return nil
+}
+
+func (d *dnsEntry) WaitCleanup(ctx context.Context) error {
+	return kutil.WaitUntilResourceDeleted(ctx, d.client, d.entry(), 5*time.Second)
+}
+
+// entry returs a empty DNSEntry used for deletion.
```
ialidzhikov:
```suggestion
// entry returns an empty DNSEntry used for deletion.
```

pkg/operation/shoot/types.go
```diff
@@ -46,13 +47,35 @@ type Shoot struct {
...
 	MachineDeployments        []extensionsv1alpha1.MachineDeployment
 }
 
+// Components contains different components deployed
+type Components struct {
+	DNS   *DNS
+	Nginx *Nginx
+}
+
+// DNS contains references to inter and external DNSProvider and DNSEntry deployers.
```
ialidzhikov:
```suggestion
// DNS contains references to internal and external DNSProvider and DNSEntry deployers.
```

ialidzhikov (changes requested):

pkg/operation/botanist/controlplane.go
mvladev: I've updated the PR with commit 2ae56c8 - now this functionality is included.

rfranzke:
@timuthy as you recently touched the whole DNS stuff - can you take a look here, please? :)

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -437,44 +425,64 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation, errorContext *er
...
-			waitUntilInfrastructureDeleted,
-		)
-
 		destroyInternalDomainDNSRecord = g.Add(flow.Task{
-			Name:         "Destroying internal domain DNS record",
-			Fn:           flow.TaskFn(botanist.DestroyInternalDomainDNSRecord).DoIf(dnsEnabled),
-			Dependencies: flow.NewTaskIDs(syncPoint),
+			Name:         "Destroying internal DNS entry",
+			Fn:           flow.TaskFn(component.OpWaiter(botanist.Shoot.Components.DNS.InternalEntry).Destroy),
+			Dependencies: flow.NewTaskIDs(syncPointCleaned, destroyNginxIngressDNSRecord),
```
timuthy: Why this is `destroyNginxIngressDNSRecord` a dependency? I know it was before due to `syncPoint` but does it still make sense?

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -437,44 +425,64 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation, errorContext *er
...
-			Fn:           botanist.WaitUntilSeedNamespaceDeleted,
-			Dependencies: flow.NewTaskIDs(deleteNamespace),
-		})
+		}),
+	}
+
+	for k, v := range botanist.Shoot.Components.DNS.Additional {
+		dnsProviders = append(dnsProviders, g.Add(flow.Task{
+			Name:         fmt.Sprintf("Destroying additional DNS provider %q", k),
+			Fn:           flow.TaskFn(component.OpWaiter(v).Destroy).DoIf(wakeupRequired),
```
timuthy: This example demonstrates a drawback which the design currently has. Before we needed only one deletion step/call to remove all DNS providers at [once](https://github.com/gardener/gardener/blob/a09ca85a330dfc1abcc2c09a91ae96a616def16a/pkg/operation/botanist/dns.go#L168) and now we need multiple because it is assumed that deploying and destroying is synchronous.

Additionally I think the `wakeupRequired` is not correct here because in the shoot reconcile flow providers are deployed right after the the KAS service is ready.

https://github.com/gardener/gardener/pull/1980/files#diff-a83ea876d4c50b98a963ab312faafde9R376

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -437,44 +425,64 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation, errorContext *er
...
-		deleteDNSProviders = g.Add(flow.Task{
-			Name:         "Deleting DNS providers",
-			Fn:           flow.TaskFn(botanist.DeleteDNSProviders).DoIf(dnsEnabled),
+	)
+
+	dnsProviders := flow.TaskIDSlice{
+		g.Add(flow.Task{
+			Name:         "Destroying External DNS provider",
+			Fn:           flow.TaskFn(component.OpWaiter(botanist.Shoot.Components.DNS.ExternalProvider).Destroy),
+			Dependencies: flow.NewTaskIDs(destroyExternalDomainDNSRecord),
```
timuthy: Since Gardener cannot directly influence which DNS provider controls which entry, every provider deletion should probably have a dependencies to the deletion of every entry, i.e. `destroyExternalDomainDNSRecord`, `destroyExternalDomainDNSRecord`, `destroyNginxIngressDNSRecord`.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -277,11 +272,16 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
...
-			Name:         "Ensuring ingress DNS record",
-			Fn:           flow.TaskFn(botanist.EnsureIngressDNSRecord).DoIf(dnsEnabled && managedExternalDNS && !o.Shoot.HibernationEnabled).RetryUntilTimeout(defaultInterval, 10*time.Minute),
+		nginxLBReady = g.Add(flow.Task{
+			Name:         "Waiting until nginx ingress LoadBalancer is ready",
+			Fn:           flow.TaskFn(botanist.WaitUntilNginxIngressServiceIsReady).DoIf(botanist.Shoot.NginxIngressEnabled()),
 			Dependencies: flow.NewTaskIDs(deployManagedResources),
 		})
+		_ = g.Add(flow.Task{
+			Name:         "Deploying nginx ingress domain record",
+			Fn:           flow.TaskFn(botanist.DeployNginxIngressDNSEntry),
```
timuthy: I'm not sure yet what to thing about this deploy logic. It was kind of confusing first because you don't immediately see what's going on:
1. Why doesn't it have a `DoIf(botanist.Shoot.NginxIngressEnabled())` like the depending step?
1. It is called "deploy" but due to the [OpDestroy](https://github.com/gardener/gardener/blob/15252f52140816dbf3c67a8009c7160242ebe99e/pkg/operation/botanist/component/deploy.go#L21) function a deploy sometimes also means that it is destroyed.

Elegant and concise, no doubts but also confusing for the reader. Maybe the other @gardener/gardener-maintainers can add their thoughts?! At least a better naming would help.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -307,6 +307,16 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 			Fn:           flow.TaskFn(botanist.HibernateControlPlane).RetryUntilTimeout(defaultInterval, 2*time.Minute).DoIf(o.Shoot.HibernationEnabled),
 			Dependencies: flow.NewTaskIDs(initializeShootClients, deploySeedMonitoring, deploySeedLogging, deployClusterAutoscaler),
 		})
+		_ = g.Add(flow.Task{
+			Name:         "Destroying External DNS Entry",
```
timuthy: W.r.t. my previous comment: Now you use a destroy function directly and take the decision in the flow whether to execute this step or not.

pkg/operation/shoot/types.go
```diff
@@ -46,13 +47,35 @@ type Shoot struct {
...
+	Nginx *Nginx
+}
+
+// DNS contains references to internal and external DNSProvider and DNSEntry deployers.
+type DNS struct {
+	ExternalProvider component.DeployWaiter
+	ExternalEntry    component.DeployWaiter
+	InternalProvider component.DeployWaiter
+	InternalEntry    component.DeployWaiter
+	Additional       map[string]component.DeployWaiter
```
timuthy:
```suggestion
	AdditionalProviders       map[string]component.DeployWaiter
```

pkg/operation/botanist/botanist.go
```diff
@@ -53,10 +56,22 @@ func New(o *operation.Operation) (*Botanist, error) {
...
+	o.Shoot.Components.DNS.InternalProvider = b.DefaultInternalDNSProvider(b.K8sSeedClient.Client())
+	o.Shoot.Components.DNS.InternalEntry = b.DefaultInternalDNSEntry(b.K8sSeedClient.Client())
+
+	o.Shoot.Components.DNS.Additional, err = b.AdditionalDNSProviders(context.TODO(), b.K8sGardenClient.Client(), b.K8sSeedClient.Client())
+	if err != nil {
+		return nil, err
+	}
+
+	o.Shoot.Components.Nginx.DNSEntry = b.DefaultNginxIngressDNSEntry(b.K8sSeedClient.Client())
+
```
timuthy: Can you do this initialization when the operation is [created](https://github.com/gardener/gardener/blob/d70ed0f5d1a6e3523bd2cde07f5e707bec417dbf/pkg/operation/operation.go#L62)? The `Components` should be useable w/o the botanist which also prevents nil pointer dereferences.

timuthy (changes requested):
Hey @mvladev, I really appreciate your PR and the segregation efforts. I've left a few questions / comments.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
ialidzhikov: I had the same comment about the naming. Ref https://github.com/gardener/gardener/pull/1980#discussion_r400883023.

timuthy: Oh sorry @ialidzhikov I missed that :(

rfranzke: +1, agreed as well. I'm also missing the `dnsEnabled` check.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
rfranzke: +1, a little bit confusing. Also, because suddenly the `Destroy` task is called (while the `Deploy`) task was called earlier. I see that this is the case where the shoot gets hibernated, though, maybe we can add a comment somewhere to make this clear.

pkg/operation/botanist/controlplane.go
```diff
@@ -222,6 +222,14 @@ func (b *Botanist) WakeUpControlPlane(ctx context.Context) error {
...
 	}
 
+	if err := b.DeployInternalDNS(ctx); err != nil {
+		return err
+	}
+
+	if err := b.DeployExternalDNS(ctx); err != nil {
+		return err
+	}
+
```
rfranzke, line 230: What about the ingress DNS?

pkg/operation/botanist/dns/dns_suite_test.go
```diff
@@ -0,0 +1,58 @@
...
+	. "github.com/onsi/gomega"
+)
+
+func TestDns(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Dns Suite")
+}
+
+const (
+	chartsRootPath = "../../../../charts"
```
rfranzke: use `filepath.Join`?

pkg/operation/botanist/dns/dnsentry.go
```diff
@@ -0,0 +1,134 @@
...
+	client    crclient.Client
+	waiter    retry.Ops
+}
+
+func (d *dnsEntry) Deploy(ctx context.Context) error {
+	return d.Apply(ctx, d.chartPath, d.shootNamespace, d.values.Name, kubernetes.Values(d.values))
+}
+
+func (d *dnsEntry) Destroy(ctx context.Context) error {
+	if err := d.client.Delete(ctx, d.entry()); err != nil && !apierrors.IsNotFound(err) {
```
rfranzke:
```suggestion
	return client.IgnoreNotFound(d.client.Delete(ctx, d.entry()))
```

pkg/operation/botanist/dns/dnsprovider.go
```diff
@@ -0,0 +1,141 @@
...
+	client    crclient.Client
+	waiter    retry.Ops
+}
+
+func (d *dnsProvider) Deploy(ctx context.Context) error {
+	return d.Apply(ctx, d.chartPath, d.shootNamespace, d.values.Name, kubernetes.Values(d.values))
+}
+
+func (d *dnsProvider) Destroy(ctx context.Context) error {
+	if err := d.client.Delete(ctx, d.provider()); err != nil && !apierrors.IsNotFound(err) {
```
rfranzke:
```suggestion
	return client.IgnoreNotFound(d.client.Delete(ctx, d.provider()))
```

pkg/operation/botanist/waiter.go
```diff
@@ -51,7 +51,25 @@ func (b *Botanist) WaitUntilKubeAPIServerServiceIsReady(ctx context.Context) err
 			// TODO(AC): This is a quite optimistic check / we should differentiate here
 			return retry.MinorError(fmt.Errorf("kube-apiserver service deployed in the Seed cluster is not ready: %v", err))
 		}
-		b.Operation.APIServerAddress = loadBalancerIngress
+		b.SetAPIServerEdgeAddress(loadBalancerIngress)
```
rfranzke: What is an edge address?

pkg/operation/operation.go
```diff
@@ -115,6 +115,7 @@ func newOperation(
 		K8sGardenClient:        k8sGardenClient,
 		K8sGardenCoreInformers: k8sGardenCoreInformers,
 		ChartApplierGarden:     kubernetes.NewChartApplier(renderer, applier),
+		ChartsRootPath:         common.ChartPath,
```
rfranzke: Why is this needed when we have the `common.ChartPath` constant?

pkg/operation/shoot/shoot.go
```diff
@@ -85,6 +85,11 @@ func New(k8sGardenClient kubernetes.Interface, k8sGardenCoreInformers gardencore
 		WantsClusterAutoscaler: false,
 
 		Extensions: extensions,
+
+		Components: &Components{
```
rfranzke: I really like this! :)

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -437,44 +425,64 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation, errorContext *er
...
+		waitUntilControlPlaneDeleted,
+		waitUntilControlPlaneExposureDeleted,
+		destroyNginxIngressDNSRecord,
+		destroyExternalDomainDNSRecord,
+		destroyInternalDomainDNSRecord,
+		waitUntilInfrastructureDeleted,
+		dnsProviders,
 	)
+
+	deleteNamespace := g.Add(flow.Task{
```
rfranzke: use a new `var () ` block - improves readability

pkg/gardenlet/controller/shoot/shoot_control_delete.go
rfranzke: +1, I don't think this dependency is needed

rfranzke (changes requested):
Sorry for joining late.. cool PR, I like some ideas behind it but also have doubts, mainly those already detected by the other colleagues.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
mvladev: > Why doesn't it have a DoIf(botanist.Shoot.NginxIngressEnabled()) like the depending step?

It's not needed - by default, when `botanist.DeployNginxIngressDNSEntry` is called it would actually try to remove the dns entry.

Only if DNS is enabled and after the LB in the Shoot cluster gets its IP then the correct value is set https://github.com/gardener/gardener/blob/1e120c562fbfe786a2ef24d509dae4f07d27e40d/pkg/operation/botanist/addons.go#L81-L97

pkg/operation/botanist/controlplane.go
mvladev: Nginx Ingress is not needed for successful deletion?

pkg/operation/operation.go
mvladev: Because for tests, this won't work.

pkg/gardenlet/controller/shoot/shoot_control_delete.go
timuthy: Please explain why `destroyNginxIngressDNSRecord` is still a dependency?

mvladev:
Have another look at the PR. I've fixed lots of the comments.

pkg/operation/botanist/botanist.go
mvladev: The seed client is not initialized before, so it's not possible without moving the client initialization as well.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
timuthy: > It's not needed - by default, when botanist.DeployNginxIngressDNSEntry is called it would actually try to remove the dns entry.

This concern wasn't about necessity but about readability. The step is called "deploy" (not only in the code, also shown in the status and log file) and you usually expect something to be deployed. It's confusing if you go down the call stack and realize that this actual "deploy" step destroys resources. I think @ialidzhikov and @rfranzke have a similar opinion.

`ProcessIngressDNSRecord`, `EnsureIngressDNSRecord` or something else would be better, imo.

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -458,13 +446,13 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation, errorContext *er
 		)
 
 		destroyInternalDomainDNSRecord = g.Add(flow.Task{
-			Name:         "Destroying internal domain DNS record",
-			Fn:           flow.TaskFn(botanist.DestroyInternalDomainDNSRecord).DoIf(dnsEnabled),
-			Dependencies: flow.NewTaskIDs(syncPoint),
+			Name:         "Destroying internal DNS entry",
+			Fn:           flow.TaskFn(component.OpWaiter(botanist.Shoot.Components.DNS.InternalEntry).Destroy),
+			Dependencies: flow.NewTaskIDs(syncPoint, destroyNginxIngressDNSRecord),
```
timuthy:
```suggestion
			Dependencies: flow.NewTaskIDs(syncPoint),
```

`destroyNginxIngressDNSRecord` is part of `syncPoint`

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -373,9 +383,18 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
...
 			Dependencies: flow.NewTaskIDs(deployKubeControllerManager, deployControlPlane, deployControlPlaneExposure),
 		})
-		f = g.Compile()
 	)
 
+	for k, v := range botanist.Shoot.Components.DNS.AdditionalProviders {
+		_ = g.Add(flow.Task{
+			Name:         fmt.Sprintf("Ensuring additional DNSProvider %q", k),
+			Fn:           flow.TaskFn(component.OpWaiter(v).Deploy).DoIf(!o.Shoot.HibernationEnabled),
+			Dependencies: flow.NewTaskIDs(waitUntilKubeAPIServerServiceIsReady),
```
timuthy:
```suggestion
			Dependencies: flow.NewTaskIDs(deployInternalDomainDNSRecord, deployExternalDomainDNSRecord, deployIngressDomainDNSRecord),
```

Best effort to let the internal/external provider create the internal/external records (dns records do not have a ref spec to a provider).

timuthy (changes requested):

test/gomega/deep_derivative.go
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
schrodit: we should move that into the testframework `test/framework/gomega`

test/gomega/kubernetes_not_found.go
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
schrodit: we should move that into the testframework `test/framework/gomega`

test/gomega/matches.go
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
schrodit: we should move that into the testframework `test/framework/gomega`

ialidzhikov:
/needs/rebase

timuthy (commented):
After another round of review I don't have any essential comments about this PR. @mvladev can you please double check the unresolved comments and rebase? Afterwards, I suggest to test this change again in our local setup and finally merge it. /cc @rfranzke @ialidzhikov

ialidzhikov:
/reviewed/do-not-merge

ialidzhikov:
Hold until v1.4.0 is released.

rfranzke:
/kind/enhancement

rfranzke:
/needs/review
/needs/rebase

mvladev:
rebased

rfranzke:
/area/networking
/size/xl
/topology/seed
/needs/review
/platform/all
/priority/normal

mvladev:
@timuthy @rfranzke @ialidzhikov can you have another look?

mvladev:
bump

pkg/operation/shoot/types.go
```diff
@@ -69,6 +72,26 @@ type Shoot struct {
 	ETCDEncryption *etcdencryption.EncryptionConfig
 }
 
+// Components contains different components deployed
+type Components struct {
+	DNS   *DNS
+	Nginx *Nginx
```
rfranzke: Hm, why is `Nginx` separate attribute and not part of the `DNS` struct? Wouldn't it be more consistent to either
a) add it to the `DNS` struct as well 
b) separate external, internal, and additional DNS into separate structs?

pkg/operation/botanist/dns/dnsprovider.go
```diff
@@ -0,0 +1,137 @@
...
+
+		status = provider.Status.State
+		if msg := provider.Status.Message; msg != nil {
+			message = *msg
+		}
+
+		d.logger.Infof("Waiting for %q DNS provider to be ready... (status=%s, message=%s)", d.values.Name, status, message)
+		return retry.MinorError(fmt.Errorf("DNS provider %q is not ready (status=%s, message=%s)", d.values.Name, status, message))
+	}); err != nil {
+		return gardencorev1beta1helper.DetermineError(err, fmt.Sprintf("Failed to create DNS provider for %q DNS record: %q (status=%s, message=%s)", d.values.Name, err.Error(), status, message))
```
rfranzke: FYI: [This change](https://github.com/gardener/gardener/pull/2336/files#diff-33d2a8cacfcd2dda84699a11862f3ab0) should be reflected here as well as #2336 will probably be merged first.

pkg/operation/botanist/dns/dnsentry.go
```diff
@@ -0,0 +1,130 @@
...
+
+		status = entry.Status.State
+		if msg := entry.Status.Message; msg != nil {
+			message = *msg
+		}
+
+		d.logger.Infof("Waiting for %q DNS record to be ready... (status=%s, message=%s)", d.values.Name, status, message)
+		return retry.MinorError(fmt.Errorf("DNS record %q is not ready (status=%s, message=%s)", d.values.Name, status, message))
+	}); err != nil {
+		return gardencorev1beta1helper.DetermineError(err, fmt.Sprintf("Failed to create %q DNS record: %q (status=%s, message=%s)", d.values.Name, err.Error(), status, message))
```
rfranzke, line 121: Same comment like below when waiting for the DNSProvider to become ready.

rfranzke (commented):
Minor comments, generally OK, need to play with it a little bit, but I agree that we should get it in this week. @gardener/gardener-maintainers - any objections?

mvladev:
I've moved Nginx to the DNS struct and  #2336

rfranzke (commented):
Thanks Martin, no further comments now, will play with it a little bit and come back.

rfranzke (approved):
I've tested several scenarios and all of them worked perfectly.
/lgtm

PR #1983, opened 2020-02-27
Improve VPA recommendations

wyb1:
Sets the VPA recommender flag `--recommendation-margin-fraction` to `0.05`. Default value is `0.15`. This flag adds a buffer to the recommendation based on the value that is set. This means that the VPA should now make recommendations that are 10% lower. I tested this out and it lowered the recommendations for some containers, specifically prometheus, elasticsearch-logging and gardener-resource-manager, by about 10% (as expected). Below are the recommendations (CPU millicores) for these containers before and after this flag change.
Screen Shot 2020-02-27 at 13 59 04
The rest of the containers were not affected by this change because they were either already getting the minimum recommendation from VPA or HVPA. [Kube-apiserver](https://github.com/gardener/gardener/blob/master/charts/seed-controlplane/charts/kube-apiserver/templates/kube-apiserver-hvpa.yaml#L78) and [etcd](https://github.com/gardener/gardener/blob/master/charts/seed-controlplane/charts/etcd/templates/etcd-hvpa.yaml#L74) get specific minimum recommendations set by HVPA. Perhaps we can change this HVPA setting for the kube-apiserver and etcd since they decrease our utilization the most. @ggaurav10 @amshuman-kr 

Fixes #

`--history-length=24h` flag has been removed since it does nothing.

**Release note**:

```improvement operator
Vertical Pod Autoscaler provides 10% lower recommendations to improve resource utilization.
```
```improvement operator
HVPA `minAllowed` values for kube-apiserver and etcd have been lowered to improve resource utilization.
```

### DIFF
--- a/charts/seed-controlplane/charts/etcd/templates/etcd-hvpa.yaml
+++ b/charts/seed-controlplane/charts/etcd/templates/etcd-hvpa.yaml
@@ -67,11 +67,11 @@ spec:
           containerPolicies:
             - containerName: etcd
               maxAllowed:
-                memory: 30G
-                cpu: "4"
+                memory: {{ .Values.hvpa.maxAllowed.memory }}
+                cpu:  {{ .Values.hvpa.maxAllowed.cpu }}
               minAllowed:
-                memory: 1000M
-                cpu: 300m
+                memory: {{ .Values.hvpa.minAllowed.memory }}
+                cpu: {{ .Values.hvpa.minAllowed.cpu }}
             - containerName: backup-restore
               mode: "Off"
   weightBasedScalingIntervals:
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -1050,9 +1050,6 @@ func (b *Botanist) DeployETCD(ctx context.Context) error {
 			"checksum/secret-etcd-server-tls": b.CheckSums["etcd-server-tls"],
 			"checksum/secret-etcd-client-tls": b.CheckSums["etcd-client-tls"],
 		},
-		"hvpa": map[string]interface{}{
-			"enabled": hvpaEnabled,
-		},
 		"maintenanceWindow": b.Shoot.Info.Spec.Maintenance.TimeWindow,
 		"storageCapacity":   b.Seed.GetValidVolumeSize("10Gi"),
 	}
@@ -1067,6 +1064,23 @@ func (b *Botanist) DeployETCD(ctx context.Context) error {
 		if role == common.EtcdRoleMain {
 			// etcd-main emits extensive (histogram) metrics
 			etcd["metrics"] = "extensive"
+			etcd["hvpa"] = map[string]interface{}{
+				"enabled": hvpaEnabled,
+				"minAllowed": map[string]interface{}{
+					"cpu":    "200m",
+					"memory": "700M",
+				},
+			}
+		}
+
+		if role == common.EtcdRoleEvents {
+			etcd["hvpa"] = map[string]interface{}{
+				"enabled": hvpaEnabled,
+				"minAllowed": map[string]interface{}{
+					"cpu":    "50m",
+					"memory": "200M",
+				},
+			}
 		}
 
 		foundEtcd := true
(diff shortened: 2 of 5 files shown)


timuthy (dismissed):
/lgtm

ggaurav10:
> --history-length=24h flag has been removed since it does nothing.

According to my understanding, VPA recommender accumulates the usage samples for this length of time, and builds histogram models for CPU and memory usage. The default value is 8 days, however, the reason for decreasing it to 24 hours was that we didn't want to consider any usage that could have been higher more than 24 hours ago to be considered in new recommendations.

This would also prevent cases of erroneous recommendations when a shoot with same name is created under a same project, since VPA uses labels to gather usage data and for such cases labels/namespace remains the same.
Please correct me if I have missed something here. If not can we please keep history length 24 hours, and reassess this parameter when we see any adverse affect.

wyb1:
@ggaurav10 you are correct that the default is 8 days but this is only if Prometheus is chosen as the historical data provider. Since no prometheus address is given to the recommender it ignores this setting. See [here](https://github.com/kubernetes/autoscaler/blob/6a929ab7e9f556c6e6bb11c8cd393476d798f8c5/vertical-pod-autoscaler/pkg/recommender/main.go#L67) in the code of the recommender. Therefore `--history-length` is currently ignored and does nothing.

ggaurav10 (dismissed):

timuthy:
@wyb1 can you force push to trigger the cla assistant? :(

wyb1:
@timuthy I will change the minAllowed set by hvpa after syncing with @ggaurav10 This should also improve our utilization and I think also belongs in this PR. It should also retrigger the cla assistant :)

wyb1:
I reduced the minAllowed values for kube-apiserver and etcd. PTAL @ggaurav10

ggaurav10 (dismissed):

charts/seed-controlplane/charts/etcd/templates/etcd-hvpa.yaml
```diff
@@ -69,9 +69,16 @@ spec:
               maxAllowed:
                 memory: 30G
                 cpu: "4"
+{{- if eq .Values.role "main"}}
```
timuthy: Can you do this in the code instead because there we already have the distinction of main and events.
https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/controlplane.go#L1067

wyb1: Added in the code. PTAL @timuthy

pkg/operation/botanist/controlplane.go
```diff
@@ -1067,6 +1064,27 @@ func (b *Botanist) DeployETCD(ctx context.Context) error {
...
 			etcd["metrics"] = "extensive"
+			etcd["hvpa"] = map[string]interface{}{
+				"hvpa": map[string]interface{}{
+					"enabled": hvpaEnabled,
+					"minAllowed": map[string]interface{}{
+						"cpu":    "200m",
+						"memory": "700M",
+					},
+				},
+			}
```
ggaurav10:
```suggestion
			etcd["hvpa"] = map[string]interface{}{
					"enabled": hvpaEnabled,
					"minAllowed": map[string]interface{}{
						"cpu":    "200m",
						"memory": "700M",
					},
			}
```

pkg/operation/botanist/controlplane.go
```diff
@@ -1067,6 +1064,27 @@ func (b *Botanist) DeployETCD(ctx context.Context) error {
...
+		if role == common.EtcdRoleEvents {
+			etcd["hvpa"] = map[string]interface{}{
+				"hvpa": map[string]interface{}{
+					"enabled": hvpaEnabled,
+					"minAllowed": map[string]interface{}{
+						"cpu":    "50m",
+						"memory": "200M",
+					},
+				},
+			}
```
ggaurav10: same as above

ggaurav10 (changes requested):

ggaurav10 (approved):

timuthy (approved):
/lgtm

PR #1988, opened 2020-02-28
Fix merge apply functions for Service

mvladev:

The `Service` applier was not handling all use-cases around services:

- Headless `ClusterIP` services  (`.spec.clusterIP=1.2.3.4` ->  `.spec.clusterIP=None`)
- `ClusterIP`/`NodePort`/`LoadBalancer` <=> `ExternalName` type
- `LoadBalancer`/`NodePort` -> `ClusterIP`
- `.spec.healthCheckNodePort` should only be defaulted to the previous value if both the new and old `Service` are of type `LoadBalancer` with  `.spec.externalTrafficPolicy: Local` and the current value of the new service's `.spec.healthCheckNodePort` is `0`

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/client/kubernetes/apply.go
+++ b/pkg/client/kubernetes/apply.go
@@ -121,32 +121,82 @@ func (c *Applier) deleteObject(ctx context.Context, desired *unstructured.Unstru
 var DefaultApplierOptions = ApplierOptions{
 	MergeFuncs: map[schema.GroupKind]MergeFunc{
 		corev1.SchemeGroupVersion.WithKind("Service").GroupKind(): func(newObj, oldObj *unstructured.Unstructured) {
-			// We do not want to overwrite a Service's `.spec.clusterIP', `.spec.healthCheckNodePort` or '.spec.ports[*].nodePort' values.
-			oldPorts := oldObj.Object["spec"].(map[string]interface{})["ports"].([]interface{})
-			newPorts := newObj.Object["spec"].(map[string]interface{})["ports"].([]interface{})
-			ports := []map[string]interface{}{}
-
-			// Check whether ports of the newObj have also been present previously. If yes, take the nodePort
-			// of the existing object.
-			for _, newPort := range newPorts {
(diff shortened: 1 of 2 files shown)


rfranzke:
Thanks, can you also file a PR to https://github.com/gardener/gardener-resource-manager/blob/master/pkg/controller/managedresources/merger.go?

mvladev:
> Thanks, can you also file a PR to https://github.com/gardener/gardener-resource-manager/blob/master/pkg/controller/managedresources/merger.go?

Yes, I'll do.

pkg/client/kubernetes/apply.go
```diff
@@ -121,32 +121,82 @@ func (c *Applier) deleteObject(ctx context.Context, desired *unstructured.Unstru
...
+				_ = unstructured.SetNestedField(newObj.Object, newSvcType, "spec", "type")
+			}
+
+			oldSvcType, found, _ := unstructured.NestedString(oldObj.Object, "spec", "type")
+			if !found {
+				oldSvcType = string(corev1.ServiceTypeClusterIP)
+
+			}
+
+			switch newSvcType {
```
timuthy, line 136: Imo, no `switch` necessary because the second `case` in line 172 is obsolet if you use an `if` condition for the first case.

mvladev: Then I'll have to do 

```go
if newSvcType== string(corev1.ServiceTypeLoadBalancer) && 
  newSvcType==string(corev1.ServiceTypeNodePort) &&
  newSvcType!=string(corev1.ServiceTypeClusterIP) { ..
```

switch makes more sense in this case

mvladev:
@rfranzke PR in https://github.com/gardener/gardener-resource-manager/pull/35

timuthy (approved):
/lgtm

rfranzke (approved):
/lgtm

let's wait with the merge until v1.1 was cut (planned for today)

PR #1991, opened 2020-03-02
TokenRequest and TokenRequestProjection enabled by default

mvladev:

Both feature gates are now enabled by default for supported versions.

- `--api-audiences` now defaults to `kubernetes`.
- `--service-account-issuer` now defaults the external URI of the cluster.
- `--service-account-key-file` now also always includes the optional key which the user might set in `.spec.kubernetes.kubeAPIServer.serviceAccountConfig.signingKeySecretName.name`
- `--service-account-signing-key-file` now always defaults to same key used by `kube-controller-manager`

Fixes #

 `--api-audiences` defaulting to `kubernetes` means that end-users don't have to modify their `TokenRequest` if they want to have tokens for more than 1 audiences and on different clusters:

```json
{
    "apiVersion": "authentication.k8s.io/v1",
    "kind": "TokenRequest",
    "metadata": {
      "name": "foo",
      "namespace": "default"
    },
    "spec": {
      "audiences": [
        "kubernetes",
        "another-system"
      ],
      "boundObjectRef": {
        "apiVersion": "v1",
        "kind": "Pod",
        "name": "nginx",
        "uid": "f02dee11-e541-4414-b0cb-461152750c37"
      },
      "expirationSeconds": 6000
    }
  }
```

```
kubectl create  -f token-request.json --raw /api/v1/namespaces/default/serviceaccounts/default/token
```

resulting token:

```json
{
  "alg": "RS256",
  "kid": "8Kznq-rxIE-mu-92CIRP0Fz4HqH8-NVhZnc8M7i0N5Q"
}
{
  "aud": [
    "kubernetes",
    "another-system"
  ],
  "exp": 1583141054,
  "iat": 1583135054,
  "iss": "https://foo.bar",
  "kubernetes.io": {
    "namespace": "default",
    "pod": {
      "name": "nginx",
      "uid": "f02dee11-e541-4414-b0cb-461152750c37"
    },
    "serviceaccount": {
      "name": "default",
      "uid": "15d0fa8c-a79c-44cc-a51f-d9d8e6445fa1"
    }
  },
  "nbf": 1583135054,
  "sub": "system:serviceaccount:default:default",
  "iatStr": "2020-03-02T07:44:14Z",
  "expStr": "2020-03-02T09:24:14Z",
  "nbfStr": "2020-03-02T07:44:14Z"
}
```

**Release note**:

```noteworthy user
`TokenRequest` and `TokenRequestProjection` feature gates are now enabled by default. By default, service account tokens now have audience - `kubernetes` and issuer - the external address of the API server. This allows for [service account token volume projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection) to be used in the cluster. `.spec.kubernetes.kubeAPIServer.audiences` and `.spec.kubernetes.kubeAPIServer.serviceAccountConfig.issuer` can still be used to override those values if needed.
```


rfranzke (commented):
Thanks Martin, looks good. Can you rebase and squash the commits? And why are there 10k+ deleted lines in the openapi? I think something went wrong during the generation?

mvladev:
> And why are there 10k+ deleted lines in the openapi? I think something went wrong during the generation?

Everything looks okay. kubectl works, there is no complaint about the generation. 

Screenshot 2020-03-03 at 8 24 40

mvladev:
Actually the deleted lines are from `pkg/client/core/openapi/zz_generated.openapi.go` which should not be there in the first place

rfranzke:
If I run `make generate` I get the changes back. Why does this file exist? I can't remember that/when it was added..

mvladev:
It's not used anywhere so I've removed it. I don't see a reason for such generation only for one group.

rfranzke:
Can you rebase again?

mvladev:
Rebased to fix conflicts

rfranzke (approved):

PR #1992, opened 2020-03-02
Assign initial `requests` for kube-apiserver based on minNodes

ggaurav10:
Assign initial `requests` for kube-apiserver based on minNodes.
Currently Gardener assigns initial `requests` based on maxNodes. If a cluster is created with minNodes = 1 and maxNodes = 100, then we would end up allocating large values for `kube-apiserver` deployment. This was fine when we did not have vertical autoscaling capabilities for kube-apsierver. However, with HVPA, we can assign lower values initially, and depend on HVPA to scale the deployment whenever required.

Fixes #

**Release note**:

```improvement operator
Assign initial `requests` for kube-apiserver based on cluster autoscaler minNodes instead of maxNodes
```

### DIFF
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -762,7 +762,12 @@ func (b *Botanist) DeployKubeAPIServer() error {
 			defaultValues["replicas"] = 0
 		}
 
-		cpuRequest, memoryRequest, cpuLimit, memoryLimit := getResourcesForAPIServer(b.Shoot.GetMinNodeCount())
+		var cpuRequest, memoryRequest, cpuLimit, memoryLimit string
+		if hvpaEnabled {
+			cpuRequest, memoryRequest, cpuLimit, memoryLimit = getResourcesForAPIServer(b.Shoot.GetMinNodeCount())
+		} else {
+			cpuRequest, memoryRequest, cpuLimit, memoryLimit = getResourcesForAPIServer(b.Shoot.GetMaxNodeCount())
+		}
 		defaultValues["apiServerResources"] = map[string]interface{}{
 			"limits": map[string]interface{}{
 				"cpu":    cpuLimit,
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -762,7 +762,7 @@ func (b *Botanist) DeployKubeAPIServer() error {
 			defaultValues["replicas"] = 0
 		}
 
-		cpuRequest, memoryRequest, cpuLimit, memoryLimit := getResourcesForAPIServer(b.Shoot.GetNodeCount())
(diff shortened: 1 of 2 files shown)


dguendisch:
cc @schrodit @OlegLoewen : we need to monitor our testmachinery cluster once this change is rolled out. We intentionally put `maxNodes` on a high value (because back then no hvpa was there and otherwise the api server / etcd was sized too small for our needs).

wyb1 (commented):
Shouldn't this also change the `minAllowed` by HVPA? Since the current minAllowed will most likely be greater than the initial requests set here.

ggaurav10:
Currently the `minAllowed` is 400M and 300m for memory and cpu respectively, which is lower than what Gardener would set to the smallest cluster (800M, 800m).
IMO, this value should be fine so that in case, the usage drops after bootstrapping the cluster, kube-apiserver can still scale down.
WDYT?

wyb1:
> IMO, this value should be fine so that in case, the usage drops after bootstrapping the cluster, kube-apiserver can still scale down.


Since most clusters are small and use way less than the minAllowed amount, wouldn't this be a great way to dynamically set the minAllowed? This would greatly improve utilization IMO.

wyb1:
I get that Gardener sets the value higher than the minAllowed (800M, 800m), but couldn't this be changed to a much lower value?

ggaurav10:
> I get that Gardener sets the value higher than the minAllowed (800M, 800m), but couldn't this be changed to a much lower value?

> dynamically set the minAllowed?

I agree that even lower `minAllowed` should be helpful. However, I am not sure about the ROI on dynamically setting it.
VPA recommendations are anyway based on usage, and `minAllowed` is a way to low-cap it. We can keep it as low as possible keeping in mind that random noise or usual apiserver's periodic maintenance tasks  don't result in scaling even for a smallest cluster. 
According to my understanding, for dynamic `minAllowed`, we would need to gather some data about the base resource usage for different cluster sizes.
This kind of t-shirt sizing was done when we had introduced `maxNodes` based resource allocation.
Please let me know if you have any ideas too for how to set the `minAllowed` dynamically.

timuthy:
I like the idea @wyb1 but let's get there gradually with careful observation.

rfranzke (approved):

timuthy (approved):
/lgtm
Let's merge it for now and follow up on what @wyb1 said. Okay @wyb1 @ggaurav10?

amshuman-kr:
Sorry for commenting after the merge. But let's avoid making too many scaling related changes in one single gardener release. We already have #1968 and #1983. If things go south it would be hard to figure out which change caused it.

Please do not misunderstand. I agree that we need to make multiple changes to improve resource utilisation. But let's iterate over incremental releases rather than clubbing many changes into bigger releases.

Issue #1844, opened 2020-01-24
Admission plugin should validate extension types in resources


rfranzke:
**What would you like to be added**:
We should have an admission controller that triggers for `CREATE` and `UPDATE` requests for `BackupBucket`, `Seed`, and `Shoot` resources. It should validate whether the extension types in the `Shoot` are actually registered in the system.

Concretely, 

* for `Shoot`s:
  * the `.spec.provider.type` must be part of the `.spec.resources[]` for `kind=Infrastructure`, `kind=ControlPlane`, `kind=Worker`
  * the `.spec.networking.type` must be part of the `.spec.resources[]` for `kind=Networking`
  * the `.spec.provider.workers[].machine.image.name` must be part of the `.spec.resources[]` for `kind=OperatingSystemConfig`
  * the `.spec.extensions[].type` must be part of the `.spec.resources[]` for `kind=Extension`,

* for `BackupBucket`s:
  * the `.spec.provider` must be  part of the `.spec.resources[]` for `kind=BackupBucket`, `kind=BackupEntry`,

* for `Seed`s:
  * the `.spec.backup.provider` must be  part of the `.spec.resources[]` for `kind=BackupBucket`, `kind=BackupEntry`

in the `ControllerRegistration`s.

**Why is this needed**:
`Shoot` resources with invalid/non-registered types should not be created as they can rather be reconciled nor deleted, and they unnecessarily pollute the system.

PR #1993, opened 2020-03-02
Only deploy required extensions

### LINKED ISSUE #1944: Only deploy required extension controllers to seeds
Over a year ago we have introduced the `ControllerRegistration` and `ControllerInstallation` resources that are used to register extension controllers and to describe installation demands of a specific registration to a specific seed.
The first, naive implementation was done in the interest of time and simply deploys every registered extension controller to every known seed cluster - regardless of whether the extension controller is required there.

Let's now enhance this implementation, i.e., only deploy those extensions to the seeds if needed. The solution should be dynamic, i.e., if an extension is no longer required it should be deleted again. 

**Why is this needed**:
Cost savings, minimal required deployments on seeds to reduce failure/error potential, optimized seed node utilization

rfranzke:
With this PR the `ControllerRegistration` controller in the gardener-controller-manager is entirely reworked. It is now more intelligent, i.e., it will only deploy those extension controller that are required to reconcile the `BackupBucket`s, `BackupEntry`s, `Shoot`s scheduled to the respective `Seed`s. If an extension controller is no longer required (maybe because all shoots were deleted) then the `ControllerRegistration` controller will automatically delete the extension.

Fixes #1944

* I added field selectors for the seed name to `BackupBucket` / `BackupEntry` resources for the `v1beta1` API version.
* I'm sometimes using actual LIST calls to the API server instead of listing from the cache as the latter one does not allow using field selectors.

**Release note**:

```noteworthy operator
Gardener will no longer deploy all extension controller to all seed clusters. Instead, it will dynamically deploy/delete them based on the scheduled `BackupBucket`s, `BackupEntry`s, and `Shoot`s.
```

### DIFF
--- a/pkg/controllermanager/controller/controllerregistration/controllerregistration.go
+++ b/pkg/controllermanager/controller/controllerregistration/controllerregistration.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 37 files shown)


pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -0,0 +1,452 @@
...
+	)
+
+	f := func() error {
+		kutil.SetMetaDataLabel(&controllerInstallation.ObjectMeta, common.SeedSpecHash, seedSpecHash)
+		kutil.SetMetaDataLabel(&controllerInstallation.ObjectMeta, common.RegistrationSpecHash, registrationSpecHash)
+		controllerInstallation.Spec = installationSpec
+		return nil
+	}
+
+	if installationName, ok := registrationNameToInstallationName[registrationName]; ok {
```
danielfoehrKn: To me this passage is a little hard to read - having the create/update operation + executing the mutating functions depending on if the ControllerInstallation already exists or not.
To me this is more clear WDYT:
``` golang
       if installationName, exists := registrationNameToInstallationName[registrationName]; exists {
		// installation already exists
		controllerInstallation.Name = installationName
	} else {
		// setting generated name as the installation does not exist yet
		controllerInstallation.GenerateName = controllerRegistration.Name + "-"
	}

	_, err = controllerutil.CreateOrUpdate(ctx, c, controllerInstallation, f)
	return err
```

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -0,0 +1,452 @@
...
+	for result := range out {
+		wantedKindTypeCombinations = wantedKindTypeCombinations.Union(result)
+	}
+
+	return wantedKindTypeCombinations
+}
+
+// computeControllerRegistrationMaps computes two maps. The first map maps the name of a ControllerRegistration to
+// the *gardencorev1alpha1.ControllerRegistration object. The second map maps the registered kind/type combinations
+// to the supporting *gardencorev1alpha1.ControllerRegistration object.
```
danielfoehrKn:
```suggestion
// to the supporting * gardencorev1beta1.ControllerRegistration object.
```

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -0,0 +1,452 @@
...
+
+	for result := range out {
+		wantedKindTypeCombinations = wantedKindTypeCombinations.Union(result)
+	}
+
+	return wantedKindTypeCombinations
+}
+
+// computeControllerRegistrationMaps computes two maps. The first map maps the name of a ControllerRegistration to
+// the *gardencorev1alpha1.ControllerRegistration object. The second map maps the registered kind/type combinations
```
danielfoehrKn:
```suggestion
// the * gardencorev1beta1.ControllerRegistration object. The second map maps the registered kind/type combinations
```

danielfoehrKn (changes requested):

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
rfranzke: Valid point and nice suggestion, I was doing it like this initially, though, the problem was that `CreateOrUpdate` does not supporting the creation for objects with `GenerateName`. I've added small comments to explain why it is like this, hope it is clear now.

danielfoehrKn (dismissed):

timebertt:
Can you please rebase?

rfranzke:
Why? I don’t see any conflicts.

timebertt:
There are no conflicts. But this PR is based on a version without `etcd-druid` and now I can't test it locally (or at least not easily) because the changes of the druid to the etcd sts don't seem to be backwards compatible.

rfranzke:
I can rebase it tomorrow (no access now). You can test it locally by rebasing locally after checking out the branch.

timebertt:
Sure, thanks :)

pkg/controllermanager/controller/controllerregistration/controllerregistration.go
```diff
@@ -137,21 +175,23 @@ func (c *Controller) Run(ctx context.Context, workers int) {
...
+	c.seedQueue.ShutDown()
 
 	for {
-		if c.controllerRegistrationQueue.Len() == 0 && c.seedQueue.Len() == 0 && c.numberOfRunningWorkers == 0 {
+		if c.controllerRegistrationQueue.Len() == 0 && c.seedQueue.Len() == 0 && c.controllerRegistrationSeedQueue.Len() == 0 && c.numberOfRunningWorkers == 0 {
 			logger.Logger.Debug("No running ControllerRegistration worker and no items left in the queues. Terminated ControllerRegistration controller...")
 			break
 		}
-		logger.Logger.Debugf("Waiting for %d ControllerRegistration worker(s) to finish (%d item(s) left in the queues)...", c.numberOfRunningWorkers, c.controllerRegistrationQueue.Len()+c.seedQueue.Len())
+		logger.Logger.Debugf("Waiting for %d ControllerRegistration worker(s) to finish (%d item(s) left in the queues)...", c.numberOfRunningWorkers, c.controllerRegistrationQueue.Len()+c.seedQueue.Len()+c.controllerRegistrationSeedQueue.Len()+c.controllerRegistrationSeedQueue.Len())
```
timebertt: Duplicated `c.controllerRegistrationSeedQueue.Len()`:
```suggestion
		logger.Logger.Debugf("Waiting for %d ControllerRegistration worker(s) to finish (%d item(s) left in the queues)...", c.numberOfRunningWorkers, c.controllerRegistrationQueue.Len()+c.seedQueue.Len()+c.controllerRegistrationSeedQueue.Len())
```

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+
+	if err := deployNeededInstallations(ctx, logger, c.k8sGardenClient.Client(), seed, wantedControllerRegistrationNames, controllerRegistrationNameToObject, registrationNameToInstallationName); err != nil {
+		return err
+	}
+	return deleteUnneededInstallations(ctx, logger, c.k8sGardenClient.Client(), wantedControllerRegistrationNames, registrationNameToInstallationName)
+}
+
+// computeKindTypesForBackupBucket computes the list of wanted kind/type combinations for extension resources based on the
+// the list of existing BackupBackup resources.
```
timebertt:
```suggestion
// the list of existing BackupBucket resources.
```

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -0,0 +1,458 @@
...
+func computeWantedControllerRegistrationNames(
+	wantedKindTypeCombinations sets.String,
+	kindTypeToControllerRegistrationName map[string]string,
+) (sets.String, error) {
+	names := sets.NewString()
+
+	for _, requiredExtension := range wantedKindTypeCombinations.UnsortedList() {
+		name, ok := kindTypeToControllerRegistrationName[requiredExtension]
+		if !ok {
+			return nil, fmt.Errorf("need to install an extension controller for %q but no appropriate ControllerRegistration found", requiredExtension)
```
timebertt, line 305: This early exit will break the whole controllerregistration reconciliation, if an unknown type is used in one of the Shoots/BackupBuckets/BackupEntries.
This way an end-user can easily break the reconciliation by for example creating a shoot with `spec.networking.type=foo` and new Shoots with a valid type which is not installed on the seed yet, will never get reconciled.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -0,0 +1,458 @@
...
+) (map[string]string, error) {
+	registrationNameToInstallationName := make(map[string]string)
+
+	for _, controllerInstallation := range controllerInstallationList.Items {
+		if controllerInstallation.Spec.SeedRef.Name != seedName {
+			continue
+		}
+
+		if _, ok := controllerRegistrationNameToObject[controllerInstallation.Spec.RegistrationRef.Name]; !ok {
+			return nil, fmt.Errorf("ControllerRegistration %q does not exist", controllerInstallation.Spec.RegistrationRef.Name)
```
timebertt, line 329: Not sure about this one.
If the referenced ControllerRegistration is (for some reason) not existing anymore, this will also block the whole controllerregistration reconciliation.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -0,0 +1,458 @@
...
+	c client.Client,
+	seed *gardencorev1beta1.Seed,
+	wantedControllerRegistrations sets.String,
+	controllerRegistrationNameToObject map[string]*gardencorev1beta1.ControllerRegistration,
+	registrationNameToInstallationName map[string]string,
+) error {
+	for _, registrationName := range wantedControllerRegistrations.UnsortedList() {
+		logger.Infof("Deploying wanted ControllerInstallation for %q", registrationName)
+
+		if err := deployNeededInstallation(ctx, c, seed, registrationName, controllerRegistrationNameToObject, registrationNameToInstallationName); err != nil {
```
timebertt: Why do you pass down the whole `controllerRegistrationNameToObject` map? Can't you simply pass down the needed registration object? Same for the installation map...

```suggestion
		if err := deployNeededInstallation(ctx, c, seed, registrationName, controllerRegistrationNameToObject[registrationName], registrationNameToInstallationName[registrationName]); err != nil {
```

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -0,0 +1,458 @@
...
+		controllerInstallation.Name = installationName
+		_, err := controllerutil.CreateOrUpdate(ctx, c, controllerInstallation, mutate)
+		return err
+	}
+
+	// The installation does not exist yet, hence, we set `GenerateName` which will automatically append a random suffix to
+	// the name. Unfortunately, the `CreateOrUpdate` function does not support creating an object that does not specify a name
+	// but only `GenerateName`, thus, we call `Create` directly.
+	controllerInstallation.GenerateName = controllerRegistration.Name + "-"
+	mutate()
```
timebertt:
```suggestion
	_ = mutate()
```

timebertt (changes requested):
Looks pretty good so far, really nice PR :)
I'm not completely through the PR, yet, though I left you some early comments.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
rfranzke: Exactly, this was done intentionally and will be prevented by #1844 (I'll start working on it next week). It should not even be possible to submit such a broken object to the API server.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
rfranzke: Yes, what would you expect? This is a system error that should never happen. Silently ignoring it and continuing with the rest would be wrong, I guess.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
timebertt: Yes, saw the issue. If we add the validation for extension types before the next release, then it's fine for me.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
timebertt: Agreed, this will anyways only happen if the operator removes the registration forcefully (i.e. manually removes the finalizer). So, I guess it's fine because the operator shouldn't need to do that in any case.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control_test.go
```diff
@@ -0,0 +1,687 @@
...
+			obj := &gardencorev1beta1.Seed{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: seedName,
+				},
+			}
+
+			c.controllerRegistrationSeedControl = &fakeControllerRegistrationSeedControl{}
+			c.seedLister = newFakeSeedLister(c.seedLister, obj, nil, nil)
+
+			Expect(c.reconcileControllerRegistrationSeedKey(seedName)).To(BeNil())
```
timebertt: Can you keep it consistent with the other `Expect`s?
```suggestion
			Expect(c.reconcileControllerRegistrationSeedKey(seedName)).NotTo(HaveOccurred())
```

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control_test.go
```diff
@@ -0,0 +1,687 @@
...
+			},
+		}
+	)
+
+	Describe("#computeKindTypesForBackupBuckets", func() {
+		It("should return empty results for empty input", func() {
+			kindTypes, buckets := computeKindTypesForBackupBuckets(nil, seedName)
+
+			Expect(kindTypes.Len()).To(BeZero())
+			Expect(buckets).To(HaveLen(0))
```
timebertt:
```suggestion
			Expect(buckets).To(BeEmpty())
```

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control_test.go
```diff
@@ -0,0 +1,687 @@
...
+		})
+
+		It("should correctly compute the result", func() {
+			kindTypes, buckets := computeKindTypesForBackupBuckets(backupBucketList, seedName)
+
+			Expect(kindTypes).To(Equal(sets.NewString(
+				extensionsv1alpha1.BackupBucketResource+"/"+backupBucket2.Spec.Provider.Type,
+				extensionsv1alpha1.BackupBucketResource+"/"+backupBucket3.Spec.Provider.Type,
+			)))
+			Expect(buckets).To(Equal(buckets))
```
timebertt:
```suggestion
			kindTypes, bs := computeKindTypesForBackupBuckets(backupBucketList, seedName)

			Expect(kindTypes).To(Equal(sets.NewString(
				extensionsv1alpha1.BackupBucketResource+"/"+backupBucket2.Spec.Provider.Type,
				extensionsv1alpha1.BackupBucketResource+"/"+backupBucket3.Spec.Provider.Type,
			)))
			Expect(bs).To(Equal(buckets))
```

pkg/controllermanager/controller/controllerregistration/controllerregistration.go
```diff
@@ -23,85 +23,109 @@ import (
...
 
 	controller := &Controller{
-		k8sGardenClient:               k8sGardenClient,
-		k8sGardenCoreInformers:        gardenCoreInformerFactory,
-		seedControl:                   NewDefaultSeedControl(k8sGardenClient, gardenCoreInformerFactory, recorder, config, controllerRegistrationLister, controllerInstallationLister, controllerRegistrationQueue),
-		controllerRegistrationControl: NewDefaultControllerRegistrationControl(k8sGardenClient, gardenCoreInformerFactory, recorder, config, seedLister, controllerRegistrationLister, controllerInstallationLister),
-		config:                        config,
-		recorder:                      recorder,
+		k8sGardenClient: k8sGardenClient,
+		secrets:         secrets,
```
timebertt: Seems like `Controller.secrets` is not used anywhere (only assigned here).
```suggestion
```

pkg/operation/botanist/botanist.go
```diff
@@ -64,32 +61,43 @@ func New(o *operation.Operation) (*Botanist, error) {
...
+		for _, kindType := range requiredExtensions.UnsortedList() {
+			split := strings.Split(kindType, "/")
+			if len(split) != 2 {
+				return fmt.Errorf("unexpected required extension: %q", kindType)
 			}
-			if extensionTypes.Len() == 0 {
-				delete(requiredExtensions, extensionKind)
+			extensionKind, extensionType := split[0], split[1]
+
+			if helper.IsResourceSupported(controllerRegistration.Spec.Resources, extensionKind, extensionType) && helper.IsControllerInstallationSuccessful(controllerInstallation) {
```
timebertt, line 99: As discussed, we should also check if the `ControllerInstallation` is healthy.

timebertt (changes requested):
Rest looks good as well.
Just some minor suggestions :)

timebertt (approved):
Looks perfect now, thanks!
/approve

PR #1994, opened 2020-03-02
external domain points to api server LB instead of internal domain

danielfoehrKn:

External domain points to api server Lb instead of internal domain.
Enables a  Shoot using a private hosted zone for the internal domain, to provide an external domain (e.g for the kubeconfig) of a custom dns provider. 
Current setup: External domain -> internal domain -> api server LB.   This does currently not work because the internal domain is not publicly resolvable (CNAME resolve will always fail).

Now: 
External domain  -> api server LB
Internal domain -> -> api server LB

Like today, requires public Apiserver LB of the Shoot.
Fixes #

**Release note**:

```improvement operator
External domain now points to API Server LoadBalancer instead of the internal domain. This is helpful when a Shoot uses a private hosted zone for the internal domain.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -144,7 +144,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 		deployExternalDomainDNSRecord = g.Add(flow.Task{
 			Name:         "Deploying external domain DNS record",
 			Fn:           flow.TaskFn(botanist.DeployExternalDomainDNSRecord).DoIf(dnsEnabled && managedExternalDNS && !o.Shoot.HibernationEnabled),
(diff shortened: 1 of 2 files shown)


rfranzke (dismissed):

rfranzke:
Rebase needed

rfranzke (approved):

timebertt (approved):

mvladev:
Then, why do we need 2 DNS records pointing to the same endpoint?

rfranzke:
One domain that is used internally in our system components (kubelet, kube-proxy, cloud-config-downloader, ...), and one domain that is used in the customer's/end-user's kubeconfig.

danielfoehrKn:
> Then, why do we need 2 DNS records pointing to the same endpoint?
@mvladev 
Internal Shoot communication can go over a private hosted zone while the kubeconfig contains the external domain.

PR #1998, opened 2020-03-03
Remove functionless DNS providers

timuthy:
Before we introduced the support for multiple DNS providers #1959 a shoot with the following DNS config was valid:
```
dns:
  domain: default.domain.example.com
  providers
  - type: aws-route53
```
Although the shoot defines a **default domain** it has a provider **without** a `secretName` which means that this provider is kept in the manifest but without any function.

With this PR, Gardener removes such functionless providers when shoots are created or updated. The automatic removal is done in favor of a validation to stay compatible.

A panic in the conversion logic for project roles has been fixes along the way.

**Release note**:

```action user
Since Gardener now supports multiple DNS providers it's not valid to let shoot use a **default** domain and a DNS provider without a `spec.dns.providers[0].secretName`. These **functionless** providers are removed by Gardener automatically, but will be forbidden in the future. Please make sure that you **don't use a DNS provider without** `spec.dns.providers[*].secretName`.
```


timebertt (approved):
/approve

danielfoehrKn (approved):

PR #1999, opened 2020-03-04
Do not duplicate `role` to `roles` in versioned `Project` objects

rfranzke:
This PR prevents that the `.spec.members[].role` field is duplicated to the `.spec.members[].roles[]` in versioned `Project `resources.
Otherwise, the API would not be backwards compatible as changing the `.spec.members[].role` is not enough anymore (because the same value is part of the `.spec.members[].roles[]` and must be removed there as well).

Example:

```
spec:
  members:
  - name: foo
    role: admin
    roles:
    - admin
    - viewer
```

Setting `.spec.members[0].role=viewer` does still preserve the `admin` role for the user because it exists in `.spec.members[0].roles[0]`.

This PR fixes this problem in the following way:
* During conversion from `{v1beta1,v1alpha1}` to `internal`, the `.spec.members[].role` is always added to the head of `.spec.members[].roles[]`
* During conversion from `internal` to `{v1beta1,v1alpha1}`, the head of `.spec.members[].roles[]` is used for `.spec.members[].role` and then truncated.

--> In the `internal` version, the `.spec.members[].roles` always contains all the roles.
--> In the `{v1beta1,v1alpha1}` versions the first role is part of `.spec.members[].role`, and all the others are part of `.spec.members[].roles[]`.

/cc @petersutter @timuthy 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/docs/extensions/project-roles.md
+++ b/docs/extensions/project-roles.md
@@ -22,10 +22,10 @@ spec:
   - apiGroup: rbac.authorization.k8s.io
     kind: User
     name: alice.doe@example.com
+    role: admin
     roles:
-    - "admin"
-    - "owner"
-    - "extension:foo"
+    - owner
+    - extension:foo
 ```
 
 The project controller will, for every extension role, create a `ClusterRole` with name `name: gardener.cloud:extension:project:<projectName>:<roleName>`, i.e., for above example: `name: gardener.cloud:extension:project:dev:foo`.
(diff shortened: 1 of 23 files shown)


timuthy (commented):
To my understanding this implies that we cannot deprecate the `role` field, right?

rfranzke:
Hm, yes, we can do the same like for the `.spec.owner` field - remove it as part of `core.gardener.cloud/v1`.

timuthy:
Okay let's do it this way. The new push will also make the license/cla check happy (hopefully) ;)

danielfoehrKn (dismissed):

timuthy (approved):
/lgtm

Issue #1564, opened 2019-10-30
☂️- Issue for Gardener Test Cases


schrodit:
### Motivation
Since Gardener has to manage hundreds or even thousands of clusters in full automation, there is no human involved anymore once deployed. A bug in the controller manager may have catastrophic effects on a productive landscape. Consequently, we must protect ourselves against this worst case scenario with a likewise fully automated output qualification.

The goal is that we can release gardener with confidence and ensure that a gardener installation is working correctly.
This issue should track necessary test cases and their priority in order to achieve this goal and improve the overall test coverage of gardener.

### Test Cases
- [x] Test the manual scale out of a worker in a running cluster (https://github.com/gardener/gardener/pull/1510)
- [x] Test a full reconciliation (maintenance and reconcile) of a shoot cluster (https://github.com/gardener/gardener/pull/1510)
- [x] Test hibernation (https://github.com/gardener/gardener/pull/1276)
- [x] Test wake up of a hibernated shoot (https://github.com/gardener/gardener/pull/1276)
- [x] Validate that rbac is enabled for the gardener cluster (https://github.com/gardener/gardener/pull/1139)
- [x] Test a kubernetes version update
  - [x] Test Minor version update (https://github.com/gardener/gardener/pull/1203)
  - [x] Test Patch Versions update
- [x] Test autoscaler scale up and down across all cloudproviders
- [ ] Test Project creation and deletion
  - [ ] Test Permissions on projects
- [ ] Test multizone cluster

ialidzhikov:
Not sure if it is the right issue, but I just wanted to suggest that we could also add test for  `.spec.kubeProxy.mode`. Afaik we currently do not test with `.spec.kubeProxy.mode: IPVS`, right?

timuthy:
Have you already thought of adding test cases for the different operating systems that we support on major cloud providers? I know the test matrix easily grows here but at least the most important permutations should be covered.

dguendisch:
> Not sure if it is the right issue, but I just wanted to suggest that we could also add test for .spec.kubeProxy.mode. Afaik we currently do not test with .spec.kubeProxy.mode: IPVS, right?

Right, we don't customize `kubeProxy:mode` yet, so we test whatever is the current default.
@ialidzhikov do you happen to have an idea on how much usage there is for `IPVS` currently?
[...]

PR #2000, opened 2020-03-04
Test creation and deletion of hibernated shoot

timebertt:
Co-authored-by: Tim Schrodi <tim.schrodi@sap.com>

This PR adds a flag `start-hibernated` to the `create-shoot` test which will set `spec.hibernation.enabled` to `true`.
It also adds `TestDef`s and Test Cases for hibernating and waking up a Shoot.
These additions can be used to test the creation and deletion of a hibernated Shoot.

Part of #1564 

**Release note**:

```improvement operator
NONE
```


OlegLoewen (dismissed):

test/framework/shootframework.go
```diff
@@ -190,6 +190,10 @@ func (f *ShootFramework) AddShoot(ctx context.Context, shootName, shootNamespace
 
 	f.Shoot = shoot
 
+	if f.Shoot.Spec.Hibernation != nil && *f.Shoot.Spec.Hibernation.Enabled {
```
danielfoehrKn: I think we need a null check for f.Shoot.Spec.Hibernation.Enabled here as well. 
 f.Shoot.Spec.Hibernation.Enabled can be null in the case that the flag "start-hibernated" is unset or false and the Shoot has a Schedule defined but the optional field enabled is not set. 

Or in the create test you make sure shootObject.Spec.Hibernation.Enabled is always set.

timebertt: Good catch, fixed :)

dguendisch (approved):

rfranzke (approved):

danielfoehrKn (approved):

PR #2009, opened 2020-03-05
Revert "Fix manifest sorting in chart renderer"

### LINKED ISSUE #2008: chart.FileContent requires 'manifest-%d' suffix
**What happened**:
In #1947 we add the `manifest-%d` suffix to the chart files to allow proper splitting and sorting of files that contain multiple manifests.
But this breaks the `FileContent` invocations such as:

```golang
	release, err := a.ChartRenderer().Render(chartPath, releaseName, namespace, values)
	if err != nil {
		panic(err)
	}

	foo := release.FileContent("foo.yaml") 
```

The manifest name is already set to `chart-name/templates/foo.yaml_manifest-0` but the `FileContent` tries to find `chart-name/templates/foo.yaml` which results in empty string.

See above.

See above.

**Environment**:

- Gardener version: v1.1.1
- Others:

ialidzhikov:
This reverts commit 67abd81c3c5875232bde1432ab1ee1db49118227.

Basically #1952 splits the kube-state-metrics charts into separate manifests - so it should not regress #1939.

Fixes #2008

**Release note**:

```improvement operator
An issue preventing to get the file content of manifest using the FileContent func of RenderedChart in the `pkg/chartrenderer` package is now fixed.
```

### DIFF
--- a/pkg/chartrenderer/default.go
+++ b/pkg/chartrenderer/default.go
@@ -28,7 +28,6 @@ import (
 	"k8s.io/helm/pkg/engine"
 	"k8s.io/helm/pkg/manifest"
 	chartapi "k8s.io/helm/pkg/proto/hapi/chart"
-	"k8s.io/helm/pkg/releaseutil"
 	"k8s.io/helm/pkg/timeconv"
 )
 
@@ -149,14 +148,7 @@ func (r *chartRenderer) renderResources(ch *chartapi.Chart, values chartutil.Val
 		}
 	}
 
-	manifestsMap := map[string]string{}
-	for origKey, file := range files {
-		for key, m := range releaseutil.SplitManifests(file) {
-			manifestsMap[origKey+"_"+key] = m
-		}
-	}
-
-	manifests := manifest.SplitManifests(manifestsMap)
+	manifests := manifest.SplitManifests(files)
 	manifests = SortByKind(manifests)
 
 	return &RenderedChart{


timebertt:
Thanks @ialidzhikov for catching this.
Where is the `FileContent` func used? I don't see any usage in `g/g` or `g/grm`. Is it used in the extensions?
If not, I would rather propose to remove the function instead of reverting the change.

If it is used somewhere, I would agree with you and go ahead with this PR.
I guess, then we would have to keep an eye on not putting multiple manifests into one tpl file, right? Otherwise, the manifest sorting will fail again.

ialidzhikov:
> Where is the FileContent func used? I don't see any usage in g/g or g/grm. Is it used in the extensions?

Yes, it is used in the extensions. Ref https://github.com/gardener/gardener-extension-provider-aws/blob/eef271f30c7e2abe03b5867b4e51111f1f9f9f86/pkg/controller/infrastructure/actuator_reconcile.go#L76-L78

> I guess, then we would have to keep an eye on not putting multiple manifests into one tpl file, right?  Otherwise, the manifest sorting will fail again.

Yes, right. We should put 1 manifest per file.

timebertt:
> Yes, it is used in the extensions.

Alright, got it, makes sense for terraformer.
Then let's go ahead and revert the change.

timebertt (approved):
/approve

PR #2010, opened 2020-03-05
Calculate default domain during seed assignment

timuthy:
Assign a default domain to a shoot not only when the shoot is created but also on update when the shoot is assigned to a seed.

/cc @ialidzhikov - thanks for reporting.

**Release note**:

```improvement operator
An issue has been fixed which prevented shoot clusters without a `spec.dns.domain` configuration from being scheduled on a seed.
```


ialidzhikov (approved):
Thank you for the fix!

/lgtm

rfranzke (approved):

PR #2019, opened 2020-03-09
Add  kubelet-preferred-address-types=InternalIP in metrics-server

jia-jerry:
Fix issue that metric-server can't scrape metrics of node in Alicloud. 
Fixes # https://github.com/gardener/gardener-extension-provider-alicloud/issues/39

It is not possibe to implement a webhook to change the deployment of metrics-server. Because the deployment is managed by gardener resource manager.
**Release note**:

```improvement operator
Metrics-server can scrape node's metrics now.
```

### DIFF
--- a/charts/shoot-core/components/charts/metrics-server/templates/deployment.yaml
+++ b/charts/shoot-core/components/charts/metrics-server/templates/deployment.yaml
@@ -45,6 +45,8 @@ spec:
         # nobody user only can write in home folder
         - --cert-dir=/home/certdir
         - --secure-port=8443
+        # Hostname can not be always resolved in all hyperscalers, such as Alibaba Cloud. Need to set internal IP
+        - kubelet-preferred-address-types=InternalIP
         # See https://github.com/kubernetes-incubator/metrics-server/issues/25 and https://github.com/kubernetes-incubator/metrics-server/issues/130
         # The kube-apiserver and the kubelet use different CAs, however, the metrics-server assumes the CAs are the same.
         # We should remove this flag once it is possible to specify the CA of the kubelet.


rfranzke (commented):
Why do we do this in g/g globally? I guess it’s rather something specific to Alicloud.

jia-jerry:
> Why do we do this in g/g globally? I guess it’s rather something specific to Alicloud.

Option 1. Alicloud provider extension webhook changes the metrics server deployment, the resource manager will reconcile it back. As the result, the pod will be destroyed and started again and agiain. 
Option 2. Alicloud provider extension webhook changes the secret managedresource-shoot-core. It is not easy to change it because the key of the data is not intuitive and it is not maintainable.

The currently one always works on  all hyperscalers.

rfranzke:
> Option 1. Alicloud provider extension webhook changes the metrics server deployment, the resource manager will reconcile it back. As the result, the pod will be destroyed and started again and agiain.

That's wrong, see https://github.com/gardener/gardener/blob/master/docs/extensions/shoot-webhooks.md. If you have a webhook on the `metrics-server` Deployment then it will not be reconciled back. It will always be injected.

> The currently one always works on all hyperscalers.

I'm fine changing it in general here in g/g, however, we need some confidence. Have you run the integration tests with all supported platforms? /cc @majst01 @Gerrit91 - would this work for you as well?

jia-jerry:
> That's wrong, see https://github.com/gardener/gardener/blob/master/docs/extensions/shoot-webhooks.md. If you have a webhook on the `metrics-server` Deployment then it will not be reconciled back. It will always be injected.
> 

OK. I made a mistake :)

charts/shoot-core/components/charts/metrics-server/templates/deployment.yaml
```diff
@@ -45,6 +45,8 @@ spec:
         # nobody user only can write in home folder
         - --cert-dir=/home/certdir
         - --secure-port=8443
+        # Hostname can not be always resolved in all hyperscalers, such as Alibaba Cloud. Need to set internal IP
+        - kubelet-preferred-address-types=InternalIP
```
Gerrit91, line 49: Isn't this missing a `--` option flag indicator?

Gerrit91:
I am not sure, if this is still something the committer wants to have added globally, we will try it out. But maybe he is already convinced to implement this as a provider extension webhook?

rfranzke:
@jia-jerry can you outline your plans? :)

jia-jerry:
> @jia-jerry can you outline your plans? :)

I will change it in the extension. Let's keep the defaut setting of the metrics server. Will close it. Thank you.

timuthy:
FYI: https://github.com/gardener/gardener/pull/2468

PR #2021, opened 2020-03-09
Fix an issue that occurs if the shoot does not get a seed assigned

schrodit:
Fixed a panic in the shoot framework if no seed was assigned.

Added the seed state to the test framework dump

**Release note**:

```improvement developer
A bug in the test framework has been fixed that panicked if no seed was assigned to the shoot.
```
```improvement developer
The current seed state was added to the test framework's state dump.
```


timebertt (approved):
/lgtm

PR #2022, opened 2020-03-10
[hotfix] Prevent CA from downscaling node running etcd pod

shreyas-s-rao:
Signed-off-by: Shreyas Rao <shreyas.sriganesh.rao@sap.com>

This hotfix PR adds annotation `"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"` to etcd pod (instead of etcd sts) so that cluster-autoscaler does not downscale any node running etcd pod.

Fixes #

This will cause the etcd pods to restart, but only in maintenance window.

**Release note**:

```improvement operator
Prevent cluster-autoscaler from downscaling any node running etcd pod.
```

### DIFF
--- a/charts/seed-controlplane/charts/etcd/templates/etcd-statefulset.yaml
+++ b/charts/seed-controlplane/charts/etcd/templates/etcd-statefulset.yaml
@@ -1,8 +1,6 @@
 apiVersion: {{ include "statefulsetversion" . }}
 kind: StatefulSet
 metadata:
-  annotations:
-    "cluster-autoscaler.kubernetes.io/safe-to-evict": "false"
   name: etcd-{{ .Values.role }}
   namespace: {{ .Release.Namespace }}
   labels:
@@ -21,6 +19,7 @@ spec:
   template:
     metadata:
       annotations:
+        "cluster-autoscaler.kubernetes.io/safe-to-evict": "false"
         checksum/configmap-etcd-bootstrap-config: {{ include (print $.Template.BasePath "/configmap-etcd-bootstrap.yaml") . | sha256sum }}
 {{- if .Values.podAnnotations }}
 {{ toYaml .Values.podAnnotations | indent 8 }}


timebertt (approved):
/lgtm

vpnachev:
Is this change merged in the master branch ?

vpnachev:
Actually, in the master we are using etcd druid https://github.com/gardener/gardener/pull/1762 now, so it cannot be implemented there, right?

shreyas-s-rao:
@vpnachev correct. This PR makes sense only for release-v1.1 branch, because druid isn't part of this

vpnachev (approved):

vpnachev:
Please add a release note `user noteworthy` that the etcd will be restarted in the next maintenance window.

ialidzhikov:
Does the etcd-druid adds this annotation? Are we sure that etcd-druid won't regress this? Do we want both etcd-main and etcd-events to be critical or only etcd-main as I see this PR will make them both critical?

swapnilgm:
> Does the etcd-druid adds this annotation? Are we sure that etcd-druid won't regress this? 

Yes, druid add this annotation ([check here](https://github.com/gardener/etcd-druid/blob/master/charts/etcd/templates/etcd-statefulset.yaml#L38)). 

> Do we want both etcd-main and etcd-events to be critical or only etcd-main as I see this PR will make them both critical?

i prefer both.

swapnilgm:
@amshuman-kr Does it makes sense to add the node labels in seed bootstrap and make use of node affinity in etcd statefulset , so that one can separate out the etcd pods to isolated nodes. And such workerpool can be configured separately to autoscaler. This prevents the auto-scalling impact, which was generally introduced from other control plane components based on shoot workload. We discussed this vaguely sometime back. This has another benefit of configuring the etcd related workpool more IO reach rather than compute/memory reach. WDYS?

I expect this to be small change and won't extend hotfix timeline too much. And avoid additional change to etcd in future. 

cc @hardikdr

amshuman-kr:
> Do we want both etcd-main and etcd-events to be critical or only etcd-main as I see this PR will make them both critical?

Good catch. I don't see why etcd-events should block CA.

amshuman-kr:
> Does it makes sense to add the node labels in seed bootstrap and make use of node affinity in etcd statefulset , so that one can separate out the etcd pods to isolated nodes. And such workerpool can be configured separately to autoscaler. This prevents the auto-scalling impact, which was generally introduced from other control plane components based on shoot workload. We discussed this vaguely sometime back. This has another benefit of configuring the etcd related workpool more IO reach rather than compute/memory reach. WDYS?

I have been thinking about heterogenous worker pools for seeds for a long time. To me, it makes sense to schedule different kinds of workloads to different dedicated worker pools in the seeds. It makes sense to schedule etcd to dedicated pool so that rest of the pools are not free to scale down independently. In future, we could think about having a pool of preemptive/spot instances for _normal_ controller pods.

swapnilgm:
@amshuman-kr I assume this as affirmative to have affinity change in this PR.
@shreyas-s-rao  Will you please do the change?  Basically add ndoeaffinity to etcd sts as 
```
  affinity:
    nodeAffinity:
      preferredDuringSchedulingIgnoredDuringExecution: //SOFT affinity
      - weight: 1
        preference:
          matchExpressions:
          - key: node-restriction.kubernetes.io/allowed-component
            operator: In
            values:
            - etcd-main
           - etcd-events
```
@georgekuruvillak : It makes sense for druid to also expose the affinity parameter on etcd CRD.

**Regarding node labels** : Now thinking over it in deep, hardcoding node labels in seed bootstrap is doesn't sound good. Cause gardener don't expect any knowledge about the seed cluster worker nodes configuration. Some labeling could be applied for shootSeed, but that can be changed later in code without change to etcd.

swapnilgm:
For node labelling, i have considered the [ node-isolation-restriction](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-isolation-restriction) feature and allow future common labelling pattern regarding affinity for other components like kube-apiserver as well. Please feel free to suggest any other, if above mentioned lable key-values doesn't work.

rfranzke:
@swapnilgm @amshuman-kr can we move the discussion to another ticket? It's out of scope of this PR.

rfranzke (approved):

rfranzke:
Are you guys OK with getting this PR in now?

timuthy:
We should do the **absolute minimum** to stabilize this release. Any tweaks should be done in further releases to not put the `v1.1.x` at stake.

swapnilgm:
I'm fine with keeping discussion away from this PR and merging this PR. I was considering about adding  node affinity in same release. But if, majority feel it's risky then, let's keep it out this release.

swapnilgm (approved):
LGTM

amshuman-kr:
> I'm fine with keeping discussion away from this PR and merging this PR. I was considering about adding node affinity in same release. But if, majority feel it's risky then, let's keep it out this release.

I do feel it is risky to do it in one shot. Also, we will need to introduce multiple worker pools in the seeds with some taints and tolerations rather than node affinity for this to work well.

PR #2026, opened 2020-03-10
[release-v1.1] Revise migration steps for primary DNS provider

timuthy:
This PR revises the migration logic for setting the primary DNS provider of shoot clusters introduced by #1959. Primary DNS providers are now set right before the reconcile flow instead of setting it through an admission plugin. The motivation for this change is to guarantee the "reconcile in maintenance only" feature.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/apis/core/v1alpha1/types_shoot.go
+++ b/pkg/apis/core/v1alpha1/types_shoot.go
@@ -218,12 +218,11 @@ type DNSProvider struct {
 	Primary *bool `json:"primary,omitempty"`
 	// SecretName is a name of a secret containing credentials for the stated domain and the
 	// provider. When not specified, the Gardener will use the cloud provider credentials referenced
-	// by the Shoot and try to find respective credentials there. Specifying this field may override
+	// by the Shoot and try to find respective credentials there (primary provider only). Specifying this field may override
 	// this behavior, i.e. forcing the Gardener to only look into the given secret.
 	// +optional
 	SecretName *string `json:"secretName,omitempty"`
-	// Type is the DNS provider type for the Shoot. Only relevant if not the default domain is used for
-	// this shoot.
+	// Type is the DNS provider type.
 	// +optional
 	Type *string `json:"type,omitempty"`
 	// Zones contains information about which hosted zones shall be included/excluded for this provider.
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -16,17 +16,13 @@ package shoot
 
 import (
 	"context"
-	"fmt"
 	"time"
-
-	"k8s.io/utils/pointer"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	"github.com/gardener/gardener/pkg/controllerutils"
 	"github.com/gardener/gardener/pkg/operation"
 	botanistpkg "github.com/gardener/gardener/pkg/operation/botanist"
-	"github.com/gardener/gardener/pkg/operation/garden"
 	"github.com/gardener/gardener/pkg/utils"
 	"github.com/gardener/gardener/pkg/utils/errors"
 	"github.com/gardener/gardener/pkg/utils/flow"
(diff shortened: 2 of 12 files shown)


rfranzke (commented):
Looks OK to me, but I didn't have a chance to review the initial PR, so maybe @ialidzhikov and @tim-ebert can give an opinion as well?

ialidzhikov:
What was the thing that breaks the "reconcile in maintenance only"? The health check update requests to Shoot labels and on the update request the admission was setting dnsprovider as primary in the Shoot spec?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -53,11 +56,27 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
...
+
 	// TODO: timuthy - Only required for migration and can be removed in a future version
-	if o.Shoot.ExternalClusterDomain != nil && garden.DomainIsDefaultDomain(*o.Shoot.ExternalClusterDomain, o.Garden.DefaultDomains) != nil {
-		o.Logger.Info("Migration step - setting primary DNS provider")
-		if err := kutil.SubmitEmptyPatch(context.TODO(), o.K8sGardenClient.Client(), o.Shoot.Info); err != nil {
-			return gardencorev1beta1helper.NewWrappedLastErrors(gardencorev1beta1helper.FormatLastErrDescription(err), err)
+	if dnsEnabled {
+		var (
+			primaryDNSProvider = gardencorev1beta1helper.FindPrimaryDNSProvider(o.Shoot.Info.Spec.DNS.Providers)
+			usesDefaultDomain  = o.Shoot.ExternalClusterDomain != nil && garden.DomainIsDefaultDomain(*o.Shoot.ExternalClusterDomain, o.Garden.DefaultDomains) != nil
```
timebertt: `garden.DomainIsDefaultDomain` should be improved to test for `"."+defaultDomain.Domain` suffix like you already do in the dns admission plugin.
https://github.com/gardener/gardener/blob/90fdb8f3afdf1999b9e3938910cda87d02b6c2f2/pkg/operation/garden/garden.go#L113


Please also add an `Entry` to the corresponding unit test for the problematic case like
```go
Entry("no default domain but with same suffix", "foo.foobar.com", []*garden.Domain{defaultDomain}, BeNil()),
```
here: https://github.com/gardener/gardener/blob/5dfca6bfa03e4ee8486562d38f10f861350d822f/pkg/operation/garden/garden_test.go#L171

plugin/pkg/shoot/dns/admission.go
```diff
@@ -201,27 +207,43 @@ func (d *DNS) Admit(ctx context.Context, a admission.Attributes, o admission.Obj
...
 
-func managePrimaryDNSProvider(dns *core.DNS, defaultDomains []string) error {
-	var usesDefaultDomain bool
-	for _, domain := range defaultDomains {
-		if strings.HasSuffix(*dns.Domain, "."+domain) {
-			usesDefaultDomain = true
-			break
-		}
+func checkPrimaryDNSProvider(dns *core.DNS, defaultDomains []string) error {
+	if dns == nil || dns.Domain == nil {
```
timebertt: Just a small improvement:

```suggestion
	if dns == nil || dns.Domain == nil || len(dns.Providers) == 0 {
```

timebertt (changes requested):
Overall the PR looks good to me.
I would just propose some minor changes.

timebertt:
> What was the thing that breaks the "reconcile in maintenance only"? The health check update requests to Shoot labels and on the update request the admission was setting dnsprovider as primary in the Shoot spec?

Exactly. Now, the first dns provider of an existing Shoot will only get promoted as the `primary` one if the Shoot is scheduled (`CREATE` with `spec.seedName` already set, or `UPDATE` with `spec.seedName="" -> spec.seedName="<seed-name>"`).

plugin/pkg/shoot/dns/admission.go
```diff
@@ -163,29 +188,10 @@ func (d *DNS) Admit(ctx context.Context, a admission.Attributes, o admission.Obj
 		return nil
```
timebertt: I think we should not return here, in order to also promote the `unmanaged` provider as a `primary` one.

timebertt (approved):
Looks good to me now :)

ialidzhikov (approved):
/lgtm

timuthy:
Please wait before merging this PR. I want to verify the `unmanaged` provider use case.

timuthy:
@rfranzke @ialidzhikov @tim-ebert thanks for you reviews.
I had to move the removal of functionless DNS providers f70d964 from the "shootStrategy" and it is now done by the migration logic as well. This prevents that proper DNS providers w/o a `secretName` (primary providers) aren't removed.

timuthy:
Update: I optimized the enqueuing: If DNS providers are migrated the shoot is immediately re-queued before running the entire reconcile flow.

plugin/pkg/shoot/dns/admission_test.go
```diff
@@ -119,6 +119,19 @@ var _ = Describe("dns", func() {
...
 
+			attrs := admission.NewAttributesRecord(shootCopy, nil, core.Kind("Shoot").WithVersion("version"), shootCopy.Namespace, shootCopy.Name, core.Resource("shoots").WithVersion("version"), "status", admission.Create, &metav1.CreateOptions{}, false, nil)
+
+			err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(*shootCopy).To(Equal(*shootBefore))
+		})
+
+		It("should do nothing because the shoot status is updated", func() {
```
ialidzhikov: I think the test case names are switched. This one needs to be:
```suggestion
		It("should do nothing because the shoot does not specify a seed (create)", func() {
```

And the above test case needs to be 
```
It("should do nothing because the shoot status is updated", func() {
```

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -218,11 +218,11 @@ type DNSProvider struct {
...
 	// SecretName is a name of a secret containing credentials for the stated domain and the
 	// provider. When not specified, the Gardener will use the cloud provider credentials referenced
-	// by the Shoot and try to find respective credentials there. Specifying this field may override
+	// by the Shoot and try to find respective credentials there (primary provider only). Specifying this field may override
 	// this behavior, i.e. forcing the Gardener to only look into the given secret.
 	// +optional
 	SecretName *string `json:"secretName,omitempty"`
-	// Type is the DNS provider type for the Shoot. Only relevant if not the default domain is used for
+	// Type is the DNS provider type.
 	// this shoot.
```
timebertt:
```suggestion

```

plugin/pkg/shoot/dns/admission_test.go
timuthy: Yes, makes sense. Thank you 👍

ialidzhikov (approved):
/lgtm

timebertt (commented):
just minor left over in the API doc

timebertt (approved):
/lgtm now
Thank you for the fixes :)

PR #2027, opened 2020-03-10
Add integration test for clusterautoscaler

timebertt:
This PR adds a clusterautoscaler test to the shoot integration test suite.

The test creates a `reserve-capacity` deployment, which requests half of the Node's allocatable capacity. It then scales up the deployment and waits for one new node to be added to the cluster. And then it scales down the deployment again and waits for one node to be removed from the cluster.

Part of #1564

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/test/integration/shoots/operations/clusterautoscaler.go
+++ b/test/integration/shoots/operations/clusterautoscaler.go
@@ -1,0 +1,179 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+	Overview
+		- Tests Shoot cluster autoscaling
+
+	AfterSuite
+		- Cleanup Workload in Shoot
+
+	Test:
+		1. Create a deployment with Pods each requesting half the capacity of a single node
+		2. Scale up the deployment and see one node added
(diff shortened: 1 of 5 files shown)


timebertt:
Fixed failing `make check`

test/integration/shoots/operations/clusterautoscaler.go
```diff
@@ -0,0 +1,179 @@
...
+
+const (
+	reserveCapacityTemplateName        = "reserve-capacity.yaml.tpl"
+	reserveCapacityDeploymentName      = "reserve-capacity"
+	reserveCapacityDeploymentNamespace = metav1.NamespaceDefault
+
+	scaleDownDelayAfterAdd = 1 * time.Minute
+	scaleDownUnneededTime  = 1 * time.Minute
+	testTimeout            = 30 * time.Minute
+	scaleUpTimeout         = 10 * time.Minute
```
schrodit: This timeout could be problematic as it can happen that on some infrastructure VM provisioning takes longer than 10 Minutes. We also saw in previous tests that noone is loocking after tests that ran into a timeout.
Therefore, I would propose to increase the timeouts to 15-20min? WDYT?

schrodit (changes requested):
Just a minor comment.
Rest looks prettty good and is already tested with the testmachinery.

test/integration/shoots/operations/clusterautoscaler.go
timebertt: Would you only increase the `scaleUpTimeout` or also the `scaleDownTimeout`?
Which value should I use for `testTimeout`? I think this should have some margin over `scaleUpTimeout + scaleDownTimeout`...

danielfoehrKn (commented):
Looks good to me

test/integration/shoots/operations/clusterautoscaler.go
schrodit: I would also suggest to increase the scale down time.
So lets make:
scaleUpTimeout: 20min
scaleDownTimeout: 20min
testTimeout: 50min (with 10 minutes margin)

Lets make the timeouts a little bit higher as we do not test the performance of the mcm or any infractructure. Therefore the timeouts are just for safety that the test are not running infinitely.

timebertt: Alright, done :)
I set `testTimeout` to 60min now, because the test is changing the Shoot spec twice (before + after) and waits for the Shoot to be completely reconciled afterwards. So I guess, it makes sense to have a margin of 20min.

timebertt:
@schrodit @danielfoehrKn PTAL :)

schrodit (approved):
lgtm

PR #1935, opened 2020-02-14
update kyma-addon to kyma 1.10

a-thaler:
**What this PR does / why we need it**:
Updates the kyma addon to Kyma 1.10 with the following benefits:
- Support for kubernetes 1.16
- Gardener certificate logic removed from addon as it is part of kyma-installer
- Kyma version has improvements in stability

**Which issue(s) this PR fixes**:
Follow-Up on #1754

**Special notes for your reviewer**:
Only GCP and Azure are supported. 

**Release note**:
```improvement operator
Update of the temporary, experimental Kyma addon to latest Kyma version 1.10. It can be installed onto shoot clusters out-of-the-box by annotating the `Shoot` with `experimental.addons.shoot.gardener.cloud/kyma=enabled`. Be aware that we won't provide upgrades or customization, and that this addon is temporary and will be removed in a future version of Gardener again. Its purpose is to ease the Kyma installation and to show-case which features it provides. It is by no means a production-ready setup. Also, please note that, once enabled, the Kyma addon can never be disabled again. The only way to get rid of it is to delete the shoot cluster. You can check the status of the installation by using `kubectl get installation/kyma-installation -o jsonpath="{'Status: '}{.status.state}{', description: '}{.status.description}"`.
```

PR #2028, opened 2020-03-10
Prevent kyma addon to pick to wrong kyma version

ialidzhikov:
Currently the master branch of g/g tries to install kyma@1.8.0 which fails. As of #1935 the correct version should be kyma@1.10. Currently the parent `values.yaml` has 1.8.0 and the subchart ` values.yaml` has 1.10.0. Looks like the value in the parent chart is overriding the value in the subchart. 
The PR just removes the parent values to avoid double-maintenance and to have the charts applied with the correct kyma version.

Fixes #

**Release note**:

```improvement user
An issue preventing kyma 1.10.0 to be installed is now fixed.
```

### DIFF
--- a/charts/shoot-addons-kyma/values.yaml
+++ b/charts/shoot-addons-kyma/values.yaml
@@ -1,13 +1,2 @@
 kyma:
   enabled: false
-  jobs:
-    image: eu.gcr.io/kyma-project/test-infra/alpine-kubectl:v20190325-ff66a3a
-    delay: 30
-  kyma:
-    version: "1.8.0"
-    subdomain: "kyma"
-  requires:
-    k8s:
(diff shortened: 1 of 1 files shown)


ialidzhikov:
cc @a-thaler

rfranzke:
Thanks!
Will this be cherry-picked to the release-v1.1 branch as well?

ialidzhikov:
> Thanks!
> Will this be cherry-picked to the release-v1.1 branch as well?

Proposed #2031.

a-thaler:
Looks good to me.

Indeed I have not seen that second values.yaml when creating https://github.com/gardener/gardener/pull/1935/files. Removing the value duplication will avoid that for the future.

PR #2034, opened 2020-03-11
Ensure backwards compatibility for DNS section

timuthy:
This PR ensures the backwards compatibility of the shoot's DNS section before multiple DNS providers were supported. Previously it was possible to apply shoot manifests w/o `primary: true` providers but since the admission logic has been touched in release `v1.1.3`, a re-apply of the same shoot manifest leads to an error:

```
spec.dns.providers: Forbidden: removing a primary provider is not allowed
```

The admission logic now re-adds the primary provider in case it was removed.

**Release note**:

```improvement user
A compatibility issue has been fixed which prevented shoot manifests from being re-applied w/o a primary DNS provider.
```

### DIFF
--- a/plugin/pkg/shoot/dns/admission.go
+++ b/plugin/pkg/shoot/dns/admission.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


timebertt (approved):
Thanks, Tim!
/approve

PR #2035, opened 2020-03-11
GEP 10 - Container runtime extensions

### LINKED ISSUE #1454: Container Runtime Options

nimrodoron:
GEP-10 implementation Creating new Container runtime extensions

Fixes #1454

**Release note**:

```noteworthy user
It is now possible to configure additional container runtimes like gVisor (if previously registered by the Gardener landscape administrator) for shoot clusters. It is possible via the the `.spec.provider.workers[].cri.containerRuntimes` list. Example value: `[{type: gvisor}]`. It can be set to each worker pool in the shoot cluster.
```

```noteworthy developer
A new container runtime extensibility contract has been introduced that allows to develop external container runtime extension controllers. Information for developing such an extension can be found [here](https://github.com/gardener/gardener/tree/master/docs/extensions/containerruntime.md).
```


pkg/apis/core/types_shoot.go
```diff
@@ -711,6 +711,9 @@ type Volume struct {
 type CRI struct {
 	// The name of the CRI library
 	Name CRIName
+	// ContainerRuntimes is the list of the required container runtimes supported for a worker pool.
+	// +optional
```
rfranzke:
```suggestion
```

optional tag not needed in internal API

pkg/apis/core/types_shoot.go
```diff
@@ -720,6 +723,12 @@ const (
 	CRINameContainerD CRIName = "containerd"
 )
 
+// ContainerRuntime contains information about worker's available container runtime
+type ContainerRuntime struct {
+	// Type is the type of the Container Runtime.
+	Type string
```
rfranzke, line 728:
```suggestion
	Type string
	// ProviderConfig is the configuration passed to the ContainerRuntime resource.
	ProviderConfig *ProviderConfig
```

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -867,6 +867,9 @@ type Volume struct {
 type CRI struct {
 	// The name of the CRI library
 	Name CRIName `json:"name"`
+	// ContainerRuntimes is the list of the required container runtimes supported for a worker pool.
+	// +optional
+	ContainerRuntimes []ContainerRuntime `json:"dataVolumes,omitempty"`
```
rfranzke:
```suggestion
	ContainerRuntimes []ContainerRuntime `json:"containerRuntimes,omitempty"`
```

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -876,6 +879,12 @@ const (
 	CRINameContainerD CRIName = "containerd"
 )
 
+// ContainerRuntime contains information about worker's available container runtime
+type ContainerRuntime struct {
+	// Type is the type of the Container Runtime.
+	Type string `json:"type"`
```
rfranzke, line 884:
```suggestion
	Type string `json:"type"`
	// ProviderConfig is the configuration passed to the ContainerRuntime resource.
	// +optional
	ProviderConfig *ProviderConfig `json:"providerConfig,omitempty"`
```

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -864,6 +864,9 @@ type Volume struct {
 type CRI struct {
 	// The name of the CRI library
 	Name CRIName `json:"name"`
+	// ContainerRuntimes is the list of the required container runtimes supported for a worker pool.
+	// +optional
+	ContainerRuntimes []ContainerRuntime `json:"dataVolumes,omitempty"`
```
rfranzke:
```suggestion
	ContainerRuntimes []ContainerRuntime `json:"containerRuntimes,omitempty"`
```

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -873,6 +876,12 @@ const (
 	CRINameContainerD CRIName = "containerd"
 )
 
+// ContainerRuntime contains information about worker's available container runtime
+type ContainerRuntime struct {
+	// Type is the type of the Container Runtime.
+	Type string `json:"type"`
```
rfranzke, line 881:
```suggestion
	Type string `json:"type"`
	// ProviderConfig is the configuration passed to the ContainerRuntime resource.
	// +optional
	ProviderConfig *ProviderConfig `json:"providerConfig,omitempty"`
```

pkg/apis/core/validation/shoot.go
```diff
@@ -1216,5 +1216,26 @@ func ValidateCRI(CRI *core.CRI, fldPath *field.Path) field.ErrorList {
 		allErrs = append(allErrs, field.NotSupported(fldPath.Child("name"), CRI.Name, avaliableWorkerCRINames.List()))
 	}
 
+	if  CRI.ContainerRuntimes != nil {
```
rfranzke: run `make format`

pkg/apis/extensions/v1alpha1/types_containerruntime.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// GetExtensionStatus implements Object.
+func (i *ContainerRuntime) GetExtensionStatus() Status {
+	return &i.Status
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ExtensionList is a list of Extension resources.
```
rfranzke:
```suggestion
// ContainerRuntimeList is a list of ContainerRuntime resources.
```

replace `Extension` with `ContainerRuntime` everywhere

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -351,6 +351,16 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation, errorContext *er
 			Fn:           botanist.WaitUntilExtensionResourcesDeleted,
 			Dependencies: flow.NewTaskIDs(deleteExtensionResources),
 		})
+		deleteContainerRuntimeResources = g.Add(flow.Task{
+			Name:         "Delete stale container runtime resources",
+			Fn:           botanist.DeleteContainerRuntimeResources,
+			Dependencies: flow.NewTaskIDs(initializeShootClients),
```
rfranzke:
```suggestion
			Dependencies: flow.NewTaskIDs(initializeShootClients, cleanKubernetesResources, cleanShootNamespaces),
```

I guess the container runtimes should only be deleted after the Kubernetes cluster has been cleaned, right?

pkg/operation/shoot/shoot.go
```diff
@@ -431,3 +436,51 @@ func ToNetworks(s *gardencorev1beta1.Shoot) (*Networks, error) {
 		APIServer: apiserver,
 	}, nil
 }
+
+func calculateContainerRuntimes(gardenClient client.Client, shoot *gardencorev1beta1.Shoot, seedNamespace string) (map[string]extensionsv1alpha1.ContainerRuntime, error) {
```
rfranzke: Hm, why do we need these two functions here and the `ContainerRuntimesMap` at all? I guess you copied from the `Extension` resource, however, here the difference is that some `Extension` resources are globally enabled without the need to be in the `Shoot`. For `ContainerRuntimes`, however, all the desired `type`s are in the Shoot manifest, so we don't need this stuff here, I assume.

pkg/operation/botanist/worker.go
```diff
@@ -83,6 +83,15 @@ func (b *Botanist) DeployWorker(ctx context.Context) error {
 			}
 		}
 
+		if worker.CRI != nil {
```
rfranzke: Just for me to understand: Can you quickly explain why we add these labels?

rfranzke (changes requested):
Nice PR, thank you.
Apart from the inline suggestions, please

1. Add the computation of required `ContainerRuntime` extensions to this function: https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/botanist.go#L104-L152
1. Add a new document for the `ContainerRuntime` extension contract to https://github.com/gardener/gardener/tree/master/docs/extensions (similar to https://github.com/gardener/gardener/blob/master/docs/extensions/network.md, for example)
1. Add the new document to the document index (https://github.com/gardener/gardener/blob/master/docs/README.md#extensions)

rfranzke:
Also, please add a release note for end-users (where to configure container runtimes), operators (information that new container runtime extensions are now supported), and developers (pointing to the documentation for the new container runtime extensions).

pkg/gardenlet/controller/shoot/shoot_control_delete.go
nimrodoron: Yes

pkg/operation/shoot/shoot.go
nimrodoron: Well, they do exist in shoot manifest, but for each worker separately, so there can be duplicated or unavailable. Here i basically get a distinct list of container runtimes and check for the registered controller registrations

pkg/operation/botanist/worker.go
nimrodoron: Yes, in the container runtime extensions the installation is done by daemon set only for the workers (nodes) who has those annotations.

nimrodoron:
> Also, please add a release note for end-users (where to configure container runtimes), operators (information that new container runtime extensions are now supported), and developers (pointing to the documentation for the new container runtime extensions).

I will, i first want everything to work and i will add this as well.

pkg/operation/botanist/worker.go
rfranzke: Understood, makes sense. However, where are the nodes labelled accordingly so that the `DaemonSet`s can define a proper label selector?

pkg/operation/shoot/shoot.go
rfranzke: Yes, but why is this needed? I think you can simply, in the `DeployContainerRuntimes` function, loop over all worker pools and all container runtimes, and then create the `ContainerRuntime` objects in the seed. You don't need any `ControllerRegistration`s (this is to be checked elsewhere and out-of-scope of this PR). Or?

nimrodoron: Ok, so the ControllerRegistration check will be done instead in the:
botanist: RequiredExtensionsExist?

If i will remove this function i will have to calculate twice the set of container runtimes. 
in the RequiredExtensionsExist and in the DeployContainerRuntimes. i can also change this function just to save a set of required container runtimes on the shoot. Than i will iterate over this set in both of the above functions, I indeed doesn't need the ControllerRegistration resource.

pkg/operation/botanist/worker.go
nimrodoron: Well, i assumed all labels set here to the worker will also be set to the node:
`// Labels is a map of key/value pairs for labels for all the `Node` objects in this worker pool.
	// +optional
	Labels map[string]string `json:"labels,omitempty"``
Actually it is different from what i did in the POC (Manipulated the Kubelet flags so the nodes where labelled. I was still planning to check this.
If it doesn't work like that i will revert it to the _Kubelet

pkg/apis/core/v1beta1/types_shoot.go
rfranzke: ping

pkg/apis/extensions/v1alpha1/types_containerruntime.go
```diff
@@ -0,0 +1,71 @@
...
+// ContainerRuntime is a specification for a container runtime resource.
+type ContainerRuntime struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              ContainerRuntimeSpec   `json:"spec"`
+	Status            ContainerRuntimeStatus `json:"status"`
+}
+
+// GetExtensionSpec implements Object.
```
rfranzke, line 44: everything in this file should renamed from `Extension` to `ContainerRuntime`

pkg/operation/shoot/shoot.go
rfranzke: > Ok, so the ControllerRegistration check will be done instead in the:
botanist: RequiredExtensionsExist?

Yes, exactly, that's enough.

> If i will remove this function i will have to calculate twice the set of container runtimes.
in the RequiredExtensionsExist and in the DeployContainerRuntimes. i can also change this function just to save a set of required container runtimes on the shoot. Than i will iterate over this set in both of the above functions, I indeed doesn't need the ControllerRegistration resource.

Hm, I don't understand. In the `RequiredExtensionExist` function you check whether a `ControllerRegistration` exists that supports the given `types` for the `ContainerRuntime` kind.
In the `DeployContainerRuntime` function you loop over each worker pool and then over each container runtime, and simply create the `ContainerRuntime` CRDs based on the provided `type` and `providerConfig`. Nothing to compute here. Yes, you have to loop in both functions, but that's how it's done for the other extension resources as well at the moment. I don't think it's a problem. :)

pkg/operation/botanist/worker.go
rfranzke: OK, no, it doesn't work like that. In the PR we discussed that the node labels can be done via the [kubelet flags](https://github.com/gardener/gardener/blob/master/charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags#L22-L26), so let's move the labels here.

pkg/apis/extensions/v1alpha1/types_containerruntime.go
nimrodoron: What do you want me to change in the types_shoot?

rfranzke: I don't understand the question. In this file, `types_containerruntime.go`, replace `Extension` with `ContainerRuntime`.

nimrodoron: Yes, this i understood, i meant the ping above :)

rfranzke: See the hidden conversation above: Add the `providerConfig` field and run `make generate` again

nimrodoron: The extension can't be changed to containerruntime, It should implement the Object interface like all other extensions.

rfranzke: WDYM? I can't follow. I am talking about https://github.com/gardener/gardener/pull/2035#discussion_r390959259

nimrodoron: I lost track a little bit here... Please make sure those 2 files  are ok now

charts/seed-operatingsystemconfig/original/values.yaml
```diff
@@ -5,7 +5,10 @@ osc:
   secretName: cpu-worker-0
   sshKey: "ssh-rsa"
   cri:
-    containerdName: containerd
+    labels:
+      name: worker.gardener.cloud/criname
```
rfranzke:
```suggestion
      name: worker.gardener.cloud/cri-name
```

pkg/operation/botanist/botanist.go
```diff
@@ -109,7 +109,16 @@ func (b *Botanist) computeRequiredExtensions() map[string]sets.String {
 		if worker.Machine.Image != nil {
 			machineImagesSet.Insert(string(worker.Machine.Image.Name))
 		}
+		if worker.CRI != nil && worker.CRI.ContainerRuntimes != nil {
```
rfranzke:
```suggestion
		if worker.CRI != nil {
```

not needed, you range over `nil`

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+// limitations under the License.
+
+package botanist
+
+import (
+	"context"
+	"fmt"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+
+	v1alpha1constants "github.com/gardener/gardener/pkg/apis/core/v1alpha1/constants"
```
rfranzke:
```suggestion
	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```

use `v1beta1 API only in controller/gardenlet

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+							}
+
+							if err := health.CheckExtensionObject(req); err != nil {
+								b.Logger.WithError(err).Errorf("Container runtime %s/%s did not get ready yet", namespace, name)
+								return retry.MinorError(err)
+							}
+
+							return retry.Ok()
+						}); err != nil {
+							return gardencorev1alpha1helper.DetermineError(fmt.Sprintf("failed waiting for container runtime %s to be ready: %v", name, err))
```
rfranzke:
```suggestion
							return gardencorev1beta1helper.DetermineError(fmt.Sprintf("failed waiting for container runtime %s to be ready: %v", name, err))
```

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+
+package botanist
+
+import (
+	"context"
+	"fmt"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+
+	v1alpha1constants "github.com/gardener/gardener/pkg/apis/core/v1alpha1/constants"
+	gardencorev1alpha1helper "github.com/gardener/gardener/pkg/apis/core/v1alpha1/helper"
```
rfranzke:
```suggestion
```

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
```
rfranzke, line 24:
```suggestion
```

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+				if lastErr := status.LastError; lastErr != nil {
+					b.Logger.Errorf("Container runtime %s did not get deleted yet, lastError is: %s", name, lastErr.Description)
+					lastError = lastErr
+				}
+
+				return retry.MinorError(gardencorev1beta1helper.WrapWithLastError(fmt.Errorf("container runtime %s is still present", name), lastError))
+			}); err != nil {
+				message := fmt.Sprintf("Failed waiting for container runtime delete")
+				if lastError != nil {
+					return gardencorev1alpha1helper.DetermineError(fmt.Sprintf("%s: %s", message, lastError.Description))
```
rfranzke:
```suggestion
					return gardencorev1beta1helper.DetermineError(fmt.Sprintf("%s: %s", message, lastError.Description))
```

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+				}
+				return gardencorev1alpha1helper.DetermineError(fmt.Sprintf("%s: %s", message, err.Error()))
```
rfranzke:
```suggestion
				return gardencorev1beta1helper.DetermineError(fmt.Sprintf("%s: %s", message, err.Error()))
```

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+								Name:      containerRuntime.Type,
+								Namespace: b.Shoot.SeedNamespace,
+							},
+						}
+					)
+
+					fns = append(fns, func(ctx context.Context) error {
+
+						_, err := controllerutil.CreateOrUpdate(ctx, b.K8sSeedClient.Client(), &toApply, func() error {
+							metav1.SetMetaDataAnnotation(&toApply.ObjectMeta, v1alpha1constants.GardenerOperation, v1alpha1constants.GardenerOperationReconcile)
```
rfranzke:
```suggestion
							metav1.SetMetaDataAnnotation(&toApply.ObjectMeta, v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationReconcile)
```

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// DeployContainerRuntimeResources creates the `Container runtime` resource in the shoot namespace in the seed
+// cluster. Gardener waits until an external controller did reconcile the resources successfully.
+func (b *Botanist) DeployContainerRuntimeResources(ctx context.Context) error {
+	fns := []flow.TaskFn{}
+	requiredContainerRuntimeTypes := sets.NewString()
+	for _, worker := range b.Shoot.Info.Spec.Provider.Workers {
+		if worker.CRI != nil && worker.CRI.ContainerRuntimes != nil {
```
rfranzke:
```suggestion
		if worker.CRI != nil {
```

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+							},
+						}
+					)
+
+					fns = append(fns, func(ctx context.Context) error {
+
+						_, err := controllerutil.CreateOrUpdate(ctx, b.K8sSeedClient.Client(), &toApply, func() error {
+							metav1.SetMetaDataAnnotation(&toApply.ObjectMeta, v1alpha1constants.GardenerOperation, v1alpha1constants.GardenerOperationReconcile)
+
+							toApply.Spec.Type = containerRuntimeType
```
rfranzke: You should also pass the `ProviderConfig` to `toApply.Spec`

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	return flow.Parallel(fns...)(ctx)
+}
+
+// DeleteStaleContainerRuntimeResources deletes unused container runtime resources from the shoot namespace in the seed.
+func (b *Botanist) DeleteStaleContainerRuntimeResources(ctx context.Context) error {
+	wantedContainerRuntimes := sets.NewString()
+	for _, worker := range b.Shoot.Info.Spec.Provider.Workers {
+		if worker.CRI != nil && worker.CRI.ContainerRuntimes != nil {
```
rfranzke:
```suggestion
		if worker.CRI != nil {
```

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+
+// WaitUntilContainerRuntimeResourcesReady waits until all container runtime resources report `Succeeded` in their last operation state.
+// The state must be reported before the passed context is cancelled or a container runtime's timeout has been reached.
+// As soon as one timeout has been overstepped the function returns an error, further waits on container runtime will be aborted.
+func (b *Botanist) WaitUntilContainerRuntimeResourcesReady(ctx context.Context) error {
+	fns := []flow.TaskFn{}
+	rquiredContainerRuntimeTypes := sets.NewString()
+
+	for _, worker := range b.Shoot.Info.Spec.Provider.Workers {
+		if worker.CRI != nil && worker.CRI.ContainerRuntimes != nil {
```
rfranzke:
```suggestion
		if worker.CRI != nil {
```

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,214 @@
...
+
+	return flow.Parallel(fns...)(ctx)
+}
+
+// WaitUntilContainerRuntimeResourcesReady waits until all container runtime resources report `Succeeded` in their last operation state.
+// The state must be reported before the passed context is cancelled or a container runtime's timeout has been reached.
+// As soon as one timeout has been overstepped the function returns an error, further waits on container runtime will be aborted.
+func (b *Botanist) WaitUntilContainerRuntimeResourcesReady(ctx context.Context) error {
+	fns := []flow.TaskFn{}
+	rquiredContainerRuntimeTypes := sets.NewString()
```
rfranzke:
```suggestion
	requiredContainerRuntimeTypes := sets.NewString()
```

pkg/apis/extensions/v1alpha1/types_containerruntime.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ Object = (*ContainerRuntime)(nil)
+
+// ExtensionResource is a constant for the name of the Extension resource.
+const ContainerRuntimeResource = "ContainerRuntime"
+const CRINameWorkerLabel = "worker.gardener.cloud/criname"
```
rfranzke:
```suggestion
const CRINameWorkerLabel = "worker.gardener.cloud/cri-name"
```

pkg/apis/extensions/v1alpha1/types_containerruntime.go
```diff
@@ -0,0 +1,71 @@
...
+const ContainerRuntimeNameWorkerLabel = "worker.gardener.cloud/containerruntime/%s=true"
```
rfranzke:
```suggestion
const ContainerRuntimeNameWorkerLabel = "containerruntime.worker.gardener.cloud/%s=true"
```

usually, there is this naming scheme/pattern

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -320,6 +334,13 @@ func (b *Botanist) deployOperatingSystemConfigsForWorker(ctx context.Context, ma
 		criConfig := map[string]interface{}{
 			"name": worker.CRI.Name,
 		}
+		if worker.CRI.ContainerRuntimes != nil {
```
rfranzke:
```suggestion
		if len(worker.CRI.ContainerRuntimes) > 0 {
```

rfranzke (changes requested):

charts/seed-operatingsystemconfig/original/values.yaml
```diff
@@ -18,8 +21,10 @@ kubernetes:
   version: 1.11.2
 worker:
   name: cpu-worker
-# cri:
+  cri:
 #   name: containerd
+#   labels:
+#   - worker.gardener.cloud/containerruntime/gvisor=true
```
danielfoehrKn: Why does the node label not look like this : "worker.gardener.cloud/containerruntime=gvisor".  This way we would not have to introduce a new label for each runtime. 
[RuntimeClass.scheduling.NodeSelector](https://github.com/kubernetes/kubernetes/blob/f0323a2030c7adae0e0965a7d3b455dd416472a0/staging/src/k8s.io/api/node/v1beta1/types.go#L84)   should be working as well with the aforementioned label structure. What was the rational behind structuring the label like that?
The label was explicitly mentioned in the [GEP](https://github.com/gardener/gardener/blob/master/docs/proposals/10-shoot-additional-container-runtimes.md) - so I am wondering if I miss something here.

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -164,13 +164,27 @@ func (b *Botanist) deployOperatingSystemConfigsForWorker(ctx context.Context, ma
...
+	}
+
 	originalConfig["osc"] = map[string]interface{}{
 		"type":                 machineImage.Name,
 		"purpose":              extensionsv1alpha1.OperatingSystemConfigPurposeReconcile,
 		"reloadConfigFilePath": common.CloudConfigFilePath,
 		"secretName":           secretName,
 		"customization":        customization,
 		"sshKey":               string(sshKey),
+		"criConfig":            criConfig,
```
danielfoehrKn: i think this should be:
```suggestion
		"cri":            criConfig,
```
[see here](https://github.com/gardener/gardener/blob/ba17ac74b277e3e2b9403d85ea9509fb2005b65f/charts/seed-operatingsystemconfig/original/templates/node/_node-cri.labels#L2)

pkg/apis/core/validation/shoot.go
```diff
@@ -1216,5 +1216,26 @@ func ValidateCRI(CRI *core.CRI, fldPath *field.Path) field.ErrorList {
...
+func ValidateContainerRuntimes(containerRuntime []core.ContainerRuntime, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	crSet := make(map[string]bool)
+
+	for i, cr := range containerRuntime {
+		if len(cr.Type) == 0 {
+			allErrs = append(allErrs, field.Required(fldPath.Index(i).Child("type"), "must specify a container runtime type"))
+		}
+		if crSet[cr.Type] {
+			allErrs = append(allErrs, field.Duplicate(fldPath.Index(i).Child("type"), "must specify different type, already exist"))
```
danielfoehrKn: I think it would be nice to also state the value of the duplicate type in the error msg.

pkg/apis/extensions/v1alpha1/types_containerruntime.go
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
danielfoehrKn:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

danielfoehrKn (changes requested):

charts/seed-operatingsystemconfig/original/values.yaml
rfranzke: I think the reason is that there can be multiple container runtimes and not only one.

danielfoehrKn: Thanks Rafael, makes sense (as both these labels get added to the node eventually)

pkg/operation/botanist/containerruntime.go
rfranzke: This suggestion was not yet applied/resolved.

pkg/operation/botanist/containerruntime.go
rfranzke: This suggestion was not yet applied/resolved.

pkg/operation/botanist/containerruntime.go
rfranzke: This suggestion was not yet applied/resolved.

pkg/operation/shoot/shoot.go
```diff
@@ -460,6 +460,11 @@ func ComputeRequiredExtensions(shoot *gardencorev1beta1.Shoot, seed *gardencorev
 		if pool.Machine.Image != nil {
 			requiredExtensions.Insert(fmt.Sprintf("%s/%s", extensionsv1alpha1.OperatingSystemConfigResource, pool.Machine.Image.Name))
 		}
+		if pool.CRI != nil {
```
rfranzke, line 463: Thanks! :)

nimrodoron: Done

pkg/apis/extensions/v1alpha1/types_containerruntime.go
```diff
@@ -0,0 +1,71 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ Object = (*ContainerRuntime)(nil)
+
+// ExtensionResource is a constant for the name of the Extension resource.
```
timuthy: Please adjust docstring and add one each for the `const`s below.

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,215 @@
...
+		fns = append(fns, func(ctx context.Context) error {
+			if err := retry.UntilTimeout(ctx, DefaultInterval, shoot.ExtensionDefaultTimeout, func(ctx context.Context) (bool, error) {
+				if err := b.K8sSeedClient.Client().Get(ctx, kutil.Key(namespace, name), &extensionsv1alpha1.ContainerRuntime{}); err != nil {
+					if apierrors.IsNotFound(err) {
+						return retry.Ok()
+					}
+					return retry.SevereError(err)
+				}
+
+				if lastErr := status.LastError; lastErr != nil {
```
timuthy: Can you use the `status` from the retrieved object in line 190?

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -164,13 +164,27 @@ func (b *Botanist) deployOperatingSystemConfigsForWorker(ctx context.Context, ma
 
 	sshKey := b.Secrets[v1beta1constants.SecretNameSSHKeyPair].Data[secrets.DataKeySSHAuthorizedKeys]
 
+	criNamesConfig := map[string]interface{}{
+		"containerd": extensionsv1alpha1.CRINameContainerD,
+	}
+
+	workerNameLabel := map[string]interface{}{
+		"name": extensionsv1alpha1.CRINameWorkerLabel,
```
timuthy:
```suggestion
		"workerlabel": extensionsv1alpha1.CRINameWorkerLabel,
```

timuthy (changes requested):

pkg/apis/extensions/v1alpha1/types_operatingsystemconfig.go
```diff
@@ -218,3 +218,6 @@ const (
 	// CRINameContainerD is a constant for ContainerD CRI name
 	CRINameContainerD = "containerd"
 )
+
+// ContainerDRuntimeContainersBinFolder is the folder where Container Runtime binaries should be saved for ContainerD usage
+const ContainerDRuntimeContainersBinFolder = "/var/bin/cr"
```
rfranzke:
```suggestion
const ContainerDRuntimeContainersBinFolder = "/var/bin/containerruntimes"
```

rfranzke: Does it make sense to add a field to the `extensions.gardener.cloud/v1alpha1.ContainerRuntime` spec?

```
spec:
  binaryFolder: <path-to-binary-folder>
```

Instead of using this constant here? It'll make it more explicit for the CR extension where to put the binaries.

charts/seed-operatingsystemconfig/original/values.yaml
```diff
@@ -5,7 +5,11 @@ osc:
   secretName: cpu-worker-0
   sshKey: "ssh-rsa"
   cri:
-    containerdName: containerd
+    containerRuntimesBinaryPath: /var/bin/cr
```
rfranzke:
```suggestion
    containerRuntimesBinaryPath: /var/bin/containerruntimes
```

charts/seed-bootstrap/templates/extensions/crd-containerruntime.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: containerruntimes.extensions.gardener.cloud
```
rfranzke, line 4:
```suggestion
  name: containerruntimes.extensions.gardener.cloud
  labels:
    gardener.cloud/deletion-protected: "true"
```

charts/seed-operatingsystemconfig/original/templates/scripts/_init-containerd.sh
```diff
@@ -12,4 +12,16 @@
           mkdir -p /etc/containerd
           containerd config default > "$FILE"
         fi
+
+        BIN_PATH={{ .Values.osc.cri.containerRuntimesBinaryPath }}
+        mkdir -p $BIN_PATH
+
+        ENV_FILE=/etc/systemd/system/containerd.service.d/30-env_config.conf
```
danielfoehrKn, line 19: Would'nt it be possible to create a regular drop-in file via osc.files? I think this way we could avoid a daemon-reload.

pkg/apis/extensions/v1alpha1/types_operatingsystemconfig.go
nimrodoron: Well, I had this dilemma as well.
I decided to put it here because it is ContainerD configurations (CRI config) not exactly CR extensions but CR Extensions do use it.
I can move it to CR it you prefer.

rfranzke: Hm, yes, CR extensions use it, so it's probably safer to provide them the path where to put the binaries via the spec. This way they don't need to rely on a constant. The constant, however, can be kept so that gardener can use it while crafting the `ContainerRuntime` resource. WDYT?

nimrodoron: Ok, so do you mean that i would extended the Container runtime crd to contain the Path as well?

rfranzke: Yes, maybe add a new field `.spec.binaryPath` to the `ContainerRuntime` CRD .

nimrodoron: Ok i will

charts/seed-operatingsystemconfig/original/values.yaml
```diff
@@ -18,8 +22,10 @@ kubernetes:
   version: 1.11.2
 worker:
   name: cpu-worker
-# cri:
+#  cri:
```
vpnachev:
```suggestion
# cri:
```

docs/extensions/containerruntime.md
```diff
@@ -0,0 +1,37 @@
+# Gardener Container Runtime Extension
+
+At the lowest layers of a Kubernetes node is the software that, among other things, starts and stops containers. It is called “Container Runtime”. 
+The most widely known container runtime is Docker, but it is not alone in this space. In fact, the container runtime space has been rapidly evolving. 
+
+Kubernetes support different container runtimes using  Container Runtime Interface (CRI) – a plugin interface which enables kubelet to use a wide variety of container runtimes.
+
+Gardener supports creation of Worker machines using CRI, more information can be found here: [CRI Support](https://https://github.com/gardener/gardener/blob/master/docs/extensions/operatingsystemconfig.md)).
```
vpnachev:
```suggestion
Gardener supports creation of Worker machines using CRI, more information can be found here: [CRI Support](operatingsystemconfig.md#cri-support).
```

pkg/apis/extensions/v1alpha1/types_containerruntime.go
```diff
@@ -0,0 +1,77 @@
...
+var _ Object = (*ContainerRuntime)(nil)
+
+// ContainerRuntimeResource is a constant for the name of the Container Runtime Extension resource.
+const ContainerRuntimeResource = "ContainerRuntime"
+
+// CRINameWorkerLabel is the name of the label describing the CRI name used in this node.
+const CRINameWorkerLabel = "worker.gardener.cloud/cri-name"
+
+// ContainerRuntimeNameWorkerLabel is a label describing a Container Runtime which should be supported on the node.
+const ContainerRuntimeNameWorkerLabel = "containerruntime.worker.gardener.cloud/%s"
```
vpnachev: Put all constants in a const block
```golang
const (
...
)
```

```suggestion
const (
// ContainerRuntimeResource is a constant for the name of the Container Runtime Extension resource.
ContainerRuntimeResource = "ContainerRuntime"
// CRINameWorkerLabel is the name of the label describing the CRI name used in this node.
CRINameWorkerLabel = "worker.gardener.cloud/cri-name"
// ContainerRuntimeNameWorkerLabel is a label describing a Container Runtime which should be supported on the node.
ContainerRuntimeNameWorkerLabel = "containerruntime.worker.gardener.cloud/%s"
)
```

pkg/apis/extensions/v1alpha1/types_containerruntime.go
```diff
@@ -0,0 +1,77 @@
...
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ContainerRuntime is a specification for a container runtime resource.
+type ContainerRuntime struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
```
vpnachev, line 38: Why is `+optional` needed?

pkg/apis/extensions/v1alpha1/types_containerruntime.go
```diff
@@ -0,0 +1,77 @@
...
+func (i *ContainerRuntime) GetExtensionStatus() Status {
+	return &i.Status
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ContainerRuntimeList is a list of ContainerRuntime resources.
+type ContainerRuntimeList struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
```
vpnachev, line 59: Is it needed here, too?

vpnachev (changes requested):
minor nits and questions

pkg/apis/extensions/v1alpha1/types_containerruntime.go
rfranzke: I think we are doing this for most types, e.g. https://github.com/gardener/gardener/blob/master/pkg/apis/core/v1beta1/types_cloudprofile.go#L30, although it's not needed in general. This was the upstream k/k style of doing it.

docs/extensions/containerruntime.md
```diff
@@ -0,0 +1,37 @@
+# Gardener Container Runtime Extension
+
+At the lowest layers of a Kubernetes node is the software that, among other things, starts and stops containers. It is called “Container Runtime”. 
+The most widely known container runtime is Docker, but it is not alone in this space. In fact, the container runtime space has been rapidly evolving. 
+
+Kubernetes support different container runtimes using  Container Runtime Interface (CRI) – a plugin interface which enables kubelet to use a wide variety of container runtimes.
```
rfranzke:
```suggestion
Kubernetes supports different container runtimes using Container Runtime Interface (CRI) – a plugin interface which enables kubelet to use a wide variety of container runtimes.
```

docs/extensions/containerruntime.md
```diff
@@ -0,0 +1,37 @@
...
+
+Gardener supports creation of Worker machines using CRI, more information can be found here: [CRI Support](https://https://github.com/gardener/gardener/blob/master/docs/extensions/operatingsystemconfig.md)).
+
+## Motivation
+
+Prior to the `Container Runtime Extensibility` concept, Gardener used Docker as the only
+container runtime to use in Worker machines. Because of the wide variety of different container runtimes
```
rfranzke:
```suggestion
container runtime to use in shoot worker machines. Because of the wide variety of different container runtimes
```

docs/extensions/containerruntime.md
```diff
@@ -0,0 +1,37 @@
...
+offers multiple important features (For example enhanced security concepts) it is important to enable end users to use other container runtimes as well.
+
+## The ContainerRuntime Extensions Resource
```
rfranzke:
```suggestion
## The `ContainerRuntime` Extension Resource
```

docs/extensions/containerruntime.md
```diff
@@ -0,0 +1,37 @@
...
+spec:
+  binaryPath: /var/bin/containerruntimes
+  type: gvisor
+```
+
+## Supporting a new Container Runtime Provider
+
+To add support for another container runtime (e.g., gvisor, kata-containers, etc.) a container runtime extension controller needs to be implemented. It should support Gardener's supported CRI plugins.
+
+The container runtime extension should install the necessary resources in the shoot and copy the runtime binaries to the relevant Worker machines in path: `spec.binaryPath`.
```
rfranzke:
```suggestion
The container runtime extension should install the necessary resources into the shoot cluster (e.g., `RuntimeClass`es), and it should copy the runtime binaries to the relevant worker machines in path: `spec.binaryPath`. 
```

rfranzke: Can you elaborate here on the labels that Gardener adds to every "relevant worker machine" so that extension developers know how to find them? Also, I guess makes sense to shortly explain the example of how to install such binaries on a node (run a DaemonSet that copies them over from a Docker registry, avoid downloading from internet to also cater with isolated environments).

pkg/operation/botanist/containerruntime.go
```diff
@@ -0,0 +1,215 @@
...
+						toApply = extensionsv1alpha1.ContainerRuntime{
+							ObjectMeta: metav1.ObjectMeta{
+								Name:      containerRuntime.Type,
+								Namespace: b.Shoot.SeedNamespace,
+							},
+						}
+					)
+
+					fns = append(fns, func(ctx context.Context) error {
+
```
rfranzke:
```suggestion
```

rfranzke (changes requested):
Looks great, a few more minor suggestions. Afterwards, I'll definitely have to play with it. Looking forward! :)

1. I think we forgot to adapt https://github.com/gardener/gardener/blob/master/docs/extensions/operatingsystemconfig.md#contract-operatingsystemconfig-resource ("The operating system must have built-in Docker support.") - can you also add containerd here?
1. Can you add an example how to configure container runtimes [to this file](https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml#L31-L32) (maybe similar to [this](https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml#L214-L218))?

Just a heads-up: If #2066 gets merged before this PR then the deletion calls for `ContainerRuntime`s in `pkg/operation/botanist/containerruntime.go` have to be adapted [similar to this](https://github.com/gardener/gardener/pull/2066/files#diff-be894e0792d0d6563ec0b8bbcaa1073eR79-R81).

pkg/apis/extensions/v1alpha1/types_containerruntime.go
nimrodoron: @rfranzke Do you want me to remove it or keep the current standard?

pkg/apis/extensions/v1alpha1/types_containerruntime.go
nimrodoron: @rfranzke  same question as above

nimrodoron:
> 1. https://github.com/gardener/gardener/blob/master/docs/extensions/operatingsystemconfig.md#contract-operatingsystemconfig-resource
I didn't put it in purpose, it is written in the CRI section if the OS extension support this CRI.
Do you think we should it there as well?

pkg/apis/extensions/v1alpha1/types_containerruntime.go
rfranzke: Hm, I don't care too much. For me both is okay.

nimrodoron: Ok so i keep it like that.

pkg/apis/extensions/v1alpha1/types_containerruntime.go
nimrodoron: Same as above

rfranzke:
Ah, I see, makes sense, thank you @nimrodoron :)

danielfoehrKn (changes requested):

charts/seed-operatingsystemconfig/original/templates/scripts/_init-containerd.sh
nimrodoron: Hi no because the file is generated with a script to concat the $PATH variable. Each OS has a different PATH and i didn't find a way to concat the PATH in a static drop in file.

danielfoehrKn: ok got it thanks!

rfranzke (commented):
PR looks great now, will check it out soon.

rfranzke (changes requested):
One thing I forgot: Can you extend the `ExtensionValidation` admission plugin  so that it checks that only registered `ContainerRuntime` resources can be used in `Shoot`?
Basically, similar to how the `OperatingSystemConfig`s are checked: https://github.com/gardener/gardener/blob/master/plugin/pkg/global/extensionvalidation/admission.go#L254-L260

nimrodoron:
> One thing I forgot: Can you extend the `ExtensionValidation` admission plugin so that it checks that only registered `ContainerRuntime` resources can be used in `Shoot`?
> Basically, similar to how the `OperatingSystemConfig`s are checked: https://github.com/gardener/gardener/blob/master/plugin/pkg/global/extensionvalidation/admission.go#L254-L260

Done

plugin/pkg/global/extensionvalidation/admission.go
```diff
@@ -252,6 +252,12 @@ func (e *ExtensionValidator) validateShoot(kindToTypesMap map[string]sets.String
 	}
 
 	for i, worker := range spec.Provider.Workers {
+		if worker.CRI != nil {
```
rfranzke, line 255: Great thanks. Can we also have a unit test, please? :)

nimrodoron: Done as well

rfranzke (approved):
/lgtm, played with it locally 👍

danielfoehrKn (approved):
Looks good to me. Though I have not made an end-to-end test yet :) 
I guess that is only possible when the gvisor extension is ready.

PR #2046, opened 2020-03-14
Added new metric to monitor CPU throttling

ggaurav10:
Added new metric to monitor CPU throttling

Fixes #

**Release note**:

```improvement operator
Added new metric to monitor CPU throttling
```

### DIFF
--- a/charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
+++ b/charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
@@ -5,6 +5,12 @@ groups:
   - record: seed:container_cpu_usage_seconds_total:sum_by_container
     expr: sum(rate(container_cpu_usage_seconds_total[5m])) by (container)
 
+  - record: seed:container_cpu_cfs_throttled_seconds_total:sum_by_container
+    expr: sum(rate(container_cpu_cfs_throttled_seconds_total[5m])) by (container)
+
+  - record: seed:container_cpu_cfs_throttled_periods_total:sum_by_container
+    expr: sum(rate(container_cpu_cfs_throttled_periods_total[5m])) by (container)
+
   - record: seed:container_memory_working_set_bytes:sum_by_container
     expr: sum(container_memory_working_set_bytes) by (container)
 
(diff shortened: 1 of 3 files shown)


vpnachev:
@ggaurav10 could you explain why these metrics are needed?
My concern is to not overload prometheus performance.

rfranzke (dismissed):

ggaurav10:
yes @vpnachev, this was my concern as well, hence the delay between making the change and raising a PR :)
Recent issue in our seeds, and a possibility of CPU throttling was the motivation behind finally raising this PR. This would help in figuring out if any of the pods are/were cpu throttled. I am not sure if there is any other way this could be figured out, since unlike memory (oomkills) there is no other trace of cpu throttling.

amshuman-kr (changes requested):
@ggaurav10 Thanks for adding the metric.

Can you please also add seed-wide aggregate metrics (sum, sum_by_container and sum_cp) [here](https://github.com/gardener/gardener/blob/master/charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml)?

LGTM otherwise.

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
```diff
@@ -5,6 +5,12 @@ groups:
   - record: seed:container_cpu_usage_seconds_total:sum_by_container
     expr: sum(rate(container_cpu_usage_seconds_total[5m])) by (container)
 
+  - record: seed:container_cpu_cfs_throttled_seconds_total:sum_by_container
+    expr: sum(rate(container_cpu_cfs_throttled_seconds_total[5m])) by (container)
```
amshuman-kr, line 9: Does is make sense to aggregate the rate? `container_cpu_cfs_throttled_seconds_total` is not about how much CPU cores are throttled, right?

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
```diff
@@ -5,6 +5,12 @@ groups:
...
+
+  - record: seed:container_cpu_cfs_throttled_periods_total:sum_by_container
+    expr: sum(rate(container_cpu_cfs_throttled_periods_total[5m])) by (container)
```
amshuman-kr, line 12: Same as above.

amshuman-kr (commented):
Is the aggregation based on rate appropriate?

amshuman-kr (approved):
Thanks @ggaurav10 for the links.

https://rancher.com/docs/rancher/v2.x/en/cluster-admin/tools/monitoring/expression/
https://opensource.zalando.com/skipper/operation/operation/

wyb1 (approved):
/lgtm

vpnachev (approved):
/approve

mwennrich:
So this will be in the next release?

rfranzke:
@mwennrich yes, planned for this week.

PR #2049, opened 2020-03-16
ExtensionValidation admission plugin

### LINKED ISSUE #1844: Admission plugin should validate extension types in resources

rfranzke:
With this PR we add a new `ExtensionValidation` admission plugin to the Gardener API server. The admission plugin is creating on `CREATE` and `UPDATE` operations of `BackupBucket`s, `BackupEntry`s, `Seed`s, and `Shoot`s. It validates for the various extension types in the respective `spec`s that a corresponding `ControllerRegistration` exists whose `spec` indicates that it is responsible for the extension type(s).

Fixes #1844

**Release note**:

```improvement operator
It is no longer possible to submit `BackupBucket, `BackupEntry`, or `Seed` resources whose specifications target extension types that are not registered in the system.
```
```improvement user
It is no longer possible to submit `Shoot` resources whose specification targets extension types that are not registered in the system.
```

### DIFF
--- a/plugin/pkg/global/extensionvalidation/admission.go
+++ b/plugin/pkg/global/extensionvalidation/admission.go
@@ -1,0 +1,311 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package extensionvalidation
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/gardener/gardener/pkg/apis/core"
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	externalcoreinformers "github.com/gardener/gardener/pkg/client/core/informers/externalversions"
+	corev1beta1listers "github.com/gardener/gardener/pkg/client/core/listers/core/v1beta1"
+
+	dnsv1alpha1 "github.com/gardener/external-dns-management/pkg/apis/dns/v1alpha1"
+	admissioninitializer "github.com/gardener/gardener/pkg/apiserver/admission/initializer"
+	"github.com/hashicorp/go-multierror"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/admission"
+)
+
+const (
+	// PluginName is the name of this admission plugin.
+	PluginName = "ExtensionValidator"
+)
+
+// Register registers a plugin.
+func Register(plugins *admission.Plugins) {
+	plugins.Register(PluginName, NewFactory)
+}
+
+// NewFactory creates a new PluginFactory.
+func NewFactory(config io.Reader) (admission.Interface, error) {
+	return New()
+}
+
+// ExtensionValidator contains listers and admission handler.
+type ExtensionValidator struct {
+	*admission.Handler
+	controllerRegistrationLister corev1beta1listers.ControllerRegistrationLister
+	backupBucketLister           corev1beta1listers.BackupBucketLister
+	readyFunc                    admission.ReadyFunc
+}
+
+var (
+	_ = admissioninitializer.WantsExternalCoreInformerFactory(&ExtensionValidator{})
+
+	readyFuncs []admission.ReadyFunc
+)
+
+// New creates a new ExtensionValidator admission plugin.
+func New() (*ExtensionValidator, error) {
+	return &ExtensionValidator{
+		Handler: admission.NewHandler(admission.Create, admission.Update),
+	}, nil
+}
+
+// AssignReadyFunc assigns the ready function to the admission handler.
+func (e *ExtensionValidator) AssignReadyFunc(f admission.ReadyFunc) {
+	e.readyFunc = f
+	e.SetReadyFunc(f)
+}
+
+// SetExternalCoreInformerFactory sets the external garden core informer factory.
+func (e *ExtensionValidator) SetExternalCoreInformerFactory(f externalcoreinformers.SharedInformerFactory) {
+	controllerRegistrationInformer := f.Core().V1beta1().ControllerRegistrations()
+	e.controllerRegistrationLister = controllerRegistrationInformer.Lister()
+
+	backupBucketInformer := f.Core().V1beta1().BackupBuckets()
+	e.backupBucketLister = backupBucketInformer.Lister()
+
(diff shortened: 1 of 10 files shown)


plugin/pkg/global/extensionvalidation/admission.go
```diff
@@ -0,0 +1,307 @@
...
+	extensionKind string
+	extensionType string
+	message       string
+}
+
+type requiredExtensions []requiredExtension
+
+func (r requiredExtensions) areRegistered(kindToTypesMap map[string]sets.String) error {
+	for _, requiredExtension := range r {
+		if err := isExtensionRegistered(kindToTypesMap, requiredExtension.extensionKind, requiredExtension.extensionType, requiredExtension.message); err != nil {
```
timebertt, line 279: Can you remove this early exit and use a multierror to collect all invalid extension types?

plugin/pkg/global/extensionvalidation/admission_test.go
```diff
@@ -0,0 +1,381 @@
...
+				registerAllExtensions()
+
+				controllerRegistration := createControllerRegistrationForKindType(registration.extensionKind, registration.extensionType)
+				gardenExternalCoreInformerFactory.Core().V1beta1().ControllerRegistrations().Informer().GetStore().Delete(controllerRegistration)
+
+				attrs := admission.NewAttributesRecord(seed, nil, core.Kind("Seed").WithVersion("version"), seed.Namespace, seed.Name, core.Resource("seeds").WithVersion("version"), "", admission.Create, &metav1.DeleteOptions{}, false, nil)
+
+				err := admissionHandler.Validate(context.TODO(), attrs, nil)
+
+				Expect(err).To(HaveOccurred(), fmt.Sprintf("expected that extension %s is not registered", controllerRegistration.Name))
```
timebertt, line 206: Can you add something like this to ensure the validation fails because of the current `registration`?

```suggestion
				Expect(err).To(HaveOccurred(), fmt.Sprintf("expected that extension %s is not registered", controllerRegistration.Name))
				Expect(err).To(ContainSubstring(controllerRegistration.Name))
```

plugin/pkg/global/extensionvalidation/admission_test.go
```diff
@@ -0,0 +1,381 @@
...
+				registerAllExtensions()
+
+				controllerRegistration := createControllerRegistrationForKindType(registration.extensionKind, registration.extensionType)
+				gardenExternalCoreInformerFactory.Core().V1beta1().ControllerRegistrations().Informer().GetStore().Delete(controllerRegistration)
+
+				attrs := admission.NewAttributesRecord(shoot, nil, core.Kind("Shoot").WithVersion("version"), shoot.Namespace, shoot.Name, core.Resource("shoots").WithVersion("version"), "", admission.Create, &metav1.DeleteOptions{}, false, nil)
+
+				err := admissionHandler.Validate(context.TODO(), attrs, nil)
+
+				Expect(err).To(HaveOccurred(), fmt.Sprintf("expected that extension %s is not registered", controllerRegistration.Name))
```
timebertt, line 305: Same here

plugin/pkg/global/extensionvalidation/admission_test.go
```diff
@@ -0,0 +1,381 @@
...
+			}
+		})
+
+		It("should prevent the object from being created because no extension type is not registered", func() {
```
timebertt:
```suggestion
		It("should prevent the object from being created because no extension type is registered", func() {
```

plugin/pkg/global/extensionvalidation/admission_test.go
```diff
@@ -0,0 +1,381 @@
...
+
+				attrs := admission.NewAttributesRecord(seed, nil, core.Kind("Seed").WithVersion("version"), seed.Namespace, seed.Name, core.Resource("seeds").WithVersion("version"), "", admission.Create, &metav1.DeleteOptions{}, false, nil)
+
+				err := admissionHandler.Validate(context.TODO(), attrs, nil)
+
+				Expect(err).To(HaveOccurred(), fmt.Sprintf("expected that extension %s is not registered", controllerRegistration.Name))
+			}
+		})
+
+		It("should prevent the object from being created because no extension type is not registered", func() {
```
timebertt:
```suggestion
		It("should prevent the object from being created because no extension type is registered", func() {
```

plugin/pkg/global/extensionvalidation/admission_test.go
```diff
@@ -0,0 +1,381 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(dr.Handles(admission.Create)).To(BeTrue())
+			Expect(dr.Handles(admission.Update)).To(BeTrue())
+			Expect(dr.Handles(admission.Connect)).NotTo(BeTrue())
+			Expect(dr.Handles(admission.Delete)).NotTo(BeTrue())
+		})
+	})
+
+	Describe("#ValidateInitialization", func() {
+		It("should return error if no ShootStateLister is set", func() {
```
timebertt:
```suggestion
		It("should return error if ControllerRegistrationLister and BackupBucketLister are not set", func() {
```

plugin/pkg/global/extensionvalidation/admission_test.go
```diff
@@ -0,0 +1,381 @@
...
+			dr, _ := New()
+
+			err := dr.ValidateInitialization()
+
+			Expect(err).To(HaveOccurred())
+		})
+
+		It("should not return error if ShootStateLister and core client is set", func() {
```
timebertt:
```suggestion
		It("should not return error if ControllerRegistrationLister and BackupBucketLister are set", func() {
```

timebertt (changes requested):
Thanks, /lgtm overall!
Just some small suggestions :)

timebertt (approved):
Thanks!

/approve

plugin/pkg/shoot/validator/admission.go
```diff
@@ -148,30 +148,28 @@ func (v *ValidateShoot) Admit(ctx context.Context, a admission.Attributes, o adm
...
 		}
+		oldShoot = old
 
 		// do not ignore metadata updates to detect and prevent removal of the gardener finalizer or unwanted changes to annotations
-		if reflect.DeepEqual(newShoot.Spec, oldShoot.Spec) && reflect.DeepEqual(newShoot.ObjectMeta, oldShoot.ObjectMeta) {
+		if reflect.DeepEqual(shoot.Spec, oldShoot.Spec) && reflect.DeepEqual(shoot.ObjectMeta, oldShoot.ObjectMeta) {
 			return nil
 		}
-
-		if newShoot.Spec.Provider.Type != oldShoot.Spec.Provider.Type {
```
timuthy, line 167: Can you shortly explain why this check is removed?

rfranzke: It is already validated in the static validation (`pkg/core/validation)` and isn't necessary to be validated here in the dynamic validation again.

timuthy (approved):
/lgtm

Kudos for the readability!

PR #2051, opened 2020-03-16
Add Component interfaces and associated methods

mvladev:

It adds high-level interfaces which can be used by every single component in the Shoot reconciliation to have its life-cycle fully encapsulated, mocked for testing and isolated from other components

Individual components such as `DNSProvider` or `DNSEntry` can implement one of the interfaces `Deployer`, `Waiter` or/and `DeployWaiter`

Fixes #

**Release note**:

```improvement developer
Added high-level interfaces which can be used by different components in the Shoot reconciliation. It allows of life-cycle to be fully encapsulated, mocked for testing and isolation from other components
```


pkg/operation/botanist/component/deploy.go
```diff
@@ -0,0 +1,131 @@
...
+func NoOp() DeployWaiter { return &deploy{} }
+
+type deploy struct {
+	invert bool
+	wait   bool
+	dw     []DeployWaiter
+}
+
+func (d *deploy) Deploy(ctx context.Context) error {
+	if d.invert {
```
rfranzke, line 58: What is the purpose of this? There is already a `Destroy` function - why would I set `invert=true` and call `Deploy` instead of calling `Destroy` directly?

If I look at above functions `OpDestroy`, etc. then I assume that the `Destroy` function of the `Deployer` interface is never called? I'm confused, tbh. Can we get rid of the `invert` field and make it rather explicit what is called?

pkg/operation/botanist/component/deploy.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, dw := range d.dw {
+		if dw == nil {
+			continue
+		}
+
+		if err := dw.Deploy(ctx); err != nil {
+			return err
+		}
+
+		if d.wait {
```
rfranzke: Hm, why is `wait` a "global" setting here? What if I want to wait only for a few dedicated `Deploy` calls while I want to simply run `Deploy` without waiting for the others?

pkg/operation/botanist/component/interfaces.go
```diff
@@ -0,0 +1,41 @@
...
+	// Destroy already deployed component.
+	Destroy(ctx context.Context) error
+}
+
+// Waiter waits for life-cycle operations of a component to finish.
+type Waiter interface {
+	// Wait for deployment to finish and component to report ready.
+	Wait(ctx context.Context) error
+	// WaitCleanup for destruction to finish and component to be fully removed.
+	WaitCleanup(ctx context.Context) error
```
timuthy, line 34: Why do we need a separate `WaitCleanup` function instead of reusing `Wait` for deploy and destroy?

pkg/operation/botanist/component/deploy.go
mvladev: The principle is simple - if a component was installed in reconcilationA then removed in reconcilationB (remove DNS entries, DNSProviders and so), the reconciler has to explicitly know the previous state and call `Destroy` explicitly to cleanup left over resources. With this, you can construct a component based on the actual state you want to have in the system.

Example with some pseudo code - when constructing the botanist you already know the desired state of the `Foo` component - to deploy or to remove. With this the create will simply call `Foo.Deploy()` and this would clean up old resources.

```go

type Botanist struct {
	Foo component.DeployWaiter
}

func NewBotanist() *Botanist {
	b := &Botanist{}
	if fooEnabled {
		b.Foo = NewFoo()
	} else {
		b.Foo = OpDestroy(NewFoo())
	}

	return b
}


// Create step 
_ = g.Add(flow.Task{
		Name:         "Ensure Foo",
		Fn:           flow.TaskFn(botanist.Foo.Deploy)
})
```

pkg/operation/botanist/component/deploy.go
mvladev: This is mainly used to combine some steps which are always combined together - for example, deploy DNSProvider, wait for it to be ready, then deploy DNSEntry, wait for it to be ready:


```go
_ = g.Add(flow.Task{
		Name: "Deploying internal domain DNS record",
		Fn: flow.TaskFn(func(ctx context.Context) error {
				return component.OpWaiter(
						botanist.Shoot.Components.DNS.InternalProvider,
						botanist.Shoot.Components.DNS.InternalEntry,
				).Deploy(ctx)
		}).DoIf(wakeupRequired),
		Dependencies: flow.NewTaskIDs(wakeUpControlPlane),
})
```

pkg/operation/botanist/component/deploy.go
rfranzke: I see... but wouldn't the following be as convenient as above while allowing to remove the `invert` logic?

```golang
type Botanist struct {}

func NewBotanist() *Botanist {
	b := &Botanist{}

	var ensureFooTask flow.TaskFn
	if fooEnabled {
		ensureFooTask = flow.TaskFn(NewFoo().Deploy)
	} else {
		ensureFooTask = flow.TaskFn(NewFoo().Destroy)
	}

	return b
}


// Create step 
_ = g.Add(flow.Task{
		Name:         "Ensure Foo",
		Fn:           ensureFooTask,
})
```

pkg/operation/botanist/component/deploy.go
rfranzke: Sure, there are use-cases where this perfectly fits. I was more looking at the `Deploy` function from a general point of view. As it is now I would not be able to use it if I only want to wait for some `Deploy` steps and not for others. Wouldn't it make sense to at least allow the possibility to specify the wait behaviour per `DeployWaiter` interface in the list (optionally, default is "always wait")?

mvladev: > Sure, there are use-cases where this perfectly fits. I was more looking at the Deploy function from a general point of view. As it is now I would not be able to use it if I only want to wait for some Deploy steps and not for others. 

This is why I've separated the life-cycle operations into 4 steps `Deploy`, `Wait`, `Destroy`, `WaitCleanup`. With the helper functions `OpDestroy`, `OpDestroyAndWait`, `OpWaiter` you can combine different steps - `Deploy + Wait` and multiple `DeployWaiter` into various combinations without the need for modifying the underlying component implementation.

pkg/operation/botanist/component/interfaces.go
mvladev: Mainly because it would force each component into implement some internal logic for keeping the state of which operation it should call. What would happen if `Deploy` is called and then `Destroy`. `Wait` would either have to use some mutex sync to keep state of the last operation and possibly block.

pkg/operation/botanist/component/deploy.go
rfranzke: OK, I see. Thanks 👍

pkg/operation/botanist/component/deploy.go
mvladev: I also tried playing with this, but the problem with this approach is that in some cases a component have to be deployed, then removed (deploying kube-apiserver for example when deleting).

rfranzke: Hm, can you elaborate more why this does not work, maybe again with pseudo-code (in comparison to how it work with the current approach)?

mvladev: I've renamed the field to `deleteOnly` to make more clear.

rfranzke (approved):

timuthy (approved):

PR #2059, opened 2020-03-17
fix containerd config generation & mtu service

### LINKED ISSUE #2058: Containerd config error
**What happened**:
Currently the OperatingSystemConfig original generated by the Gardener contains a bug when using containerd.
The systemd drop -in file is malformed.
[This is the code in question](https://github.com/gardener/gardener/blob/ed24799c78bebc1d5d66c5ee542e0da03f9b1cc8/charts/seed-operatingsystemconfig/original/templates/containerd/_containerd.service#L1):  

**Result**
- containerd is being executed without the default config file
  - this will become a problem when introducing the ContainerRuntimes because these are being configured in the containerd config file
- the mtu-customizer.service is not being created
- the containerd.service is in error state because it cannot load/ malconfigured a drop-in file
- the update-ca-certs.service failed due to dependency on docker.service. Docker.service cannot be restarted due to containerd.service error.

A default containerd config file should be generated.

1. enable CRI for the shoot & use an ubuntu image
2. Make sure to use an update ubuntu-os extension containing the containerd changes
3. reconcile shoot
4. get the OperatingSystemConfig secret containing the original config
5. Observe that the mtu-customizer.service is malformed
image

6. Get a shell to the node  
7.  execute `systemctl list-units --type=service | grep mtu-customizer` 
Observe that there is no mtu-customizer service running

8. execute ` ls etc/containerd`
Observe that there is no config.yaml for containerd.

9. execute `systemctl status containerd.service`
Observe that it is in an error state due to a drop-in file.

**Environment**:

- Gardener version: (current master)
1.17.3
AWS
- Others:
Ubuntu machine image

danielfoehrKn:
Introduces a new systemd service that generates the containerd config file.
The containerd.service  is being configured to use this config file.

Fixes #2058 

**Release note**:

```improvement operator
NONE
```


danielfoehrKn:
@nimrodoron I have changed the the systemd units to ensure that the containerd config file is being created before the execution of the containerd.service. PTAL

charts/seed-operatingsystemconfig/original/templates/osc.yaml
```diff
@@ -38,14 +38,15 @@ spec:
...
-{{ include "containerd-service" . | indent 2 }}
+{{ include "containerd-config-generator" . | indent 2 }}
 {{- end }}
 {{- end }}
   files:
 {{ include "root-certs" . | indent 2 }}
 {{- if .Values.worker.cri }}
 {{- if eq .Values.worker.cri.name .Values.osc.cri.containerdName }}
 {{ include "init-containerd-config-script" . | indent 2 }}
+{{ include "containerd-require-config-generator" . | indent 2 }}
```
rfranzke:
```suggestion
{{- if .Values.worker.cri }}
{{- if eq .Values.worker.cri.name .Values.osc.cri.containerdName }}
{{ include "containerd-require-config-generator" . | indent 2 }}
{{- end }}
{{- end }}
```

charts/seed-operatingsystemconfig/original/templates/containerd/containerd-config-generator.service
```diff
@@ -0,0 +1,15 @@
...
+    [Unit]
+    Description=Containerd config generator
+    [Install]
+    WantedBy=multi-user.target
+    [Service]
+    Type=oneshot
+    RemainAfterExit=yes
+    ExecStart=/opt/bin/init-containerd
+{{- end -}}
+
```
rfranzke:
```suggestion
```

charts/seed-operatingsystemconfig/original/templates/containerd/_containerd.service
```diff
@@ -1,10 +0,0 @@
-{{- define "containerd-service" -}}
-- name: containerd.service
-  command: restart
-  enable: true
-  dropIns:
```
rfranzke, line 5: @nimrodoron does the whole containerd instrumentation still work now? I guess the service was named `containerd.service` on purpose and designed as drop-in?

rfranzke (changes requested):

charts/seed-operatingsystemconfig/original/templates/containerd/_containerd.service
nimrodoron: First ContainerD was started with the default configurations of the image and later on it was restarted after the configurations has changed.
There is also important order of startup of ContainerD and Docker, After this change it should be checked that both of them are running ok. 
Finally the ContainerD is running with the default config.toml so there is not problem with Container Runtimes and also docker restarted successfully.

nimrodoron: By the way, For Ubuntu, the default configuration is similar to the generated config.toml.
The socket path is not changed (Like in CoreOS or SuSe) So the ContainerD works well before and after the restart. Also ContainerD starting well also if the config file is not there with default configurations.

rfranzke: Does it mean you are OK with this change?

nimrodoron: Well, for Ubuntu at the moment it seems ok if it works. But At the end of the initialization ContainerD & Docker worked fine, we both checked, and ContainerD did use the config.toml.
Are you sure there is a bug after the initialization is finished?

For CoreOS and SuSe where the config.toml changing the socket path and docker should be restart after the ContainerD with the new configurations so i'm not sure if this will work well, but currently we didn't push them because of other reasons...

charts/seed-operatingsystemconfig/original/templates/osc.yaml
danielfoehrKn: Why would that be required? We already have that check in place?

charts/seed-operatingsystemconfig/original/templates/containerd/_containerd.service
danielfoehrKn: I am not quite sure if I understand what you mean here. 
The changes introduced in this PR make sure that
- the config file is generated
- introduces a dependency from the containerd.service to the config-generator.service. That means on each cloud-config update, the generator service will be restarted to generate the config file and then the dependent containerd.service will be restarted as well. 
I think that is the same thing you aimed just realized in a different way
. 
Problem was, that there was a configuration error of the drop-in file for the containerd.service so that the mtu-customizer service was actually placed inside the 10-pre-xxx.conf file & not the desired content.
I introduced another service to have a separation between the generation of the config file and the containerd execution that is logically dependent on it. 

Regarding the socket path change on CoreOS and Suse: the socket path is configured in the config toml as you mentioned. The generator service, consequently also the containerd service should be restarted when the cloud-config changes (by the cloud-config execution script).  Also the PartOf dependecy makes sure that the generation of the config file is complete before containerd is restarted, so that it is configured with the new socket path.


Let me know if that makes sense and of course please test it. Also check that without the PR there are several units in an erroneous state.

nimrodoron: Ok, sounds pretty much similar. 

Regarding the details in the bug:
I didn't check the mtu-customizer but:
8. execute ls etc/containerd
Observe that there is no config.yaml for containerd.

9. execute systemctl status containerd.service
Observe that it is in an error state due to a drop-in file.

From my checks it worked ok the file created and containerd service worked well.
If containerd didn't work the Kubelet should have fail as well and the machine won't join the cluster, which did happen.

danielfoehrKn: The erroneous drop-in file nor the absence of the config file prevent the containerd process from starting. From the kubelets perspective the containerd socket was reachable and usable under its default socket path.

The problems will arise when configuring the non-existing config.toml with ContainerRuntimes

nimrodoron: If containerd service is down the socket is not reachable and Kubelet can't work.

I have just created a cluster with Ubuntu.
config.toml was created successfully.
containerd.service status is running successfully.

`containerd.service - containerd container runtime
   Loaded: loaded (/lib/systemd/system/containerd.service; enabled; vendor preset: enabled)
  Drop-In: /etc/systemd/system/containerd.service.d
           └─10-exec-start-pre-init-config.conf, 11-exec_config.conf
   Active: active (running) since Wed 2020-03-18 10:32:22 UTC; 1h 27min ago
     Docs: https://containerd.io
  Process: 5323 ExecStartPre=/opt/bin/init-containerd (code=exited, status=0/SUCCESS)
  Process: 5321 ExecStartPre=/sbin/modprobe overlay (code=exited, status=0/SUCCESS)
 Main PID: 5392 (containerd)
    Tasks: 417
   CGroup: /system.slice/containerd.service
           ├─ 5392 /usr/bin/containerd --config=/etc/containerd/config.toml`

I don't find mtu-customizer service as you mentioned.

I don't mind if you apply this change if it solving the mtu-customizer, i don't know this service.

nimrodoron: Ok, talking with @rfranzke i understood the bug is on AWS. I checked on GCP where everything was fine. I will check the fix on GCP.

danielfoehrKn: ok makes sense. Apparently the mtu-customizer service is only added on AWS, thats why it has not been observed on GCP. Thanks for checking

nimrodoron: Ok I checked your change in GCP as well, Works. great :)

nimrodoron (dismissed):

charts/seed-operatingsystemconfig/original/templates/containerd/containerd-config-generator.service
zanetworker: this just calls the contanerd init, maybe a better name `containerd-initializer.service`?

rfranzke (dismissed):

charts/seed-operatingsystemconfig/original/templates/containerd/containerd-require-config-generator.yaml
```diff
@@ -0,0 +1,12 @@
...
+- path: /etc/systemd/system/containerd.service.d/10-require-config-generator.conf
+  permissions: 0644
+  content:
+    inline:
+      encoding: ""
+      data: |
+        [Unit]
+        After=containerd-config-generator.service
+        Requires=containerd-config-generator.service
+        PartOf=containerd-config-generator.service
```
zanetworker: do your really need all three `After,Requires,PartOf` here?

charts/seed-operatingsystemconfig/original/templates/containerd/containerd-config-generator.service
```diff
@@ -0,0 +1,15 @@
...
+  enable: true
+  content: |
+    [Unit]
+    Description=Containerd config generator
+    [Install]
+    WantedBy=multi-user.target
+    [Service]
+    Type=oneshot
+    RemainAfterExit=yes
+    ExecStart=/opt/bin/init-containerd
```
zanetworker: This just calls `/opt/bin/init-containerd` maybe a better name would be `containerd-initializer.service`?

danielfoehrKn: Thanks for the suggestion, though I think containerd-config-generator.service is more precise at the moment as it really just starts the config generation. 
If there would be more initialization going on I would agree.
Would that be fine with you @zanetworker ?

charts/seed-operatingsystemconfig/original/templates/containerd/containerd-require-config-generator.yaml
danielfoehrKn: After is required for the timely ordering. So that the generator with type=oneshot completes before the containerd.service is started/restarted.

PartOf is needed so that the containerd service is restarted when the generator service is (also stopped but it is a type=oneshot service so its not relevant). 

Requires is needed so that the generator service is started when the containerd service is started.  Also instead of 'Wants' as this lets the containerd.service fail if the config generation.service failed to start. 

This is my current understanding. During my tests that worked, however I might be missing something here.  Would appreciate a verification.

charts/seed-operatingsystemconfig/original/templates/containerd/containerd-config-generator.service
zanetworker: I am fine, it was just a suggestion based on the fact that context of the script is not clear  from the binary executed, but again fine by me :)

charts/seed-operatingsystemconfig/original/templates/containerd/containerd-require-config-generator.yaml
zanetworker: Again a suggestion to avoid redundancy (having a script executed twice), based on my understanding `Requires` includes all cases but `PartOf` is only about starting and stopping, but I could be wrong.

charts/seed-operatingsystemconfig/original/templates/containerd/containerd-config-generator.service
danielfoehrKn: I changed my mind, I ll rename it as you suggested. There might be more changes included here in the future that goes beyond generating the config file.

charts/seed-operatingsystemconfig/original/templates/containerd/containerd-require-config-generator.yaml
danielfoehrKn: I tested it and based on that PartOf can be removed. Thanks!

danielfoehrKn:
PTAL - incorporated changes by @zanetworker

rfranzke (approved):

zanetworker (approved):
/lgtm

PR #2060, opened 2020-03-18
Added HPA metrics to kube-state-metrics in garden ns.

amshuman-kr:
**What this PR does / why we need it**:
Added HPA metrics to kube-state-metrics in garden ns. These can help in analysing HPA behaviour and in analysing the benefits and impact of reducing HPA scale down stabilization period (currently `24h`) to a lower value. Ideally, we should redule the HPA scale down stabilization duration to resource utilization and to avoid over-provisioning for longer periods.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
For the time being, I have not added any aggregate rules. If I find something useful as an aggregate metrics, I will file a separate PR later.

**Release note**:

```improvement operator
Added HPA metrics to kube-state-metrics in garden ns. These can help in analysing HPA behaviour and in analysing the benefits and impact of reducing HPA scale down stabilization period (currently `24h`) to a lower value.
```

PR #2061, opened 2020-03-18
Fix project controller test

mvladev:

Ordering of items in the queue is not guaranteed, because the list of projects is not always guaranteed to be in the same order.

Related https://github.com/gardener/gardener/pull/2060

**Release note**:

```improvement operator
NONE
```


timebertt (approved):
/lgtm, thanks for fixing

PR #2066, opened 2020-03-19
Gardener Seed Admission Controller

### LINKED ISSUE #2053: Seed Admission Webhook preventing accidental deletion of extension CR(D)s
Let's add a new separate binary to g/g that serves a validation webhook in the seed clusters in order to prevent accidental deletion of extension CR(D)s. It shall only allow the deletion if it was confirmed before (similar to how shoot/project validation is prevented).

**Why is this needed**:
Deleting an extension CRD can have catastrophic and hard-to-recover consequences, hence, let's protect them from accidental deletion.

rfranzke:
This PR introduces a new component, the Gardener Seed admission controller. It is deployed by the Gardenlet as part of its seed bootstrapping phase and, consequently, running in every seed cluster. It's main purpose is to serve webhooks (validating or mutating) in order to admit or deny certain requests to the seed's API server.
Concretely, it prevents the undesired deletion of `CustomResourceDefinitions` labeled with `gardener.cloud/deletion-protected=true`, and most custom resources of the `extensions.gardener.cloud/v1alpha1` API group if they were not previously annotated with `confirmation.gardener.cloud/deletion=true`.

Example logs:

```
$ kubectl -n shoot--foo--bar delete infra bar
Error from server: admission webhook "seed.admission.core.gardener.cloud" denied the request: must have a "confirmation.gardener.cloud/deletion" annotation to delete
```

Fixes #2053

The PR is split in three commits: The first commit introduces the gardener-seed-admission-controller, the second commit enhances the gardenlet code to a) deploy it as part of the seed bootstrap, and b) confirm the deletion for the extension resources prior to sending `DELETE` calls. The third commit adds a script for local development that adds an image vector overwrite for the Gardener components in the `charts/images.yaml`.

PS: It might happen that, when you check out this PR and start the Gardenlet locally, the gardener-seed-admission-controller cannot be deployed in your seeds because of `ImagePullBackOff`. This is because the Gardener CI pipeline has not yet been adapted to automatically build an image for this new component. This can only happen after the PR is merged. You can use the image `eu.gcr.io/gardener-project/gardener/seed-admission-controller:v1.2.0-dev` for testing the admission controller.

**Release note**:

```noteworthy operator
A new component has been introduced, the gardener-seed-admission-controller. It runs in every seed cluster and prevents the undesired deletion of `CustomResourceDefinitions` labeled with `gardener.cloud/deletion-protected=true`, and most custom resources of the `extensions.gardener.cloud/v1alpha1` API group if they were not previously annotated with `confirmation.gardener.cloud/deletion=true`.
```


charts/seed-bootstrap/templates/gardener-seed-admission-controller/validatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - apiGroups:
+      - apiextensions.k8s.io
+      apiVersions:
+      - v1alpha1
+      - v1beta1
+      - v1
+      operations:
+      - DELETE
+      resources:
+      - customresourcedefinitions
```
mvladev, line 19: Why don't create an additional rule with `objectSelector` selecting `gardener.cloud/deletion-protected: "true"`

charts/seed-bootstrap/templates/gardener-seed-admission-controller/validatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,44 @@
...
...bVYwWlhNd2dnRWlNQTBHQ1NxR1NJYjNEUUVCCkFRVUFBNElCRHdBd2dnRUtBb0lCQVFDeWkwUUdPY3YyYlRmM044T0xOOTdSd3NnSDZRQXI4d1NwQU9ydHRCSmcKRm5mblUyVDFSSGd4bTdxZDE5MFdMOERDaHYwZFpmNzZkNmVTUTRacmpqeUFyVHp1ZmI0RHRQd2crVldxN1h2RgpCTnluKzJoZjRTeVNrd2Q2azdYTGhVVFJ4MDQ4SWJCeUM0ditGRXZtb0xBd3JjMGQwRzE0ZWM2c25EKzdqTzdlCmt5a1EvTmdBT0w3UDZrRHM5ejYrYk9mZ0YwbkdOK2JtZVdRcUplalIwdCtPeVFEQ3g1L0ZNdFVmRVZSNVFYODAKYWVlZmdwM0pGWmI2ZkF3OUtoTHRkUlYzRlAwdHo2aFMrZTRTZzBtd0FBT3FpalpzVjg3a1A1R1l6anRjZkExMgpsRFlsL25iMUd0VnZ2a1FENDlWblY3bURubDZtRzNMQ01OQ05INldsWk52M0FnTUJBQUdqUWpCQU1BNEdBMVVkCkR3RUIvd1FFQXdJQkJqQVBCZ05WSFJNQkFmOEVCVEFEQVFIL01CMEdBMVVkRGdRV0JCU0ZBM0x2Sk0yMWQ4cXMKWlZWQ2U2UnJUVDl3aVRBTkJna3Foa2lHOXcwQkFRc0ZBQU9DQVFFQW5zL0VKM3lLc2p0SVNvdGVRNzE0cjJVbQpCTVB5VVlUVGRSSEQ4TFpNZDNSeWt2c2FjRjJsMnk4OE56NndKY0F1b1VqMWg4YUJEUDVvWFZ0Tm1GVDlqeWJTClRYclJ2V2krYWVZZGI1NTZuRUE1L2E5NGUrY2IrQ2szcXkvMXhnUW9TNDU3QVpRT0Rpc0RaTkJZV2tBRnMyTGMKdWNwY0F0WEp0SXRoVm03RmpvQUhZY3NyWTA0eUFpWUVKTEQwMlRqVURYZzRpR09HTWtWSGRtaGF3QkRCRjNBagplc2ZjcUZ3amk2SnlBS0ZSQUNQb3d5a1FPTkZ3VVNvbTg5dVlFU1NDSkZ2TkNrOU1KbWpKMlB6RFV0NkN5cFI0CmVwRmRkMWZYTHd1d243ZnZQTW1KcUQzSHRMYWxYMUFabVBrK0JJOGV6ZkFpVmNWcW5USlFNWGxZUHBZZTlBPT0KLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo=
```
mvladev, line 44: Why is this hard-coded?

charts/seed-bootstrap/templates/gardener-seed-admission-controller/validatingwebhookconfiguration.yaml
rfranzke: Object selectors only work with Kubernetes 1.15+. Our minimum seed Kubernetes version is 1.11, though. Hence, we can not use it yet.

charts/seed-bootstrap/templates/gardener-seed-admission-controller/validatingwebhookconfiguration.yaml
rfranzke: This CA certificate is only used to validate the server certificate served by the admission webhook. Generating a dedicated certificate per seed doesn't increase security. It's only used to satisfy the webhook contract (which is that HTTPS on port 443 must be served). Consequently, the server certificate served by the admission webhook is always the same as well.

charts/seed-bootstrap/templates/gardener-seed-admission-controller/validatingwebhookconfiguration.yaml
mvladev: Well for `<1.15` it'll send all objects as before, but it'll filter them out in `1.15+`.

rfranzke: OK, I can add it. Though, it doesn't allow us to remove our own label-filtering logic. I'll add a TODO there that this logic can be removed once the minimum seed version is bumped to >= 1.15.

charts/seed-bootstrap/charts/fluentd-es/templates/fluent-bit-configmap.yaml
```diff
@@ -357,7 +364,7 @@ data:
         Time_Format %Y-%m-%d %H:%M:%S.%L
 
     [PARSER]
-        Name        geacParser
+        Name        gsacParser
```
timebertt, line 367: It seems like the `etcd-backup-restore` FILTER in Line 106 also has to be changed to `gsacParser`

charts/images.yaml
```diff
@@ -4,10 +4,14 @@
...
 images:
-# Gardenlet
+# Gardener components
 - name: gardenlet
   sourceRepository: github.com/gardener/gardener
   repository: eu.gcr.io/gardener-project/gardener/gardenlet
+- name: gardener-seed-admission-controller
+  sourceRepository: github.com/gardener/gardener
+  repository: eu.gcr.io/gardener-project/gardener/seed-admission-controller
+  tag: "v1.2.0-dev"
```
timebertt: Can you explain, why you are specifying a tag here and not reusing the same mechanism we use for the `gardenlet` deploying itself with the same version?

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// NewCommandStartGardenerSeedAdmissionController creates a *cobra.Command object with default parameters
+func NewCommandStartGardenerSeedAdmissionController(ctx context.Context) *cobra.Command {
+	opts := &Options{}
+
+	cmd := &cobra.Command{
+		Use:   "gardener-seed-admission-controller",
+		Short: "Launch the Gardener seed admission controller",
+		Long:  `The Gardener seed admission controller serves as validation webhook endpoint for resources in the seed clusters.`,
```
timebertt:
```suggestion
		Long:  `The Gardener seed admission controller serves a validation webhook endpoint for resources in the seed clusters.`,
```

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
```diff
@@ -0,0 +1,241 @@
...
+
+	opts.AddFlags(cmd.Flags())
+	return cmd
+}
+
+// run runs the Gardener seed admission controller. This should never exit.
+func run(ctx context.Context, bindAddress string, port int, certPath, keyPath, kubeconfigPath string) {
+	scheme := runtime.NewScheme()
+	corev1.AddToScheme(scheme)
+	admissionregistrationv1beta1.AddToScheme(scheme)
```
timebertt:
```suggestion
	utilruntime.Must(corev1.AddToScheme(scheme))
	utilruntime.Must(admissionregistrationv1beta1.AddToScheme(scheme))
```

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
```diff
@@ -0,0 +1,241 @@
...
+	if contentType := r.Header.Get("Content-Type"); contentType != "application/json" {
+		err := fmt.Errorf("contentType=%s, expect %s", contentType, "application/json")
+		logger.Errorf(err.Error())
+		respond(w, &admissionv1beta1.AdmissionResponse{Result: &metav1.Status{Message: err.Error()}})
+		return receivedReview, err
+	}
+
+	// Deserialize HTTP request body into admissionv1beta1.AdmissionReview object.
+	if _, _, err := g.codecs.Decode(body, nil, &receivedReview); err != nil {
+		logger.Errorf(err.Error())
```
timebertt, line 205: Can you also add a error message here?
```suggestion
		err = fmt.Errorf("invalid request body (error decoding body): %+v", err)
		logger.Errorf(err.Error())
```

charts/seed-bootstrap/templates/gardener-seed-admission-controller/validatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - DELETE
+      resources:
+      - backupbuckets
+      - backupentries
+      - controlplanes
+      - extensions
+      - infrastructures
+      - networks
+      - operatingsystemconfigs
+      - workers
```
timebertt: I assume, you left out the `Cluster` resource on purpose, because no one should act upon it, right?
But for example the infra/cp/network/worker controllers fetch the `cluster` resource first before reconciling. So if somebody would delete the `Cluster` CRD / a `Cluster` CR, all controllers would fail to reconcile. So I guess, this is also something worth protecting.

docs/concepts/seed-admission-controller.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Validating Webhooks
+
+#### Unconfirmed Deletion Prevention
+As part of Gardener's [extensibility concepts](../extensions/overview.md) a lot of `CustomResourceDefinition`s are deployed to the seed clusters that serve as extension points for provider-specific controllers.
+For example, the [`Infrastructure` CRD](../extensions/infrastructure.md) triggers the provider extension to prepare the IaaS infrastructure of the underlying cloud provider for a to-be-created shoot cluster.
+Consequently, these extension CRDs have a lot of power and control large portions of the end-user's shoot cluster.
+Accidental or undesired deletions of those resource can cause tremendous and hard-to-recover-from outages and should be prevented.
+
+The Gardener seed admission controller is registered a `ValidatingWebhookConfiguration` for `CustomResourceDefintions` and most (custom) resources in the `extensions.gardener.cloud/v1alpha1` API group.
```
timebertt:
```suggestion
The Gardener seed admission controller is registered as a `ValidatingWebhookConfiguration` for `CustomResourceDefintions` and most (custom) resources in the `extensions.gardener.cloud/v1alpha1` API group.
```

charts/seed-bootstrap/templates/gardener-seed-admission-controller/validatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,44 @@
...
+  labels:
+    app: gardener
+    role: seed-admission-controller
+webhooks:
+  - name: seed.admission.core.gardener.cloud
+    rules:
+    - apiGroups:
+      - apiextensions.k8s.io
+      apiVersions:
+      - v1alpha1
```
timebertt: I don't see you checking for `apiextensions.k8s.io/v1alpha1` in `pkg/seedadmission/extension_crds_test.go` so we should remove `v1alpha1` here.
```suggestion
```

pkg/seedadmission/extension_crds.go
```diff
@@ -0,0 +1,155 @@
...
+		err = c.Get(ctx, kutil.Key(request.Namespace, request.Name), o)
+		obj = o
+	}
+
+	if err != nil {
+		return client.IgnoreNotFound(err)
+	}
+
+	entryLogger := gardenlogger.
+		NewFieldLogger(logger, "resource", fmt.Sprintf("%s/%s/%s", request.Kind.Group, request.Kind.Version, request.Kind.Kind)).
```
timebertt, line 101:
```suggestion
		NewFieldLogger(logger, "resource", fmt.Sprintf("%s/%s/%s", request.Resource.Group, request.Resource.Version, request.Resource.Resource)).
```

pkg/operation/common/utils_test.go
```diff
@@ -479,4 +486,146 @@ var _ = Describe("common", func() {
...
+			expectedObj := obj.DeepCopy()
+			expectedObj.Annotations = map[string]string{ConfirmationDeletion: "true"}
+
+			k8sSeedRuntimeClient.EXPECT().Get(ctx, gomock.AssignableToTypeOf(client.ObjectKey{}), obj)
+			k8sSeedRuntimeClient.EXPECT().Update(ctx, expectedObj)
+
+			Expect(ConfirmDeletion(ctx, k8sSeedRuntimeClient, obj)).To(Succeed())
+		})
+
+		It("should add the deletion confirmation annotation for an object with labels", func() {
```
timebertt:
```suggestion
		It("should add the deletion confirmation annotation for an object with annotations", func() {
```

pkg/operation/common/utils_test.go
```diff
@@ -479,4 +486,146 @@ var _ = Describe("common", func() {
...
+		BeforeEach(func() {
+			ctrl = gomock.NewController(GinkgoT())
+			k8sSeedRuntimeClient = mockclient.NewMockClient(ctrl)
+		})
+
+		AfterEach(func() {
+			ctrl.Finish()
+		})
+
+		It("should add the deletion confirmation annotation for an object without labels", func() {
```
timebertt:
```suggestion
		It("should add the deletion confirmation annotation for an object without annotations", func() {
```

pkg/operation/common/utils_test.go
```diff
@@ -479,4 +486,146 @@ var _ = Describe("common", func() {
...
+
+				Expect(CheckIfDeletionIsConfirmed(obj)).To(Succeed())
+			})
+		})
+	})
+
+	Describe("#ConfirmDeletion", func() {
+		var (
+			ctrl                 *gomock.Controller
+			k8sSeedRuntimeClient *mockclient.MockClient
```
timebertt:
```suggestion
			c *mockclient.MockClient
```

pkg/operation/common/utils_test.go
```diff
@@ -479,4 +486,146 @@ var _ = Describe("common", func() {
...
+				ctx = context.TODO()
+				obj = &corev1.Namespace{}
+			)
+
+			expectedObj := obj.DeepCopy()
+			expectedObj.Annotations = map[string]string{ConfirmationDeletion: "true"}
+
+			k8sSeedRuntimeClient.EXPECT().Get(ctx, gomock.AssignableToTypeOf(client.ObjectKey{}), obj)
+			k8sSeedRuntimeClient.EXPECT().Update(ctx, expectedObj).Return(apierrors.NewConflict(corev1.Resource("namespaces"), "", errors.New("conflict")))
+			k8sSeedRuntimeClient.EXPECT().Get(ctx, gomock.AssignableToTypeOf(client.ObjectKey{}), obj)
```
timebertt:
```suggestion
			k8sSeedRuntimeClient.EXPECT().Get(ctx, gomock.AssignableToTypeOf(client.ObjectKey{}), obj)
			k8sSeedRuntimeClient.EXPECT().Update(ctx, expectedObj)
```

pkg/operation/common/utils_test.go
```diff
@@ -479,4 +486,146 @@ var _ = Describe("common", func() {
...
+				ctx = context.TODO()
+				obj = &corev1.Namespace{}
+			)
+
+			expectedObj := obj.DeepCopy()
+			expectedObj.Annotations = map[string]string{ConfirmationDeletion: "true"}
+
+			k8sSeedRuntimeClient.EXPECT().Get(ctx, gomock.AssignableToTypeOf(client.ObjectKey{}), obj).Return(apierrors.NewNotFound(corev1.Resource("namespaces"), ""))
+
+			Expect(ConfirmDeletion(ctx, k8sSeedRuntimeClient, obj)).To(Succeed())
```
timebertt:
```suggestion
			expectedObj := obj.DeepCopy()

			k8sSeedRuntimeClient.EXPECT().Get(ctx, gomock.AssignableToTypeOf(client.ObjectKey{}), obj).Return(apierrors.NewNotFound(corev1.Resource("namespaces"), ""))

			Expect(ConfirmDeletion(ctx, k8sSeedRuntimeClient, obj)).To(Succeed())
			Expect(obj).To(Equal(expectedObj))
```

charts/seed-bootstrap/templates/gardener-seed-admission-controller/validatingwebhookconfiguration.yaml
timebertt: Also what about the CRs in `dns.gardener.cloud`? You marked the CRDs as deletion protected but do not check the CRs here. So somebody could accidentally delete all the dnsproviders/entries and we would loos all the Shoots' dns entries would also be pretty bad. Though, this would not be so hard-to-recover-from as with the other resources, because it can self-heal on the next reconciliation. But I think, this is also something worth protecting.

timebertt: I think we should also protect the `ManagedResource` CRD/CRs, because we use it to deploy all the extension controllers and also some extensions are depending on it to deploy stuff into seed/shoot which might be critical. So if we loose the mr CRD or all mr CRs, we would also have a mess....

charts/seed-bootstrap/templates/gardener-seed-admission-controller/deployment.yaml
```diff
@@ -0,0 +1,65 @@
...
+        resources:
+{{ toYaml .Values.gardenerSeedAdmissionController.resources | indent 10 }}
+        {{- end }}
+        resources:
+          requests:
+            cpu: 20m
+            memory: 50Mi
+          limits:
+            cpu: 100m
+            memory: 100Mi
```
timebertt: Duplicated key `resources` if `. Values.gardenerSeedAdmissionController.resources` is specified.
```suggestion
```

charts/seed-bootstrap/templates/gardener-seed-admission-controller/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,16 @@
...
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: gardener-seed-admission-controller
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: gardener
+    role: seed-admission-controller
+spec:
+  maxUnavailable: {{ sub (int .Values.gardenerSeedAdmissionController.replicas) 1 }}
```
timebertt: Why are you not simply using `minAvailable`?
```suggestion
  minAvailable: 1
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,571 @@
...
+						request.OldObject = runtime.RawExtension{Raw: objJSON}
+						testDeletionConfirmed(ctx, nil, logger, request, resource)
+					}
+				})
+			})
+
+			Context("other resources", func() {
+				It("should prevent the deletion because deletion is not confirmed", func() {
+					for _, resource := range otherResources {
+						objJSON := getObjectJSONWithLabelsAnnotations(obj, resource, deletionProtectedLabels, nil)
```
timebertt:
```suggestion
						objJSON := getObjectJSONWithLabelsAnnotations(obj, resource, nil, nil)
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,571 @@
...
+						request.OldObject = runtime.RawExtension{Raw: objJSON}
+						testDeletionUnconfirmed(ctx, nil, logger, request, resource)
+					}
+				})
+
+				It("should admit the deletion because deletion is confirmed", func() {
+					for _, resource := range otherResources {
+						objJSON := getObjectJSONWithLabelsAnnotations(obj, resource, deletionProtectedLabels, deletionConfirmedAnnotations)
```
timebertt:
```suggestion
						objJSON := getObjectJSONWithLabelsAnnotations(obj, resource, nil, deletionConfirmedAnnotations)
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,571 @@
...
+						request.Object = runtime.RawExtension{Raw: objJSON}
+						testDeletionConfirmed(ctx, nil, logger, request, resource)
+					}
+				})
+			})
+
+			Context("other resources", func() {
+				It("should prevent the deletion because deletion is not confirmed", func() {
+					for _, resource := range otherResources {
+						objJSON := getObjectJSONWithLabelsAnnotations(obj, resource, deletionProtectedLabels, nil)
```
timebertt:
```suggestion
						objJSON := getObjectJSONWithLabelsAnnotations(obj, resource, nil, nil)
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,571 @@
...
+						request.Object = runtime.RawExtension{Raw: objJSON}
+						testDeletionUnconfirmed(ctx, nil, logger, request, resource)
+					}
+				})
+
+				It("should admit the deletion because deletion is confirmed", func() {
+					for _, resource := range otherResources {
+						objJSON := getObjectJSONWithLabelsAnnotations(obj, resource, deletionProtectedLabels, deletionConfirmedAnnotations)
```
timebertt:
```suggestion
						objJSON := getObjectJSONWithLabelsAnnotations(obj, resource, nil, deletionConfirmedAnnotations)
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,571 @@
...
+
+func testDeletionConfirmed(
+	ctx context.Context,
+	c *mockclient.MockClient,
+	logger *logrus.Logger,
+	request *admissionv1beta1.AdmissionRequest,
+	resource metav1.GroupVersionResource,
+) {
+	request.Resource = resource
+	Expect(ValidateExtensionDeletion(ctx, c, logger, request)).To(BeNil(), resourceToId(resource))
```
timebertt:
```suggestion
	Expect(ValidateExtensionDeletion(ctx, c, logger, request)).To(Succeed(), resourceToId(resource))
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,571 @@
...
+
+		Context("object must be looked up (DELETE COLLECTION)", func() {
+			var obj *unstructured.UnstructuredList
+
+			BeforeEach(func() {
+				obj = &unstructured.UnstructuredList{}
+				obj.SetKind("List")
+			})
+
+			It("should return an error because the GET call failed", func() {
```
timebertt:
```suggestion
			It("should return an error because the LIST call failed", func() {
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,571 @@
...
+					c.EXPECT().List(ctx, obj, client.InNamespace(request.Namespace)).Return(apierrors.NewNotFound(core.Resource(resource.Resource), "name"))
+
+					Expect(ValidateExtensionDeletion(ctx, c, logger, request)).To(BeNil(), resourceToId(resource))
+				}
+			})
+
+			Context("custom resource definitions", func() {
+				BeforeEach(func() {
+					request.Kind = metav1.GroupVersionKind{Kind: "CustomResourceDefinition"}
+					request.Kind.Kind = "CustomResourceDefinition"
```
timebertt:
```suggestion
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,571 @@
...
+					request.Name = "foo"
+					request.Namespace = "bar"
+				})
+
+				It("should prevent the deletion because deletion is not confirmed", func() {
+					for _, resource := range otherResources {
+						prepareRequestAndObjectWithResource(request, obj, resource)
+
+						c.EXPECT().Get(ctx, kutil.Key(request.Namespace, request.Name), obj).DoAndReturn(func(_ context.Context, _ client.ObjectKey, o runtime.Object) error {
+							prepareObjectWithLabelsAnnotations(o, resource, deletionProtectedLabels, nil)
```
timebertt:
```suggestion
							prepareObjectWithLabelsAnnotations(o, resource, nil, nil)
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,571 @@
...
+						testDeletionUnconfirmed(ctx, c, logger, request, resource)
+					}
+				})
+
+				It("should admit the deletion because deletion is confirmed", func() {
+					for _, resource := range otherResources {
+						prepareRequestAndObjectWithResource(request, obj, resource)
+
+						c.EXPECT().Get(ctx, kutil.Key(request.Namespace, request.Name), obj).DoAndReturn(func(_ context.Context, _ client.ObjectKey, o runtime.Object) error {
+							prepareObjectWithLabelsAnnotations(o, resource, deletionProtectedLabels, deletionConfirmedAnnotations)
```
timebertt:
```suggestion
							prepareObjectWithLabelsAnnotations(o, resource, nil, deletionConfirmedAnnotations)
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,571 @@
...
+					request.Namespace = "bar"
+				})
+
+				It("should prevent the deletion because deletion is not confirmed", func() {
+					for _, resource := range otherResources {
+						prepareRequestAndObjectWithResource(request, obj, resource)
+						obj.SetKind(obj.GetKind() + "List")
+
+						c.EXPECT().List(ctx, obj, client.InNamespace(request.Namespace)).DoAndReturn(func(_ context.Context, o runtime.Object, _ ...client.ListOption) error {
+							prepareObjectWithLabelsAnnotations(o, resource, deletionProtectedLabels, nil)
```
timebertt:
```suggestion
							prepareObjectWithLabelsAnnotations(o, resource, nil, nil)
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,571 @@
...
+					}
+				})
+
+				It("should admit the deletion because deletion is confirmed", func() {
+					for _, resource := range otherResources {
+						prepareRequestAndObjectWithResource(request, obj, resource)
+						obj.SetKind(obj.GetKind() + "List")
+
+						c.EXPECT().List(ctx, obj, client.InNamespace(request.Namespace)).DoAndReturn(func(_ context.Context, o runtime.Object, _ ...client.ListOption) error {
+							prepareObjectWithLabelsAnnotations(o, resource, deletionProtectedLabels, deletionConfirmedAnnotations)
```
timebertt:
```suggestion
							prepareObjectWithLabelsAnnotations(o, resource, nil, deletionConfirmedAnnotations)
```

timebertt (changes requested):
Thanks for the PR, Rafael!
Looks pretty good, though I have some comments here and there (mainly regarding the protected CRs/CRDs and the unit tests).
Also wanna give it a test in my local setup tomorrow, will come back with the results after that :)

charts/images.yaml
rfranzke: I'm confused, we discussed this on the phone two days ago. The reason is that for non-released versions the Gardenlet thinks its own version is `vX.Y.Z-dev`, and for this no image tag exists. Hence, if you'd check out the `master` branch (or any non-released commit) then it won't work. Hence, we decided together that specifying the version here is the best approach for now. After the `v1.2` release the version here will always be up-to-date with the latest released version (see `prepare_release` CI script).

charts/seed-bootstrap/templates/gardener-seed-admission-controller/validatingwebhookconfiguration.yaml
rfranzke: > I assume, you left out the Cluster resource on purpose, because no one should act upon it, right?

Right.

> So I guess, this is also something worth protecting.

The purpose of the protection is prevent catastrophic situations. Deleting the `Cluster` resource will only make the reconciliation fail as it only serves as  kind of configmap. Hence, no need for protecting it.

> Also what about the CRs in dns.gardener.cloud? You marked the CRDs as deletion protected but do not check the CRs here.

On purpose as well for now. If we would protect it we would have to change the DNS controller manager as it deletes the DNS entries. We don't want this at this point in time. Hence, I only protected the CRD for now. I'm open to later also protect the custom resources as well in case the DNS controller manager adopts the deletion confirmation logic. But I don't want to depend on it now.

> Though, this would not be so hard-to-recover-from as with the other resources, because it can self-heal on the next reconciliation.

Right.

> I think we should also protect the ManagedResource CRD/CRs, because we use it to deploy all the extension controllers and also some extensions are depending on it to deploy stuff into seed/shoot which might be critical.

Agreed to protect the CRD itself, will do. However, I won't protect the CRs as it again would require adoption in all the extension controllers (= not backwards compatible). Same rationale like for the DNS objects.

pkg/seedadmission/extension_crds.go
rfranzke: Why are you suggesting this?

pkg/seedadmission/extension_crds_test.go
rfranzke: When do we use `BeNil()` and when do we use `Succeed()`?

charts/seed-bootstrap/charts/fluentd-es/templates/fluent-bit-configmap.yaml
rfranzke: Nice catch... wtf.. why is it using it in the first place... anyways, changed.

charts/seed-bootstrap/templates/gardener-seed-admission-controller/rbac.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - extensions.gardener.cloud
+  resources:
+  - backupbuckets
+  - backupentries
+  - controlplanes
+  - extensions
+  - infrastructures
+  - networks
+  - operatingsystemconfigs
+  - workers
```
danielfoehrKn, line 27: What about ContainerRuntime resources? Is that something to be added at a later point?

pkg/operation/common/utils_test.go
rfranzke: After the first `Update()` call the object is already annotated, hence, the retry function detects no difference and exits early (before it sends another `Update()` call). I'm sure how to reset the object (remove the annotation) in the second `Get()` call again.

docs/concepts/seed-admission-controller.md
danielfoehrKn:
```suggestion
The Gardener seed admission controller is registered a `ValidatingWebhookConfiguration` for `CustomResourceDefinitions` and most (custom) resources in the `extensions.gardener.cloud/v1alpha1` API group.
```

charts/seed-bootstrap/templates/gardener-seed-admission-controller/deployment.yaml
rfranzke: nice catch, thanks

charts/seed-bootstrap/templates/gardener-seed-admission-controller/poddisruptionbudget.yaml
rfranzke: Hm..., too easy to read? 😛Fixed, thanks.

danielfoehrKn (commented):
Looking good. Will run a more thorough test soon.

charts/seed-bootstrap/templates/gardener-seed-admission-controller/rbac.yaml
rfranzke: I didn't add it because it's not yet merged. But yes, it makes sense to already add it now in order to prevent follow-up effort. It doesn't harm here if the #2035 is not yet merged. Done, thanks.
PS: I also asked to add the protection label to the new CRD here: https://github.com/gardener/gardener/pull/2035#discussion_r395293604

charts/images.yaml
timebertt: Sorry, I must have missed that discussion on the phone, there was a bad connection during the meeting 😞
Anyways, got it. Thanks for explaining.

rfranzke: 😄 Alright, np. :)

pkg/seedadmission/extension_crds.go
```diff
@@ -0,0 +1,157 @@
...
+func ValidateExtensionDeletion(ctx context.Context, c client.Client, logger *logrus.Logger, request *admissionv1beta1.AdmissionRequest) error {
+	// Ignore all resources other than our expected ones
+	switch request.Resource {
+	case
+		metav1.GroupVersionResource{Group: apiextensionsv1beta1.SchemeGroupVersion.Group, Version: apiextensionsv1beta1.SchemeGroupVersion.Version, Resource: "customresourcedefinitions"},
+		metav1.GroupVersionResource{Group: apiextensionsv1.SchemeGroupVersion.Group, Version: apiextensionsv1.SchemeGroupVersion.Version, Resource: "customresourcedefinitions"},
+
+		metav1.GroupVersionResource{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "backupbuckets"},
+		metav1.GroupVersionResource{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "backupentries"},
+		metav1.GroupVersionResource{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "controlplanes"},
```
timebertt, line 51: I see, you have added `containerruntimes` to the `ValidatingWebhookConfiguration`. I figure, you might also wanna add it here:
```suggestion
		metav1.GroupVersionResource{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "controlplanes"},
		metav1.GroupVersionResource{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "containerruntimes"},
```

pkg/seedadmission/extension_crds_test.go
```diff
@@ -0,0 +1,570 @@
...
+		})
+
+		var (
+			resources = []metav1.GroupVersionResource{
+				{Group: apiextensionsv1beta1.SchemeGroupVersion.Group, Version: apiextensionsv1beta1.SchemeGroupVersion.Version, Resource: "customresourcedefinitions"},
+				{Group: apiextensionsv1.SchemeGroupVersion.Group, Version: apiextensionsv1.SchemeGroupVersion.Version, Resource: "customresourcedefinitions"},
+
+				{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "backupbuckets"},
+				{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "backupentries"},
+				{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "controlplanes"},
```
timebertt, line 75: I see, you have added `containerruntimes` to the `ValidatingWebhookConfiguration`. I figure, you might also wanna add it here:
```suggestion
				{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "controlplanes"},
				{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "containerruntimes"},
```

pkg/seedadmission/extension_crds.go
timebertt: Hm, I thought you used `request.Kind` here unintentionally, as you are mostly checking/working with `request.Resource` and also named the log key `"resource"`. I just thought, it would make more sense to also put the values of `request.Resource` into the log instead of `request.Kind`. But not strong opinion on this one... :)

pkg/operation/common/utils_test.go
```diff
@@ -479,4 +486,146 @@ var _ = Describe("common", func() {
...
+				expectedObj = obj.DeepCopy()
+			)
+
+			c.EXPECT().Get(ctx, gomock.AssignableToTypeOf(client.ObjectKey{}), obj).Return(apierrors.NewNotFound(corev1.Resource("namespaces"), ""))
+
+			Expect(ConfirmDeletion(ctx, c, obj)).To(Succeed())
+			Expect(obj).To(Equal(expectedObj))
+		})
+
+		It("should retry and conflict and add the deletion confirmation annotation", func() {
```
timebertt:
```suggestion
		It("should retry on conflict and add the deletion confirmation annotation", func() {
```

pkg/operation/common/utils_test.go
timebertt: Got it. What about doing it like this? Would that make sense?

```go
c.EXPECT().Get(ctx, gomock.AssignableToTypeOf(client.ObjectKey{}), obj)
c.EXPECT().Update(ctx, expectedObj).Return(apierrors.NewConflict(corev1.Resource("namespaces"), "", errors.New("conflict")))
c.EXPECT().Get(ctx, gomock.AssignableToTypeOf(client.ObjectKey{}), expectedObj).DoAndReturn(func(_ context.Context, _ client.ObjectKey, o runtime.Object) error {
	baseObj.DeepCopyInto(o.(*corev1.Namespace))
	return nil
})
c.EXPECT().Update(ctx, expectedObj)

Expect(ConfirmDeletion(ctx, c, obj)).To(Succeed())
```

pkg/seedadmission/extension_crds_test.go
timebertt: No strong opinion here, but it seems like it is good practice to use `HaveOccurred()` and `Succeed()` for asserting (non-)presence of `error`s instead of `BeNil()` for the sake of making the test a little bit easier to understand. But as said, really no strong opinion here :)

charts/seed-bootstrap/templates/gardener-seed-admission-controller/validatingwebhookconfiguration.yaml
timebertt: > The purpose of the protection is prevent catastrophic situations. Deleting the `Cluster` resource will only make the reconciliation fail as it only serves as kind of configmap. Hence, no need for protecting it.

Ok, fair enough.

> If we would protect it we would have to change the DNS controller manager as it deletes the DNS entries. We don't want this at this point in time.

Oh yeah, got it. I didn't consider the implication on the dns controller manager's own deletion logic.

> However, I won't protect the CRs as it again would require adoption in all the extension controllers (= not backwards compatible).

Same here, also didn't think about the deletion logic in the controllers. But agreed, protecting the dns/managedresource CRDs is already a good measure to prevent catastrophic situations 👍

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
```diff
@@ -0,0 +1,242 @@
...
+func NewCommandStartGardenerSeedAdmissionController(ctx context.Context) *cobra.Command {
+	opts := &Options{}
+
+	cmd := &cobra.Command{
+		Use:   "gardener-seed-admission-controller",
+		Short: "Launch the Gardener seed admission controller",
+		Long:  `The Gardener seed admission controller serves a validation webhook endpoint for resources in the seed clusters.`,
+		Run: func(cmd *cobra.Command, args []string) {
+			utilruntime.Must(opts.validate(args))
+			opts.run(ctx)
```
timebertt, line 112: WDYT about printing all the flags before starting the controller, like the kubernetes binaries do? (Could also be added to our other binaries). Just for convenience and debugging purposes...

```suggestion
			
			logger.Infof("Starting Gardener seed admission controller...")
			cmd.Flags().VisitAll(func(flag *pflag.Flag) {
				logger.Infof("FLAG: --%s=%s", flag.Name, flag.Value)
			})
			
			opts.run(ctx)
```

This will print out something like:
```
time="2020-03-20T11:10:22+01:00" level=info msg="Starting Gardener seed admission controller..."
time="2020-03-20T11:10:22+01:00" level=info msg="FLAG: --bind-address=0.0.0.0"
time="2020-03-20T11:10:22+01:00" level=info msg="FLAG: --help=false"
time="2020-03-20T11:10:22+01:00" level=info msg="FLAG: --kubeconfig=/Users/d067603/.kube/configs/local/kind-1.yaml"
time="2020-03-20T11:10:22+01:00" level=info msg="FLAG: --port=9443"
time="2020-03-20T11:10:22+01:00" level=info msg="FLAG: --tls-cert-path=example/seed-admission-controller/tls.crt"
time="2020-03-20T11:10:22+01:00" level=info msg="FLAG: --tls-private-key-path=example/seed-admission-controller/tls.key"
```

pkg/seedadmission/extension_crds.go
```diff
@@ -0,0 +1,157 @@
...
+		o := &unstructured.Unstructured{}
+		err = json.Unmarshal(request.OldObject.Raw, o)
+		obj = o
+
+	case request.Object.Raw != nil:
+		o := &unstructured.Unstructured{}
+		err = json.Unmarshal(request.Object.Raw, o)
+		obj = o
+
+	case request.Name == "":
```
timebertt, line 80: In which case is this `case` used? It seems to me like the apiserver always calls the webhook for each object when deleting all or label selected resources.

example/seed-admission-controller/10-validatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,38 @@
...
+        - backupentries
+        - containerruntimes
+        - controlplanes
+        - extensions
+        - infrastructures
+        - networks
+        - operatingsystemconfigs
+        - workers
+    failurePolicy: Fail
+    objectSelector: {}
```
timebertt:
```suggestion
```

pkg/seedadmission/extension_crds.go
rfranzke: This is used for DELETE COLLECTION calls. The namespace controller sometimes calls this when it cleans all resources inside a namespace.

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
```diff
@@ -0,0 +1,242 @@
...
+		logger.Errorf(err.Error())
+		respond(w, &admissionv1beta1.AdmissionResponse{Result: &metav1.Status{Message: err.Error()}})
+		return receivedReview, err
+	}
+
+	// Deserialize HTTP request body into admissionv1beta1.AdmissionReview object.
+	if _, _, err := g.codecs.Decode(body, nil, &receivedReview); err != nil {
+		err = fmt.Errorf("invalid request body (error decoding body): %+v", err)
+		logger.Errorf(err.Error())
+		respond(w, &admissionv1beta1.AdmissionResponse{Result: &metav1.Status{Message: err.Error()}})
```
timebertt: Hmm, this actually returns `HTTP 200` because `response.result.code` is not set, which seems kind of odd to me.
I would propose to use `sigs.k8s.io/controller-runtime/pkg/webhook/admission.Response` as a parameter for `respond()` and construct responses with the provided helper functions like `admission.Errored()` and `admission.Allowed()` which set the status code and error message correctly for you.

```suggestion
		respond(w, admission.Errored(http.StatusInternalServerError, err))
```

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
```diff
@@ -0,0 +1,242 @@
...
+	// If the request field is empty then do not admit (invalid body).
+	if receivedReview.Request == nil {
+		err := fmt.Errorf("invalid request body (missing admission request)")
+		logger.Errorf(err.Error())
+		respond(w, &admissionv1beta1.AdmissionResponse{Result: &metav1.Status{Message: err.Error()}})
+		return receivedReview, err
+	}
+	// If the request does not indicate the correct operation (DELETE) we allow the review without further doing.
+	if receivedReview.Request.Operation != admissionv1beta1.Delete {
+		respond(w, &admissionv1beta1.AdmissionResponse{})
```
timebertt: This will actually not allow the request (although with the provided webhookconfigs no request with `Operation!=DELETE` will be sent to the webhook).
```suggestion
		respond(w, admission.Allowed("operation is no DELETE operation"))
```

timebertt (commented):
Finished testing now, works completely fine for me.
I have only one more suggestion on how to handle the admission response.

pkg/seedadmission/extension_crds.go
timebertt: Alright, got it.

charts/seed-bootstrap/templates/gardener-seed-admission-controller/validatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - infrastructures
+      - networks
+      - operatingsystemconfigs
+      - workers
+    failurePolicy: Fail
+    {{- if semverCompare ">= 1.15-0" .Capabilities.KubeVersion.GitVersion -}}
+    objectSelector:
+      matchLabels:
+        gardener.cloud/deletion-protected: "true"
+    {{- end -}}
```
timebertt: This will render something like this:
```yaml
    failurePolicy: FailobjectSelector:
      matchLabels:
        gardener.cloud/deletion-protected: "true"namespaceSelector: {}
    clientConfig:
```


```suggestion
    {{- if semverCompare ">= 1.15-0" .Capabilities.KubeVersion.GitVersion }}
    objectSelector:
      matchLabels:
        gardener.cloud/deletion-protected: "true"
    {{- end }}
```

timebertt (changes requested):
Sorry, found one more bug.

charts/images.yaml
rfranzke: I changed my mind. I'll change the code in a way that the `charts/images.yaml` does not contain the tag by default. In this case, the Gardenlet will deploy GSAC with its current version (if it is not a dev version). If the current version is a dev version then it will deploy it with the last stable version.

rfranzke:
Wow, thanks for your thorough and intensive review @tim-ebert, kudos! I'm impressed! I hope I've addressed everything now - can you please check it again?

timebertt:
You're welcome. Will take another look on Monday :)

pkg/version/version.go
```diff
@@ -57,3 +58,40 @@ func Get() apimachineryversion.Info {
...
+	return version, nil
+}
+
+func constructVersion(withVPrefix bool, major, minor, patch int64) string {
+	prefix := ""
+	if withVPrefix {
+		prefix = "v"
+	}
+
+	nv, _ := semver.NewVersion(fmt.Sprintf("v%d.%d.%d", major, minor, patch))
```
timebertt:
```suggestion
	nv, _ := semver.NewVersion(fmt.Sprintf("%s%d.%d.%d", prefix, major, minor, patch))
```

pkg/version/version_test.go
```diff
@@ -0,0 +1,56 @@
...
+		Entry("unparseable version", "-dev", "", true),
+		Entry("patch dev version > 0", "v1.2.3-dev", "v1.2.2", false),
+		Entry("patch dev version > 0", "v1.0.3-dev", "v1.0.2", false),
+		Entry("patch dev version > 0", "v0.0.3-dev", "v0.0.2", false),
+		Entry("patch dev version = 0", "v1.2.0-dev", "v1.1.0", false),
+		Entry("minor dev version > 0", "v1.1.0-dev", "v1.0.0", false),
+		Entry("minor dev version > 0", "v0.1.0-dev", "v0.0.0", false),
+		Entry("minor dev version = 0", "v1.0.0-dev", "v0.0.0", false),
+		Entry("major dev version > 0", "v2.0.0-dev", "v1.0.0", false),
+		Entry("major dev version = 0", "v0.0.0-dev", "v0.0.0-dev", false),
```
timebertt: Can you add entries for versions without a `v` Prefix?

pkg/version/version.go
```diff
@@ -57,3 +58,40 @@ func Get() apimachineryversion.Info {
...
+
+	if v.Patch() > 0 {
+		return constructVersion(withVPrefix, v.Major(), v.Minor(), v.Patch()-1), nil
+	}
+	if v.Minor() > 0 {
+		return constructVersion(withVPrefix, v.Major(), v.Minor()-1, v.Patch()), nil
+	}
+	if v.Major() > 0 {
+		return constructVersion(withVPrefix, v.Major()-1, v.Minor(), v.Patch()), nil
+	}
```
timebertt: As discussed offline, I would propose to calculate the `vx.y.z-dev-<commit>` image tag in case the running gardenlet version is a `-dev` version instead of this logic.

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	mux.HandleFunc("/webhooks/validate-extension-crd-deletion", seedAdmissionController.validateExtensionCRDDeletion)
+
+	srv := &http.Server{
+		Addr:    fmt.Sprintf("%s:%d", bindAddress, port),
+		Handler: mux,
+	}
+
+	go func() {
+		logger.Infof("Starting Gardener seed admission controller...")
```
timebertt: Duplicated to the log in line 107.
```suggestion
```

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
```diff
@@ -0,0 +1,241 @@
...
+		err := fmt.Errorf("invalid request body (missing admission request)")
+		logger.Errorf(err.Error())
+		respond(w, admission.Errored(http.StatusBadRequest, err))
+		return receivedReview, err
+	}
+
+	// If the request does not indicate the correct operation (DELETE) we allow the review without further doing.
+	if receivedReview.Request.Operation != admissionv1beta1.Delete {
+		respond(w, admission.Allowed("operation is no DELETE operation"))
+		return receivedReview, nil
```
timebertt: You have to return a non-nil error in order to early exit and not run the validation itself.
```suggestion
		return receivedReview, fmt.Errorf("operation is no DELETE operation")
```

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
```diff
@@ -0,0 +1,241 @@
...
+	return receivedReview, nil
+}
+
+func (g *GardenerSeedAdmissionController) validateExtensionCRDDeletion(w http.ResponseWriter, r *http.Request) {
+	receivedReview, err := g.handleAdmissionReview(w, r)
+	if err != nil {
+		return
+	}
+
+	if err := seedadmission.ValidateExtensionDeletion(context.Background(), g.client, logger, receivedReview.Request); err != nil {
```
timebertt: The default timeout for validating webhooks is 30s, after which a request is cancelled by the kube-apiserver if the webhook server hasn't responded. Let's use a proper context with a timeout of 30s to avoid running zombie operations/checks.

```suggestion
	ctx, cancel := context.WithTimeout(r.Context(), 30 * time.Second)
	defer cancel()

	if err := seedadmission.ValidateExtensionDeletion(ctx, g.client, logger, receivedReview.Request); err != nil {
```

pkg/gardenlet/controller/shoot/seed_registration_control.go
```diff
@@ -599,10 +594,18 @@ func deployGardenlet(ctx context.Context, k8sGardenClient kubernetes.Interface,
...
+	gardenletImage, err := imageVector.FindImage("gardenlet")
+	if err != nil {
+		return err
+	}
 	var (
 		repository = gardenletImage.String()
-		tag        = version.Get().GitVersion
+		tag        string
 	)
+	tag, err = version.GetStableVersion(version.Get().GitVersion)
```
timebertt: As discussed, this will do a downgrade of deployed `gardenlet`s if the root `gardenlet` is running a `-dev` version and the last minor has patch releases, which were deployed before.

timebertt (changes requested):
I only have some smaller comments left now.

timebertt (approved):
Very nice, looks perfect now :) 
Gave it a spin in my local setup again, runs as expected!

/approve

rfranzke:
Let's wait with the merge of this PR until v1.2 has been cut.

PR #2067, opened 2020-03-19
Ensure worker pools are compatible with system components

rfranzke:
In order for the shoot's system components to run there must be at least one worker pool that either has no taints or only taints with `PreferNoSchedule` effect. Otherwise, end-users can exclude them from their nodes, causing the shoot reconciliation to fail:

```
Events:
  Type     Reason             Age                   From                Message
  ----     ------             ----                  ----                -------
  Warning  FailedScheduling   <unknown>             default-scheduler   0/2 nodes are available: 2 node(s) had taints that the pod didn't tolerate.
  Warning  FailedScheduling   <unknown>             default-scheduler   0/2 nodes are available: 2 node(s) had taints that the pod didn't tolerate.
  Normal   NotTriggerScaleUp  11s (x11 over 2m31s)  cluster-autoscaler  pod didn't trigger scale-up (it wouldn't fit if a new node is added): 1 node(s) had taints that the pod didn't tolerate
```

/cc @ggaurav10 

**Release note**:

```noteworthy user
It must now be ensured that the `Shoot` specification contains at least one worker pool with either no taints or only taints with `PreferNoSchedule` effect. This is to make sure that the system components (CoreDNS, vpn-shoot, etc.) can be scheduled correctly.
```

### DIFF
--- a/pkg/apis/core/validation/shoot.go
+++ b/pkg/apis/core/validation/shoot.go
@@ -1032,10 +1032,13 @@ func validateTaintEffect(effect *corev1.TaintEffect, allowEmpty bool, fldPath *f
 
 // ValidateWorkers validates worker objects.
 func ValidateWorkers(workers []core.Worker, fldPath *field.Path) field.ErrorList {
-	allErrs := field.ErrorList{}
+	var (
+		allErrs = field.ErrorList{}
 
-	workerNames := make(map[string]bool)
-	atLeastOneActivePool := false
+		workerNames                        = make(map[string]bool)
+		atLeastOneActivePool               = false
+		atLeastOnePoolWithCompatibleTaints = len(workers) == 0
+	)
 
 	for i, worker := range workers {
 		if worker.Minimum != 0 && worker.Maximum != 0 {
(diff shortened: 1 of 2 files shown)


pkg/apis/core/validation/shoot.go
```diff
@@ -1046,12 +1049,31 @@ func ValidateWorkers(workers []core.Worker, fldPath *field.Path) field.ErrorList
...
+			for _, taint := range worker.Taints {
+				if taint.Effect != corev1.TaintEffectPreferNoSchedule {
+					onlyPreferNoScheduleEffectTaints = false
+					break
+				}
+			}
+			if onlyPreferNoScheduleEffectTaints {
+				atLeastOnePoolWithCompatibleTaints = true
+			}
+		}
```
```suggestion
		} else if len(worker.Taints) == 1 && worker.Tains[0].Effect ==  corev1.TaintEffectPreferNoSchedule {
		        atLeastOnePoolWithCompatibleTaints = true
		}
```

ggaurav10: This block can go inside the following `if` check, right?
```
if atLeastOnePoolWithCompatibleTaints != true
```
Basically no need to process all worker groups if we have found atleast 1 favourable worker group.

timuthy (changes requested):
LGTM, just a suggestion.

ggaurav10 (commented):
One minor suggestion. Looks good otherwise.

pkg/apis/core/validation/shoot.go
rfranzke: Why do we need to handle this case specially? It's covered by the general case already.

rfranzke: done

timuthy: It replaces the case entirely to get rid of the for loop.

timuthy: But now it's resolved anyway since it's been moved to a switch case.

rfranzke: I don't get it, sorry. What if I have `NoExecute` taints in the first three pools, the fourth one has `PreferNoSchedule` and the next five pools have `NoExecute` as well?

pkg/apis/core/validation/shoot.go
```diff
@@ -1046,12 +1049,33 @@ func ValidateWorkers(workers []core.Worker, fldPath *field.Path) field.ErrorList
 			allErrs = append(allErrs, field.Duplicate(fldPath.Index(i).Child("name"), worker.Name))
 		}
 		workerNames[worker.Name] = true
+
+		switch {
+		case atLeastOnePoolWithCompatibleTaints:
+		case len(worker.Taints) == 0:
```
ggaurav10, line 1055:
```suggestion
		case atLeastOnePoolWithCompatibleTaints, len(worker.Taints) == 0:
```
This would be the correct way to mention multiple cases simultaneously.

ggaurav10 (changes requested):

pkg/apis/core/validation/shoot.go
rfranzke: This was not my intention, although it doesn't make a big difference:
1. If `atLeastOnePoolWithCompatibleTaints` is `true` then nothing is done (first case)
1. If `len(worker.Taints) == 0` is `true` then `atLeastOnePoolWithCompatibleTaints` is set to `true` (second case)
1. Otherwise, the loop is executed (third case).

ggaurav10: ah yes. Thanks for correcting.

ggaurav10 (approved):

pkg/apis/core/validation/shoot.go
timuthy: Agree. Sorry for the confusion.

rfranzke: As discussed, will add more test cases.

timuthy (approved):

PR #2070, opened 2020-03-19
Re-enable memory-based HPA for apiserver at 120 percent average utilisation.

amshuman-kr:
#1968 disabled memory-based HPA for shoot `kube-apiserver` to improve resource utilisation. But there are some usage patterns where memory utilisation can increase without a proportionate increase in CPU utilisation which can lead to `kube-apiserver` pods getting `OOMKilled`. Though [the](https://github.com/gardener/hvpa-controller/pull/57) [recent](https://github.com/gardener/hvpa-controller/pull/61) [improvements](https://github.com/gardener/hvpa-controller/pull/63) in [gardener/hvpa-controller](https://github.com/gardener/hvpa-controller) have made sure that the `kube-apiserver` recovers from such `OOMKills`, there is still some adverse affect on some usage patterns where memory utilisation is disproportionate to the CPU utilisation.

This PR re-enables memory-based HPA for kube-apiserver but at 120 percent average utilisation. This is to still avoid horizontally scaling for trivial loads but still scale horizontally if memory load is disproportionate to the CPU load.

The CPU utilisation configuration remains unchanged.

Fixes #

It would be preferable release this PR with #2070.

**Release note**:

```improvement operator
Re-enabled memory-based HPA for apiserver but at 120 percent average utilisation. This is to still avoid horizontally scaling for trivial loads but still scale horizontally if memory load is disproportionate to the CPU load. The CPU utilisation configuration remains unchanged.
```

### DIFF
--- a/charts/seed-controlplane/charts/kube-apiserver/values.yaml
+++ b/charts/seed-controlplane/charts/kube-apiserver/values.yaml
@@ -91,7 +91,7 @@ minReplicas: 1
 hpa:
   targetAverageUtilizationCpu: 80
   memoryMetricForHpaEnabled: false
-  targetAverageUtilizationMemory: 80
+  targetAverageUtilizationMemory: 120
 
 auditConfig:
   auditPolicy: ""
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -659,7 +659,7 @@ func (b *Botanist) DeployKubeAPIServerService(ctx context.Context) error {
 // DeployKubeAPIServer deploys kube-apiserver deployment.
 func (b *Botanist) DeployKubeAPIServer(ctx context.Context) error {
 	hvpaEnabled := gardenletfeatures.FeatureGate.Enabled(features.HVPA)
-	memoryMetricForHpaEnabled := false
+	memoryMetricForHpaEnabled := true
 
 	if b.ShootedSeed != nil {
 		// Override for shooted seeds


amshuman-kr:
cc @rfranzke @ggaurav10

rfranzke (approved):
/lgtm implementation-wise, @ggaurav10 can you review content-wise?

ggaurav10:
The memory metric was removed to reduce cost of seeds. If we are introducing it back, should we atleast consider reducing `maxReplicas` to 3 from 4?

Even though this PR's change will make HPA "less reluctant" to scale up, this change will still end up increasing the cost of seeds. By reducing `maxReplicas` to 3, we can mitigate some of the increase in cost. From past experience, when we use memory metric, we end up with `maxReplicas` of apiservers for almost all clusters. Reducing it to 3 would save us cost of 25% of apiservers.

Also, 4 replicas were required at the highest load that we saw in our perf tests. In case we are not comfortable in reducing this number, we need to relook at `limitsRequestsGapScaleParams` because current `limitsRequestsGapScaleParams.Memory.Value == 5G` at max load (25G memory) will not leave any room for graceful VPA updates with HPA threshold of 120%. This is because currently HVPA-controller scales vertically only if HPA is "limited".

I prefer to reduce `maxReplicas`, since we haven't seen the kind of load in real life that we put in our test runs - basically I am preferring `cost` over `stability under hypothetical load conditions`.
WDYT?

Also, we need to relook into stabilisation or frequent rolling updates of apiserver. Currently `scaleUpStabilization` and `scaleDownStabilization` are set too small, because of which we have seen too frequent vertical scaling recently. We need to increase this, may be in a separate PR.

rfranzke:
> From past experience, when we use memory metric, we end up with maxReplicas of apiservers for almost all clusters. 

This should definitely be avoided. If the memory metric leads to this situation again, we should rather not introduced it. Even if we lower the max replicas. Just my high-level pov.

amshuman-kr:
@ggaurav10 You raised multiple points. Please point out if I have not addressed all of them.

Regarding reducing `maxReplicas` to 3. I am, in principle, fine. But it will reduce the maximum size of the cluster (~220 nodes or 1k request rates). We can mitigate by increasing the `limitsRequestsGapScaleParams` configuration but without running a perf test, it is difficult to be sure how much increase would be enough. Perhaps, we can use the new min/max metrics I have added in #2080 to evaluate if there are indeed clusters out there which need such a scale. WDYT?

> I prefer to reduce maxReplicas, since we haven't seen the kind of load in real life that we put in our test runs - basically I am preferring cost over stability under hypothetical load conditions.
WDYT?

I agree with this prioritization as long as the disruption is not consistent to the end-users. The memory-scaling removal was consistently disruptive to some end-users.

> Also, we need to relook into stabilisation or frequent rolling updates of apiserver. Currently scaleUpStabilization and scaleDownStabilization are set too small, because of which we have seen too frequent vertical scaling recently. We need to increase this, may be in a separate PR.

This is a separate topic and we should increase the `minChange` configuration (and the `limitsRequestsGapScaleParams` to cope with such an increase). We should address it in a separate PR. I will work on it.

amshuman-kr:
> This should definitely be avoided. If the memory metric leads to this situation again, we should rather not introduced it. Even if we lower the max replicas. Just my high-level pov.

I agree. Unfortunately, the only way to be sure this does not happen is to use the `Scale` sub-resource (the support for which @ggaurav10  is already working on), to scale only vertically first and scale horizontally only at the highest end of the scale.

However, some of the new metrics in #2080 should help checking if the behaviour re-introducing memory-based scaling at 120% makes it as bad (in terms of average number of apiserver replicas per active shoot) as at 80%. From what I have seen, it should be better than at 80%. But that is based on `6h` data which is not at all sufficient. With #2080 we should have data over longer periods.

rfranzke:
OK, so you suggest to proceed for one release with the 120% metric, and then re-evaluate whether it can be kept or should be removed again for the next release?

amshuman-kr:
> OK, so you suggest to proceed for one release with the 120% metric, and then re-evaluate whether it can be kept or should be removed again for the next release?

Yes. Also, if #2080 is included in the same release, we can also evaluate if it is safe to reduce `maxReplicas` to 3 because we will have that data too over a longer period.

ggaurav10:
> it will reduce the maximum size of the cluster (~220 nodes or 1k request rates)

As I said, I would like to prefer cost saving over this "hypothetical load condition" which we have never seen, and are no-where close to seeing in near future. We have hardly (== never?) seen a request rate of 600 per second.
It's just my preference, keeping the real world scenarios in mind.

amshuman-kr:
> As I said, I would like to prefer cost saving over this "hypothetical load condition" which we have never seen, and are no-where close to seeing in near future. We have hardly (== never?) seen a request rate of 600 per second.

🙂 I am, in principle fine. As long as we re-adjust the advice we have put out there that we can support ~220 nodes. But we don't know what to re-adjust it to (at least, not yet).

Also, we don't have any alerts if anyone is reaching close to the limits of the supported scale. I think we should prioritize _Alerts when some percentage threshold of the maxAllowed is reached for requests of any container in the targetRef. Prio 2_ part of [gardener/hvpa-controller#30](https://github.com/gardener/hvpa-controller/issues/30).

ggaurav10:
> As long as we re-adjust the advice we have put out there that we can support ~220 nodes

Would this statement not hold for changing the HPA threshold as well, since HPA is slow to react now?

Also, we are a long way from the time when we did the 220 nodes perf tests. After that we have introduced hoards of changes like `limits` scaling (without this the `limits` was set to very very high, because of which there were less/**no** disruptions (oomkills, cpu throttling etc.) in case the usage spikes. With `limits` scaling we have set lower limits, and now we rely on overriding stabilisation parameters in case of such disruptions.

Also, we are considering the changes in stabilisation parameters, which will further change the overall behaviour.

Unrelated PS:
I am not sure where we have put such an advice officially out there. Please let me know. Thanks.

amshuman-kr:
> Would this statement not hold for changing the HPA threshold as well, since HPA is slow to react now?

Nope. Because scaling would happen before the limits are hit.

> I am not sure where we have put such an advice officially out there. Please let me know. Thanks.

We have mentioned it in the only [published tests](https://github.com/gardener/hvpa-controller/tree/master/docs/performance-tests/220%20node%20cluster#summary--highlights) and also a few times in the internal and external (gardener) slack channels.

But as I said, it might be OK reduce `maxReplicas` to 3. We can either chose to either publish an update (but we don't have a good number to publish yet). Or we can choose not to update right now and update only after the scale sub-resource support is in at which time we may not have to update at all. But this depends on the timeline. We can chose this path if everyone agrees.

ggaurav10:
As i mentioned, the number 220 nodes is already invalidated since we have introduced hoards of changes after our last perf runs. Explained in my last comment with an example of `limits` scaling.

The numbers published earlier (both at the hvpa repo link, and slack channels) are still correct, but not for the latests releases of both Gardener and hvpa-controller. The number has already changed, with or without reducing `maxReplicas`.

ggaurav10 (approved):
Approving the PR with a disclaimer that this change would result in higher costs of seeds, which can be mitigated by:
1. Reducing the `maxReplicas` for apiserver (preferably from 4 to 3), and/or
1. Reducing the HPA scale down stabilisation period (currently set to 24 hours)

rfranzke:
Hm, wait a sec, if you still have concerns then of the mitigations you mentioned should be included into this PR as well, or?

ggaurav10:
The PR looks good to me technically. But I am not sure if I alone should take the decision to hold the PR when cost comes into the picture. This should be a collective decision IMO. :)

amshuman-kr:
Like I said above, If we all agree, I will reduce `maxReplicas` to 3 in this PR itself. Also, we won't publish any updates to the maximum supported scale (though it will reduce) until we run a performance test with the scale sub-resource support.

Is this fine with everyone?

rfranzke:
Personally, I would prefer keeping `maxReplicas=4` as it allows for larger clusters. I think what definitely needs to be avoided is that, if we enable the memory HPA metric, even small clusters start to unnecessarily scale horizontally. This needs to be avoided independent of the `maxReplicas` number. Is this ensured with the `120%` value?

amshuman-kr:
> Reducing the HPA scale down stabilisation period (currently set to 24 hours)

I can't include this change in this PR. It will have to be on the landscape repo. Also, I will need data from #2080 to make that change to file that PR.

amshuman-kr:
> I think what definitely needs to be avoided is that, if we enable the memory HPA metric, even small clusters start to unnecessarily scale horizontally. This needs to be avoided independent of the maxReplicas number. Is this ensured with the 120% value?

It is possible this PR might increase the avg number of replicas per active shoot. But not as much as memory HPA at 80%. But there is no absolute guarantee as the currently available data of `6h` is not enough to say with certainty. 

Another possibility is to merge #2080 first and make a release. And based on the data from #2080 we can see if this PR reliably will not increase replicas for smaller clusters. WDYT?

rfranzke:
> Another possibility is to merge #2080 first and make a release. And based on the data from #2080 we can see if this PR reliably will not increase replicas for smaller clusters. WDYT?

Yes, I wanted to suggest this. Let's first gather the metrics and then make the change. @ggaurav10 mentioned that it won't be beneficial for the majority of clusters anyway (just for corner cases, if I understood correctly, which are important as well, of course).

amshuman-kr:
👍 Should I mark it as WIP and park it for now? Or do you want me to close it and raise it again later?

rfranzke:
Thanks, let's rather close it for now and reopen if applicable. Thanks Amshu! :)

PR #2071, opened 2020-03-19
Validate project namespace is not in use

timuthy:
This PR checks that a namespace specified in projects resources (`project.spec.namespace`) is not already taken by another project.

**Release note**:

```noteworthy user
Gardener now checks that a namespace specified in project resources (`project.spec.namespace`) is not already taken by another project.
```

### DIFF
--- a/plugin/pkg/global/resourcereferencemanager/admission.go
+++ b/plugin/pkg/global/resourcereferencemanager/admission.go
@@ -267,7 +267,8 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o 
 			return nil
 		}
 		// Set createdBy field in Project
-		if a.GetOperation() == admission.Create {
+		switch a.GetOperation() {
+		case admission.Create:
 			project.Spec.CreatedBy = &rbacv1.Subject{
 				APIGroup: "rbac.authorization.k8s.io",
 				Kind:     rbacv1.UserKind,
@@ -288,6 +289,16 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o 
 				}
 
 				project.Spec.Owner = owner
+			}
+
+			err = r.ensureProjectNamespace(project)
+		case admission.Update:
+			oldProject, ok := a.GetOldObject().(*core.Project)
+			if !ok {
+				return apierrors.NewBadRequest("could not convert old resource into Project object")
+			}
+			if oldProject.Spec.Namespace == nil && project.Spec.Namespace != nil {
+				err = r.ensureProjectNamespace(project)
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -308,6 +308,18 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
 				})
 			}
 		}
+
```
rfranzke: I guess it makes sense to only perform this list + for loop if the operation is either `CREATE`, or if it is `UPDATE` **and** the namespace has changed.

rfranzke (changes requested):

plugin/pkg/global/resourcereferencemanager/admission.go
timuthy: Done

rfranzke:
Can you drop commit 4439cbae5e424db7768b14690298304377b941aa? It's already merged as part of #2072.

timuthy:
> Can you drop commit 4439cba? It's already merged as part of #2072.

Okay, should be good now :D

rfranzke (approved):
/lgtm, thank you! :)

plugin/pkg/global/resourcereferencemanager/admission_test.go
```diff
@@ -600,6 +603,84 @@ var _ = Describe("resourcereferencemanager", func() {
...
+				Expect(err).To(Not(HaveOccurred()))
+			})
+
+			It("should allow specifying a namespace which is not in use (update)", func() {
+				projectOld := project.DeepCopy()
+				projectCopy := project.DeepCopy()
+				project.Spec.Namespace = pointer.StringPtr("garden-foo")
+				projectCopy.Name = "project-2"
+				projectCopy.Spec.Namespace = pointer.StringPtr("garden-bar")
+				gardenCoreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&projectCopy)
```
timebertt, line 627: On update the informer will already have the old project, right?
So, I guess, it makes sense to add it also in the test, because you are listing all projects in `ensureProjectNamespace ()`.
```suggestion
				gardenCoreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&projectOld)
				gardenCoreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&projectCopy)
```

plugin/pkg/global/resourcereferencemanager/admission_test.go
```diff
@@ -600,6 +603,84 @@ var _ = Describe("resourcereferencemanager", func() {
 					},
 				})))
 			})
+
+			It("should allow specifying a namespace which is not in use (create)", func() {
+				project.Spec.Namespace = pointer.StringPtr("garden-foo")
+				projectCopy := project.DeepCopy()
+				projectCopy.Name = "project-2"
+				projectCopy.Spec.Namespace = pointer.StringPtr("garden-bar")
+				gardenCoreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&projectCopy)
```
timebertt, line 612: This call actually errors because you are adding a pointer to `projectCopy` (which is already a `*core.Project`).
The test still compiles as the parameter is an `interface{}`, but the call errors and there will be no object in the store. 
But the test is therefore incorrect and will still pass, because you ignore the returned error.

I know, we do that in a lot of tests and IMHO we should stop ignoring errors returned by calls like this and always wrap such setup methods in `Expect(...).To(Succeed())`. Otherwise we may end up with incorrect tests that will pass.

Can you do that for all new test cases you added here?

```suggestion
				Expect(gardenCoreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(projectCopy)).To(Succeed())
```

timebertt (changes requested):
Sorry, I wasn't fast enough.
Still, I would like to request some changes in the tests :)

plugin/pkg/global/resourcereferencemanager/admission_test.go
timebertt: addressed with #2075

PR #2073, opened 2020-03-19
Add alpha scaling annotation to circumvent node criteria

rfranzke:
This PR adds an alpha annotation that can be used for shoots in order to decide the initial kube-apiserver size. As of today this is decided based on the minimum or maximum number of worker nodes. Though, this is only an approximation and does not work generally. Eventually, HVPA (if activated) takes care of scaling, though, for some use-cases an initially largely proportioned API server is required.

/cc @ggaurav10 @amshuman-kr 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/common/types.go
+++ b/pkg/operation/common/types.go
@@ -290,6 +290,13 @@ const (
 	// SecretRefChecksumAnnotation is the annotation key for checksum of referred secret in resource spec.
 	SecretRefChecksumAnnotation = "checksum/secret.data"
 
+	// ShootAlphaScalingAPIServerClass is a constant for an annotation on the shoot stating the initial API server class.
+	// It influences the size of the initial resource requests/limits.
+	// Possible values are [small, medium, large, xlarge, 2xlarge].
+	// Note that this annotation is alpha and can be removed anytime without further notice. Only use it if you know
+	// what you do.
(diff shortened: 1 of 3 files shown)


danielfoehrKn (dismissed):

ggaurav10 (dismissed):

pkg/operation/common/types.go
```diff
@@ -290,6 +290,12 @@ const (
 	// SecretRefChecksumAnnotation is the annotation key for checksum of referred secret in resource spec.
 	SecretRefChecksumAnnotation = "checksum/secret.data"
 
+	// ShootAlphaScalingAPIServerSize is a constant for an annotation on the shoot stating the initial API server size.
+	// Possible values are [small, medium, large, xlarge, 2xlarge].
+	// Note that this annotation is alpha and can be removed anytime without further notice. Only use it if you know
+	// what you do.
+	ShootAlphaScalingAPIServerSize = "alpha.kube-apiserver.scaling.shoot.gardener.cloud/size"
```
vpnachev: why not
```suggestion
	ShootAlphaScalingAPIServerSize = "alpha.kube-apiserver.scaling.shoot.gardener.cloud/class"
```

rfranzke: Makes sense, thanks. I was first using size but later decided to call it 'class' and forgot to rename it. PTAL again :)

vpnachev (approved):
/approve

PR #1842, opened 2020-01-23
Add support for nodeless local Gardener setup

zanetworker:
**What this PR does / why we need it**:
This PR adds support for a `nodeless` local Gardener setup.  With this setup, Kubernetes on docker is not required, the only requirement is `Docker`. All the components for this cluster will run as containers. 

For more information, please have a look at the documentation (`hack/local-garden/README.md`).

**Which issue(s) this PR fixes**:
Fixes #1812

**Special notes for your reviewer**:

**Release note**:

```improvement user
Developers can now make use of a `nodeless` local gardener setup. Simply run `make local-garden-up` and get a nodeless Garden cluster. 
```

Issue #1605, opened 2019-11-11
Protobuf generation is missing


rfranzke:
**What happened**:
Using Gardener on a 1.16 Kubernetes cluster does not work properly because the kube-controller-manager only uses `protobuf` in 1.16, and the Gardener does not yet support this content type.

**What you expected to happen**:
Gardener should provide protobuf definitons.

**How to reproduce it (as minimally and precisely as possible)**:
Deploy Gardener on 1.16 cluster and try to delete a namespace. See error logs in the kube-controller-manager like:

```
E1111 09:24:51.733918       1 namespace_controller.go:148] deletion of namespace extension-provider-openstack-7ld84 failed: [object *v1beta1.ShootList does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message, object *v1alpha1.ShootList does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message, object *v1beta1.SecretBindingList does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message, object *v1alpha1.OpenIDConnectPresetList does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message, object *v1beta1.QuotaList does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message, object *v1alpha1.SecretBindingList does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message, object *v1alpha1.PlantList does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message, object *v1alpha1.QuotaList does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message, object *v1alpha1.BackupEntryList does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message]
```

**Environment**:

- Gardener version: 0.31.3
- Kubernetes version (use `kubectl version`): 1.16.2
- Cloud provider or hardware configuration: AWS

vlerenc:
@mvladev Didn't you want to build that right away? Can someone else help?

ialidzhikov:
@vlerenc, I will check today.

ialidzhikov:
Ok, we reproduced this with @vpnachev . In my local setup (minikube v1.16) 
```
k create ns foo
k delete ns foo
```

The ns status
```yaml
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: "2019-12-10T14:02:34Z"
  deletionTimestamp: "2019-12-10T14:02:55Z"
  name: bar
[...]

PR #2075, opened 2020-03-20
Use K8s 1.15 for nodeless local Gardener

timuthy:
Currently the nodeless local Gardener setup (https://github.com/gardener/gardener/pull/1842) for development purposes uses K8s 1.17 but Gardener is not ready to run versions > 1.15 yet (see https://github.com/gardener/gardener/issues/1605). This makes it impossible to successfully delete a project and its related namespace. Let's switch back to 1.15 as long as https://github.com/gardener/gardener/issues/1605 or  is not resolved.

/cc @zanetworker

I revised the admission tests along the way @tim-ebert.

**Release note**:

```noteworthy developer
⚠️ The local nodeless Gardener setup (for local development) has been reverted to K8s 1.15 in order to be consistent with the highest version supported by Gardener in a production environment.
```

### DIFF
--- a/docs/usage/supported_k8s_versions.md
+++ b/docs/usage/supported_k8s_versions.md
@@ -4,8 +4,9 @@ Currently, the Gardener supports the following Kubernetes versions:
 
 ## Garden cluster version
 
-:warning: The minimum version of the garden cluster that can be used to run Gardener is **`1.10.x`**.
+:warning: The Kubernetes version of a garden cluster that can be used to run Gardener must be in the range of **`1.10.x`** to  **`1.15.x`**.
 The reason for that is that the least supported Kubernetes version in Gardener is `1.10`.
+Furthermore, Gardener's [extension API server](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/) does not support version >= `1.16` yet.
 
 ## Seed cluster versions
 
--- a/hack/local-garden/run-kube-apiserver
+++ b/hack/local-garden/run-kube-apiserver
@@ -11,13 +11,14 @@ LABEL=${1:-local-garden}
 
 SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 
-IMAGE=eu.gcr.io/gardener-project/3rd/k8s_gcr_io/kube-apiserver:v1.17.1
+IMAGE=k8s.gcr.io/hyperkube:v1.15.11
 MOUNTS="-v $SCRIPTPATH/certificates/certs:/certs -v $SCRIPTPATH/certificates/keys:/keys"
 ETCD_PORT=12379
 LISTEN_PORT="2443"
 
 echo "Starting gardener-dev kube-apiserver..!"
-docker run -d --name kube-apiserver -l $LABEL --add-host $HOSTS_LINE --network gardener-dev --rm -p $LISTEN_PORT:$LISTEN_PORT $MOUNTS  $IMAGE /usr/local/bin/kube-apiserver \
+docker run -d --name kube-apiserver -l $LABEL --add-host $HOSTS_LINE --network gardener-dev --rm -p $LISTEN_PORT:$LISTEN_PORT $MOUNTS  $IMAGE /hyperkube \
+  kube-apiserver \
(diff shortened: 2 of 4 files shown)


plugin/pkg/global/resourcereferencemanager/admission_test.go
```diff
@@ -198,7 +198,7 @@ var _ = Describe("resourcereferencemanager", func() {
 		Context("tests for SecretBinding objects", func() {
 			It("should accept because all referenced objects have been found (secret found in cache)", func() {
 				kubeInformerFactory.Core().V1().Secrets().Informer().GetStore().Add(&secret)
```
timebertt: Is there a reason why you didn't wrap the `Add()` calls to `kubeInformerFactory` in `Expect`s like the others?

docs/usage/supported_k8s_versions.md
```diff
@@ -4,8 +4,9 @@ Currently, the Gardener supports the following Kubernetes versions:
 
 ## Garden cluster version
 
-:warning: The minimum version of the garden cluster that can be used to run Gardener is **`1.10.x`**.
+:warning: The Kubernetes version of a garden cluster that can be used to run Gardener must be in the range of **`1.10.x`** to  **`1.15.x`**.
 The reason for that is that the least supported Kubernetes version in Gardener is `1.10`.
+Furthermore, Gardener's [extension API server](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/) does not support version >= `1.16` yet.
```
rfranzke, line 9: Out of curiosity: Why not? Because of the admission webhooks? In this case, can you enhance [this code](https://github.com/gardener/gardener/blob/master/pkg/operation/garden/garden.go#L299-L306)?

timebertt (commented):
Thanks for fixing!
/lgtm in general

hack/local-garden/run-kube-apiserver
```diff
@@ -11,13 +11,14 @@ LABEL=${1:-local-garden}
 
 SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 
-IMAGE=eu.gcr.io/gardener-project/3rd/k8s_gcr_io/kube-apiserver:v1.17.1
+IMAGE=eu.gcr.io/gardener-project/k8s.gcr.io/hyperkube:v1.15.10
```
rfranzke:
```suggestion
IMAGE=k8s.gcr.io/hyperkube:v1.15.10
```

Can we use an upstream image?

hack/local-garden/run-kube-controller-manager
```diff
@@ -4,12 +4,13 @@ LABEL=${1:-local-garden}
 
 SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 
-IMAGE=eu.gcr.io/gardener-project/3rd/k8s_gcr_io/kube-controller-manager:v1.17.1
+IMAGE=eu.gcr.io/gardener-project/k8s.gcr.io/hyperkube:v1.15.10
```
rfranzke:
```suggestion
IMAGE=k8s.gcr.io/hyperkube:v1.15.10
```

Can we use an upstream image?

rfranzke (changes requested):

docs/usage/supported_k8s_versions.md
timuthy: Hm, I thought because of #1605?

rfranzke: Ah, right, thanks for the reminder. :)

plugin/pkg/global/resourcereferencemanager/admission_test.go
timuthy: I just missed them. Thanks for pointing out :)

zanetworker:
makes sense, I think we should make this configurable as well with a default of latest supported  version of Gardener.

zanetworker:
btw, I think some of the flags for the certificates are changed as well, did you get a chance to validate that? 

Update: probably for <1.15. Tested it now and looks good :)

rfranzke (approved):

timebertt (approved):
Thanks, /lgtm now :)

PR #2077, opened 2020-03-20
Enable ScheduleDaemonSetPods featuregate by default for v1.11

### LINKED ISSUE #2076: Enable ScheduleDaemonSetPods feature gate for 1.11 shoots
**What happened**:
For < 1.12 shoot clusters the daemon set pods are not scheduled by the kube-scheduler but by the kube-controller-manager. This can lead to situations where pods with a high priority cannot be started properly.

(Daemon set) Pods with high priority can be properly scheduled.

1. Create a 1.11 cluster
1. Deploy pods into it that request almost all resources
1. Try to deploy a daemon set with `system-cluster-critical` priority class

**How fix it**:
Since 1.11, Kubernetes has introduced a `ScheduleDaemonSetPods` feature gate that makes the kube-scheduler schedule the pods instead of the KCM. This fixes the above mentioned problem.
The feature gate is enabled by default beginning with Kubernetes v1.12.
Let's add the feature gate to both the kube-controller-manager and kube-scheduler configuration by default for 1.11 clusters.

**Environment**:

- Gardener version: v1.1.5
- Kubernetes version (use `kubectl version`): v1.11.10
- Cloud provider or hardware configuration: AWS

zanetworker:
For < 1.12 shoot clusters the daemon set pods are not scheduled by the kube-scheduler but by the kube-controller-manager. This can lead to situations where pods with a high priority cannot be started properly. This PR adds `ScheduleDaemonSetPods` by default for  v1.11 clusters. 

Fixes #2076 

**Release note**:

```improvement operator
ScheduleDaemonSetPods feature gate is now enabled by default for 1.11 clusters to ensure correct scheduling of system-critical-pods
```

### DIFF
--- a/charts/seed-controlplane/charts/kube-apiserver/templates/_helpers.tpl
+++ b/charts/seed-controlplane/charts/kube-apiserver/templates/_helpers.tpl
@@ -8,7 +8,7 @@
 {{- if semverCompare "< 1.12" .Values.kubernetesVersion }}
 - --feature-gates=TokenRequest=true
 {{- end }}
-{{- if semverCompare "1.11" .Values.kubernetesVersion }}
+{{- if semverCompare "1.11.x" .Values.kubernetesVersion }}
 - --feature-gates=TokenRequestProjection=true
 {{- end }}
 {{- end -}}
--- a/charts/seed-controlplane/charts/kube-controller-manager/templates/_helpers.tpl
+++ b/charts/seed-controlplane/charts/kube-controller-manager/templates/_helpers.tpl
@@ -2,6 +2,9 @@
 {{- if .Values.featureGates }}
 - --feature-gates={{ range $feature, $enabled := .Values.featureGates }}{{ $feature }}={{ $enabled }},{{ end }}
 {{- end }}
+{{- if semverCompare "1.11.x" .Values.kubernetesVersion }}
+- --feature-gates=ScheduleDaemonSetPods=true
+{{- end }}
 {{- end -}}
 
 {{- define "kube-controller-manager.port" -}}
--- a/charts/seed-controlplane/charts/kube-scheduler/templates/_helpers.tpl
+++ b/charts/seed-controlplane/charts/kube-scheduler/templates/_helpers.tpl
@@ -5,6 +5,9 @@
 {{- if semverCompare "< 1.11" .Values.kubernetesVersion }}
 - --feature-gates=PodPriority=true
 {{- end }}
+{{- if semverCompare "1.11.x" .Values.kubernetesVersion }}
+- --feature-gates=ScheduleDaemonSetPods=true
+{{- end }}
 {{- end -}}
 
 {{- define "kube-scheduler.componentconfigversion" -}}


charts/seed-controlplane/charts/kube-scheduler/templates/_helpers.tpl
```diff
@@ -6,6 +6,9 @@
 - --feature-gates=PodPriority=true
 {{- end }}
 {{- end -}}
+{{- if semverCompare "1.11" .Values.kubernetesVersion }}
+- --feature-gates=ScheduleDaemonSetPods=true
+{{- end }}
```
rfranzke:
```suggestion
{{- if semverCompare "1.11.x" .Values.kubernetesVersion }}
- --feature-gates=ScheduleDaemonSetPods=true
{{- end }}
{{- end -}}
```

charts/seed-controlplane/charts/kube-controller-manager/templates/_helpers.tpl
```diff
@@ -2,6 +2,9 @@
 {{- if .Values.featureGates }}
 - --feature-gates={{ range $feature, $enabled := .Values.featureGates }}{{ $feature }}={{ $enabled }},{{ end }}
 {{- end }}
+{{- if semverCompare "1.11" .Values.kubernetesVersion }}
```
rfranzke:
```suggestion
{{- if semverCompare "1.11.x" .Values.kubernetesVersion }}
```

rfranzke (changes requested):

charts/seed-controlplane/charts/kube-apiserver/templates/_helpers.tpl
```diff
@@ -8,7 +8,7 @@
 {{- if semverCompare "< 1.12" .Values.kubernetesVersion }}
 - --feature-gates=TokenRequest=true
 {{- end }}
-{{- if semverCompare "1.11" .Values.kubernetesVersion }}
+{{- if semverCompare "1.11.x" .Values.kubernetesVersion }}
```
rfranzke, line 11: nice catch! :)

rfranzke (approved):
/lgtm

rfranzke:
@zanetworker we have seen for a 1.11.10 cluster having this feature gate enabled that the kube-scheduler does not schedule the DaemonSet pods for newly created nodes with the reasoning "node not ready". However, for the node to get ready the DaemonSet pods for kube-proxy + calico need to be installed (both running in host network). But as the kube-scheduler does not schedule them the node can never get ready.

zanetworker:
Closing this PR based on our discussion. 

For 1.10 and 1.11 clusters enabling this feature-gate won't help much as the `kube-scheduler` takes node readiness into consideration  and fails to schedule Networking pods (e.g., Calico) which will result in a broken setup due to `FailedPlacement`.  As a result, we recommend upgrading to higher versions of kubernetes for clusters with these issues where the feature-gate is enabled by default.

PR #2080, opened 2020-03-20
Aggregate metrics for containers at scaling and limiting threshold.

amshuman-kr:
Added the following aggregate metrics (by containers).

1. To keep track of the number of containers whose resource usage is over scaling, limiting thresholds, vpa recommendations.
1. To keep track of the number containers which have pending scale down or scale up recommendations.
1. To keep track of minimum and maximum values for usage, requests, limits and VPA recommendations for containers across the cluster.
1. To keep track of the number of containers that have requests over a threshold percentage of their VPA maxAllowed configuration.

This can help evaluation and analyse different aspects and performance of horizontal and vertical scaling.

Fixes #

**Release note**:

```improvement operator
Added the following aggregate metrics (by containers). Firstly, to keep track of the number of containers whose resource usage is over scaling, limiting thresholds, vpa recommendations. Secondly, to keep track of the number containers which have pending scale down or scale up recommendations. Thirdly, to keep track of minimum and maximum values for usage, requests, limits and VPA recommendations for containers across the cluster. Fourthly, to keep track of the number of containers that have requests over a threshold percentage of their VPA maxAllowed configuration.
```

### DIFF
--- a/charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
+++ b/charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
@@ -36,6 +36,182 @@ groups:
   - record: seed:hvpa_status_applied_vpa_recommendation_target:sum_by_container
     expr: sum(hvpa_status_applied_vpa_recommendation{recommendation="target"}) by (container, resource)
 
+  # Recording rules for the count of apiserver containers which have CPU usage at or above 80% of their request.
+  # 1. This helps evaluate if memory-based HPA should be re-introduced at 120%. And if it is re-opened what will be the impact on the avg number of replicas per active shoot in the seed clusters.
+  # 2. This helps evaluate the benefit in reducing HPA scale down stabilization period (currently, 24h) to gain better resource utilization.
+  - record: seed:apiserver_containers_over_cpu_scaling_threshold_total:count_by_container
+    expr: count(sum(rate(container_cpu_usage_seconds_total{container="kube-apiserver"}[5m])) by (namespace) / sum(kube_pod_container_resource_requests_cpu_cores{container="kube-apiserver"}) by (namespace) >= 0.8)
+
+  # Recording rules for the count of apiserver containers which have memory usage at or above 120% of their request.
+  # 1. This helps evaluate if memory-based HPA should be re-introduced at 120%. And if it is re-opened what will be the impact on the avg number of replicas per active shoot in the seed clusters.
+  # 2. This helps evaluate the benefit in reducing HPA scale down stabilization period (currently, 24h) to gain better resource utilization.
+  - record: seed:apiserver_containers_over_memory_scaling_threshold_total:count_by_container
+    expr: count(sum(container_memory_working_set_bytes{container="kube-apiserver"}) by (namespace) / sum(kube_pod_container_resource_requests_memory_bytes{container="kube-apiserver"}) by (namespace) >= 1.2)
+
+  # Recording rules for the count of containers which have CPU usage below 80% of their limits, aggregated by containers.
+  # How many shoots and components are not at risk of being affected (CPU-throttled or OOMKilled) if scaling (horizontally or vertically) is blocked due to any reason (bug, deliberate or wrong configuration).
+  - record: seed:containers_below_cpu_limiting_threshold_total:count_by_container
+    expr: count(max(rate(container_cpu_usage_seconds_total[5m])) by (container, namespace) / max(kube_pod_container_resource_limits_cpu_cores) by (container, namespace) < 0.8) by (container)
+
+  # Recording rules for the count of containers which have memory usage below 80% of their limits, aggregated by containers.
+  # How many shoots and components are not at risk of being affected (CPU-throttled or OOMKilled) if scaling (horizontally or vertically) is blocked due to any reason (bug, deliberate or wrong configuration).
+  - record: seed:containers_below_memory_limiting_threshold_total:count_by_container
+    expr: count(max(container_memory_working_set_bytes) by (container, namespace) / max(kube_pod_container_resource_limits_memory_bytes) by (container, namespace) < 0.8) by (container)
+
+  # Recording rules for the count of containers which have CPU usage below 100% of their VPA recommendations, aggregated by containers.
+  # How many shoots and components are not affected by the observed case where VPA was not recommending higher values even when usage was consistently above the present VPA recommendation.
+  - record: seed:containers_below_cpu_recommendation_total:count_by_container
+    expr: count(max(rate(container_cpu_usage_seconds_total[5m])) by (container, namespace) / max(vpa_status_recommendation{recommendation="target", resource="cpu"}) by (container, namespace) < 1) by (container)
+
+  # Recording rules for the count of containers which have memory usage below 100% of their VPA recommendations, aggregated by containers.
+  # How many shoots and components are not affected by the observed case where VPA was not recommending higher values even when usage was consistently above the present VPA recommendation.
+  - record: seed:containers_below_memory_recommendation_total:count_by_container
+    expr: count(max(container_memory_working_set_bytes) by (container, namespace) / max(vpa_status_recommendation{recommendation="target", resource="memory"}) by (container, namespace) < 1) by (container)
+
+  # Recording rules for the count of containers which have CPU usage at or above 80% of their limits, aggregated by containers.
+  # How many shoots and components are at risk of being affected (CPU-throttled or OOMKilled) if scaling (horizontally or vertically) is blocked due to any reason (bug, deliberate or wrong configuration).
+  - record: seed:containers_over_cpu_limiting_threshold_total:count_by_container
+    expr: count(max(rate(container_cpu_usage_seconds_total[5m])) by (container, namespace) / max(kube_pod_container_resource_limits_cpu_cores) by (container, namespace) >= 0.8) by (container)
+
+  # Recording rules for the count of containers which have memory usage at or above 80% of their limits, aggregated by containers.
+  # How many shoots and components are at risk of being affected (CPU-throttled or OOMKilled) if scaling (horizontally or vertically) is blocked due to any reason (bug, deliberate or wrong configuration).
+  - record: seed:containers_over_memory_limiting_threshold_total:count_by_container
+    expr: count(max(container_memory_working_set_bytes) by (container, namespace) / max(kube_pod_container_resource_limits_memory_bytes) by (container, namespace) >= 0.8) by (container)
+
+  # Recording rules for the count of containers which have CPU usage at or above 100% of their VPA recommendations, aggregated by containers.
+  # How many shoots and components are affected by the observed case where VPA was not recommending higher values even when usage was consistently above the present VPA recommendation.
+  - record: seed:containers_over_cpu_recommendation_total:count_by_container
+    expr: count(max(rate(container_cpu_usage_seconds_total[5m])) by (container, namespace) / max(vpa_status_recommendation{recommendation="target", resource="cpu"}) by (container, namespace) >= 1) by (container)
+
+  # Recording rules for the count of containers which have memory usage at or above 100% of their VPA recommendations, aggregated by containers.
+  # How many shoots and components are affected by the observed case where VPA was not recommending higher values even when usage was consistently above the present VPA recommendation.
+  - record: seed:containers_over_memory_recommendation_total:count_by_container
+    expr: count(max(container_memory_working_set_bytes) by (container, namespace) / max(vpa_status_recommendation{recommendation="target", resource="memory"}) by (container, namespace) >= 1) by (container)
+
+  # Recording rules for the count of containers which have CPU requests below 100% of their VPA recommendations, aggregated by containers.
+  # How many shoots and components are adversely affected by any delay in scaling up (due to bugs or deliberate/wrong configuration of HPA, VPA, HVPA).
+  - record: seed:containers_cpu_scale_up_pending_total:count_by_container
+    expr: count(max(kube_pod_container_resource_requests_cpu_cores) by (container, namespace) / max(vpa_status_recommendation{recommendation="target", resource="cpu"}) by (container, namespace) < 1) by (container)
+
+  # Recording rules for the count of containers which have memory requests below 100% of their VPA recommendations, aggregated by containers.
+  # How many shoots and components are adversely affected by any delay in scaling up (due to bugs or deliberate/wrong configuration of HPA, VPA, HVPA).
+  - record: seed:containers_memory_scale_up_pending_total:count_by_container
+    expr: count(max(kube_pod_container_resource_requests_memory_bytes) by (container, namespace) / max(vpa_status_recommendation{recommendation="target", resource="memory"}) by (container, namespace) < 1) by (container)
+
+  # Recording rules for the count of containers which have CPU requests above 100% of their VPA recommendations, aggregated by containers.
+  # How many shoots and components are costing extra resources by any delay in scaling down (due to bugs or deliberate/wrong configuration of HPA, VPA, HVPA).
+  - record: seed:containers_cpu_scale_down_pending_total:count_by_container
+    expr: count(max(kube_pod_container_resource_requests_cpu_cores) by (container, namespace) / max(vpa_status_recommendation{recommendation="target", resource="cpu"}) by (container, namespace) > 1) by (container)
+
+  # Recording rules for the count of containers which have memory requests at or above 100% of their VPA recommendations, aggregated by containers.
+  # How many shoots and components are costing extra resources by any delay in scaling down (due to bugs or deliberate/wrong configuration of HPA, VPA, HVPA).
+  - record: seed:containers_memory_scale_down_pending_total:count_by_container
+    expr: count(max(kube_pod_container_resource_requests_memory_bytes) by (container, namespace) / max(vpa_status_recommendation{recommendation="target", resource="memory"}) by (container, namespace) > 1) by (container)
+
+  # Recording rules for the count of containers which have CPU requests below 60% of their VPA maxAllowed configuration, aggregated by containers.
+  # How many shoots are below 60% of the configured maxAllowed VPA recommendations.
+  - record: seed:containers_below_cpu_max_allowed_threshold_total:count_by_container
+    expr: count(max(kube_pod_container_resource_requests_cpu_cores) by (container, namespace) / max(vpa_spec_container_resource_policy_allowed{allowed="max", resource="cpu"}) by (container, namespace) < 0.6) by (container)
+
+  # Recording rules for the count of containers which have memory requests below 60% of their VPA maxAllowed configuration, aggregated by containers.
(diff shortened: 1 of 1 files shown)


amshuman-kr:
cc @ggaurav10 @wyb1

amshuman-kr:
Added min and max metrics for usage, requests, limits and recommendations aggregated by containers to help tune VPA configurations for these components and also help while deciding if dedicated worker pools might help for some of these components.

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
```diff
@@ -36,6 +36,102 @@ groups:
...
+  # Recording rules for the count of containers which have CPU requests above 100% of their VPA recommendations, aggregated by containers.
+  - record: seed:containers_cpu_scale_down_pending_total:count_by_container
+    expr: count(max(kube_pod_container_resource_requests_cpu_cores) by (container, namespace) / max(vpa_status_recommendation{recommendation="target", resource="cpu"}) by (container, namespace) > 1) by (container)
+
+  # Recording rules for the count of containers which have memory requests at or above 100% of their VPA recommendations, aggregated by containers.
+  - record: seed:containers_memory_scale_down_pending_total:count_by_container
+    expr: count(max(kube_pod_container_resource_requests_memory_bytes) by (container, namespace) / max(vpa_status_recommendation{recommendation="target", resource="memory"}) by (container, namespace) > 1) by (container)
+
+  # Recording rules for the minimum CPU usage aggregated by container
+  - record: seed:container_cpu_usage_seconds_total:max_by_container
```
wyb1:
```suggestion
  - record: seed:container_cpu_usage_seconds_total:min_by_container
```

Many recording rules below this are named `*_max_by_container` even though the operator used is `min`. Is this intended?

amshuman-kr: Oops. Missed the substitution. Will fix shortly.

amshuman-kr:
@wyb1 I have fixed the missed substitution. Can you PTAL?

amshuman-kr:
Added two more metrics to keep track of the number of containers that have requests over a threshold percentage of their VPA maxAllowed configuration. Will help in keeping track of clusters with higher load.

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
```diff
@@ -36,6 +36,110 @@ groups:
...
+  # Recording rules for the count of containers which have CPU usage at or above 80% of their request, aggregated by containers.
+  - record: seed:containers_over_cpu_scaling_threshold_total:count_by_container
+    expr: count(max(rate(container_cpu_usage_seconds_total[5m])) by (container, namespace) / max(kube_pod_container_resource_requests_cpu_cores) by (container, namespace) >= 0.8) by (container)
+
+  # Recording rules for the count of containers which have memory usage at or above 120% of their request, aggregated by containers.
+  - record: seed:containers_over_memory_scaling_threshold_total:count_by_container
+    expr: count(max(container_memory_working_set_bytes) by (container, namespace) / max(kube_pod_container_resource_requests_memory_bytes) by (container, namespace) >= 1.2) by (container)
+
+  # Recording rules for the count of containers which have CPU usage at or above 80% of their limits, aggregated by containers.
+  - record: seed:containers_over_cpu_limiting_threshold_total:count_by_container
```
wyb1, line 73: What is the difference between this and `seed:containers_over_cpu_scaling_threshold_total:count_by_container`? The expressions are the same.

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
```diff
@@ -36,6 +36,110 @@ groups:
...
+    expr: count(max(rate(container_cpu_usage_seconds_total[5m])) by (container, namespace) / max(kube_pod_container_resource_limits_cpu_cores) by (container, namespace) >= 0.8) by (container)
+
+  # Recording rules for the count of containers which have memory usage at or above 80% of their limits, aggregated by containers.
+  - record: seed:containers_over_memory_limiting_threshold_total:count_by_container
+    expr: count(max(container_memory_working_set_bytes) by (container, namespace) / max(kube_pod_container_resource_limits_memory_bytes) by (container, namespace) >= 0.8) by (container)
```
wyb1, line 79: I'm not sure if `max(container_memory_working_set_bytes) by (container, namespace)` is the correct expression. For example if we have a namespace with multiple containers with the same name and they are all above the threshold, only 1 will be counted instead of all of them.

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
```diff
@@ -36,6 +36,110 @@ groups:
...
+
+  # Recording rules for the count of containers which have CPU usage at or above 100% of their VPA recommendations, aggregated by containers.
```
wyb1, line 81: I don't think recording rules is the right approach for these metrics (or at least how it is implemented right now). What if we want to know the count of containers that have CPU usage at or above 80%? 50%? Or if we want to switch it around and know how many containers use less than 100%? To do this we would have to add the metrics and wait for a new gardener release. Also having this many recording rules is confusing and it is easy to lose oversight. Perhaps we can find a better way to do this.

wyb1 (commented):
We can't keep making recording rules for these metrics. Soon there will be no more oversight of these metrics. To counteract this I propose 2 things:
1. There should be dashboards for the metrics in the seed grafana (the grafana running in the garden ns).
2. The recording rules would be much easier to write and understand if they came from [histograms](https://prometheus.io/docs/practices/histograms/). Then we could, for example, have metrics that would tell us 90% of containers under VPA are using 60% of their requests recommended by VPA. Then we only need 1 recording rule (i.e. a metric that says X% of containers under VPA have value Y.) per resource. We would basically be defining SLOs for VPA/HVPA. The problem with this is that the metrics have to come from another exporter. So either our vpa exporter or a new exporter.

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
amshuman-kr: No. The expressions are different.

`seed:containers_over_cpu_scaling_threshold_total:count_by_container` is based on `kube_pod_container_resource_requests_cpu_cores` indicating how many namespaces have this container at or above the threshold where they might be scaled. 0.8 = 80% because that is the utilization percentage when HPA is suppose scale the replicas. This is useful to evaluate how many shoot namespaces are at threshold of HPA scaling.

`seed:containers_over_cpu_limiting_threshold_total:count_by_container` is based on indicating how many namespaces have this container at or above the threshold where their resources might be limited by docker/kubelet. This can help in analysing how many shoots are at risk of being affected if scaling (horizontally or vertically) is blocked due to any reason (bug, deliberate or wrong configuration).

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
amshuman-kr: The rationale behind using `max` is to be conservative. The idea is that if _any_ of the replicas of the container in the namespace is at or above the threshold of limits then we count it. Not just if the total of usage is above the total limits. This is especially important because we have seen that Kubernetes `Services` do a poor job of load balancing across replicas.

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
amshuman-kr: Agreed about the general goodness of using histograms for this. But I don't see any standard histogram metrics which I can use for this purpose. Please let me know if you are aware of any.

If you notice, all of the metrics below are at 100%. The rationale is to help tune the VPA and HPA to see it is recommending values right (safe enough and simultaneously not too generous). For this, 100% is sufficient and we do not need the intermediate percentiles.

In general, it would be desirable to create histogram metrics with quantile/percentiles. But that would be larger effort and will require more time. I would rather get these aggregate metrics in now and evaluate/tune VPA and HPA. We can work on histogram custom metrics later and replace these metrics. WDYT?

amshuman-kr:
> Soon there will be no more oversight of these metrics.

I volunteer to keep the oversight.

To counteract this I propose 2 things:

> There should be dashboards for the metrics in the seed grafana (the grafana running in the garden ns).

But we cannot create the dashboard without these metrics. Or did you propose that we create the dashboards with these metrics?

> The recording rules would be much easier to write and understand if they came from histograms. Then we could, for example, have metrics that would tell us 90% of containers under VPA are using 60% of their requests recommended by VPA. Then we only need 1 recording rule (i.e. a metric that says X% of containers under VPA have value Y.) per resource. We would basically be defining SLOs for VPA/HVPA. The problem with this is that the metrics have to come from another exporter. So either our vpa exporter or a new exporter.

Agreed about the general goodness of using histograms for this. But I don't see any standard histogram metrics which I can use for this purpose. Please let me know if you are aware of any.

If you notice, almost all of the metrics below are at 100%. The rationale is to help tune the VPA and HPA to see it is recommending values right (safe enough and simultaneously not too generous). For this, 100% is sufficient and we do not need the intermediate percentiles.

The only exceptions are `seed:containers_over_cpu_scaling_threshold_total:count_by_container` and `seed:containers_over_memory_scaling_threshold_total:count_by_container` which are at 80% and 120% explicitly for evaluating HPA behaviour. This is to evaluate the need and impact of re-opening #2070.

In general, it would be desirable to create histogram metrics with quantile/percentiles. But that would be larger effort and will require more time. I would rather get these aggregate metrics in now and evaluate/tune VPA and HPA. We can work on histogram custom metrics later and replace these metrics. WDYT?

amshuman-kr:
Listing the concrete use-case behind each of the new metrics, in case it helps.

- `seed:containers_over_cpu_scaling_threshold_total:count_by_container` and `seed:containers_over_memory_scaling_threshold_total:count_by_container`
   1. Help evaluate if #2070 needs to be re-opened. And if it is re-opened what will be the impact on the avg number of replicas per active shoot in the seed clusters.
   2. Help evaluate the benefit in reducing HPA scale down stabilization period (currently, `24h`) to gain better resource utilization.
- `seed:containers_over_cpu_limiting_threshold_total:count_by_container` and `seed:containers_over_memory_limiting_threshold_total:count_by_container`
  - How many shoots and components are at risk of being affected (CPU-throttled or `OOMKilled`) if scaling (horizontally or vertically) is blocked due to any reason (bug, deliberate or wrong configuration).
- `seed:containers_over_cpu_recommendation_total:count_by_container` and `seed:containers_over_memory_recommendation_total:count_by_container`
  - How many shoots and components are affected by the case noticed by @ggaurav10 where VPA was not recommending higher values even when usage was consistently above the present VPA recommendation.
- `seed:containers_cpu_scale_up_pending_total:count_by_container` and `seed:containers_memory_scale_up_pending_total:count_by_container`
   - How many shoots and components are adversely affected by any delay in scaling up (due to bugs or deliberate/wrong configuration of HPA, VPA, HVPA).
- `seed:containers_cpu_scale_down_pending_total:count_by_container` and `seed:containers_memory_scale_down_pending_total:count_by_container`
   - How many shoots and components are costing extra resources by any delay in scaling down (due to bugs or deliberate/wrong configuration of HPA, VPA, HVPA).
- `seed:containers_over_cpu_max_allowed_threshold_total:count_by_container` and `seed:containers_over_memory_max_allowed_threshold_total:count_by_container`
   - How many shoots are above 60% of the configured `maxAllowed` VPA recommendations. Currently, there seem to be none. But we need to know early if our end-users start getting close to the lists of the scale of the cluster that we have tested the performance so far. I will also create an alert in another PR for the individual shoot clusters based on this threshold.
- `seed:container_cpu_usage_seconds_total:min_by_container`, `seed:container_memory_working_set_bytes:min_by_container`, `seed:kube_pod_container_resource_requests_cpu_cores:min_by_container`,
`seed:kube_pod_container_resource_requests_memory_bytes:min_by_container`,
`seed:kube_pod_container_resource_limits_cpu_cores:min_by_container`,
`seed:kube_pod_container_resource_limits_memory_bytes:min_by_container`,
`seed:vpa_status_recommendation:min_by_container`
   1. How far down can we configure VPA `minAllowed` for the different components. We have some components that can be scaled down further.
   1. Get insight into grouping components once multiple (probably dedicated) worker-pools are introduced in seed clusters.
- `seed:container_cpu_usage_seconds_total:max_by_container`,
`seed:container_memory_working_set_bytes:max_by_container`,
`seed:kube_pod_container_resource_requests_cpu_cores:max_by_container`,
`seed:kube_pod_container_resource_requests_memory_bytes:max_by_container`,
`seed:kube_pod_container_resource_limits_cpu_cores:max_by_container`,
`seed:kube_pod_container_resource_limits_memory_bytes:max_by_container,
`seed:vpa_status_recommendation:max_by_container`
   1. How far down can we configure VPA `maxAllowed` for the different components. We have some components (especially, the non-shoot workload) that need to be scaled up more.
   1. Get insight into grouping components once multiple (probably dedicated) worker-pools are introduced in seed clusters.

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
```diff
@@ -36,6 +36,110 @@ groups:
   - record: seed:hvpa_status_applied_vpa_recommendation_target:sum_by_container
     expr: sum(hvpa_status_applied_vpa_recommendation{recommendation="target"}) by (container, resource)
 
+  # Recording rules for the count of containers which have CPU usage at or above 80% of their request, aggregated by containers.
+  - record: seed:containers_over_cpu_scaling_threshold_total:count_by_container
+    expr: count(max(rate(container_cpu_usage_seconds_total[5m])) by (container, namespace) / max(kube_pod_container_resource_requests_cpu_cores) by (container, namespace) >= 0.8) by (container)
+
+  # Recording rules for the count of containers which have memory usage at or above 120% of their request, aggregated by containers.
+  - record: seed:containers_over_memory_scaling_threshold_total:count_by_container
+    expr: count(max(container_memory_working_set_bytes) by (container, namespace) / max(kube_pod_container_resource_requests_memory_bytes) by (container, namespace) >= 1.2) by (container)
```
ggaurav10: As discussed separately, since this is supposed to provide info on HPA scaling, it should use average utilisation across namespace, and not max.
```
count(avg(rate(container_cpu_usage_seconds_total[5m])) by (container, namespace) / max(kube_pod_container_resource_requests_cpu_cores) by (container, namespace) >= 0.8) by (container)
```
or
```
count(sum(rate(container_cpu_usage_seconds_total[5m])) by (container, namespace) / sum(kube_pod_container_resource_requests_cpu_cores) by (container, namespace) >= 0.8) by (container)
```

Also, it might be better to filter based on relevant container, and use for eg. `container_cpu_usage_seconds_total{container="kube-apiserver"}`, so that these calculations are not done for all the control plane containers. From what I have observed, letting prometheus do lot of calculations triggers VPA update on prometheus making it unavailable for several minutes.

Also, since the thresholds (80%,120%) used in these expressions are very specific to apiserver's HPA, it makes more sense to filter based on only apiserver container.

It may also be worthwhile to add a comment somewhere that these expressions provide misleading data when apiserver is going through rolling update because of vertical scale up.

ggaurav10 (changes requested):

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
amshuman-kr: Already made the change yesterday. But forgot to push. Done now. I have also added comments throughout (the same as what I mentioned in this PR discussion). PTAL.

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
ggaurav10: What if we change the expression to:
```
max(rate(container_cpu_usage_seconds_total[5m])) by (container, namespace) / max(vpa_status_recommendation{recommendation="target", resource="cpu"}) by (container, namespace)
```
Basically, remove the comparison and the count calculation. It can be further granulised, but may not really be required as of now.
This will enable us to query all kinds of percentages and counts as mentioned by @wyb1 
Although, on the downside, this might increase the disk usage for prometheus, but I think we should be able to live with that.
Same thing can be done for lot of other recording rules here providing more flexibility, for eg, to figure out a favourable threshold percentage for HPA, we might need to compare against 130% also, and not just 80% or 120%.
WDYT?

amshuman-kr: Nope. Cardinality = number of namespaces. Not good for aggregate prometheus.

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
wyb1: Ok, it will still work just making sure that you were aware of this. Then `max` is good 👍

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
wyb1: Ah ok my bad. I should have compared them side by side. Then the metrics are fine.

charts/seed-bootstrap/prometheus-rules/recording-rules.rules.yaml
wyb1: > We can work on histogram custom metrics later and replace these metrics. WDYT?

I understand you need these metrics to tune vpa/hvpa sooner rather than later so I am ok with this.

Howerver, I would at least like a metric that gives us the complement. So we know how many containers are using less than 100% of their given requests. This way we can account for all containers using VPA. Does this make sense?

ggaurav10: > Nope. Cardinality = number of namespaces. Not good for aggregate prometheus.

Just wanted to point out the benefits of added flexibility of comparisons by changing the expression (+ no need to roll out new release if we realise that we might need data for 130% threshold comparison for HPA for eg.).
After weighing pros and cons, if it still doesn't suit, then this works too 👍

amshuman-kr: > Howerver, I would at least like a metric that gives us the complement. So we know how many containers are using less than 100% of their given requests. This way we can account for all containers using VPA. Does this make sense?

Yes. That makes sense. I will add the complementary metrics.

amshuman-kr: > After weighing pros and cons, if it still doesn't suit, then this works too

The histogram approach would be the right long term solution.

amshuman-kr: @wyb1 I have added the complementary metrics.

amshuman-kr:
@wyb1 I have added the complementary metrics. Can you PTAL?

ggaurav10 (approved):

wyb1 (approved):
/lgtm

rfranzke (approved):

amshuman-kr:
❤️

Issue #1578, opened 2019-11-01
Optimize etcd encryption handling


rfranzke:
With https://github.com/gardener/gardener/pull/1066 we have introduced etcd encryption for shoot clusters. Today, in order to activate the encryption for new shoot clusters we require two reconciliation flows. In the first flow the encryption configuration is created but set to inactivate, and only in the second flow it's made active. However, it would be possible to combine that into one flow, at least for new shoots (for existing shoots we should keep it as it is).

Also, we should parallelize the secret rewriting in the shoot cluster in case the etcd encryption config has changed. This has already been done with https://github.com/gardener/gardener/pull/1391 but was reverted with 8420df1a0a02151bd14966057eae760d42d52773 due to instabilities. Let's fix them and reintroduce it.

rfranzke:
#2084 fixed the first part of this issue, so only the parallel re-writing of secrets is open now.
@plkokanov any plans for the implementation?

PR #2084, opened 2020-03-23
Persist etcd encryption in shootstate

plkokanov:
This PR makes it so that ETCD encryption data is generated and saved in the shoot state and can then be used to generate and  deploy an ETCD encryption secret.

Fixes first part of #1578
Part of #1631 

**Release note**:

```improvement operator
ETCD encryption data is persisted in the ShootState
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -98,7 +98,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 			Name: "Syncing shoot cluster information to seed",
 			Fn:   flow.TaskFn(botanist.SyncClusterResourceToSeed).RetryUntilTimeout(defaultInterval, defaultTimeout),
 		})
-		_ = g.Add(flow.Task{
+		ensureShootStateExists = g.Add(flow.Task{
 			Name: "Ensuring that ShootState exists",
 			Fn:   flow.TaskFn(botanist.EnsureShootStateExists).RetryUntilTimeout(defaultInterval, defaultTimeout),
 		})
@@ -195,22 +195,23 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 		generateEncryptionConfigurationMetaData = g.Add(flow.Task{
 			Name:         "Generating etcd encryption configuration",
 			Fn:           flow.TaskFn(botanist.GenerateEncryptionConfiguration).DoIf(enableEtcdEncryption),
-			Dependencies: flow.NewTaskIDs(deployNamespace),
+			Dependencies: flow.NewTaskIDs(deployNamespace, ensureShootStateExists),
 		})
 		persistETCDEncryptionConfiguration = g.Add(flow.Task{
 			Name:         "Persisting etcd encryption configuration in ShootState",
 			Fn:           flow.TaskFn(botanist.PersistEncryptionConfiguration).DoIf(enableEtcdEncryption),
-			Dependencies: flow.NewTaskIDs(deployNamespace, generateEncryptionConfigurationMetaData),
+			Dependencies: flow.NewTaskIDs(deployNamespace, ensureShootStateExists, generateEncryptionConfigurationMetaData),
 		})
+		// TODO: This can be removed in a future version once all etcd encryption configuration secrets have been cleaned up.
 		_ = g.Add(flow.Task{
 			Name:         "Removing old etcd encryption configuration secret from garden cluster",
 			Fn:           flow.TaskFn(botanist.RemoveOldETCDEncryptionSecretFromGardener),
-			Dependencies: flow.NewTaskIDs(generateEncryptionConfigurationMetaData),
+			Dependencies: flow.NewTaskIDs(persistETCDEncryptionConfiguration),
 		})
 		createOrUpdateETCDEncryptionConfiguration = g.Add(flow.Task{
 			Name:         "Applying etcd encryption configuration",
 			Fn:           flow.TaskFn(botanist.ApplyEncryptionConfiguration).DoIf(enableEtcdEncryption),
-			Dependencies: flow.NewTaskIDs(deployNamespace, generateEncryptionConfigurationMetaData, persistETCDEncryptionConfiguration),
+			Dependencies: flow.NewTaskIDs(deployNamespace, ensureShootStateExists, generateEncryptionConfigurationMetaData, persistETCDEncryptionConfiguration),
 		})
 		deployKubeAPIServer = g.Add(flow.Task{
 			Name:         "Deploying Kubernetes API server",
@@ -240,7 +241,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 		_ = g.Add(flow.Task{
 			Name:         "Rewriting Shoot secrets if EncryptionConfiguration has changed",
 			Fn:           flow.TaskFn(botanist.RewriteShootSecretsIfEncryptionConfigurationChanged).DoIf(enableEtcdEncryption && !o.Shoot.HibernationEnabled).RetryUntilTimeout(defaultInterval, 15*time.Minute),
-			Dependencies: flow.NewTaskIDs(initializeShootClients, createOrUpdateETCDEncryptionConfiguration),
+			Dependencies: flow.NewTaskIDs(initializeShootClients, ensureShootStateExists, createOrUpdateETCDEncryptionConfiguration),
 		})
 		_ = g.Add(flow.Task{
 			Name:         "Deploying Kubernetes scheduler",
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -192,15 +192,30 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 			Fn:           flow.TaskFn(botanist.WaitUntilControlPlaneReady),
 			Dependencies: flow.NewTaskIDs(deployControlPlane),
 		})
-		createOrUpdateEtcdEncryptionConfiguration = g.Add(flow.Task{
+		generateEncryptionConfigurationMetaData = g.Add(flow.Task{
+			Name:         "Generating etcd encryption configuration",
+			Fn:           flow.TaskFn(botanist.GenerateEncryptionConfiguration).DoIf(enableEtcdEncryption),
+			Dependencies: flow.NewTaskIDs(deployNamespace),
+		})
+		persistETCDEncryptionConfiguration = g.Add(flow.Task{
+			Name:         "Persisting etcd encryption configuration in ShootState",
+			Fn:           flow.TaskFn(botanist.PersistEncryptionConfiguration).DoIf(enableEtcdEncryption),
+			Dependencies: flow.NewTaskIDs(deployNamespace, generateEncryptionConfigurationMetaData),
+		})
+		_ = g.Add(flow.Task{
+			Name:         "Removing old etcd encryption configuration secret from garden cluster",
+			Fn:           flow.TaskFn(botanist.RemoveOldETCDEncryptionSecretFromGardener),
+			Dependencies: flow.NewTaskIDs(generateEncryptionConfigurationMetaData),
+		})
+		createOrUpdateETCDEncryptionConfiguration = g.Add(flow.Task{
 			Name:         "Applying etcd encryption configuration",
 			Fn:           flow.TaskFn(botanist.ApplyEncryptionConfiguration).DoIf(enableEtcdEncryption),
-			Dependencies: flow.NewTaskIDs(deployNamespace),
+			Dependencies: flow.NewTaskIDs(deployNamespace, generateEncryptionConfigurationMetaData, persistETCDEncryptionConfiguration),
 		})
 		deployKubeAPIServer = g.Add(flow.Task{
 			Name:         "Deploying Kubernetes API server",
 			Fn:           flow.TaskFn(botanist.DeployKubeAPIServer).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(deploySecrets, deployETCD, waitUntilEtcdReady, waitUntilKubeAPIServerServiceIsReady, waitUntilControlPlaneReady, createOrUpdateEtcdEncryptionConfiguration).InsertIf(!staticNodesCIDR, waitUntilInfrastructureReady),
+			Dependencies: flow.NewTaskIDs(deploySecrets, deployETCD, waitUntilEtcdReady, waitUntilKubeAPIServerServiceIsReady, waitUntilControlPlaneReady, createOrUpdateETCDEncryptionConfiguration).InsertIf(!staticNodesCIDR, waitUntilInfrastructureReady),
 		})
 		waitUntilKubeAPIServerIsReady = g.Add(flow.Task{
 			Name:         "Waiting until Kubernetes API server reports readiness",
@@ -225,7 +240,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 		_ = g.Add(flow.Task{
 			Name:         "Rewriting Shoot secrets if EncryptionConfiguration has changed",
 			Fn:           flow.TaskFn(botanist.RewriteShootSecretsIfEncryptionConfigurationChanged).DoIf(enableEtcdEncryption && !o.Shoot.HibernationEnabled).RetryUntilTimeout(defaultInterval, 15*time.Minute),
-			Dependencies: flow.NewTaskIDs(initializeShootClients, createOrUpdateEtcdEncryptionConfiguration),
+			Dependencies: flow.NewTaskIDs(initializeShootClients, createOrUpdateETCDEncryptionConfiguration),
 		})
 		_ = g.Add(flow.Task{
 			Name:         "Deploying Kubernetes scheduler",
--- a/pkg/operation/botanist/etcdencryption.go
+++ b/pkg/operation/botanist/etcdencryption.go
@@ -22,7 +22,7 @@ import (
 	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
 	gardencorev1alpha1helper "github.com/gardener/gardener/pkg/apis/core/v1alpha1/helper"
 	"github.com/gardener/gardener/pkg/operation/common"
-	encryptionconfiguration "github.com/gardener/gardener/pkg/operation/etcdencryption"
+	"github.com/gardener/gardener/pkg/operation/etcdencryption"
 	"github.com/gardener/gardener/pkg/utils"
 	"github.com/gardener/gardener/pkg/utils/infodata"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
@@ -47,43 +47,25 @@ func (b *Botanist) GenerateEncryptionConfiguration(ctx context.Context) error {
 		secret = nil
 	}
 
-	gardenerResourceList := gardencorev1alpha1helper.GardenerResourceDataList(b.ShootState.Spec.Gardener)
-	etcdEncryption, err := encryptionconfiguration.GetETCDEncryptionConfig(gardenerResourceList)
-	if err != nil {
-		return err
-	}
-
-	if etcdEncryption == nil {
-		etcdEncryption, err = generateETCDEncryption(secret)
+	if b.Shoot.ETCDEncryption == nil {
+		var err error
+		b.Shoot.ETCDEncryption, err = generateETCDEncryption(secret)
 		if err != nil {
 			return err
 		}
 	}
 
-	forcePlainTextSecrets := kutil.HasMetaDataAnnotation(secret, common.EtcdEncryptionForcePlaintextAnnotationName, "true")
-	etcdEncryption.SetForcePlainTextResources(forcePlainTextSecrets)
+	if secret != nil {
+		forcePlainTextSecrets := kutil.HasMetaDataAnnotation(secret, common.EtcdEncryptionForcePlaintextAnnotationName, "true")
+		b.Shoot.ETCDEncryption.SetForcePlainTextResources(forcePlainTextSecrets)
+	}
 
-	b.Shoot.ETCDEncryption = etcdEncryption
 	return nil
 }
 
 // PersistEncryptionConfiguration adds the encryption configuration to the ShootState.
 func (b *Botanist) PersistEncryptionConfiguration(ctx context.Context) error {
-	shootState := &gardencorev1alpha1.ShootState{ObjectMeta: kutil.ObjectMeta(b.Shoot.Info.Namespace, b.Shoot.Info.Name)}
-	if _, err := controllerutil.CreateOrUpdate(ctx, b.K8sGardenClient.Client(), shootState, func() error {
-		gardenerResourceList := gardencorev1alpha1helper.GardenerResourceDataList(b.ShootState.Spec.Gardener)
-		err := infodata.UpsertInfoData(&gardenerResourceList, common.ETCDSecretsEncryptionConfigDataName, b.Shoot.ETCDEncryption)
-		if err != nil {
-			return err
-		}
-		shootState.Spec.Gardener = gardenerResourceList
-		return nil
-	}); err != nil {
-		return err
-	}
-
-	b.ShootState = shootState
-	return nil
+	return b.persistEncryptionConfigInShootState(ctx)
 }
 
 // ApplyEncryptionConfiguration creates or updates a secret on the Seed
@@ -94,20 +76,16 @@ func (b *Botanist) ApplyEncryptionConfiguration(ctx context.Context) error {
 		secret = &corev1.Secret{ObjectMeta: kutil.ObjectMeta(b.Shoot.SeedNamespace, common.EtcdEncryptionSecretName)}
 		conf   *apiserverconfigv1.EncryptionConfiguration
 	)
-	encryptionData, err := encryptionconfiguration.GetETCDEncryptionConfig(b.ShootState.Spec.Gardener)
-	if err != nil {
-		return err
-	}
-	if encryptionData == nil {
+	if b.Shoot.ETCDEncryption == nil {
 		return errors.New("Could not find etcd encryption configuration in ShootState")
 	}
 
-	conf = encryptionconfiguration.NewEncryptionConfiguration(encryptionData)
-	_, err = controllerutil.CreateOrUpdate(ctx, b.K8sSeedClient.Client(), secret, func() error {
-		if encryptionData.ForcePlainTextResources {
+	conf = etcdencryption.NewEncryptionConfiguration(b.Shoot.ETCDEncryption)
+	_, err := controllerutil.CreateOrUpdate(ctx, b.K8sSeedClient.Client(), secret, func() error {
+		if b.Shoot.ETCDEncryption.ForcePlainTextResources {
 			kutil.SetMetaDataAnnotation(secret, common.EtcdEncryptionForcePlaintextAnnotationName, "true")
 		}
-		return encryptionconfiguration.UpdateSecret(secret, conf)
+		return etcdencryption.UpdateSecret(secret, conf)
 	})
 	if err != nil {
 		return err
@@ -140,7 +118,7 @@ func (b *Botanist) RemoveOldETCDEncryptionSecretFromGardener(ctx context.Context
 }
 
 func confChecksum(conf *apiserverconfigv1.EncryptionConfiguration) (string, error) {
-	data, err := encryptionconfiguration.Write(conf)
+	data, err := etcdencryption.Write(conf)
 	if err != nil {
 		return "", err
 	}
@@ -152,13 +130,7 @@ func confChecksum(conf *apiserverconfigv1.EncryptionConfiguration) (string, erro
 // encryption configuration changed. Rewriting here means that a patch request is sent that forces
 // the etcd to encrypt them with the new configuration.
 func (b *Botanist) RewriteShootSecretsIfEncryptionConfigurationChanged(ctx context.Context) error {
-	gardenerResourceList := gardencorev1alpha1helper.GardenerResourceDataList(b.ShootState.Spec.Gardener)
-	etcdEncryption, err := encryptionconfiguration.GetETCDEncryptionConfig(gardenerResourceList)
-	if err != nil {
-		return err
-	}
-
-	if !etcdEncryption.RewriteResources {
+	if !b.Shoot.ETCDEncryption.RewriteResources {
 		return nil
 	}
 
@@ -194,21 +166,8 @@ func (b *Botanist) RewriteShootSecretsIfEncryptionConfigurationChanged(ctx conte
 	}
 	b.Logger.Info("Successfully removed all added secret labels in the shoot after etcd encryption config changed")
 
-	etcdEncryption.RewriteResources = false
-	if err = infodata.UpsertInfoData(&gardenerResourceList, common.ETCDSecretsEncryptionConfigDataName, etcdEncryption); err != nil {
-		return err
-	}
-
-	shootState := &gardencorev1alpha1.ShootState{ObjectMeta: kutil.ObjectMeta(b.Shoot.Info.Namespace, b.Shoot.Info.Name)}
-	if _, err = controllerutil.CreateOrUpdate(ctx, b.K8sGardenClient.Client(), shootState, func() error {
-		shootState.Spec.Gardener = gardenerResourceList
-		return nil
-	}); err != nil {
-		return err
-	}
-
-	b.ShootState = shootState
-	return nil
+	b.Shoot.ETCDEncryption.RewriteResources = false
+	return b.persistEncryptionConfigInShootState(ctx)
 }
 
 func (b *Botanist) updateShootLabelsForEtcdEncryption(ctx context.Context, labelRequirement *labels.Requirement, mutateLabelsFunc func(m metav1.Object)) []error {
@@ -230,8 +189,25 @@ func (b *Botanist) updateShootLabelsForEtcdEncryption(ctx context.Context, label
 	return errorList
 }
 
-func generateETCDEncryption(secret *corev1.Secret) (*encryptionconfiguration.ETCDEncryptionConfig, error) {
-	encryptionConfig := &encryptionconfiguration.ETCDEncryptionConfig{}
+func (b *Botanist) persistEncryptionConfigInShootState(ctx context.Context) error {
+	shootState := &gardencorev1alpha1.ShootState{ObjectMeta: kutil.ObjectMeta(b.Shoot.Info.Namespace, b.Shoot.Info.Name)}
+	if _, err := controllerutil.CreateOrUpdate(ctx, b.K8sGardenClient.Client(), shootState, func() error {
+		gardenerResourceList := gardencorev1alpha1helper.GardenerResourceDataList(b.ShootState.Spec.Gardener)
+		if err := infodata.UpsertInfoData(&gardenerResourceList, common.ETCDEncryptionConfigDataName, b.Shoot.ETCDEncryption); err != nil {
+			return err
+		}
+		shootState.Spec.Gardener = gardenerResourceList
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	b.ShootState = shootState
+	return nil
+}
+
+func generateETCDEncryption(secret *corev1.Secret) (*etcdencryption.EncryptionConfig, error) {
+	encryptionConfig := &etcdencryption.EncryptionConfig{}
 	if secret != nil {
 		if err := encryptionConfig.AddEncryptionKeyFromSecret(secret); err != nil {
 			return nil, err
--- a/pkg/operation/botanist/etcdencryption.go
+++ b/pkg/operation/botanist/etcdencryption.go
@@ -16,14 +16,15 @@ package botanist
 
 import (
 	"context"
-	"crypto/rand"
+	"errors"
 	"fmt"
-	"time"
 
-	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
+	gardencorev1alpha1helper "github.com/gardener/gardener/pkg/apis/core/v1alpha1/helper"
 	"github.com/gardener/gardener/pkg/operation/common"
 	encryptionconfiguration "github.com/gardener/gardener/pkg/operation/etcdencryption"
 	"github.com/gardener/gardener/pkg/utils"
+	"github.com/gardener/gardener/pkg/utils/infodata"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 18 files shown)


pkg/operation/botanist/etcdencryption.go
```diff
@@ -36,85 +38,197 @@ import (
...
 )
 
-// ApplyEncryptionConfiguration creates or updates a secret on the Seed
-// which contains the encryption configuration that is necessary to encrypt the
-// Kubernetes secrets in etcd.
-//
-// To mitigate data loss to a certain degree, the secret is also synced to the Garden cluster.
-func (b *Botanist) ApplyEncryptionConfiguration(ctx context.Context) error {
-	if err := b.syncEncryptionConfigurationFromGardenCluster(ctx); err != nil {
+// SyncEncryptionConfiguration syncs an already generated etcd encryptio	n configuration secret to the shootstate
```
danielfoehrKn:
```suggestion
// SyncEncryptionConfiguration syncs an already generated etcd encryption configuration secret to the shootstate
```

pkg/operation/botanist/etcdencryption.go
```diff
@@ -36,85 +38,197 @@ import (
...
+// TODO: This step can be removed in the future, once all shoots have started using the shoot state to persist resources.
+func (b *Botanist) SyncEncryptionConfiguration(ctx context.Context) error {
+	secret := &corev1.Secret{ObjectMeta: kutil.ObjectMeta(b.Shoot.SeedNamespace, common.EtcdEncryptionSecretName)}
+	err := b.K8sSeedClient.Client().Get(ctx, kutil.KeyFromObject(secret), secret)
```
danielfoehrKn: The secret is always read even though it is not required if encryptionKey != nil && secretsEncryptionState != nil

pkg/operation/botanist/etcdencryption.go
```diff
@@ -36,85 +38,197 @@ import (
...
+	_, err = controllerutil.CreateOrUpdate(ctx, b.K8sGardenClient.Client(), shootState, func() error {
+		gardenerResourceDataList := gardencorev1alpha1helper.GardenerResourceDataList(shootState.Spec.Gardener)
+		forcePlaintextSecrets := kutil.HasMetaDataAnnotation(secret, common.EtcdEncryptionForcePlaintextAnnotationName, "true")
+		encryptionStateInfoData, err := infodata.GetInfoData(gardenerResourceDataList, common.ETCDSecretsEncryptionStateName)
+		if err != nil {
+			return err
+		}
+		etcdEncryptionState, ok := encryptionStateInfoData.(*encryptionconfiguration.ETCDEncryptionState)
+		if !ok {
+			return fmt.Errorf("Could not convert GardenerResourceData entry %s to encryptionconfiguration.ETCDEncryptionState", common.ETCDSecretsEncryptionStateName)
```
danielfoehrKn:
```suggestion
			return fmt.Errorf("could not convert GardenerResourceData entry %s to encryptionconfiguration.ETCDEncryptionState", common.ETCDSecretsEncryptionStateName)
```

pkg/operation/botanist/etcdencryption.go
```diff
@@ -36,85 +38,197 @@ import (
...
+		}
+		etcdEncryptionState.ForcePlainTextResources = forcePlaintextSecrets
```
danielfoehrKn: Can you help me with the following: forcePlaintextSecrets in the ShootState is only set 
- during the Secrets [rewrite if the checksum changed ](https://github.com/gardener/gardener/blob/753f00bbd5a0745361022aa698faf701c63c3971/pkg/operation/botanist/etcdencryption.go#L279) -> always set to true
- in the migration code when there is no etcdEncryptionState (the marked code) -> calculate from the etcd secret

Who sets the initial checksum on the etcd-encryption-secret & who deploys it initially?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -190,10 +190,25 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 			Fn:           flow.TaskFn(botanist.WaitUntilControlPlaneReady),
 			Dependencies: flow.NewTaskIDs(deployControlPlane),
 		})
+		syncETCDEncryptionConfiguration = g.Add(flow.Task{
+			Name:         "Syncing etcd encryption configuration to ShootState",
```
danielfoehrKn: In each function here we are retrieving the ShootState plus sometimes the encryption secret from the seed. Then extract the etcd--enrcyption-key and etcd-encryption-state from the ShootState. 
Then either generate something (key) or take information from the seed secret to update the ShootState.

I do not know if the controller-runtime underneath is caching these calls, but that seems to me like too many of the same request being made here. 
I would try to unify the functions
- Syncing etcd encryption configuration to ShootState
- Syncing force plain text secrets config
- Generating etcd encryption secret
THen only request the seed secret once and update the ShootState once.

pkg/operation/etcdencryption/encryptionconfiguration.go
```diff
@@ -233,3 +217,19 @@ func UpdateSecret(secret *corev1.Secret, conf *apiserverconfigv1.EncryptionConfi
 	secret.Data[common.EtcdEncryptionSecretFileName] = confData
 	return nil
 }
+
+// GetSecretKeyForResources returns the AESCBC key whcih is used to encrypt the resource.
```
danielfoehrKn:
```suggestion
// GetSecretKeyForResources returns the AESCBC key which is used to encrypt the resource.
```

danielfoehrKn (changes requested):
Thanks! I think it would also make sense to point out in the release notes that
- instead of persisting the encryption data in a copy-secret in the Garden, it is now persistet & migrated  into the ShootState


Also: [Method](https://github.com/gardener/gardener/blob/753f00bbd5a0745361022aa698faf701c63c3971/pkg/operation/botanist/etcdencryption.go#L360) is unused

pkg/openapi/openapi_generated.go
```diff
@@ -1375,7 +1375,7 @@ func schema_pkg_apis_core_v1alpha1_ClusterAutoscaler(ref common.ReferenceCallbac
 				Properties: map[string]spec.Schema{
 					"scaleDownDelayAfterAdd": {
 						SchemaProps: spec.SchemaProps{
-							Description: "ScaleDownDelayAfterAdd defines how long after scale up that scale down evaluation resumes (default: 10 mins).",
+							Description: "ScaleDownDelayAfterAdd defines how long after scale up that scale down evaluation resumes (default: 1 hour).",
```
rfranzke: Why do these values change?

pkg/utils/infodata/infodata.go
```diff
@@ -0,0 +1,97 @@
...
+func Unmarshal(entry *gardencorev1alpha1.GardenerResourceData) (InfoData, error) {
+	unmarshaller := GetUnmarshaller(TypeVersion(entry.Type))
+	if unmarshaller == nil {
+		return nil, fmt.Errorf("unknown info data type %q", entry.Type)
+	}
+	data, err := unmarshaller(entry.Data.Raw)
+	if err != nil {
+		return nil, fmt.Errorf("cannot unmarshal data set %q of type %q: %s", entry.Name, entry.Type, err)
+	}
+	return data, err
```
rfranzke:
```suggestion
	return data, nil
```

pkg/utils/infodata/infodata.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// GetInfoData retrieves the go representation of an object from the GardenerResourceDataList
+func GetInfoData(resourceDataList gardencorev1alpha1helper.GardenerResourceDataList, name string) (InfoData, error) {
+	resourceData := resourceDataList.Get(name)
+	if resourceData == nil {
+		return nil, nil
```
rfranzke, line 76: Shouldn't this return an error because the provided name was not found?

pkg/utils/infodata/infodata.go
```diff
@@ -0,0 +1,97 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *
+ */
+
+package infodata
```
rfranzke, line 19: Unit tests are missing for the whole package

pkg/operation/etcdencryption/etcdencryptioninfodata.go
```diff
@@ -0,0 +1,67 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *
+ */
+
+package encryptionconfiguration
```
rfranzke: unit tests are missing here

pkg/operation/etcdencryption/etcdencryptiondata.go
```diff
@@ -0,0 +1,76 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *
+ */
+
+package encryptionconfiguration
```
rfranzke: unit tests are missing

pkg/operation/etcdencryption/encryptionconfiguration_test.go
```diff
@@ -234,4 +245,67 @@ resources:
 			Expect(secret.Data[common.EtcdEncryptionSecretFileName]).To(Equal(yamlData))
 		})
 	})
+
+	Describe("ETCD Encryption InfoData", func() {
```
rfranzke: Can you move the tests into a file that starts with the same filename like the one defining the actual code? I think these tests here, for example, would belong in `etcdencryptiondata_test.go`.

pkg/operation/etcdencryption/encryptionconfiguration.go
```diff
@@ -48,6 +48,33 @@ func init() {
...
+	} else {
+		providerConfigs = append(providerConfigs, identityProviderConfiguration, aescbcProviderConfiguration)
+	}
+
+	resourceConfig := apiserverconfigv1.ResourceConfiguration{
+		Resources: []string{common.EtcdEncryptionEncryptedResourceSecrets},
+		Providers: providerConfigs,
+	}
+
+	conf := &apiserverconfigv1.EncryptionConfiguration{
```
rfranzke:
```suggestion
	return &apiserverconfigv1.EncryptionConfiguration{
```

pkg/operation/botanist/etcdencryption.go
plkokanov: You're right, initially I had it the other way around, but swapped to fetch the secret first and to only execute the step if the secret is available or not.
I'll switch it again -> seems better tbh

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
plkokanov: I'll check how I can optimise these, generally I'd like to keep `Syncing etcd encryption configuration to ShootState` as a separate step as it will be really simple to remove in the near future (we won't need it once all shoots have updated the encryption metadata to their shootstates.

pkg/operation/botanist/etcdencryption.go
plkokanov: I assume that instead of `forcePlaintextSecrets` you mean `isEncryptionActive` as that is what you linked.
> in the migration code when there is no etcdEncryptionState (the marked code) -> calculate from the etcd secret

This will be removed in the future.

Here's the way that the ETCD encryption secret was generated previously (before the shootstate)
On first reconciliation (creation)
1. Generate a passive encryption configuration and generate a secret object from it.
2. Calculate the checksum of the encryption configuration and only save it in memory
3. Rewrite the shoot secrets and [add the checksum annotation to the secret object](https://github.com/gardener/gardener/blob/753f00bbd5a0745361022aa698faf701c63c3971/pkg/operation/boanist/etcdencryption.go#L336).

On second reconciliation:
1. See that the secret object has a checksum annotation
2. Change the passive encryption configuration to active, update the secret
3. Recalculate the checksum and save it in memory
4. Rewrite shoot secrets with enabled encryption and [update the checksum as it has changed](https://github.com/gardener/gardener/blob/753f00bbd5a0745361022aa698faf701c63c3971/pkg/operation/botanist/etcdencryption.go#L336).

I've tried to keep the same workflow as I'm not sure if there will be any side effects if I change it slightly.
But indeed, I should not update `isEncryptionActive` every time. I used to update the checksum to the etcd encryption secret in the garden cluster every time but this isn't needed here. Thanks for pointing it out :).

danielfoehrKn: Thanks for the detailed explanation!

pkg/operation/botanist/etcdencryption.go
rfranzke:
```suggestion
	if err := b.K8sSeedClient.Client().Get(ctx, kutil.KeyFromObject(secret), secret); client.IgnoreNotFound(err) != nil {
		return err
	}
```

pkg/operation/botanist/etcdencryption.go
```diff
@@ -36,85 +38,197 @@ import (
...
+			name, key, err := encryptionconfiguration.GetSecretKeyForResources(conf, common.EtcdEncryptionEncryptedResourceSecrets)
+			if err != nil {
+				return err
+			}
+			encryptionKey = &encryptionconfiguration.ETCDEncryptionKey{
+				Key:  key,
+				Name: name,
+			}
+			infodata.UpsertInfoData(&gardenerResourceDataList, common.ETCDSecretsEncryptionKeyDataName, encryptionKey)
+		} else {
```
rfranzke: This else branch can never be reached, right?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
rfranzke: 1. Why is the sync performed before the generation of the etcd secret?
1. Why are there so many different tasks in the flow? Can't they combined into a general "sync data" step? This would also work well with @danielfoehrKn's comment.

pkg/operation/botanist/etcdencryption.go
```diff
@@ -36,85 +38,197 @@ import (
...
 }
 
-func (b *Botanist) syncEncryptionConfigurationFromGardenCluster(ctx context.Context) error {
-	gardenSecret := &corev1.Secret{}
-	if err := b.K8sGardenClient.Client().Get(ctx, common.GardenEtcdEncryptionSecretKey(b.Shoot.Info.Namespace, b.Shoot.Info.Name), gardenSecret); err != nil {
+// SyncForcePlaintextSecrets syncs the annotation which determines whether secret resources should be kept unencrypted in the
+// Shoot's ETCD from the ETCD encryption configuration secret.
+func (b *Botanist) SyncForcePlaintextSecrets(ctx context.Context) error {
+	secret := &corev1.Secret{ObjectMeta: kutil.ObjectMeta(b.Shoot.SeedNamespace, common.EtcdEncryptionSecretName)}
+	err := b.K8sSeedClient.Client().Get(ctx, kutil.KeyFromObject(secret), secret)
```
rfranzke:
```suggestion
	if err := b.K8sSeedClient.Client().Get(ctx, kutil.KeyFromObject(secret), secret); client.IgnoreNotFound(err) != nil {
		return err
	}
```

pkg/operation/botanist/etcdencryption.go
```diff
@@ -36,85 +38,197 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	b.ShootState = shootState
+	return nil
+}
+
+// GenerateEncryptionConfiguration generates data for encryption configuration and saves it in the shootstate
+func (b *Botanist) GenerateEncryptionConfiguration(ctx context.Context) error {
```
rfranzke: I would not bundle the generation and the store-in-state part into a single function. Instead, one function should generate the stuff, another one takes it and puts it into the state. Then you can also get rid of all the duplication in this function (most stuff is already in the first function).

pkg/operation/botanist/etcdencryption.go
```diff
@@ -123,25 +237,7 @@ func (b *Botanist) createOrUpdateEncryptionConfiguration(ctx context.Context) (*
...
 		return encryptionconfiguration.UpdateSecret(secret, conf)
 	})
-	if err != nil {
-		return nil, err
-	}
 
-	return secret, err
-}
-
-func (b *Botanist) syncEncryptionConfigurationToGarden(ctx context.Context, controlPlaneSecret *corev1.Secret) error {
```
rfranzke, line 133: Good point, so we can clean up the `<shootname>.etcd-encryption` secrets in the project namespaces.

pkg/operation/botanist/etcdencryption.go
```diff
@@ -164,16 +260,36 @@ func (b *Botanist) RewriteShootSecretsIfEncryptionConfigurationChanged(ctx conte
 		return b.CheckSums[common.EtcdEncryptionSecretName]
 	}()
 
-	if err := b.rewriteShootSecretsIfChecksumChanged(ctx, checksum); err != nil {
```
rfranzke, line 167: why was this changed?

pkg/operation/botanist/etcdencryption.go
```diff
@@ -164,16 +260,36 @@ func (b *Botanist) RewriteShootSecretsIfEncryptionConfigurationChanged(ctx conte
...
+	err := b.rewriteShootSecretsIfChecksumChanged(ctx, checksum)
+	if err != nil {
 		return err
 	}
 
-	gardenSecret := &corev1.Secret{ObjectMeta: kutil.ObjectMetaFromKey(common.GardenEtcdEncryptionSecretKey(b.Shoot.Info.Namespace, b.Shoot.Info.Name))}
-	_, err := controllerutil.CreateOrUpdate(ctx, b.K8sGardenClient.Client(), gardenSecret, func() error {
-		kutil.SetMetaDataAnnotation(gardenSecret, common.EtcdEncryptionChecksumAnnotationName, checksum)
+	shootState := &gardencorev1alpha1.ShootState{ObjectMeta: kutil.ObjectMeta(b.Shoot.Info.Namespace, b.Shoot.Info.Name)}
+	_, err = controllerutil.CreateOrUpdate(ctx, b.K8sGardenClient.Client(), shootState, func() error {
```
rfranzke: Again, a lot of duplicate code. Can't we build a helper function that takes the relevant information about the etcd encryption data and stores it in the shootstate?

pkg/operation/botanist/etcdencryption.go
```diff
@@ -36,85 +38,197 @@ import (
...
 		}
 
-		// When firstly created, the encryption configuration secret does not have a checksum annotation yet. This annotation will
-		// only be added after all shoot secrets have been rewritten. In order to allow a smooth transition from un-encrypted to encrypted
-		// etcd data we first make the configuration inactive, i.e., put the `identity` provider as first list in the entry. In the next
-		// reconciliation we will detect that the annotation is set and then we can make it active, i.e., moving the `identity` provider
-		// to the second list item. The reason for this is that new API servers would otherwise start with an active configuration and would
-		// try to decrypt secrets in the etcd store (which would fail because they are not yet encrypted). Be aware that this will only happen
-		// once during the first introduction of the encryption configuration.
-		firstCreationOfEncryptionConfiguration := !metav1.HasAnnotation(secret.ObjectMeta, common.EtcdEncryptionChecksumAnnotationName)
```
rfranzke, line 101: @plkokanov can you elaborate on this change? Where is this logic gone?
Actually, we discussed with @timuthy yesterday that we want to get rid of this anyways (we don't need to make the encryption inactive if the cluster is new - we can directly make it active). If I understand your changes correctly the logic is gone now, but what was your motivation to remove it?

rfranzke (changes requested):

pkg/openapi/openapi_generated.go
plkokanov: code generation changed them for some reason. Need to check why it happened and revert it.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
plkokanov: The sync is only intended to fetch already existing etcd encryption configuration from an existing secret so that it doesn't get regenerated in the Generation step. This will be removed later on.

pkg/operation/botanist/etcdencryption.go
plkokanov: They would be removed automatically as they have shoot owner references.
However, I could add logic to clean up our shooted seeds and other shoots which won't be deleted for a long time.

pkg/operation/botanist/etcdencryption.go
plkokanov: It's not gone, just handled differently: 
1. When we generate the key and state for the encryption config metadata, the state has `IsEncryptionActive` set to false.
2. When the encryption config for the secret is generated/deployed it checks the `IsEncryptionActive` and  creates it either in passive or active mode (first reconciliation will be passive)
3. Once the shoot secrets are rewritten `IsEncryptionActive` is changed to true
4. Then on next reconciliation the encryption config generation step will change the configuration to active.

To facilitate the new logic I can just remove the `isEncryptionActive` flag and only use `ForcePlainTextResources`

pkg/operation/botanist/etcdencryption.go
plkokanov: Shouldn't have changed, something I missed during a rebase I guess.

pkg/operation/botanist/etcdencryption.go
rfranzke: I think it makes sense as you are anyways touching the code. @timuthy  WDYT?

pkg/operation/botanist/etcdencryption.go
plkokanov: On second reconciliations they will no longer be nil, but come to think of it I don't see it as completely necessary to log this on every reconciliation.

pkg/utils/infodata/infodata.go
plkokanov: Depending on the calling code if we get `nil` we can make a decision whether to generate new data then update the shootstate or do nothing. Otherwise we have to check the error type.

pkg/operation/botanist/etcdencryption.go
timuthy: Yes, good suggestion. I would appreciate if you can do it in this PR @plkokanov.

plkokanov:
Updated the PR, it's ready for another look.

ialidzhikov:
Can you rebase to latest master to make ci/cd happy?

pkg/operation/botanist/etcdencryption.go
```diff
@@ -36,113 +35,106 @@ import (
...
+	secret := &corev1.Secret{}
+	if err := b.K8sSeedClient.Client().Get(ctx, kutil.Key(b.Shoot.SeedNamespace, common.EtcdEncryptionSecretName), secret); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+		secret = nil
 	}
-	secret, err := b.createOrUpdateEncryptionConfiguration(ctx)
+
+	gardenerResourceList := gardencorev1alpha1helper.GardenerResourceDataList(b.ShootState.Spec.Gardener)
```
rfranzke: This function is doing too much. From its name it suggests that it generates the configuration. But it also does update it in the `ShootState`. Either improve the function's name or split it into two functions, one that generates (or takes the existing configuration), and another function that updates it in the `ShootState`.

pkg/operation/etcdencryption/encryptionconfiguration.go
```diff
@@ -233,3 +173,18 @@ func UpdateSecret(secret *corev1.Secret, conf *apiserverconfigv1.EncryptionConfi
 	secret.Data[common.EtcdEncryptionSecretFileName] = confData
 	return nil
 }
+
+// GetSecretKeyForResources returns the AESCBC key which is used to encrypt the resource.
```
rfranzke:
```suggestion
// GetSecretKeyForResources returns the AESCBC key name and AESCBC key secret which is used to encrypt the resource. If the AESCBC is not found then it returns empty strings.
```

pkg/operation/etcdencryption/etcdencryptioninfodata.go
```diff
@@ -0,0 +1,176 @@
...
+	etcdKey := &ETCDEncryptionKey{
+		Key:  key.Secret,
+		Name: key.Name,
+	}
+	return etcdKey, nil
+}
+
+// GenerateOrSyncEncryptionStateFromSecret syncs the ETCDEncryptionState data from the provided secret if it is not nil and the gardenerResourceDataList.
+// Otherwise, it generates a new ETCDEncryptionState.
+func GenerateOrSyncEncryptionStateFromSecret(gardenerResourceList gardencorev1alpha1helper.GardenerResourceDataList, secret *corev1.Secret) (*ETCDEncryptionState, error) {
```
rfranzke: I found it confusing that this is generated from secret. Actually, you only use the secret to detect whether "force plain text" is enforced. Can you pass this information to the function and rename it?

```suggestion
func GenerateOrSyncEncryptionState(gardenerResourceList gardencorev1alpha1helper.GardenerResourceDataList, forcePlaintextSecrets bool) (*ETCDEncryptionState, error) {
```

pkg/operation/botanist/etcdencryption.go
```diff
@@ -36,113 +35,106 @@ import (
...
+// RemoveOldETCDEncryptionSecretFromGardener removes the etcd encryption configuration secret from the Shoot's namespace in the garden cluster as it is no longer necessary.
+// This step can be removed in the future after all secrets have been cleaned up.
+func (b *Botanist) RemoveOldETCDEncryptionSecretFromGardener(ctx context.Context) error {
+	etcdSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      common.GardenEtcdEncryptionSecretName(b.Shoot.Info.Name),
+			Namespace: b.Shoot.Info.Namespace,
+		},
+	}
+	err := b.K8sGardenClient.Client().Delete(ctx, etcdSecret)
```
rfranzke:
```suggestion
	return client.IgnoreNotFound(b.K8sGardenClient.Client().Delete(ctx, etcdSecret))
```

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -192,15 +192,24 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 			Fn:           flow.TaskFn(botanist.WaitUntilControlPlaneReady),
 			Dependencies: flow.NewTaskIDs(deployControlPlane),
 		})
-		createOrUpdateEtcdEncryptionConfiguration = g.Add(flow.Task{
+		generateEncryptionConfigurationMetaData = g.Add(flow.Task{
+			Name:         "Generating etcd encryption configuration",
+			Fn:           flow.TaskFn(botanist.GenerateEncryptionConfiguration).DoIf(enableEtcdEncryption),
+			Dependencies: flow.NewTaskIDs(deployNamespace),
+		})
+		_ = g.Add(flow.Task{
```
rfranzke, line 206: I think it makes sense to make this function depending on the task that is syncing the encryption configuration to the `ShootState` - in this case `generateEncryptionConfigurationMetaData`

pkg/operation/botanist/etcdencryption.go
```diff
@@ -36,113 +35,106 @@ import (
...
+	if err != nil {
+		return err
+	}
 
-	_, err := controllerutil.CreateOrUpdate(ctx, b.K8sSeedClient.Client(), secret, func() error {
-		var err error
-		conf, err = encryptionconfiguration.ReadSecret(secret)
+	conf = encryptionconfiguration.NewEncryptionConfiguration(etcdKeyData.Name, etcdKeyData.Key, etcdStateData.ForcePlainTextResources)
+	_, err = controllerutil.CreateOrUpdate(ctx, b.K8sSeedClient.Client(), secret, func() error {
 		if err != nil {
```
rfranzke: I don't get this line - why is the first thing that you do in this transformation function an error check?

pkg/operation/etcdencryption/etcdencryptiondata.go
```diff
@@ -0,0 +1,76 @@
...
+// ETCDKeyData holds the key and its name used to encrypt resources in ETCD
+type ETCDKeyData struct {
+	Key  string `json:"key"`
+	Name string `json:"name"`
+}
+
+// ETCDEncryptionStateData holds information whether a key is active or not and whether resources should be forcefully kept in plain text
+type ETCDEncryptionStateData struct {
+	ForcePlainTextResources bool `json:"forcePlainTextResources"`
+	RewriteResources        bool `json:"rewriteResources"`
```
rfranzke: Why is this something that needs to be stored in the state? Shouldn't it be found out dynamically during reconciliation by comparing the checksums?

pkg/operation/etcdencryption/etcdencryptiondata.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/gardener/gardener/pkg/utils/infodata"
+)
+
+// ETCDKeyDataType is the type used to denote an ETCDKeyData structure in the ShootState
+const ETCDKeyDataType = infodata.TypeVersion("ETCDKeyData")
```
rfranzke:
```suggestion
const ETCDKeyDataType = infodata.TypeVersion("etcdKeyData")
```

pkg/operation/etcdencryption/etcdencryptiondata.go
```diff
@@ -0,0 +1,76 @@
...
+
+// ETCDEncryptionStateDataType is the type used to denote an ETCDEncryptionStateData structure in the ShootState
+const ETCDEncryptionStateDataType = infodata.TypeVersion("ETCDEncryptionStateData")
```
rfranzke:
```suggestion
const ETCDEncryptionStateDataType = infodata.TypeVersion("etcdEncryptionStateData")
```

pkg/operation/etcdencryption/etcdencryptiondata.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// ETCDKeyData holds the key and its name used to encrypt resources in ETCD
+type ETCDKeyData struct {
+	Key  string `json:"key"`
+	Name string `json:"name"`
+}
+
+// ETCDEncryptionStateData holds information whether a key is active or not and whether resources should be forcefully kept in plain text
+type ETCDEncryptionStateData struct {
```
rfranzke: Why not merging this into `ETCDKeyData`?

pkg/operation/etcdencryption/etcdencryptioninfodata.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+// ETCDEncryptionKey holds the the key and its name used to encrypt resources in ETCD.
+type ETCDEncryptionKey struct {
+	Key  string
+	Name string
+}
+
+// ETCDEncryptionState holds information whether a key is active or not and whether resources should be forcefully kept in plain text
+type ETCDEncryptionState struct {
```
rfranzke: Both structs here already exist. Why defining them again?

pkg/operation/etcdencryption/etcdencryptioninfodata.go
```diff
@@ -0,0 +1,176 @@
...
+	return &ETCDEncryptionState{forcePlainTextResources, rewriteResources}, nil
+}
+
+// GetETCDEncryptionState retrieves the ETCDEncryptionState from the gardenerResourceDataList.
+func GetETCDEncryptionState(gardenerResourceDataList gardencorev1alpha1helper.GardenerResourceDataList) (*ETCDEncryptionState, error) {
+	infoData, err := infodata.GetInfoData(gardenerResourceDataList, common.ETCDSecretsEncryptionStateName)
+	if err != nil {
+		return nil, err
+	}
+	if infoData != nil {
```
rfranzke: early exit

```suggestion
	if infoData == nil {
		return nil, nil
	}
```

pkg/operation/etcdencryption/etcdencryptioninfodata.go
```diff
@@ -0,0 +1,176 @@
...
+	return nil, nil
+}
+
+// GetETCDEncryptionKey retrieves the ETCDEncryptionKey from the gardenerResourceDataList.
+func GetETCDEncryptionKey(gardenerResourceDataList gardencorev1alpha1helper.GardenerResourceDataList) (*ETCDEncryptionKey, error) {
+	infoData, err := infodata.GetInfoData(gardenerResourceDataList, common.ETCDSecretsEncryptionKeyDataName)
+	if err != nil {
+		return nil, err
+	}
+	if infoData != nil {
```
rfranzke: same here, early exit

pkg/operation/etcdencryption/etcdencryptioninfodata.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// UpsertETCDDataInShootState upserts the provided ETCDEncryptionKey and ETCDEncryptionState if they are not nil into the gardenerResourceList.
+func UpsertETCDDataInShootState(gardenerResourceList *gardencorev1alpha1helper.GardenerResourceDataList, etcdEncryptionKey *ETCDEncryptionKey, etcdEncryptionState *ETCDEncryptionState) error {
+	if etcdEncryptionKey != nil {
+		if err := infodata.UpsertInfoData(gardenerResourceList, common.ETCDSecretsEncryptionKeyDataName, etcdEncryptionKey); err != nil {
+			return err
+		}
+	}
+	if etcdEncryptionState != nil {
```
rfranzke: same here, early exit

pkg/utils/infodata/infodata.go
rfranzke: ping

rfranzke (changes requested):

pkg/operation/etcdencryption/etcdencryptioninfodata.go
plkokanov: The idea was to have one struct act as the external (which will be saved in a json format and we could have multiple versions) the other one is the internal, to be used for the generation.

rfranzke: I don't get why we need an external and an internal version. Can you explain?

pkg/operation/etcdencryption/etcdencryptiondata.go
plkokanov: If the encryption configuration was regenerated, but the migration is triggered before the step which rewrites the secret, there would be no way on the destination seed to compare the (now newly generated) checksum as we won't have access to the secret on the source seed.

pkg/operation/etcdencryption/etcdencryptiondata.go
plkokanov: Was wondering about that, my thoughts were that you could have multiple key entries (also for different providers), but one 
 or different number of StateData.

pkg/operation/etcdencryption/etcdencryptiondata.go
rfranzke: OK, makes sense. Thanks!

pkg/operation/etcdencryption/etcdencryptiondata.go
rfranzke: Hm, would that be a use-case when the key is rotated? Do you mean this?

pkg/operation/etcdencryption/etcdencryptioninfodata.go
plkokanov: That was @mandelsoft's proposal (at least for the certificates). But I guess he wasn't really sure if we need to support multiple external versions as well.

pkg/operation/botanist/etcdencryption.go
plkokanov: If I split it then I would have to keep the generated data in the Botanist structure, or did you mean something else. TBH i'd favour splitting.

pkg/operation/etcdencryption/etcdencryptioninfodata.go
rfranzke: Hm, this doesn't help me now, I can't follow/have too less information.

pkg/operation/botanist/etcdencryption.go
plkokanov: left it by accident

pkg/operation/botanist/etcdencryption.go
rfranzke: Yes, maybe store it in the `pkg/operation/shoot.Shoot` struct for now?

pkg/operation/etcdencryption/etcdencryptiondata.go
plkokanov: Yes, however now that I think a bit more on that case, we might need an additional way to say which key is currently the active one. Not sure if that can be kept only in code.
Anyway, in that case you would have two KeyData structs for the keys and one StateData telling you whether the resources should be encrypted as a whole and/or rewritten.

rfranzke: OK, thanks. Maybe it makes sense to prepare the data structures for the shoot state already to support future rotation of keys?

pkg/operation/common/types.go
```diff
@@ -131,6 +131,9 @@ const (
 	// is accepted.
 	GardenerDeletionProtected = "gardener.cloud/deletion-protected"
 
+	// ETCDSecretsEncryptionConfigDataName is the name of ShootState data entry holding the current key and encryption state used to encrypt shoot secrets
+	ETCDSecretsEncryptionConfigDataName = "etcdSecretsEncryptionConfiguration"
```
```suggestion
	ETCDEncryptionConfigDataName = "etcdSecretsEncryptionConfiguration"
```

Remove "secrets" - maybe we will encrypt even more in the future

pkg/operation/etcdencryption/encryptionconfiguration.go
rfranzke: ping

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -192,15 +192,30 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
...
+			Name:         "Generating etcd encryption configuration",
+			Fn:           flow.TaskFn(botanist.GenerateEncryptionConfiguration).DoIf(enableEtcdEncryption),
+			Dependencies: flow.NewTaskIDs(deployNamespace),
+		})
+		persistETCDEncryptionConfiguration = g.Add(flow.Task{
+			Name:         "Persisting etcd encryption configuration in ShootState",
+			Fn:           flow.TaskFn(botanist.PersistEncryptionConfiguration).DoIf(enableEtcdEncryption),
+			Dependencies: flow.NewTaskIDs(deployNamespace, generateEncryptionConfigurationMetaData),
+		})
+		_ = g.Add(flow.Task{
```
rfranzke, line 206: Can you add a `// TODO` comment here that this task can be removed in a future version?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -192,15 +192,30 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
...
+			Name:         "Removing old etcd encryption configuration secret from garden cluster",
+			Fn:           flow.TaskFn(botanist.RemoveOldETCDEncryptionSecretFromGardener),
+			Dependencies: flow.NewTaskIDs(generateEncryptionConfigurationMetaData),
```
```suggestion
			Dependencies: flow.NewTaskIDs(persistETCDEncryptionConfiguration),
```

pkg/operation/botanist/etcdencryption.go
```diff
@@ -158,36 +152,21 @@ func confChecksum(conf *apiserverconfigv1.EncryptionConfiguration) (string, erro
...
-		kutil.SetMetaDataAnnotation(gardenSecret, common.EtcdEncryptionChecksumAnnotationName, checksum)
-		return nil
-	})
-	return err
-}
-
-func (b *Botanist) rewriteShootSecretsIfChecksumChanged(ctx context.Context, checksum string) error {
-	secret := &corev1.Secret{}
-	if err := b.K8sSeedClient.Client().Get(ctx, kutil.Key(b.Shoot.SeedNamespace, common.EtcdEncryptionSecretName), secret); err != nil {
+	gardenerResourceList := gardencorev1alpha1helper.GardenerResourceDataList(b.ShootState.Spec.Gardener)
```
rfranzke: Instead of parsing the `b.ShootState`, can't you use `b.Shoot.ETCDEncryption` now?

pkg/operation/etcdencryption/etcdencryptiondata.go
```diff
@@ -0,0 +1,69 @@
...
+
+// ETCDEncryptionConfigData holds information whether a key is active or not and whether resources should be forcefully kept in plain text
+type ETCDEncryptionConfigData struct {
+	EncryptionKeys          []ETCDEncryptionKeyData `json:"encryptionKeys"`
+	ForcePlainTextResources bool                    `json:"forcePlainTextResources"`
+	RewriteResources        bool                    `json:"rewriteResources"`
+}
+
+// ETCDEncryptionUnmarshal unmarshals an ETCDKeyData json.
+func ETCDEncryptionUnmarshal(bytes []byte) (infodata.InfoData, error) {
```
rfranzke:
```suggestion
func Unmarshal(bytes []byte) (infodata.InfoData, error) {
```

It's already part of the `encryptionconfiguration` package, so calling `encryptionconfiguration.Unmarshal()` is more readable :) It might make sense to revisit also other functions of package and evaluate whether similar renamings can/should be applied.

pkg/operation/etcdencryption/etcdencryptiondata.go
```diff
@@ -0,0 +1,69 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *
+ */
+
+package encryptionconfiguration
```
rfranzke: The folder is called `etcdencryption` and the package is called `encryptionconfiguration`. Please bring this in sync.

rfranzke (changes requested):
Thanks @plkokanov, looks way way better now, way more readable, great. Just having a few minor suggestions now. I still need to test it out once all the comments have been addressed.

pkg/operation/botanist/etcdencryption.go
plkokanov: During restoration we can skip the generation and upsert-to-shootstate step so `b.Shoot.ETCDEncryption` will be nil.
If we don't skip those steps then I can take it from `b.Shoot.ETCDEncryption`.

rfranzke: OK, why not simply also setting `b.Shoot.ETCDEncryption` configuration? We already have this field now, so I'd say it makes sense to only parse the `ShootState` at most once, store the to-be-used encryption configuration in `b.Shoot.ETCDEncryption` and use it from there always?

plkokanov: Yep was also thinking of that. Wondering if I should just create a `ShootState` struct similar to the `Shoot` one and keep all relevant information there - `b.ShootState.ETCDEncryptionConfig` and the `ShootState` would just be in `b.ShootState.Info`. Could be useful for the Secrets/Certificates although I'm not sure if I can think of a concrete case right now. Maybe it makes sense to create that struct if necessary with the Certificates PR.

rfranzke: Yes, sounds good, let's check it again with the certificate PR. For now I'd keep the structure as is so that we can proceed with this PR.

rfranzke (approved):
/lgtm, thank you very much for addressing all the comments! :)

PR #2093, opened 2020-03-25
ControllerInstallation care controller caters with non-existing MR objects

rfranzke:
Prevent undesired/unhelpful logs like

```
time="2020-03-19T11:09:18Z" level=error msg="managedresources.resources.gardener.cloud \"my-extension-4qw9n\" not found" controllerinstallation-care=my-extension-4qw9n
```

when a `ControllerInstallation` gets deleted / an extension controller gets undeployed.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/controllerinstallation/controllerinstallation_care_control.go
+++ b/pkg/gardenlet/controller/controllerinstallation/controllerinstallation_care_control.go
@@ -59,6 +59,10 @@ func (c *Controller) reconcileControllerInstallationCareKey(key string) error {
 		return err
 	}
 
+	if controllerInstallation.DeletionTimestamp != nil {
+		return nil
+	}
+
 	if err := c.careControl.Care(controllerInstallation, key); err != nil {
 		return err
 	}
--- a/plugin/pkg/global/extensionvalidation/admission_test.go
+++ b/plugin/pkg/global/extensionvalidation/admission_test.go
@@ -248,7 +248,7 @@ var _ = Describe("ExtensionValidator", func() {
 						},
 						{
 							CRI: &core.CRI{Name: "cri",
-								ContainerRuntimes: []core.ContainerRuntime{{Type: "cr1",}, {Type: "cr2",}}},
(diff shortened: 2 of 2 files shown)


pkg/gardenlet/controller/controllerinstallation/controllerinstallation_care_control.go
```diff
@@ -105,7 +106,7 @@ func (c *defaultCareControl) Care(controllerInstallationObj *gardencorev1beta1.C
 	}
 
 	managedResource := &resourcesv1alpha1.ManagedResource{}
-	if err := k8sSeedClient.Client().Get(ctx, kutil.Key(v1beta1constants.GardenNamespace, controllerInstallation.Name), managedResource); err != nil {
+	if err := k8sSeedClient.Client().Get(ctx, kutil.Key(v1beta1constants.GardenNamespace, controllerInstallation.Name), managedResource); client.IgnoreNotFound(err) != nil {
```
timuthy: Now we skip the early exit. Isn't that a problem?

timebertt: If you ignore the `NotFound` error here, there will be no `managedresources.resources.gardener.cloud \"my-extension-4qw9n\" not found` logs, but instead the care controller will try to set the `Installed` condition to `InstallationPending` in the lines below, which competes with the conditions written by the `ControllerInstallation` controller in the [deletion flow](https://github.com/gardener/gardener/blob/dd06c0e10c9143cb70787e24fb0a832982b410a2/pkg/gardenlet/controller/controllerinstallation/controllerinstallation_control.go#L345).

I think you should rather add a check, if the `ControllerInstallation` is deleted to line 53 and exit early there, instead of ignoring the `NotFound` error here.

https://github.com/gardener/gardener/blob/dd06c0e10c9143cb70787e24fb0a832982b410a2/pkg/gardenlet/controller/controllerinstallation/controllerinstallation_care_control.go#L52-L57

timebertt (changes requested):

timebertt (approved):
Thanks
/approve

timuthy (approved):

PR #2094, opened 2020-03-25
increase timeout for care test

danielfoehrKn:
Increases the timeout for the care-controller integration test and flags the test as beta() and serial().

This is to avoid failures in case the condition is not met within the current timeout (5 minutes).
Fixes #

@schrodit @dguendisch 
Can you let me know if that helps?
**Release note**:

```improvement operator

```

### DIFF
--- a/test/integration/shoots/care/shoot_care.go
+++ b/test/integration/shoots/care/shoot_care.go
@@ -41,14 +41,14 @@ import (
 )
 
 const (
-	timeout = 10 * time.Minute
+	timeout = 20 * time.Minute
 )
 
 var _ = ginkgo.Describe("Shoot Care testing", func() {
 
 	f := testframework.NewShootFramework(nil)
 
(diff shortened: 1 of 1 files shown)


test/integration/shoots/care/shoot_care.go
```diff
@@ -41,14 +41,14 @@ import (
 )
 
 const (
-	timeout = 10 * time.Minute
+	timeout = 20 * time.Minute
```
schrodit, line 44: We should increase to at least 30min? as line 81 and line 65 have at least 30 minutes timeout

schrodit (changes requested):

rfranzke:
As discussed, let's close the PR - the test's purpose is actually to succeed only if the shoot condition is properly set to `False` after a short time. If it takes longer then either our code or the test is buggy.

Please open a new PR with either fixes to the code or to the test.

dguendisch:
Recall that this PR also set the Beta label to the test, so keep in mind that without further changes, the next gardener release will fail in integration tests.

rfranzke:
@danielfoehrKn can you first please create a separate PR only for labelling the test as beta to not hold the release?

PR #2098, opened 2020-03-26
Restart control plane controllers during maintenance

### LINKED ISSUE #1862: Auto-restart controller managers in shoot maintenance time window
We could auto-restart (kill pods) of the controller-manager as self-repair mechanism for shoot clusters with problems. We often see that the {kube,cloud}-controller-managers are hanging and only a pod restart helps. The root cause there is unknown and probably some deadlock situation in the k/k codebase itself, though, the auto-restart could function as work around until the root cause is fixed upstream.

**Why is this needed**:
Reduce manual ops effort and improve auto-repair capabilities of shoots.

timuthy:
This PR adds the possibility to restart control plane pods by:
1.) manually adding the shoot task `shoot.gardener.cloud/tasks: restartControlPlanePods`
2.) automatically during the shoot's maintenance time window

Contract:
Every pod in the shoot namespace of the seed which is labeled with `maintenance.gardener.cloud/restart: true` is affected by this restart (Pod deletion).

Fixes #1862

PRs which label the CCM will follow in the corresponding extension provider repository.

**Release note**:

```improvement operator
Control plane controllers of shoot clusters can now be automatically restarted during the shoot's maintenance time window. This is used to refresh the state and cache of probably long running containers and thus can circumvent potential dead-locks or starving routines. The feature can be enabled via the `ControllerManagerConfiguration` under `.controllers.shootMaintecance.enableShootControlPlaneRestarter` (see `example/20-componentconfig-gardener-controller-manager.yaml`).
```
```noteworthy developer
Gardener now deletes every pod of the shoot control plane with the label `maintenance.gardener.cloud/restart: true` if a shoot is assigned the task annotation `shoot.gardener.cloud/tasks: restartControlPlanePods`. This is used to refresh the state and cache of probably long running containers and thus can circumvent potential dead-locks or starving routines. If you are a developer of a Gardener extension please check the necessity and possibility of labelling such pods. For example, a running instance of the Cloud-Controller-Manager can safely be deleted/restarted while the shoot profits from a refreshed instance. More information can be found in `docs/extensions/shoot-maintenance.md`
```

### DIFF
--- a/charts/seed-controlplane/charts/kube-controller-manager/templates/kube-controller-manager.yaml
+++ b/charts/seed-controlplane/charts/kube-controller-manager/templates/kube-controller-manager.yaml
@@ -1,22 +1,3 @@
-apiVersion: v1
-kind: Service
-metadata:
-  name: kube-controller-manager
-  namespace: {{ .Release.Namespace }}
-  labels:
-    app: kubernetes
-    role: controller-manager
-spec:
-  type: ClusterIP
-  clusterIP: None
-  ports:
-  - name: metrics
-    port: {{ include "kube-controller-manager.port" . }}
-    protocol: TCP
-  selector:
-    app: kubernetes
-    role: controller-manager
----
 apiVersion: {{ include "deploymentversion" . }}
 kind: Deployment
 metadata:
@@ -46,6 +27,9 @@ spec:
         networking.gardener.cloud/to-dns: allowed
         networking.gardener.cloud/to-shoot-apiserver: allowed
         networking.gardener.cloud/from-prometheus: allowed
+{{- if .Values.podLabels }}
+{{ toYaml .Values.podLabels | indent 8 }}
+{{- end }}
     spec:
       tolerations:
       - effect: NoExecute
--- a/docs/extensions/shoot-maintenance.md
+++ b/docs/extensions/shoot-maintenance.md
@@ -1,0 +1,12 @@
+# Shoot maintenance
+Shoots configure a maintenance time window in which Gardener performs maintenance related tasks like infrastructure reconciliation, restarting pods, updating K8s or machine image versions.
+
+## Restart control plane controllers
+Gardener operators can make Gardener restart/delete certain control plane pods during a shoot's maintenance.
+This feature helps to automatically solve service denials of controllers due to stale caches, dead-locks or starving routines.
+
+Please note that these are exceptional cases but they are observed from time to time.
+Gardener, for example, takes this precautionary measure for `Kube-Controller-Manager` pods. 
+
+Extension controllers can extend the amount of pods being affected by these restarts.
+If your Gardener extension manages pods of a shoot's control plane (shoot namespace in seed) and it could potentially profit from a regular restart please consider labeling it with `maintenance.gardener.cloud/restart: true`.
--- a/example/20-componentconfig-gardener-controller-manager.yaml
+++ b/example/20-componentconfig-gardener-controller-manager.yaml
@@ -17,6 +17,7 @@ controllers:
   # shootMonitorPeriod: 300s
   shootMaintenance:
     concurrentSyncs: 5
+  # enableShootControlPlaneRestarter: true
   shootHibernation:
     concurrentSyncs: 5
   shootQuota:
--- a/pkg/apis/core/v1beta1/constants/types_constants.go
+++ b/pkg/apis/core/v1beta1/constants/types_constants.go
@@ -267,4 +267,7 @@ const (
 	// EventResourceReferenced indicates that the resource deletion is in waiting mode because the resource is still
 	// being referenced by at least one other resource (e.g. a SecretBinding is still referenced by a Shoot)
 	EventResourceReferenced = "ResourceReferenced"
+
+	// LabelPodMaintenanceRestart is a constant for a label that describes that a pod should be restarted during maintenance.
+	LabelPodMaintenanceRestart = "maintenance.gardener.cloud/restart"
 )
--- a/pkg/controllermanager/apis/config/types.go
+++ b/pkg/controllermanager/apis/config/types.go
@@ -144,6 +144,9 @@ type ShootMaintenanceControllerConfiguration struct {
 	// ConcurrentSyncs is the number of workers used for the controller to work on
 	// events.
 	ConcurrentSyncs int
+	// EnableShootControlPlaneRestarter configures whether adequate pods of the shoot control plane are restarted during maintenance.
+	// +optional
+	EnableShootControlPlaneRestarter *bool
 }
 
 // ShootQuotaControllerConfiguration defines the configuration of the
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -991,6 +991,9 @@ func (b *Botanist) DeployKubeControllerManager(ctx context.Context) error {
 			"checksum/secret-kube-controller-manager-server": b.CheckSums[common.KubeControllerManagerServerName],
 			"checksum/secret-service-account-key":            b.CheckSums["service-account-key"],
 		},
+		"podLabels": map[string]interface{}{
+			v1beta1constants.LabelPodMaintenanceRestart: "true",
+		},
 	}
 
 	if b.Shoot.HibernationEnabled == b.Shoot.Info.Status.IsHibernated {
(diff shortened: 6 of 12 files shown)


rfranzke:
Can you also add documentation to `docs/extensions`?

timuthy:
> Can you also add documentation to docs/extensions?

Done. Thanks for reminding 👍

charts/seed-controlplane/charts/kube-controller-manager/templates/kube-controller-manager.yaml
```diff
@@ -46,6 +46,9 @@ spec:
         networking.gardener.cloud/to-dns: allowed
```
timebertt, line 27: As you are already on it, can you split up the manifests in this file into multiple files?

timebertt (commented):
/lgtm in general

WDYT about also restarting the `cluster-autoscaler`, `gardener-resource-manager` and the `kube-scheduler`?

rfranzke:
> WDYT about also restarting the cluster-autoscaler, gardener-resource-manager and the kube-scheduler?

I'd introduce this later on in case we need it. So far we only saw CCM/KCM/MCM getting stuck sometimes, but not the other components yet.

docs/extensions/shoot-maintenance.md
```diff
@@ -0,0 +1,10 @@
+# Shoot maintenance
+Shoots configure a maintenance time window in which Gardener performs maintenance related tasks like infrastructure reconciliation, updating K8s or machine image versions.
+
+## Restart control plane controllers
+It is possible to configure Gardener to restart/delete certain control plane pods during a shoot's maintenance.
```
rfranzke:
```suggestion
As an extension controller it is possible to make Gardener restarting/deleting certain control plane pods during a shoot's maintenance.
```

docs/extensions/shoot-maintenance.md
```diff
@@ -0,0 +1,10 @@
+# Shoot maintenance
+Shoots configure a maintenance time window in which Gardener performs maintenance related tasks like infrastructure reconciliation, updating K8s or machine image versions.
```
rfranzke: Can you also add that it restarts pods to this sentence?

timebertt (dismissed):
> I'd introduce this later on in case we need it. So far we only saw CCM/KCM/MCM getting stuck sometimes, but not the other components yet.

Alright then, /approve

docs/extensions/shoot-maintenance.md
```diff
@@ -0,0 +1,10 @@
...
+
+## Restart control plane controllers
+It is possible to configure Gardener to restart/delete certain control plane pods during a shoot's maintenance.
+Stale states and caches are refreshed and thus solve potential dead-locks or starving routines.
+
+Gardener, for example, takes this precautionary measure for `Kube-Controller-Manager` pods.
```
rfranzke, line 9: Maybe add the more reasoning here - usually, this should not be necessary as controllers becoming hung should not happen in the first place. It's only an action we take to reduce ops effort because we have observed that indeed some controllers may become hung (KCM) - hard to reproduce and, thus, this is only a work-around until the controller is eventually fixed.

docs/extensions/shoot-maintenance.md
```diff
@@ -0,0 +1,10 @@
+# Shoot maintenance
+Shoots configure a maintenance time window in which Gardener performs maintenance related tasks like infrastructure reconciliation, updating K8s or machine image versions.
+
+## Restart control plane controllers
+It is possible to configure Gardener to restart/delete certain control plane pods during a shoot's maintenance.
+Stale states and caches are refreshed and thus solve potential dead-locks or starving routines.
+
+Gardener, for example, takes this precautionary measure for `Kube-Controller-Manager` pods. 
+
+If your Gardener extension manages deployments of a shoot's control plane (shoot namespace in seed) and it could potentially profit from a regular restart please consider labeling your pods with `maintenance.gardener.cloud/restart: true`.
```
rfranzke: Generally, this applies for pods, not deployments. It would also work for Statefulset/other workload resources.

example/20-componentconfig-gardener-controller-manager.yaml
```diff
@@ -17,6 +17,7 @@ controllers:
   # shootMonitorPeriod: 300s
   shootMaintenance:
     concurrentSyncs: 5
+  # restartControlPlaneControllers: true
```
rfranzke: Generally, it just restarts pods with a certain label, independent of whether they are controllers or control plane components. Can we find a more general name? Maybe `enableLabeledPodsRestart` or something similar?

pkg/apis/core/v1beta1/constants/types_constants.go
```diff
@@ -267,4 +267,7 @@ const (
 	// EventResourceReferenced indicates that the resource deletion is in waiting mode because the resource is still
 	// being referenced by at least one other resource (e.g. a SecretBinding is still referenced by a Shoot)
 	EventResourceReferenced = "ResourceReferenced"
+
+	// ControlPlaneMaintenanceRestart is a constant for a label that describes that a pod should be restarted during maintenance.
+	ControlPlaneMaintenanceRestart = "maintenance.gardener.cloud/restart"
```
rfranzke:
```suggestion
	LabelPodMaintenanceRestart = "maintenance.gardener.cloud/restart"
```

pkg/controllermanager/apis/config/types.go
```diff
@@ -144,6 +144,9 @@ type ShootMaintenanceControllerConfiguration struct {
 	// ConcurrentSyncs is the number of workers used for the controller to work on
 	// events.
 	ConcurrentSyncs int
+	// RestartControlPlaneControllers restarts the shoot control plane controller if set to `true`.
+	// +optional
+	RestartControlPlaneControllers *bool
```
rfranzke: same like above

pkg/operation/botanist/controlplane.go
```diff
@@ -1301,3 +1304,13 @@ func determineSchedule(shoot *gardencorev1beta1.Shoot, schedule string, f func(*
 	creationHour := shoot.CreationTimestamp.Hour()
 	return fmt.Sprintf(schedule, creationMinute, creationHour), nil
 }
+
+// RestartControlPlaneControllers restarts controllers of a shoot control plane.
+func (b *Botanist) RestartControlPlaneControllers(ctx context.Context) error {
```
```suggestion
func (b *Botanist) RestartLabeledPods(ctx context.Context) error {
```

pkg/operation/common/types.go
```diff
@@ -349,6 +349,9 @@ const (
 	// ShootTaskDeployInfrastructure is a name for a Shoot's infrastructure deployment task.
 	ShootTaskDeployInfrastructure = "deployInfrastructure"
 
+	// ShootTaskRestartControlPlaneControllers is a name for a Shoot task which is dedicated to restart related control plane controllers.
+	ShootTaskRestartControlPlaneControllers = "restartControlPlaneControllers"
```
```suggestion
	ShootTaskRestartLabeledPods = "restartLabeledPods"
```

rfranzke (changes requested):

example/20-componentconfig-gardener-controller-manager.yaml
timuthy: > it just restarts pods with a certain label

Imo, this is an implementation detail. The field should be self explainable so that an operator knows what it does. If we just call it something like `labeledPodRestart` then it is quite ambiguous because of: 
a) which labeled Pods?
b) in the shoot, in the seed, or both?

We can rename it but I don't want to be too generic. Maybe something like `enableRestartControlPlaneComponents`. WDYT?

pkg/operation/common/types.go
timuthy: See https://github.com/gardener/gardener/pull/2098#discussion_r399120264

example/20-componentconfig-gardener-controller-manager.yaml
rfranzke: What about `enableShootControlPlaneRestarter`?

docs/extensions/shoot-maintenance.md
timuthy: Actually the operator configures if the pods are restarted. I changed the text a bit to make it more clear.

timuthy:
Thanks for your reviews. PTAL.

pkg/operation/common/types.go
```diff
@@ -349,6 +349,9 @@ const (
 	// ShootTaskDeployInfrastructure is a name for a Shoot's infrastructure deployment task.
 	ShootTaskDeployInfrastructure = "deployInfrastructure"
 
+	// ShootTaskRestartControlPlaneControllers is a name for a Shoot task which is dedicated to restart related control plane controllers.
```
```suggestion
	// ShootTaskRestartControlPlanePods is a name for a Shoot task which is dedicated to restart related control plane pods.
```

rfranzke (commented):
Thanks! :) /lgtm now, just a minor nit in the doc string.

rfranzke (approved):
/lgtm

timebertt (approved):
/lgtm

PR #2099, opened 2020-03-26
GEP 10 - Container runtime extensions - addded cr config validation

nimrodoron:
Left over from: https://github.com/gardener/gardener/pull/2035
Part of #1454
Check that container runtimes with the same type can't have different configurations

### DIFF
--- a/pkg/apis/core/validation/shoot.go
+++ b/pkg/apis/core/validation/shoot.go
@@ -1076,6 +1076,8 @@ func ValidateWorkers(workers []core.Worker, fldPath *field.Path) field.ErrorList
 		allErrs = append(allErrs, field.Forbidden(fldPath, fmt.Sprintf("at least one worker pool must exist having either no taints or only the %q taint", corev1.TaintEffectPreferNoSchedule)))
 	}
 
+	allErrs = append(allErrs, ValidateContainerRuntimesConfigurations(workers, fldPath.Child("workers"))...)
+
 	return allErrs
 }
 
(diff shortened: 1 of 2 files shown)


pkg/apis/core/validation/shoot.go
```diff
@@ -1260,3 +1262,24 @@ func ValidateContainerRuntimes(containerRuntime []core.ContainerRuntime, fldPath
...
+
+func ValidateContainerRuntimesConfigurations(workers []core.Worker, fldPath *field.Path) field.ErrorList {
+	definedContainerRuntimesMap := map[string]core.ContainerRuntime{}
+	allErrs := field.ErrorList{}
+
+	for i, worker := range workers {
+		if worker.CRI != nil {
+			for j, cr := range worker.CRI.ContainerRuntimes {
+				if val, ok := definedContainerRuntimesMap[cr.Type]; ok {
+					if &cr.ProviderConfig != &val.ProviderConfig {
```
```suggestion
					if !apiequality.Semantic.DeepEqual(cr.ProviderConfig, val.ProviderConfig) {
```

pkg/apis/core/validation/shoot.go
```diff
@@ -1260,3 +1262,24 @@ func ValidateContainerRuntimes(containerRuntime []core.ContainerRuntime, fldPath
 
 	return allErrs
 }
+
+func ValidateContainerRuntimesConfigurations(workers []core.Worker, fldPath *field.Path) field.ErrorList {
```
rfranzke, line 1266: Can you add a doc string?

PR #2100, opened 2020-03-26
Delete DNS entries when shoot is hibernated

### LINKED ISSUE #2039: Delete DNSEntries once a cluster is hibernated
**What happened**:

1. `Cluster-A` is created. It's `LoadBalancer-A` IP address is (as an example) `1.2.3.4`.
1. `Cluster-A` is hibernated. `DNSEntries` are still available and point to `1.2.3.4`.
1. `Loadbalancer-A` is deleted and cloud provider removes the `1.2.3.4` IP address assignment.
1. `Cluster-B` is created. It's `LoadBalancer-B` IP address is `1.2.3.4` (the cloud provider now reuses the IP address).

```
openssl s_client -showcerts -connect 1.2.3.4:443 | openssl x509 -text

X509v3 Subject Alternative Name:
    DNS:kube-apiserver
    DNS:kube-apiserver.CLUSTER-B
    DNS:kube-apiserver.CLUSTER-B.svc
    DNS:CLUSTER-B.INTERNAL-DOMAIN
    DNS:kubernetes
    DNS:kubernetes.default
    DNS:kubernetes.default.svc
    DNS:kubernetes.default.svc.cluster.local
    DNS:CLUSTER-B.EXTERNAL
    IP Address:127.0.0.1
    IP Address:1.2.3.4
    IP Address:10.223.0.1
```

`DNSEntry` is removed once a cluster is hibernated and no traffic is send to the wrong cluster.

see above 

**Environment**:

- Gardener version: master
- Cloud provider or hardware configuration: GCP specifically, but it can apply for all
- Others:

timuthy:
This PR deletes DNS entries when a shoot enters hibernation.

Fixes #2039

**Release note**:

```improvement operator
DNS entries are now deleted when a shoot enters hibernation which prevents outdated DNS records.
```

### DIFF
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -31,6 +31,7 @@ import (
 	gardenletfeatures "github.com/gardener/gardener/pkg/gardenlet/features"
 	"github.com/gardener/gardener/pkg/operation/common"
 	"github.com/gardener/gardener/pkg/utils"
+	"github.com/gardener/gardener/pkg/utils/flow"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	"github.com/gardener/gardener/pkg/utils/kubernetes/health"
 	"github.com/gardener/gardener/pkg/utils/retry"
@@ -288,6 +289,14 @@ func (b *Botanist) HibernateControlPlane(ctx context.Context) error {
 
 	if !b.Shoot.DisableDNS {
 		if err := c.Delete(ctx, &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: v1beta1constants.DeploymentNameKubeAPIServer, Namespace: b.Shoot.SeedNamespace}}, kubernetes.DefaultDeleteOptions...); client.IgnoreNotFound(err) != nil {
+			return err
+		}
+
+		if err := flow.Parallel(
+			func(ctx context.Context) error { return b.DestroyInternalDomainDNSRecord(ctx) },
+			func(ctx context.Context) error { return b.DestroyExternalDomainDNSRecord(ctx) },
+			func(ctx context.Context) error { return b.DestroyIngressDNSRecord(ctx) },
+		)(ctx); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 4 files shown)


rfranzke:
Is there a reason for not deleting the ingress DNS entry as well?

timuthy:
> Is there a reason for not deleting the ingress DNS entry as well?

https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/addons.go#L44

rfranzke:
😕 Does it make sense to move this now also to the place where you delete the other DNS records?

timuthy:
PTAL

pkg/operation/botanist/controlplane.go
```diff
@@ -290,6 +291,19 @@ func (b *Botanist) HibernateControlPlane(ctx context.Context) error {
 		if err := c.Delete(ctx, &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: v1beta1constants.DeploymentNameKubeAPIServer, Namespace: b.Shoot.SeedNamespace}}, kubernetes.DefaultDeleteOptions...); client.IgnoreNotFound(err) != nil {
 			return err
 		}
+
+		fns := []flow.TaskFn{
+			func(ctx context.Context) error { return b.DestroyInternalDomainDNSRecord(ctx) },
+			func(ctx context.Context) error { return b.DestroyExternalDomainDNSRecord(ctx) },
+		}
+
+		if b.Shoot.NginxIngressEnabled() {
```
rfranzke: Is this check needed? I think we would have a problem if a user changes in one shot the shoot spec to `hibernated.enabled=true`, `nginx-ingress.enabled=false`, or? Then the nginx-ingress DNS record would remain?

timuthy: Hm I think the extended condition `.DoIf(dnsEnabled && managedExternalDNS && !o.Shoot.HibernationEnabled)` prevents this.
https://github.com/gardener/gardener/blob/cf98a923f51ef3c2ea556032b46642d2869c0c09/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go#L280

It's just a proposal to save some calls but since not found errors are ignored we can also execute it every time. I don't have a strong opinion.

timuthy: Yes nice catch. I'll remove it.

rfranzke (approved):
/lgtm

PR #2101, opened 2020-03-26
GEP 10 - Container runtime extensions - addded cr config validation

nimrodoron:
Left over from: #2035
Part of #1454
Check that container runtimes with the same type can't have different configurations


rfranzke (dismissed):
/lgtm

pkg/apis/core/validation/shoot.go
```diff
@@ -1260,3 +1262,25 @@ func ValidateContainerRuntimes(containerRuntime []core.ContainerRuntime, fldPath
 
 	return allErrs
 }
+
+// This function checks that all container runtimes with the same type have the same configurations
```
```suggestion
// ValidateContainerRuntimesConfigurations checks that all container runtimes with the same type have the same configurations.
```

danielfoehrKn (dismissed):

rfranzke (approved):

PR #2105, opened 2020-03-26
GEP 5 - introduce version classification, validation and admission

### LINKED ISSUE #1577: Implement versioning policy contract (GEP-5)

danielfoehrKn:
This PR is part of GEP5 .
- Adds the field classification to the expirable version 

Validation
- do not allow adding versions with expiration date in the past
- only allow one 'supported' version per major & minor version combination

Admission 
-  ignore preview versions for defaulting
- when classification is specified, only allow valid values

Part of #1577 

Please also [checkout the documentation](https://github.com/gardener/gardener/blob/05e7905e2cb40e2f9f70c466281614a2a31b0008/docs/operations/versioning.md)
**Release note**:

```improvement operator
Kubernetes and Machine image versions now have the optional field "classification" with allowed values of [preview, supported, deprecated, expired]. Please see GEP5 (https://github.com/gardener/gardener/docs/proposals/05-versioning-policy.md) and the operations documentation under https://github.com/gardener/gardener/blob/master/docs/operations/versioning.md.  
```


danielfoehrKn:
Need to remove the deepcopy.go files that partly contain changes from another branch

danielfoehrKn:
Please review this PR first as it contains the prerequisite for the other GEP5 PR's.
After this is merged, I will rebase the others.

pkg/apis/core/types_cloudprofile.go
```diff
@@ -150,3 +153,24 @@ const (
...
+	ClassificationPreview VersionClassification = "preview"
+	// ClassificationSupported indicates that a patch version is the default version for the particular minor version.
+	// There is always exactly one supported Kubernetes patch version for every still maintained Kubernetes minor version.
+	// Supported versions are eligible for the automated Kubernetes patch version update for shoot clusters in Gardener.
+	ClassificationSupported VersionClassification = "supported"
+	// ClassificationDeprecated indicates that a patch version should not be used anymore, should be updated to a new version
+	// and will eventually expire.
+	// The time after which a deprecated version expires can be configured in the `kubernetesVersionManagement` and `machineImageVersionManagement` section of the CloudProfile Controller config.
+	// Every version that is neither in preview nor supported is deprecated.
+	// All patch versions of not supported minor versions are deprecated.
```
rfranzke: I don't understand this sentence - what do you mean with "not supported minor versions"?

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -44,11 +45,65 @@ func ValidateCloudProfileUpdate(newProfile, oldProfile *core.CloudProfile) field
...
 	return allErrs
 }
 
+// ValidateCloudProfileSpecUpdate validates the spec update of a CloudProfile
+func ValidateCloudProfileSpecUpdate(new *core.CloudProfileSpec, old *core.CloudProfileSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	allErrs = append(allErrs, validateCloudProfileVersionsUpdate(new.Kubernetes.Versions, old.Kubernetes.Versions, fldPath.Child("kubernetes", "versions"))...)
+
+	for index, _ := range old.MachineImages {
+		if len(new.MachineImages) - 1  >= index {
```
rfranzke: Why do we work with the index here? What if the order will be changed? Shouldn't we rather look for the machine image name and do the comparison based on it?

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -44,11 +45,65 @@ func ValidateCloudProfileUpdate(newProfile, oldProfile *core.CloudProfile) field
...
+
+	return allErrs
+}
+
+// ValidateCloudProfileAddedVersions validates versions added to the CloudProfile
+func ValidateCloudProfileAddedVersions(versions []core.ExpirableVersion, addedVersions map[int]core.ExpirableVersion, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for index, version := range addedVersions {
+		// Do not allow adding addedVersions with expiration date in the past
+		if  version.ExpirationDate != nil && version.ExpirationDate.Time.UTC().Before(time.Now()) {
```
rfranzke:
```suggestion
		if  version.ExpirationDate != nil && version.ExpirationDate.Time.UTC().Before(time.Now().UTC()) {
```

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -44,11 +45,65 @@ func ValidateCloudProfileUpdate(newProfile, oldProfile *core.CloudProfile) field
...
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(index), version.Version, fmt.Sprintf("unable to add version '%s'. Adding a version with expiration date in the past is not allowed", version.Version)))
```
rfranzke:
```suggestion
			allErrs = append(allErrs, field.Forbidden(fldPath.Index(index), version.Version, fmt.Sprintf("unable to add version %q. Adding a version with expiration date in the past is not allowed", version.Version)))
```

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -44,11 +45,65 @@ func ValidateCloudProfileUpdate(newProfile, oldProfile *core.CloudProfile) field
...
+			continue
+		}
+
+		if version.Classification != nil && *version.Classification == core.ClassificationSupported {
+			currentSemVer, err := semver.NewVersion(version.Version)
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(fldPath.Index(index), version.Version, fmt.Sprintf("unable to add version '%s'. Version should be semVer compatible", version.Version)))
```
rfranzke:
```suggestion
				allErrs = append(allErrs, field.Invalid(fldPath.Index(index), version.Version, fmt.Sprintf("unable to add version '%s'. Version must be a semantic version", version.Version)))
```

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +167,63 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
...
+		if version.Classification == nil || *version.Classification != classification {
+			continue
+		}
+		result = append(result, version)
+	}
+	return result
+}
+
+// FilterVersionsWithSameMajorMinor filters the given versions slice for versions having the same major and minor version as the given version
+func FilterVersionsWithSameMajorMinor(version semver.Version, versions []core.ExpirableVersion) ([]core.ExpirableVersion, error) {
```
rfranzke:
```suggestion
func FindVersionsWithSameMajorMinor(versions []core.ExpirableVersion, version semver.Version) ([]core.ExpirableVersion, error) {
```

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +167,63 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
...
+		if _, found := newVersions[x.Version]; !found {
+			diff[index] = x
+		}
+	}
+	return diff
+}
+
+// GetAddedVersions finds versions that have been added in the new compared to the new version slice.
+// returns a map associating the index in the new version slice with the version
+func GetAddedVersions(old []core.ExpirableVersion, new []core.ExpirableVersion) map[int]core.ExpirableVersion {
```
rfranzke: This function looks almost identical as the above function. Can you move the logic out and just parametrize it accordingly to prevent duplication?

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +167,63 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
 	}
 	return nil
 }
+
+// GetRemovedVersions finds versions that have been removed in the old compared to the new version slice.
+// returns a map associating the index in the old version slice with the version
+func GetRemovedVersions(old []core.ExpirableVersion, new []core.ExpirableVersion) map[int]core.ExpirableVersion {
+	newVersions := make(map[string]struct{}, len(old))
```
rfranzke:
```suggestion
	newVersions := make(map[string]struct{}, len(new))
```

pkg/apis/core/validation/cloudprofile.go
rfranzke: Why is this check only performed if the classification is supported?

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -44,11 +45,65 @@ func ValidateCloudProfileUpdate(newProfile, oldProfile *core.CloudProfile) field
...
+
+			filteredVersions, err := helper.FilterVersionsWithSameMajorMinor(*currentSemVer, helper.FilterVersionsWithClassification(versions, core.ClassificationSupported))
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(fldPath.Index(index), currentSemVer.String(), fmt.Sprintf("unable to add version '%s' : '%v'", version.Version, err)))
+				continue
+			}
+
+			// do not allow adding multiple supported versions per minor version
+			if len(filteredVersions) > 0 {
+				allErrs = append(allErrs, field.Invalid(fldPath.Index(index), currentSemVer.String(), fmt.Sprintf("unable to add version '%s' with classification '%s'. Only one '%s' version is allowed  per minor version", version.Version, core.ClassificationSupported, core.ClassificationSupported)))
```
rfranzke:
```suggestion
				allErrs = append(allErrs, field.Forbidden(fldPath.Index(index), currentSemVer.String(), fmt.Sprintf("unable to add version %q with classification %q. Only one %q version is allowed  per minor version", version.Version, core.ClassificationSupported, core.ClassificationSupported)))
```

plugin/pkg/shoot/validator/admission.go
```diff
@@ -519,9 +519,16 @@ func validateKubernetesVersionConstraints(constraints []core.ExpirableVersion, s
 		}
 
 		if getLatestPatchVersion {
```
rfranzke: Unrelated to your change, just saw it: Can you fix line 511 and convert both times to UTC?

rfranzke (changes requested):

rfranzke:

pkg/apis/core/helper/helpers.go
rfranzke: Also, as far as I can see the index is only used for the field path error. If possible, can you enhance the data structure so that you can reuse in the admission plugin in #2106?

pkg/apis/core/types_cloudprofile.go
danielfoehrKn: Thanks for spotting. Removed it (that information should not be there and is part of the optional version management).

pkg/apis/core/validation/cloudprofile.go
danielfoehrKn: Yes, changed it.

pkg/apis/core/validation/cloudprofile.go
danielfoehrKn: This is only to obtain the semantic version and is already seperatly checked here for [machine versions](https://github.com/gardener/gardener/blob/deb59ec01131a39b18be35267466590219027078/pkg/apis/core/validation/cloudprofile.go#L255-L254 ):  (will adjust error message here too) and here [via regex for Kubernetes versions](https://github.com/gardener/gardener/blob/deb59ec01131a39b18be35267466590219027078/pkg/apis/core/validation/cloudprofile.go#L154-L153).

pkg/apis/core/helper/helpers.go
danielfoehrKn: Returning a sets.Set  containing the versions to be able to reuse them in the admission plugin.
Thanks

pkg/apis/core/helper/helpers.go
```diff
@@ -17,6 +17,8 @@ package helper
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/util/sets"
```
timebertt: Move to 3rd party import block below

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +169,61 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
...
+
+// GetAddedVersions finds versions that have been added in the new compared to the new version slice.
+// returns a map associating the index in the new version slice with the version and a string set containing the versions.
+func GetAddedVersions(old []core.ExpirableVersion, new []core.ExpirableVersion) (map[int]core.ExpirableVersion, sets.String) {
+	return getVersionDiff(new, old)
+}
+
+// getVersionDiff gets versions that are in v1 but not in v2.
+// Returns versions mapped to its index in v1 and also a string set containing the versions
+func getVersionDiff(v1, v2 []core.ExpirableVersion) (map[int]core.ExpirableVersion, sets.String) {
```
timebertt: Why don't you only return a `map[string]int`, so you can simply check, if a given version was added/removed, and get the index for the field path?
```suggestion
func getVersionDiff(v1, v2 []core.ExpirableVersion) (map[string]int) {
```

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +169,61 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
...
+	v2Versions := make(map[string]struct{}, len(v2))
+	for _, x := range v2 {
+		v2Versions[x.Version] = struct{}{}
+	}
+	diff := make(map[int]core.ExpirableVersion, len(v1))
```
timebertt: You don't know at this point, how long the diff map will be.
```suggestion
	diff := map[int]core.ExpirableVersion{}
```

pkg/apis/core/helper/helpers_test.go
```diff
@@ -15,12 +15,15 @@
 package helper_test
 
 import (
+	"github.com/Masterminds/semver"
```
timebertt: Move down to 3rd party import block

pkg/apis/core/types_cloudprofile.go
```diff
@@ -150,3 +153,21 @@ const (
 	// VolumeClassPremium is a constant for the premium volume class.
 	VolumeClassPremium string = "premium"
 )
+
+// VersionClassification is the logical state of a version according to https://github.com/gardener/gardener/blob/master/docs/proposals/05-versioning-policy.md
+type VersionClassification string
+
+var SupportedVersionClassifications = sets.NewString(string(ClassificationPreview), string(ClassificationSupported), string(ClassificationDeprecated))
```
timebertt: Can you remove this line?
Supported value sets should be in the validation logic like here https://github.com/gardener/gardener/blob/c87b1e95dac172d3ac00617ed08d5e47cc512eda/pkg/apis/core/validation/project.go#L135 or here https://github.com/gardener/gardener/blob/8fb9d96f4ed16e1fd4d3069cfc20b0282d7ef2b5/pkg/apis/core/validation/seed.go#L98

pkg/apis/core/types_cloudprofile.go
```diff
@@ -150,3 +153,21 @@ const (
...
+
+const (
+	// ClassificationPreview indicates that a version has recently been added and not promoted to "Supported" yet.
+	// ClassificationPreview versions will not be considered for automatic Kubernetes patch version updates.
```
timebertt:
```suggestion
	// ClassificationPreview versions will not be considered for automatic Kubernetes and Machine Image patch version updates.
```

pkg/apis/core/types_cloudprofile.go
```diff
@@ -150,3 +153,21 @@ const (
...
+	ClassificationPreview VersionClassification = "preview"
+	// ClassificationSupported indicates that a patch version is the default version for the particular minor version.
+	// Supported versions are eligible for the automated Kubernetes patch version update for shoot clusters in Gardener.
```
timebertt:
```suggestion
	// ClassificationSupported indicates that a patch version is the default version for the particular minor version.
	// Supported versions are eligible for automated Kubernetes and Machine Image patch version update for shoot clusters in Gardener.
```

timebertt: > ... is the default version for the particular minor version

If I understood it correctly, if the Version Management is disabled, the `Supported` version is also the overall default, right?

pkg/apis/core/types_cloudprofile.go
```diff
@@ -150,3 +153,21 @@ const (
...
+	ClassificationSupported VersionClassification = "supported"
+	// ClassificationDeprecated indicates that a patch version should not be used anymore, should be updated to a new version
+	// and will eventually expire.
+	// The time after which a deprecated version expires can be configured in the `kubernetesVersionManagement` and `machineImageVersionManagement` section of the CloudProfile Controller config.
```
timebertt: Hmm, isn't the "time after which a deprecated version expires" defined by `core.ExpirableVersion.ExpirationDate`.
If I see this correctly, this has nothing to do with the automated version management.

pkg/apis/core/v1alpha1/types_cloudprofile.go
```diff
@@ -179,3 +182,19 @@ const (
 	// VolumeClassPremium is a constant for the premium volume class.
 	VolumeClassPremium string = "premium"
 )
+
+// VersionClassification is the logical state of a version according to https://github.wdf.sap.corp/kubernetes/kube-docs/wiki/Versioning-Policy
+type VersionClassification string
+
+const (
+	// ClassificationPreview indicates that a version has recently been added and not promoted to "Supported" yet.
```
timebertt, line 190: Please adapt this version accordingly to the internal version

pkg/apis/core/v1beta1/types_cloudprofile.go
```diff
@@ -179,3 +182,19 @@ const (
 	// VolumeClassPremium is a constant for the premium volume class.
 	VolumeClassPremium string = "premium"
 )
+
+// VersionClassification is the logical state of a version according to https://github.com/gardener/gardener/blob/master/docs/proposals/05-versioning-policy.md
+type VersionClassification string
+
+const (
+	// ClassificationPreview indicates that a version has recently been added and not promoted to "Supported" yet.
```
timebertt, line 190: Please adapt this version accordingly to the internal version

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -44,11 +45,68 @@ func ValidateCloudProfileUpdate(newProfile, oldProfile *core.CloudProfile) field
...
 
 	allErrs = append(allErrs, apivalidation.ValidateObjectMetaUpdate(&newProfile.ObjectMeta, &oldProfile.ObjectMeta, field.NewPath("metadata"))...)
+	allErrs = append(allErrs, ValidateCloudProfileSpecUpdate(&newProfile.Spec, &oldProfile.Spec, field.NewPath("spec"))...)
 	allErrs = append(allErrs, ValidateCloudProfile(newProfile)...)
 
 	return allErrs
 }
 
+// ValidateCloudProfileSpecUpdate validates the spec update of a CloudProfile
+func ValidateCloudProfileSpecUpdate(new *core.CloudProfileSpec, old *core.CloudProfileSpec, fldPath *field.Path) field.ErrorList {
```
timebertt:
```suggestion
func ValidateCloudProfileSpecUpdate(new, old *core.CloudProfileSpec, fldPath *field.Path) field.ErrorList {
```

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -44,11 +45,68 @@ func ValidateCloudProfileUpdate(newProfile, oldProfile *core.CloudProfile) field
...
+		}
+
+		if version.Classification != nil && *version.Classification == core.ClassificationSupported {
+			currentSemVer, err := semver.NewVersion(version.Version)
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(fldPath.Index(index), version.Version, fmt.Sprintf("unable to add version '%s'. Version must be a semantic version", version.Version)))
+				continue
+			}
+
+			filteredVersions, err := helper.FindVersionsWithSameMajorMinor(helper.FilterVersionsWithClassification(versions, core.ClassificationSupported), *currentSemVer, )
```
timebertt:
```suggestion
			filteredVersions, err := helper.FindVersionsWithSameMajorMinor(helper.FilterVersionsWithClassification(versions, core.ClassificationSupported), *currentSemVer)
```

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -44,11 +45,68 @@ func ValidateCloudProfileUpdate(newProfile, oldProfile *core.CloudProfile) field
...
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(fldPath.Index(index), currentSemVer.String(), fmt.Sprintf("unable to add version '%s' : '%v'", version.Version, err)))
```
timebertt: Can you enhance this error to make it clear, what the actual error was?

pkg/apis/core/validation/cloudprofile_test.go
```diff
@@ -504,4 +562,127 @@ var _ = Describe("CloudProfile Validation Tests ", func() {
...
 		})
 	})
+
+	var (
+		cloudProfileNew *core.CloudProfile
+		cloudProfileOld *core.CloudProfile
+		dateInThePast   = &metav1.Time{Time: time.Now().AddDate(-5, 0, 0)}
+	)
+
+	Describe("#ValidateCloudProfile Spec Update", func() {
```
timebertt:
```suggestion
	Describe("#ValidateCloudProfileSpecUpdate", func() {
```

pkg/apis/core/validation/cloudprofile_test.go
```diff
@@ -504,4 +562,127 @@ var _ = Describe("CloudProfile Validation Tests ", func() {
...
+			Expect(errorList).To(ConsistOf(PointTo(MatchFields(IgnoreExtras, Fields{
+				"Type":   Equal(field.ErrorTypeForbidden),
+				"Field":  Equal("spec.kubernetes.versions[2]"),
+				"Detail": ContainSubstring("Adding a version with expiration date in the past is not allowed"),
+			}))))
+		})
+
+		It("do not allow adding a second supported version", func() {
+			versions[0].Classification = &supportedClassification
+			// supported version is higher than the preview version
```
timebertt: WDYM with `the preview version`?

plugin/pkg/shoot/validator/admission.go
```diff
@@ -519,9 +519,16 @@ func validateKubernetesVersionConstraints(constraints []core.ExpirableVersion, s
...
+			// ignore preview versions for defaulting
+			if versionConstraint.Classification != nil && *versionConstraint.Classification == core.ClassificationPreview {
+				// do not include preview version in the valid values when using defaulting
+				validValues = validValues[:len(validValues)-1]
```
timebertt: This seems wrong to me. How do you know, that the current `versionConstraint` is the last entry in `validValues`?

plugin/pkg/shoot/validator/admission_test.go
```diff
@@ -990,6 +990,41 @@ var _ = Describe("validator", func() {
...
 			})
 
+			It("should default a major.minor kubernetes version only to non-preview versions", func() {
+				shoot.Spec.Kubernetes.Version = "1.6"
+				preview := core.ClassificationPreview
+				previewVersion := core.ExpirableVersion{Version: "1.6.6", Classification: &preview}
+				highestNonPreviewPatchVersion := core.ExpirableVersion{Version: "1.6.5"}
+				cloudProfile.Spec.Kubernetes.Versions = append(cloudProfile.Spec.Kubernetes.Versions, previewVersion, highestNonPreviewPatchVersion, core.ExpirableVersion{Version: "1.7.1"}, core.ExpirableVersion{Version: "1.7.2"})
+
+				coreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&project)
```
timebertt: Can you please wrap all calls to the informer store with `Expect(...).To(Succeed())` to avoid incorrect tests?
```suggestion
				Expect(coreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&project)).To(Succeed())
```

plugin/pkg/shoot/validator/admission_test.go
```diff
@@ -990,6 +990,41 @@ var _ = Describe("validator", func() {
...
+				cloudProfile.Spec.Kubernetes.Versions = []core.ExpirableVersion{previewVersion, highestNonPreviewPatchVersion, {Version: "1.7.1"}, {Version: "1.7.2"}}
+
+				coreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&project)
+				coreInformerFactory.Core().InternalVersion().CloudProfiles().Informer().GetStore().Add(&cloudProfile)
+				coreInformerFactory.Core().InternalVersion().Seeds().Informer().GetStore().Add(&seed)
+				attrs := admission.NewAttributesRecord(&shoot, nil, core.Kind("Shoot").WithVersion("version"), shoot.Namespace, shoot.Name, core.Resource("shoots").WithVersion("version"), "", admission.Create, &metav1.CreateOptions{}, false, nil)
+
+				err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+				Expect(err).To(HaveOccurred())
```
timebertt, line 1025: Can you test if the returned error is the expected one?

timebertt (changes requested):
Overall it looks quite good.
I left you some comments here and there.

pkg/apis/core/types_cloudprofile.go
danielfoehrKn: yes, correct.

pkg/apis/core/types_cloudprofile.go
danielfoehrKn: The core.ExpirableVersion.ExpirationDate when using version management is configurable in the  `kubernetesVersionManagement`.
Though that is not expressed in this comment, will update. Thanks

pkg/apis/core/validation/cloudprofile_test.go
danielfoehrKn: leftover comment, thanks

plugin/pkg/shoot/validator/admission.go
danielfoehrKn: It can only be, it has been [previously added here](https://github.com/gardener/gardener/blob/70b7ea5c9831489463276478936d19295795d8d2/plugin/pkg/shoot/validator/admission.go#L515)

pkg/apis/core/types_cloudprofile.go
timebertt: Ok, can you then also adapt the doc string accordingly?

pkg/apis/core/types_cloudprofile.go
timebertt: > The core.ExpirableVersion.ExpirationDate when using version management is configurable in the kubernetesVersionManagement

Sure, but my point was though, that the time after which a version expires is configured by the `ExpirationDate` field (can be set individually per version or even manually) and is in general not related to the automated version management. The version management, will rather use this field to expire versions automatically (using a pre-defined default expiration time configured in the controller config).

plugin/pkg/shoot/validator/admission.go
timebertt: Ah, sure. My bad... 

Still, I think, appending an item to a slice, then removing it directly afterwards if certain conditions are met, is not good style.
WDYT about doing it like this?
```go
if !getLatestPatchVersion {
  validValues = append(validValues, versionConstraint.Version)
} else {
  // ignore preview versions for defaulting
  if versionConstraint.Classification == nil || *versionConstraint.Classification != core.ClassificationPreview {
    validValues = append(validValues, versionConstraint.Version)
  }
  // and so on...
}
```

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +169,61 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
...
+// GetAddedVersions finds versions that have been added in the new compared to the new version slice.
+// returns a map associating the index in the new version slice with the version and a string set containing the versions.
+func GetAddedVersions(old []core.ExpirableVersion, new []core.ExpirableVersion) (map[int]core.ExpirableVersion, sets.String) {
+	return getVersionDiff(new, old)
+}
+
+// getVersionDiff gets versions that are in v1 but not in v2.
+// Returns versions mapped to its index in v1 and also a string set containing the versions
+func getVersionDiff(v1, v2 []core.ExpirableVersion) (map[int]core.ExpirableVersion, sets.String) {
+	v2Versions := make(map[string]struct{}, len(v2))
```
timuthy: Why don't you use a `set` here?

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -44,11 +45,68 @@ func ValidateCloudProfileUpdate(newProfile, oldProfile *core.CloudProfile) field
...
+		// Do not allow adding addedVersions with expiration date in the past
+		if  version.ExpirationDate != nil && version.ExpirationDate.Time.UTC().Before(time.Now().UTC()) {
+			allErrs = append(allErrs, field.Forbidden(fldPath.Index(index), fmt.Sprintf("unable to add version %q. Adding a version with expiration date in the past is not allowed", version.Version)))
+			continue
+		}
+
+		if version.Classification != nil && *version.Classification == core.ClassificationSupported {
+			currentSemVer, err := semver.NewVersion(version.Version)
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(fldPath.Index(index), version.Version, fmt.Sprintf("unable to add version '%s'. Version must be a semantic version", version.Version)))
```
timuthy:
```suggestion
				allErrs = append(allErrs, field.Invalid(fldPath.Index(index), version.Version, fmt.Sprintf("unable to add version %q. Version must be a semantic version", version.Version)))
```

pkg/apis/core/validation/cloudprofile.go
timuthy:
```suggestion
				allErrs = append(allErrs, field.Invalid(fldPath.Index(index), currentSemVer.String(), fmt.Sprintf("unable to add version %q : '%v'", version.Version, err)))
```

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +169,61 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
...
+		result = append(result, version)
+	}
+	return result
+}
+
+// FindVersionsWithSameMajorMinor filters the given versions slice for versions having the same major and minor version as the given version
+func FindVersionsWithSameMajorMinor(versions []core.ExpirableVersion, version semver.Version) ([]core.ExpirableVersion, error) {
+	var result []core.ExpirableVersion
+	for _, v := range versions {
+		semVer, err := semver.NewVersion(v.Version)
```
timuthy, line 217: I wonder if we can ignore this error since the semver check is done by the validator.

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -44,11 +45,68 @@ func ValidateCloudProfileUpdate(newProfile, oldProfile *core.CloudProfile) field
...
+
+			filteredVersions, err := helper.FindVersionsWithSameMajorMinor(helper.FilterVersionsWithClassification(versions, core.ClassificationSupported), *currentSemVer, )
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(fldPath.Index(index), currentSemVer.String(), fmt.Sprintf("unable to add version '%s' : '%v'", version.Version, err)))
+				continue
+			}
+
+			// do not allow adding multiple supported versions per minor version
+			if len(filteredVersions) > 0 {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Index(index), fmt.Sprintf("unable to add version %q with classification %q. Only one %q version is allowed  per minor version", version.Version, core.ClassificationSupported, core.ClassificationSupported)))
```
timuthy: Why is this only checked for update but not for create events?

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -100,6 +158,16 @@ func validateKubernetesSettings(kubernetes core.KubernetesSettings, fldPath *fie
 		} else {
 			versionsFound.Insert(version.Version)
 		}
+		allErrs = append(allErrs, validateExpirableVersion(version, idxPath)...)
+	}
+
+	return allErrs
+}
+
+func validateExpirableVersion(version core.ExpirableVersion, fldPath *field.Path) field.ErrorList {
```
timuthy: Should we also checkt that one `supported` version is available if at least one version specifies a classification?

plugin/pkg/shoot/validator/admission.go
timuthy: I'm not quite sure how this is related to the `getLatestPatchVersion` branch. Why not checking the `preview` classification when the version is added to the `validValues` in line 515?

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +169,61 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
 	}
 	return nil
 }
+
+// GetRemovedVersions finds versions that have been removed in the old compared to the new version slice.
+// returns a map associating the index in the old version slice with the version and a string set containing the versions.
+func GetRemovedVersions(old []core.ExpirableVersion, new []core.ExpirableVersion) (map[int]core.ExpirableVersion, sets.String) {
```
timuthy: I think we should avoid using multiple return values for returning redundant data. It's kind of hard to understand why these different "collections" are returned and only because you need exactly a `map` here whereas you need a `set` there shouldn't be the reason.

timuthy (changes requested):

pkg/apis/core/types_cloudprofile.go
danielfoehrKn: Ok makes sense. I just removed that sentence as it might lead to confusion plus it is a duplication of what is already statet in the version itself.

pkg/apis/core/helper/helpers.go
danielfoehrKn: Yes makes sense, thanks

pkg/apis/core/validation/cloudprofile.go
danielfoehrKn: > I wonder if we can ignore this error since the semver check is done by the validator

Not returning an error any more according to this suggestion - so this error is no longer needed.

plugin/pkg/shoot/validator/admission.go
danielfoehrKn: @timuthy A preview version is a valid version. A shoot can be created with it. Otherwise no shoot can use a preview version.

danielfoehrKn: I adjusted it. PTAL

pkg/apis/core/validation/cloudprofile.go
danielfoehrKn: Good point. Moved the check to cover both create and update events.

pkg/apis/core/validation/cloudprofile.go
danielfoehrKn: I think that is up to the operator to decide. If using the VersionManagement or the VersionMaintenance that is already made sure through reconciliation.

pkg/apis/core/helper/helpers.go
danielfoehrKn: adjusted it.

danielfoehrKn:
Adressed the comments, thanks! PTAL

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +168,57 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
 	}
 	return nil
 }
+
+// GetRemovedVersions finds versions that have been removed in the old compared to the new version slice.
+// returns a map associating the index in the old version slice with the version and a string set containing the versions.
```
timebertt:
```suggestion
// returns a map associating the index in the old version slice with the version.
```

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +168,57 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
...
+func GetRemovedVersions(old []core.ExpirableVersion, new []core.ExpirableVersion) map[string]int {
+	return getVersionDiff(old, new)
+}
+
+// GetAddedVersions finds versions that have been added in the new compared to the new version slice.
+// returns a map associating the index in the new version slice with the version and a string set containing the versions.
```
timebertt:
```suggestion
// returns a map associating the index in the new version slice with the version.
```

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +168,57 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
...
+func GetAddedVersions(old []core.ExpirableVersion, new []core.ExpirableVersion) map[string]int {
+	return getVersionDiff(new, old)
+}
+
+// getVersionDiff gets versions that are in v1 but not in v2.
+// Returns versions mapped to its index in v1 and also a string set containing the versions
```
timebertt:
```suggestion
// Returns versions mapped to their index in v1.
```

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +168,57 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
...
+func getVersionDiff(v1, v2 []core.ExpirableVersion) map[string]int {
+	v2Versions := sets.String{}
+	for _, x := range v2 {
+		v2Versions.Insert(x.Version)
+	}
+	diff := map[string]int{}
+	for index, x := range v1 {
+		if found := v2Versions.Has(x.Version); !found {
```
timuthy:
```suggestion
		if !v2Versions.Has(x.Version) {
```

pkg/apis/core/types_cloudprofile.go
```diff
@@ -150,3 +152,19 @@ const (
 	// VolumeClassPremium is a constant for the premium volume class.
 	VolumeClassPremium string = "premium"
 )
+
+// VersionClassification is the logical state of a version according to https://github.com/gardener/gardener/blob/master/docs/proposals/05-versioning-policy.md
```
timebertt: Can you please use a link to the documentation you wrote instead of the Proposal?
```suggestion
// VersionClassification is the logical state of a version according to https://github.com/gardener/gardener/blob/master/docs/operations/versioning.md
```

pkg/apis/core/types_cloudprofile.go
```diff
@@ -150,3 +152,19 @@ const (
...
+type VersionClassification string
+
+const (
+	// ClassificationPreview indicates that a version has recently been added and not promoted to "Supported" yet.
+	// ClassificationPreview versions will not be considered for automatic Kubernetes and Machine Image patch version updates.
+	ClassificationPreview VersionClassification = "preview"
+	// ClassificationSupported indicates that a patch version is the recommended version for a shoot.
+	// Using VersionMaintenance (see: https://github.com/gardener/gardener/docs/operation/versioning.md) there is one supported version per maintained minor version.
```
timebertt, line 164:
```suggestion
	// Using VersionMaintenance (see: https://github.com/gardener/gardener/blob/master/docs/operation/versioning.md) there is one supported version per maintained minor version.
```

timebertt: Also, https://github.com/gardener/gardener/pull/2105#discussion_r400273690 is not addressed.
Please also mention Machine Images here, as the version policy will also be applied to them.

pkg/apis/core/v1alpha1/types_cloudprofile.go
timebertt, line 190: Please adapt this version accordingly to the internal version

pkg/apis/core/v1beta1/types_cloudprofile.go
timebertt, line 190: Please adapt this version accordingly to the internal version

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -44,11 +45,50 @@ func ValidateCloudProfileUpdate(newProfile, oldProfile *core.CloudProfile) field
...
+				allErrs = append(allErrs, validateCloudProfileVersionsUpdate(newImage.Versions, oldImage.Versions, fldPath.Child("machineImages").Index(index).Child("versions"))...)
+			}
+		}
+	}
+
+	return allErrs
+}
+
+// ValidateCloudProfileAddedVersions validates versions added to the CloudProfile
+func ValidateCloudProfileAddedVersions(versions []core.ExpirableVersion, addedVersions map[string]int, fldPath *field.Path) field.ErrorList {
```
timuthy, line 105: At the moment this validation is only done in case of an update isn't it also relevant when you create a `CloudProfile`?

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -100,6 +140,33 @@ func validateKubernetesSettings(kubernetes core.KubernetesSettings, fldPath *fie
...
+		if err != nil {
+			// check is already performed by caller, avoid duplicate error
+			return allErrs
+		}
+
+		filteredVersions := helper.FindVersionsWithSameMajorMinor(helper.FilterVersionsWithClassification(versions, core.ClassificationSupported), *currentSemVer)
+
+		// do not allow adding multiple supported versions per minor version
+		if len(filteredVersions) > 0 {
+			allErrs = append(allErrs, field.Forbidden(fldPath, fmt.Sprintf("unable to add version %q with classification %q. Only one %q version is allowed  per minor version", version.Version, core.ClassificationSupported, core.ClassificationSupported)))
```
timebertt:
```suggestion
			allErrs = append(allErrs, field.Forbidden(fldPath, fmt.Sprintf("unable to add version %q with classification %q. Only one %q version is allowed per minor version", version.Version, core.ClassificationSupported, core.ClassificationSupported)))
```

plugin/pkg/shoot/validator/admission.go
```diff
@@ -508,11 +508,18 @@ func validateKubernetesVersionConstraints(constraints []core.ExpirableVersion, s
 	validValues := []string{}
```
timebertt: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices states:

> the nil slice is the preferred style

Please use a nil slice here instead of initializing a zero-length slice.

```suggestion
	var validValues []string
```

plugin/pkg/shoot/validator/admission.go
```diff
@@ -508,11 +508,18 @@ func validateKubernetesVersionConstraints(constraints []core.ExpirableVersion, s
...
 	var latestVersion *semver.Version
 	for _, versionConstraint := range constraints {
-		if versionConstraint.ExpirationDate != nil && versionConstraint.ExpirationDate.Time.Before(time.Now()) {
+		if versionConstraint.ExpirationDate != nil && versionConstraint.ExpirationDate.Time.UTC().Before(time.Now().UTC()) {
 			continue
```
timebertt, line 511: Can we return an error, if the Shoot's used k8s version is expired? Or are we already doing that elsewhere?

plugin/pkg/shoot/validator/admission.go
```diff
@@ -508,11 +508,18 @@ func validateKubernetesVersionConstraints(constraints []core.ExpirableVersion, s
...
+		if !getLatestPatchVersion {
+			validValues = append(validValues, versionConstraint.Version)
+		} else {
+			// ignore preview versions for defaulting
+			if versionConstraint.Classification != nil && *versionConstraint.Classification == core.ClassificationPreview {
+				continue
+			}
+		}
 
 		if versionConstraint.Version == shootVersion {
```
timebertt, line 522: Can you move this if block up, before the previous one? If I see this correctly, the Shoot won't be able to use a Preview version otherwise...

plugin/pkg/shoot/validator/admission.go
```diff
@@ -508,11 +508,18 @@ func validateKubernetesVersionConstraints(constraints []core.ExpirableVersion, s
...
+		if versionConstraint.ExpirationDate != nil && versionConstraint.ExpirationDate.Time.UTC().Before(time.Now().UTC()) {
 			continue
 		}
 
-		validValues = append(validValues, versionConstraint.Version)
+		if !getLatestPatchVersion {
+			validValues = append(validValues, versionConstraint.Version)
+		} else {
+			// ignore preview versions for defaulting
+			if versionConstraint.Classification != nil && *versionConstraint.Classification == core.ClassificationPreview {
```
timuthy: Can you explain why you want to ignore preview versions? Let's suppose `1.18.0` is a `preview` version and a shoot is created with `1.18`. Why wouldn't you want to default it to `1.18.0`? Instead the user only gets non 1.18 versions in the error message.

timebertt (changes requested):
Thanks for addressing my comments, looking pretty good now.
Only some nits left :)

timuthy (changes requested):
Thanks for addressing the feedback. I found some leftovers and have a few questions.

plugin/pkg/shoot/validator/admission.go
danielfoehrKn: I see what your are saying, though I think that is not related to this PR. It would require some rework in the shoot admission. Though it is not a major effort, I would like to have that in a dedicated  PR.

plugin/pkg/shoot/validator/admission.go
danielfoehrKn: The Shoot can use a preview version. I wrote a new test for it in the admission_test.
Preview versions are only filtered here when the Shoot only provides a major.minor version like 1.6

pkg/apis/core/helper/helpers.go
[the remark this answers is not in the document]
danielfoehrKn: Pointed out by @tim-ebert , thanks though - adjusted it!

pkg/apis/core/validation/cloudprofile.go
danielfoehrKn: To my understanding, If this check is  in the Spec validation, than that would block any  update of the CloudProfile when containing an expired version. I do not think forcing the CloudProfile to be expired version free is operator friendly. 
Rather I would only validate that when adding new version to an existing profile. 

I think that the underlying problem is that,  when updating the CloudProfile the [same checks as when initially creating the CloudProfile are also performed](https://github.com/gardener/gardener/blob/90b9eb1905851dadd3713d2e65bfbe7b5717a03f/pkg/apis/core/validation/cloudprofile.go#L49).

plugin/pkg/shoot/validator/admission.go
danielfoehrKn: Because that is the nature of preview versions (also agreed upon with @vlerenc ) - they should be opt-in. They should not be defaulted to. If they want to have a preview version, even if there are only preview version for the minor version, they should make a concious effort and set the specific version.
In my opinion we should not default the stakeholders cluster to a version that is not thoroughly tested.
Gardener should default to stable versions and tested versions - we call them "supported".

This defaulting feature was to my knowledge created from the need of staekholders to create up to date clusters from their CI pipelines to test their workload on. Providing them with preview clusters is IMHO not helpful in this context.

pkg/apis/core/v1alpha1/types_cloudprofile.go
danielfoehrKn: Please check if the doc strings are the way you suggested it - thanks!

plugin/pkg/shoot/validator/admission.go
timebertt: Fine for me 👍 Can you open an issue to track that?

pkg/apis/core/v1alpha1/types_cloudprofile.go
timebertt: Thanks, the doc strings look good to me now 👍

plugin/pkg/shoot/validator/admission.go
timebertt: Oh yes, sorry.

timebertt (commented):
Thanks for addressing the comments.
PR is looking completely good for me now :)
Except that I am not completely sure what to think about the part with the preview version defaulting.

plugin/pkg/shoot/validator/admission.go
timuthy: Okay, let's say it this way: `1.18.0` is in `preview` and it's also the latest release you offer. `1.17.3` is another version offered.

As far as I see the logic you'll get the following results:
1. Create a shoot with `1.18` --> Error message that `1.17.3` is a valid version.
2. Create a shoot with `1.19.1` --> Error message that `1.17.3` and `1.18.0` are valid versions.

Can you double check that pls? And if that's the case, it's a little strange, isn't it?

pkg/apis/core/validation/cloudprofile.go
timuthy: > I do not think forcing the CloudProfile to be expired version free is operator friendly.

I don't understand why it's operator friendly to allow specifying expired version during creation whereas it's not for updates (or the other way around). What's the use case to add an already expired version through creation and what's the motivation to prevent it thru updates?

plugin/pkg/shoot/validator/admission.go
danielfoehrKn: At the moment it is intended to be exactly how you described in your example. 
I do not think that this is a strange behaviour.

1. 1.18 ask Gardener to default the version for the Shoot. Gardener does not default to preview versions -> 1.17.3 is the latest version it can default to.
2. 1.19.1 does not use defaulting, so preview versions can be selected on an opt-in basis. Thats why giving the valid values of 1.18.0 and 1.17.3 is correct in this context.

@rfranzke maybe you have an opinion here as well

pkg/apis/core/validation/cloudprofile.go
danielfoehrKn: > I think that the underlying problem is that, when updating the CloudProfile the same checks as when initially creating the CloudProfile are also performed.

The problem is, the way the validation **is designed at the moment:**
- Create: ValidateCloudProfileSpec
- Update: ValidateCloudProfileSpecUpdate() & ValidateCloudProfileSpec()

So if I want to introduce a check during CloudProfile Create() I have to check that in the ValidateCloudProfileSpec(). As newly created version do not have previous field values,  I would have to introduce a check that no version can have an expiration date in the past. 

**The problem:** A version X has an expiration date Y. Time passes after Y and the operator wants to make a change to the CloudProfile completely unrelated to the versions. That will not work as the validation will throw an error that the CloudProfile is not valid. The operator has to fix the expired version first before being able to perform his intendet CloudProfile update. Thats what I meant with "not operator friendly".

To sum it up: The way the validation is currently designed is in my opinion causing problems here. There is a coupling between the update validation logic of the CloudProfile and the Creation validation logic.

I propose changing the validation logic by introducing a separate function ValidateCloudProfileCreation() [that calls the current validation function  ](https://github.com/gardener/gardener/blob/90b9eb1905851dadd3713d2e65bfbe7b5717a03f/pkg/apis/core/validation/cloudprofile.go#L34).
This way we can validate in the ValidateCloudProfileCreation() that no version has an expiration date in the past and in the Update we do the same for added versions.
I ll update the PR accordingly.

timuthy: Sounds good 👍

danielfoehrKn: I updated it - [PTAL here](https://github.com/gardener/gardener/blob/0921db6fe993a98c2eff652588ded6fedabf7de7/pkg/apis/core/validation/cloudprofile.go#L34) and [here](https://github.com/gardener/gardener/blob/0921db6fe993a98c2eff652588ded6fedabf7de7/pkg/apis/core/validation/cloudprofile_test.go#L697)

plugin/pkg/shoot/validator/admission.go
danielfoehrKn: Please see [this issue.](https://github.com/gardener/gardener/issues/2136)

plugin/pkg/shoot/validator/admission.go
timuthy: > 1.18 ask Gardener to default the version for the Shoot.

I'm not sure if users really use Gardener to indirectly ask what patch version it's capable of defaulting to. Imo, this is a convenience for the user and if we can't default a version why not showing all the versions the user can pick for his cluster (like done in the 2nd scenario). Easier for the user to understand and easier to maintain.

But sure, let's wait for other opinions or have an offline discussion :)

pkg/apis/core/validation/cloudprofile.go
rfranzke: Hm, but what if we want to phase out a minor versions, 1.10 for example? We would only specify deprecated versions, no supported ones. Would that work?

plugin/pkg/shoot/validator/admission.go
rfranzke: I tend to agree with @timuthy - we should a) make the error messages clear and helpful, and b) prevent to implement any hard-to-maintain special cases. So, it's fine to not default to 1.18.0 if it's the only version and it's not yet supported/still in preview, but then let's show this in the error message ("supported versions are 1.18.0 (preview, must be selected explicitly) and 1.17.3").

danielfoehrKn: Ok, I think that makes sense, this way the user knows what is going on. I ll adjust it.
Thanks for fir reviewing so thoroughly!

pkg/apis/core/validation/cloudprofile.go
danielfoehrKn: When using no VersionManagement or VersionManagement, the operator can specify the versions without restrictions.

Using VersionManagement, the reconcile enforces one supported version independent of the minor version.

Using also VersionMaintenance, the reconcile enforces one supported version per maintained minor version. That means you cannot phase out maintained minor versions. However the number of maintained minor versions can be adjusted on the fly in order to make any minor version unmaintained. 
Does that create problems?

rfranzke: > Using VersionManagement, the reconcile enforces one supported version independent of the minor version.

How do you phase out a version with this restriction?

danielfoehrKn: You can phase out any version using VersionManagement? As long as it is not the latest non-preview patch version of the latest minor version.

rfranzke: I don't get it. If we offer 1.10 and 1.11 and we don't want to offer 1.10 anymore, how do we do it? You are saying that you enforce at least one supported version. However, what we want is to set an expiration date to every patch version of 1.10 so that we can delete it from the `CloudProfile` after all expiration dates are expired.

plugin/pkg/shoot/validator/admission.go
danielfoehrKn: Added a separate commit containing the improved error message. PTAL

pkg/apis/core/validation/cloudprofile.go
danielfoehrKn: Given you example, the  CloudProfile only contains minor versions 1.10 and 1.11. We want to expire all version from minor 1.10. 

Scenario 
1) using VersionManagement.
All but the latest patch version from 1.10 are deprecated automatically.  The operator needs to explicitly set an expiration date to the latest patch version as a safety measure.

2) VersionMaintenance. 
- Need to adjust the configuration of the CloudProfile controller so that 1.10 is an unmaintained minor version. 
- then it is the same as VersionManagement.

Is that more clear now?

rfranzke: Not really. Again, you said "Using VersionManagement, the reconcile enforces one supported version independent of the minor version." (see https://github.com/gardener/gardener/pull/2105#discussion_r403992734). It contradicts to 100% the first point.

rfranzke: OK, after a call with @danielfoehrKn it became clear that the statement above does **not** mean that there has to be one supported versions **per** minor version. Instead, there must be one supported version in all the versions.

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +168,57 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
 	}
 	return nil
 }
+
+// GetRemovedVersions finds versions that have been removed in the old compared to the new version slice.
+// returns a map associating the version with its index in the in the old version slice.
+func GetRemovedVersions(old []core.ExpirableVersion, new []core.ExpirableVersion) map[string]int {
```
rfranzke:
```suggestion
func GetRemovedVersions(old, new []core.ExpirableVersion) map[string]int {
```

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +168,57 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
...
+	return getVersionDiff(old, new)
+}
+
+// GetAddedVersions finds versions that have been added in the new compared to the new version slice.
+// returns a map associating the version with its index in the in the old version slice.
+func GetAddedVersions(old []core.ExpirableVersion, new []core.ExpirableVersion) map[string]int {
```
rfranzke:
```suggestion
func GetAddedVersions(old, new []core.ExpirableVersion) map[string]int {
```

pkg/apis/core/helper/helpers.go
```diff
@@ -167,3 +168,57 @@ func FindWorkerByName(workers []core.Worker, name string) *core.Worker {
...
+	}
+	return result
+}
+
+// FindVersionsWithSameMajorMinor filters the given versions slice for versions other the given one, having the same major and minor version as the given version
+func FindVersionsWithSameMajorMinor(versions []core.ExpirableVersion, version semver.Version) []core.ExpirableVersion {
+	var result []core.ExpirableVersion
+	for _, v := range versions {
+		// semantic version already checked by validator
+		semVer, _ := semver.NewVersion(v.Version)
```
rfranzke: Why is the error ignored?

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -100,6 +170,33 @@ func validateKubernetesSettings(kubernetes core.KubernetesSettings, fldPath *fie
...
 		}
+		allErrs = append(allErrs, validateExpirableVersion(kubernetes.Versions, version, idxPath)...)
+	}
+
+	return allErrs
+}
+
+var supportedVersionClassifications = sets.NewString(string(core.ClassificationPreview), string(core.ClassificationSupported), string(core.ClassificationDeprecated))
+
+func validateExpirableVersion(versions []core.ExpirableVersion, version core.ExpirableVersion, fldPath *field.Path) field.ErrorList {
```
rfranzke:
```suggestion
func validateExpirableVersion(version core.ExpirableVersion, allVersions []core.ExpirableVersion, fldPath *field.Path) field.ErrorList {
```

plugin/pkg/shoot/validator/admission.go
```diff
@@ -385,9 +385,13 @@ func validateProvider(c *validationContext) field.ErrorList {
 		)...)
 	}
 
-	ok, validKubernetesVersions, versionDefault := validateKubernetesVersionConstraints(c.cloudProfile.Spec.Kubernetes.Versions, c.shoot.Spec.Kubernetes.Version, c.oldShoot.Spec.Kubernetes.Version)
+	ok, isDefaulted, validKubernetesVersions, versionDefault := validateKubernetesVersionConstraints(c.cloudProfile.Spec.Kubernetes.Versions, c.shoot.Spec.Kubernetes.Version, c.oldShoot.Spec.Kubernetes.Version)
 	if !ok {
-		allErrs = append(allErrs, field.NotSupported(field.NewPath("spec", "kubernetes", "version"), c.shoot.Spec.Kubernetes.Version, validKubernetesVersions))
+		err := field.NotSupported(field.NewPath("spec", "kubernetes", "version"), c.shoot.Spec.Kubernetes.Version, validKubernetesVersions)
+		if isDefaulted {
+			err.Detail = fmt.Sprintf("unable to default version. No higher non-preview version found. In case there is a higher preview version available for that minor, it has to be selected explicitly -  %s", err.Detail)
```
rfranzke: What does "no higher non-preview version" mean? I'm confused by the word "higher" ?!

rfranzke (changes requested):
Looks great, minor suggestions. :)

plugin/pkg/shoot/validator/admission.go
danielfoehrKn: meaning there is no version x > version to default.  
WDYT about "no greater non-preview version"

pkg/apis/core/helper/helpers.go
danielfoehrKn: Initially I removed this error check after a suggestion, but looking over it again, you are right. Even though the semVer check is being done for mache images (pkg/apis/core/validation/cloudprofile.go#L285-L284) and Kubernetes versions (pkg/apis/core/validation/cloudprofile.go#L163-L162) already, the error can still happen and should be checked. In addition the helper function cannot assume, that it only receives valid semantic versions. 

Thanks

plugin/pkg/shoot/validator/admission.go
rfranzke: Hm, maybe I'm missing something. We are defaulting the Kubernetes version when somebody is using `1.18`, for example. We try to find the highest patch version of `1.18` that is `supported`, right?
Hence, if we can't find such a version, wouldn't a message like "unable to default version - couldn't find a patch version for 1.18 with \"supported\" classification - in case there is a \"preview\"-classified version please select it explicitly" be more accurate?

danielfoehrKn: It only looks for the highest non-preview version with the same major.minor. That can be a supported version or be deprecated.

rfranzke: OK, understood. Then:

"unable to default version - couldn't find a suitable patch version for 1.18 (either \"supported\" or \"deprecated\" classification and a non-expired expiration date) - in case there is a \"preview\"-classified version please select it explicitly"

danielfoehrKn: Actually a version can also have no classification - can we leave the supported and deprecated out? 

"unable to default version - couldn't find a suitable patch version for 1.18. Suitable patch versions have a non-expired expiration date and are no "preview" versions. "Preview"-classified versions have to be selected explicitly"

rfranzke: sounds good!

danielfoehrKn: PTAL

rfranzke (dismissed):
/lgtm

timuthy (dismissed):

timebertt (dismissed):
Thank, looks completely good for me now 👍

pkg/mock/client-go/discovery/mocks.go
```diff
@@ -6,7 +6,7 @@ package discovery
 
 import (
 	gomock "github.com/golang/mock/gomock"
-	openapi_v2 "github.com/googleapis/gnostic/OpenAPIv2"
+	OpenAPIv2 "github.com/googleapis/gnostic/OpenAPIv2"
```
timebertt: Please revert this change, this file is generated and CI fails here.

danielfoehrKn: Sorry, mockgen was outdated

rfranzke (dismissed):

PR #2106, opened 2020-03-26
[GEP-5]  do not allow deletion of used Kubernetes and OS versions

### LINKED ISSUE #1577: Implement versioning policy contract (GEP-5)

danielfoehrKn:
This PR is part of GEP5 .
- Do not allow the deletion of a version from the CloudProfile which is in use by any Shoot.

Part of #1577 

✅ Depends on #2105

**Release note**:

```noteworthy operator
The Gardener API server now blocks the removal of Kubernetes and machine image versions from the `CloudProfile` which are still in use by shoot clusters. 
```

### DIFF
--- a/plugin/pkg/global/resourcereferencemanager/admission.go
+++ b/plugin/pkg/global/resourcereferencemanager/admission.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sync"
 	"time"
 
 	"github.com/gardener/gardener/pkg/apis/core"
@@ -30,10 +31,12 @@ import (
 	"github.com/gardener/gardener/pkg/operation/common"
 	"github.com/gardener/gardener/plugin/pkg/utils"
 
+	"github.com/hashicorp/go-multierror"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/admission"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 	kubeinformers "k8s.io/client-go/informers"
@@ -317,6 +320,94 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o 
 						core.ProjectMemberOwner,
 					},
 				})
+			}
+		}
+	case core.Kind("CloudProfile"):
+		cloudProfile, ok := a.GetObject().(*core.CloudProfile)
+		if !ok {
+			return apierrors.NewBadRequest("could not convert resource into CloudProfile object")
+		}
+		if utils.SkipVerification(operation, cloudProfile.ObjectMeta) {
+			return nil
+		}
+		if a.GetOperation() == admission.Update {
+			oldCloudProfile, ok := a.GetOldObject().(*core.CloudProfile)
+			if !ok {
+				return apierrors.NewBadRequest("could not convert old resource into CloudProfile object")
+			}
+
+			// getting Kubernetes versions that have been removed from the CloudProfile
+			removedKubernetesVersions := sets.StringKeySet(helper.GetRemovedVersions(oldCloudProfile.Spec.Kubernetes.Versions, cloudProfile.Spec.Kubernetes.Versions))
+
+			// getting Machine image versions that have been removed from the CloudProfile
+			removedMachineImageVersions := map[string]sets.String{}
+			for _, oldImage := range oldCloudProfile.Spec.MachineImages {
+				imageFound := false
+				for _, newImage := range cloudProfile.Spec.MachineImages {
+					if oldImage.Name == newImage.Name {
+						imageFound = true
+						removedMachineImageVersions[oldImage.Name] = sets.StringKeySet(helper.GetRemovedVersions(oldImage.Versions, newImage.Versions))
+					}
+				}
+				if !imageFound {
+					for _, version := range oldImage.Versions {
+						if removedMachineImageVersions[oldImage.Name] == nil {
+							removedMachineImageVersions[oldImage.Name] = sets.NewString()
+						}
+						removedMachineImageVersions[oldImage.Name] = removedMachineImageVersions[oldImage.Name].Insert(version.Version)
+					}
+				}
+			}
+
+			if len(removedKubernetesVersions) > 0 || len(removedMachineImageVersions) > 0 {
+				shootList, err1 := r.shootLister.List(labels.Everything())
+				if err1 != nil {
+					return apierrors.NewInternalError(fmt.Errorf("could not list shoots to verify that Kubernetes and/or Machine image version can be removed: %v", err1))
+				}
+
+				var (
+					channel = make(chan error)
+					wg      sync.WaitGroup
+				)
+				wg.Add(len(shootList))
+
+				for _, s := range shootList {
+					if s.Spec.CloudProfileName != cloudProfile.Name {
+						wg.Done()
+						continue
+					}
+
+					go func(shoot *core.Shoot) {
+						defer wg.Done()
+
+						if removedKubernetesVersions.Has(shoot.Spec.Kubernetes.Version) {
+							channel <- fmt.Errorf("unable to delete Kubernetes version %q from CloudProfile %q - version is still in use by shoot '%s/%s'", shoot.Spec.Kubernetes.Version, shoot.Spec.CloudProfileName, shoot.Namespace, shoot.Name)
+						}
+						for _, worker := range shoot.Spec.Provider.Workers {
+							if worker.Machine.Image == nil {
+								continue
+							}
+							// happens if Shoot runs an image that does not exist in the old CloudProfile - in this case: ignore
+							if _, ok := removedMachineImageVersions[worker.Machine.Image.Name]; !ok {
+								continue
+							}
+
+							if removedMachineImageVersions[worker.Machine.Image.Name].Has(worker.Machine.Image.Version) {
+								channel <- fmt.Errorf("unable to delete Machine image version '%s/%s' from CloudProfile %q - version is still in use by shoot '%s/%s' by worker %q", worker.Machine.Image.Name, worker.Machine.Image.Version, shoot.Spec.CloudProfileName, shoot.Namespace, shoot.Name, worker.Name)
+							}
+						}
+					}(s)
+				}
+
+				// close channel when wait group has 0 counter
+				go func() {
+					wg.Wait()
+					close(channel)
+				}()
+
+				for channelResult := range channel {
+					err = multierror.Append(err, channelResult)
+				}
 			}
 		}
 	}
--- a/plugin/pkg/global/resourcereferencemanager/admission_test.go
+++ b/plugin/pkg/global/resourcereferencemanager/admission_test.go
@@ -684,5 +684,318 @@ var _ = Describe("resourcereferencemanager", func() {
 				})))
 			})
 		})
+
+		Context("CloudProfile - Update Kubernetes versions", func() {
+			versions := []core.ExpirableVersion{
+				{Version: "1.17.2"},
+				{Version: "1.17.1"},
+				{Version: "1.17.0"},
+			}
+			shootOne := shootBase
+			shootOne.Name = "shoot-One"
+			shootOne.Spec.Provider.Type = "aws"
+			shootOne.Spec.CloudProfileName = "aws-profile"
+			shootOne.Spec.Kubernetes.Version = "1.17.2"
+
+			shootTwo := shootOne
+			shootTwo.Name = "shoot-Two"
+			shootTwo.Spec.Kubernetes.Version = "1.17.1"
+			var (
+				cloudProfile = core.CloudProfile{
+					ObjectMeta: metav1.ObjectMeta{Name: "aws-profile"},
+					Spec: core.CloudProfileSpec{
+						Kubernetes: core.KubernetesSettings{
+							Versions: versions,
+						},
+					},
+				}
+			)
+			It("should accept if no kubernetes version has been removed", func() {
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootOne)).To(Succeed())
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootTwo)).To(Succeed())
+
+				attrs := admission.NewAttributesRecord(&cloudProfile, &cloudProfile, core.Kind("CloudProfile").WithVersion("version"), "", cloudProfile.Name, core.Resource("CloudProfile").WithVersion("version"), "", admission.Update, &metav1.UpdateOptions{}, false, defaultUserInfo)
+
+				err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			It("should accept removal of kubernetes version that is not in use by any shoot", func() {
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootOne)).To(Succeed())
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootTwo)).To(Succeed())
+
+				cloudProfileNew := cloudProfile
+				cloudProfileNew.Spec = core.CloudProfileSpec{
+					Kubernetes: core.KubernetesSettings{
+						Versions: []core.ExpirableVersion{
+							{Version: "1.17.2"},
+							{Version: "1.17.1"},
+						},
+					},
+				}
+
+				attrs := admission.NewAttributesRecord(&cloudProfileNew, &cloudProfile, core.Kind("CloudProfile").WithVersion("version"), "", cloudProfile.Name, core.Resource("CloudProfile").WithVersion("version"), "", admission.Update, &metav1.UpdateOptions{}, false, defaultUserInfo)
+
+				err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			It("should reject removal of kubernetes versions that are still in use by shoots", func() {
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootOne)).To(Succeed())
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootTwo)).To(Succeed())
+
+				cloudProfileNew := cloudProfile
+				cloudProfileNew.Spec = core.CloudProfileSpec{
+					Kubernetes: core.KubernetesSettings{
+						Versions: []core.ExpirableVersion{
+							{Version: "1.17.2"},
+						},
+					},
+				}
+
+				attrs := admission.NewAttributesRecord(&cloudProfileNew, &cloudProfile, core.Kind("CloudProfile").WithVersion("version"), "", cloudProfile.Name, core.Resource("CloudProfile").WithVersion("version"), "", admission.Update, &metav1.UpdateOptions{}, false, defaultUserInfo)
+
+				err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("1.17.1"))
+			})
+		})
+
+		Context("CloudProfile - Update Machine image versions", func() {
+			versions := []core.ExpirableVersion{
+				{Version: "1.17.3"},
+				{Version: "1.17.2"},
+				{Version: "1.17.1"},
+				{Version: "1.17.0"},
+				{Version: "1.16.0"},
+			}
+			shootOne := shootBase
+			shootOne.Name = "shoot-One"
+			shootOne.Spec.Provider.Type = "aws"
+			shootOne.Spec.CloudProfileName = "aws-profile"
+			shootOne.Spec.Provider.Workers = []core.Worker{
+				{
+					Name: "coreos-worker",
+					Machine: core.Machine{
+						Image: &core.ShootMachineImage{
+							Name:    "coreos",
+							Version: "1.17.3",
+						},
+					},
+				},
+			}
+
+			shootTwo := shootOne
+			shootTwo.Name = "shoot-Two"
+			shootTwo.Spec.Provider.Workers = []core.Worker{
+				{
+					Name: "ubuntu-worker-1",
+					Machine: core.Machine{
+						Image: &core.ShootMachineImage{
+							Name:    "ubuntu",
+							Version: "1.17.2",
+						},
+					},
+				},
+				{
+					Name: "ubuntu-worker-2",
+					Machine: core.Machine{
+						Image: &core.ShootMachineImage{
+							Name:    "ubuntu",
+							Version: "1.17.1",
+						},
+					},
+				},
+			}
+
+			var (
+				cloudProfile = core.CloudProfile{
+					ObjectMeta: metav1.ObjectMeta{Name: "aws-profile"},
+					Spec: core.CloudProfileSpec{
+						MachineImages: []core.MachineImage{
+							{
+								Name:     "coreos",
+								Versions: versions,
+							},
+							{
+								Name:     "ubuntu",
+								Versions: versions,
+							},
+						},
+					},
+				}
+			)
+			It("should accept if no machine image version has been removed", func() {
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootOne)).To(Succeed())
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootTwo)).To(Succeed())
+
+				attrs := admission.NewAttributesRecord(&cloudProfile, &cloudProfile, core.Kind("CloudProfile").WithVersion("version"), "", cloudProfile.Name, core.Resource("CloudProfile").WithVersion("version"), "", admission.Update, &metav1.UpdateOptions{}, false, defaultUserInfo)
+
+				err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+				Expect(err).NotTo(HaveOccurred())
+			})
+			It("should accept removal of a machine version that is not in use by any shoot", func() {
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootOne)).To(Succeed())
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootTwo)).To(Succeed())
+
+				newVersions := []core.ExpirableVersion{
+					{Version: "1.17.3"},
+					{Version: "1.17.2"},
+					{Version: "1.17.1"},
+				}
+
+				// new cloud profile has version 1.17.0 and 1.16.0 removed. These are not in use by any worker of any shoot
+				cloudProfileNew := cloudProfile
+				cloudProfileNew.Spec = core.CloudProfileSpec{
+					MachineImages: []core.MachineImage{
+						{
+							Name:     "coreos",
+							Versions: newVersions,
+						},
+						{
+							Name:     "ubuntu",
+							Versions: newVersions,
+						},
+					},
+				}
+
+				attrs := admission.NewAttributesRecord(&cloudProfileNew, &cloudProfile, core.Kind("CloudProfile").WithVersion("version"), "", cloudProfile.Name, core.Resource("CloudProfile").WithVersion("version"), "", admission.Update, &metav1.UpdateOptions{}, false, defaultUserInfo)
+
+				err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			It("should reject removal of a machine image version that is in use by a shoot", func() {
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootOne)).To(Succeed())
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Shoots().Informer().GetStore().Add(&shootTwo)).To(Succeed())
+
+				newVersions := []core.ExpirableVersion{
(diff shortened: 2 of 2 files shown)


rfranzke:

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -481,3 +546,12 @@ func (r *ReferenceManager) lookupSecret(namespace, name string) error {
 
 	return nil
 }
+
+func getRemovedVersions(oldVersions []core.ExpirableVersion, newVersions []core.ExpirableVersion) sets.String {
```
rfranzke: Why don't you reuse your helper function? https://github.com/gardener/gardener/pull/2105/files#diff-0a71cc18fe106e50d1a5aca2b5330aafR173

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,68 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+		if a.GetOperation() == admission.Update {
+			oldCloudProfile, ok := a.GetOldObject().(*core.CloudProfile)
+			if !ok {
+				return apierrors.NewBadRequest("could not convert old resource into CloudProfile object")
+			}
+
+			// getting Kubernetes versions that have been removed from the CloudProfile
+			removedKubernetesVersions := getRemovedVersions(oldCloudProfile.Spec.Kubernetes.Versions, cloudProfile.Spec.Kubernetes.Versions)
+
+			// getting Machine image versions that have been removed from the CloudProfile
```
rfranzke, line 342: Same here - why not reusing your helper function? https://github.com/gardener/gardener/pull/2105/files#diff-0a71cc18fe106e50d1a5aca2b5330aafR173

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,68 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+						if removedMachineImageVersions[oldImage.Name] == nil {
+							removedMachineImageVersions[oldImage.Name] = sets.NewString()
+						}
+						removedMachineImageVersions[oldImage.Name] = removedMachineImageVersions[oldImage.Name].Insert(version.Version)
+					}
+				}
+			}
+
+			if len(removedKubernetesVersions) > 0 || len(removedMachineImageVersions) > 0 {
+				shootList, e := r.shootLister.List(labels.Everything())
```
rfranzke:
```suggestion
				shootList, err := r.shootLister.List(labels.Everything())
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,68 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+				if e != nil {
+					return err
```
rfranzke: ❓

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,68 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+				}
+				for _, shoot := range shootList {
+					if shoot.Spec.CloudProfileName != cloudProfile.Name {
+						continue
+					}
+
+					if removedKubernetesVersions.Has(shoot.Spec.Kubernetes.Version) {
+						err = multierror.Append(err, fmt.Errorf("unable to delete Kubernetes version '%s' from CloudProfile. Version is still in use by a Shoot (Name: '%s', Namespace: '%s', CloudProfile: '%s')", shoot.Spec.Kubernetes.Version, shoot.Name, shoot.Namespace, shoot.Spec.CloudProfileName))
```
rfranzke:
```suggestion
						err = multierror.Append(err, fmt.Errorf("unable to delete Kubernetes version %q from CloudProfile - version is still in use by shoot %s/%s", shoot.Spec.Kubernetes.Version, shoot.Namespace, shoot.Name))
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,68 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+					}
+					for _, worker := range shoot.Spec.Provider.Workers {
+						// happens if Shoot already runs a version that does not exist in the CloudProfile - in this case: ignore
+						if _, ok := removedMachineImageVersions[worker.Machine.Image.Name]; !ok {
```
rfranzke: `worker.Machine.Image` is a pointer - `nil` check is missing

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,68 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+							continue
+						}
+						if removedMachineImageVersions[worker.Machine.Image.Name].Has(worker.Machine.Image.Version) {
+							err = multierror.Append(err, fmt.Errorf("unable to delete Machine image version (%s-%s) from CloudProfile. Version is still in use by a Shoot (Shoot Name: '%s', Namespace: '%s', Worker: '%s', CloudProfile: '%s')", worker.Machine.Image.Name, worker.Machine.Image.Version, shoot.Name, shoot.Namespace, worker.Name, shoot.Spec.CloudProfileName))
```
rfranzke:
```suggestion
							err = multierror.Append(err, fmt.Errorf("unable to delete Machine image version %s/%s from CloudProfile - version is still in use by shoot %s/%s", worker.Machine.Image.Name, worker.Machine.Image.Version, shoot.Namespace, shoot.Name))
```

rfranzke (changes requested):

plugin/pkg/global/resourcereferencemanager/admission.go
danielfoehrKn: Thats whats being done one line below? 
[As mentioned here](https://github.com/gardener/gardener/pull/2105#discussion_r399186975) I will use the returned set.Set from the updated helper to get rid of the then redundant code below.

plugin/pkg/global/resourcereferencemanager/admission.go
danielfoehrKn: see other comment

plugin/pkg/global/resourcereferencemanager/admission.go
danielfoehrKn: return BadRequest() instead. The reason why I just returned the error is because this is [how it is done here](https://github.com/gardener/gardener/blob/2c33877a1315fb2d655cb5a3fc93561767a457fe/plugin/pkg/global/resourcereferencemanager/admission.go#L229-L228)

plugin/pkg/global/resourcereferencemanager/admission.go
danielfoehrKn: adjusted it, however kept the CloudProfile name, because I think that is valuable information when editing multiple CloudProfiles with one patch.

plugin/pkg/global/resourcereferencemanager/admission.go
danielfoehrKn: right, the defaulting of the ShootMachineImage  in the shootAdmissionPlugin happens after - thanks!

plugin/pkg/global/resourcereferencemanager/admission.go
danielfoehrKn: the multierror is already called err, this would override it / not work.
Called the error now err1 to make it more understandable.

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,71 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+						}
+						removedMachineImageVersions[oldImage.Name] = removedMachineImageVersions[oldImage.Name].Insert(version.Version)
+					}
+				}
+			}
+
+			if len(removedKubernetesVersions) > 0 || len(removedMachineImageVersions) > 0 {
+				shootList, err1 := r.shootLister.List(labels.Everything())
+				if err1 != nil {
+					return apierrors.NewBadRequest("could not list shoots to verify Kubernetes and Machine image version")
```
rfranzke: Also return the `err1` here/its error message

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,71 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+				}
+				for _, shoot := range shootList {
```
rfranzke: As this can become quite large can you please parallelize the body of the for loop to speed up the computations?

rfranzke (changes requested):
Looks good overall! Nice :)

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +323,98 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+
+				var (
+					channel = make(chan error)
+					wg      sync.WaitGroup
+				)
+				wg.Add(len(shootList))
+
+				for _, s := range shootList {
+					shoot := s.DeepCopy()
+					go func(ctx context.Context, removedKubernetesVersions sets.String, removedMachineImageVersions map[string]sets.String) {
```
rfranzke: Hm, why do you pass `removedKubernetesVersions` and `removedMachineImageVersions` here but not `shoot` or `cloudProfile`, for example?

danielfoehrKn: Thanks for spotting.

rfranzke (dismissed):
/lgtm

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -19,8 +19,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sync"
 	"time"
 
+	"github.com/hashicorp/go-multierror"
```
timebertt: Move down to 3rd party import block

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +323,98 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+						}
+						removedMachineImageVersions[oldImage.Name] = removedMachineImageVersions[oldImage.Name].Insert(version.Version)
+					}
+				}
+			}
+
+			if len(removedKubernetesVersions) > 0 || len(removedMachineImageVersions) > 0 {
+				shootList, err1 := r.shootLister.List(labels.Everything())
+				if err1 != nil {
+					return apierrors.NewBadRequest(fmt.Sprintf("could not list shoots to verify Kubernetes and Machine image version: %s", err1.Error()))
```
timebertt:
```suggestion
					return apierrors.NewBadRequest(fmt.Sprintf("could not list shoots to verify that Kubernetes and/or Machine image versions can be removed: %s", err1.Error()))
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +323,98 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+				// close channel when wait group has 0 counter
+				go func() {
+					wg.Wait()
+					close(channel)
+				}()
+
+				for channelResult := range channel {
+					if channelResult != nil {
+						err = multierror.Append(err, channelResult)
+					}
```
timebertt: `multierror.Append()` already checks if the err is `nil`.
```suggestion
					err = multierror.Append(err, channelResult)
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +323,98 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+							channel <- nil
+							return
+						}
+
+						if removedKubernetesVersions.Has(shoot.Spec.Kubernetes.Version) {
+							channel <- fmt.Errorf("unable to delete Kubernetes version %q from CloudProfile %q  - version is still in use by shoot %q/%q", shoot.Spec.Kubernetes.Version, shoot.Spec.CloudProfileName, shoot.Name, shoot.Namespace)
+						}
+						for _, worker := range shoot.Spec.Provider.Workers {
+							if worker.Machine.Image == nil {
+								channel <- nil
```
timebertt: I don't see the point of sending a `nil` error.
```suggestion
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +323,98 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+								return
+							}
+							// happens if Shoot already runs a version that does not exist in the CloudProfile - in this case: ignore
+							if _, ok := removedMachineImageVersions[worker.Machine.Image.Name]; !ok {
+								channel <- nil
```
timebertt: Same here
```suggestion
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +323,98 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+
+				var (
+					channel = make(chan error)
+					wg      sync.WaitGroup
+				)
+				wg.Add(len(shootList))
+
+				for _, s := range shootList {
+					shoot := s.DeepCopy()
+					go func(ctx context.Context, shoot *core.Shoot, cloudProfile *core.CloudProfile, removedKubernetesVersions sets.String, removedMachineImageVersions map[string]sets.String) {
```
timebertt: `ctx` is unused
```suggestion
					go func(shoot *core.Shoot, cloudProfile *core.CloudProfile, removedKubernetesVersions sets.String, removedMachineImageVersions map[string]sets.String) {
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +323,98 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+						defer wg.Done()
+
+						if shoot.Spec.CloudProfileName != cloudProfile.Name {
+							channel <- nil
+							return
+						}
```
timebertt: Can we execute this check before starting the goroutine and before `DeepCopy`ing the shoot to reduce unnecessary operations?

plugin/pkg/global/resourcereferencemanager/admission.go
timebertt: Also can you explain, why you are passing the other variables by parameter? I don't see any benefit from it.

plugin/pkg/global/resourcereferencemanager/admission_test.go
```diff
@@ -684,5 +684,239 @@ var _ = Describe("resourcereferencemanager", func() {
...
+				attrs := admission.NewAttributesRecord(&cloudProfileNew, &cloudProfile, core.Kind("CloudProfile").WithVersion("version"), "", cloudProfile.Name, core.Resource("CloudProfile").WithVersion("version"), "", admission.Update, &metav1.UpdateOptions{}, false, defaultUserInfo)
+
+				err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("1.17.1"))
+				Expect(err.Error()).To(ContainSubstring(shootTwo.Spec.Provider.Workers[1].Machine.Image.Name))
+			})
+
+			It("should reject removal of a whole machine image  which versions are in use by a shoot", func() {
```
timebertt:
```suggestion
			It("should reject removal of a whole machine image which versions are in use by a shoot", func() {
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +323,98 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+
+						if removedKubernetesVersions.Has(shoot.Spec.Kubernetes.Version) {
+							channel <- fmt.Errorf("unable to delete Kubernetes version %q from CloudProfile %q  - version is still in use by shoot %q/%q", shoot.Spec.Kubernetes.Version, shoot.Spec.CloudProfileName, shoot.Name, shoot.Namespace)
+						}
+						for _, worker := range shoot.Spec.Provider.Workers {
+							if worker.Machine.Image == nil {
+								channel <- nil
+								return
```
timebertt: What if following workers have an image set, which is removed? I think, you won't detect it, if you do an early exit here.
```suggestion
								continue
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +323,98 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+							}
+							// happens if Shoot already runs a version that does not exist in the CloudProfile - in this case: ignore
+							if _, ok := removedMachineImageVersions[worker.Machine.Image.Name]; !ok {
+								channel <- nil
+								return
```
timebertt: Same here
```suggestion
								continue
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +323,98 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+					go func(ctx context.Context, shoot *core.Shoot, cloudProfile *core.CloudProfile, removedKubernetesVersions sets.String, removedMachineImageVersions map[string]sets.String) {
+						defer wg.Done()
+
+						if shoot.Spec.CloudProfileName != cloudProfile.Name {
+							channel <- nil
+							return
+						}
+
+						if removedKubernetesVersions.Has(shoot.Spec.Kubernetes.Version) {
+							channel <- fmt.Errorf("unable to delete Kubernetes version %q from CloudProfile %q  - version is still in use by shoot %q/%q", shoot.Spec.Kubernetes.Version, shoot.Spec.CloudProfileName, shoot.Name, shoot.Namespace)
```
timebertt:
```suggestion
							channel <- fmt.Errorf("unable to delete Kubernetes version %q from CloudProfile %q - version is still in use by shoot %q/%q", shoot.Spec.Kubernetes.Version, shoot.Spec.CloudProfileName, shoot.Name, shoot.Namespace)
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +323,98 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+								channel <- nil
+								return
+							}
+							// happens if Shoot already runs a version that does not exist in the CloudProfile - in this case: ignore
+							if _, ok := removedMachineImageVersions[worker.Machine.Image.Name]; !ok {
+								channel <- nil
+								return
+							}
+							if removedMachineImageVersions[worker.Machine.Image.Name].Has(worker.Machine.Image.Version) {
+								channel <- fmt.Errorf("unable to delete Machine image version %q/%q from CloudProfile %q - version is still in use by shoot %q/%q by worker %q", worker.Machine.Image.Name, worker.Machine.Image.Version, shoot.Spec.CloudProfileName, shoot.Name, shoot.Namespace, worker.Name)
```
timebertt: Using `%q/%q` looks kind of weird to me (e.g. `"ubuntu"/"1.17.1"`). WDYT?
```suggestion
								channel <- fmt.Errorf("unable to delete Machine image version '%s/%s' from CloudProfile %q - version is still in use by shoot '%s/%s' by worker %q", worker.Machine.Image.Name, worker.Machine.Image.Version, shoot.Spec.CloudProfileName, shoot.Namespace, shoot.Name, worker.Name)
```

plugin/pkg/global/resourcereferencemanager/admission_test.go
```diff
@@ -684,5 +684,239 @@ var _ = Describe("resourcereferencemanager", func() {
...
+		Context("CloudProfile - Update Kubernetes versions", func() {
+			versions := []core.ExpirableVersion{
+				{Version: "1.17.2"},
+				{Version: "1.17.1"},
+				{Version: "1.17.0"},
+			}
+			shootOne := shootBase
+			shootOne.Name = "shoot-One"
+			shootOne.Spec.Provider.Type = "aws"
+			shootOne.Spec.Kubernetes.Version = "1.17.2"
```
timebertt, line 698: Can you add a CloudProfile name to Shoots and CloudProfile?
```suggestion
			shootOne.Spec.Kubernetes.Version = "1.17.2"
			shootOne.Spec.CloudProfileName = "aws"
```

plugin/pkg/global/resourcereferencemanager/admission_test.go
```diff
@@ -684,5 +684,239 @@ var _ = Describe("resourcereferencemanager", func() {
...
+							},
+						},
+					},
+				}
+
+				attrs := admission.NewAttributesRecord(&cloudProfileNew, &cloudProfile, core.Kind("CloudProfile").WithVersion("version"), "", cloudProfile.Name, core.Resource("CloudProfile").WithVersion("version"), "", admission.Update, &metav1.UpdateOptions{}, false, defaultUserInfo)
+
+				err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+				Expect(err).To(HaveOccurred())
```
timebertt, line 762: Can you please test, that expected error really occured?
```suggestion
				Expect(err).To(HaveOccurred())
				Expect(err.Error()).To(ContainSubstring(fmt.Sprintf("unable to delete Kubernetes version %q from CloudProfile %q  - version is still in use by shoot %q/%q", "1.17.1", ...)))
```

plugin/pkg/global/resourcereferencemanager/admission_test.go
```diff
@@ -684,5 +684,239 @@ var _ = Describe("resourcereferencemanager", func() {
 				})))
 			})
 		})
+
+		Context("CloudProfile - Update Kubernetes versions", func() {
```
timebertt, line 688: Can you add a test, that checks that `Admit` succeeds if k8s/machine image versions weren't changed?

danielfoehrKn:
Can we go ahead with this PR?

timebertt (changes requested):
Overall it looks good, but I left you small suggestions.

plugin/pkg/global/resourcereferencemanager/admission.go
danielfoehrKn: [multierror err will be initialized neverthless](https://github.com/gardener/gardener/blob/e5a97cf136d73afe0267191a67845a8914c6b75d/vendor/github.com/hashicorp/go-multierror/append.go#L13-L12) - so the check in if err != nil will yield a wrong result

plugin/pkg/global/resourcereferencemanager/admission.go
danielfoehrKn: I think you are right, I only need to hand over the shoot because it is executed in the loop.

plugin/pkg/global/resourcereferencemanager/admission.go
danielfoehrKn: thanks, good catch. I added a test case for that scenario.

plugin/pkg/global/resourcereferencemanager/admission.go
timebertt: Yeah, but as you are now only sending non-nil errors through the channel, I think you can remove the check.

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,97 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+						}
+						removedMachineImageVersions[oldImage.Name] = removedMachineImageVersions[oldImage.Name].Insert(version.Version)
+					}
+				}
+			}
+
+			if len(removedKubernetesVersions) > 0 || len(removedMachineImageVersions) > 0 {
+				shootList, err1 := r.shootLister.List(labels.Everything())
+				if err1 != nil {
+					return apierrors.NewBadRequest(fmt.Sprintf("could not list shoots to verify that Kubernetes and/or Machine image version can be removed: %s", err1.Error()))
```
timebertt:
```suggestion
					return apierrors.NewInternalError(fmt.Errorf("could not list shoots to verify that Kubernetes and/or Machine image version can be removed: %v", err1))
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,97 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+				wg.Add(len(shootList))
+
+				for _, s := range shootList {
+					if s.Spec.CloudProfileName != cloudProfile.Name {
+						wg.Done()
+						continue
+					}
+
+					shoot := s.DeepCopy()
+					go func(shoot *core.Shoot) {
```
timebertt: I think copying the pointer should be enough here. We should avoid unnecessary calls to `DeepCopy()` in places where we loop over all Shoots (which can be a lot in some gardener installations).
```suggestion
					shoot := s
					go func() {
```

plugin/pkg/global/resourcereferencemanager/admission.go
timebertt: Hmm, sry. I don't see it. Also, if I replace `continue` with `return` no test fails.

plugin/pkg/global/resourcereferencemanager/admission.go
timebertt: Can you also add a unit test for this case here?

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,97 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+						defer wg.Done()
+
+						if removedKubernetesVersions.Has(shoot.Spec.Kubernetes.Version) {
+							channel <- fmt.Errorf("unable to delete Kubernetes version %q from CloudProfile %q - version is still in use by shoot %q/%q", shoot.Spec.Kubernetes.Version, shoot.Spec.CloudProfileName, shoot.Name, shoot.Namespace)
```
timebertt:
```suggestion
							channel <- fmt.Errorf("unable to delete Kubernetes version %q from CloudProfile %q - version is still in use by shoot '%s/%s'", shoot.Spec.Kubernetes.Version, shoot.Spec.CloudProfileName, shoot.Namespace, shoot.Name)
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,97 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+							if worker.Machine.Image == nil {
+								continue
+							}
+							// happens if Shoot runs an image that does not exist in the old CloudProfile - in this case: ignore
+							if _, ok := removedMachineImageVersions[worker.Machine.Image.Name]; !ok {
+								continue
+							}
+
+							if removedMachineImageVersions[worker.Machine.Image.Name].Has(worker.Machine.Image.Version) {
+								channel <- fmt.Errorf("unable to delete Machine image version'%s/%s' from CloudProfile %q - version is still in use by shoot '%s/%s' by worker %q", worker.Machine.Image.Name, worker.Machine.Image.Version, shoot.Spec.CloudProfileName, shoot.Namespace, shoot.Name, worker.Name)
```
timebertt:
```suggestion
								channel <- fmt.Errorf("unable to delete Machine image version '%s/%s' from CloudProfile %q - version is still in use by shoot '%s/%s' by worker %q", worker.Machine.Image.Name, worker.Machine.Image.Version, shoot.Spec.CloudProfileName, shoot.Namespace, shoot.Name, worker.Name)
```

timebertt (changes requested):
Some last suggestions :)

danielfoehrKn:
PTAL

plugin/pkg/global/resourcereferencemanager/admission.go
timebertt: ping

plugin/pkg/global/resourcereferencemanager/admission.go
timebertt: ping

plugin/pkg/global/resourcereferencemanager/admission.go
timebertt: ping

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,94 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+						}
+						removedMachineImageVersions[oldImage.Name] = removedMachineImageVersions[oldImage.Name].Insert(version.Version)
+					}
+				}
+			}
+
+			if len(removedKubernetesVersions) > 0 || len(removedMachineImageVersions) > 0 {
+				shootList, err1 := r.shootLister.List(labels.Everything())
+				if err1 != nil {
+					return apierrors.NewInternalError(fmt.Errorf("could not list shoots to verify that Kubernetes and/or Machine image version can be removed: %v", err1.Error()))
```
timebertt:
```suggestion
					return apierrors.NewInternalError(fmt.Errorf("could not list shoots to verify that Kubernetes and/or Machine image version can be removed: %v", err1))
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -319,6 +322,94 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o
...
+					if s.Spec.CloudProfileName != cloudProfile.Name {
+						wg.Done()
+						continue
+					}
+
+					go func(shoot *core.Shoot) {
+						defer wg.Done()
+
+						if removedKubernetesVersions.Has(shoot.Spec.Kubernetes.Version) {
+							channel <- fmt.Errorf("unable to delete Kubernetes version %q from CloudProfile %q - version is still in use by shoot %q/%q", shoot.Spec.Kubernetes.Version, shoot.Spec.CloudProfileName, shoot.Namespace, shoot.Name)
```
timebertt:
```suggestion
							channel <- fmt.Errorf("unable to delete Kubernetes version %q from CloudProfile %q - version is still in use by shoot '%s/%s'", shoot.Spec.Kubernetes.Version, shoot.Spec.CloudProfileName, shoot.Namespace, shoot.Name)
```

timebertt (changes requested):
Thanks, two small nits left.

timebertt (approved):
Thanks, a lot!
Looks great now :)
/approve

rfranzke (approved):

PR #2107, opened 2020-03-26
[GEP-5] version management

### LINKED ISSUE #1577: Implement versioning policy contract (GEP-5)

danielfoehrKn:
This PR is part of GEP5 .

Introduces an optional version management implemented as a part of the CloudProfile reconciliation.
I recommend checking out the documentation under docs/operations to get an overview.

Part of #1577

✅ Depends on #2105

Only review commit "GEP5 - version management" as the other commit is a prerequisite and already being review [here.](https://github.com/gardener/gardener/pull/2105)
**Release note**:

```improvement operator
 Version Management for Machine image and Kubernetes versions is available. Version Management  is an optional feature configurable in the Gardener Controller Manager config and part of the CloudProfile reconciliation. For more information please check out the documentation under https://github.com/gardener/gardener/blob/master/docs/operations/versioning.md 
```


rfranzke:
Also, please drop the commit that is part of #2105 if possible.

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
+# Kubernetes and Machine image versions in Gardener
```
rfranzke, line 1:
```suggestion
# Kubernetes and Machine Image Versions in Gardener
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Motivation
+
+On the one hand-side Gardener is responsible for managing the Kubernetes and Machine image versions of its Shoot clusters.
```
rfranzke, line 5:
```suggestion
On the one hand, Gardener is responsible for managing the Kubernetes and machine image versions of its Shoot clusters.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+On the other hand-side, Gardener needs to be configured and updated based on the availability and support of the Kubernetes and Machine image version it provides.
```
rfranzke, line 6:
```suggestion
On the other hand, the Gardener configuration needs to be maintained with the Kubernetes and machine image versions (based on availability and support) that shall be offered for Shoot clusters.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+
+For instance the Kubernetes community releases **minor** versions roughly every three months and usually maintains **three minor** versions (the actual and the last two) with bug fixes and security updates.
```
rfranzke, line 8:
```suggestion
For instance, the Kubernetes community releases **minor** versions roughly every three months and usually maintains **three minor** versions (the actual and the last two) with bug fixes and security updates. 
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+## Motivation
+
+On the one hand-side Gardener is responsible for managing the Kubernetes and Machine image versions of its Shoot clusters.
+On the other hand-side, Gardener needs to be configured and updated based on the availability and support of the Kubernetes and Machine image version it provides.
+
+For instance the Kubernetes community releases **minor** versions roughly every three months and usually maintains **three minor** versions (the actual and the last two) with bug fixes and security updates. 
+Patch releases are done more frequently.
+
+Gardener needs to be configured accordingly to reflect the "logical state" of a version. 
+It should be possible to define the Kubernetes or Machine image versions that still receive bug fixes and security patches and also vice-versa to define the version that are out-of-maintenance and are potentially vulnerable.
```
rfranzke, line 12:
```suggestion
It should be possible to define the Kubernetes or machine image versions that still receive bug fixes and security patches and also vice-versa to define the version that are out-of-maintenance and are potentially vulnerable.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+Moreover, this allows Gardener to "understand" the current state of a version and act upon it (more information in the following sections). 
+
+Because the manual maintenance can be cumbersome, Gardener can automate certain activities
```
rfranzke, line 15:
```suggestion
As the manual maintenance can be cumbersome, Gardener can automate certain activities 
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+        expirationDate: "2020-04-25T09:30:40Z"
+        version: 1.16.4
+      - classification: supported
+        version: 1.15.7
+      - classification: deprecated
+        expirationDate: "2020-06-09T14:01:39Z"
+        version: 1.15.6
+```
+ 
+ ## Version Requirements (Kubernetes and Machine image)
```
rfranzke, line 85:
```suggestion
 ## Version Requirements (Kubernetes and machine image)
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ 
+ The Gardener API server enforces the following requirements for versions:
```
rfranzke, line 87: Remove all leading whitespaces in the following lines

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+  
+ **Deletion a version** 
+ - A version that is in use by a Shoot cannot be deleted from the CloudProfile.
```
rfranzke, line 90:
```suggestion
 - A version that is in use by a Shoot cannot be removed from the `CloudProfile`.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ **Adding a version** 
+ - A version must not have an expiration date in the past.
+ - There can be only one `supported` version (there can be one supported version per Minor version if `Version Maintenance` is configured).
+ 
+ ## Forceful migration of expired versions
+ 
+ If a Shoot is running a version after its expiration date has passed, it will be forcefully migrated during it maintenance time.
+ This happens **even if the owner has opted out of automatic cluster updates!**
+
+For **Machine images**, the Shoots worker pools will be updated to the latest `non-preview` version of the pools respective image.
```
rfranzke, line 101:
```suggestion
For **machine image versions**, the Shoots' worker pools will be updated to the latest `non-preview` version of the pools respective image.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+For **Kubernetes versions**, the forceful update picks the latest 'non-preview' patch version of the current minor version. 
+
+ If the cluster is already on the latest patch version and the latest patch version is also expired, 
+ it will continue with the latest patch version of the **next consecutive minor Kubernetes version**, so **it will result in an 
+ update of a minor Kubernetes version!** 
+ 
+ If that's expired as well, the update process repeats until a non-expired Kubernetes version is reached.
+  
+ **Depending on the circumstances described above, it can happen that the cluster receives multiple consecutive minor Kubernetes version updates!**
+ As a Kubernetes version update cannot skip a minor version, the CloudProfile has to be maintained properly.
```
rfranzke, line 112:
```suggestion
 As a Kubernetes version update cannot skip a minor version, the `CloudProfile` has to be maintained properly.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Version Management
+
+Using `Version Management` can help with managing the Kubernetes and Machine Image versions in the CloudProfile. It is not required and not enabled by default.
```
rfranzke, line 116:
```suggestion
Using `Version Management` can help with managing the Kubernetes and machine Image versions in the `CloudProfile`. It is not required and not enabled by default.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+
+Characteristics of using `Version Management` are:
+- Initial classification of versions in the CloudProfile having no classification.
```
rfranzke, line 119:
```suggestion
- Newly added versions in the `CloudProfile` don't have a classification.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+- Enforces one `supported` version that has no expiration date
```
rfranzke, line 120:
```suggestion
- Enforces exactly one `supported` version without an expiration date
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+- Automatically deprecates versions with the correct `expiration date` (defined in `controllers.cloudProfile.kubernetesVersionManagement.expirationDuration` for Kubernetes versions and `controllers.cloudProfile.machineImageVersionManagement.expirationDuration` for machine images)
```
rfranzke, line 121:
```suggestion
- Automatically deprecates versions with the correct `expiration date` (defined in `controllers.cloudProfile.kubernetesVersionManagement.expirationDuration` for Kubernetes versions and `controllers.cloudProfile.machineImageVersionManagement.expirationDuration` for machine image versions)
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+- Reconciliation when the CloudProfile has been updated
```
rfranzke, line 122:
```suggestion
- `Reconciliation` when the CloudProfile has been updated
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+For instance, Kubernetes releases **minor** versions roughly every three months and usually maintains **three minor** versions (the actual and the last two). 
+This means Kubernetes clusters having a version that is not `maintained` any more, run into the risk of having severe bugs or known security problems. 
+Considering Kubernetes minor versions: 1.18 (being the most recent), 1.17, 1.16, 1.15 and 1.14. Then versions 1.18, 1.17, 1.16 are `maintained` and versions 1.15 and 1.14 are `unmaintained`
+
+The operator of Gardener can optionally define `maintained` minor versions through the configuration file of the Gardener Controller Manager.
+This can be done by specifying the number of maintained minor versions starting from the most recent. Given the Kubernetes project example above, the amount of maintained minor versions is `3`.
+The configuration is located in the Gardener Controller Manager config at: `controllers.cloudProfile.kubernetesVersionManagement.versionMaintenance` and `controllers.cloudProfile.machineImageVersionManagement.versionMaintenance`.
+
+
```
rfranzke, line 155:
```suggestion
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+`Version Maintenance` helps with the handling of `maintained` and `unmaintained` versions, specifically:
+ - Enforces one `supported` version per `maintained` **minor version** (versus only one `supported` version without `Version Maintenance`)
+   - be aware that if there is no supported version for a maintained minor version, it will pick the **latest** version independent of the classification and sets it to `supported`.
+ - Deprecates all versions with an `unmaintained` minor (no matter what classification they have).
+ - Sets the correct deprecation date for `unmaintained` versions (`controllers.cloudProfile.kubernetesVersionManagement.expirationDuration`).
+ - Does not set an expiration date on the latest version with an `unmaintained` version. Otherwise, an undesired and automated forceful minor Kubernetes version upgrade is possible. This is potentially harmful for the deployed workload.
```
rfranzke, line 161:
```suggestion
 - Does not set an expiration date on the latest version with an `unmaintained` version. Otherwise, an undesired and automated forceful minor Kubernetes version upgrade is possible. This is potentially harmful for the deployed workload. If a Gardener operator wants to phase out a minor version he/she has to set the expiration date manually.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+      expirationDuration: 2880h # defaults to 120 days / 4 months
+```
+
+## Duties of the Gardener Operator
+
+Even though Gardener strives for high automation, we think that it leads to safer operations 
+having a human operator in charge of the following operations: 
+
+- Adding `preview` versions
+- Promotion of a `preview` version to `supported`. We think this should be done on a case-by-case basis, hence Gardener deliberately does not offer an automation in this case.
```
rfranzke, line 195: Hm, I don't like this too much. Yes, it's safer this way of course, but given the frequency of patch releases this might involve quite a lot of manual action...

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+- Deprecate a `supported` version
+   - This is of particular importance when using `Version Management`. As it defines an initial classification for versions with no classification, 
+ it can happen that it tries to introduce another `supported` version even though there is already one present. That is forbidden and blocked by the Gardener API server. 
+ That situation occurs when specifying an unclassified version that is the overall latest version (or the latest version of a minor version when using `Version Maintenance`).
+ It is the operator duty to make sure that this does not happen. Specifically when introducing a new latest version without a classification, the current `supported` version should be manually deprecated ahead of time.
```
rfranzke, line 200:
```suggestion
 It is the operator's duty to make sure that this does not happen. Specifically when introducing a new latest version without a classification, the current `supported` version should be manually deprecated ahead of time.
```

rfranzke: > should be manually deprecated ahead of time.

Does this mean that the deprecation date has to be entered manually as well?

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+- Deletion of versions. Before deletion, the operator has to make sure the version is not in use by any Shoot. The recommended way to achieve this, is to deprecate the version and set an expiration date. 
+Clusters with `auto-update` configured will automatically update during their next maintenance. The remainder is forcefully updated during maintenance time after the version is finally expired.
+- Setting an `expiration date` on the latest patch version of an `unmaintained`  Kubernetes minor version. This is to prevent that Gardener automatically forces minor version upgrades for Clusters.
+    - After the deprecation date passed, it will force a Kubernetes minor version update during the next maintenance time window of the Shoot.
+
+## Common scenarios for the Gardener Operator
+
+The following describes common scenarios the Gardener operator regularly faces when managing the Kubernetes and Machine image versions of a Gardener installation.
+We assume a CloudProfile that looks schematically like this:
```
rfranzke, line 209:
```suggestion
We assume a `CloudProfile` that looks schematically like this:
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ It is the operator duty to make sure that this does not happen. Specifically when introducing a new latest version without a classification, the current `supported` version should be manually deprecated ahead of time.
+- Deletion of versions. Before deletion, the operator has to make sure the version is not in use by any Shoot. The recommended way to achieve this, is to deprecate the version and set an expiration date. 
+Clusters with `auto-update` configured will automatically update during their next maintenance. The remainder is forcefully updated during maintenance time after the version is finally expired.
+- Setting an `expiration date` on the latest patch version of an `unmaintained`  Kubernetes minor version. This is to prevent that Gardener automatically forces minor version upgrades for Clusters.
+    - After the deprecation date passed, it will force a Kubernetes minor version update during the next maintenance time window of the Shoot.
+
+## Common scenarios for the Gardener Operator
+
+The following describes common scenarios the Gardener operator regularly faces when managing the Kubernetes and Machine image versions of a Gardener installation.
```
rfranzke, line 208:
```suggestion
The following describes common scenarios the Gardener operator regularly faces when managing the Kubernetes and machine image versions of a Gardener installation.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+1.15.1 : deprecated with no expiration date
+1.15.0 : deprecated with expiration date that is not expired
+1.14.0 : deprecated with expiration date already expired
+```
+For `Version Maintenance`, the Gardener Controller Manager configuration in `controllers.cloudProfile.kubernetesVersionManagement.versionMaintenance` is set to `3`.
+
+**Scenarios**
+
+* * *
+***A new patch version has been released and should be added to the CloudProfile.***
```
rfranzke, line 225:
```suggestion
***A new patch version has been released and should be added to the `CloudProfile`.***
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+
+**Example**: Kubernetes releases version 1.18.4. 
+ 
+ -  Add the version and set the classification to `preview`. There can be multiple preview versions.
+ 
+**Example 2**: Kubernetes releases version 1.17.1. 
+ 
+ - With `Version Management` enabled: Version 1.17.1 will be deprecated wit an expiration date right away.
```
rfranzke, line 233:
```suggestion
 - With `Version Management` enabled: Version 1.17.1 will be deprecated with an expiration date right away.
```

rfranzke: Why will it be deprecated right away? Do you mean 1.17.0?

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ 
+ - With `Version Maintenance` enabled: Version 1.17.1 has a maintained minor version. 
+    If it is classified as `preview`, it stays unchanged. If added without a classification, version 1.17.1 will be classified as `supported` and 
+    version 1.17.0 will deprecated with the expiration date for `supported` versions. 
+
+* * *
+***A new minor release version has been released and should be added to the CloudProfile.***
```
rfranzke, line 240:
```suggestion
***A new minor release version has been released and should be added to the `CloudProfile`.***
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+
+**Example**: Kubernetes releases version 1.19.0.
+
+ - Deprecate the current `supported` version, optionally with an expiration date. Afterwards add the new version as a `preview` version.
```
rfranzke, line 244: I don't get it. Can you enhance this explanation?

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ 
+ - `Version Management` enabled: Add version 1.19.0 without a classification. 1.19.0 will be classified as `supported` and version 1.18.3 will be expired with the configured expiration date.
```
rfranzke, line 246: Why is 1.19.0 supported right away? Shouldn't it be `preview`?
Also, where is the expiration date for 1.18.3 coming from?

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ - With `Version Maintenance` enabled: Add version 1.19.0 without a classification. `Version Maintenance` makes sure that versions now having an `unmaintained` minor, will be accordingly deprecated. 1.19.0 will be classified as `supported` and version 1.16.0 will be deprecated having no expiration date.
+ 
+* * *
+***A `preview` version should be promoted to `supported`.***
+
+**Example**: Kubernetes version 1.18.4 should be promoted to `supported`.
+
+ - Deprecate the currently `supported` version 1.18.3, optionally with a expiration date. Afterwards change the classification of the version 1.18.4 from `preview` to `supported`.
+ 
+ - `Version Management` enabled: Deprecate the current `supported` version 1.18.3. An expiration date is being set automatically. Afterwards change the classification of the version 1.18.4 from `preview` to `supported`
```
rfranzke, line 257: Can it be done in one call?

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ - With `Version Maintenance` enabled: Same as for `Version Management`, but make sure to deprecate the `supported` version of the particular minor version (in this example: 1.18.3).
+ 
+* * *
+***The `supported` version should be deprecated.***
+
+**Example**: Version 1.18.3 should be deprecated.
+
+ - Deprecate version 1.18.3 by setting the classification `deprecated`, optionally with a expiration date.
+ 
+ - `Version Management` enabled: By just adding a new version e.g 1.18.4 without a classification, Version 1.18.3 will be deprecated and 1.18.4 becomes the new `supported` version.
```
rfranzke, line 268:
```suggestion
 - `Version Management` enabled: By just adding a new version, e.g. 1.18.4, without a classification, version 1.18.3 will be deprecated and 1.18.4 becomes the new `supported` version.  
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ - With `Version Maintenance` enabled:  Similar as for `Version Management`. Having one `supported` version per `maintained` minor version is enforced. 
+ 
+* * *
+**The current `supported` version has problems and should be replaced with the previous `supported` version.**
+
+**Example**: Version 1.18.3 has problems.
+
+a) Version is not in use by any Shoot.
+ - delete the version from the CloudProfile. 
+ - Manually change the classification of the previously `supported` version.
```
rfranzke, line 280:
```suggestion
 - Manually change the classification of the previously `supported` version to `supported` again.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ - `Version Management` and `Version Maintenance`: nothing else has to be done. 
+
+b) Shoots are already running the erroneous version
```
rfranzke, line 283:
```suggestion
b) Shoots are already running the 1.18.3 version
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ -  Make sure there is a higher non-preview version available in the CloudProfile.
```
rfranzke, line 284:
```suggestion
 -  Make sure there is a higher non-preview version available in the `CloudProfile`.
```

rfranzke: What if this is not the case? For example, when 1.18.4 is not yet released?

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ -  Manually change the classification of the erroneous version to `deprecated` with an `expiration date` in the past. Change the version to `supported` again.
+ - `Version Management` and `Version Maintenance`: The current `supported` version needs to be labeled as `preview` with an `expiration date` in the past. 
+    Because one supported version is enforced, the latest `deprecated` version will automatically be classified as `supported` again.
+    Shoots running the `preview` version with an `expiration date` in the past will be forcefully updated. 
+    
+
+c) Shoots are already running the erroneous version and it is the overall latest version:
+ - Shoots cannot upgrade to a newer version as there is none available. Also a downgrade is not supported. 
+   To reduce the blast-radius, label the version as `deprecated` with an `expiration date` in the past to prevent `auto-updates`, defaulting and new cluster creation with that version.
+   The operator has to wait for a newer patch version or delete the Shoot clusters in question. Only then, the version can be deleted from the CloudProfile.
```
rfranzke, line 294:
```suggestion
   The operator has to wait for a newer patch version or delete the Shoot clusters in question. Only then, the version can be deleted from the `CloudProfile`.
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ - `Version Management` and `Version Maintenance`: same approach, but label the version as `preview`, otherwise the automation will reconcile it to be the `supported` version.
```
rfranzke, line 295:
```suggestion
 - `Version Management` and `Version Maintenance`: same approach, but label the version as `preview`, otherwise the controller will reconcile it to be the `supported` version.
```

rfranzke: Will it automatically mark the next-lower patch version as `supported` again? What if it has had problems as well?

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+ 
+* * *
+***A version should be removed from the CloudProfile, is already expired but cannot be deleted.***
```
rfranzke, line 298:
```suggestion
***A version should be removed from the `CloudProfile`, is already expired but cannot be deleted.***
```

docs/operations/versioning.md
```diff
@@ -0,0 +1,305 @@
...
+
+**Example**: Clusters are using version 1.14.0. 
+ 
+ - Version 1.14.0 can only be removed it is not in use by any Shoot anymore.
+ - Wait 24 hours to make sure that the maintenance of all Shoots has run - this should have forceully upgraded the clusters.
+ - Make sure the maintenance was successful (check the events on the Shoot). A misconfigured CloudProfile can block the maintenance operations.
```
rfranzke, line 304:
```suggestion
 - Make sure the maintenance was successful (check the events on the Shoot). A misconfigured `CloudProfile` can block the maintenance operations.
```

rfranzke (changes requested):
Just read the docs for now, and my head is already buzzing... Thanks for the detailed description and examples, I don't think I have understood each and every scenario. Overall, I think it's a lot of operational complexity that is being introduced now. So many special cases/situations...

/cc @gardener/gardener-maintainers @vlerenc - can you please check (at least the docs) as well? What is your impression/opinion?

vlerenc:
@danielfoehrKn Could you please accept/reject the already made proposals (by @rfranzke) and then I'll fetch the PR and go through the description as well? Overall, with the exception of a few fuzzy statements in the wiki document, `docs/operations/versioning.md` should match the wiki, I guess. 

@rfranzke, was the wiki also confusing/too much? I am not sure how to simplify things more - our end users want a way to plan ahead and together we kind agreed on said document as a compromise.

rfranzke:
@vlerenc I'd say yes, the description in GEP5 itself was a bit confusing and challenging as well. Anyways, I was more talking about the concrete implementation now. I understood that you haven't checked @danielfoehrKn's markdown document yet - maybe go through it (even without the incorporation of my primarily cosmetic suggestions), I guess you'll see what I mean. I don't have concrete proposals now where/what to do better, I just read it and am concerned about the operational complexity that comes with it.

rfranzke:
We just had a chat with @danielfoehrKn @tim-ebert @timuthy and asked ourselves whether it makes more sense to externalise the behaviour that this PR introduces. Concretely, the suggestion was to put it into a separate binary that takes an existing `CloudProfile` plus the various configuration options (deprecation times, etc.) as input. It will generate a `CloudProfile` with the proper `expirationDate`s and `classifications` and returns it to the caller.

The arguments/rationale behind it was that the current behaviour 

1. conflicts with the GitOps approach because the `CloudProfile`'s `spec` is modified during runtime and might conflict with future applications of the committed manifest
1. is counter-intuitive from a usage pov as the `CloudProfile`'s `spec` contains a desired state from the end-user, however, it is modified over time by a controller
1. does not necessarily need to be in GCM itself - for end-users only the already merged `classifications` field and the `expirationDate` field is relevant - how those values are generated/where they are coming from doesn't matter to them, i.e., we could also externalise them into a separate binary

We haven't thought it through, probably @danielfoehrKn can tell best whether this would help making the usage more easy. Definitely, the current approach makes GitOps hard/impossible. Would such a separate binary that allows to provider higher-level abstractions of versions make the operations easier?

Generally, the e2e scenario could look as follows:

1. Operator stores the complete `CloudProfile` in git.
1. Operator calls the separate binary with the path to the YAML plus the configurations (e.g., when he wants to add a new version then he could use a new command line flag `--add-version=1.18.3` ❔)
1. Operator gets back the `CloudProfile` with updated `classification` and `expirationDate`s and stores it in git again
1. The `CloudProfile` gets applied via regular `kubectl apply -f <path>`.

docs/operations/versioning.md
danielfoehrKn: Thats one of the cases we discussed a while back with @vlerenc - we do not want to have an automation in place that could be harmful in case we "forget"about it. Lets say there is a preview version that has not been checked thoroughly - nevertheless after 2 weeks it will be promoted to supported ...

docs/operations/versioning.md
danielfoehrKn: No.

docs/operations/versioning.md
danielfoehrKn: No, using VersionManagement enforces only one supported version overall. The supported version in this example is already 1.18.3 - so the version 1.17.1 will be deprecated accordingly.

docs/operations/versioning.md
danielfoehrKn: I do not know what to enhance if I do not know what part is not understandable. 

This describes the manual steps involved if not using VersionManagement or VersionMaintenance.

docs/operations/versioning.md
danielfoehrKn: 1.19.0 can be specified as preview and no automation will touch it. 
These examples describe what the automation does. So adding the version 1.19.0 without a classification triggers the automation (either VersionManagement or VersionMaintenance) to determine an initial classification.

danielfoehrKn: The expiration date for 1.18.3 in the case of VersionManagement is configured in the Gardener ControllerManager config [as described above](https://github.com/gardener/gardener/blob/aaca736472c700f9cd9f8b4dbec06b9d2f4f90c0/docs/operations/versioning.md#L138).

docs/operations/versioning.md
danielfoehrKn: see [below describing that usecase.](https://github.com/gardener/gardener/blob/aaca736472c700f9cd9f8b4dbec06b9d2f4f90c0/docs/operations/versioning.md#L291)

danielfoehrKn:
This optional functionality intents to make the life of a Gardener Operator more seamless. 

If we think about the Gardener Operator as a person that edits the CloudProfile.yaml file, I think having an automation in place that makes sure the version adhere to a certain policy, can make their life easier. This way they do not have to manually touch & expire every version, ... .  If that is implemented as a controller in the GCM or as an external binary serves the same purpose for me.
 
Though I do understand that when someone use a GitOps based approach, having a reconciliation on the CloudProfile might present a challenge as single versions need to be patched individually. 

I am open to 
1) not introduce the optional Version Management capabilities 
2) create a separate binary that generates the CloudProfile yaml based on certain parameters (e.g the same as in this PR is configured in the config file). 

The thing for 2)  is that I would like to make sure is  that everyone agrees that this functionality is indeed useful for operating Gardener. If we are happy with the status quo, just editing each version in the CloudProfile manually (e.g via GitOps), then we might as well leave it the way it is today.   Otherwise  there is no point in investing additional effort into it. 
I do not think it makes sense to further invest in this PR until we have reached a consent here. 
@rfranzke @vlerenc

vlerenc:
Dear colleagues, shall I still read `versioning.md`? Is it unclear what we try to achieve? Basically the goals from our internal versioning policy, i.e. a few "controls" that make sure we remain true to our contract, i.e. to our deprecation and expiration policy, e.g. do not allow to remove a supported version or expire a supported version not sooner than in 4 months, etc. That's the main goal, i.e. to protect us/ops against fatal mistakes that could roll thousands of clusters. I would have thought, an admission web hook is fine/well invested time to protect us from mistakes and help ops doing everything right.

P.S.: That said, introducing classifications and acting on them is part of Gardener like e.g. not actively rolling into previews or rolling into whatever is expected once something expires.

rfranzke:
This PR is not about validation - it implements a controller that is automatically setting the expiration dates, changes the classifications (e.g., from `supported` to `deprecated`, etc.). This has a few drawbacks, see my https://github.com/gardener/gardener/pull/2107#issuecomment-613337799.

My understanding was that maintaining these fields/values is something that should better be automated because a) the risk of making mistakes is high (ok, can be mitigated with validation) and b) is a larger maintenance effort if it needs to be done manually (if the list of Kubernetes/OS versions is large).

If this was never the goal and we explicitly want/force operators to do the maintenance manually then we should only implement validation to enforce the configured policies are kept.

timebertt:
I have to admit, I am also still confused regarding the version management/maintenance in all of the special cases. I've read through those parts of the documentation again and agree that this would introduce quite some complexity, especially if combined with a GitOps approach.
I tend to prefer to put these helpers in a separate binary, as suggested by @rfranzke, so that it is indeed useful for an operator and does not get in his way instead.

vlerenc:
Hmm... while I am also confused by the separation between management and maintenance (after the first pass I had to read it again as I missed that there are two different m-words), I am even more confused by the GitOps argument. We kind of have that today and it is tedious and dangerous.

The minimal goal is prevent grave mistakes like removing or expiring a version well before the time we promise in our contract and rolling clusters forcefully. The extended goal is to make it less tedious.

In that sense, I don't think a dedicated executable to modify the cloud profiles "outside" the Garden cluster + GitOps even fulfills the minimal requirement, which is to make sure nobody can break the cloud profiles/configuration.

Overall, the document lists reasonable tasks. Maybe we can have a call?

rfranzke:
For the validation parts there are other PRs like #2106, for example. It is not complete yet and we don't have any validation for "preventing grave mistakes like removing or expiring a version well before the time we promise in our contract", for example, but it's outside of the scope of this particular PR. [cc @danielfoehrKn - these are valid points that we should be addressed]

Here we are implementing automatisms/controllers that change the `CloudProfile`s in the garden cluster over time.
This is conflicting with what we (and probably most other people) do with GitOps. Today, the desired `CloudProfile` (maybe having several classifications and expiration dates for some versions) is stored in Git. Here we have our desired state. Now, with this PR, over time the `CloudProfile` in Git diverges from the actual `CloudProfile` in the system, and this breaks the GitOps approach. Also, these automatisms lead to the complicated steps that need to be performed for certain operations as described in the document.

I understand that you mainly care about preventing us/operators from breaking the contract. This is totally fine and valid. Though, here we are already looking at something that targets what you described as "extended goal".

danielfoehrKn:
Closing this - agreed to implement this functionality not as a K8 controller but rather with an external binary that generates a CloudProfile yaml manifest to not break the GitOps approach. Please see #1577

PR #2108, opened 2020-03-26
[GEP-5] Changes to the Shoot Maintenance Controller

### LINKED ISSUE #1577: Implement versioning policy contract (GEP-5)

danielfoehrKn:
This PR is part of GEP5 .
Introduce changes to the Shoot Maintenance Controller.
- allows force upgrades of the minor version
- sends events on the Shoot stating which version have been upgraded and why 

I recommend checking out the documentation under docs/operations that is [included here](https://github.com/gardener/gardener/pull/2107).

Part of #1577

✅ Depends on #2105

**Release note**:

```noteworthy user
Shoot clusters can now receive forceful minor upgrades when using an expired Kubernetes version. Versions with `preview` classification are excluded from auto-update functionality for Kubernetes and machine image versions.
```

### DIFF
--- a/pkg/apis/core/v1beta1/helper/helper.go
+++ b/pkg/apis/core/v1beta1/helper/helper.go
@@ -16,17 +16,16 @@ package helper
 
 import (
 	"fmt"
-	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
-	"github.com/gardener/gardener/pkg/logger"
 	versionutils "github.com/gardener/gardener/pkg/utils/version"
 
 	"github.com/Masterminds/semver"
-	errors "github.com/pkg/errors"
+	"github.com/pkg/errors"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -579,17 +578,6 @@ func ShootUsesUnmanagedDNS(shoot *gardencorev1beta1.Shoot) bool {
 	return shoot.Spec.DNS != nil && len(shoot.Spec.DNS.Providers) > 0 && shoot.Spec.DNS.Providers[0].Type != nil && *shoot.Spec.DNS.Providers[0].Type == "unmanaged"
 }
 
-// GetMachineImagesFor returns a list of all machine images for a given shoot.
-func GetMachineImagesFor(shoot *gardencorev1beta1.Shoot) []*gardencorev1beta1.ShootMachineImage {
-	var workerMachineImages []*gardencorev1beta1.ShootMachineImage
-	for _, worker := range shoot.Spec.Provider.Workers {
-		if worker.Machine.Image != nil {
-			workerMachineImages = append(workerMachineImages, worker.Machine.Image)
-		}
-	}
-	return workerMachineImages
-}
-
 // DetermineMachineImageForName finds the cloud specific machine images in the <cloudProfile> for the given <name> and
 // region. In case it does not find the machine image with the <name>, it returns false. Otherwise, true and the
 // cloud-specific machine image will be returned.
@@ -617,33 +605,18 @@ func ShootMachineImageVersionExists(constraint gardencorev1beta1.MachineImage, i
 	return false, 0
 }
 
-// DetermineLatestMachineImageVersion determines the latest MachineImageVersion from a MachineImage
-func DetermineLatestMachineImageVersion(image gardencorev1beta1.MachineImage) (*semver.Version, gardencorev1beta1.ExpirableVersion, error) {
-	var (
-		latestSemVerVersion       *semver.Version
-		latestMachineImageVersion gardencorev1beta1.ExpirableVersion
-	)
-
-	for _, imageVersion := range image.Versions {
-		v, err := semver.NewVersion(imageVersion.Version)
-		if err != nil {
-			return nil, gardencorev1beta1.ExpirableVersion{}, fmt.Errorf("error while parsing machine image version '%s' of machine image '%s': version not valid: %s", imageVersion.Version, image.Name, err.Error())
-		}
-		if latestSemVerVersion == nil || v.GreaterThan(latestSemVerVersion) {
-			latestSemVerVersion = v
-			latestMachineImageVersion = imageVersion
-		}
-	}
-	return latestSemVerVersion, latestMachineImageVersion, nil
-}
-
-// GetShootMachineImageFromLatestMachineImageVersion determines the latest version in a machine image and returns that as a ShootMachineImage
-func GetShootMachineImageFromLatestMachineImageVersion(image gardencorev1beta1.MachineImage) (*semver.Version, gardencorev1beta1.ShootMachineImage, error) {
-	latestSemVerVersion, latestImage, err := DetermineLatestMachineImageVersion(image)
+// GetLatestQualifyingShootMachineImage determines the latest qualifying version in a machine image and returns that as a ShootMachineImage
+// A version qualifies if its classification is not preview and the version is not expired.
+func GetLatestQualifyingShootMachineImage(image gardencorev1beta1.MachineImage, predicates ...VersionPredicate) (bool, *gardencorev1beta1.ShootMachineImage, error) {
+	predicates = append(predicates, FilterExpiredVersion())
+	qualifyingVersionFound, latestImageVersion, err := GetLatestQualifyingVersion(image.Versions, predicates...)
 	if err != nil {
-		return nil, gardencorev1beta1.ShootMachineImage{}, err
+		return false, nil, err
 	}
-	return latestSemVerVersion, gardencorev1beta1.ShootMachineImage{Name: image.Name, Version: &latestImage.Version}, nil
+	if !qualifyingVersionFound {
+		return false, nil, nil
+	}
+	return true, &gardencorev1beta1.ShootMachineImage{Name: image.Name, Version: &latestImageVersion.Version}, nil
 }
 
 // UpdateMachineImages updates the machine images in place.
@@ -651,7 +624,6 @@ func UpdateMachineImages(workers []gardencorev1beta1.Worker, machineImages []*ga
 	for _, machineImage := range machineImages {
 		for idx, worker := range workers {
 			if worker.Machine.Image != nil && machineImage.Name == worker.Machine.Image.Name {
-				logger.Logger.Infof("Updating worker images of worker '%s' from version %s to version %s", worker.Name, *worker.Machine.Image.Version, *machineImage.Version)
 				workers[idx].Machine.Image = machineImage
 			}
 		}
@@ -670,58 +642,6 @@ func KubernetesVersionExistsInCloudProfile(cloudProfile *gardencorev1beta1.Cloud
 		}
 	}
 	return false, gardencorev1beta1.ExpirableVersion{}, nil
-}
-
-// DetermineLatestKubernetesPatchVersion finds the latest Kubernetes patch version in the <cloudProfile> compared
-// to the given <currentVersion>. In case it does not find a newer patch version, it returns false. Otherwise,
-// true and the found version will be returned.
-func DetermineLatestKubernetesPatchVersion(cloudProfile *gardencorev1beta1.CloudProfile, currentVersion string) (bool, string, error) {
-	ok, newerVersions, _, err := determineNextKubernetesVersions(cloudProfile, currentVersion, "~")
-	if err != nil || !ok {
-		return ok, "", err
-	}
-	sort.Strings(newerVersions)
-	return true, newerVersions[len(newerVersions)-1], nil
-}
-
-// DetermineNextKubernetesMinorVersion finds the next available Kubernetes minor version in the <cloudProfile> compared
-// to the given <currentVersion>. In case it does not find a newer minor version, it returns false. Otherwise,
-// true and the found version will be returned.
-func DetermineNextKubernetesMinorVersion(cloudProfile *gardencorev1beta1.CloudProfile, currentVersion string) (bool, string, error) {
-	ok, newerVersions, _, err := determineNextKubernetesVersions(cloudProfile, currentVersion, "^")
-	if err != nil || !ok {
-		return ok, "", err
-	}
-	sort.Strings(newerVersions)
-	return true, newerVersions[0], nil
-}
-
-// determineKubernetesVersions finds newer Kubernetes versions in the <cloudProfile> compared
-// with the <operator> to the given <currentVersion>. The <operator> has to be a github.com/Masterminds/semver
-// range comparison symbol. In case it does not find a newer version, it returns false. Otherwise,
-// true and the found version will be returned.
-func determineNextKubernetesVersions(cloudProfile *gardencorev1beta1.CloudProfile, currentVersion, operator string) (bool, []string, []gardencorev1beta1.ExpirableVersion, error) {
-	var (
-		newerVersions       = []gardencorev1beta1.ExpirableVersion{}
-		newerVersionsString = []string{}
-	)
-
-	for _, version := range cloudProfile.Spec.Kubernetes.Versions {
-		ok, err := versionutils.CompareVersions(version.Version, operator, currentVersion)
-		if err != nil {
-			return false, []string{}, []gardencorev1beta1.ExpirableVersion{}, err
-		}
-		if version.Version != currentVersion && ok {
-			newerVersions = append(newerVersions, version)
-			newerVersionsString = append(newerVersionsString, version.Version)
-		}
-	}
-
-	if len(newerVersions) == 0 {
-		return false, []string{}, []gardencorev1beta1.ExpirableVersion{}, nil
-	}
-
-	return true, newerVersionsString, newerVersions, nil
 }
 
 // SetMachineImageVersionsToMachineImage sets imageVersions to the matching imageName in the machineImages.
@@ -781,4 +701,155 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
 		return &providers[0]
 	}
 	return nil
+}
+
+type VersionPredicate func(expirableVersion gardencorev1beta1.ExpirableVersion, version *semver.Version) (bool, error)
+
+// GetKubernetesVersionForPatchUpdate finds the latest Kubernetes patch version for its minor version in the <cloudProfile> compared
+// to the given <currentVersion>. Preview and expired versions do not qualify for the kubernetes patch update. In case it does not find a newer patch version, it returns false. Otherwise,
+// true and the found version will be returned.
+func GetKubernetesVersionForPatchUpdate(cloudProfile *gardencorev1beta1.CloudProfile, currentVersion string) (bool, string, error) {
+	currentSemVerVersion, err := semver.NewVersion(currentVersion)
+	if err != nil {
+		return false, "", err
+	}
+
+	qualifyingVersionFound, latestVersion, err := GetLatestQualifyingVersion(cloudProfile.Spec.Kubernetes.Versions, FilterDifferentMajorMinorVersion(*currentSemVerVersion), FilterSameVersion(*currentSemVerVersion), FilterExpiredVersion())
+	if err != nil {
+		return false, "", err
+	}
+	// latest version cannot be found. Do not return an error, but allow for minor upgrade if Shoot's machine image version is expired.
+	if !qualifyingVersionFound {
+		return false, "", nil
+	}
+
+	return true, latestVersion.Version, nil
+}
+
+// GetKubernetesVersionForMinorUpdate finds a Kubernetes version in the <cloudProfile> that qualifies for a Kubernetes minor level update given a <currentVersion>.
+// A qualifying version is a non-preview version having the minor version increased by exactly one version.
+// In case the consecutive minor version has only expired versions, picks the latest expired version (will do another minor update during the next maintenance time)
+// If a version can be found, returns true and the qualifying patch version of the next minor version.
+// In case it does not find a version, it returns false.
+func GetKubernetesVersionForMinorUpdate(cloudProfile *gardencorev1beta1.CloudProfile, currentVersion string) (bool, string, error) {
+	currentSemVerVersion, err := semver.NewVersion(currentVersion)
+	if err != nil {
+		return false, "", err
+	}
+
+	qualifyingVersionFound, latestVersion, err := GetLatestQualifyingVersion(cloudProfile.Spec.Kubernetes.Versions, FilterNonConsecutiveMinorVersion(*currentSemVerVersion), FilterSameVersion(*currentSemVerVersion), FilterExpiredVersion())
+	if err != nil {
+		return false, "", err
+	}
+	if !qualifyingVersionFound {
+		// in case there are only expired versions in the consecutive minor version, pick the latest expired version
+		qualifyingVersionFound, latestVersion, err = GetLatestQualifyingVersion(cloudProfile.Spec.Kubernetes.Versions, FilterNonConsecutiveMinorVersion(*currentSemVerVersion), FilterSameVersion(*currentSemVerVersion))
+		if err != nil {
+			return false, "", err
+		}
+		if !qualifyingVersionFound {
+			return false, "", nil
+		}
+	}
+
+	return true, latestVersion.Version, nil
+}
+
+// GetLatestQualifyingVersion returns the latest expirable version from a set of expirable versions
+// A version qualifies if its classification is not preview and the optional predicate does not filter out the version.
+// If the predicate returns true, the version is not considered for the latest qualifying version.
+func GetLatestQualifyingVersion(versions []gardencorev1beta1.ExpirableVersion, predicate ...VersionPredicate) (qualifyingVersionFound bool, latest *gardencorev1beta1.ExpirableVersion, err error) {
+	latestSemanticVersion := &semver.Version{}
+	var latestVersion *gardencorev1beta1.ExpirableVersion
+OUTER:
+	for _, v := range versions {
+		if v.Classification != nil && *v.Classification == gardencorev1beta1.ClassificationPreview {
+			continue
+		}
+
+		semver, err := semver.NewVersion(v.Version)
+		if err != nil {
+			return false, nil, fmt.Errorf("error while parsing version '%s': %s", v.Version, err.Error())
+		}
+
+		for _, p := range predicate {
+			if p == nil {
+				continue
+			}
+
+			shouldFilter, err := p(v, semver)
+			if err != nil {
+				return false, nil, fmt.Errorf("error while evaluation predicate: '%s'", err.Error())
+			}
+			if shouldFilter {
+				continue OUTER
+			}
+		}
+
+		if semver.GreaterThan(latestSemanticVersion) {
+			latestSemanticVersion = semver
+			// avoid DeepCopy
+			latest := v
+			latestVersion = &latest
+		}
+	}
+	// unable to find qualified versions
+	if latestVersion == nil {
+		return false, nil, nil
+	}
+	return true, latestVersion, nil
+}
+
+// FilterDifferentMajorMinorVersion returns a VersionPredicate(closure) that evaluates whether a given version v has a different same major.minor version compared to the currentSemVerVersion
+// returns true if v has a different major.minor version
+func FilterDifferentMajorMinorVersion(currentSemVerVersion semver.Version) VersionPredicate {
+	return func(_ gardencorev1beta1.ExpirableVersion, v *semver.Version) (bool, error) {
+		isWithinRange, err := versionutils.CompareVersions(v.String(), "~", currentSemVerVersion.String())
+		if err != nil {
+			return true, err
+		}
+		return !isWithinRange, nil
+	}
+}
+
+// FilterNonConsecutiveMinorVersion returns a VersionPredicate(closure) that evaluates whether a given version v has a consecutive minor version compared to the currentSemVerVersion
+// returns true if v does not have a consecutive minor version
+func FilterNonConsecutiveMinorVersion(currentSemVerVersion semver.Version) VersionPredicate {
+	return func(_ gardencorev1beta1.ExpirableVersion, v *semver.Version) (bool, error) {
+		isWithinRange, err := versionutils.CompareVersions(v.String(), "^", currentSemVerVersion.String())
+		if err != nil {
+			return true, err
+		}
+
+		if !isWithinRange {
+			return true, nil
+		}
+
+		hasIncorrectMinor := currentSemVerVersion.Minor()+1 != v.Minor()
+		return hasIncorrectMinor, nil
+	}
+}
+
+// FilterSameVersion returns a VersionPredicate(closure) that evaluates whether a given version v is equal to the currentSemVerVersion
+// returns true it it is equal
+func FilterSameVersion(currentSemVerVersion semver.Version) VersionPredicate {
+	return func(_ gardencorev1beta1.ExpirableVersion, v *semver.Version) (bool, error) {
+		return v.Equal(&currentSemVerVersion), nil
+	}
+}
+
+// FilterLowerVersion returns a VersionPredicate(closure) that evaluates whether a given version v is lower than the currentSemVerVersion
+// returns true if it is lower
+func FilterLowerVersion(currentSemVerVersion semver.Version) VersionPredicate {
+	return func(_ gardencorev1beta1.ExpirableVersion, v *semver.Version) (bool, error) {
+		return v.LessThan(&currentSemVerVersion), nil
+	}
+}
+
+// FilterExpiredVersion returns a closure that evaluates whether a given expirable version is expired
+// returns true it it is expired
+func FilterExpiredVersion() func(expirableVersion gardencorev1beta1.ExpirableVersion, version *semver.Version) (bool, error) {
+	return func(expirableVersion gardencorev1beta1.ExpirableVersion, _ *semver.Version) (bool, error) {
+		return expirableVersion.ExpirationDate != nil && (time.Now().UTC().After(expirableVersion.ExpirationDate.UTC()) || time.Now().UTC().Equal(expirableVersion.ExpirationDate.UTC())), nil
+	}
 }
--- a/pkg/apis/core/v1beta1/helper/helpers_test.go
+++ b/pkg/apis/core/v1beta1/helper/helpers_test.go
@@ -17,10 +17,6 @@ package helper_test
 import (
 	"time"
 
-	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
-	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
-	. "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
-
 	"github.com/Masterminds/semver"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
@@ -29,12 +25,17 @@ import (
 	"github.com/onsi/gomega/types"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/pointer"
+
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	. "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 )
 
 var _ = Describe("helper", func() {
 	var (
-		trueVar  = true
-		falseVar = false
+		trueVar                 = true
+		falseVar                = false
+		expirationDateInThePast = metav1.Time{Time: time.Now().AddDate(0, 0, -1)}
 	)
 
 	Describe("errors", func() {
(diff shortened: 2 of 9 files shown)


danielfoehrKn:
Edit: Need to update the individual PRs to make the make verify work.

rfranzke:
Also, please drop the commit that is part of #2105 if possible.

danielfoehrKn:
> Can you add a release note, please?
> Also, please drop the commit that is part of #2105 if possible.

I think for now you can just select the mentioned commit. I will rebase as soon as #2105 is merged. 
Is that ok?

Also this PR does not depend on #2107 - the changes to the Maintenance Controller do not depend on the Version Management.

rfranzke:
Converted to a draft PR as its dependencies are not yet resolved. Please mark as 'ready for review' once they are resolved.

rfranzke:
@danielfoehrKn the description states

> ❌Depends on #2107

If this is correct, why is the PR ready for review?
If it's incorrect and doesn't depend on it then please remove it from the description.

vendor/github.com/gobuffalo/envy/.env
```diff
@@ -0,0 +1,5 @@
+# This is a comment
```
timebertt: Where does this come from?

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,120 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+	}
+
+	return true, latestVersion.Version, nil
+}
+
+// FilterQualifyingVersions removes versions from a set of expirable versions that do not qualify for updates during maintenance time
+// Removes preview versions, expired versions and versions based on a predicate
+// filters a version if the predicate returns true
+// returns a map[version]ExpirableVersion and a set of versions containing the filtered versions
+func FilterQualifyingVersions(versions []gardencorev1beta1.ExpirableVersion, predicate func(a semver.Version) (bool, error)) (map[string]gardencorev1beta1.ExpirableVersion, sets.String, error) {
```
timebertt: Normally I would expect that, if a predicate function returns true for a specific element, it is included and not filtered _out_.
I was also kind of confused by the doc string, which only says `filters a version if the predicate returns true`

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,120 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+		if v.ExpirationDate != nil && (time.Now().UTC().After(v.ExpirationDate.UTC()) || time.Now().UTC().Equal(v.ExpirationDate.UTC())) {
+			continue
+		}
+
+		if predicate != nil {
+			semver, err := semver.NewVersion(v.Version)
+			if err != nil {
+				return nil, nil, fmt.Errorf("error while parsing version '%s': %s", v.Version, err.Error())
+			}
+			shouldFilter, err := predicate(*semver)
```
timebertt: Hmm, why don't you use `a *semver.Version` as a parameter for predicate, if you anyways only pass a pointer?

pkg/apis/core/v1beta1/helper/helper.go
timebertt: Also why do you return both a map and a set? On the one hand, this is unneeded duplication, if I see it correctly. On the other hand, you only use the map to get the latest version, so why don't return only the latest version here directly?

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,120 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+
+// GetKubernetesVersionForPatchUpdate finds the latest Kubernetes patch version for its minor version in the <cloudProfile> compared
+// to the given <currentVersion>. Preview versions do not qualify for the kubernetes auto update. In case it does not find a newer patch version, it returns false. Otherwise,
+// true and the found version will be returned.
+func GetKubernetesVersionForPatchUpdate(cloudProfile *gardencorev1beta1.CloudProfile, currentVersion string) (bool, string, error) {
+	v, err := semver.NewVersion(currentVersion)
+	if err != nil {
+		return false, "", err
+	}
+	currentSemVerVersion := *v
```
timebertt: Why do you need an extra variable here? Can't you work with the pointer?

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,120 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+// returns a map[version]ExpirableVersion and a set of versions containing the filtered versions
+func FilterQualifyingVersions(versions []gardencorev1beta1.ExpirableVersion, predicate func(a semver.Version) (bool, error)) (map[string]gardencorev1beta1.ExpirableVersion, sets.String, error) {
+	filteredVersions := make(map[string]gardencorev1beta1.ExpirableVersion, len(versions))
+	filtered := sets.String{}
+	for _, v := range versions {
+		if v.Classification != nil && *v.Classification == gardencorev1beta1.ClassificationPreview {
+			continue
+		}
+
+		if v.ExpirationDate != nil && (time.Now().UTC().After(v.ExpirationDate.UTC()) || time.Now().UTC().Equal(v.ExpirationDate.UTC())) {
```
timebertt: We should also filter out versions, that are less than the current version, right?

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,120 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+// If a version can be found, returns true and the latest patch version of the next minor version.
+// In case it does not find a version, it returns false.
+func GetKubernetesVersionForMinorUpdate(cloudProfile *gardencorev1beta1.CloudProfile, currentVersion string) (bool, string, error) {
+	v, err := semver.NewVersion(currentVersion)
+	if err != nil {
+		return false, "", err
+	}
+	currentSemVerVersion := *v
+
+	predicate := func(v semver.Version) (bool, error) {
```
timebertt: Suggestion: WDYT about moving these Predicates to separate functions like `SameMinorVersionPredicate` or similar names to improve readability and testability?

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -18,6 +18,16 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/sirupsen/logrus"
```
timebertt: Move down to 3rd party import block

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -239,25 +283,33 @@ func hasMaintainNowAnnotation(shoot *gardencorev1beta1.Shoot) bool {
...
+			if err != nil {
+				return nil, nil, err
+			}
+
+			if !shouldBeUpdated {
+				return nil, nil, nil
+			}
+
+			message := fmt.Sprintf("image of worker-pool '%s' from version '(%s/%s)' to version '(%s/%s)'. Reason: %s", worker.Name, workerImage.Name, *workerImage.Version, updatedMachineImage.Name, *updatedMachineImage.Version, *reason)
+			reasonForUpdate = &message
```
timebertt: It seems like you are overwriting the reason here, if there are multiple pools and in that case, you will only have one message to output.

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -239,25 +283,33 @@ func hasMaintainNowAnnotation(shoot *gardencorev1beta1.Shoot) bool {
...
-		if err != nil {
-			return nil, err
-		}
-
-		if shouldBeUpdated {
-			shootMachineImagesForUpdate = append(shootMachineImagesForUpdate, shootMachineImage)
+	var reasonForUpdate *string
+	for _, worker := range shoot.Spec.Provider.Workers {
+		workerImage := worker.Machine.Image
+		if workerImage != nil {
```
timebertt: `worker.Machine.Image` can not be nil, because it is defaulted, right?
I'm just asking to make sure, we deliver patch updates to workers without a specific machine image selected (i.e. the default image)

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -191,42 +193,84 @@ func (c *defaultMaintenanceControl) Maintain(shootObj *gardencorev1beta1.Shoot,
...
+			return &latestPatchVersionNewMinor, &msg, nil
 		}
 	}
-	return nil, nil
+	return nil, nil, nil
 }
 
-func shouldKubernetesVersionBeUpdated(shoot *gardencorev1beta1.Shoot, profile *gardencorev1beta1.CloudProfile) (bool, error) {
-	versionExistsInCloudProfile, offeredVersion, err := gardencorev1beta1helper.KubernetesVersionExistsInCloudProfile(profile, shoot.Spec.Kubernetes.Version)
+func shouldKubernetesVersionBeUpdated(shoot *gardencorev1beta1.Shoot, profile *gardencorev1beta1.CloudProfile) (bool, *string, bool, error) {
```
timebertt: Can you please give the return parameters names, if you have multiple `bool`s, or at least add a doc string to make it easier to understand?

pkg/apis/core/v1beta1/helper/helper.go
timebertt: I am not sure, but I think the behaviour as described [here](https://github.com/gardener/gardener/blob/aaca736472c700f9cd9f8b4dbec06b9d2f4f90c0/docs/operations/versioning.md#forceful-migration-of-expired-versions) will not work:
> If that's expired as well, the update process repeats until a non-expired Kubernetes version is reached.
Depending on the circumstances described above, it can happen that the cluster receives multiple consecutive minor Kubernetes version updates! 

How will you do two consecutive minor upgrades (with the first upgrade being to an expired version) if you filter out all expired versions?

timebertt (changes requested):
I am not fully through yet (haven't looked at the tests), but I already have some comments.

pkg/apis/core/v1beta1/helper/helper.go
danielfoehrKn: mhm to me a predicate is just a function that evaluates to a boolean - in this filtering context case if it should be filtered out. Of course we could also inert the logic ...
What is confusing about the docs string?

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
danielfoehrKn: Shoot machine images are always set during maintenance time. If not set during creation, a default one is assigned. I removed the nil check.

pkg/apis/core/v1beta1/helper/helper.go
danielfoehrKn: Refactored the predicates and adjusted the docs string.

vendor/github.com/gobuffalo/envy/.env
danielfoehrKn: deleted it.

timebertt:
There are still conflicts in `test/framework/worker_utils.go`. Can you please rebase and resolve them?

danielfoehrKn:
PTAL

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,126 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+		latestVersion, err = GetLatestQualifyingVersion(cloudProfile.Spec.Kubernetes.Versions, FilterForConsecutiveMinor(*currentSemVerVersion), FilterSameVersion(*currentSemVerVersion))
+		if err != nil {
+			return false, "", nil
+		}
+	}
+
+	return true, latestVersion.Version, nil
+}
+
+// GetLatestQualifyingVersion returns the latest MachineImage version from a set of expirable versions
```
timebertt:
```suggestion
// GetLatestQualifyingVersion returns the latest expirable version from a set of expirable versions
```

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,126 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+// true and the found version will be returned.
+func GetKubernetesVersionForPatchUpdate(cloudProfile *gardencorev1beta1.CloudProfile, currentVersion string) (bool, string, error) {
+	currentSemVerVersion, err := semver.NewVersion(currentVersion)
+	if err != nil {
+		return false, "", err
+	}
+
+	latestVersion, err := GetLatestQualifyingVersion(cloudProfile.Spec.Kubernetes.Versions, FilterForSameMajorMinorVersion(*currentSemVerVersion), FilterSameVersion(*currentSemVerVersion), FilterExpiredVersionPredicate())
+	if err != nil {
+		return false, "", nil
```
timebertt, line 723: Why are you not returning the error here? What if there was an error like version not parseable or whatever?
Maybe you can make `GetLatestQualifyingVersion()` to return `nil` as a first return parameter and use that as an indicator that there was no qualifying version found instead of returning an error in that case.

pkg/apis/core/v1beta1/helper/helper.go
timebertt: Nice, clear now and also much easier to understand now, thank you :)

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,126 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+func GetKubernetesVersionForMinorUpdate(cloudProfile *gardencorev1beta1.CloudProfile, currentVersion string) (bool, string, error) {
+	currentSemVerVersion, err := semver.NewVersion(currentVersion)
+	if err != nil {
+		return false, "", err
+	}
+
+	latestVersion, err := GetLatestQualifyingVersion(cloudProfile.Spec.Kubernetes.Versions, FilterForConsecutiveMinor(*currentSemVerVersion), FilterSameVersion(*currentSemVerVersion), FilterExpiredVersionPredicate())
+	if err != nil {
+		// in case there are only expired versions in the consecutive minor version, pick the latest expired version
+		latestVersion, err = GetLatestQualifyingVersion(cloudProfile.Spec.Kubernetes.Versions, FilterForConsecutiveMinor(*currentSemVerVersion), FilterSameVersion(*currentSemVerVersion))
```
timebertt: Like the composability here with the different predicates 👍
Now it's quite easy to understand, what's going on here :)

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,126 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+				continue OUTER
+			}
+		}
+
+		filteredVersions[v.Version] = v
+	}
+	if len(filteredVersions) == 0 {
+		return gardencorev1beta1.ExpirableVersion{}, fmt.Errorf("no latest qualifying version could be determined")
+	}
+	latestVersion := sets.StringKeySet(filteredVersions).List()[len(filteredVersions)-1]
```
timebertt: Hmm, why are you using a map here? I think, this would even only work, if the versions are correctly sorted in the cloudprofile. 
WDYT about only storing the latest qualifying version which was found, and when you find another qualifying version, check if it is newer and overwrite it?

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,126 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+	return filteredVersions[latestVersion], nil
+}
+
+// FilterForConsecutiveMinor returns a closure that evaluates whether a given version v has a consecutive minor version compared to the currentSemVerVersion
+// returns true if v has a consecutive minor version
```
timebertt: wrong doc string

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,126 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+func FilterForSameMajorMinorVersion(currentSemVerVersion semver.Version) func(expirableVersion gardencorev1beta1.ExpirableVersion, version *semver.Version) (bool, error) {
```
timebertt: WDYT about introducing a type alias for the predicates? You write the func type out many times now.

```suggestion
func FilterForSameMajorMinorVersion(currentSemVerVersion semver.Version) VersionPredicate {
```

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +699,126 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+// returns true it it is equal
+func FilterSameVersion(currentSemVerVersion semver.Version) func(expirableVersion gardencorev1beta1.ExpirableVersion, version *semver.Version) (bool, error) {
+	return func(_ gardencorev1beta1.ExpirableVersion, v *semver.Version) (bool, error) {
+		return v.Equal(&currentSemVerVersion), nil
+	}
+}
+
+// FilterExpiredVersionPredicate returns a closure that evaluates whether a given expirable version is expired
+// returns true it it is expired
+func FilterExpiredVersionPredicate() func(expirableVersion gardencorev1beta1.ExpirableVersion, version *semver.Version) (bool, error) {
```
timebertt: Can you rename to keep the predicate names consistent?
```suggestion
func FilterExpiredVersion() func(expirableVersion gardencorev1beta1.ExpirableVersion, version *semver.Version) (bool, error) {
```

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -18,6 +18,16 @@ import (
 	"fmt"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
```
timebertt: sort imports

pkg/apis/core/v1beta1/helper/helper.go
timebertt: Ping, is this resolved?

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -191,42 +193,92 @@ func (c *defaultMaintenanceControl) Maintain(shootObj *gardencorev1beta1.Shoot,
...
+		sendMaintenanceCompletedEvent(c.recorder, shoot, fmt.Sprintf("%s %s %s.", msg, "Updated ", *reasonForKubernetesUpdate))
+	}
+
+	if updatedMachineImages != nil {
+		for _, reason := range reasonForImageUpdatePerPool {
+			sendMaintenanceCompletedEvent(c.recorder, shoot, fmt.Sprintf("%s %s %s.", "Maintenance for worker pool completed.", "Updated ", reason))
+		}
+	}
+
+	shootLogger.Infof("[SHOOT MAINTENANCE] completed for %q/%q", shoot.Namespace, shoot.Name)
```
timebertt:
```suggestion
	shootLogger.Infof("[SHOOT MAINTENANCE] completed for '%s/%s'", shoot.Namespace, shoot.Name)
```

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -191,42 +193,92 @@ func (c *defaultMaintenanceControl) Maintain(shootObj *gardencorev1beta1.Shoot,
...
+			return &latestPatchVersionNewMinor, &msg, nil
 		}
 	}
-	return nil, nil
+	return nil, nil, nil
 }
 
-func shouldKubernetesVersionBeUpdated(shoot *gardencorev1beta1.Shoot, profile *gardencorev1beta1.CloudProfile) (bool, error) {
-	versionExistsInCloudProfile, offeredVersion, err := gardencorev1beta1helper.KubernetesVersionExistsInCloudProfile(profile, shoot.Spec.Kubernetes.Version)
+func shouldKubernetesVersionBeUpdated(shoot *gardencorev1beta1.Shoot, profile *gardencorev1beta1.CloudProfile) (shouldBeUpdated bool, updateVersion *string, isExpired bool, error error) {
```
timebertt:
```suggestion
func shouldKubernetesVersionBeUpdated(shoot *gardencorev1beta1.Shoot, profile *gardencorev1beta1.CloudProfile) (shouldBeUpdated bool, reason *string, isExpired bool, error error) {
```

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -272,27 +330,38 @@ func determineMachineImage(cloudProfile *gardencorev1beta1.CloudProfile, shootMa
...
+		// check if already on latest version
+		latestShootMachineImage, err := gardencorev1beta1helper.GetLatestQualifyingShootMachineImage(*machineImage)
 		if err != nil {
-			return false, nil, fmt.Errorf("failure while updating machineImage to the latest version: %s", err.Error())
+			return false, nil, nil, fmt.Errorf("failed to determine latest machine image in cloud profile: %s", err.Error())
 		}
 
-		return true, shootMachineImage, nil
-	}
+		if latestShootMachineImage.Version == shootMachineImage.Version {
```
timebertt: Why are you filtering the same version here and not in your helper like you do for k8s versions?

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -239,25 +291,31 @@ func hasMaintainNowAnnotation(shoot *gardencorev1beta1.Shoot) bool {
...
 
-		if shouldBeUpdated {
-			shootMachineImagesForUpdate = append(shootMachineImagesForUpdate, shootMachineImage)
+		if !shouldBeUpdated {
+			return nil, nil, nil
 		}
-	}
 
-	return shootMachineImagesForUpdate, nil
+		message := fmt.Sprintf("image of worker-pool '%s' from version '(%s/%s)' to version '(%s/%s)'. Reason: %s", worker.Name, workerImage.Name, *workerImage.Version, updatedMachineImage.Name, *updatedMachineImage.Version, *reason)
```
timebertt: The image name will always stay the same, correct? Or am I mistaken?

timebertt (changes requested):

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
danielfoehrKn: Updated it thanks, overlooked it.

danielfoehrKn: Because then we can determine the case that a machine image [has already the required version ](https://github.com/gardener/gardener/blob/6b625621ee85e19edfafce8ffc0161edfcf61bdc/pkg/controllermanager/controller/shoot/shoot_maintenance_control.go#L346)(so no need for an update request)

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
danielfoehrKn: Yes correct, I just found it nice to prefix it with the machine image name - but I can change that.

timebertt: Sure, including the image name is good, I think. I just noticed that the message will include it twice :)

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
timebertt: Got it 👍

pkg/apis/core/v1beta1/helper/helper.go
danielfoehrKn: I added a flag so that the case 1) error occured 2) no qualified version found can be distinguished.

pkg/apis/core/v1beta1/helper/helpers_test.go
```diff
@@ -595,365 +663,427 @@ var _ = Describe("helper", func() {
...
-			"1.11.0",
-			"1.12.9",
-			true,
-		),
-
-		Entry("no new version",
-			gardencorev1beta1.CloudProfile{
-				Spec: gardencorev1beta1.CloudProfileSpec{
-					Kubernetes: kubernetesSettings,
+		FEntry("Expect error, because no latest version could be found",
```
timebertt: `cd pkg && ginkgo unfocus`.
```suggestion
		Entry("Expect error, because no latest version could be found",
```

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -18,6 +18,15 @@ import (
...
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/retry"
+	"github.com/sirupsen/logrus"
```
timebertt: sort imports, CI is failing here

pkg/apis/core/v1beta1/helper/helpers_test.go
```diff
@@ -595,365 +663,427 @@ var _ = Describe("helper", func() {
...
-			Expect(image.Version).To(PointTo(Equal(latestVersion)))
-		})
-
-		It("should return an error for invalid semVerVersion", func() {
-			inputImage := gardencorev1beta1.MachineImage{
-				Name: "coreos",
-				Versions: []gardencorev1beta1.ExpirableVersion{
-					{Version: "0.0.XX"},
-				},
+	DescribeTable("#GetLatestQualifyingShootMachineImage",
```
timebertt: Can you move that down under `Describe("Version helper", func() {`?

pkg/apis/core/v1beta1/helper/helpers_test.go
```diff
@@ -595,365 +663,427 @@ var _ = Describe("helper", func() {
...
+					Version: "1.17.1",
+				},
+				false,
+			),
+			Entry("Expect error, because no latest version could be found",
+				[]gardencorev1beta1.ExpirableVersion{},
+				false,
+				nil,
+				false,
+			),
```
timebertt: Hmm, you're saying "Expect error" but expect error is actually set to `false`.

pkg/apis/core/v1beta1/helper/helpers_test.go
```diff
@@ -595,365 +663,427 @@ var _ = Describe("helper", func() {
...
-			"1.12.9",
 			true,
-		),
-
-		Entry("no new version",
-			gardencorev1beta1.CloudProfile{
-				Spec: gardencorev1beta1.CloudProfileSpec{
-					Kubernetes: kubernetesSettings,
-				},
+			gardencorev1beta1.ShootMachineImage{
```
timebertt: Test panics here.
```suggestion
			&gardencorev1beta1.ShootMachineImage{
```

pkg/apis/core/v1beta1/helper/helpers_test.go
```diff
@@ -595,365 +663,427 @@ var _ = Describe("helper", func() {
...
+						{Version: "1.15.0"},
+						{Version: "1.14.4"},
+						{Version: "1.12.2"},
+					},
+					"",
+					false,
+				),
+			)
+			DescribeTable("Test version filter predicates",
+				func(predicate func(expirableVersion gardencorev1beta1.ExpirableVersion, version *semver.Version) (bool, error), version *semver.Version, expirableVersion gardencorev1beta1.ExpirableVersion, expectFilterVersion, expectError bool) {
```
timebertt:
```suggestion
				func(predicate VersionPredicate, version *semver.Version, expirableVersion gardencorev1beta1.ExpirableVersion, expectFilterVersion, expectError bool) {
```

pkg/apis/core/v1beta1/helper/helpers_test.go
```diff
@@ -595,365 +663,427 @@ var _ = Describe("helper", func() {
...
-			constraint        gardencorev1beta1.MachineImage
-			shootMachineImage gardencorev1beta1.ShootMachineImage
-		)
-		BeforeEach(func() {
-			constraint = gardencorev1beta1.MachineImage{
-				Name: "coreos",
-				Versions: []gardencorev1beta1.ExpirableVersion{
-					{
-						Version: "0.0.2",
+			DescribeTable("#GetKubernetesVersionForMinorUpdate",
```
timebertt, line 886: Can you add some cases here and also for machine images and predicates that lesser versions are not considered?

pkg/apis/core/v1beta1/helper/helpers_test.go
```diff
@@ -595,365 +663,427 @@ var _ = Describe("helper", func() {
...
+					gardencorev1beta1.ExpirableVersion{
+						ExpirationDate: &expirationDateInThePast,
+					},
+					true,
+					false,
+				),
+				Entry("Should not filter version.",
+					FilterExpiredVersion(),
+					nil,
+					gardencorev1beta1.ExpirableVersion{},
```
timebertt, line 1102: Can you add a non-expired expiration date here?
```suggestion
					gardencorev1beta1.ExpirableVersion{ExpirationDate: &metav1.Time{Time: time.Now().Add(time.Hour)}},
```

pkg/controllermanager/controller/shoot/shoot_maintenance_test.go
```diff
@@ -26,6 +27,9 @@ import (
 	"k8s.io/utils/pointer"
 )
 
+var (
+	previewClassification = gardencorev1beta1.ClassificationPreview
```
timebertt: Can you move that to the other variables declarations inside `Describe`?

pkg/controllermanager/controller/shoot/shoot_maintenance_test.go
```diff
@@ -247,24 +283,84 @@ var _ = Describe("Shoot Maintenance", func() {
...
+			// expire shoots kubernetes version 1.0.0
+			cloudProfile.Spec.Kubernetes.Versions[5].ExpirationDate = &expirationDateInThePast
+			shoot.Spec.Kubernetes = gardencorev1beta1.Kubernetes{Version: "1.0.0"}
+
+			// mark latest version 1.02 as preview
+			cloudProfile.Spec.Kubernetes.Versions[3].Classification = &previewClassification
+
+			version, _, err := MaintainKubernetesVersion(shoot, cloudProfile, shootLogger)
+
+			Expect(err).To(BeNil())
```
timebertt: `s/Expect(err).To(BeNil())/Expect(err).NotTo(HaveOccurred())/g`
```suggestion
			Expect(err).NotTo(HaveOccurred())
```

timebertt (changes requested):
Productive code looks quite good now, have some comments regarding the tests.

pkg/apis/core/v1beta1/helper/helpers_test.go
danielfoehrKn: thanks

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
danielfoehrKn: ok, that is a job for [importsort](https://github.com/danielfoehrKn/importsort)  :)

pkg/apis/core/v1beta1/helper/helpers_test.go
danielfoehrKn: Don't you think that is already tested here? We are handing over a set of versions in a CloudProfile and expect the latest qualifying version to be returned. So that is implicitly higher than a lower version of the shoot. Though I ll add a test for the FilterForSameMajorMinorVersion() predicate.

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -617,41 +605,24 @@ func ShootMachineImageVersionExists(constraint gardencorev1beta1.MachineImage, i
...
-		}
-	}
-	return latestSemVerVersion, latestMachineImageVersion, nil
-}
-
-// GetShootMachineImageFromLatestMachineImageVersion determines the latest version in a machine image and returns that as a ShootMachineImage
-func GetShootMachineImageFromLatestMachineImageVersion(image gardencorev1beta1.MachineImage) (*semver.Version, gardencorev1beta1.ShootMachineImage, error) {
-	latestSemVerVersion, latestImage, err := DetermineLatestMachineImageVersion(image)
+// GetLatestQualifyingShootMachineImage determines the latest qualifying version in a machine image and returns that as a ShootMachineImage
+// A version qualifies if it's classification is not preview and the version is not expired.
```
rfranzke:
```suggestion
// A version qualifies if its classification is not preview and the version is not expired.
```

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +701,146 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+	// unable to find qualified versions
+	if latestVersion == nil {
+		return false, nil, nil
+	}
+	return true, latestVersion, nil
+}
+
+// FilterForSameMajorMinorVersion returns a VersionPredicate(closure) that evaluates whether a given version v has the same major.minor version compared to the currentSemVerVersion
+// returns true if v has a the same major.minor version
+func FilterForSameMajorMinorVersion(currentSemVerVersion semver.Version) VersionPredicate {
```
rfranzke:
```suggestion
func FilterDifferentMajorMinorVersion(currentSemVerVersion semver.Version) VersionPredicate {
```

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +701,146 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+		if err != nil {
+			return true, err
+		}
+		return !isWithinRange, nil
+	}
+}
+
+// FilterForConsecutiveMinor returns a VersionPredicate(closure) that evaluates whether a given version v has a consecutive minor version compared to the currentSemVerVersion
+// returns true if v does not have a consecutive minor version
+func FilterForConsecutiveMinor(currentSemVerVersion semver.Version) VersionPredicate {
```
rfranzke:
```suggestion
func FilterNonConsecutiveMinorVersion(currentSemVerVersion semver.Version) VersionPredicate {
```

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -191,42 +192,92 @@ func (c *defaultMaintenanceControl) Maintain(shootObj *gardencorev1beta1.Shoot,
...
 
+	msg := "Maintenance completed."
+	// no update to Kubernetes and Machine Image version --> one event
+	if updatedKubernetesVersion == nil && updatedMachineImages == nil {
+		sendMaintenanceCompletedEvent(c.recorder, shoot, fmt.Sprintf("%s No action required.", msg))
+		return nil
+	}
+
+	if updatedKubernetesVersion != nil {
+		sendMaintenanceCompletedEvent(c.recorder, shoot, fmt.Sprintf("%s %s %s.", msg, "Updated ", *reasonForKubernetesUpdate))
```
rfranzke:
```suggestion
		sendMaintenanceCompletedEvent(c.recorder, shoot, fmt.Sprintf("Updated %s %s.", msg, *reasonForKubernetesUpdate))
```

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -191,42 +192,92 @@ func (c *defaultMaintenanceControl) Maintain(shootObj *gardencorev1beta1.Shoot,
...
 		return err
 	}
-	msg := "Completed; updated the Shoot specification successfully."
-	shootLogger.Infof("[SHOOT MAINTENANCE] %s", msg)
-	c.recorder.Eventf(shoot, corev1.EventTypeNormal, gardencorev1beta1.ShootEventMaintenanceDone, "%s", msg)
 
+	msg := "Maintenance completed."
+	// no update to Kubernetes and Machine Image version --> one event
+	if updatedKubernetesVersion == nil && updatedMachineImages == nil {
+		sendMaintenanceCompletedEvent(c.recorder, shoot, fmt.Sprintf("%s No action required.", msg))
```
rfranzke:
```suggestion
		sendMaintenanceCompletedEvent(c.recorder, shoot, fmt.Sprintf("%s - no action required.", msg))
```

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -191,42 +192,92 @@ func (c *defaultMaintenanceControl) Maintain(shootObj *gardencorev1beta1.Shoot,
...
+		return nil
+	}
+
+	if updatedKubernetesVersion != nil {
+		sendMaintenanceCompletedEvent(c.recorder, shoot, fmt.Sprintf("%s %s %s.", msg, "Updated ", *reasonForKubernetesUpdate))
+	}
+
+	if updatedMachineImages != nil {
+		for _, reason := range reasonForImageUpdatePerPool {
+			sendMaintenanceCompletedEvent(c.recorder, shoot, fmt.Sprintf("%s %s %s.", "Maintenance for worker pool completed.", "Updated ", reason))
```
rfranzke:
```suggestion
			sendMaintenanceCompletedEvent(c.recorder, shoot, fmt.Sprintf("Maintenance for worker pool completed. Updated %s.", reason))
```

rfranzke (commented):
Nice PR, I like it, well readable. Thanks, just a few nits.

pkg/apis/core/v1beta1/helper/helpers_test.go
timebertt: Sorry, I had to think it through again. Yes, for `GetKubernetesVersionForMinorUpdate` and `GetKubernetesVersionForPatchUpdate` you are doing the check via semver constraint checks (`~`/`^`) in the Predicates and that is now tested, perfect :).
For machine image versions you are not considering the current version in the helper, so thats also fine, no need for a check here.

But if ISIC the shoot maintenance controller will actually downgrade the shoots machine image the current image is expired and only lesser version exist, that are not expired.
For example this test for the maintenance controller, will fail:
```go
It("should not downgrade - even if highest version is expired", func() {
	highestExpiredVersion := gardencorev1beta1.ExpirableVersion{
		Version:        "1.1.2",
		Classification: &deprecatedClassification,
		ExpirationDate: &expirationDateInThePast,
	}
	cloudProfile.Spec.MachineImages[0].Versions = append(cloudProfile.Spec.MachineImages[0].Versions, highestExpiredVersion)
	shoot.Spec.Provider.Workers[0].Machine.Image.Version = &highestExpiredVersion.Version
	workerImages, _, err := MaintainMachineImages(testlogger, shoot, cloudProfile)

	Expect(len(workerImages)).NotTo(Equal(0))
	Expect(workerImages[0].Name).To(Equal(cloudProfile.Spec.MachineImages[0].Name))
	Expect(*workerImages[0].Version).To(Equal("1.1.2"))
})
```

pkg/apis/core/v1beta1/helper/helpers_test.go
timebertt: ping

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
timebertt: ping

pkg/controllermanager/controller/shoot/shoot_maintenance_test.go
timebertt: There are still quite a lot of `Expect(err).To(BeNil())` checks, can you replace them with `Expect(err).NotTo(HaveOccurred())`?

pkg/apis/core/v1beta1/helper/helpers_test.go
danielfoehrKn: Thanks for taking the time and checking. Yes when the CloudProfile is misconfigured that can happen and should be prevented.  I added a check.

I can see from your code that you would not expect an error in that scenario. While I think that there should be no maintenance error event on the Shoot (nothing the Shoot can do about it), I think the reconcile should throw and log an error to be retried eventually. The operator would need to step in and provide a proper version. 
This would also be in line how it is currently implemented in the Kubernetes version upgrade.

WDYT

timebertt: Actually, the code was just for demonstrating what I meant, haven't given it so much thought ;) 
Yeah, I think at least an error log would be appropriate for that case, but no strong opinion here.

danielfoehrKn:
PTAL

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/Masterminds/semver"
```
tim-ebert, line 21: sort imports

timebertt: sort imports

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -272,27 +331,46 @@ func determineMachineImage(cloudProfile *gardencorev1beta1.CloudProfile, shootMa
...
 		if err != nil {
-			return false, nil, fmt.Errorf("failure while updating machineImage to the latest version: %s", err.Error())
+			return false, nil, nil, fmt.Errorf("an error occured while determining the latest Shoot Machine Image for machine image %q: %s", machineImage.Name, err.Error())
 		}
 
-		return true, shootMachineImage, nil
-	}
+		// there should always be a qualifying version (at least the Shoot's machine image version itself).
+		if !qualifyingVersionFound {
+			return false, nil, nil, fmt.Errorf("no latest qualifying Shoot machine image could be determined for machine image %q. This is most likely a misconfiguration in the CloudProfile. Make sure the machine image in the CloudProfile has at least one version that is not expired, not in preview and greater or equal to the current Shoot image version %q", machineImage.Name, *shootMachineImage.Version)
```
timebertt, line 358: Hmm, I think, this error will also be returned, if the shoot uses a preview image version, regardless of wether there are other supported versions or not, correct?
Can you enhance the logic to not return this message in that case?

pkg/controllermanager/controller/shoot/shoot_maintenance_test.go
```diff
@@ -196,22 +218,46 @@ var _ = Describe("Shoot Maintenance", func() {
 		It("should return an error - cloud profile has no matching (machineImage.name) machine image defined", func() {
 			cloudProfile.Spec.MachineImages = cloudProfile.Spec.MachineImages[1:]
 
-			_, err := MaintainMachineImages(shoot, cloudProfile, machineCurrentImages)
+			_, _, err := MaintainMachineImages(testlogger, shoot, cloudProfile)
 
 			Expect(err).NotTo(BeNil())
 		})
+
+		It("should return an error and not downgrade shoot - even if highest version is expired", func() {
```
timebertt: Nice, can you also do that for the case, that the shoot uses a preview version?

timebertt: Btw, what is the desired behaviour, if the Shoot uses a preview version and there is a higher supported version? (Not sure, if that would be even a valid configuration of a CloudProfile - I'm losing track of all the different scenarios and rules 🙈).
Will it then be upgraded to the higher supported version?

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -782,3 +702,154 @@ func FindPrimaryDNSProvider(providers []gardencorev1beta1.DNSProvider) *gardenco
...
+// FilterSameVersion returns a VersionPredicate(closure) that evaluates whether a given version v is equal to the currentSemVerVersion
+// returns true it it is equal
+func FilterSameVersion(currentSemVerVersion semver.Version) VersionPredicate {
+	return func(_ gardencorev1beta1.ExpirableVersion, v *semver.Version) (bool, error) {
+		return v.Equal(&currentSemVerVersion), nil
+	}
+}
+
+// FilterLowerVersion returns a VersionPredicate(closure) that evaluates whether a given version v is lower to the currentSemVerVersion
+// returns true it it is lower
```
timebertt:
```suggestion
// FilterLowerVersion returns a VersionPredicate(closure) that evaluates whether a given version v is lower than the currentSemVerVersion
// returns true if it is lower
```

timebertt (changes requested):
Nice, only nits left :)

pkg/controllermanager/controller/shoot/shoot_maintenance_test.go
danielfoehrKn: Yes it would upgrade to the higher supported version. At the moment we do not validate against such configuration in the CloudProfile - I guess we should -as that makes no sense.

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
danielfoehrKn: Yup, good catch. Added a test case for that. Hopefully now all cases are covered :)

pkg/controllermanager/controller/shoot/shoot_maintenance_test.go
timebertt: Ok, will you take care of that?

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
danielfoehrKn: test case for [the logic is here](https://github.com/gardener/gardener/blob/2e5c630352e0e5dc5db661b309a51b705e550fc1/pkg/controllermanager/controller/shoot/shoot_maintenance_test.go#L240).

pkg/controllermanager/controller/shoot/shoot_maintenance_test.go
danielfoehrKn: test case for [the logic is here](https://github.com/gardener/gardener/blob/2e5c630352e0e5dc5db661b309a51b705e550fc1/pkg/controllermanager/controller/shoot/shoot_maintenance_test.go#L240).

pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
danielfoehrKn: Added the [actual check here](https://github.com/gardener/gardener/blob/2e5c630352e0e5dc5db661b309a51b705e550fc1/pkg/controllermanager/controller/shoot/shoot_maintenance_control.go#L351).

and the test for it here:  test case for [the logic is here](https://github.com/gardener/gardener/blob/c1ea382c6f648c9bab8e5a8844d7c7f2ce341e18/pkg/controllermanager/controller/shoot/shoot_maintenance_test.go#L240).

pkg/controllermanager/controller/shoot/shoot_maintenance_test.go
danielfoehrKn: See above.

schrodit (approved):
lgtm from test framework side

timebertt (approved):
Nice, thanks for addressing my comments!
Looks good from my POV 🚀
/approve

rfranzke (approved):

PR #2118, opened 2020-03-30
Write UID to shoot status after operation creation

### LINKED ISSUE #1926: Shoot with wrong ExternalDomain cannot be created nor deleted
**What happened**:
We have met this problem recently:
A user tried to create a Shoot with a customized domain name, however, he did not also provide the dns credential needed for his customized domain.
His default domain should be like `<shoot_name>.<project_name>.shoot.<gardener_domain>`, however, this is what he set `<shoot_name>.<some_customized_name>.<gardener_domain>`.

So the creation of this Shoot failed with following error as expected:
```
Could not initialize a new operation for Shoot deletion: unable to figure out which secret should be used for dns
```

And then the user tried to delete this Shoot. Gardener went on by skipping the above dns check and continued with creating the resources, and in the end, was blocked by the same check at the beginning of Shoot deletion steps. And thus the deletion failed.

The fact that although the dns domain of the Shoot is not correct, the infrastructure resources are created anyway, is causing a lot of trouble since the Shoot cannot be deleted now.

We have tried several ways to solve this problem and found that manual and force deletion of the Shoot will not work: we can of course manually clean up the resources created for the Shoot, and then try to delete the Shoot specification itself. But the shoots.core.gardener.cloud resource cannot be deleted because:
1. After recent changes in admission controller, finalizers of Shoot cannot be removed unless Shoot deletion flow succeeded.
2. Manually force delete corresponding registry in Etcd doesn't help, we can still get the resource using kubectl.

The only way we found to solve this problem is to change its dns domain to a correct one. However, this cannot be done easily via simply changing the field in the specification, because it is immutable.
**We have to manually force update the field directly in Etcd.**

1. Perhaps the creation of such kind (with dns problem, for example) of Shoot should be refused? Because their fields in Shoot specification are immutable, users can't possibly change them if something goes wrong.
2. If the creation goes wrong when the creation has not yet begun. We should directly remove the Shoot api-resource instead of first create the infrastructure resources and delete them again.

1. Create a Shoot, customize the dns domain in Shoot.yaml. Use `<shoot_name>.<some_customized_name>.<gardener_domain>` instead of `<shoot_name>.<project_name>.shoot.<gardener_domain>`.
2. Delete the Shoot.

I didn't try this on other landscapes or for other cloud providers, because removal of the problematic Shoot is brutal and may cause problem if not cautious enough. But this is a gardenlet problem, so it most likely will happen on every landscape and for every cloud provider.

**Environment**:
ac-canary.

- Gardener version: v1.0.2
- Kubernetes version (use `kubectl version`): client 1.16.3, server 1.15.7
- Cloud provider or hardware configuration: coreos on alicloud
- Others:

timuthy:
The following PR improves a few points before a shoot is considered for reconciliation / deletion:
- The shoot `uid` is written to the shoot `status` right after the operation for the shoot could be initialized
- `shoot.spec.seedName == nil` conditions are removed since a filter is already applied before calling the event handlers https://github.com/gardener/gardener/blob/b375d42f6006d863a2f5cf1ec835a46c23174ce1/pkg/gardenlet/controller/shoot/shoot.go#L120

Fixes #1926

**Release note**:

```improvement operator
Shoot clusters for which an operation could never be created (e.g. because of invalid DNS configuration) can now be deleted right away.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control.go
@@ -164,6 +164,9 @@ func (c *Controller) updateShootStatusProcessing(shoot *gardencorev1beta1.Shoot,
 				Description:    message,
 				LastUpdateTime: metav1.Now(),
 			}
+			if len(shoot.Status.UID) == 0 {
+				shoot.Status.UID = shoot.UID
+			}
 			return shoot, nil
 		})
 	return err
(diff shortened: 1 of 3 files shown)


rfranzke (approved):

timebertt (approved):
/lgtm

PR #2128, opened 2020-04-02
Change ContainerRuntime deployment

### LINKED ISSUE #2117: Enabling CRI on existing worker pool fails
**What happened**:

Adding gvisor CRI to an existing Shoot worker pool does not install gvisor. 

Gvisor is successfully installed on the node
- Daemonset "containerd-gvisor" in the Shoot has been scheduled
- Worker nodes are properly annotated

1) Create a Shoot with containerd as container runtime.
2) After, add the gvisor CRI to the workers section like so:

```yaml
 cri:
   name: containerd
   containerRuntimes:
    - type: gvisor
```
3) Observe that the update is allowed by the API server
4) Wait until the Shoot has been reconciled successfully
5) Check that the Containerruntime resource has been successfully reconciled & the HealthCheck is ok for the deployed deamonSet
6) Check on the machine that runsc is not installed in /var/bin/containerruntimes

**Reason**

The problem is that the way the node labels are currently set, does not work when updating an existing worker pool to use gvisor.
[Currently only the kubelet adds the required label ](https://github.com/gardener/gardener/blob/51634b550b0819af24d2dd1a08d5a15288b48d46/charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags#L25)'containerruntime.worker.gardener.cloud/gvisor=true'   to the node. However that is only being done during node creation.
Relying on the kubelet to add the node label is therefore not sufficient.

> 5) Check that the Containerruntime resource has been successfully reconciled & the HealthCheck is ok for the deployed deamonSet

- Gvisor extension deploys the DaemonSet and then considers the reconciliation as successful
- HealthCheck for the Deamonset is successful because: 

```yaml
NAMESPACE     NAME                    DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR                                                                                  AGE
kube-system   calico-node             1         1         1       1            1           beta.kubernetes.io/os=linux                                                                    3h39m
kube-system   containerd-gvisor       0         0         0       0            0           containerruntime.worker.gardener.cloud/gvisor=true,worker.gardener.cloud/cri-name=containerd   3h20m
```
The gvisor daemonset expect the label "containerruntime.worker.gardener.cloud/gvisor=true" which does not exist.
Though the daemonset has not been applied.

**Possible Solution** 

The worker extension resource is able to add labels to nodes in the particular worker group. 
This can be configured in the Shoot via: 
```yaml
labels:
   containerruntime.worker.gardener.cloud/gvisor: "true"
```
So as a current workaround you can just add the label manually via the Shoot.yaml.

As a proper bugfix though I would like to propose removing the kubelet created labels and relying solely on the labels defined in the worker.  The Gardener, if CRI is configured, can just implicitly create those labels in the Worker spec.

I can take over this task.

**Environment**:

- Gardener version: Master
- Kubernetes version (use `kubectl version`): 1.17.1
- Cloud provider or hardware configuration: AWS - ubuntu
- Others:

danielfoehrKn:

- Deploy  one ContainerRuntime per Pool per container runtime. 
- Add ContainerRuntime required labels during Worker deployment instead of kubelet node registration 

Please see issue #2117 for more details.

This is the prerequisite for the gvisor extension to handle a ContainerRuntime resource per worker pool. 

Fixes #2117 

Introduces new fields in the ContainerRuntime CRD
- **workerGroup**: contains the name of the WorkerGroup / Pool this ContainerRuntime configuration belongs to
- **workerGroupSelector** contains the label selector to select the nodes of that pool. Technically this could be computed by each extension, however when adjusting the labels in the gardener, this would require manual adjustment in each cri extension -> decoupling.  
For instance at the moment we are using "worker.garden.sapcloud.io/group" and "worker.gardener.cloud/pool". When Gardener decides to drop the deprecated sapcloud label, we do not have to adjust the extensions. This mechanism can also be handy for testing.

I am open to changing the field names, maybe "workerPoolName" or similar is better.
**Release note**:

```improvement operator
Gardener now deploys one ContainerRuntime CRD per WorkerPool per container runtime. This enables container runtimes to be configurable on the worker pool level.
```
```improvement operator
Fixed a bug that prevented an existing worker pool using containerd to be updated with a container runtime (e.g gvisor).
```


pkg/apis/extensions/v1alpha1/types_containerruntime.go
```diff
@@ -66,6 +66,10 @@ type ContainerRuntimeList struct {
 type ContainerRuntimeSpec struct {
 	// BinaryPath is the Worker's machine path where container runtime extensions should copy the binaries to.
 	BinaryPath string `json:"binaryPath"`
+	// WorkerGroup specifies the worker group the ContainerRuntime should be available for.
+	WorkerGroup string `json:"workerGroup"`
+	// WorkerGroupSelector is the label selector the extension uses to match the nodes belonging to the worker group.
+	WorkerGroupSelector metav1.LabelSelector `json:"workerGroupSelector"`
```
rfranzke: Can we combine them and call them "worker pool"?

```yaml
workerPool:
  name: <name>
  selector:
    matchLabels:
      ...
```

rfranzke:
FYI: https://github.com/gardener/gardener/pull/2130 might be interesting to you... maybe we can merge it first and you rebase on top of it?

rfranzke:
Now that we deploy a dedicated `ContainerRuntime` per pool, can you revert #2101 please?

pkg/apis/extensions/v1alpha1/types_containerruntime.go
danielfoehrKn: Adjusted it.

pkg/operation/botanist/worker.go
```diff
@@ -78,6 +78,17 @@ func (b *Botanist) DeployWorker(ctx context.Context) error {
 			}
 		}
 
+		// add CRI labels selected by the RuntimeClass
+		if worker.CRI != nil && len(worker.CRI.ContainerRuntimes) > 0 {
```
rfranzke, line 82: Nice idea maintaining the labels this way! 💯 Great!

rfranzke (dismissed):
/lgtm :)

pkg/operation/botanist/containerruntime.go
```diff
@@ -39,85 +40,79 @@ import (
...
-						toApply = extensionsv1alpha1.ContainerRuntime{
-							ObjectMeta: metav1.ObjectMeta{
-								Name:      containerRuntime.Type,
-								Namespace: b.Shoot.SeedNamespace,
-							},
+				fns = append(fns, func(ctx context.Context) error {
+					_, err := controllerutil.CreateOrUpdate(ctx, b.K8sSeedClient.Client(), &toApply, func() error {
+						metav1.SetMetaDataAnnotation(&toApply.ObjectMeta, v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationReconcile)
+						toApply.Spec.BinaryPath = extensionsv1alpha1.ContainerDRuntimeContainersBinFolder
+						toApply.Spec.Type = cr.Type
```
zanetworker, line 64: doesn't this mean that we could have more than one container runtime per pool? or is this desired behavior?

danielfoehrKn: It is designed to have one ContainerRuntime CRD per Pool per type. For instance for worker with name "worker-1" having gVisor and Kata container runtime configured:  worker-1-gVisor & worker-1-kata

zanetworker: okay got it, I just got confused from the description of the PR 

>Deploy one ContainerRuntime per Pool per CRI. 

I should have noticed the capital letter initials :D

danielfoehrKn: The description was confusing, sorry. Updated it :)

danielfoehrKn:
@zanetworker can you review the changes in this PR / are there more topics to address?  Other work in the extensions depend on it. Thanks!

zanetworker:
@danielfoehrKn maybe it makes sense to modify / add the example of the shoot with the updated worker pool? otherwise lgtm.

danielfoehrKn:
> @danielfoehrKn maybe it makes sense to modify / add the example of the shoot with the updated worker pool? otherwise lgtm.

There are no changes in the Shoot - only the ContainerRuntime CRD has added fields :)

zanetworker:
> > @danielfoehrKn maybe it makes sense to modify / add the example of the shoot with the updated worker pool? otherwise lgtm.
> 
> There are no changes in the Shoot - only the ContainerRuntime CRD has added fields :)

I meant the worker configuration, which is probably documentation in the provider config.

zanetworker (dismissed):

rfranzke:
@zanetworker probably means https://github.com/gardener/gardener/blob/master/docs/extensions/containerruntime.md should be updated? :)

zanetworker:
https://github.com/gardener/gardener/pull/2128/files#diff-f030bf723b64aeefc41cc0f8f5b178dfR71 this is missing if you are going to update the documentation in the PR.

danielfoehrKn:
Added the documentation. PTAL

docs/extensions/containerruntime.md
```diff
@@ -26,7 +26,15 @@ metadata:
 spec:
   binaryPath: /var/bin/containerruntimes
   type: gvisor
+  workerPool:
+        name: worker-ubuntu
```
rfranzke: correct indentation

rfranzke (approved):

zanetworker (approved):

PR #2131, opened 2020-04-02
Revert "GEP 10 - Container runtime extensions - addded cr config validation"

danielfoehrKn:
This [PR](https://github.com/gardener/gardener/pull/2128) introduces the deployment of on ContainerRuntime CRD per WorkerPool per runtime.
WorkerPools can have different provider configs, so this check is not needed any more.

Fixes #

**Release note**:

```improvement operator

```


rfranzke:
Thanks Daniel, I think it should be part of your other PR #2128, WDYT?

PR #2133, opened 2020-04-03
Add "Do not edit" description to all Shoot artifacts

mvladev:

it adds the following annotation to all resources managed by `resource-manager`
```yaml
metadata:
  annotations:
    gardener.cloud/description: |
      DO NOT EDIT - This resource is managed by Gardener.
```

Fixes #

**Release note**:

```improvement user
All resources managed by `resource-manager` now have `gardener.cloud/description` annotation with `DO NOT EDIT` warning.
```


rfranzke:
Thanks @mvladev - nice idea. Though, wouldn't it make more sense to let GRM add the annotation itself instead of manually adding it to every single resource?

mvladev:
> Though, wouldn't it make more sense to let GRM add the annotation itself instead of manually adding it to every single resource?

I thought about it, but Some resources already [have this annotation](https://github.com/gardener/gardener/pull/2133/files#diff-e1f39896c426bda625f992d220ce4629L6-R9). Merging both would be a problem.

As an alternative, GRM could use different annotation.

rfranzke:
Yeah, maybe GRM can use `resource-manager.gardener.cloud/description` is something else prefixed with `resource-manager.gardener.cloud/` ?

mvladev:
Closing in favor of gardener/gardener-resource-manager#46

PR #2135, opened 2020-04-03
Do not try to sync extension resources if cluster object cannot be found

plkokanov:
This PR makes it so that extensions resources are not synced by the ShootState sync controller if the corresponding `Cluster` resource does not exist. This means that the Shoot and ShootState have most likely been deleted.

None

None

**Release note**:

```improvement operator
Extension resources are no longer synced by the ShootState Sync Controller if the corresponding `Cluster` resource does not exist.
```


rfranzke (approved):
Thanks! :)

danielfoehrKn (approved):

PR #2137, opened 2020-04-03
[GEP-10] Add validation for CRIs and CRs

vpnachev:
Now every machine image version can have optional lists of
- supported container runtime interfaces, which is limited to the well know CRIs - currently only `containerd`, and silently `docker`.
- supported container runtimes

Fixes #

**Release note**:

```improvement operator
Now every machine image version in the `CloudProfile` can specify list of supported container runtime interfaces and container runtimes.
```
```action operator
With this version of Gardener, a validation for supported container runtimes (CR) and container runtime interfaces(CRI) per machine image version has been introduced. To prevent disruptions for creation and update of shoot cluster using non default CR and CRI, the `CloudProfiles` needs to be enhanced with the list of supported CR and CRI per machine image version. An example can be found [here](https://github.com/gardener/gardener/blob/master/example/30-cloudprofile.yaml)
```


timebertt:
Thanks for the PR @vpnachev!
Can you add an example [here](https://github.com/gardener/gardener/blob/master/example/30-cloudprofile.yaml)?

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -189,6 +189,12 @@ func validateMachineImages(machineImages []core.MachineImage, fldPath *field.Pat
 			if err != nil {
 				allErrs = append(allErrs, field.Invalid(versionsPath.Child("version"), machineVersion.Version, "could not parse version. Use SemanticVersioning. In case there is no semVer version for this image use the extensibility provider (define mapping in the ControllerRegistration) to map to the actual non-semVer version"))
 			}
+
+			for j, cri := range machineVersion.ContainerRuntimeInterfaces {
+				if !avaliableWorkerCRINames.Has(string(cri)) {
```
zanetworker: should we also cross check CRIs with container runtimes supported / CRI?

pkg/apis/core/types_cloudprofile.go
```diff
@@ -78,8 +78,20 @@ type KubernetesSettings struct {
...
+}
+
+// MachineImageVersion is an expirable version with list of supported container runtimes and interfaces
+type MachineImageVersion struct {
+	// Version is the version identifier.
+	Version string
+	// ExpirationDate defines the time at which this version expires.
+	ExpirationDate *metav1.Time
+	// ContainerRuntimeInterfaces is list of container runtime interface names that are available for this image version.
+	ContainerRuntimeInterfaces []string
```
danielfoehrKn: I am not sure if we should put the Container Runtime and CRI information into each version - this will mean a lot of duplicate information when the CRI and runtime list grows. Maybe we can find a better way - e.g on the  a dedicated CRI section with reference to the images.

Not every container runtime interface can work with every container runtime (e.g CRI-O cannot work with gvisor et al.). I do not see a way to express that here. I think thats another reason for a dedicated CRI section instead.

Also I think it is kind of nice to have the expirable versions for machine images and kubernetes versions. This is also being taken advantage of when implementing the [Version Management as part of GEP5](https://github.com/gardener/gardener/pull/2107). 

```yaml
  cri:
    - name: containerd
      machineImage:
        - name: ubuntu
          versionConstraint: ">=16.04.0"
      containerRuntimes:
        - gvisor
        - kata
    - name: cri-o
      machineImage:
        - name: ubuntu
          versions:
            - "18.04.0"
            - "19.04.0"
            - "19.10.0"
      containerRuntimes:
        - podman
```

Regarding the example above: gVisor for instance needs Linux Kernel version > 4.14.77+, and Ubuntu 16.04 that is required for containerd is already [based on kernel 4.4](https://wiki.ubuntu.com/XenialXerus/ReleaseNotes). It seems to me that when containerd is supported, then the container runtimes are also supported -> no need to separately restrict the container runtimes to the os.  I would have to check that for others too.

zanetworker: and that would also address my comment to cross-check CRI with Container runtimes. One point though, some runtimes (e.g., gvisor) would require a certain version of the OS, and although containerd is right on there, the os version required by containerd can be less than the one required by gvisor and so it can lead to unexpected side-effects, i.e., you cant assume that gvisor is supported as long as containerd is on the system. 

One way to work around this is to always have the minimum version of the image be the minimum supported version of all the supported container runtimes i.e., min(gvisor,kata)

danielfoehrKn: Yes I think that would make sense. This way we are not overcomplicating it. Also we only miss out on some small portion of the versions.

vpnachev: First of all CRI is maintained in-tree by the gardener, https://github.com/gardener/gardener/blob/82af7cd270c092cd07a32455d86c38643a299e74/charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.service#L9-L16

And nowadays the CRI can be either omitted or a list with a single element `containerd`. And container runtimes rely on `containerd` following the contract in the [GEP](https://github.com/gardener/gardener/blob/master/docs/proposals/10-shoot-additional-container-runtimes.md)

> The first phase would be to allow the change of the Kubelet configuration as described here so that Kubernetes would use containerd instead of the default Dockershim

So the contract in the GEP implies that every container runtime, except docker, must be working with `containerd`. 

I was thinking of a list of unsupported CR(I)s and started the implementation in this way, but then the operator is responsible to explicitly disable broken CR(I)s for particular image and version. Thus I switched to list of enabled CR(I)s. 



> I am not sure if we should put the Container Runtime and CRI information into each version - this will mean a lot of duplicate information when the CRI and runtime list grows.

We can make CRI configuration also on image level if it is applicable to all versions. What do you think?

> Maybe we can find a better way - e.g on the a dedicated CRI section with reference to the images.

How does this help to keep the list short? Yes, semver version restrictions can be used, but this does not guarantee a shorter list.  The cloud profile is a resource that maintain all constraints for the environment, thus I prefer it to be as explicit as possible. 

> Also I think it is kind of nice to have the expirable versions for machine images and kubernetes versions. This is also being taken advantage of when implementing the Version Management as part of GEP5.

I do not understand, every image version still have its own `expirationDate` field so what do you mean?

pkg/apis/core/validation/cloudprofile.go
vpnachev: No, the [GEP](https://github.com/gardener/gardener/blob/master/docs/proposals/10-shoot-additional-container-runtimes.md#proposal) implies that every CR rely on `containerd`.

pkg/apis/core/types_cloudprofile.go
danielfoehrKn: > And container runtimes rely on containerd following the contract in the GEP

Yes, but containerd is not the only CRI available, maybe we will add for instance CRI-O in the future or other ones. I think that below can be interpreted  that crio-o can use gVisor as the container runtime, but only the combination containerd -> gvisor && containerd-> kata are allowed. :
```yaml 
    containerRuntimeInterfaces:
     - containerd
     - cri-o
     containerRuntimes:
     - gvisor
     - kata-containers
```

> How does this help to keep the list short? Yes, semver version restrictions can be used, but this does not guarantee a shorter list. The cloud profile is a resource that maintain all constraints for the environment, thus I prefer it to be as explicit as possible.

Because we do not repeat the information which CRI and container runtime is support for each and every version.For instance containerd and gvisor can work with any ubuntu if the linux kernel version is > 4.14.77. So we do not have to repeat the same information that containerd is supported every time.
Rather we can express that with a semVer constraint. 
If that is not applicable, e.g crio-o and ubuntu (see above) - we can still specify single versions. 

>  I do not understand, every image version still have its own expirationDate field so what do you mean?

GEP5 simplifies the handling of machine image versions and kubernetes versions by using common functionality for expirableVersions. If machine images do not use expirableVersions any more, that cannot be done / or needs a hacky conversion. 
[See here](https://github.com/gardener/gardener/pull/2107/files#diff-9cb87d91362ac5df8e525e234ee40a3eR47).

vpnachev: > I think that below can be interpreted that crio-o can use gVisor as the container runtime, but only the combination containerd -> gvisor && containerd-> kata are allowed. 

Generally you are right, but `gvisor` and `kata` extensions can implement a validation webhook to check whether they are configured with proper CRI.

I am afraid that if we move CRI configs to dedicated field, we will more or less repeat the image names and versions there. So we are just moving the duplicated information.

danielfoehrKn: I feel like the approach with Webhooks is too cumbersome - we do not need that with the above mentioned approach.

> I am afraid that if we move CRI configs to dedicated field, we will more or less repeat the image names and versions there. So we are just moving the duplicated information.

Some redundancy is not avoidable here I think. But I do not see why moving the CRI information like describe in my previous comment increases redundancy, it reduces it compared to moving it to the image or the machine image version.

I feel like we need another opinion here.

rfranzke: Valid points, thanks for bringing up the doubts. After thinking about it I tend to agree more with @vpnachev. The proposal of @danielfoehrKn looks like a lot of double maintenance wrt machine images (it's better/easier-to-maintain to store the CRI/container runtime information directly where a machine image version is defined).

Though, while I understand the webhook validation suggestion, I think for this use-case it's not perfectly applicable. From end-user's pov you don't see a potential limitation (e.g., if gvisor cannot be used with cri-o) - but that's one important use-case of the `CloudProfile`.

So, what about this:

```yaml
spec:
  machineImages:
  - name ubuntu:
    versions:
    - version: 1.2.3
      cri:
      - name: containerd
        containerRuntimes:
        - type: gvisor
        - type: kata
      - name: cri-o
        containerRuntimes:
        - type: kata
``` 

> Also I think it is kind of nice to have the expirable versions for machine images and kubernetes versions. This is also being taken advantage of when implementing the Version Management as part of GEP5.

If @vpnachev embeds the `core(/{v1alpha,v1beta1}).ExpireableVersion` type into the `core(/{v1alpha,v1beta1}).MachineImageVersion` type then you, @danielfoehrKn, should still be able to work with the one `ExpireableVersion` type used in both Kubernetes and machine image versions for your GEP5 PRs.

danielfoehrKn: Alright, so we'll have more duplicate information, but less maintenance. When adding a version,  we do not have to adjust another portion of the ClouProfile, but instead can just add the version. Fine with me.

> @danielfoehrKn, should still be able to work with the one ExpireableVersion type used in both Kubernetes and machine image versions for your GEP5 PRs.

Good suggestion. That would significantly decrease the needed refactoring.

rfranzke:
/kind/enhancement

rfranzke:
/area/usability
/area/robustness
/topology/garden
/platform/all
/priority/normal
/needs/changes

pkg/apis/core/v1alpha1/types_cloudprofile.go
```diff
@@ -97,10 +97,18 @@ type KubernetesSettings struct {
...
-	Versions []ExpirableVersion `json:"versions" patchStrategy:"merge" patchMergeKey:"version" protobuf:"bytes,2,rep,name=versions"`
+	Versions []MachineImageVersion `json:"versions" patchStrategy:"merge" patchMergeKey:"version" protobuf:"bytes,2,rep,name=versions"`
+}
+
+// MachineImageVersion is an expirable version with list of supported container runtimes and interfaces
+type MachineImageVersion struct {
+	ExpirableVersion `json:",inline" protobuf:"bytes,2,opt,name=expirableVersion"`
+	// CRI list of supported container runtime and interfaces supported by this version
+	// +optional
+	CRI []CRI `json:"cri,omitempty" protobuf:"bytes,1,opt,name=cri"`
```
rfranzke: Shouldn't it be?

```suggestion
	ExpirableVersion `json:",inline" protobuf:"bytes,1,opt,name=expirableVersion"`
	// CRI list of supported container runtime and interfaces supported by this version
	// +optional
	CRI []CRI `json:"cri,omitempty" protobuf:"bytes,2,rep,name=cri"`
```

pkg/apis/core/v1beta1/types_cloudprofile.go
```diff
@@ -97,10 +97,18 @@ type KubernetesSettings struct {
...
-	Versions []ExpirableVersion `json:"versions" patchStrategy:"merge" patchMergeKey:"version" protobuf:"bytes,2,rep,name=versions"`
+	Versions []MachineImageVersion `json:"versions" patchStrategy:"merge" patchMergeKey:"version" protobuf:"bytes,2,rep,name=versions"`
+}
+
+// MachineImageVersion is an expirable version with list of supported container runtimes and interfaces
+type MachineImageVersion struct {
+	ExpirableVersion `json:",inline" protobuf:"bytes,2,opt,name=expirableVersion"`
+	// CRI list of supported container runtime and interfaces supported by this version
+	// +optional
+	CRI []CRI `json:"cri,omitempty" protobuf:"bytes,1,opt,name=cri"`
```
rfranzke: same question like above

pkg/apis/core/validation/cloudprofile.go
```diff
@@ -290,7 +297,14 @@ func validateMachineImages(machineImages []core.MachineImage, fldPath *field.Pat
...
 			}
-			allErrs = append(allErrs, validateExpirableVersion(machineVersion, image.Versions, versionsPath)...)
+
+			allErrs = append(allErrs, validateExpirableVersion(machineVersion.ExpirableVersion, helper.ToExpirableVersions(image.Versions), versionsPath)...)
+
+			for j, cri := range machineVersion.CRI {
+				if !avaliableWorkerCRINames.Has(string(cri.Name)) {
+					allErrs = append(allErrs, field.NotSupported(versionsPath.Child("cri").Index(j), cri, avaliableWorkerCRINames.List()))
+				}
+			}
```
rfranzke: You could also check for duplicates in `.cri` and `.cri.containerRuntimes`.

plugin/pkg/shoot/validator/admission.go
```diff
@@ -766,6 +768,78 @@ func validateMachineImagesConstraints(constraints []core.MachineImage, image, ol
...
+
+	if worker.CRI == nil {
+		return nil
+	}
+
+	var machineImage *core.MachineImage
+	var machineVersion *core.MachineImageVersion
+
+	for _, image := range constraints {
+		if image.Name == worker.Machine.Image.Name {
```
rfranzke, line 784: Hm, nil check for `image`?

rfranzke (changes requested):
Great PR, thank you!
Apart from the inline comments, do you think it makes to extend https://github.com/gardener/gardener/blob/master/plugin/pkg/global/extensionvalidation/admission.go to prevent operators from creating `CloudProfile`s referencing container runtime types for which no `ControllerRegistration` exists? Well, now that I write/ask it, probably it comes with additional implementation effort (what if a `ControllerRegistration` for e.g. `gvisor` is removed but the referencing `CloudProfile`s are not updated? We need extra logic to prevent future update calls from failing). Maybe we shouldn't do it as `CloudProfile`s are anyways controlled by operators/admins, i.e., the same people that create `ControllerRegistration`s. What is your opinion?

pkg/apis/core/v1alpha1/types_cloudprofile.go
vpnachev: Actually `ExpirableVersion` should be fully inlined, just like TypeMeta is.

pkg/apis/core/validation/cloudprofile.go
vpnachev: done

plugin/pkg/shoot/validator/admission.go
vpnachev: It is highly unlikely to happen due to the defaulting, but I have added a check just in case.

vpnachev (commented):
> ... do you think it makes to extend https://github.com/gardener/gardener/blob/master/plugin/pkg/global/extensionvalidation/admission.go to prevent operators from creating CloudProfiles referencing container runtime types for which no ControllerRegistration exists? ... Maybe we shouldn't do it as CloudProfiles are anyways controlled by operators/admins, i.e., the same people that create ControllerRegistrations. What is your opinion?


To be honest, I think that the CR(I) fields in the cloud profile per machine image version should be more for the machine image itself, not for the controllers.
On the other side, we already have a validation on shoot creation/update that prevents usage of container runtime for which is missing controller registration 

I tried with cloud profile with `gvisor` enabled but have not installed ControllerRegistration for it and got the following error.
```
$ k apply -f aws-shoot.yaml
Error from server (Forbidden): error when creating "aws-shoot.yaml": shoots.core.gardener.cloud "local-aws" is forbidden: 1 error occurred:
        * given Shoot uses non-registered container runtime type: spec.provider.workers[0].cri.containerRuntimes[0].type ("gvisor")
```

PS. I know I am replying a bit late and the above validation might have been implemented after your comment.
PPS. It was implemented earlier with https://github.com/gardener/gardener/commit/51634b550b0819af24d2dd1a08d5a15288b48d46#diff-044e8b580dfa3e3a0d43cece0ecd29f9

plugin/pkg/shoot/validator/admission.go
```diff
@@ -766,6 +768,83 @@ func validateMachineImagesConstraints(constraints []core.MachineImage, image, ol
...
+	}
+
+	if machineImage == nil {
+		return nil
+	}
+
+	for _, version := range machineImage.Versions {
+		if version.Version == worker.Machine.Image.Version {
+			machineVersion = &version
+		}
```
```suggestion
			machineVersion = &version
			break
		}
```

plugin/pkg/shoot/validator/admission.go
```diff
@@ -766,6 +768,83 @@ func validateMachineImagesConstraints(constraints []core.MachineImage, image, ol
...
+	var (
+		allErrors = field.ErrorList{}
+		validCRIs = []string{}
+		foundCRI  *core.CRI
+	)
+
+	for _, criConstraint := range constraints {
+		validCRIs = append(validCRIs, string(criConstraint.Name))
+		if cri.Name == criConstraint.Name {
+			foundCRI = &criConstraint
```
rfranzke, line 820:
```suggestion
			foundCRI = &criConstraint
			break
```

rfranzke (commented):
Thanks for updating the PR - small suggestions, otherwise /lgtm

rfranzke (approved):
/lgtm

rfranzke:
Can you update the release note that it's now a required action to write the CRI + container runtime types into the `CloudProfile` to not disrupt creation of new shoots (earlier Gardener versions didn't have this validation)?

plugin/pkg/shoot/validator/admission.go
vpnachev: Done

vpnachev:
> Can you update the release note that it's now a required action to write the CRI + container runtime types into the CloudProfile to not disrupt creation of new shoots (earlier Gardener versions didn't have this validation)?

Good catch - added.

PR #2138, opened 2020-04-03
Make image version optional

### LINKED ISSUE #1960: Make machine image version optional
Make the machine image version optional and default it to the latest available version for the selected distribution.

**Why is this needed**:
Ease shoot creation without necessity to find proper version in the `CloudProfile` resource.

vpnachev:
Make image version optional

Fixes #1960 

**Release note**:

```improvement user
Now only machine image name can be specified in the shoot manifest, Gardener will default the version to the latest available version for the specified image.
```

### DIFF
--- a/plugin/pkg/shoot/validator/admission.go
+++ b/plugin/pkg/shoot/validator/admission.go
@@ -305,15 +305,6 @@ func (v *ValidateShoot) Admit(ctx context.Context, a admission.Attributes, o adm
 		}
 	}
 
-	// General approach with machine image defaulting in this code: Try to keep the machine image
-	// from the old shoot object to not accidentally update it to the default machine image.
-	// This should only happen in the maintenance time window of shoots and is performed by the
-	// shoot maintenance controller.
-	image, err := getDefaultMachineImage(cloudProfile.Spec.MachineImages)
-	if err != nil {
-		return apierrors.NewBadRequest(err.Error())
-	}
-
 	if !reflect.DeepEqual(oldShoot.Spec.Provider.InfrastructureConfig, shoot.Spec.Provider.InfrastructureConfig) {
 		if shoot.ObjectMeta.Annotations == nil {
 			shoot.ObjectMeta.Annotations = make(map[string]string)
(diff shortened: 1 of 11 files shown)


danielfoehrKn:
Can we wait for [GEP 5](https://github.com/gardener/gardener/pull/2105) to be discussed and merged? It has implications for this PR.
Thinks to consider
- default to preview versions or not (currently under discussion) - part of GEP5
- aligning it with defaulting of Kubernetes versions when only specifying major.minor version. Maybe we should be consistent for Machine image versions and Kubernetes versions.

vpnachev:
> aligning it with defaulting of Kubernetes versions when only specifying major.minor version. Maybe we should be consistent for Machine image versions and Kubernetes versions.

Kubernetes versions are following versioning schema similar to semver, machine images are not doing this. Also, from our point of view minor/major OS image upgrade is not considered as backward incompatible. Thus specifying only `major.minor` does not make sense.  

I prefer to get this PR in now.

timuthy:
> Thus specifying only major.minor does not make sense.

Do you only think of shoot update events? Because it may be a valid use case when creating a shoot.

danielfoehrKn:
Alright, no need to wait for GEP 5, just want to make sure that all open questions are addressed.

vpnachev:
@timuthy could you elaborate?

Machine image version is defaulted only when a worker pool is created, e.g. completely new shoot or new worker pool for already existing shoot. The maintenance controller can update the version to the latest one depending on auto update settings.

timuthy:
I just haven't understood your statement:

> Thus specifying only major.minor does not make sense.

But let's move that discussion to the GEP-5 related PRs. Your input towards this would be very valuable.

plugin/pkg/shoot/validator/admission.go
```diff
@@ -723,9 +738,24 @@ func validateMachineImagesConstraints(constraints []core.MachineImage, image, ol
 	return false, validValues
 }
 
-func getOldWorkerMachineImageOrDefault(workers []core.Worker, name string, defaultImage *core.ShootMachineImage) *core.ShootMachineImage {
-	if oldWorker := helper.FindWorkerByName(workers, name); oldWorker != nil && oldWorker.Machine.Image != nil {
-		return oldWorker.Machine.Image
+func getOldWorkerMachineImageOrDefault(oldWorkers []core.Worker, worker core.Worker, images []core.MachineImage) (*core.ShootMachineImage, error) {
```
timuthy: Can we name it `ensureMachineImage` or similar? Because now it also handles the `worker.Machine.Image != nil` cases which was previously done by the caller.

timuthy (commented):
Looks generally fine, nice feature! Just one nit.

plugin/pkg/shoot/validator/admission.go
vpnachev: done

timuthy (dismissed):
/lgtm

Can you resolve the merge conflicts? Then we can merge it.

vpnachev:
Conflicts are resolved now

timuthy (dismissed):

rfranzke (approved):

PR #2140, opened 2020-04-05
Add tests for FileContent func

ialidzhikov:
In the past we had a small regression with this FileContent func - ref gardener/gardener#2008. The PR adds test to prevent similar case.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/chartrenderer/default_test.go
+++ b/pkg/chartrenderer/default_test.go
@@ -1,0 +1,101 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package chartrenderer_test
+
+import (
+	"path/filepath"
+
+	"github.com/gardener/gardener/pkg/chartrenderer"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/version"
+	"k8s.io/client-go/discovery/fake"
+	"k8s.io/client-go/testing"
+	"k8s.io/helm/pkg/engine"
+)
+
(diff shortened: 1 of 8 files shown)


pkg/chartrenderer/default_test.go
```diff
@@ -0,0 +1,102 @@
...
+	})
+
+	Describe("#Render", func() {
+		It("should return err when chartPath is missing", func() {
+			_, err := renderer.Render(filepath.Join("testdata", "missing"), "missing", "default", map[string]string{})
+			Expect(err).To(HaveOccurred())
+		})
+
+		It("should return rendered chart", func() {
+			chart, err := renderer.Render(alpineChartPath, "missing", "default", map[string]string{})
```
timuthy:
```suggestion
			chart, err := renderer.Render(alpineChartPath, "alpine", "default", map[string]string{})
```

pkg/chartrenderer/default_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+			files := chart.Files()
+			Expect(files).To(HaveLen(1))
+			Expect(files).To(HaveKeyWithValue("alpine/templates/alpine-pod.yaml", alpinePod))
+		})
+	})
+
+	Describe("#FileContent", func() {
+		It("should return empty string when template file is missing", func() {
+			chart, err := renderer.Render(alpineChartPath, "alpine", "default", map[string]string{})
```
timuthy, line 86:
```suggestion
			chart, err := renderer.Render(alpineChartPath, "missing", "default", map[string]string{})
```

timuthy (commented):
Nice! Looks good. Only one proposal (nit) to be more consistent with the chart names.

pkg/chartrenderer/default_test.go
ialidzhikov: The missing in this test is is not the chartPath but the filename passed to FileContent func.

plkokanov:
Looks good!

pkg/chartrenderer/default_test.go
```diff
@@ -0,0 +1,102 @@
...
+				Major: "1",
+				Minor: "16",
+			},
+		}
+
+		capabilities, err := chartrenderer.DiscoverCapabilities(fakeDiscovery)
+		Expect(err).ToNot(HaveOccurred())
+
+		renderer = chartrenderer.New(engine.New(), capabilities)
+		Expect(err).ToNot(HaveOccurred())
```
timebertt: No error returned in the previous call.
```suggestion
```

timebertt (changes requested):
Thanks, only one nit :)

pkg/chartrenderer/default_test.go
ialidzhikov: Nice catch, thank you!

rfranzke (approved):

timebertt (approved):
/lgtm

PR #2150, opened 2020-04-07
Component descriptor script works for hyperkube

rfranzke:
Component descriptor script works for hyperkube


ialidzhikov (approved):
/lgtm

Issue #1576, opened 2019-11-01
Gardenlet


rfranzke:
Similar to how Kubernetes has the Kubelet to manage pods on nodes we want to have a Gardenlet to manage shoots on seeds.
The idea is to run the shoot reconciliation flow (+ all other controllers communicating with the seed) inside the seed itself, allowing for better network performance and scalability. The gardener-controller-manager will remain the control loops for things that are not requiring to communicate with the seed's API servers.

The following things need to be done for an initial implementation:

* [x] Gardenlet should have its own componentconfig
* Garden-controller-manager keeps following control loops:
  * [x] `CloudProfile` controller
  * [x] `ControllerRegistration` controller
  * [x] `Plant` controller
  * [x] `Project` controller
  * [x] `Quota` controller
  * [x] `SecretBinding` controller
  * [x] `Shoot` maintenance controller
  * [x] `Shoot` hibernation controller
  * [x] `Shoot` quota controller
* Gardenlet gets following control loops:
  * [x] `BackupBucket` controller
  * [x] `BackupEntry` controller
  * [x] `ControllerInstallation` controller
  * [x] `Seed` controller
  * [x] `Shoot` controller
  * [x] `Shoot` care controller
* [x] Gardenlet sends regular heartbeat and maintains the `GardenletReady` condition on `Seed`s
* [x] Garden-controller-manager gets a new seed lifecycle controller that turns the `GardenletReady` condition to `Unknown` if the gardenlet stops sending heartbeats
* [x] Helm chart for Gardenlet
* [x] Gardenlet requests a valid client certificate for the garden cluster with a bootstrap token and the `CertificateSigningRequest` resource
* [x] Gardenlet can deploy its own Helm chart for shooted seeds
* [x] Documentation & Migration Guide

mvladev:
Can you elaborate on 

> Gardenlet sends regular heartbeat and maintains the GardenletReady condition on Seeds

I'll strongly recommend to use the [Lease](https://github.com/kubernetes/api/blob/665c8a257c1af277521b08dd43d5c73570405ef0/coordination/v1/types.go#L23-L37) resource for storing heartbeats and update the status only when something really breaks. See the [Node Heartbeat KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/0009-node-heartbeat.md).

rfranzke:
[...]

PR #2151, opened 2020-04-07
Seed heartbeat controller uses Lease object to report Seeds' readiness

### LINKED ISSUE #1722: Use `Lease` objects for seed heartbeats
With #1601 we have introduced the gardenlet, a component comparable with the kubelet in Kubernetes.

Today, the gardenlet sends a heartbeat every `20s` to its responsible `Seed` objects by updating the `GardenletReady` condition.
This process is comparable to the heartbeats sent by the kubelet for its `Node` object.
However, the Kubernetes community has proposed an implemented another, more scalable mechanism, which is using `Lease` resources for the heartbeat (see [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/0009-node-heartbeat.md)).

@mvladev suggested in [this comment](https://github.com/gardener/gardener/issues/1576#issuecomment-550970997) to do the same.

Kubernetes creates a dedicated namespace in every cluster, the `kube-node-lease` namespaces, which holds all the `Lease`s:

```bash
$ k -n kube-node-lease get lease -o yaml
apiVersion: v1
items:
- apiVersion: coordination.k8s.io/v1
  kind: Lease
  metadata:
    creationTimestamp: "2019-10-25T08:03:11Z"
    name: docker-desktop
    namespace: kube-node-lease
    ownerReferences:
    - apiVersion: v1
      kind: Node
      name: docker-desktop
      uid: e35a28d4-f6fd-11e9-98c7-025000000001
    resourceVersion: "666283"
    selfLink: /apis/coordination.k8s.io/v1/namespaces/kube-node-lease/leases/docker-desktop
    uid: e35464b4-f6fd-11e9-98c7-025000000001
  spec:
    holderIdentity: docker-desktop
    leaseDurationSeconds: 40
    renewTime: "2019-11-11T06:30:03.034444Z"
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

We should probably do it the same way, create a dedicated `garden-seed-lease` namespace (we have to prevent `Project` resources with the `seed-lease` name), and write all the `Lease` objects for the `Seed`s here.

swilen-iwanow:
Changes the current seed lifecycle controller to use Lease objects
Fixes #1722

**Release note**:

```improvement operator
Seed lifecycle controller uses Lease object to report Seeds' readiness
```
```action operator
The minimum required Kubernetes version for the garden cluster is now `1.12`. Please make sure this your garden cluster is of at least this version before upgrading Gardener.
```


pkg/apis/core/validation/project.go
```diff
@@ -31,6 +33,9 @@ func ValidateProject(project *core.Project) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	allErrs = append(allErrs, apivalidation.ValidateObjectMeta(&project.ObjectMeta, false, ValidateName, field.NewPath("metadata"))...)
+	if project.Name == "seed-lease" {
```
rfranzke: Hmm, what if somebody has already such a project? Should we rather call the namespace `garden--seed-lease`? Then we don't need to validate this here. WDYT?

pkg/gardenlet/controller/lease/lease.go
```diff
@@ -0,0 +1,153 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
```diff
@@ -16,9 +16,11 @@ package seed
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	gardenconstants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```
rfranzke:
```suggestion
	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
```diff
@@ -98,21 +131,29 @@ func (c *defaultControl) Reconcile(seedObj *gardencorev1beta1.Seed, key string)
...
 		return true, nil
 	}
 
-	for _, condition := range seed.Status.Conditions {
-		// If the `GardenletReady` condition is not yet `Unknown` then check when it most recently sent a heartbeat and wait for the
-		// configured `monitorPeriod` before proceeding with any action.
-		if condition.Type == gardencorev1beta1.SeedGardenletReady && condition.Status != gardencorev1beta1.ConditionUnknown && !condition.LastUpdateTime.UTC().Before(time.Now().UTC().Add(-c.config.Controllers.Seed.MonitorPeriod.Duration)) {
+	seedLogger.Debugf("Setting status for seed %q to 'Unknown' as gardenlet stopped reporting seed status.", seed.Name)
+	observedSeedLease, err := c.k8sGardenClient.Kubernetes().CoordinationV1beta1().Leases(gardenconstants.GardenSeedLeaseNamespace).Get(seed.Name, metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
```
rfranzke:
```suggestion
	if client.IgnoreNotFound(err) != nil {
```

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
```diff
@@ -98,21 +131,29 @@ func (c *defaultControl) Reconcile(seedObj *gardencorev1beta1.Seed, key string)
...
+
+	// New seeds don't have conditions - gardenlet never reported anything yet. Wait for grace period.
 	if len(seed.Status.Conditions) == 0 {
 		return true, nil
 	}
 
-	for _, condition := range seed.Status.Conditions {
-		// If the `GardenletReady` condition is not yet `Unknown` then check when it most recently sent a heartbeat and wait for the
-		// configured `monitorPeriod` before proceeding with any action.
-		if condition.Type == gardencorev1beta1.SeedGardenletReady && condition.Status != gardencorev1beta1.ConditionUnknown && !condition.LastUpdateTime.UTC().Before(time.Now().UTC().Add(-c.config.Controllers.Seed.MonitorPeriod.Duration)) {
```
rfranzke, line 109: Maybe I missed it - where is the monitor period now evaluated?

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
```diff
@@ -98,21 +131,29 @@ func (c *defaultControl) Reconcile(seedObj *gardencorev1beta1.Seed, key string)
...
+	seedLogger.Debugf("Setting status for seed %q to 'Unknown' as gardenlet stopped reporting seed status.", seed.Name)
```
rfranzke: Why was this log moved up here? At this point in time it's not yet decided, is it?

pkg/gardenlet/controller/lease/lease.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"fmt"
+
+	coordinationv1 "k8s.io/api/coordination/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	coordclientset "k8s.io/client-go/kubernetes/typed/coordination/v1"
+
```
rfranzke:
```suggestion
```

pkg/gardenlet/controller/lease/lease_test.go
```diff
@@ -0,0 +1,182 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/gardenlet/controller/seed/seed_lease_control.go
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/gardenlet/controller/seed/seed_lease_control.go
```diff
@@ -0,0 +1,135 @@
...
+	condition := helper.GetCondition(seed.Status.Conditions, gardencorev1beta1.SeedGardenletReady)
+	if condition == nil {
+		err := c.initGardenletReadyCondition(seed)
+		if err != nil {
+			return err
+		}
+	}
+
+	clock := clockpkg.RealClock{}
+	seedOwnerReference := buildSeedOwnerReference(seed)
```
rfranzke: Is the kubelet doing the same thing? Actually, `Seed` is a cluster-wide object while `Lease` is namespaced. Kubernetes does not support owner references for these mixed object types.

pkg/gardenlet/controller/seed/seed_lease_control.go
```diff
@@ -0,0 +1,135 @@
...
+	leaseControl := lease.NewLeaseController(clock, c.k8sGardenClient.Kubernetes(), leaseResyncSeconds, seed.GetName(), gardenercorev1betaconstants.GardenSeedLeaseNamespace, *seedOwnerReference)
+	if err := leaseControl.Sync(); err != nil {
+		return err
+	}
+
+	if condition.Status == "Unknown" {
```
rfranzke: Can you use the constant instead of `"Unknown"`?

pkg/gardenlet/controller/seed/seed_lease_control.go
```diff
@@ -0,0 +1,135 @@
...
+	return err
+}
+
+func (c *Controller) checkSeedConnection(seed *gardencorev1beta1.Seed) error {
+	client, err := seedpkg.GetSeedClient(context.TODO(), c.k8sGardenClient.Client(), c.config.SeedClientConnection.ClientConnectionConfiguration, c.config.SeedSelector == nil, seed.Name)
+	if err != nil {
+		return err
+	}
+
+	_, err = client.Kubernetes().CoreV1().Pods("").List(v1.ListOptions{})
```
rfranzke: Why do we list the pods? Can we just call `/healthz` endpoint of the API server?

pkg/gardenlet/controller/seed/seed_lease_control.go
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
```
rfranzke, line 2: Can we have a unit test for this file?

rfranzke (changes requested):
Thanks for the PR, highly appreciate it! :)

ialidzhikov:
Can you also rebase to latest master to make ci/cd happy?

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
swilen-iwanow: The object is being reconciled roughly on 10 seconds. If the lease was not updated during those 10 seconds we consider that the gardenlet condition is unknown.

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
swilen-iwanow: Yes, moved it by accident during rebase.  Will fix

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
rfranzke: I've seen that - the question above was about the "why", and what is now happening with the monitor period?

pkg/gardenlet/controller/seed/seed_lease_control.go
swilen-iwanow: Yes, as far as I remember they were doing this for the nodes. https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/nodelease/controller.go#L219

rfranzke:
@mvladev as you suggested it back then, can you please take a look? :)

mvladev:
I'll look into it tomorrow. I'm busy today.

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
swilen-iwanow: I am not using the MonitorPeriod, just checking if the stored lease is older than the observed. But i can change it if you would like to compare it to time.Now() as before.

rfranzke: When we introduced the monitor period we discussed that it might be helpful to have this configurable (mainly for setups where the timeout may be prolonged, e.g., cross-continent seeds), so I'd prefer to keep this part as it was before.

swilen-iwanow: I see, yeah makes sense. I will change it then

pkg/apis/core/validation/project.go
swilen-iwanow: It is gonna work. Also it wouldn't need a change in the dashboard to validate this in the frontend. But I will leave the check, not to be able to create a project with such namespace.

rfranzke: This check is also executed when existing projects are updated. Hence, if such a project already exists in a gardener landscape then it can never be updated again...

swilen-iwanow: So how can we prevent users to specify garden--seed-lease in their `Projects`

rfranzke: As suggested - use `garden--seed-lease` as namespace - all project namespaces look like `garden-<name>`, and `<name>` cannot start with `-` (I hope, please double check). WDYT?

swilen-iwanow: Yes yes I agree with that I meant the validation of the Project's namespace. Shouldn't we make it impossible to specify garden--seed-lease as a namespace when you are creating Projects via kubectl for example

rfranzke: Ah, ok, I see. Yes, you can validate this in this file - just check that `.spec.namespace` is not `garden--seed-lease` (yes, this is also backwards-incompatible, but the likelihood that somebody has such a namespace that he bound to a project is ~0).

pkg/gardenlet/controller/seed/seed_lease_control.go
swilen-iwanow: Just an update: I am trying, but it is starting to look really bad 😁

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
```diff
@@ -98,21 +101,23 @@ func (c *defaultControl) Reconcile(seedObj *gardencorev1beta1.Seed, key string)
...
+	// New seeds don't have conditions - gardenlet never reported anything yet. Wait for grace period.
 	if len(seed.Status.Conditions) == 0 {
 		return true, nil
 	}
 
-	for _, condition := range seed.Status.Conditions {
-		// If the `GardenletReady` condition is not yet `Unknown` then check when it most recently sent a heartbeat and wait for the
-		// configured `monitorPeriod` before proceeding with any action.
-		if condition.Type == gardencorev1beta1.SeedGardenletReady && condition.Status != gardencorev1beta1.ConditionUnknown && !condition.LastUpdateTime.UTC().Before(time.Now().UTC().Add(-c.config.Controllers.Seed.MonitorPeriod.Duration)) {
+	observedSeedLease, err := c.k8sGardenClient.Kubernetes().CoordinationV1beta1().Leases(v1beta1constants.GardenSeedLeaseNamespace).Get(seed.Name, metav1.GetOptions{})
```
rfranzke: Can you use the controller-runtime client here?

rfranzke (commented):
Thanks @swilen-iwanow, only one minor comment. @mvladev can you check as well, please?

pkg/apis/core/validation/project_test.go
```diff
@@ -101,6 +101,17 @@ var _ = Describe("Project Validation Tests", func() {
 			}))))
 		})
 
+		It("should forbid Projects with namespace garden-seed-lease", func() {
```
ialidzhikov:
```suggestion
		It("should forbid Projects with namespace garden--seed-lease", func() {
```

rfranzke (commented):
Looks great now, thanks, still need to test it.
@mvladev any feedback?

swilen-iwanow:
> Looks great now, thanks, still need to test it.
> @mvladev any feedback?

@rfranzke We spoke earlier this week. And he gave me some feedback that I have to implement.

rfranzke:
What's the feedback/what do you have to implement?

pkg/apis/core/validation/project.go
```diff
@@ -65,6 +67,9 @@ func ValidateProjectUpdate(newProject, oldProject *core.Project) field.ErrorList
 func ValidateProjectSpec(projectSpec *core.ProjectSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
+	if projectSpec.Namespace != nil && *projectSpec.Namespace == constants.GardenSeedLeaseNamespace {
+		allErrs = append(allErrs, field.Forbidden(fldPath.Child("namespace"), "Project namespace "+constants.GardenSeedLeaseNamespace+" is reserved by Gardener"))
```
ialidzhikov:
```suggestion
		allErrs = append(allErrs, field.Forbidden(fldPath.Child("namespace"), fmt.Sprintf("Project namespace %s is reserved by Gardener", constants.GardenSeedLeaseNamespace)))
```

pkg/gardenlet/controller/seed/seed_lease_control_test.go
```diff
@@ -0,0 +1,216 @@
...
+	mockversioned "github.com/gardener/gardener/pkg/mock/gardener/client/core/clientset/versioned"
+	mockgardencorev1beta1 "github.com/gardener/gardener/pkg/mock/gardener/client/core/clientset/versioned/typed/core/v1beta1"
+	mock "github.com/gardener/gardener/pkg/mock/gardener/client/kubernetes"
+	mockio "github.com/gardener/gardener/pkg/mock/go/io"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+
+	// . "github.com/onsi/ginkgo/extensions/table"
```
ialidzhikov:
```suggestion
```

pkg/gardenlet/controller/seed/seed_lease_control_test.go
```diff
@@ -0,0 +1,216 @@
...
+		seedRESTClient.EXPECT().Get().Return(request.AbsPath("/healthz"))
+		body.EXPECT().Read(gomock.Any()).Return(0, io.EOF).AnyTimes()
+		body.EXPECT().Close().AnyTimes()
+
+		//update seed condition
+		k8sGardenClient.EXPECT().GardenCore().Return(gardenVersioned)
+		gardenVersioned.EXPECT().CoreV1beta1().Return(gardenIface).AnyTimes()
+		gardenIface.EXPECT().Seeds().Return(seedIface).AnyTimes()
+		seedIface.EXPECT().Get(seedName, metav1.GetOptions{}).Return(seed, nil)
+		// seedIface.EXPECT().UpdateStatus(gomock.Any()).AnyTimes()
```
ialidzhikov:
```suggestion
```

pkg/gardenlet/controller/seed/seed_lease_control_test.go
```diff
@@ -0,0 +1,216 @@
...
+		controller = Controller{k8sGardenClient: k8sGardenClient, seedLister: lister, seedLeaseQueue: seedLeaseQueue, getSeedClient: fakeClientSeedFunc, config: gardenletConfig, seedLeaseControl: seedLeaseControl}
+		Expect(indexer.Add(seed)).NotTo(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		seedLeaseQueue.ShutDown()
+		syncFail = nil
+	})
+
+	Context("seed retreival", func() {
```
ialidzhikov:
```suggestion
	Context("seed retrieval", func() {
```

pkg/gardenlet/controller/seed/seed.go
```diff
@@ -40,19 +43,22 @@ import (
...
+	componentbaseconfig "k8s.io/component-base/config"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 // Controller controls Seeds.
 type Controller struct {
 	k8sGardenClient        kubernetes.Interface
 	k8sGardenCoreInformers gardencoreinformers.SharedInformerFactory
+	getSeedClient          getSeedClient
```
ialidzhikov, line 54: What is the motivation for this? Can't we have just `k8sSeedClient kubernetes.Interface` and in the test pass a fake/mock client? Why it needs to be a func that provides a seedClient?

ialidzhikov (commented):
Nits.

pkg/apis/core/v1alpha1/constants/types_constants.go
```diff
@@ -211,6 +211,10 @@ const (
 	// It is also used by the gardener-apiserver.
 	GardenNamespace = "garden"
 
+	// GardenSeedLeaseNamespace is the namespace in which Gardenlet will report Seeds'
+	// status using Lease resources for each Seed
+	GardenSeedLeaseNamespace = "garden--seed-lease"
```
mvladev: I would rather have something `gardener-system-seed-lease`.

pkg/apis/core/validation/project.go
mvladev: What happens if a project already exists which uses this NS - It would break the said Project.

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
```diff
@@ -98,34 +104,53 @@ func (c *defaultControl) Reconcile(seedObj *gardencorev1beta1.Seed, key string)
...
+	if client.IgnoreNotFound(err) != nil {
+		return false, err
+	}
+
+	if observedSeedLease != nil && observedSeedLease.Spec.RenewTime != nil {
+		if observedSeedLease.Spec.RenewTime.After(time.Now().UTC().Add(-c.config.Controllers.Seed.MonitorPeriod.Duration)) {
+			return true, nil
+		}
+
+		err = c.k8sGardenClient.Client().Get(ctx, kutil.Key(v1beta1constants.GardenSeedLeaseNamespace, seed.Name), observedSeedLease)
```
mvladev: `.Get(...)` mutates `observedSeedLease`. You should not mutate items in the cache - do a `DeepCopy` and use the coped resource for any mutations.

pkg/gardenlet/controller/seed/seed_lease_control.go
```diff
@@ -0,0 +1,139 @@
...
+		return err
+	}
+
+	if err := c.checkSeedConnection(seed); err != nil {
+		return fmt.Errorf("[SEED LEASE] cannot establish connection with Seed %s: %v", key, err)
+	}
+
+	condition := helper.GetCondition(seed.Status.Conditions, gardencorev1beta1.SeedGardenletReady)
+	if condition == nil {
+		condition, err = c.initGardenletReadyCondition(seed)
```
mvladev: Don't mutate cache objects. Deep copy them.

Why is it doing any mutation in the `reconcileSeedLeaseKey`? Any such operations should only be called in methods behind a workqueue.

pkg/gardenlet/controller/seed/seed_lease_control.go
```diff
@@ -0,0 +1,139 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	seedOwnerReference := buildSeedOwnerReference(seed)
+	if err := c.seedLeaseControl.Sync(seed.GetName(), *seedOwnerReference); err != nil {
```
mvladev: Each controller should be isolated from each other. Better add it to the workqueue of that controller.

pkg/gardenlet/controller/seed/seed_lease_control.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (c *Controller) updateSeedStatus(seed *gardencorev1beta1.Seed, conditionGardenletReady gardencorev1beta1.Condition) error {
+	_, err := kutil.TryUpdateSeedConditions(c.k8sGardenClient.GardenCore(), retry.DefaultBackoff, seed.ObjectMeta,
+		func(seed *gardencorev1beta1.Seed) (*gardencorev1beta1.Seed, error) {
+			seed.Status.Conditions = helper.MergeConditions(seed.Status.Conditions, conditionGardenletReady)
+			return seed, nil
+		},
+	)
+	return err
+}
```
mvladev: The methods here are making update requests to the API server:

1. For objects which don't need update
1. With retry. Retry logic should be handled by workqueue.

mvladev (changes requested):

pkg/apis/core/v1alpha1/constants/types_constants.go
swilen-iwanow: @rfranzke wdyt?

rfranzke: Fine for me

pkg/apis/core/validation/project.go
rfranzke: Was similarly discussed in https://github.com/gardener/gardener/pull/2151#discussion_r405290085

pkg/gardenlet/controller/seed/seed.go
swilen-iwanow: This is used to mock the result of this function https://github.com/gardener/gardener/blob/master/pkg/operation/seed/seed.go#L116, which creates new kubernetes client. It seemed easier to provide it as a dependency.

pkg/apis/core/v1alpha1/constants/types_constants.go
```diff
@@ -211,6 +211,10 @@ const (
 	// It is also used by the gardener-apiserver.
 	GardenNamespace = "garden"
 
+	// GardenSeedLeaseNamespace is the namespace in which Gardenlet will report Seeds'
+	// status using Lease resources for each Seed
+	GardenSeedLeaseNamespace = "gardener-system-seed-lease"
```
rfranzke:
```suggestion
	GardenerSeedLeaseNamespace = "gardener-system-seed-lease"
```

pkg/apis/core/v1beta1/constants/types_constants.go
```diff
@@ -219,6 +219,10 @@ const (
 	// It is also used by the gardener-apiserver.
 	GardenNamespace = "garden"
 
+	// GardenSeedLeaseNamespace is the namespace in which Gardenlet will report Seeds'
+	// status using Lease resources for each Seed
+	GardenSeedLeaseNamespace = "gardener-system-seed-lease"
```
rfranzke:
```suggestion
	GardenerSeedLeaseNamespace = "gardener-system-seed-lease"
```

pkg/apis/core/validation/project.go
```diff
@@ -15,10 +15,13 @@
 package validation
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/gardener/gardener/pkg/apis/core"
 
+	"github.com/gardener/gardener/pkg/apis/core/v1alpha1/constants"
```
rfranzke:
```suggestion
	"github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```

Though, probably it's better to add the `GardenerSeedLeaseNamespace` constant to `pkg/apis/core` as well (and move it from `pkg/apis/core/{v1alpha1,v1beta1/constants` to `pkg/apis/core/{v1alpha1,v1beta1}`).

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
```diff
@@ -98,34 +105,54 @@ func (c *defaultControl) Reconcile(seedObj *gardencorev1beta1.Seed, key string)
...
-		// If the `GardenletReady` condition is not yet `Unknown` then check when it most recently sent a heartbeat and wait for the
-		// configured `monitorPeriod` before proceeding with any action.
-		if condition.Type == gardencorev1beta1.SeedGardenletReady && condition.Status != gardencorev1beta1.ConditionUnknown && !condition.LastUpdateTime.UTC().Before(time.Now().UTC().Add(-c.config.Controllers.Seed.MonitorPeriod.Duration)) {
+	observedSeedLease, err := c.leaseLister.Leases(v1beta1constants.GardenSeedLeaseNamespace).Get(seed.Name)
+	if client.IgnoreNotFound(err) != nil {
+		return false, err
+	}
+
+	if observedSeedLease != nil && observedSeedLease.Spec.RenewTime != nil {
+		if observedSeedLease.Spec.RenewTime.After(time.Now().UTC().Add(-c.config.Controllers.Seed.MonitorPeriod.Duration)) {
```
rfranzke:
```suggestion
		if observedSeedLease.Spec.RenewTime.UTC().After(time.Now().UTC().Add(-c.config.Controllers.Seed.MonitorPeriod.Duration)) {
```

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
```diff
@@ -98,34 +105,54 @@ func (c *defaultControl) Reconcile(seedObj *gardencorev1beta1.Seed, key string)
...
+			return true, nil
+		}
+
+		latestLeaseObject := &coordinationv1.Lease{}
+		err = c.k8sGardenClient.Client().Get(ctx, kutil.Key(v1beta1constants.GardenSeedLeaseNamespace, seed.Name), latestLeaseObject)
+		if err != nil {
```
rfranzke:
```suggestion
		if err := c.k8sGardenClient.Client().Get(ctx, kutil.Key(v1beta1constants.GardenSeedLeaseNamespace, seed.Name), latestLeaseObject); err != nil {
```

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
```diff
@@ -98,34 +105,54 @@ func (c *defaultControl) Reconcile(seedObj *gardencorev1beta1.Seed, key string)
...
+			return false, err
+		}
+
+		if latestLeaseObject.Spec.RenewTime.After(time.Now().UTC().Add(-c.config.Controllers.Seed.MonitorPeriod.Duration)) {
```
rfranzke:
```suggestion
		if latestLeaseObject.Spec.RenewTime.UTC().After(time.Now().UTC().Add(-c.config.Controllers.Seed.MonitorPeriod.Duration)) {
```

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
```diff
@@ -98,34 +105,54 @@ func (c *defaultControl) Reconcile(seedObj *gardencorev1beta1.Seed, key string)
...
+		return false, err
+	}
+
+	if observedSeedLease != nil && observedSeedLease.Spec.RenewTime != nil {
+		if observedSeedLease.Spec.RenewTime.After(time.Now().UTC().Add(-c.config.Controllers.Seed.MonitorPeriod.Duration)) {
+			return true, nil
+		}
+
+		latestLeaseObject := &coordinationv1.Lease{}
+		err = c.k8sGardenClient.Client().Get(ctx, kutil.Key(v1beta1constants.GardenSeedLeaseNamespace, seed.Name), latestLeaseObject)
```
rfranzke: So this is to not accidentally make the wrong decision due to outdated cache, hence, we fetch it again from the system to double-check?

pkg/gardenlet/controller/seed/seed_lease_control.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	result := client.RESTClient().Get().AbsPath("/healthz").Do()
+	if result.Error() != nil {
+		return fmt.Errorf("Failed to execute call to Kubernetes API Server: %v", result.Error())
+	}
+
+	var statusCode int
+	result.StatusCode(&statusCode)
+	if statusCode != http.StatusOK {
+		return fmt.Errorf("API Server returned unexepected status code: %v", statusCode)
```
rfranzke:
```suggestion
		return fmt.Errorf("API Server returned unexpected status code: %d", statusCode)
```

rfranzke (changes requested):
Still looks good, only nits

pkg/gardenlet/controller/seed/seed_lease_control.go
```diff
@@ -0,0 +1,121 @@
...
+	bldr.WithStatus(gardencorev1beta1.ConditionTrue)
+	bldr.WithReason("GardenletReady")
+	bldr.WithMessage("Gardenlet is posting ready status.")
+
+	if newCondition, update := bldr.WithNowFunc(metav1.Now).Build(); update {
+		seed.Status.Conditions = helper.MergeConditions(seedCopy.Status.Conditions, newCondition)
+		if err := c.k8sGardenClient.Client().Status().Update(context.TODO(), seed); err != nil {
+			return err
+		}
+	}
```
rfranzke, line 88: @mvladev in your suggestion https://github.com/gardener/gardener/issues/1576#issuecomment-550968489 you mentioned that you want the seed status to only be updated in case something breaks, probably to prevent too many unnecessary UPDATE calls to the seed object. However, as far as I can see we are still updating the `GardenletReady` condition every `2s` here, so I'm wondering about the benefit of this PR/change in general. Isn't it now more or less similar like before (actually, it's worse because we are now making twice as many calls to the `Lease` objects)?

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
swilen-iwanow: Yes

pkg/gardenlet/controller/seed/seed_lease_control.go
swilen-iwanow: Kubernetes, also benefits from this approach as the `Node` resource might become very big and updating it becomes expensive.

rfranzke: I understand, but we are still updating the `Seed` object every time to write the condition into it, or?

swilen-iwanow: No, if only it is not set already or it is with different status than `True`.

rfranzke: OK, then this is it that I was missing. Thanks for the explanation.

docs/usage/supported_k8s_versions.md
```diff
@@ -14,6 +14,6 @@ The reason for that is that we require CRD status subresources for the extension
 
 ## Shoot cluster versions
 
-Gardener itself is capable of spinning up clusters with Kubernetes versions **`1.10`** up to **`1.18`**.
+Gardener itself is capable of spinning up clusters with Kubernetes versions **`1.12`** up to **`1.18`**.
```
rfranzke:
```suggestion
Gardener itself is capable of spinning up clusters with Kubernetes versions **`1.10`** up to **`1.18`**.
```

docs/usage/supported_k8s_versions.md
```diff
@@ -4,8 +4,8 @@ Currently, the Gardener supports the following Kubernetes versions:
 
 ## Garden cluster version
 
-:warning: The minimum version of the garden cluster that can be used to run Gardener is **`1.10.x`**.
-The reason for that is that the least supported Kubernetes version in Gardener is `1.10`.
+:warning: The minimum version of the garden cluster that can be used to run Gardener is **`1.12.x`**.
+The reason for that is that the least supported Kubernetes version in Gardener is `1.12`.
```
rfranzke:
```suggestion
```

Explain the reason here (`Lease` object)

cmd/gardener-apiserver/app/gardener_apiserver.go
```diff
@@ -263,6 +267,25 @@ func (o Options) run(stopCh <-chan struct{}) error {
 		return err
 	}
 
+	if err := server.GenericAPIServer.AddPostStartHook("bootstrap-gardener-cluster", func(context genericapiserver.PostStartHookContext) error {
```
rfranzke:
```suggestion
	if err := server.GenericAPIServer.AddPostStartHook("bootstrap-garden-cluster", func(context genericapiserver.PostStartHookContext) error {
```

cmd/gardener-apiserver/app/gardener_apiserver.go
```diff
@@ -263,6 +267,25 @@ func (o Options) run(stopCh <-chan struct{}) error {
...
+
+		_, err = kubeClient.CoreV1().Namespaces().Create(&corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: gardencorev1beta1.GardenerSeedLeaseNamespace,
+			},
+		})
+		if err != nil && !apierrors.IsAlreadyExists(err) {
+			return err
+		}
+		return nil
```
rfranzke: first handle error case properly

```suggestion
		if _, err := kubeClient.CoreV1().Namespaces().Get(gardencorev1beta1.GardenerSeedLeaseNamespace, metav1.GetOptions{}); client.IgnoreNotFound(err) != nil {
			return err
		}
		if apierrors.IsNotFound(err) {
			// the namespace already exists
			return nil
		}

		_, err = kubeClient.CoreV1().Namespaces().Create(&corev1.Namespace{
			ObjectMeta: metav1.ObjectMeta{
				Name: gardencorev1beta1.GardenerSeedLeaseNamespace,
			},
		})
		return err
```

rfranzke (changes requested):
Nits.
Still need to play with it locally.

rfranzke:
Can you also add an "action" release note for operators wrt the garden cluster minimum version bump?

pkg/controllermanager/controller/seed/seed_lifecycle_control.go
```diff
@@ -98,34 +104,53 @@ func (c *defaultControl) Reconcile(seedObj *gardencorev1beta1.Seed, key string)
...
+	if client.IgnoreNotFound(err) != nil {
+		return false, err
+	}
+
+	if observedSeedLease != nil && observedSeedLease.Spec.RenewTime != nil {
+		if observedSeedLease.Spec.RenewTime.UTC().After(time.Now().UTC().Add(-c.config.Controllers.Seed.MonitorPeriod.Duration)) {
+			return true, nil
+		}
+
+		latestLeaseObject := &coordinationv1.Lease{}
```
mvladev, line 123: Can you a comment that in this case, the cache might be old and therefore a call to the API server must be done to ensure that the lease has really expired?

So people don't have wonder why you are doing this call here.

docs/usage/supported_k8s_versions.md
swilen-iwanow: The Lease object is only used in the Garden cluster, Shoot clusters still can use older k8s versions, right?

rfranzke: Yes

rfranzke (dismissed):
/lgtm

@mvladev any more comments?

mvladev (changes requested):

rfranzke (dismissed):

rfranzke:
/kind/enhancement

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/huandu/xstrings v1.3.0
 	github.com/json-iterator/go v1.1.9
+	github.com/mesos/mesos-go v0.0.10 // indirect
```
ialidzhikov: Why this is added?

swilen-iwanow: I accidentally used apierrors from this repo, because of the Auto Import of VSCode, which I guess added it to the go.mod, but didn't remove it as I fixed the import. Sorry

rfranzke (approved):

rfranzke:
@mvladev final opinion? :)

/reviewed/do-not-merge
until g/g@v1.4.0 is released, cc @ialidzhikov

mvladev (approved):

PR #2152, opened 2020-04-07
check-generate.sh calls make rules

rfranzke:
check-generate.sh calls make rules


ialidzhikov (approved):
/lgtm

PR #2155, opened 2020-04-07
Make machine image optional

ialidzhikov:
Currently trying to apply a Shoot without a machine image:
```
$ k apply -f shoot.yaml
error: error validating "shoot.yaml": error validating data: ValidationError(Shoot.spec.provider.workers[0].machine.image): missing required field "version" in com.github.gardener.gardener.pkg.apis.core.v1beta1.ShootMachineImage; if you choose to ignore these errors, turn validation off with --validate=false
```

It looks like kubectl is using the openapi schema to validate the resources.

Ref #1960

**Release note**:

```improvement operator
NONE
```


vpnachev (dismissed):
/lgtm

nice catch

rfranzke:
Should the type be changed to pointer? Should we add omitempty tags?

ialidzhikov:
> Should the type be changed to pointer? Should we add omitempty tags?

It is now changed to `*string`, `omitempty` also added.

ialidzhikov:
@rfranzke , @mvladev , I changed the Version field in the external type to `*string` and the respective field in the internal type is `string`. 
If you have any objections/comments from api-conventions side, please share.
My goal is to have the Version field optional (and to be omitted, if needed, on POST) and to be defaulted in an admission plugin.

vpnachev (approved):
/lgtm

rfranzke (approved):

rfranzke:
FYI @danielfoehrKn

danielfoehrKn (approved):

PR #2156, opened 2020-04-07
removing controllerregistration admission plugin

danielfoehrKn:

The controllerregistration admission plugin enforces that a particular kind-type combination (e.g Infrastrucutre - aws) cannot be registered by multiple controllerregistrations. 
However there are use cases when that is required. 

**Example when that is required:** 
- Gardener uses private hosted zone as the internal domain for Shoot clusters. This is the case for aws-gov-cloud. The regular provider extensions do not support private zones. 
- In order to support private zones, we have another extension (private-zone-extension), also acting on (Infrastructure - aws) , assigning the created VPC to the private hosted zone

To support the above mentioned use-case, there are currently controller-registrations that have no kind-type association. However with the [recent changes introduced in the Gardener Controller Manager](https://github.com/gardener/gardener/commit/8359e1ae278c1f121d1d3db69a96c1fa6e0ce4a5), controllerinstallations of such a registration are being indentified as not needed and consequently removed.

**To sum it up**
- Gardener should not enforce that a particular kind-type combination (e.g Infrastrucutre - aws) cannot be registered by multiple controllerregistrations. 
- Gardener  should validate that every controllerregistration has at least one  kind-type association.

Fixes #

**Release note**:

```improvement operator

```


ialidzhikov (changes requested):
Admission plugin can be disabled with the `--disable-admission-plugins` flag of the API server. There is no need to remove anything from the codebase.

@danielfoehrKn, any objections to go with the flag and close this PR?

danielfoehrKn:
Sure lets close this PR. Thanks for the suggestion.

PR #2159, opened 2020-04-07
Make admission plugins configurable in gardener-apiserver chart

ialidzhikov:
Make admission plugins configurable in gardener-apiserver chart.

Ref https://github.com/gardener/gardener/pull/2156

**Release note**:

```improvement operator
gardener-apiserver chart does now allow configuring the `--disable-admission-plugins` and  `--enable-admission-plugins` flags.
```

### DIFF
--- a/charts/gardener/controlplane/values.yaml
+++ b/charts/gardener/controlplane/values.yaml
@@ -93,6 +93,10 @@ global:
         memory:
           value: "5G"
           percentage: 80
+    # List of admission plugins that should be disabled although they are in the default enabled plugins list.
+  # disableAdmissionPlugins: []
+    # List of admission plugins to be enabled in addition to default enabled ones.
+  # enableAdmissionPlugins: []
 
     audit:
  #    dynamicConfiguration: false                             Enables dynamic audit configuration. This feature also requires the DynamicAuditing feature flag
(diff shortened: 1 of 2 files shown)


charts/gardener/controlplane/values.yaml
```diff
@@ -93,6 +93,10 @@ global:
         memory:
           value: "5G"
           percentage: 80
+    # Comma-delimited list of admission plugins that should be disabled although they are in the default enabled plugins list.
+    disableAdmissionPlugins: ""
```
rfranzke: Would it be more convenient to make this a list of strings and then join them with `,` in the template?

timebertt: +1
Maybe we can keep it similar to the way we do it in the Shoot's kube-apiserver chart.
https://github.com/gardener/gardener/blob/ca6d53cd358f2c415101aa63b1212841cddb140f/charts/seed-controlplane/charts/kube-apiserver/templates/kube-apiserver.yaml#L79

charts/gardener/controlplane/values.yaml
```diff
@@ -93,6 +93,10 @@ global:
         memory:
           value: "5G"
           percentage: 80
+    # List of admission plugins that should be disabled although they are in the default enabled plugins list.
+    disableAdmissionPlugins: []
```
rfranzke: Just to make sure these values are optional, can you comment them? 
```suggestion
  # disableAdmissionPlugins: []
```

rfranzke (approved):

timebertt (approved):

danielfoehrKn (approved):
Thanks for this PR!

PR #2160, opened 2020-04-08
Added integration test for scale-to-zero feature of cluster autoscaler

hardikdr:
**What this PR does / why we need it**: Adds Integration test for the scale-to-zero feature of autoscaler.

Fixes #

**Special notes for your reviewer**: Skipping the test for other providers except aws and azure. 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/test/framework/resources/templates/reserve-capacity.yaml.tpl
+++ b/test/framework/resources/templates/reserve-capacity.yaml.tpl
@@ -21,6 +21,12 @@ spec:
       terminationGracePeriodSeconds: 5
       nodeSelector:
         worker.gardener.cloud/pool: {{ .WorkerPool }}
+{{- if .TolerationKey }}
+      tolerations:
+      - key: {{ .TolerationKey }}
+        operator: "Exists"
+        effect: "NoSchedule"
+{{- end }}
       containers:
       - name: pause-container
         image: gcr.io/google_containers/pause-amd64:3.1
--- a/test/integration/shoots/operations/clusterautoscaler.go
+++ b/test/integration/shoots/operations/clusterautoscaler.go
@@ -38,6 +38,7 @@ import (
 	"github.com/gardener/gardener/test/framework"
 
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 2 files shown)


dguendisch (dismissed):

test/integration/shoots/operations/clusterautoscaler.go
```diff
@@ -165,6 +166,125 @@ var _ = ginkgo.Describe("Shoot clusterautoscaler testing", func() {
...
+			}
+			s.Spec.Kubernetes.ClusterAutoscaler.ScaleDownDelayAfterAdd = &metav1.Duration{Duration: scaleDownDelayAfterAdd}
+			s.Spec.Kubernetes.ClusterAutoscaler.ScaleDownUnneededTime = &metav1.Duration{Duration: scaleDownUnneededTime}
+
+			return nil
+		})
+		framework.ExpectNoError(err)
+
+		// Reset shoot spec to original values after test
+		defer func() {
```
danielfoehrKn, line 210: Sorry unrelated to this particular function, but a general issue.

@dguendisch @schrodit  can we make it possible in the framework to use the ginkgo.AfterSuite(). That is not possible any more because the framework already uses it to execute common functionality. 
I find it difficult to work around that and mostly end up cramping all the test code within on ginko.it() and separating it with ginko.By() (like in this case) - only to be able to execute cleanup functionality. 
I would either wish to have 
- an option to disable the framework to execute the ginko.AfterSuite()
- or the ability to register functions in a generic way that are also executed in the AfterSuite()

WDYT & is that feasible?

test/integration/shoots/operations/clusterautoscaler.go
```diff
@@ -165,6 +166,125 @@ var _ = ginkgo.Describe("Shoot clusterautoscaler testing", func() {
...
+		if shoot.Spec.Provider.Type != "aws" && shoot.Spec.Provider.Type != "azure" {
+			ginkgo.Skip("not applicable")
+		}
+
+		// Create a dedicated worker-pool for cluster autoscaler.
+		testWorkerPool := shoot.Spec.Provider.Workers[0]
+		testWorkerPool.Name = testWorkerPoolName
+		testWorkerPool.Minimum = 0
+		testWorkerPool.Maximum = 2
+		testWorkerPool.Labels = map[string]string{"worker.gardener.cloud/pool": testWorkerPoolName}
```
danielfoehrKn: Isn't that label added by the kubelet already during node registration? Or why is there a need to add that label explicitly?

hardikdr:
cc @tim-ebert

test/integration/shoots/operations/clusterautoscaler.go
```diff
@@ -165,6 +166,125 @@ var _ = ginkgo.Describe("Shoot clusterautoscaler testing", func() {
...
+		framework.ExpectNoError(err)
+
+		// Reset shoot spec to original values after test
+		defer func() {
+			ginkgo.By("reverting shoot spec changes by test")
+			err := f.UpdateShoot(ctx, func(s *corev1beta1.Shoot) error {
+				s.Spec.Kubernetes.ClusterAutoscaler = origClusterAutoscalerConfig
+
+				// Remove the dedicated ca-test workerpool
+				s.Spec.Provider.Workers = origWorkers
```
danielfoehrKn: The test is Serial() thats why this is feasible I guess. 
What if you change the logic to only remove the added worker pool instead of completely overwriting it?  This way the test may be able  to work in non-serial mode? (pool also has taints so that no other pod is scheduled there).
Or could the changed autoscaler config create problems in parallely executed tests?

danielfoehrKn (dismissed):

test/framework/resources/templates/reserve-capacity.yaml.tpl
```diff
@@ -21,6 +21,10 @@ spec:
       terminationGracePeriodSeconds: 5
       nodeSelector:
         worker.gardener.cloud/pool: {{ .WorkerPool }}
+      tolerations:
+      - key: {{ .TolerationKey }}
+        operator: "Exists"
+        effect: "NoSchedule"
```
timebertt, line 28: This will add toleration for all keys to the deployment used in the other clusterautoscaler test (because the `.TolerationKey` is not set there).
Can you therefore wrap it in an `if` like this?
```suggestion
{{- if .TolerationKey }}
      tolerations:
      - key: {{ .TolerationKey }}
        operator: "Exists"
        effect: "NoSchedule"
{{- end }}
```

test/integration/shoots/operations/clusterautoscaler.go
```diff
@@ -165,6 +166,125 @@ var _ = ginkgo.Describe("Shoot clusterautoscaler testing", func() {
...
+		testWorkerPool.Minimum = 0
+		testWorkerPool.Maximum = 2
+		testWorkerPool.Labels = map[string]string{"worker.gardener.cloud/pool": testWorkerPoolName}
+		testTaint := corev1.Taint{
+			Key:    testWorkerPoolName,
+			Effect: corev1.TaintEffectNoSchedule,
+			Value:  testWorkerPoolName,
+		}
+
+		testWorkerPool.Taints = append(testWorkerPool.Taints, testTaint)
```
timebertt: Can you maybe set the test taint directly on the worker pool instead of appending, so you will have only one taint even if the original worker pool might be configured differently in the future?
```suggestion
testWorkerPool.Taints = []corev1.Taint{
			{
				Key:    testWorkerPoolName,
				Effect: corev1.TaintEffectNoSchedule,
				Value:  testWorkerPoolName,
			},
		}
```

test/integration/shoots/operations/clusterautoscaler.go
```diff
@@ -165,6 +166,125 @@ var _ = ginkgo.Describe("Shoot clusterautoscaler testing", func() {
...
+			workerPoolName              = shoot.Spec.Provider.Workers[0].Name
+			testWorkerPoolName          = "ca-test"
+			origWorkers                 = shoot.Spec.Provider.Workers
+		)
+		if shoot.Spec.Provider.Type != "aws" && shoot.Spec.Provider.Type != "azure" {
+			ginkgo.Skip("not applicable")
+		}
+
+		// Create a dedicated worker-pool for cluster autoscaler.
+		testWorkerPool := shoot.Spec.Provider.Workers[0]
```
timebertt:
```suggestion
		testWorkerPool := origWorkers[0]
```

test/integration/shoots/operations/clusterautoscaler.go
```diff
@@ -165,6 +166,125 @@ var _ = ginkgo.Describe("Shoot clusterautoscaler testing", func() {
...
+				return nil
+			})
+			framework.ExpectNoError(err)
+		}()
+
+		nodeList, err := framework.GetAllNodesInWorkerPool(ctx, f.ShootClient, &workerPoolName)
+		framework.ExpectNoError(err)
+
+		origNodeCount := len(nodeList.Items)
+		gomega.Expect(origNodeCount).To(gomega.BeEquivalentTo(origWorkers[0].Minimum), "shoot should have minimum node count before the test")
```
timebertt: I think, you would need to test here, if the **newly added** worker pool and not the original one has the minimum amount of nodes, right?

timebertt (changes requested):
Thanks @hardikdr for adding this test!
It already looks good overall, I just have some minor suggestions.

test/integration/shoots/operations/clusterautoscaler.go
hardikdr: This is actually related to the scale-to-zero feature. This label is for future node-objects, which will be created under this worker/machine-deployment. 
Cluster-autoscaler reads this from MachineDeployment and decides if reserve-excess-capacity pods[with label affinity]  will be able to schedule on the future nodes under this Machinedeployment, specifically when current worker size is 0.
I hope that's understandable, essentially it's dependent on how autoscaler works.

test/framework/resources/templates/reserve-capacity.yaml.tpl
hardikdr: yes, sounds good.

test/integration/shoots/operations/clusterautoscaler.go
hardikdr: That's pretty well-thought, thanks for the suggestion.

test/integration/shoots/operations/clusterautoscaler.go
danielfoehrKn: If I understand it correctly, this label is required because the cluster-autoscaler reads it from the  machine deployment? 
Is Gardener creating these labels during the Worker creation or who is setting these labels on the deployment? If yes, is Gardener always setting thee labels because they are required for the autoscaler feature? 
If that is the case, we can get rid of the[ kubelet flag to not have redundancy.](https://github.com/gardener/gardener/blob/51634b550b0819af24d2dd1a08d5a15288b48d46/charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.flags#L23)

timebertt: Got your point.
The `worker.gardener.cloud/pool` label is added by the kubelet itself (via flags configured in the systemd unit), but the cluster-autoscaler needs to know which labels a node in a specific pool will have, so it will look at the MachineDeployment, right?
So if I understand correctly what you are saying, an enduser who wants to use the scale-from-zero feature would need to explicitly add this label to the worker pool in his Shoot spec?
If that's the case, I would propose to enhance the Worker actuator to always add the labels, a pool will have, to the MachineDeployment.

rfranzke:
rebase to latest `master` please
also, please squash the commits :)

test/integration/shoots/operations/clusterautoscaler.go
hardikdr: >So if I understand correctly what you are saying, an end-user who wants to use the scale-from-zero feature would need to explicitly add this label to the worker pool in his Shoot spec?

Not really. I added this label only because I wanted the reserve-excess-capacity pods to be scheduled on this worker-pool only. In general, end-users don't need to add this label in shoot-spec unless their workload has a specific affinity towards the worker-pool. 
- Technically it could be any `test-label` and not necessarily worker pool-specific. 

>If that's the case, I would propose to enhance the Worker actuator to always add the labels, a pool will have, to the MachineDeployment.

So I think worker-actuator doesn't need any changes.

test/integration/shoots/operations/clusterautoscaler.go
hardikdr: >Or could the changed autoscaler config create problems in parallely executed tests?

It should not really be an issue running it in parallel in the future.  I'll make a change to remove the exact worker pool than overwriting to the original one, good point. Thanks.

test/integration/shoots/operations/clusterautoscaler.go
timebertt: > I added this label only because I wanted the reserve-excess-capacity pods to be scheduled on this worker-pool only

Sure. And what if the enduser wants to do the same and run some portion of his workload on a dedicated worker pool, which should be scaled down to / up from zero? Then they would need to explicitly add this label to the worker pool in his Shoot spec, correct?

test/integration/shoots/operations/clusterautoscaler.go
hardikdr: There is actually a catch here. I wanted to calculate the `requestCPUMilli` and `requestMemoryMegs` for the reserve-excess-capacity pods. I would need a node-object for that, but the new testWorkerPool doesn't have it as it's scaled to zero .. :D
- I'll fix it by re-fetching the nodeList from origWorkerPool, I hope that'd be fine.

test/integration/shoots/operations/clusterautoscaler.go
hardikdr: >Then they would need to explicitly add this label to the worker pool in his Shoot spec, correct? 

Yes, that's a very valid point, and possibly a gap also I think.

test/integration/shoots/operations/clusterautoscaler.go
timebertt: Sure, saw that coming :D 
So I think, using a node object from the old worker pool to calculate the resource requests, and checking that the new worker pool is at 0 at the beginning should be fine then.

test/integration/shoots/operations/clusterautoscaler.go
timebertt: Ok, then WDYT about enhancing the worker actuator to add the default pool labels to the MachineDeployment(s)? This would solve this problem, right?

Does mcm add then add these labels from MachineDeployments to the Node objects then? If so, then we should indeed avoid duplication, as @danielfoehrKn suggested (see https://github.com/gardener/gardener/pull/2160#discussion_r405413204).

hardikdr: >Does mcm add then add these labels from MachineDeployments to the Node objects then? 

Yes. I generally agree to set the worker-pool label via MachineDeployment, in that case, autoscaler would be able to read them and workload having an affinity to this label can basically trigger the scale-up of zeroed worker pools.

Though, I would also expect end-users to completely control the labels, if their workload is dependent on them via affinity.

timebertt: Perfect, then let's adapt this. But anyways not in this PR, I will open an issue for that.

> Though, I would also expect end-users to completely control the labels, if their workload is dependent on them via affinity.

Sure, of course, we should add the default labels (like `worker.gardener.cloud/pool` and so on) and any additionally specified worker pool labels from the Shoot spec to the MachineDeployment.

@rfranzke Should we adapt the worker contract and add that the worker controller is responsible for adding the node labels to avoid duplication? Or is there a specific reason, why we add the labels through kubelet flags?

rfranzke: Yes, sounds good. But I suggest to do it like @danielfoehrKn did it for the container runtime labels - let Gardener create the `Worker` extension object with the respective labels. Then we don't have to revendor the generic worker actuator in all extensions and get it out of the box. Also, it makes more sense to put it there as Gardener can control this label centrally - it's not necessary that every worker extension does it on its own.

https://github.com/gardener/gardener/pull/2128/files#diff-873d79d4ef156d5ab5a81e3dce9a0a41 - @tim-ebert you want to take this?

test/integration/shoots/operations/clusterautoscaler.go
hardikdr: done.

test/integration/shoots/operations/clusterautoscaler.go
hardikdr: done.

hardikdr:
> rebase to latest `master` please
> also, please squash the commits :)

Done.

test/integration/shoots/operations/clusterautoscaler.go
timebertt: Yep, that's exactly what I meant. I can take this 👍

rfranzke: Is this a prerequisite for this PR?

timebertt: No, we can go ahead with this one in the meantime.
But once we have introduced the labels change, we can remove this line setting the label explicitly here.

timebertt (dismissed):
Thanks @hardikdr 
Looks great now!
/approve

dguendisch (dismissed):

test/integration/shoots/operations/clusterautoscaler.go
hardikdr: Is there anything specific to be discussed here at current PR?

timebertt: I don't think so. Can we proceed with this PR @danielfoehrKn ?

test/integration/shoots/operations/clusterautoscaler.go
timebertt: There you go: https://github.com/gardener/gardener/pull/2166

timebertt: I tested both clusterautoscaler integration tests with #2166, and they worked out perfectly fine. It also worked without setting the worker pool label explicitly in your test. So we can go ahead and remove it :)

hardikdr: Sure, I removed the label from the worker pool, It would be nice if you could have a final check.

timebertt (approved):

test/integration/shoots/operations/clusterautoscaler.go
danielfoehrKn: Go ahead please.

danielfoehrKn (approved):

test/integration/shoots/operations/clusterautoscaler.go
schrodit: I already thought of the second options and wanted to implement that as soon as it would be needed.
But in this specific use case the ginkgo aftersuite is not suitable as we want the cleanup after this test has finished and not the complete suite.
Now there are several tests running in parallel and even for serial tests the execution order is not controllable.
This means that aftersuite would not solve the problem as we want a clean environment after this test is completed, not after all tests.

Possible options then are:
- defer functions
- or `BeforeEach` in its own context

Does it make sense?

danielfoehrKn: yup, thats how I have done it so far. I see your point, just from my test framework "user perspective" it would be nice to be able to register normal Ginko.AfterSuites or a similar functionality provided by the test framework. 
Just FYI, maybe that could be improved at some point.

PR #2162, opened 2020-04-08
Add providerConfig to Seed object

rfranzke:
The `Seed` object didn't have a `.spec.provider.providerConfig` field yet. Some extensions might need to have seed-specific configuration, and they can now use this field to define it.
During extension installation the complete seed spec is now passed to the extension's Helm chart so that the extension has access to each and every field.

/cc @MartinWeindel @mandelsoft 

**Release note**:

```improvement developer
The `Seed` object does now allow to specify a provider-specific and seed-specific configuration in the `.spec.provider.providerConfig` field.
```
```improvement developer
Gardener's `ControllerInstallation` controller does now pass the complete specification of the `Seed` object to the extension's Helm chart with the `.gardener.seed.spec` value. This way extension's can find out every information about the seed they get installed to.
```


MartinWeindel (approved):
/lgtm

timuthy (approved):
/lgtm

timebertt (approved):
/lgtm

pkg/apis/core/v1alpha1/types_seed.go
```diff
@@ -147,8 +147,11 @@ type ShootNetworks struct {
 type SeedProvider struct {
 	// Type is the name of the provider.
 	Type string `json:"type" protobuf:"bytes,1,opt,name=type"`
+	// ProviderConfig is the configuration passed to Seed resource.
+	// +optional
+	ProviderConfig *ProviderConfig `json:"providerConfig,omitempty" protobuf:"bytes,2,opt,name=providerConfig"`
 	// Region is a name of a region.
-	Region string `json:"region" protobuf:"bytes,2,opt,name=region"`
+	Region string `json:"region" protobuf:"bytes,3,opt,name=region"`
```
ialidzhikov, line 154: FYI: we cannot change in future the field numbers if we want to stay backwards compatible. Ref https://developers.google.com/protocol-buffers/docs/proto#updating

I think now is good as the protobufs are still not released.

rfranzke: OK, thanks for pointing this out, I wasn't aware. Thanks!

PR #2166, opened 2020-04-09
Set Node Labels via Worker instead of kubelet-flags

timebertt:
With this PR, the default Node labels (k8s node role, worker pool name and cri name) are added via the Worker resource instead of setting them via a kubelet flag.
Gardener add these labels that were previously added via kubelet flags directly to the worker spec and the provider extensions will be responsible for adding the specified labels to the worker nodes.
All extensions using the generic Worker actuator to create MachineDeployments will properly set the labels on the MachineDeployments out of the box, which will then be propagated to the Node objects by mcm.

This enables endusers to properly use the scale-from-zero feature when running scaled workload with label selectors for the default worker pool labels (`worker.gardener.cloud/pool`). (see https://github.com/gardener/gardener/pull/2160#discussion_r405358138)
cc @hardikdr 

Fixes #

cc @danielfoehrKn Please check, if the the CRI labels are still set correctly.

**Release note**:

```improvement operator
Gardener now sets the default Node labels via the Worker CRD instead of via kubelet flags.
```
```improvement developer
Gardener now sets the default Node labels via the Worker CRD instead of via kubelet flags. Please make sure that your worker controller properly propagates the specified labels to the Node objects.
```

### DIFF
--- a/pkg/operation/botanist/worker.go
+++ b/pkg/operation/botanist/worker.go
@@ -28,6 +28,7 @@ import (
 	"github.com/gardener/gardener/pkg/utils/kubernetes/health"
 	"github.com/gardener/gardener/pkg/utils/retry"
 	"github.com/gardener/gardener/pkg/utils/secrets"
+	versionutils "github.com/gardener/gardener/pkg/utils/version"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


pkg/operation/botanist/worker.go
```diff
@@ -78,53 +84,67 @@ func (b *Botanist) DeployWorker(ctx context.Context) error {
...
 		}
 
+		if workerPool.Labels == nil {
+			workerPool.Labels = map[string]string{}
+		}
+
+		// k8s node role labels
+		if k8sVersionLessThan115 {
+			workerPool.Labels["kubernetes.io/role"] = "node"
+			workerPool.Labels["node-role.kubernetes.io/node="] = ""
```
rfranzke:
```suggestion
			workerPool.Labels["node-role.kubernetes.io/node"] = ""
```

rfranzke (changes requested):
nit

pkg/operation/botanist/worker.go
timebertt: done

rfranzke (approved):

hardikdr (approved):

danielfoehrKn (approved):
Tested with containerd and gVisor CRI - working.
Thanks for picking it up.

PR #2167, opened 2020-04-09
Extend the infrastructure actuator interfece with Migrate and Restore

vlvasilev:
With this PR we add new functionality to Infrastructure actuator interface consisting of Migrate and Restore. Each provider must implement the corresponding functions.
Migrate must remove all related resources of the shoot control plane without deleting the IaaS resources.
Restore must read the state from Infrastructure.Status.State and base on it to make the proper shoot control plane resources(e.g config maps and secrets)
Part of [# 1631](https://github.com/gardener/gardener/issues/1631) 

**Release note**:

```improvement developer
Extend the infrastructure actuator interface with Migrate and Restore
```

### DIFF
--- a/extensions/pkg/controller/infrastructure/reconciler.go
+++ b/extensions/pkg/controller/infrastructure/reconciler.go
@@ -22,6 +22,7 @@ import (
 	"github.com/gardener/gardener/extensions/pkg/util"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"github.com/go-logr/logr"
@@ -40,8 +41,12 @@ import (
 const (
 	// EventInfrastructureReconciliation an event reason to describe infrastructure reconciliation.
 	EventInfrastructureReconciliation string = "InfrastructureReconciliation"
-	// EventInfrastructureDeleton an event reason to describe infrastructure deletion.
-	EventInfrastructureDeleton string = "InfrastructureDeleton"
+	// EventInfrastructureDeletion an event reason to describe infrastructure deletion.
+	EventInfrastructureDeletion string = "InfrastructureDeletion"
+	// EventInfrastructureMigration an event reason to describe infrastructure migration.
+	EventInfrastructureMigration string = "InfrastructureMigration"
+	// EventInfrastructureRestoration an event reason to describe infrastructure restoration.
+	EventInfrastructureRestoration string = "InfrastructureRestoration"
 )
 
 type reconciler struct {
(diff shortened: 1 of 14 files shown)


extensions/pkg/controller/infrastructure/reconciler.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/gardener/gardener/extensions/pkg/util"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	gardencorev1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```
rfranzke:
```suggestion
	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```

extensions/pkg/controller/utils.go
```diff
@@ -333,3 +335,31 @@ func GetVerticalPodAutoscalerObject() *unstructured.Unstructured {
 	obj.SetKind("VerticalPodAutoscaler")
 	return obj
 }
+
+// RemoveAnnotation removes an annotation key passed as annotation
+func RemoveAnnotation(ctx context.Context, c client.Client, obj runtime.Object, annotation string) error {
```
rfranzke, line 340: Can we have unit tests for these two functions here?

extensions/pkg/terraformer/state.go
```diff
@@ -160,3 +163,45 @@ func sniffJSONStateVersion(stateConfigMap []byte) (uint64, error) {
...
 }
+
+// Initialize implements
+func (f StateConfigMapInitializerFunc) Initialize(ctx context.Context, c client.Client, namespace, name string) error {
+	return f(ctx, c, namespace, name)
+}
+
+// CreateState create terraform state config map and use empty state
+// It does not create or update state ConfigMap if alredy exists
+func CreateState(ctx context.Context, c client.Client, namespace, name string) error {
```
rfranzke, line 174: cc @ialidzhikov - can you double-check these changes to the `terraformer` pkg here?

rfranzke (changes requested):
Only minor comments - looks good!
In order to test it I just have to take e.g. https://github.com/gardener/gardener-extension-provider-aws/pull/62 and revendor g/g to this PR, right?

extensions/pkg/predicate/predicate.go
```diff
@@ -173,7 +173,7 @@ func LastOperationNotSuccessful() predicate.Predicate {
 		}
 
 		lastOp := acc.GetExtensionStatus().GetLastOperation()
-		return lastOp == nil ||
+		return lastOp != nil &&
```
rfranzke: I'm not sure about all the consequences of this. Looping in @timuthy - WDYT?

rfranzke (commented):
Thanks for addressing the comments - no further suggestions from my side at the moment. Waiting for feedback from @ialidzhikov and @timuthy

extensions/pkg/predicate/predicate.go
timuthy: Good point. I think it is necessary to cover cases in which the last operation is not set at all, e.g. if one controller instance stops operation very early (maybe because the pod is evicted or it crashed) and the other starting instance must begin with the reconciliation again. The change above will prevent this in the future. What's the motivation for that change at all?

swilen-iwanow: If the pod crashes before setting the operation, Gardenlet will treat the infra as unsuccessful and annotate the object for reconciliation again right? The motivation was that when you create an Infrastructure object without annotations it was reconciled anyway, but we need to update its status with the state from the ShootState before starting the reconciliation. Because of this predicate the controller starts reconciliation, and we thought this check might not be correct because it states "LastOperationNotSuccessful", which in case of nil returns true. Anyway I am already adding new annotation to handle the wait-for-state scenario and @vlvasilev will prob revert this.

rfranzke: > The motivation was that when you create an Infrastructure object without annotations it was reconciled anyway

This is a valid point - indeed, when the operation annotation is not ignored then this should not happen.

timuthy: > If the pod crashes before setting the operation, Gardenlet will treat the infra as unsuccessful and annotate the object for reconciliation again right?

This only happens after the Gardener flow step times out. It'll then retry again, i.e. sets the operation annotation during the next reconciliation flow.

rfranzke: @vlvasilev @swilen-iwanow what's the status?

swilen-iwanow: We are going to revert this, but still have to implement the wait-for-state annotation handling for the extension controllers. Should we do this in this PR or open a separate.

rfranzke: Do you mean for the restoration case? Probably it should be this PR, or?

rfranzke:
PR looks good now, waiting for @ialidzhikov to double-check the Terraformer changes.
I couldn't play with it yet, I guess https://github.com/gardener/gardener-extension-provider-aws/pull/62 is outdated, or?
Also, what about https://github.com/gardener/gardener-extension-provider-aws/pull/62/files#r407865431? Should it be added to this PR?

rfranzke:
@ialidzhikov currently doesn't have the capacity to check it but agreed with merging it as @vlvasilev @plkokanov @swilen-iwanow thorougly tested

rfranzke (approved):
/approve

PR #1567, opened 2019-10-30
WIP: introduce scheduling method Selector

mandelsoft:
**What this PR does / why we need it**:
The new scheduling method `Selector` allows to make use of the new label selector field
`providers` to select seeds running on different providers depending of the provider list list and matching labels. It prefers matching seeds if the load is the same.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement operator
If automatic cross provider scheduling is desired, it  is possible know to specify a new seed selector field `providers` for a cloud profile to enable scheduling on seeds running on different providers.
It is observed if the scheduling method `Selector` has been chosen, which is the new default now.
```

PR #2169, opened 2020-04-09
introduce scheduling method Selector

mandelsoft:
This is a remake #1567 
A comprehensive solution seems to be in far distance, therefore here a new try based an the latest enhancements of the scheduler.

The new scheduling method `Selector` allows to make use of the new label selector field
`providers` to select seeds running on different providers depending of the provider list list and matching labels. It prefers matching seeds if the load is the same.

Fixes #

**Release note**:

```improvement operator
If automatic cross provider scheduling is desired, it  is possible know to specify a new seed selector field `providers` for a cloud profile to enable scheduling on seeds running on different providers.
It is observed if the scheduling method `Selector` has been chosen, which is the new default now.
```


pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -218,12 +219,21 @@ func filterSeedsMatchingSeedSelector(cloudProfile *gardencorev1beta1.CloudProfil
...
 	var candidates []*gardencorev1beta1.Seed
+	var providers []string
 
+	if cloudProfile.Spec.SeedSelector != nil {
+		providers = cloudProfile.Spec.SeedSelector.Providers
+	}
 	switch {
 	case shoot.Spec.Purpose != nil && *shoot.Spec.Purpose == gardencorev1beta1.ShootPurposeTesting:
 		candidates = determineCandidatesOfSameProvider(seedList, shoot)
+	case strategy == config.Selector:
```
rfranzke: Hm, why is it a separate strategy? I think this should be the default behaviour in all strategies. If `.spec.seedSelector.providers[]` was provided in the `CloudProfile` then they will be considered. The strategy is only used later to find the best match. WDYT? As you suggested, it probably makes more sense to move the scheduling strategy out of the gardener-scheduler configuration and into the `CloudProfile`, but let's do this in a separate PR.

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -218,12 +219,21 @@ func filterSeedsMatchingSeedSelector(cloudProfile *gardencorev1beta1.CloudProfil
...
+		if matchProvider(cloudProfile.Spec.Type, shoot.Spec.Provider.Type, providers) {
```
rfranzke: The signature of the `matchProvider` function suggests that the first argument is the seed's provider, but you are passing the provider of the `CloudProfile`. I'm confused.

Also, why are we executing this function here? We are calling it again inside `addCanditatesWithMatchingProviders`?!

rfranzke (changes requested):

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: This was to always start the list of candidates with matching region seeds if the shoots provider is in the list of accepted providers. This should prefer those seeds

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: You are right. Compared to the old base for the pull request, this behaviour has changed. Now the label selector is always matched prior to the concrete strategy. This should be kept and additionally the providers from the optional list should be filtered. I'll change this.

mandelsoft: Hmm, this will not work, because the other strategies check the type. This check could be removed, because it would be  filtered already by the `matchProvider` filter together with the labels. But anyway it cannot be combined with `SameRegion` strategy, because the regions of different providers will probably never match. Therefore this provider filter would be completely without function. Then thereis only the `MinimalDistance` strategy that potentially makes sense, but this is strange strategy so far. May be we should swicth here to the Levenshtein-Distance.

I've tested it (without Levensthein). It yields the same results compared to the Selector strategy

rfranzke: OK, good points, agreed that the `SameRegion` strategy doesn't make sense cross-provider.

What do you think about the following:

1. As suggested earlier, don't introduce a new strategy, but adapt the two existing ones:
1. The `SameRegion` prefers seeds having providers of the same type over other providers. If a seed of the same type exists then it chooses one with the same region. If no seed with the same provider exists then it chooses any seed with another matching provider, independent of the region (yes, it's not too intuitive for now, but until the whole region/minimal distance stuff is sorted out this is maybe an acceptable behaviour?)
1. The `MinimalDistance` strategy prefers seeds having providers of the same type and then chooses the one with the smallest distance (lexicographically, for now). If no such seed exists it chooses a seed of another matching provider and, again, finds the best region fit lexicographically.

pkg/scheduler/controller/shoot/scheduler_control.go
rfranzke: Hm, can it be made more readable? It's not too straight-forward to understand if you just read these lines.

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: This would counteract the name *SameReaion*. Instead I left the `SameRegion` as it is and introduced a `BestRegion` which is a combination of both. It uses `SameRegion` if such a seed exsists. Otherwise it falls back to `MinimalDistance`.

At the same time I improved the `MinimalDistance` strategy. It splits the region names into two parts. The **base region** and an **orientation** (north, south, east, west, central). The distance is then based on the levenshtein distance of the base region plus an adjustment for mismatching orientations. This assures that two regions differing only by their orientation are nearer to each other than two completely different regions.

mandelsoft:
New version (see comment above) introducing a better minimal distance strategy

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: This strategy has been deleted completely now.

pkg/apis/core/v1alpha1/types_cloudprofile.go
```diff
@@ -83,6 +83,16 @@ type CloudProfileSpec struct {
 	VolumeTypes []VolumeType `json:"volumeTypes,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,9,rep,name=volumeTypes"`
 }
 
+// SeedSelector contains constraints for selecting seed to be usable for shoots using a profile
+type SeedSelector struct {
+	// LabelSelector is optional and can be used to select seeds by their label settings
+	// +optional
+	*metav1.LabelSelector `json:",inline,omitempty" protobuf:"bytes,1,opt,name=labelSelector"`
```
ialidzhikov, line 93: I cannot find an example in k/k for embedded pointer field with tags `,inline,omitempty`. The only field with tags `,inline,omitempty` is this one https://github.com/kubernetes/kubernetes/blob/d8799658082ce4e3733968042248de024bd0154e/staging/src/k8s.io/api/networking/v1beta1/types.go#L154-L160 and it is not a pointer.

The non-pointer way works:
```golang
	// LabelSelector is optional and can be used to select seeds by their label settings
	// +optional
	metav1.LabelSelector `json:",inline,omitempty" protobuf:"bytes,1,opt,name=labelSelector"`
```

It looks like the go-to-protobuf generator does not (or cannot) handle embedded field of type poitner.

rfranzke: Should we then simply duplicate the content of the `LabelSelector` until this limitation is fixed?

```golang
type SeedSelector struct {
	// MatchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels
  // map is equivalent to an element of matchExpressions, whose key field is "key", the
  // operator is "In", and the values array contains only "value". The requirements are ANDed.
  // +optional
  MatchLabels map[string]string `json:"matchLabels,omitempty" protobuf:"bytes,1,rep,name=matchLabels"`
  // MatchExpressions is a list of label selector requirements. The requirements are ANDed.
  // +optional
  MatchExpressions []metav1.LabelSelectorRequirement `json:"matchExpressions,omitempty" protobuf:"bytes,2,rep,name=matchExpressions"`
	// Providers is optional and can be used by restricting seeds by their provider type. '*' can be used to enable seeds regardless of their provider type.
	// +optional
	Providers []string `json:"providers,omitempty" protobuf:"bytes,3,rep,name=providers"`
}

File: pkg/scheduler/controller/shoot/distance_test.go [hunk 4]
```diff
@@ -0,0 +1,76 @@
+package shoot
```
rfranzke, line 15: license header is missing

pkg/scheduler/apis/config/types.go
```diff
@@ -24,6 +24,8 @@ import (
 const (
 	// SameRegion Strategy determines a seed candidate for a shoot only if the cloud profile and region are identical
 	SameRegion CandidateDeterminationStrategy = "SameRegion"
+	// BestRegion Strategy prefers a matching region, only if no such seed exsists it tries to find the nearest one.
+	BestRegion CandidateDeterminationStrategy = "BestRegion"
```
rfranzke: Can you also add it here: https://github.com/gardener/gardener/blob/99bf01f0935d1333d284c60251717ea62102f7e4/pkg/scheduler/apis/config/v1alpha1/types.go#L26

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -19,6 +19,10 @@ import (
 	"fmt"
 	"strings"
 
+	// "github.com/texttheater/golang-levenshtein/levenshtein"
```
rfranzke: remove it?

pkg/scheduler/controller/shoot/distance_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Scheduler_Control", func() {
+	Context("orientation", func() {
+		It("handles name without orientation", func() {
+			n, o := orientation("europe")
```
rfranzke: Can we user better variable names than `n` and `o`?

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -218,14 +228,35 @@ func filterSeedsMatchingSeedSelector(cloudProfile *gardencorev1beta1.CloudProfil
 	return matchingSeeds, nil
 }
 
-func getCandidates(shoot *gardencorev1beta1.Shoot, seedList []*gardencorev1beta1.Seed, strategy config.CandidateDeterminationStrategy) ([]*gardencorev1beta1.Seed, error) {
+func filterSeedsMatchingProviders(cloudProfile *gardencorev1beta1.CloudProfile, shoot *gardencorev1beta1.Shoot, seedList []*gardencorev1beta1.Seed) ([]*gardencorev1beta1.Seed, error) {
+	var providers []string
```
rfranzke:
```suggestion
	var allowedSeedProviders []string
```

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -306,38 +350,89 @@ func determineCandidatesWithMinimalDistanceStrategy(seeds []*gardencorev1beta1.S
...
 
 	for _, seed := range seeds {
-		if seed.DeletionTimestamp == nil && seed.Spec.Provider.Type == shoot.Spec.Provider.Type && !gardencorev1beta1helper.TaintsHave(seed.Spec.Taints, gardencorev1beta1.SeedTaintInvisible) && common.VerifySeedReadiness(seed) {
+		if seed.DeletionTimestamp == nil && !gardencorev1beta1helper.TaintsHave(seed.Spec.Taints, gardencorev1beta1.SeedTaintInvisible) && common.VerifySeedReadiness(seed) {
 			seedRegion := seed.Spec.Provider.Region
-			var matchingCharacters int
-			for i := 0; i <= len(shootRegion); i++ {
-				prefix := shootRegion[:i]
-				if strings.HasPrefix(seedRegion, prefix) {
+			/*
```
rfranzke: Can you remove this old code instead of commenting it out?

rfranzke (changes requested):

rfranzke:
@danielfoehrKn can you please have a look?

pkg/apis/core/v1alpha1/types_cloudprofile.go
mandelsoft: It is also possible to omit the `*`, This has the same problem, all the coding checks for a nil pointer (optional), so all the tests and others must be adapted.

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -180,11 +184,17 @@ func determineSeed(shoot *gardencorev1beta1.Shoot, seedLister gardencorelisters.
...
-	candidates, err := getCandidates(shoot, filteredSeeds, strategy)
+	filteredSeeds, err = filterSeedsMatchingProviders(cloudProfile, shoot, filteredSeeds)
 	if err != nil {
 		return nil, err
 	}
 
+	candidates, err := getCandidates(cloudProfile, shoot, filteredSeeds, strategy)
+	if err != nil {
+		return nil, err
+
```
danielfoehrKn:
```suggestion
```

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -278,11 +309,24 @@ func getSeedWithLeastShootsDeployed(seedList []*gardencorev1beta1.Seed, shootLis
...
+	}
+	return false
+}
+
 func determineCandidatesOfSameProvider(seedList []*gardencorev1beta1.Seed, shoot *gardencorev1beta1.Shoot) []*gardencorev1beta1.Seed {
 	var candidates []*gardencorev1beta1.Seed
 	// Determine all candidate seed clusters matching the shoot's provider and region.
 	for _, seed := range seedList {
 		if seed.DeletionTimestamp == nil && seed.Spec.Provider.Type == shoot.Spec.Provider.Type && !gardencorev1beta1helper.TaintsHave(seed.Spec.Taints, gardencorev1beta1.SeedTaintInvisible) && common.VerifySeedReadiness(seed) {
+
```
danielfoehrKn:
```suggestion
```

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -306,38 +350,89 @@ func determineCandidatesWithMinimalDistanceStrategy(seeds []*gardencorev1beta1.S
 	}
 
 	var (
-		maxMatchingCharacters int
-		shootRegion           = shoot.Spec.Region
-		candidates            []*gardencorev1beta1.Seed
+		// maxMatchingCharacters int
```
danielfoehrKn:
```suggestion
```

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -218,14 +228,35 @@ func filterSeedsMatchingSeedSelector(cloudProfile *gardencorev1beta1.CloudProfil
...
+	var matchingSeeds []*gardencorev1beta1.Seed
+	for _, seed := range seedList {
+		if seed.DeletionTimestamp == nil && matchProvider(seed.Spec.Provider.Type, shoot.Spec.Provider.Type, providers) && !gardencorev1beta1helper.TaintsHave(seed.Spec.Taints, gardencorev1beta1.SeedTaintInvisible) && common.VerifySeedReadiness(seed) {
+			matchingSeeds = append(matchingSeeds, seed)
+		}
+	}
+	return matchingSeeds, nil
+}
+
+func getCandidates(cloudProfile *gardencorev1beta1.CloudProfile, shoot *gardencorev1beta1.Shoot, seedList []*gardencorev1beta1.Seed, strategy config.CandidateDeterminationStrategy) ([]*gardencorev1beta1.Seed, error) {
```
danielfoehrKn:
```suggestion
func getCandidates(shoot *gardencorev1beta1.Shoot, seedList []*gardencorev1beta1.Seed, strategy config.CandidateDeterminationStrategy) ([]*gardencorev1beta1.Seed, error) {
```
unused

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -218,14 +228,35 @@ func filterSeedsMatchingSeedSelector(cloudProfile *gardencorev1beta1.CloudProfil
...
-func getCandidates(shoot *gardencorev1beta1.Shoot, seedList []*gardencorev1beta1.Seed, strategy config.CandidateDeterminationStrategy) ([]*gardencorev1beta1.Seed, error) {
+func filterSeedsMatchingProviders(cloudProfile *gardencorev1beta1.CloudProfile, shoot *gardencorev1beta1.Shoot, seedList []*gardencorev1beta1.Seed) ([]*gardencorev1beta1.Seed, error) {
+	var providers []string
+	if cloudProfile.Spec.SeedSelector != nil {
+		providers = cloudProfile.Spec.SeedSelector.Providers
+	}
+
+	var matchingSeeds []*gardencorev1beta1.Seed
+	for _, seed := range seedList {
+		if seed.DeletionTimestamp == nil && matchProvider(seed.Spec.Provider.Type, shoot.Spec.Provider.Type, providers) && !gardencorev1beta1helper.TaintsHave(seed.Spec.Taints, gardencorev1beta1.SeedTaintInvisible) && common.VerifySeedReadiness(seed) {
```
danielfoehrKn: `seed.DeletionTimestamp == nil && !gardencorev1beta1helper.TaintsHave(seed.Spec.Taints, gardencorev1beta1.SeedTaintInvisible) && common.VerifySeedReadiness(seed) `   is checked in three different places. 
See [here](https://github.com/gardener/gardener/blob/547bdffba5d0487d1ab453801613109094cabfa4/pkg/scheduler/controller/shoot/scheduler_control.go#L328-L327) and [here](https://github.com/gardener/gardener/blob/547bdffba5d0487d1ab453801613109094cabfa4/pkg/scheduler/controller/shoot/scheduler_control.go#L340). 
I think it makes sense to introduce another function that filters out seed with deletion timestamps, with a taint and seeds that are not ready. This way we do not have to make the same check over again.

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: This was just for discussion of my proposed change :-)

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -293,7 +337,7 @@ func determineCandidatesOfSameProvider(seedList []*gardencorev1beta1.Seed, shoot
 func determineCandidatesWithSameRegionStrategy(seedList []*gardencorev1beta1.Seed, shoot *gardencorev1beta1.Shoot) []*gardencorev1beta1.Seed {
 	var candidates []*gardencorev1beta1.Seed
 	for _, seed := range seedList {
-		if seed.DeletionTimestamp == nil && seed.Spec.Provider.Type == shoot.Spec.Provider.Type && seed.Spec.Provider.Region == shoot.Spec.Region && !gardencorev1beta1helper.TaintsHave(seed.Spec.Taints, gardencorev1beta1.SeedTaintInvisible) && common.VerifySeedReadiness(seed) {
+		if seed.DeletionTimestamp == nil && seed.Spec.Provider.Region == shoot.Spec.Region && !gardencorev1beta1helper.TaintsHave(seed.Spec.Taints, gardencorev1beta1.SeedTaintInvisible) && common.VerifySeedReadiness(seed) {
```
danielfoehrKn: Can you explain why you removed the provider check here? 
For SameRegion  that means if two cloud providers name their regions identical, a Shoot with provider AWS may end up on GCP because the load of Shoot is smaller. Thats a behaviour change. Have I missed that we want to change that?

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -218,14 +228,35 @@ func filterSeedsMatchingSeedSelector(cloudProfile *gardencorev1beta1.CloudProfil
...
 	switch {
 	case shoot.Spec.Purpose != nil && *shoot.Spec.Purpose == gardencorev1beta1.ShootPurposeTesting:
 		candidates = determineCandidatesOfSameProvider(seedList, shoot)
 	case strategy == config.SameRegion:
 		candidates = determineCandidatesWithSameRegionStrategy(seedList, shoot)
+	case strategy == config.BestRegion:
+		candidates = determineCandidatesWithSameRegionStrategy(seedList, shoot)
+		if len(candidates) > 0 {
+			return candidates, nil
+		}
```
danielfoehrKn:
```suggestion
				filtered := determineCandidatesWithSameRegionStrategy(seedList, shoot)
		if len(filtered) > 0 {
			return candidates, nil
		}
```
This is to not override the candidates so that you can have the provider check in the determineCandidatesWithSameRegionStrategy().
Also see my other comment there.

danielfoehrKn (changes requested):

mandelsoft:
yepp, I've fixed the *protobug* generater and added a forked version just for testing.
protobuf generation now succeeds (at least locally).

Could you verify that the api server communication still works, then I would open a PR for the `kubernetes/code-generator` project

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: yepp, but this was the case even before my change

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: It does not overwrite anything. Every *case* set the `candidates` variable. It is not filtered or reused, therefore there is no reason to use another variable for a dedicated case.

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: This was indeed an error, I delete too much during requested refactoring

mandelsoft:
Opened a pull request kubernetes/kubernetes#90296 for the protobuf fix.
So far only for master. Let's see what happens.
As discussed with @rfranzke we will need it for 1.16.

pkg/apis/core/v1alpha1/types_cloudprofile.go
mandelsoft: I finally managed to fix the protobuf generator ( see comment below)

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -306,38 +368,66 @@ func determineCandidatesWithMinimalDistanceStrategy(seeds []*gardencorev1beta1.S
...
+	}
+	return name, ""
+}
+
+// distance calculates a formal distance between two region names observing
+// some usual orientation keywords. It is based on the levenshtein distance
+// of the regions base names plus the difference based on the orientation.
+// regions with the same base but different orientations are basically nearer
+// to each other than two completely diffrent regions.
```
danielfoehrKn:
```suggestion
// to each other than two completely different regions.
```

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -306,38 +368,66 @@ func determineCandidatesWithMinimalDistanceStrategy(seeds []*gardencorev1beta1.S
...
+			minDistance = dist
+			candidates = []*gardencorev1beta1.Seed{seed}
 		}
 	}
 	return candidates
 }
 
+var orientations = []string{"north", "south", "east", "west", "central"}
+
+// distance extracts an orientation relative to a base from a region name
```
danielfoehrKn:
```suggestion
// orientation extracts an orientation relative to a base from a region name
```

pkg/scheduler/apis/config/types.go
```diff
@@ -24,6 +24,8 @@ import (
 const (
 	// SameRegion Strategy determines a seed candidate for a shoot only if the cloud profile and region are identical
 	SameRegion CandidateDeterminationStrategy = "SameRegion"
+	// BestRegion Strategy prefers a matching region, only if no such seed exsists it tries to find the nearest one.
```
danielfoehrKn:
```suggestion
	// BestRegion Strategy prefers a matching region, only if no such seed exists it tries to find the nearest one.
```

pkg/scheduler/apis/config/v1alpha1/types.go
```diff
@@ -19,13 +19,17 @@ import (
...
 	componentbaseconfigv1alpha1 "k8s.io/component-base/config/v1alpha1"
+
+	"github.com/gardener/gardener/pkg/scheduler/apis/config"
 )
 
 const (
 	// SameRegion Strategy determines a seed candidate for a shoot only if the cloud profile and region are identical
-	SameRegion CandidateDeterminationStrategy = "SameRegion"
+	SameRegion CandidateDeterminationStrategy = CandidateDeterminationStrategy(config.SameRegion)
+	// BestRegion Strategy prefers a matching region, only if no such seed exsists it tries to find the nearest one.
```
danielfoehrKn:
```suggestion
	// BestRegion Strategy prefers a matching region, only if no such seed exists it tries to find the nearest one.
```

pkg/apis/core/types_cloudprofile.go
```diff
@@ -61,13 +61,23 @@ type CloudProfileSpec struct {
...
 }
 
+// SeedSelector contains constraints for selecting seed to be usable for shoots using a profile
+type SeedSelector struct {
+	// LabelSelector is optional and can be used to select seeds by their label settings
+	// +optional
+	*metav1.LabelSelector
+	// Providers is optional and can be used by restricting seeds by their provider type. '*' can be used to enable seeds regardless of their provider type.
+	// +optional
+	Providers []string
```
danielfoehrKn:
```suggestion
	ProviderTypes []string
```
WDYT about renaming it? Because it is consistently found in Spec.Provider.type in Shoot & Seed

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -193,10 +205,29 @@ func determineSeed(shoot *gardencorev1beta1.Shoot, seedLister gardencorelisters.
...
+		return nil, fmt.Errorf("none of the %d seeds is valid for scheduling (not deleting, visible and ready)", len(seedList))
+	}
+	return matchingSeeds, nil
+}
+
 func filterSeedsMatchingSeedSelector(cloudProfile *gardencorev1beta1.CloudProfile, seedList []*gardencorev1beta1.Seed) ([]*gardencorev1beta1.Seed, error) {
 	selector := &metav1.LabelSelector{}
-	if cloudProfile.Spec.SeedSelector != nil {
-		selector = cloudProfile.Spec.SeedSelector
+	if cloudProfile.Spec.SeedSelector != nil && cloudProfile.Spec.SeedSelector.LabelSelector != nil {
```
danielfoehrKn: can we not exit here if cloudProfile.Spec.SeedSelector.LabelSelector == nil ?. If the CloudProfile does not specify a label selector, none of them are filtered out based on that criteria? 
This way we do not have to loop over the seeds.

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -218,6 +247,21 @@ func filterSeedsMatchingSeedSelector(cloudProfile *gardencorev1beta1.CloudProfil
 	return matchingSeeds, nil
 }
 
+func filterSeedsMatchingProviders(cloudProfile *gardencorev1beta1.CloudProfile, shoot *gardencorev1beta1.Shoot, seedList []*gardencorev1beta1.Seed) ([]*gardencorev1beta1.Seed, error) {
+	var possibleProviders []string
+	if cloudProfile.Spec.SeedSelector != nil {
```
danielfoehrKn, line 247: same thing here, can we not just return all seeds if  cloudProfile.Spec.SeedSelector.Providers empty?

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -226,6 +270,12 @@ func getCandidates(shoot *gardencorev1beta1.Shoot, seedList []*gardencorev1beta1
...
 	case strategy == config.SameRegion:
 		candidates = determineCandidatesWithSameRegionStrategy(seedList, shoot)
+	case strategy == config.BestRegion:
+		candidates = determineCandidatesWithSameRegionStrategy(seedList, shoot)
+		if len(candidates) > 0 {
+			return candidates, nil
+		}
+		fallthrough
 	case strategy == config.MinimalDistance:
 		candidates = determineCandidatesWithMinimalDistanceStrategy(seedList, shoot)
```
danielfoehrKn: If a scheduler is used the "BestRegion" strategy the function determineCandidatesWithSameRegionStrategy() is always executed twice (once in the switch and once after the fallthrough in the func [determineCandidatesWithMinimalDistanceStrategy](https://github.com/gardener/gardener/blob/a3d1c24b0c042c241cc0a4030846baa5ebe6d16d/pkg/scheduler/controller/shoot/scheduler_control.go#L366)).  In my opinion we should avoid e.g by adjusting the  determineCandidatesWithMinimalDistanceStrategy() to not call determineCandidatesWithSameRegionStrategy(). In the "BestSeed" we would then just call the determineCandidatesWithMinimalDistanceStrategy() directly.

```suggestion
              candidates = determineCandidatesWithSameRegionStrategy(seedList, shoot)
		if len(candidates) > 0 {
			return candidates, nil
		}
		candidates = determineCandidatesWithMinimalDistanceStrategy(seedList, shoot)
```

danielfoehrKn (changes requested):
The cross-provider filtering & the new minimal distance implementation I think are a nice improvement for the scheduler.

One thing that I do not quite understand is:  What is the need to introduce another scheduling strategy "BestSeed" - that does nothing else then calling the SameRegion() Strategy and then the Minimal DistanceStrategy(). That is exactly what the MinimalDistance()Strategy already does today.

For MinimalDistance I would expect
1) prefer same provider type and region 
2) prefer same provider - region over region with same name from other provider. E.g aws & gcp both have a region us-west-1.
3) if that is not the case, only then schedule cross provider

Regarding 2) - I do not think that is possible with the current implementation yet. Currently it only calculates the distance based on the region names without taking the provider into consideration or having a special handling for the described case.

Additionally, if we decide to keep "BestRegion" then I think we also need unit tests that check that it has the expected functionality -same region + minimalDistance (independent of the tests of each Strategy)

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -306,38 +368,66 @@ func determineCandidatesWithMinimalDistanceStrategy(seeds []*gardencorev1beta1.S
...
+	}
+	return name, ""
+}
+
+// distance calculates a formal distance between two region names observing
+// some usual orientation keywords. It is based on the levenshtein distance
+// of the regions base names plus the difference based on the orientation.
+// regions with the same base but different orientations are basically nearer
+// to each other than two completely diffrent regions.
+func distance(seed, shoot string) (int, int) {
```
timuthy: Can we get rid of second return type (levenshtein distance)? I think it makes the signature more complex and it's not really used except by some unit tests.

pkg/scheduler/controller/shoot/distance_test.go
```diff
@@ -0,0 +1,90 @@
...
+			Expect(leven).To(Equal(6))
+			Expect(dist).To(Equal(leven*2 + 1))
+		})
+
+		It("handles identical base with diffent orientation", func() {
+			dist, leven := distance("europe-west", "europe-east")
+			Expect(leven).To(Equal(0))
+			Expect(dist).To(Equal(2))
+		})
+
```
timuthy, line 77: Can you add another test like this:

```
It("handles identical base with diffent orientation", func() {
			dist, leven := distance("europe-west", "europe-central")
			Expect(leven).To(Equal(0))
			Expect(dist).To(Equal(2))
		})
```

It points out that `europe-central` has the same distance as `europe-east` (or any other `europa-<orient>`) to `europe-west`.

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/texttheater/golang-levenshtein/levenshtein"
```
timuthy: Please move to 3rd import block.

pkg/apis/core/types_cloudprofile.go
timuthy: What do you think about defaulting this to `spec.type` if no provider is set? It'd make [matchProvider](https://github.com/gardener/gardener/pull/2169/files?file-filters%5B%5D=.go&file-filters%5B%5D=.md&file-filters%5B%5D=.mod&file-filters%5B%5D=.proto&file-filters%5B%5D=.report&file-filters%5B%5D=.sum&file-filters%5B%5D=.txt#diff-9084f8a80e30e2e4bc3aeb5546a21bdaR333) more explicit.

timuthy (changes requested):
Nice PR!  I added a few comments / questions. Can you please also update the release notes which mentions the new strategy `BestRegion`? It would also be nice to see the new strategy being documented [here](https://github.com/gardener/gardener/blob/master/docs/concepts/scheduler.md).

rfranzke:
/kind/enhancement

rfranzke:
@mandelsoft can you check the latest comments and update the PR?

pkg/scheduler/controller/shoot/distance_test.go
mandelsoft: yep, this is the case and the test already exists. Any better Idea to incorporate orientation differences into the global distance value.

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: I've removed BestRegion again according to your overall remark.

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: if possibleProviders does not contain anything, there is still a filtering, it selects only seeds with a provider matching the shoots provider

pkg/apis/core/types_cloudprofile.go
mandelsoft: done

pkg/scheduler/apis/config/v1alpha1/types.go
mandelsoft: deleted

mandelsoft:
> That is exactly what the MinimalDistance()Strategy already does today.

@danielfoehrKn You are completely right, I missed the first line in minimal distance.
But nevertheless don't you think, that we should offer both:
- only based on difference  `MinimalDistance` and
- Preferred a matching region and then minimal distance?

The general problem here is the word: it should *prefer*. Basically there no such mechanism in the scheduler so far. It just selects the seed with the smallest number of shoots from a list of possible shoots. So, what should *prefer* mean?  If we use only matching seed, if such seeds exist, the others will never be selected, even if they are empty and the matching ones are *full*.  Basically we are missing some kind of weight, that is assigned to every potential seed and that is multiplied with the number of shoots. A preselection as done so far in *MinimalDistance* does not make sense.

Should I try to add such a weight in this PR or is it better to handle this separately.

In * MinimalDistance* a variant then would be to use the distance somehow as part of the weight, so that seeds could be used, even if they are further away if the nearer ones are quite busy.

danielfoehrKn:
@mandelsoft I am also for introducing weights - I would opt for  introducing them in another PR.

With the current implementation we would get cross-provider scheduling based on the new minimal distance strategy what is definitely an improvement.

In another PR the scheduling then could be made more configurable via weights. This would also be nice in case we want to adjust the scheduling behaviour based on the current load on the seeds.

Just to check: What would be possible weights?
- [number_of_shoots] *  [distance] 

However I am not sure how to include a preference for the case that regions in different cloud providers have the same name (one provider is the provider of the Shoot). Maybe another weight?-> [see post above](https://github.com/gardener/gardener/pull/2169#pullrequestreview-396391545).
We could create a dedicated issue to follow up.
WDYT

mandelsoft:
That sounds good.

The weight shout somehow include the size/capacity of the seed. Because of the autoscaling
this seems to difficult. But there is a useful maximum capacity. Therefore the weight should incorporate the quotient of the actual size and this maximum capacity (may be this must be a configuration option of the seed)

We still have the protobuf problem. I've opened a pull request there, but absolutely nothing happens. The only alternative would be to use a forked version with the fix, instead.

Changing the intended manifest structure just to work around the problem would be quite ugly.

rfranzke:
/hold

until https://github.com/kubernetes/kubernetes/pull/90296 got feedback

rfranzke:
/area/usability
/topology/garden
/platform/all
/priority/normal
/needs/rebase

rfranzke:
The Kubernetes PR has been merged 🎉 Can we pin the `k8s.io/code-generator` dependency to this commit (I hope we can skip versions as we don't want to fork and cherry-pick it) and rebase this PR?

timuthy:
@mandelsoft can you please check the unresolved comments and rebase?

mandelsoft:
Unfortunately even 1.17 is too old for the code generator to update to the latest version.

```
# github.com/gardener/gardener/pkg/client/core/clientset/versioned/typed/core/v1beta1
../../../client/core/clientset/versioned/typed/core/v1beta1/backupbucket.go:72:5: too many arguments in call to c.client.Get().Resource("backupbuckets").Name(name).VersionedParams(&options, scheme.ParameterCodec).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1beta1/backupbucket.go:88:5: too many arguments in call to c.client.Get().Resource("backupbuckets").VersionedParams(&opts, scheme.ParameterCodec).Timeout(timeout).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1beta1/backupbucket.go:104:8: too many arguments in call to c.client.Get().Resource("backupbuckets").VersionedParams(&opts, scheme.ParameterCodec).Timeout(timeout).Watch
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1beta1/backupbucket.go:114:5: too many arguments in call to c.client.Post().Resource("backupbuckets").VersionedParams(&opts, scheme.ParameterCodec).Body(backupBucket).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1beta1/backupbucket.go:127:5: too many arguments in call to c.client.Put().Resource("backupbuckets").Name(backupBucket.ObjectMeta.Name).VersionedParams(&opts, scheme.ParameterCodec).Body(backupBucket).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1beta1/backupbucket.go:142:5: too many arguments in call to c.client.Put().Resource("backupbuckets").Name(backupBucket.ObjectMeta.Name).SubResource("status").VersionedParams(&opts, scheme.ParameterCodec).Body(backupBucket).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1beta1/backupbucket.go:153:5: too many arguments in call to c.client.Delete().Resource("backupbuckets").Name(name).Body(&opts).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1beta1/backupbucket.go:168:5: too many arguments in call to c.client.Delete().Resource("backupbuckets").VersionedParams(&listOpts, scheme.ParameterCodec).Timeout(timeout).Body(&opts).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1beta1/backupbucket.go:181:5: too many arguments in call to c.client.Patch(pt).Resource("backupbuckets").Name(name).SubResource(subresources...).VersionedParams(&opts, scheme.ParameterCodec).Body(data).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1beta1/backupentry.go:75:5: too many arguments in call to c.client.Get().Namespace(c.ns).Resource("backupentries").Name(name).VersionedParams(&options, scheme.ParameterCodec).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1beta1/backupentry.go:75:5: too many errors
note: module requires Go 1.14
# github.com/gardener/gardener/pkg/client/core/clientset/versioned/typed/core/v1alpha1
../../../client/core/clientset/versioned/typed/core/v1alpha1/backupbucket.go:72:5: too many arguments in call to c.client.Get().Resource("backupbuckets").Name(name).VersionedParams(&options, scheme.ParameterCodec).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1alpha1/backupbucket.go:88:5: too many arguments in call to c.client.Get().Resource("backupbuckets").VersionedParams(&opts, scheme.ParameterCodec).Timeout(timeout).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1alpha1/backupbucket.go:104:8: too many arguments in call to c.client.Get().Resource("backupbuckets").VersionedParams(&opts, scheme.ParameterCodec).Timeout(timeout).Watch
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1alpha1/backupbucket.go:114:5: too many arguments in call to c.client.Post().Resource("backupbuckets").VersionedParams(&opts, scheme.ParameterCodec).Body(backupBucket).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1alpha1/backupbucket.go:127:5: too many arguments in call to c.client.Put().Resource("backupbuckets").Name(backupBucket.ObjectMeta.Name).VersionedParams(&opts, scheme.ParameterCodec).Body(backupBucket).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1alpha1/backupbucket.go:142:5: too many arguments in call to c.client.Put().Resource("backupbuckets").Name(backupBucket.ObjectMeta.Name).SubResource("status").VersionedParams(&opts, scheme.ParameterCodec).Body(backupBucket).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1alpha1/backupbucket.go:153:5: too many arguments in call to c.client.Delete().Resource("backupbuckets").Name(name).Body(&opts).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1alpha1/backupbucket.go:168:5: too many arguments in call to c.client.Delete().Resource("backupbuckets").VersionedParams(&listOpts, scheme.ParameterCodec).Timeout(timeout).Body(&opts).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1alpha1/backupbucket.go:181:5: too many arguments in call to c.client.Patch(pt).Resource("backupbuckets").Name(name).SubResource(subresources...).VersionedParams(&opts, scheme.ParameterCodec).Body(data).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1alpha1/backupentry.go:75:5: too many arguments in call to c.client.Get().Namespace(c.ns).Resource("backupentries").Name(name).VersionedParams(&options, scheme.ParameterCodec).Do
	have (context.Context)
	want ()
../../../client/core/clientset/versioned/typed/core/v1alpha1/backupentry.go:75:5: too many errors
note: module requires Go 1.14
FAIL	github.com/gardener/gardener/pkg/scheduler/controller/shoot [build failed]
```

ialidzhikov:
@mandelsoft , see https://twitter.com/ibuildthecloud/status/1230542321508184064.
https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20200123-client-go-ctx.md is added with v1.18

rfranzke:
I am open for a separate PR that bumps our Kubernetes dependencies to v1.18 now that we have bumped them to 1.17 already. However, will your code-generator fix work even with v1.18? It's merged into today's `master` branch of k/k, so actually targetted for 1.19, right?

timuthy:
https://github.com/kubernetes/kubernetes/pull/91731 finally got merged. Can you please rebase and re-vendor @mandelsoft?

rfranzke:
ping @mandelsoft

pkg/apis/core/types_shoot.go
```diff
@@ -83,7 +83,7 @@ type ShootSpec struct {
 	// SeedName is the name of the seed cluster that runs the control plane of the Shoot.
 	SeedName *string
 	// SeedSelector is an optional selector which must match a seed's labels for the shoot to be scheduled on that seed.
-	SeedSelector *metav1.LabelSelector
+	SeedSelector *SeedSelector
```
rfranzke, line 86: As discussed, I would not change this field for now and only limit the adaptation of the seed selector to the `CloudProfile` (unless there is a good reason to do it also in the `Shoot`).

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -175,61 +176,98 @@ func determineSeed(shoot *gardencorev1beta1.Shoot, seedLister gardencorelisters.
...
+	if err == nil {
+		filteredSeeds, err = filterCandidates(shoot, filteredSeeds)
 	}
+	if err == nil {
+		filteredSeeds, err = applyStrategy(shoot, filteredSeeds, strategy)
+	}
+	if err == nil {
+		return getSeedWithLeastShootsDeployed(filteredSeeds, shootList)
+	}
+	return nil, err
```
rfranzke: TBH, I get the motivation for this, but I would still prefer the standard style where the error is handled first. This is what we use all over the places.

rfranzke (commented):
Can you please adapt

* https://github.com/gardener/gardener/blob/master/docs/concepts/scheduler.md
* https://github.com/gardener/gardener/blob/master/example/30-cloudprofile.yaml#L11-L16

Also, kindly squash the commits appropriately.

pkg/scheduler/apis/config/types.go
```diff
@@ -24,6 +24,8 @@ import (
 const (
 	// SameRegion Strategy determines a seed candidate for a shoot only if the cloud profile and region are identical
 	SameRegion CandidateDeterminationStrategy = "SameRegion"
+	// BestRegion Strategy prefers a matching region, only if no such seed exists it tries to find the nearest one.
+	BestRegion CandidateDeterminationStrategy = "BestRegion"
```
danielfoehrKn: BestRegion Strategy [is missing here](https://github.com/gardener/gardener/blob/a546bdb06a753ae7bf912a12d8a9f078990226bf/pkg/scheduler/apis/config/types.go#L46-L45)  and [here](https://github.com/gardener/gardener/blob/a546bdb06a753ae7bf912a12d8a9f078990226bf/pkg/scheduler/apis/config/v1alpha1/types.go#L48-L47) leading to validation errors.

pkg/apis/core/v1alpha1/types_cloudprofile.go
```diff
@@ -84,6 +84,16 @@ type CloudProfileSpec struct {
...
+	// Providers is optional and can be used by restricting seeds by their provider type. '*' can be used to enable seeds regardless of their provider type.
+	// +optional
+	ProviderTypes []string `json:"providers,omitempty" protobuf:"bytes,2,rep,name=providerTypes"`
```
danielfoehrKn:
```suggestion
	ProviderTypes []string `json:"providerTypes,omitempty" protobuf:"bytes,2,rep,name=providerTypes"`
```

pkg/apis/core/types_cloudprofile.go
```diff
@@ -73,6 +73,16 @@ func (c *CloudProfile) GetProviderType() string {
 	return c.Spec.Type
 }
 
+// SeedSelector contains constraints for selecting seed to be usable for shoots using a profile
+type SeedSelector struct {
+	// LabelSelector is optional and can be used to select seeds by their label settings
+	// +optional
+	*metav1.LabelSelector
+	// Providers is optional and can be used by restricting seeds by their provider type. '*' can be used to enable seeds regardless of their provider type.
```
danielfoehrKn:
```suggestion
	// ProviderTypes contains a list of allowed provider types
	// used by the Gardener scheduler to restricting seeds by their provider type and enable cross-provider scheduling - by default, Shoots are only scheduled on Seeds having the same provider type.
	// '*' can be used to enable seeds regardless of their provider type.
```

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -175,61 +176,98 @@ func determineSeed(shoot *gardencorev1beta1.Shoot, seedLister gardencorelisters.
...
+
+func applyStrategy(shoot *gardencorev1beta1.Shoot, seedList []*gardencorev1beta1.Seed, strategy config.CandidateDeterminationStrategy) ([]*gardencorev1beta1.Seed, error) {
 	var candidates []*gardencorev1beta1.Seed
 
 	switch {
 	case shoot.Spec.Purpose != nil && *shoot.Spec.Purpose == gardencorev1beta1.ShootPurposeTesting:
 		candidates = determineCandidatesOfSameProvider(seedList, shoot)
 	case strategy == config.SameRegion:
 		candidates = determineCandidatesWithSameRegionStrategy(seedList, shoot)
+	case strategy == config.BestRegion:
```
danielfoehrKn: Currently there are three strategies
 - same region -> same cloud provider + same region name of Seed and Shoot
 - bestRegion -> same region, if no match: minimal distance
 - minimal distance -> determine minimal distance (levenshtein distance) across provider types

Why do we need the "bestRegion" strategy? I think it would be easier to just have "minimalDistance" that (like before) executes the "sameRegion" and if unsuccessful uses the levenshtein distance.

If "bestRegion" region is wanted, then I am missing unit test for it - even though it is just a combination of two strategies.

Additionally, the current behaviour of the "minimalDistance" has changed as "sameRegion" is not executed any more.

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -175,61 +176,98 @@ func determineSeed(shoot *gardencorev1beta1.Shoot, seedLister gardencorelisters.
...
+		return nil, fmt.Errorf("none of the %d seeds is valid for scheduling (not deleting, visible and ready)", len(seedList))
+	}
+	return matchingSeeds, nil
 }
 
-func filterSeedsMatchingSeedSelector(seedList []*gardencorev1beta1.Seed, labelSelector *metav1.LabelSelector, kind string) ([]*gardencorev1beta1.Seed, error) {
-	selector := &metav1.LabelSelector{}
-	if labelSelector != nil {
-		selector = labelSelector
+func filterSeedsMatchingSeedSelector(seedList []*gardencorev1beta1.Seed, seedSelector *gardencorev1beta1.SeedSelector, kind string) ([]*gardencorev1beta1.Seed, error) {
```
danielfoehrKn:
```suggestion
func filterSeedsMatchingLabelSelector(seedList []*gardencorev1beta1.Seed, seedSelector *gardencorev1beta1.SeedSelector, kind string) ([]*gardencorev1beta1.Seed, error) {
```

would prefer that name as this check only uses the label selector of the provided Seed selector.

danielfoehrKn (dismissed):

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: a matching region always has minimal distance. Therefore it should always be preferred.
I've removed the BestRegion again.

Best region is deleted again

pkg/apis/core/types_shoot.go
mandelsoft: Why do you want a different behaviour for the seed selector for shoots and profiles? I makes sense to allow shoots to additionally restrict seeds according to their provider type.

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: Just trying ....

pkg/scheduler/apis/config/types.go
mandelsoft: deleted

pkg/scheduler/controller/shoot/scheduler_control.go
danielfoehrKn: > a matching region always has minimal distance. Therefore it should always be preferred.

but two cloud providers can name their regions identically. In this case I would prefer using the same cloud provider (what same region strategy does).

pkg/apis/core/types_shoot.go
rfranzke: Hm, okay, agreed, I don't have a good argument against it. Let's keep it then.

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: I could incorporate the provider in the distance, so that same provider always is closer

ialidzhikov:
@mandelsoft could you please check `make check-generate`?

pkg/apis/core/types_cloudprofile.go
```diff
@@ -73,6 +73,18 @@ func (c *CloudProfile) GetProviderType() string {
 	return c.Spec.Type
 }
 
+// SeedSelector contains constraints for selecting seed to be usable for shoots using a profile
+type SeedSelector struct {
+	// LabelSelector is optional and can be used to select seeds by their label settings
+	// +optional
```
ialidzhikov: By convention we do not apply the `// +optional` to the internal types. You could remove it.

```suggestion
```

pkg/apis/core/types_cloudprofile.go
```diff
@@ -73,6 +73,18 @@ func (c *CloudProfile) GetProviderType() string {
...
+	*metav1.LabelSelector
+	// ProviderTypes contains a list of allowed provider types used by the Gardener scheduler to restricting seeds by
+	// their provider type and enable cross-provider scheduling.
+	// By default, Shoots are only scheduled on Seeds having the same provider type.
+	// +optional
```
ialidzhikov: By convention we do not apply the `// +optional` to the internal types. You could remove it.

```suggestion
```

pkg/apis/core/v1alpha1/types_cloudprofile.go
```diff
@@ -84,6 +84,16 @@ type CloudProfileSpec struct {
 	VolumeTypes []VolumeType `json:"volumeTypes,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,9,rep,name=volumeTypes"`
 }
 
+// SeedSelector contains constraints for selecting seed to be usable for shoots using a profile
+type SeedSelector struct {
+	// LabelSelector is optional and can be used to select seeds by their label settings
+	// +optional
+	*metav1.LabelSelector `json:",inline,omitempty" protobuf:"bytes,1,opt,name=labelSelector"`
+	// Providers is optional and can be used by restricting seeds by their provider type. '*' can be used to enable seeds regardless of their provider type.
```
ialidzhikov, line 94: Why the doc string of the v1alpha1 is quite different from the doc string of the internal type. The internal doc string does not specify anything about the `*` special case.

pkg/scheduler/apis/config/v1alpha1/types.go
```diff
@@ -19,13 +19,15 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	componentbaseconfigv1alpha1 "k8s.io/component-base/config/v1alpha1"
+
+	"github.com/gardener/gardener/pkg/scheduler/apis/config"
 )
 
 const (
 	// SameRegion Strategy determines a seed candidate for a shoot only if the cloud profile and region are identical
-	SameRegion CandidateDeterminationStrategy = "SameRegion"
+	SameRegion CandidateDeterminationStrategy = CandidateDeterminationStrategy(config.SameRegion)
```
ialidzhikov: Hmm, why it is done like this? I think we just duplicate the consts for echa apiversion - see for example `pkg/apis/core/v1alpha1/constants/types_constants.go` and `pkg/apis/core/v1beta1/constants/types_constants.go`.

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -306,51 +354,83 @@ func determineCandidatesOfSameProvider(seedList []*gardencorev1beta1.Seed, shoot
...
+// some usual orientation keywords. It is based on the levenshtein distance
+// of the regions base names plus the difference based on the orientation.
+// regions with the same base but different orientations are basically nearer
+// to each other than two completely unrelated regions.
+func distance(seed, shoot string) int {
+	d, _ := _distance(seed, shoot)
+	return d
+}
+
+func _distance(seed, shoot string) (int, int) {
```
ialidzhikov: There is no convention in this project to prefix private funcs with underscore.

ialidzhikov (changes requested):

pkg/scheduler/controller/shoot/distance_test.go
```diff
@@ -0,0 +1,90 @@
...
+			Expect(orient).To(Equal("west"))
+		})
+		It("handles orientation in the middle", func() {
+			base, orient := orientation("middle-west-europe")
+			Expect(base).To(Equal("middle-:-europe"))
+			Expect(orient).To(Equal("west"))
+		})
+	})
+
+	Context("orientation", func() {
```
ialidzhikov: There is already a context with the same name above. Probably you wanted to name it `distance`?

```suggestion
	Context("distance", func() {
```

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -291,11 +327,23 @@ func getSeedWithLeastShootsDeployed(seedList []*gardencorev1beta1.Seed, shootLis
 	return bestCandidate, nil
 }
 
+func matchProvider(seed, shoot string, providers []string) bool {
```
ialidzhikov: You could improve the naming

```suggestion
func matchProvider(seedProviderType, shootProviderType string, possibleProviders []string) bool {
```

pkg/apis/core/types_cloudprofile.go
```diff
@@ -73,6 +73,18 @@ func (c *CloudProfile) GetProviderType() string {
 	return c.Spec.Type
 }
 
+// SeedSelector contains constraints for selecting seed to be usable for shoots using a profile
+type SeedSelector struct {
+	// LabelSelector is optional and can be used to select seeds by their label settings
+	// +optional
+	*metav1.LabelSelector
+	// ProviderTypes contains a list of allowed provider types used by the Gardener scheduler to restricting seeds by
```
ialidzhikov, line 82: Could you please update `docs/concepts/scheduler.md` as it currently states as limitation
```
- Currently, shoots can only scheduled to seeds from the same cloud provider (`spec.provider.type`), however that is not a technical limitation and might be changed in the future.
```

which is something that this PR aims to improve.

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -306,51 +354,83 @@ func determineCandidatesOfSameProvider(seedList []*gardencorev1beta1.Seed, shoot
...
+// of the regions base names plus the difference based on the orientation.
+// regions with the same base but different orientations are basically nearer
+// to each other than two completely unrelated regions.
+func distance(seed, shoot string) int {
+	d, _ := _distance(seed, shoot)
+	return d
+}
+
+func _distance(seed, shoot string) (int, int) {
+	seed_base, seed_orient := orientation(seed)
```
ialidzhikov:
```suggestion
	seedBase, seedOrientation := orientation(seed)
```

Ref https://golang.org/doc/effective_go.html#mixed-caps

pkg/scheduler/apis/config/v1alpha1/types.go
mandelsoft: Basically I'm convinced that it is always better to avoid duplicating constant literals that must be identical for a correct behaviour of a program. But if you want I can change this.

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: for testing the internal values are required, but I introduced an `end user`function that has only one return value

.ci/generate
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
rfranzke: What is this script? Why do we need it? Isn't it the same like `make generate`?

docs/concepts/scheduler.md
```diff
@@ -31,23 +31,27 @@ However, the Gardener Scheduler on the other hand does not need a TLS configurat
...
-The scheduler takes into consideration the region names of all available seeds in the cluster of the desired infrastructure and picks the regions that match lexicographically the best (starting from the left letter to right letter of the region name).
-E.g. if the shoots wants a cluster in AWS eu-north-1, the scheduler picks all Seeds in region AWS eu-central-1, because at least the continent “eu-“ matches (even better with region instances like AWS ap-southeast-1 and AWS ap-southeast-2).
+2. *Minimal Distance strategy* \
+The Gardener Scheduler tries to find a valid seed with minimal distance to the shoot's intended region.
+The distance is calculated based on the Levenshtein distance of the region. Therefore the region name
+is split into a base name and an orientation. Possible orientations are `north`, `south`, `east`, `west` and `central`.
+The distance then is twice the Levenshtein distance of the region's base name plus a correction value based on the
+orientation and the provider.\
+If the orientations of shoot and seed candidate match, the correction value is 0, if they differ it is 2 and if
+either the seed's or the shoot`s region does not have an orientation it is 1.
```
rfranzke:
```suggestion
either the seed's or the shoot's region does not have an orientation it is 1.
```

rfranzke (commented):
Only minor comments..

docs/concepts/scheduler.md
```diff
@@ -76,11 +80,13 @@ It allows the user to provide a label selector that must match the labels of `Se
 The labels on `Seed`s are usually controlled by Gardener administrators/operators - end users cannot add arbitrary labels themselves.
 If provided, the Gardener Scheduler will only consider those seeds as "suitable" whose labels match those provided in the `.spec.seedSelector` of the `Shoot`.
 
+By default only seeds with the same provider than the shoot are selected. By adding a `providerTypes` field to the `seedSelector`
+a dedicated  set of possible providers (`*` means all provider types) can be selected.
```
timuthy:
```suggestion
a dedicated set of possible providers (`*` means all provider types) can be selected.
```

.ci/generate
mandelsoft: The script is intended to be used as step command for an optional `generate` concourse pipeline. It can be added to the
pipeline definition in forked repository to be able to update a branch with the uptodate generated files.

In my dev environment the generation of the controller-registration created something that does not match the version created by the verify step in central builds. For such cases it would be helpful to be able to execute and update the generation in the central build, also.

timebertt: Which version of helm are you using? helm 3 generates a different controller-registration than helm 2.
You can run `make install-requirements` to install the same version used in the pipeline, so you won't run into any verification issues.

Also, I don't think it's a good idea to run `hack/generate.sh` in the pipeline (especially not with `./...` which includes `vendor/...`), as `hack/check-generate.sh` will then not be able to detect, if generated files are not up-to-date and generated changes are simply ignored, which in turn might lead to undetected bugs in the committed code.
I see, you only want to execute it in your forked repository's pipeline. But then, I don't get, why we would need to add it here in g/g.

pkg/apis/core/v1beta1/types_cloudprofile.go
```diff
@@ -84,6 +86,16 @@ type CloudProfileSpec struct {
...
 }
 
+// SeedSelector contains constraints for selecting seed to be usable for shoots using a profile
+type SeedSelector struct {
+	// LabelSelector is optional and can be used to select seeds by their label settings
+	// +optional
+	*metav1.LabelSelector `json:",inline,omitempty" protobuf:"bytes,1,opt,name=labelSelector"`
+	// Providers is optional and can be used by restricting seeds by their provider type. '*' can be used to enable seeds regardless of their provider type.
+	// +optional
+	ProviderTypes []string `json:"providerTypes,omitempty" protobuf:"bytes,2,rep,name=providerTypes"`
```
timuthy, line 96: What do you think about defaulting this to spec.type if no provider is set? Imo, this is what people usually want to do.

timuthy (commented):
Sorry, I have to address my comment again since it hasn't been answered: https://github.com/gardener/gardener/pull/2169#discussion_r411433926

.ci/generate
mandelsoft: It's not a problem of `make install-requirements`. I've called it multiple times, it does not make a difference.

The vendor folder is ignored for go package pats. I always call it this way for my projects. But Rafael already proposed just to call `make generate`. I've changed this.

ialidzhikov: > I see, you only want to execute it in your forked repository's pipeline. But then, I don't get, why we would need to add it here in g/g.

+1 for removing it.

docs/concepts/scheduler.md
```diff
@@ -31,23 +31,27 @@ However, the Gardener Scheduler on the other hand does not need a TLS configurat
...
 If it cannot find a suitable Seed, it adds an event to the shoot stating, that it is unschedulable.
 
-### 2. Minimal Distance strategy
-
-As a first step, the `SameRegion` strategy is being executed.
-If no seed in the same region could be found, the scheduler uses a lexicographical approach to determine a suitable seed cluster.
-This leverages the fact that most cloud providers (except from Azure) use geographically aligned region names.
-The scheduler takes into consideration the region names of all available seeds in the cluster of the desired infrastructure and picks the regions that match lexicographically the best (starting from the left letter to right letter of the region name).
-E.g. if the shoots wants a cluster in AWS eu-north-1, the scheduler picks all Seeds in region AWS eu-central-1, because at least the continent “eu-“ matches (even better with region instances like AWS ap-southeast-1 and AWS ap-southeast-2).
+2. *Minimal Distance strategy* \
```
ialidzhikov: Nit: I would vote for empty lines instead of `\`. You can use indentation of 3 spaces if you want the text to be aligned according to the list item.

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -306,51 +354,83 @@ func determineCandidatesOfSameProvider(seedList []*gardencorev1beta1.Seed, shoot
...
+		// replace
+		if dist < minDistance {
+			minDistance = dist
+			candidates = []*gardencorev1beta1.Seed{seed}
 		}
 	}
 	return candidates
 }
 
+var orientations = []string{"north", "south", "east", "west", "central"}
```
ialidzhikov: Nit: I would vote to move the funcs that calculate distance between two regions into a separate file in this pkg (`distance.go`) as it is kind of utility functionality that can be separated. I think I would be cleaner as this file already has the core scheduling logic and is already quite long.

pkg/scheduler/apis/config/v1alpha1/types.go
ialidzhikov: I would vote for reverting this as end-user might want to import `pkg/scheduler/apis/config/v1alpha1` pkg and with similar reference, he would get transitively also the internal package - `pkg/scheduler/apis/config`. As end user you would want to get under your `vendor/` only the external (versioned) package.

ialidzhikov (commented):
Thank you for addressing the previous comments. Minor nits now.

.ci/generate
timebertt: I am not completely against adding the file here, but I think, there might be better options.
For example, you might look into GitHub Actions (which is quite easy to setup) to run generation jobs on your fork. There you have every freedom of configuring jobs and run steps as you wish, so you/we don't have to add a `.ci` script. Bonus: you can even let Actions directly commit to your fork very easily, if you want 😉

pkg/scheduler/controller/shoot/scheduler_control.go
mandelsoft: moved to distance.do

mandelsoft:
@ialidzhikov, I've removed the reference

.ci/generate
mandelsoft: I've removed the file again.

rfranzke (approved):
/lgtm

.ci/generate
timebertt: Hmm, it's still here 🤔

timuthy (approved):

ialidzhikov:
@mandelsoft , could you please squash to single commit? Then `lgtm` from my side.

mandelsoft:
@ialidzhikov, because scripts are always taken from the content branch, there is no way to get them from the meta/ci branch, which would be repository local

ialidzhikov:
> @ialidzhikov, because scripts are always taken from the content branch, there is no way to get them from the meta/ci branch, which would be repository local

Hmm, how is that preventing squashing to single commit? 🤔

.ci/generate
mandelsoft: Hmm, I don't see it anymore

ialidzhikov (approved):
Nevermind, I hope we can do `squash and merge` while merging.

/lgtm

mandelsoft:
I hadn't squashed it yet, because of github PR comments in code.
Just use squash-and merge

PR #2170, opened 2020-04-09
[GEP-11] API Server Proxy for Inverting Seed-to-Shoot Connectivity

zanetworker:
This PR is a proposal for utilizing API server proxy to invert the Seed-to-shoot connection. 

cc @mvladev 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/docs/proposals/11-apiserver-network-proxy.md
+++ b/docs/proposals/11-apiserver-network-proxy.md
@@ -1,0 +1,159 @@
+# Utilize API Server Network Proxy to Invert Seed-to-Shoot Connectivity
+
+- [Utilize API Server Network Proxy to Invert Seed-to-Shoot Connectivity](#utilize-api-server-network-proxy-to-invert-seed-to-shoot-connectivity)
+  - [Problem](#problem)
+  - [Proposal](#proposal)
+    - [API Server Network Proxy](#api-server-network-proxy)
+  - [Challenges](#challenges)
+    - [Prometheus to Shoot connectivity](#prometheus-to-shoot-connectivity)
+      - [Possible Solutions](#possible-solutions)
+      - [Port-forwarder Sidecar](#port-forwarder-sidecar)
+      - [Proxy Client Sidecar](#proxy-client-sidecar)
+      - [Proxy sub-resource](#proxy-sub-resource)
+    - [Proxy-server Loadbalancer Sharing and Re-advertising](#proxy-server-loadbalancer-sharing-and-re-advertising)
+      - [Possible Solution](#possible-solution)
+    - [Summary](#summary)
+
+## Problem
+
+Gardener's architecture for Kubernetes clusters relies on having the control-plane (e.g., kube-apiserver, kube-scheduler, kube-controller-manager, etc.) and the data-plane  (e.g., kube-proxy, kubelet, etc.) of the cluster residing in separate places, this provides many benefits but poses some challenges, especially when API-server to system components communication is required. This problem is solved today in Gardener by [making use of OpenVPN](https://github.com/gardener/vpn) to establish a VPN connection from the seed to the shoot. To do so, the following steps are required:
+
+- Create a Loadbalancer service on the shoot.
+- Add a sidecar to the API server pod which knows the address of the newly created Loadbalancer.
+- Establish a connection over the internet to the VPN Loadbalancer
+- Install additional iptables rules that would redirect all the IPs of the shoot (i.e., service, pod, node CIDRs) to the established VPN tunnel
+
+
+There are however quite a few problems with the above approach, here are some:
+
+- Every shoot would require an additional loadbalancer, this accounts for addition overhead in terms of both costs and troubleshooting efforts.
+- Private access use-cases would not be possible without having a seed residing in the same private domain as a hard requirement. For example, have a look at [this issue](https://github.com/gardener/gardener-extension-provider-gcp/issues/56)
+- Providing a public endpoint to access components in the shoot poses a security risk.
+
+
+## Proposal
+
+There are mutliple ways to tackle the directional connectivity issue mentioned above, one way would be to invert the connection between the API server and the system components, i.e., instead of having the API server side-car establish a tunnel, we would have an agent residing in the shoot cluster initiate the connection itself. This way we don't need a Loadbalancer for every shoot and from the security perspective, there is no ingress from outside, only controlled egress.
+
+We want to replace this:
+
+`APIServer | VPN-seed ---> internet ---> LB --> VPN-Shoot (4314) --> Pods | Nodes | Services`
+
+With this:
+
+`APIServer <-> Proxy-Server <--- internet <--- Proxy-Agent --> Pods | Nodes | Services`
+
+
+### API Server Network Proxy
+
+To solve this issue we can utilize the [apiserver-network-proxy](https://github.com/kubernetes-sigs/apiserver-network-proxy) upstream implementation. Which provides a reference implementation for a reverse streaming server. The way it works is as follows:
+
+- Proxy agent connects to proxy server to establish a sticky connection.
+- Traffic to the proxy server (residing in the seed) gets then re-directed to the agent (residing in the shoot) which forwards the traffic to in-cluster components.
+
+The initial motivation for the apiserver-network-proxy project is to get rid of provider-specific implementations that reside in the API-server (e.g., SSH), but it turns out that
+it has other interesting use-cases such as data-plane connection decoupling, which is the main use-case for this proposal.
+
+Starting with **Kubernetes 1.18** it's possible to make use of an `--egress-selector-config-file` flag, this helps point the API-server to traffic hook points based on traffic direction. For example, in the config below the API server would have to forward all cluster related traffic (e.g., logs, port-forward, exec, ...etc.) to the **proxy-server** which then knows how to forward traffic to the shoot. For the rest of the traffic, e.g. API server to ETCD or other control-plane components `direct` is used which means legacy routing method, i.e., by-pass the proxy.
+```yaml
+  egress-selector-configuration.yaml: |-
+    apiVersion: apiserver.k8s.io/v1alpha1
+    kind: EgressSelectorConfiguration
+    egressSelections:
+    - name: cluster
+      connection:
+        proxyProtocol: httpConnect
+        transport:
+          tcp:
+            url: https://proxy-server:8131
+    - name: master
+      connection:
+        proxyProtocol: direct
+    - name: etcd
+      connection:
+        proxyProtocol: direct
+```
+
+## Challenges
+
+
+### Prometheus to Shoot connectivity
+One challenge remains to completely eliminate the need for a VPN connection. In today's Gardener setup, each control-plane has a Prometheus instance that directly scrapes cluster components such as CoreDNS, Kubelets, cadvisor, etc. This works because in addition to the VPN side car attached to the API server pod, we have another one attached to prometheus which knows how to forward traffic to these endpoints. Once the VPN is eliminated, it is required to find other means to forward traffic to these components.
+
+#### Possible Solutions
+
+There are currently two ways to solve this problem:
+
+- Attach a port-forwarder side-car to prometheus.
+- Utilize the proxy subresource on the API server.
+
+#### Port-forwarder Sidecar
+
+With this solution each prometheus instance would have a side-car that has the kubeconfig of the shoot cluster, and which establishes a port-forward connection to the endpoints residing in the shoot.
+
+There are a many problems with this approach:
+
+- the port-forward connection is not reliable.
+- the connection would break if the API server instance dies.
+- requires an additional component.
+- would need to expose every pod / service via port-forward.
+
+
+```console
+Prom Pod (Prometheus -> Port-forwarder) <-> APIServer -> Proxy-server <--- internet <--- Proxy-Agent --> Pods | Nodes | Services
+```
+
+#### Proxy Client Sidecar
+
+Another solution would be to implement a proxy-client as a sidecar for every component that wishes to communicate with the shoot cluster. For this to work, means to re-direct / inject that proxy to handle the component's traffic is necessary (e.g., additional IPtable rules).
+
+```console
+Prometheus Pod (Prometheus -> Proxy) <-> Proxy-Server <--- internet <--- Proxy-Agent --> Pods | Nodes | Services
+```
+
+The problem with this approach is that it requires an additional sidecar (along with traffic redirection) to be attached to every client that wishes to communicate with the shoot cluster, this can cause:
+
+- additional maintenance efforts (extra code).
+- other side-effects (e.g., if istio sidecar injection is enabled)
+
+#### Proxy sub-resource
+
+Kubernetes supports proxying requests to nodes, services, and pod endpoints in the shoot cluster. This proxy connection can be utilized for scraping the necessary endpoints in the shoot.
+
+This approach requires less components and is more reliable than the port-forward solution, however, it relies on having the API server supporting proxied connection for the required endpoints.
+
+```console
+Prometheus  <-> APIServer <-> Proxy-Server <--- internet <--- Proxy-Agent --> Pods | Nodes | Services
+```
+
+As simple as it is, it has a downside that it relies on the availability of the API server.
+
+### Proxy-server Loadbalancer Sharing and Re-advertising
+
+With the proxy-server in place, we need to provide means to enable the proxy-agent in the shoot to establish the connection with the server. As a result, we need to provide a public endpoint through which this channel of communication can be established, i.e., we need a Loadbalancer(s).
+
+
+#### Possible Solution
+
+Using a Loadbalancer / proxy server would not make sense since this is a pain-point we are trying to eliminate in the first-place, doing so just moves the costs to the control-plane. A possible solution is to communicate over a shared loadbalancer in the seed, similar to what has been proposed [here](https://github.com/gardener/gardener/blob/master/docs/proposals/08-shoot-apiserver-via-sni.md), this way we can prevent the extra-costs for load-balancers.
+
+With this in mind, we still have other pain-points, namely:
+
+- Advertising Loadbalancer public IPs to the shoot.
+- Directing the traffic to the corresponding shoot proxy-server.
+
+For advertising the Loadbalancer IP, a DNS entry can be created for the proxy loadbalancer (or re-use the DNS entry for the SNI proxy), along with necessary certificates, which is then used to connect to the loadbalancer. At this point we can decide on either one of the two approaches:
+
+1. One Proxy / API server with a shared loadbalancer.
+2. Use one proxy server for all agents.
+
+In the first case, we will probably need a proxy for the proxy-server that knows how to direct traffic to the correct proxy server based on the corresponding shoot cluster. In the second case, we don't need another proxy if the proxy server is cluster-aware, i.e., can pool and identify connections coming from the same cluster and peer them with the correct API. Unfortunately, the second case is not supported today.
+
+### Summary
+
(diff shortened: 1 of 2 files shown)


docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,108 @@
...
+    - [Challenges](#challenges)
+    - [Solutions](#solutions)
+      - [Port-forwarder Sidecar](#port-forwarder-sidecar)
+      - [Proxy sub-resource](#proxy-sub-resource)
+    - [Summary](#summary)
+
+
+## Problem
+
+Gardener's productive setup for Kubernetes clusters relies on having the control-plane (e.g., kube-apiserver, kube-scheduler, kube-controller-manager, etc.) and the data-plane  (e.g., kube-proxy, kubelet, etc.) of the cluster residing in separate places, this provides many benefits but poses some challenges especially when API-server to system components communication is required. This problem is solved today in Gardener by [making use of OpenVPN](https://github.com/gardener/vpn) to establish a VPN connection from the seed to the shoot. To do so, the following steps are required:
```
vlerenc: Isn't "network domains" the better term for "places"?

vlerenc: ", this provides..." starts a new sentence I believe.

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,108 @@
...
+
+- Create a Loadbalancer service on the shoot.
+- Add a side-care to the API server pod which knows the address of the newly created Loadbalancer.
+- Establish a connection over the internet to the VPN Loadbalancer
+- Install additional iptables rules that would redirect all the IPs of the shoot (i.e., service, pod, node CIDRs) to the established VPN tunnel
+
+
+There are however quite a few problems with the above approach, here are some:
```
vlerenc, line 27: Thank you for considering these! :-)

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,108 @@
...
+
+### API Server Network Proxy
+
+To solve this issue we can utilize the [apiserver-network-proxy](apiserver-network-proxy) upstream implementation. Which provides a reference implementation for a reverse streaming server. The way it works is as follows:
+
+- Proxy agent connects to proxy server to establish a sticky connection.
+- Traffic to the proxy server gets then re-directed to the agent which forwards the traffic to in-cluster components.
+
+The initial motivation for the apiserver-network-proxy project is to get rid of provider-specific implementations that reside in the API-server (e.g., SSH), but it turns out that
+it has other interesting use-cases such as data-plane connection decoupling, which is the main use-case for this proposal.
```
vlerenc, line 55: Awesome.

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,108 @@
...
+      connection:
+        proxyProtocol: direct
+    - name: etcd
+      connection:
+        proxyProtocol: direct
+```
+
+### Challenges
+
+One challenge remains to completely eliminate the need for a VPN connection. In today's Gardener setup, each control-plane has a Prometheus instance that directly scrapes cluster components such as CoreDNS, Kubelets, cadvisor, etc. This works because in addition to the VPN side car attached to the API server pod, we have another one attached to prometheus which knows how to forward traffic to these endpoints. Once the VPN is eliminated, it is required to find other means to forward traffic to these components.
```
vlerenc: cc @wyb1

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,108 @@
...
+- Attach a port-forwarder side-car to prometheus.
+- Utilize the proxy subresource on the API server.
+
+#### Port-forwarder Sidecar
+
+With this solution each prometheus instance would have a side-car that has the kubeconfig of the shoot cluster, and which estalishes a port-forward connection to the endpoints residing in the shoot.
+
+There are a couple of problems with this approach:
+
+- the port-forward connection is not reliable.
```
vlerenc, line 96: How often would a connection be established? It's always one IP/port at a time to scrape one component, so aren't these rather short-lived queries?

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,108 @@
...
+- the connection would break if the API server instance dies.
```
vlerenc, line 97: OK, however, as far as I have understood our control plane auto-scaling, which is the main source for dying API servers, that's handled in most cases gracefully (OOM isn't of course, but shouldn't happen often).

But of course you are right. We don't want/need false positive alerts. Weighting it, I love getting rid of the VPN LB for the reasons stated above and we could maybe tune the monitoring alerts (wait longer/for more missed probes, which should reduce the likelihood of false positives, right?).

vlerenc (commented):
Thank you for the GEP and description. This sounds really, really good. Do you have an effort estimation as well?

docs/proposals/11-apiserver-network-proxy.md
rfranzke:
```suggestion
Gardener's architecture for Kubernetes clusters relies on having the control-plane (e.g., kube-apiserver, kube-scheduler, kube-controller-manager, etc.) and the data-plane  (e.g., kube-proxy, kubelet, etc.) of the cluster residing in separate places, this provides many benefits but poses some challenges, especially when API-server to system components communication is required. This problem is solved today in Gardener by [making use of OpenVPN](https://github.com/gardener/vpn) to establish a VPN connection from the seed to the shoot. To do so, the following steps are required:
```

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,108 @@
...
+      - [Proxy sub-resource](#proxy-sub-resource)
+    - [Summary](#summary)
+
+
+## Problem
+
+Gardener's productive setup for Kubernetes clusters relies on having the control-plane (e.g., kube-apiserver, kube-scheduler, kube-controller-manager, etc.) and the data-plane  (e.g., kube-proxy, kubelet, etc.) of the cluster residing in separate places, this provides many benefits but poses some challenges especially when API-server to system components communication is required. This problem is solved today in Gardener by [making use of OpenVPN](https://github.com/gardener/vpn) to establish a VPN connection from the seed to the shoot. To do so, the following steps are required:
+
+- Create a Loadbalancer service on the shoot.
+- Add a side-care to the API server pod which knows the address of the newly created Loadbalancer.
```
rfranzke:
```suggestion
- Add a sidecar to the API server pod which knows the address of the newly created Loadbalancer.
```

Maybe also mention Prometheus here?

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,108 @@
...
+
+- Proxy agent connects to proxy server to establish a sticky connection.
+- Traffic to the proxy server gets then re-directed to the agent which forwards the traffic to in-cluster components.
+
+The initial motivation for the apiserver-network-proxy project is to get rid of provider-specific implementations that reside in the API-server (e.g., SSH), but it turns out that
+it has other interesting use-cases such as data-plane connection decoupling, which is the main use-case for this proposal.
+
+Starting **Kubernetes 1.18** its possible to make use of an ```--egress-selector-config-file``` flag, this helps point the API-server to traffic hook points based on traffic direction, for example, in the config below, the API server would have to forward all cluster related traffic (e.g., logs, port-forward, exec, ...etc.) to the **proxy-server** which then knows how to forward traffic to the shoot. For the rest of the traffic e.g., API server to ETCD or other control-plane components `direct` is used which means legacy routing method, i.e., by-pass the proxy.
```
rfranzke:
```suggestion
Starting with **Kubernetes 1.18** it's possible to make use of an `--egress-selector-config-file` flag, this helps point the API-server to traffic hook points based on traffic direction. For example, in the config below the API server would have to forward all cluster related traffic (e.g., logs, port-forward, exec, ...etc.) to the **proxy-server** which then knows how to forward traffic to the shoot. For the rest of the traffic, e.g. API server to ETCD or other control-plane components `direct` is used which means legacy routing method, i.e., by-pass the proxy.
```

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,108 @@
...
+One challenge remains to completely eliminate the need for a VPN connection. In today's Gardener setup, each control-plane has a Prometheus instance that directly scrapes cluster components such as CoreDNS, Kubelets, cadvisor, etc. This works because in addition to the VPN side car attached to the API server pod, we have another one attached to prometheus which knows how to forward traffic to these endpoints. Once the VPN is eliminated, it is required to find other means to forward traffic to these components.
+
+
+### Solutions
+
+There are currently two ways to solve this problem:
+
+- Attach a port-forwarder side-car to prometheus.
+- Utilize the proxy subresource on the API server.
+
```
rfranzke, line 89: What about simply installing ip routes in the Prometheus network namespace (using a sidecar) that redirects traffic to the shoot networks to the proxy in the seed? Basically, this would be the same solution like for the kube-apiserver with the only difference that it doesn't/can't reuse the upstream network egress configuration.

rfranzke (commented):
Nice proposal, looking forward! :)

docs/proposals/11-apiserver-network-proxy.md
zanetworker: thought about it, but it wouldn't work that easily, for that you would need an additional sidecar that acts as proxy to the proxy that knows how talk HTTP Connect, so basically we have another level of re-direction and another component that we have maintain, besides, it would be nice to get rid of iptables as much as possible :)

docs/proposals/11-apiserver-network-proxy.md
zanetworker: The port-forward connection would need to live for the entire life-time of the shoot even if the API server dies. Ideally, it would be established just once, but if the API-server dies (which acts as our message relayer), the connection would break.

zanetworker:
@vlerenc I already have a PoC that works (maybe I can try to demo that in one of the next meetings), the missing piece is getting prometheus to keep all the endpoints it monitors without changing the implementation that much.

docs/proposals/11-apiserver-network-proxy.md
mvladev: The Port-forward won't work at all because the Prometheus must talk to let's say every `node-exporter` pod in the cluster.

It's designed only for debugging. In our case `proxy` should be used.

zanetworker: well technically it would work, if prometheus uses the pod role and discovers all the pods with a node-exporter label, it would work, or? 

and of course its not practical because if we use direct addresses we would have to port-forward to all node-exporters.

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,107 @@
...
+`APIServer | VPN-seed ---> internet ---> LB --> VPN-Shoot (4314) --> Pods | Nodes | Services`
+
+With this:
+
+`APIServer <-> Proxy-Server <--- internet <--- Proxy-Agent --> Pods | Nodes | Services`
+
+
+### API Server Network Proxy
+
+To solve this issue we can utilize the [apiserver-network-proxy](apiserver-network-proxy) upstream implementation. Which provides a reference implementation for a reverse streaming server. The way it works is as follows:
```
timebertt: Non-functional link

zanetworker:
I will still add some changes to the GEP, will ping once I do.

zanetworker:
Alright, updates are in.

rfranzke:
@mvladev @DockToFuture can you have another look? If there are no further comments/questions/suggestions I suggest to merge this PR this week?

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,134 @@
...
+
+`APIServer <-> Proxy-Server <--- internet <--- Proxy-Agent --> Pods | Nodes | Services`
+
+
+### API Server Network Proxy
+
+To solve this issue we can utilize the [apiserver-network-proxy](https://github.com/kubernetes-sigs/apiserver-network-proxy) upstream implementation. Which provides a reference implementation for a reverse streaming server. The way it works is as follows:
+
+- Proxy agent connects to proxy server to establish a sticky connection.
+- Traffic to the proxy server gets then re-directed to the agent which forwards the traffic to in-cluster components.
```
rfranzke: Can you make more clear where (in seed/shoot) which component is running?

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,134 @@
...
+
+- the port-forward connection is not reliable.
+- the connection would break if the API server instance dies.
+- requires an additional component.
+
+#### Proxy sub-resource
+
+Kubernetes supports proxying requests to nodes, services, and pod endpoints in the shoot cluster. This proxy connection can be utilized for scraping the necessary endpoints in the shoot.
+
+This approach requires less-compenents and is more reliable than the port-forward solution, however, it relies on having the API server supporting proxied connection for the required endpoints.
```
rfranzke: It also comes with additional load on the API server (which we don't have today), right? Is this negligible?
Which solution do you prefer?

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,134 @@
...
+For advertising the Loadbalancer IP, a DNS entry can be created for the proxy loadbalancer, along with necessary certificates, which is then used to connect to the loadbalancer. At this point we can decided on either one of the two approaches:
+
+1. One Proxy / API server with a shared loadbalancer.
+2. Use one proxy server for all agents.
+
+In the first case, we will probably need a proxy for the proxy-server that knows how to direct traffic to the correct proxy server based on the corresponding shoot cluster. In the second case, we don't need another proxy if the proxy server is cluster-aware, i.e., can pool and identify connections coming from the same cluster and peer them with the correct API. Unfortunately, the second case is not supported today.
+
+### Summary
+
+- API server proxy can be utilized to invert the connection.
```
rfranzke:
```suggestion
- API server proxy can be utilized to invert the connection (only for clusters >= 1.18, for older clusters the old VPN solution will remain).
```

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,134 @@
...
+
+- Advertising Loadbalancer public IPs to the shoot.
+- Directing the traffic to the corresponding shoot proxy-server.
+
+For advertising the Loadbalancer IP, a DNS entry can be created for the proxy loadbalancer, along with necessary certificates, which is then used to connect to the loadbalancer. At this point we can decided on either one of the two approaches:
+
+1. One Proxy / API server with a shared loadbalancer.
+2. Use one proxy server for all agents.
+
+In the first case, we will probably need a proxy for the proxy-server that knows how to direct traffic to the correct proxy server based on the corresponding shoot cluster. In the second case, we don't need another proxy if the proxy server is cluster-aware, i.e., can pool and identify connections coming from the same cluster and peer them with the correct API. Unfortunately, the second case is not supported today.
```
mvladev: I would suggest to do a solution very similar to what [API server SNI](https://github.com/gardener/gardener/blob/master/docs/proposals/08-shoot-apiserver-via-sni.md) is trying to achieve:

```
Shoot-A Proxy-Agent -> LB (shared) -> SNI (proxy) -> Shoot-A Proxy-Server
Shoot-B Proxy-Agent -> LB (shared) -> SNI (proxy) -> Shoot-B Proxy-Server
```

When combined with API server SNI, then you'll end with only 1 LB needed for the Seed cluster.
Even better - the API server DNS entry can be reused, but have a different port:

- `Shoot-A Proxy-Agent` is configured to talk to `api.<shoot-a>.<project>.internal.<domain>:9443`
- `Shoot-B Proxy-Agent` is configured to talk to `api.<shoot-b>.<project>.internal.<domain>:9443`

The SNI proxy would know that `api.<shoot-a>.<project>.internal.<domain>:9443` points to `Shoot-A Proxy-Server` and would simply forward the traffic to it.
There won't be a need for any `PROXY protocol` in this case.

docs/proposals/11-apiserver-network-proxy.md
zanetworker: it adds additional load to the API-server yes but I don't think its significant, also the scrape interval from prometheus has been halved  so it shouldn't be much. This solution is the most achievable, using port-forward solution is also putting load on the API but would require knowledge of all services in the shoot cluster that we need to port-forward to, in addition to a port-forwarder sidecar, so I prefer the proxy solution in this case.

docs/proposals/11-apiserver-network-proxy.md
zanetworker: yup thats the first approach, but what you are  saying is even share the loadbalancer of the API-server and  increase the replicas by adding other services if required. I think that should work, we just need to know the loadbalancer limits.

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,134 @@
...
+#### Possible Solutions
+
+There are currently two ways to solve this problem:
+
+- Attach a port-forwarder side-car to prometheus.
+- Utilize the proxy subresource on the API server.
+
+#### Port-forwarder Sidecar
+
+With this solution each prometheus instance would have a side-car that has the kubeconfig of the shoot cluster, and which estalishes a port-forward connection to the endpoints residing in the shoot.
```
plkokanov:
```suggestion
With this solution each prometheus instance would have a side-car that has the kubeconfig of the shoot cluster, and which establishes a port-forward connection to the endpoints residing in the shoot.
```

docs/proposals/11-apiserver-network-proxy.md
```diff
@@ -0,0 +1,134 @@
...
+One challenge remains to completely eliminate the need for a VPN connection. In today's Gardener setup, each control-plane has a Prometheus instance that directly scrapes cluster components such as CoreDNS, Kubelets, cadvisor, etc. This works because in addition to the VPN side car attached to the API server pod, we have another one attached to prometheus which knows how to forward traffic to these endpoints. Once the VPN is eliminated, it is required to find other means to forward traffic to these components.
+
+#### Possible Solutions
+
+There are currently two ways to solve this problem:
+
+- Attach a port-forwarder side-car to prometheus.
+- Utilize the proxy subresource on the API server.
+
+#### Port-forwarder Sidecar
```
mvladev, line 90: I would remove this entire section. The more sane option would be to deploy a very simple sidecar proxy to the Prometheus which is going to have a Dialer which would simply do a HTTP CONNECT call to the Proxy-Server, similar to what [K8S's API server is doing](https://github.com/kubernetes/kubernetes/blob/bac9351c64671ce4d5198d431c97bf1ccd72752f/staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go#L104-L129). This would establish the TCP connection which can then be used to send the actual data of the request.

The setup would be:

```
Prom Pod (Prometheus -> Proxy) <-> Proxy-Server <--- internet <--- Proxy-Agent --> Pods | Nodes | Services
```

mvladev (changes requested):

docs/proposals/11-apiserver-network-proxy.md
zanetworker: yup thats a third option, which I actually tried doing but its more code overhead (i.e., another piece to maintain, extra certificates if the client is a sidecar and not baked in prometheus). I can add it as a third option but I don't think that will be our first go-to approach.

I left the the port-forward section because imo  it makes sense to involve the reader in the decision process.

rfranzke:
If there are no objections then I'd like to merge this PR this week. Any final comments?

vpnachev (approved):
sounds good to me

rfranzke (approved):

mvladev (approved):

PR #2195, opened 2020-04-17
care controller - Health check condition message

danielfoehrKn:

The care controller collects the conditions from extensions in the Shoot (Worker, Network, Extension) and checks if there are unsuccessful health checks reported by the extension. 

This PR enhances the error message  on the HealthCondition of the Shoot. Specifically, the Extension kind, name and namespace are being added. 

This makes it more clear where the actual unhealthy report is coming from. 
image

Instead the message will be: 

"Extension/Worker/..  CRD (namespace/name) reports failing healthcheck : <error mesage from CRD>"

Fixes #

Though Gardener already reports the ExtensionType via the condition.reason field, it was not immediately evident to me, that this is the case. 
image

"ExtensionUnhealthyReport" - I feel like the term  "Extension" is overloaded in the Gardener context -> it was not clear to me that the Extension CRD was meant. Thats why I propose adding it also into the message itself.

**Release note**:

```improvement operator

```


rfranzke (approved):

timebertt (approved):
/lgtm

PR #2198, opened 2020-04-20
Increase limits for metrics-server

rfranzke:
For a cluster with many pods the metrics-server will run out-of-memory. I saw that it required ~550 MB for a cluster with ~12.5k pods.
Let's increase the limits so that it can support such scenarios.

**Release note**:

```improvement user
The CPU and memory limits for the metrics-server have been slightly increased to support large clusters.
```


timebertt (approved):
/lgtm

PR #2200, opened 2020-04-20
respect http proxy settings in test framework utils

MartinWeindel:
If in the context of integration tests the shoot cluster is only reachable via an HTTP proxy, the TestGardenerSuite started with the
test machinery fails even if HTTP_PROXY/HTTPS_PROXY environment variables are set.
```
Get https://api.tm87h-wns.it.shoot.dev.k8s-hana.ondemand.com/api/v1/namespaces/kubernetes-dashboard/services/https:kubernetes-dashboard:/proxy: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```
I assume, the problem is the missing proxy settings in the HTTP client, which is fixed in the standard way with this PR.

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/test/framework/http_utils.go
+++ b/test/framework/http_utils.go
@@ -18,14 +18,20 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"github.com/pkg/errors"
 	"net/http"
 	"time"
+
+	"github.com/pkg/errors"
 )
 
 // HTTPGet performs an HTTP GET request with context
 func HTTPGet(ctx context.Context, url string) (*http.Response, error) {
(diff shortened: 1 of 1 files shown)


schrodit (changes requested):
we should also add the proxy here https://github.com/gardener/gardener/blob/31da4f39f62bc52c62bf481bf18599baee5b67b5/test/framework/http_utils.go#L29
rest lgtm

dguendisch (approved):

rfranzke (approved):

schrodit (approved):
lgtm

PR #2201, opened 2020-04-20
Skip the containerd test for non-ubuntu worker pool

ialidzhikov:
Skip the containerd test for non-ubuntu worker pool.

Fixes #

**Release note**:

```improvement developer
The `containerd` test is now skipped for worker pools that are not using the `ubuntu` operating system.
```

### DIFF
--- a/test/integration/shoots/operations/containerruntime.go
+++ b/test/integration/shoots/operations/containerruntime.go
@@ -35,10 +35,17 @@ var _ = ginkgo.Describe("Shoot container runtime testing", func() {
 		shoot := f.Shoot
 
 		if len(shoot.Spec.Provider.Workers) == 0 {
-			ginkgo.Skip("at least one worker pool is required in the test shoot.")
+			ginkgo.Skip("at least one worker pool is required in the test shoot")
 		}
 
-		containerdWorker := shoot.Spec.Provider.Workers[0].DeepCopy()
+		worker := shoot.Spec.Provider.Workers[0]
+		// containerD is supported only with Ubuntu OS for now.
+		// TODO: adapt/remove this when containerD is available on other OS.
+		if worker.Machine.Image.Name != "ubuntu" {
+			ginkgo.Skip("worker with machine image 'ubuntu' is required")
+		}
+
+		containerdWorker := worker.DeepCopy()
 
 		allowedCharacters := "0123456789abcdefghijklmnopqrstuvwxyz"
 		id, err := gardenerutils.GenerateRandomStringFromCharset(3, allowedCharacters)
@@ -60,11 +67,11 @@ var _ = ginkgo.Describe("Shoot container runtime testing", func() {
 			ginkgo.By("removing containerd worker pool after test execution")
 			err := f.UpdateShoot(ctx, func(s *gardencorev1beta1.Shoot) error {
 				var workers []gardencorev1beta1.Worker
-				for _, worker := range s.Spec.Provider.Workers {
-					if worker.Name == workerPoolName {
+				for _, current := range s.Spec.Provider.Workers {
+					if current.Name == workerPoolName {
 						continue
 					}
-					workers = append(workers, worker)
+					workers = append(workers, current)
 				}
 				s.Spec.Provider.Workers = workers
 				return nil
--- a/example/05-project-dev.yaml
+++ b/example/05-project-dev.yaml
@@ -27,6 +27,6 @@ spec:
 # purpose: "Experimenting with Gardener"
   # The `spec.namespace` field is optional and will be initialized if unset - the resulting
   # namespace will be generated and look like "garden-dev-<random-chars>", e.g. "garden-dev-5z43z".
-  # If the namespace is set then the namespace must be labelled with `garden.sapcloud.io/role: project`
-  # and `project.garden.sapcloud.io/name: <project-name>` (<project-name>=dev in this case).
+  # If the namespace is set then the namespace must be labelled with `gardener.cloud/role: project`
+  # and `project.gardener.cloud/name: <project-name>` (<project-name>=dev in this case).
   namespace: garden-dev
--- a/test/framework/test_description.go
+++ b/test/framework/test_description.go
@@ -1,11 +1,26 @@
+// Copyright 2020 Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package framework
 
 import (
 	"context"
 	"fmt"
+	"time"
+
 	"github.com/onsi/ginkgo"
 	"k8s.io/apimachinery/pkg/util/sets"
-	"time"
 )
 
 // TestDescription labels tests according to the provided labels in the expected order.
--- a/test/framework/utils.go
+++ b/test/framework/utils.go
@@ -16,17 +16,18 @@ package framework
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"reflect"
+	"regexp"
+
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
 	"github.com/hashicorp/go-multierror"
 	"github.com/onsi/ginkgo"
 	"github.com/pkg/errors"
-	"io/ioutil"
 	apimachineryRuntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
-	"os"
-	"reflect"
-	"regexp"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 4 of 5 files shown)


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.1.5
 	github.com/coreos/etcd v3.3.17+incompatible
 	github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f
-	github.com/gardener/etcd-druid v0.1.12
+	github.com/gardener/etcd-druid v0.1.15
```
danielfoehrKn: Can we have these dependency updates in another PR?

test/integration/shoots/operations/containerruntime.go
```diff
@@ -54,6 +54,9 @@ var _ = ginkgo.Describe("Shoot container runtime testing", func() {
 			Name:              extensionsv1alpha1.CRINameContainerD,
 			ContainerRuntimes: nil,
 		}
+		// containerD is supported only with Ubuntu OS for now.
+		// TODO: adapt/remove this when containerD is available on other OS.
+		containerdWorker.Machine.Image.Name = "ubuntu"
```
danielfoehrKn: I think we also have to set the machine.image.versions to nil as we are copying the existing worker pool. This way we can take advantage of the machine image version defaulting.

danielfoehrKn (changes requested):

go.mod
ialidzhikov: I don't see anything wrong but..

test/integration/shoots/operations/containerruntime.go
ialidzhikov: Nice catch! I force-pushed.

go.sum
```diff
@@ -112,6 +112,7 @@ github.com/docopt/docopt-go v0.0.0-20180111231733-ee0de3bc6815/go.mod h1:WwZ+bS3
 github.com/dsnet/compress v0.0.1 h1:PlZu0n3Tuv04TzpfPbrnI0HW/YwodEXDS+oPKahKF0Q=
 github.com/dsnet/compress v0.0.1/go.mod h1:Aw8dCMJ7RioblQeTqt88akK31OvO8Dhf5JflhBbQEHo=
 github.com/dsnet/golib v0.0.0-20171103203638-1ea166775780/go.mod h1:Lj+Z9rebOhdfkVLjJ8T6VcRQv3SXugXy999NBtR9aFY=
+github.com/elazarl/goproxy v0.0.0-20170405201442-c4fc26588b6e h1:p1yVGRW3nmb85p1Sh1ZJSDm4A4iKLS5QNbvUHMgGu/M=
```
danielfoehrKn: where is that coming from? Otherwise lgtm

danielfoehrKn (dismissed):
Thanks!

test/integration/shoots/operations/containerruntime.go
```diff
@@ -51,6 +51,11 @@ var _ = ginkgo.Describe("Shoot container runtime testing", func() {
 			Name:              extensionsv1alpha1.CRINameContainerD,
 			ContainerRuntimes: nil,
 		}
+		// containerD is supported only with Ubuntu OS for now.
+		// TODO: adapt/remove this when containerD is available on other OS.
+		containerdWorker.Machine.Image = &gardencorev1beta1.ShootMachineImage{
```
timebertt: Hmm, I'm not sure about that, but I think it would be better to instead skip the test if the original worker pool (the one you copy). That way, we would prevent running the test multiple times for coreos, suse and so on, but with the same stuff we test.
But I am not sure, if we actually run the complete shoot suite for every operating system. @schrodit @dguendisch Can you clarify, please?

ialidzhikov: Generally the right approach would be to skip/fail the test when the passed OS is not the required one - kind of precondition for the test.
As far as I understood, currently there is no mechanism from tm side to restrict test to given OS, that why I am proposing this workaround.

danielfoehrKn: I think running the integration tests multiple test with different OS is an implementation detail on the TestMachinery side? Lets see what @schrodit @dguendisch say

timebertt: Yes, agreed, that's kind of what I am suggesting. I would propose to check the image name of the original worker pool and if it isn't `ubuntu` call `ginkgo.Skip()`. WDYT?

schrodit: Generally both cases are valid.
I would rather go with tim's proposal as otherwise we would run the test for all non ubuntu tests.
Because it would mean that we test containerd on ubuntu even the test flavor specifies that we would want to test only coreos or suse.

dguendisch: > if we actually run the complete shoot suite for every operating system.

Yes we do. And as you've written, we currently cannot skip certain tests from TM side for different OSes. Hence I like @tim-ebert 's suggestion to check the original worker pool and and only add the containerd one when there's a chance of success and skip it otherwise.

danielfoehrKn: @ialidzhikov  WDYT?

ialidzhikov: It makes sense. I adapted the PR to skip the test for non-ubuntu worker pool.

schrodit (approved):
lgtm

timebertt (approved):
Nice, thanks @ialidzhikov!
/approve

test/integration/shoots/operations/containerruntime.go
timebertt: Thanks for clarifying :)

test/integration/shoots/operations/containerruntime.go
```diff
@@ -35,10 +35,17 @@ var _ = ginkgo.Describe("Shoot container runtime testing", func() {
...
 		if len(shoot.Spec.Provider.Workers) == 0 {
-			ginkgo.Skip("at least one worker pool is required in the test shoot.")
+			ginkgo.Skip("at least one worker pool is required in the test shoot")
 		}
 
-		containerdWorker := shoot.Spec.Provider.Workers[0].DeepCopy()
+		worker := shoot.Spec.Provider.Workers[0]
+		// containerD is supported only with Ubuntu OS for now.
+		// TODO: adapt/remove this when containerD is available on other OS.
+		if worker.Machine.Image.Name != "ubuntu" {
```
rfranzke, line 44: Hm, @ialidzhikov I missed this and I don't like that this is hard-coded here. Does it make sense to make this configurable via command line flag which machine images are supported? Can you do that?

timebertt: Hmm, we also do it like that here:
https://github.com/gardener/gardener/blob/29da00ee28e332fb7348a55ebfbe3de1c1f49988/test/integration/shoots/operations/clusterautoscaler.go#L178-L180

How would this command line flag look like? Who would configure it / where will it be configured which tests should be enabled/skipped for which provider/os/...?
As @ialidzhikov mentioned:
> currently there is no mechanism from tm side to restrict test to given OS

I think, if we have such mechanism, this test would be a good fit for that.
@schrodit WDYT how to proceed on that?

rfranzke: > Who would configure it / where will it be configured which tests should be enabled/skipped for which provider/os/...?

Somebody has to trigger the tests. Here those flags shall be configured.

schrodit: I think we also have some aliclouds specifics.
So we have that problem on multiple tests.

I think we should have such mechanism that skip tests for specific providers or only run them for a small set of providers.

But I think that this should be configured in gardener so that test owners can directly configure the applicable shoot flavors when writing the tests.
With this we would have also simple versioning of that mapping file.
Maybe we can think of a specific configuration file that is evaluated by the testsuite and can run tests based on this config.

WDYT? @dguendisch

rfranzke: I don't know, I'm not a fan of configuring such things in gardener/gardener. Instead, it should be configured in the extensions repos which tests they are supporting/not supporting. Otherwise, we are again having provider-specifics in g/g and this is something that we wanted to get rid of with extensibility...

dguendisch: > Maybe we can think of a specific configuration file that is evaluated by the testsuite and can run tests based on this config.

Excellent idea! It avoids having to hard-code cloud-provider mappings.
When this file can be provided by the testmachinery, we could store such mappings along the mappings of flavors (in the landscape configuration).
I'd tend to prefer a more centralized storage of test-to-flavor/cloudprovider-applicability in our landscape configs as opposed to the test itself, simply because it gives us way better turnaround times in changing these configurations/mappings. So often we need to enable this test/flavor combination here or disable this test/flavor there because of some temporal thingy or because of a new flavor or test-dimension.

schrodit: A configuration that is supported by the gardener framework but reside in another repo to skip tests is valid.
Then the testmachinery would provide such config during runtime and we could maintain that file somewhere central.
This would give us the flexibility to skip tests in case of errors and temporary issues.

But regarding tests that only apply for specific cloudproviders.
I think we should have an inidication in the test itself that it is only supposed to work with a fixed set of cloudproviders.
Otherwise when the configuration is in a magic file outside g/g people do not know that these tests are only supposed to run on specific cloudproviders/flavors.

> Otherwise, we are again having provider-specifics in g/g and this is something that we wanted to get rid of with extensibility...

I get you're point. But then I these tests are provider specific and should be moved to the extension repos?

rfranzke: > I think we should have an inidication in the test itself that it is only supposed to work with a fixed set of cloudproviders.
Otherwise when the configuration is in a magic file outside g/g people do not know that these tests are only supposed to run on specific cloudproviders/flavors.

I would try to flag such tests as "disabled by default" and add documentation that they have to explicitly enabled because they test stuff where provider-specific support/implementation is a prerequisite. Hence, the tests should only run if explicitly stated that the underlying extensions support the test case. Which of such tests are support and which not should be again maintained in the extensions repos.

> I get you're point. But then I these tests are provider specific and should be moved to the extension repos?

I wouldn't do that because it would result in a lot of duplication. What about above proposal?

Issue #1751, opened 2019-12-18
Improve Shoot Hibernation Prevention Check


vpnachev:
**What would you like to be added**:

Recently we have introduced a validation(https://github.com/gardener/gardener/pull/1642) that prevents shoot hibernation in case the cluster has a validating or mutating webhook which might block creation of new pods or nodes. The motivation is that such clusters cannot be waken up later.

The validation [logic](https://github.com/gardener/gardener/blob/ca5c68dd4102c519a31ca5f52462874c3a194803/pkg/operation/botanist/constraints_check.go#L59-L138) can be improved by using [ShouldCallHook](https://github.com/kubernetes/kubernetes/blob/127c47caf49bec4a59a0500f99b1b1d4cdbf340e/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook.go#L143-L204). 
The unit [tests]( https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/generic/webhook_test.go) provide good examples how it can be used.

So, the following api requests must not be blocked by any webhook
*  creations of pods in `kube-system` namespace labeled with `origin=gardener`.
* creation of nodes

**Why is this needed**:
There are clusters with not so harmful admission webhooks and the hibernation for them should be possible.

timuthy:
We can also think about extending the list above by `endpoints` in `kube-system`. Or at least some other implementation to prevent the following scenario:

- Cluster with `Mutatingwebhookconfiguration` on `endpoints`
- Cluster entered hibernation
- The endpoints of the service backing the `Mutatingwebhookconfiguration` still existed
- KCM tried to renew it's leader election (`renew-deadline` 10s) but didn't succeed because the Kube-Apiserver timed out calling the webhook
- Eventually, all running `Pods` in the cluster couldn't be garbage collected and the shoot wasn't able to completely hibernate.

rfranzke:
As suggest by @tim-ebert it makes also sense to improve the "check if endpoints exist" check. Custom operators might manage `Endpoint` objects on their own. Hence, let's ignore all `Endpoint`s that don't belong to `Service`s (identifiable by looking at the `Service` object with the same name like the `Endpoint` object - if no label selector is present then the `Endpoint` was not created by KCM endpoint controller).

loewenstein:
Is there an eta for this issue?

timebertt:
[...]

PR #2205, opened 2020-04-20
Ignore endpoints not reconciled by kcm during hibernation

timebertt:
This PR improves the "check if endpoints referencing Pod IPs exist" check, which is run during hibernation. Now, gardener checks if the Endpoints object is reconciled by `kube-controller-manager`, otherwise it ignores and does not block the hibernation. This is done to support hibernating shoot clusters with custom operators, which reconcile Endpoints objects themselves (e.g. knative).

Part of #1751

**Release note**:

```improvement user
Gardener does not block hibernation of a Shoot Cluster anymore, in case it contains Endpoints objects that are reconciled by a custom operator (e.g. knative).
```

### DIFF
--- a/pkg/operation/botanist/waiter.go
+++ b/pkg/operation/botanist/waiter.go
@@ -35,6 +35,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


pkg/utils/kubernetes/kubernetes.go
```diff
@@ -199,3 +199,16 @@ func GetLoadBalancerIngress(ctx context.Context, client client.Client, namespace
 
 	return "", errors.New("`.status.loadBalancer.ingress[]` has an element which does neither contain `.ip` nor `.hostname`")
 }
+
+// IsEndpointsReconciledByKubeControllerManager checks if a given Endpoints object is reconciled by kube-controller-manager.
+// It looks at the corresponding Service (same name and namespace) and checks whether a selector is specified or not.
+func IsEndpointsReconciledByKubeControllerManager(ep corev1.Endpoints, services []corev1.Service) bool {
+	for _, service := range services {
```
rfranzke: This might be very costly if there are many endpoints/services in the cluster. What about creating a map[string]*corev1.Service that maps `namespace/name` to a service object?

timebertt: Good point, I adapted the logic to use a simple string set for storing, which Endpoints objects should be ignored. PTAL :)

rfranzke (dismissed):

timuthy (dismissed):

pkg/operation/botanist/waiter.go
```diff
@@ -336,12 +337,30 @@ func (b *Botanist) WaitUntilEndpointsDoNotContainPodIPs(ctx context.Context) err
...
+		}
+
+	OUTER:
 		for _, endpoints := range endpointsList.Items {
 			for _, subset := range endpoints.Subsets {
 				for _, address := range subset.Addresses {
 					if podsNetwork.Contains(net.ParseIP(address.IP)) {
-						msg := fmt.Sprintf("waiting until there are no Endpoints containing Pod IPs in the shoot cluster..."+
-							"there is still at least one Endpoints containing a Pod's IP: %s/%s, IP: %s", endpoints.Namespace, endpoints.Name, address.IP)
+						if epsNotReconciledByKCM.Has(fmt.Sprintf("%s/%s", endpoints.Namespace, endpoints.Name)) {
```
vpnachev: You can move this check between these 2 lines
```golang
	for _, endpoints := range endpointsList.Items {
			for _, subset := range endpoints.Subsets {
```

vpnachev (commented):
minor nit

pkg/operation/botanist/waiter.go
timebertt: Good suggestion, thanks! PTAL :)

vpnachev (approved):
/lgtm

rfranzke (approved):

PR #2206, opened 2020-04-21
Add local ssd disk interface support

qiffang:
In https://github.com/gardener/machine-controller-manager/pull/414, we add local ssd disks support.
This PR enables user to add local ssd disks interface,  the local ssd disk of GCP supports two interfaces: NVME and SCSI. See details in the below documents: 

- https://cloud.google.com/compute/docs/disks/local-ssd#create_local_ssd

- https://pkg.go.dev/google.golang.org/api@v0.17.0/compute/v1?tab=doc#AttachedDisk

I will update `gardener-extension-provider-gcp` to receive interface parameter after this PR is accepted.
Currently, I hard code the special interface `SCSI` in `gardener-extension-provider-gcp` in my local environment.


pkg/apis/core/types_shoot.go
```diff
@@ -706,6 +706,8 @@ type Volume struct {
 	VolumeSize string
 	// Encrypted determines if the volume should be encrypted.
 	Encrypted *bool
+	// Interface is the interface of local ssd disk.
+	Interface *string
```
rfranzke, line 710: Should we call this `LocalSSDInterface` to make it more clear what this field is used for?

pkg/apis/extensions/v1alpha1/types_worker.go
```diff
@@ -153,6 +153,9 @@ type Volume struct {
 	// Encrypted determines if the volume should be encrypted.
 	// +optional
 	Encrypted *bool `json:"encrypted,omitempty"`
+	// Interface is the interface of local ssd disk.
+	// +optional
+	Interface  *string `json:"interface,omitempty"`
```
rfranzke, line 158: Same here, should it be `LocalSSDInterface`?

However, is this field something that is potentially applicable to all providers (not only GCP)? If yes, then it makes certainly sense to introduce it here in gardener/gardener (please also enhance https://github.com/gardener/gardener/blob/master/docs/extensions/worker.md in this case). However, if it's more something that is only relevant to GCP then it should probably be done as part of GCP's `WorkerConfig`, see e.g. an example for AWS: https://github.com/gardener/gardener-extension-provider-aws/blob/master/pkg/apis/aws/v1alpha1/types_worker.go#L24-L50

WDYT?

rfranzke (changes requested):
Thanks for your contribution!

pkg/apis/extensions/v1alpha1/types_worker.go
qiffang: Currently, it is closely related to GCP, but i think it will also need to another cloud platform.
What's your suggestion?

rfranzke: > but i think it will also need to another cloud platform.

Why do you think that? How would a potential usage for AWS/Azure look like, for example? I just want to prevent that we add a provider-specific field to the Gardener API. It should be completely free of provider specifics. If the field is general enough then it's fine to add it here, otherwise something like https://github.com/gardener-attic/gardener-extensions/commit/ddba59f549c2ed50d49bf207242bd56adf126bf2 would make more sense.

qiffang: Ok, for aws, we can specify a machine which support local ssd.
Yeah, you are right, it is better to do as part of GCP's `WorkerConfig `.

But how to specify the `WorkConfig` when user create shoot cluster. I do not see there is a `WorkConfig` field in shoot struct of v1beta1

rfranzke: There is this field https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml#L49-L50 - it's similar to `infrastructureConfig` and `controlPlaneConfig`.

qiffang: Got it, i will move the interface field to GCP's `WorkerConfig` as your suggestion.

rfranzke: Sounds great, thank you very much for your efforts! :) Shall we close this PR then?

qiffang: Yes, i will close it.
By the way, i build the `gardener-extension-provider-gcp` by `make generate`, it failed
```
> Generate
charts/gardener-extension-provider-gcp/doc.go:15: running "../../vendor/github.com/gardener/gardener/extensions/hack/generate-controller-registration.sh": exit status 1
I0421 15:54:28.241924    6545 main.go:123] parsing go packages in directory .
I0421 15:54:34.763045    6545 main.go:225] using package=github.com/gardener/gardener-extension-provider-gcp/pkg/apis/config/v1alpha1
I0421 15:54:34.768536    6545 main.go:161] written to ../../../../hack/api-reference/config.md
Generating deepcopy funcs
Generating defaulters
Generating conversions
Generating deepcopy funcs
Generating defaulters
Generating conversions
I0421 15:55:14.983831   12855 main.go:123] parsing go packages in directory .
I0421 15:55:20.724491   12855 main.go:225] using package=github.com/gardener/gardener-extension-provider-gcp/pkg/apis/gcp/v1alpha1
I0421 15:55:20.732766   12855 main.go:161] written to ../../../../hack/api-reference/api.md
make: *** [Makefile:117: generate] Error 1
```

Do you have any idea?

rfranzke: Probably you have to run `make install-requirements`, but I suggest to move this discussion to the g/gardener-extension-provider-gcp repo :)

PR #2210, opened 2020-04-21
Create dedicated secrets of cloud-config for each worker group

### LINKED ISSUE #2168: Workers count is limited by cloud-config Secret size limit
**What happened**:
Currently it is not possible to have more than 20-25 (depending on the OS) workers in `.spec.provider.workers`.

1. Create a Shoot with 25 Worker entries under `.spec.provider.workers`.

2. Ensure that the `Deploying managed resources` step fails with:

```yaml
status:
  lastErrors:
  - description: 'task "Deploying managed resources" failed: retry failed with context
      deadline exceeded, last error: Secret "managedresource-shoot-cloud-config-execution"
      is invalid: data: Too long: must have at most 1048576 characters'
    lastUpdateTime: "2020-04-09T11:08:27Z"
    taskID: Deploying managed resources
```

**Environment**:

- Gardener version: v1.2.0
- Others:

aylei:
Signed-off-by: Aylei <rayingecho@gmail.com>

Currently it is not possible to have more than 20-25 (depending on the OS) workers because the number of worker's cloud configs are limited by the size of the secret. This PR divide the cloud config to dedicated managed resources to eliminate this limitation.

I've tested this PR on eks v1.14.9 and managed to create 31 worker groups.

**edit:**
After I changed the implementation I tested with two worker groups and it works fine

```
spec:
  injectLabels:
    shoot.gardener.cloud/no-cleanup: "true"
  keepObjects: false
  secretRefs:
  - name: managedresource-shoot-cloud-config-rbac
  - name: managedresource-shoot-cloud-config-execution-worker-x934f
  - name: managedresource-shoot-cloud-config-execution-worker-ahhgt
```

Fixes #2168 

**Release note**:

```improvement operator
Fix a bug that limits the workers count of a single shoot.
```


pkg/operation/botanist/addons.go
```diff
@@ -129,14 +130,26 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
...
+			"addons":                  {false, b.generateOptionalAddonsChart},
 			// TODO: Just a temporary solution. Remove this in a future version once Kyma is moved out again.
 			"addons-kyma": {false, b.generateTemporaryKymaAddonsChart},
 		}
 	)
+	bootstrapTokenSecret, err := kutil.ComputeBootstrapToken(context.TODO(), b.K8sShootClient.Client(), utils.ComputeSHA256Hex([]byte(time.Now().Format("2006-01-02")))[:6], "A bootstrap token generated by Gardener.", 48*time.Hour)
+	if err != nil {
+		return fmt.Errorf("error computing bootstrap token for shoot cloud config: %+v", err)
+	}
+	// deploy cloud config execution scripts for each worker group
```
danielfoehrKn:
```suggestion
	//  for each worker pool add managed resource containing a secret with the cloud config execution script
```

danielfoehrKn (changes requested):
Thanks for your contribution.
I looked over it and tested it (with two worker pools) - looks good!

I only noticed one thing: the managed resource "shoot-cloud-config-execution" that now should be replaced with a managed resource per worker pool, is not being cleaned up.
However, just deleting the managed resource comes with a problem: deleting a managed resource also deletes its depending resources.

Maybe it could be done like this:
After the new managed resources are already applied, delete the managed resource  "shoot-cloud-config-execution" . The deleted secrets in the kube-system namespace of the shoot are re-applied right away by the other managed resources. 
This is only one option on top of my head. Maybe just deleting the finalizers and the secret is another.

The cleanup would be only required as a migration step and can be deleted in a future version if Gardener.

aylei:
@danielfoehrKn Thanks for your feedback! Another solution is keep the name "shoot-cloud-config-execution" instead of renaming it to "shoot-cloud-config-rbac". I changed the name in favor of more accurate semantic, but regarding the concern of cleanup, keep the original name would be a better idea.

How do you think?

pkg/operation/botanist/addons.go
```diff
@@ -129,14 +130,26 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
...
-			"addons":                       {false, b.generateOptionalAddonsChart},
+			"shoot-cloud-config-rbac": {false, b.generateCloudConfigRBACChart},
+			"shoot-core":              {false, b.generateCoreAddonsChart},
+			"shoot-core-namespaces":   {true, b.generateCoreNamespacesChart},
+			"addons":                  {false, b.generateOptionalAddonsChart},
 			// TODO: Just a temporary solution. Remove this in a future version once Kyma is moved out again.
 			"addons-kyma": {false, b.generateTemporaryKymaAddonsChart},
 		}
 	)
+	bootstrapTokenSecret, err := kutil.ComputeBootstrapToken(context.TODO(), b.K8sShootClient.Client(), utils.ComputeSHA256Hex([]byte(time.Now().Format("2006-01-02")))[:6], "A bootstrap token generated by Gardener.", 48*time.Hour)
```
```suggestion
	bootstrapTokenSecret, err := kutil.ComputeBootstrapToken(ctx, b.K8sShootClient.Client(), utils.ComputeSHA256Hex([]byte(time.Now().Format("2006-01-02")))[:6], "A bootstrap token generated by Gardener.", 48*time.Hour)
```

charts/shoot-cloud-config/values.yaml
```diff
@@ -1,26 +1,16 @@
...
-#   secretName: cloud-config-cpu-worker2-4av4a
-#   cloudConfig: generated-original-cloud-config-data
-#   command: /usr/bin/reload --path=<path>
-#   units:
-#   - kubelet.service
-#   kubeletDataVolume:
-#     name: vol2
-#     size: 9375000000
-#     type: io1
+images:
```
rfranzke, line 1: Hm, should we comment all the fields again? Any particular reason why you uncommented them?

rfranzke (commented):
@danielfoehrKn has brought up a good point, thanks for catching it.

> Another solution is keep the name "shoot-cloud-config-execution" instead of renaming it to "shoot-cloud-config-rbac". I changed the name in favor of more accurate semantic, but regarding the concern of cleanup, keep the original name would be a better idea.

A `ManagedResource` can reference multiple secrets. What do you think about keeping the `shoot-cloud-config-execution` `ManagedResource` but create a dedicated `Secret` for each pool, and then reference all of them in the `shoot-cloud-config-execution` `ManagedResource`?

PR looks great, btw, thanks a lot! :)

charts/shoot-cloud-config/values.yaml
aylei: `check-charts.sh` expects all charts could be rendered with the default values, and I found that other charts used dummy values to pass the check

aylei:
> A `ManagedResource` can reference multiple secrets. What do you think about keeping the `shoot-cloud-config-execution` `ManagedResource` but create a dedicated `Secret` for each pool, and then reference all of them in the `shoot-cloud-config-execution` `ManagedResource`?

Good call! I wouldn't have been worried the naming issue if I knew this before

aylei:
@rfranzke @danielfoehrKn I've changed the implementation to keep the `shoot-cloud-config-execution` managed resources and only break down the secrets, PTAL

pkg/utils/secrets/basic_auth.go
```diff
@@ -101,7 +101,7 @@ func (s *BasicAuthSecretConfig) GenerateBasicAuth() (*BasicAuth, error) {
 	return basicAuth, nil
 }
 
-// SecretData computes the data map which can be used in a Kubernetes secret.
+// AsSecretData computes the data map which can be used in a Kubernetes secret.
```
rfranzke:
```suggestion
// SecretData computes the data map which can be used in a Kubernetes secret.
```

pkg/utils/secrets/certificates.go
```diff
@@ -154,7 +154,7 @@ func (s *CertificateSecretConfig) GenerateCertificate() (*Certificate, error) {
 	return certificateObj, nil
 }
 
-// SecretData computes the data map which can be used in a Kubernetes secret.
+// AsSecretData computes the data map which can be used in a Kubernetes secret.
```
rfranzke:
```suggestion
// SecretData computes the data map which can be used in a Kubernetes secret.
```

pkg/utils/secrets/control_plane.go
```diff
@@ -89,7 +89,7 @@ func (s *ControlPlaneSecretConfig) GenerateControlPlane() (*ControlPlane, error)
 	return controlPlane, nil
 }
 
-// SecretData computes the data map which can be used in a Kubernetes secret.
+// AsSecretData computes the data map which can be used in a Kubernetes secret.
```
rfranzke:
```suggestion
// SecretData computes the data map which can be used in a Kubernetes secret.
```

pkg/utils/secrets/rsa_private_key.go
```diff
@@ -83,7 +83,7 @@ func (s *RSASecretConfig) GenerateRSAKeys() (*RSAKeys, error) {
 	return rsa, nil
 }
 
-// SecretData computes the data map which can be used in a Kubernetes secret.
+// AsSecretData computes the data map which can be used in a Kubernetes secret.
```
rfranzke:
```suggestion
// SecretData computes the data map which can be used in a Kubernetes secret.
```

pkg/utils/secrets/static_token.go
```diff
@@ -95,7 +95,7 @@ func (s *StaticTokenSecretConfig) GenerateStaticToken() (*StaticToken, error) {
 	}, nil
 }
 
-// SecretData computes the data map which can be used in a Kubernetes secret.
+// AsSecretData computes the data map which can be used in a Kubernetes secret.
```
rfranzke:
```suggestion
// SecretData computes the data map which can be used in a Kubernetes secret.
```

pkg/operation/botanist/addons.go
```diff
@@ -172,7 +167,47 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
...
+		if err := manager.
+			NewSecret(b.K8sSeedClient.Client()).
+			WithNamespacedName(b.Shoot.SeedNamespace, secretName).
+			WithKeyValues(data).
+			Reconcile(ctx); err != nil {
+			return err
+		}
+		cloudConfigManagedResource.WithSecretRef(secretName)
+	}
+	return cloudConfigManagedResource.Reconcile(ctx)
```
rfranzke: I like it very much, thanks :)

rfranzke (changes requested):

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -442,23 +434,31 @@ func (b *Botanist) generateCloudConfigExecutionChart() (*chartrenderer.RenderedC
...
+			"kubernetesVersion": b.Shoot.Info.Spec.Kubernetes.Version,
+			"worker":            w,
+		}
+		var err error
+		config, err = b.InjectShootShootImages(config, common.HyperkubeImageName)
+		if err != nil {
+			return nil, err
+		}
+		return b.ChartApplierShoot.Render(filepath.Join(common.ChartPath, "shoot-cloud-config"), releaseName, metav1.NamespaceSystem, config)
 	}
```
rfranzke, line 449: I played with it locally, works great. I think the only thing that is missing is what @danielfoehrKn already pointed out earlier:

The old `managedresource-shoot-cloud-config-execution` secret remains orphaned in the system for already existing shoots. While it's not harming in any way it's probably still better to simply delete it in order to prevent confusing of operators that are not that familiar with the code. This deletion code can be added after this line and should have a `TODO` comment that it can be removed in a future version of Gardener.

pkg/utils/secrets/basic_auth.go
aylei: Thanks for correcting these!

pkg/operation/botanist/operatingsystemconfig.go
aylei: Good catch, what do you think about keeping the secret name `managedresource-shoot-cloud-config-execution` for the rbac resources? This way we don't have to worry about migration

rfranzke: Hm, I'd say it's easier to follow if the name of the secret containing the `shoot-cloud-config-exeuction-rbac` manifests has also `rbac` in its name. Also, if we would overwrite it like this the `cloud-config-<worker-pool-name>` secret in the `kube-system` namespace of the shoot would be deleted (it would be re-applied by the new managed resource that you create for the worker pool, but it would be shortly not present). This probably doesn't harm too much, but summing those two facts up I'd vote for the small migration logic. WDYT?

aylei: Make sense, let's do the migration

aylei: Addressed in https://github.com/gardener/gardener/pull/2210/commits/c53a9e855e4eb567b9c1372c9d4243d4ee370313

pkg/operation/botanist/addons.go
```diff
@@ -172,6 +168,60 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
...
+}
+
+// deleteLegacyCloudConfigSecret deletes the legacy cloud-config-execution secret created by gardenlet prior to v1.3.0
+func (b *Botanist) deleteLegacyCloudConfigSecret() error {
+	secretName := "managedresource-shoot-cloud-config-execution"
+	err := b.K8sSeedClient.Kubernetes().CoreV1().Secrets(b.Shoot.SeedNamespace).Delete(secretName, metav1.NewDeleteOptions(0))
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
 	return nil
```
```suggestion
	return client.IgnoreNotFound(b.K8sSeedClient.Kubernetes().CoreV1().Secrets(b.Shoot.SeedNamespace).Delete(secretName, metav1.NewDeleteOptions(0)))
```

rfranzke (commented):
Thank you very much! Looks good, final suggestion. Also, can you please squash your commits into a single one to get it ready to be merged? :)

aylei:
> Thank you very much! Looks good, final suggestion. Also, can you please squash your commits into a single one to get it ready to be merged? :)

Sure

rfranzke (dismissed):

danielfoehrKn (dismissed):

pkg/chartrenderer/default.go
```diff
@@ -187,3 +187,13 @@ func (c *RenderedChart) FileContent(filename string) string {
 	}
 	return ""
 }
+
+// AsSecretData returns all rendered manifests that is capable for used as data of a secret
+func (c *RenderedChart) AsSecretData() map[string][]byte {
+	data := make(map[string][]byte, len(c.Files()))
+	for fileName, fileContent := range c.Files() {
+		key := strings.Replace(fileName, "/", "_", -1)
```
timebertt: WDYT, about using `strings.ReplaceAll`. It should yield the same results, but is probably nicer to read :)
```suggestion
		key := strings.ReplaceAll(fileName, "/", "_")
```

pkg/operation/botanist/addons.go
```diff
@@ -172,6 +168,60 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
...
+		return err
+	}
+	// TODO: remove this in the future
+	return b.deleteLegacyCloudConfigSecret()
+}
+
+// deleteLegacyCloudConfigSecret deletes the legacy cloud-config-execution secret created by gardenlet prior to v1.3.0
+func (b *Botanist) deleteLegacyCloudConfigSecret() error {
+	secretName := "managedresource-shoot-cloud-config-execution"
+	err := b.K8sSeedClient.Kubernetes().CoreV1().Secrets(b.Shoot.SeedNamespace).Delete(secretName, metav1.NewDeleteOptions(0))
```
timebertt: Any specific reason, why you set `deletionGracePeriodSeconds=0`?

timebertt (commented):
Thanks @aylei for your contribution. Appreciate it!
I have just two minor comments/questions, otherwise /lgtm

pkg/operation/botanist/addons.go
```diff
@@ -172,6 +168,60 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
 		}
 	}
 
+	// Create the cloud config managed resource that contains:
```
ialidzhikov: Can we move the `shoot-cloud-config-execution` ManagedResource creation into a dedicated func - for example `deployCloudConfigExectionManagedResource`. WDYT?

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -395,25 +395,17 @@ func (b *Botanist) applyAndWaitForShootOperatingSystemConfig(ctx context.Context
...
-	bootstrapTokenSecret, err := kutil.ComputeBootstrapToken(context.TODO(), b.K8sShootClient.Client(), utils.ComputeSHA256Hex([]byte(time.Now().Format("2006-01-02")))[:6], "A bootstrap token generated by Gardener.", 48*time.Hour)
-	if err != nil {
-		return nil, err
-	}
-	workers := make([]map[string]interface{}, 0, len(b.Shoot.Info.Spec.Provider.Workers))
-	for _, worker := range b.Shoot.Info.Spec.Provider.Workers {
-
+// getGenerateCloudConfigExecutionChartFunc returns a function that renders the gardener-resource-manager configuration for
+// the cloud config user data of the given worker.
+func (b *Botanist) getGenerateCloudConfigExecutionChartFunc(releaseName string, worker gardencorev1beta1.Worker, tokenSecret *corev1.Secret) func() (*chartrenderer.RenderedChart, error) {
```
ialidzhikov, line 400: Hmm, is there any specify reason to return `func() (*chartrenderer.RenderedChart, error)` but not simply `(*chartrenderer.RenderedChart, error)` as the other similar `generate<Name>Chart` funcs return `(*chartrenderer.RenderedChart, error)`?

pkg/operation/botanist/addons.go
```diff
@@ -172,6 +168,60 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
...
+	for name, renderChartFunc := range cloudConfigCharts {
+		renderedChart, err := renderChartFunc()
+		if err != nil {
+			return fmt.Errorf("error rendering %q chart: %+v", name, err)
+		}
+		data := renderedChart.AsSecretData()
+
+		secretName := "managedresource-" + name
+		if err := manager.
+			NewSecret(b.K8sSeedClient.Client()).
```
ialidzhikov: We currently don't have clean up logic for secrets which are no longer used. For example if you have 2 workers in the Shoot spec and you remove 1 of them, its ManagedResource secret won't be deleted.

ialidzhikov (commented):
Thank you for the PR!

rfranzke:
/kind/enhancement
/status/author-action

pkg/operation/botanist/addons.go
aylei: Good catch!

pkg/operation/botanist/operatingsystemconfig.go
aylei: I did this because `generateCloudConfigExecutionChart` requires some input parameters and I want to make the function signature to be consistent with other `generate<Name>Chart` funcs. So I used this function to return a `func() (*chartrenderer.RenderedChart, error)` which would capture the inputs from the context.

pkg/operation/botanist/addons.go
aylei: There is not specific reason actually, I will change it to unset

pkg/operation/botanist/addons.go
aylei: Sure

pkg/operation/botanist/addons.go
aylei: Addressed in https://github.com/gardener/gardener/pull/2210/commits/b8aadd8ebe56c6a226687d3f640295056f05c25c

aylei:
I'm separating the newest commit to make the addressing of comments clear, will squash the changes if it looks good for you @ialidzhikov @tim-ebert . And the secrets cleanup has been tested on EKS.

BTW, just out of curiosity, why didn't we enable the `squash and merge` support of Github? This would be really convenient as I experienced.

pkg/operation/botanist/addons.go
aylei: Addressed in https://github.com/gardener/gardener/commit/b8aadd8ebe56c6a226687d3f640295056f05c25c

pkg/operation/botanist/addons.go
```diff
@@ -172,7 +176,123 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
...
 	}
 
-	return nil
+	return b.deployCloudConfigExectionManagedResource(ctx)
+}
+
+// deployCloudConfigExectionManagedResource create the cloud config managed resource that contains:
+// 1. A secret containing the dedicated cloud config execution script for each worker group
+// 2. A secret containing some shared RBAC policies for downloading the cloud config execution script
+func (b *Botanist) deployCloudConfigExectionManagedResource(ctx context.Context) error {
```
```suggestion
func (b *Botanist) deployCloudConfigExecutionManagedResource(ctx context.Context) error {
```

pkg/operation/botanist/addons.go
```diff
@@ -172,7 +176,123 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
 		}
 	}
 
-	return nil
+	return b.deployCloudConfigExectionManagedResource(ctx)
+}
+
+// deployCloudConfigExectionManagedResource create the cloud config managed resource that contains:
```
```suggestion
// deployCloudConfigExectionManagedResource creates the cloud config managed resource that contains:
```

pkg/operation/botanist/addons.go
```diff
@@ -172,7 +176,123 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
...
+	for _, secret := range secretList.Items {
+		if !wantedSecretNames.Has(secret.Name) {
+			toDelete := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      secret.Name,
+					Namespace: secret.Namespace,
+				},
+			}
+			fns = append(fns, func(ctx context.Context) error {
+				if err := common.ConfirmDeletion(ctx, c, toDelete); err != nil {
```
rfranzke: This is not needed for the `Secret` deletion

pkg/operation/botanist/addons.go
```diff
@@ -172,7 +176,123 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
...
+func (b *Botanist) deleteLegacyCloudConfigSecret(ctx context.Context) error {
+	c := b.K8sSeedClient.Client()
+
+	legacySecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "managedresource-shoot-cloud-config-execution",
+			Namespace: b.Shoot.SeedNamespace,
+		},
+	}
+	if err := common.ConfirmDeletion(ctx, b.K8sSeedClient.Client(), legacySecret); err != nil {
```
rfranzke: This is not needed for the secret  deletion

rfranzke (commented):
Thanks for updating, a few nits. @ialidzhikov can you check again?

pkg/operation/botanist/addons.go
```diff
@@ -172,7 +176,123 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
 		}
 	}
 
-	return nil
+	return b.deployCloudConfigExectionManagedResource(ctx)
```
ialidzhikov:
```suggestion
	return b.deployCloudConfigExecutionManagedResource(ctx)
```

pkg/operation/botanist/addons.go
```diff
@@ -172,7 +176,123 @@ func (b *Botanist) DeployManagedResources(ctx context.Context) error {
...
+}
+
+// deployCloudConfigExectionManagedResource creates the cloud config managed resource that contains:
```
ialidzhikov:
```suggestion
// deployCloudConfigExecutionManagedResource creates the cloud config managed resource that contains:
```

pkg/operation/botanist/addons.go
aylei: Thanks! I just changed it locally

pkg/operation/botanist/addons.go
aylei: Addressed in https://github.com/gardener/gardener/pull/2210/commits/b850b90295d18f227fa810fa08b1e67e5661d075

pkg/operation/botanist/addons.go
aylei: Addressed in https://github.com/gardener/gardener/pull/2210/commits/b850b90295d18f227fa810fa08b1e67e5661d075

ialidzhikov:
@aylei , could you please squash to single commit?

aylei:
> @aylei , could you please squash to single commit?

sure and done

ialidzhikov (approved):
/lgtm

rfranzke (approved):

danielfoehrKn (approved):

PR #2212, opened 2020-04-22
Add error codes to conditions

rfranzke:
As of today we only set error codes during reconciliation. However, the shoot health checks may also detect situations in which error codes make sense to populate. Hence, this PR adds a new `codes []ErrorCode` field to conditions. Gardener will fetch the codes from the extensions as well and merge them into the shoot conditions.

Gardener itself is not yet using error codes for conditions, will work on it in a separate PR.

**Release note**:

```improvement user
Shoot conditions may now also specify a list of error codes, similar to `.status.lastError.codes`.
```
```improvement developer
Extension resource conditions may now also specify a list of error codes, similar to `.status.lastError.codes`. Gardener will pick them up and merge them into the shoot conditions.
```


timebertt (approved):
Nice PR :)
/lgtm

danielfoehrKn:
Looks good. I like that by the fact that the extensions use the same condition definition - there is actually not so much to do.

danielfoehrKn (approved):

PR #2213, opened 2020-04-22
Make the pod affinity of `fluentd` preferred.

Kristian-ZH:
Make the pod affinity of `fluentd` preferred.

Fixes #

@vlvasilev @vpnachev @ialidzhikov 

**Release note**:

```improvement operator
The `podAntiAffinity` of the `fluentd` statefulset deployed in the seed clusters is now a soft requirement.
```

### DIFF
--- a/charts/seed-bootstrap/charts/fluentd-es/templates/fluentd-es-statefulset.yaml
+++ b/charts/seed-bootstrap/charts/fluentd-es/templates/fluentd-es-statefulset.yaml
@@ -29,18 +29,20 @@ spec:
       {{- if gt (int .Values.fluentd.replicaCount ) 1 }}
       affinity:
         podAntiAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-          - labelSelector:
-              matchExpressions:
-              - key: app
-                operator: In
-                values:
-                - fluentd-es
-              - key: role
-                operator: In
-                values:
-                - logging
-            topologyKey: "kubernetes.io/hostname"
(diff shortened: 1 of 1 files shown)


vpnachev (changes requested):
Please, add a proper release note.

danielfoehrKn:
Fluent Bit is deployed as a Daemonset that's why the Affinity is not needed any more?

Kristian-ZH:
> Fluent Bit is deployed as a Daemonset that's why the Affinity is not needed any more?

Before, we have used fluentd as log-collector. For this reason it had to be on every node. Not the fluent-bit does this work and we do not need affinity on fluentd level

Kristian-ZH:
> Fluent Bit is deployed as a Daemonset that's why the Affinity is not needed any more?

\cc @vlvasilev  Can you explain, please why have we used the affinity before.

vlvasilev:
In the past the meta data for a particular log entry was retrieved by fluentd. This was making extremely high amount of request. Thus the internet traffic in a particular nod was flooded and to mitigated this we made the pod affinity so no two fluentd pod appears in a node at one time. After refactoring this meta data gathering was left to fluent-bit. At first place this affinity is no longer needed and second on a Seed with low number of nodes this will not allow proper horizontal scale of the fluentd.

Kristian-ZH:
After a discussion with my colleagues about the affinity, we decided that it is a better option to make it `preffered` than remove it. 

So the PR is updated now

vpnachev (dismissed):
/lgtm

vpnachev (approved):
/lgtm

PR #2217, opened 2020-04-22
Improve Cluster synchronization to seed

rfranzke:
This PR refactors the `Cluster` synchronization to the seed when a shoot shall be reconciled or deleted. Earlier it was possible that the shoot's status is set to `Error` in case the `Cluster` sync failed even if the shoot is outside of its maintenance time window.

With this PR, we have refactored the way how `Garden`, `Seed`, `Shoot`, and `Operation` objects are built (using the builder pattern). Also, we have refactored the shoot controller to only sync the `Cluster` resource right before the reconciliation starts. In the `reconcileShoot` function it's also synced right after the reconciliation succeeded. Also, if a cluster has `Failed` status or shall be ignored the `Cluster` resource is synced in order to allow the extension controllers to get to know about this status.

/cc @timuthy

**Release note**:

```improvement user
An issue that caused the `Shoot`'s `.status.lastOperation.state` to be set to `Error` although no actual reconciliation operation is executed for the `Shoot` has been fixed.
```

### DIFF
--- a/pkg/gardenlet/controller/seed/seed_control.go
+++ b/pkg/gardenlet/controller/seed/seed_control.go
@@ -297,7 +297,11 @@ func (c *defaultControl) ReconcileSeed(obj *gardencorev1beta1.Seed, key string) 
 	// Initialize conditions based on the current status.
 	conditionSeedBootstrapped := gardencorev1beta1helper.GetOrInitCondition(seed.Status.Conditions, gardencorev1beta1.SeedBootstrapped)
 
-	seedObj, err := seedpkg.New(c.k8sGardenClient, c.k8sGardenCoreInformers.Core().V1beta1(), seed)
+	seedObj, err := seedpkg.
+		NewBuilder().
+		WithSeedObject(seed).
+		WithSeedSecretFromClient(ctx, c.k8sGardenClient.Client()).
+		Build()
 	if err != nil {
 		message := fmt.Sprintf("Failed to create a Seed object (%s).", err.Error())
 		conditionSeedBootstrapped = gardencorev1beta1helper.UpdatedCondition(conditionSeedBootstrapped, gardencorev1beta1.ConditionUnknown, gardencorev1beta1.ConditionCheckError, message)
--- a/pkg/gardenlet/controller/shoot/shoot_care_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_care_control.go
@@ -134,7 +134,17 @@ func (c *defaultCareControl) Care(shootObj *gardencorev1beta1.Shoot, key string)
 
 	shootLogger.Debugf("[SHOOT CARE] %s", key)
 
-	operation, err := operation.New(shoot, c.config, shootLogger, c.k8sGardenClient, c.k8sGardenCoreInformers, c.identity, c.secrets, c.imageVector)
+	operation, err := operation.
+		NewBuilder().
+		WithLogger(shootLogger).
+		WithConfig(c.config).
+		WithGardenerInfo(c.identity).
+		WithSecrets(c.secrets).
+		WithImageVector(c.imageVector).
+		WithGardenFrom(c.k8sGardenCoreInformers, shoot.Namespace).
+		WithSeedFrom(c.k8sGardenCoreInformers, *shoot.Spec.SeedName).
+		WithShootFrom(c.k8sGardenCoreInformers, shoot).
+		Build(context.TODO(), c.k8sGardenClient)
 	if err != nil {
 		shootLogger.Errorf("could not initialize a new operation: %s", err.Error())
 		return nil // We do not want to run in the exponential backoff for the condition checks.
--- a/pkg/gardenlet/controller/shoot/shoot_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control.go
@@ -201,7 +201,17 @@ func (c *Controller) deleteShoot(shoot *gardencorev1beta1.Shoot, logger *logrus.
 		return reconcile.Result{}, c.updateShootStatusDeleteSuccess(shoot)
 	}
 
-	o, err := operation.New(shoot, c.config, logger, c.k8sGardenClient, c.k8sGardenCoreInformers.Core().V1beta1(), c.identity, c.secrets, c.imageVector)
+	o, err := operation.
+		NewBuilder().
+		WithLogger(logger).
+		WithConfig(c.config).
+		WithGardenerInfo(c.identity).
+		WithSecrets(c.secrets).
+		WithImageVector(c.imageVector).
+		WithGardenFrom(c.k8sGardenCoreInformers.Core().V1beta1(), shoot.Namespace).
+		WithSeedFrom(c.k8sGardenCoreInformers.Core().V1beta1(), *shoot.Spec.SeedName).
+		WithShootFrom(c.k8sGardenCoreInformers.Core().V1beta1(), shoot).
+		Build(context.TODO(), c.k8sGardenClient)
 	if err != nil {
 		return reconcile.Result{}, utilerrors.WithSuppressed(err, c.updateShootStatusError(shoot, fmt.Sprintf("Could not initialize a new operation for Shoot deletion: %s", err.Error())))
 	}
@@ -303,7 +313,17 @@ func (c *Controller) reconcileShoot(shoot *gardencorev1beta1.Shoot, logger *logr
 	// make sure that the latest version of the shoot object is used as the basis for next operations
 	shoot = updatedShoot
 
-	o, err := operation.New(shoot, c.config, logger, c.k8sGardenClient, c.k8sGardenCoreInformers.Core().V1beta1(), c.identity, c.secrets, c.imageVector)
+	o, err := operation.
+		NewBuilder().
+		WithLogger(logger).
+		WithConfig(c.config).
+		WithGardenerInfo(c.identity).
+		WithSecrets(c.secrets).
+		WithImageVector(c.imageVector).
+		WithGardenFrom(c.k8sGardenCoreInformers.Core().V1beta1(), shoot.Namespace).
+		WithSeedFrom(c.k8sGardenCoreInformers.Core().V1beta1(), *shoot.Spec.SeedName).
+		WithShootFrom(c.k8sGardenCoreInformers.Core().V1beta1(), shoot).
+		Build(context.TODO(), c.k8sGardenClient)
 	if err != nil {
 		if failedOrIgnored {
 			// do not set error from operation initialization in Shoot status if Shoot would not be reconciled anyways
--- a/pkg/gardenlet/controller/shoot/shoot_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control.go
@@ -23,11 +23,14 @@ import (
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"github.com/gardener/gardener/pkg/controllerutils"
 	"github.com/gardener/gardener/pkg/logger"
 	"github.com/gardener/gardener/pkg/operation"
 	"github.com/gardener/gardener/pkg/operation/common"
 	"github.com/gardener/gardener/pkg/operation/garden"
+	seedpkg "github.com/gardener/gardener/pkg/operation/seed"
+	shootpkg "github.com/gardener/gardener/pkg/operation/shoot"
 	"github.com/gardener/gardener/pkg/utils"
 	utilerrors "github.com/gardener/gardener/pkg/utils/errors"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
@@ -37,11 +40,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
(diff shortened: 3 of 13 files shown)


timuthy:
@gardener/gardener-maintainers has someone capacity to double check? @rfranzke and I paired for this PR but another round of review would be helpful.

rfranzke:
/kind/bug

rfranzke:
/exp/expert
/platform/all

rfranzke:
/exp/expert

rfranzke:
/platform/all

vlerenc:
/exp/expert
/platform/all

vlerenc:
/kind/epic
/kind/poc

vlerenc:
/kind/epic
/kind/poc

vlerenc:
/kind/epic
/kind/poc

vlerenc:
Just a note on the bot labels up there:
- the first is not valid for PR
- the second is not defined in he bot manager

timebertt:
Will take a look today, sorry for the delay

pkg/gardenlet/controller/seed/seed_control.go
```diff
@@ -297,7 +297,11 @@ func (c *defaultControl) ReconcileSeed(obj *gardencorev1beta1.Seed, key string)
 	// Initialize conditions based on the current status.
 	conditionSeedBootstrapped := gardencorev1beta1helper.GetOrInitCondition(seed.Status.Conditions, gardencorev1beta1.SeedBootstrapped)
 
-	seedObj, err := seedpkg.New(c.k8sGardenClient, c.k8sGardenCoreInformers.Core().V1beta1(), seed)
```
ialidzhikov, line 300: The New and NewFromName funcs form github.com/gardener/gardener/pkg/operation/seed pkg seems to be no longer used. Can we remove them?

pkg/operation/seed/seed.go
```diff
@@ -53,6 +53,68 @@ import (
...
 )
 
+// NewBuilder returns a new Builder.
+func NewBuilder() *Builder {
+	return &Builder{
+		seedObjectFunc: func() (*gardencorev1beta1.Seed, error) { return nil, fmt.Errorf("seed object is required but not set") },
+	}
+}
+
+// WithSeedObject sets the seedObject attribute at the Builder.
```
timebertt: Can you check the doc strings of all the introduced builders, please? There are some occurrences stating that the method sets the `*Func` attribute of the builder and others stating it sets the attribute without the `Func` suffix (I guess, this is just a leftover)
```suggestion
// WithSeedObject sets the seedObjectFunc attribute at the Builder to return the given Seed.
```

ialidzhikov (commented):
The behaviour I described in https://github.com/gardener/gardener-extension-provider-aws/pull/62#pullrequestreview-400915000 is coming from this PR.
I think the reason is that the Cluster initially is created without .status.lastOperation and this prevents the `ShootNotFailed` predicate to be satisfied?

pkg/operation/shoot/shoot.go
```diff
@@ -41,82 +40,156 @@ import (
...
-	if err != nil {
-		return nil, err
+// WithShootObject sets the shootObject attribute at the Builder.
+func (b *Builder) WithShootObject(shootObject *gardencorev1beta1.Shoot) *Builder {
+	b.shootObjectFunc = func() (*gardencorev1beta1.Shoot, error) { return shootObject, nil }
+	return b
+}
+
+// WithShootObjectFromLister sets the shootObject attribute at the Builder after fetching it from the given lister.
+func (b *Builder) WithShootObjectFromLister(shootLister gardencorelisters.ShootLister, namespace, name string) *Builder {
```
timebertt, line 65: This method is unused.

pkg/operation/operation.go
```diff
@@ -50,100 +49,210 @@ import (
...
+
+	renderer, err := chartrenderer.NewForConfig(k8sGardenClient.RESTConfig())
+	if err != nil {
+		return nil, err
+	}
+	applier, err := kubernetes.NewApplierForConfig(k8sGardenClient.RESTConfig())
+	if err != nil {
+		return nil, err
+	}
+	operation.ChartApplierGarden = kubernetes.NewChartApplier(renderer, applier)
```
timebertt: It seems like `operation.ChartApplierGarden` is not used anymore.

pkg/operation/seed/seed.go
```diff
@@ -53,6 +53,68 @@ import (
...
+func (b *Builder) WithSeedObject(seedObject *gardencorev1beta1.Seed) *Builder {
+	b.seedObjectFunc = func() (*gardencorev1beta1.Seed, error) { return seedObject, nil }
+	return b
+}
+
+// WithSeedObjectFromLister sets the seedObject attribute at the Builder after fetching it from the given lister.
+func (b *Builder) WithSeedObjectFromLister(seedLister gardencorelisters.SeedLister, seedName string) *Builder {
```
timebertt, line 69: I have to admit, I am a bit confused by the different `With*From*` builder methods. Sometimes we have `*FromLister`, sometimes `*FromClient`, sometimes just `*From` (which will take an informer).

This one for example, I think, can be replaced with `WithSeedObjectFromClient`, as it is only used in `WithSeedFrom` alongside `WithSeedSecretFromClient` ~and the client for the garden cluster is anyways cached~.

Can we further simplify the different possibilities to pass clients to the builders (preferably using controller-runtime clients where possible and reasonable, as we anyways drift more and more towards using that), to make the builders even easier to use? (also see other comments on the respective methods, where I think, that I've found some room for simplification).

pkg/gardenlet/controller/shoot/shoot_care_control.go
```diff
@@ -134,7 +134,17 @@ func (c *defaultCareControl) Care(shootObj *gardencorev1beta1.Shoot, key string)
...
+	operation, err := operation.
+		NewBuilder().
+		WithLogger(shootLogger).
+		WithConfig(c.config).
+		WithGardenerInfo(c.identity).
+		WithSecrets(c.secrets).
+		WithImageVector(c.imageVector).
+		WithGardenFrom(c.k8sGardenCoreInformers, shoot.Namespace).
+		WithSeedFrom(c.k8sGardenCoreInformers, *shoot.Spec.SeedName).
+		WithShootFrom(c.k8sGardenCoreInformers, shoot).
```
timebertt, line 146: These 3 methods are only used here and in the function they add to the builder, we already have a garden client (see line 147). So I think, we should not need to pass the informer here.
Also, in general it would probably be more convenient to pass only the names to the builder methods and pass the garden client once to `Build()` and propagate it to the respective builder funcs there. @rfranzke @timuthy WDYT?

```suggestion
		WithGardenAndShootNamespace(shoot.Namespace).
		WithSeedName(*shoot.Spec.SeedName).
		WithShootName(shoot).
```

pkg/operation/garden/garden.go
```diff
@@ -39,28 +39,75 @@ import (
...
+
+// WithProject sets the project attribute at the Builder.
+func (b *Builder) WithProject(project *gardencorev1beta1.Project) *Builder {
+	b.projectFunc = func() (*gardencorev1beta1.Project, error) { return project, nil }
+	return b
+}
+
+// WithProjectFromLister sets the project attribute after fetching it from the lister.
+func (b *Builder) WithProjectFromLister(projectLister gardencorelisters.ProjectLister, namespace string) *Builder {
+	b.projectFunc = func() (*gardencorev1beta1.Project, error) {
```
timebertt, line 58:
```suggestion
func (b *Builder) WithProjectNamespace(namespace string) *Builder {
	b.projectFunc = func(ctx context.Context, gardenClient client.Client) (*gardencorev1beta1.Project, error) {
```

pkg/operation/seed/seed.go
```diff
@@ -53,6 +53,68 @@ import (
...
+	b.seedObjectFunc = func() (*gardencorev1beta1.Seed, error) { return seedLister.Get(seedName) }
```
timebertt, line 70:
```suggestion
func (b *Builder) WithSeedName(seedName string) *Builder {
	b.seedObjectFunc = func(ctx context.Context, gardenClient client.Client) (*gardencorev1beta1.Seed, error) {
```

pkg/operation/seed/seed.go
```diff
@@ -53,6 +53,68 @@ import (
...
+	return b
+}
+
+// WithSeedSecret sets the seedSecret attribute at the Builder.
+func (b *Builder) WithSeedSecret(seedSecret *corev1.Secret) *Builder {
```
timebertt, line 75: This method is unused.

pkg/operation/seed/seed.go
```diff
@@ -53,6 +53,68 @@ import (
...
+	b.seedSecretFunc = func(*corev1.SecretReference) (*corev1.Secret, error) { return seedSecret, nil }
+	return b
+}
+
+// WithSeedSecretFromClient sets the seedSecret attribute at the Builder after reading it with the client.
+func (b *Builder) WithSeedSecretFromClient(ctx context.Context, c client.Client) *Builder {
```
timebertt, line 81: Can we remove this method in favor of adding a client to `Build()` and getting the secret from the client as a "default implementation", if no `seedSecretFunc` has been set, or something similar?

pkg/operation/shoot/shoot.go
```diff
@@ -41,82 +40,156 @@ import (
...
+}
+
+// WithCloudProfileObject sets the cloudProfileFunc attribute at the Builder.
+func (b *Builder) WithCloudProfileObject(cloudProfileObject *gardencorev1beta1.CloudProfile) *Builder {
+	b.cloudProfileFunc = func(string) (*gardencorev1beta1.CloudProfile, error) { return cloudProfileObject, nil }
+	return b
+}
+
+// WithCloudProfileObjectFromLister sets the cloudProfileObject attribute at the Builder after fetching it from the given lister.
+func (b *Builder) WithCloudProfileObjectFromLister(cloudProfileLister gardencorelisters.CloudProfileLister) *Builder {
```
timebertt, line 77: Can we remove this method in favor of adding a default implementation to `Build()`, that gets the cloudprofile, if no `cloudProfileFunc` has been set, or something similar?

pkg/operation/shoot/shoot.go
```diff
@@ -41,82 +40,156 @@ import (
...
+}
+
+// WithShootSecret sets the shootSecretFunc attribute at the Builder.
+func (b *Builder) WithShootSecret(secret *corev1.Secret) *Builder {
+	b.shootSecretFunc = func(context.Context, client.Client, string, string) (*corev1.Secret, error) { return secret, nil }
+	return b
+}
+
+// WithShootSecretFromLister sets the cloudProfileObject attribute at the Builder after fetching it from the given lister.
+func (b *Builder) WithShootSecretFromSecretBindingLister(secretBindingLister gardencorelisters.SecretBindingLister) *Builder {
```
timebertt, line 89: Same here

pkg/operation/operation.go
```diff
@@ -50,100 +49,210 @@ import (
...
+
+// WithGarden sets the gardenFunc attribute at the Builder.
+func (b *Builder) WithGarden(g *garden.Garden) *Builder {
+	b.gardenFunc = func(_ map[string]*corev1.Secret) (*garden.Garden, error) { return g, nil }
+	return b
+}
+
+// WithGardenFrom sets the gardenFunc attribute at the Builder which will build a new Garden object.
+func (b *Builder) WithGardenFrom(k8sGardenCoreInformers gardencoreinformers.Interface, namespace string) *Builder {
+	b.gardenFunc = func(secrets map[string]*corev1.Secret) (*garden.Garden, error) {
```
timebertt, line 101:
```suggestion
func (b *Builder) WithGardenAndShootNamespace(namespace string) *Builder {
	b.gardenFunc = func(ctx context.Context, gardenClient client.Client, secrets map[string]*corev1.Secret) (*garden.Garden, error) {
```

timebertt (changes requested):
I have some suggestions to further simplify the added builders.
Mainly, it would be nice to a) have default implementations for e.g. getting the cloud profile and secret binding to remove unnecessary calls to builder methods and b) rather pass a client to `Build()` and remove the clients/informers/listers from the builder methods.

The rest, actually looks good wrt the cluster sync.

pkg/operation/shoot/shoot.go
rfranzke: It's unused in our current code, yes, but isn't it fair to offer it as part of a builder for other packages that might want to use it?

pkg/operation/seed/seed.go
rfranzke: I'm not too sure about how the controller-runtime cache works, maybe you can shed some light on it. I want to prevent that we make unnecessary external calls with the controller-runtime client as we are already watching seeds, etc., and can perfectly read the object from the informers cache. However, as said I'm not sure about the cache of controller-runtime: When does it use it (in contrast to making an actual API call)? We are not watching e.g. seeds using the controller-runtime yet, i.e., do we already have such a cache yet?

pkg/operation/seed/seed.go
rfranzke: See https://github.com/gardener/gardener/pull/2217#discussion_r417070944

pkg/operation/seed/seed.go
rfranzke: The reason behind it was that we wanted to prevent that a caller has to always pass the client to `Build()` (which would be unnecessary if he already has the objects at hand and can simply pass them using `WithSeedObject`, for example.

pkg/operation/shoot/shoot.go
rfranzke: Why?

pkg/operation/shoot/shoot.go
rfranzke: Why?

rfranzke:
> The behaviour I described in gardener/gardener-extension-provider-aws#62 (review) is coming from this PR.
I think the reason is that the Cluster initially is created without .status.lastOperation and this prevents the ShootNotFailed predicate to be satisfied?

Hm maybe, will check. Nice catch anyways. The `ShootNotFailed` predicate can be improved. As a mitigation until all extension have revendored we can probably add a "fake status" to the shoot manifest that is inside the `Cluster` object.

rfranzke:
> Mainly, it would be nice to a) have default implementations for e.g. getting the cloud profile and secret binding to remove unnecessary calls to builder methods

Default implementations require that certain arguments are always passed to `Build()` (e.g., the client). We thought/discussed about it and didn't like it too much. For example, when you already have the `CloudProfile` and use the `WithCloudProfileObject` builder function, and the only purpose of the context/client passed to `Build()` is to have a default implementation then it's the opposite of convenient for the caller. Consequently, we introduced `WithCloudProfileFrom` where clients/listers can be passed only if it needs to be read.

> and b) rather pass a client to Build() and remove the clients/informers/listers from the builder methods.

Depends on https://github.com/gardener/gardener/pull/2217#discussion_r417071667

pkg/operation/shoot/shoot.go
timebertt: Yep, sure.

pkg/operation/shoot/shoot.go
timebertt: FMPOV, calling this method and passing only an extra lister is unnecessary, as you have to pass a client to `Build` anyways. If you call this method, you're basically expressing, that you don't have the `CloudProfile` object yet and want it to be retrieved on `Build()`. This could be done implicitly (by not calling `WithCloudProfileObject`) to remove the duplication of clients.

rfranzke: See https://github.com/gardener/gardener/pull/2217#issuecomment-620994942 for a comment about default implementations

timebertt: Yes, but the shoot builder's `Build` method anyways takes a client.

timebertt:
> > Mainly, it would be nice to a) have default implementations for e.g. getting the cloud profile and secret binding to remove unnecessary calls to builder methods
> 
> Default implementations require that certain arguments are always passed to `Build()` (e.g., the client). We thought/discussed about it and didn't like it too much. For example, when you already have the `CloudProfile` and use the `WithCloudProfileObject` builder function, and the only purpose of the context/client passed to `Build()` is to have a default implementation then it's the opposite of convenient for the caller. Consequently, we introduced `WithCloudProfileFrom` where clients/listers can be passed only if it needs to be read.

Well, good point. I agree with you, that wouldn't be convenient.
~Would it be an option to add a `WithClient` method to the builder, to pass a Client for all the default implementations? (Yes, the context would have to be stored in the builder, which is bad practice...)
If not, I would keep passing the client to the builder methods, but maybe streamline all builders to use controller-runtime clients.~
Edited: obsolete with https://github.com/gardener/gardener/pull/2217#discussion_r417100845

pkg/operation/seed/seed.go
timebertt: Sorry, it was a misconception from my side, that the controller-runtime client for the garden cluster uses a cache (it only caches rest clients, but not API objects). That way, we would indeed cause a lot of unnecessary external API calls, which we should prevent.

So, you are probably right, using informers/listers is our best guess for now.
But as soon, as we have cached controller-runtime clients (which we will need anyways, as we are making more and more requests through them), we can probably switch all of these builder methods to controller-runtime clients, to unify and further simplify.

pkg/gardenlet/controller/shoot/shoot_care_control.go
timebertt: obsolete with https://github.com/gardener/gardener/pull/2217#discussion_r417100845

pkg/operation/garden/garden.go
timebertt: obsolete with https://github.com/gardener/gardener/pull/2217#discussion_r417100845

pkg/operation/seed/seed.go
timebertt: obsolete with https://github.com/gardener/gardener/pull/2217#discussion_r417100845

pkg/operation/seed/seed.go
timebertt: Sure, good point.

pkg/operation/shoot/shoot.go
timebertt: Also obsolete with https://github.com/gardener/gardener/pull/2217#discussion_r417100845, as we rather want to use the lister's cache instead of making unnecessary external API calls.

pkg/operation/operation.go
timebertt: obsolete with https://github.com/gardener/gardener/pull/2217#discussion_r417100845

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -389,22 +503,42 @@ func (c *Controller) reconcileShoot(shoot *gardencorev1beta1.Shoot, logger *logr
 		}
 	}
 
-	if err := c.runReconcileShootFlow(o, operationType); err != nil {
+	if err := c.runReconcileShootFlow(o); err != nil {
```
danielfoehrKn, line 523: If there is an error during reconcile, it will return an error ->  leads to another reconcile -> syncs the Cluster resource containing the Shoot to the Seed. Is that the reason for not syncing the Cluster resource to the Seed here?

rfranzke: Hm, you are asking because we are syncing the `Cluster` resource again when the reconciliation is successful but not when it is errorneous, right? Yes, basically, it would be more consistent to sync it here as well. @timuthy WDYT?

timuthy: Good point but I wouldn't probably do it, at least not just for consistency reasons. The cluster resource is a rather big resource and an extra sync does not seem beneficial.

pkg/operation/seed/seed.go
rfranzke: Yes, +1, sounds great, thanks.

rfranzke:
Can you take another look?

@ialidzhikov for the problem you encountered I added https://github.com/gardener/gardener/pull/2217/files#diff-8f27551340700250a13bbbe0ea0507b0R143-R154 in the `shoot_control.go`,  I hope it helps until #2265 is revendored by the extensions. WDYT?

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/sirupsen/logrus"
```
timebertt: Move down to 3rd party import block

timebertt (commented):
Nit, otherwise /lgtm

timebertt (approved):

ialidzhikov:
> @ialidzhikov for the problem you encountered I added https://github.com/gardener/gardener/pull/2217/files#diff-8f27551340700250a13bbbe0ea0507b0R143-R154 in the shoot_control.go, I hope it helps until #2265 is revendored by the extensions. WDYT?

I still reproduce it with the master branch. So I think we have to revendor the extensions

PR #2220, opened 2020-04-22
Adding gVisor Integration tests 

danielfoehrKn:
Adding integration tests for the gVisor container runtime integration in Gardener.

To test for the first release of [gVisor extension](https://github.com/gardener/gardener-extension-runtime-gvisor) .

Fixes #

Has the same suboptimal behaviour as discussed in this [PR](https://github.com/gardener/gardener/pull/2201#discussion_r411907231) - is restricted to Ubuntu OS. 
I think we can still go ahead with this PR as the already existing containerd integration test is configured the same way. As soon as there is a solution, these two integration tests can be easily adjusted. 

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/apis/core/types_shoot.go
+++ b/pkg/apis/core/types_shoot.go
@@ -723,6 +723,13 @@ const (
 	CRINameContainerD CRIName = "containerd"
 )
 
+// ContainerRuntimeName is a type alias for the ContainerRuntime name string.
+type ContainerRuntimeName string
+
+const (
+	ContainerRuntimeGVisor ContainerRuntimeName = "gvisor"
+)
+
 // ContainerRuntime contains information about worker's available container runtime
 type ContainerRuntime struct {
 	// Type is the type of the Container Runtime.
--- a/test/integration/shoots/operations/containerruntime.go
+++ b/test/integration/shoots/operations/containerruntime.go
@@ -17,6 +17,8 @@ package operations
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
+	"time"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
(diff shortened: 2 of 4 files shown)


test/integration/shoots/operations/containerruntime.go
```diff
@@ -107,19 +86,240 @@ var _ = ginkgo.Describe("Shoot container runtime testing", func() {
...
 	response, err := rootPodExecutor.Execute(ctx, command)
 	framework.ExpectNoError(err)
 	g.Expect(response).ToNot(g.BeNil())
 	g.Expect(string(response)).To(g.Equal(fmt.Sprintf("%s\n", expected)))
 }
+
+// deployGVisorPod deploys a pod using the gVisor RuntimeClass.
+func deployGVisorPod(ctx context.Context, c client.Client) (*corev1.Pod, error) {
+	allowedCharacters := "0123456789abcdefghijklmnopqrstuvwxyz"
+	id, err := gardenerutils.GenerateRandomStringFromCharset(3, allowedCharacters)
```
ialidzhikov, line 291: You could use .metadata.generateName for the Pod name.

ialidzhikov (changes requested):

test/integration/shoots/operations/containerruntime.go
danielfoehrKn: true, then I ll also adjust that in the test-framework where it is used similarely.

pkg/apis/core/types_shoot.go
```diff
@@ -723,6 +723,13 @@ const (
 	CRINameContainerD CRIName = "containerd"
 )
 
+// ContainerRuntimeName is a type alias for the ContainerRuntime name string.
+type ContainerRuntimeName string
+
+const (
+	ContainerRuntimeGVisor ContainerRuntimeName = "gvisor"
```
timebertt, line 730: Why is this here?
I don't think, there should be any known provider names in Gardener's API...
Btw, isn't this a clear case for an integration test, that should be part of the respective Extension repo?

dguendisch (approved):
integrationtest-wise lgtm

test/integration/shoots/operations/containerruntime.go
```diff
@@ -107,19 +86,240 @@ var _ = ginkgo.Describe("Shoot container runtime testing", func() {
...
+				{
+					Name:  "gvisor-container",
+					Image: "busybox",
+					Command: []string{
+						"sleep",
+						"10000000",
+					},
+					Resources:                corev1.ResourceRequirements{},
+					TerminationMessagePath:   "/dev/termination-log",
+					TerminationMessagePolicy: corev1.TerminationMessageReadFile,
```
ialidzhikov, line 314: These are the default values for `Resources`, `TerminationMessagePath` and  `TerminationMessagePolicy `, right? You could remove them.

danielfoehrKn:
Decided that this integration test rather belongs [in the respective extension repo](https://github.com/gardener/gardener-extension-runtime-gvisor).

danielfoehrKn:
Opened in the [gVisor-extension.](https://github.com/gardener/gardener-extension-runtime-gvisor/pull/6)

PR #2222, opened 2020-04-23
Refactor interaction with extension CRs

rfranzke:
We had duplicated the logic for deletion, waiting for readiness, and waiting for deletion multiple times for the various extension resources. This PR maintains the relevant code only once and makes it callable/reusable to prevent this duplication.

/kind cleanup

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/common/extensions.go
+++ b/pkg/operation/common/extensions.go
@@ -1,0 +1,263 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/gardener/gardener/pkg/api/extensions"
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/utils/flow"
+	"github.com/gardener/gardener/pkg/utils/kubernetes/health"
+	"github.com/gardener/gardener/pkg/utils/retry"
+
+	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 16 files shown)


danielfoehrKn (commented):
Tested and looks good so far.

pkg/operation/common/extensions.go
```diff
@@ -0,0 +1,263 @@
...
+	ctx context.Context,
+	c client.Client,
+	logger *logrus.Entry,
+	newObjFunc func() runtime.Object,
+	kind string,
+	namespace string,
+	name string,
+	interval time.Duration,
+	timeout time.Duration,
+	postReadyFunc func(runtime.Object) error,
```
timuthy, line 50: The `postReadyFunc` seems like a step that isn't necessary. What do you think about leaving out `postReadyFunc`  and instead pass the `runtime.Object` in favor of `newObjFunc`, i.e. the object is then filled by `WaitUntilExtensionCRReady`. Afterwards the caller has the correctly filled `obj` and calls any post steps by himself?

pkg/operation/common/extensions.go
```diff
@@ -0,0 +1,263 @@
...
+		if !ok {
+			return fmt.Errorf("expected extensionsv1alpha1.Object but got %T", obj)
+		}
+
+		if o.GetDeletionTimestamp() == nil {
+			return nil
+		}
+
+		if predicateFunc != nil && predicateFunc(o) {
+			fns = append(fns, func(ctx context.Context) error {
```
timuthy, line 192: I know that you didn't change the logic but I wondered if we can operate on the list directly instead of adding funcs which again wait for a single deletion?

timuthy (changes requested):
Looks good in general, only a few conceptional suggestions.

pkg/operation/common/extensions.go
rfranzke: I was doing it like this earlier, but the problem was that for the `waitUntil*` functions we always should have a new object into which we load the actual state from the server. Hence, the `newObjFunc` was added in favor of the `runtime.Object`.

pkg/operation/common/extensions.go
rfranzke: Hm, what do you suggest concretely when you say "operate on the list directly"?

pkg/operation/common/extensions.go
timuthy: I think this is also up to the caller. It's the same for the Controller-Runtime client funcs. The `newObjFunc` isn't a strong prevention mechanism, as it can also return any `runtime.Object`.

rfranzke: I faced a problem when I did it like this. Concretely, e.g. in the infrastructure reconcilation the operation annotation is added. While waiting for it to be ready it fetched it the first time while the operation annotation was still present. Subsequent fetches (where the operation annotation was not present anymore) didn't overwrite the object correctly. Only when always instantiating a new object (like it was before) it was functioning correctly.

pkg/operation/common/extensions.go
timuthy: Like getting the list --> apply the predicates --> if elements still remain, run the check again. So my proposal is to use the elements from the list instead of retrieving them separately again.

pkg/operation/common/extensions.go
timuthy: Sure, thanks for explaining. I didn't consider this case.

pkg/operation/common/extensions.go
rfranzke: Ok, understood, thanks. I tried it, but I ended up having again two places where the `lastError` of the extension object is evaluated. I can introduce helper functions for that, however, as this PR is a preparation for follow-up work that properly evaluates the error codes and I don't know how this will look like, would it be okay to keep it like this for now? I keep it in mind and check whether it can be implemented after the error codes improvements.

timuthy: Yep, sound reasonable. Thanks for your effort checking this.

timuthy (approved):
Thanks for addressing my comments. With the tests from @danielfoehrKn should be good to merge.

PR #2223, opened 2020-04-23
Fix bug in CheckCSIConditions function

rfranzke:
This PR fixes a bug in the `csimigration.CheckCSIConditions` function.
If the `csiMigrationVersion` is for example `1.18` then the `isHigherThanCSIMigrationVersion` is only `false` when the provided shoot Kubernetes version is `<= 1.18.0`. If it is `1.18.1` then it is `true` which means that it immediately returns `true` for `csiMigrationComplete` instead of checking if the annotation is present on the cluster.

/kind bug

**Release note**:

```improvement developer
A bug in the CSI migration controller has been fixed that may cause the `CSIMigration<Provider>Complete` feature gate to be set to early.
```


ialidzhikov (approved):
/lgtm

PR #2229, opened 2020-04-24
Remove shoot tasks only if executed

### LINKED ISSUE #2228: Controller restart in maintenance time window does not work
**What happened**:
With #2098, #2116 a feature was contributed that enables auto-restart of controllers in the shoot maintenance time window. However, it seems that there is a bug. I'm having a shoot that was reconciled `8h` ago: 

```yaml
  lastOperation:
    description: Shoot cluster state has been successfully reconciled.
    lastUpdateTime: '2020-04-23T21:01:23Z'
    progress: 100
    state: Succeeded
    type: Reconcile
```

However, the {kube,cloud}-controller-managers run since 31h:

```shell
$ kubectl -n shoot--foo--bar get pod | grep controller-m
cloud-controller-manager-749ccfc56f-ndjt7     1/1     Running     0          31h
kube-controller-manager-84d47674b6-4w8pw      1/1     Running     0          31h
```

The controller-managers should have been restarted `8h` ago. The 

Create a shoot and enable the `enableShootControlPlaneRestarter` flag in GCM config:

```yaml
    apiVersion: controllermanager.config.gardener.cloud/v1alpha1
    kind: ControllerManagerConfiguration
    controllers:
      shootMaintenance:
        concurrentSyncs: 200
        enableShootControlPlaneRestarter: true
```

/cc @timuthy 

**Environment**:

- Gardener version: v1.3.0
- Kubernetes version (use `kubectl version`): v1.14.0
- Cloud provider or hardware configuration: AWS

rfranzke:
The shoot reconciliation flow was removing the tasks annotation at the end of a successful shoot reconciliation, no matter whether it actually respected it in this shoot reconciliation.
This lead to race conditions if new tasks were added while an existing reconciliation is in progress. Concretely, sometimes the shoot-maintenance was not executed completely. The shoot reconciliation started a second time, however, without any tasks as they were removed from the first reconcilation.
This PR changes the behaviour so that only tasks are removed if they were really executed.

Fixes #2228

**Release note**:

```improvement operator
A race condition that led to incomplete maintenance operations for shoot clusters has been fixed.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -84,14 +84,16 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
 	}
 
 	var (
-		defaultTimeout                 = 30 * time.Second
-		defaultInterval                = 5 * time.Second
-		dnsEnabled                     = !o.Shoot.DisableDNS
-		managedExternalDNS             = o.Shoot.ExternalDomain != nil && o.Shoot.ExternalDomain.Provider != "unmanaged"
-		managedInternalDNS             = o.Garden.InternalDomain != nil && o.Garden.InternalDomain.Provider != "unmanaged"
-		allowBackup                    = o.Seed.Info.Spec.Backup != nil
-		staticNodesCIDR                = o.Shoot.Info.Spec.Networking.Nodes != nil
-		restartControlPlaneControllers = controllerutils.HasTask(o.Shoot.Info.Annotations, common.ShootTaskRestartControlPlanePods)
+		defaultTimeout     = 30 * time.Second
+		defaultInterval    = 5 * time.Second
+		dnsEnabled         = !o.Shoot.DisableDNS
+		managedExternalDNS = o.Shoot.ExternalDomain != nil && o.Shoot.ExternalDomain.Provider != "unmanaged"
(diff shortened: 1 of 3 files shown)


pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -381,6 +383,27 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation, operationType
...
+	newShoot, err := kutil.TryUpdateShootAnnotations(c.k8sGardenClient.GardenCore(), retry.DefaultRetry, o.Shoot.Info.ObjectMeta,
+		func(shoot *gardencorev1beta1.Shoot) (*gardencorev1beta1.Shoot, error) {
+			var tasksToRemove []string
+
+			if requestInfrastructureReconciliation {
+				tasksToRemove = append(tasksToRemove, common.ShootTaskDeployInfrastructure)
+			}
+			if requestControlPlanePodsRestart {
+				tasksToRemove = append(tasksToRemove, common.ShootTaskRestartControlPlanePods)
+			}
```
timuthy: What do you think about saving all tasks before running the flow and looping through them here instead of doing this only for these specific ones?

rfranzke: +1, will do

rfranzke: done, PTAL

timuthy (approved):
/lgtm

PR #2233, opened 2020-04-24
Confine shoot spec update rollout to maintenance

### LINKED ISSUE #2196: Optionally roll out Shoot spec changes only in the maintenance time window
We should make it optionally configurable in the `Shoot` that changes to the `spec` are only applied during the Shoot's maintenance time window.
Let's introduce `.spec.maintenance.confineSpecUpdateRollout` (defaults to `false`). If set to `true` then changes to the `spec` won't trigger the shoot reconciliation flow. However, setting the `gardener.cloud/operation=reconcile` annotation shall enforce the rollout immediately (despite of this flag).

**Why is this needed**:
Some users want to update their Shoot specifications, however, they don't want to roll out the changes immediately but in their maintenance time window (in case the update results in a rolling update of the worker nodes or restart of etcd, for example).

rfranzke:
This PR adds a new optional field `.spec.maintenance.confineSpecUpdateRollout` to the `Shoot`. If it is set to `true` then any update to the spec will not cause a reconciliation directly. Internally, this is achieved by not incrementing the `.metadata.generation`. Instead, the changes are rolled out by the reconciliation that happens in the shoot's maintenance time window anyways.

Fixes #2196

* I added docs for shoot maintenance along the way. 
* Also, you might ask yourself (like I asked myself 😉) whether it's really a good idea to not bumped the `.metadata.generation` to achieve this. I was iterating back and forth, but eventually I decided for "yes". The main reasons were:
  * a) `.metadata.generation` is mainly a measurement for controllers to detect whether they have to reconcile. As we control both (API server + controller) we can keep this knowledge in sync.
  * b) End-users might want to compare `.metadata.generation` and `.status.observedGeneration` to check whether their changes got already applied. However, when they set the `.spec.maintenance.confineSpecUpdateRollout=true` then they know already that the changes won't get applied immediately, so it's no issue.
  * c) This was actually my main concern: Convention-wise this approach is violating, however, we are already "violating" (I'd rather call it "customizing") when we bump the generation. IMO this is the advantage of developing and maintaing our own API server (instead of using CRDs).

**Release note**:

```noteworthy user
It is now possible to confine rollouts of changes/updates to the `Shoot` specification to the individual maintenance time window. You can set `.spec.maintenance.confineSpecUpdateRollout=true` to achieve the desired behaviour. Please consult [this document](https://github.com/rfranzke/gardener/blob/feature/confine-spec-updates/docs/usage/shoot_maintenance.md#confine-specification-changesupdates-roll-out) to get all information about it.
```


docs/usage/shoot_maintenance.md
```diff
@@ -0,0 +1,80 @@
...
+Via the `.spec.maintenance.confineSpecUpdateRollout` field you can control whether you want to make Gardener roll out changes/updates to your shoot specification only during the maintenance time window.
+It is `false` by default, i.e., any change to your shoot specification triggers a reconciliation (even outside of the maintenance time window).
+This is helpful if you want to update your shoot but don't want the changes to be applied immediately. One example use-case would be a Kubernetes version upgrade that you want to roll out during the maintenance time window.
+Any update to the specification will not increase the `.metadata.generation` of the `Shoot` which is something you should be aware of.
+Also, even if Gardener administrators/operators have not enabled the "reconciliation in maintenance time window only" configuration (as mentioned above) then your shoot will only reconcile in the maintenance time window.
+The reason is that Gardener cannot differentiate between create/update/reconcile operations.
+
+⚠️  If `confineSpecUpdateRollout=true`, please note that if you change the maintenance time window itself then it will only be considered in the next but one maintenance.
```
timuthy:
```suggestion
⚠️  If `confineSpecUpdateRollout=true`, please note that if you change the maintenance time window itself then it will only be effective after the upcoming maintenance.
```

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -94,6 +94,10 @@ func (c *Controller) respectSyncPeriodOverwrite() bool {
 	return controllerutils.BoolPtrDerefOr(c.config.Controllers.Shoot.RespectSyncPeriodOverwrite, false)
 }
 
+func (c *Controller) confineSpecUpdateRollout(maintenance *gardencorev1beta1.Maintenance) bool {
```
timuthy:
```suggestion
func confineSpecUpdateRollout(maintenance *gardencorev1beta1.Maintenance) bool {
```

timuthy (changes requested):
Looks good in general, just minor suggestions in the code. I didn't test the changes but I'm afraid that this change breaks the scheduled hibernation feature, doesn't it? If yes, the least we should do is to prevent both features being used at the same time. WDYT?

rfranzke:
> I didn't test the changes but I'm afraid that this change breaks the scheduled hibernation feature, doesn't it? If yes, the least we should do is to prevent both features being used at the same time. WDYT?

Nice catch! Yes, it would break it, indeed. We could forbid it per validation, right. The other option would be to make an exception for hibernation/wake-up and roll them out immediately as those operations are disruptive anyways (and there isn't a use-case where people would want their shoots from being hibernated/woken-up in the maintenance time window only (and even if they can achieve that more accurately by defining a hibernation schedule)). The only downside would be an inconsistency, though, on the other hand hibernation is a strong use-case and preventing it in general might be too harsh, probably unnecessary harsh.

timuthy:
> We could forbid it per validation, right. The other option would be to make an exception for hibernation/wake-up and roll them out immediately

I went back and forth with thoughts about this and tend to vote for the second option as well. Let's mention it in the documentation and ideally also in the doc string of the `confineSpecUpdateRollout` field.

pkg/registry/core/shoot/strategy.go
```diff
@@ -143,6 +144,14 @@ func mustIncreaseGeneration(oldShoot, newShoot *core.Shoot) bool {
 	return false
 }
 
+func mustIncreaseGenerationForSpecChanges(oldShoot, newShoot *core.Shoot) bool {
+	if newShoot.Spec.Maintenance != nil && newShoot.Spec.Maintenance.ConfineSpecUpdateRollout != nil && *newShoot.Spec.Maintenance.ConfineSpecUpdateRollout {
+		return !apiequality.Semantic.DeepEqual(helper.HibernationIsEnabled(oldShoot), helper.HibernationIsEnabled(newShoot))
```
timuthy: Any special reason you use `DeepEqual`?

```suggestion
		return helper.HibernationIsEnabled(oldShoot) != helper.HibernationIsEnabled(newShoot)
```

timuthy (changes requested):
Thanks for addressing my comments. Just one Nit, but looks good apart from that.

timuthy (approved):

PR #2236, opened 2020-04-27
Fix the default dns assignment issue of `metadata.generateName`

### LINKED ISSUE #1783: Support `metadata.generateName` as alternative to `metadata.name`
**What happened**:
I tried to create a shoot with generated name and got an error.
```
apiVersion: garden.sapcloud.io/v1beta1
kind: Shoot
metadata:
  generateName: prefix  <-------------
  annotations:
    garden.sapcloud.io/purpose: evaluation
spec: <shoot spec>
```

```
error: error when retrieving current configuration of:
Resource: "garden.sapcloud.io/v1beta1, Resource=shoots", GroupVersionKind: "garden.sapcloud.io/v1beta1, Kind=Shoot"
Name: "", Namespace: "<project-namespace>"
Object: ...
from server for: "generate-shoot-name/shoot.yml": resource name may not be empty
```

I expected to get a Shoot whose name is generated with "prefix" as a prefix.

1. Login to the Gardener Dashboard
1. Click the "+" sign to create a Shoot
1. Switch to YAML tab
1. remove `metadata.name` and replace it with a `metadata.generateName` key

**Environment**:

- Gardener version: 1.35.4
- Kubernetes version (use `kubectl version`): n/a
- Cloud provider or hardware configuration: n/a
- Others:

swapnilgm:
- - Shoot name length limit restriction is applied on `generateName` with random suffix length fixed to 5 as in the kubernetes. 
- Default DNS  name is generated using same name generator for shoot. But it will differ from generated shoot name, cause the generated name is not available at the time DNS admission. And generator function will generate different name in each attempt, So, Dns name and shoot name will differ. Currently i don't see the good way to sync these two. Suggestions are welcome.

Fixes #1783 

**Release note**:

```improvement operator
Issues for Shoot with `metadata.generateName` are fixed. Shoot name length limit restriction is applied on `generateName` with random suffix length fixed to 5 as in the kubernetes. Default DNS  name is generated using same name generator for shoot. ⚠️ But it will differ from generated shoot name.  
```

### DIFF
--- a/plugin/pkg/shoot/dns/admission.go
+++ b/plugin/pkg/shoot/dns/admission.go
@@ -29,6 +29,7 @@ import (
 	coreinformers "github.com/gardener/gardener/pkg/client/core/informers/internalversion"
 	corelisters "github.com/gardener/gardener/pkg/client/core/listers/core/internalversion"
 	"github.com/gardener/gardener/pkg/operation/common"
+	gardenerutils "github.com/gardener/gardener/pkg/utils"
 	admissionutils "github.com/gardener/gardener/plugin/pkg/utils"
 
 	corev1 "k8s.io/api/core/v1"
@@ -289,6 +290,13 @@ func assignDefaultDomainIfNeeded(shoot *core.Shoot, projectLister corelisters.Pr
 		if shootDomain != nil && strings.HasSuffix(*shootDomain, "."+domain) {
 			// Check that the specified domain matches the pattern for default domains, especially in order
 			// to prevent shoots from "stealing" domain names for shoots in other projects
+			if len(shoot.GenerateName) > 0 && (len(shoot.Name) == 0 || strings.HasPrefix(shoot.Name, shoot.GenerateName)) {
+				// Case where shoot name is generated or to be generated
+				if !strings.HasSuffix(*shootDomain, fmt.Sprintf(".%s.%s", project.Name, domain)) {
+					return apierrors.NewBadRequest("shoot with 'metadata.generateName' uses a default domain but does not match expected scheme: <random-subdomain>.<project-name>.<default-domain>")
+				}
+				return nil
+			}
 			if *shootDomain != fmt.Sprintf("%s.%s.%s", shoot.Name, project.Name, domain) {
 				return apierrors.NewBadRequest("shoot uses a default domain but does not match expected scheme: <shoot-name>.<project-name>.<default-domain>")
 			}
@@ -301,7 +309,14 @@ func assignDefaultDomainIfNeeded(shoot *core.Shoot, projectLister corelisters.Pr
 			if shoot.Spec.DNS == nil {
 				shoot.Spec.DNS = &core.DNS{}
 			}
-			generatedDomain := fmt.Sprintf("%s.%s.%s", shoot.Name, project.Name, domain)
+			shootDNSName := shoot.Name
+			if len(shoot.Name) == 0 && len(shoot.GenerateName) > 0 {
+				shootDNSName, err = gardenerutils.GenerateRandomStringFromCharset(len(shoot.GenerateName)+5, "0123456789abcdefghijklmnopqrstuvwxyz")
+				if err != nil {
+					return apierrors.NewInternalError(err)
+				}
+			}
+			generatedDomain := fmt.Sprintf("%s.%s.%s", shootDNSName, project.Name, domain)
 			shoot.Spec.DNS.Domain = &generatedDomain
 			return nil
 		}
--- a/plugin/pkg/shoot/dns/admission_test.go
+++ b/plugin/pkg/shoot/dns/admission_test.go
@@ -471,6 +471,91 @@ var _ = Describe("dns", func() {
 
 				Expect(err).To(MatchError(apierrors.NewBadRequest("shoot domain field .spec.dns.domain must be set if provider != unmanaged and assigned to a seed which does not disable DNS")))
 			})
+
+			Context("#Shoot GenerateName used", func() {
+				BeforeEach(func() {
+					shoot.Name = ""
+					shoot.GenerateName = "demo-"
+				})
+
+				It("should set different default domain for multiple shoots with same generate name", func() {
+					shootCopy := shoot.DeepCopy()
+
+					Expect(kubeInformerFactory.Core().V1().Secrets().Informer().GetStore().Add(&defaultDomainSecret)).To(Succeed())
+					Expect(coreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&project)).To(Succeed())
+					Expect(coreInformerFactory.Core().InternalVersion().Seeds().Informer().GetStore().Add(&seed)).To(Succeed())
+					attrs := admission.NewAttributesRecord(&shoot, nil, core.Kind("Shoot").WithVersion("version"), shoot.Namespace, shoot.Name, core.Resource("shoots").WithVersion("version"), "", admission.Create, &metav1.CreateOptions{}, false, nil)
+
+					err := admissionHandler.Admit(context.TODO(), attrs, nil)
+					Expect(err).To(Not(HaveOccurred()))
+
+					Expect(kubeInformerFactory.Core().V1().Secrets().Informer().GetStore().Add(&defaultDomainSecret)).To(Succeed())
+					Expect(coreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&project)).To(Succeed())
+					Expect(coreInformerFactory.Core().InternalVersion().Seeds().Informer().GetStore().Add(&seed)).To(Succeed())
+					attrs = admission.NewAttributesRecord(shootCopy, nil, core.Kind("Shoot").WithVersion("version"), shootCopy.Namespace, shootCopy.Name, core.Resource("shoots").WithVersion("version"), "", admission.Create, &metav1.CreateOptions{}, false, nil)
+
+					err = admissionHandler.Admit(context.TODO(), attrs, nil)
+
+					Expect(err).To(Not(HaveOccurred()))
+
+					Expect(*shoot.Spec.DNS.Domain).NotTo(Equal(*shootCopy.Spec.DNS.Domain))
+				})
+
+				It("should generate a default domain with shoot name for the shoot (no domain)", func() {
+					shoot.Name = "foo"
+					Expect(kubeInformerFactory.Core().V1().Secrets().Informer().GetStore().Add(&defaultDomainSecret)).To(Succeed())
+					Expect(coreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&project)).To(Succeed())
+					Expect(coreInformerFactory.Core().InternalVersion().Seeds().Informer().GetStore().Add(&seed)).To(Succeed())
+					attrs := admission.NewAttributesRecord(&shoot, nil, core.Kind("Shoot").WithVersion("version"), shoot.Namespace, shoot.Name, core.Resource("shoots").WithVersion("version"), "", admission.Create, &metav1.CreateOptions{}, false, nil)
+
+					err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+					Expect(err).NotTo(HaveOccurred())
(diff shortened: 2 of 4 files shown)


rfranzke:
Converting to "draft PR" as it has "WIP" label

plugin/pkg/shoot/validator/admission.go
```diff
@@ -200,8 +200,16 @@ func (v *ValidateShoot) Admit(ctx context.Context, a admission.Attributes, o adm
 		// project name. These checks should only be performed for CREATE operations (we do not want to reject changes to existing
 		// Shoots in case the limits are changed in the future).
 		var lengthLimit = 21
-		if len(project.Name+shoot.Name) > lengthLimit {
-			return apierrors.NewBadRequest(fmt.Sprintf("the length of the shoot name and the project name must not exceed %d characters (project: %s; shoot: %s)", lengthLimit, project.Name, shoot.Name))
+		//TODO: Check if this is still the case. Document the resource naming convention on infra and related limits.
```
rfranzke: Yes, this is still the case. :)

plugin/pkg/shoot/dns/admission.go
```diff
@@ -301,7 +302,11 @@ func assignDefaultDomainIfNeeded(shoot *core.Shoot, projectLister corelisters.Pr
 			if shoot.Spec.DNS == nil {
 				shoot.Spec.DNS = &core.DNS{}
 			}
-			generatedDomain := fmt.Sprintf("%s.%s.%s", shoot.Name, project.Name, domain)
+			shootName := shoot.Name
+			if len(shoot.Name) == 0 && len(shoot.GenerateName) > 0 {
+				shootName = shootregistry.Strategy.NameGenerator.GenerateName(shoot.GenerateName)
```
rfranzke: Hm, this name might be different from the name of the actual shoot later on, right? The `GenerateName()` function will be executed again and probably generate a different result?
Should we simply compute a random string here instead of using `.metadata.generateName`?

plugin/pkg/shoot/validator/admission.go
```diff
@@ -200,8 +200,16 @@ func (v *ValidateShoot) Admit(ctx context.Context, a admission.Attributes, o adm
...
+		if len(shoot.Name) == 0 && len(shoot.GenerateName) > 0 {
+			var randomLength = 5
+			if len(project.Name+shoot.GenerateName) > lengthLimit-randomLength {
```
rfranzke, line 205: Looks good!

rfranzke (commented):
Why there is an empty commit "https://github.com/gardener/gardener/pull/2236/commits/eedf4de620a742fbfacc5e97a67401d88e493910"?

Also, please add test cases. :)

plugin/pkg/shoot/dns/admission.go
swapnilgm: > Hm, this name might be different from the name of the actual shoot later on, right? The `GenerateName()` function will be executed again and probably generate a different result?

Yes. Correct. I digged in check for different hooks on resource admission, but didn't find any solution to get Object with generateName or atleast exact generated name of shoot. 

> Should we simply compute a random string here instead of using `.metadata.generateName`?

Bringing this discussion up was one of the reason for marking this PR WIP. I'm fine with generating random string here itself with generateName prefix. But i thought using name generator would be better idea. Since it will try to follow the same pattern for Shoot name. And it will be single point of code change, if we feel like introducing separate implementation of name generator considering the name validation is dependent on project name.

Do you still want me to change it?

rfranzke: Not sure, I'd be confused as end-user if my shoot name ends up with `shoot-foo12` but the DNS is `shoot-bar89`. Then it'd be better to simply call the DNS name `asdjnjknasd28` instead of tampering with the shoot name.
I don't get the point wrt "single point of code change" - what do you mean by that?

swapnilgm: Makes sense. Then I will generate random string.

>I don't get the point wrt "single point of code change" - what do you mean by that?
Ignore it.

plugin/pkg/shoot/dns/admission.go
```diff
@@ -289,6 +290,13 @@ func assignDefaultDomainIfNeeded(shoot *core.Shoot, projectLister corelisters.Pr
 		if shootDomain != nil && strings.HasSuffix(*shootDomain, "."+domain) {
 			// Check that the specified domain matches the pattern for default domains, especially in order
 			// to prevent shoots from "stealing" domain names for shoots in other projects
+			if len(shoot.GenerateName) > 0 && (len(shoot.Name) == 0 || strings.HasPrefix(shoot.Name, shoot.GenerateName)) {
```
rfranzke, line 293: Hm, what is the second case here (`strings.HasPrefix(shoot.Name, shoot.GenerateName)`)? Can you explain? If both `.metadata.generateName` and `metadata.name` are specified then `.metadata.name` will win (i.e., we would have to execute the below/already existing check). What is the `HasPrefix` case?

plugin/pkg/shoot/dns/admission.go
```diff
@@ -301,7 +309,15 @@ func assignDefaultDomainIfNeeded(shoot *core.Shoot, projectLister corelisters.Pr
 			if shoot.Spec.DNS == nil {
 				shoot.Spec.DNS = &core.DNS{}
 			}
-			generatedDomain := fmt.Sprintf("%s.%s.%s", shoot.Name, project.Name, domain)
+			shootDNSName := shoot.Name
+			if len(shoot.Name) == 0 && len(shoot.GenerateName) > 0 {
+				random, err := gardenerutils.GenerateRandomStringFromCharset(15, "0123456789abcdefghijklmnopqrstuvwxyz")
```
rfranzke: Why "15"? Shouldn't it be `len(shoot.GenerateName)+5`?

plugin/pkg/shoot/dns/admission.go
```diff
@@ -301,7 +309,15 @@ func assignDefaultDomainIfNeeded(shoot *core.Shoot, projectLister corelisters.Pr
...
+				if err != nil {
+					return apierrors.NewInternalError(err)
+				}
+				shootDNSName = fmt.Sprintf("shoot-%s", random)
```
rfranzke: I don't think we need this prefix here. let's just take the random name computed above.

plugin/pkg/shoot/validator/admission.go
```diff
@@ -200,8 +200,15 @@ func (v *ValidateShoot) Admit(ctx context.Context, a admission.Attributes, o adm
 		// project name. These checks should only be performed for CREATE operations (we do not want to reject changes to existing
 		// Shoots in case the limits are changed in the future).
 		var lengthLimit = 21
-		if len(project.Name+shoot.Name) > lengthLimit {
-			return apierrors.NewBadRequest(fmt.Sprintf("the length of the shoot name and the project name must not exceed %d characters (project: %s; shoot: %s)", lengthLimit, project.Name, shoot.Name))
+		if len(shoot.Name) == 0 && len(shoot.GenerateName) > 0 {
+			var randomLength = 5
+			if len(project.Name+shoot.GenerateName) > lengthLimit-randomLength {
+				return apierrors.NewBadRequest(fmt.Sprintf("the length of the shoot generateName and the project name must not exceed %d characters (project: %s; shoot with generateName: %s)", lengthLimit, project.Name, shoot.GenerateName))
```
```suggestion
				return apierrors.NewBadRequest(fmt.Sprintf("the length of the shoot generateName and the project name must not exceed %d characters (project: %s; shoot with generateName: %s)", lengthLimit-randomLength, project.Name, shoot.GenerateName))
```

rfranzke (changes requested):
Can you also rebase (probably you have to update 1.14) and make the CI happy?

rfranzke:
/kind/enhancement

plugin/pkg/shoot/dns/admission.go
swapnilgm: No specific calculation to reach this number. We can reduce or increase. I am not aware of any restriction on DNS subdomain. I picked it at random but reasonably safe. Just didn't thought too long random dns name is not needed. but too small will limit the range of unique DNS name.

Generatename and DNS subdomain don't have dependency anyways. So i didn't get why do you expect this to dependent on len(GenerateName)?

rfranzke: > Generatename and DNS subdomain don't have dependency anyways. So i didn't get why do you expect this to dependent on len(GenerateName)?

No specific reason, it doesn't matter too much. It just feels more natural to have the length similar to the length of the shoot name instead of arbitrarily choosing `15`

plugin/pkg/shoot/dns/admission.go
swapnilgm: >If both .metadata.generateName and metadata.name are specified then .metadata.name will win (i.e., we would have to execute the below/already existing check)

It's the precise case.  This hasPrefix check guarantees that, even if generateName is specifed and Shoot name is not generated then go for string domain scheme, where ShootName should match.
```
Shoot.Name ="bar-23"
Shoot.GenerateName = "foo-"
```

It for the case, where shoot  name and shoot generateName both are specified but shoot name isn't generated 

For the case, in update call,
```
Shoot.Name ="foo-23"
Shoot.GenerateName = "foo-"
```
We can't differentiate whether the Shoot.Name is generated or user specified, so domain may or may not have ShootName.So, we apply check till project name and not shoot name.

rfranzke: OK, understood, thanks, makes sense. 👍

swapnilgm:
@rfranzke  Thanks for review. I have addressed the comment. PTAL.
Locally the `make verify` runs without error. PR don't have any changes to go.mod.
But ci job is failing with following error: 
```
> Generate
go: inconsistent vendoring in /go/src/github.com/gardener/gardener:
	golang.org/x/tools@v0.0.0-20200428211428-0c9eba77bc32: is explicitly required in go.mod, but vendor/modules.txt indicates golang.org/x/tools@v0.0.0-20200408032209-46bd65c8538f

run 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory
```
Is vendoring on master broken? - Doesn't look like so. `Make revendor ` also had no change on go.mod or go.sum. 

Not sure whats wrong with ci build environment. Any quick hint would be nice.

rfranzke:
Can you run `make install-requirements revendor verify`?

plugin/pkg/shoot/dns/admission_test.go
```diff
@@ -471,6 +471,91 @@ var _ = Describe("dns", func() {
...
+					Expect(kubeInformerFactory.Core().V1().Secrets().Informer().GetStore().Add(&defaultDomainSecret)).To(Succeed())
+					Expect(coreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&project)).To(Succeed())
+					Expect(coreInformerFactory.Core().InternalVersion().Seeds().Informer().GetStore().Add(&seed)).To(Succeed())
+					attrs = admission.NewAttributesRecord(shootCopy, nil, core.Kind("Shoot").WithVersion("version"), shootCopy.Namespace, shootCopy.Name, core.Resource("shoots").WithVersion("version"), "", admission.Create, &metav1.CreateOptions{}, false, nil)
+
+					err = admissionHandler.Admit(context.TODO(), attrs, nil)
+
+					Expect(err).To(Not(HaveOccurred()))
+
+					Expect(*shoot.Spec.DNS.Domain).NotTo(Equal(*shootCopy.Spec.DNS.Domain))
```
rfranzke, line 501: Shouldn't this here be a separate test (in a separate `It()`)?

rfranzke (commented):
Nit in the test

plugin/pkg/shoot/dns/admission_test.go
swapnilgm: No. It's a single test. Are you interpreting line 484 to 490 and line 490 to 501 different  `IT` cases?

rfranzke: I see, forget my comment. Understood now. Thanks!

rfranzke:
Can you please rebase and squash the commits?

swapnilgm:
Done.

rfranzke (approved):

PR #2237, opened 2020-04-27
Extend error code mapping

rfranzke:
This PR extends our error code mapping with more error messages that now properly map to known codes.
Also, it introduces two new error codes, `ERR_INFRA_RESOURCES_DEPLETED` and `ERR_CONFIGURATION_PROBLEM`.

* In a future step I'll partly move the provider-specific error codes from g/g to the various provider extension repositories.
* I removed the error helpers from `core/v1alpha1/helper` package to prevent double-maintenance.

**Release note**:

```noteworthy user
There are two new error codes that help categorizing a problem occurred during a shoot reconciliation or health check: `ERR_INFRA_RESOURCES_DEPLETED` indicates that the underlying infrastructure does not have enough resources anymore, and `ERR_CONFIGURATION_PROBLEM` indicates that the user has misconfigured something and should double-check the specification.
```


timuthy (approved):
/lgtm

PR #2238, opened 2020-04-27
Allow switching kube-proxy mode for k8s cluster > 1.16

### LINKED ISSUE #2225: Lift restriction that prevents switching kube proxy mode
With https://github.com/gardener/gardener/pull/961 we have added validation that prevents changing the kube-proxy mode as there were upstream issue. Those issues are resolved with https://github.com/kubernetes/kubernetes/pull/78775 as of 1.16, so we can lift our restriction as well.

Please only prevent switching the mode if the `v1.14.1 <= kubernetes version < 1.16`.

https://github.com/gardener/gardener/blob/a835e06efa0b641de6802ac1f615c08eba23151a/pkg/apis/core/validation/shoot.go#L326-L328

**Why is this needed**:
More flexibility

DockToFuture:
This PR lifts the restrictions on `kube-proxy` to allow switching of kube-proxy mode (`IPTables`, `IPVS`) for k8s cluster > 1.16.  Further the `iptables` and `ipvs` rules of the previous mode are cleaned up in an init container before running `kube-proxy` in the new mode.

PR https://github.com/kubernetes/kubernetes/pull/78775 fixed the cleanup error in `v1.16` which led in the past to the restrictions. 

Fixes #2225

**Release note**:

```improvement operator
Restrictions on kube-proxy are lifted to allow switching of kube-proxy mode (IPTables, IPVS) for k8s cluster > 1.16. 
```

### DIFF
--- a/charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
+++ b/charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
@@ -1,0 +1,31 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kube-proxy-cleanup-script
+  namespace: kube-system
+  labels:
+    app: kubernetes
+    gardener.cloud/role: system-component
+    origin: gardener
+    role: proxy
+data:
+  cleanup.sh: |
+    #!/bin/sh -e
+    OLD_KUBE_PROXY_MODE="$(cat "$1")"
+    if [ -z "${OLD_KUBE_PROXY_MODE}" ] || [ "${OLD_KUBE_PROXY_MODE}" = "${KUBE_PROXY_MODE}" ]; then
+      echo "${KUBE_PROXY_MODE}" >"$1"
+      echo "Nothing to cleanup - the mode didn't change."
+      exit 0
+    fi
+    {{- if semverCompare "< 1.17" .Values.kubernetesVersion }}
+    /hyperkube
+    {{- if semverCompare "< 1.15" .Values.kubernetesVersion }}
+    proxy
+    {{- else }}
+    kube-proxy
+    {{- end }}
+    {{- else }}
+    /usr/local/bin/kube-proxy
+    {{- end }} --v=2 --cleanup --proxy-mode="${OLD_KUBE_PROXY_MODE}"
+    echo "${KUBE_PROXY_MODE}" >"$1"
--- a/charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
+++ b/charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
@@ -20,6 +20,7 @@ spec:
         scheduler.alpha.kubernetes.io/critical-pod: ''
         checksum/secret-kube-proxy: {{ include (print $.Template.BasePath "/kube-proxy-secret.yaml") . | sha256sum }}
         checksum/configmap-componentconfig: {{ include (print $.Template.BasePath "/componentconfig.yaml") . | sha256sum }}
+        checksum/configmap-kube-proxy-cleanup-script: {{ include (print $.Template.BasePath "/kube-proxy-cleanup-script.yaml") . | sha256sum }}
 {{- if .Values.podAnnotations }}
 {{ toYaml .Values.podAnnotations | indent 8 }}
 {{- end }}
@@ -45,6 +46,30 @@ spec:
           /sbin/sysctl -w net.ipv6.conf.all.disable_ipv6=1 || echo "IPv6 is already disabled. Doing nothing..."
         securityContext:
           privileged: true
+{{- end }}
+{{- if (semverCompare ">= 1.16" .Values.kubernetesVersion) }}
+      initContainers:
+      - name: cleanup
+        image: {{ index .Values.images "kube-proxy" }}
+        env:
+        - name: KUBE_PROXY_MODE
+          value: {{ include "kube-proxy.mode" . }}
+        imagePullPolicy: IfNotPresent
+        command:
+        - sh
+        - -c
+        - /script/cleanup.sh /var/lib/kube-proxy/mode
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - name: kube-proxy-cleanup-script
+          mountPath: /script
+        - name: kernel-modules
+          mountPath: /lib/modules
+        - name: kube-proxy-dir
+          mountPath: /var/lib/kube-proxy
+        - name: kube-proxy-mode
+          mountPath: /var/lib/kube-proxy/mode
 {{- end }}
       priorityClassName: system-cluster-critical
       tolerations:
@@ -114,3 +139,15 @@ spec:
       - name: kernel-modules
         hostPath:
           path: /lib/modules
+      - name: kube-proxy-cleanup-script
+        configMap:
+          name: kube-proxy-cleanup-script
+          defaultMode: 0777
+      - name: kube-proxy-dir
+        hostPath:
+          path: /var/lib/kube-proxy
+          type: DirectoryOrCreate
+      - name: kube-proxy-mode
+        hostPath:
+          path: /var/lib/kube-proxy/mode
+          type: FileOrCreate
--- a/pkg/apis/core/validation/shoot.go
+++ b/pkg/apis/core/validation/shoot.go
@@ -323,7 +323,7 @@ func validateKubeProxyModeUpdate(newConfig, oldConfig *core.KubeProxyConfig, ver
 	if oldConfig != nil {
 		oldMode = *oldConfig.Mode
 	}
-	if ok, _ := versionutils.CheckVersionMeetsConstraint(version, ">= 1.14.1"); ok {
+	if ok, _ := versionutils.CheckVersionMeetsConstraint(version, ">= 1.14.1, < 1.16"); ok {
 		allErrs = append(allErrs, apivalidation.ValidateImmutableField(newMode, oldMode, fldPath.Child("mode"))...)
 	}
 	return allErrs
--- a/pkg/apis/core/validation/shoot_test.go
+++ b/pkg/apis/core/validation/shoot_test.go
@@ -1469,10 +1469,30 @@ var _ = Describe("Shoot Validation Tests", func() {
 				mode := core.ProxyMode("IPVS")
 				shoot.Spec.Kubernetes.Version = "1.14.1"
 				shoot.Spec.Kubernetes.KubeProxy.Mode = &mode
-
 				errorList := ValidateShoot(shoot)
-
 				Expect(errorList).To(HaveLen(2))
+			})
+
+			It("should be successful when using kubernetes version 1.16.1 and proxy mode is changed", func() {
+				mode := core.ProxyMode("IPVS")
+				kubernetesConfig := core.KubernetesConfig{}
+				config := core.KubeProxyConfig{
+					KubernetesConfig: kubernetesConfig,
+					Mode:             &mode,
+				}
+				shoot.Spec.Kubernetes.KubeProxy = &config
+				shoot.Spec.Kubernetes.Version = "1.16.1"
+				oldMode := core.ProxyMode("IPTables")
+				oldConfig := core.KubeProxyConfig{
+					KubernetesConfig: kubernetesConfig,
+					Mode:             &oldMode,
+				}
+				shoot.Spec.Kubernetes.KubeProxy.Mode = &mode
+				oldShoot := shoot.DeepCopy()
+				oldShoot.Spec.Kubernetes.KubeProxy = &oldConfig
+
+				errorList := ValidateShootSpecUpdate(&shoot.Spec, &oldShoot.Spec, false, field.NewPath("spec"))
+				Expect(errorList).To(BeEmpty())
 			})
 		})
 


rfranzke (dismissed):

ialidzhikov:
@DockToFuture, can you update the PR description with the corresponding k/k issue and provide details about it?

DockToFuture:
Will update it, please wait with merging.

rfranzke:
/kind/enhancement

rfranzke:
@DockToFuture what's the status of this PR?

DockToFuture:
Before we switch the `proxy-mode` we have to run `kube-proxy` with the `--cleanup` flag and wait until it successfully terminates otherwise the `ipvs rules` or `iptables rules` won't be properly purged. It will not work after the mode is switched. Therefore I'm looking for a proper way how to implement it.

vpnachev:
An init container with `kube-proxy --cleanup` should do the job, what do you think?

DockToFuture:
This cleans up the current mode not the one from which `kube-proxy` has switched.

zanetworker:
would a prestop hook to do the cleanup help here?

rfranzke:
A prestop hook would be executed whenever the kube-proxy pod stops, so I don't think that's what we want. Instead, would it be viable to go with @vpnachev's idea? Could we add an init container that reads the configured mode from the kube-proxy config and then decides whether to cleanup?

Basically: When the mode is `iptables` then check if there are IPVS artifacts on the node, and if yes then run the `--cleanup` job, if not simply `exit 0`. If the mode is `IPVS` then check if there are iptable artifacts on the node, and if yes then run the `--cleanup` job, if not simply `exit 0`.

If that's not possible then we would have to "remember" the old mode and then decide in code when to spawn the `--cleanup` job. However, this would be more complex. So can we achieve above suggestion?

vpnachev:
> When the mode is iptables then check if there are IPVS artifacts on the node, and if yes then run the --cleanup job, if not simply exit 0. If the mode is IPVS then check if there are iptable artifacts on the node, and if yes then run the --cleanup job, if not simply exit 0.

This was exactly what I was thinking last night. I am just not sure how to properly detect existing artifacts, so maybe the init container can dump the current active mode in a file on the host, e.g. `/kube-proxy-mode`.

```bash
OLD_KUBE_PROXY_MODE=$(cat /kube-proxy-mode)
[[ ${OLD_KUBE_PROXY_MODE} == ${KUBE_PROXY_MODE} ]] && exit 0
kube-proxy --clean-up || exit $?
echo ${KUBE_PROXY_MODE} >/kube-proxy-mode
```  

The `KUBE_PROXY_MODE` variable can be set by gardener in the daemonset, or can be read  from `/var/lib/kube-proxy-config/config.yaml` file.

rfranzke:
+💯 @vpnachev

zanetworker:
@rfranzke the --cleanup flag cleans up both IPVS and IPtables according to the documentation, do we need the check here? Or is that not the case @DockToFuture 

```
--cleanup
  If true cleanup iptables and ipvs rules and exit.
```

rfranzke:
You need the check because you only want to run the cleanup if you switch the mode. Otherwise, we end up with the same problem like the pre-stop hook - it would be executed everytime when kube-proxy starts.  😁

rfranzke:
@DockToFuture why have you introduced a new field in the status? Does @vpnachev's proposal not work https://github.com/gardener/gardener/pull/2238#issuecomment-623889699?

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -45,6 +45,28 @@ spec:
           /sbin/sysctl -w net.ipv6.conf.all.disable_ipv6=1 || echo "IPv6 is already disabled. Doing nothing..."
         securityContext:
           privileged: true
+{{- end }}
+{{- if and .Values.enableCleanup (ne .Values.previousMode "") (semverCompare ">= 1.16" .Values.kubernetesVersion) }}
```
vpnachev: Doing it this way, every time the `kube-proxy` pod is restarted it will clean the previous rules.
This might cause service disruptions for a short period of time.

DockToFuture: Only when switching the kube-proxy mode and a previous mode was set. After the next reconcile cleanup is set to false and init container will be removed from the ds.

rfranzke: The next reconciliation might happen only in the next maintenance time window. Intermittent restarts will lead to service disruptions as @vpnachev said.

DockToFuture:
> @DockToFuture why have you introduced a new field in the status? Does @vpnachev's proposal not work [#2238 (comment)](https://github.com/gardener/gardener/pull/2238#issuecomment-623889699)?

Did remove it, because with the proposal of @vpnachev it is not necessary any more.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -45,6 +45,24 @@ spec:
           /sbin/sysctl -w net.ipv6.conf.all.disable_ipv6=1 || echo "IPv6 is already disabled. Doing nothing..."
         securityContext:
           privileged: true
+{{- end }}
+{{- if (semverCompare ">= 1.16" .Values.kubernetesVersion) }}
```
rfranzke, line 50: Why is the init container only added for >= 1.16? Also, I've forgotten why switching the kube-proxy mode is prohibited only during 1.14.1-1.16 versions. What about earlier versions (< 1.14)? Does the switch work here? If yes, why is this init container not needed?

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -114,3 +132,7 @@ spec:
       - name: kernel-modules
         hostPath:
           path: /lib/modules
+      - name: kube-proxy-init-script
```
rfranzke: Can you add a checksum annotation to the pod template for this configmap mount?

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -45,6 +45,24 @@ spec:
           /sbin/sysctl -w net.ipv6.conf.all.disable_ipv6=1 || echo "IPv6 is already disabled. Doing nothing..."
         securityContext:
           privileged: true
+{{- end }}
+{{- if (semverCompare ">= 1.16" .Values.kubernetesVersion) }}
+      initContainers:
```
rfranzke, line 51: `initContainers` key is already defined in [line 35](https://github.com/gardener/gardener/pull/2238/files#diff-0d19dce54b9a94959c40888d94978cb7R35).

charts/shoot-core/components/charts/kube-proxy/templates/psp/kube-proxy-init-script.yaml
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kube-proxy-init-script
+  namespace: kube-system
+data:
+  init.sh: |
+    #!/bin/sh
+    OLD_KUBE_PROXY_MODE=$(cat /kube-proxy-mode);
+    if [ A${OLD_KUBE_PROXY_MODE} = A${KUBE_PROXY_MODE} ]; then
```
rfranzke: Hm, what is this expression? What does it mean? What's the `A`?

charts/shoot-core/components/charts/kube-proxy/templates/psp/kube-proxy-init-script.yaml
```diff
@@ -0,0 +1,24 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kube-proxy-init-script
+  namespace: kube-system
+data:
+  init.sh: |
+    #!/bin/sh
+    OLD_KUBE_PROXY_MODE=$(cat /kube-proxy-mode);
```
rfranzke: As far as I can see this file only exists inside the ephemeral storage of the container. Shouldn't it be on the host/VM itself, i.e., shouldn't you mount a host path where it can be stored?

rfranzke (changes requested):

charts/shoot-core/components/charts/kube-proxy/templates/psp/kube-proxy-init-script.yaml
DockToFuture: `sh` does not allow the comparison of an empty strings and will return an error.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
DockToFuture: but only if this condition is true `{{- if and .Values.enableIPVS (semverCompare "< 1.14" .Values.kubernetesVersion) }}`

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
DockToFuture: Sure also thought about that, but i come to the conclusion that we won't change this script. What is the benefit?

charts/shoot-core/components/charts/kube-proxy/templates/psp/kube-proxy-init-script.yaml
rfranzke: What about making it more explicit, e.g:

```shell
#!/bin/sh

function foo {
  if [ -z "$1" ] || [ "$1" == "expected-string" ]; then
    echo "exit 0"
  fi
}

foo "$(cat /non-existing-file)"
foo "expected-string"
```

? Both invocations of `foo` return `exit 0`.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
rfranzke: Can you look into the future 🔮 😃? OK, joking aside, you're right, it's probably more a theoretical case in which the annotation would help. I'm thinking about this script having a bug that will make the init container crash loop. Without the checksum annotation fixing the script won't help as it wouldn't trigger a roll-out. As said, probably very unlikely, but I'd still add the checksum, just to be safe (it doesn't harm having it, does it?).

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
DockToFuture: kube proxy used autocleanup in the prior versions, see https://github.com/kubernetes/kubernetes/pull/76109

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
DockToFuture: No, will add it. :)

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
rfranzke: Thank you!

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -114,3 +135,11 @@ spec:
       - name: kernel-modules
         hostPath:
           path: /lib/modules
+      - name: kube-proxy-init-script
+        configMap:
+          name: kube-proxy-init-script
+          defaultMode: 0777
+      - name: kube-proxy-mode
+        hostPath:
+          path: /kube-proxy-mode
```
rfranzke: Probably we should not put in at root level but rather in some other folder. In comparison to the kubelet directory, what about

```suggestion
          path: /var/lib/kube-proxy
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-init-script.yaml
```diff
@@ -0,0 +1,25 @@
...
+    /hyperkube
+    {{- if semverCompare "< 1.15" .Values.kubernetesVersion }}
+    proxy
+    {{- else }}
+    kube-proxy
+    {{- end }}
+    {{- else }}
+    /usr/local/bin/kube-proxy
+    {{- end }} --v=2 --cleanup --proxy-mode=${OLD_KUBE_PROXY_MODE} || exit $?;
+    echo ${KUBE_PROXY_MODE} >/kube-proxy-mode;
```
rfranzke:
```suggestion
    OLD_KUBE_PROXY_MODE="$(cat "$1")";
    if [ -z "${OLD_KUBE_PROXY_MODE}" ] || [ "${OLD_KUBE_PROXY_MODE}" == "${KUBE_PROXY_MODE}" ]; then
      echo "${KUBE_PROXY_MODE}" > "$1";
      exit 0
    fi;
    {{- if semverCompare "< 1.17" .Values.kubernetesVersion }}
    /hyperkube
    {{- if semverCompare "< 1.15" .Values.kubernetesVersion }}
    proxy
    {{- else }}
    kube-proxy
    {{- end }}
    {{- else }}
    /usr/local/bin/kube-proxy
    {{- end }} --v=2 --cleanup --proxy-mode="${OLD_KUBE_PROXY_MODE}" || exit $?;
    echo "${KUBE_PROXY_MODE}" > "$1";
```

About this line:

> `--proxy-mode="${OLD_KUBE_PROXY_MODE}" || exit $?;`

What's the rationale behind it? Why do you need this `|| exit` part?

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -45,6 +46,26 @@ spec:
...
+{{- end }}
+{{- if (semverCompare ">= 1.16" .Values.kubernetesVersion) }}
+      initContainers:
+      - name: kube-proxy-cleanup
+        image: {{ index .Values.images "kube-proxy" }}
+        env:
+        - name: KUBE_PROXY_MODE
+          value: {{ include "kube-proxy.mode" . }}
+        imagePullPolicy: IfNotPresent
+        command: ['sh', '-c', "/script/init.sh"]
```
rfranzke:
```suggestion
        command:
        - sh
        - -c
        - /script/init.sh
        - /var/lib/kube-proxy/mode
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -45,6 +46,26 @@ spec:
...
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - name: kube-proxy-init-script
+          mountPath: /script/init.sh
+          subPath: init.sh
+        - name: kernel-modules
+          mountPath: /lib/modules
+        - name: kube-proxy-mode
+          mountPath: /kube-proxy-mode
```
rfranzke:
```suggestion
          mountPath: /var/lib/kube-proxy
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-init-script.yaml
vpnachev: If the clean-up fails, it has to be retried while the file on the host still contains the old kube-proxy mode.

rfranzke: Sure, but can't we just set the `-e` flag, i.e., `#!/bin/sh -e`? Then line 25 won't be executed if the cleanup process fails.

vpnachev: Sure, with `-e` it is the same behavior.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -20,6 +20,7 @@ spec:
         scheduler.alpha.kubernetes.io/critical-pod: ''
         checksum/secret-kube-proxy: {{ include (print $.Template.BasePath "/kube-proxy-secret.yaml") . | sha256sum }}
         checksum/configmap-componentconfig: {{ include (print $.Template.BasePath "/componentconfig.yaml") . | sha256sum }}
+        checksum/configmap-kube-proxy-init-script: {{ include (print $.Template.BasePath "/kube-proxy-init-script.yaml") . | sha256sum }}
```
vpnachev:
```suggestion
        checksum/configmap-kube-proxy-cleanup-script: {{ include (print $.Template.BasePath "/kube-proxy-cleanup-script.yaml") . | sha256sum }}
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -45,6 +46,26 @@ spec:
...
+        image: {{ index .Values.images "kube-proxy" }}
+        env:
+        - name: KUBE_PROXY_MODE
+          value: {{ include "kube-proxy.mode" . }}
+        imagePullPolicy: IfNotPresent
+        command: ['sh', '-c', "/script/init.sh"]
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - name: kube-proxy-init-script
```
vpnachev:
```suggestion
        - name: kube-proxy-cleanup-script
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-init-script.yaml
```diff
@@ -0,0 +1,25 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kube-proxy-init-script
```
vpnachev:
```suggestion
  name: kube-proxy-cleanup-script
```

Please, add some labels, too.
```yaml
    app: kubernetes
    gardener.cloud/role: system-component
    origin: gardener
    role: proxy
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-init-script.yaml
```diff
@@ -0,0 +1,25 @@
...
+  namespace: kube-system
+data:
+  init.sh: |
```
vpnachev:
```suggestion
  cleanup.sh: |
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-init-script.yaml
```diff
@@ -0,0 +1,25 @@
...
+    #!/bin/sh
+    OLD_KUBE_PROXY_MODE=$(cat /kube-proxy-mode);
```
vpnachev: Semicolons are not required in shell scripts at the end of the lines.
```suggestion
    OLD_KUBE_PROXY_MODE=$(cat /kube-proxy-mode)
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-init-script.yaml
```diff
@@ -0,0 +1,25 @@
...
+    if [ -z ${OLD_KUBE_PROXY_MODE} ] || [ ${OLD_KUBE_PROXY_MODE} = ${KUBE_PROXY_MODE} ]; then
```
vpnachev:
```suggestion
    if [ -z ${OLD_KUBE_PROXY_MODE} ] || [ ${OLD_KUBE_PROXY_MODE} = ${KUBE_PROXY_MODE} ]
    then
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-init-script.yaml
```diff
@@ -0,0 +1,25 @@
...
+    {{- if semverCompare "< 1.17" .Values.kubernetesVersion }}
+    /hyperkube
+    {{- if semverCompare "< 1.15" .Values.kubernetesVersion }}
+    proxy
+    {{- else }}
+    kube-proxy
+    {{- end }}
+    {{- else }}
+    /usr/local/bin/kube-proxy
+    {{- end }} --v=2 --cleanup --proxy-mode=${OLD_KUBE_PROXY_MODE} || exit $?;
```
vpnachev: `--proxy-mode=${OLD_KUBE_PROXY_MODE}` I think this parameter is obsolete.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -45,6 +46,26 @@ spec:
...
+          mountPath: /script/init.sh
+          subPath: init.sh
```
vpnachev:
```suggestion
          mountPath: /script
```

It would be even better if you mount it under `/var/lib/kube-proxy/script`

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
vpnachev:
```
s/init.sh/cleanup.sh/
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -114,3 +135,11 @@ spec:
       - name: kernel-modules
         hostPath:
           path: /lib/modules
+      - name: kube-proxy-init-script
```
vpnachev:
```suggestion
      - name: kube-proxy-cleanup-script
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -114,3 +135,11 @@ spec:
...
+        configMap:
+          name: kube-proxy-init-script
```
vpnachev:
```suggestion
          name: kube-proxy-cleanup-script
```

vpnachev (commented):
Some minor nits, basically rename `init` to `cleanup` .

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-init-script.yaml
DockToFuture: That's true but does not work properly with env variable.

vpnachev: I wanted to say that you do not need 
```
--proxy-mode=${OLD_KUBE_PROXY_MODE}
```
when
```
--cleanup
```
 is set, do you?

DockToFuture: I tested it and it was only properly working when it was set. In theory cleanup should do both but when testing it always skipped `ipvs` cleanup, when not explicitly set.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
```diff
@@ -0,0 +1,30 @@
...
+  namespace: kube-system
+  labels:
+    app: kubernetes
+    gardener.cloud/role: system-component
+    origin: gardener
+    role: proxy
+data:
+  cleanup.sh: |
+    #!/bin/sh -e
+    OLD_KUBE_PROXY_MODE=$(cat "$1")
```
rfranzke:
```suggestion
    OLD_KUBE_PROXY_MODE="$(cat "$1")"
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
```diff
@@ -0,0 +1,30 @@
...
+    if [ -z ${OLD_KUBE_PROXY_MODE} ] || [ ${OLD_KUBE_PROXY_MODE} == ${KUBE_PROXY_MODE} ]; then
```
rfranzke:
```suggestion
    if [ -z "${OLD_KUBE_PROXY_MODE}" ] || [ "${OLD_KUBE_PROXY_MODE}" == "${KUBE_PROXY_MODE}" ]; then
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
```diff
@@ -0,0 +1,30 @@
...
+      echo ${KUBE_PROXY_MODE} >"$1"
```
rfranzke:
```suggestion
      echo "${KUBE_PROXY_MODE}" >"$1"
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
```diff
@@ -0,0 +1,30 @@
...
+    {{- if semverCompare "< 1.17" .Values.kubernetesVersion }}
+    /hyperkube
+    {{- if semverCompare "< 1.15" .Values.kubernetesVersion }}
+    proxy
+    {{- else }}
+    kube-proxy
+    {{- end }}
+    {{- else }}
+    /usr/local/bin/kube-proxy
+    {{- end }} --v=2 --cleanup --proxy-mode=${OLD_KUBE_PROXY_MODE}
```
rfranzke:
```suggestion
    {{- end }} --v=2 --cleanup --proxy-mode="${OLD_KUBE_PROXY_MODE}"
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
```diff
@@ -0,0 +1,30 @@
...
+    echo ${KUBE_PROXY_MODE} >"$1"
```
rfranzke:
```suggestion
    echo "${KUBE_PROXY_MODE}" >"$1"
```

rfranzke (changes requested):

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
rfranzke: ping, this is not resolved yet

DockToFuture: I forgot this, thx for checking. Is now resolved.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
```diff
@@ -0,0 +1,30 @@
...
+    {{- if semverCompare "< 1.17" .Values.kubernetesVersion }}
+    /hyperkube
+    {{- if semverCompare "< 1.15" .Values.kubernetesVersion }}
+    proxy
+    {{- else }}
+    kube-proxy
+    {{- end }}
+    {{- else }}
+    /usr/local/bin/kube-proxy
+    {{- end }} --v=2 --cleanup --proxy-mode="${OLD_KUBE_PROXY_MODE}"
```
rfranzke, line 30:
```
W0513 06:07:39.225390       8 server.go:213] WARNING: all flags other than --config, --write-config-to, and --cleanup are deprecated. Please begin using a config file ASAP.
```

According to this message the `--proxy-mode` flag should not be needed, correct?

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
```diff
@@ -0,0 +1,30 @@
...
+    app: kubernetes
+    gardener.cloud/role: system-component
+    origin: gardener
+    role: proxy
+data:
+  cleanup.sh: |
+    #!/bin/sh -e
+    OLD_KUBE_PROXY_MODE="$(cat "$1")"
+    if [ -z "${OLD_KUBE_PROXY_MODE}" ] || [ "${OLD_KUBE_PROXY_MODE}" == "${KUBE_PROXY_MODE}" ]; then
+      echo "${KUBE_PROXY_MODE}" >"$1"
```
vpnachev, line 17:
```suggestion
      echo "${KUBE_PROXY_MODE}" >"$1"
      echo "there is nothing to cleanup, exiting"
```

rfranzke, line 17:
```suggestion
      echo "${KUBE_PROXY_MODE}" >"$1"
      echo "Nothing to cleanup - the mode didn't change."
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -45,6 +46,31 @@ spec:
           /sbin/sysctl -w net.ipv6.conf.all.disable_ipv6=1 || echo "IPv6 is already disabled. Doing nothing..."
         securityContext:
           privileged: true
+{{- end }}
+{{- if (semverCompare ">= 1.16" .Values.kubernetesVersion) }}
+      initContainers:
+      - name: kube-proxy-cleanup
```
rfranzke:
```suggestion
      - name: cleanup
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -45,6 +46,31 @@ spec:
...
+        command:
+        - sh
+        - -c
+        - /script/cleanup.sh /var/lib/kube-proxy/mode
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - name: kube-proxy-cleanup-script
+          mountPath: /script/cleanup.sh
+          subPath: cleanup.sh
```
vpnachev:
```suggestion
          mountPath: /script
```

rfranzke (changes requested):
Tested it out, works great, just a few left-over cosmetics, otherwise lgtm.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -114,3 +140,15 @@ spec:
...
+          name: kube-proxy-cleanup-script
+          defaultMode: 0777
+      - name: kube-proxy-dir
+        hostPath:
+          path: /var/lib/kube-proxy
+          type: DirectoryOrCreate
+      - name: kube-proxy-mode
+        hostPath:
+          path: /var/lib/kube-proxy/mode
+          type: FileOrCreate
```
vpnachev, line 153: `kube-proxy-dir` is enough.
```suggestion
```

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
```diff
@@ -45,6 +46,31 @@ spec:
...
+        volumeMounts:
+        - name: kube-proxy-cleanup-script
+          mountPath: /script/cleanup.sh
+          subPath: cleanup.sh
+        - name: kernel-modules
+          mountPath: /lib/modules
+        - name: kube-proxy-dir
+          mountPath: /var/lib/kube-proxy
+        - name: kube-proxy-mode
+          mountPath: /var/lib/kube-proxy/mode
```
vpnachev, line 72:
```suggestion
```

vpnachev (changes requested):
Thank you!

It looks very well now, I have just a few minor change requests.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
rfranzke: duplicate of https://github.com/gardener/gardener/pull/2238#discussion_r424193815

vpnachev: Ops, Rafael also has a similar proposal above.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-cleanup-script.yaml
DockToFuture: Yes, but unfortunately it is not working without the `--proxy-mode` flag. Everytime it returns nil for the ipvs interface, in the  CleanupLeftovers func. https://github.com/kubernetes/kubernetes/blob/77a86a52c0ec88bed49e752831c7820a98ca1c69/pkg/proxy/ipvs/proxier.go#L776.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
DockToFuture: It will not work if I skip creating the `/var/lib/kube-proxy/mode` file because `cat` will return an error for a non-existing file. This is the case if we create a new cluster.

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
DockToFuture: See comment above

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
DockToFuture: Changed it. :)

rfranzke (dismissed):

charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
vpnachev: You can use 
```bash
touch "$1"
```
at the beginning of the cleanup script.

DockToFuture: Thats true, but I guess `FileOrCreate` function was exactly created for this purpose.

vpnachev: I found it very strange to mount 2 different volumes just because of a single file, but I am fine.

vpnachev (approved):
/lgtm

PR #2242, opened 2020-04-28
Health checker computes error codes where applicable

rfranzke:
This PR adapts the health checker used by the shoot care controller to compute error codes where applicable.
Concretely,

1. the `.status.lastError` in `Etcd` resources is evaluated
1. the `.status.conditions[].message`s in `Node` resources are evaluated
1. node conditions indicating disk/memory/pid pressure result in `ERR_CONFIGURATION_PROBLEM`

Also, the conditions are now always pardoned if the shoot is in creation/deletion process (earlier, the first error occurred during such reconciliations did not pardon the conditions anymore).

* The error codes computed by extension health check controllers are already taken into consideration (was implemented as part of #2212).
* The migration logic due to introduction of etcd-druid has been removed now

**Release note**:

```improvement user
The shoot health check controller has been improved to produce error codes (if applicable) to the `.status.conditions[].codes` that help categorizing observed problems.
```

### DIFF
--- a/pkg/apis/core/v1beta1/helper/errors.go
+++ b/pkg/apis/core/v1beta1/helper/errors.go
@@ -51,7 +51,7 @@ var (
 	insufficientPrivilegesRegexp = regexp.MustCompile(`(?i)(AccessDenied|Forbidden|deny|denied)`)
 	dependenciesRegexp           = regexp.MustCompile(`(?i)(PendingVerification|Access Not Configured|accessNotConfigured|DependencyViolation|OptInRequired|DeleteConflict|Conflict|inactive billing state|ReadOnlyDisabledSubscription|is already being used|InUseSubnetCannotBeDeleted|VnetInUse)`)
 	resourcesDepletedRegexp      = regexp.MustCompile(`(?i)(not available in the current hardware cluster|InsufficientInstanceCapacity|SkuNotAvailable|ZonalAllocationFailed)`)
-	configurationProblemRegexp   = regexp.MustCompile(`(?i)(AzureBastionSubnet|not supported in your requested Availability Zone|InvalidParameterValue|notFound|NetcfgInvalidSubnet|InvalidSubnet|Invalid value)`)
+	configurationProblemRegexp   = regexp.MustCompile(`(?i)(AzureBastionSubnet|not supported in your requested Availability Zone|InvalidParameterValue|notFound|NetcfgInvalidSubnet|InvalidSubnet|Invalid value|KubeletHasInsufficientMemory|KubeletHasDiskPressure|KubeletHasInsufficientPID)`)
 )
 
 // DetermineError determines the Garden error code for the given error and creates a new error with the given message.
--- a/pkg/operation/botanist/health_check.go
+++ b/pkg/operation/botanist/health_check.go
@@ -16,6 +16,7 @@ package botanist
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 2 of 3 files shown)


ialidzhikov (approved):
/lgtm

pkg/apis/core/v1beta1/helper/errors.go
```diff
@@ -51,7 +51,7 @@ var (
 	insufficientPrivilegesRegexp = regexp.MustCompile(`(?i)(AccessDenied|Forbidden|deny|denied)`)
 	dependenciesRegexp           = regexp.MustCompile(`(?i)(PendingVerification|Access Not Configured|accessNotConfigured|DependencyViolation|OptInRequired|DeleteConflict|Conflict|inactive billing state|ReadOnlyDisabledSubscription|is already being used|InUseSubnetCannotBeDeleted|VnetInUse)`)
 	resourcesDepletedRegexp      = regexp.MustCompile(`(?i)(not available in the current hardware cluster|InsufficientInstanceCapacity|SkuNotAvailable|ZonalAllocationFailed)`)
-	configurationProblemRegexp   = regexp.MustCompile(`(?i)(AzureBastionSubnet|not supported in your requested Availability Zone|InvalidParameterValue|notFound|NetcfgInvalidSubnet|InvalidSubnet|Invalid value)`)
+	configurationProblemRegexp   = regexp.MustCompile(`(?i)(AzureBastionSubnet|not supported in your requested Availability Zone|InvalidParameterValue|notFound|NetcfgInvalidSubnet|InvalidSubnet|Invalid value|KubeletHasInsufficientMemory|KubeletHasDiskPressure|KubeletHasInsufficientPID)`)
```
danielfoehrKn, line 54: Mhm, why is KubeletHasInsufficientMemory necessarely a (shoot) - configuration problem? To my knowledge that can temporarely happen if the node has pods scheduled with burstable or best-effort QoS class (c-groups do not have memory limit and can therefore overshoot).  I think that is more a workload problem rather than a Shoot configuration issue  like suggested here.

rfranzke: The name "configuration problem" does not specify where/what the configuration problem is. It's an "umbrella" code that can be used in multiple situations.
The rationale behind this PR is to help operators that work with the Dashboard and the list of "clusters with issues" by hiding shoots where an operator cannot help anyways. If the kubelet has insufficient memory, for example, then the operator can't do anything. As you said, it's a workload (resources requests/limits configuration) problem and out of the operator's control. As the Dashboard/Gardener is using error codes to indicate user/infrastructure problems I think this code is a good fit for such kind of situations.

rfranzke:
/kind/enhancement

pkg/apis/core/v1beta1/helper/errors.go
danielfoehrKn: >   It's an "umbrella" code that can be used in multiple situations.

Yes the operator does not care if it is a Shoot configuration problem or a workload related problem. 
However not only operators but also end-users are typically using the Gardener dashboard / check the Shoot conditions to check for the health of their Shoot. For them I think it is helpful to differentiate between these two cases. 
Shouldn't be the  dashboard independent of the Gardener? So the Dashboard could just filter both of these cases for the operator - but at least the Shoot owner could differentiate.

rfranzke: Why is "configuration problem" not good enough? It's just an approximation of the care controller to categorize errors. In order to understand more/the concrete error is not part of the error code but of the condition reason/message. I don't see the necessity to introduce a second error code.

rfranzke: @danielfoehrKn are you okay with proceeding for now? Does my argumentation make sense?

danielfoehrKn: yes, go ahead.

PR #2244, opened 2020-04-28
Eliminate useless extensionsv1alpha1.Last{Operation,Error} interfaces

rfranzke:
The `extensionsv1alpha1.LastOperation` and `extensionv1alpha1.LastError` interfaces were useless as there is (now and in the foreseeable future) only one type/implementation of the `LastOperation`/`LastError` types.
This PR cleans them up and adapts the code that were using them.

/cc @timuthy 
/kind/cleanup

**Release note**:

```improvement developer
The `extensionsv1alpha1.Last{Operation,Error}` interfaces were removed - the respective `GetLast{Operation,Error}()` functions do now return the objects directly instead of the old interfaces.
```


timuthy (approved):
/lgtm

PR #2247, opened 2020-04-28
Extend controller actuators for migrate and restore operations

swilen-iwanow:
Extends the generic actuators, for the extension controllers of Control Plane, Backup Entry and Operation System Config custom resorces (CR), with operations restore and migrate.
Fixes #

Later a PR for each provider extension and operating system extension will be filled that will revendor those changes.

**Release note**:

```improvement developer
ControlPlane, BackupEntry and OperatingSystemConfig controllers support operations for migrate and restore.
```

### DIFF
--- a/extensions/pkg/controller/backupentry/genericactuator/actuator.go
+++ b/extensions/pkg/controller/backupentry/genericactuator/actuator.go
@@ -108,3 +108,13 @@ func (a *actuator) deployEtcdBackupSecret(ctx context.Context, be *extensionsv1a
 func (a *actuator) Delete(ctx context.Context, be *extensionsv1alpha1.BackupEntry) error {
 	return a.backupEntryDelegate.Delete(ctx, be)
 }
+
+// Restore restores the update of a BackupEntry
+func (a *actuator) Restore(ctx context.Context, be *extensionsv1alpha1.BackupEntry) error {
+	return a.deployEtcdBackupSecret(ctx, be)
+}
+
+// Migrate migrates the BackupEntry
+func (a *actuator) Migrate(ctx context.Context, be *extensionsv1alpha1.BackupEntry) error {
+	return nil
+}
--- a/extensions/pkg/controller/backupentry/genericactuator/actuator.go
+++ b/extensions/pkg/controller/backupentry/genericactuator/actuator.go
@@ -111,7 +111,7 @@ func (a *actuator) Delete(ctx context.Context, be *extensionsv1alpha1.BackupEntr
 
 // Restore restores the update of a BackupEntry
 func (a *actuator) Restore(ctx context.Context, be *extensionsv1alpha1.BackupEntry) error {
-	return a.deployEtcdBackupSecret(ctx, be)
+	return a.Reconcile(ctx, be)
 }
 
 // Migrate migrates the BackupEntry
--- a/extensions/pkg/controller/controlplane/genericactuator/actuator.go
+++ b/extensions/pkg/controller/controlplane/genericactuator/actuator.go
@@ -508,6 +508,8 @@ func marshalWebhooks(webhooks []admissionregistrationv1beta1.MutatingWebhook, na
 	return buf.Bytes(), nil
 }
 
+// Restore reconciles the given controlplane and cluster, restoring the additional Shoot
+// control plane components as needed.
 func (a *actuator) Restore(
 	ctx context.Context,
 	cp *extensionsv1alpha1.ControlPlane,
@@ -516,6 +518,8 @@ func (a *actuator) Restore(
 	return a.Reconcile(ctx, cp, cluster)
 }
 
+// Migrate reconciles the given controlplane and cluster, deleting the additional
+// control plane components as needed.
 func (a *actuator) Migrate(
 	ctx context.Context,
 	cp *extensionsv1alpha1.ControlPlane,
--- a/extensions/pkg/controller/controlplane/genericactuator/actuator.go
+++ b/extensions/pkg/controller/controlplane/genericactuator/actuator.go
@@ -507,3 +507,19 @@ func marshalWebhooks(webhooks []admissionregistrationv1beta1.MutatingWebhook, na
 
 	return buf.Bytes(), nil
 }
+
+func (a *actuator) Restore(
+	ctx context.Context,
+	cp *extensionsv1alpha1.ControlPlane,
+	cluster *extensionscontroller.Cluster,
+) (bool, error) {
+	return a.Reconcile(ctx, cp, cluster)
+}
+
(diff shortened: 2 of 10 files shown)


extensions/pkg/controller/backupentry/genericactuator/actuator.go
```diff
@@ -108,3 +108,13 @@ func (a *actuator) deployEtcdBackupSecret(ctx context.Context, be *extensionsv1a
 func (a *actuator) Delete(ctx context.Context, be *extensionsv1alpha1.BackupEntry) error {
 	return a.backupEntryDelegate.Delete(ctx, be)
 }
+
+// Restore restores the update of a BackupEntry
+func (a *actuator) Restore(ctx context.Context, be *extensionsv1alpha1.BackupEntry) error {
+	return a.deployEtcdBackupSecret(ctx, be)
```
rfranzke: What about calling `a.Reconcile`?

extensions/pkg/controller/controlplane/genericactuator/actuator.go
```diff
@@ -507,3 +507,19 @@ func marshalWebhooks(webhooks []admissionregistrationv1beta1.MutatingWebhook, na
 
 	return buf.Bytes(), nil
 }
+
+func (a *actuator) Restore(
```
rfranzke, line 513: Can you add doc strings to these functions?

extensions/pkg/controller/controlplane/reconciler.go
```diff
@@ -96,18 +101,27 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, err
 	}
 
-	if cp.DeletionTimestamp != nil {
+	operationType := gardencorev1beta1helper.ComputeOperationType(cp.ObjectMeta, cp.Status.LastOperation)
+
+	switch {
```
rfranzke, line 106: It's now the third or fourth time that I see this switch which always does the same checks. Would it make sense to introduce a helper function that can be called by every controller and that returns the respective operation type? You could then call `r.migrate`, `r.delete`, etc., based on the returned operation type?

extensions/pkg/controller/operatingsystemconfig/oscommon/actuator/actuator_migrate.go
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

extensions/pkg/controller/operatingsystemconfig/oscommon/actuator/actuator_restore.go
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

rfranzke:
Can you also add a release note for developers?

extensions/pkg/controller/controlplane/reconciler.go
swilen-iwanow: I am not sure if I understand you correctly. The switch is not working with operationType, but a set of conditions. If we move this logic to а separate function that returns an operationType then another switch should be used in the consumer function.

rfranzke: Yes, right, the idea was a separate function that gets the `extensionsv1alpha1.Object` as input and generate the operationType as output. You would still need this switch here, however, the "computations" for the various cases would be moved into this separate function and can be reused across all controllers. WDYT, would something like this make sense?

rfranzke:
/kind/enhancement

extensions/pkg/controller/controlplane/reconciler.go
swilen-iwanow: I think this will lower the code duplication, but decrease simplicity and agility to change this switch in the future for a particular controller. Anyway what operation type should the function return if it is already migrated, I mean like in the first case? Also we still don't have lastOperationType=Restore, because it was introduced with the flow. Maybe do it like this now and change it after?

rfranzke: Yes, sure, agreed, makes sense. Just wanted to throw it in so that we can give it a thought.

rfranzke (dismissed):
/lgtm

vlvasilev (dismissed):
/lgtm

extensions/pkg/controller/controlplane/reconciler.go
plkokanov: We could do it with a factory which returns functions/structs -> MigrateOperation, DeleteOperation, RestoreOperation, ReconcileOperation or NilOperation (for the first case) then for different controllers we could just extend the factory if changes are required. But I'm ok with doing this later on.

extensions/pkg/controller/operatingsystemconfig/reconciler.go
```diff
@@ -195,19 +210,87 @@ func (r *reconciler) delete(ctx context.Context, osc *extensionsv1alpha1.Operati
...
+	if !hasFinalizer {
+		r.logger.Info("Deleting operating system config causes a no-op as there is no finalizer.", "osc", osc.Name)
+		return reconcile.Result{}, nil
+	}
+
+	if err := r.updateStatusProcessing(ctx, osc, gardencorev1beta1.LastOperationTypeDelete, "Deleting the operating system config"); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if err := r.actuator.Delete(ctx, osc); err != nil {
```
plkokanov: Shouldn't this be `r.actuator.Migrate()`

plkokanov (changes requested):

extensions/pkg/controller/operatingsystemconfig/reconciler.go
swilen-iwanow: Sure it must be, thanks!

plkokanov (approved):
/lgtm

rfranzke (approved):

Issue #2197, opened 2020-04-17
Gardenlet should detect outdated HealthCheck conditions in Extension CRDs 


danielfoehrKn:
**What would you like to be added**:

The Gardenlet should detect outdated HealthCheck conditions in the Extension CRDs in the seed.

**Why is this needed**:
The care controller in the gardenlet collects health check conditions from extension CRDs in the Shoot (Worker, Network, Extension) .
Extension providers perform health checks of their deployed resources and write back the health check report into the ExtensionCRD.Status.Conditions.

In case the extension provider for any reason stops to perform health checks, the conditions are still present, and the Gardenlet views them as the actual state (unknowingly that the health check info is outdated). The Gardenlet does not evaluate  the condition.lastUpdate time of a health check condition.

PR #2249, opened 2020-04-28
Fix health check library predicates

### LINKED ISSUE #2241: Healthcheck controller stops tracking CR after hibernation
**What happened**:
With the improvement gardenlet to detect outdated healthcheck conditions (ref #2197) we noticed that after hibernation the healthcheck actuator do no longer requeue the extension CR.

1. Create Shoot

2. Hibernate the Shoot.

3. Wake up the Shoot.

4. Ensure that the extension CR conditions will be outdated. For example the Worker CR:
```
$ k -n shoot--foo--bar get worker baz -o yaml

  status:
    conditions:
    - lastTransitionTime: "2020-04-28T00:29:03Z"
      lastUpdateTime: "2020-04-28T00:29:03Z"
      message: Shoot is hibernated
      reason: HealthCheckSuccessful
      status: "True"
      type: SystemComponentsHealthy
    - lastTransitionTime: "2020-04-28T00:29:02Z"
      lastUpdateTime: "2020-04-28T00:29:02Z"
      message: Shoot is hibernated
      reason: HealthCheckSuccessful
      status: "True"
      type: EveryNodeReady
    - lastTransitionTime: "2020-04-28T00:29:02Z"
      lastUpdateTime: "2020-04-28T00:29:02Z"
      message: Shoot is hibernated
      reason: HealthCheckSuccessful
      status: "True"
      type: ControlPlaneHealthy
    lastOperation:
      description: Successfully reconciled worker
      lastUpdateTime: "2020-04-28T06:15:37Z"
      progress: 100
      state: Succeeded
      type: Reconcile
```

**Environment**:

- Gardener version: master
- Others:

danielfoehrKn:

With the improvement gardenlet to detect outdated healthcheck conditions (ref #2197) we noticed that after hibernation the healthcheck actuator do no longer requeue the extension CR.
Fixes the health check predicates to stop performing health checks when the shoot should be hibernated and starts the checks again when it has fully woken up. 

Fixes #2241

Needs adjustments in extensions during the healthcheck registration.

**Release note**:

```improvement operator
Fixed a bug in the healthcheck library that prevents checks after a Shoot has been woken up from hibernation. Gardener extensions require a minor change during the healthcheck registration.
```

### DIFF
--- a/extensions/pkg/controller/healthcheck/controller.go
+++ b/extensions/pkg/controller/healthcheck/controller.go
@@ -18,11 +18,12 @@ import (
 	"fmt"
 
 	healthcheckconfig "github.com/gardener/gardener/extensions/pkg/controller/healthcheck/config"
+	extensionshandler "github.com/gardener/gardener/extensions/pkg/handler"
 	extensionspredicate "github.com/gardener/gardener/extensions/pkg/predicate"
-
 	"github.com/gardener/gardener/pkg/api/extensions"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"github.com/gardener/gardener/pkg/utils"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -56,6 +57,8 @@ type AddArgs struct {
 	// The Gardenlet reads the conditions on the extension Resource.
 	// Through this mechanism, the extension can contribute to the Shoot's HealthStatus.
 	registeredExtension *RegisteredExtension
+	// GetExtensionObjListFunc returns a list of the runtime.Object representation of the extension to register
+	GetExtensionObjListFunc GetExtensionObjectListFunc
 }
 
 // DefaultAddArgs are the default Args for the health check controller.
@@ -68,35 +71,36 @@ type DefaultAddArgs struct {
 
 // RegisteredExtension is a registered extensions that the HealthCheck Controller watches.
 // The field extension  contains any extension object
-// The field healthConditionType contains all distinct healthCondition types (extracted from the healthCheck).
+// The field healthConditionTypes contains all distinct healthCondition types (extracted from the healthCheck).
 // They are being used as the .type field of the Condition that the HealthCheck controller writes to the extension Resource.
 // The field groupVersionKind stores the GroupVersionKind of the extension resource
 type RegisteredExtension struct {
-	extension           extensionsv1alpha1.Object
-	getExtensionObjFunc GetExtensionObjectFunc
-	healthConditionType []string
-	groupVersionKind    schema.GroupVersionKind
+	extension            extensionsv1alpha1.Object
+	getExtensionObjFunc  GetExtensionObjectFunc
+	healthConditionTypes []string
+	groupVersionKind     schema.GroupVersionKind
 }
 
 // DefaultRegistration configures the default health check NewActuator to execute the provided health checks and adds it to the provided controller-runtime manager.
-// the NewActuator reconciles a single extension with a specific type and writes conditions for each distinct healthConditionType.
+// the NewActuator reconciles a single extension with a specific type and writes conditions for each distinct healthConditionTypes.
 // extensionType (e.g aws) defines the spec.type of the extension to watch
 // kind defines the GroupVersionKind of the extension
-// register returns a runtime.Object representation of the extension to register
+// GetExtensionObjListFunc returns a list of the runtime.Object representation of the extension to register
+// getExtensionObjFunc returns a runtime.Object representation of the extension to register
 // mgr is the controller runtime manager
 // opts contain config for the healthcheck controller
 // custom predicates allow for fine-grained control which resources to watch
-// healthChecks defines the checks to execute mapped to the healthConditionType its contributing to (e.g checkDeployment in Seed -> ControlPlaneHealthy).
+// healthChecks defines the checks to execute mapped to the healthConditionTypes its contributing to (e.g checkDeployment in Seed -> ControlPlaneHealthy).
 // register returns a runtime representation of the extension resource to register it with the controller-runtime
-func DefaultRegistration(extensionType string, kind schema.GroupVersionKind, getExtensionObjFunc GetExtensionObjectFunc, mgr manager.Manager, opts DefaultAddArgs, customPredicates []predicate.Predicate, healthChecks []ConditionTypeToHealthCheck) error {
-	predicates := DefaultPredicates()
-	predicates = append(predicates, customPredicates...)
+func DefaultRegistration(extensionType string, kind schema.GroupVersionKind, getExtensionObjListFunc GetExtensionObjectListFunc, getExtensionObjFunc GetExtensionObjectFunc, mgr manager.Manager, opts DefaultAddArgs, customPredicates []predicate.Predicate, healthChecks []ConditionTypeToHealthCheck) error {
+	predicates := append([]predicate.Predicate{}, customPredicates...)
 
 	args := AddArgs{
-		ControllerOptions: opts.Controller,
-		Predicates:        predicates,
-		Type:              extensionType,
-		SyncPeriod:        opts.HealthCheckConfig.SyncPeriod,
+		ControllerOptions:       opts.Controller,
+		Predicates:              predicates,
+		Type:                    extensionType,
+		SyncPeriod:              opts.HealthCheckConfig.SyncPeriod,
+		GetExtensionObjListFunc: getExtensionObjListFunc,
 	}
 
 	if err := args.RegisterExtension(getExtensionObjFunc, getHealthCheckTypes(healthChecks), kind); err != nil {
@@ -119,26 +123,16 @@ func (a *AddArgs) RegisterExtension(getExtensionObjFunc GetExtensionObjectFunc, 
 	}
 
 	a.registeredExtension = &RegisteredExtension{
-		extension:           acc,
-		healthConditionType: conditionTypes,
-		groupVersionKind:    kind,
-		getExtensionObjFunc: getExtensionObjFunc,
+		extension:            acc,
+		healthConditionTypes: conditionTypes,
+		groupVersionKind:     kind,
+		getExtensionObjFunc:  getExtensionObjFunc,
 	}
 	return nil
 }
 
 func (a *AddArgs) GetExtensionGroupVersionKind() schema.GroupVersionKind {
 	return a.registeredExtension.groupVersionKind
-}
-
-// DefaultPredicates returns the default predicates.
-func DefaultPredicates() []predicate.Predicate {
-	return []predicate.Predicate{
-		extensionspredicate.ShootNotFailed(),
-		// watch: only requeue on spec change to prevent infinite loop
-		// health checks are being executed every 'sync period' anyways
-		predicate.GenerationChangedPredicate{},
-	}
 }
 
 // Register the extension resource. Must be of type extensionsv1alpha1.Object
@@ -161,11 +155,21 @@ func add(mgr manager.Manager, args AddArgs) error {
 	if err != nil {
 		return err
 	}
+
+	log.Log.Info("Registered health check controller", "kind", args.registeredExtension.groupVersionKind.Kind, "type", args.Type, "health check type", args.registeredExtension.healthConditionTypes, "sync period", args.SyncPeriod.Duration.String())
+
+	// add type predicate to only watch registered resource (e.g ControlPlane) with a certain type (e.g aws)
 	predicates := extensionspredicate.AddTypePredicate(args.Predicates, args.Type)
 
-	log.Log.Info("Registered health check controller", "Kind", args.registeredExtension.groupVersionKind.Kind, "type", args.Type, "health check type", args.registeredExtension.healthConditionType, "sync period", args.SyncPeriod.Duration.String())
+	if err := ctrl.Watch(&source.Kind{Type: args.registeredExtension.getExtensionObjFunc()}, &handler.EnqueueRequestForObject{}, predicates...); err != nil {
+		return err
+	}
 
-	return ctrl.Watch(&source.Kind{Type: args.registeredExtension.getExtensionObjFunc()}, &handler.EnqueueRequestForObject{}, predicates...)
+	// watch Cluster of Shoot provider type (e.g aws)
+	// this is to be notified when the Shoot is being hibernated (stop health checks) and wakes up (start health checks again)
+	return ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.Cluster{}}, &extensionshandler.EnqueueRequestsFromMapFunc{
+		ToRequests: extensionshandler.SimpleMapper(extensionshandler.ClusterToObjectMapper(args.GetExtensionObjListFunc, predicates), extensionshandler.UpdateWithNew),
+	})
 }
 
 func getHealthCheckTypes(healthChecks []ConditionTypeToHealthCheck) []string {
--- a/extensions/pkg/controller/healthcheck/healtcheck_actuator.go
+++ b/extensions/pkg/controller/healthcheck/healtcheck_actuator.go
@@ -147,7 +147,7 @@ func (a *Actuator) ExecuteHealthCheckFunctions(ctx context.Context, request type
 				}
 
 				if !preCheckFunc(obj, cluster) {
-					a.logger.Info("Skipping health check as pre check function returned false", "conditionType", healthConditionType)
+					a.logger.V(6).Info("Skipping health check as pre check function returned false", "condition type", healthConditionType)
 					channel <- channelResult{
 						healthCheckResult: &SingleCheckResult{
 							IsHealthy: true,
--- a/extensions/pkg/controller/healthcheck/reconciler.go
+++ b/extensions/pkg/controller/healthcheck/reconciler.go
@@ -21,12 +21,12 @@ import (
 	"github.com/gardener/gardener/extensions/pkg/controller"
 	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
 	"github.com/gardener/gardener/extensions/pkg/util"
-
 	"github.com/gardener/gardener/pkg/api/extensions"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	gardenv1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -101,7 +101,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 
 	if acc.GetDeletionTimestamp() != nil {
-		r.logger.Info("Do not perform HealthCheck for extension resource. Extension is being deleted.", "name", acc.GetName(), "namespace", acc.GetNamespace())
+		r.logger.V(6).Info("Do not perform HealthCheck for extension resource. Extension is being deleted.", "name", acc.GetName(), "namespace", acc.GetNamespace())
 		return reconcile.Result{}, nil
 	}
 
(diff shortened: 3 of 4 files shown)


extensions/pkg/controller/healthcheck/controller.go
```diff
@@ -161,11 +154,17 @@ func add(mgr manager.Manager, args AddArgs) error {
 	if err != nil {
 		return err
 	}
-	predicates := extensionspredicate.AddTypePredicate(args.Predicates, args.Type)
 
 	log.Log.Info("Registered health check controller", "Kind", args.registeredExtension.groupVersionKind.Kind, "type", args.Type, "health check type", args.registeredExtension.healthConditionType, "sync period", args.SyncPeriod.Duration.String())
 
-	return ctrl.Watch(&source.Kind{Type: args.registeredExtension.getExtensionObjFunc()}, &handler.EnqueueRequestForObject{}, predicates...)
+	// add typ predicate to only watch registered resource (e.g ControlPlane) with a certain type (e.g aws)
```
rfranzke:
```suggestion
	// add type predicate to only watch registered resource (e.g ControlPlane) with a certain type (e.g aws)
```

extensions/pkg/controller/healthcheck/controller.go
```diff
@@ -131,16 +134,6 @@ func (a *AddArgs) GetExtensionGroupVersionKind() schema.GroupVersionKind {
 	return a.registeredExtension.groupVersionKind
 }
 
-// DefaultPredicates returns the default predicates.
-func DefaultPredicates() []predicate.Predicate {
```
rfranzke, line 135: Hm, why are they deleted? They weren't used?

extensions/pkg/controller/healthcheck/reconciler.go
```diff
@@ -132,7 +137,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 func (r *reconciler) performHealthCheck(ctx context.Context, request reconcile.Request, rawExtension unstructured.Unstructured) (reconcile.Result, error) {
 	healthCheckResults, err := r.actuator.ExecuteHealthCheckFunctions(ctx, types.NamespacedName{Namespace: request.Namespace, Name: request.Name})
 	if err != nil {
-		r.logger.Info("Failed to execute healthChecks. Updating each HealthCheckCondition for the extension resource to ConditionCheckError.", "kind", r.registeredExtension.groupVersionKind.Kind, "health condition type", r.registeredExtension.healthConditionType, "name", request.Name, "namespace", request.Namespace, "error", err.Error())
+		r.logger.Info("Failed to execute healthChecks. Updating each HealthCheckCondition for the extension resource to ConditionCheckError.", "kind", r.registeredExtension.groupVersionKind.Kind, "health condition types", r.registeredExtension.healthConditionType, "name", request.Name, "namespace", request.Namespace, "error", err.Error())
```
rfranzke: Why "types"? It is still only one type that is printed here, or?

extensions/pkg/controller/healthcheck/reconciler.go
```diff
@@ -123,6 +123,11 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				return r.resultWithRequeue(), err
 			}
 		}
+		r.logger.Info("Do not perform HealthCheck for extension resource. Shoot is hibernated.", "name", acc.GetName(), "namespace", acc.GetNamespace(), "kind", acc.GetObjectKind().GroupVersionKind().Kind)
+		return reconcile.Result{}, err
+	}
+	if cluster.Shoot.Status.IsHibernated {
```
rfranzke: Can't this be combined with line 119?

`if controller.IsHibernated(cluster) || cluster.Shoot.Status.IsHibernated {`

extensions/pkg/controller/healthcheck/reconciler.go
```diff
@@ -123,6 +123,11 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+		r.logger.Info("Do not perform HealthCheck for extension resource. Shoot is waking up.", "name", acc.GetName(), "namespace", acc.GetNamespace(), "kind", acc.GetObjectKind().GroupVersionKind().Kind)
 		return reconcile.Result{}, err
```
rfranzke:
```suggestion
		return reconcile.Result{}, nil
```

pkg/operation/common/extensions.go
```diff
@@ -188,22 +188,24 @@ func WaitUntilExtensionCRsDeleted(
...
-					logger,
-					newObjFunc,
-					kind,
-					o.GetNamespace(),
-					o.GetName(),
-					interval,
-					timeout,
-				)
-			})
+		if predicateFunc != nil && !predicateFunc(o) {
```
rfranzke: Can you remove the changes from this PR?

extensions/pkg/controller/healthcheck/reconciler.go
```diff
@@ -123,6 +123,11 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				return r.resultWithRequeue(), err
 			}
 		}
+		r.logger.Info("Do not perform HealthCheck for extension resource. Shoot is hibernated.", "name", acc.GetName(), "namespace", acc.GetNamespace(), "kind", acc.GetObjectKind().GroupVersionKind().Kind)
+		return reconcile.Result{}, err
```
rfranzke:
```suggestion
		return reconcile.Result{}, nil
```

extensions/pkg/controller/healthcheck/reconciler.go
```diff
@@ -123,6 +123,11 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				return r.resultWithRequeue(), err
```
rfranzke: Why not 
```suggestion
				return reconcile.Result{}, err
```
here?

rfranzke (changes requested):

rfranzke:
/kind/bug

extensions/pkg/controller/healthcheck/reconciler.go
ialidzhikov: Isn't this actually the place where we need to use `r.resultWithRequeue()` as the Shoot is waking up and we need to start performing health checks?

ialidzhikov: Or we rely that there will be a update event for the Cluster where the Shoot will have `.spec.hibernation.enabled=false` and `.status.hibernated=false`?

extensions/pkg/controller/healthcheck/controller.go
danielfoehrKn: As the health check is continuously executed (default every 30 seconds) there should be no reason to watch the individual Extension resources. Rather the health check controllers are now only watching the Cluster resource. This is to get updates when the Shoot should be hibernated (stop the health check) and has woken up again (start again).

extensions/pkg/controller/healthcheck/reconciler.go
danielfoehrKn: No, types contains a slice of [SystemComponentsHealthy, NodesHealth, ...]. This error occurs when the healthChecks could not be executed because the client could not be created .

extensions/pkg/controller/healthcheck/reconciler.go
danielfoehrKn: That would also update the conditions on the extension again (indicating hibernation). I do not think that this is necessary, as the cluster was previously hibernated, so the conditions are already in the right state. Plus no one would read the updated extensions, as the care-controlller does not read extension conditions when the Shoot is hibernated.

extensions/pkg/controller/healthcheck/reconciler.go
rfranzke: In this case, can we rename `healthConditionType` to `healthConditionTypes`, please?

extensions/pkg/controller/healthcheck/reconciler.go
rfranzke: I was thinking to keep it in sync/similar to what the care controller does: https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/health_check.go#L905-L908

extensions/pkg/controller/healthcheck/controller.go
rfranzke: I see.  As we only watch the `Cluster` objects now, how does the cache of the controller-runtime work? Would we now execute an actual `List()` call for the extension objects every time instead of relying on the cache?

extensions/pkg/controller/healthcheck/reconciler.go
danielfoehrKn: Because it failed to update the extension conditions to "hibernated" - it should do that before stopping the health checks. This is to show that it stopped the health-checks deliberately due to hibernation.

extensions/pkg/controller/healthcheck/reconciler.go
danielfoehrKn: Yes, It is watching the Cluster and starts / stops the health checks based on the Shoot hibernation. In case we would miss the one crucial event that sets the status.isHibernated == false, well then it would start whenever there is another change to the Cluster resource - that would typically be during maintenance ? Plus Gardener is able to detect such outdated health check reports. WDYT

extensions/pkg/controller/healthcheck/reconciler.go
rfranzke: As the error is returned then it should be requeued automatically anyways (with exponential backoff instead of fixed delay), or? Am I missing something?

extensions/pkg/controller/healthcheck/reconciler.go
danielfoehrKn: >  the Shoot is waking up and we need to start performing health checks

I thought it would make sense to start the health checks when the Shoot is fully woken up again (status.isHibernated == false) - otherwise health checks would always fail for the period of time when the reconcile / redeploy of resources has not completed.

ialidzhikov: Is there Cluster update event when the Shoot gets fully woken up?

rfranzke: Nice catch @ialidzhikov. Actually, current g/g master doesn't have it, but it will be introduced https://github.com/gardener/gardener/pull/2217. Consequently, the code should be adapted to be backwards compatible.

extensions/pkg/controller/healthcheck/reconciler.go
danielfoehrKn: missed that we are returning the error too - will adapt.

extensions/pkg/controller/healthcheck/reconciler.go
danielfoehrKn: So to be backwards compatible the health checks should start as soon as the Shoot.Spec.Hibernation.Enabled == false?

rfranzke: Probably yes, at least for some time. :(

ialidzhikov: I tested the current PR and the health checks do not start after wake up (Cluster Shoot `.status.hibernated=true`)

danielfoehrKn: I added a[ remark in the original issue](https://github.com/gardener/gardener/issues/2241#issuecomment-621103812) and will adapt the code.

extensions/pkg/controller/healthcheck/reconciler.go
danielfoehrKn: Cannot do the check  `cluster.Shoot.Status.IsHibernated` due to the same reasons as stated [here](https://github.com/gardener/gardener/pull/2249#discussion_r417142588)?

extensions/pkg/controller/healthcheck/reconciler.go
danielfoehrKn: @ialidzhikov ok, due to local setup issues I was only testing on the Cluster resource . It does not start after wakeup due to  #2217 right?

ialidzhikov: Yes

ialidzhikov: Is it ready for review? What was changed to address this comment?

danielfoehrKn: Checking the scenarios currently by using it in the aws-provider.

extensions/pkg/controller/healthcheck/controller.go
danielfoehrKn: that would not be good and should be clarified before merging.

extensions/pkg/controller/healthcheck/reconciler.go
danielfoehrKn: Got caught up in something - so the hibernation and wake up scenario look good. However the initial start of the checks does not work when only watching the cluster. I think I have to add a watch on the actual extension resource again.

extensions/pkg/controller/healthcheck/controller.go
danielfoehrKn: Added the watch on the extension resource again - so it is the same behaviour as before but an additional watch on the Cluster. Would still be interesting to know if and what the client is caching. It is about[ this code here](https://github.com/gardener/gardener/blob/e408106d170496145fac1abe5cf2e1ccae3088fd/extensions/pkg/controller/healthcheck/reconciler.go#L93). Maybe @tim-ebert can shed some light.

extensions/pkg/controller/healthcheck/reconciler.go
danielfoehrKn: Ok ready for review / testing now. Tested it with provider-aws.

rfranzke (commented):
Looks good now code-wise, but I didn't test it out.
@ialidzhikov WDYT?

ialidzhikov:
Hi @danielfoehrKn, thank you for updating the PR. What is changed? We now remove the `GenerationChangedPredicate` and and this helps extension CR to be enqueued on wake up?

danielfoehrKn:
> Hi @danielfoehrKn, thank you for updating the PR. What is changed? We now remove the `GenerationChangedPredicate` and and this helps extension CR to be enqueued on wake up?

Removed the ShootFailed and Generation changed predicates. And added a watch on the cluster.

extensions/pkg/controller/healthcheck/controller.go
timebertt: Yep, looking into it

ialidzhikov:
> Removed the ShootFailed and Generation changed predicates. And added a watch on the cluster.

I can see that. My question was rather what is enqueueing the extension CR after wake up?

rfranzke:
@ialidzhikov I think it's the `Cluster` watch that maps to the extension CR?

ialidzhikov:
> @ialidzhikov I think it's the `Cluster` watch that maps to the extension CR?

But as we discussed yesterday in this PR, currently there is no cluster sync/cluster update event after wake up, right?

rfranzke:
At the beginning of each shoot reconciliation there is a `Cluster` sync, and as the wake-up is an reconciliation as well it should be fine.

ialidzhikov:
> At the beginning of each shoot reconciliation there is a `Cluster` sync, and as the wake-up is an reconciliation as well it should be fine.

I see. Thanks for clarifying.

extensions/pkg/controller/healthcheck/controller.go
```diff
@@ -161,11 +155,21 @@ func add(mgr manager.Manager, args AddArgs) error {
 	if err != nil {
 		return err
 	}
+
+	log.Log.Info("Registered health check controller", "Kind", args.registeredExtension.groupVersionKind.Kind, "type", args.Type, "health check type", args.registeredExtension.healthConditionTypes, "sync period", args.SyncPeriod.Duration.String())
```
ialidzhikov: Nit

```suggestion
	log.Log.Info("Registered health check controller", "kind", args.registeredExtension.groupVersionKind.Kind, "type", args.Type, "health check type", args.registeredExtension.healthConditionTypes, "sync period", args.SyncPeriod.Duration.String())
```

Ref https://github.com/kubernetes-sigs/controller-runtime/blob/master/TMP-LOGGING.md#logging-structured-values-key-value-pairs

extensions/pkg/controller/healthcheck/healtcheck_actuator.go
```diff
@@ -147,7 +147,7 @@ func (a *Actuator) ExecuteHealthCheckFunctions(ctx context.Context, request type
 				}
 
 				if !preCheckFunc(obj, cluster) {
-					a.logger.Info("Skipping health check as pre check function returned false", "conditionType", healthConditionType)
+					a.logger.V(6).Info("Skipping health check as pre check function returned false", "conditionType", healthConditionType)
```
ialidzhikov: Nit

```suggestion
					a.logger.V(6).Info("Skipping health check as pre check function returned false", "condition type", healthConditionType)
```

extensions/pkg/controller/healthcheck/controller.go
timebertt: I think, removing the `DefaultsPredicates` is not a good idea.
Notice the comment, that was there before:
```
// watch: only requeue on spec change to prevent infinite loop	
// health checks are being executed every 'sync period' anyways
```
IIUC and if we remove the default predicates, we will have many unnecessary/unwanted requeues.

timebertt: Wrt to the cache topic, here is what I've found in short:
- under the hood controller-runtime uses `k8s.io/client-go/tools/cache.SharedIndexInformer` for caching
- `controller.Watch(&source.Kind{ ...` will basically result in starting a watch for the given resource via an informer and add an event handler, just like we do for example here:
https://github.com/gardener/gardener/blob/c934fb27dd50489f11330f3fb954fbeb136bd087/pkg/gardenlet/controller/shoot/shoot.go#L119-L126
- `manager.New()` creates two clients
  - one without a cache (direct call to `client.New()`), which can be used for reading directly from the API and injected into controllers/reconcilers by implementing this interface:
https://github.com/gardener/gardener/blob/c934fb27dd50489f11330f3fb954fbeb136bd087/vendor/sigs.k8s.io/controller-runtime/pkg/runtime/inject/inject.go#L45-L47
  - and another one with a cache (if not configured differently via the manager options), which is injected via `InjectClient()` and which we use all over the place in our extensions and extensions library. See for example [here](https://github.com/gardener/gardener-extension-provider-aws/blob/f9b9f455c6af7b6e915f3f2de4a438cb4f663bdf/vendor/sigs.k8s.io/controller-runtime/pkg/manager/manager.go#L317-L332).
This client is a `client.DelegatingClient`, which will use a cache composed of `SharedIndexInformer`s for each `GroupVersionKind`. The trap here is, that for requests with `*unstructured.Unstructured` and `*unstructured.UnstructuredList` it reads directly from the API and not from the cache ([ref](https://github.com/kubernetes-sigs/controller-runtime/blob/fe0e7596fb90a87d024e1d80266b087adbbd36b8/pkg/client/split.go#L46-L61)). All other requests with typed objects/lists, will use the cache, which will be kept in sync by the respective informers.

timebertt: TL;DR:
> Would we now execute an actual List() call for the extension objects every time instead of relying on the cache?

If you mean this call here, no, we would use the client's cache.
https://github.com/gardener/gardener/blob/d311408c6c9a25bc6e204d9599b745168ad2b97b/extensions/pkg/handler/mapper.go#L66

timebertt: The bigger problem is this `Get` call here, as this is an unstructured request and will therefore not use the client's cache:
https://github.com/gardener/gardener/blob/e408106d170496145fac1abe5cf2e1ccae3088fd/extensions/pkg/controller/healthcheck/reconciler.go#L90-L98

Bad news is, this `Get` call is done once per sync period (`30s` as Daniel said) for each extension type, for which a health check is registered.
Good news is, this is easy to fix, as we would only have to replace the first to lines in `Reconcile` by this:
```go
rawExtension := r.registeredExtension.getExtensionObjFunc()
```
Should I open a separate PR for that, as this has nothing to do with this PR?

rfranzke: Very nice digging @tim-ebert!
If it's only this one line then it can probably be included in this PR, but it's also fine to keep it separate if @danielfoehrKn prefers that. Please sync with him 👍 

One question, you said that the `SharedIndexInformer`s are created under the hood when creating a new managers, and that such an index informer is used for every `GroupVersionKind`. How does this work internally, i.e., how does the controller-runtime know which objects to cache? According to my understanding, keeping such a cache up-to-date requires a WATCH for the objects, right? I don't believe it starts a WATCH for ALL API objects by default, or? So how does this work?

timebertt: The `SharedIndexInformer`s are actually created on demand by the cache (wich basically is nothing more than a set of informers). So every time you call `client.Get()`/`client.List()` on a cached client or `controller.Watch()`, the cache checks, if it already has a corresponding informer for the respectie gvk or it creates a new one, which will then basically result in a `WATCH` call to the API

timebertt: Ok, I'll open a new PR for fixing the get call, as it requires more work than a single line and this PR has still open questions, if I see it correctly.

ialidzhikov:
/needs/rebase

ialidzhikov:
Is there anything else (except the nit comments) blocking the merge of this PR?

rfranzke:
Rebase needed

ialidzhikov (approved):
/lgtm

rfranzke (approved):

PR #2258, opened 2020-04-29
Skip infrastructure reconciliation for hibernated shoots

### LINKED ISSUE #2253: ☂️The Infrastructure is reconciled constantly when a Shoot is hibernated and terraform apply pod returns error
**What happened**:
After triggering hibernation the terraform pod is spawn, although it is not needed. If error occurs during execution of this pod this will continue each 30 seconds. During this infinite loop a logs will be created which will load the fluentd buffer with no end point messages, because elastic-search pod is scaled to 0. 
No terraform should be triggered unless Infrastructure is marked for deletion.
Make your own shoot in Gardener WebUI with valid credentials. 
Hibernate the shoot.
Change the credentials in cloudprovider secret with fake ones.
Annotate the Infrastructure with `gardener.cloud/operation=reconcile`

**Environment**:

- Gardener version: 1.3.1
- Kubernetes version (use `kubectl version`): 1.15.11
- Cloud provider or hardware configuration: azure
- Others:

rfranzke:
The infrastructure reconciliation task will now only be added if the shoot is not hibernated. Also, the gardenlet will request infrastructure reconciliation when the shoot is waking up.

Fixes #2253

**Release note**:

```improvement user
The infrastructure reconciliation for hibernated shoots is now skipped.
```

### DIFF
--- a/pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
+++ b/pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
@@ -172,10 +172,11 @@ func (c *defaultMaintenanceControl) Maintain(shootObj *gardencorev1beta1.Shoot, 
 			return nil, fmt.Errorf("auto update section of Shoot %s/%s changed mid-air", s.Namespace, s.Name)
 		}
 
-		delete(s.Annotations, v1beta1constants.GardenerOperation)
-		controllerutils.AddTasks(s.Annotations, common.ShootTaskDeployInfrastructure)
 		s.Annotations[v1beta1constants.GardenerOperation] = common.ShootOperationReconcile
 
+		if !gardencorev1beta1helper.HibernationIsEnabled(s) {
+			controllerutils.AddTasks(s.Annotations, common.ShootTaskDeployInfrastructure)
+		}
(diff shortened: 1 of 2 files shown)


pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
```diff
@@ -172,10 +172,11 @@ func (c *defaultMaintenanceControl) Maintain(shootObj *gardencorev1beta1.Shoot,
 			return nil, fmt.Errorf("auto update section of Shoot %s/%s changed mid-air", s.Namespace, s.Name)
 		}
 
-		delete(s.Annotations, v1beta1constants.GardenerOperation)
```
danielfoehrKn, line 175: Looks good, the only thing : why is the operations not removed any more?

rfranzke: It doesn't make sense to explicitly remove it because, in the next line, we add it again with the "reconcile" value. Hence, we can overwrite it directly and get the same result.

timebertt (approved):
/lgtm

PR #2261, opened 2020-04-30
"Required" condition for ControllerInstallation to prevent early deletion

### LINKED ISSUE #2218: Gardener Controller Manager should not delete ControllerInstallation right away
**What happened**:

When removing the Containerruntime support for a Shoot, the GCM determined that the gVisor extension is not required any more (only Shoot using ContainerRuntimes).
The gVisor extension is removed immediately.

The reconcile of the Shoot fails, because the ContainerRuntime will never be reconciled/ deleted.
The GCM waits for the Shoot to be reconciled successfully before removing the ControllerInstallation.

1. Create a Shoot with gVisor as a Container Runtime
2. Make sure it is the only Shoot having gVisor enabled.
3. Delete gVisor for the worker pool.
4. Observe the ControllerInstallation for extension-gvisor is deleted right away
5. Observe that the Shoot reconcile does not complete
image

Even though I first encountered that behaviour using the gVisor ControllerRegistration, it is not specific to it.  A general solution is needed.

**Environment**:

- Gardener version: master
- Others:

rfranzke:
With #1993 the logic of the GCM was adapted to only deploy the required extension controllers to the seeds (instead of all like it was earlier). The way it is done is by reacting on `BackupBucket`/`BackupEntry`/`Shoot` objects and analyzing the `spec`. The outcome is the list of required `ControllerRegistrations`.

However, some of the fields in the (shoot) `spec` allow to dynamically add or remove certain extension types, for example, `.spec.extensions[]` or `.spec.provider.workers[].cri.containerRuntimes[]`. Above mentioned logic might decide that a certain `ControllerRegistration` is no longer required although the corresponding extension CR object still exists in the seed (because the shoot was not yet reconciled which would delete it). This results in shoot reconciliations being stuck because the responsible extension controllers were already deleted by GCM.

As per architecture of Gardener, GCM runs in the garden cluster while the Gardenlet runs in the seed clusters. The latter might be unreachable (network-wise) from the garden cluster. Consequently, GCM cannot "look" into the seed and check if extension CRs still exist before taking the "delete extension controller" decision.

This PR introduces a new `Required` condition type for the `ControllerInstallation`s. The gardenlet will set it to `True` if the seed still has extension objects that the extension controller backed by the `ControllerInstallation` is responsible for. Once all extension objects are deleted this condition will be set to `False`. The GCM takes this condition now into account when computing the list of required extensions. This ensures that extension controllers are only deleted once all their CRs are deleted.

Fixes #2218

The PR leverages the `federatedseed` controller that was earlier contributed by @swilen-iwanow @plkokanov @vlvasilev as part of #1631. So far, it was only having one control loop, the `ShootState` synchronization for extension objects.
As part of this PR I've slightly refactored it to allow adding a new control loop which also watches the extension objects and computes the mentioned `Required` condition.
The `federatedseed` controller spawns for every seed cluster a new controller that watches all extension objects in the seed. This new controller now has the two mentioned control loops. 

/cc @danielfoehrKn @tim-ebert 

**Release note**:

```improvement operator
An issue has been fixed that may have resulted in the deletion of extension controllers from a seed cluster although the seed still had extension objects that the extension controller was responsible for. 
```


pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+
+			controllerRegistration := &gardencorev1beta1.ControllerRegistration{}
+			if err := c.k8sGardenClient.Client().Get(ctx, kutil.Key(controllerInstallation.Spec.RegistrationRef.Name), controllerRegistration); err != nil {
+				return reconcile.Result{}, err
+			}
+
+			controllerInstallationRequired[controllerInstallation] = false
+			for _, resource := range controllerRegistration.Spec.Resources {
+				if requiredTypes, ok := c.requiredKindTypes[resource.Kind]; ok && requiredTypes != nil && requiredTypes.Has(resource.Type) {
```
plkokanov: Shouldn't we also lock reads from map so that they don't happen during concurrent writes (with RWMutex for reads/writes)?

rfranzke:
/kind/bug

plkokanov (changes requested):
Thanks, really good PR  :)

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
rfranzke: Yeah, probably it's better to have it, thanks for spotting. I added it now, PTAL.

plkokanov (dismissed):

pkg/apis/core/types_controllerinstallation.go
```diff
@@ -69,4 +69,7 @@ const (
 	ControllerInstallationInstalled ConditionType = "Installed"
 	// ControllerInstallationValid is a condition type for indicating whether the installation request is valid.
 	ControllerInstallationValid ConditionType = "Valid"
+	// ControllerInstallationRequired is a condition type for indicating that the respective extension controller is
+	// still required on the seed cluster as extension custom resources still exist.
```
timebertt:
```suggestion
	// still required on the seed cluster as corresponding extension resources still exist.
```

pkg/controllermanager/controller/controllerregistration/controllerinstallation_control_test.go
```diff
@@ -0,0 +1,126 @@
...
+			Expect(queue.items[0]).To(Equal(seedName))
+		})
+
+		It("should not add the new obj to the queue because required condition unchanged", func() {
+			oldObj := &gardencorev1beta1.ControllerInstallation{}
+			newObj := &gardencorev1beta1.ControllerInstallation{}
+
+			c.controllerInstallationUpdate(oldObj, newObj)
+
+			Expect(queue.Len()).To(Equal(0))
```
timebertt:
```suggestion
			Expect(queue.Len()).To(BeZero())
```

pkg/gardenlet/controller/federatedseed/extensions/artifact.go
```diff
@@ -0,0 +1,144 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package extensions
+
+import (
+	"fmt"
+
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
```
timebertt: Move down to 3rd party import block

pkg/gardenlet/controller/federatedseed/extensions/extensions.go
```diff
@@ -0,0 +1,162 @@
...
+	controllerInstallationControl controllerInstallationControl
+	shootStateControl             shootStateControl
+}
+
+// NewController creates new controller that syncs extensions states to ShootState
+func NewController(ctx context.Context, gardenClient, seedClient kubernetes.Interface, seedName string, extensionsInformers extensionsinformers.SharedInformerFactory, log *logrus.Entry, recorder record.EventRecorder) *Controller {
+	controller := &Controller{
+		log: log,
+
+		waitGroup: sync.WaitGroup{},
```
timebertt: `sync.WaitGroup` does not have to be initialized, I think. The zero value is already usable.
```suggestion
```

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
```diff
@@ -0,0 +1,133 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type controllerInstallationControl struct {
+	k8sGardenClient kubernetes.Interface
+	seedClient      kubernetes.Interface
+	seedName        string
+	log             *logrus.Entry
+
+	lock              *sync.RWMutex
```
timebertt: Any specific reason, why you are using a pointer to a `sync.RWMutex`?
```suggestion
	lock              sync.RWMutex
```

pkg/gardenlet/controller/federatedseed/extensions/extensions.go
```diff
@@ -0,0 +1,162 @@
...
+
+// Controller watches the extension resources and has several control loops.
+type Controller struct {
+	log *logrus.Entry
+
+	waitGroup              sync.WaitGroup
+	workerCh               chan int
+	numberOfRunningWorkers int
+
+	controllerArtifacts           *controllerArtifact
```
timebertt: I think you don't need a pointer here, as `controllerArtifact` anyways contains only reference-type values (map and slice), correct?
```suggestion
	controllerArtifacts           controllerArtifact
```

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
```diff
@@ -0,0 +1,133 @@
...
+
+		controllerInstallationList := &gardencorev1beta1.ControllerInstallationList{}
+		if err := c.k8sGardenClient.Client().List(ctx, controllerInstallationList); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		controllerInstallationRequired := map[*gardencorev1beta1.ControllerInstallation]bool{}
+
+		for _, obj := range controllerInstallationList.Items {
+			controllerInstallation := obj.DeepCopy()
```
timebertt: Any specific reason, why you are `DeepCopy`ing the installation here?

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
```diff
@@ -0,0 +1,133 @@
...
+			return nil
+		}); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		c.lock.Lock()
+		c.requiredKindTypes[kind] = requiredTypes
+		c.lock.Unlock()
+
+		// Step 2: List all existing controller installation objects for the seed cluster this controller is responsible
```
timebertt: Hmm, this second step is basically independent of the `kind` this reconciler is acting on, but is executed for every extensions kind we have.
WDYT about adding a second queue and worker for ControllerInstallations, that executes only the second step in a periodical manner (like every 10s) for each installation, instead of going through each installation for every added/deleted extension object?

pkg/gardenlet/controller/federatedseed/seed_control.go
```diff
@@ -217,27 +217,27 @@ func (f *federatedSeedControllerManager) createControllers(ctx context.Context,
...
-	shootStateSyncController := shootstate.NewSyncController(ctx, k8sGardenClient, k8sSeedClient, extensionsClient, extensionsInformers, config, seedLogger, recorder)
+
+	extensionsController := extensions.NewController(ctx, k8sGardenClient, k8sSeedClient, seedName, extensionsInformers, seedLogger, recorder)
 
 	childCtx, cancelFunc := context.WithCancel(ctx)
-	logger.Logger.Infof("Run ShootStateSync controller for seed: %s", seedName)
-	if err := shootStateSyncController.Run(childCtx, *config.Controllers.ShootStateSync.ConcurrentSyncs); err != nil {
-		logger.Logger.Infof("There was error running the ShootStateSync controller for seed: %s", seedName)
+	logger.Logger.Infof("Run extensions controller for seed: %s", seedName)
+	if err := extensionsController.Run(childCtx, *config.Controllers.ShootStateSync.ConcurrentSyncs, *config.Controllers.ShootStateSync.ConcurrentSyncs); err != nil {
```
timebertt: Should we use separate config fields for the number of ShootState workers and ControllerInstallation workers?

timebertt (commented):
Looks quite good in general.
Only some nits and one general question wrt the new controller.

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
rfranzke: This will result in linting issues:

```
$ make check
> Check
Executing golangci-lint
pkg/gardenlet/controller/federatedseed/extensions/extensions.go:101:46: copylocks: call of s.createControllerInstallationWorkers copies lock value: github.com/gardener/gardener/pkg/gardenlet/controller/federatedseed/extensions.controllerInstallationControl contains sync.RWMutex (govet)
		s.createControllerInstallationWorkers(ctx, s.controllerInstallationControl)
		                                           ^
pkg/gardenlet/controller/federatedseed/extensions/extensions.go:112:87: copylocks: createControllerInstallationWorkers passes lock by value: github.com/gardener/gardener/pkg/gardenlet/controller/federatedseed/extensions.controllerInstallationControl contains sync.RWMutex (govet)
func (s *Controller) createControllerInstallationWorkers(ctx context.Context, control controllerInstallationControl) {
                                                                                      ^
make: *** [check] Error 1
```

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
rfranzke: The `ControllerInstallation` object is later stored in the `controllerInstallationRequired` map as key. I have to use the pointer type for the map key type (otherwise `==` operator is not implemented for the map key type which results in a compilation error). `obj` is only of type `gardencorev1beta1.ControllerInstallation`, and using `&obj` would take the address of the iteration variable, leading to undesired results.

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
rfranzke: Hm, I didn't like the periodic idea too much as it exposes the risk of creating new race conditions. But you were bringing up a valid point which was that this reconciler was doing too much (not only handling the `kind` it was created for but all others as well).
I refactored it a little bit and changed it in the following way:

1. Compute all required types of existing extension objects for the reconciler `kind`.
1. List all `ControllerRegistration`s and store the names + extension types of those that are responsible for resources of the reconciler `kind`.
1. List all `ControllerInstallation`s, filter out those of different seeds and those that are referencing unrelated registrations (as of the previous step). Now only those `ControllerInstallation`s are handled that reference a registration that is responsible for the reconciler `kind`. If its supported extension types are part of the required list (computed in step 1) then set the `Required` condition to `True`, otherwise to `False`.

This way the reconcilers are only doing the job they were created for, and I also got rid of the `sync.RWMutex`.

Also, these loops are not happening too often (only if new extension objects are created/deleted or if the `.spec.type` changes (which never happens)).

pkg/gardenlet/apis/config/v1alpha1/types.go
```diff
@@ -133,7 +136,7 @@ type GardenletControllerConfiguration struct {
 	// ShootCare defines the configuration of the ShootCare controller.
 	// +optional
 	ShootCare *ShootCareControllerConfiguration `json:"shootCare,omitempty"`
-	// ShootStateSync defines the configuration of the ShootStateController controller
+	// ShootStateSync defines the configuration of the ShootStateController controller/
```
timebertt:
```suggestion
	// ShootStateSync defines the configuration of the ShootStateController controller.
```

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
timebertt: Can you maybe rather use the installation's name as the key for the map like in the registration map above? This seems cleaner to me.

rfranzke:
/status/author-action

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
rfranzke: As discussed in our call, the object is needed later to update the status. Hence, keeping it as it is for now.

rfranzke:
@tim-ebert I've made the changes we discussed. Kindly PTAL again.

charts/gardener/gardenlet/charts/runtime/templates/configmap-componentconfig.yaml
```diff
@@ -69,6 +69,10 @@ data:
         concurrentSyncs: {{ required ".Values.global.gardenlet.config.controllers.controllerInstallationCare.concurrentSyncs is required" .Values.global.gardenlet.config.controllers.controllerInstallationCare.concurrentSyncs }}
         syncPeriod: {{ required ".Values.global.gardenlet.config.controllers.controllerInstallationCare.syncPeriod is required" .Values.global.gardenlet.config.controllers.controllerInstallationCare.syncPeriod }}
       {{- end }}
+      {{- if .Values.global.gardenlet.config.controllers.controllerInstallationCare }}
```
timebertt:
```suggestion
      {{- if .Values.global.gardenlet.config.controllers. controllerInstallationRequired }}
```

timebertt (commented):
One nit, /lgtm otherwise

charts/gardener/gardenlet/charts/runtime/templates/configmap-componentconfig.yaml
rfranzke: Thanks for catching, done.

timebertt (dismissed):
Thanks, /approve

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
```diff
@@ -0,0 +1,213 @@
...
+		message = "extension objects exist in the seed having the kind/type combinations the controller is responsible for"
+	)
+
+	if !required {
+		status = gardencorev1beta1.ConditionFalse
+		reason = "NoExtensionObjects"
+		message = "no " + message
+	}
+
+	return kutil.TryUpdateStatus(ctx, retry.DefaultBackoff, c, controllerInstallation, func() error {
```
danielfoehrKn, line 209: Why not just setting a deletion timestamp on the ControllerInstallation instead of updating the Condition? In the end the Gardenlet in the Seed is the only one who know if the ControllerInstallation is still required.

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+}
+
+func updateControllerInstallationRequiredCondition(ctx context.Context, c client.Client, controllerInstallation *gardencorev1beta1.ControllerInstallation, required bool) error {
+	var (
+		conditionRequired = gardencorev1beta1helper.GetOrInitCondition(controllerInstallation.Status.Conditions, gardencorev1beta1.ControllerInstallationRequired)
+
+		status  = gardencorev1beta1.ConditionTrue
+		reason  = "ExtensionObjectsExist"
+		message = "extension objects exist in the seed having the kind/type combinations the controller is responsible for"
```
danielfoehrKn: Can we make that more clear? Also including the concrete Extension kinds and types would be helpful.
Something like: "Extension objects exists  on seed 'seed-aws' that require the Controller Installation: [aws-containerrruntime, aws-infrastructure]."

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
```diff
@@ -0,0 +1,213 @@
...
+	)
+
+	if !required {
+		status = gardencorev1beta1.ConditionFalse
+		reason = "NoExtensionObjects"
+		message = "no " + message
```
danielfoehrKn: Also adjust accordingly.

danielfoehrKn (changes requested):
Works perfectly - thanks. Just some minor comments.

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
rfranzke: I don't want to mix responsibilities here. There is the `ControllerRegistration` controller in the Gardener Controller Manager that is responsible for creating and deleting `ControllerInstallation`s at the right point in time. Especially with #2278 it will evaluate way more settings (e.g., deployment policy) that are relevant for the decision - not only if there are still extension resources pending.

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
rfranzke: done

timebertt (approved):

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
```diff
@@ -0,0 +1,216 @@
...
+		if required == nil {
+			if !allKindsCalculated {
+				// if required wasn't set yet then but not all kinds were calculated then the it's not possible to
+				// decide yet whether it's required or not
+				return reconcile.Result{}, nil
+			}
+
+			// if required wasn't set yet then but all kinds were calculated then the installation is no longer required
+			required = pointer.BoolPtr(false)
+			message = "no extension objects exist in the seed having the kind/type combinations the controller is responsible for"
```
danielfoehrKn, line 182:
```suggestion
			message = "No extension objects exist in the seed having the kind/type combinations the controller is responsible for."
```

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
```diff
@@ -0,0 +1,216 @@
...
+		} else if required != nil && *required {
+			message = fmt.Sprintf("extension objects still exist in the seed: %+v", requiredKindTypes.UnsortedList())
```
danielfoehrKn, line 184:
```suggestion
			message = fmt.Sprintf("Extension objects still exist in the seed: %+v.", requiredKindTypes.UnsortedList())
```

danielfoehrKn (changes requested):

pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go
rfranzke: Kubelet doesn't use this style either and I'd keep the message style similar to [Golang error strings](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

rfranzke:
Any further comments? Otherwise, I'd like to merge this PR.

timebertt:
Nope, not from my side. 
Go ahead :)

PR #2260, opened 2020-04-30
[ci:component:github.com/gardener/etcd-backup-restore:0.8.0->v0.9.0]

gardener-robot-ci-2:
*Release Notes*:
``` improvement operator github.com/gardener/etcd-backup-restore #222 @swapnilgm
Skip the first full snapshot on start,  if initial delta snapshot is taken and last full snapshot is recent enough than 24hr.
```

``` improvement operator github.com/gardener/etcd-backup-restore #217 @shreyas-s-rao
Add documentation to force restore etcd data.
```

``` improvement developer github.com/gardener/etcd-backup-restore #215 @shreyas-s-rao
Fix integration test setup script.
```

``` noteworthy user github.com/gardener/etcd-backup-restore #214 @shreyas-s-rao
HTTP API for triggering out-of-schedule full and delta snapshots now returns snapshot metadata in response body in JSON format.
```

``` improvement user github.com/gardener/etcd-backup-restore #214 @shreyas-s-rao
Added new HTTP API for fetching details of latest full and delta snapshots, in JSON format.
```

``` improvement user github.com/gardener/etcd-backup-restore #211 @shreyas-s-rao
Add metrics `etcdbr_snapstore_latest_deltas_total` and `etcdbr_snapstore_latest_deltas_revisions_total` to provide information about the delta snapshots since the latest full snapshot in the snapstore.
```

``` improvement operator github.com/gardener/etcd-backup-restore #210 @ialidzhikov
The release tags from now are prefixed with `v`.
```

``` improvement operator github.com/gardener/etcd-backup-restore #208 @swapnilgm
Configuring backup-restore server using config file is now supported.
```

PR #2036, opened 2020-03-11
Allow injecting image vector overwrite into etcd-druid

rfranzke:
**What this PR does / why we need it**:
With #1762 we have introduced the etcd-druid in `v0.1.3` which is responsible for deploying the etcd `StatefulSet`s, `ConfigMap`s, etc. However, we are still controlling both etcd and the backup-restore sidecar image versions in our `charts/images.yaml`. This is quite dangerous as these versions might not fit to the manifests/configuration applied by the etcd-druid. Instead, it's better to let the etcd-druid control the versions.

@georgekuruvillak is working on etcd-druid side to also support an image vector, similar to Gardener. Consequently, it would also support an image vector overwrite in case an operator wants to customize the image locations. However, so far there is no mechanism to provide such an image vector overwrite to components that are deployed by Gardener.

This PR allows to specify image vector overwrites for components directly deployed by Gardenlet that support image vector overwritings (planned only for etcd-druid, right now).

**Special notes for your reviewer**:
In its current `v0.1.3` version, the etcd-druid is not yet capable of picking up the images from the provided overwrite. @georgekuruvillak is working on introducing the same, and once a new etcd-druid lands in g/g we can remove both `etcd` and `backup-restore` images from our `charts/images.yaml`. Together with this change the `Etcd` will no longer contain `.spec.etcd.image` and `.spec.backup.image` - instead, the druid will implicitly control which images are deployed.

**Release note**:

```improvement operator
It is now possible to specify an image vector overwrite for image vectors of components directly deployed by Gardener (e.g., the etcd-druid will have its own image vector in the future that operators might want to overwrite with custom image locations). Please consult [this document](https://github.com/gardener/gardener/blob/master/docs/deployment/image_vector.md#image-vectors-for-dependent-components) to get more information.
```

PR #2262, opened 2020-04-30
Remove etcd and etcd-backup-restore images

rfranzke:
After the introduction of `etcd-druid` we decided that it does not make sense to let Gardener control the versions of the `etcd` and `etcd-backup-restore` images as their manifests/configuration are controlled by `etcd-druid`. It is necessary for a successful deployment that the version works well with the manifests/configuration.
Consequently, `etcd-druid` was meanwhile made capable of controlling these versions using the image vector concept (see gardener/etcd-druid#23 and gardener/etcd-druid#24).
Gardener already supports overwriting the image vector for components it deploys, see #2036 and https://github.com/gardener/gardener/blob/master/docs/deployment/image_vector.md#image-vectors-for-dependent-components.
This PR now removes the `etcd` and `etcd-backup-restore` images from Gardener so that `etcd-druid` can take over.

/cc @shreyas-s-rao @swapnilgm @georgekuruvillak - once this PR is merged you need to release a new version of etcd-druid when you want Gardener to use a newer/different version of etcd or etcd-backup-restore. There won't be PRs such as #2260 anymore to this repository. Instead, they should be opened to the [gardener/etcd-druid](https://github.com/gardener/etcd-druid) repository. I opened gardener/etcd-druid#50 to track this.

ℹ️Softly depends on #2263

**Release note**:

```noteworthy operator
Gardener's image vector does no longer contain `etcd` and `etcd-backup-restore`. If you want to overwrite these versions you must do it via the `etcd-druid`. Please consult [this document](https://github.com/gardener/gardener/blob/master/docs/deployment/image_vector.md#image-vectors-for-dependent-components) for more information.
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -31,14 +31,6 @@ images:
   tag: "0.4.0"
 
 # Seed controlplane
-- name: etcd
-  sourceRepository: github.com/etcd-io/etcd
-  repository: quay.io/coreos/etcd
-  tag: v3.3.17
-- name: etcd-backup-restore
-  sourceRepository: github.com/gardener/etcd-backup-restore
-  repository: eu.gcr.io/gardener-project/gardener/etcdbrctl
-  tag: "v0.9.0"
 - name: hyperkube # used for kubectl + kubelet binaries on the worker nodes
   sourceRepository: github.com/kubernetes/kubernetes
   repository: k8s.gcr.io/hyperkube
--- a/charts/seed-controlplane/charts/etcd/templates/etcd.yaml
+++ b/charts/seed-controlplane/charts/etcd/templates/etcd.yaml
@@ -33,7 +33,6 @@ spec:
     networking.gardener.cloud/to-public-networks: allowed
     networking.gardener.cloud/to-private-networks: allowed
   etcd:
-    image: {{ index .Values.images "etcd" }}
 {{- if .Values.etcd.resources }}
     resources:
 {{ toYaml .Values.etcd.resources | indent 6 }}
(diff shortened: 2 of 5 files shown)


rfranzke:
Important note: The version of `etcd-druid` that Gardener is using is still referencing `etcd-backup-restore:0.8.0`. Hence, merging this PR without a new `etcd-druid` release would result in Gardener not deploying the `v0.9.0` version.

@shreyas-s-rao @georgekuruvillak @swapnilgm can you please release a new druid version now that https://github.com/gardener/etcd-druid/pull/51 is merged?

rfranzke:
Thanks!

swapnilgm (changes requested):
Simply removing injection won't work right? Since it won't remove the configured etcd image from existing etcd resource. Since druid will see that for exsiting shoot etcd/backup image is already confgirued in CRD, so if won't apply it from the imagevctor.  We will have to delete and recreate old etcd's with image injection.

rfranzke:
What do you mean? I have removed the `image` field of the `Etcd` resource, it's part of the PR, so after shoot reconciliation this field will be omitted and the version of druid will be taken.

> Since druid will see that for exsiting shoot etcd/backup image is already confgirued in CRD, so if won't apply it from the imagevctor. We will have to delete and recreate old etcd's with image injection.

I don't understand - can you explain what you mean? Did you test it locally? In my tests, I was reconciling a shoot and it didn't restart (as expected, because I didn't specify an imagevector overwrite for druid). Then I added such an imagevector overwrite and I saw druid being restarted. In the subsequent shoot reconciliation also the etcds got restarted.

swapnilgm:
ok Ignore it. I had misunderstanding about how gardener applies the charts/templates.

swapnilgm (approved):
lgtm

ialidzhikov (approved):
/lgtm

Issue #2145, opened 2020-03-30
ShootNotFailed predicate not working as expected


rfranzke:
**What happened**:
The [`ShootNotFailed` predicate](https://github.com/gardener/gardener-extensions/blob/master/pkg/controller/controlplane/controller.go#L65) does not work as expected.
If a reconciliation starts and errors then it gets requeued with exponential backoff. However, the predicates are bypassed, hence, reconciliation will be tried forever (unless the controller restarts - then the queue is empty and predicates are considered).

**What you expected to happen**:
If the `lastOperation.state` value of the `Cluster.shoot.status` is changed to `Failed` then reconciliation should not happen anymore unless the `state` is no longer `Failed`.

**How to reproduce it (as minimally and precisely as possible)**:
1. Trigger a reconciliation that is erroring (and requeued with exp. backoff)
1. Change the `lastOperation.state` to `Failed` in the `Cluster` resource
1. See the controller not stopping reconciling but continuing.

The same could be fixed for the https://github.com/gardener/etcd-druid as well, see https://github.com/gardener/etcd-druid/pull/33#discussion_r399256650.

rfranzke:
/close
as related PRs have been merged

PR #2265, opened 2020-04-30
Fix ShootNotFailed extension predicate

rfranzke:
The `ShootNotFailed` predicate was flawed and did not work as expected. This PR fixes this and adds proper unit tests to prevent future regressions. Detected as part of the review for #2217, thanks @ialidzhikov for spotting!

Related to #2145 ❓

/cc @danielfoehrKn, probably is related to #2145 which is what you are currently looking at.

**Release note**:

```improvement developer
The `ShootNotFailed` predicate in the extensions library does now work as expected.
```


rfranzke:
/kind/bug

extensions/pkg/predicate/predicate_test.go
```diff
@@ -12,25 +12,47 @@
...
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	"github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	mockcache "github.com/gardener/gardener/pkg/mock/controller-runtime/cache"
+	mockclient "github.com/gardener/gardener/pkg/mock/controller-runtime/client"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+
+	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
```
timuthy, line 27: Was that change intended? If yes, why?

timuthy (commented):
Looks good, only one question.

extensions/pkg/predicate/predicate_test.go
rfranzke: Yes, intended. I changed the package of the `predicate_test.go` file from `predicate_test` to `predicate` to be able to test the unexported function.
Keeping the dot import for `gomega` would now introduce a name conflict as the `predicate` package already defines an `Or` function (similarly to the `gomega` package).

timuthy (approved):
/lgtm

timebertt (approved):
/lgtm

PR #2270, opened 2020-04-30
Improve webhook hibernation validation

### LINKED ISSUE #1751: Improve Shoot Hibernation Prevention Check

mvladev:

Improve hibernation webhook check filter

Webhooks with `failurePolicy: Fail` stop hibernation if they select any of the following resources:

- `apiservices`
- `apiservices/status`
- `certificatesigningrequests`
- `certificatesigningrequests/approval`
- `certificatesigningrequests/status`
- `clusterrolebindings`
- `clusterroles`
- `configmaps` (only for `kube-system` namespace)
- `controllerrevisions` (only for `kube-system` namespace)
- `customresourcedefinitions`
- `customresourcedefinitions/status`
- `daemonsets` (only for `kube-system` namespace)
- `daemonsets/status` (only for `kube-system` namespace)
- `deployments` (only for `kube-system` namespace)
- `deployments/scale` (only for `kube-system` namespace)
- `endpoints`
- `leases`
- `namespaces`
- `namespaces/finalize`
- `namespaces/status`
- `networkpolicies` (only for `kube-system` namespace)
- `nodes`
- `nodes/status`
- `pods` (only for `kube-system` namespace and `shoot.gardener.cloud/no-cleanup=true,orgin=gardener` labels)
- `pods/status` (only for `kube-system` namespace and `shoot.gardener.cloud/no-cleanup=true,orgin=gardener` labels)
- `podsecuritypolicies`
- `priorityclasses`
- `replicasets` (only for `kube-system` namespace)
- `replicasets/scale` (only for `kube-system` namespace)
- `replicasets/status` (only for `kube-system` namespace)
- `rolebindings` (only for `kube-system` namespace)
- `roles` (only for `kube-system` namespace)
- `secrets` (only for `kube-system` namespace)
- `serviceaccounts` (only for `kube-system` namespace)
- `services`
- `services/status`

Fixes #1751

**Release note**:

```noteworthy user
Hibernation blocking due to Mutating/Validating webhooks is now improved and enforced for hooks with `failurePolicy: Fail` and operation `CREATE`, `UPDATE` or `*` for the following resources:

- `apiservices`
- `apiservices/status`
- `certificatesigningrequests`
- `certificatesigningrequests/approval`
- `certificatesigningrequests/status`
- `clusterrolebindings`
- `clusterroles`
- `configmaps` (only for `kube-system` namespace)
- `controllerrevisions` (only for `kube-system` namespace)
- `customresourcedefinitions`
- `customresourcedefinitions/status`
- `daemonsets` (only for `kube-system` namespace)
- `daemonsets/status` (only for `kube-system` namespace)
- `deployments` (only for `kube-system` namespace)
- `deployments/scale` (only for `kube-system` namespace)
- `endpoints`
- `leases`
- `namespaces`
- `namespaces/finalize`
- `namespaces/status`
- `networkpolicies` (only for `kube-system` namespace)
- `nodes`
- `nodes/status`
- `pods` (only for `kube-system` namespace and `shoot.gardener.cloud/no-cleanup=true,orgin=gardener` labels)
- `pods/status` (only for `kube-system` namespace and `shoot.gardener.cloud/no-cleanup=true,orgin=gardener` labels)
- `podsecuritypolicies`
- `priorityclasses`
- `replicasets` (only for `kube-system` namespace)
- `replicasets/scale` (only for `kube-system` namespace)
- `replicasets/status` (only for `kube-system` namespace)
- `rolebindings` (only for `kube-system` namespace)
- `roles` (only for `kube-system` namespace)
- `secrets` (only for `kube-system` namespace)
- `serviceaccounts` (only for `kube-system` namespace)
- `services`
- `services/status`
```

### DIFF
--- a/pkg/operation/botanist/matchers/matcher.go
+++ b/pkg/operation/botanist/matchers/matcher.go
@@ -1,0 +1,252 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package matchers
+
+import (
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	botanistconstants "github.com/gardener/gardener/pkg/operation/botanist/constants"
+	"github.com/gardener/gardener/pkg/operation/common"
+
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	appsv1 "k8s.io/api/apps/v1"
+	appsv1beta1 "k8s.io/api/apps/v1beta1"
+	appsv1beta2 "k8s.io/api/apps/v1beta2"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	autoscalingv2beta1 "k8s.io/api/autoscaling/v2beta1"
+	autoscalingv2beta2 "k8s.io/api/autoscaling/v2beta2"
+	certificatesv1beta1 "k8s.io/api/certificates/v1beta1"
+	coordinationv1 "k8s.io/api/coordination/v1"
+	coordinationv1beta1 "k8s.io/api/coordination/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	eventsv1beta1 "k8s.io/api/events/v1beta1"
+	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	networkingv1 "k8s.io/api/networking/v1"
+	networkingv1beta1 "k8s.io/api/networking/v1beta1"
+	nodev1alpha1 "k8s.io/api/node/v1alpha1"
+	nodev1beta1 "k8s.io/api/node/v1beta1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	rbacv1alpha1 "k8s.io/api/rbac/v1alpha1"
+	rbacv1beta1 "k8s.io/api/rbac/v1beta1"
+	schedulingv1 "k8s.io/api/scheduling/v1"
+	schedulingv1alpha1 "k8s.io/api/scheduling/v1alpha1"
+	schedulingv1beta1 "k8s.io/api/scheduling/v1beta1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
+	apiregistrationv1beta1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1beta1"
+	metricsv1alpha1 "k8s.io/metrics/pkg/apis/metrics/v1alpha1"
+	metricsv1beta1 "k8s.io/metrics/pkg/apis/metrics/v1beta1"
+)
+
+var (
+	kubeSystemLabels = labels.Set{
+		common.ShootNoCleanup:            "true",
+		v1beta1constants.LabelRole:       metav1.NamespaceSystem,
+		v1beta1constants.GardenerPurpose: metav1.NamespaceSystem,
+	}
+	podsLabels = labels.Set{
+		common.ShootNoCleanup:                           "true",
+		botanistconstants.ManagedResourceLabelKeyOrigin: botanistconstants.ManagedResourceLabelValueGardener,
+	}
+	// WebhookConstraintMatchers contains a list of all api resources which can break
+	// the waking up of a cluster.
+	WebhookConstraintMatchers = []WebhookConstraintMatcher{
+		{GVR: schema.GroupVersionResource{Group: "discovery.k8s.io", Version: "v1alpha1", Resource: "endpointslices"}, NamespaceLabels: kubeSystemLabels},
+		{GVR: schema.GroupVersionResource{Group: "discovery.k8s.io", Version: "v1beta1", Resource: "endpointslices"}, NamespaceLabels: kubeSystemLabels},
+		// v1 is still not available, but the APi will be promoted soon
+		{GVR: schema.GroupVersionResource{Group: "discovery.k8s.io", Version: "v1", Resource: "endpointslices"}, NamespaceLabels: kubeSystemLabels},
+
+		{GVR: corev1.SchemeGroupVersion.WithResource("pods"), NamespaceLabels: kubeSystemLabels, ObjectLabels: podsLabels},
+		{GVR: corev1.SchemeGroupVersion.WithResource("pods"), NamespaceLabels: kubeSystemLabels, ObjectLabels: podsLabels, Subresource: "status"},
+		{GVR: corev1.SchemeGroupVersion.WithResource("configmaps"), NamespaceLabels: kubeSystemLabels},
+		{GVR: corev1.SchemeGroupVersion.WithResource("endpoints")},
+		{GVR: corev1.SchemeGroupVersion.WithResource("events")},
+		{GVR: corev1.SchemeGroupVersion.WithResource("secrets"), NamespaceLabels: kubeSystemLabels},
+		{GVR: corev1.SchemeGroupVersion.WithResource("serviceaccounts"), NamespaceLabels: kubeSystemLabels},
+		{GVR: corev1.SchemeGroupVersion.WithResource("services")},
+		{GVR: corev1.SchemeGroupVersion.WithResource("services"), Subresource: "status"},
+		{GVR: corev1.SchemeGroupVersion.WithResource("nodes"), ClusterScoped: true},
+		{GVR: corev1.SchemeGroupVersion.WithResource("nodes"), ClusterScoped: true, Subresource: "status"},
+		{GVR: corev1.SchemeGroupVersion.WithResource("namespaces"), ClusterScoped: true},
+		{GVR: corev1.SchemeGroupVersion.WithResource("namespaces"), ClusterScoped: true, Subresource: "status"},
+		{GVR: corev1.SchemeGroupVersion.WithResource("namespaces"), ClusterScoped: true, Subresource: "finalize"},
+
+		{GVR: eventsv1beta1.SchemeGroupVersion.WithResource("events")},
+
+		{GVR: appsv1.SchemeGroupVersion.WithResource("controllerrevisions"), NamespaceLabels: kubeSystemLabels},
+		{GVR: appsv1.SchemeGroupVersion.WithResource("daemonsets"), NamespaceLabels: kubeSystemLabels},
+		{GVR: appsv1.SchemeGroupVersion.WithResource("daemonsets"), NamespaceLabels: kubeSystemLabels, Subresource: "status"},
+		{GVR: appsv1.SchemeGroupVersion.WithResource("deployments"), NamespaceLabels: kubeSystemLabels},
+		{GVR: appsv1.SchemeGroupVersion.WithResource("deployments"), NamespaceLabels: kubeSystemLabels, Subresource: "scale"},
+		{GVR: appsv1.SchemeGroupVersion.WithResource("replicasets"), NamespaceLabels: kubeSystemLabels},
+		{GVR: appsv1.SchemeGroupVersion.WithResource("replicasets"), NamespaceLabels: kubeSystemLabels, Subresource: "status"},
+		{GVR: appsv1.SchemeGroupVersion.WithResource("replicasets"), NamespaceLabels: kubeSystemLabels, Subresource: "scale"},
+
+		{GVR: appsv1beta1.SchemeGroupVersion.WithResource("controllerrevisions"), NamespaceLabels: kubeSystemLabels},
+		{GVR: appsv1beta1.SchemeGroupVersion.WithResource("daemonsets"), NamespaceLabels: kubeSystemLabels},
+		{GVR: appsv1beta1.SchemeGroupVersion.WithResource("daemonsets"), NamespaceLabels: kubeSystemLabels, Subresource: "status"},
+		{GVR: appsv1beta1.SchemeGroupVersion.WithResource("deployments"), NamespaceLabels: kubeSystemLabels},
+		{GVR: appsv1beta1.SchemeGroupVersion.WithResource("deployments"), NamespaceLabels: kubeSystemLabels, Subresource: "scale"},
+		{GVR: appsv1beta1.SchemeGroupVersion.WithResource("replicasets"), NamespaceLabels: kubeSystemLabels},
+		{GVR: appsv1beta1.SchemeGroupVersion.WithResource("replicasets"), NamespaceLabels: kubeSystemLabels, Subresource: "status"},
+		{GVR: appsv1beta1.SchemeGroupVersion.WithResource("replicasets"), NamespaceLabels: kubeSystemLabels, Subresource: "scale"},
+
+		{GVR: appsv1beta2.SchemeGroupVersion.WithResource("controllerrevisions"), NamespaceLabels: kubeSystemLabels},
+		{GVR: appsv1beta2.SchemeGroupVersion.WithResource("daemonsets"), NamespaceLabels: kubeSystemLabels},
+		{GVR: appsv1beta2.SchemeGroupVersion.WithResource("daemonsets"), NamespaceLabels: kubeSystemLabels, Subresource: "status"},
+		{GVR: appsv1beta2.SchemeGroupVersion.WithResource("deployments"), NamespaceLabels: kubeSystemLabels},
+		{GVR: appsv1beta2.SchemeGroupVersion.WithResource("deployments"), NamespaceLabels: kubeSystemLabels, Subresource: "scale"},
+		{GVR: appsv1beta2.SchemeGroupVersion.WithResource("replicasets"), NamespaceLabels: kubeSystemLabels},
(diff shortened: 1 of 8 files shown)


timebertt:
@mvladev WDYT about marking webhooks with rules for `endpoints` in the `kube-system` namespace as problematic as suggested in https://github.com/gardener/gardener/issues/1751#issuecomment-570268980?

@vpnachev can you sync with @mvladev about your work/progress on this topic, as you assigned yourself to #1751?

rfranzke:
/kind/enhancement

mvladev:
I've updated the PR with almost all resources which can break the cluster.

pkg/operation/botanist/matchers/matcher.go
```diff
@@ -0,0 +1,252 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
+	apiregistrationv1beta1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1beta1"
+	metricsv1alpha1 "k8s.io/metrics/pkg/apis/metrics/v1alpha1"
+	metricsv1beta1 "k8s.io/metrics/pkg/apis/metrics/v1beta1"
+)
+
+var (
+	kubeSystemLabels = labels.Set{
+		"shoot.gardener.cloud/no-cleanup": "true",
```
rfranzke:
```suggestion
		common.ShootNoCleanup: "true",
```

pkg/operation/botanist/matchers/matcher.go
```diff
@@ -0,0 +1,252 @@
...
+		"role":                            "kube-system",
+		"gardener.cloud/purpose":          "kube-system",
```
rfranzke:
```suggestion
		v1beta1.GardenerPurpose:          metav1.NamespaceSystem,
```

pkg/operation/botanist/matchers/matcher.go
```diff
@@ -0,0 +1,252 @@
...
+	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
+	apiregistrationv1beta1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1beta1"
+	metricsv1alpha1 "k8s.io/metrics/pkg/apis/metrics/v1alpha1"
+	metricsv1beta1 "k8s.io/metrics/pkg/apis/metrics/v1beta1"
+)
+
+var (
+	kubeSystemLabels = labels.Set{
+		"shoot.gardener.cloud/no-cleanup": "true",
+		"role":                            "kube-system",
```
rfranzke:
```suggestion
		"role":                            metav1.NamespaceSystem,
```

pkg/operation/botanist/matchers/matcher.go
```diff
@@ -0,0 +1,252 @@
...
+		"gardener.cloud/purpose":          "kube-system",
+	}
+	podsLabels = labels.Set{
+		"shoot.gardener.cloud/no-cleanup": "true",
```
rfranzke:
```suggestion
		common.ShootNoCleanup: "true",
```

pkg/operation/botanist/matchers/matcher.go
```diff
@@ -0,0 +1,252 @@
...
+
+		{GVR: nodev1alpha1.SchemeGroupVersion.WithResource("runtimeclasses"), ClusterScoped: true},
+		{GVR: nodev1beta1.SchemeGroupVersion.WithResource("runtimeclasses"), ClusterScoped: true},
+
+		{GVR: schedulingv1.SchemeGroupVersion.WithResource("priorityclasses"), ClusterScoped: true},
+		{GVR: schedulingv1alpha1.SchemeGroupVersion.WithResource("priorityclasses"), ClusterScoped: true},
+		{GVR: schedulingv1beta1.SchemeGroupVersion.WithResource("priorityclasses"), ClusterScoped: true},
+
+		{GVR: storagev1.SchemeGroupVersion.WithResource("storageclasses"), ClusterScoped: true},
+		{GVR: storagev1beta1.SchemeGroupVersion.WithResource("storageclasses"), ClusterScoped: true},
```
rfranzke: This list seems too large to me. Can you explain how you "calculated" it? I don't think we should include each and every GVK that we apply in the shoot. Instead, we should rather strip this down to only those that REALLY can break the wake-up, i.e. webhook for pod creation in kube-system, etc. If a `StorageClass` cannot be created/updated then it will be visible in the shoot conditions later on and the user can take care.

WDYT @vpnachev @tim-ebert ?

rfranzke (changes requested):

mvladev:
I've updated the PR.

> This list seems too large to me. Can you explain how you "calculated" it? I don't think we should include each and every GVK that we apply in the shoot. Instead, we should rather strip this down to only those that REALLY can break the wake-up, i.e. webhook for pod creation in kube-system, etc. If a StorageClass cannot be created/updated then it will be visible in the shoot conditions later on and the user can take care.

Here is the rationale on why I've added those resources:

- `apiservices`, `apiservices/status` breaks `metrics-server`
- `certificatesigningrequests`, `certificatesigningrequests/approval`, `certificatesigningrequests/status` breaks kubelet joining the cluster
- `clusterrolebindings`, `clusterroles`, `rolebindings`, `roles` (only for `kube-system` namespace) breaks `kube-proxy`, `dns`...
- `configmaps` (only for `kube-system` namespace), breaks `kube-dns`
- `customresourcedefinitions`, `customresourcedefinitions/status` - breaks networking extensions
- `controllerrevisions`, `daemonsets`, `daemonsets/status`,`deployments`,`deployments/scale`, `replicasets` , `replicasets/scale` , `replicasets/status`  (only for `kube-system` namespace) break all workloads in `kube-system`
- `endpoints` - breaks `kube-proxy`, apiserver automatic discovery.
- `endpointslices` (only for `kube-system` namespace) - breaks `kube-proxy` in large clusters
- `events` - doesn't explicitly break anything, but no debugging would be possible
- `horizontalpodautoscalers` (only for `kube-system` namespace), `horizontalpodautoscalers/status` (only for `kube-system` namespace) breaks dns
- `leases` - breaks kubelet healthchecks
- `podmetrics` (only for `kube-system` namespace) - breaks scaling for coredns
- `nodemetrics` - breaks metrics server
- `namespaces`, `namespaces/finalize`, `namespaces/status` breaks apiserver 
- `networkpolicies` (only for `kube-system` namespace) - ,breaks everything in kube-system
- `nodes` `nodes/status` breaks kubelet
- `pods`, `pods/status` (only for `kube-system` namespace and `shoot.gardener.cloud/no-cleanup=true,orgin=gardener` labels) breaks workloads in kube-system
- `podsecuritypolicies` - breaks every workload in the cluster
- `priorityclasses` - breaks every workload in kube-system
- `runtimeclasses` - breaks every workload in kube-system
- `secrets` (only for `kube-system` namespace) - breaks every workload in kube-system
- `serviceaccounts` (only for `kube-system` namespace) - breaks every workload in kube-system
- `services`,`services/status` - breaks every workload in kube-system and kube-apiserver advertising

rfranzke:
We do so many things break the wake-up? Once the webhook pods in question are up again then the rest will probably self-heal if it was blocked before. We "just" have to make sure that the pods can come up, or?
Generally, you might be right that all these things can break a shoot in general, but this can also be "achieved" during runtime and without hibernation at all. The check you are adapting, however, shall detect configurations that would prevent the wake-up.

timebertt:
> This list seems too large to me. Can you explain how you "calculated" it? I don't think we should include each and every GVK that we apply in the shoot. Instead, we should rather strip this down to only those that REALLY can break the wake-up, i.e. webhook for pod creation in kube-system, etc. If a `StorageClass` cannot be created/updated then it will be visible in the shoot conditions later on and the user can take care.
> 
> WDYT @vpnachev @tim-ebert ?

I agree with @rfranzke. The original purpose of the hibernation constraint just was to prevent issues, where users harm themselves in a way that the cluster is not able to be woken up without intervention of an operator. As mentioned, most of the resources you added, will be able to self-heal once the system-components are up. So I think, we should keep this check as small as possible to not restrict the users too much and only keep checks for those resources, which indeed prevent wake-up of system-components (`nodes`, `pods` and `endpoints` in `kube-system`). Mostly everything else will self-heal once `kcm` and `kube-scheduler` are up and running.
If we observe issues with further resources that block wake-up, I am completely open to adding them to the checks.

mvladev:
From the list above, can you list what should **not** be in the things we block?

timebertt:
Most of the resources you list, are not created/updated during wake-up, because they are already in the system, so they would not break the cluster wake-up. That's why I would drop these, as they only put additional restrictions on the user.

FMPOV resources, that are created/updated and therefore required for wake-up are:
- `nodes`
- `pods` in the `kube-system` namespace
- `endpoints` in the `kube-system` namespace (kcm/scheduler leaderelection)
- `csrs` with subresources (node bootstrap)
- `leases`
@rfranzke am I missing something?

@mvladev WDYT about trimming down the list to the mentioned resources?

rfranzke:
Didn't you say `configmaps` as well because some pod is doing leader election with it?

timebertt:
> Didn't you say `configmaps` as well because some pod is doing leader election with it?

I thought so at first, but double-checked and it doesn't seem to be the case. (CoreDNS isn't doing any leader election of course, because it's not a singleton...)

mvladev:
Those resources can be skipped I guess

- ~`apiservices`, `apiservices/status` breaks `metrics-server`~
- ~`endpointslices` (only for `kube-system` namespace) - breaks `kube-proxy` in large clusters~
- ~`horizontalpodautoscalers` (only for `kube-system` namespace), `horizontalpodautoscalers/status` (only for `kube-system` namespace) breaks dns~
- ~`podmetrics` (only for `kube-system` namespace) - breaks scaling for coredns~
- ~`nodemetrics` - breaks metrics server~
- ~`events` - doesn't explicitly break anything, but no debugging would be possible~
- ~`priorityclasses`~
- ~`runtimeclasses` - breaks every workload in kube-system~

But the rest can interfere with waking up / waking cluster from old gardener version:

- `certificatesigningrequests`, `certificatesigningrequests/approval`, `certificatesigningrequests/status` kubelet creates this once it attempts to join the cluster. If blocked - no `Nodes` can be created. Broken cluster.
- `clusterrolebindings`, `clusterroles`, `rolebindings`, `roles` (only for `kube-system` namespace). Cluster might be reconciled at version `x` of gardener and woken up at version `y` where the system controllers might require additional permissions - broken cluster. Also the `bootstrap-roles` reconciler in kube-apiserver will fail and the apiserver won't start at all.
- `configmaps` (only for `kube-system` namespace) - Cluster might be reconciled at version `x` of gardener and woken up at version `y` where the configuration of an component is changed. 
- `customresourcedefinitions`, `customresourcedefinitions/status` - breaks networking extensions, also Cluster might be reconciled at version `x` of gardener and woken up at version `y` where a new CRD is needed. 
- `controllerrevisions`, `daemonsets`, `daemonsets/status`,`deployments`,`deployments/scale`, `replicasets` , `replicasets/scale` , `replicasets/status`  (only for `kube-system` namespace) breaks all workloads in `kube-system`, bricking the cluster.
- `endpoints` -  apiserver automatic discovery (`kubernetes` endpoint in `default` namespace, breaks leader election of `kube-controller-manager`, `kube-scheduler` creating deadlock.
- `leases` - breaks kubelet healthchecks in newer kubernetes versions.
- `namespaces`, `namespaces/finalize`, `namespaces/status` breaks the cluster once the kube-apiserver is upgraded + woken combo from a version which uses leases as node healthchecks and needs to create the `kube-node-lease` namespace.
- `networkpolicies` (only for `kube-system` namespace) - Cluster might be reconciled at version `x` of gardener and woken up at version `y` where an additional connectivity between component is required, breaking the cluster. 
- `nodes` `nodes/status` breaks kubelet.
- `pods`, `pods/status` (only for `kube-system` namespace and `shoot.gardener.cloud/no-cleanup=true,orgin=gardener` labels) breaks workloads in kube-system
- `podsecuritypolicies` -  Cluster might be reconciled at version `x` of gardener and woken up at version `y` where an additional/updated PSP is needed and can cause the `pods` to not be created at all.
- `secrets` (only for `kube-system` namespace) - secrets might need updates (new certificate etc)
- `serviceaccounts` (only for `kube-system` namespace) - Cluster might be reconciled at version `x` of gardener and woken up at version `y` where an additional serviceaccount is needed and can cause the `pods` to not be created at all.
- `services`,`services/status` - breaks every workload in kube-system and kube-apiserver advertising in `default` namespace

timebertt:
> * `controllerrevisions`, `daemonsets`, `daemonsets/status`,`deployments`,`deployments/scale`, `replicasets` , `replicasets/scale` , `replicasets/status`  (only for `kube-system` namespace) breaks all workloads in `kube-system`, bricking the cluster.

Why? Is there a change to any of those resources during wake-up?

timebertt:
> But the rest can interfere with waking cluster from old gardener version

Hmm, have you seen an occurrence of such a case? 
As this will happen very rarely, I would not add these restrictions to the hibernation prevention check. If such an issue occurs, the user/operator can simply fix the cluster by pressing the reconcile button and everything will self-heal again. And as hibernated clusters, most likely don't run any productive workload, I would not see this as critical...

mvladev:
> > * `controllerrevisions`, `daemonsets`, `daemonsets/status`,`deployments`,`deployments/scale`, `replicasets` , `replicasets/scale` , `replicasets/status`  (only for `kube-system` namespace) breaks all workloads in `kube-system`, bricking the cluster.
> 
> Why? Is there a change to any of those resources during wake-up?

Mainly when the cluster is hibernated at version `x` of the gardener and woken at version `y` where changes are needed.

> > Hmm, have you seen an occurrence of such a case?
>
> As this will happen very rarely, I would not add these restrictions to the hibernation prevention check. If such an issue occurs, the user/operator can simply fix the cluster by pressing the reconcile button and everything will self-heal again. And as hibernated clusters, most likely don't run any productive workload, I would not see this as critical...

I see this as it can cause a possible breakage. It's better to avoid it in the first place.

timebertt:
@rfranzke can you give your opinion on the hibernation after gardener upgrade topic?

rfranzke:
Hm, @mvladev that's valid point, theoretically, but isn't it also kind of artificial? I can't think of a situation where this has happened in the past. Do you have an example?
Weighing the likelihood of such situation to happen with the level of improvement end-users would get if they could hibernate their shoots I would probably be pragmatic and allow them to hibernate their clusters.

mvladev:
> Hm, @mvladev that's valid point, theoretically, but isn't it also kind of artificial? I can't think of a situation where this has happened in the past. Do you have an example?

I've seen several clusters which have been hibernated for a quite a while - while several gardener releases have happened in-between. 

As you mentioned the hibernation feature is mainly used for development / test clusters and therefore it's better to avoid any kind of Gardener operator involvement.

Several Kubernetes features such as Mutating/Validating webhooks are _footgun_ and I want to limit the blast radius. The fact that something has _not happened_ in the past, doesn't mean that _it will not_ in the future.

rfranzke:
Sure, this is a quite common scenario, agreed. Though, I was more asking about examples for

> [...] woken at version y where changes are needed.

Which changes are you thinking of here? We "only" need calico, kube-proxy, coredns, vpn-shoot, and the webhook pod to come up and be reachable. I cant think of a situation where it was actually necessary to change the deployment/daemonset/... to make the wake-up being successful, but maybe I'm missing something.
I'm happy to extend this list in the future if we indeed see the necessity.

> As you mentioned the hibernation feature is mainly used for development / test clusters and therefore it's better to avoid any kind of Gardener operator involvement.

Right, and as it is a great feature to save costs/resources for those clusters wouldn't it be good to keep our restrictions to the bare minimum we can cater with?

mvladev:
Most of those restrictions are limited to the `kube-system` namespace as recommended by the [admission webhook documentations](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#avoiding-operating-on-the-kube-system-namespace).

End-users won't be affected by it as they should not have such webhooks in the first place.

rfranzke:
OK, that'd be fine to me if this is restricted to `kube-system`. However, I see also other, global resources in your GVK list like `events`, `clusterroles`, etc. Can they be cleaned up?

mvladev:
Many of those resources must be reconciled by the API server and if they cannot, it'll report bad health check and cannot start:

- `clusterrole` + `clusterrolebinding` mutating webhooks which are not running can break the API server [/healthz/poststarthook/rbac/bootstrap-roles](https://github.com/kubernetes/kubernetes/blob/25651408aeadf38c3df7ea8c760e7519fd37d625/pkg/registry/rbac/rest/storage_rbac.go#L62)

- `apiservice` + `apiservice/status` break `/healthz/poststarthook/apiservice-openapi-controller`, 
`/healthz/poststarthook/apiservice-registration-controller` and 
`/healthz/poststarthook/apiservice-status-available-controller`
- `proiorityclasses` break [/healthz/poststarthook/scheduling/bootstrap-system-priority-classes](https://github.com/kubernetes/kubernetes/blob/25651408aeadf38c3df7ea8c760e7519fd37d625/pkg/registry/scheduling/rest/storage_scheduling.go#L43)

rfranzke:
OK, I think these are valid points, thanks. And yes, as most resources are restricted to `kube-system` is should be okay. It would be great if you could comment the code with the reason why it is blocked, i.e., with information similar to https://github.com/gardener/gardener/pull/2270#issuecomment-625725446.
What's the reason for blocking services globally? Shouldn't `default` and `kube-system` namespaces be enough?

mvladev:
> What's the reason for blocking services globally? Shouldn't default and kube-system namespaces be enough?

The default NS is not labeled at the moment. Also, there is a repair ClusterIP code which is executing every single time when the API server is starting as part of [/readyz/poststarthook/bootstrap-controller](https://github.com/kubernetes/kubernetes/blob/25651408aeadf38c3df7ea8c760e7519fd37d625/pkg/master/controller.go#L155). It must be allowed to edit bad services.

mvladev:
I've updated the PR - have another look.

rfranzke (approved):
I'm okay with starting like this for now, thanks for updating the PR.
@tim-ebert WDYT?

timebertt:
Yep, fine for me, too.

PR #2254, opened 2020-04-29
Add K8S gomega error matchers.

mvladev:
**What this PR does / why we need it**:

New matchers for K8S errors are added:

- `Expect(err).To(BeNotFoundError())`
- `Expect(err).To(BeForbiddenError())`
- `Expect(err).To(BeBadRequestError())`

Makes testing easier as you are no longer required to do:

```go
Expect(err).To(HaveOccurred())
Expect(apierrors.IsNotFound(err)).To(BeTrue())
```

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement operator
NONE
```

PR #2271, opened 2020-05-02
Remove default delete options when deleting extensions

rfranzke:
With the refactoring of #2222 the default delete options were introduced when deleting extension resources. However, the delete options weren't provided before this PR (only for `OperatingSystemConfig` resources, but even here there was no specific rationale behind it).

The reason for removing it now again is that @ialidzhikov detected extension resources that were blocked with `foregroundDeletion` finalizer for seeds < 1.16. Basically, https://github.com/kubernetes/kubernetes/issues/77081 is hit which is only fixed in Kubernetes >= 1.16. Hence, let's restore the old behaviour for now.

Also, the PR fixes a linting problem in the `plugin/pkg/shoot/validator/admission_test.go` file introduced with #2254.

/cc @ialidzhikov, thanks for spotting

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/operatingsystemconfig_test.go
+++ b/pkg/operation/botanist/operatingsystemconfig_test.go
@@ -18,7 +18,6 @@ import (
 	"context"
 
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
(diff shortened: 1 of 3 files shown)


ialidzhikov (approved):
Thank you for digging and fixing it!

/lgtm

PR #2277, opened 2020-05-04
expand extension actuator interface with migrate and restore

vlvasilev:
With this PR we extend the Extensions actuator interface with Migrate and Restore functionality.
 
Part of [# 1631](https://github.com/gardener/gardener/issues/1631) 

**Release note**:

```improvement developer
Extend the Extensions actuator interface with Migrate and Restore
```

### DIFF
--- a/extensions/pkg/controller/extension/reconciler.go
+++ b/extensions/pkg/controller/extension/reconciler.go
@@ -25,6 +25,7 @@ import (
 	"github.com/gardener/gardener/extensions/pkg/util"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


extensions/pkg/controller/extension/reconciler.go
```diff
@@ -173,39 +174,43 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+	case ex.Annotations[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationRestore:
+		if result, err = r.restore(r.ctx, ex, operationType); err != nil {
+			return result, err
+		}
+		return reconcile.Result{Requeue: r.resync != 0, RequeueAfter: r.resync}, nil
+	default:
+		if result, err = r.reconcile(r.ctx, ex, operationType); err != nil {
+			return result, err
+		}
+		return reconcile.Result{Requeue: r.resync != 0, RequeueAfter: r.resync}, nil
```
rfranzke: Hm, why do we have this special handing with the error here? I don't see this for the infrastructure controller, for example: https://github.com/gardener/gardener/blob/master/extensions/pkg/controller/infrastructure/reconciler.go#L104-L115

Can you explain?

vlvasilev: Monkey see monkey do.
This error handling was there, before the switch.

plkokanov: Hmm I guess we should keep it then? Otherwise looks good.

rfranzke: cc @timuthy - can you give an opinion? I think you were contributing it initially (hope I remember correctly)

rfranzke:
/kind/enhancement

extensions/pkg/controller/extension/reconciler.go
timuthy: Are you specifically asking for the re-queuing? The motivation is that an extension can make use of an automatic re-queuing to allow clean-up jobs, for example. But I don't see a necessity for doing this in the restore case.

rfranzke: OK, thanks. Yes, agreed, for restore it's probably not needed. @vlvasilev can you adapt accordingly?

swilen-iwanow (approved):
lgtm

extensions/pkg/controller/extension/reconciler.go
vlvasilev: Done

rfranzke (approved):

rfranzke:
/lgtm
/reviewed/do-not-merge
until g/g@v1.4.0 is released, cc @ialidzhikov

PR #2278, opened 2020-05-04
More control for extension controller deployment

### LINKED ISSUE #1581: Disable extensions that are globally enabled
Today, it is possible to globally enable an extension via its `ControllerRegistration` resource. This means that all shoot clusters automatically get this extension without the necessity to explicitly activate it in the `.spec.extensions[]` list.

However, some shoots might want to opt out of our globally enabled extensions, but this is not possible today.
Let's introduce `.spec.extensions[].enabled={true,false}` (defaulted to `true`) which can be used to opt-out of these extensions.

cc @MartinWeindel

### LINKED ISSUE #2180: Allow better control for extension controller deployment
1. Let's lift the restriction that there is only exactly one registered extension for each `kind/type` combination.
1. Allow to not specify any `.spec.resources` (if not yet allowed)
1. Let's enhance the `ControllerRegistration` specification to allow better control for how/when extensions are deployed:

```yaml
---
kind: ControllerRegistration
metadata:
  name: os-coreos
spec:
  resources:
  - kind: OperatingSystemConfig
    type: coreos
    primary: true # new
  deployment:
    policy: OnDemand # new
    seedSelector: # new
      matchLabels:
        foo: bar
    type: helm
    providerConfig:
      chart: |
        H4sIFAAAAAAA/yk...
      values:
        foo: bar
```

* The `.spec.resources[].primary` field defaults to `true`. The validation should ensure that there is only exactly one registration with `primary=true` for each `kind/type` combination.
* The `.spec.deployment.policy` defaults to `OnDemand` and controls the policy for the deployment. `OnDemand` is what is currently implemented. Another allowed value is `Always` that tells that extension shall be deployed to each seed independent of whether it's needed for the particular reconciliation of a `BackupBucket`/`BackupEntry`/`Shoot`. 
* The `.spec.deployment.seedSelector` field controls which seeds are targeted. By default all seeds are targeted as the field is optional.

**Why is this needed**:
With #1993 the extension controllers registered via `ControllerRegistration` resources are no longer deployed by default to every seed cluster. Instead, the needed extensions for the reconciliation of a `BackupBucket`/`BackupEntry`/`Shoot` are computed and (un-)installed on-demand.
Also, currently the code is assuming that for each `kind/type` combination there is exactly one registration. While there should only be exactly one controller responsible for the reconciliation of an extension resource this breaks scenarios where secondary controllers (that just listen for example `Infrastructure` resource) can be installed as well.
We should allow better control so that more use-cases can be implemented.

Also interesting/related in this context: #1581

rfranzke:
This PR improves the control options for extension controller deployments. It introduces a deployment policy, seed selector, primary configuration, and allows shoots the possibility to opt-out of globally enabled extensions.

Fixes #2180
Fixes #1581

/cc @amshuman-kr @MartinWeindel @danielfoehrKn 

**Release note**:

```noteworthy operator
The `ControllerRegistration` object does now allow better control for deployment/deletion of extension controllers to seed clusters (policy, seed selector, ...). You might want to look into [this document](https://github.com/gardener/gardener/blob/master/docs/extensions/controllerregistration.md). 
```
```improvement user
It is now possible to opt-out of/disable globally enabled extensions for `Shoot`s by setting `.spec.extensions[] = {type: <extension-type>, disabled: true}`.
```


rfranzke:
/kind/enhancement

docs/extensions/controllerregistration.md
```diff
@@ -154,3 +166,38 @@ In the above example, Gardener itself does not understand the AWS-specific provi
...
+
+The `reconcileTimeout` tells Gardener how long it should wait during its shoot reconciliation flow for the `Extension/foo`'s reconciliation to finish.
+
+### Deployment configuration options
+
+The `.spec.deployment` resource allows to configure a deployment `policy`.
+There are the following policies:
+
+* `OnDemand` (default): Gardener will demand the deployment and deletion of the extension controller to/from seed clusters dynamically. It will automatically determine (based on other resources like `Shoot`s) whether it is required and decide accordingly.
+* `Always`: Gardener will demand the deployment of the extension controller to seed clusters independent of whether it is actually required or not.
```
timuthy: Can you give the reader an example when this might be useful?

pkg/apis/core/v1alpha1/types_controllerregistration.go
```diff
@@ -57,15 +58,20 @@ type ControllerRegistrationSpec struct {
 // kind (aws-route53, gcp, auditlog, ...).
 type ControllerResource struct {
 	// Kind is the resource kind, for example "OperatingSystemConfig".
-	Kind string `json:"kind" protobuf:"bytes,1,opt,name=kind"`
+	Kind string `json:"kind" protobuf:"bytes,1,rep,name=kind"`
```
timuthy: `rep` stands for [repeated](https://github.com/golang/protobuf/blob/e9dc0d72c1a35f349f6d6871842199ddaaaf9edc/proto/properties.go#L160) and is rather adequate for arrays if my research is correct. I'm not sure about the proper tag for a required field, though. K8s, for instance, defines an `opt` tag for required fields at some places (see [here](https://github.com/kubernetes/api/blob/master/networking/v1beta1/types.go#L305)) while at other places they don't have one at all (see [here](https://github.com/kubernetes/api/blob/master/networking/v1/types.go#L156)).

@ialidzhikov can you shed more light on this?

timuthy (commented):
I'll finish the review later today but already have some comments I want to submit.

docs/extensions/controllerregistration.md
```diff
@@ -16,10 +16,21 @@ spec:
...
   - kind: OperatingSystemConfig
     type: coreos
+    primary: true
 ```
 
 This information tells Gardener that there is an extension controller that can handle `OperatingSystemConfig` resources of type `coreos`.
-It will now create `ControllerInstallation` resources:
+
+Also, it specifies that this controller is the primary one responsible for the lifecycle of the `OperatingSystemConfig` resource.
+Setting `primary` to `false` would allow to register additional, secondary controllers that may also watch/react on the `OperatingSystemConfig/coreos` resources, however, only the primary controller may change/update the `status` of the extension object.
```
danielfoehrKn, line 25: > only the primary controller may change/update the `status`

Why is there such a restriction? That is also not possible to enforce?  Secondary controllers must be able to alter the status of the Extension resource (for instance update a condition / add an additional condition) .

docs/extensions/controllerregistration.md
```diff
@@ -16,10 +16,21 @@ spec:
...
+Secondary controllers might be helpful in scenarios where additional tasks need to be completed which are not part of the reconciliation logic of the primary controller but separated out into a dedicated extension.
```
danielfoehrKn, line 29: > only the primary controller may change/update the `status`

  Secondary controllers must be able to alter the status of the primary Extension resource (for instance update a condition / add an additional condition) .  There is also already a case where this is being done.

I think the definition for a 'primary' controller should rather be, that it is the only controller to alter the 'status.lastOperation' of an Extension CRD. Secondary controllers would still be able to write conditions.

Also in certain cases, in addition to the standard Gardener extension reconciliation, there can be another (you call it secondary) extension that is required for the Shoot to be successfully created / deleted  / reconciled. 
This is for instance the case when deploying Gardener in an environment that relies on private DNS zones that need to be associated with the Shoot VPC and Seed VPC. The secondary controllers would also reconcile the infrastructure CRD. If this VPC assignment is missing because the reconcile of the secondary controller errors, the worker nodes cannot join in a later step. Hence there are use-cases for (secondary) extension controllers that are essential to the Shoot flow. 
. The above mentioned use case is only one, and we do not know in which environment people want to operate Gardener in. 

In my opinion, we need to have a way of integration essential secondary controllers into the Gardener reconcile flow. Maybe the extension controller can write a certain condition.Type on the ExtensionCRD that if existing, needs to indicate success, in order for the Gardenlet to  continue the reconcile.
 I do not think that more than one controller should reconcile the  'status.LastOperation' though.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -269,46 +269,67 @@ func computeKindTypesForShoots(
...
-		names.Insert(name)
+		controllerRegistrationNames.Insert(names...)
+	}
+
+	// filter controller registrations with non-matching seed selector
+	names, err := controllerRegistrationNamesWithMatchingSeedLabelSelector(controllerRegistrationNames.UnsortedList(), controllerRegistrationNameToObject, seedLabels)
+	if err != nil {
+		return nil, err
 	}
+	controllerRegistrationNames = controllerRegistrationNames.Intersection(names)
```
danielfoehrKn: Why do you need an intersect? controllerRegistrationNames contains all required names. Based on that you determine the ControllerRegistrations that have a matching seed selector. Is that not what you need already?

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -140,7 +140,7 @@ func (c *defaultControllerRegistrationSeedControl) Reconcile(obj *gardencorev1be
 	}
 
 	var (
-		controllerRegistrationNameToObject, kindTypeToControllerRegistrationName = computeControllerRegistrationMaps(controllerRegistrationList)
+		controllerRegistrationNameToObject, kindTypeToControllerRegistrationNames, alwaysPolicyControllerRegistrationNames = computeControllerRegistrationMaps(controllerRegistrationList)
```
timuthy: Imo, reading the code is quite hard with the different maps and sets being built and passed here and there. What do you think about introducing a struct like:

```
type controllerRegistration struct {
  registration *gardencorev1beta1.ControllerRegistration
  alwaysDeploy bool
  kindType string
} 
```

The function `computeControllerRegistrationMaps` can then return `map[string]controllerRegistration`. The only thing we'd miss is `kindTypeToControllerRegistrationNames` but this map is anyway only required in `computeWantedControllerRegistrationNames`, i.e. let's build it there for sake of simplification.

docs/extensions/controllerregistration.md
rfranzke: done

pkg/apis/core/v1alpha1/types_controllerregistration.go
rfranzke: Yeah, thanks for pointing it out and checking, I reverted this change again. It's anyways unrelated to this PR.

docs/extensions/controllerregistration.md
rfranzke: > Secondary controllers must be able to alter the status of the primary Extension resource (for instance update a condition / add an additional condition) . 

Sure, I made this more clear now.

About the rest: Yes, you are probably right, but this is not the goal of this PR. It's indeed a first step into this direction, but out-of-scope for now. I suggest to open a new issue once this PR is merged that covers the feature request you outlined above.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -454,3 +475,47 @@ func convertObjToMap(in interface{}) (map[string]interface{}, error) {
 
 	return out, nil
 }
+
+func id(extensionKind, extensionType string) string {
```
timuthy: Should we also make `shoot.ComputeRequiredExtensions` use this function?

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
rfranzke: Nice catch, yes, indeed, you're right. It was a left-over from intermediate work. I've removed it now.

pkg/operation/shoot/shoot.go
```diff
@@ -453,12 +456,17 @@ func MergeExtensions(registrations []gardencorev1beta1.ControllerRegistration, e
 
 	// Extensions defined in Shoot resource.
 	for _, extension := range extensions {
-		obj, ok := typeToExtension[extension.Type]
-		if ok {
+		if obj, ok := typeToExtension[extension.Type]; ok {
+			if extension.Disabled != nil && *extension.Disabled {
```
timuthy:
```suggestion
			if utils.IsTrue(extension.Disabled) {
```

pkg/operation/shoot/shoot.go
```diff
@@ -525,8 +533,15 @@ func ComputeRequiredExtensions(shoot *gardencorev1beta1.Shoot, seed *gardencorev
 	requiredExtensions.Insert(fmt.Sprintf("%s/%s", extensionsv1alpha1.NetworkResource, shoot.Spec.Networking.Type))
 	requiredExtensions.Insert(fmt.Sprintf("%s/%s", extensionsv1alpha1.WorkerResource, shoot.Spec.Provider.Type))
 
+	var disabledExtensions = sets.NewString()
 	for _, extension := range shoot.Spec.Extensions {
-		requiredExtensions.Insert(fmt.Sprintf("%s/%s", extensionsv1alpha1.ExtensionResource, extension.Type))
+		id := fmt.Sprintf("%s/%s", extensionsv1alpha1.ExtensionResource, extension.Type)
+
+		if extension.Disabled != nil && *extension.Disabled {
```
timuthy:
```suggestion
		if utils.IsTrue(extension.Disabled) {
```

docs/extensions/controllerregistration.md
```diff
@@ -154,3 +169,38 @@ In the above example, Gardener itself does not understand the AWS-specific provi
...
+  - kind: Extension
+    type: foo
+    primary: true
+    globallyEnabled: true
+    reconcileTimeout: 30s
+```
+
+The `globallyEnabled=true` option specifies that the `Extension/foo` object shall be created by default for all shoots (unless they opted out by setting `.spec.extensions[].enabled=false` in the `Shoot` spec).
+
+The `reconcileTimeout` tells Gardener how long it should wait during its shoot reconciliation flow for the `Extension/foo`'s reconciliation to finish.
```
timuthy, line 194: I just double checked and realized that this setting does not have any affect at the moment. Since you're already improving the extension handling, can you add the timeout parameter to the extension deployment [here](https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/extension.go#L79) and to the deletion [here](https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/extension.go#L126). Otherwise I will open a separate PR if you prefer to not do it here.

timuthy (changes requested):
Nice PR! I've added some questions and suggestions but looks good in general.

docs/extensions/controllerregistration.md
danielfoehrKn: I have created an [issue here.](https://github.com/gardener/gardener/issues/2302)

docs/extensions/controllerregistration.md
rfranzke: I'll pass it in the wait-for-reconcile-ready function, but I don't know how to pass it in the wait-for-deletion case (second link you shared). We don't read the timeout yet, we simply wait all existing `Extension` resources with the same timeout. It's not straight-forward to fix this IMO, because when the respective `ControllerRegistration` is changed (supported for the `Extension` kind/type is removed, for example), then also the information about the `reconcileTimeout` is gone. However, as the field is already called `*reconcile*Timeout` it's maybe fair for now to only use it during the reconciliation? If you still want to fix this (maybe the last known configured timeout should be stored as annotation on the `Extension` object directly (to not lose the information), or so ..) then I suggest to follow up with it separately.

timuthy: Makes sense, thank you 👍

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
rfranzke: I've done it with https://github.com/gardener/gardener/pull/2278/commits/d1086235093267a84398d69d0cd350889c98459b, do you think it's better now?

timuthy: Yes, thank you ❤️

pkg/apis/core/validation/controllerregistration.go
```diff
@@ -84,5 +118,15 @@ func ValidateControllerRegistrationSpecUpdate(new, old *core.ControllerRegistrat
 		return allErrs
 	}
 
+	kindTypeToPrimary := make(map[string]*bool, len(old.Resources))
+	for _, resource := range old.Resources {
+		kindTypeToPrimary[resource.Kind+resource.Type] = resource.Primary
+	}
+	for i, resource := range new.Resources {
+		if primary, ok := kindTypeToPrimary[resource.Kind+resource.Type]; ok {
+			allErrs = append(allErrs, apivalidation.ValidateImmutableField(resource.Primary, primary, fldPath.Child("resources").Index(i).Child("primary"))...)
```
timuthy, line 127: I'm not sure if this validation makes sense because an operator can easily circumvent it by removing the resource from the list and adding it again afterwards. WDYT?

rfranzke: IMO it makes sense as it prevents obvious misconfiguration. If somebody uses the workaround you mentioned then it's probably/hopefully a conscious decision and the person is aware of the consequences.

timuthy: Okay, fair enough

timuthy (dismissed):

docs/extensions/controllerregistration.md
```diff
@@ -16,10 +16,24 @@ spec:
...
     type: coreos
+    primary: true
 ```
 
 This information tells Gardener that there is an extension controller that can handle `OperatingSystemConfig` resources of type `coreos`.
-It will now create `ControllerInstallation` resources:
+
+Also, it specifies that this controller is the primary one responsible for the lifecycle of the `OperatingSystemConfig` resource.
+Setting `primary` to `false` would allow to register additional, secondary controllers that may also watch/react on the `OperatingSystemConfig/coreos` resources, however, only the primary controller may change/update the main `status` of the extension object (that are used to "communicate" with the Gardenlet).
+Particularly, only the primary controller may be set `.status.lastOperation`, `.status.lastError`, `.status.observedGeneration`, and `.status.state`.
```
timebertt:
```suggestion
Particularly, only the primary controller may set `.status.lastOperation`, `.status.lastError`, `.status.observedGeneration`, and `.status.state`.
```

docs/extensions/controllerregistration.md
```diff
@@ -154,3 +169,38 @@ In the above example, Gardener itself does not understand the AWS-specific provi
...
+### Deployment configuration options
+
+The `.spec.deployment` resource allows to configure a deployment `policy`.
+There are the following policies:
+
+* `OnDemand` (default): Gardener will demand the deployment and deletion of the extension controller to/from seed clusters dynamically. It will automatically determine (based on other resources like `Shoot`s) whether it is required and decide accordingly.
+* `Always`: Gardener will demand the deployment of the extension controller to seed clusters independent of whether it is actually required or not. This might be helpful if you want to add a new component/controller to all seed clusters by default. Another use-case is to minimize the durations until extension controllers are deployed and ready in case you have highly fluctuating seed clusters.
+
+Also, the `.spec.deployment.seedSelector` allows to specify a label selector for seed clusters.
+Only if it matches to the labels of a seed then it will be deployed to it.
```
timebertt:
```suggestion
Only if it matches the labels of a seed then it will be deployed to it.
```

timebertt (commented):
Nits in the docs

timuthy (approved):

PR #2279, opened 2020-05-04
Do not reconcile extensions of failed shoot clusters

danielfoehrKn:
Extensions CRDs for Shoots with lastOperation.Failed should not be reconciled.
Introduces a check (in addition to the controller-runtime predicate) that checks the Shoot inlined in the Cluster CRD for last.operation == failed. If a Shoot is failed, stops the reconcile. 

Also see [this issue](https://github.com/gardener/gardener/issues/2145).
Fixes #

**Release note**:

```improvement operator
Fixes a bug in the extension library of all extension resources that lead to not stopping the reconciliation of extension resources when the Shoot is in 'failed' state (Shoot.Status.lastOperation.state = Failed). 
```

### DIFF
--- a/extensions/pkg/controller/backupbucket/reconciler.go
+++ b/extensions/pkg/controller/backupbucket/reconciler.go
@@ -87,9 +87,20 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, err
 	}
 
+	shoot, err := extensionscontroller.GetShoot(r.ctx, r.client, request.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if extensionscontroller.IsShootFailed(shoot) {
+		r.logger.Info("Stop reconciling BackupBucket of failed Shoot.", "namespace", request.Namespace, "name", bb.Name)
+		return reconcile.Result{}, nil
+	}
+
 	if bb.DeletionTimestamp != nil {
 		return r.delete(r.ctx, bb)
 	}
+
 	return r.reconcile(r.ctx, bb)
 }
 
--- a/extensions/pkg/controller/cluster.go
+++ b/extensions/pkg/controller/cluster.go
@@ -132,6 +132,21 @@ func ShootFromCluster(decoder runtime.Decoder, cluster *extensionsv1alpha1.Clust
 	return shoot, nil
 }
 
+// GetShoot tries to read Gardener's Cluster extension resource in the given namespace and return the embedded Shoot resource.
+func GetShoot(ctx context.Context, c client.Client, namespace string) (*gardencorev1beta1.Shoot, error) {
+	cluster := &extensionsv1alpha1.Cluster{}
+	if err := c.Get(ctx, kutil.Key(namespace), cluster); err != nil {
+		return nil, err
+	}
+
+	decoder, err := NewGardenDecoder()
+	if err != nil {
+		return nil, err
+	}
+
+	return ShootFromCluster(decoder, cluster)
+}
+
 // NewGardenDecoder returns a new Garden API decoder.
 func NewGardenDecoder() (runtime.Decoder, error) {
 	return serializer.NewCodecFactory(gardenscheme).UniversalDecoder(), nil
--- a/extensions/pkg/controller/containerruntime/reconciler.go
+++ b/extensions/pkg/controller/containerruntime/reconciler.go
@@ -107,6 +107,10 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+	if extensionscontroller.IsFailed(cluster) {
+		r.logger.Info("Stop reconciling ContainerRuntime of failed Shoot.", "namespace", request.Namespace, "name", cr.Name)
+		return reconcile.Result{}, nil
+	}
 
 	operationType := gardencorev1beta1helper.ComputeOperationType(cr.ObjectMeta, cr.Status.LastOperation)
 
--- a/extensions/pkg/controller/extension/reconciler.go
+++ b/extensions/pkg/controller/extension/reconciler.go
@@ -23,11 +23,11 @@ import (
 	extensionshandler "github.com/gardener/gardener/extensions/pkg/handler"
 	extensionspredicate "github.com/gardener/gardener/extensions/pkg/predicate"
 	"github.com/gardener/gardener/extensions/pkg/util"
-
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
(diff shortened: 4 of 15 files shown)


timuthy (commented):
Looks good in general 👍 Can you please double check the unit tests and also I'd suggest to add a log entry if a resource is ignored because of a failed shoot.

extensions/pkg/controller/backupbucket/reconciler.go
```diff
@@ -90,6 +90,16 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 	}
+
+	shoot, err := extensionscontroller.GetShoot(r.ctx, r.client, request.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if extensionscontroller.IsShootFailed(shoot) {
+		return reconcile.Result{}, nil
+	}
```
rfranzke: Shouldn't this be moved up to line 89?

extensions/pkg/controller/cluster.go
```diff
@@ -132,6 +132,26 @@ func ShootFromCluster(decoder runtime.Decoder, cluster *extensionsv1alpha1.Clust
...
+	if err != nil {
+		return nil, err
+	}
+
+	shoot, err := ShootFromCluster(decoder, cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	return shoot, nil
```
rfranzke:
```suggestion
	return ShootFromCluster(decoder, cluster)
```

extensions/pkg/controller/containerruntime/reconciler.go
```diff
@@ -107,6 +107,10 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return r.delete(r.ctx, cr, cluster)
 	}
 
+	if extensionscontroller.IsFailed(cluster) {
+		return reconcile.Result{}, nil
+	}
+
```
rfranzke: Shouldn't this be moved up to line 105?

extensions/pkg/controller/extension/reconciler.go
```diff
@@ -168,15 +168,21 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 	}
 
+	shoot, err := extensionscontroller.GetShoot(r.ctx, r.client, request.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if extensionscontroller.IsShootFailed(shoot) {
+		return reconcile.Result{}, nil
+	}
```
rfranzke: Shouldn't this be moved up to line 172?

extensions/pkg/controller/shoot.go
```diff
@@ -59,6 +60,16 @@ func IsHibernated(cluster *Cluster) bool {
...
 
+// IsFailed returns true if the shoot is failed, or false otherwise.
+func IsFailed(cluster *Cluster) bool {
+	return IsShootFailed(cluster.Shoot)
+}
+
+func IsShootFailed(shoot *gardencorev1beta1.Shoot) bool {
+	lastOperation := shoot.Status.LastOperation
+	return lastOperation != nil && lastOperation.State == gardencorev1beta1.LastOperationStateFailed
+}
```
rfranzke, line 75: Can you add small unit tests for these functions?

rfranzke (changes requested):

extensions/pkg/controller/backupbucket/reconciler.go
danielfoehrKn: I though:
1) Checking for a deletion timestamp on the CRD is cheaper than retrieving & parsing the Cluster CRD ->  in case the Extension CRD should be deleted, not necessary to do the following operation 
2) Ability to delete CRDs of failed clusters manually without changing the Cluster CRD first.

What would be the reason for moving it before the deletion check?

extensions/pkg/controller/containerruntime/reconciler.go
danielfoehrKn: [same question as here](https://github.com/gardener/gardener/pull/2279#discussion_r419915471)

extensions/pkg/controller/extension/reconciler.go
danielfoehrKn: [same question as here](https://github.com/gardener/gardener/pull/2279#discussion_r419915471)

extensions/pkg/controller/backupbucket/reconciler.go
rfranzke: I don't understand. The motivation behind the `Failed` status is that nothing is happening (not even deletion) unless the user manually confirms that it can be retried. 
For example, consider a shoot that is stuck in deletion for 24h because of invalid credentials or whatever other reason. Gardenlet would transition the status to `Failed` and would not try to delete it again until the user acts. The same behaviour should be implemented by the extensions.

extensions/pkg/controller/backupbucket/reconciler.go
```diff
@@ -90,6 +90,17 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	if bb.DeletionTimestamp != nil {
 		return r.delete(r.ctx, bb)
 	}
+
+	shoot, err := extensionscontroller.GetShoot(r.ctx, r.client, request.Namespace)
```
rfranzke, line 90: Should also be moved up above line 90.

extensions/pkg/controller/containerruntime/reconciler.go
```diff
@@ -107,6 +107,11 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return r.delete(r.ctx, cr, cluster)
 	}
 
+	if extensionscontroller.IsFailed(cluster) {
```
rfranzke, line 110: To be moved up above line 106

extensions/pkg/controller/extension/reconciler.go
```diff
@@ -168,15 +168,22 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-		result reconcile.Result
-		err    error
-	)
+	var result reconcile.Result
 
 	if ex.DeletionTimestamp != nil {
 		return r.delete(r.ctx, ex)
 	}
 
+	shoot, err := extensionscontroller.GetShoot(r.ctx, r.client, request.Namespace)
```
rfranzke, line 174: To be moved up above line 173

rfranzke (changes requested):

rfranzke:
/kind/bug

rfranzke:
/status/author-action

rfranzke (dismissed):

timuthy (dismissed):
Nice!

timuthy:
/reviewed/do-not-merge

timuthy:
As discussed, let's merge after the v1.4.0 release.

rfranzke:
Could you please rebase to get this PR merged?
/needs/rebase

rfranzke (approved):

PR #2281, opened 2020-05-05
Adapt workaround for broken ShootNotFailed predicate

rfranzke:
The `ShootNotFailed` predicate was broken and got fixed with #2265. However, to give extension controllers enough time to revendor we have added a temporary workaround in #2217. It was faking the last operation for newly created shoots. However, @ialidzhikov reported in https://github.com/gardener/gardener/pull/2217#issuecomment-623637931 that it does not work properly.
This PR adapts the workaround to allow extensio controllers to properly react for newly created shoots (where the `status` of the `Shoot` is not yet fully set up).

/cc @ialidzhikov 

**Release note**:

```improvement operator
NONE
```


ialidzhikov (approved):
Thank you!

/lgtm

PR #2283, opened 2020-05-05
Add global cleanuphandler and possible after test options to the testsuite testcases

### LINKED ISSUE #2272: Shoot care test clean up logic is not executed on interrupt signal
**What happened**:
The Shoot care test scales the kube-apiserver deploy to 0 replicas and waits until this is reflected in the  appropriate Shoot condition. In a defer, the test executes clean up logic and scales back the kube-apiserver replicas to the original count. Ref https://github.com/gardener/gardener/blob/v1.3.1/test/integration/shoots/care/shoot_care.go.

We see cases in which the whole test suite times out and this may sometimes interrupt the Shoot care test during its execution. As the clean up logic is executed in `defer` (not in `AfterSuite`), the test is interrupted and the defer is not executed leaving a Shoot with 0 kube-apiserver replicas (similar Shoot subsequently will fail to delete because of this reason).

gingko ensures that it will execute `AfterSuite` function when interrupt signal is received. Our clean up logic should be in `AfterSuite` block, not in `defer`.

1. Run the test.

```
go test -timeout=0 -mod=vendor ./test/suites/shoot \
      --v -ginkgo.v -ginkgo.progress -ginkgo.noColor \
      --report-file=/tmp/report.json \
      -kubecfg=$KUBECONFIG \
      -shoot-name=$SHOOT_NAME \
      -project-namespace=$PROJECT_NAMESPACE \
      -ginkgo.focus="Should observe failed health condition"
```

2. Interrupt.

```
=== RUN   TestGardenerSuite
Running Suite: Shoot Test Suite
===============================
Random Seed: 1588519529
Will run 1 of 19 specs

S
------------------------------
Shoot Care testing
  [BETA] [SERIAL] [SHOOT] Should observe failed health condition in the Shoot when scaling down the API Server of the Shoot
  /home/go/src/github.com/gardener/gardener/test/framework/gingko_utils.go:26
[BeforeEach] Shoot Care testing
  /home/go/src/github.com/gardener/gardener/test/framework/commonframework.go:59
[BeforeEach] Shoot Care testing
  /home/go/src/github.com/gardener/gardener/test/framework/shootframework.go:79
[It] [BETA] [SERIAL] [SHOOT] Should observe failed health condition in the Shoot when scaling down the API Server of the Shoot
  /home/go/src/github.com/gardener/gardener/test/framework/gingko_utils.go:26
time="2020-05-03T18:25:35+03:00" level=info msg="Waiting for shoot aws-local to have expected condition (APIServerAvailable: False). Currently: (APIServerAvailable: True)"
^C
---------------------------------------------------------
Received interrupt.  Running AfterSuite...
^C again to terminate immediately

Ran 1 of 2 Specs in 24.908 seconds
FAIL! -- 1 Passed | 0 Failed | 0 Pending | 1 Skipped
FAIL	github.com/gardener/gardener/test/suites/shoot	25.580s
```

3. Ensure that the clean up logic is not executed.

**Environment**:

- Gardener version:
- Others:

schrodit:

Currently most testcases implement their cleanup logic with `defer` functions which does not cleanup in every case and may leak resources.

Therefore, two new features are added to the testframework to be able to 
- set cleanup functions per test
- set global test cleanup for the `AfterSuite`

A *global cleanup* handler can be now easily added by registering the cleanup function with `framework.AddCleanupAction(fn func())`.
I can also be removed if not needed with`framework.RemoveCleanupAction(fn func())`
Explination: this global cleanup actions store is needed as there can only be one `ginkgo.AfterSuite` which means that we have to invoke all cleanup handlers there.

*cleanup functions* for a specific test can be defined by adding a cleanup option to the testdefinition `It()`. An example can be found below.

The `AfterTest` cleanup function is defined by creating a new context just for the test and register a `ginkgo.AfterEach`. 
As ginkgo's `AfterEach` is not invoked when the test is interrupt(see this issue onsi/ginkgo#222 ), the cleanup function is also registered as a global aftersuite cleanup.
This registered aftersuite cleanup is removed from the global again when it is executed as aftersuite cleanup so that the cleanup function does not run multiple times.

```go
f := framework.NewShootFramework(nil)
f.Serial().CIt("My test", func(ctx context.Context) {
  // my test goes here
  Expect(val).ToBe(somevalue)
}, 1 * time.Minute,
framework.WithCAfterTests(func(ctx context.Context) {
  // my cleanup code goes here
}, 30 * time.Seconds))
```

Fixes https://github.com/gardener/gardener/issues/2272

**Release note**:

```improvement developer
It is now possible to add a global cleanup function for integration tests.
```
```improvement developer
It is now possible to add a dedicated `AfterTest` function to test cases to run a specific function when the test has finished.
```

### DIFF
--- a/test/framework/cleanup.go
+++ b/test/framework/cleanup.go
@@ -1,0 +1,60 @@
+// Copied from https://github.com/kubernetes/kubernetes/blob/de66c643c479f420b1902f8897636ea41ac2b3c3/test/e2e/framework/cleanup.go
+/*
+Copyright 2016 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+import "sync"
+
+// CleanupActionHandle is an integer pointer type for handling cleanup action
+type CleanupActionHandle *int
+
+var cleanupActionsLock sync.Mutex
+var cleanupActions = map[CleanupActionHandle]func(){}
+
+// AddCleanupAction installs a function that will be called in the event of the
+// whole test being terminated.  This allows arbitrary pieces of the overall
+// test to hook into SynchronizedAfterSuite().
+func AddCleanupAction(fn func()) CleanupActionHandle {
+	p := CleanupActionHandle(new(int))
+	cleanupActionsLock.Lock()
+	defer cleanupActionsLock.Unlock()
+	cleanupActions[p] = fn
+	return p
+}
+
+// RemoveCleanupAction removes a function that was installed by
+// AddCleanupAction.
+func RemoveCleanupAction(p CleanupActionHandle) {
+	cleanupActionsLock.Lock()
+	defer cleanupActionsLock.Unlock()
+	delete(cleanupActions, p)
+}
+
+// RunCleanupActions runs all functions installed by AddCleanupAction.  It does
+// not remove them (see RemoveCleanupAction) but it does run unlocked, so they
+// may remove themselves.
+func RunCleanupActions() {
+	list := []func(){}
+	func() {
+		cleanupActionsLock.Lock()
+		defer cleanupActionsLock.Unlock()
+		for _, fn := range cleanupActions {
+			list = append(list, fn)
+		}
+	}()
+	// Run unlocked.
+	for _, fn := range list {
+		fn()
+	}
+}
--- a/test/framework/test_options.go
+++ b/test/framework/test_options.go
@@ -1,0 +1,125 @@
+// Copyright 2020 Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package framework
+
+import (
+	"context"
+	"time"
+
+	"github.com/onsi/ginkgo"
+)
+
+// TestOptions contains options to add additional functionality
+// or cleanup handlers to a testcase.
+type TestOptions struct {
+	// afterTests holds a list of all registered AfterTest functions
+	// that are executed when the test has finished.
+	AfterTests afterTests
+
+	// CAfterTests holds a list of all registered contextified AfterTest functions
+	// that are executed when the test has finished.
+	CAfterTests []cAfterTestOption
+}
+
+// ApplyOptions applies the given test options on these options.
+func (o *TestOptions) ApplyOptions(opts []TestOption) *TestOptions {
+	for _, opt := range opts {
+		opt.ApplyToTestOptions(o)
+	}
+	return o
+}
+
+// Complete registers all test options that are configured.
+// it should be a function that configures a ginkgo test case
+// This should get called when all options are applied.
+func (o *TestOptions) Complete(it func()) {
+
+	if len(o.AfterTests) == 0 && len(o.CAfterTests) == 0 {
+		it()
+		return
+	}
+
+	// Create a new context so that the afterTests function only runs once after this one testcase.
+	// Otherwise the after tests would run after every test case in the outer context.
+	ginkgo.Context("", func() {
+		it()
+
+		for _, aftertest := range o.AfterTests {
+			// register afterTest to global aftersuite in case the test interrupts
+			var h CleanupActionHandle
+			ginkgo.BeforeEach(func() {
+				h = AddCleanupAction(aftertest)
+			})
+			ginkgo.AfterEach(func() {
+				RemoveCleanupAction(h)
+				aftertest()
+			})
+		}
+
+		for _, caftertest := range o.CAfterTests {
+			// register afterTest to global aftersuite in case the test interrupts
+			var h CleanupActionHandle
+			ginkgo.BeforeEach(func() {
+				h = AddCleanupAction(func() {
+					contextify(caftertest.Body, caftertest.Timeout)()
+				})
+			})
+			CAfterEach(func(ctx context.Context) {
+				RemoveCleanupAction(h)
+				caftertest.Body(ctx)
+			}, caftertest.Timeout)
+		}
+	})
+}
+
+// cAfterTestOption contains options for contextified after test function.
+type cAfterTestOption struct {
+	Body    func(ctx context.Context)
+	Timeout time.Duration
+}
+
+// ApplyToTestOptions adds contextified after test functions to test options
+func (at *cAfterTestOption) ApplyToTestOptions(opts *TestOptions) {
+	opts.CAfterTests = append(opts.CAfterTests, *at)
+}
+
+// TestOption is some configuration that modifies options for testcase.
+type TestOption interface {
+	// ApplyToTestOptions applies this configuration to the given test options.
+	ApplyToTestOptions(*TestOptions)
+}
+
+// afterTests are functions that should run when a test has finished
+type afterTests []func()
+
+// ApplyToTestOptions adds after test functions to test options
+func (at afterTests) ApplyToTestOptions(opts *TestOptions) {
+	opts.AfterTests = append(opts.AfterTests, at...)
+}
+
+// WithAfterTests adds functions to the current test that are called
+// when the test has finished
+func WithAfterTests(funcs ...func()) TestOption {
+	return afterTests(funcs)
+}
+
+// WithAfterTests adds contextified functions to the current test that are called
+// when the test has finished
+func WithCAfterTest(body func(ctx context.Context), timeout time.Duration) TestOption {
+	return &cAfterTestOption{
+		Body:    body,
+		Timeout: timeout,
+	}
+}
--- a/test/integration/shoots/care/shoot_care.go
+++ b/test/integration/shoots/care/shoot_care.go
@@ -43,30 +43,13 @@ const (
 )
 
 var _ = ginkgo.Describe("Shoot Care testing", func() {
-
-	f := framework.NewShootFramework(nil)
+	var (
+		f            = framework.NewShootFramework(nil)
+		origReplicas *int32
+		err          error
+	)
 
 	f.Beta().Serial().CIt("Should observe failed health condition in the Shoot when scaling down the API Server of the Shoot", func(ctx context.Context) {
-		var (
-			origReplicas *int32
-			err          error
-		)
-
-		defer func(ctx context.Context) {
-			if origReplicas != nil {
-				f.Logger.Infof("Test cleanup: Scale API Server to '%d' replicas again", *origReplicas)
-				origReplicas, err = framework.ScaleDeployment(timeout, f.SeedClient.Client(), origReplicas, gardencorev1beta1constants.DeploymentNameKubeAPIServer, f.ShootSeedNamespace())
-				gomega.Expect(err).ToNot(gomega.HaveOccurred())
-
-				// wait for healthy condition
-				f.Logger.Infof("Test cleanup: wait for Shoot Health condition '%s' to become healthy again", gardencorev1beta1.ShootAPIServerAvailable)
-				err = f.WaitForShootCondition(ctx, 20*time.Second, 5*time.Minute, gardencorev1beta1.ShootAPIServerAvailable, gardencorev1beta1.ConditionTrue)
-				gomega.Expect(err).ToNot(gomega.HaveOccurred())
-
-				f.Logger.Info("Test cleanup successful")
-			}
-		}(ctx)
-
 		cond := helper.GetCondition(f.Shoot.Status.Conditions, gardencorev1beta1.ShootAPIServerAvailable)
 		gomega.Expect(cond).ToNot(gomega.BeNil())
 		gomega.Expect(cond.Status).To(gomega.Equal(gardencorev1beta1.ConditionTrue))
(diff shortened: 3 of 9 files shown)


dguendisch (dismissed):

ialidzhikov:
ginkgo will execute the AfterSuite on `SIGINT`. Will it also execute it on `SIGTERM`? What is the termination signal sent from test-machinery to the suite after test suite timeout occurs?

schrodit:
`SIGTERM` is handled as `SIGINT` see https://github.com/onsi/ginkgo/blob/05454150228890f52539c074418de9c6a923e28a/internal/specrunner/spec_runner.go#L220

The testmachinery uses the default `activeDeadlineSeconds` of kubernetes which to my understanding calls a `SIGTERM` to PID 1 (which is the test in our case) and waits for the grace period.
So also execution in the testmachinery/kubernetes should be correclty handled

test/framework/cleanup.go
```diff
@@ -0,0 +1,60 @@
+// Copied from https://github.com/kubernetes/kubernetes/blob/de66c643c479f420b1902f8897636ea41ac2b3c3/test/e2e/framework/cleanup.go
```
danielfoehrKn, line 1: Is it necessary to copy that file - cant we use a package import?

danielfoehrKn (commented):
Generally looks good to me

test/framework/cleanup.go
rfranzke: Let's please not start vendoring k/k - it's a pain. cc @mvladev

rfranzke (dismissed):

rfranzke:
/kind/test

test/framework/test_options.go
```diff
@@ -0,0 +1,118 @@
...
+			CAfterEach(func(ctx context.Context) {
+				RemoveCleanupAction(h)
+				caftertest.Body(ctx)
+			}, caftertest.Timeout)
+		}
+	})
+}
+
+// CAfterTestOptions contains options for contextified after test function.
+type CAfterTestOptions struct {
```
timebertt: Can we keep these private?
```suggestion
type cAfterTestOptions struct {
```

test/framework/test_options.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+// Test options contains options to add additional functionality
+// or cleanup handlers to a testcase.
+type TestOptions struct {
+	// AfterTests holds a list of all registered AfterTest functions
+	// that are executed when the test has finished.
+	AfterTests []func()
+
+	// AfterTests holds a list of all registered contextified AfterTest functions
```
timebertt:
```suggestion
	// CAfterTests holds a list of all registered contextified AfterTest functions
```

test/framework/test_options.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// TestOption is some configuration that modifies options for testcase.
+type TestOption interface {
+	// ApplyToTest applies this configuration to the given test options.
+	ApplyToTest(*TestOptions)
+}
+
+// AfterTests are functions that should run when a test has finished
+type AfterTests []func()
```
timebertt: Can we keep these private?
```suggestion
type afterTests []func()
```

test/framework/test_options.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/onsi/ginkgo"
+	"time"
+)
+
+// Test options contains options to add additional functionality
+// or cleanup handlers to a testcase.
+type TestOptions struct {
+	// AfterTests holds a list of all registered AfterTest functions
+	// that are executed when the test has finished.
+	AfterTests []func()
```
timebertt:
```suggestion
	AfterTests afterTests
```

test/framework/test_options.go
```diff
@@ -0,0 +1,118 @@
...
+func (o *TestOptions) Complete(it func()) {
+
+	if len(o.AfterTests) == 0 && len(o.CAfterTests) == 0 {
+		it()
+		return
+	}
+
+	// Create a new context so that the AfterTests function only runs once after this one testcase.
+	// Otherwise the after tests would run after every test case in the outer context.
+	ginkgo.Context("", func() {
```
timebertt, line 56: Hmmm, this will add a "" line to ginkgo failure message, right? If we can't avoid it (which I assume), can we at least give it a name (like "AfterTest wrapper context" or whatever) to not add more confusion to failure messages?

test/integration/shoots/care/shoot_care.go
```diff
@@ -43,16 +43,26 @@ const (
...
+		zero := int32(0)
+		origReplicas, err = framework.ScaleDeployment(timeout, f.SeedClient.Client(), &zero, gardencorev1beta1constants.DeploymentNameKubeAPIServer, f.ShootSeedNamespace())
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
-		defer func(ctx context.Context) {
+		// wait for unhealthy condition
+		err = f.WaitForShootCondition(ctx, 20*time.Second, 5*time.Minute, gardencorev1beta1.ShootAPIServerAvailable, gardencorev1beta1.ConditionFalse)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	}, timeout,
+		framework.WithCAfterTests(func(ctx context.Context) {
```
timebertt:
```suggestion
	}, timeout, framework.WithCAfterTests(func(ctx context.Context) {
```

test/framework/test_options.go
```diff
@@ -0,0 +1,118 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package framework
+
+import (
+	"context"
+	"github.com/onsi/ginkgo"
```
timebertt: Can you sort the imports, please?

test/framework/test_options.go
```diff
@@ -0,0 +1,118 @@
...
+
+// ApplyToTest adds contextified after test functions to test options
+func (at *CAfterTestOptions) ApplyToTest(opts *TestOptions) {
+	opts.CAfterTests = append(opts.CAfterTests, *at)
+}
+
+// TestOption is some configuration that modifies options for testcase.
+type TestOption interface {
+	// ApplyToTest applies this configuration to the given test options.
+	ApplyToTest(*TestOptions)
```
timebertt:
```suggestion
	ApplyToTestOptions(*TestOptions)
```

timebertt (changes requested):
Nice to have these cleanup actions! Thanks @schrodit :)

One thing, I don't quite get (maybe you had your reasons...) is, why these cleanup functions are so tightly coupled with individual `*It`s.
In general, I think it would be nice to be able to configure cleanup functions on the `Context` level. For example: consider the [clusterautoscaler test](https://github.com/gardener/gardener/blob/master/test/integration/shoots/operations/clusterautoscaler.go). It has two `Its` and both are doing 2 cleanup actions, which are pretty similar (scale down the deployment and revert changes to the shoot spec). It would be nice, to have only 1 cleanup action on the context level which can cover both test cases. WDYT?

schrodit:
> In general, I think it would be nice to be able to configure cleanup functions on the Context level.
This is already now possible. You can just use `AfterEach` inside the context and ginkgo will already run the afteach function after every test in the context.
That's why I create a new context wrapper to only run the cleanup after one specific test.

test/framework/test_options.go
schrodit: The context text does not really appear in the output if the test were successful.
But you're right. It would appear as empty line in the error output.
I have no strong opinion on that. If the empty line is to confusing I can add text to the context.

See this example:
```go
func TestFramework(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "Framework Test Suite")
}

var _ = Describe("test", func() {

	It("Without context", func(){
		Expect(true).To(BeTrue())
	})

	Context("my context", func() {
		It("With named context", func(){
			Expect(true).To(BeTrue())
		})
	})

	Context("", func() {
		It("With empty context", func(){
			Expect(true).To(BeTrue())
		})
	})

})
```

Would produce:
```
=== RUN   TestFramework
Running Suite: Framework Test Suite
===================================
Random Seed: 1588749141
Will run 3 of 3 specs

test 
  Without context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:16
[It] Without context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:16
•
------------------------------
test my context 
  With named context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:21
[It] With named context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:21
•
------------------------------
test  
  With empty context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:27
[It] With empty context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:27
•
Ran 3 of 3 Specs in 0.000 seconds
SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestFramework (0.00s)
```

With failures:
```
=== RUN   TestFramework
Running Suite: Framework Test Suite
===================================
Random Seed: 1588749317
Will run 3 of 3 specs

test 
  Without context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:16
[It] Without context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:16

• Failure [0.000 seconds]
test
/xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:14
  Without context [It]
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:16

  Expected
      <bool>: true
  to be false

  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:17
------------------------------
test my context 
  With named context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:21
[It] With named context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:21

• Failure [0.000 seconds]
test
/xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:14
  my context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:20
    With named context [It]
    /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:21

    Expected
        <bool>: true
    to be false

    /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:22
------------------------------
test  
  With empty context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:27
[It] With empty context
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:27

• Failure [0.000 seconds]
test
/xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:14
  
  /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:26
    With empty context [It]
    /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:27

    Expected
        <bool>: true
    to be false

    /xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:28
------------------------------


Summarizing 3 Failures:

[Fail] test [It] Without context 
/xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:17

[Fail] test my context [It] With named context 
/xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:22

[Fail] test  [It] With empty context 
/xxx/go/src/github.com/gardener/test-infra/tmp/main_test.go:28

Ran 3 of 3 Specs in 0.001 seconds
FAIL! -- 0 Passed | 3 Failed | 0 Pending | 0 Skipped
--- FAIL: TestFramework (0.00s)
FAIL

Process finished with exit code 1
```

timebertt: Thanks for testing this!
As I am seeing the output from above now, I would probably leave the context string empty, as you have done it. It might be even more confusing to add a text there...

test/framework/test_options.go
```diff
@@ -0,0 +1,119 @@
...
+			CAfterEach(func(ctx context.Context) {
+				RemoveCleanupAction(h)
+				caftertest.Body(ctx)
+			}, caftertest.Timeout)
+		}
+	})
+}
+
+// cAfterTestOptions contains options for contextified after test function.
+type cAfterTestOptions struct {
```
timebertt: This actually contains only _one_ option.
```suggestion
type cAfterTestOption struct {
```

test/framework/test_options.go
```diff
@@ -0,0 +1,119 @@
...
+
+// WithAfterTests adds functions to the current test that are called
+// when the test has finished
+func WithAfterTests(funcs ...func()) TestOption {
+	return afterTests(funcs)
+}
+
+// WithAfterTests adds contextified functions to the current test that are called
+// when the test has finished
+func WithCAfterTests(body func(ctx context.Context), timeout time.Duration) TestOption {
```
timebertt:
```suggestion
func WithCAfterTest(body func(ctx context.Context), timeout time.Duration) TestOption {
```

timebertt (commented):
Thanks for addressing my comments. Only one nit left.

> This is already now possible. You can just use AfterEach inside the context and ginkgo will already run the afteach function after every test in the context.

Ok, I wasn't aware.
But if I understood correctly, these will not be run, if the test is interrupted/terminated, right?
So WDYT about reusing the logic you added to add a seconder wrapper which allows users to run cleanup actions after each test but as a fallback in `AfterSuite` on termination?

schrodit:
> So WDYT about reusing the logic you added to add a seconder wrapper which allows users to run cleanup actions after each test but as a fallback in AfterSuite on termination?

Good idea. I can add a helper function to address the ginkgo issue also in general for after each.
But if the aftereach function should be run for all test in the context we would need to call a `BeforeEach` function that registers the cleanup function as global action and remove it with the aftereach call.
Because otherwise the cleanup action would be removed after the first `AfterTest`.

I will open a seperate PR for that as it envolves some more things to consider.

timebertt (dismissed):
/approve

ialidzhikov:
Does it make sense to update also the other tests using defer for cleanup 
- https://github.com/gardener/gardener/blob/1c5d78e086d9d66bc5a23b20a095f6860136d8d0/test/integration/shoots/applications/networking.go#L82-L90

- https://github.com/gardener/gardener/blob/1c5d78e086d9d66bc5a23b20a095f6860136d8d0/test/integration/shoots/logging/seed_log.go#L80-L90

- https://github.com/gardener/gardener/blob/1c5d78e086d9d66bc5a23b20a095f6860136d8d0/test/integration/shoots/maintenance/maintenance_test.go#L280-L285

- https://github.com/gardener/gardener/blob/1c5d78e086d9d66bc5a23b20a095f6860136d8d0/test/integration/shoots/maintenance/maintenance_test.go#L331-L336

- https://github.com/gardener/gardener/blob/1c5d78e086d9d66bc5a23b20a095f6860136d8d0/test/integration/shoots/operations/clusterautoscaler.go#L95-L104

- and others

schrodit:
@ialidzhikov I adjusted the other tests as well. But I will leave the maintenance test as is as it still uses the old framework. When we refactor this one I will also update this test.

I also found a bug in the execution that has been addressed with another `BeforeEach` function to register the global cleanup action.

PTAL

timebertt (approved):
Thanks, @schrodit!
/lgtm

rfranzke:
Any objections merging this PR? @danielfoehrKn @ialidzhikov

ialidzhikov:
> Any objections merging this PR? @danielfoehrKn @ialidzhikov

No objections, sorry for not being able to re-review.

PR #2284, opened 2020-05-05
Fix Shoot status transitioning to Failed too early

ialidzhikov:
Fix Shoot status transitioning to Failed too early.

Steps to reproduce the issue:
1. Create a Shoot with OS extension which is currently not present on the Seed

2. Delete the OS extension controller deployment to make sure that the extension ManagedResource won't be healthy to be able to reproduce clearly the issue

3. Ensure that the Shoot is marked as Failed after 1m as the required OS extension is not ready. We rather want to mark the Shoot as Failed after retries for 24h.

Fixes #

**Release note**:

```improvement operator
NONE
```


timuthy (approved):
/lgtm

Thanks for taking care.

ialidzhikov:
/kind/bug

PR #2286, opened 2020-05-05
Increase limits for CoreDNS + blackbox-exporter

rfranzke:
We sometimes observe that the blackbox-exporter cannot properly report the API server availability as either itself gets CPU-throttled or CoreDNS gets CPU-throttled or its request times out (probably related to the CPU-throttling itself, but may also occur for shoots where the worker nodes are far away from the control plane).

This PR adapts the limits slightly and increases the timeout for the BBX. It should prevent false negative API reports from the BBX.

* /cc @vlerenc
* We are having a `HorizontalPodAutoscaler` for CoreDNS that scales between 2-5 replicas and for `80%` target CPU average utilization. However, this seems to not work too well, so we should revisit this again.

**Release note**:

```improvement operator
The CPU limits for the `coredns` and `blackbox-exporter` deployments in the shoot have been slightly increased to prevent false negative API server availability reports.
```

### DIFF
--- a/charts/shoot-core/components/charts/coredns/values.yaml
+++ b/charts/shoot-core/components/charts/coredns/values.yaml
@@ -19,7 +19,7 @@ deployment:
       imagePullPolicy: IfNotPresent
       resources:
         limits:
-          cpu: 100m
+          cpu: 250m
           memory: 100Mi
         requests:
           cpu: 50m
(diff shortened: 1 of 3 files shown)


vlerenc:
Thanks a lot @rfranzke. And also to you @ggaurav10.

zanetworker (approved):
/lgtm but I am not sure how does the timeout solve the issue.

ggaurav10 (approved):

rfranzke:
/kind/bug

wyb1 (approved):

PR #2290, opened 2020-05-06
Fix duplicate Extension CRD reconcile in extension library

### LINKED ISSUE #2179: Infrastructure can be reconciled multiple times by Cluster CR updates
**What happened**:
Currently the infrastructure controller watches also for Cluster resource and maps events. This can lead Infrastucture to be enqueued for reconciliation when it is already reconciling.

https://github.com/gardener/gardener/blob/57ec6754b99f561be5e126cfee4c5f6d0b87b6b8/extensions/pkg/controller/infrastructure/controller.go#L97-L101

1. Start provider extension controller and apply an Infrastructure CR.

2. Update the Cluster CR spec while the Infrastructure is being reconciled (`.status.lastOperation.state=Processing`).

3. Ensure that the Cluster update event will enqueue the Infrastructure for 1 more reconciliation.

danielfoehrKn:
Gardenlet sets the operation annotation ( gardener.cloud/operation: reconcile) on extension CRDs to trigger a reconcile (if the extension does not deliberately ignore the annotation). This is for the Gardenlet to be able to control the reconcile times -e.g only during maintenance time.

If the extension respects the operation-annotation, the extensions now only watch the Extension CRD and no auxiliary resources as it was previously the case. 
 
In turn the Gardenlet must make sure that the watching clients in the extension receive an update event when the extension should be reconciled. Though the Gardenlet already sets the operation annotation, clients could sometimes miss the update event, leading to no/ delayed reconciles. In the following reconcile, the Gardenlet requests another reonciliation, but the operation annotation is already present. No update event will be send to the extensions -> Extension is not being reconciled anymore. 

To circumvent that issue, the Gardenlet now sets an annotation on the Extension CRDs  in addition to the operation-annotation that guarantees an update event (contains timestamp).

Fixes #2179

As the extension controller registration changes - there is a small code change required when revendoring Gardener in the Extensions.

I have split it in two commits with one containing the changes in the Gardenlet and the other the changes in the extensions.
The commits have a relationship to each other.  The extension controllers now only watch the Extension CRD, in turn the Gardenlet must add a new annotation to guarantee the update event.
**Release note**:

```improvement operator
The Gardenlet sets an additional annotation to Extension CRDs during reconciliation to guarantee an update event for the watching clients.
```
```improvement operator
Fixes a bug in the extension libraries that could lead to duplicate reconciliation of extension resources. When respecting the operation annotation set by the Gardenlet during reconciliation, extension controllers now only watch the Extension CRD.
```

### DIFF
--- a/extensions/pkg/controller/backupbucket/controller.go
+++ b/extensions/pkg/controller/backupbucket/controller.go
@@ -47,6 +47,10 @@ type AddArgs struct {
 	Predicates []predicate.Predicate
 	// Type is the type of the resource considered for reconciliation.
 	Type string
+	// IgnoreOperationAnnotation specifies whether to ignore the operation annotation or not.
+	// If the annotation is not ignored, the extension controller will only reconcile
+	// with a present operation annotation typically set during a reconcile (e.g in the maintenance time) by the Gardenlet
+	IgnoreOperationAnnotation bool
 }
 
 // DefaultPredicates returns the default predicates for a controlplane reconciler.
@@ -75,18 +79,20 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
 func Add(mgr manager.Manager, args AddArgs) error {
 	args.ControllerOptions.Reconciler = NewReconciler(mgr, args.Actuator)
 	predicates := extensionspredicate.AddTypePredicate(args.Predicates, args.Type)
-	return add(mgr, args.ControllerOptions, predicates)
+	return add(mgr, args, predicates)
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
-func add(mgr manager.Manager, options controller.Options, predicates []predicate.Predicate) error {
-	ctrl, err := controller.New(ControllerName, mgr, options)
+func add(mgr manager.Manager, args AddArgs, predicates []predicate.Predicate) error {
+	ctrl, err := controller.New(ControllerName, mgr, args.ControllerOptions)
 	if err != nil {
 		return err
 	}
 
-	if err := ctrl.Watch(&source.Kind{Type: &corev1.Secret{}}, &extensionshandler.EnqueueRequestsFromMapFunc{ToRequests: extensionshandler.SimpleMapper(SecretToBackupBucketMapper(predicates), extensionshandler.UpdateWithNew)}); err != nil {
-		return err
+	if args.IgnoreOperationAnnotation {
+		if err := ctrl.Watch(&source.Kind{Type: &corev1.Secret{}}, &extensionshandler.EnqueueRequestsFromMapFunc{ToRequests: extensionshandler.SimpleMapper(SecretToBackupBucketMapper(predicates), extensionshandler.UpdateWithNew)}); err != nil {
+			return err
+		}
 	}
 
 	return ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.BackupBucket{}}, &handler.EnqueueRequestForObject{}, predicates...)
--- a/extensions/pkg/controller/extension/reconciler.go
+++ b/extensions/pkg/controller/extension/reconciler.go
@@ -64,6 +64,10 @@ type AddArgs struct {
 	Resync time.Duration
 	// Type is the type of the resource considered for reconciliation.
 	Type string
+	// IgnoreOperationAnnotation specifies whether to ignore the operation annotation or not.
+	// If the annotation is not ignored, the extension controller will only reconcile
+	// with a present operation annotation typically set during a reconcile (e.g in the maintenance time) by the Gardenlet
+	IgnoreOperationAnnotation bool
 }
 
 // Add adds an Extension controller to the given manager using the given AddArgs.
@@ -101,13 +105,15 @@ func add(mgr manager.Manager, args AddArgs) error {
 
 	predicates := extensionspredicate.AddTypePredicate(args.Predicates, args.Type)
 
-	if err := ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.Extension{}}, &handler.EnqueueRequestForObject{}, predicates...); err != nil {
-		return err
+	if args.IgnoreOperationAnnotation {
+		if err := ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.Cluster{}}, &extensionshandler.EnqueueRequestsFromMapFunc{
+			ToRequests: extensionshandler.SimpleMapper(ClusterToExtensionMapper(predicates...), extensionshandler.UpdateWithNew),
+		}); err != nil {
+			return err
+		}
 	}
 
-	return ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.Cluster{}}, &extensionshandler.EnqueueRequestsFromMapFunc{
-		ToRequests: extensionshandler.SimpleMapper(ClusterToExtensionMapper(predicates...), extensionshandler.UpdateWithNew),
-	})
+	return ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.Extension{}}, &handler.EnqueueRequestForObject{}, predicates...)
 }
 
 // reconciler reconciles Extension resources of Gardener's
--- a/extensions/pkg/controller/infrastructure/controller.go
+++ b/extensions/pkg/controller/infrastructure/controller.go
@@ -49,6 +49,10 @@ type AddArgs struct {
 	Type string
 	// WatchBuilder defines additional watches on controllers that should be set up.
 	WatchBuilder extensionscontroller.WatchBuilder
+	// IgnoreOperationAnnotation specifies whether to ignore the operation annotation or not.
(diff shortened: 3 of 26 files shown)


rfranzke:
/kind/bug

pkg/operation/botanist/worker.go
```diff
@@ -136,6 +136,7 @@ func (b *Botanist) DeployWorker(ctx context.Context) error {
 
 	_, err = controllerutil.CreateOrUpdate(ctx, b.K8sSeedClient.Client(), worker, func() error {
 		metav1.SetMetaDataAnnotation(&worker.ObjectMeta, v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationReconcile)
+		metav1.SetMetaDataAnnotation(&worker.ObjectMeta, v1beta1constants.GardenerReconciliationRequest, time.Now().UTC().String())
```
rfranzke: It should be done for operating system configs as well.

rfranzke:
What about the timestamp annotation when triggering a deletion of an extension CRD?

extensions/pkg/controller/backupbucket/controller.go
```diff
@@ -42,16 +42,17 @@ type AddArgs struct {
 	// The options.Reconciler is always overridden with a reconciler created from the
 	// given actuator.
 	ControllerOptions controller.Options
-	// Predicates are the predicates to use.
-	// If unset, GenerationChangedPredicate will be used.
-	Predicates []predicate.Predicate
```
rfranzke, line 47: Why is this removed from the `AddArgs`? I think the idea was that the caller can provide custom provides here.

extensions/pkg/controller/backupbucket/controller.go
```diff
@@ -42,16 +42,17 @@ type AddArgs struct {
...
+	// IgnoreOperationAnnotation specifies whether to ignore the operation annotation or not.
+	// If the annotation is not ignored, the extension controller will only reconcile
+	// with a present operation annotation typically set during a reconcile (e.g in the maintenance time) by the Gardenlet
+	IgnoreOperationAnnotation bool
 }
 
 // DefaultPredicates returns the default predicates for a controlplane reconciler.
-func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
-	if ignoreOperationAnnotation {
+func (a AddArgs) DefaultPredicates() []predicate.Predicate {
```
rfranzke: Why is this now a function of the `AddArgs`? Can't we keep that the caller has to provide the "ignore operation annotation" via argument?

extensions/pkg/controller/infrastructure/controller.go
```diff
@@ -76,28 +77,32 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
...
 	}
 
-	if err := ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.Cluster{}}, &extensionshandler.EnqueueRequestsFromMapFunc{
-		ToRequests: extensionshandler.SimpleMapper(ClusterToInfrastructureMapper(predicates), extensionshandler.UpdateWithNew),
-	}); err != nil {
-		return err
+	// do not watch cluster if respect operation annotation to prevent unwanted reconciliations in case the operation annotation
+	// is already present & the extension CRD is already deleting
+	if args.IgnoreOperationAnnotation {
+		if err := ctrl.Watch(&source.Kind{Type: &extensionsv1alpha1.Cluster{}}, &extensionshandler.EnqueueRequestsFromMapFunc{
```
ialidzhikov, line 104: When I was discussing the issue with @timuthy, he shared the following scenario:
```
Suppose you have a cluster in deletion and after some time it transitions to Failed  state. Then the op or user can trigger a retry which resets the Failed  state, right? Therefore, the deletion must continue somehow but it’s not triggered by operation: reconcile  because it’s in deletion
```

Does this path from above work when we remove the Cluster watch with this PR?

ialidzhikov:
You could also update the description that the PR targets to fix #2179.

extensions/pkg/controller/infrastructure/controller.go
rfranzke: Yes, because we wanted to add the timestamp as annotation also for deletion (which will trigger the controller and make it enqueueing the extension object again for deletion), see https://github.com/gardener/gardener/pull/2290#issuecomment-624627927.

danielfoehrKn: Will add it to the deletion too.

extensions/pkg/controller/backupbucket/controller.go
danielfoehrKn: Added it again. Now during extension registration, the OperationAnnotation flag has to be handed over twice, but thats the price. Once for the predicates and once for the AddArgs.

pkg/operation/botanist/worker.go
danielfoehrKn: Adjusted.

pkg/apis/core/v1beta1/constants/types_constants.go
```diff
@@ -103,6 +103,12 @@ const (
 	// GardenerOperationReconcile is a constant for the value of the operation annotation describing a reconcile
 	// operation.
 	GardenerOperationReconcile = "reconcile"
+	// GardenerReconciliationRequest is a constant for an annotation on a resource that describes the timestamp when a reconcile has been requested.
+	// It is only used to guarantee an update event for watching clients in case the operation-annotation is already present.
+	GardenerReconciliationRequest = "gardener.cloud/reconciliation-request"
+	// GardenerDeletionRequest is a constant for an annotation on a resource that describes the timestamp when a deletion has been requested.
+	// It is only used to guarantee an update event for watching clients in case the deletion-annotation is already present.
+	GardenerDeletionRequest = "gardener.cloud/deletion-request"
```
rfranzke: I don't think we need two annotations, just use `gardener.cloud/timestamp` - it can be updated for both reconciliations and deletions.

extensions/pkg/controller/extension/reconciler.go
```diff
@@ -58,23 +58,25 @@ type AddArgs struct {
 	// The options.Reconciler is always overridden with a reconciler created from the
 	// given actuator.
 	ControllerOptions controller.Options
-	// Predicates are the predicates to use.
-	Predicates []predicate.Predicate
```
rfranzke: Can you re-add the predicates here as well, please?

extensions/pkg/controller/extension/reconciler.go
```diff
@@ -58,23 +58,25 @@ type AddArgs struct {
...
 	Type string
+	// IgnoreOperationAnnotation specifies whether to ignore the operation annotation or not.
+	// If the annotation is not ignored, the extension controller will only reconcile
+	// with a present operation annotation typically set during a reconcile (e.g in the maintenance time) by the Gardenlet
+	IgnoreOperationAnnotation bool
 }
 
 // Add adds an Extension controller to the given manager using the given AddArgs.
-func Add(mgr manager.Manager, args AddArgs) error {
+func Add(mgr manager.Manager, args AddArgs, predicates []predicate.Predicate) error {
```
rfranzke:
```suggestion
func Add(mgr manager.Manager, args AddArgs) error {
```

extensions/pkg/controller/network/controller.go
```diff
@@ -46,6 +46,10 @@ type AddArgs struct {
 	Predicates []predicate.Predicate
 	// Type is the type of the resource considered for reconciliation.
 	Type string
+	// IgnoreOperationAnnotation specifies whether to ignore the operation annotation or not.
+	// If the annotation is not ignored, the extension controller will only reconcile
+	// with a present operation annotation typically set during a reconcile (e.g in the maintenance time) by the Gardenlet
+	IgnoreOperationAnnotation bool
```
rfranzke, line 52: You missed to only conditionally enable the `Cluster` mapper below in the `add` function.

extensions/pkg/controller/reconciler.go
```diff
@@ -81,6 +81,7 @@ func (o *operationAnnotationWrapper) Reconcile(request reconcile.Request) (recon
 	if annotations[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationReconcile {
 		withOpAnnotation := obj.DeepCopyObject()
 		delete(annotations, v1beta1constants.GardenerOperation)
+		delete(annotations, v1beta1constants.GardenerReconciliationRequest)
```
rfranzke: Can you remove it again? We didn't want to make it an official contract that extensions "have" to follow. The Gardenlet "just" adds the annotation, but it isn't needed to be removed again. Also, I think now that we have the timestamp annotation it could make operations  somewhat easier in some cases.

rfranzke (changes requested):

extensions/pkg/controller/reconciler.go
danielfoehrKn: ok, I see, will do.

rfranzke (approved):
Nice, thank you very much!

/lgtm

PR #2300, opened 2020-05-07
Change port of GSAC from 9443 to 443

rfranzke:
This PR changes the port of the GSAC deployment + service from `9443` to `443`. The reason is that the GKE clusters assume that webhooks are exposed at port `443`.

See https://github.com/kubernetes/kubernetes/issues/79739

**Release note**:

```improvement operator
The validating webhook for the Gardener Seed Admission controller is now exposed at port `443`, allowing it to function properly in GKE clusters.
```


rfranzke:
See also: https://github.com/gardener-attic/gardener-extensions/commit/460df4a2392a67929607504910eb0d6118e94faa

ialidzhikov (approved):
/lgtm

PR #2305, opened 2020-05-08
Allow K8s pod to exclude their logs from the fluent-bit log processor

vlvasilev:
With this PR when the installation of the logging stack is skipped for a particular shoot, the pods in that shoot control plane are annotated so their logs to be excluded from processing.
Thus fluentd-es will be  partially unloaded.
This PR does not handle pods spawned by extensions controllers.
Fixes #

**Release note**:

```improvement operator
Allow K8s pod to exclude their logs from the fluent-bit log processor when logging is turned off for those pods
```

### DIFF
--- a/charts/seed-bootstrap/charts/fluentd-es/templates/fluent-bit-configmap.yaml
+++ b/charts/seed-bootstrap/charts/fluentd-es/templates/fluent-bit-configmap.yaml
@@ -241,6 +241,7 @@ data:
         Labels              Off
         Annotations         Off
         tls.verify          Off
+        K8S-Logging.Exclude On
 
     [FILTER]
         Name record_modifier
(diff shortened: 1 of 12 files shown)


rfranzke:
Hm, not sure if that's the right approach. Wouldn't it be more convenient to let the Gardener Seed Admission Controller do the job and inject the pod annotation via webhook? This would also cover the pods created by extensions.

vlvasilev:
@rfranzke Am I allowed to add mutating functionality in admission controller?
If so I will try to do it in the Gardener Seed Admission Controller.

rfranzke:
Sure, why not :) That was the rationale of adding it. So far it only had one use case (CRD deletion prevention), but now we have a new one.
Are you okay with closing this PR then in favor of a new one that adapts GSAC?

PR #2309, opened 2020-05-11
Add /healthz endpoint for Gardenlet

rfranzke:
This PR enhances the Gardenlet with a HTTPS server that serves the `/healthz` and `/metrics` endpoints. The `/healthz` endpoint returns `200 OK` if the Gardenlet can regularly renew its seed lease. Otherwise, it will return `500 Internal Server Error`.

**Release note**:

```action operator
The Gardenlet does now run a HTTPS server that serves a `/healthz` and `/metrics` endpoint. You should generate a server certificate for the `gardenlet`, `gardenlet.garden`, `gardenlet.garden.svc` hosts. The bind address, port, and TLS certificate paths are configurable [in its component config](https://github.com/gardener/gardener/blob/master/example/20-componentconfig-gardenlet.yaml#L75-L81). Also, the `gardenlet` Helm chart was enhanced with a liveness probe that targets the `/healthz` endpoint.
```


rfranzke:
/kind/enhancement

timuthy:
Nice PR!
Can you elaborate why we expose the new endpoints via HTTP? I think especially the `/metrics` endpoint should be exposed via HTTPS only, shouldn't it?

rfranzke:
No specific reason, [it just was like this before](https://github.com/gardener/gardener/blob/master/pkg/server/server.go#L80). I can also move both to HTTPS, should be no problem.

pkg/gardenlet/apis/config/v1alpha1/defaults.go
```diff
@@ -114,6 +114,13 @@ func SetDefaults_GardenletConfiguration(obj *GardenletConfiguration) {
 		v := DefaultKubernetesLogLevel
 		obj.KubernetesLogLevel = &v
 	}
+
+	if len(obj.Server.HTTPS.BindAddress) == 0 {
+		obj.Server.HTTPS.BindAddress = "0.0.0.0"
+	}
+	if obj.Server.HTTPS.Port == 0 {
+		obj.Server.HTTPS.Port = 2020
```
timuthy:
```suggestion
		obj.Server.HTTPS.Port = 2720
```
To be consistent with `charts/gardener/gardenlet/values.yaml`.

cmd/gardenlet/app/gardenlet.go
```diff
@@ -377,8 +382,24 @@ func (g *Gardenlet) Run(ctx context.Context) error {
...
 	leaderElectionCtx, leaderElectionCancel := context.WithCancel(context.Background())
 
+	// Initialize /healthz manager.
+	g.HealthManager = healthz.NewPeriodicHealthz(seedcontroller.LeaseResyncGracePeriodSeconds * time.Second)
+	g.HealthManager.Set(true)
+
+	// Start HTTP server.
+	go server.
+		NewBuilder().
+		WithBindAddress(g.Config.Server.HTTPS.BindAddress).
```
timuthy, line 392: Potential nil pointer dereference since `g.Config.Server` is optional.

pkg/gardenlet/apis/config/v1alpha1/defaults.go
```diff
@@ -114,6 +114,13 @@ func SetDefaults_GardenletConfiguration(obj *GardenletConfiguration) {
 		v := DefaultKubernetesLogLevel
 		obj.KubernetesLogLevel = &v
 	}
+
+	if len(obj.Server.HTTPS.BindAddress) == 0 {
```
timuthy, line 121: Potential nil pointer dereference.

pkg/server/server.go
```diff
@@ -20,69 +20,49 @@ import (
...
+	bindAddress  string
+	port         int
+	tlsCertPath  *string
+	tlsKeyPath   *string
+	handlers     map[string]http.Handler
+	handlerFuncs map[string]http.HandlerFunc
 
-	"k8s.io/client-go/tools/cache"
-)
+	serverType string
```
timuthy: Unused variable.

timuthy (changes requested):

cmd/gardenlet/app/gardenlet.go
rfranzke: I've added some config validation that is executed beforehand so that we can safely access these fields here.

timuthy (commented):
Looks good to me in general.

I noticed one issue with the health check though, if the Gardenlet is responsible for more than one seed and the lease of one of the seeds cannot be renewed. This scenario will end up in a constantly swapping health state. OTOH, such a setup is not recommended for production use cases, so I can understand if you don't want to tackle it. WDYT?

rfranzke:
Do you have a suggestion how what the expected behaviour should be and how to implement it?

timuthy:
> Do you have a suggestion how what the expected behaviour should be and how to implement it?

Well, the health manager needs to be aware of its callers. Eventually only the caller identity (e.g. seed name) which once set `c.healthManager.Set(false)` can reset the whole health status of the manager to `true` again. Other `c.healthManager.Set(true)` calls will be ignored in the meantime.

rfranzke:
I don't think the health manager should be responsible for this. Its sole task should be to store/manage the health status, but the actual information (healthy/unhealthy) should come from the caller.
I was thinking about a separate Go routine in the seed controller which regularly (e.g., every `5s` calculates for how many seeds the gardenlet is responsible). The seed lease reconciler reports the individual seed status into a shared map. The mentioned Go routine loops over it, and if all seeds renewed their lease then it sets the health status to `true`, otherwise it sets it to `false` - in both cases it resets the map for next iteration. WDYT?

rfranzke:
I implemented as discussed via phone/https://github.com/gardener/gardener/pull/2309#issuecomment-627179774, PTAL again.

pkg/gardenlet/controller/seed/seed.go
```diff
@@ -208,6 +220,61 @@ func (c *Controller) Run(ctx context.Context, workers int) {
...
+	if seedName != "" {
+		expectedHealthReports = 1
+	} else {
+		seedLabelSelector, err = metav1.LabelSelectorAsSelector(c.config.SeedSelector)
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	for {
```
timuthy, line 240: Now with this for loop, do you think it is still relevant to use a `periodicHealthz`? Seems like we already have a reset mechanism with `c.leaseMap = make(map[string]bool)` here.

The reason I ask is because of the time intervals in different go routines we're about to introduce: `periodicHealthz` (20 sec.) - this loop (10 sec.) - `seed_lease_control` (2 sec.). Maybe we cloud simplify this handling if we went w/o `periodicHealthz`.

rfranzke: I think it's still relevant, yes. What if this controller breaks/blocks for some reason? We still want the `/healthz` to be set to `false` in this case.

timuthy: Okay, I'm not quite sure how this can cause an issue since we are doing it in a Go routine but also no objections against introducing it for safety reason 👍 Thanks for letting know.

timuthy (dismissed):
/lgtm

Can you check the failing pipeline job?

pkg/gardenlet/controller/seed/seed.go
rfranzke: Yeah, I agree, it's not a very good example, but I think it's still safer to let it automatically reset by an independent mechanism.

timuthy (dismissed):
/lgtm

rfranzke:
/hold
I have to adapt the gardenlet self-deployment to also configure the new server section.

rfranzke:
/area/quality
/area/robustness
/size/l
/topology/seed
/needs/changes
/platform/all
/priority/normal

rfranzke:
/unhold
/drop needs changes
/needs/review
/cc @timuthy

timuthy (approved):

vpnachev:
Now when #2361 is merged, I've removed the release note

```
action developer
Developers have to run `make dev-setup` in order to populate the Gardenlet certificates to their `dev` directories.
```

PR #2310, opened 2020-05-11
Add PDB and anti-affinity for kube-apiserver

### LINKED ISSUE #2294: Increase kube-apiserver replicas and add anti-affinity
Increase the default number of replicas for the API server deployment to 3 and add anti-affinity rules for the apiserver pods such that they are each running on a different node

**Why is this needed**:
If the node where the apiserver is running is not working, this causes access to the apiserver for the shoot to go down

rfranzke:
This PR adds an anti-affinity rule based on the node hostname for the kube-apiserver deployment of shoots. It also adds a pod disruption budget that states that at most one replica might be unavailable at any given point in time.

Fixes #2294 

**Release note**:

```improvement operator
The `kube-apiserver` deployment of shoots is now scheduled with anti-affinity and a pod disruption budget of `1`.
```


charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
```diff
@@ -44,6 +44,20 @@ spec:
         networking.gardener.cloud/to-shoot-networks: allowed
         networking.gardener.cloud/from-prometheus: allowed
     spec:
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
```
vlerenc: Do we really need to `require` it or can we also `prefer` it? In most cases `prefer` should yield the same result, except for small/bin-packed clusters and then it would block, which is something that may become dangerous to the stability of the control plane. Even if 2 pods end up on the same node, normal eviction would work even with the PDB (one by one within the node drain timeout) and only if the node dies, multiple API servers would be gone, but that's highly unlikely and even if, API servers start fast and if there are more than the ones on that particular node, then the others might be able to support the load for a brief time window. Bottom line, I would go with `prefer` as it's closer to what we had before/less risky. WDYT?

vlerenc:
Thanks @rfranzke. I have only the suggestion to consider `prefer` over `require`, see inline comment.

timebertt (commented):
I only have the same question as @vlerenc wrt `prefer`/`required`.
Rest looks good.

charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
rfranzke: Sure, makes sense, thanks for spotting.

timebertt (approved):
/lgtm

vlerenc:
@rfranzke @tim-ebert I just took a look at an API server (because I checked another thing, the hostPort Konnectivity-Service issue) and am wondering whether I missed here something in the review. The anti-affinity is global/across all control planes, because I don't see a control plane label selector at pod level. Is that so? (I guess it's not relevant and maybe not even true, but theoretically this may have helped to hide the hostPort clash for longer).

That in turn means, that the scheduler will first place 1 API server per node and then we are back at square 1 and they will get assigned randomly again. What we wanted was to *avoid placing API servers of the same control plane onto the same node*, but we can easily place 20 API servers onto the same node as long as they belong to other control planes.

WDYT?

rfranzke:
You're right. We would need a shoot-specific label on the kube-apiserver so achieve the desired behaviour.
/reopen

vlerenc:
Thanks @rfranzke (and for the smile @tim-ebert :-)). OK, good that we spotted it. What a clashing `hostPort` can do. ;-)

As for the robot issue, maybe I should ignore the command based on the state. The validation error was: `"Validation Failed", "errors": [{"resource": "PullRequest", "code": "custom", "field": "state", "message": "state cannot be changed. The pull request cannot be reopened."}`

rfranzke:
Sure. Still, I find it kind of "surprising" that these anti-affinities are considered cross namespace although `Pod` is a namespaced object. Anyways, will work on the fix.

vlerenc:
Thanks @rfranzke .

vlerenc:
@rfranzke made aware of the docs in https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#an-example-of-a-pod-that-uses-pod-affinity and that the scheduler has some implicit logic that applies to namespaces, so everything is good as the API servers per control plane have anti-affinity, but only per control plane/within one namespace. Beyond, they can be scheduled on the same node/this has no effect. I tried with 3 pods that all have the same label and `requiredDuringSchedulingIgnoredDuringExecution` anti-affinity (`required` to proof the point). Only one pod can be started in one NS, but another identical pod in another NS is not conflicting/considered:
```
ns-1          api-1                                                 1/1     Running   0          70s
ns-1          api-2                                                 0/1     Pending   0          15s
ns-2          api-1                                                 1/1     Running   0          11s
```
Thanks for checking and sorry for the question/disturbance all / @zanetworker / @tim-ebert .

Issue #2193, opened 2020-04-16
Project Tolerations to support limiting access to Seed usage


rfranzke:
**What would you like to be added**:
Today, the `Seed` can have several taints specified in the `.spec.taints[]` list. One of them is `seed.gardener.cloud/protected`. If set, only shoots in the `garden` namespace are allowed to specify the seed in the `.spec.seedName`. However, the `garden` namespace is hard-coded (see #401) which is not good. Also, there is a new use case where specific seeds shall be restricted to specific (multiple) projects only.

Let's adapt our `Project` resource with a new `.spec.tolerations[]` field:

```yaml
---
apiVersion: core.gardener.cloud/v1beta1
kind: Project
metadata:
  name: my-project
spec:
  ...
  tolerations:
  - key: <some-key>
```

If all taints of a `Seed` are tolerated by the `Project` then the `Shoot` is allowed to use it. Today, we also have other taints (`invisible`, `disable-dns`, `excess-capacity`). These are more settings of the seed and don't fit to the taint/toleration concept. Hence, we should move them out of the `.spec.taints[]` to a new `.spec.settings[]` field.

Now a `Seed` can have the `<seed-name>.seed.gardener.cloud/protected` taint. If the project has the `<seed-name>.seed.gardener.cloud/protected` toleration then a `Shoot` is allowed to be scheduled on it.

In order to prevent that end-users (which are having `update`,`patch` privileges) add arbitrary `tolerations` we should use a custom RBAC verb. Only with it it would be allowed to change the `.spec.tolerations` field:

```yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: gardener.cloud:project-tolerations
rules:
- apiGroups:
  - core.gardener.cloud
  resources:
  - projects
  verbs:
  - tolerations
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: gardener.cloud:project-tolerations
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: gardener.cloud:project-tolerations
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: gardener-admin
```

The scheduler and our admission should be enhanced to respect these new taints/tolerations concepts.

PS: In a future iteration we could think about introducing a new namespaced `Seed` resource, but this is out of scope of this issue.

**Why is this needed**:
[...]

Issue #1953, opened 2020-02-19
☂️-Issue for network cost optimizations across Gardener components


zanetworker:
Network cost contribute a large portion of our total costs, therefore, it makes sense to monitor and optimize networking to reflect lower costs.

The optimizations should cover a large umbrella including but not limited to:

**Seeds**
- [x] Gardenlet (https://github.com/gardener/gardener/pull/1928)
- [x] Resource Manager  (https://github.com/gardener/gardener/pull/1931, https://github.com/gardener/gardener-resource-manager/pull/40, https://github.com/gardener/gardener-resource-manager/pull/47)
- [x] Dependency Watchdog - talks to the shoots loadbalancers (https://github.com/gardener/gardener/pull/1954)
- [x] Gardener Extensions (health checks) – library imported by every extension and uses the loadbalancer for to access the shoot APIs (https://github.com/gardener/gardener-extensions/pull/602)
- [x] Shoot controlplanes
   - [x] Cloud-controller-manager?
- [x] Monitoring @wyb1
   - [x] Seed Prometheus (#1976)
   - [x] Aggregate Prometheus (#1976)
   - [x] Grafana (access to dashboard repositiories - raw.github.usercontent.com and grafana.com) (#1977)
- [x] OS Processes
   - [x] GCP  (google_accounts_daemon, google_clock_skew_daemon, google_network_daemon)
   - [x] Azure (waagent daemon) -> `/usr/share/oem/bin/waagent`
   - [x] Other providers (needs more investigation)
- [x] Image Pulls (utilize caching or p2p registeries?)

**Garden**:
- [x] Seed Kube-apiservers -> Catalog + Metrics server
- [x] Auditlog pods -> api.auditlog.sap.hana.ondemand.com (@vpnachev)
- [x] Prometheus + Garden Promentheus (@wyb1)
- [x] Seed Kube-controller-managers
- [x] Seed Cloud-controller-managers

**Possible fixes:** 

- Fix the health-checks to access the API-servers of the shoots in-cluster instead of going through the load-balancer.
- Reduce Prometheus scrape intervals, especially non-availability alert metrics.
- Grafana (can we use an offline dashboard library, do we really need access to the internet, should we block by network policies).
- Kube-proxy + other system components + cloud-controller-managers (we probably won’t be able to influence much here, can we really?).
- Processes (some can be stopped I guess, others seems to be important such as the waagent and the google_accounts_damons)

If there are other components that you think are making egress connections, please add them to the list. 

[...]

PR #2315, opened 2020-05-12
Add settings for seeds and deprecate hard-coded taints

rfranzke:
This PR adds new settings fields in the seed spec in order to proceed with deprecating the hard-coded taints in the core API. As part of #2193 we envision to add support for taints/tolerations for seeds/shoots similar to nodes/pods.
The `protected` taint is planned to be deprecated as part of a future PR as well.

Part of #2193

**Release note**:

```noteworthy operator
The following taints have been deprecated and will be removed in a future version:
* `seed.gardener.cloud/disable-capacity-reservation` in favour of the new `.spec.settings.excessCapacityReservation.enabled` field.
* `seed.gardener.cloud/invisible` in favour of the new `.spec.settings.scheduling.visible` field.
* `seed.gardener.cloud/disable-dns` in favour of the new `.spec.settings.shootDNS.enabled` field.
The `.controllers.seed.reserveExcessCapacity` setting in the component config of the Gardenlet has been removed in favour of the new settings field mentioned earlier.
Please check [this document](https://github.com/gardener/gardener/blob/master/docs/usage/seed_settings.md) for further information about the seed settings.
```


rfranzke:
/kind/api-change

rfranzke:
/area/usability
/size/xl
/topology/garden
/needs/review
/platform/all
/priority/normal

rfranzke:
ping

example/50-seed.yaml
```diff
@@ -35,11 +35,16 @@ spec:
   #   services: 100.64.0.0/13
     blockCIDRs:
     - 169.254.169.254/32
+  settings:
+    excessCapacity:
+      enabled: true # this seed will deploy excess-capacity-reservation pods
+    scheduling:
+      visible: true # the gardener-scheduler will consider this seed for shoots
+    shootDNS:
+      enabled: true # all shoots on this seed will use DNS, if disabled they'll just the plain IPs/hostnames
```
vpnachev:
```suggestion
      enabled: true # all shoots on this seed will use DNS, if disabled they'll just use the plain IPs/hostnames
```

example/50-seed.yaml
```diff
@@ -35,11 +35,16 @@ spec:
...
 # taints:
-# - key: seed.gardener.cloud/disable-dns                  # all shoots on this seed won't use any DNS, just the plain IPs/hostnames
-# - key: seed.gardener.cloud/protected                    # only shoots in the `garden` namespace can use this seed
-# - key: seed.gardener.cloud/invisible                    # the gardener-scheduler won't consider this seed for shoots
-# - key: seed.gardener.cloud/disable-capacity-reservation # this seed will not deploy excess-capacity-reservation pods
+# - key: seed.gardener.cloud/protected # only shoots in the `garden` namespace can use this seed
```
vpnachev, line 46: I think it makes sense to still keep the old taints here, but prefix them with 
```
DEPRECATED, please use `spec.settings...` field
```

pkg/apis/core/helper/helpers.go
```diff
@@ -133,6 +133,21 @@ func TaintsHave(taints []core.SeedTaint, key string) bool {
 	return false
 }
 
+// SeedSettingExcessCapacityReservationEnabled returns true if the 'excess capacity reservation' setting is enabled.
+func SeedSettingExcessCapacityReservationEnabled(settings *core.SeedSettings) bool {
+	return settings != nil && settings.ExcessCapacityReservation != nil && settings.ExcessCapacityReservation.Enabled
```
vpnachev:
```suggestion
	return settings == nil || settings.ExcessCapacityReservation == nil || settings.ExcessCapacityReservation.Enabled
```

My point is that by default the excess capacity is enabled, thus this method should return `true` even when excess capacity is not explicitly enabled. 

The same comment is applicable for visible and dns.

pkg/apis/core/types_seed.go
```diff
@@ -139,6 +141,39 @@ type SeedProvider struct {
...
 }
 
+// SeedSettings contains certain settings for this seed cluster.
+type SeedSettings struct {
+	// ExcessCapacityReservation controls the excess capacity reservation for shoot control planes in the seed.
+	ExcessCapacityReservation *SeedSettingExcessCapacityReservation
+	// Scheduling controls settings for scheduling decisions for the seed.
+	Scheduling *SeedSettingScheduling
+	// ShootDNS controls the shoot DNS settings for the seed.
+	ShootDNS *SeedSettingShootDNS
```
vpnachev, line 151:
```suggestion
	// DNS controls the shoot DNS settings for the seed.
	DNS *SeedSettingDNS
```

In the future DNS resources can be used for system components, for example the monitoring and logging in the `garden` namespace, and then this name will be misleading.

pkg/apis/core/v1alpha1/defaults.go
```diff
@@ -85,6 +85,43 @@ func SetDefaults_VolumeType(obj *VolumeType) {
...
+	}
+
+	if obj.Spec.Settings.ExcessCapacityReservation == nil {
+		enabled := true
+		for _, taint := range obj.Spec.Taints {
+			if taint.Key == DeprecatedSeedTaintDisableCapacityReservation {
+				enabled = false
+			}
+		}
+		obj.Spec.Settings.ExcessCapacityReservation = &SeedSettingExcessCapacityReservation{Enabled: enabled}
```
vpnachev, line 101: I have no strong opinion, but can you avoid explicitly setting the value when it is `true`?
Syncing the seed spec to the every cluster in the seed is contributing to https://github.com/gardener/gardener/issues/1953 and avoiding to set the explicit values will help a bit to the network optimization.

pkg/apis/core/v1beta1/types_seed.go
```diff
@@ -163,21 +202,25 @@ type SeedTaint struct {
 	Value *string `json:"value,omitempty" protobuf:"bytes,2,opt,name=value"`
 }
 
+// TODO: Remove these taints in the next API version in favor of the .spec.settings field.
```
vpnachev: API or g/g version?

vpnachev (commented):
Thank you!
Like always, looks very good.

I have just a few minor comments/questions.

pkg/apis/core/helper/helpers.go
rfranzke: Thanks for being so careful, indeed!

pkg/apis/core/types_seed.go
rfranzke: Hm, not sure whether I get this one. We already have `.spec.dns` in the `Seed`. Now having `.spec.settings.dns` will be misleading imo. Hence, I explicitly called it `ShootDNS` because all the configuration related to the seed DNS is/will be part of `.spec.dns` already.

What do you mean by "for example the monitoring and logging in the garden namespace"?

pkg/apis/core/v1alpha1/defaults.go
rfranzke: I'm not really sure what you are suggesting here and why. Do you argue that it is not necessary setting it to `true` because the default (if `.spec.settings==nil` or `.spec.settings.excessCapacityReservation==nil` is anyways `true`)? If yes, then: I still think it's more explicit to set the field as it'll be the new default.

With respect to the argument that it'll help with optimizing the network traffic: Theoretically yes, of course, but I don't think we are at a point yet where we have to worry about these few bytes ;-) If it's really that costly then we can also optimize by syncing resources like the `Seed` only once and not as part of each `Cluster` resource.

--> I would like to stay explicit here if there is no strong reason against it.

pkg/apis/core/v1beta1/types_seed.go
rfranzke: _API_ version ;-) I made it more explicit.

gardener-botmanager:
Command "/cc" failed with "Review cannot be requested from pull request author.".
<details>
<summary>Additional Information</summary>

```
Redacted in public. Check backend logs.
```
</details>

pkg/apis/core/types_seed.go
vpnachev: I forgot about `seed.spec.dns`, you can ignore my comment above.

pkg/apis/core/v1alpha1/defaults.go
vpnachev: You got my point.
I am fine with keeping it like this.

vpnachev (approved):
/approve

Issue #2304, opened 2020-05-08
Configure cluster autoscaler safe-to-evict annotation only for etcd-main


swapnilgm:
**What would you like to be added**:
Gardener should only apply the `cluster-autoscaler.kubernetes.io/safe-to-evict=false` annotation from etcd-main crd. And not the etcd-events. 
**Why is this needed**:
Etcd-events is not that critical component. And kube-apiserver does not have blocker in case etcd-events rescheduling takes time in absence of such annotation. This will help saving costs because of underused nodes on which etcd-events pods will be scheduled.
 
With https://github.com/gardener/etcd-druid/issues/54, the etcd-druid will no longer add such annotation by default to etcd pod. And the annotation should be explicitly configured on etcd resources.

rfranzke:
@swapnilgm @georgekuruvillak @amshuman-kr do you take care of this one?

georgekuruvillak:
Ill take it up.

swapnilgm:
Closed via #2317

PR #2317, opened 2020-05-12
safe-to-evict annotation only added for etcd-main

georgekuruvillak:
This PR removes `safe-to-evict` annotation from `etcd-events` pods. `etcd-main` pods continue to have the annotation. This prevents `cluster-autoscaler` causing re-scheduling of `etcd-main` pods while it is fine to do so for `etcd-events` pods.
Fixes #
#2304 

**Release note**:

```improvement operator
Removed `safe-to-evict` cluster-autoscaler annotation from `etcd-events` pods. 
```

### DIFF
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -1090,11 +1090,6 @@ func (b *Botanist) DeployETCD(ctx context.Context) error {
 			v1beta1constants.GardenerOperation: v1beta1constants.GardenerOperationReconcile,
 			v1beta1constants.GardenerTimestamp: time.Now().UTC().String(),
 		},
-		"podAnnotations": map[string]interface{}{
-			"checksum/secret-etcd-ca":          b.CheckSums[v1beta1constants.SecretNameCAETCD],
-			"checksum/secret-etcd-server-cert": b.CheckSums[common.EtcdServerTLS],
-			"checksum/secret-etcd-client-tls":  b.CheckSums[common.EtcdClientTLS],
-		},
 		"storageCapacity": b.Seed.GetValidVolumeSize("10Gi"),
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/operation/botanist/controlplane.go
```diff
@@ -1169,6 +1171,7 @@ func (b *Botanist) DeployETCD(ctx context.Context) error {
 			}
 
 		case common.EtcdRoleEvents:
+			delete(podAnnotations, "cluster-autoscaler.kubernetes.io/safe-to-evict")
```
swapnilgm: Not required, Right? `podAnnotation` is newly declared map variable and doesn't include this annotation.

georgekuruvillak: It is outside the for loop. So its same for both. Because etcd-main is processed first, the annotation is added. I can rearrange the etcd order; first to reconcile etcd-events and then etcd-main.

swapnilgm: Got your point. But then don't make it order dependent. Instead please move `podAnnotation` declaration inside for loop.

swapnilgm (approved):
Thank you for the PR.
/lgtm 
/do-not-merge
This pr rerquires release of etcd-druid. But we are expecting etcdbr release with metrics initialisation fix soon. Which will add etcdbr image change fix to etcd-druid. Let's wait till it and not merge immediately.

rfranzke (approved):

rfranzke:
@swapnilgm why can't we merge it yet? It won't have any effect with etcd-druid, but the change is not directly depending on it, right? Or will it break any existing behaviour when merging now?

rfranzke:
/kind/enhancement
/area/auto-scaling
/size/xs

swapnilgm:
hmm. Theoretically, there won't be any issue. For etcd-main sts, while template rendering there will be two entries for annotation, i am not sure about the template rendering behaviour in that case. Will have to test it with old image then.

rfranzke:
Hm, I guess it should be no issue because the annotations is a `map` and  the entries are identical, so they will only exist once. Can you confirm so that we can go ahead with the merge?

ialidzhikov:
I also find it confusing as the real fix for the issue will come with this change + etcd-druid release. Why you just don't combine this change  + etcd-druid version bump (when available)? IMO it is much easier to track. ;)

swapnilgm:
> Hm, I guess it should be no issue because the annotations is a map and the entries are identical, so they will only exist once. Can you confirm so that we can go ahead with the merge?

Yeah. That's why i said theoretically yes. But anyways i tested it and it works fine. So, this can go without druid release. Anyways, i'm will trigger druid release in while, so let's have both together.

>I also find it confusing as the real fix for the issue will come with this change + etcd-druid release. Why you just don't combine this change + etcd-druid version bump (when available)? IMO it is much easier to track. ;)

Yes. Initial plan was same. Soon, i'll trigger release for druid and we make this PR merge to auto-generated branch for druid version update.

rfranzke:
Depends on #2334

PR #2320, opened 2020-05-13
Мutate shoot control plane pods annotation so to be excluded from log processing.

vlvasilev:
With this PR when a shoot is for testing purpose or it has been set to hibernate the Gardener-Seed-Admission-Controller will annotate the shoot's pod with `fluentbit.io/exclude=true` which will exclude the pod's logs from the fluent-bit log processor.

**Release note**:

```improvement operator
Exclude logs from testing purpose or hibernated shoots
```

### DIFF
--- a/charts/seed-bootstrap/templates/gardener-seed-admission-controller/mutatingwebhookconfiguration.yaml
+++ b/charts/seed-bootstrap/templates/gardener-seed-admission-controller/mutatingwebhookconfiguration.yaml
@@ -1,0 +1,37 @@
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
+metadata:
+  labels:
+    app: gardener
+    role: seed-admission-controller
+  name: gardener-seed-admission-controller
+webhooks:
+- name: pods.seed.admission.core.gardener.cloud
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - pods
+    scope: '*'
+  admissionReviewVersions:
+  - v1beta1
+  clientConfig:
+    caBundle: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUMrakNDQWVLZ0F3SUJBZ0lVVHAzWHZocldPVk04WkdlODZZb1hNVi9VSjdBd0RRWUpLb1pJaHZjTkFRRUwKQlFBd0ZURVRNQkVHQTFVRUF4TUthM1ZpWlhKdVpYUmxjekFlRncweE9UQXlNamN4TlRNME1EQmFGdzB5TkRBeQpNall4TlRNME1EQmFNQlV4RXpBUkJnTlZCQU1UQ210MVltVnlibVYwWlhNd2dnRWlNQTBHQ1NxR1NJYjNEUUVCCkFRVUFBNElCRHdBd2dnRUtBb0lCQVFDeWkwUUdPY3YyYlRmM044T0xOOTdSd3NnSDZRQXI4d1NwQU9ydHRCSmcKRm5mblUyVDFSSGd4bTdxZDE5MFdMOERDaHYwZFpmNzZkNmVTUTRacmpqeUFyVHp1ZmI0RHRQd2crVldxN1h2RgpCTnluKzJoZjRTeVNrd2Q2azdYTGhVVFJ4MDQ4SWJCeUM0ditGRXZtb0xBd3JjMGQwRzE0ZWM2c25EKzdqTzdlCmt5a1EvTmdBT0w3UDZrRHM5ejYrYk9mZ0YwbkdOK2JtZVdRcUplalIwdCtPeVFEQ3g1L0ZNdFVmRVZSNVFYODAKYWVlZmdwM0pGWmI2ZkF3OUtoTHRkUlYzRlAwdHo2aFMrZTRTZzBtd0FBT3FpalpzVjg3a1A1R1l6anRjZkExMgpsRFlsL25iMUd0VnZ2a1FENDlWblY3bURubDZtRzNMQ01OQ05INldsWk52M0FnTUJBQUdqUWpCQU1BNEdBMVVkCkR3RUIvd1FFQXdJQkJqQVBCZ05WSFJNQkFmOEVCVEFEQVFIL01CMEdBMVVkRGdRV0JCU0ZBM0x2Sk0yMWQ4cXMKWlZWQ2U2UnJUVDl3aVRBTkJna3Foa2lHOXcwQkFRc0ZBQU9DQVFFQW5zL0VKM3lLc2p0SVNvdGVRNzE0cjJVbQpCTVB5VVlUVGRSSEQ4TFpNZDNSeWt2c2FjRjJsMnk4OE56NndKY0F1b1VqMWg4YUJEUDVvWFZ0Tm1GVDlqeWJTClRYclJ2V2krYWVZZGI1NTZuRUE1L2E5NGUrY2IrQ2szcXkvMXhnUW9TNDU3QVpRT0Rpc0RaTkJZV2tBRnMyTGMKdWNwY0F0WEp0SXRoVm03RmpvQUhZY3NyWTA0eUFpWUVKTEQwMlRqVURYZzRpR09HTWtWSGRtaGF3QkRCRjNBagplc2ZjcUZ3amk2SnlBS0ZSQUNQb3d5a1FPTkZ3VVNvbTg5dVlFU1NDSkZ2TkNrOU1KbWpKMlB6RFV0NkN5cFI0CmVwRmRkMWZYTHd1d243ZnZQTW1KcUQzSHRMYWxYMUFabVBrK0JJOGV6ZkFpVmNWcW5USlFNWGxZUHBZZTlBPT0KLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo=
+    service:
+      name: gardener-seed-admission-controller
+      namespace: garden
+      path: /webhooks/mutate-pods
+  failurePolicy: Ignore
+  matchPolicy: Exact
+  namespaceSelector:
+    matchLabels:
+      gardener.cloud/role: shoot
+  objectSelector: {}
+  reinvocationPolicy: Never
+  sideEffects: None
+  timeoutSeconds: 10
--- a/pkg/seedadmission/extension_crds.go
+++ b/pkg/seedadmission/extension_crds.go
@@ -16,7 +16,6 @@ package seedadmission
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
@@ -24,7 +23,6 @@ import (
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	gardenlogger "github.com/gardener/gardener/pkg/logger"
 	"github.com/gardener/gardener/pkg/operation/common"
-	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 
 	"github.com/sirupsen/logrus"
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
@@ -32,7 +30,6 @@ import (
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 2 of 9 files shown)


rfranzke:
/area/logging
/kind/enhancement

charts/seed-bootstrap/templates/gardener-seed-admission-controller/mutatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,42 @@
...
+    service:
+      name: gardener-seed-admission-controller
+      namespace: garden
+      path: /webhooks/mutate-pods
+      port: 443
```
rfranzke: not neeeded

```suggestion
```

charts/seed-bootstrap/templates/gardener-seed-admission-controller/mutatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,42 @@
...
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - pods
+    scope: '*'
+  failurePolicy: Ignore
+  namespaceSelector:
+    matchLabels:
+      gardener.cloud/role: shoot
```
rfranzke:
```suggestion
```

Duplicates of lines 34 ff.

pkg/seedadmission/extension_crds.go
```diff
@@ -58,45 +58,19 @@ func ValidateExtensionDeletion(ctx context.Context, c client.Client, logger *log
...
-	}
+	var operation string
 
+	obj, err := getRequestObject(ctx, c, request)
 	if err != nil {
 		return client.IgnoreNotFound(err)
 	}
 
+	if strings.HasSuffix(obj.GetObjectKind().GroupVersionKind().Kind, "List") {
+		operation = "DELETE OPERATION"
```
rfranzke:
```suggestion
		operation = "DELETECOLLECTION"
```

pkg/seedadmission/extension_crds.go
```diff
@@ -156,3 +130,41 @@ func crdMustBeConsidered(logger *logrus.Entry, labels map[string]string) bool {
 
 	return true
 }
+
+func getRequestObject(ctx context.Context, c client.Client, request *admissionv1beta1.AdmissionRequest) (runtime.Object, error) {
+
```
rfranzke:
```suggestion
```

pkg/seedadmission/extension_crds.go
```diff
@@ -156,3 +130,41 @@ func crdMustBeConsidered(logger *logrus.Entry, labels map[string]string) bool {
 
 	return true
 }
+
+func getRequestObject(ctx context.Context, c client.Client, request *admissionv1beta1.AdmissionRequest) (runtime.Object, error) {
```
rfranzke: Can you move this function into a new, separate file now that it's no longer solely required by the extension CRD handler?

pkg/seedadmission/pods.go
```diff
@@ -0,0 +1,151 @@
...
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//MutatePod mutates pod if mutation conditions are met
```
rfranzke:
```suggestion
// MutatePod mutates pod if mutation conditions are met.
```

pkg/seedadmission/pods.go
```diff
@@ -0,0 +1,151 @@
...
+	patch, err := admitMutatePodAndMutate(ctx, c, entryLogger, obj, request.Namespace)
+	if err != nil {
+		logger.Errorf("Error mutating pod %s/%s: %v", request.Name, request.Namespace, err)
+		return nil, err
+	}
+
+	return patch, nil
+}
+
+// admitMutatePod checks if the object mutation is confirmed and return the pathces needed. If the given object is a list of objects then it
```
rfranzke:
```suggestion
// admitMutatePod checks if the object mutation is allowed and return the patches needed. If the given object is a list of objects then it
```

pkg/seedadmission/pods.go
```diff
@@ -0,0 +1,151 @@
...
+// performs the check for every single object.
+func admitMutatePodAndMutate(ctx context.Context, c client.Client, logger *logrus.Entry, obj runtime.Object, namespace string) ([]jsonpatch.JsonPatchOperation, error) {
```
rfranzke:
```suggestion
func mutatePod(ctx context.Context, c client.Client, logger *logrus.Entry, obj runtime.Object, namespace string) ([]jsonpatch.JsonPatchOperation, error) {
```
?

pkg/seedadmission/pods.go
rfranzke: Can it even happen that the object is a list of objects?

pkg/seedadmission/pods.go
```diff
@@ -0,0 +1,151 @@
...
+					key: value,
+				},
+			})
+		} else {
+			patch = append(patch, jsonpatch.JsonPatchOperation{
+				Operation: "replace",
+				Path:      "/metadata/annotations/" + key,
+				Value:     value,
+			})
+		}
```
rfranzke: Doesn't this overwrite the whole annotations if there is at least one existing annotation?
Shouldn't it rather be:

```suggestion
		if originalAnnotations == nil {
			patch = append(patch, jsonpatch.JsonPatchOperation{
				Operation: "add",
				Path:      "/metadata/annotations",
				Value: map[string]string{
					key: value,
				},
			})
		} else if originalAnnotations[key] == "" {
		  patch = append(patch, jsonpatch.JsonPatchOperation{
				Operation: "add",
				Path:      "/metadata/annotations/" + key,
				Value: map[string]string{
					key: value,
				},
			})
		} else {
			patch = append(patch, jsonpatch.JsonPatchOperation{
				Operation: "replace",
				Path:      "/metadata/annotations/" + key,
				Value:     value,
			})
		}
```

pkg/seedadmission/pods_test.go
```diff
@@ -0,0 +1,510 @@
...
+
+		var (
+			resource    = metav1.GroupVersionResource{Group: corev1.SchemeGroupVersion.Group, Version: corev1.SchemeGroupVersion.Version, Resource: "pods"}
+			fooResource = metav1.GroupVersionResource{Group: "foo", Version: "bar", Resource: "baz"}
+			key         = "fluentbit.io/exclude"
+		)
+
+		testingPurpuse := core.ShootPurpose("testing")
+		developmentPurpuse := core.ShootPurpose("development")
+		//Hibernation := core.Hibernation{Enabled: pointer.BoolPtr(true)}
```
rfranzke: If it's not needed then remove it

pkg/seedadmission/pods_test.go
```diff
@@ -0,0 +1,510 @@
...
+		NotHibernation := core.Hibernation{Enabled: pointer.BoolPtr(false)}
```
rfranzke: These variables should start with lower case

rfranzke (changes requested):
Very nice PR again, thank you!

pkg/seedadmission/pods.go
vlvasilev: It doesn't overwrite the existing annotations. 
According to the documentation [link](http://jsonpatch.com/#add) `add` operation insert an object into an array. During the test all of the annotation persisted after mutation.

rfranzke: OK, sure, makes sense. Thank you!

pkg/seedadmission/pods.go
vlvasilev: I don't know. I saw it from the extensions_crds.go and just copied the logic.

rfranzke: Probably it cannot happen. I think that's special with `DELETE` operation?! cc @timuthy @mvladev @vpnachev @ialidzhikov

timuthy: I'm not aware of it either. Also `kubectl` for example transforms a `kubectl create -f list-of-shoots.yaml` to single `POST` requests.

vlvasilev: @rfranzke  Should I remove the List processing option?

rfranzke: I'd remove it as it's code that can not be reached :) thanks!

rfranzke:
Can you fix the failing CI build, please?

charts/seed-bootstrap/templates/gardener-seed-admission-controller/mutatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,37 @@
...
+      path: /webhooks/mutate-pods
+  failurePolicy: Ignore
+  matchPolicy: Exact
+  namespaceSelector:
+    matchLabels:
+      gardener.cloud/role: shoot
+  objectSelector: {}
+  reinvocationPolicy: Never
+  sideEffects: None
+  timeoutSeconds: 30
```
vpnachev:
```suggestion
  timeoutSeconds: 10
```

Keep the timeout low. 30 seconds is also the client time out and usually the client interrupt the connection before the API. The result is that every request matched by the webhook fails, even though the failure policy is `ignore`.

vlvasilev: Done

pkg/seedadmission/pods.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, err
+	}
+
+	acc, err := meta.Accessor(pod)
+	if err != nil {
+		return nil, err
+	}
+
+	if !isMutationRequired {
+		logger.Infof("Skipping mutation for %s/%s due to policy check", acc.GetNamespace(), acc.GetName())
```
rfranzke: Should we move this to debug log level? I guess a lot of pods are created in the seed for which the mutating is not required. This will flood the logs unnecessarily. WDYT?

pkg/seedadmission/pods.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, nil
+	}
+
+	originalAnnotations := acc.GetAnnotations()
+	annotationsToAdd := map[string]string{
+		"fluentbit.io/exclude": "true",
```
rfranzke, line 88:
```suggestion
		"fluentbit.io~1exclude": "true",
```

rfranzke (changes requested):

pkg/seedadmission/pods.go
vlvasilev: What is the use of `~1` 
This change makes all of the CREAT request invalid.
After deleting all of the pod in the shoot control plane namespace no other pods are created.

charts/seed-bootstrap/templates/gardener-seed-admission-controller/mutatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: admissionregistration.k8s.io/v1
```
rfranzke:
```suggestion
apiVersion: admissionregistration.k8s.io/v1beta1
```

Can you please use the `v1beta1` version to also serve seed clusters with a lower Kubernetes version? I'm seeing

```
time="2020-05-22T07:24:42+02:00" level=error msg="Seed bootstrapping failed: failed to apply manifests: 1 error occurred: could not apply object of kind \"MutatingWebhookConfiguration\" \"default/gardener-seed-admission-controller\": no matches for kind \"MutatingWebhookConfiguration\" in version \"admissionregistration.k8s.io/v1\"" seed=aws
```

pkg/seedadmission/pods.go
```diff
@@ -0,0 +1,132 @@
...
+				Operation: "add",
+				Path:      "/metadata/annotations",
+				Value: map[string]string{
+					key: value,
+				},
+			})
+		} else {
+			patch = append(patch, jsonpatch.JsonPatchOperation{
+				Operation: "replace",
+				Path:      "/metadata/annotations/" + key,
```
rfranzke:
```suggestion
				Path:      "/metadata/annotations/" + strings.ReplaceAll(key, "/", "~1"),
```

@vlvasilev wrt to https://github.com/gardener/gardener/pull/2320/files#r428491172:

`~1` is used to escape `/` in a JSON patch. Without it this `replace` operation will fail:

```
error: pods "cloud-controller-manager-b5ccb7948-48vps" could not be patched: Internal error occurred: Internal error occurred: jsonpatch replace operation does not apply: doc is missing path: /metadata/annotations/fluentbit.io/exclude
```

You're right, it's not needed for the above `add` operation, but during replace you have to escape `/`. I tested and verified it locally.

rfranzke (changes requested):

pkg/seedadmission/pods.go
vlvasilev: Thanks for the explanation.

rfranzke (dismissed):
Thanks for addressing the comments! Again, a very nice PR :)
/lgtm

rfranzke (approved):

PR #2324, opened 2020-05-13
Fix retry cycle time for shoots

timuthy:
This PR fixes an issue which reset the `.status.retryCycleStartTime` in several cases too often and thus prevented erroneous shoot clusters from transitioning to a `Failed` state.

The feature `confineSpecUpdateRollout` (https://github.com/gardener/gardener/pull/2233) for shoots in a failed state is now respected as well. Normally a `.spec` update resets the failed status and with `confineSpecUpdateRollout` this should only happen in the maintenance time window.

Additionally, @rfranzke and I suggest to decrease the default value for the Gardenlet setting `retryDuration` from 24h to 12h with the assumption that shoots failing for ~12h most often need an intervention by the shoot owner, i.e. it isn't reasonable to retry for another 12h w/o any spec update.

**Release note**:

```improvement operator
An issue has been fixed with prevented failed shoots from being excluded from reconciliation after the [retryDuration](https://github.com/gardener/gardener/blob/a46946fdce6bf75e0ce18272e7762f5765cced39/example/20-componentconfig-gardenlet.yaml#L33) is exceeded.
```
```noteworthy operator
The default value for [retryDuration](https://github.com/gardener/gardener/blob/a46946fdce6bf75e0ce18272e7762f5765cced39/example/20-componentconfig-gardenlet.yaml#L33) has been changed from 24h to 12h. Hence, Gardenlet tries to reconcile shoots with erroneous operations for 12 hours (by default). After this period of time only the `retry` operation, a `.spec` change, or a rollout of a new Gardenlet version re-triggers a reconciliation.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -432,8 +432,7 @@ func (c *Controller) updateShootStatusReconcileStart(o *operation.Operation, ope
 
 	if o.Shoot.Info.Status.RetryCycleStartTime == nil ||
 		o.Shoot.Info.Generation != o.Shoot.Info.Status.ObservedGeneration ||
-		o.Shoot.Info.Status.Gardener.Version == version.Get().GitVersion ||
-		(o.Shoot.Info.Status.LastOperation != nil && o.Shoot.Info.Status.LastOperation.State == gardencorev1beta1.LastOperationStateFailed) {
+		o.Shoot.Info.Status.Gardener.Version != version.Get().GitVersion {
 
 		now := metav1.NewTime(time.Now().UTC())
 		retryCycleStartTime = &now
--- a/plugin/pkg/shoot/validator/admission.go
+++ b/plugin/pkg/shoot/validator/admission.go
@@ -31,6 +31,7 @@ import (
 	corelisters "github.com/gardener/gardener/pkg/client/core/listers/core/internalversion"
 	"github.com/gardener/gardener/pkg/controllerutils"
 	"github.com/gardener/gardener/pkg/operation/common"
+	"github.com/gardener/gardener/pkg/utils"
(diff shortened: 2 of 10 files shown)


rfranzke:
/area/robustness
/kind/bug
/needs/review
/size/xs

plugin/pkg/shoot/validator/admission.go
```diff
@@ -319,6 +320,12 @@ func (v *ValidateShoot) Admit(ctx context.Context, a admission.Attributes, o adm
 		controllerutils.AddTasks(shoot.ObjectMeta.Annotations, common.ShootTaskDeployInfrastructure)
 	}
 
+	if shoot.Spec.Maintenance != nil && utils.IsTrue(shoot.Spec.Maintenance.ConfineSpecUpdateRollout) &&
```
rfranzke, line 323: Can you add a test case for this?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -430,10 +431,13 @@ func (c *Controller) updateShootStatusReconcile(o *operation.Operation, operatio
 func (c *Controller) updateShootStatusReconcileStart(o *operation.Operation, operationType gardencorev1beta1.LastOperationType) error {
 	var retryCycleStartTime *metav1.Time
 
+	fmt.Printf("RetryCycleStartTime %v, Generation %v, ObservedGeneration %v, Gardener status version %v, Gardener version %v, Last Operation %v\n",
```
rfranzke: Left over, can you remove it?

rfranzke (changes requested):

timebertt (commented):
/lgtm apart from @rfranzke's comments

rfranzke:
/area/robustness
/kind/bug
/needs/review
/size/xs

rfranzke (approved):
/lgtm
/needs/second-opinion

timebertt (approved):
/lgtm

PR #2330, opened 2020-05-14
Gardenlet - fix Seed enqueue when Bootstrap status changed

danielfoehrKn:

Fixes the Seed enqueue in the federated seed controller for Seeds that, when the Gardenlet starts, are not bootstrapped successfully. The Seed's condition looks like this:

```
 "conditions": [
            {
                "type": "Bootstrapped",
                "status": "False",
            }
```

The problem is that the federated seed controller's check if the Seed's status changed to True is flawed.

```
 "conditions": [
            {
                "type": "Bootstrapped",
                "status": "True",
            }
```
In fact that means that no federated seed controllers are started for the Seed --> no ShootStateSync or ControllerInstallation checks are being performed.

Fixes #

**Release note**:

```improvement operator
Fix the Seed enqueue in the federated seed controller for initially not-bootstrapped Seeds. Now Seeds are properly enqueued when their status changes from `not-bootstrapped` to `successfully bootstrapped`.
```

### DIFF
--- a/pkg/gardenlet/controller/federatedseed/seed_control.go
+++ b/pkg/gardenlet/controller/federatedseed/seed_control.go
@@ -105,23 +105,14 @@ func (c *Controller) addSeedToQueue(obj interface{}) {
 	c.seedQueue.Add(key)
 }
 
-func (c *Controller) seedUpdate(oldObj, newObj interface{}) {
-	var (
-		newSeedObj = newObj.(*gardencorev1beta1.Seed)
-		oldSeedObj = oldObj.(*gardencorev1beta1.Seed)
-	)
+func (c *Controller) seedUpdate(_, newObj interface{}) {
+	var newSeedObj = newObj.(*gardencorev1beta1.Seed)
 
 	newSeedBootstrappedCondition := helper.GetCondition(newSeedObj.Status.Conditions, gardencorev1beta1.SeedBootstrapped)
-	oldSeedBootstrappedCondition := helper.GetCondition(oldSeedObj.Status.Conditions, gardencorev1beta1.SeedBootstrapped)
-
-	if seedChangedToBootstrapped(newSeedBootstrappedCondition, oldSeedBootstrappedCondition) {
+	_, found := c.federatedSeedControllerManager.controllers[newSeedObj.Name]
+	if !found && newSeedBootstrappedCondition != nil && newSeedBootstrappedCondition.Status == gardencorev1beta1.ConditionTrue {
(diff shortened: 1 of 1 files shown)


timebertt (approved):
/lgtm

timebertt:
/kind/bug

timebertt:
/topology/seed

pkg/gardenlet/controller/federatedseed/seed_control.go
```diff
@@ -105,25 +105,16 @@ func (c *Controller) addSeedToQueue(obj interface{}) {
...
-	)
+func (c *Controller) seedUpdate(_, newObj interface{}) {
+	var newSeedObj = newObj.(*gardencorev1beta1.Seed)
 
 	newSeedBootstrappedCondition := helper.GetCondition(newSeedObj.Status.Conditions, gardencorev1beta1.SeedBootstrapped)
-	oldSeedBootstrappedCondition := helper.GetCondition(oldSeedObj.Status.Conditions, gardencorev1beta1.SeedBootstrapped)
-
-	if seedChangedToBootstrapped(newSeedBootstrappedCondition, oldSeedBootstrappedCondition) {
+	_, found := c.federatedSeedControllerManager.controllers[newSeedObj.Name]
+	if !found && newSeedBootstrappedCondition != nil && newSeedBootstrappedCondition.Status == gardencorev1beta1.ConditionTrue {
```
rfranzke, line 113: Hm, why didn't you just fix the bug in the old code? I think it makes sense to only enqueue on update if the condition switches to `True`, or?

Also, the check whether the controller is already started is now duplicated here (the reconciler is already checking it).

/cc @swilen-iwanow @plkokanov @vlvasilev

danielfoehrKn: I was thinking that what if we miss an update event? The controller would not be started any more. This way we could even miss the one particular update and would still start the controller at a later point in time. I just find it uncomfortable to rely only on one update event instead just checking if the desired state (having a controller) is fulfilled or not.

Yes the check is duplicated, but it is not an expensive one (lookup key in a map) vs [executing this function](https://github.com/gardener/gardener/blob/46a88c15db5397c573eba6f90f7cdba03327f361/pkg/gardenlet/controller/federatedseed/seed_control.go#L183-L182) and then performing the check.

rfranzke: OK, agreed. Makes sense, thanks!

rfranzke (approved):

PR #2331, opened 2020-05-14
Reduce the amount federated seed workers

danielfoehrKn:

The current default settings for the ShootState controller and the ControllerInstallationRequired Controller in the federated seed controller lead to ca. 280 workers being created.
By reducing the default settings and fixing a bug during worker creation, the amount of workers created is reduced to the amount of extensions watched per Seed - around 20 per Seed.

Also do not update the ShootState if the Extension State is null. 

Fixes #

**Release note**:

```improvement operator
Adjust default concurrent-sync settings for the Gardenlet controllers `ShootState` and `ControllerInstallationRequired`  causing too many goroutines to be created. Also fixed a bug during worker creation to further reduce the amount of workers being created. 
```

### DIFF
--- a/example/20-componentconfig-gardenlet.yaml
+++ b/example/20-componentconfig-gardenlet.yaml
@@ -28,7 +28,7 @@ controllers:
     concurrentSyncs: 20
     syncPeriod: 30s
   controllerInstallationRequired:
-    concurrentSyncs: 20
+    concurrentSyncs: 1
   shoot:
     concurrentSyncs: 20
     syncPeriod: 1h
@@ -53,7 +53,7 @@ controllers:
     - type: EveryNodeReady
       duration: 5m
   shootStateSync:
-    concurrentSyncs: 5
+    concurrentSyncs: 1
     syncPeriod: 30s
   seed:
     concurrentSyncs: 5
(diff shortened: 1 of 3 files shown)


rfranzke (dismissed):
/lgtm
/cc @plkokanov @swilen-iwanow @vlvasilev 
/needs second-opinion

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
```diff
@@ -83,6 +83,11 @@ func (s *shootStateControl) createShootStateSyncReconcileFunc(ctx context.Contex
 		purpose := extensionObject.GetExtensionSpec().GetExtensionPurpose()
 		state := extensionObject.GetExtensionStatus().GetState()
 
+		if state == nil {
+			s.log.Debugf("Skip Shoot's %s %s extension state with name %s - no state found.", shoot.Name, kind, name)
+			return reconcile.Result{}, nil
+		}
```
swilen-iwanow: I am a bit worried about this. As far as I remember I've also added such logic, but then we decided to remove the state from the ShootState if it set to nil in the Extension's state. @plkokanov, wdyt?

plkokanov: I guess we could change the logic a bit to work like this:
if the extension state in the shoot state was not nil, but in the current reconciliation it is nil, we should update the ShootState and delete the extension state from it [check this line](https://github.com/gardener/gardener/pull/2331/files/f304afd50eb3d162a025fac78f5f33bc1fd64b1c#diff-5fbd8d379e4dd574afdac838e68337f2R110). Otherwise we do not update if the state in the current reconciliation is nil.

Basically if we could add a check before calling CreateOrUpdate would be great imo.

rfranzke: Hm, this sounds unrelated to this PR. So probably @danielfoehrKn should revert this change for now and either he or you can open a new one?

danielfoehrKn: I would appreciate if @swilen-iwanow or @plkokanov could take over this issue / give it another thought as you are the experts on the topic. 
I ll revert the change.

rfranzke (approved):
/lgtm

swilen-iwanow (approved):
lgtm

PR #2012, opened 2020-03-06
Introduce flow for preparing Shoot's Control Plane for migration

swilen-iwanow:
**What this PR does / why we need it**:
Adds additional flow to the Shoot's lifecycle which is going to prepare the Shoot's Control plane for migration, and adapts the reconciliation flow to restore Shoot's extension resources from the ShootState.
**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
1. The API Server still doesn't permit changes to the Seed name. We will change this when everything around the CPM is merged.
2. The adaptation of the reconciliation flow is partial, because secrets are done in another PR. This one is only for the Extension Resources

**Release note**:

```improvement operator
NONE
```

PR #2333, opened 2020-05-14
Set shoot to `Failed` if non-retryable error codes are detected

rfranzke:
The shoot controller will set the shoot status to `Failed` if it detects an "unauthorized" or "configuration problem" error. It doesn't make sense to auto-retry in these cases because the user needs to change something (either the secret or some configuration, e.g. in the shoot spec).

On the way, we have refactored the functions and removed the duplication for shoot status updates.

**Release note**:

```noteworthy user
Shoot clusters which are flagged with `ERR_INFRA_UNAUTHORIZED` or `ERR_CONFIGURATION_PROBLEM` error codes are now automatically set to `Failed` status. This means that they won't be retried automatically unless you annotate the `Shoot` with [`gardener.cloud/operation=retry`](https://github.com/gardener/gardener/blob/master/docs/usage/shoot_operations.md#retry-failed-operation). All other error codes will lead to automatic retries for at most `12h` before the shoot is set to `Failed`.
```

/area robustness
/area quality
/area scalability
/kind enhancement
/priority normal
/exp intermediate
/needs review
/topology garden

### DIFF
--- a/pkg/apis/core/v1beta1/helper/errors.go
+++ b/pkg/apis/core/v1beta1/helper/errors.go
@@ -190,3 +190,16 @@ func LastErrorWithTaskID(description string, taskID string, codes ...gardencorev
 		},
 	}
 }
+
+// HasNonRetryableErrorCode returns true if at least one of given list of last errors has at least one error code that
+// indicates that an automatic retry would not help fixing the problem.
+func HasNonRetryableErrorCode(lastErrors ...gardencorev1beta1.LastError) bool {
+	for _, lastError := range lastErrors {
+		for _, code := range lastError.Codes {
+			if code == gardencorev1beta1.ErrorInfraUnauthorized || code == gardencorev1beta1.ErrorConfigurationProblem {
+				return true
+			}
+		}
+	}
+	return false
+}
--- a/pkg/apis/core/v1beta1/helper/errors_test.go
+++ b/pkg/apis/core/v1beta1/helper/errors_test.go
@@ -27,43 +27,60 @@ import (
 	. "github.com/onsi/gomega/types"
 )
 
-var _ = Describe("helper", func() {
-	Describe("errors", func() {
-		Describe("#DetermineError", func() {
-			DescribeTable("appropriate error should be determined",
-				func(err error, msg string, expectedErr error) {
-					Expect(DetermineError(err, msg)).To(Equal(expectedErr))
-				},
+var _ = Describe("errors", func() {
+	DescribeTable("#DetermineError",
+		func(err error, msg string, expectedErr error) {
+			Expect(DetermineError(err, msg)).To(Equal(expectedErr))
+		},
 
-				Entry("no error", nil, "foo", errors.New("foo")),
-				Entry("no code to extract", errors.New("foo"), "", errors.New("foo")),
-				Entry("unauthorized", errors.New("unauthorized"), "", NewErrorWithCodes("unauthorized", gardencorev1beta1.ErrorInfraUnauthorized)),
-				Entry("unauthorized with coder", NewErrorWithCodes("", gardencorev1beta1.ErrorInfraUnauthorized), "", NewErrorWithCodes("", gardencorev1beta1.ErrorInfraUnauthorized)),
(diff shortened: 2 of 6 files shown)


pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -557,6 +520,141 @@ func (c *Controller) durationUntilNextShootSync(shoot *gardencorev1beta1.Shoot)
...
+		return reconcile.Result{}, utilerrors.WithSuppressed(err, updateErr)
+	}
+
+	return reconcile.Result{
+		Requeue:      true,
+		RequeueAfter: 15 * time.Second, // prevent ddos-ing the seed
+	}, nil
+}
+
+func (c *Controller) updateShootStatusOperationStart(shoot *gardencorev1beta1.Shoot, shootNamespace string) (*gardencorev1beta1.Shoot, error) {
```
rfranzke, line 543: FYI @swilen-iwanow, you might want to use these functions as part of #2012 as well (instead of duplicating (parts of) them again)

pkg/apis/core/v1beta1/helper/errors.go
```diff
@@ -190,3 +190,16 @@ func LastErrorWithTaskID(description string, taskID string, codes ...gardencorev
 		},
 	}
 }
+
+// HasNotRetryableErrorCode returns true if at least one of given list of last errors has at least one error code that
+// indicates that an automatic retry would not help fixing the problem.
+func HasNotRetryableErrorCode(lastErrors ...gardencorev1beta1.LastError) bool {
```
timebertt: Just a naming suggestion here, no strong opinion on that one.
```suggestion
func HasNonRetryableErrorCode(lastErrors ...gardencorev1beta1.LastError) bool {
```

pkg/apis/core/v1beta1/helper/errors_test.go
```diff
@@ -27,43 +27,60 @@ import (
...
+		errorWithoutCodes                = gardencorev1beta1.LastError{}
+	)
+
+	DescribeTable("#HasNotRetryableErrorCode",
+		func(lastErrors []gardencorev1beta1.LastError, matcher GomegaMatcher) {
+			Expect(HasNotRetryableErrorCode(lastErrors...)).To(matcher)
+		},
+
+		Entry("no error given", nil, BeFalse()),
+		Entry("only errors with retryable error codes", []gardencorev1beta1.LastError{unauthorizedError, configurationProblemError}, BeTrue()),
```
timebertt:
```suggestion
		Entry("only errors with non-retryable error codes", []gardencorev1beta1.LastError{unauthorizedError, configurationProblemError}, BeTrue()),
```

pkg/apis/core/v1beta1/helper/errors_test.go
```diff
@@ -27,43 +27,60 @@ import (
...
+		Entry("only errors with not retryable error codes", []gardencorev1beta1.LastError{infraInsufficientPrivilegesError, infraQuotaExceededError, infraDependenciesError, infraResourcesDepletedError, cleanupClusterResourcesError}, BeFalse()),
```
timebertt:
```suggestion
		Entry("only errors with retryable error codes", []gardencorev1beta1.LastError{infraInsufficientPrivilegesError, infraQuotaExceededError, infraDependenciesError, infraResourcesDepletedError, cleanupClusterResourcesError}, BeFalse()),
```

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -557,6 +520,141 @@ func (c *Controller) durationUntilNextShootSync(shoot *gardencorev1beta1.Shoot)
...
+			shoot.Status.LastOperation.LastUpdateTime = metav1.NewTime(time.Now().UTC())
+			return shoot, nil
+		},
+	)
+	if updateErr != nil {
+		return reconcile.Result{}, utilerrors.WithSuppressed(err, updateErr)
+	}
+
+	return reconcile.Result{
+		Requeue:      true,
```
timebertt: Obsolete with `RequeueAfter`.
```suggestion
```

timebertt (commented):
/lgtm
Only some minor nits.

timebertt (approved):

PR #2336, opened 2020-05-15
Exit shoot flow when observing extension error

timuthy:
This PR makes Gardener to exit the shoot reconciliation flow as soon as a step notices an error in one of resources reconciled in the seed cluster (e.g. infrastructure, etcd). Now, errors are reported quicker and not only after certain timeout has been exceeded.

**Release note**:

```improvement operator
The error reporting for shoot clusters has been improved. Configuration problems or similar issues which occur during shoot reconciliation are now instantly visible in the shoot resource.
```

/area usability
/kind enhancement
/priority normal

### DIFF
--- a/pkg/utils/kubernetes/health/health.go
+++ b/pkg/utils/kubernetes/health/health.go
@@ -19,18 +19,18 @@ import (
 	"net/http"
 	"time"
 
-	"k8s.io/apimachinery/pkg/runtime"
-
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"github.com/gardener/gardener/pkg/utils"
+	"github.com/gardener/gardener/pkg/utils/retry"
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/rest"
 )
@@ -265,10 +265,10 @@ func CheckSeed(seed *gardencorev1beta1.Seed, identity *gardencorev1beta1.Gardene
 }
 
 // CheckExtensionObject checks if an extension Object is healthy or not.
-func CheckExtensionObject(o runtime.Object) error {
+func CheckExtensionObject(o runtime.Object) (bool, error) {
 	obj, ok := o.(extensionsv1alpha1.Object)
 	if !ok {
-		return fmt.Errorf("expected extensionsv1alpha1.Object but got %T", o)
+		return retry.SevereError(fmt.Errorf("expected extensionsv1alpha1.Object but got %T", o))
 	}
 
 	status := obj.GetExtensionStatus()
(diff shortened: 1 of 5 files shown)


pkg/utils/kubernetes/health/health.go
```diff
@@ -19,6 +19,8 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/gardener/gardener/pkg/utils/retry"
```
timebertt: Can you sort the imports, please?

Site note:
@gardener/gardener-maintainers We should probably add some extended import formatting tool to `make format`, so we can stop manually sorting the imports according to golang code style guidelines.
AFAIK @danielfoehrKn has a tool which could do the job: https://github.com/danielfoehrKn/importsort
Can we use that, WDYT?

timebertt (commented):
Thank you!
/lgtm
Only one nit.

pkg/utils/kubernetes/health/health.go
rfranzke: Hm, yeah, we should have a tool + verification for it. However, probably it makes sense to stick to already existing (and well-adopted) tools like `goimports` instead of creating/maintaining our own? I'm open to adapt our conventions so that `goimports` can be used out-of-the-box. WDYT?

timebertt: Sure, I would also prefer to stick to `goimports` for example. But AFAIK, it doesn't offer any possibility to properly sort imports as we do (e.g. third-party block last, all first-party/third-party imports in same block and so on), right?
Do you propose to accept any arbitrary sorting accepted by `goimports` now?

timebertt (dismissed):

pkg/utils/kubernetes/health/health.go
rfranzke: Yes, that's why @danielfoehrKn created his own tool. I think `goimports` doesn't do it arbitrarily but adds the third-party imports before the repository-local imports. So, yes, I'm open to change this convention to get rid of the many "please sort the imports" review comments. :)

pkg/operation/botanist/dns.go
```diff
@@ -220,9 +219,13 @@ func (b *Botanist) waitUntilDNSProviderReady(ctx context.Context, name string) e
...
 			message = *msg
 		}
+		providerErr := fmt.Errorf("DNS provider %q is not ready (status=%s, message=%s)", name, status, message)
 
 		b.Logger.Infof("Waiting for %q DNS provider to be ready... (status=%s, message=%s)", name, status, message)
-		return retry.MinorError(fmt.Errorf("DNS provider %q is not ready (status=%s, message=%s)", name, status, message))
+		if status == dnsv1alpha1.STATE_ERROR || status == dnsv1alpha1.STATE_INVALID {
+			return retry.SevereError(providerErr)
+		}
+		return retry.MinorError(providerErr)
```
rfranzke: We missed adapting the `WaitForDNSEntryReady` function accordingly, can you change as well similar to this change?
https://github.com/gardener/gardener/pull/2336/files#diff-33d2a8cacfcd2dda84699a11862f3ab0L274-L282

rfranzke (changes requested):

pkg/operation/botanist/dns.go
timuthy: Nice catch!

pkg/utils/kubernetes/health/health.go
timebertt: Yeah, probably we should just agree on using `goimports` from now on and stop the "please sort the imports" review comments as long as it is accepted by `goimports`.

danielfoehrKn: Yep agree. Though I just automated the tool usage by adding it to my local pre-commit hook  ;(

danielfoehrKn (dismissed):

timebertt (dismissed):

rfranzke (dismissed):
/lgtm

rfranzke (approved):
/lgtm
/remove needs rebase
/remove status author-action

ghost:
@rfranzke Command "/remove needs rebase" failed with "Label does not exist".
<details>
<summary>Additional Information</summary>

```
Redacted in public. Check backend logs.
```
</details>

rfranzke:
@vlerenc FYI: https://github.com/gardener/gardener/pull/2336#issuecomment-631450535

vlerenc:
Hmm... @rfranzke, when you submit a review, multiple events reach the bot, the first one succeeds, the next one doesn't (label is no longer set). OK, bad, not sure now how to handle that. Add is idempotent, remove is not in GitHub. Adding an added label does nothing, but removing a removed label fails. I will look into it, but not right away. Maybe I can deal with it, when I try to bundle the actions. I hope you can live with it. What I can do easily is silence the problem/not comment on error. ;-)

rfranzke:
😄 I can certainly live with it. I forgot that `/lgtm` will remove these labels automatically, so I was thinking it's another problem. Thanks for looking into it.

PR #2266, opened 2020-04-30
Configurable Gardenlet Stale health check verification

danielfoehrKn:
Gardenlet does not check for stale extension health check reports per default.

**What this PR does / why we need it**:

The Gardenlet introduced checks for stale extension reports with PR #2215 .
However the Gardener extensions are using the health check libraries that can produce outdated-health check reports. This should be fixed with this [PR](https://github.com/gardener/gardener/pull/2249).
To not couple the Gardener version with the extension versions, for now the Gardener does NOT check for outdated health checks per default. 

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement operator
For backwards-compatibility reasons, the gardenlet does not  check for stale extension reports per default. To enable, the field controllers.shootCare.staleExtensionHealthCheckThreshold  in the Gardenlet configuration file (https://github.com/gardener/gardener/blob/master/example/20-componentconfig-gardenlet.yaml)can be set.
```

PR #2337, opened 2020-05-15
Remove staleExtensionHealthCheckThreshold from gardenlet default values.yaml

ialidzhikov:
As of #2266 the gardenlet check for stale extension healthchecks needs to be disabled by default for several releases to ensure backwards compatibility.
But currently `staleExtensionHealthCheckThreshold` is part of gardenlet default values.yaml - if you need to disable the check, you have to explicitly set the value to `null` in the values file passed to helm install or helm upgrade.

Fixes #

**Release note**:

```improvement operator
The check for stale extension healthchecks is now also disabled by default in gardenlet chart values.yaml.
```

/area quality
/kind enhancement
/priority normal

### DIFF
--- a/charts/gardener/gardenlet/values.yaml
+++ b/charts/gardener/gardenlet/values.yaml
@@ -82,7 +82,6 @@ global:
         shootCare:
           concurrentSyncs: 5
           syncPeriod: 30s
-          staleExtensionHealthCheckThreshold: 5m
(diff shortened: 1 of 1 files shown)


danielfoehrKn (approved):
Thanks.

rfranzke (approved):

PR #2340, opened 2020-05-18
Add seedSelector for shoot scheduling decisions

### LINKED ISSUE #2306: `.spec.seedSelector` for `Shoot` resources
Give users the ability to select access restrictions (for gardener operators) for their shoot, depending on the selected region.
E.g. on the UI it could look like the following (texts are just an example and for better understanding of the requested feature and will be configurable in the Dashboard)
Screenshot 2020-05-08 at 15 35 43
https://github.com/gardener/dashboard/issues/689

**UPDATE**: see https://github.com/gardener/gardener/issues/2306#issuecomment-626781778 for latest proposal

~~The selected options should be added to the shoot spec as following:~~
```yaml
kind: Shoot
metadata:
  namespace: garden-bar
  name: foo
spec:
  accessRestrictions:
    eu-access-standard:
    - eu-access-24-7-node
    foo: []
```

~~In addition, the `region` of the `CloudProfile` has to be extended with an `accessRestrictions` field.~~
- ~~This way we can validate the shoot resource and only allow the change if the region has a corresponding accessRestrictions field in the CloudProfile~~
- ~~On the UI the checkboxes would only appear only for the respective regions.~~

```yaml
kind: CloudProfile
name: aws
regions:
  - name: ca-central-1
    zones:
    - name: ca-central-1a
    - name: ca-central-1b
  - name: eu-central-1
    accessRestrictions:
      - eu-access-standard
      - foo
    zones:
    - name: eu-central-1a
    - name: eu-central-1b
    - name: eu-central-1c
  - name: eu-north-1
    accessRestrictions:
      - eu-access-standard
    zones:
    - name: eu-north-1a
    - name: eu-north-1b
    - name: eu-north-1c
```

~~As described above the shoot validation needs to be extended so that in case an `accessRestriction` is set, it ensures that the selected region has a corresponding accessRestrictions field in the corresponding `CloudProfile`~~

**Why is this needed**:
In order to fulfil GDPR requirements, shoot owners should have the possibility to restrict access to their shoots in order to protect personal data.

rfranzke:
This PR adds a `.spec.seedSelector` field for `Shoot`s (similar to the `.spec.nodeSelector` for `Pod`s).
Also, it allows to configure arbitrary labels per region in the `CloudProfile`'s `.spec.regions[]` list.

Fixes #2306

**Release note**:

```improvement operator
It is now possible to add arbitrary labels to the `.spec.regions[].labels` field in the `CloudProfile`, e.g. to provide more information about reliability, access restrictions, etc.
```
```improvement user
The `Shoot` specification does now have a new `.spec.seedSelector` field which allows to provide label selector. Only seeds whose labels match will be considered for scheduling decisions.
```

/area usability
/kind api-change
/kind enhancement
/priority normal


docs/concepts/scheduler.md
```diff
@@ -69,6 +69,13 @@ It filters out `Seed`s
 
 After this filtering process the least utilized seed, i.e., the one with the least number of shoot control planes, will be the winner and written to the `.spec.seedName` field of the `Shoot`.
 
+## `seedSelector` field in the `Shoot` specification
+
+Similar to the `.spec.nodeSelector` field in `Pod`s, the `Shoot` specification has an optional `.spec.seedSelector` field.
+It allows the user to provide a label selector that must match the labels of `Seed`s in order to be scheduled to one of them.
+The labels on `Seed`s are usually controlled by Gardener administrators/operators - end users cannot add arbitrary labels themselves.
+If provided, the Gardener scheduler will only consider those seeds as "suitable" whose labels match those provided in the `.psec.seedSelector` of the `Shoot`.
```
timebertt:
```suggestion
If provided, the Gardener scheduler will only consider those seeds as "suitable" whose labels match those provided in the `.spec.seedSelector` of the `Shoot`.
```

docs/concepts/scheduler.md
```diff
@@ -69,6 +69,13 @@ It filters out `Seed`s
 
 After this filtering process the least utilized seed, i.e., the one with the least number of shoot control planes, will be the winner and written to the `.spec.seedName` field of the `Shoot`.
 
+## `seedSelector` field in the `Shoot` specification
```
timebertt, line 72: Can you update Line 51?
`Subsequently the Gardener Controller Manager picks up and starts to create the cluster on the specified seed.` 
`Gardener Controller Manager` -> `Gardenlet`

timebertt: Also there is a missing closing bracket in line 55.

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -175,12 +175,16 @@ func determineSeed(shoot *gardencorev1beta1.Shoot, seedLister gardencorelisters.
 		return nil, err
 	}
 
-	filteredSeeds, err := filterSeedsMatchingSeedSelector(cloudProfile, seedList)
+	seedsMatchingCloudProfileSelector, err := filterSeedsMatchingSeedSelector(seedList, cloudProfile.Spec.SeedSelector, "CloudProfile")
+	if err != nil {
+		return nil, err
+	}
+	seedsMatchingShootSelector, err := filterSeedsMatchingSeedSelector(seedsMatchingCloudProfileSelector, shoot.Spec.SeedSelector, "Shoot")
```
timebertt, line 182: Hmm, this will yield an unhelpful message, if there are Seeds that are matching the `seedSelector` of the Shoot, but not the one of the CloudProfile.
I.e. it will say "none out of the <n> seeds has the matching labels required by seed selector of 'Shoot'" although, all Seeds might actually match the Shoot's selector.

Can you evaluate both selectors individually and then combine the filtered set of Seeds here to be able to give a more helpful error message?

timebertt (commented):
Very nice:)
Only minor comments from my side.

pkg/scheduler/controller/shoot/scheduler_control.go
rfranzke: I haven't really understood what you are suggesting, can you explain again? The selectors are already evaluated individually. Also, I'm not too sure whether that's really worth the "effort" because there are other criteria where the seed selector labels may match but other preconditions don't. For example, I can have both an AWS and GCP seed with the same `foo=bar` labels, but when creating a GCP shoot with `foo=bar` seed selector then the AWS seed is not even considered. Isn't the `CloudProfile` filtering of the same quality?

timebertt (approved):

PR #2341, opened 2020-05-18
Add developer documentation about API changes

rfranzke:
This PR adds a checklist containing the steps that should be considered when changing the API of Gardener or the API of a component configuration.

**Release note**:

```improvement operator
NONE
```

/area dev-productivity
/area documentation
/area open-source
/kind enhancement
/priority normal

### DIFF
--- a/docs/development/changing-the-api.md
+++ b/docs/development/changing-the-api.md
@@ -1,0 +1,39 @@
+# Extending the API
+
+This document describes the steps that need to be performed when changing the API.
+It provides guidance for API changes to both (Gardener system in general or component configurations).
+
+Generally, as Gardener is a Kubernetes-native extension, it follows the same API conventions and guidelines like Kubernetes itself.
+[This document](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md) as well as [this document](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md) already provide a good overview and general explanation of the basic concepts behind it.
+We are following the same approaches.
+
+## Gardener API
+
+The Gardener API is defined in `pkg/apis/{core,extensions,settings}` directories and is the main point of interaction with the system.
+It must be ensured that the API is always backwards-compatible.
+If fields shall be removed permanently from the API then a proper deprecation period must be adhered to so that end-users have enough time adapt their clients.
+
+**Checklist** when changing the API:
+
+1. Modify the field(s) in the respective Golang files of all external and the internal version.
+    1. Make sure new fields are being added as "optional" fields, i.e., they are of pointer types, they have the `// +optional` comment, and they have the `omitempty` JSON tag.
+    1. Make sure that the existing field numbers in the protobuf tags are not changed.
+1. If necessary then implement/adapt the conversion logic defined in the versioned APIs (e.g., `pkg/apis/core/v1beta1/conversions*.go`).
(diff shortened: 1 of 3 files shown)


timebertt (dismissed):
Nice, good idea 👍 
/approve

docs/development/changing-the-api.md
```diff
@@ -0,0 +1,37 @@
...
+
+The Gardener API is defined in `pkg/apis/{core,extensions,settings}` directories and is the main point of interaction with the system.
+It must be ensured that the API is always backwards-compatible.
+If fields shall be removed permanently from the API then a proper deprecation period must be adhered to so that end-users have enough time adapt their clients.
+
+**Checklist** when changing the API:
+
+1. Modify the field(s) in the respective Golang files of all external and the internal version. Make sure new fields are being added as "optional" fields, i.e., they are of pointer types, they have the `// +optional` comment, and they have the `omitempty` JSON tag.
+1. If necessary then implement/adapt the conversion logic defined in the versioned APIs (e.g., `pkg/apis/core/v1beta1/conversions*.go`).
+1. If necessary then implement/adapt defaulting logic defined in the versioned APIs (e.g., `pkg/apis/core/v1beta1/defaults*.go`).
```
timuthy, line 22: This has been a common pitfall for me when I extended the shoot API and defaulting/admission caused a spec change which again caused immediate but unwanted reconciliations.

I suggest to add a side note like: "Please be aware that defaulting might lead to a `Generation` change for existing objects and as a result the resource might be reconciled out of band. Example: Gardener implements advanced reconciliation decisions for shoot clusters, e.g. reconciliation only happens in maintenance time windows. On the other hand, it regularly adds labels to the shoot clusters which identify their health status and if a new defaulting logic is introduced, this label update request results in an increased generation as well."

timuthy (commented):
Just one suggestion from my own lessons learned.

docs/development/changing-the-api.md
```diff
@@ -0,0 +1,37 @@
...
+1. Run the code generation: `make install-requirements generate`
+1. If necessary then implement/adapt validation logic defined in the internal API (e.g., `pkg/apis/core/validation/validation*.go`).
+1. If necessary then adapt the exemplary YAML manifests of the Gardener resources defined in `example/*.yaml`.
+1. In most cases it makes sense to add/adapt the documentation for administrators/operators and/or end-users in the `docs` folder to provide information on purpose and usage of the added/changed fields.
+1. When opening the pull request then always add a release note so that end-users are becoming aware of the changes.
```
ialidzhikov, line 27: Probably we could add also 
```
1. Don't change the field numbers in the protobuf tag for any existing fields.
```

docs/development/changing-the-api.md
rfranzke: Hm, can you explain more how this situation can occur? I just tried to reproduce it locally by adding some defaulting logic for an optional field, started the API server, and then changed a label of the shoot. The generation did not increase. Is it only related to admission plugins?

timuthy: Oh, yes probably a corner case.
I encountered this because the defaulting option was only added to a version (core.gardener.cloud/v1alpha1) which was not the stored version of the object (garden.sapcloud.io/v1beta1). We can leave it out - sorry for the confusion.

danielfoehrKn (approved):

timuthy (approved):

timebertt (approved):

vpnachev (approved):
Very nice!

/approve

PR #2344, opened 2020-05-18
Removed the old test framework

schrodit:

The old test framework has been now completely removed.
The remaining packages that still used the old framework (maintenance test and network policies) have been adjusted to use the new framework.

:warning: The gardener repo does only contain the generator code for the network policies.
Therefore the tests have to be regenerated when the new gardener version in merged into the repositories. Therefore I will open follow up PR's as soon as gardener is released (example for aws https://github.com/gardener/gardener-extension-provider-aws/pull/95).

During refactoring I also noticed that the networkingpolicies are outdated:
- Dependency Watchdog is now part of the seed components in the garden namespace
- Networkpolicies to the respective extension are added
- With the newly introduced csi plugins some networkpolicies are different for shoot clusters with csi enabled and for clusters without csi.

I now fixed the first 2 issues but we also have to clarify who is responsible for maintaining these tests in the future.
 cc @mvladev @dguendisch 

Additionally: removed the terstdefinition for maintenance and scheduler test as are we currently 
unable to run them. We will read them as soon as the installer is ready.

Fixes #

**Release note**:

```action developer
The old testframework has been completely removed. All extension repositories that previously used the networkolicies generator have to be regenerated when this change is vendored.
```

/area testing
/kind cleanup
/priority normal


rfranzke (approved):
/needs second-opinion

dguendisch (approved):

Issue #2321, opened 2020-05-13
Extend the Shoot API so that users can provide secret data to extension configurations via references


plkokanov:
**What would you like to be added**:
Extend the `Shoot` API to support adding secrets data required by various `Shoot` components (dns, provider configurations, extension configurations) via secret references.
 
**Why is this needed**:
Currently some extension configurations could require secret data which has to be specified directly in the `Shoot`'s spec. For example:
```
apiVersion: shoot-cert-service.extensions.config.gardener.cloud/v1alpha1
kind: Configuration
issuerName: gardener
restrictIssuer: true # restrict issuer to any sub-domain of shoot.spec.dns.domain (default)
acme:
  email: john.doe@example.com
  server: https://acme-v02.api.letsencrypt.org/directory
# privateKey: | # Optional key for Let's Encrypt account.
#   -----BEGIN BEGIN RSA PRIVATE KEY-----
#   ...
#   -----END RSA PRIVATE KEY-----
```
To avoid this, references to secrets should be used instead. This means that the way it is currently done for the `dns` configuration should be changed so that required secrets are defined in a homogeneous way.

**Proposal**:
The `Shoot` spec has a new field similar to `Pod`'s `Volumes` which lists all the secrets that can be used by `Shoot` components. The example below shows how this could look for the certificate service extension and dns configuration:

```
apiVersion: core.gardener.cloud/v1beta1
kind: Shoot
spec:
  dns:
    domain: crazy-botany.core.my-custom-domain.com
    providers:
    - type: aws-route53
       secretName: domain-secret
  extensions:
  - type: shoot-cert-service
    providerConfig:
      apiVersion: shoot-cert-service.extensions.config.gardener.cloud/v1alpha1
      kind: Configuration
      issuerName: gardener
      restrictIssuer: true # restrict issuer to any sub-domain of shoot.spec.dns.domain (default)
      acme:
        email: john.doe@example.com
        server: https://acme-v02.api.letsencrypt.org/directory
        privateKeyRef:
          secretName: private-key
  kubernetes:
    version: 1.18.2
  ...
  Secrets:
  - name: private-key
    secretName: my-lets-encrypt-private-key
  - name: domain-secret
    secretName: my-domain-secret
```

The secrets are created by the user and stored in the `project`'s namespace on the garden cluster. During reconciliation they are copied to the `Shoot`'s namespace on the seed cluster.

**Note**
If we need to provide other objects then we could change the name of the spec.`Secrets` field and it could look like this
[...]

PR #2350, opened 2020-05-18
Persist and migrate secrets required by extension controllers

### LINKED ISSUE #2327: Persist secrets required by extension controllers so that they can be used during Control Plane Migration
Extend the ShootState, ShootState Sync Controller and the Status of Extension CRs so that secret resources required by extension controllers can be persisted in the ShootState and migrated from the source seed to the destination seed.

**Why is this needed**:
Currently extension controllers can save their state in the `status.state` field of the corresponding extension resource. This is synced by the `ShootState Sync Controller` to the `ShootState` and used to restore the state of the extension controller on the destination seed during control plane migration. However, if the extension controller has to persist a secret, it cannot be directly saved to the `status.state` as it contains confidential information. 

**Proposal**:
To handle such cases an additional field is added to the `status` of extension CRs similar to Pod `Volumes` - `status.Secrets` (if it is required to handle other resources similarly then we could change the name to e.g. `status.Resources`) 

- The `ShootState Sync Controller` will read the new field, copy the data of the corresponding secrets and add them to the `ShootState` resource. 
- The control plane migration's restoration phase has an additional step which gets all such secrets from the `ShootState` and deploys them on the `Seed` in the `Shoot`'s namespace.

Here's an example with the `Extension` CR used by the certificate extension controller
```
apiVersion: extensions.gardener.cloud/v1alpha1
kind: Extension
metadata:
  name: "extension-certificate-service"
  namespace: shoot--project--abc
spec:
  type: shoot-cert-service
  providerConfig:
    apiVersion: shoot-cert-service.extensions.config.gardener.cloud/v1alpha1
    kind: Configuration
    issuerName: gardener
    restrictIssuer: true # restrict issuer to any sub-domain of shoot.spec.dns.domain (default)
    acme:
      email: john.doe@example.com
      server: https://acme-v02.api.letsencrypt.org/directory
      privateKeyRef:
        name: private-key
status:
  observedGeneration: 2
  state:
    # some data
    privateKeySecretRef: my-lets-encrypt-private-key
  secretReferences:
  - name: private-key
     secretName: my-lets-encrypt-private-key
```

stoyanr:
Extends the ShootState and the Status of extension CRs, as well as the corresponding controller and migration logic, so that resources (usually secrets) required by extension controllers can be persisted in the ShootState and migrated from the source seed to the destination seed.

Fixes #2327, see also #2321

* The PR introduces one new utility, `utis/object.go`. I would like to add tests for this utility in a subsequent PR.

**Release note**:
```improvement operator
Resources (usually secrets) required by extension controllers can now be added to the extension status, persisted in the ShootState, and migrated from the source seed to the destination seed.
```

/area control-plane-migration
/kind task
/priority normal


rfranzke:
FYI: I converted the PR to a "draft PR" as it has dependencies that need to be resolved before it can be merged.
I'll try to check it asap.

swilen-iwanow (commented):
Nice! No comments from my POV.

pkg/apis/core/v1alpha1/helper/shootstate_list.go
```diff
@@ -94,3 +96,36 @@ func (g *GardenerResourceDataList) Upsert(data *gardencorev1alpha1.GardenerResou
 	}
 	*g = append(*g, *data)
 }
+
+// ResourceDataList is a list of ResourceData
+type ResourceDataList []gardencorev1alpha1.ResourceData
+
+// Delete deletes an item from the list
+func (r *ResourceDataList) Delete(ref *autoscalingv1.CrossVersionObjectReference) {
```
rfranzke, line 104: Can you add some small unit tests for these new functions?

pkg/apis/extensions/v1alpha1/types.go
```diff
@@ -39,6 +40,8 @@ type Status interface {
 	GetLastError() *gardencorev1beta1.LastError
 	// GetState retrieves the State of the extension
 	GetState() *runtime.RawExtension
+	// GetResources retrieves the list of named resource references referred to in the state by their names.
+	GetResources() []gardencorev1alpha1.NamedResourceReference
```
rfranzke:
```suggestion
	GetResources() []gardencorev1beta1.NamedResourceReference
```

pkg/apis/extensions/v1alpha1/types_defaults.go
```diff
@@ -63,6 +64,9 @@ type DefaultStatus struct {
 	// State can be filled by the operating controller with what ever data it needs.
 	// +optional
 	State *runtime.RawExtension `json:"state,omitempty"`
+	// Resources holds a list of named resource references that can be referred to in the state by their names.
+	// +optional
+	Resources []gardencorev1alpha1.NamedResourceReference `json:"resources,omitempty"`
```
rfranzke:
```suggestion
	Resources []gardencorev1beta1.NamedResourceReference `json:"resources,omitempty"`
```

pkg/apis/extensions/v1alpha1/types_defaults.go
```diff
@@ -99,3 +103,8 @@ func (d *DefaultStatus) GetObservedGeneration() int64 {
 func (d *DefaultStatus) GetState() *runtime.RawExtension {
 	return d.State
 }
+
+// GetResources implements Status.
+func (d *DefaultStatus) GetResources() []gardencorev1alpha1.NamedResourceReference {
```
rfranzke:
```suggestion
func (d *DefaultStatus) GetResources() []gardencorev1beta1.NamedResourceReference {
```

pkg/operation/common/extensions.go
```diff
@@ -125,7 +126,20 @@ func DeleteExtensionCR(
...
+	}
+
+	extensionStatus := obj.GetExtensionStatus()
+	if extensionStatus != nil {
+		for _, resource := range extensionStatus.GetResources() {
+			if err := utils.DeleteObjectByRef(ctx, c, &resource.ResourceRef, namespace); err != nil {
+				return err
+			}
+		}
+	}
```
rfranzke: Hmm, I'm not sure about this - shouldn't the extension controller itself by responsible for cleaning up these referred resources on deletion as well as on migration?

pkg/utils/object.go
```diff
@@ -0,0 +1,224 @@
...
+// from runtime.RawExtension.
+func CreateOrUpdateObject(ctx context.Context, c client.Client, gvk schema.GroupVersionKind, name, namespace string, content map[string]interface{}) error {
+	// Initialize the object
+	key := client.ObjectKey{Namespace: namespace, Name: name}
+	obj := &unstructured.Unstructured{}
+	obj.SetGroupVersionKind(gvk)
+	obj.SetName(name)
+	obj.SetNamespace(namespace)
+
+	// Create or update the object with retries for optimistic concurrency
```
rfranzke: What about using the [`retry.OnConflict`](https://github.com/gardener/gardener/blob/master/vendor/k8s.io/client-go/util/retry/util.go#L82:1) for this?

pkg/utils/object.go
```diff
@@ -0,0 +1,224 @@
...
+			found = false
+		}
+
+		// Set object content
+		if content != nil {
+			obj.SetUnstructuredContent(mergeObjectContents(obj.UnstructuredContent(),
+				filterMetadata(content, add(systemMetadataFields, "namespace", "name")...)))
+		}
+
+		// Create or update the object
```
rfranzke: Why not calling `controllerutil.CreateOrUpdate`?

pkg/utils/object.go
```diff
@@ -0,0 +1,224 @@
...
+
+		// Delete the object
+		err := c.Delete(ctx, obj)
+		if err != nil && (!apierrors.IsConflict(err) || retries == maxRetries) {
+			return errors.Wrapf(err, "could not delete object %s %s", gvk, key)
+		}
+
+		done = err == nil
+	}
+	return nil
```
rfranzke: Wouldn't

```suggestion
	return client.IgnoreNotFound(c.Delete(ctx, obj))
```

be enough? How can a conflict happen during delete?

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
```diff
@@ -114,20 +143,50 @@ func shouldSkipExtensionObjectSync(extensionObject extensionsv1alpha1.Object, ki
...
+			CrossVersionObjectReference: *ref,
+			Data:                        *data,
+		})
+	} else {
+		list.Delete(ref)
+	}
+	shootState.Spec.Resources = list
+}
+
+func getResourceByName(resources []gardencorev1alpha1.NamedResourceReference, name string) *gardencorev1alpha1.NamedResourceReference {
```
rfranzke: This could be put into a general API helper (`pkg/apis/core/v1beta1/helper`) as it'll probably be needed more often in the future, right?

rfranzke (changes requested):
Very nice PR, thanks!
Apart from the inline comments I'm wondering whether changes to the `Shoot` object should be excluded from this PR as there is no implementation behind these fields. Probably it's better to add them as part of #2321, or?

pkg/apis/extensions/v1alpha1/types.go
stoyanr: I will rework as suggested. Just a question: since there is `NamedResourceReference` in both `v1alpha1` and `v1beta1`, what's the reason to prefer beta over alpha in a version that is itself alpha? Is it simply that it's the latest / more stable one, or is there something more?

rfranzke: Yes, right, `v1beta1` is the latest + default version used by Gardener, that's the only reason.

pkg/apis/core/v1alpha1/helper/shootstate_list.go
stoyanr: Done

pkg/apis/extensions/v1alpha1/types.go
stoyanr: Thanks, done.

pkg/apis/extensions/v1alpha1/types_defaults.go
stoyanr: Done

pkg/apis/extensions/v1alpha1/types_defaults.go
stoyanr: Done

pkg/operation/common/extensions.go
stoyanr: Removed this code as suggested.

pkg/utils/object.go
stoyanr: Done

pkg/utils/object.go
stoyanr: Done

pkg/utils/object.go
stoyanr: It seems a conflict can't happen during delete, sorry, my bad. Done as suggested.

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
stoyanr: Done

stoyanr:
@rfranzke Thanks for your review! I believe I addressed all inline comments now. I added the changes to the `Shoot` object to justify adding `NamedObjectReference` in `shoot.go`, and especially in version `v1beta1`. The other option would have been to introduce it in `shootstate.go` and so only in version `v1alpha1`, and then refactor the whole thing in the upcoming PR #2360. I felt adding it where it would end up already at this point made more sense, I can still change it if you see an issue with this.

rfranzke:
Thanks! Makes sense, especially as PR #2360 is already open and the implementation ready. Can you please rebase to get rid of the conflicts, and potentially squash the PRs? Will take another look asap.

stoyanr:
@rfranzke You mean squash the 2 commits? I thought having the feedback in a separate commit would make re-reviewing easier for you (at least it does for me). Should I stick to 1 commit per PR in the future, i.e. is this a policy?

stoyanr:
Rebased and squashed

rfranzke (commented):
Looks great, thank you for addressing the suggestions! Also tried it out locally - works well.
@stoyanr can you add a release note for (extension) developers about the change?

@swilen-iwanow can you take another look as well, please?
/needs second-opinion

FYI @plkokanov @swilen-iwanow - I haven't found a clear document about what extension controllers have to do for the control plane migration - maybe you can work on such document and elaborate a little bit on the `ShootState`, `status.state`, `.status.resources`, etc.

stoyanr:
@rfranzke Added a release note as requested.

rfranzke (approved):
Thanks for the release note and your thorough tests!
/lgtm
As @swilen-iwanow already approved yesterday we can go ahead with the merge.

Issue #2325, opened 2020-05-13
Double logging in provider extensions with same information


ialidzhikov:
**What happened**:
Currently provider-azure's infrastructure_controller (most probably applicable also for other provider extensions) logs terraform errors as:
```
[...]

PR #2351, opened 2020-05-18
Prevent errors to be logged multiple times

ialidzhikov:

/area quality
/kind enhancement
/priority normal

controller-runtime.Reconciler errors are logged by the framework. See https://github.com/kubernetes-sigs/controller-runtime/blob/198197fe0fd7553cadd8afcd3338f8ee4b8a7e3f/pkg/internal/controller/controller.go#L258. Currently we have cases in which given error is logged multiple times - once by the actuator in the extension controller, once by the reconciler in g/g and once by the controller-runtime.
This PR removes superfluous logging in the extension reconcilers when error is supposed to be logged by controller-runtime.

Ref #2325

**Release note**:

```improvement operator
Extension Reconcilers do no longer log the same error which the controller-runtime is supposed to log too. This should prevent given error to appear multiple times in the logs.
```

```action developer
Extension controllers using `extensions/pkg/controller/error.RequeueAfterError` need to make sure that the cause error is properly logged on their side.
```

### DIFF
--- a/extensions/pkg/controller/backupbucket/reconciler.go
+++ b/extensions/pkg/controller/backupbucket/reconciler.go
@@ -106,8 +106,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 func (r *reconciler) reconcile(ctx context.Context, bb *extensionsv1alpha1.BackupBucket) (reconcile.Result, error) {
 	if err := extensionscontroller.EnsureFinalizer(ctx, r.client, FinalizerName, bb); err != nil {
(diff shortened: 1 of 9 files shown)


rfranzke (approved):
/lgtm

Out of curiosity: Why you don't use `errors.Wrapf`?

ialidzhikov:
> Out of curiosity: Why you don't use errors.Wrapf?

I have no strong opinion - if there is a convention/agreement to use `errors.Wrap`, I can adapt this PR.

rfranzke:
I'm fine with both, no strong opinion.

PR #2359, opened 2020-05-20
[GEP-7] Adds functionality to persist/load cp secrets to/from ShootState

plkokanov:

/area control-plane
/kind enhancement
/priority normal

This PR adds functionality to persist control plane secrets in the ShootState resources. 
As a first step all control-plane secrets which were previously generated during reconciliation (in the `DeploySecrets` flow step) are added to the ShootState - this can be removed in a future version.
Generally the old `DeploySecrets` step is split into 2 - the first one will generate all wanted secrets' metadata and add them to the `ShootState`; the second one will take all secrets' metadata from the `ShootState` and use the secret configs to generate the actual secrets in the control plane.

Part of #1631 

**Release note**:

```improvement operator
Secrets deployed by gardener in the `Shoot`'s Control Plane are now saved/loaded to/from the ShootState.
```


pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -202,9 +202,20 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation) *gardencorev1bet
...
+		loadSecrets = g.Add(flow.Task{
+			Name:         "Loading existing secrets into ShootState",
+			Fn:           flow.TaskFn(botanist.LoadExistingSecretsIntoShootState),
+			Dependencies: flow.NewTaskIDs(ensureShootStateExists),
+		})
+		generateSecrets = g.Add(flow.Task{
+			Name:         "Generating secrets and saving them into ShootState",
+			Fn:           flow.TaskFn(botanist.GenerateAndSaveSecrets),
+			Dependencies: flow.NewTaskIDs(loadSecrets, ensureShootStateExists),
+		})
```
swilen-iwanow, line 216: I see that this step adds the generated secrets to the ShootState, but do we need this generation of secrets if the namespace is in termination and we are not deploying them in the following step.

pkg/operation/controlplanesecrets/secrets_loader.go
```diff
@@ -0,0 +1,108 @@
...
+	return s.loadExistingSecretIntoResourceList(s.APIServerBasicAuthConfig)
+}
+
+// LoadStaticTokenFromSecret loads infodata for the static token from the existing kubernetes secret into this SecretLoader's GardenerResourceDataList
+func (s *SecretsLoader) LoadStaticTokenFromSecret() error {
+	return s.loadExistingSecretIntoResourceList(s.StaticTokenConfig)
+}
+
+func (s *SecretsLoader) loadExistingSecretIntoResourceList(secretConfig secrets.ConfigInterface) error {
+	if s.ExistingSecrets == nil {
```
swilen-iwanow: The function caller iteration wouldn't break and still going to test this. Maybe we should somehow differentiate, if this is nil or make it outside the function.

pkg/operation/botanist/secrets.go
```diff
@@ -1222,3 +401,7 @@ func dnsNamesForEtcd(namespace string) []string {
 	names = append(names, dnsNamesForService(fmt.Sprintf("%s-client", v1beta1constants.ETCDEvents), namespace)...)
 	return names
 }
+
+func shouldRotateBasicAuthCredentials(shootWantsBasicAuth, shootHasBasicAuth bool) bool {
+	return !shootWantsBasicAuth && shootHasBasicAuth || shootWantsBasicAuth && !shootHasBasicAuth
```
swilen-iwanow: Could it be simply?
```suggestion
	return shootWantsBasicAuth != shootHasBasicAuth
```

pkg/operation/controlplanesecrets/secrets_deployer.go
```diff
@@ -0,0 +1,228 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// SecretsDeployer defines methods which can be used to deploy Secrets. The configuration from these secrets is taken from the passed in
+// GardenerResourceDataList. The deployed secrets are saved in the DeployedSecrets.
+type SecretsDeployer struct {
+	*SecretsManager
+	APIServerBasicAuth *secrets.BasicAuth
+	StaticToken        *secrets.StaticToken
```
swilen-iwanow: Why are these exported? If we don't need them to change outside the deployer maybe we can make it private and remove the checks  `if s.StaticToken == nil {...}` in the exported functions. Wdyt, does it make sense?

pkg/operation/controlplanesecrets/secrets_generator.go
```diff
@@ -0,0 +1,96 @@
...
+func NewSecretsGenerator(
+	gardenerResourceDataList gardencorev1alpha1helper.GardenerResourceDataList,
+	apiserverBasicAuthConfig secrets.ConfigInterface,
+	staticTokenConfig secrets.ConfigInterface,
+	certificateAuthorityConfigs map[string]*secrets.CertificateSecretConfig,
+	secretConfigGenerator SecretConfigGeneratorFunc,
+) *SecretsGenerator {
+	s := &SecretsGenerator{&SecretsManager{}}
+	s.CertificateAuthorities = map[string]*secrets.Certificate{}
+	s.CertificateAuthorities = map[string]*secrets.Certificate{}
```
swilen-iwanow: You have these twice.

pkg/operation/controlplanesecrets/controlplane_secrets_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+package controlplanesecrets_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestEtcdEncryption(t *testing.T) {
```
rfranzke:
```suggestion
func TestControlPlaneSecrets(t *testing.T) {
```

pkg/operation/botanist/wanted_secrets.go
```diff
@@ -0,0 +1,696 @@
+package botanist
```
rfranzke, line 15: license header..

pkg/operation/botanist/secrets.go
```diff
@@ -1222,3 +401,7 @@ func dnsNamesForEtcd(namespace string) []string {
...
+}
```
rfranzke:
```suggestion
func shouldRotateBasicAuthCredentials(shootWantsBasicAuth, shootHasBasicAuth bool) bool {
	return shootWantsBasicAuth != shootHasBasicAuth
}
```
Given this simplification and the confusing function name we probably don't need the function at all.

pkg/operation/controlplanesecrets/secrets_deployer.go
```diff
@@ -0,0 +1,228 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: Probably this whole package should be moved to `pkg/operation/botanist/shootsecrets` (some secrets are not related to the control plane, hence, `controlplanesecrets` is not entirely accurate).

pkg/operation/controlplanesecrets/secrets_manager.go
```diff
@@ -0,0 +1,49 @@
...
+	APIServerBasicAuthConfig    secrets.ConfigInterface
+	StaticTokenConfig           secrets.ConfigInterface
+	CertificateAuthorityConfigs map[string]*secrets.CertificateSecretConfig
+	SecretConfigGenerator       SecretConfigGeneratorFunc
+
+	CertificateAuthorities map[string]*secrets.Certificate
+}
+
+// LoadCertificateAuthorities will load the CertificateAuthorities from the available GardenerResourceDataList
+func (s *SecretsManager) LoadCertificateAuthorities() error {
```
rfranzke: Now that you have, very exemplarily, added unit tests for most new functions, can you also add one for this function?

pkg/operation/controlplanesecrets/secrets_loader.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// SecretsLoader contains methods which read already existing secrets and load them into the provided GardenerResourceDataList
+type SecretsLoader struct {
+	*SecretsManager
+	ExistingSecrets map[string]*corev1.Secret
+}
+
+// NewSecretsLoader creates a new SecretsLoader.
+func NewSecretsLoader(
```
rfranzke: Can't you pass the `SecretsManager` object instead of all the individual parts?

pkg/utils/secrets/basic_auth.go
```diff
@@ -73,6 +75,45 @@ func (s *BasicAuthSecretConfig) Generate() (Interface, error) {
 	return s.GenerateBasicAuth()
 }
 
+// GenerateInfoData implements ConfigInterface.
+func (s *BasicAuthSecretConfig) GenerateInfoData() (infodata.InfoData, error) {
+	password, err := utils.GenerateRandomString(s.PasswordLength)
```
rfranzke, line 79: Hm, I'm a little confused. Now we have multiple functions generating a password, and also multiple functions that can load an existing basic auth data. Is it possible to harmonize this to make it easier for a consumer to know what to call when?

pkg/utils/secrets/infodata/basic_auth.go
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: All these new files+functions of this `infodata` package seem to be tightly coupled to the `pkg/utils/secrets` package. Does it make sense to merge them into it?

pkg/utils/secrets/vpn_tlsauth.go
```diff
@@ -0,0 +1,104 @@
+package secrets
```
rfranzke, line 15: license header

pkg/operation/controlplanesecrets/secrets_generator.go
```diff
@@ -0,0 +1,96 @@
...
+package controlplanesecrets
+
+import (
+	gardencorev1alpha1helper "github.com/gardener/gardener/pkg/apis/core/v1alpha1/helper"
+	"github.com/gardener/gardener/pkg/utils/infodata"
+	"github.com/gardener/gardener/pkg/utils/secrets"
+)
+
+// SecretsGenerator extends the SecretsManager with methods which can generate InfoData from the given secret configurations.
+type SecretsGenerator struct {
```
rfranzke: What's the purpose of this struct? It's only using the `SecretsManager` internally. Does it make sense to remove it and move the struct methods to the `SecretsManager`, instead?

pkg/operation/controlplanesecrets/secrets_generator.go
```diff
@@ -0,0 +1,96 @@
...
+	secretConfigs, err := s.SecretConfigGenerator(nil, nil, s.CertificateAuthorities)
+	if err != nil {
+		return err
+	}
+	for _, config := range secretConfigs {
+		if err := generateSecretInfoData(&s.GardenerResourceDataList, config); err != nil {
+			return err
+		}
+	}
+	return err
```
rfranzke:
```suggestion
	return nil
```

pkg/operation/controlplanesecrets/secrets_deployer.go
```diff
@@ -0,0 +1,228 @@
...
+	StaticToken        *secrets.StaticToken
+
+	ExistingSecrets map[string]*corev1.Secret
+	DeployedSecrets map[string]*corev1.Secret
+
+	mutex *sync.RWMutex
+}
+
+// NewSecretsDeployer creates a new SecretsDeployer.
+func NewSecretsDeployer(
```
rfranzke: Does it make sense to only pass the `SecretsManager` object instead of all its related attributes?

pkg/operation/controlplanesecrets/secrets_deployer.go
```diff
@@ -0,0 +1,228 @@
...
+	s.SecretConfigGenerator = secretConfigGenerator
+	s.ExistingSecrets = existingSecrets
+	s.DeployedSecrets = deployedSecrets
+	s.mutex = mutex
+
+	return s
+}
+
+// LoadBasicAuth loads the APIServerBasicAuth secret from the corresponding InfoData
+func (s *SecretsDeployer) LoadBasicAuth() error {
```
rfranzke: Hm, conceptually, I find the `Load*` functions in the `SecretsDeployer` somewhat confusing when there already is a `SecretsLoader`. Can this be improved?

pkg/operation/controlplanesecrets/secrets_generator.go
```diff
@@ -0,0 +1,96 @@
...
+	s.GardenerResourceDataList = gardenerResourceDataList
+	s.APIServerBasicAuthConfig = apiserverBasicAuthConfig
+	s.StaticTokenConfig = staticTokenConfig
+	s.CertificateAuthorityConfigs = certificateAuthorityConfigs
+	s.SecretConfigGenerator = secretConfigGenerator
+	return s
+}
+
+// GenerateAPIServerBasicAuth will generate InfoData for the APIServerBasicAuth
+func (s *SecretsGenerator) GenerateAPIServerBasicAuth() error {
```
swilen-iwanow: I think we could have a single exported method called `Generate` which is doing the work of the methods bellow. We just return the errors in the client code, anyways. It then will have another boolean field to determine if it should generate basic auth secrets, called `withBasicAuth` for example.

pkg/operation/botanist/secrets.go
```diff
@@ -15,838 +15,189 @@
...
 	}
 
-	wantedSecretsList, err := b.generateWantedSecrets(basicAuthAPIServer, staticToken, certificateAuthorities)
+	b.ShootState = shootState
+
+	return nil
+}
+
+// DeploySecrets takes all existing secrets from the ShootState resource and deploys them in the shoot's control plane.
+func (b *Botanist) DeploySecrets(ctx context.Context) error {
```
rfranzke, line 139: Is my understanding correct that this function reads the `ShootState` again after it was previously generated by `GenerateAndSaveSecrets`? If yes, can we avoid these network calls by only reading the `ShootState` once and then passing it internally to the next function?

pkg/operation/controlplanesecrets/secrets_generator.go
plkokanov: Wanted to keep the functions related to generating secrets separate from the SecretsDeployer struct, since it also uses thee SecretsManager struct. However, now I'm thinking of changing this: 
rename `SecretsManager` to `SecretsConfig` and just use it as a common struct for the secret configurations. Then pass it to the SecretsGenerator, SecretsDeployer and SecretsLoader. This way I could even have it created only once.

pkg/operation/botanist/secrets.go
plkokanov: It reads the saved struct from the botanist. So what you suggested is already happening :)

pkg/gardenlet/controller/shoot/shoot_control_delete.go
plkokanov: I guess all of the secrets functions should not be executed in that case.

pkg/operation/controlplanesecrets/secrets_deployer.go
plkokanov: We need at least the StaticToken exported for as it has to be saved for later usage during the DeploySecrets step.

pkg/operation/controlplanesecrets/secrets_generator.go
plkokanov: I had this idea as well. Generally could look something like (same for generator, deployer and loader):
```
secretsManager := NewSecretsManager().WithStaticToken(...).WithGeneratorFunc(...).WithCAConfigs(...)
if shootWantsBasicAuth {
    secretsManager := secretsManager.WithAPIServerBasicAuth(...)
}
secretsDeployer := NewSecretsDeployer().WithManager(secretsManager).WithExistingSecrets(...)
if err := secretsDeployer.Deploy(); err != nil {
...
}
mutex.Lock()
defer mutex.Unlock()
for name, secret := range secretsDeployer.DeployedSecrets {
  b.Secrets[name] = secret
}
return nil
```

Might as well change the `SecretsManager` to `SecretsConfig` I think something like that makes more sense.

pkg/utils/secrets/infodata/basic_auth.go
plkokanov: If we want to have the infodata only in Gardener then I guess I should try to make them more independent, as the secret configs are used in some extension controllers (however, i don't have any good ideas to that currently). Otherwise I could add them to the same package (simiar to the etcdencryption)

pkg/utils/secrets/basic_auth.go
stoyanr: I also personally find the division of responsibilities between `SecretsManager`, `SecretsLoader`, `SecretsGenerator`, and `SecretsDeployer` a bit confusing. I think this is the root cause for some of the other comments and might cause issues in the future as well. I think here is no very good way to do such a fine-granular division in this case. I would prefer combining all these 4 types into a single one (e.g. `SecretsManager`) having all `Load`, `Generate`, and `Deploy` methods. It would turn out a bit biggish, but I see this as the "lesser evil" in this case.

pkg/utils/secrets/infodata/basic_auth.go
stoyanr: I would agree with @rfranzke, the `infodata` package is too tightly coupled to deserve an own package and the split of the code between the 2 packages makes reading and understanding the code harder. So I would also vote for a single package (and set of files).

pkg/operation/controlplanesecrets/secrets_generator.go
stoyanr: Why not combine everything together, see above.

pkg/operation/controlplanesecrets/secrets_deployer.go
stoyanr: I also personally find the division of responsibilities between `SecretsManager`, `SecretsLoader`, `SecretsGenerator`, and `SecretsDeployer` a bit confusing. I think this is the root cause for some of the other comments and might cause issues in the future as well. I think here is no very good way to do such a fine-granular division in this case. I would prefer combining all these 4 types into a single one (e.g. `SecretsManager`) having all `Load`, `Generate`, and `Deploy` methods. It would turn out a bit biggish, but I see this as the "lesser evil" in this case

pkg/utils/secrets/basic_auth.go
stoyanr: @rfranzke I think these functions are indeed used in different situations, so perhaps they should stay different but maybe the documentation comments when each of them is used should be added?

pkg/operation/controlplanesecrets/secrets_generator.go
stoyanr: Agree with @swilen-iwanow, and I would say perhaps the same is true for `Deploy`, and `Load`. See my other proposal about combining everything together in one type, e.g. `SecretsManager`. This type could then expose methods such as `Generate`, `Deploy`, and `Load`, and many of the current methods would no longer be exported.

pkg/operation/controlplanesecrets/secrets_deployer_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+	BeforeEach(func() {
+		ctrl = gomock.NewController(GinkgoT())
+		k8sClient = mockclient.NewMockClient(ctrl)
+
+		gardenerResourceDataList = gardencorev1alpha1helper.GardenerResourceDataList{
+			v1alpha1.GardenerResourceData{
+				Name: common.BasicAuthSecretName,
+				Type: string(secretsinfodata.BasicAuthDataType),
+				Data: runtime.RawExtension{Raw: []byte("{\"password\":\"asdf\"}")},
```
ialidzhikov: Nit - you could use backticks to get rid of escaping

```suggestion
				Data: runtime.RawExtension{Raw: []byte(`{"password":"asdf"}`)},
```

pkg/operation/controlplanesecrets/secrets_deployer_test.go
```diff
@@ -0,0 +1,221 @@
...
+			},
+			v1alpha1.GardenerResourceData{
+				Name: common.StaticTokenSecretName,
+				Type: string(secretsinfodata.StaticTokenDataType),
+				Data: runtime.RawExtension{Raw: []byte("{\"tokens\":[{\"userName\":\"foo\",\"userID\":\"fooID\",\"groups\":[\"group1\",\"group2\"],\"token\":\"foo\"}]}")},
+			},
+			v1alpha1.GardenerResourceData{
+				Name: "ca",
+				Type: string(secretsinfodata.CertificateDataType),
+				Data: runtime.RawExtension{Raw: []byte(fmt.Sprintf("{\"privateKey\":\"%s\",\"certificate\":\"%s\"}", []byte(cakey), []byte(cacert)))},
```
ialidzhikov: Nit - I guess the `[]byte` casting here is unnecessary:

```suggestion
				Data: runtime.RawExtension{Raw: []byte(fmt.Sprintf("{\"privateKey\":\"%s\",\"certificate\":\"%s\"}", cakey, cacert))},
```

pkg/operation/controlplanesecrets/secrets_deployer_test.go
```diff
@@ -0,0 +1,221 @@
...
+				secrets.DataKeyCertificateCA: b,
+				secrets.DataKeyPrivateKeyCA:  a,
+			},
+			Type: corev1.SecretTypeOpaque,
+		}
+
+		k8sClient.EXPECT().Create(context.TODO(), caSecret).Return(nil)
+		err := secretsDeployer.LoadCertificateAuthorities()
+		Expect(err).NotTo(HaveOccurred())
+		defer GinkgoRecover()
```
ialidzhikov: Hm, what is the need for `GinkgoRecover` usage here? Is it a leftover?

pkg/utils/secrets/vpn_tlsauth.go
```diff
@@ -0,0 +1,104 @@
...
+	} else {
+		key, err = generateKeyDefault()
+	}
+	return
+}
+
+// SecretData computes the data map which can be used in a Kubernetes secret.
+func (v *VPNTLSAuth) SecretData() map[string][]byte {
+	data := map[string][]byte{
+		"vpn.tlsauth": v.TLSAuthKey,
```
ialidzhikov:
```suggestion
		DataKeyVPNTLSAuth: v.TLSAuthKey,
```

pkg/utils/secrets/vpn_tlsauth_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+		Describe("#Generate", func() {
+			It("should properly generate VPNTLSAuth object", func() {
+				obj, err := vpnTLSAuthConfig.Generate()
+				Expect(err).NotTo(HaveOccurred())
+
+				vpnTLSAuth, ok := obj.(*VPNTLSAuth)
+				Expect(ok).To(BeTrue())
+
+				Expect(len(vpnTLSAuth.TLSAuthKey)).ToNot(BeNil())
```
ialidzhikov:
```suggestion
				Expect(vpnTLSAuth.TLSAuthKey).ToNot(BeEmpty())
```

plkokanov:
Updated the PR and rebased. 

@zanetworker I had to change the code which adds an additional username to the static token so as to include it during the generation of the static token's infodata. It might be a good idea to take a look if everything is ok [secrets_manager.go:201](https://github.com/gardener/gardener/blob/e1034ed0d095c47febc0c31ae1e12375162b7b90/pkg/operation/shootsecrets/secrets_manager.go#L201). There's a test case for it here: [secrets_manager_test.go:192](https://github.com/gardener/gardener/blob/e1034ed0d095c47febc0c31ae1e12375162b7b90/pkg/operation/shootsecrets/secrets_manager_test.go#L192)

zanetworker:
@plkokanov  will try to test it.

pkg/operation/botanist/wanted_secrets.go
```diff
@@ -0,0 +1,718 @@
...
+		CertType:   secrets.CACert,
+	},
+	v1beta1constants.SecretNameCAMetricsServer: {
+		Name:       v1beta1constants.SecretNameCAMetricsServer,
+		CommonName: "metrics-server",
+		CertType:   secrets.CACert,
+	},
+}
+
+// generateWantedSecrets returns a list of Secret configuration objects satisfying the secret config intface,
```
ialidzhikov:
```suggestion
// generateWantedSecrets returns a list of Secret configuration objects satisfying the secret config interface,
```

pkg/operation/shootsecrets/secrets_manager.go
```diff
@@ -0,0 +1,386 @@
...
+	"github.com/gardener/gardener/pkg/operation/common"
+	"github.com/gardener/gardener/pkg/utils/infodata"
+	"github.com/gardener/gardener/pkg/utils/secrets"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// SecretConfigGeneratorFunc ...
```
ialidzhikov: I guess this is a leftover

pkg/utils/secrets/private_key_infodata_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+
+	"github.com/gardener/gardener/pkg/utils"
+
+	. "github.com/gardener/gardener/pkg/utils/secrets"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("RrivateKey InfoData", func() {
```
ialidzhikov:
```suggestion
var _ = Describe("PrivateKey InfoData", func() {
```

pkg/utils/secrets/private_key_infodata_test.go
```diff
@@ -0,0 +1,50 @@
...
+	var (
+		privateKeyJSON     = []byte(fmt.Sprintf(`{"privateKey":"%s"}`, utils.EncodeBase64([]byte("foo"))))
+		privateKeyInfoData = &PrivateKeyInfoData{
+			PrivateKey: []byte("foo"),
+		}
+	)
+
+	Describe("#UnmarshalPrivateKey", func() {
+		It("should properly unmarshal RrivateKeyJSONData into PrivateKeyInfoData", func() {
```
ialidzhikov:
```suggestion
		It("should properly unmarshal PrivateKeyJSONData into PrivateKeyInfoData", func() {
```

pkg/utils/secrets/private_key_infodata.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// PrivateKeyDataType is the type used to denote an PrivateKeyJSONData structure in the ShootState
+const PrivateKeyDataType = infodata.TypeVersion("privateKey")
+
+func init() {
+	infodata.Register(PrivateKeyDataType, UnmarshalPrivateKey)
+}
+
+// PrivateKeyJSONData is the json representation of CertificateInfoData used to store private key in the ShootState
```
ialidzhikov:
```suggestion
// PrivateKeyJSONData is the json representation of PrivateKeyInfoData used to store private key in the ShootState
```

pkg/operation/botanist/wanted_secrets.go
```diff
@@ -0,0 +1,718 @@
...
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	"github.com/gardener/gardener/pkg/features"
+	gardenletfeatures "github.com/gardener/gardener/pkg/gardenlet/features"
+	"github.com/gardener/gardener/pkg/operation/common"
+	"github.com/gardener/gardener/pkg/utils/secrets"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/authentication/user"
+)
+
+var basicAuthSecretAPIServer = &secrets.BasicAuthSecretConfig{
```
ialidzhikov, line 31: Hmm, I tried to enable basic auth for newly created Shoot but the kube-apiserver failed to start with:
```
  Warning  FailedMount  7s (x15 over 14m)  kubelet, ip-10-250-11-255.eu-west-1.compute.internal  MountVolume.SetUp failed for volume "kube-apiserver-basic-auth" : secret "kube-apiserver-basic-auth" not found
```

Is this reproducible also for you?

pkg/operation/botanist/wanted_secrets.go
```diff
@@ -0,0 +1,718 @@
...
+		kubecfgToken, err = staticToken.GetTokenForUsername(common.KubecfgUsername)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Secret definition for kubecfg
+	secretList = append(secretList, &secrets.ControlPlaneSecretConfig{
+		CertificateSecretConfig: &secrets.CertificateSecretConfig{
+			Name:      common.KubecfgSecretName,
```
ialidzhikov, line 549: Hmm, I get the following entry in the ShootState for the kubecfg secret.
```
  - data:
      certificate: null
      privateKey: null
    name: kubecfg
    type: certificate
```

Is that expected?

pkg/operation/botanist/wanted_secrets.go
plkokanov: Missed adding it in the deploy step. Thanks for finding it out. Couldn't complete all the tests as my dev setup has problems currently.

pkg/operation/botanist/wanted_secrets.go
plkokanov: Yep, the kubecfg secret doesn't have a certificate and private key right now. 
TBH we don't really need to add it to the shootstate right now as all of the secret infodata required for the kubecfg secret is in the staticToken and the APIServerBasicAuth infodata. Initially I let it stay but I think I can improve it since there really is no need to have it in the shootstate now that I think about it.

zanetworker:
@plkokanov the konnectivity-server secrets are no longer created in this PR (e.g., KonnectivityServerKubeconfig, and the server certificate): 

```go 
	secretList = append(secretList,
			&secrets.ControlPlaneSecretConfig{
				CertificateSecretConfig: &secrets.CertificateSecretConfig{
					Name:      common.KonnectivityServerKubeconfig,
					SigningCA: certificateAuthorities[v1beta1constants.SecretNameCACluster],
				},

				BasicAuth: basicAuthAPIServer,
				Token:     konnectivityServerToken,

				KubeConfigRequest: &secrets.KubeConfigRequest{
					ClusterName:  b.Shoot.SeedNamespace,
					APIServerURL: fmt.Sprintf("%s.%s", v1beta1constants.DeploymentNameKubeAPIServer, b.Shoot.SeedNamespace),
				},
			},
			&secrets.ControlPlaneSecretConfig{
				CertificateSecretConfig: &secrets.CertificateSecretConfig{
					Name:       common.KonnectivityServerCertName,
					CommonName: common.KonnectivityServerCertName,
					DNSNames:   konnectivityServerDNSNames,

					CertType:  secrets.ServerCert,
					SigningCA: certificateAuthorities[v1beta1constants.SecretNameCACluster],
				},
			})
```

So new shoots won't be created because the secret won't be mounted (as it does not exist) and the APIServer pod will be stuck in `ContainerCreating`

plkokanov:
> @plkokanov the konnectivity-server secrets are no longer created in this PR (e.g., KonnectivityServerKubeconfig, and the server certificate):
> 
> ```go
> 	secretList = append(secretList,
> 			&secrets.ControlPlaneSecretConfig{
> 				CertificateSecretConfig: &secrets.CertificateSecretConfig{
> 					Name:      common.KonnectivityServerKubeconfig,
> 					SigningCA: certificateAuthorities[v1beta1constants.SecretNameCACluster],
> 				},
> 
> 				BasicAuth: basicAuthAPIServer,
> 				Token:     konnectivityServerToken,
> 
> 				KubeConfigRequest: &secrets.KubeConfigRequest{
> 					ClusterName:  b.Shoot.SeedNamespace,
> 					APIServerURL: fmt.Sprintf("%s.%s", v1beta1constants.DeploymentNameKubeAPIServer, b.Shoot.SeedNamespace),
> 				},
> 			},
> 			&secrets.ControlPlaneSecretConfig{
> 				CertificateSecretConfig: &secrets.CertificateSecretConfig{
> 					Name:       common.KonnectivityServerCertName,
> 					CommonName: common.KonnectivityServerCertName,
> 					DNSNames:   konnectivityServerDNSNames,
> 
> 					CertType:  secrets.ServerCert,
> 					SigningCA: certificateAuthorities[v1beta1constants.SecretNameCACluster],
> 				},
> 			})
> ```
> 
> So new shoots won't be created because the secret won't be mounted (as it does not exist) and the APIServer pod will be stuck in `ContainerCreating`

Thanks for checking it out. I'll re-add them. And see if there are any other changes with the generateWantedSecretConfigs func that I might have missed.

pkg/operation/botanist/wanted_secrets.go
ialidzhikov: Now it is working again. Thanks.

pkg/operation/shootsecrets/secrets_manager.go
```diff
@@ -0,0 +1,394 @@
...
+		close(results)
+	}()
+	for out := range results {
+		if out.err != nil {
+			errorList = append(errorList, out.err)
+			continue
+		}
+		s.DeployedSecrets[out.secret.Name] = out.secret
+	}
+	// Wait and check wether an error occurred during the parallel processing of the Secret creation.
```
ialidzhikov:
```suggestion
	// Wait and check whether an error occurred during the parallel processing of the Secret creation.
```

pkg/operation/botanist/wanted_secrets.go
ialidzhikov: Okay, thanks.

ialidzhikov (commented):
/lgtm

I played locally with the PR and operations like reconcile of existing cluster, deletion of existing cluster, switch on/off of kube-apiserver basic auth were ok.

cc @rfranzke for the codewise changes
cc @zanetworker for the konnectivity-server secrets

rfranzke (dismissed):

rfranzke (dismissed):

rfranzke:
@plkokanov could you kindly rebase? There are no conflicts, but @tim-ebert PR got merged, so just to be sure to be up-to-date.
@zanetworker will have another look at the PR

plkokanov:
Rebased and ran a quick test to check that the reconciliation is still working properly.

zanetworker:
@plkokanov so I did another round of testing and I am seeing issues with the token appending functionality (which was removed from this PR). Basically, the token is not being appended to  an existing static token secret which leads to authorization issues.

plkokanov:
assi

> @plkokanov so I did another round of testing and I am seeing issues with the token appending functionality (which was removed from this PR). Basically, the token is not being appended to an existing static token secret which leads to authorization issues.

Thats because of [this](https://github.com/gardener/gardener/pull/2359/files#diff-4beaaf961a45e1ffff087b35d73bf95fR377) line. Perhaps it makes sense to check if any secret has changes in its Data field and update it, or I could just limit it to the staticToken @zanetworker wdyt?

zanetworker:
So far I see no other use case for other secrets so I would say the static token should be okay for now.  Also removing the data tokens from the secrets is something that is needed (but I am adding that in another [PR](https://github.com/gardener/gardener/pull/2484)) . 

You can just copy the logic from the current `AppendStaticToken` function. You can also have a look at https://github.com/gardener/gardener/pull/2484

plkokanov:
I think I managed to adapt to your future changes (at lest for the tokens addition/removal), I'll test it and add them to the PR, then it will be easier to rebase if your PR gets merged in first.

zanetworker:
sure, let me know what order you prefer. I can also wait till this one is merged.

plkokanov:
@zanetworker updated and rebased the PR.
Tested it by running the reconciliation without the konnectivity feature, then enabling it and reconciling and then again disabling and reconciling.

rfranzke (approved):

PR #2251, opened 2020-04-28
[GEP-11] Invert tunnel direction for >= 1.18 clusters

zanetworker:
**What this PR does / why we need it**:
So far we have been using OpenVPN to establish a tunnel from seed to shoot. This PR inverts the tunnel connection to be established from shoot to seed using the API server network proxy. For more information, please check the corresponding [GEP](https://github.com/gardener/gardener/pull/2170).  
  
**Special notes for Reviewer**
To address [this comment](https://github.com/gardener/gardener/pull/2251#discussion_r417075006), I refactored the static-token generation logic to append new tokens (otherwise a regeneration of all secrets making use of the token is required). 

**Release note**:

```noteworthy operator
The Gardenlet supports a new `KonnectivityTunnel` feature gate (disabled by default). If enabled then the network connection between the shoot control plane in the seed and the shoot worker nodes will be established from shoot->seed for >= 1.18 shoot clusters (instead of seed->shoot like earlier). Furthermore, in this case the additional "vpn-shoot" load balancer in the shoot will no longer be needed. Please note that the feature is in alpha state and might be promoted in future Gardener releases.
```

PR #2360, opened 2020-05-20
Reconcile and delete Shoot resource references

### LINKED ISSUE #2321: Extend the Shoot API so that users can provide secret data to extension configurations via references

stoyanr:
/area control-plane-migration
/kind task
/priority normal

Extends the Shoot API, as well as the corresponding reconciliation and deletion logic, to support references to resources (usually secrets) required by various Shoot components (DNS, extension provider configurations, etc.).

Fixes #2321, see also #2327.

* Doesn't change `DNS` in Shoot as recommended in #2321, this could be added later in a separate PR.

**Release note**:
```improvement operator
References to resources (usually secrets) can now be added to the Shoot and referred to by name in various Shoot components, e.g. extension provider configurations. 
```

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -87,6 +87,7 @@
   * Generic (non-essential) extensions
     * [`Extension` resource](extensions/extension.md)
 * [Extending project roles](extensions/project-roles.md)
+* [Referenced resources](extensions/referenced-resources.md)
 
 ## Testing
 
--- a/docs/extensions/referenced-resources.md
+++ b/docs/extensions/referenced-resources.md
@@ -1,0 +1,46 @@
+# Referenced Resources
+
+The Shoot resource can include a list of resources (usually secrets) that can be referenced by name in extension `providerConfig` and other Shoot sections, for example:
+
+```yaml
+kind: Shoot
+apiVersion: core.gardener.cloud/v1beta1
+metadata:
+  name: crazy-botany
+  namespace: garden-dev
+  ...
+spec:
+  ...
+  extensions:
+  - type: foobar
+    providerConfig:
+      apiVersion: foobar.extensions.gardener.cloud/v1alpha1
+      kind: FooBarConfig
+      foo: bar
+      secretRef: foobar-secret
+  resources:
+  - name: foobar-secret
+    resourceRef:
+      apiVersion: v1
+      kind: Secret
+      name: my-foobar-secret
+```
+
+Gardener expects to find these referenced resources in the project namespace (e.g. `garden-dev`) and will copy them to the Shoot namespace in the Seed cluster when reconciling a Shoot, adding a prefix to their names to avoid naming collisions with Gardener's own resources. 
+
+Extension controllers can resolve the references to these resources by accessing the Shoot via the `Cluster` resource. To properly read a referenced resources, extension controllers should use the utility function `GetObjectByReference` from the `extensions/pkg/controller` package, for example:
+
+```go
+    ...
+    ref = &autoscalingv1.CrossVersionObjectReference{
+        APIVersion: "v1",
+        Kind:       "Secret",
+        Name:       "foo",
+    }
+    secret := &corev1.Secret{}
+    if err := controller.GetObjectByReference(ctx, client, ref, "shoot--test--foo", secret); err != nil {
+        return err
+    }
+    // Use secret
+    ...
+```
--- a/example/90-shoot.yaml
+++ b/example/90-shoot.yaml
@@ -219,6 +219,7 @@ spec:
   #   apiVersion: foobar.extensions.gardener.cloud/v1alpha1
   #   kind: FooBarConfig
   #   foo: bar
+  #   secretRef: foobar-secret # See the resources section below for the actual reference
   # disabled: true # only relevant for extensions that were marked as 'globally enabled' by Gardener administrators
   networking:
     type: calico
@@ -270,3 +271,10 @@ spec:
 # seedSelector:
 #   matchLabels:
 #     foo: bar
+# List resources referenced by providerConfig and other sections, if any
+# resources:
+# - name: foobar-secret
+#   resourceRef:
+#     apiVersion: v1
+#     kind: Secret
+#     name: my-foobar-secret
--- a/extensions/pkg/controller/utils.go
+++ b/extensions/pkg/controller/utils.go
@@ -22,12 +22,14 @@ import (
 
 	controllererror "github.com/gardener/gardener/extensions/pkg/controller/error"
 	"github.com/gardener/gardener/extensions/pkg/util"
+	"github.com/gardener/gardener/pkg/operation/botanist"
 
 	resourcemanagerv1alpha1 "github.com/gardener/gardener-resource-manager/pkg/apis/resources/v1alpha1"
 	"github.com/gardener/gardener/pkg/api/extensions"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -272,25 +274,6 @@ func exponentialBackoff(ctx context.Context, backoff wait.Backoff, condition wai
 	return wait.ErrWaitTimeout
 }
 
-// WaitUntilResourceDeleted deletes the given resource and then waits until it has been deleted. It respects the
-// given interval and timeout.
-func WaitUntilResourceDeleted(ctx context.Context, c client.Client, obj runtime.Object, interval time.Duration) error {
-	key, err := client.ObjectKeyFromObject(obj)
-	if err != nil {
-		return err
-	}
-
-	return wait.PollImmediateUntil(interval, func() (done bool, err error) {
-		if err := c.Get(ctx, key, obj); err != nil {
-			if apierrors.IsNotFound(err) {
-				return true, nil
-			}
-			return false, err
-		}
-		return false, nil
-	}, ctx.Done())
-}
-
 // WatchBuilder holds various functions which add watch controls to the passed Controller.
 type WatchBuilder []func(controller.Controller) error
 
@@ -362,4 +345,10 @@ func IsMigrated(obj runtime.Object) bool {
 	return lastOp != nil &&
 		lastOp.Type == gardencorev1beta1.LastOperationTypeMigrate &&
 		lastOp.State == gardencorev1beta1.LastOperationStateSucceeded
+}
+
+// GetObjectByReference gets an object by the given reference, in the given namespace.
+// If the object kind doesn't match the given reference kind this will result in an error.
+func GetObjectByReference(ctx context.Context, c client.Client, ref *autoscalingv1.CrossVersionObjectReference, namespace string, obj runtime.Object) error {
+	return c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: botanist.ReferencedResourcesPrefix + ref.Name}, obj)
 }
--- a/extensions/pkg/controller/utils_test.go
+++ b/extensions/pkg/controller/utils_test.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/gardener/gardener/extensions/pkg/controller"
 	mockclient "github.com/gardener/gardener/pkg/mock/controller-runtime/client"
+	"github.com/gardener/gardener/pkg/operation/botanist"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
@@ -27,6 +28,7 @@ import (
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -268,6 +270,38 @@ var _ = Describe("Utils", func() {
 				State: gardencorev1beta1.LastOperationStateSucceeded,
 			}
 			Expect(controller.IsMigrated(worker)).To(BeTrue())
+		})
+	})
+
+	Describe("#GetObjectByReference", func() {
+		var (
+			ctx = context.TODO()
+			ref = &autoscalingv1.CrossVersionObjectReference{
+				APIVersion: "v1",
+				Kind:       "Secret",
+				Name:       "foo",
+			}
+			namespace = "shoot--test--foo"
+			refSecret = &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      botanist.ReferencedResourcesPrefix + "foo",
+					Namespace: namespace,
+				},
+				Data: map[string][]byte{
+					"foo": []byte("bar"),
+				},
+			}
+		)
+
+		It("should call client.Get and return the result", func() {
+			secret := &corev1.Secret{}
+			c.EXPECT().Get(ctx, client.ObjectKey{Namespace: namespace, Name: botanist.ReferencedResourcesPrefix + ref.Name}, secret).DoAndReturn(
+				func(_ context.Context, _ client.ObjectKey, secret *corev1.Secret) error {
+					refSecret.DeepCopyInto(secret)
+					return nil
+				})
+			Expect(controller.GetObjectByReference(ctx, c, ref, namespace, secret)).To(Succeed())
+			Expect(secret).To(Equal(refSecret))
 		})
 	})
 })
--- a/pkg/apiserver/admission/initializer/types.go
+++ b/pkg/apiserver/admission/initializer/types.go
@@ -22,6 +22,7 @@ import (
 
 	"k8s.io/apiserver/pkg/admission"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
+	"k8s.io/client-go/dynamic"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 )
@@ -62,6 +63,12 @@ type WantsKubeClientset interface {
 	admission.InitializationValidator
 }
 
+// WantsDynamicClient defines a function which sets a dynamic client for admission plugins that need it.
+type WantsDynamicClient interface {
+	SetDynamicClient(dynamic.Interface)
+	admission.InitializationValidator
+}
+
 // WantsAuthorizer defines a function which sets an authorizer for admission plugins that need it.
 type WantsAuthorizer interface {
 	SetAuthorizer(authorizer.Authorizer)
@@ -78,6 +85,8 @@ type pluginInitializer struct {
 
 	kubeInformers kubeinformers.SharedInformerFactory
 	kubeClient    kubernetes.Interface
+
+	dynamicClient dynamic.Interface
 
 	authorizer authorizer.Authorizer
 }
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -127,20 +127,25 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 				return taskIDs
 			}(),
 		})
+		deployReferencedResources = g.Add(flow.Task{
+			Name:         "Deploying referenced resources",
+			Fn:           flow.TaskFn(botanist.DeployReferencedResources).RetryUntilTimeout(defaultInterval, defaultTimeout),
+			Dependencies: flow.NewTaskIDs(deployNamespace),
+		})
 		deployInternalDomainDNSRecord = g.Add(flow.Task{
 			Name:         "Deploying internal domain DNS record",
 			Fn:           flow.TaskFn(botanist.DeployInternalDNS).DoIf(!o.Shoot.HibernationEnabled),
-			Dependencies: flow.NewTaskIDs(waitUntilKubeAPIServerServiceIsReady),
+			Dependencies: flow.NewTaskIDs(deployReferencedResources, waitUntilKubeAPIServerServiceIsReady),
 		})
 		deployExternalDomainDNSRecord = g.Add(flow.Task{
 			Name:         "Deploying external domain",
 			Fn:           flow.TaskFn(botanist.DeployExternalDNS).DoIf(!o.Shoot.HibernationEnabled),
-			Dependencies: flow.NewTaskIDs(waitUntilKubeAPIServerServiceIsReady),
+			Dependencies: flow.NewTaskIDs(deployReferencedResources, waitUntilKubeAPIServerServiceIsReady),
 		})
 		deployInfrastructure = g.Add(flow.Task{
 			Name:         "Deploying Shoot infrastructure",
 			Fn:           flow.TaskFn(botanist.DeployInfrastructure).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(deploySecrets, deployCloudProviderSecret),
+			Dependencies: flow.NewTaskIDs(deploySecrets, deployCloudProviderSecret, deployReferencedResources),
 		})
 		waitUntilInfrastructureReady = g.Add(flow.Task{
 			Name:         "Waiting until shoot infrastructure has been reconciled",
@@ -174,7 +179,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		deployControlPlane = g.Add(flow.Task{
 			Name:         "Deploying shoot control plane components",
 			Fn:           flow.TaskFn(botanist.DeployControlPlane).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(deploySecrets, deployCloudProviderSecret, waitUntilInfrastructureReady),
+			Dependencies: flow.NewTaskIDs(deploySecrets, deployCloudProviderSecret, deployReferencedResources, waitUntilInfrastructureReady),
 		})
 		waitUntilControlPlaneReady = g.Add(flow.Task{
 			Name:         "Waiting until shoot control plane has been reconciled",
@@ -215,7 +220,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		deployControlPlaneExposure = g.Add(flow.Task{
 			Name:         "Deploying shoot control plane exposure components",
 			Fn:           flow.TaskFn(botanist.DeployControlPlaneExposure).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(waitUntilKubeAPIServerIsReady),
+			Dependencies: flow.NewTaskIDs(deployReferencedResources, waitUntilKubeAPIServerIsReady),
 		})
 		waitUntilControlPlaneExposureReady = g.Add(flow.Task{
 			Name:         "Waiting until Shoot control plane exposure has been reconciled",
@@ -250,7 +255,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		computeShootOSConfig = g.Add(flow.Task{
 			Name:         "Computing operating system specific configuration for shoot workers",
 			Fn:           flow.TaskFn(botanist.ComputeShootOperatingSystemConfig).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(initializeShootClients, waitUntilInfrastructureReady),
+			Dependencies: flow.NewTaskIDs(deployReferencedResources, initializeShootClients, waitUntilInfrastructureReady),
 		})
 		deployGardenerResourceManager = g.Add(flow.Task{
 			Name:         "Deploying gardener-resource-manager",
@@ -260,7 +265,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		deployNetworking = g.Add(flow.Task{
 			Name:         "Deploying shoot network plugin",
 			Fn:           flow.TaskFn(botanist.DeployNetwork).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(deployGardenerResourceManager, computeShootOSConfig),
+			Dependencies: flow.NewTaskIDs(deployReferencedResources, deployGardenerResourceManager, computeShootOSConfig),
 		})
 		waitUntilNetworkIsReady = g.Add(flow.Task{
 			Name:         "Waiting until shoot network plugin has been reconciled",
@@ -275,7 +280,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		deployWorker = g.Add(flow.Task{
 			Name:         "Configuring shoot worker pools",
 			Fn:           flow.TaskFn(botanist.DeployWorker).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(deployCloudProviderSecret, waitUntilInfrastructureReady, initializeShootClients, computeShootOSConfig, waitUntilNetworkIsReady),
+			Dependencies: flow.NewTaskIDs(deployCloudProviderSecret, deployReferencedResources, waitUntilInfrastructureReady, initializeShootClients, computeShootOSConfig, waitUntilNetworkIsReady),
 		})
 		waitUntilWorkerReady = g.Add(flow.Task{
 			Name:         "Waiting until shoot worker nodes have been reconciled",
@@ -330,7 +335,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		deployExtensionResources = g.Add(flow.Task{
 			Name:         "Deploying extension resources",
 			Fn:           flow.TaskFn(botanist.DeployExtensionResources).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(initializeShootClients),
+			Dependencies: flow.NewTaskIDs(deployReferencedResources, initializeShootClients),
 		})
 		_ = g.Add(flow.Task{
 			Name:         "Waiting until extension resources are ready",
@@ -355,7 +360,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		deployContainerRuntimeResources = g.Add(flow.Task{
 			Name:         "Deploying container runtime resources",
 			Fn:           flow.TaskFn(botanist.DeployContainerRuntimeResources).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(initializeShootClients),
+			Dependencies: flow.NewTaskIDs(deployReferencedResources, initializeShootClients),
 		})
 		_ = g.Add(flow.Task{
 			Name:         "Waiting until container runtime resources are ready",
--- a/pkg/operation/botanist/resources.go
+++ b/pkg/operation/botanist/resources.go
@@ -1,0 +1,66 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package botanist
+
+import (
+	"context"
+	"fmt"
+
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	"github.com/gardener/gardener/pkg/utils"
+	managedresources "github.com/gardener/gardener/pkg/utils/managedresources"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// ManagedResourceName is the name of the managed resource used to deploy referenced resources to the Seed cluster.
+	ManagedResourceName = "referenced-resources"
+	// ReferencedResourcesPrefix is the prefix used when copying referenced resources to the Shoot namespace in the Seed,
+	// to avoid naming collisions with resources managed by Gardener.
+	ReferencedResourcesPrefix = "ref-"
+)
+
+// DeployReferencedResources reads all referenced resources from the Garden cluster and writes a managed resource to the Seed cluster.
+func (b *Botanist) DeployReferencedResources(ctx context.Context) error {
+	// Read referenced objects into a slice of unstructured objects
+	var unstructuredObjs []*unstructured.Unstructured
+	for _, resourceRef := range b.Shoot.ResourceRefs {
+		// Read the resource from the Garden cluster
+		obj, err := utils.GetObjectByRef(ctx, b.K8sGardenClient.Client(), &resourceRef, b.Shoot.Info.Namespace)
+		if err != nil {
+			return err
+		}
+		if obj == nil {
+			return fmt.Errorf("object not found %v", resourceRef)
+		}
+
+		// Create an unstructured object and append it to the slice
+		unstructuredObj := &unstructured.Unstructured{Object: obj}
+		unstructuredObj.SetNamespace(b.Shoot.SeedNamespace)
+		unstructuredObj.SetName(ReferencedResourcesPrefix + unstructuredObj.GetName())
+		unstructuredObjs = append(unstructuredObjs, unstructuredObj)
+	}
+
+	// Create managed resource from the slice of unstructured objects
+	return managedresources.CreateManagedResourceFromUnstructured(ctx, b.K8sSeedClient.Client(), b.Shoot.SeedNamespace, ManagedResourceName,
+		v1beta1constants.SeedResourceManagerClass, unstructuredObjs, false, nil)
+}
+
+// DestroyReferencedResources deletes the managed resource containing referenced resources from the Seed cluster.
+func (b *Botanist) DestroyReferencedResources(ctx context.Context) error {
+	return client.IgnoreNotFound(managedresources.DeleteManagedResource(ctx, b.K8sSeedClient.Client(), b.Shoot.SeedNamespace, ManagedResourceName))
+}
--- a/pkg/operation/shoot/types.go
+++ b/pkg/operation/shoot/types.go
@@ -25,6 +25,7 @@ import (
 	"github.com/gardener/gardener/pkg/operation/etcdencryption"
 	"github.com/gardener/gardener/pkg/operation/garden"
 
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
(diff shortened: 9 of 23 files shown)


rfranzke:
As discussed via Slack:
/hold
/status author-action
/needs changes

ghost:
@stoyanr Only code owners are allowed to unhold a PR.

rfranzke:
/unhold
/remove status/author-action

@stoyanr can you check the CI failure?
> Unformatted files detected:
pkg/apiserver/admission/initializer/initializer.go

stoyanr:
@rfranzke CI failure fixed

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -135,7 +140,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		deployExternalDomainDNSRecord = g.Add(flow.Task{
 			Name:         "Deploying external domain",
 			Fn:           flow.TaskFn(botanist.DeployExternalDNS).DoIf(!o.Shoot.HibernationEnabled),
-			Dependencies: flow.NewTaskIDs(waitUntilKubeAPIServerServiceIsReady),
+			Dependencies: flow.NewTaskIDs(deployReferencedResources, waitUntilKubeAPIServerServiceIsReady),
```
rfranzke, line 143: Probably we should make every extension resource depending on this task as well, WDYT? Now that we add support for referencing resources from the Garden cluster the extensions could start making use of it. We shouldn't deploy the extension CRs before these referenced resources were synced.

pkg/operation/botanist/resources.go
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/operation/botanist/resources.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/gardener/gardener/pkg/utils"
+)
+
+// DeployReferencedResources reads all referenced resources from the Garden cluster and writes them to the Seed cluster.
+func (b *Botanist) DeployReferencedResources(ctx context.Context) error {
```
rfranzke, line 38: I guess this function lacks the deletion of those resources that were part of the `.spec.resources[]` in a previous shoot version and are now removed.
Probably it makes sense to leverage the Gardener-Resource-Manager deployed in every seed's `garden` namespace. It listens for `ManagedResource`s with `.spec.class=seed`. You could create such an object and let the GRM manage the referred resources in the seed. This way you don't have to remember the resources yourself to know what to clean up.

rfranzke (changes requested):
Very nice PR, thank you very much!
Apart from the inline comments, could you please
1. adapt the [example `Shoot` resource](https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml) with the new `.spec.resources` field.
1. add some short document to https://github.com/gardener/gardener/tree/master/docs/extensions that explains to extension developers how they can use these resource references.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
stoyanr: Done

pkg/operation/botanist/resources.go
stoyanr: Done

pkg/operation/botanist/resources.go
stoyanr: Done

stoyanr:
@rfranzke Thanks for your review! Addressed all inline comments, adapted the example, and added a short document as suggested.

docs/extensions/referenced-resources.md
```diff
@@ -0,0 +1,29 @@
+# Referenced Resources
```
rfranzke, line 1: Nit: Can you add the document to the index, https://github.com/gardener/gardener/blob/master/docs/README.md#extensions?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
rfranzke: The internal DNS domain deployment task is not yet depending on it.

pkg/operation/botanist/resources.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/gardener/gardener/extensions/pkg/controller"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	"github.com/gardener/gardener/pkg/utils"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const (
+	ManagedResourceName = "referenced-resources"
```
rfranzke, line 31: As the const is exported, can you add a doc string?

pkg/operation/botanist/resources.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	// Create managed resource from the slice of unstructured objects
+	return controller.CreateManagedResourceFromUnstructured(ctx, b.K8sSeedClient.Client(), b.Shoot.SeedNamespace, ManagedResourceName,
+		v1beta1constants.SeedResourceManagerClass, unstructuredObjs, false, nil)
+}
+
+// DestroyReferencedResources deletes the managed resource containing referenced resources from the Seed cluster.
+func (b *Botanist) DestroyReferencedResources(ctx context.Context) error {
+	return controller.DeleteManagedResource(ctx, b.K8sSeedClient.Client(), b.Shoot.SeedNamespace, ManagedResourceName)
```
rfranzke: Probably we should gracefully handle the `IsNotFound` error here, right?

```suggestion
	return client.IgnoreNotFound(controller.DeleteManagedResource(ctx, b.K8sSeedClient.Client(), b.Shoot.SeedNamespace, ManagedResourceName))
```

rfranzke (changes requested):
Thanks, very nice, just a few nits.

example/90-shoot.yaml
```diff
@@ -269,3 +270,10 @@ spec:
 # seedSelector:
 #   matchLabels:
 #     foo: bar
+# List resources referenced by providerConfig and other sections, if any
+#  resources:
+#  - name: foobar-secret
+#    resourceRef:
+#      apiVersion: v1
+#      kind: Secret
+#      name: my-foobar-secret
```
timebertt:
```suggestion
# resources:
# - name: foobar-secret
#   resourceRef:
#     apiVersion: v1
#     kind: Secret
#     name: my-foobar-secret
```

pkg/operation/shoot/types.go
```diff
@@ -70,6 +71,8 @@ type Shoot struct {
 	MachineDeployments        []extensionsv1alpha1.MachineDeployment
 
 	ETCDEncryption *etcdencryption.EncryptionConfig
+
+	ResourceRefs map[string]autoscalingv1.CrossVersionObjectReference
```
timebertt, line 75: Why is this a `autoscalingv1.CrossVersionObjectReference`? 
Did I overlook something in the reviews of #2350?
```suggestion
	ResourceRefs map[string]corev1.ObjectReference
```

pkg/operation/botanist/resources.go
```diff
@@ -0,0 +1,60 @@
...
+			return err
+		}
+		if obj == nil {
+			return fmt.Errorf("object not found %v", resourceRef)
+		}
+
+		// Create an unstructured object and append it to the slice
+		unstructuredObj := &unstructured.Unstructured{Object: obj}
+		unstructuredObj.SetNamespace(b.Shoot.SeedNamespace)
+		unstructuredObjs = append(unstructuredObjs, &unstructured.Unstructured{Object: obj})
```
timebertt: Hmmm, can we maybe add some well-known prefix to the name of the resources we copy to the seed?
If the user specifies a resource name, we already use, e.g. Secret `gardener`, we will overwrite this resource, which can lead to problematic issues...

```suggestion
		unstructuredObj := &unstructured.Unstructured{Object: obj}
		unstructuredObj.SetName(SomeWellKnownPrefix + unstructuredObj.GetName())
		unstructuredObj.SetNamespace(b.Shoot.SeedNamespace)
		unstructuredObjs = append(unstructuredObjs, &unstructured.Unstructured{Object: obj})
```

timebertt (commented):
Thanks for the PR.
Looks good in general, but just had a quick look.

pkg/operation/shoot/types.go
rfranzke: I think the reason is that `autoscalingv1.CrossVersionObjectReference` only exposes `APIGroup`, `Kind`, and `Name` while `corev1.ObjectReference` exposes more fields that would remain unused.

stoyanr: @tim-ebert @rfranzke Yes, `CrossVersionObjectReference` fits much better from my perspective, and also it was initially suggested in #2321. It was introduced in #2350. I don't think we would gain much my changing it to a different reference type at this point.

pkg/operation/botanist/resources.go
stoyanr: @tim-ebert @rfranzke I have some doubts. If we add a prefix when copying the resource, then the original reference in the Shoot would not be valid anymore (as it would use the unprefixed name), so any extension controller trying to use it would fail, unless they know to use the prefix. How exactly do you suggest to make this work?

docs/extensions/referenced-resources.md
stoyanr: Done

pkg/operation/botanist/resources.go
stoyanr: Done

pkg/operation/botanist/resources.go
stoyanr: Done, although it's superfluous, the deletion logic inside `DeleteManagedResource` already checks for `NotFound`.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
stoyanr: Done

example/90-shoot.yaml
stoyanr: Done

stoyanr:
@tim-ebert @rfranzke Thanks, rebased and addressed your comments except the last 2, please see my comments.

pkg/operation/shoot/types.go
timebertt: Yep, right, agreed. I just wondered and wanted to double-check.

pkg/operation/botanist/resources.go
timebertt: What about introducing a helper func like `GetObjectByResourceReference` which just takes a resource ref (which can be taken from the Shoot section of the Cluster resource) and returns the relating resource from the Seed? This function can then be used by the extension controllers and basically hides the prefix constant/naming mechanism from the caller, so they don't need the know the specifics...

stoyanr: That could work, except that this helper func would have to return the object as `unstructured.Unstructured`, (or as `map[string]interface{}`) because we can't make any assumptions about its actual type. On the other hand, an extension controller could read the object as a strongly-typed struct, since they can make such assumptions. If were writing an extension controller, I would strongly prefer the strongly-typed option. If you are absolutely sure we need to do this, I would rather write a helper func `GetShootResourceReference(string)` that would return a correct `CrossVersionObjectReferece` by its name in the Shoot resource. I still have doubts however that all this would be useful enough to justify making the life of an extension controller developer harder in one way or another.

timebertt: > I still have doubts however that all this would be useful enough to justify making the life of an extension controller developer harder in one way or another.

The purpose of this change is not to be useful in any way, it's rather a requirement to avoid collision of names, right?

> If were writing an extension controller, I would strongly prefer the strongly-typed option.

I think, this should still be possible. You can make this helper func accept a resource ref and also a `runtime.Object`, so the caller can pass an `*corev1.Secret` for example and it gets filled by the library function (just like the controller-runtime client does it in `client.Get` and `client.List`). This way, it is strongly-typed and the caller just has to check if the resource ref points to a GroupKind that is supported (which they have to do either way) and then pass the ref and a pointer to the respective golang type to the helper func to retrieve the actual object. WDYT? Am I missing something?

docs/README.md
```diff
@@ -86,6 +86,7 @@
   * Generic (non-essential) extensions
     * [`Extension` resource](extensions/extension.md)
 * [Extending project roles](extensions/project-roles.md)
+* [Referened resources](extensions/referenced-resources.md)
```
danielfoehrKn:
```suggestion
* [Referenced resources](extensions/referenced-resources.md)
```

pkg/apiserver/admission/initializer/types.go
```diff
@@ -62,6 +64,18 @@ type WantsKubeClientset interface {
 	admission.InitializationValidator
 }
 
+// WantsDiscoveryClient defines a function which sets a discovery client for admission plugins that need it.
+type WantsDiscoveryClient interface {
+	SetDiscoveryClient(discovery.DiscoveryInterface)
+	admission.InitializationValidator
+}
+
+// WantsDynamicClient defines a function which sets a dynamic client client for admission plugins that need it.
```
danielfoehrKn:
```suggestion
// WantsDynamicClient defines a function which sets a dynamic client for admission plugins that need it.
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -526,6 +543,48 @@ func (r *ReferenceManager) ensureShootReferences(shoot *core.Shoot) error {
...
+
+		// Parse APIVersion to GroupVersion
+		gv, err := schema.ParseGroupVersion(resource.ResourceRef.APIVersion)
+		if err != nil {
+			return err
+		}
+
+		// Check if the resource is namespaced
+		if !apiResource.Namespaced {
+			return errors.New("Shoot cannot reference a resource that is not namespaced")
```
danielfoehrKn:
```suggestion
			return fmt.Errorf("failed to resolve shoot resource reference %q. Cannot reference a resource that is not namespaced", resource.Name)
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -526,6 +543,48 @@ func (r *ReferenceManager) ensureShootReferences(shoot *core.Shoot) error {
...
+			Verb:            "get",
+			APIGroup:        gv.Group,
+			APIVersion:      gv.Version,
+			Resource:        apiResource.Name,
+			Namespace:       shoot.Namespace,
+			Name:            resource.ResourceRef.Name,
+			ResourceRequest: true,
+		}
+		if decision, _, _ := r.authorizer.Authorize(readAttributes); decision != authorizer.DecisionAllow {
+			return errors.New("Shoot cannot reference a resource you are not allowed to read")
```
danielfoehrKn:
```suggestion
			return errors.New("shoot cannot reference a resource you are not allowed to read")
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -526,6 +543,48 @@ func (r *ReferenceManager) ensureShootReferences(shoot *core.Shoot) error {
...
 	}
 
+	for _, resource := range shoot.Spec.Resources {
+		// Get the APIResource for the current resource
+		apiResource, err := r.getAPIResource(resource.ResourceRef.APIVersion, resource.ResourceRef.Kind)
+		if err != nil {
+			return err
+		}
+		if apiResource == nil {
+			return fmt.Errorf("could not find API resource with version %s and kind %s", resource.ResourceRef.APIVersion, resource.ResourceRef.Kind)
```
danielfoehrKn:
```suggestion
			return fmt.Errorf("shoot resource reference %q could not be resolved for API resource with version %q and kind %q", resource.Name, resource.ResourceRef.APIVersion, resource.ResourceRef.Kind)
```

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -526,6 +543,48 @@ func (r *ReferenceManager) ensureShootReferences(shoot *core.Shoot) error {
...
+			Name:            resource.ResourceRef.Name,
+			ResourceRequest: true,
+		}
+		if decision, _, _ := r.authorizer.Authorize(readAttributes); decision != authorizer.DecisionAllow {
+			return errors.New("Shoot cannot reference a resource you are not allowed to read")
+		}
+
+		// Check if the resource actually exists
+		if err := r.lookupResource(gv.WithResource(apiResource.Name), shoot.Namespace, resource.ResourceRef.Name); err != nil {
+			return err
```
danielfoehrKn:
```suggestion
			return fmt.Errorf("failed to resolve shoot resource reference %q: %v", resource.Name, err)
```

danielfoehrKn (commented):
Only minor comments

pkg/operation/botanist/resources.go
stoyanr: OK, did as suggested.

stoyanr:
@rfranzke @tim-ebert @danielfoehrKn Addressed all comments and suggestions once again.

danielfoehrKn (dismissed):
\lgtm
 have not tested it though

extensions/pkg/controller/utils.go
```diff
@@ -363,3 +371,16 @@ func IsMigrated(obj runtime.Object) bool {
...
 }
+
+// GetObjectByReference gets an object by the given reference, in the given namespace.
+// If the object kind doesn't match the given reference kind this will result in an error.
+func GetObjectByReference(ctx context.Context, c client.Client, ref *autoscalingv1.CrossVersionObjectReference, namespace string, obj runtime.Object) error {
+	gvk, err := utils.GVKFromCrossVersionObjectReference(ref)
+	if err != nil {
+		return err
+	}
+	if !reflect.DeepEqual(obj.GetObjectKind().GroupVersionKind(), gvk) {
```
timebertt: This will not work, as `GetObjectKind()` returns an empty GVK for empty objects, e.g. `&corev1.Secret{}` (which is probably, what the caller wants to pass to this helper function):

```go
It("will error", func() {
	err := controller.GetObjectByReference(nil, nil, &autoscalingv1.CrossVersionObjectReference{
		APIVersion: "v1",
		Kind:       "Secret",
		Name:       "foo",
	}, "shoot--test--foo", &corev1.Secret{})
	Expect(err).To(MatchError(ContainSubstring("doesn't match reference kind")))
})
```

What you need here, is actually a `runtime.Scheme`, which can give you the GVK for any registered golang representation of an API type.
However, I would not do such a check here and rather directly call `c.Get` with the provided `runtime.Object`. The caller has to check, if the resource reference has the expected type anyways and otherwise `c.Get` will return an error for you.
So you can basically trim down this function to:
```go
func GetObjectByReference(ctx context.Context, c client.Client, ref *autoscalingv1.CrossVersionObjectReference, namespace string, obj runtime.Object) error {
	return c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: ReferencedResourcesPrefix + ref.Name}, obj)
}
```

pkg/operation/botanist/resources.go
```diff
@@ -0,0 +1,63 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package botanist
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/gardener/gardener/extensions/pkg/controller"
```
timebertt: Hmm, can we avoid references from `pkg` to `extensions/pkg`, please?
You can maybe move the functions you need to somewhere in `pkg` and then reference them from `extensions/pkg` again...

pkg/operation/botanist/resources.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		if obj == nil {
+			return fmt.Errorf("object not found %v", resourceRef)
+		}
+
+		// Create an unstructured object and append it to the slice
+		unstructuredObj := &unstructured.Unstructured{Object: obj}
+		unstructuredObj.SetNamespace(b.Shoot.SeedNamespace)
+		unstructuredObj.SetName(controller.ReferencedResourcesPrefix + unstructuredObj.GetName())
+		unstructuredObjs = append(unstructuredObjs, &unstructured.Unstructured{Object: obj})
```
timebertt: Why are you constructing a new unstructured object? Ideally, you don't need to do that at any place and just operate on the same object in place completely from the call where you obtain it from a client all the way through (maybe with a `DeepCopy()` in between).
```suggestion
		unstructuredObjs = append(unstructuredObjs, unstructuredObj)
```

extensions/pkg/controller/utils.go
```diff
@@ -47,6 +49,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	// ReferencedResourcesPrefix is the prefix used when copying referenced resources to the Shoot namespace in the Seed,
+	// to avoid naming collisions with resources managed by Gardener.
+	ReferencedResourcesPrefix = "ref-"
```
timebertt: I think, this const should rather be in `pkg/operation/botanist` or somewhere under `pkg`.

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -570,3 +629,23 @@ func (r *ReferenceManager) lookupSecret(namespace, name string) error {
 
 	return nil
 }
+
+func (r *ReferenceManager) getAPIResource(groupVersion, kind string) (*metav1.APIResource, error) {
+	resources, err := r.discoveryClient.ServerResourcesForGroupVersion(groupVersion)
```
timebertt: You don't need an extra discovery client for that, it is already part of `kubernetes.Interface`.
Can you remove `r.discoveryClient` and the corresponding methods and fields in the initializer.?
```suggestion
	resources, err := r.kubeClient.Discovery().ServerResourcesForGroupVersion(groupVersion)
```

plugin/pkg/global/resourcereferencemanager/admission_test.go
```diff
@@ -194,6 +225,15 @@ var _ = Describe("resourcereferencemanager", func() {
 			fakeAuthorizer = fakeAuthorizerType{}
 			admissionHandler.SetAuthorizer(fakeAuthorizer)
 
+			discoveryClient = &discoveryfake.FakeDiscovery{Fake: &testing.Fake{Resources: discoveryClientResources}}
+			admissionHandler.SetDiscoveryClient(discoveryClient)
+
+			scheme = runtime.NewScheme()
+			_ = corev1.AddToScheme(scheme)
```
timebertt: Please don't ignore errors in tests:
```suggestion
			Expect(corev1.AddToScheme(scheme)).To(Succeed())
```

plugin/pkg/global/resourcereferencemanager/admission_test.go
```diff
@@ -529,6 +571,36 @@ var _ = Describe("resourcereferencemanager", func() {
...
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Seeds().Informer().GetStore().Add(&seed)).To(Succeed())
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().SecretBindings().Informer().GetStore().Add(&secretBinding)).To(Succeed())
+				Expect(kubeInformerFactory.Core().V1().ConfigMaps().Informer().GetStore().Add(&configMap)).To(Succeed())
+
+				user := &user.DefaultInfo{Name: "disallowed-user"}
+				attrs := admission.NewAttributesRecord(&shoot, nil, core.Kind("Shoot").WithVersion("version"), shoot.Namespace, shoot.Name, core.Resource("shoots").WithVersion("version"), "", admission.Create, &metav1.CreateOptions{}, false, user)
+
+				err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+				Expect(err).To(HaveOccurred())
```
timebertt:
```suggestion
				Expect(err).To(MatchError("shoot cannot reference a resource you are not allowed to read"))
```

plugin/pkg/global/resourcereferencemanager/admission_test.go
```diff
@@ -529,6 +571,36 @@ var _ = Describe("resourcereferencemanager", func() {
...
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().Seeds().Informer().GetStore().Add(&seed)).To(Succeed())
+				Expect(gardenCoreInformerFactory.Core().InternalVersion().SecretBindings().Informer().GetStore().Add(&secretBinding)).To(Succeed())
+				Expect(kubeInformerFactory.Core().V1().ConfigMaps().Informer().GetStore().Add(&configMap)).To(Succeed())
+
+				user := &user.DefaultInfo{Name: allowedUser}
+				attrs := admission.NewAttributesRecord(&shoot, nil, core.Kind("Shoot").WithVersion("version"), shoot.Namespace, shoot.Name, core.Resource("shoots").WithVersion("version"), "", admission.Create, &metav1.CreateOptions{}, false, user)
+
+				err := admissionHandler.Admit(context.TODO(), attrs, nil)
+
+				Expect(err).To(HaveOccurred())
```
timebertt:
```suggestion
				Expect(err).To(MatchError(ContainSubstring("failed to resolve shoot resource reference")))
```

docs/extensions/referenced-resources.md
```diff
@@ -0,0 +1,29 @@
...
+      secretRef: foobar-secret
+  resources:
+  - name: foobar-secret
+    resourceRef:
+      apiVersion: v1
+      kind: Secret
+      name: my-foobar-secret
+```
+
+Gardener expects to find these referenced resources in the project namespace (e.g. `garden-dev`) and will copy them to the Shoot namespace in the Seed cluster when reconciling a Shoot. Extension controllers can resolve the references to these resources by accessing the Shoot via the `Cluster` resource.
```
timebertt: Please adapt the docs according to your changes with the prefix and the helper function (`GetObjectByReference`).

timebertt (dismissed):
Finally looked through the PR in detail. Sorry for the late review.
I left you some comments here and there.

plugin/pkg/global/resourcereferencemanager/admission_test.go
stoyanr: `Expect(err).To(HaveOccurred())` is the way errors are checked in this test (and many similar tests). I think checking specific error messages makes tests more brittle, as such messages may change. I would also keep error checking consistent throughout the file.

plugin/pkg/global/resourcereferencemanager/admission_test.go
stoyanr: See above.

stoyanr:
@tim-ebert Thanks for your review! While I do appreciate the detailed and thoughtful feedback, and I am definitely going to implement the changes you suggested, I do have a few concerns:

* This PR has been opened for about 2 weeks now. 8 days ago, I got a review with "only a few minor comments" from @rfranzke, which I addressed. Then I got a review from you, with comments that are not quite minor. I implemented all changes you requested, but now I get another review, with many comments not related to the last changes that I did. I don't think this is how it's supposed to work. 
* Some of your suggestions, such as your request to change `Expect(err).To(HaveOccurred())` are not consistent with the way things are done in this particular test. There is no way for me to suspect that you or somebody else might want me to do this differently from the rest of the code. From my perspective, such changes, if they should be introduced at all, should be done in a consistent way in separate "refactoring" PRs. Otherwise, one PR would do it one way and another in a different way, resulting in code that's inconsistent and harder to maintain. 

@rfranzke We've been having internal discussions in the Sofia team why PRs take sometimes so long to merge. Based on my experience here, I would like to have a broader discussion on the topic, as I do see some room for improvement beyond Sofia.

pkg/operation/botanist/resources.go
stoyanr: There is no reason, I just overlooked this, thanks for noticing!

pkg/operation/botanist/resources.go
stoyanr: I wasn't aware about this. Is this documented somewhere? I would very much like to be able to avoid such "forbidden" dependencies myself, what's the best way to go about it besides feedback during a review?

stoyanr: @tim-ebert When taking a second look, I am not really sure how to do this, and it's actually why I did it like this in the first place. The reconciliation flow calls the newly introduced `botanist.DeployReferencedResources` and `botanist.DestoryManagedResources` functions. They have to be in `botanist` I think, similarly to most other functions called from the flow. If I move them to `extensions/pkg/controller`, then the dependencies to that package will pop-up in `gardenlet/controller/shoot`, which is not much better. These 2 functions in turn call `controller.CreateManagedResourceFromUnstructured` and `controller.DeleteManagedResources` to deploy referenced resources as a managed resource, as requested by @rfranzke. So I would need to move the entire `managedresources.go` from `extensions/pkg/controller` to somewhere under `pkg`, and I am not sure this is such a great idea. Please advise.

extensions/pkg/controller/utils.go
stoyanr: This const can't be moved unless we get rid of the dependency from `pkg` to `extensions/pkg` (see above), since this would lead to circular dependencies.

extensions/pkg/controller/utils.go
stoyanr: Done

docs/extensions/referenced-resources.md
stoyanr: Done

stoyanr:
@tim-ebert Included all your feedback except the ones regarding dependencies from `pkg` to `extensions/pkg`, see my comments above.

docs/extensions/referenced-resources.md
```diff
@@ -0,0 +1,46 @@
...
+      apiVersion: v1
+      kind: Secret
+      name: my-foobar-secret
+```
+
+Gardener expects to find these referenced resources in the project namespace (e.g. `garden-dev`) and will copy them to the Shoot namespace in the Seed cluster when reconciling a Shoot, adding a prefix to their names to avoid naming collisions with Gardener's own resources. 
+
+Extension controllers can resolve the references to these resources by accessing the Shoot via the `Cluster` resource. To properly read a referenced resources, extension controllers should use the utility function `GetObjectByReference` from the `extensions/pkg/controller` package, for example:
+
+```go
```
timebertt, line 33: Thanks, very nice!

extensions/pkg/controller/utils_test.go
```diff
@@ -270,4 +271,36 @@ var _ = Describe("Utils", func() {
...
+					Name:      controller.ReferencedResourcesPrefix + "foo",
+					Namespace: namespace,
+				},
+				Data: map[string][]byte{
+					"foo": []byte("bar"),
+				},
+			}
+		)
+
+		It("will error", func() {
```
timebertt:
```suggestion
		It("should correctly get the referenced resource", func() {
```

extensions/pkg/controller/utils_test.go
```diff
@@ -270,4 +271,36 @@ var _ = Describe("Utils", func() {
...
+			secret := &corev1.Secret{}
+			c.EXPECT().Get(ctx, client.ObjectKey{Namespace: namespace, Name: controller.ReferencedResourcesPrefix + ref.Name}, secret).DoAndReturn(
+				func(_ context.Context, _ client.ObjectKey, secret *corev1.Secret) error {
+					*secret = *refSecret
```
timebertt: I think `DeepCopy`/`DeepCopyInto` is probably the cleaner way to copy API objects.
```suggestion
					refSecret.DeepCopyInto(secret)
```

pkg/operation/botanist/resources.go
rfranzke: > I wasn't aware about this. Is this documented somewhere? I would very much like to be able to avoid such "forbidden" dependencies myself, what's the best way to go about it besides feedback during a review?

No, it's not documented anywhere, as the gardener-extensions library got only merged into g/g recently, and we probably have to learn first how to deal with this new structure. See also https://github.com/gardener/gardener/pull/2251#discussion_r435203263 where I just yesterday made the same comment. If you like you can certainly start a new document under the `docs` folder that describes such things, that would certainly be helpful, also for outside contributors. Though, generally, I'm afraid we'll never reach the situation where everything is perfectly documented because the code structure isn't static, so there such suggestions might occur in the future as well, I assume.

> So I would need to move the entire managedresources.go from extensions/pkg/controller to somewhere under pkg, and I am not sure this is such a great idea. Please advise.

Hm, I don't know of anything that would speak against moving it from `extensions/pkg` to `pkg`. Now that the Gardener core itself is also more and more relying on the `ManagedResource`s it probably is fair to move it.

rfranzke:
Not sure what @tim-ebert's opinion is, here my thoughts about your comments:

> This PR has been opened for about 2 weeks now. 8 days ago, I got a review with "only a few minor comments" from @rfranzke, which I addressed.

Well, I agree that having an open PR for "so long" is far from being optimal, but I don't know how to help. Most people of @gardener/gardener-maintainers are busy/overloaded and are reviewing PRs as good as possible and as time/workload allows. More than once I asked also other contributors for help with reviewing the PRs, but so far it only improved slightly. We simply need more people that are actively looking into the PRs so that the review process gets better distributed.

> Then I got a review from you, with comments that are not quite minor. I implemented all changes you requested, but now I get another review, with many comments not related to the last changes that I did. I don't think this is how it's supposed to work.

What do you suggest, instead? I actually liked that not only me was checking the PR, that's a good thing as different people spot different things. I also don't think that continuous reviews should be limited to only those changes that were recently added. Personally, I'm often only spotting things after some iterations and not directly when I'm firstly looking at it. Also, with every new commit/change there is potential that things got added/removed that are not optimal and need to be changed again. Isn't that the natural process of a PR? However, I agree that this cannot be done endlessly as - as a contributor - you want to get your PR in within a reasonable time frame/review cycles. But I don't think we have to discuss about this yet - yes, the PR was opened for 14 days (mainly for the reasons mentioned above), but when checking the review cycles I don't think we are there yet to discuss that the PR is open for too long.
Anyways, do you have some concrete ideas/proposals what to improve? How should it work from your point of view?

> Some of your suggestions, such as your request to change Expect(err).To(HaveOccurred()) are not consistent with the way things are done in this particular test. There is no way for me to suspect that you or somebody else might want me to do this differently from the rest of the code.

Agreed, this happens in most PRs and will probably always happen. I don't see something "bad" with it - if you don't think the suggestion is valid then simply comment on it with your rationale and ask the reviewer to check again, perfectly fine.
Also, I think it will never be the case that a contributor can suspect/know these things in advance. It is also not necessary. That's where reviews help. Yes, with good documentation you can minimize the review cycles, but then again you have to maintain this documentation. In the end, I think it should be balanced in a healthy way.

About the comments: Naturally, they are individual per reviewer and even then they are not always consistent. I'm just thinking about myself and my reviews. I'm trying to, but I doubt that when you'd compare all my comments it would always be consistent.
So yes, agreed that nitpicking can be avoided as much as possible, or at least reviewed without "request changes" mode.
It's also true that not everything in the code is perfectly consistent, especially in tests. Of course, striving for it makes sense, but given the large number of contributors and size of the project I doubt that such perfect consistency will ever be achieved.

> From my perspective, such changes, if they should be introduced at all, should be done in a consistent way in separate "refactoring" PRs. Otherwise, one PR would do it one way and another in a different way, resulting in code that's inconsistent and harder to maintain.

Agreed, makes sense. Though, I don't consider it too harmful to suggest such things in open PRs - if you don't want to implement it because it would be even more inconsistent then simply comment on it and that's it.

stoyanr:
@rfranzke Thanks for your answers. While I understand the reasons you give, I don't subscribe yet to "everything's perfect and should be as it is". Especially seeing that's not only this PR, and knowing that I am not the only one frustrated. I do have some ideas, but I would propose to discuss this in a wider round. The comments to this particular PR are not the proper place IMO.

Don't get me wrong, I am not saying we have fundamental issues, and I completely agree that most of the comments / feedback to this PR have been very helpful. I do think however that there's room for improvement. Hopefully you don't mind to continue this discussion.

pkg/operation/botanist/resources.go
stoyanr: There are probably quite a number of external dependencies (in extensions, etc.) to these `controller` functions for dealing with managed resources. If I move them to `pkg`, all of that would need to be adapted next time Gardener is revendored. Is this ok?
Also, any particular place under `pkg` that you would suggest? None of the existing packages seems like a great fit to me. Perhaps a new package under `utils`, e.g. `utils/managedresources`?

stoyanr: Another thing: in `managedresources.go` there are a few references to other functions from `extensions/pkg`, that are used also elsewhere. I will have to move them as well. I am afraid that may trigger a "chain reaction" of moving. Hopefully this is ok.

rfranzke:
Certainly, nothing will ever be perfect. :) and sure, if there are concrete proposal/ideas to discuss then please feel free to setup a meeting.

pkg/operation/botanist/resources.go
rfranzke: Yes, something like `pkg/utils/managedresources` sounds good to me. As for the adaptations of the package paths in extensions, it should be okay as extensions have to actively revendor to be affected. We can probably make them aware by adding a `action developer` release note to this PR.
Regarding the other functions, I'm not exactly sure which of them you mean, but probably it's okay as well to move them and adapt the calls accordingly.
Generally, I think we still have a lot of stuff in `extensions/pkg` that could be moved to `pkg` now, so any small iteration on refactorings like this are appreciated! We simply "copied" the gardener/gardener-extensions repo to gardener/gardener and didn't spend any time on such refactorings yet, unfortunately.

timebertt:
Thanks @rfranzke for also expressing your thoughts on the matter, I completely agree with your comments.

Still, here are some additional thoughts of mine, I hope that you, @stoyanr, can understand:
> This PR has been opened for about 2 weeks now. 8 days ago, I got a review with "only a few minor comments" from @rfranzke, which I addressed. Then I got a review from you, with comments that are not quite minor.

I totally understand, that you are frustrated by having your PR open for so long with no or only some progress. As @rfranzke already mentioned, most of the maintainers here are quite busy and so was I at the time you opened this PR. Still, I got curious about your PR and wanted to see, if you already considered the thing with regard to conflicting names. As I didn't have time to look into the PR in full detail, I still wanted to let you know of my concerns upfront instead of letting you wait even longer for feedback, that should be addressed (which is actually one of your concerns, I see). But as I didn't have the time to fully check your PR, I just "commented" on your PR instead of blocking it by "requesting changes". But I see, you and @rfranzke also agreed on addressing the issue, so I assume, this was valid feedback. If you have any suggestions, what could be done differently to handle such a situation, I'm open for it.

> I implemented all changes you requested, but now I get another review, with many comments not related to the last changes that I did. I don't think this is how it's supposed to work.

Nevertheless, I wanted to take the time to check your PR again in detail and I see, I made some suggestions that were agreed on improving by you and others (see @rfranzke's reactions), which other reviewers didn't catch in the first place. I definitely didn't want to be mean by requesting these changes so late after the PR was opened, so I apologized for joining late. But as my suggestions were accepted, I think they were valid and don't see anything "bad" about it.

> From my perspective, such changes, if they should be introduced at all, should be done in a consistent way in separate "refactoring" PRs. Otherwise, one PR would do it one way and another in a different way, resulting in code that's inconsistent and harder to maintain

Regarding my suggestions, that were inconsistent in the way with how things are done in the other test cases: I didn't notice that and would have been completely fine with you commenting on them to let me know about it and stating that this should be addressed separately or whatever. That were only minor suggestions anyways, so I actually didn't mean to be nitpicking. Sorry, if you perceived it that way.

> We've been having internal discussions in the Sofia team why PRs take sometimes so long to merge. Based on my experience here, I would like to have a broader discussion on the topic, as I do see some room for improvement beyond Sofia.

We are always happy to have other people reviewing as well, as we are kind of short-handed. So if you believe, PRs take too long to be reviewed and merged, you are more than welcome to lend a hand here and there and help with reviews.
But I see, there were reviews requested also from other people, that seemingly didn't have the time to check the PR either (though, it was not only the case in this PR but is more a general notion sometimes).
So if you indeed have ideas and suggestions on how to improve the PR and review process, I would be happy to discuss them.

Bottom line: I hope you can understand my points. I am definitely open to discuss these issues, if someone has concrete ideas/suggestions.

timebertt (commented):
> @tim-ebert Included all your feedback except the ones regarding dependencies from `pkg` to `extensions/pkg`, see my comments above.

Thanks, I checked again, looks really good now. Only two small nits in the tests.
Wrt to the dependencies thing, I would also go for `pkg/utils/managedresources`, this sounds good to me.
Once you have adapted, I will take another look and give a test run :)

pkg/operation/botanist/resources.go
stoyanr: OK, I tried to do as suggested. I put that in a separate commit so you could evaluate the impact more easily. I moved most managed resources utility functions to `pkg/utils/managedresources` as suggested, in fact all of them except `RenderChartAndCreateManagedResource` because that one uses `chart.Render` and moving that would result in a much bigger refactoring, which I believe is better left outside the scope of this PR. There are no more dependencies from `pkg` to `pkg/extensions`.

extensions/pkg/controller/utils.go
stoyanr: Done after getting rid of the dependency, see above.

extensions/pkg/controller/utils_test.go
stoyanr: Noticed that and already changed it.

stoyanr:
@tim-ebert @rfranzke Everything addressed now I think, please check again.

pkg/utils/object.go
```diff
@@ -120,7 +120,7 @@ func DeleteObject(ctx context.Context, c client.Client, gvk schema.GroupVersionK
 	return client.IgnoreNotFound(c.Delete(ctx, obj))
 }
 
-func gvkFromCrossVersionObjectReference(ref *autoscalingv1.CrossVersionObjectReference) (schema.GroupVersionKind, error) {
+func GVKFromCrossVersionObjectReference(ref *autoscalingv1.CrossVersionObjectReference) (schema.GroupVersionKind, error) {
```
rfranzke: Doc string is missing for exported function

rfranzke (commented):
As discussed via Slack, changes look good to me.
Only a validation for duplicate resource ref names is missing, i.e., following configuration should not be accepted by the gardener-apiserver:

```yaml
 91   resources:
 92   - name: foo
 93     resourceRef:
 94       apiVersion: v1
 95       kind: Secret
 96       name: bar
 97   - name: foo
 98     resourceRef:
 99       apiVersion: v1
100       kind: Secret
101       name: baz
```

Other than that, works perfectly in my environment, played with it. Thanks for the PR.

pkg/utils/object.go
stoyanr: Made this function unexported again as it's actually not used outside the package anymore.

stoyanr:
@rfranzke Added the missing validation, and a test for it. Hopefully it should be ok now.

rfranzke (dismissed):
/lgtm

rfranzke (approved):

PR #2364, opened 2020-05-20
Prevent Shoot to be reconciled twice on creation

ialidzhikov:

/area quality
/kind bug
/priority normal

Currently we observe Shoots reconciling twice on creation. 

In the shootUpdate func, we ignore updates when `metadata.generation == status.observedGeneration`.
Ref https://github.com/gardener/gardener/blob/23a965d14aec956bd35fd614d45caf2979e9780c/pkg/gardenlet/controller/shoot/shoot_control.go#L74-L82

Currently in `reconcileShoot` func we have 2 update requests that are executed when the Shoot status.observedGeneration is 0 - the update request for the finalizer and the status update request for status.uid.

`reconcileShoot` func returns after the status.uid update request and in this way "exits early" and allows immediate 2nd reconciliation happening from the update request (in this way prevents reconcile after the flow).
https://github.com/gardener/gardener/blob/23a965d14aec956bd35fd614d45caf2979e9780c/pkg/gardenlet/controller/shoot/shoot_control.go#L467-L476

This PR applies the same for the finalizer update in the beginning of `reconcileShoot` func and in this way prevents additional reconcile after the flow execution.

Fixes #

**Release note**:

```improvement user
An issue causing an additional reconciliation after successful flow execution for newly created Shoots is now fixed.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control.go
@@ -42,7 +42,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/pointer"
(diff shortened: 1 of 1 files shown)


pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -407,22 +407,24 @@ func (c *Controller) reconcileShoot(logger *logrus.Entry, shoot *gardencorev1bet
...
-		if finalizers.Has(gardencorev1beta1.GardenerName) {
-			return curShoot, nil
-		}
+	if !controllerutils.HasFinalizer(shoot, gardencorev1beta1.GardenerName) {
+		// need retry logic, because the scheduler is acting on it at the same time and cached object might not be up to date
+		_, err := kutil.TryUpdateShoot(c.k8sGardenClient.GardenCore(), retry.DefaultBackoff, shoot.ObjectMeta, func(curShoot *gardencorev1beta1.Shoot) (*gardencorev1beta1.Shoot, error) {
+			finalizers := sets.NewString(curShoot.Finalizers...)
+			if finalizers.Has(gardencorev1beta1.GardenerName) {
+				return curShoot, nil
+			}
```
rfranzke: Can you get rid of this check now in favour of line 410? It looks a little confusing. Alternatively, replace the whole 412-422 with https://github.com/gardener/gardener/blob/9ee208f92d0e3d560f4ee64f688c3a4355b19f20/pkg/controllerutils/finalizers.go#L34 - WDYT?

rfranzke (commented):
Thanks for digging into it, nice finding!

rfranzke (approved):
/lgtm
/needs second-opinion
/cc @vpnachev

vpnachev (approved):
/lgtm

PR #2377, opened 2020-05-25
Add missing worker predicates for state sync controller

rfranzke:

/area control-plane-migration
/kind bug
/priority normal

Earlier, the `Worker` state sync controller wasn't applying extension type predicates to the mapped `Worker` objects.  This led to the fact that a controller, actually responsible for only type `foo`, was modifying the `.status.state` of `Worker` objects having another type.
This PR fixes the problem by adding the type predicate to the mapper.

/cc @vlvasilev @swilen-iwanow @plkokanov 

**Release note**:

```improvement operator
NONE
```


timuthy (approved):
/lgtm

PR #2378, opened 2020-05-25
Add utility function for cleaning up leaked worker `ClusterRole`s

### LINKED ISSUE #2144: Clean up Leaked ClusterRole Resources
Previously we had leaking cluster roles in the seed clusters
```yaml
extensions.gardener.cloud:{{ .Values.providerName }}:{{ .Release.Namespace }}:machine-controller-manager
```

It was fixed for newly created shoots with https://github.com/gardener/gardener-extensions/pull/378, but a clean-up logic to delete the orphaned cluster roles was not implemented.

Note, there is a logic for the deletion of every chart resource, but it relies that the template contains the real name of the resource with properly evaluated place holders, which is not the case for the machine-controller-manager Cluster Role:
https://github.com/gardener/gardener/blob/de93728c2e1aaca721f73c01dbadb232f9d5f861/pkg/utils/chart/chart.go#L124-L157

rfranzke:

/area quality
/kind cleanup
/priority normal

In an early state of the `Worker` extension we had a bug that was leaking `ClusterRole`s for the machine-controller-manager. It was fixed with https://github.com/gardener-attic/gardener-extensions/pull/378 but the leaked resources weren't cleaned up yet.
This PR adds a utility function to the generic `Worker` actuator package that can be called by affected extension controllers. It cleans up those orphaned resources.

Fixes #2144

**Release note**:

```improvement developer
A utility function for cleaning up orphaned `ClusterRole`s for the `machine-controller-manager` was added to the `extensions/pkg/controller/worker/genericactuator` package.
```

### DIFF
--- a/extensions/pkg/controller/backupentry/reconciler.go
+++ b/extensions/pkg/controller/backupentry/reconciler.go
@@ -99,7 +99,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 
 	if extensionscontroller.IsShootFailed(shoot) {
-		r.logger.Info("Stop reconciling BackupEntry of failed Shoot.", "namespace", shootTechnicalID, "name", be.Name)
+		r.logger.Info("Stop reconciling BackupEntry of failed Shoot.", "name", be.Name)
 		return reconcile.Result{}, nil
 	}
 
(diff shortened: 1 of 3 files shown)


timuthy (approved):
/lgtm

PR #2379, opened 2020-05-25
Improve update event predicates for ControllerInstallationRequired controller

### LINKED ISSUE #2332: Gardenlet `ControllerInstallationRequired` controller does not consider DNS Providers
**What happened**:

The Gardenlet does not reconcile the registered DNS Providers in the federated seed controller.

Gardenlet should reconcile the DNS Provider to include it in the decision making whether a ControllerInstallation is required or not.

The DNSProvider CRD cannot be enqueued as an extension object.

[This commit ](https://github.com/gardener/gardener/commit/fc5c52001d98a5f1ffe8ab73e731a2cd868b5e89) cosmetically fixed the occuring panic in the Gardenlet.
The panic happens because the [DNSProvider is being registered as an extension object](https://github.com/gardener/gardener/blob/1bb22255b2142b59b7e94f2f67e320666fd7be2e/pkg/gardenlet/controller/federatedseed/extensions/artifact.go#L65-L64) - though it is none. 

**Environment**:

- Gardener version:
- Others:

/area disaster-recovery
/area control-plane-migration
/kind bug
/priority normal

rfranzke:

/area quality
/kind enhancement
/priority normal
/size s
/topology seed

This PR improves the update event predicate inside the `ControllerInstallationRequired` controller to allow object-tailored predicate functions during the object registration.

Fixes #2332

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/federatedseed/extensions/artifact.go
+++ b/pkg/gardenlet/controller/federatedseed/extensions/artifact.go
@@ -22,7 +22,6 @@ import (
 
 	dnsv1alpha1 "github.com/gardener/external-dns-management/pkg/apis/dns/v1alpha1"
 	dnsinformers "github.com/gardener/external-dns-management/pkg/client/dns/informers/externalversions"
-	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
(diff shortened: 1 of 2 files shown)


danielfoehrKn (approved):
/lgtm

Thanks!

**Unrelated to this PR**
Although I understand that the DNS Controller Manager and its resources are not closely tied to Gardener, it not being part of the extensions api-group causes some extra logic needed here. It cannot be treated in a uniform way as all the other extension resources.
Imagine there are other resources besides the DNSProvider that are relevant for the ControllerInstallation - we would need extra code for each single one of them. [This condition would become quite long](https://github.com/gardener/gardener/blob/46a88c15db5397c573eba6f90f7cdba03327f361/pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go#L68-L67).

rfranzke:
I totally agree with you, @danielfoehrKn. I was never a fan of having the DNS types separately, but I understand @mandelsoft argument that the dns-controller-manager can and is being used totally independent of Gardener. If you have a good idea for a solution combining these both requirements I would be more than happy to "fix" this.

PR #2381, opened 2020-05-25
Metal-Pod was renamed to metal-stack

majst01:
/area documentation
/kind cleanup
/priority normal

Metal-Pod is no longer valid, renamed to metal-stack. Fix documentation.


rfranzke (approved):
/lgtm

thanks :)

PR #2384, opened 2020-05-25
Tolerations for Shoots and Projects

### LINKED ISSUE #401: Configurable list for namespaces which may use protected seeds
The Seed spec allows to configure `.spec.protected={true,false}` which means that only Shoots that have been deployed to the `garden` namespace may use this respective Seed cluster.

We should make this list of namespaces configurable, either globally (in the configuration of the SeedManager admission plugin) or per Seed (maybe in the resource directly, maybe also in the admission plugin's configuration).

### LINKED ISSUE #2193: Project Tolerations to support limiting access to Seed usage

rfranzke:

/area usability
/area documentation
/kind enhancement
/priority normal

This PR adds support for tolerations for `Shoot`s, similar to tolerations for `Pod`s.

Also, it supports specifying a whitelist + defaults for on `Project` level. Only users with the `modify-spec-tolerations-whitelist` custom verb are allowed to change the whitelist. This is validated by a new `CustomVerbAuthorizer` admission plugin.

The new `ShootTolerationRestriction` admission plugin validates that only whitelisted tolerations might be used by shoots. It also defaults them according to the default tolerations configured in the `Project`. It also allows globally configured defaults + whitelist with its own admission plugin configuration.

I added some documentation about all the existing admission plugins along the way.

Fixes #2193
Fixes #401

**Release note**:

```noteworthy user
`Shoot` resources do now support specifying tolerations for taints on seeds in the `.spec.tolerations` field. Only tolerations that were whitelisted by the corresponding `Project`'s `.spec.tolerations.whitelist` field, or by the global configuration (controlled by Gardener administrators) are allowed to be used. Please read more about it in [this document](https://github.com/gardener/gardener/blob/master/docs/usage/tolerations.md).
```
```improvement operator
A new [document](https://github.com/gardener/gardener/blob/master/docs/concepts/apiserver_admission_plugins.md) providing an overview of the existing in-tree admission plugins in Gardener has been added.
```


docs/concepts/apiserver_admission_plugins.md
```diff
@@ -0,0 +1,113 @@
+# Admission Plugins
+
+Similar to the kube-apiserver, the gardener-apiserver comes with a few in-tree managed admission plugins.
+If you want to get an overview of the what and why of admission plugins then [this document](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/) might be a good start.
+
+This document enlists all existing admission plugins with a short explanation of what it is responsible for.
```
timuthy:
```suggestion
This document lists all existing admission plugins with a short explanation of what it is responsible for.
```

pkg/apis/core/helper/helpers.go
```diff
@@ -133,6 +134,41 @@ func TaintsHave(taints []core.SeedTaint, key string) bool {
...
+		if taint.Key == core.DeprecatedSeedTaintDisableDNS || taint.Key == core.DeprecatedSeedTaintInvisible || taint.Key == core.DeprecatedSeedTaintDisableCapacityReservation {
+			continue
+		}
+		relevantTaints = append(relevantTaints, taint)
+	}
+
+	if len(relevantTaints) == 0 {
+		return true
+	}
+	if len(tolerations) == 0 {
```
timuthy:
```suggestion
	if len(relevantTaints) > len(tolerations) {
```

pkg/apis/core/helper/helpers_test.go
```diff
@@ -93,6 +93,96 @@ var _ = Describe("helper", func() {
...
+			[]core.SeedTaint{
+				{Key: "foo"},
+				{Key: "bar", Value: pointer.StringPtr("baz")},
+			},
+			[]core.Toleration{
+				{Key: "foo", Value: pointer.StringPtr("bar")},
+				{Key: "bar", Value: pointer.StringPtr("foo")},
+			},
+			false,
+		),
```
timuthy, line 183:
```suggestion
		Entry("taints > tolerations",
			[]core.SeedTaint{
				{Key: "foo"},
				{Key: "bar", Value: pointer.StringPtr("baz")},
			},
			[]core.Toleration{
				{Key: "bar", Value: pointer.StringPtr("baz")},
			},
			false,
		),
	),
```

pkg/apis/core/v1alpha1/defaults.go
```diff
@@ -181,6 +191,10 @@ func SetDefaults_Shoot(obj *Shoot) {
 		p := ShootPurposeEvaluation
 		obj.Spec.Purpose = &p
 	}
+
+	if obj.Namespace == v1beta1constants.GardenNamespace {
+		addTolerations(&obj.Spec.Tolerations, Toleration{Key: SeedTaintProtected})
```
timuthy, line 199: Can you add a comment which explains the purpose of this action? I think it is or will not be obvious in the future.

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -211,6 +213,41 @@ func TaintsHave(taints []gardencorev1beta1.SeedTaint, key string) bool {
...
+		if taint.Key == gardencorev1beta1.DeprecatedSeedTaintDisableDNS || taint.Key == gardencorev1beta1.DeprecatedSeedTaintInvisible || taint.Key == gardencorev1beta1.DeprecatedSeedTaintDisableCapacityReservation {
+			continue
+		}
+		relevantTaints = append(relevantTaints, taint)
+	}
+
+	if len(relevantTaints) == 0 {
+		return true
+	}
+	if len(tolerations) == 0 {
```
timuthy:
```suggestion
	if len(relevantTaints) > len(tolerations) {
```

plugin/pkg/global/customverbauthorizer/admission.go
```diff
@@ -0,0 +1,135 @@
...
+		if !ok {
+			return apierrors.NewBadRequest("could not convert old resource into Project object")
+		}
+	}
+
+	if apiequality.Semantic.DeepEqual(oldObj.Spec.Tolerations, obj.Spec.Tolerations) {
+		return nil
+	}
+
+	return c.authorize(ctx, a, "modify-spec-tolerations-whitelist", "modify .spec.tolerations.whitelist")
```
timuthy: `modify-spec-tolerations-whitelist`, can you make it a `const`?

plugin/pkg/shoot/tolerationrestriction/admission.go
```diff
@@ -0,0 +1,252 @@
...
+	return nil
+}
+
+func (t *TolerationRestriction) admitShoot(shoot *core.Shoot) error {
+	project, err := admissionutils.GetProject(shoot.Namespace, t.projectLister)
+	if err != nil {
+		return apierrors.NewBadRequest(fmt.Sprintf("could not find referenced project: %+v", err.Error()))
+	}
+
+	// do not change shoot tolerations if they specify a key already
```
timuthy: Rather move comment to line 173?

pkg/apis/core/helper/helpers.go
```diff
@@ -133,6 +134,41 @@ func TaintsHave(taints []core.SeedTaint, key string) bool {
...
+	taintKeyValues := sets.NewString()
+	for _, taint := range relevantTaints {
+		taintKeyValues.Insert(utils.IDForKeyWithOptionalValue(taint.Key, taint.Value))
+	}
+
+	for _, toleration := range tolerations {
+		if taintKeyValues.Has(toleration.Key) {
+			continue
+		}
+		if !taintKeyValues.Has(utils.IDForKeyWithOptionalValue(toleration.Key, toleration.Value)) {
```
timuthy: This means that all taints must contain all tolerations but it should be the other way around, shouldn't it? At the moment, a seed is rejected if the shoot has more tolerations than the seed has taints.

timuthy (changes requested):
Nice PR! I left a few comments.

pkg/apis/core/helper/helpers.go
rfranzke: Very nice catch, thanks for finding it, indeed. I've inverted the logic as you outlined and added another test case for it.

timuthy (commented):
I tested your changes again and it works very well! One restriction we should keep in mind is that because we only write the default tolerations during the creation, a `kubectl apply` with new tolerations will delete the ones we defaulted during creation.

For example:
1. Add defaults to project
2. `kubectl apply shoot.yaml` w/o tolerations
3. Ensure that default tolerations were added to shoot resource
4. Edit `shoot.yaml` and add a new (white listed) toleration
5. `kubectl apply shoot.yaml`
6. Observe that the default toleration is gone.

Unfortunately, `patchStrategy:"merge" patchMergeKey:"key"` struct tags don't help either since `kubectl apply` doesn't use the [strategic merge patch](https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/#notes-on-the-strategic-merge-patch). Don't consider it a blocker but I wanted to hear your opinion or maybe you have an idea how to circumvent it.

rfranzke:
Thanks for testing! First of all, we should probably add the patch strategy anyways, right?
As for the situation you explained: I would think this expected behaviour, or? If you specify tolerations yourself (like you did in step 4) then those are being considered the relevant ones. The defaults are only added if you don't specify any toleration at all.

timuthy:
Thanks for your thoughts. As discussed, let's handle it the way you proposed it 👍

timuthy (dismissed):
/lgtm

timuthy (approved):
/lgtm

PR #2396, opened 2020-05-27
Add flag 'allowPrivilegedContainers' to the Shoot creation framework

danielfoehrKn:

/area testing
/kind enhancement
/priority normal

Add flag 'allowPrivilegedContainers' to the Shoot creation framework.
This is needed to test Shoot creation having .Spec.Kubernetes.allowPrivilegedContainers = false. (defaults to true).

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/test/framework/shootcreationframework.go
+++ b/test/framework/shootcreationframework.go
@@ -33,31 +33,33 @@ var shootCreationCfg *ShootCreationConfig
 type ShootCreationConfig struct {
 	GardenerConfig *GardenerConfig
 
-	shootKubeconfigPath          string
-	seedKubeconfigPath           string
-	testShootName                string
-	testShootPrefix              string
-	shootMachineImageName        string
-	shootMachineType             string
-	shootMachineImageVersion     string
-	cloudProfile                 string
-	seedName                     string
-	shootRegion                  string
-	secretBinding                string
-	shootProviderType            string
-	shootK8sVersion              string
-	externalDomain               string
-	workerZone                   string
-	networkingPods               string
-	networkingServices           string
-	networkingNodes              string
-	startHibernatedFlag          string
-	startHibernated              bool
-	infrastructureProviderConfig string
-	controlPlaneProviderConfig   string
-	networkingProviderConfig     string
-	workersConfig                string
-	shootYamlPath                string
+	shootKubeconfigPath           string
+	seedKubeconfigPath            string
+	testShootName                 string
+	testShootPrefix               string
+	shootMachineImageName         string
+	shootMachineType              string
+	shootMachineImageVersion      string
+	cloudProfile                  string
+	seedName                      string
+	shootRegion                   string
+	secretBinding                 string
+	shootProviderType             string
+	shootK8sVersion               string
+	externalDomain                string
+	workerZone                    string
+	networkingPods                string
+	networkingServices            string
+	networkingNodes               string
+	startHibernatedFlag           string
+	startHibernated               bool
+	allowPrivilegedContainersFlag string
+	allowPrivilegedContainers     *bool
+	infrastructureProviderConfig  string
+	controlPlaneProviderConfig    string
+	networkingProviderConfig      string
+	workersConfig                 string
+	shootYamlPath                 string
 }
 
 // ShootCreationFramework represents the shoot test framework that includes
@@ -131,6 +133,14 @@ func validateShootCreationConfig(cfg *ShootCreationConfig) {
 			ginkgo.Fail("startHibernated is not a boolean value")
 		}
 		cfg.startHibernated = parsedBool
+	}
+
+	if StringSet(cfg.allowPrivilegedContainersFlag) {
+		parsedBool, err := strconv.ParseBool(cfg.allowPrivilegedContainersFlag)
+		if err != nil {
+			ginkgo.Fail("allowPrivilegedContainers is not a boolean value")
+		}
+		cfg.allowPrivilegedContainers = &parsedBool
 	}
 
 	if !StringSet(cfg.infrastructureProviderConfig) {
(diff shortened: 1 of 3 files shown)


test/framework/shootcreationframework.go
```diff
@@ -300,6 +318,8 @@ func RegisterShootCreationFrameworkFlags() *ShootCreationConfig {
 	flag.StringVar(&newCfg.networkingServices, "networking-services", "", "the spec.networking.services to use for this shoot. Optional.")
 	flag.StringVar(&newCfg.networkingNodes, "networking-nodes", "", "the spec.networking.nodes to use for this shoot. Optional.")
 	flag.StringVar(&newCfg.startHibernatedFlag, "start-hibernated", "", "the spec.hibernation.enabled to use for this shoot. Optional.")
+	flag.StringVar(&newCfg.allowPrivilegedContainersFlag, "allow-privileged-containers", "", "the spec.kubernetes.allowPrivilegedContainers to use for this shoot. Optional.")
```
timebertt:
```suggestion
	flag.StringVar(&newCfg.allowPrivilegedContainersFlag, "allow-privileged-containers", "", "the spec.kubernetes.allowPrivilegedContainers to use for this shoot. Optional, defaults to true.")
```

timebertt (commented):
Can you add the flag to the `CreateShoot` TestDef after this line?
https://github.com/gardener/gardener/blob/a6d9dd94feec9031587d07b5df09ab682d4ec6c7/.test-defs/CreateShoot.yaml#L34

Otherwise, lgtm

schrodit (approved):
lgtm

schrodit (changes requested):
the flag should be exposed in the testdef to be configurable in the testmachinery
https://github.com/gardener/gardener/blob/master/.test-defs/CreateShoot.yaml#L12

schrodit (approved):
lgtm

timebertt (approved):
Nice, thanks for taking care!
/lgtm

test/framework/shootcreationframework.go
```diff
@@ -300,6 +318,8 @@ func RegisterShootCreationFrameworkFlags() *ShootCreationConfig {
 	flag.StringVar(&newCfg.networkingServices, "networking-services", "", "the spec.networking.services to use for this shoot. Optional.")
 	flag.StringVar(&newCfg.networkingNodes, "networking-nodes", "", "the spec.networking.nodes to use for this shoot. Optional.")
 	flag.StringVar(&newCfg.startHibernatedFlag, "start-hibernated", "", "the spec.hibernation.enabled to use for this shoot. Optional.")
+	flag.StringVar(&newCfg.allowPrivilegedContainersFlag, "allow-privileged-containers", "", "the spec.kubernetes.allowPrivilegedContainers to use for this shoot. Optional, defaults to true.")
```
ialidzhikov, line 321: Why we have `allowPrivilegedContainersFlag` and `allowPrivilegedContainers`?
Can't we just have `flag.BoolVar` for `allowPrivilegedContainers`?

ialidzhikov (changes requested):
for the comment above

test/framework/shootcreationframework.go
danielfoehrKn: I asked myself the same question. 
I concluded that the problem is the following: 
 - flag.BoolVar cannot take nil as the default value -> the flag value is thus defaulted to true
 - the [shoot resource template file](https://github.com/gardener/gardener/blob/d5ece2ec84428b8cfd5082af49e7aaa25a7f2b62/test/framework/resources/templates/default-shoot.yaml#L1)  specifies the Spec.Kubernetes. allowPrivilegedContainers=false
 - without setting the flag, the value in the shoot template resource will always be overridden. This means that you would have to know that you have to specify the --allow-privileged-containers=false flag for the test execution in order to be able to set the field to false. 

With the current implementation, the template values are honored if the flag is not set during test execution. If the flag is set, it will override the template values. 

@tim-ebert thats why you designed it this way I assume?

ialidzhikov: > the shoot resource template file specifies the Spec.Kubernetes. allowPrivilegedContainers=false

The template file does not specify `allowPrivilegedContainers`, it is rather defaulted to `true` on apiserver side when not explicitly specified.

danielfoehrKn: I am talking about the case when the template file (you can set a custom one) does specify the flag.

schrodit: the problem here is the testmachinery integration.
It uses an environment variable to set the value: https://github.com/gardener/gardener/pull/2396/files#diff-77ce71c9c5a09d09add3eb7fd4df8d8cR35

If the env var is not set, the flag would be set to `""` which is not a valid boolean although you want to handle it as `not set`.
Therefore we need a string that we can set to false if `== ""` and otherwise parse the boolean.

timebertt: Ah, now I remember. Thanks for clarifying @schrodit!

ialidzhikov: I see now. Thanks for the explanation.
I was thinking about having optional arg handling in the test definition. But it has the drawback that the test definition is getting harder to read with similar addition:

```bash
OPTIONAL_ARGS=()
if [[ ! -z "$ALLOW_PRIVILEGED_CONTAINERS" ]]; then
  OPTIONAL_ARGS+=(-allow-privileged-containers="$ALLOW_PRIVILEGED_CONTAINERS")
fi

if [[ ! -z "$FOO" ]]; then
  OPTIONAL_ARGS+=(-foo="$FOO")
fi

go test -timeout=0 -mod=vendor ./test/system/shoot_creation
    --v -ginkgo.v -ginkgo.progress -ginkgo.noColor
    -verbose=debug
    -kubecfg=$TM_KUBECONFIG_PATH/gardener.config
    -seed-kubecfg-path=$TM_KUBECONFIG_PATH/seed.config
    -shoot-kubecfg-path=$TM_KUBECONFIG_PATH/shoot.config
    # ...
    -start-hibernated=$START_HIBERNATED
    ${OPTIONAL_ARGS[@]}
```

Specifying the default value in the test definition is also not desired as then we would end with multiple places that specifies default:

```bash
ALLOW_PRIVILEGED_CONTAINERS=${ALLOW_PRIVILEGED_CONTAINERS:-true}

go test -timeout=0 -mod=vendor ./test/system/shoot_creation
    --v -ginkgo.v -ginkgo.progress -ginkgo.noColor
    -verbose=debug
    -kubecfg=$TM_KUBECONFIG_PATH/gardener.config
    -seed-kubecfg-path=$TM_KUBECONFIG_PATH/seed.config
    -shoot-kubecfg-path=$TM_KUBECONFIG_PATH/shoot.config
    # ...
    -start-hibernated=$START_HIBERNATED
    -allow-privileged-containers=$ALLOW_PRIVILEGED_CONTAINERS
```

I don't like much the current handling with flag.StringVar for a bool but I don't have something better to suggest for this PR.

ialidzhikov (approved):

PR #2397, opened 2020-05-27
Be more conservative with errors of 3rd party components

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal

The change with #2336 reveals sporadic errors too fast. This PR changes the behaviour and reverts to a more conservative approach when evaluating errors of 3rd party components.

**Release note**:

```improvement operator
The handling when evaluating errors from third-party controllers/components is more conservative now. It will wait until a certain grace period is exceed before actually revealing and reporting the error to the `Shoot` resource.
```

### DIFF
--- a/pkg/operation/common/extensions.go
+++ b/pkg/operation/common/extensions.go
@@ -44,6 +44,7 @@ func WaitUntilExtensionCRReady(
 	namespace string,
 	name string,
 	interval time.Duration,
+	severeThreshold time.Duration,
 	timeout time.Duration,
 	postReadyFunc func(runtime.Object) error,
 ) error {
@@ -57,6 +58,7 @@ func WaitUntilExtensionCRReady(
 		namespace,
 		name,
 		interval,
+		severeThreshold,
 		timeout,
 		postReadyFunc,
 	)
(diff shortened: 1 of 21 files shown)


pkg/utils/retry/retry_test.go
```diff
@@ -260,6 +262,17 @@ var _ = Describe("Retry", func() {
 		})
 	})
 
+	DescribeTable("#MinorOrSevereError",
+		func(retryCountUntilSevere, threshold int, err error, matcher types.GomegaMatcher) {
+			done, err := MinorOrSevereError(retryCountUntilSevere, threshold, err)
+			Expect(done).To(matcher)
+			Expect(err).To(BeIdenticalTo(err))
```
timebertt:
```suggestion
			done, err2 := MinorOrSevereError(retryCountUntilSevere, threshold, err)
			Expect(done).To(matcher)
			Expect(err2).To(BeIdenticalTo(err))
```

pkg/operation/common/extensions.go
```diff
@@ -68,26 +70,36 @@ func WaitUntilObjectReadyWithHealthFunction(
...
 	name string,
 	interval time.Duration,
+	severeThreshold time.Duration,
 	timeout time.Duration,
 	postReadyFunc func(runtime.Object) error,
 ) error {
-	var lastObservedError error
+	var (
+		errorWithCode         *gardencorev1beta1helper.ErrorWithCodes
+		lastObservedError     error
```
timebertt, line 85: `lastObservedError` is never set 🤔

timebertt (commented):
Nits, otherwise lgtm

pkg/operation/common/extensions.go
rfranzke: Nice catch, it got accidentally removed in #2336

timebertt (approved):
Thanks for fixing!
/lgtm

PR #2400, opened 2020-05-28
[GEP-7] Adds correct LastOperationType to extension CRs

plkokanov:

/area control-plane-migration
/kind bug
/priority normal

The restore function of the infrastructure controller was improperly setting the lastOperationType to Migrate after a successful restore.

None

None

**Release note**:

```improvement operator
The lastOperation of the infrastructure resource after successful restoration is no longer set to Migrate=Success.
```

### DIFF
--- a/extensions/pkg/controller/infrastructure/reconciler.go
+++ b/extensions/pkg/controller/infrastructure/reconciler.go
@@ -114,7 +114,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	case infrastructure.DeletionTimestamp != nil:
 		return r.delete(r.ctx, infrastructure, cluster)
 	case infrastructure.Annotations[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationRestore:
-		return r.restore(r.ctx, infrastructure, cluster, operationType)
+		return r.restore(r.ctx, infrastructure, cluster)
 	default:
 		return r.reconcile(r.ctx, infrastructure, cluster, operationType)
 	}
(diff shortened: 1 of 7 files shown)


extensions/pkg/controller/infrastructure/reconciler.go
```diff
@@ -230,7 +230,7 @@ func (r *reconciler) restore(ctx context.Context, infrastructure *extensionsv1al
 		return reconcile.Result{}, fmt.Errorf("%s: %+v", msg, err)
 	}
 
-	err := r.updateStatusSuccess(ctx, infrastructure, gardencorev1beta1.LastOperationTypeMigrate, EventInfrastructureRestoration, "Successfully restored infrastructure")
+	err := r.updateStatusSuccess(ctx, infrastructure, operationType, EventInfrastructureRestoration, "Successfully restored infrastructure")
```
ialidzhikov: Is the `operationType` passed to restore func is `LastOperationTypeRestore`. Checking the `ComputeOperationType` func, it is not the case, I think. Don't we want to set it to `LastOperationTypeRestore` instead?

plkokanov: @vlvasilev what was the initial requirement to have `operationType` passed to the restore/reconcile functions instead of just hard coding the operation type? Do we need to have the operationType as `Create` during the initial restoration?

vlvasilev: There was no `LastOperationTypeRestore` and at the time this PR been made there was no need for it. Simply the addition of this constant would serve only as information string back then.

ialidzhikov: What about now? Do we want to use `LastOperationTypeRestore` or use `operationType` which I think is something different from `LastOperationTypeRestore`?

plkokanov: Generally we want to use LastOperationTypeRestore. I'll adapt the PR accordingly but it might take some time to test it out.

plkokanov:
Modified the lastOperation to now only be Restore for the restore function. I still need to revendor, build the extensions locally and test.

rfranzke (approved):
/lgtm

plkokanov:
Just tested this together with #2404 and #2359 and managed to run a successful migration

swilen-iwanow (approved):
lgtm

ialidzhikov (approved):
/lgtm

PR #2402, opened 2020-05-28
Fix retaining existing pod annotations in GSAC

timuthy:

/area control-plane
/kind bug
/priority normal

This PR fixes the patch calculation for pods belonging to a **test shoot** in the GSAC. These pods are supposed to be annotated with `"fluentbit.io/exclude": "true"` (see #2320). However, the previously calculated patch caused any existing annotations of a Pod to be dropped if an update (create) request didn't already contain the `"fluentbit.io/exclude"` annotation.

**Release note**:

```improvement operator
An issues has been resolved which caused missing annotations for control plane pods of test shoots (`.spec.purpose: testing`).
```

### DIFF
--- a/pkg/seedadmission/pods.go
+++ b/pkg/seedadmission/pods.go
@@ -88,23 +88,22 @@ func mutateShootControlPlanePodAnnotations(ctx context.Context, c client.Client,
 		"fluentbit.io/exclude": "true",
 	}
 
-	for key, value := range annotationsToAdd {
-		if originalAnnotations == nil || originalAnnotations[key] == "" {
-			patch = append(patch, jsonpatch.JsonPatchOperation{
-				Operation: "add",
-				Path:      "/metadata/annotations",
-				Value: map[string]string{
-					key: value,
-				},
-			})
-		} else {
+	if originalAnnotations != nil {
+		for key, value := range annotationsToAdd {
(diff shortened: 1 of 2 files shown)


vlvasilev:
@timuthy How can I reproduce it? No pods initially are set with this annotation although the pods get annotate successfully keeping the other annotation. I have create a busybox pod with annotation without `"fluentbit.io/exclude": "true"` and after creation the pod kept his initial annotation plus the fluent-bit related one.

rfranzke:
@vlvasilev The problem is that if other annotations exist already then your old code will overwrite the whole `annotations` section of the pod. You can try creating a pod with an annotation `foo=bar`. After your webhook kicks in then the `foo=bar` will be removed. @timuthy change fixes this. :)

/assign @vlvasilev

vlvasilev:
@rfranzke Sorry. I guess I was looking the automatically added annotations during the test and since in the documentation says that `add` works as an insert operation I didn't notice the bug.

vlvasilev (approved):
/lgtm

ghost:
@vlvasilev Only code owners may change review labels.

ghost:
@vlvasilev Only code owners may change review labels.

ialidzhikov (approved):
/lgtm

ialidzhikov:
Nice catch @timuthy . Thank you for fixing!

PR #2404, opened 2020-05-29
[GEP-7] Enhance migration, fix BackupEntry annotation bug

swilen-iwanow:

/area control-plane-migration
/kind enhancement
/kind bug
/priority normal

Enhances the migration flow by adding steps for: 
- Kubernetes API Server scale-up, in case of hibernated Shoot
- Annotating BackupEntry for migration (there was a bug that was annotating every BackupEntry with migrate in the Seed)

Fixes partially #1631  

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -346,7 +346,7 @@ func (b *Botanist) ScaleGardenerResourceManagerToOne(ctx context.Context) error 
 	return kubernetes.ScaleDeployment(ctx, b.K8sSeedClient.Client(), kutil.Key(b.Shoot.SeedNamespace, v1beta1constants.DeploymentNameGardenerResourceManager), 1)
 }
 
-// PrepareKubeAPIServerForMigration deletes the kube-apiserverand deletes its hvpa
+// PrepareKubeAPIServerForMigration deletes the kube-apiserver and deletes its hvpa
 func (b *Botanist) PrepareKubeAPIServerForMigration(ctx context.Context) error {
 	if err := b.K8sSeedClient.Client().Delete(ctx, &hvpav1alpha1.Hvpa{ObjectMeta: metav1.ObjectMeta{Name: v1beta1constants.DeploymentNameKubeAPIServer, Namespace: b.Shoot.SeedNamespace}}); client.IgnoreNotFound(err) != nil && !meta.IsNoMatchError(err) {
 		return err
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -247,6 +247,24 @@ func (b *Botanist) WakeUpControlPlane(ctx context.Context) error {
 	return nil
 }
 
+// WakeUpKubeAPIServer creates a service and ensures API Server is scaled up
+func (b *Botanist) WakeUpKubeAPIServer(ctx context.Context) error {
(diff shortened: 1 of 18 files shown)


pkg/registry/core/shoot/strategy.go
```diff
@@ -172,6 +168,11 @@ func (shootStatusStrategy) PrepareForUpdate(ctx context.Context, obj, old runtim
 	newShoot := obj.(*core.Shoot)
 	oldShoot := old.(*core.Shoot)
 	newShoot.Spec = oldShoot.Spec
+
+	if lastOperation := newShoot.Status.LastOperation; lastOperation != nil && lastOperation.Type == core.LastOperationTypeRestore &&
```
rfranzke, line 172: Why was that moved out of the `mustIncreaseGeneration` function?

rfranzke (commented):
Generally looks good, just one question.

/invite @plkokanov @vlvasilev @stoyanr

pkg/registry/core/shoot/strategy.go
swilen-iwanow: In the previous function shoot.oldStatus is set to the shoot.newStatus meaning lastOperation was old not Restore=Pending, so it needed one more update in order to increase the generation.

rfranzke: I see, thank you!

swilen-iwanow:
@vlvasilev could you drop a review, mainly for the worker actuator : )

rfranzke (dismissed):
/lgtm
/needs second-opinion

pkg/operation/botanist/controlplane.go
```diff
@@ -321,12 +346,13 @@ func (b *Botanist) ScaleGardenerResourceManagerToOne(ctx context.Context) error
 	return kubernetes.ScaleDeployment(ctx, b.K8sSeedClient.Client(), kutil.Key(b.Shoot.SeedNamespace, v1beta1constants.DeploymentNameGardenerResourceManager), 1)
 }
 
-// PrepareControlPlaneDeploymentsForMigration scales the kube-apiserver, kube-controller-manager and gardener-resource-manager to zero and deletes the hvpa for the kube-apiserver
-func (b *Botanist) PrepareControlPlaneDeploymentsForMigration(ctx context.Context) error {
+// PrepareKubeAPIServerForMigration deletes the kube-apiserverand deletes its hvpa
```
vlvasilev:
```suggestion
// PrepareKubeAPIServerForMigration deletes the kube-apiserver and deletes its hvpa
```

pkg/operation/botanist/controlplane.go
```diff
@@ -321,12 +346,13 @@ func (b *Botanist) ScaleGardenerResourceManagerToOne(ctx context.Context) error
...
+func (b *Botanist) PrepareKubeAPIServerForMigration(ctx context.Context) error {
```
vlvasilev, line 350: Maybe it will be better to put the deletion of the hvpa in separated function as the deletion of the kupe-apiserver deployment

swilen-iwanow: I am not sure what will be the actual benefit except that it could be reused in future. But I think in most cases deleting the KubeAPI server will require cleaning the hvpa as well, so you will need to add both steps again? WDYT?

vlvasilev (approved):
/LGTM

rfranzke (approved):

Issue #2120, opened 2020-03-31
Worker pool with swap enabled


vasu1124:
**What would you like to be added**:
Add the option to order worker nodes with swap **enabled**:
* start kubelet with `--fail-swap-on=false`
* bootstrap to configure variable amount of swap space on root disk
* enable swap
* taint node

**Why is this needed**:
The ABAP Server makes use of extended (shared) memory that typically is sized beyond available memory and can deal with swap effectively. The feature is required for cost efficiency.

We are aware of https://github.com/kubernetes/kubernetes/issues/53533. 
The interference with required system components/pods with high priority should be taken into consideration.

vpnachev:
Could you elaborate these steps
> * bootstrap to configure variable amount of swap space on root disk
> * taint node

We are seeing users complaining that the local disk is out of space, so how we can take space of this disk for the swap partition? How much it should be? Maybe this should be exposed in the API.

I believe a label is the better option, not taint. Or maybe both, but with what effect should be the taint?

>  The ABAP Server makes use of extended (shared) memory that typically is sized beyond available memory and can deal with swap effectively. The feature is required for cost efficiency.

This does not looks to me very container friendly. How much such containers can be running on a single node?


PS. I am afraid that this feature might introduce a lot of trouble for both - Gardener users and operators, but it makes sense to be evaluated.

vasu1124:
How much swap? The official recommendation is  2x RAM upto 320GB, depending on the physical RAM:
https://launchpad.support.sap.com/#/notes/1597355
I think in this case, we would add an additional dedicated disk for swap. 

> This does not looks to me very container friendly. How much such containers can be running on a single node?

Indeed, it is a special use case. You can run multiple ABAP servers/processes on a single node, their parallel usage is tested even with disk/swap pressure, so packaged in containers should work too. But it will cause friction with other workloads that expect guarantees.
That is why I suggest to taint the nodes to ensure that only swap-friendly workload (with a toleration) gets deployed on that worker pool. See https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/ .

majst01:
Adding my comments here even if not involved in this special topic.
I just want to add my $0.02 from long term operations experience.
You will pay operational burden with infrastructure savings!
I am not sure how big your savings on the infrastructure side will be compared to the operational pain you will get. It is almost impossible to solve all heap fightings automatically, things will go bad, really bad, and bonus, you will get terrible disk performace as bonus.

[...]

PR #2409, opened 2020-06-02
Allow setting failSwapOn flag in the shoot.yaml file

saggir:
How to categorize this PR?

/area performance
/area documentation
/kind enhancement
/priority normal

What this PR does / why we need it:
This PR adds support for failSwapOn flag in the shoot yaml file, as part of the kubelet section

Today this parameter is set to true by default and there is no way to change it, adding the support in the shoot yaml will alow changing it to "false".
The default value is still "true"

From Kubernetes documentation: 
--fail-swap-on:
Makes the Kubelet fail to start if swap is enabled on the node. (default true) (DEPRECATED: This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.)

Release note:
`Shoot resource now supports changing the failSwapOn flag under the kubelet section, default=true`


pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -626,6 +626,9 @@ type KubeletConfig struct {
 	// +optional
 	// Default: 1m
 	ImagePullProgressDeadline *metav1.Duration `json:"imagePullProgressDeadline,omitempty" protobuf:"bytes,12,opt,name=imagePullProgressDeadline"`
+	// FailSwapOn makes the Kubelet fail to start if swap is enabled on the node. (default true).
+	// +optional
+	FailSwapOn *bool `json:"failSwapOn,omitempty" protobuf:"varint,13,opt,name=failSwapOn"`
```
rfranzke, line 631: Should we default it to `true` in the `defaults.go`?

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -623,6 +623,9 @@ type KubeletConfig struct {
 	// +optional
 	// Default: 1m
 	ImagePullProgressDeadline *metav1.Duration `json:"imagePullProgressDeadline,omitempty" protobuf:"bytes,12,opt,name=imagePullProgressDeadline"`
+	// FailSwapOn makes the Kubelet fail to start if swap is enabled on the node. (default true).
+	// +optional
+	FailSwapOn *bool `json:"failSwapOn,omitempty" protobuf:"varint,13,opt,name=failSwapOn"`
```
rfranzke, line 628: Should we default it to `true` in the `defaults.go`?

rfranzke (changes requested):
Thanks for the PR!

1. Can you please fix the file permissions?
1. Can you please squash the commits?

vpnachev:
Fixes #2120 ?

pkg/apis/core/v1alpha1/types_shoot.go
saggir: I dont think it is needed, it behaves the same as CPUCFSQuota, which dont have an entry in the defaults.go file

pkg/apis/core/v1beta1/types_shoot.go
saggir: I dont think it is needed, it behaves the same as CPUCFSQuota, which dont have an entry in the defaults.go file

saggir:
replaced with a cleaner PR - https://github.com/gardener/gardener/pull/2411

PR #2410, opened 2020-06-02
Fix hibernation order

timuthy:

/area control-plane
/kind bug
/priority normal

This PR makes Gardener delete DNS entries only after the the control plane has been hibernated. Before, the reconcile flow might stuck during `initializeShootClients` because the DNS entries were gone too early.

**Release note**:

```improvement operator
An issue has been fixed which prevented clusters from hibernating successfully.
```


timebertt (approved):
/lgtm

rfranzke (approved):

PR #2411, opened 2020-06-02
Adding support for  fail-swap-on flag in shoot.yaml

saggir:
How to categorize this PR?

/area performance
/area documentation
/kind enhancement
/priority normal

What this PR does / why we need it:
This PR adds support for failSwapOn flag in the shoot yaml file, as part of the kubelet section

Today this parameter is set to true by default and there is no way to change it, adding the support in the shoot yaml will alow changing it to "false".
The default value is still "true"

From Kubernetes documentation:
--fail-swap-on:
Makes the Kubelet fail to start if swap is enabled on the node. (default true) (DEPRECATED: This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.)

Release note:
```improvement user
The `Shoot` resource now supports changing the `failSwapOn` flag under the kubelet section, default is `true`.
```

Part of #2120


mvladev (dismissed):
Add defaulting to `true` in https://github.com/gardener/gardener/blob/master/pkg/apis/core/v1beta1/defaults.go and https://github.com/gardener/gardener/blob/master/pkg/apis/core/v1alpha1/defaults.go

Also - tons of files have changed their permissions to `755` from `644` and it's not needed.

vpnachev:
And please, update this PR do not open new one.

rfranzke (commented):
Thanks! PR as such looks good, though, I see unrelated changes in
* charts/images.yaml
* docs/README.md
* vendor/modules.txt

Can you check? Also, please squash the commits into a single one.

/reviewed ok-to-test

rfranzke:
Also, the CLA must be signed.

saggir:
Regarding the default values, I followed the way you handle the kubelet parameter "cpuCFSQuota", it is not contained in the defaults file...should I add it to the defaults.go file anyway?

rfranzke:
Personally, I have no strong opinion given the current situation that you explained (some fields are not defaulted as well). But @mvladev is right that it's better to do it. So, at least for the new fields we introduce we should probably do it, and then, in a separate PR, do the same for the old fields.

saggir:
I added the defaults, how do I sign the CLA?

pkg/apis/core/v1alpha1/defaults.go
```diff
@@ -181,6 +181,10 @@ func SetDefaults_Shoot(obj *Shoot) {
 		p := ShootPurposeEvaluation
 		obj.Spec.Purpose = &p
 	}
+
+	if obj.Spec.Kubernetes.Kubelet.FailSwapOn == nil {
```
rfranzke, line 205: A small unit test in the `defaults_test.go` so that we don't accidentally break the API in the future would be great.

pkg/apis/core/v1beta1/defaults.go
```diff
@@ -181,6 +181,10 @@ func SetDefaults_Shoot(obj *Shoot) {
 		p := ShootPurposeEvaluation
 		obj.Spec.Purpose = &p
 	}
+
+	if obj.Spec.Kubernetes.Kubelet.FailSwapOn == nil {
```
rfranzke, line 205: A small unit test in the `defaults_test.go` so that we don't accidentally break the API in the future would be great.

rfranzke:
You have to open https://cla-assistant.io/gardener/gardener?pullRequest=2411 and confirm, I guess.
Also, please squash the commits into a single one.

pkg/apis/core/v1beta1/defaults.go
saggir: I dont see a test for the SetDefaults_Shoot func... am I missing something?

rfranzke: Hm, right, I've only added it as part of https://github.com/gardener/gardener/pull/2384/files#diff-86a47c7826e138a4538e48eb597c8846R267, so maybe you can follow the same approach and just add a test for your `failSwapOn` field for now.

pkg/apis/core/v1alpha1/defaults.go
saggir: added

pkg/apis/core/v1beta1/defaults.go
saggir: added

rfranzke:
Thanks for signing the CLA and adding the test.
The changes look good to me. Could you just revert the file permission changes for the `example/90-shoot.yaml` file? Also, another squash + rebase is necessary to make this PR mergeable. Thank you!

/reviewed ok-to-test

rfranzke (dismissed):
/lgtm

rfranzke:
Please check the failing test and fix it: https://concourse.ci.gardener.cloud/builds/233356
You can also run them locally by `make test`.

saggir:
What is the issue with the verify method? the error is not clear, I generated and commited the files...

rfranzke:
@saggir you can run `make generate` locally and check the difference with `git diff`.
Concretely, please get rid of the unnecessary line breaks [here](https://github.com/gardener/gardener/pull/2411/files#diff-3e5e3d0abe429f772fd10a0a2dfd6fb6R195) and [here](https://github.com/gardener/gardener/pull/2411/files#diff-e82afa9c12ffaffab8e0dbac2d24c20eR195).

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -330,7 +333,6 @@ func (b *Botanist) deployOperatingSystemConfigsForWorker(ctx context.Context, ma
 		workerConfig["cri"] = map[string]interface{}{
 			"name": worker.CRI.Name,
 		}
-		downloaderConfig["cri"] = workerConfig["cri"]
```
rfranzke: Hm, not sure why this change happened, the line was added as part of the recently merged #2421 and should not be deleted as part of this PR. Can you please revert this line and rebase again to master?

saggir: strange, I will fix it...

rfranzke:
Can you check the nil pointer exception in your test, please? `make test` or `make verify` locally before pushing helps.

rfranzke (approved):
/lgtm

Issue #2054, opened 2020-03-16
Control plane migration: shoot specific owner ids for DNS entries


MartinWeindel:
**What would you like to be added**:
Gardener and dns-shoot-service should add shoot specific owner ids for their managed DNS entries and corresponding DNS owner objects to allow a smooth control plane migration.

**Why is this needed**:
Both Gardener and the dns-shoot-service manage DNS entries in the control plane of the shoot. If the control plane is moved to another seed, it must not happen, that DNS entries (or more important the corresponding DNS records in the DNS backend) are deleted during intermediate states. Even if they will be recreated later, this would result in unavailable services in the shoot.
There are two controllers generating DNS entries for the shoot: the gardenlet responsible for the API server and INGRESS DNS entries, and the dns shoot service managing entry objects for services and ingresses in the shoot.
There are several aspects to be considered:
- Only one DNS controller must be responsible for managing DNS entries for a shoot. This should already be handled by some hand-over mechanism of Gardener (see issue #1631)
- The DNS controller taking over responsibility must reconcile all DNS entries of the shoot, before it takes ownership of them. Otherwise it can happen, that it starts deleting DNS records of still unreconciled and therefor unknown DNS entry objects. As there are two controllers managing shoot related DNS entries, each needs an own owner ID.

** Suggested owner ids **
The owner id must be globally unique for each shoot and managing controller. Therefore it must contain
- the Garden ID
- the project name
- the shoot name
- the managing controller
The managing controller is needed, as the creation of the DNS owner is used as sync point after all DNS entries which must be migrated are written.

There is already a shoot id used as label key in the shoot-dns-service, which can be reused as a basis.
It has the form: `<garden-id>.gardener.cloud/<shoot-namespace>` 

It is suggested to append `/core` for Gardener and `/dns-shoot` for dns-shoot-service respectively.

rfranzke:
Hm, I'm confused. If you would use `<garden-id>.gardener.cloud/<shoot-namespace>/<some-suffix>` as owner ID, how can you then move the responsibility for the DNS entries to a new seed? I don't see any seed identifier encoded in the owner ID.

(Just for reference, looking at what we have today, the `TXT` record with the owner ID is `seed.gardener.cloud/<garden-id>/<shoot-name>`.)

If I understand correctly, this issue is about the following things:

[...]

PR #2416, opened 2020-06-03
[GEP-7] Add DNSOwner to Shoot operations to support migration scenario

### LINKED ISSUE #2335: Add management for DNSOwners and DNSEntries during Shoot reconciliation, migration and restoration
The normal shoot reconciliation flow should deploy `DNSOwner` objects together with `DNSEntries` and `DNSProviders` and set the `DNSEntry.spec.ownerId` to the `ownerID` field of the `DNSOwner`. The `ownerID` is specific for the shoot cluster and should not depend on the name of the `Seed`.
`DNSOwners`, `DNSEntries` and `DNSPRoviders` are CRs managed by the dns extension controller.

The shoot migration step of the Control Plane Migration should delete these resources in the following order:
1. `DNSOwner`
2. `DNSProvider`
3. `DNSEntries`

The shoot restoration phase of the Control Plane Migration should create these resources in the following order:
1. `DNSEntries`
2. `DNSProvider`
3. `DNSOwner`

**Why is this needed**:
Today `gardenlet` manages and creates the `DNSEntries` and `DNSProvider` required for the `Shoot`'s APIServer. Once the `DNSEntries` are reconciled a dns record is created which will have the `Seed`'s name as its owner id. Once a Shoot is migrated to a new seed, the dns extension controller will try to reconcile the `DNSEntries` on the new seed which will fail since it will try to overwrite the record's owner id to the name of the new `Seed`.  To avoid this error and having to recreate the dns records we need to make use of additional `DNSOwner` objects

/area control-plane-migration
/kind enhancement
/priority normal

swilen-iwanow:

/area control-plane-migration
/kind enhancement
/priority normal

During Shoot creation Gardenlet is going to deploy one more resource - DNSOwner. Currently the DNSOwner is set to be equal to the Shoot namespace name in the Seed. 

Fixes #2335

**Release note**:

```improvement operator
Gardener adds DNSOwner Resource to external and internal DNS.  
```


rfranzke (commented):
The changes themselves look good to me, though, I cannot really judge whether the used IDs (shoot namespace name in seed) are the right choice. There is https://github.com/gardener/gardener/issues/2054 suggesting something else.
@MartinWeindel @mandelsoft please check.

swilen-iwanow:
> The changes themselves look good to me, though, I cannot really judge whether the used IDs (shoot namespace name in seed) are the right choice. There is #2054 suggesting something else.
> @MartinWeindel @mandelsoft please check.

Yes @rfranzke, yesterday I came to the same conclusion, reading trough the issue again. I think the issue itself has to be updated, with the actual work that has to be done. As far as I understand from the comments, we have to implement logic that creates `ConfigMap` with cluster-identity information for each level of cluster Gardener-Seed-Shoot, am I right? Then we can change those owner IDs to be equal to the cluster-identity that is in the Shoot's `kube-system` namespace.

rfranzke:
To be honest, I have lost track a little bit in those regards. It might make sense to talk to the DNS colleagues and then summarize again the implementation plan (only high-level).

mandelsoft:
The owner ID should be taken from the shoot identity provided by gardener (see the comment from Rafael).

There is no need to use different owner ids for the internal, external and ingress dns entries. All entries managed by the gardenlet should use the same id, derived form the shoot id (for example `<shootid>-garden`. The suffix is useful because there are other DNS managers (like the shoot-dns-service) that will also use the shooot id to derive unique identities for there purpose that should not conflict with each other.

For the migration. The ownerid now is specific for a shoot, therefore the id itself does not need to be migrated. The only thing that happens is, that the OwnerID object (carrying the technical identity string) on the giving side has to be deleted and afterwards it has to be created on the taking side (with the very same id).

One tricky part is the order:
- On the giving side
  - first delete the OwnerID Object on the giving side. This prevents the DNS controller on the giving side to touch the DNS entry on the provider. even if the Entry object is deleted.
  - Afterwards the DNSEntry object can be deleted without the danger of loosing the DNS entry in the provider. This can be done at any time later, because the DNS controller does not feel responsible anymore for this object, after the Owner Object has been deleted.
- On the taking side the order is exactly the other way round.
  - First create the DNSEntry objects on the new seed.
  - Then create the OwnerObject with the very same identity used on the giving side (therefore the shoot id should be the base of this identity, to be able to choose the same one). Once this object is created the local DNS controller (on the taking side) now feels responsible for these entries in the provider. Therefore the DNSEntry objects must already be available before the Owner is added, because otherwise the DNS records would be deleted on the provider until they would be recreatred after the DNSEntry objects are again created.

rfranzke:
Do we have to
/hold
this PR until #2054 is implemented?
/needs changes
/status external-action

swilen-iwanow:
Yes I think that we should.

swilen-iwanow:
Accidentally deleted the branch in the fork, can we reopen this somehow?

rfranzke:
@swilen-iwanow I don't think it's possible, you probably have to create a new PR

swilen-iwanow:
> @swilen-iwanow I don't think it's possible, you probably have to create a new PR

Sorry about that.

PR #2418, opened 2020-06-04
update istio to 1.6.0

mvladev:

/area control-plane
/kind enhancement
/priority normal

Updates istio to `1.6.0`

Fixes #

**Release note**:

```improvement operator
Istio is updated to `1.6.0` and `status-port` of ingress gateway is changed to `15021`.
```

### DIFF
--- a/charts/istio/istio-istiod/templates/configmap.yaml
+++ b/charts/istio/istio-istiod/templates/configmap.yaml
@@ -59,7 +59,7 @@ data:
     # If true, automatically configure client side mTLS settings to match the corresponding service's
     # server side mTLS authentication policy, when destination rule for that service does not specify
     # TLS settings.
-    enableAutoMtls: false
+    enableAutoMtls: true
 
     outboundTrafficPolicy:
       # allow traffic only to services in the mesh
--- a/pkg/client/kubernetes/types.go
+++ b/pkg/client/kubernetes/types.go
@@ -77,6 +77,7 @@ func init() {
 		resourcesscheme.AddToScheme,
 		hvpav1alpha1.AddToScheme,
 		druidv1alpha1.AddToScheme,
+		apiextensionsscheme.AddToScheme,
 	)
 	utilruntime.Must(seedSchemeBuilder.AddToScheme(SeedScheme))
 
--- a/pkg/operation/seed/istio/crds.go
+++ b/pkg/operation/seed/istio/crds.go
@@ -20,11 +20,17 @@ import (
 
 	"github.com/gardener/gardener/pkg/client/kubernetes"
 	"github.com/gardener/gardener/pkg/operation/botanist/component"
+
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type crds struct {
 	kubernetes.ChartApplier
-	chartPath string
+	chartPath      string
+	client         crclient.Client
+	deprecatedCRDs []apiextensionsv1beta1.CustomResourceDefinition
 }
 
 // NewIstioCRD can be used to deploy istio CRDs.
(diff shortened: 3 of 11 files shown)


charts/istio/istio-istiod/templates/configmap.yaml
```diff
@@ -59,7 +59,7 @@ data:
     # If true, automatically configure client side mTLS settings to match the corresponding service's
     # server side mTLS authentication policy, when destination rule for that service does not specify
     # TLS settings.
-    enableAutoMtls: false
+    enableAutoMtls: true
```
zanetworker, line 62: Isn't this true by default?

mvladev: It's from `1.6` https://github.com/istio/istio/blob/418e60eda506dfc61a9e333480096b6a6f7756ed/pkg/config/mesh/mesh.go#L156

It's better to be explicit and set it to `true`

zanetworker (dismissed):

pkg/client/kubernetes/types.go
```diff
@@ -77,6 +78,7 @@ func init() {
 		resourcesscheme.AddToScheme,
 		hvpav1alpha1.AddToScheme,
 		druidv1alpha1.AddToScheme,
+		apiextensionsv1.AddToScheme,
```
rfranzke: Wouldn't
```suggestion
		apiextensionsscheme.AddToScheme,
```
be better?

pkg/operation/seed/istio/crds.go
```diff
@@ -20,26 +20,46 @@ import (
...
+			// TODO: remove this after several gardener releases
+			{ObjectMeta: metav1.ObjectMeta{Name: "meshpolicies.authentication.istio.io"}},
+			// TODO: remove this after several gardener releases
+			{ObjectMeta: metav1.ObjectMeta{Name: "policies.authentication.istio.io"}},
+		},
 	}
 }
 
 func (c *crds) Deploy(ctx context.Context) error {
+	for _, derecatedCRD := range c.deprectedCRDs {
```
rfranzke:
```suggestion
	for _, deprecatedCRD := range c.deprectedCRDs {
```

pkg/operation/seed/istio/crds.go
```diff
@@ -20,26 +20,46 @@ import (
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type crds struct {
 	kubernetes.ChartApplier
-	chartPath string
+	chartPath     string
+	client        crclient.Client
+	deprectedCRDs []apiextensionsv1.CustomResourceDefinition
```
rfranzke:
```suggestion
	deprecatedCRDs []apiextensionsv1.CustomResourceDefinition
```

pkg/operation/seed/istio/crds_test.go
```diff
@@ -86,11 +88,20 @@ var _ = Describe("#CRDs", func() {
 		ca := kubernetes.NewChartApplier(renderer, a)
 		Expect(ca).NotTo(BeNil(), "should return chart applier")
 
-		crd = NewIstioCRD(ca, chartsRootPath)
+		crd = NewIstioCRD(ca, chartsRootPath, c)
 
 	})
 
 	JustBeforeEach(func() {
+		deprectedCRDs := []apiextensionsv1.CustomResourceDefinition{
```
rfranzke: Shouldn't v1beta1 be used for backwards compatibility/support of older kubernetes versions?

rfranzke:
```suggestion
		deprecatedCRDs := []apiextensionsv1.CustomResourceDefinition{
```

rfranzke (commented):
Only nits

pkg/client/kubernetes/types.go
mvladev: It's better to use the scheme from the versioned resource

rfranzke: Why? The `apiextensionsscheme.AddToScheme` adds all versions to the scheme and not only a concrete one. This might be needed/helpful as Gardener is working with Kubernetes clusters of several versions.

mvladev:
Fixed schema imports - have another look @rfranzke

rfranzke (dismissed):
Thanks Martin!
/lgtm

rfranzke (approved):

PR #2422, opened 2020-06-05
process managed resource conditions for health checks

timuthy:

/area robustness
/kind enhancement
/priority normal

This PR adapts the health system component checks, so that now the conditions of the corresponding managed resource is considered as opposed to checking the deployment, daemon set, etc. directly. 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/health_check.go
+++ b/pkg/operation/botanist/health_check.go
@@ -35,6 +35,7 @@ import (
 	"github.com/gardener/gardener/pkg/utils/kubernetes/health"
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+	resourcesv1alpha1 "github.com/gardener/gardener-resource-manager/pkg/apis/resources/v1alpha1"
 	prometheusmodel "github.com/prometheus/common/model"
 	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
@@ -64,11 +65,9 @@ func mustGardenRoleLabelSelector(gardenRoles ...string) labels.Selector {
 }
 
 var (
-	controlPlaneSelector    = mustGardenRoleLabelSelector(v1beta1constants.GardenRoleControlPlane)
-	systemComponentSelector = mustGardenRoleLabelSelector(v1beta1constants.GardenRoleSystemComponent)
-	monitoringSelector      = mustGardenRoleLabelSelector(v1beta1constants.GardenRoleMonitoring)
-	optionalAddonSelector   = mustGardenRoleLabelSelector(v1beta1constants.GardenRoleOptionalAddon)
-	loggingSelector         = mustGardenRoleLabelSelector(v1beta1constants.GardenRoleLogging)
+	controlPlaneSelector = mustGardenRoleLabelSelector(v1beta1constants.GardenRoleControlPlane)
+	monitoringSelector   = mustGardenRoleLabelSelector(v1beta1constants.GardenRoleMonitoring)
+	loggingSelector      = mustGardenRoleLabelSelector(v1beta1constants.GardenRoleLogging)
 )
 
 // Now determines the current time.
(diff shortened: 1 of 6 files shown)


pkg/operation/botanist/health_check.go
```diff
@@ -187,9 +182,22 @@ func (b *HealthChecker) checkRequiredDaemonSets(condition gardencorev1beta1.Cond
 		actualNames.Insert(object.Name)
 	}
 
-	if missingNames := requiredNames.Difference(actualNames); missingNames.Len() != 0 {
-		c := b.FailedCondition(condition, "DaemonSetMissing", fmt.Sprintf("Missing required daemon sets: %v", missingNames.List()))
-		return &c
+	return b.checkRequiredResourceNames(condition, requiredNames, actualNames, "DaemonSetMissing", "Missing required daemon sets")
+}
+
+var relevantConditions = map[resourcesv1alpha1.ConditionType]struct{}{
```
```suggestion
var relevantManagedResourceConditions = map[resourcesv1alpha1.ConditionType]struct{}{
```

pkg/operation/botanist/health_check.go
```diff
@@ -187,9 +182,22 @@ func (b *HealthChecker) checkRequiredDaemonSets(condition gardencorev1beta1.Cond
...
+	resourcesv1alpha1.ResourcesApplied: struct{}{},
+	resourcesv1alpha1.ResourcesHealthy: struct{}{},
+}
+
+// CheckManagedResource checks the conditions of the given managed resource and reflects the state in the returned condition.
+func (b *HealthChecker) CheckManagedResource(condition gardencorev1beta1.Condition, mr *resourcesv1alpha1.ManagedResource) *gardencorev1beta1.Condition {
+	for _, cond := range mr.Status.Conditions {
+		_, relevant := relevantConditions[cond.Type]
+		if relevant && cond.Status == resourcesv1alpha1.ConditionFalse {
```
rfranzke: Shouldn't the condition also fail if the relevant type wasn't found?

pkg/operation/botanist/health_check.go
```diff
@@ -187,9 +182,22 @@ func (b *HealthChecker) checkRequiredDaemonSets(condition gardencorev1beta1.Cond
...
+}
+
+var relevantConditions = map[resourcesv1alpha1.ConditionType]struct{}{
+	resourcesv1alpha1.ResourcesApplied: struct{}{},
+	resourcesv1alpha1.ResourcesHealthy: struct{}{},
+}
+
+// CheckManagedResource checks the conditions of the given managed resource and reflects the state in the returned condition.
+func (b *HealthChecker) CheckManagedResource(condition gardencorev1beta1.Condition, mr *resourcesv1alpha1.ManagedResource) *gardencorev1beta1.Condition {
+	for _, cond := range mr.Status.Conditions {
```
timebertt, line 198: Should we also check, if `observedGeneration` of the mr is up-to-date (like [here](https://github.com/gardener/gardener-resource-manager/blob/master/pkg/health/health.go#L46))?

timebertt (commented):
Nice PR!
Only one additional comment from my side.

pkg/operation/botanist/health_check.go
timuthy: Yes, actually a good idea. The question is though what we should do in such a case? Is it helpful to report an "outdated" message instead of the actual condition?

pkg/operation/botanist/health_check.go
timuthy: Yes, good idea.

pkg/operation/botanist/health_check.go
timebertt: Yes, I would probably go for an "outdated" message as this has (probably) no correlation to a single condition, right?

timuthy: Sure, let's proceed like that 👍

timuthy:
Thanks for your reviews @rfranzke @tim-ebert. The changes are ready for another round of review.

rfranzke:
Can you check the CI failure?

pkg/operation/botanist/health_check.go
```diff
@@ -656,20 +578,23 @@ func (b *Botanist) checkSystemComponents(
...
-		return exitCondition, err
+	for name := range common.ManagedResourcesShoot {
+		mr := &resourcesv1alpha1.ManagedResource{}
+		if err := b.K8sSeedClient.Client().Get(ctx, kutil.Key(b.Shoot.SeedNamespace, name), mr); err != nil {
+			return nil, err
+		}
+
+		if mr.Generation != mr.Status.ObservedGeneration {
+			return nil, fmt.Errorf("observed generation of managed resource %s/%s outdated (%d/%d)", mr.Namespace, mr.Name, mr.Status.ObservedGeneration, mr.Generation)
+		}
```
timebertt: What's the reason, you didn't put the generation check into the `CheckManagedResource` func?

timebertt (dismissed):
/lgtm now

pkg/operation/botanist/health_check.go
timuthy: Because the surrounding function handles the error with an unkown condition. I thought that would be more appropriate.

rfranzke: Hmm, `Unknown` is usually only used we cannot determine the status of a health check, for example when we cannot talk to the Kubernetes API server. If we detect an outdated condition then we can indeed determine the status and should probably report a failed condition.

rfranzke (approved):
/lgtm

timebertt (approved):
Nice!
/lgtm

Issue #2423, opened 2020-06-05
[Flaky test] guestbook test is failing on azure


ialidzhikov:
**How to categorize this issue?**

/area testing
/kind bug
/priority normal

**What happened**:
I observe too many flakes for the guestbook test on azure.
One of the flakes that I managed to reproduce is related to `ingress-nginx` where the ingress controller hangs too long `attempting to acquire leader lease`.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
1. Run the guestbook test on newly created Azure Shoot.

```
go test -timeout=0 -mod=vendor ./test/suites/shoot \
      --v -ginkgo.v -ginkgo.progress -ginkgo.noColor \
      --report-file=/tmp/report.json \
      -kubecfg=$KUBECONFIG \
      -shoot-name=$SHOOT_NAME \
      -project-namespace=$PROJECT_NAMESPACE \
      -ginkgo.focus="should deploy guestbook app successfully"
```

2. `addons-nginx-ingress-controller` hangs with:

```
I0605 13:11:22.175670       7 leaderelection.go:249] failed to renew lease kube-system/ingress-controller-leader-nginx: failed to tryAcquireOrRenew context deadline exceeded
I0605 13:11:22.175757       7 leaderelection.go:205] attempting to acquire leader lease  kube-system/ingress-controller-leader-nginx...
```

The pod is shown as running (and the livenessProbe obviously pass)
```
$ k -n kube-system get po addons-nginx-ingress-controller-759458d496-zxgcv
NAME                                               READY   STATUS    RESTARTS   AGE
addons-nginx-ingress-controller-759458d496-zxgcv   1/1     Running   0          19m
```

despite the fact the Pod cannot acquire leader lease.

```
$ k -n kube-system get cm ingress-controller-leader-nginx -o yaml
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    control-plane.alpha.kubernetes.io/leader: '{"holderIdentity":"addons-nginx-ingress-controller-759458d496-zxgcv","leaseDurationSeconds":30,"acquireTime":"2020-06-05T13:05:44Z","renewTime":"2020-06-05T13:10:59Z","leaderTransitions":0}'
  creationTimestamp: "2020-06-05T13:05:44Z"
  name: ingress-controller-leader-nginx
  namespace: kube-system
  resourceVersion: "2814"
[...]

PR #2424, opened 2020-06-06
Guestbook

dguendisch:
/area testing
/kind test
/priority normal

Azure guestbook tests are flaky, see #2423
This PR increases the timeout to wait for the guestbook app to serve its URLs from `10min` to `20min` and the timeout to wait for the namespace to be deleted (at the end of the test) from `5min` to `10min`.
It adds additional dumping logic to dump the nginx ingress controller logs at the end of a failed test so that we have a better chance to analyze the cause of the flakes.

Maybe improves #2423, at least helps to better understand its potential causes.

### DIFF
--- a/test/framework/dump.go
+++ b/test/framework/dump.go
@@ -15,6 +15,8 @@
 package framework
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"sort"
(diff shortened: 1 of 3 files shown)


schrodit (dismissed):
lgtm

test/framework/dump.go
```diff
@@ -209,6 +210,37 @@ func (f *GardenerFramework) dumpGardenerExtension(extension v1alpha1.Object) {
...
+		if err := f.DumpLogsForPodInNamespace(ctx, ctxIdentifier, k8sClient, namespace, pod.Name); err != nil {
+			result = multierror.Append(result, err)
+		}
+	}
+	return result
+}
+
+func (f *CommonFramework) DumpLogsForPodInNamespace(ctx context.Context, ctxIdentifier string, k8sClient kubernetes.Interface, podNamespace, podName string) error {
+	f.Logger.Infof("%s [NAMESPACE %s] [POD %s] [LOGS]", ctxIdentifier, podNamespace, podName)
+	request := k8sClient.Kubernetes().CoreV1().Pods(podNamespace).GetLogs(podName, &corev1.PodLogOptions{})
```
ialidzhikov: There is already `pkg/client/kubernetes.GetPodLogs` func. We can use it, right?

test/framework/dump.go
```diff
@@ -209,6 +210,37 @@ func (f *GardenerFramework) dumpGardenerExtension(extension v1alpha1.Object) {
 	}
 }
 
+func (f *CommonFramework) DumpLogsForPodsWithLabelsInNamespace(ctx context.Context, ctxIdentifier string, k8sClient kubernetes.Interface, namespace string, labels client.MatchingLabels) error {
```
ialidzhikov: Nit: you could make it 1 idea more generic:

```suggestion
func (f *CommonFramework) DumpLogsForPodsWithLabelsInNamespace(ctx context.Context, ctxIdentifier string, k8sClient kubernetes.Interface, namespace string, opts ...client.ListOption) error {
```

test/framework/dump.go
dguendisch: Right, thanks for pointing out.

test/framework/dump.go
dguendisch: Good idea, thanks!

dguendisch:
Addressed comments, PTAL

ialidzhikov (approved):
/lgtm

PR #2431, opened 2020-06-08
Replace custom ProviderConfig type with runtime.RawExtension

### LINKED ISSUE #1917: Directly use runtime.RawExtension instead of ProviderConfig in apis
[type_utils:22](https://github.com/gardener/gardener/blob/master/pkg/apis/core/types_utils.go#L22) (and the corresponding versioned definition)
```
// ProviderConfig is a workaround for missing OpenAPI functions on runtime.RawExtension struct.
// https://github.com/kubernetes/kubernetes/issues/55890
// https://github.com/kubernetes-sigs/cluster-api/issues/137
type ProviderConfig struct {
	runtime.RawExtension
}
```
Could be removed in favour of directly using runtime.RawExtension as of kubernetes 1.16 the listed issues in the workaround have been resolved.

**Why is this needed**:
ProviderConfig was only a temporary workaround.

rfranzke:

/area dev-productivity
/kind cleanup
/priority normal

This PR cleans up our own wrapper type for `runtime.RawExtension` due to limitations in previous Kubernetes versions that got fixed meanwhile.

Fixes #1917

**Release note**:

```improvement operator
NONE
```


timebertt (commented):
In general, PR looks good.

One general question, though:
The godoc of `RawExtension` says:
```go
// RawExtension is used to hold extensions in external versions.
//
// To use this, make a field which has RawExtension as its type in your external, versioned
// struct, and Object in your internal struct. You also need to register your
// various plugin types.
```

It seems like the intended usage is to have `runtime.RawExtension` only in the external/versioned APIs and `runtime.Object` in internal APIs. You could then add conversion funcs, that decode the raw extensions into the respective golang types or fall back to `Unknown` objects. I am wondering, if that would benefit us in any way or rather complicate things and if this is even possible for our case... 🤔 
If we did it like that, would we then be able to avoid decode calls in the extensions in favor of the "baked in" conversion funcs of the scheme/decoder/serializer?

I'm fine with proceeding in the way you did it in this PR. I'm just wondering and thinking out loud.
What are your thoughts?

rfranzke:
I don't think I've understood what you suggest/ask.

> You could then add conversion funcs, that decode the raw extensions into the respective golang types 

What does this mean? Let's take the `.spec.provider.infrastructureConfig` field in `v1beta1.Shoot`. What should the conversion function to when converting to `core.Shoot`?

timebertt:
> What does this mean? Let's take the `.spec.provider.infrastructureConfig` field in `v1beta1.Shoot`. What should the conversion function to when converting to `core.Shoot`?

This will probably convert to an `Unknown` object. But as g/g itself doesn't deal with provider configs, a benefit would probably on the extensions' side. So they could register for example the respective `InfrastructureConfig` types and so on in their schemes and would then get a `{aws,gcp,...}.InfrastructureConfig` object instead of having to decode a `RawExtension` themselves.
I am not sure, if this is how it works, but at least I understood it this way.

rfranzke:
Hm, okay, I don't know either whether it would work. This requires some more investment to investigate what can be achieved.

timuthy (approved):
/lgtm

timebertt (approved):
> Hm, okay, I don't know either whether it would work. This requires some more investment to investigate what can be achieved.

Sure, let's keep it this way at first...
/lgtm

PR #2435, opened 2020-06-09
Fix deletion case for extension predicates

### LINKED ISSUE #2432: Gardener fails to delete failed shoots

/kind bug
/priority normal

**What happened**:
Gardener failed to delete a shoot because the infrastructure had depending objects. After a while the shoot transitioned into a `Failed` state which means that Gardener does not automatically retry the deletion until it is manually re-triggered (annotation `gardener.cloud/operation=retry`) by users or operators.

After the mentioned depending infra objects had been deleted, the deletion was re-triggered for the shoot but the infrastructure extension resource in the seed was not reconciled anymore.

The infra resource in the seed is reconciled, so that the infrastructure is torn down and the infra resource finally removed.

**Environment**:

- Gardener version: v1.5.2

timuthy:

/area quality
/kind bug
/priority normal

This PR fixes the deletion of shoots or their extension resources in the seed once shoots were in a `Failed` state.

It also simplifies the predicates for extension controllers. I create a list of cases (operation annotation not ignored) in which a reconciliation is desired and ask the reviewers to double check them. Please note, that those cases only consider _events_ that are send to the queue and does not cover cases in which a controller enqueues elements directly.

**Desired reconciliation:**
1. Resource creation
2. Existing resource is annotated with `gardener.cloud/operation: reconcile`
3. Resource deletion
4. Reconcile loop is interrupted (e.g. through Pod deletion), so that the next controller instance is supposed to automatically reconcile the resource (`lastOperation.state` != `Succeeded` for create events).

**Undesired reconciliation:**
1. Resource (meta, spec, status) is updated w/o `gardener.cloud/operation: reconcile` annotation and is not in deletion.
2. Shoot has `lastOperation.state: Failed`.

Fixes #2432

**Release note**:

```improvement operator
An issue has been fixed which caused failed shoot clusters to not be deleted successfully after a retry had been triggered.
```

### DIFF
--- a/extensions/pkg/controller/backupbucket/controller.go
+++ b/extensions/pkg/controller/backupbucket/controller.go
@@ -67,10 +67,6 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
 			extensionspredicate.LastOperationNotSuccessful(),
 			extensionspredicate.IsDeleting(),
 		),
-		extensionspredicate.Or(
-			extensionspredicate.HasOperationAnnotation(),
-			predicate.GenerationChangedPredicate{},
-		),
 	}
 }
 
--- a/extensions/pkg/controller/backupentry/controller.go
+++ b/extensions/pkg/controller/backupentry/controller.go
@@ -67,10 +67,6 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
 			extensionspredicate.LastOperationNotSuccessful(),
 			extensionspredicate.IsDeleting(),
 		),
-		extensionspredicate.Or(
-			extensionspredicate.HasOperationAnnotation(),
-			predicate.GenerationChangedPredicate{},
-		),
 	}
 }
 
--- a/extensions/pkg/controller/containerruntime/controller.go
+++ b/extensions/pkg/controller/containerruntime/controller.go
@@ -77,10 +77,6 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
 			extensionspredicate.IsDeleting(),
 		),
 		extensionspredicate.ShootNotFailed(),
-		extensionspredicate.Or(
-			extensionspredicate.HasOperationAnnotation(),
-			predicate.GenerationChangedPredicate{},
-		),
 	}
 }
 
--- a/extensions/pkg/controller/controlplane/controller.go
+++ b/extensions/pkg/controller/controlplane/controller.go
@@ -67,10 +67,6 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
 			extensionspredicate.IsDeleting(),
 		),
 		extensionspredicate.ShootNotFailed(),
-		extensionspredicate.Or(
-			extensionspredicate.HasOperationAnnotation(),
-			predicate.GenerationChangedPredicate{},
-		),
 	}
 }
 
--- a/extensions/pkg/controller/extension/reconciler.go
+++ b/extensions/pkg/controller/extension/reconciler.go
@@ -91,10 +91,6 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
 			extensionspredicate.IsDeleting(),
 		),
 		extensionspredicate.ShootNotFailed(),
-		extensionspredicate.Or(
-			extensionspredicate.HasOperationAnnotation(),
-			predicate.GenerationChangedPredicate{},
-		),
 	}
 }
 
--- a/extensions/pkg/controller/infrastructure/controller.go
+++ b/extensions/pkg/controller/infrastructure/controller.go
@@ -71,10 +71,6 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
 			extensionspredicate.IsDeleting(),
 		),
 		extensionspredicate.ShootNotFailed(),
-		extensionspredicate.Or(
-			extensionspredicate.HasOperationAnnotation(),
-			predicate.GenerationChangedPredicate{},
-		),
 	}
 }
 
--- a/extensions/pkg/controller/network/controller.go
+++ b/extensions/pkg/controller/network/controller.go
@@ -67,10 +67,6 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
 			extensionspredicate.IsDeleting(),
 		),
 		extensionspredicate.ShootNotFailed(),
-		extensionspredicate.Or(
-			extensionspredicate.HasOperationAnnotation(),
-			predicate.GenerationChangedPredicate{},
-		),
 	}
 }
 
--- a/extensions/pkg/controller/operatingsystemconfig/controller.go
+++ b/extensions/pkg/controller/operatingsystemconfig/controller.go
@@ -72,10 +72,6 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
 			extensionspredicate.LastOperationNotSuccessful(),
 			extensionspredicate.IsDeleting(),
 		),
-		extensionspredicate.Or(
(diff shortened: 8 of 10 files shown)


rfranzke:
Thanks for the PR, but doesn't this break the use case when the operation annotation is disabled/not used?

timuthy:
> Thanks for the PR, but doesn't this break the use case when the operation annotation is disabled/not used?

Actually, I didn't touch those cases. They still just have the `GenerationChangedPredicate`.

rfranzke:
I see, thanks. We had introduced this `Or` predicate as part of https://github.com/gardener-attic/gardener-extensions/commit/e04ae949f8694e00920de6d0871ffd920745f293 to prevent infinite reconciliations due to status updates. As this PR reverts this commit, wouldn't we get back to this problem?

timuthy:
That's the reason I've changed the `LastOperationNotSuccessful` predicate to only consider create events and return `false` for any other case. I think this predicate only makes sense to cover case (4) from the desired reconciliations.

rfranzke:
OK, thanks for the explanations. Sounds good then, I think it makes sense. Still, a dangerous change. Which kind of testing was done?
Also
/invite @tim-ebert @ialidzhikov 
for 
/second-opinion

rfranzke:
@gardener-robot @vlerenc bug? :)

timuthy:
> Still, a dangerous change.

100% agree. I tested the use cases mentioned in the description isolated for the `extenion.extensions.gardener.cloud` resource. Maybe also @danielfoehrKn can double check the use cases since he also worked with the predicates some weeks ago.

vlerenc:
I guess so @rfranzke, I have refactored a lot in the bot and that's what happens without proper tests: you break yourself. I should now. ;-) Thanks for letting me know. I'll look into it.

ialidzhikov (approved):
I played with the described cases and looks good to me.
Probably it would be safer to wait until v1.6.0 is released?

/lgtm

rfranzke:
Thanks @ialidzhikov for your time and effort.
Isn't this PR something that should be cherry-picked to at least `release-v1.5`?

danielfoehrKn:
It makes sense to me that the status infinite reconcile loop is  still prevented by the changes introduced above.

I could also reproduce the issue with a Shoot stuck in failed, then annotated with operation/retry and then the provider extension did not start reconciling the extension again.
So what I am asking myself is
 - Gardener adds [the timestamp annotation](https://github.com/gardener/gardener/blob/ba0ca2f9125fe4c229ea6801a5996cbbc48abe96/pkg/operation/common/utils.go#L770) to trigger an event. The predicate [isDeleting()](https://github.com/gardener/gardener/blob/d722a8d82fd418d3f473b8fce4150fb4d24f165a/extensions/pkg/controller/infrastructure/controller.go#L71) should evaluate to true - why did it not happen? 
- I am not sure why the changes to `LastOperationNotSuccessful ` and removing the predicate helps to retrigger the Extension reconciliation when it was failed.
```
		extensionspredicate.Or(
			extensionspredicate.HasOperationAnnotation(),
			predicate.GenerationChangedPredicate{},
		),
```  
Additionally: why does gardener not also simply add the operation annotation on the infrastructure CRD during the Shoot reconciliation. That would def. evaluate the predicate to `true` if the Operation annotation is not ignored.

I need to further test this, but maybe someone that already tested it can shed some light.

danielfoehrKn:
Another question when already targeting the predicates. When the OperationAnnotation is ignored & the Shoot is in `failed` state, [the infra CRD will only be reconciled if the Generation changed](https://github.com/gardener/gardener/blob/d722a8d82fd418d3f473b8fce4150fb4d24f165a/extensions/pkg/controller/infrastructure/controller.go#L61) . This seems wrong to me. There should be a mechanism to retry the operation without changing the spec. E.g a retry annotation.

ialidzhikov:
> Isn't this PR something that should be cherry-picked to at least release-v1.5?

I have no strong opinion. @timuthy, I guess only the `LastOperationNotSuccessful` predicate change is enough for fixing the described issue -https://github.com/gardener/gardener/issues/2432, right? If yes, could we leave the cleanup/simplification for a follow up PR?
I think it would be good to have it deployed and 👀 -ed on our dev landscape. WDYT?

timuthy:
@danielfoehrKn 
> Gardener adds the timestamp annotation to trigger an event. The predicate isDeleting() should evaluate to true - why did it not happen?

It didn't happen because of the [second OR block](https://github.com/gardener/gardener/blob/e0b05a5b8aea511342e6779e87644213bfa528b9/extensions/pkg/controller/infrastructure/controller.go#L74-L76) that looked for the operation annotation or the generation change.

> I am not sure why the changes to LastOperationNotSuccessful and removing the predicate helps to retrigger the Extension reconciliation when it was failed.

Basically for the reason mentioned above. Additionally, the changes in the `LastOperationNotSuccessful` predicate are necessary to prevent an infinite loop through status updates.

> Additionally: why does gardener not also simply add the operation annotation on the infrastructure CRD during the Shoot reconciliation.

Gardener does that during reconciliation but not when the resource is deleted.
https://github.com/gardener/gardener/blob/4049e52f6f79656cc8de3e4bdc7be30569374ca8/pkg/operation/botanist/infrastructure.go#L66

> Another question when already targeting the predicates. When the OperationAnnotation is ignored & the Shoot is in failed state, the infra CRD will only be reconciled if the Generation changed . This seems wrong to me. There should be a mechanism to retry the operation without changing the spec. E.g a retry annotation.

You have a point there. I think ignoring the operation annotation always has been a development use case only and thus we never took any sophisticated reconciliation decisions. There are other restrictions as well if you run Gardener w/ the extension and ignoring the operation annotation, e.g. no regular reconciliation of the infrastructure.

timuthy:
> > Isn't this PR something that should be cherry-picked to at least release-v1.5?
> 
> I have no strong opinion. @timuthy, I guess only the `LastOperationNotSuccessful` predicate change is enough for fixing the described issue -#2432, right? If yes, could we leave the cleanup/simplification for a follow up PR?
> I think it would be good to have it deployed and 👀 -ed on our dev landscape. WDYT?

Actually, the changes for the `LastOperationNotSuccessful` predicate only allow to get rid of the second OR block. A simple fix which is not as risky as the this change is to simply change the second OR block like this:

```
extensionspredicate.Or(
			extensionspredicate.HasOperationAnnotation(),
                         extensionspredicate.IsDeleting(),
			predicate.GenerationChangedPredicate{},
		),
```

Do you want me to file a PR for the `release-v1.5` branch (/cc @rfranzke )?

danielfoehrKn:
> It didn't happen because of the second OR block that looked for the operation annotation or the generation change. 

Thanks, forgot that all the predicates have to evaluate to `true`.  Then it makes sense.

I know this goes beyond this PR, but I feel like the decision wether something is reconciled or not is difficult to comprehend. If there would be a debug option that gives further details, that would be helpful for development and operations. E.g.
```
Reconciling Infrastructure CRD  (name/ns). The following predicates evaluated to `true`: (GenerationChanged, ShootIsNotFailed)
```

```
Not Reconciling Infrastructure CRD  (name/ns). The following predicates evaluated to `false`: Or(GenerationChanged, ShootIsNotFailed), ShootIsNotFailed.
```

rfranzke:
Shall we proceed with the PR for now as is to fix the bug? Both @timuthy and @ialidzhikov have validated the various scenarios, so it should be ok?

timebertt:
Yes, I think we can. 👍 
I also gave it some thought and didn't find any scenario, where it shouldn't work, although I can't promise that I thought about all corner cases... :)

rfranzke:
Alright, thanks! Same here.. let's hope for the best ;-)
/merge

rfranzke:
> Do you want me to file a PR for the release-v1.5 branch (/cc @rfranzke )?

@timuthy I think it makes sense, yes.

timuthy:
Backport https://github.com/gardener/gardener/pull/2445

PR #2436, opened 2020-06-09
Enable namespace release when deleting Projects

### LINKED ISSUE #879: Enable deleting a project without deleting its namespace
For the garden-setup project, I'd like to create a `garden` project linked to the `garden` namespace, in order to be able to use the `shoot.garden.sapcloud.io/use-as-seed` annotation and have the seed created automatically. 

However, since deleting a project will also delete the corresponding namespace, this causes trouble when trying to remove the project again, as the Gardener itself is within the `garden` namespace. 

Unfortunately, it seems impossible to undo the step of turning a namespace into a project namespace. 
It would be nice to have an annotation (or something else) that changes the project deletion behaviour, so that instead of deleting the namespace, the Gardener will instead undo anything it did to the namespace when it was associated with the project (e.g. deleting `ownerReference`, etc.) and leave it in the state it was in before the project was created.

rfranzke:

/area usability
/kind enhancement
/priority normal

This PR enables the possibility to release a `Namespace` when deleting a `Project`, i.e., it is not being deleted. In order to achieve this the `Namespace` must be annotated with `namespace.gardener.cloud/keep-after-project-deletion=true`.

Fixes #879

/cc @Diaphteiros 

**Release note**:

```improvement operator
It is now possible to keep a `Namespace` in the system even when the related `Project` is deleted by annotating the `Namespace` with `namespace.gardener.cloud/keep-after-project-deletion=true`.
```

### DIFF
--- a/pkg/controllermanager/controller/project/project_control_delete.go
+++ b/pkg/controllermanager/controller/project/project_control_delete.go
@@ -16,45 +16,39 @@ package project
 
 import (
 	"context"
+	"strconv"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/controllerutils"
+	"github.com/gardener/gardener/pkg/operation/common"
 
-	"github.com/sirupsen/logrus"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func (c *defaultControl) delete(project *gardencorev1beta1.Project, projectLogger logrus.FieldLogger) (bool, error) {
+func (c *defaultControl) delete(ctx context.Context, project *gardencorev1beta1.Project) (bool, error) {
 	if namespace := project.Spec.Namespace; namespace != nil {
-		alreadyDeleted, err := c.deleteNamespace(project, *namespace)
+		released, err := c.releaseNamespace(ctx, project, *namespace)
 		if err != nil {
 			c.reportEvent(project, true, gardencorev1beta1.ProjectEventNamespaceDeletionFailed, err.Error())
 			_, _ = c.updateProjectStatus(project.ObjectMeta, setProjectPhase(gardencorev1beta1.ProjectFailed))
 			return false, err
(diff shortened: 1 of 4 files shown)


pkg/operation/common/types.go
```diff
@@ -274,6 +274,11 @@ const (
 	// Deprecated: Use `NamespaceProject` instead.
 	NamespaceProjectDeprecated = "namespace.garden.sapcloud.io/project"
 
+	// NamespaceKeepAfterProjectDeletion is a constant for an annotation on a `Namespace` resource that states that the
```
timebertt: Nit
```suggestion
	// NamespaceKeepAfterProjectDeletion is a constant for an annotation on a `Namespace` resource that states that
```

pkg/controllermanager/controller/project/project_control_delete.go
```diff
@@ -74,6 +68,33 @@ func (c *defaultControl) deleteNamespace(project *gardencorev1beta1.Project, nam
...
+		delete(namespace.Labels, common.ProjectName)
+		delete(namespace.Labels, v1beta1constants.GardenRole)
+		delete(namespace.Labels, common.ProjectNameDeprecated)
+		delete(namespace.Labels, v1beta1constants.DeprecatedGardenRole)
+		for i := len(namespace.OwnerReferences) - 1; i >= 0; i-- {
+			if ownerRef := namespace.OwnerReferences[i]; ownerRef.APIVersion == gardencorev1beta1.SchemeGroupVersion.String() &&
+				ownerRef.Kind == "Project" &&
+				ownerRef.Name == project.Name &&
+				ownerRef.UID == project.UID {
+				namespace.OwnerReferences = append(namespace.OwnerReferences[:i], namespace.OwnerReferences[i+1:]...)
```
timebertt, line 91: Hmm, if you have multiple owner references that should be removed, this won't remove the second reference, but rather the one after that or result in an out of range panic, correct?
I know this will never happen, as the API server doesn't accept multiple owner references with the same field values. But for the sake of correct code, can you build up a slice of references that should be kept instead of removing items from the slice you're looping over? WDYT?

timebertt (commented):
/lgtm
Only two nits, but no "must-fix" ones :)

pkg/controllermanager/controller/project/project_control_delete.go
timuthy: AFAIK, this should work the way it is implemented because it just decreases by each iteration and doesn't use any index which is out of range.

timebertt: Oh well, you're right. I missed that 😄

timuthy (dismissed):
/lgtm
We should also think about making this the "new default" for re-used namespaces.

danielfoehrKn (dismissed):

PR #2437, opened 2020-06-09
Add network policy documentation

danielfoehrKn:

/area documentation
/kind enhancement
/priority normal

Add documentation for the network policies that Gardener deploys into the Seed's  `gardener` &`shoot-namespaces` and into the `kube-system` namespace in the Shoot.
Fixes # 1580

We should not introduce such features without proper documentation. 
**Release note**:

```improvement operator

```

### DIFF
--- a/docs/extensions/controlplane.md
+++ b/docs/extensions/controlplane.md
@@ -56,6 +56,12 @@ The second one contains the output of the [`Infrastructure` resource](infrastruc
 In order to support a new control plane provider you need to write a controller that watches all `ControlPlane`s with `.spec.type=<my-provider-name>`.
 You can take a look at the below referenced example implementation for the Alicloud provider.
 
+The control plane controller as part of the `ControlPlane` reconciliation, often deploys resources (e.g. pods/deployments) into the Shoot namespace in the `Seed` as part of its `ControlPlane` reconciliation loop.
+Because the namespace contains [network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) that per default [deny all ingress and egress traffic](https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-deny-all-ingress-and-all-egress-traffic), 
+the pods may need to have proper labels matching to the selectors of the network policies in order to allow the required network traffic. 
+Otherwise, they won't be allowed to talk to certain other components (e.g., the kube-apiserver of the shoot).
+Please [see this document](https://github.com/gardener/gardener/tree/master/docs/development/seed_network_policies.md) for more information.
+
 ## Non-provider specific information required for infrastructure creation
 
 Most providers might require further information that is not provider specific but already part of the shoot resource.
--- a/docs/README.md
+++ b/docs/README.md
@@ -15,6 +15,7 @@
     * [Gardener Seed Admission Controller](concepts/seed-admission-controller.md)
     * [Gardenlet](concepts/gardenlet.md)
 * [Backup Restore](concepts/backup-restore.md)
+* [Network Policies](concepts/network_policies.md)
 
 ## Usage
 
@@ -32,6 +33,7 @@
 * [`ManagedIstio` feature](usage/istio.md)
 * [Trusted TLS certificate for shoot control planes](usage/trusted-tls-for-control-planes.md)
 * [API Server Network Proxy Reverse Tunneling](usage/reverse-tunnel.md)
+* [Network Policies in the Shoot Cluster](usage/shoot_network_policies.md)
 
 ## Proposals
 
@@ -56,7 +58,7 @@
 * [Adding New Cloud Providers](development/new-cloud-provider.md)
 * [Extending the Monitoring Stack](development/monitoring-stack.md)
 * [How to create log parser for container into fluent-bit](development/log_parsers.md)
-* [Feature Gates in Gardener](development/feature_gates.md)
+* [Network Policies in the Seed Cluster](development/seed_network_policies.md)
 
 ## Extensions
 
@@ -103,6 +105,7 @@
 * [Deploying the Gardener and a Seed into an AKS cluster](deployment/aks.md)
 * [Overwrite image vector](deployment/image_vector.md)
 * [Migration from Gardener `v0` to `v1`](deployment/migration_v0_to_v1.md)
+* [Feature Gates in Gardener](deployment/feature_gates.md)
 
 ## Monitoring
 
--- a/docs/concepts/network_policies.md
+++ b/docs/concepts/network_policies.md
@@ -1,0 +1,18 @@
+# Network Policies in Gardener
+
+As `Seed` clusters can host the [Kubernetes control planes](https://kubernetes.io/docs/concepts/#kubernetes-control-plane) of many `Shoot` clusters, it is necessary to isolate the control planes from each other for security reasons.
+Besides deploying each control plane in its own namespace, Gardener creates [network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) to also isolate the networks. 
+Essentially, network policies make sure that pods can only talk to other pods over the network they are supposed to.
+As such, network policies are an important part of Gardener's tenant isolation.
+
+Gardener deploys network policies into
+ - each namespace hosting the Kubernetes control plane of the Shoot cluster.
+ - the namespace dedicated to Gardener seed-wide global controllers. This namespace is often called `garden` and contains e.g. the [Gardenlet](https://github.com/gardener/gardener/blob/15cae57db802cbe460ff4cb3f80c26b2fc15e26f/docs/concepts/gardenlet.md).
+ - the `kube-system` namespace in the Shoot.
+ 
+The aforementioned namespaces in the Seed contain a `deny-all` network policy that [denies all ingress and egress traffic](https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-deny-all-ingress-and-all-egress-traffic).
+This [secure by default](https://en.wikipedia.org/wiki/Secure_by_default) setting requires pods to whitelist network traffic.
+This is done by pods having [labels matching to the selectors of the network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/#networkpolicy-resource) deployed by Gardener.
+
+More details on the deployed network policies can be found in the [development](https://github.com/gardener/gardener/tree/master/docs/development/seed_network_policies.md) and [usage](https://github.com/gardener/gardener/tree/master/docs/development/shoot_network_policies.md) sections.
+ 
--- a/docs/development/seed_network_policies.md
+++ b/docs/development/seed_network_policies.md
@@ -1,0 +1,130 @@
+# Network Policies in the Seed Cluster
+
+This document describes the [Kubernetes network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) deployed by Gardener into the Seed cluster.
+For network policies deployed into the Shoot `kube-system` namespace, please see the [usage section](https://github.com/gardener/gardener/tree/master/docs/development/shoot_network_policies.md).
+
+Network policies deployed by Gardener have names and annotations describing their purpose, so this document does only highlight a subset of the policies in detail.
+
+## Network policies in the Shoot namespace in the Seed
+
+The network policies in the Shoot namespace in the Seed can roughly be grouped into policies required for the control plane components and for logging & monitoring.   
+
+The network policy `deny-all` plays a special role. This policy [denies all ingress and egress traffic](https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-deny-all-ingress-and-all-egress-traffic) from each pod in the Shoot namespace.
+So per default, a pod running in the control plane cannot talk to any other pod in the whole Seed cluster.
+This means the pod needs to have labels matching to appropriate network policies allowing it to talk to exactly the components required to execute its desired functionality.
+[This has also implications for Gardener extensions](#implications-for-gardener-extensions) that need to deploy additional components into the `Shoot's` control plane.
+
+### Network Policies for Control Plane Components
+
+This section highlights a selection of network policies that exist in the Shoot namespace in the Seed cluster.
+In general, the control plane components serve different purposes and thus need access to different pods and network ranges.
+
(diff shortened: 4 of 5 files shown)


docs/extensions/controlplane.md
```diff
@@ -56,6 +56,11 @@ The second one contains the output of the [`Infrastructure` resource](infrastruc
 In order to support a new control plane provider you need to write a controller that watches all `ControlPlane`s with `.spec.type=<my-provider-name>`.
 You can take a look at the below referenced example implementation for the Alicloud provider.
 
+The control plane provider as part of the `ControlPlane` reconciliation, often deploys resources (e.g pods/deployments) into the Shoot namespace in the `Seed`.
+Because the namespace contains [network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) that per default [deny all ingress and egress traffic](https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-deny-all-ingress-and-all-egress-traffic), 
+the pods need to use proper network policies that allow the required network traffic.
```
rfranzke:
```suggestion
the pods may need to have proper labels matching to the selectors of the network policies in order to allow the required network traffic. Otherwise, they won't be allowed to talk to certain other components (e.g., the kube-apiserver of the shoot).
```

docs/extensions/controlplane.md
```diff
@@ -56,6 +56,11 @@ The second one contains the output of the [`Infrastructure` resource](infrastruc
...
+Please [see the documentation](https://github.com/gardener/gardener/tree/master/docs/usage/network_policies.md) for more information.
```
rfranzke:
```suggestion
Please [see the document](https://github.com/gardener/gardener/tree/master/docs/usage/network_policies.md) for more information.
```

docs/extensions/controlplane.md
```diff
@@ -56,6 +56,11 @@ The second one contains the output of the [`Infrastructure` resource](infrastruc
 In order to support a new control plane provider you need to write a controller that watches all `ControlPlane`s with `.spec.type=<my-provider-name>`.
 You can take a look at the below referenced example implementation for the Alicloud provider.
 
+The control plane provider as part of the `ControlPlane` reconciliation, often deploys resources (e.g pods/deployments) into the Shoot namespace in the `Seed`.
```
rfranzke:
```suggestion
The control plane controller often deploys resources (e.g pods/deployments) into the Shoot namespace in the `Seed` as part of its `ControlPlane` reconciliation loop.
```

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
+# Network Policies in Gardener
```
rfranzke: Can you please add the document to the index (docs/README.md)?

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+
+As `Seed` clusters can host the `ControlPlane` of many `Shoot` clusters, it is necessary to isolate the `ControlPlanes` from each other.
```
rfranzke:
```suggestion
As `Seed` clusters can host the control planes of many `Shoot` clusters, it is necessary to isolate the control planes from each other for security reasons.
```

replicate the same across the doc, `ControlPlane` is usually used when the `ControlPlane` extension CRD is meant.

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+Besides deploying each `ControlPlane` in its own namespace, Gardener creates [network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) to also isolate the networks. Essentially, network policies make sure that pods can only talk to other pods over the network they are supposed to.
+As such, network policies are an important part of Gardener's tenant isolation.
+
+Apart from deploying network polices into each `Shoot` namespace hosting the `ControlPlane`, Gardener also deploys policies
+ - into the namespace dedicated to Gardener system components. This namespace is often called `garden` and contains components such as the [Gardenlet](https://github.com/gardener/gardener/blob/15cae57db802cbe460ff4cb3f80c26b2fc15e26f/docs/concepts/gardenlet.md).
```
rfranzke: "system components" is usally used for the `kube-system` pods in the shoot. Can we find another name? I don't have a good idea right now, maybe something like "seed-wide global controllers" or something similar.

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+ - into the `kube-system` namespace in the Shoot.
+
+Gardener deployed network policies have names describing their purpose, so this document does not describe each policy in detail.
```
rfranzke:
```suggestion
Network policies deployed by Gardener have names and annotations describing their purpose, so this document does only highlight a subset of the policies in detail.
```

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+## Network policies in the `Shoot` namespace in the `Seed`
+
+The network policies in the `Shoot` namespace in the `Seed` can roughly be grouped into policies required for the control plane components and for logging & monitoring.   
+
+The network policy `deny-all` plays a special role. This policy [denies all ingress and egress traffic](https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-deny-all-ingress-and-all-egress-traffic) from each pod in the `Shoot` namespace.
+So per default, a pod running in the control plane cannot talk to any other pod via IP or indirectly over DNS.
```
rfranzke:
```suggestion
So per default, a pod running in the control plane cannot talk to any other pod in the whole seed.
```

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+This means the pod needs to have appropriate network policies allowing it to talk to exactly the components required to execute its desired functionality.
```
rfranzke:
```suggestion
This means the pod needs to have labels matching to appropriate network policies allowing it to talk to exactly the components required to execute its desired functionality.
```

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+[This has also implications for Gardener extension](#implications-for-gardener-extensions) that need to deploy additional components into the `Shoot's` control plane.
```
rfranzke:
```suggestion
[This has also implications for Gardener extensions](#implications-for-gardener-extensions) that need to deploy additional components into the `Shoot's` control plane.
```

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+
+This section highlights a selection of network policies that exist in the `Shoot` namespace in the `Seed` cluster.
+In general, the control plane components serve different purposes and thus need access to different pods and network ranges.
+
+In contrast to other network policies, the policy `allow-to-shoot-networks` is tailored to the individual `Shoot` cluster, 
+because it is based on the network configuration in the `Shoot` manifest.
+It allows pods with the label `networking.gardener.cloud/to-shoot-networks=allowed` to access pods in the `Shoot` pod, 
+service and node CIDR range. This is used by the `Shoot` API Server and the prometheus pods to communicate over VPN/proxy with pods in the Shoot cluster.
+
+The policy `allow-to-blocked-cidrs` allows pods with the label `networking.gardener.cloud/to-blocked-cidrs=allowed` to access IPs that are explicitly blocked for all control planes in a `Seed`cluster (configurable via `seed.Spec.networks.blockCIDRS`). This is used for instance to block the cloud provider's metadata service.
```
rfranzke:
```suggestion
The policy `allow-to-blocked-cidrs` allows pods with the label `networking.gardener.cloud/to-blocked-cidrs=allowed` to access IPs that are explicitly blocked for all control planes in a `Seed` cluster (configurable via `spec.networks.blockCIDRS`). This is used for instance to block the cloud provider's metadata service.
```

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+
+Another network policy to be highlighted is `allow-to-seed-apiserver`.
+Some components need access to the `Seed` API Server. This can be allowed by labeling the pod with `networking.gardener.cloud/to-seed-apiserver=allowed`.
+This policy allows exactly the IPs of the `kube-apiserver` of the `Seed`.
+While all other policies have a static set of permissions (do not change during the lifecycle of the `Shoot`), the policy `allow-to-seed-apiserver` is reconciled to reflect the endpoints in the `default` namespace.
+This is required because endpoint IPs are not stable (think of scaling the `Seed` API Server pods or hibernating the `Seed` cluster in a local development environment).
```
rfranzke:
```suggestion
This is required because endpoint IPs are not necessarily stable (think of scaling the `Seed` API Server pods or hibernating the `Seed` cluster in a local development environment).
```

Usually, the IPs are stable, e.g., in mastered clusters or when the API server is backed by a load balancer. They are not necessarily stable when the seed is a shoot as well because of the "special" network topology.

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+
+Further the following network policies exist in the `Shoot` namespace.
```
rfranzke:
```suggestion
Furthermore, the following network policies exist in the `Shoot` namespace.
```

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+
+### Network policies for Logging & Monitoring 
+
+[Gardener provides out-of-the-box logging and monitoring](https://github.com/gardener/gardener/blob/master/docs/extensions/logging-and-monitoring.md) for the `ControlPlane` via a stack based on prometheus, elasticsearch, kibana and grafana.
+
+{{% notice tip %}}
+<p>Gardener currently introduces a logging stack based on [Loki](https://github.com/grafana/loki). So this section is subject to change. Please checkout [the Comunity Meeting for more information](https://www.youtube.com/watch?v=345b8xCcB-U&t=1166s).</p>
+{{% /notice %}}
```
rfranzke: Which kind of syntax is this? It's not properly rendered in the markdown: https://github.com/gardener/gardener/blob/d90cb6f1646d3078be69a4883770db3ff69d451f/docs/usage/network_policies.md

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+allow-kibana                      app=kibana-logging,garden.sapcloud.io/role=logging,role=logging 
+allow-prometheus                  app=prometheus,garden.sapcloud.io/role=monitoring,role=monitoring        
+allow-to-aggregate-prometheus     networking.gardener.cloud/to-aggregate-prometheus=allowed 
+allow-to-elasticsearch            networking.gardener.cloud/to-elasticsearch=allowed  
+allow-elasticsearch               app=elasticsearch-logging,garden.sapcloud.io/role=logging,role=logging   
+```
+
+Let's take for instance a look at the network policy `from-prometheus`.
+As part of the shoot reconciliation flow, Gardener deploys a shoot-specific Prometheus into the shoot namespace. 
+Each pod that should be scraped for logs must be labeled with `networking.gardener.cloud/from-prometheus=allowed` to allow incoming network requests by the prometheus pod.
```
rfranzke:
```suggestion
Each pod that should be scraped for metrics must be labeled with `networking.gardener.cloud/from-prometheus=allowed` to allow incoming network requests by the prometheus pod.
```

docs/usage/network_policies.md
```diff
@@ -0,0 +1,148 @@
...
+Most components of the `Shoot` cluster's control plane are therefore labeled appropriately. This includes the pods for the `cloud-controller-manager`, `kube-apiserver`, `kube-controller-manager`, `kube-scheduler` and `machine-controller-manager`.
```
rfranzke:
```suggestion
```

docs/usage/network_policies.md
rfranzke: Also, should this document be moved to `docs/extensions`? Or `docs/development`? It's not really related to the "usage of Gardener" (other than the small section about the network policies in the shoot because that's actually what might be interesting to end-users using shoots)?

rfranzke (commented):
Generally, I appreciate the document, thanks a lot. However, in my opinion it is way too detailed (which components uses which policies and why, etc.). This might make it hard to maintain. Wouldn't it be a better idea to actually list all then network policies deployed by Gardener and quickly explain their purpose, maybe with selective examples for components that are highly likely to stay, e.g., the kube-controller-manager wanting to talk to the shoot API server; but not the dependency-watchdog using "this and that" network policy to do "this and that"?

docs/usage/network_policies.md
danielfoehrKn: I was also unsure where to place it. I guess `docs/development` makes more sense

danielfoehrKn:
Thanks for the review.

> however, in my opinion it is way too detailed (which components uses which policies and why, etc.). 

Mhm ok. Thinking from a user perspective, that would be the most helpful to me. But maybe this could also be placed on the external website as a tutorial or something ("Understanding network policies in Gardner" ) WDYT @g-pavlov 

> This might make it hard to maintain.

I think we should strive for documentation that is helpful for the user/dev to get a general understanding vs having low touch docs.

> Wouldn't it be a better idea to actually list all then network policies deployed by Gardener and quickly explain their purpose

The document lists all the network policies, but they are grouped by where they are deployed to. So in any case when there is a new / deleted Network policy we would need to change the document.

Basically I tried to not cover each and every policy in detail but to only highlight certain policies to explain their purpose and why they exist.

rfranzke:
Yes, but we have to find a good balance so that the docs are still helpful, but not too detailed. With a higher likelihood the latter will lead to outdated, confusing docs which will invalidate its helpfulness. I don't suggest to rewrite the entire document, but maybe you can go over it once again and check for parts that are "too detailed" (e.g, the sentence about the dependency watchdog), and simply remove them. WDYT?

zanetworker:
@rfranzke I do agree that for `usage` purposes not all the documentation in this PR is required, however, I still think that we should keep it, my suggestion is to move part of it to `concepts/network-policies` but I also don't mind if we keep it where it is.

danielfoehrKn:
I split up the document into 
- a brief Conceptional part under `docs/concepts`
- a more detailed section for Network policies deployed to the Seed under `docs/development` 
- a brief section in `docs/usage` for the network policies deployed to the `kube-system` namespace in the Shoot

rfranzke (approved):
Thank you very much for incorporating the suggestions, I think it's way better structured and clear now!
/lgtm

Issue #2354, opened 2020-05-19
☂️Integrate vSMP MemoryOne 


hardikdr:
**How to categorize this issue?**

/area usability
/area os
/kind epic
/kind enhancement
/priority normal

**What would you like to be added**: 
This is an umbrella issue for the integration of vSMP MemoryONE operating-system based machines with Gardener. 

MemoryONE helps in the expansion of memory using non-volatile storage-class memory, which is cost-effective against expensive DRAMs.

MemoryONE is a virtual machine technology that transparently expands the system memory and then runs a standard Linux-based operating-system.
We have been requested to support jeos/chost-sles for the same. Please find more information on the product [here](https://www.scalemp.com/products/vsmp-memoryone/).

Please find the detailed process [here](https://files01.scalemp.com/downloads/others/EC2/docs/vSMP_MemoryONE_for_EC2-userguide.pdf), around how to boot up the machine based on MemoryONE VM. 

The following are the main changes required for the complete integration, which can also be used to track the progress. 
- [x] Adapt MCM to support multiple-disks for AWS. @hardikdr 
      - Tracking issue: https://github.com/gardener/machine-controller-manager/issues/462

- [x] Adapt chost/jeos extension controller to prepend the MemoryONE user data. @rfranzke  
       - Tracking issue: https://github.com/gardener/gardener-extension-os-suse-jeos/issues/18
       -  Implementation PR (Gardener): https://github.com/gardener/gardener/pull/2420
       -  Implementation PR (SuSE JeOS extension): https://github.com/gardener/gardener-extension-os-suse-jeos/pull/19

- [x] Allow setting the env-variable in kube-scheduler to overwrite the max-attachable volumes to machines. Investigate related configurations for CSI-drivers. @rfranzke 
       -  Tracking issue: https://github.com/gardener/gardener/issues/2343
       -  Implementation PR: https://github.com/gardener/gardener/pull/2413

- [x] Adapt AWS-worker extension controller to support multiple disks. @rfranzke 
       -  Implementation PR: https://github.com/gardener/gardener-extension-provider-aws/pull/105

- [x] Adapt AWS-worker extension controller to support disks with snapshots. @rfranzke 
       -  Implementation PR: https://github.com/gardener/gardener-extension-provider-aws/pull/112

- [x] Adapt/Verify Gardener APIs to allow setting multiple volumes and customization configs for AWS via shoot-resource. @rfranzke @hardikdr 
[...]

PR #1261, opened 2019-07-25
Support for worker pool OSC customization

pablochacin:
**What this PR does / why we need it**:
Add support for specifying a ProviderConfig for worker pool
images which is passed as part of the OSC and can be used
to customize it.

For example: 
```
apiVersion: core.gardener.cloud/v1alpha1
kind: Shoot
metadata:
  ...
spec:
  workers:
  - name: hana-worker-pool
    machineImage:
      name: suse-jeos
      version: 1234
      providerConfig:
        apiVersion: suse-jeos.operatingsystemconfig.extensions.gardener.cloud/v1alpha1
        kind: Customization
        environment: hana
    minimum: 5
    maximum: 250
```

This provider config is passed to the `OperatingSystemConfig`

```
apiVersion: extensions.gardener.cloud/v1alpha1
kind: OperatingSystemConfig
metadata:
  ...
spec:
  type: suse-jeos
  purpose: provision
  files:
    ...
  units:
    ...
  providerConfig:
    apiVersion: suse-jeos.operatingsystemconfig.extensions.gardener.cloud/v1alpha1
    kind: Customization
    environment: hana
``` 

**Special notes for your reviewer**:

This PR has not yet been tested due to issues setting a development environment.

**Release note**:

```noteworthy user
It is now possible to specify machine image specific configuration to customize the operating system. However, this is only possible if the respective operating system controller supports this (or, more generally, only what the controller supports can be customized). The `machineImage` section in the `Shoot` resource does now have a `providerConfig` field that can be filled with operating system specific configuration. Please consult the documentation of the respective OS-controller you are using.
```

PR #2438, opened 2020-06-09
Fix transportation of machine image providerConfig to OSC

rfranzke:

/area os
/kind bug
/priority normal

With #1261 the machine image specific provider config was introduced, however, it never worked because it wasn't correctly transported to the `OperatingSystemConfig` resource. This PR fixes the bug.

Ref #2354 

Thanks @vpnachev for spotting!
/assign @vpnachev 
/invite @vpnachev 

**Release note**:

```improvement user
A bug has been fixed that prevented the correct transportation of the machine image specific configuration in `.spec.provider.workers[].machine.image.providerConfig` to the respective extension controller.
```


vpnachev (approved):
/lgtm

PR #2444, opened 2020-06-10
Remove fixed CIDRs of integration test shoots (and rely on defaulting)

dguendisch:
/area testing
/kind enhancement
/priority normal

Removes the fixed CIDRs for pods and services so that tests can rely/reuse a seed's default CIDRs. This will help to e.g. use different CIDRs for alicloud vs. other cloudproviders without having to configure special CIDRs for a TestRun's dimension.


rfranzke (approved):
/lgtm
makes sense, thank you!

PR #2445, opened 2020-06-10
[release-v1.5] Fix deletion case for extension predicates

timuthy:

/area quality
/kind bug
/priority normal

This PR fixes the deletion of shoots or their extension resources in the seed once shoots were in a `Failed` state.

**Release note**:

```improvement operator
An issue has been fixed which caused failed shoot clusters to not be deleted successfully after a retry had been triggered.
```

### DIFF
--- a/extensions/pkg/controller/backupbucket/controller.go
+++ b/extensions/pkg/controller/backupbucket/controller.go
@@ -67,10 +67,6 @@ func DefaultPredicates(ignoreOperationAnnotation bool) []predicate.Predicate {
 			extensionspredicate.LastOperationNotSuccessful(),
 			extensionspredicate.IsDeleting(),
 		),
-		extensionspredicate.Or(
-			extensionspredicate.HasOperationAnnotation(),
-			predicate.GenerationChangedPredicate{},
-		),
 	}
 }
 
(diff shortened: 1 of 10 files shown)


danielfoehrKn (approved):

rfranzke:
Maybe I missed it, but why isn't it a 1:1 backport of #2435? This is a different commit now?

danielfoehrKn:
@rfranzke I think this is  the "safer" version (only adding a predicate to make the deletion case working) [from there](https://github.com/gardener/gardener/pull/2435#issuecomment-641920158).
Though I am not sure, thinking about it, if it would not be better to backport the exact same commit.

timuthy:
Well, that's exactly what we discussed [here](https://github.com/gardener/gardener/pull/2435#issuecomment-641920158) after the suggestion from @ialidzhikov and [your response](https://github.com/gardener/gardener/pull/2435#issuecomment-641971043).

rfranzke:
I see, probably I missed this point, sorry. I'm fine with this PR if you think it makes more sense, but why can't we simply take the same fix that we did for the `master` branch?

rfranzke (approved):
/lgtm
/ok-to-test

timuthy:
@rfranzke I have no strong opinion but know it was confusing to apply a different fix as a back port because it looked differently. Now, I applied the exact same changes we had applied to the `master` branch.

rfranzke:
Thanks Tim!

PR #2446, opened 2020-06-10
Auto-delete stale projects

### LINKED ISSUE #1997: Clean up empty projects after a period of inactivity
Deletion of empty projects after an extended period of inactivity.

**Why is this needed**:
On our Canary landscape I see a number of projects that are empty and have no signs of activity for the last 3 weeks (at least). Furthermore, some of these projects occupy apparently "desired" names (something short like `test`, `1`, `project` or similar), resulting in an increasing amount of projects with various suffixes added. Some of these projects, too, seem abandoned.

Ideally, however, we'd motivate the project creators to choose more descript project-names to avoid getting in each other's ways - and to delete them once they are not needed anymore.

rfranzke:

/area usability
/area performance
/kind enhancement
/priority normal

This PR enhances the `gardener-controller-manager` to auto-delete stale `Project`s after a configurable amount of time. `Project`s are considered stale when all of the following conditions apply:

* The `Project` is older than the configured `minimumLifetimeDays` value.
* The `Project`'s namespace does not contain any `Shoot` resources.
* The `Project`'s namespace does not contain any `Plant` resources.
* The `Project`'s namespace does not contain any `BackupEntry` resources.
* No `Secret` resource  in the `Project`'s namespace is referred by a `SecretBinding` resource that is in use by a `Shoot` resource .
* No `Quota` resource in the `Project`'s namespace is referred by a `SecretBinding` resource that is in use by a `Shoot` resource .

After the configured `staleGracePeriodDays` of a stale `Project` is exceeded a timestamp for auto-deletion will be assigned to the `Project` (based on the configured `staleExpirationTimeDays` value).
After the assigned auto-deletion timestamp is expired the `Project` will be deleted by the gardener-controller-manager.

Purposes of the configuration option:
* `minimumLifetimeDays`: Don't consider newly created `Project`s as "stale" too early to give people/end-users some time to onboard and get familiar with the system.
* `staleGracePeriodDays`: Don't compute auto-delete timestamps for stale `Project`s too early to not unnecessarily make people/end-users nervous "just because" they haven't actively used their `Project` for a given amount of time.
* `staleExpirationTimeDays`: Expiration time after which stale `Project`s are finally auto-deleted.

Fixes #1997

* Differently than discussed in #1997 I did go for fields in the `.status` (instead of annotations or conditions). The reason is mainly the the status cannot be tampered by end-users (annotation can), and the conditions concept didn't really fit for the use-case.
* ~✅ While my local dev. tests worked well so far, I still will implement some unit tests and add proper documentation. I opened the PR as draft in case somebody has early feedback on the implementation.~

**Release note**:

```noteworthy operator
The gardener-controller-manager does now automatically delete stale `Project`s that are no longer in use. By default, only `Project`s older than `30d` which are unused for at least `14d` will be auto-deleted after `90d`. However, the concrete values depend on the configuration of the respective Gardener landscape. You can find more information in [this document](https://github.com/gardener/gardener/blob/master/docs/concepts/controller-manager.md#project-controller).
```


pkg/controllermanager/controller/project/project_stale_control.go
```diff
@@ -0,0 +1,343 @@
...
+		projectLogger = newProjectLogger(project)
+	)
+
+	projectLogger.Infof("[STALE PROJECT RECONCILE]")
+
+	// Skip projects that are not older than the configured minimum lifetime in days. This allows having Projects for a
+	// certain period of time until they are checked whether they got stale.
+	if project.CreationTimestamp.UTC().Add(time.Hour * 24 * time.Duration(*c.config.Controllers.Project.MinimumLifetimeDays)).After(NowFunc().UTC()) {
+		projectLogger.Infof("[STALE PROJECT RECONCILE] Project is not older than the configured minimum %d days lifetime (%v), considering it 'not stale'", *c.config.Controllers.Project.MinimumLifetimeDays, project.CreationTimestamp.UTC())
+		return c.markProjectAsNotStale(ctx, project)
```
timuthy: Can you elaborate why it's necessary to mark a "new" project as not stale?

timuthy (commented):
Very nice PR!!! Just one question

pkg/controllermanager/controller/project/project_stale_control.go
rfranzke: Thanks for the review. My thoughts were that we don't want to "frighten" people that create new projects with an immediate "stale" notification. But this whole thing is configurable, so if someone doesn't want this behaviour then simply `minimumLifetimeDays` can be set to `0`.

timuthy: Sure, I get the motivation and it's good to have it. My comment was more logic wise because I thought a "new" project doesn't have stale timestamps at all, does it? But you probably mark it as not stale every time in case the `MinimumLifetimeDays` in the configuration has been prolonged, right? 👀

rfranzke: Ah, I see. Yes, exactly! It was created as part of @tim-ebert's thoughts (https://github.com/gardener/gardener/issues/1997#issuecomment-614436902) to support changing these fields dynamically.

rfranzke:
I've finished the unit tests and added documentation.

/unassign
/ready
/remove needs/tests needs/documentation status/author-action
/needs review

timuthy:
Thanks @rfranzke, I'm taking another look.

docs/concepts/controller-manager.md
```diff
@@ -0,0 +1,58 @@
...
+
+If a project is considered "stale" then its `.status.staleSinceTimestamp` will be set to the time when it was first detected to be stale.
+If it gets actively used again this timestamp will be removed.
+After some time the `.status.staleAutoDeleteTimestamp` will be set to a timestamp after which Gardener will auto-delete the `Project` resource if it still is not actively used.
+
+The component configuration of the Gardener Controller Manager offers to configure the following options:
+* `minimumLifetimeDays`: Don't consider newly created `Project`s as "stale" too early to give people/end-users some time to onboard and get familiar with the system. The "stale project" reconciler won't set any timestamp for such `Project`s.
+* `staleGracePeriodDays`: Don't compute auto-delete timestamps for stale `Project`s too early to not unnecessarily make people/end-users nervous "just because" they haven't actively used their `Project` for a given amount of time.
+* `staleExpirationTimeDays`: Expiration time after which stale `Project`s are finally auto-deleted.
+
```
timuthy, line 56: Can you add an explanation about what happens if the operator changes these values for already processed projects?

timuthy (commented):
I don't have any further comments except the nit. Nice PR!

pkg/controllermanager/controller/project/project_stale_control_test.go
```diff
@@ -0,0 +1,376 @@
...
+
+						project.Status.StaleSinceTimestamp = &oldTime
+						project.Status.StaleAutoDeleteTimestamp = &oldTime
+
+						expectStaleMarking(k8sGardenRuntimeClient, k8sGardenRuntimeStatusWriter, project, &staleSinceTimestamp, &staleAutoDeleteTimestamp)
+
+						Expect(control.ReconcileStaleProject(project)).To(Succeed())
+					})
+
+					It("should should delete the project if the auto delete timestamp is exceeded", func() {
```
ialidzhikov:
```suggestion
					It("should delete the project if the auto delete timestamp is exceeded", func() {
```

pkg/controllermanager/controller/project/project_stale_control_test.go
```diff
@@ -0,0 +1,376 @@
...
+							staleAutoDeleteTimestamp = metav1.Time{Time: staleSinceTimestamp.Add(24 * time.Hour * time.Duration(staleExpirationTimeDays))}
+						)
+						project.Status.StaleSinceTimestamp = &staleSinceTimestamp
+
+						expectStaleMarking(k8sGardenRuntimeClient, k8sGardenRuntimeStatusWriter, project, &staleSinceTimestamp, &staleAutoDeleteTimestamp)
+
+						Expect(control.ReconcileStaleProject(project)).To(Succeed())
+					})
+
+					It("should should extend the project's stale grace period", func() {
```
ialidzhikov:
```suggestion
					It("should extend the project's stale grace period", func() {
```

pkg/controllermanager/controller/project/project_stale_control_test.go
```diff
@@ -0,0 +1,376 @@
...
+			control StaleControlInterface
+		)
+
+		BeforeSuite(func() {
+			oldNowFunc = NowFunc
+			oldTimeNowFunc = common.TimeNow
+		})
+
+		AfterSuite(func() {
+			NowFunc = oldNowFunc
```
ialidzhikov: Wouldn't be better if we have the nowFunc as arg of `ReconcileStaleProject` func?


```golang
type StaleControlInterface interface {
	ReconcileStaleProject(project *gardencorev1beta1.Project, nowFunc func() metav1.Time) error
}
```

In this way we could prevent preserving the oldNowFunc in BeforeSuite and setting it back in AfterSuite?

pkg/controllermanager/controller/project/project_stale_control.go
```diff
@@ -0,0 +1,349 @@
...
+			}
+		}
+
+		names.Insert(secret.Name)
+	}
+
+	return names
+}
+
+// computeQuotaNames determines the names of Quotas that are of type Opaque and don't have owner references to a
```
ialidzhikov:
```suggestion
// computeQuotaNames determines the names of Quotas from the given slice.
```

ialidzhikov (dismissed):
Nice PR!
/lgtm

pkg/controllermanager/controller/project/project_stale_control_test.go
rfranzke: ok, sure, thanks, done

ialidzhikov (approved):
/lgtm

timebertt (approved):
Looks great! Thanks for adapting my suggestions :)
/lgtm

Issue #2414, opened 2020-06-03
Use cached controller-runtime client


timebertt:
**How to categorize this issue?**

/area quality networking cost
/kind enhancement

**What would you like to be added**:

In g/g we have an increasing number of API calls that are made via controller-runtime clients instead of using more efficient mechanisms like `Informers` or `Listers`.
In our current implementation, these clients — no matter if for garden, seed, plant or shoot cluster — are not backed by a cache, which means, that every call will result in a direct call to the respective API server. Because most of these calls go to API servers that are not running on the same cluster / in the same "network" (e.g. to the garden cluster or to the (shooted) seed's API server) like the caller (mainly gardenlet), they contribute to "external" traffic, which is (highly) priced by cloud providers.

Also, there are many different places and mechanisms in our code, where and how we construct such ClientSets, ChartRenderes, ChartAppliers, RESTMappers and so on, and they are very short-lived (e.g often constructed only for a single operation and discarded after the operation has completed).

As we will probably introduce more and more API calls via controller-runtime clients (because it is less complicated to use and allows more generic constructs), we should refactor how we construct/retrieve those clients and equip them with a cache, so we can save unnecessary network traffic (costs).

Related to/part of #1953

## Steps

#2822 tracks all steps for graduating the `CachedRuntimeClients` feature gate, but they can also be considered a part of this issue.

These are steps/subtasks of this issue, that are not strictly required for or related to the graduation of the feature gate but rather general improvements regarding cached controller-runtime clients:

- [x] Harmonize/centralize `RESTMapper` creation and get rid of disk cached discovery clients (#2415)
- [x] Harmonize/centralize `Applier`, `ChartRenderer` and `ChartApplier` creation (#2417)
- [x] Add a cached controller-runtime client to our clientset in `pkg/client/kubernetes` (~~will be used by default, so this will result in a breaking change for all callers~~ will be put behind a feature gate, so we can disable it if we encouter some problems while running it in a productive system) (#2449)
- [x] Add some form of a registry (`ClientMap`) which can be used by all components/controllers to retrieve clients to arbitrary clusters in a harmonized way and which keeps track of all constructed clients (#2449)
- [x] Adapt the main components (gcm, gardenlet) and all their controllers to make use of the `ClientMap` (#2449)
- [x] Adapt usages in the integration test framework (#2474)
- [x] Eliminate usages of generated client sets in `kubernetes.Interface` wherever appropriate in favor of cached controller-runtime clients
[...]

PR #2449, opened 2020-06-11
Use cached controller-runtime client

timebertt:

/area quality
/area networking
/area cost
/kind enhancement
/priority normal
/size xl

This PR refactors the way how kubernetes clients are constructed and used in g/g.
The main motivation behind it is to make use of controller-runtime clients leveraging informer-backed caches to reduce network traffic (see #2414 for more details).

As the clientsets we use are very short-lived (e.g. Seed/Shoot clients often only for a single reconciliation) and can therefore not benefit of cached clients, some form of a central registry for storing, requesting and removing ClientSets for arbitrary clusters gardener talks to is needed. Therefore this PR introduces the concept of a `ClientMap` which stores all the different clients, gcm and gardenlet use, to keep their caches alive and running. Consequently it adapts all gcm/gardenlet controllers to make use of a `ClientMap`.

Main part of #2414 

I split this PR into multiple commits, so it's a bit easier to review. The ...
1. commit adds a cached runtime client to our default ClientSet (`kubernetes.Interface`). Calls to `Interface.Client()` will return the new cached client, which has to be started before the first GET/LIST call by using the new `Start()` func on the ClientSet (otherwise those calls will return a `cache not started` error). Callers can use `DirectClient()` to retrieve a client which talks directly to the API server.
2. commit adds the new `ClientMap` concept. This is basically an interface which is implemented by different composable kinds of client maps like `SeedClientMap`, `ShootClientMap` and also `DelegatingClientMap` (which is some form of a "master ClientMap" that can be used to retrieve all kinds of clients). All implementations are backed by the same core logic of the `GenericClientMap`. It also adds Builders for constructing the `ClientMaps` on startup.
3. commit adds some fake implementations for ClientSets and ClientMaps and some unit tests for those.
4. and following commits adapt the client usages in all g/g components and integration tests.

**Release note**:

```noteworthy operator
A new FeatureGate `CachedRuntimeClients` has been added to `gardenlet`, `gardener-controller-manager` and `gardener-scheduler`. If enabled via the respective component config, the components use cached clients for their API calls wherever possible.
```
```action developer
The creation and usage of kubernetes clientsets in controllers across gardener/gardener has been refactored. Please use the newly introduced `ClientMap`s to retrieve clientsets for all kind of clusters instead of creating new clientsets each time to leverage the clients' caches.
```
```action developer
Kubernetes clientsets (`pkg/client/kubernetes.Interface`) now return a cached controller-runtime client from `Interface.Client()` if `kubernetes.UseCachedRuntimeClients` has been set to true (defaults to `false`). The cached clients have to be started before the first usage by a call to `Interface.Start()`. If you use a `ClientMap` to retrieve the clientset, this is done automatically.
```


timebertt:
Update:
- rebased, solved conflicts
- removed adaption of client usages in test framework, as @schrodit and I discovered that this part needs some more work and is better done in a separate PR

timebertt:
This sounds kind of rude, but
/uninvite @gardener/test-infra-maintainers 
😉

pkg/client/kubernetes/runtime_client.go
```diff
@@ -0,0 +1,174 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+const (
+	defaultCacheResyncPeriod = 6 * time.Hour
```
rfranzke, line 34: Just out of curiosity, why `6h`? :)

pkg/client/kubernetes/runtime_client.go
```diff
@@ -0,0 +1,174 @@
...
+// NewDirectClient creates a new client.Client which can be used to talk to the API directly (without a cache).
+func NewDirectClient(config *rest.Config, options client.Options) (client.Client, error) {
+	if err := setClientOptionsDefaults(config, &options); err != nil {
+		return nil, err
+	}
+
+	return newDirectClient(config, options)
+}
+
+func newDirectClient(config *rest.Config, options client.Options) (client.Client, error) {
```
rfranzke: This function seems not helpful to me/can be removed.

pkg/client/kubernetes/runtime_client.go
```diff
@@ -0,0 +1,174 @@
...
+// It also creates a new client cache, which will be started immediately
+func NewRuntimeClient(config *rest.Config, options client.Options, stopCh <-chan struct{}) (client.Client, error) {
+	if err := setClientOptionsDefaults(config, &options); err != nil {
+		return nil, err
+	}
+
+	return newRuntimeClient(config, options, stopCh)
+}
+
+func newRuntimeClient(config *rest.Config, options client.Options, stopCh <-chan struct{}) (client.Client, error) {
```
rfranzke: I find it a little confusing that there is `newRuntimeClient` that internally always calls `newRuntimeClientWithCache` which I wouldn't expect as a caller. Probably this can be easily resolved by renaming the functions appropriately?

pkg/client/kubernetes/runtime_client.go
```diff
@@ -0,0 +1,174 @@
...
+// for creating new controller-runtime clients with caches.
+func NewRuntimeCache(config *rest.Config, options cache.Options) (cache.Cache, error) {
+	if err := setCacheOptionsDefaults(&options); err != nil {
+		return nil, err
+	}
+
+	return newRuntimeCache(config, options)
+}
+
+func newRuntimeCache(config *rest.Config, options cache.Options) (cache.Cache, error) {
```
rfranzke: same like above, this function seems not to be helpful

pkg/client/kubernetes/fake/fake_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestFake(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Fake ClientSet Suite")
```
rfranzke:
```suggestion
	RunSpecs(t, "Client Kubernetes ClientSet Fake Suite")
```

pkg/client/kubernetes/clientmap/builder/builder_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestBuilder(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "ClientMap Builder Suite")
```
rfranzke:
```suggestion
	RunSpecs(t, "Client Kubernetes ClientMap Builder Suite")
```

(replicate for others as well, this helps to identify which suite is running when you use `ginkgo ./...` in the repo root

pkg/client/kubernetes/clientmap/builder/deleg_builder.go
```diff
@@ -0,0 +1,163 @@
...
+import (
+	"fmt"
+
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/internal"
+
+	"github.com/sirupsen/logrus"
+)
+
+type DelegatingClientMapBuilder struct {
```
rfranzke, line 28: doc string is missing

pkg/client/kubernetes/clientmap/builder/deleg_builder.go
```diff
@@ -0,0 +1,163 @@
...
+	gardenClientMapFunc func(logger logrus.FieldLogger) (clientmap.ClientMap, error)
+	seedClientMapFunc   func(gardenClients clientmap.ClientMap, logger logrus.FieldLogger) (clientmap.ClientMap, error)
+	shootClientMapFunc  func(gardenClients, seedClients clientmap.ClientMap, logger logrus.FieldLogger) (clientmap.ClientMap, error)
+	plantClientMapFunc  func(gardenClients clientmap.ClientMap, logger logrus.FieldLogger) (clientmap.ClientMap, error)
+
+	logger logrus.FieldLogger
+}
+
+func NewDelegatingClientMapBuilder() *DelegatingClientMapBuilder {
```
rfranzke, line 38: doc string is missing

pkg/client/kubernetes/clientmap/internal/deleg_clientmap.go
```diff
@@ -0,0 +1,111 @@
...
+	"context"
+	"fmt"
+
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
+)
+
+var _ clientmap.ClientMap = &DelegatingClientMap{}
+
+type DelegatingClientMap struct {
```
rfranzke: doc strings are missing for exported variables and functions

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
+)
+
+type gardenClientMap struct {
+	clientmap.ClientMap
+}
+
+func NewGardenClientMap(factory *GardenClientSetFactory, logger logrus.FieldLogger) clientmap.ClientMap {
```
rfranzke, line 37: doc strings are missing for exported variables and functions

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
```diff
@@ -0,0 +1,67 @@
...
+		kubernetes.WithClientOptions(client.Options{
+			Scheme: kubernetes.GardenScheme,
+		}),
+	)
+
+	if err != nil {
+		return nil, fmt.Errorf("failed constructing new client for Garden cluster: %w", err)
+	}
+
+	return clientSet, nil
```
rfranzke: Can this be simplified to 
```suggestion
	return NewClientSetWithConfig(
		kubernetes.WithRESTConfig(f.RESTConfig),
		kubernetes.WithClientOptions(client.Options{
			Scheme: kubernetes.GardenScheme,
		}),
	)
```

or is the error message from `NewClientSetWithConfig` not detailed enough?

(same for other occurrences in the clientmap pkg)

pkg/client/kubernetes/clientmap/internal/generic_clientmap.go
```diff
@@ -0,0 +1,210 @@
...
+)
+
+var _ clientmap.ClientMap = &GenericClientMap{}
+
+const (
+	waitForCacheSyncTimeout = 5 * time.Minute
+	maxRefreshInterval      = 5 * time.Second
+)
+
+type GenericClientMap struct {
```
rfranzke, line 42: doc strings are missing for exported variables and functions

pkg/client/kubernetes/clientmap/internal/generic_clientmap.go
```diff
@@ -0,0 +1,210 @@
...
+	clientSets map[clientmap.ClientSetKey]*clientMapEntry
+	factory    clientmap.ClientSetFactory
+
+	// mu guards concurrent access to clientSets
+	mu sync.RWMutex
```
rfranzke:
```suggestion
	lock sync.RWMutex
```

pkg/client/kubernetes/clientmap/internal/generic_clientmap.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func (cm *GenericClientMap) GetClient(ctx context.Context, key clientmap.ClientSetKey) (kubernetes.Interface, error) {
+	entry, started, found := func() (*clientMapEntry, bool, bool) {
+		cm.mu.RLock()
+		defer cm.mu.RUnlock()
+
+		entry, found := cm.clientSets[key]
+		return entry, cm.started, found
+	}()
```
rfranzke: Why is this wrapped into a func?

pkg/client/kubernetes/clientmap/internal/plant_clientmap.go
```diff
@@ -0,0 +1,82 @@
...
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
+)
+
+type plantClientMap struct {
+	clientmap.ClientMap
+}
+
+func NewPlantClientMap(factory *PlantClientSetFactory, logger logrus.FieldLogger) clientmap.ClientMap {
```
rfranzke, line 35: doc strings are missing for exported variables and functions

pkg/client/kubernetes/clientmap/internal/seed_clientmap.go
```diff
@@ -0,0 +1,101 @@
...
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
+)
+
+type seedClientMap struct {
+	clientmap.ClientMap
+}
+
+func NewSeedClientMap(factory *SeedClientSetFactory, logger logrus.FieldLogger) clientmap.ClientMap {
```
rfranzke, line 36: doc strings are missing for exported variables and functions

pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
```diff
@@ -0,0 +1,123 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	baseconfig "k8s.io/component-base/config"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type shootClientMap struct {
+	clientmap.ClientMap
+}
+
+func NewShootClientMap(factory *ShootClientSetFactory, logger logrus.FieldLogger) clientmap.ClientMap {
```
rfranzke, line 39: doc strings are missing for exported variables and functions

pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	clientOptions := client.Options{
+		Scheme: kubernetes.ShootScheme,
+	}
+
+	clientSet, err := NewClientFromSecret(ctx, seedClient, seedNamespace, secretName,
+		kubernetes.WithClientConnectionOptions(f.ClientConnectionConfig),
+		kubernetes.WithClientOptions(clientOptions),
+	)
```
rfranzke: Why is this executed always?

pkg/client/kubernetes/clientmap/builder/garden_builder.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/internal"
+
+	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/rest"
+)
+
+type GardenClientMapBuilder struct {
```
rfranzke, line 31: doc strings are missing for exported variables and functions

pkg/client/kubernetes/clientmap/builder/plant_builder.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/internal"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/keys"
+
+	"github.com/sirupsen/logrus"
+)
+
+type PlantClientMapBuilder struct {
```
rfranzke, line 31: doc strings are missing for exported variables and functions

pkg/client/kubernetes/clientmap/builder/seed_builder.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/sirupsen/logrus"
+	baseconfig "k8s.io/component-base/config"
+
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/internal"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/keys"
+)
+
+type SeedClientMapBuilder struct {
```
rfranzke, line 32: doc strings are missing for exported variables and functions

pkg/client/kubernetes/clientmap/builder/shoot_builder.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/sirupsen/logrus"
+	baseconfig "k8s.io/component-base/config"
+
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/internal"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/keys"
+)
+
+type ShootClientMapBuilder struct {
```
rfranzke, line 32: doc strings are missing for exported variables and functions

pkg/client/kubernetes/clientmap/fake/fake_builder.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
+	"github.com/gardener/gardener/pkg/client/kubernetes/fake"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ClientMapBuilder struct {
```
rfranzke, line 26: doc strings are missing for exported variables and functions

pkg/client/kubernetes/clientmap/fake/fake_clientmap.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
+	"github.com/gardener/gardener/pkg/client/kubernetes/fake"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ clientmap.ClientMap = &ClientMap{}
+
+type ClientMap struct {
```
rfranzke, line 31: doc strings are missing for exported variables and functions

cmd/gardener-controller-manager/app/gardener_controller_manager.go
```diff
@@ -243,19 +244,20 @@ func NewGardener(cfg *config.ControllerManagerConfiguration) (*Gardener, error)
...
-		kubernetes.WithClientOptions(
-			client.Options{
-				Scheme: kubernetes.GardenScheme,
-			}),
-	)
+	gardenClientMapBuilder := clientmapbuilder.NewGardenClientMapBuilder().
+		WithRESTConfig(restCfg)
+
+	clientMap, err := clientmapbuilder.NewDelegatingClientMapBuilder().
+		WithGardenClientMapBuilder(gardenClientMapBuilder).
```
rfranzke:
```suggestion
	clientMap, err := clientmapbuilder.
	  NewDelegatingClientMapBuilder().
		WithGardenClientMapBuilder(clientmapbuilder.NewGardenClientMapBuilder().WithRESTConfig(restCfg)).
```

pkg/operation/operation.go
```diff
@@ -280,65 +286,45 @@ func (o *Operation) InitializeSeedClients() error {
...
+	o.K8sShootClient = shootClient
 
-	o.ChartApplierShoot = k8sShootClient.ChartApplier()
 	return nil
 }
 
-func (o *Operation) controlPlaneHibernated() (bool, error) {
-	replicaCount, err := common.CurrentReplicaCount(o.K8sSeedClient.Client(), o.Shoot.SeedNamespace, v1beta1constants.DeploymentNameKubeAPIServer)
-	if err != nil {
+func (o *Operation) IsAPIServerRunning() (bool, error) {
```
rfranzke, line 312: doc string

pkg/operation/operation.go
```diff
@@ -280,65 +286,45 @@ func (o *Operation) InitializeSeedClients() error {
...
+	deployment := &appsv1.Deployment{}
+	if err := o.K8sSeedClient.DirectClient().Get(context.TODO(), kutil.Key(o.Shoot.SeedNamespace, v1beta1constants.DeploymentNameKubeAPIServer), deployment); err != nil {
```
rfranzke, line 315: Why you use the direct client here?

pkg/operation/common/utils.go
```diff
@@ -166,6 +166,24 @@ func ProjectForNamespace(projectLister gardencorelisters.ProjectLister, namespac
...
+		return nil, err
+	}
+
+	for _, project := range projectList.Items {
+		if project.Spec.Namespace != nil && *project.Spec.Namespace == namespaceName {
+			return &project, nil
+		}
+	}
+
+	return nil, apierrors.NewNotFound(gardencorev1beta1.Resource("Project"), fmt.Sprintf("for namespace %s", namespaceName))
```
rfranzke: Does it make sense to put this into a separate function that can be called from both `ProjectForNamespace` and `ProjectForNamespaceWithClient`?

pkg/operation/botanist/health_check.go
```diff
@@ -234,6 +235,18 @@ func shootHibernatedCondition(condition gardencorev1beta1.Condition) gardencorev
 	return gardencorev1beta1helper.UpdatedCondition(condition, gardencorev1beta1.ConditionTrue, "ConditionNotChecked", "Shoot cluster has been hibernated.")
 }
 
+func shootControlPlaneNotRunningCondition(lastOperation *gardencorev1beta1.LastOperation, condition gardencorev1beta1.Condition) gardencorev1beta1.Condition {
+	message := "Shoot control plane is not running at the moment."
+	switch {
+	case lastOperation == nil || lastOperation.Type == gardencorev1beta1.LastOperationTypeCreate:
+		message = "Shoot control plane has not been started yet."
```
```suggestion
		message = "Shoot control plane has not been fully created yet."
```

pkg/operation/botanist/health_check.go
```diff
@@ -234,6 +235,18 @@ func shootHibernatedCondition(condition gardencorev1beta1.Condition) gardencorev
...
+	case lastOperation.Type == gardencorev1beta1.LastOperationTypeDelete:
+		message = "Shoot control plane has already been deleted."
```
```suggestion
		message = "Shoot control plane has already been or is about to be deleted."
```

pkg/operation/botanist/controlplane.go
```diff
@@ -118,13 +119,24 @@ func (b *Botanist) deleteNamespace(ctx context.Context, name string) error {
...
 			Namespace: b.Shoot.SeedNamespace,
 		},
 	}
-	return client.IgnoreNotFound(b.K8sSeedClient.Client().Delete(ctx, deploy, kubernetes.DefaultDeleteOptions...))
+
+	if err := b.K8sSeedClient.Client().Delete(ctx, deploy, kubernetes.DefaultDeleteOptions...); client.IgnoreNotFound(err) != nil {
+		return err
+	}
+
+	return nil
```
```suggestion
	return client.IgnoreNotFound(b.K8sSeedClient.Client().Delete(ctx, deploy, kubernetes.DefaultDeleteOptions...))
```

pkg/operation/botanist/controlplane.go
```diff
@@ -201,7 +213,9 @@ func (b *Botanist) DeleteClusterAutoscaler(ctx context.Context) error {
 // * kube-apiserver
 // * kube-controller-manager
 func (b *Botanist) WakeUpControlPlane(ctx context.Context) error {
-	client := b.K8sSeedClient.Client()
+	// use direct client here, as cached client sometimes causes scale functions not to work properly
```
rfranzke, line 214: Hm, interesting. Is there an issue in the controller-runtime that we can link here? If not, could you open one?

pkg/operation/botanist/controlplane.go
```diff
@@ -355,21 +377,6 @@ func (b *Botanist) PrepareKubeAPIServerForMigration(ctx context.Context) error {
 	return b.DeleteKubeAPIServer(ctx)
 }
 
-func (b *Botanist) scaleControlPlaneDeploymentsToZero(ctx context.Context, c client.Client) error {
```
rfranzke, line 368: Why was this function removed? Wasn't it called anywhere?

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -688,7 +705,7 @@ func migrateDNSProviders(ctx context.Context, c client.Client, o *operation.Oper
 		}
 
 		// Remove functionless DNS providers
-		// TODO: timuhty - Validation should forbid to create functionless DNS providers in the future.
+		// TODO: timuthy - Validation should forbid to create functionless DNS providers in the future.
```
rfranzke, line 708: :-D

pkg/gardenlet/controller/backupentry/backup_entry_actuator.go
```diff
@@ -120,6 +120,9 @@ func (a *actuator) Delete(ctx context.Context) error {
 // by the current progress of the Flow execution.
 func (a *actuator) reportBackupEntryProgress(ctx context.Context, stats *flow.Stats) {
 	if err := kutil.TryUpdateStatus(ctx, kretry.DefaultRetry, a.gardenClient, a.backupEntry, func() error {
+		if a.backupEntry.Status.LastOperation == nil {
+			return fmt.Errorf("last operation of BackupEntry %s/%s is unset", a.backupEntry.Namespace, a.backupEntry.Name)
```
rfranzke, line 125: Hm, why is this an error? Can't you simply initialize it to allow it being set in the next lines?

pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go
```diff
@@ -117,6 +117,9 @@ func (a *actuator) Delete(ctx context.Context) error {
 // by the current progress of the Flow execution.
 func (a *actuator) reportBackupBucketProgress(ctx context.Context, stats *flow.Stats) {
 	if err := kutil.TryUpdateStatus(ctx, kretry.DefaultRetry, a.gardenClient, a.backupBucket, func() error {
+		if a.backupBucket.Status.LastOperation == nil {
+			return fmt.Errorf("last operation of BackupBucket %s/%s is unset", a.backupBucket.Namespace, a.backupBucket.Name)
```
rfranzke, line 122: Hm, why is this an error? Can't you simply initialize it to allow it being set in the next lines?

rfranzke (changes requested):
Honestly, kudos for this very well structured and readable PR, especially given its large size. I'm impressed! :)
I still want to play with it locally, and I don't really have much non-minor comments.

pkg/client/kubernetes/runtime_client.go
timebertt: No specific reason, maybe you can add your opinion on a sensible default.
I didn't want to go with no resync at all, but also didn't found any concrete points for a specific period.

pkg/client/kubernetes/clientmap/internal/deleg_clientmap.go
timebertt: Yes, right.
I'm wondering if these funcs and vars can be considered "exported" as they reside in the `internal` package. 🤔 
But will add doc strings anyways, would be good here, I guess.

pkg/client/kubernetes/runtime_client.go
rfranzke: I see, no problem, I don't have a suggestion as of now. I'm fine to start with it.

pkg/client/kubernetes/runtime_client.go
```diff
@@ -0,0 +1,174 @@
...
+	config, err := clientConfig.ClientConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	opts := append([]ConfigFunc{WithRESTConfig(config)}, fns...)
+	return NewDirectClientWithConfig(opts...)
+}
+
+// NewRuntimeClientWithConfig returns a new controller runtime client from a config.
```
ialidzhikov:
```suggestion
// NewDirectClientWithConfig returns a new controller runtime client from a config.
```

pkg/controllermanager/controller/shoot/shoot.go
```diff
@@ -65,7 +65,7 @@ type Controller struct {
 // NewShootController takes a Kubernetes client for the Garden clusters <k8sGardenClient>, a struct
```
ialidzhikov:
```suggestion
// NewShootController takes a collection of ClientSets <clientMap>, a struct
```

cmd/gardenlet/app/gardenlet.go
```diff
@@ -278,20 +280,32 @@ func NewGardenlet(ctx context.Context, cfg *config.GardenletConfiguration) (*Gar
...
+	seedClientMapBuilder := clientmapbuilder.NewSeedClientMapBuilder().
+		WithInCluster(cfg.SeedSelector == nil).
+		WithClientConnectionConfig(&cfg.SeedClientConnection.ClientConnectionConfiguration)
+	shootClientMapBuilder := clientmapbuilder.NewShootClientMapBuilder().
+		WithClientConnectionConfig(&cfg.ShootClientConnection.ClientConnectionConfiguration)
+
+	clientMap, err := clientmapbuilder.NewDelegatingClientMapBuilder().
+		WithGardenClientMapBuilder(gardenClientMapBuilder).
+		WithSeedClientMapBuilder(seedClientMapBuilder).
+		WithShootClientMapBuilder(shootClientMapBuilder).
```
ialidzhikov, line 296: Also looks like `WithLogger(logger).` needs to be specified for this builder as otherwise the gardenlet start fails with `Error: failed to build ClientMap: logger is required but not set`.

pkg/client/kubernetes/clientmap/builder/shoot_builder.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil, fmt.Errorf("seed client is required but not set")
+	}
+	if b.clientConnectionConfig == nil {
+		return nil, fmt.Errorf("clientConnectionConfig is required but not set")
+	}
+
+	return internal.NewShootClientMap(&internal.ShootClientSetFactory{
+		GetGardenClient:        b.gardenClientFunc,
+		GetSeedClient:          b.seedClientFunc,
+		ClientConnectionConfig: *b.clientConnectionConfig,
```
ialidzhikov, line 99: I think here we do not pass `Log` for `ShootClientSetFactory` as this is causing nil pointer.

ialidzhikov (commented):
Generally looks fine to me. I didn't know that the default controller-runtime `Client` is different from the one you get via the `GetClient()` from the manager with regards to the cache. I cannot judge every piece as probably I will need more time to go over it but posting few comments inline (I hope that they will be useful). Thank you also for the cleanup of the multiple `TODOs` and the unused code fragments.

cmd/gardener-controller-manager/app/gardener_controller_manager.go
```diff
@@ -243,19 +244,20 @@ func NewGardener(cfg *config.ControllerManagerConfiguration) (*Gardener, error)
...
+		WithPlantClientMapBuilder(clientmapbuilder.NewPlantClientMapBuilder()).
```
ialidzhikov, line 251: `WithLogger(logger).` needed also here.

timebertt:
Updates:
- addressed the first batch of review comments
- added a bunch of doc strings in `pkg/client/kubernetes/clientmap/*`
- `DelegatingClientMap`: test for nil before delegating calls to the contained ClientMaps
- rebased to fix conflicts

/status author-action

pkg/client/kubernetes/runtime_client.go
timebertt: Yep, done :)

pkg/client/kubernetes/clientmap/builder/builder_suite_test.go
timebertt: Agreed, done. I hope I didn't miss some suites.

pkg/client/kubernetes/clientmap/internal/generic_clientmap.go
timebertt: This is done to do an atomar read operation on the ClientMap while leveraging `defer`.
Basically the whole structure of the ClientMap package is heavily inspired by the structure of the controller-runtime cache package and they also do it like this in some places (e.g. [here](https://github.com/kubernetes-sigs/controller-runtime/blob/c4893e0b992fb1e6e7da29acd692a0cdf9d7aec9/pkg/cache/internal/informers_map.go#L170-L175)).
I have also seen IIFEs in other places of the kubernetes and golang packages and was confused at first. But on the second read I found it to make it easier to directly see which part of the code is executed while holding a lock (kind of like `synchronized` blocks in Java) 😄

pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
timebertt: I am not sure, what you mean... 😅

pkg/operation/operation.go
timebertt: Basically just to make sure, we're not reading from a stale cache, when checking if we should initialize a shoot client in the care controller. But no strong opinion here.

pkg/operation/common/utils.go
timebertt: Yes, it would make sense. But I don't see an easy and clean solution how :/
As `projectLister.List` returns `[]*gardencorev1beta1.Project` and `client.List()` writes to a `*gardencorev1beta1.ProjectList{}` which contains a `[]gardencorev1beta1.Project`, this helper function would need to accept a `interface{}` param, which would have to be casted again. Other option would be to copy all items of a list in one function to a slice of a type accepted by the helper function...

pkg/operation/botanist/controlplane.go
timebertt: Hmm, actually I think, this is more a problem on our side in the `kubernetes.Scale*` funcs. But I didn't investigate further, as I thought, using the direct client for such calls, that are only executed once/twice per day per Shoot should not harm too much.
What we could try, is rather use `client.Patch` for setting the replica count instead of `client.Update` as long as there is no support for the `scale` subresource in controller-runtime (see https://github.com/kubernetes-sigs/controller-runtime/issues/172), but I'm not actually sure, that this would help.

pkg/operation/botanist/controlplane.go
timebertt: Yes, I didn't see any usages.

pkg/gardenlet/controller/backupentry/backup_entry_actuator.go
timebertt: Hmm, `ReportShootProgress` also does it like this:
https://github.com/gardener/gardener/blob/ed1dd896ddc16a64e47714c08531799d6ea93dde/pkg/operation/operation.go#L408-L410
I just did it analogously...
We could initialize `Status.LastOperation` here, but this is already done before starting the flow. I just added it here, because I hit nil pointers here caused by stale cache reads.

pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go
timebertt: See above

pkg/controllermanager/controller/shoot/shoot.go
timebertt: Oh man, that was quite outdated 😄

cmd/gardenlet/app/gardenlet.go
timebertt: Good catch!

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
timebertt: Yep, done.

cmd/gardenlet/app/gardenlet.go
timebertt: Done, gcm and gardenlet should be starting up just fine again.

timebertt:
Thanks for your reviews @rfranzke and @ialidzhikov!
I have addressed most comments now, PTAL :)

/remove status/author-action
/needs review

ialidzhikov:
Could you please check `make verify`?

pkg/client/kubernetes/clientmap/internal/generic_clientmap.go
rfranzke: I see, thanks for the explanation!

pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
rfranzke: Sorry.. I meant: In line 113 you check if the `gardener-internal` secret can be used, so the lines 98-101 could have been put in the `else` block. But probably it was like this before and the clientset creation does not cause any "harm" in the form of I/O, so let's keep it like this. I'll resolve this comment.

pkg/operation/operation.go
rfranzke: Hm ok.. agreed, no other opinion either. Maybe you can add a small comment that explains why the direct client is used. 👍

pkg/operation/common/utils.go
rfranzke: Yes, I actually meant following the second approach you mentioned. I think I have done something similar in the `ControllerRegistration` controller. It's not much code, but this way the "business logic" would be captured in a single function and you only have to transform the input (project lists) to satisfy this function's signature.

pkg/operation/botanist/controlplane.go
rfranzke: Hm, okay, thanks. Let's keep it like this until the subresources are supported in the controller-runtime. Resolving this comment.

pkg/gardenlet/controller/backupentry/backup_entry_actuator.go
rfranzke: OK, agreed, thanks for the explanation!

rfranzke:
/needs rebase
/assign @ialidzhikov 

We should try to get this PR as it touches basically everything and the longer we wait the more often @tim-ebert has to rebase. :(
@timuthy can you try to review today?

timebertt:
> Could you please check `make verify`?

Done, fixed.

> We should try to get this PR as it touches basically everything and the longer we wait the more often @tim-ebert has to rebase. :(

Yep, already rebased twice because of conflicts this morning... 😞 
/remove needs/rebase

pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
timebertt: Yes, it was like this before, I just moved the code... :)

pkg/operation/common/utils.go
timebertt: Well, that makes sense. Done :)

timebertt:
All comments are resolved, PTAL :)

timebertt:
Updates:
- `GenericClientMap`: one additional change to not leak too many idling goroutines: https://github.com/tim-ebert/gardener/blob/enh/cached-client/pkg/client/kubernetes/clientmap/internal/generic_clientmap.go#L201-L207
- `kubernetes.NewClientFromSecret`: change parameter `k8sClient Interface` -> `c client.Client`

pkg/client/kubernetes/clientmap/internal/deleg_clientmap.go
```diff
@@ -0,0 +1,170 @@
...
+	return &delegatingClientMap{
+		GardenClients: gardenClientMap,
+		SeedClients:   seedClientMap,
+		ShootClients:  shootClientMap,
+		PlantClients:  plantClientMap,
+	}
+}
+
+// ErrUnknownKeyType is an error that will be returned by a delegatingClientMap if the type of the given key is unknown.
+type ErrUnknownKeyType struct {
```
timuthy:
```suggestion
type errUnknownKeyType struct {
```
Since it's in an `internal` package and not part of the API surface.

pkg/client/kubernetes/clientmap/internal/deleg_clientmap.go
```diff
@@ -0,0 +1,170 @@
...
+	key        clientmap.ClientSetKey
+}
+
+func (e *ErrUnknownKeyType) Error() string {
+	return fmt.Sprintf("call to %s with unknown ClientSetKey type: %T", e.calledFunc, e.key)
+}
+
+// ErrUnsupportedKeyType is an error that will be returned by a delegatingClientMap if it doesn't contain a ClientMap
+// that is responsible for the type of the given key.
+type ErrUnsupportedKeyType struct {
```
timuthy:
```suggestion
type errUnsupportedKeyType struct {
```

pkg/client/kubernetes/clientmap/internal/generic_clientmap.go
```diff
@@ -0,0 +1,227 @@
...
+			clientSetCancel()
+		}
+	}()
+
+	entry.cancel = clientSetCancel
+
+	entry.clientSet.Start(clientSetContext.Done())
+}
+
+func (cm *GenericClientMap) waitForClientSetCacheSync(entry *clientMapEntry, waitStopCh <-chan struct{}) bool {
```
timuthy:
```suggestion
func waitForClientSetCacheSync(entry *clientMapEntry, waitStopCh <-chan struct{}) bool {
```

pkg/client/kubernetes/clientset.go
```diff
@@ -15,73 +15,140 @@
...
+type clientSet struct {
+	config     *rest.Config
+	restMapper meta.RESTMapper
+	restClient rest.Interface
+
+	applier       Applier
+	chartApplier  ChartApplier
+	chartRenderer chartrenderer.Interface
+
+	// client is the default controller-runtime client which uses SharedIndexInformers to keeps its cache in sync
```
timuthy:
```suggestion
	// client is the default controller-runtime client which uses SharedIndexInformers to keep its cache in sync
```

timuthy (changes requested):
I don't have much to add to the previous reviews. Nice job!

pkg/client/kubernetes/clientmap/internal/generic_clientmap.go
timebertt: Nice catch!

timebertt:
Thanks for all your reviews!
I have addressed all comments, PTAL :)

rfranzke (commented):
Thanks, I don't have further comments on the code changes itself.
However, I am not able to successfully reconcile a shoot. It always times out waiting for the `BackupEntry` to be ready.

```
time="2020-06-19T07:47:56+02:00" level=info msg="Error syncing backupentry garden-dev/shoot--dev--foo-a1--bba682be-d612-4e48-93d9-1847009a4ca5: timed out waiting for the condition"
time="2020-06-19T07:47:56+02:00" level=error msg="Could not update the status after reconciliation start: timed out waiting for the condition" backupentry=shoot--dev--d065849-a2--80b3b44f-df6d-424c-807f-b1a7c61a8dcf
...
time="2020-06-19T07:48:21+02:00" level=error msg="BackupBucket /00255292-2e3b-4828-a8a3-bb2017f9213d did not get ready yet" backupentry=shoot--dev--foo-a1--bba682be-d612-4e48-93d9-1847009a4ca5 error="extension state is not succeeded but Processing"
time="2020-06-19T07:48:21+02:00" level=error msg="BackupBucket /00255292-2e3b-4828-a8a3-bb2017f9213d did not get ready yet" backupentry=shoot--dev--foo-a2--80b3b44f-df6d-424c-807f-b1a7c61a8dcf error="extension state is not succeeded but Processing"
```

Can you reproduce this?

timebertt:
Let me check

timebertt:
Yes, I can reproduce the issue. It seems like, the client's cached isn't synced fast enough for quick consecutive updates in a row (like the status updates to backupbucket and backupentry). Probably we should always go for patch updates for something like finalizers and status...
But this actually seems like a lot of effort
Other option as a quick win would be to use the direct client for the updates, we know that they will happen quickly in a row. Or at least make sure, that the final status update setting the lastOperation to succeeded is set with a direct client.

I'm wondering how they deal with this problem in k/k because I guess they mostly use Informers for the core objects. Do they only use patch requests for status updates?

timebertt:
I think, ATM this issue only happens in the BackupBucket and BackupEntry controller, as those seem to be the only ones using controller-runtime clients (and therefore now cached clients) for their status updates, all the other ones are using the "normal" direct ClientSets.

rfranzke:
> I'm wondering how they deal with this problem in k/k because I guess they mostly use Informers for the core objects. Do they only use patch requests for status updates?

No.

I'm not sure whether I can follow. We have timeouts of several seconds, so even if the cache is "slow" I would expect that it is synced/up-to-date again somewhen until the timeout is indeed reached. How does it work in the extensions library?

timuthy:
The cache delay also strikes us in every "waitForXYZ" function we have.

Let's take the following example:
A shoot maintenance is ongoing and therefore the infrastructure resource is reconciled. The shoot reconciliation flow will quickly (almost immediately) return from `"Waiting until shoot infrastructure has been reconciled"` because the infra resources taken from the cache is successful but outdated. Hence, it doesn't wait for the infra reconciliation to be completed.

```
--- cache version considered in waiting function ---
resource version: 24769770

--- actual version after annotating with `gardener.cloud/operation: reconcile` ---
resource version: 24777500
```

timebertt:
> The cache delay also strikes us in every "waitForXYZ" function we have.

Indeed, this is not good :(
What should we do wrt this? Should we explicitly check, if the `resourceVersion` of the object return from `client.Get` is at `>=` the `resourceVersion` from the passed object?

timebertt:
> I'm not sure whether I can follow. We have timeouts of several seconds, so even if the cache is "slow" I would expect that it is synced/up-to-date again somewhen until the timeout is indeed reached. How does it work in the extensions library?

Until now, we used `retry.DefaultRetry` for updates to the status in the `Backup{Entry,Bucket}` controllers, which has `Factor=1.0`. I switched to `retry.DefaultBackoff` (`Factor=5.0`) and now it works.
In the extensions library also seem to use only `retry.DeafaultBackoff`.

timuthy:
> What should we do wrt this? Should we explicitly check, if the resourceVersion of the object return from client.Get is at >= the resourceVersion from the passed object?

Hm this can work. OTOH, we don't have a "passed object" today and don't save it in the Botanist either. 
Also every function which relies on an up-to-date object is then responsible to check this. Maybe it's better for the mutating function (e.g. the one that updates the annotations) to wait for the changes to be reflected in the cache if the update is a critical one. Or we switch back to the direct client ...

timebertt:
> Hm this can work. OTOH, we don't have a "passed object" today and don't save it in the Botanist either.

I quickly tested this approach with `waitUntilBackupBucketExtensionReconciled` and `waitUntilBackupEntryExtensionReconciled` and it works fine. Luckily, those controllers store the respective object in their actuators, so we can get the current `resourceVersion` from them to pass them to a modified version of `common.WaitUntilObjectReadyWithHealthFunction`.
You're right, this probably does work so easily for the shoot flow as we don't return or store the latest version of the extension objects...

timebertt:
> Maybe it's better for the mutating function (e.g. the one that updates the annotations) to wait for the changes to be reflected in the cache if the update is a critical one.

Yes, that would also be possible.

timebertt:
Alright, as discussed on the phone, let's continue with this PR in the following way to address the discovered issues:
- use a direct client in the few cases where a controller-runtime client is used to make updates to the status (e.g. backup controllers)
- use a direct client in all `WaitFor*` funcs that are critical (e.g. to the Shoot reconcile flow)
- introduce a FeatureGate for the cached clients as a kill-switch, to disable the cache of all clients in case any catastrophic issues with the implementations are discovered while running on a real gardener system (so g/g behaves exactly like before)

This way we can make sure, we don't break too much and can slowly improve while observing the behaviour in our systems.

/needs changes
/status author-action

timebertt:
Update:

> * use a direct client in the few cases where a controller-runtime client is used to make updates to the status (e.g. backup controllers)

This one is done. I will tackle the other issues next week.

ialidzhikov:
Ref https://github.com/gardener/gardener-extension-provider-gcp/pull/112#issuecomment-647530802 - it might be interesting also for this.

timebertt:
> Ref [gardener/gardener-extension-provider-gcp#112 (comment)](https://github.com/gardener/gardener-extension-provider-gcp/pull/112#issuecomment-647530802) - it might be interesting also for this.

Thanks for finding out and mentioning here.
We have only one usage in g/g of `client.MatchingFields` and for that I already switched to the direct client because of [lack of support for multiple field selectors](https://github.com/kubernetes-sigs/controller-runtime/blob/ca25c1f1014d6db6eba745e04f180d272a854e9a/pkg/cache/internal/cache_reader.go#L98-L103).
So I assume, there is nothing left to adapt here as of now.

timebertt:
Update:
> * use a direct client in all `WaitFor*` funcs that are critical (e.g. to the Shoot reconcile flow)
> * introduce a FeatureGate for the cached clients as a kill-switch, to disable the cache of all clients in case any catastrophic issues with the implementations are discovered while running on a real gardener system (so g/g behaves exactly like before)

These are also done. What is quite nice about the FeatureGate is, I implemented it this way, so this won't result in a breaking change for extensions using our clientset, as by default the runtime client is now uncached as before.
Please take another look, while I am conducting some final tests.

/remove needs/changes
/remove status/author-action
/needs review
/unassign

timebertt:
Sorry guys. Rebasing takes longer than expected, because the PRs merged today didn't incorporate the latest changes on master.
/status author-action 
😞

timebertt:
Blocked by https://github.com/gardener/gardener/issues/2492#issuecomment-648328778
I rebased and resolved the conflicts, but now I am not able to check if Shoot reconciliation still works because of the mentioned issues.
/hold

timuthy (commented):
I cross checked the changes we discussed in our last meeting and looks good to me so far. Let's wait for your final test.

timebertt:
Rebased to solve conflicts, also triggered CI run

timebertt:
Rebased onto master to get the latest fixes, should be good now :)

timebertt:
I conducted some final tests and it looks good.
/remove status/author-action
/unassign

@timuthy @ialidzhikov @rfranzke Please take a final look, so we can get it in :)

timuthy (approved):

rfranzke (approved):
/lgtm

works perfectly with creation/deletion/hibernation/wake-up

timebertt:
🥳 🎉

PR #2450, opened 2020-06-12
[GEP-7] Don`t use tryUpdate when restoring extensions

swilen-iwanow:

/area control-plane-migration
/kind bug
/priority normal

Found a bug that was preventing `status.state` from update if a conflict occurs, when using `tryUpdate()` . For example:
For the first call of tryUpdateStatus we have the obj `worker.status.state=nil`.
And in tryUpdate we get the object from the server, transform it so `worker.status.state=something`, we check the object for changes before and after the transform and if it has changed we try to update. If a conflict occurs we fail, but our in memory object has `worker.status.state=something`.
So on the next iteration we get the object from API Server, which still has `worker.status.state=nil` in the server, but the object in memory is with `worker.status.state=something` (for some reason client.Get() merges it). On the next step we transform it which doesn't change anything and test if the transformed object is different from the one received from the server. They are now equal so we don't update and simply return.
 
Fixes #

I am not sure if a release note should be set for fixing bugs of feature that is not available yet.
**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/operation/botanist/migration.go
+++ b/pkg/operation/botanist/migration.go
@@ -102,43 +102,40 @@ func (b *Botanist) restoreExtensionObject(ctx context.Context, client client.Cli
 }
 
 func (b *Botanist) restoreExtensionObjectState(ctx context.Context, client client.Client, extensionObj extensionsv1alpha1.Object, resourceKind, resourceName string, purpose *string) error {
-	return kutil.TryUpdateStatus(ctx, k8sretry.DefaultBackoff, client, extensionObj, func() error {
-		var resourceRefs []autoscalingv1.CrossVersionObjectReference
-		if b.ShootState.Spec.Extensions != nil {
-			list := gardencorev1alpha1helper.ExtensionResourceStateList(b.ShootState.Spec.Extensions)
-			if extensionResourceState := list.Get(resourceKind, &resourceName, purpose); extensionResourceState != nil {
-				extensionStatus := extensionObj.GetExtensionStatus()
-				extensionStatus.SetState(*extensionResourceState.State)
-				extensionStatus.SetResources(extensionResourceState.Resources)
-				for _, r := range extensionResourceState.Resources {
-					resourceRefs = append(resourceRefs, r.ResourceRef)
+	var resourceRefs []autoscalingv1.CrossVersionObjectReference
+	if b.ShootState.Spec.Extensions != nil {
+		list := gardencorev1alpha1helper.ExtensionResourceStateList(b.ShootState.Spec.Extensions)
+		if extensionResourceState := list.Get(resourceKind, &resourceName, purpose); extensionResourceState != nil {
+			extensionStatus := extensionObj.GetExtensionStatus()
+			extensionStatus.SetState(extensionResourceState.State)
+			extensionStatus.SetResources(extensionResourceState.Resources)
+			for _, r := range extensionResourceState.Resources {
+				resourceRefs = append(resourceRefs, r.ResourceRef)
+			}
+		}
+	}
+	if b.ShootState.Spec.Resources != nil {
+		list := gardencorev1alpha1helper.ResourceDataList(b.ShootState.Spec.Resources)
+		for _, resourceRef := range resourceRefs {
+			resourceData := list.Get(&resourceRef)
+			if resourceData != nil {
+				obj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&resourceData.Data)
+				if err != nil {
+					return err
+				}
+				if err := utils.CreateOrUpdateObjectByRef(ctx, b.K8sSeedClient.Client(), &resourceRef, b.Shoot.SeedNamespace, obj); err != nil {
+					return err
 				}
 			}
 		}
-		if b.ShootState.Spec.Resources != nil {
-			list := gardencorev1alpha1helper.ResourceDataList(b.ShootState.Spec.Resources)
-			for _, resourceRef := range resourceRefs {
-				resourceData := list.Get(&resourceRef)
-				if resourceData != nil {
-					obj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&resourceData.Data)
-					if err != nil {
-						return err
-					}
-					if err := utils.CreateOrUpdateObjectByRef(ctx, b.K8sSeedClient.Client(), &resourceRef, b.Shoot.SeedNamespace, obj); err != nil {
-						return err
-					}
-				}
-			}
-		}
-		return nil
-	})
+	}
+	return client.Status().Update(ctx, extensionObj)
 }
 
-func annotateExtensionObjectWithOperationRestore(ctx context.Context, client client.Client, extensionObj extensionsv1alpha1.Object) error {
-	return kutil.TryUpdate(ctx, k8sretry.DefaultBackoff, client, extensionObj, func() error {
-		kutil.SetMetaDataAnnotation(extensionObj, v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationRestore)
-		return nil
-	})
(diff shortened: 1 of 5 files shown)


plkokanov (changes requested):
Thanks for taking care of this!
Can you also add the deprecation comment to tryUpdate (not sure if @mvladev won't do that), and shouldn't we wrap `annotateExtensionObjectWithOperationRestore` and `restoreExtensionObjectState` in the calling code with `RetryOnConflict` making sure to copy the object where necessary so that we don't run into the same issues as `TryUpdate`.
Or you could wrap the `Update` calls inside `annotateExtensionObjectWithOperationRestore` and `restoreExtensionObjectState` with `RetryOnConflict`.

swilen-iwanow:
> Thanks for taking care of this!
> Can you also add the deprecation comment to tryUpdate (not sure if @mvladev won't do that), and shouldn't we wrap `annotateExtensionObjectWithOperationRestore` and `restoreExtensionObjectState` in the calling code with `RetryOnConflict` making sure to copy the object where necessary so that we don't run into the same issues as `TryUpdate`.
> Or you could wrap the code inside `annotateExtensionObjectWithOperationRestore` and `restoreExtensionObjectState` with `RetryOnConflict`.

Spoking with @mvladev on Friday, and if I understood what he meant correctly is that the task itself should be retried and we should leave the retry logic to the flow framework instead of adding additional. I can add the comments for the deprecation, yes.

plkokanov:
> Spoking with @mvladev on Friday, and if I understood what he meant correctly is that the task itself should be retried and we should leave the retry logic to the flow framework instead of adding additional. I can add the comments for the deprecation, yes.

Hmm, actually makes sense since this is how we use CreateOrUpdate method.

rfranzke:
Hm, to be honest, I don't really understand the rationale behind this PR. Generally, @mvladev is right that the surrounding flow/reconciler can take over the retry, but IMO this only applies to reconciliation loops that are short/small/fast. Our shoot reconciliation is quite the opposite, so having a time-boxed retry makes sense to me.

Doesn't it make more sense to fix this `TryUpdateStatus` method instead of simply removing it? It's used in several other places and such those would also benefit from a fix.

swilen-iwanow:
> Hm, to be honest, I don't really understand the rationale behind this PR. Generally, @mvladev is right that the surrounding flow/reconciler can take over the retry, but IMO this only applies to reconciliation loops that are short/small/fast. Our shoot reconciliation is quite the opposite, so having a time-boxed retry makes sense to me.
> 
> Doesn't it make more sense to fix this `TryUpdateStatus` method instead of simply removing it? It's used in several other places and such those would also benefit from a fix.

What I meant was, not to retry the whole reconciliation loop but specific step/task of the flow with `TaskFn`. About the fix of the TryUpdateStatus it could be done as far as we agree that we want to fix it instead of deprecating it, @mvladev @rfranzke .

rfranzke:
> What I meant was, not to retry the whole reconciliation loop but specific step/task of the flow with TaskFn.

OK, that's probably fine for this special case, sure.

> About the fix of the TryUpdateStatus it could be done as far as we agree that we want to fix it instead of deprecating it

Why should we deprecate it?

swilen-iwanow:
> Why should we deprecate it? 

If there is a conflict we don't recalculate if the update should happen or not. 
1. Reconciliation of an object starts - the object status is X and field `foo=bar` 
2. Based on field `foo=bar`, we calculate the desired status to be `X+1`
3. In the mean time field foo is set to `abc` and status is updated by another controller to be Y
4. Because of the retry we'll eventually set status to X+1, while foo is abc (which is not valid)

As this is not part of the PR, we can open an issue to enhance the method or deprecate it.

rfranzke:
I don't see any reason to deprecate it just because it has a bug. It's a helpful function that is used in various places. Yes, it's out of scope of this PR, but we should fix it as soon as possible. It definitely makes sense to open an issue, thanks. As you have already spent some time on analyzing and understanding the problem, does it make sense that you also pick up the task of fixing it?

rfranzke (dismissed):
/lgtm
/needs second-opinion

swilen-iwanow:
Okay then, I will open an Issue and you can assign me.

pkg/operation/botanist/migration.go
```diff
@@ -102,43 +102,40 @@ func (b *Botanist) restoreExtensionObject(ctx context.Context, client client.Cli
...
-				extensionStatus.SetResources(extensionResourceState.Resources)
-				for _, r := range extensionResourceState.Resources {
-					resourceRefs = append(resourceRefs, r.ResourceRef)
-				}
+	var resourceRefs []autoscalingv1.CrossVersionObjectReference
+	if b.ShootState.Spec.Extensions != nil {
+		list := gardencorev1alpha1helper.ExtensionResourceStateList(b.ShootState.Spec.Extensions)
+		if extensionResourceState := list.Get(resourceKind, &resourceName, purpose); extensionResourceState != nil {
+			extensionStatus := extensionObj.GetExtensionStatus()
+			extensionStatus.SetState(*extensionResourceState.State)
```
plkokanov: @swilen-iwanow we could probably avoid dereferencing the pointer to the `extensionResourceState` and modify `SetState` to take a pointer now that we know the true cause of the error. Wdyt?

timebertt:
Hmm, I also don't quite get the purpose of this PR and I also don't get, what's the bug in the `tryUpdate` func.

> So on the next iteration we get the object from API Server, which still has worker.status.state=nil in the server, but the object in memory is with worker.status.state=something (for some reason client.Get() merges it)

I don't think, that `client.Get()` merges anything, or am I missing something here?
@vlvasilev Can you point me to the line of code, where it merges objects? Can you also write down some concrete steps to reproduce this behaviour?

Apart from that, the PR on its own is /lgtm as it only removes the retry on conflict.

timuthy:
FYI: That's the exact same reason @rfranzke added the `newObjFunc` to the WaitUntilExtensionCRReady function. https://github.com/gardener/gardener/blob/42700aecc47572ba47a38324fa973f4249eac387/pkg/operation/common/extensions.go#L42

See this discussion: https://github.com/gardener/gardener/pull/2222#discussion_r414449331

timebertt:
@timuthy, thanks for the reference.
Now, I finally got the problem. I digged into it a bit, and it seems like the underlying json decoder just decodes the fields that are present in the response sent from the API server and doesn't reset other fields (kind of expected, as this is the straight-forward approach to implement `Decode()`)... ([ref](https://github.com/gardener/gardener/blob/35113df7e8482eb72d634f7fa5f4bf18ee85fc0e/vendor/github.com/json-iterator/go/reflect_struct_decoder.go#L508)).

swilen-iwanow:
> @timuthy, thanks for the reference.
> Now, I finally got the problem. I digged into it a bit, and it seems like the underlying json decoder just decodes the fields that are present in the response sent from the API server and doesn't reset other fields (kind of expected, as this is the straight-forward approach to implement `Decode()`)... ([ref](https://github.com/gardener/gardener/blob/35113df7e8482eb72d634f7fa5f4bf18ee85fc0e/vendor/github.com/json-iterator/go/reflect_struct_decoder.go#L508)).

Yes exactly what I meant by merge. Sorry for the late response.

rfranzke:
So @tim-ebert @timuthy are you okay with proceeding with this PR?

timuthy:
I talked to @swilen-iwanow and we think it'd be a better approach to `patch` the restore annotation instead of using `client.Update`. This will eliminate a potential resource version conflict.

rfranzke (approved):
/lgtm
/needs second-opinion
/assign @timuthy @tim-ebert

plkokanov (approved):
/lgtm

timuthy (approved):
/lgtm
Thanks for addressing the comments

timebertt (approved):
/lgtm

PR #2456, opened 2020-06-15
[GEP-7] Remove DNSProviders in migration flow

swilen-iwanow:

/area control-plane-migration 
/kind enhancement 
/priority normal

Adds one more step to delete the DNSProviders explicitly during the migration process and
keeps the `ControllerInstallations` on the Seed if there is atleast one `Shoot` referring it in the shoot.spec.seedName or shoot.status.seedName. This in cases where you are migrating the last `Shoot` on the `Seed` to new `Seed`.

**Release note**:

```improvement operator
ControllerInstallations are not removed from the Seed if there is at least one Shoot referring it in the `spec.seedName` or `status.seedName`.
```

### DIFF
--- a/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
+++ b/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/gardener/gardener/pkg/apis/core"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	v1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	gardencorelisters "github.com/gardener/gardener/pkg/client/core/listers/core/v1beta1"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
@@ -123,9 +124,8 @@ func (c *defaultControllerRegistrationSeedControl) Reconcile(obj *gardencorev1be
 	if err != nil {
 		return err
 	}
-	shootList, err := c.k8sGardenClient.GardenCore().CoreV1beta1().Shoots(metav1.NamespaceAll).List(metav1.ListOptions{
-		FieldSelector: fields.SelectorFromSet(fields.Set{core.ShootSeedName: seed.Name}).String(),
-	})
+
+	shootList, err := getShoots(ctx, c.k8sGardenClient, seed)
 	if err != nil {
 		return err
 	}
@@ -226,7 +226,7 @@ func computeKindTypesForShoots(
 	ctx context.Context,
 	logger *logrus.Entry,
 	client client.Client,
-	shootList *gardencorev1beta1.ShootList,
+	shootList []gardencorev1beta1.Shoot,
 	seed *gardencorev1beta1.Seed,
 	controllerRegistrationList []*gardencorev1beta1.ControllerRegistration,
 	internalDomain *gardenpkg.Domain,
@@ -239,8 +239,8 @@ func computeKindTypesForShoots(
 		out = make(chan sets.String)
 	)
 
-	for _, shoot := range shootList.Items {
-		if shoot.Spec.SeedName == nil || *shoot.Spec.SeedName != seed.Name {
+	for _, shoot := range shootList {
+		if (shoot.Spec.SeedName == nil || *shoot.Spec.SeedName != seed.Name) && (shoot.Status.SeedName == nil || *shoot.Status.SeedName != seed.Name) {
 			continue
 		}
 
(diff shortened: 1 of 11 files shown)


pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -123,9 +123,7 @@ func (c *defaultControllerRegistrationSeedControl) Reconcile(obj *gardencorev1be
 	if err != nil {
 		return err
 	}
-	shootList, err := c.k8sGardenClient.GardenCore().CoreV1beta1().Shoots(metav1.NamespaceAll).List(metav1.ListOptions{
-		FieldSelector: fields.SelectorFromSet(fields.Set{core.ShootSeedName: seed.Name}).String(),
```
rfranzke, line 127: To optimize the list call, does it make sense to have a field selector for `.status.seedName` as well, and use it in conjunction with this one? Is this possible in the first place?

pkg/gardenlet/controller/shoot/shoot_control_prepare_migration.go
```diff
@@ -170,22 +170,22 @@ func (c *Controller) runPrepareShootControlPlaneMigration(o *operation.Operation
...
 		})
 		keepManagedResourcesObjectsInShoot = g.Add(flow.Task{
-			Name:         "Keep Managed Resources objects in the Shoot",
+			Name:         "Configuring Managed Resources objects to be kept in the Shoot",
 			Fn:           flow.TaskFn(botanist.KeepManagedResourcesObjects).DoIf(cleanupShootResources),
 			Dependencies: flow.NewTaskIDs(ensureResourceManagerScaledUp),
 		})
 		deleteAllManagedResourcesFromShootNamespace = g.Add(flow.Task{
-			Name:         "Deletes all Managed Resources  from the Shoot's namespace",
+			Name:         "Deleting all Managed Resources  from the Shoot's namespace",
```
rfranzke:
```suggestion
			Name:         "Deleting all Managed Resources from the Shoot's namespace",
```

pkg/operation/botanist/migration.go
```diff
@@ -155,7 +156,11 @@ func (b *Botanist) AnnotateBackupEntryInSeedForMigration(ctx context.Context) er
 		name        = common.GenerateBackupEntryName(b.Shoot.Info.Status.TechnicalID, b.Shoot.Info.Status.UID)
 		backupEntry = &extensionsv1alpha1.BackupEntry{}
 	)
+
 	if err := b.K8sSeedClient.Client().Get(ctx, kutil.Key(name), backupEntry); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
 		return err
```
rfranzke:
```suggestion
	if err := b.K8sSeedClient.Client().Get(ctx, kutil.Key(name), backupEntry); client.IgnoreNotFound(err) != nil {
		return err
```

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
swilen-iwanow: Yes I was thinking the same, but couldn't find a way to use `OR` for the field selectors in case the `shoot.spec.seedName` is set to `Target Seed`, but still have a `status.seedName` equal to `Source Seed`. I had another thought - to use the `status.seedName` only, now when we set the status.seedName in the beginning of the creation/reconciliation.

rfranzke: Hm sure, could work. Generally, filtering here would help a lot given the large number of shoots in a system.

swilen-iwanow: I tried with `FilterFunc` of type `status.seedName=gcp` and this will block the first shoot creation on a new seed because there is a check if extensions are ready :/ before starting the reconciliation. Maybe we can list twice and merge the lists?

rfranzke: Hm yeah, sounds reasonable. Thanks for your efforts.

pkg/apis/core/v1beta1/types_utils.go
```diff
@@ -70,3 +72,29 @@ const (
...
+	for _, v := range unionedShoots {
+		shoots = append(shoots, v)
+	}
+
+	return shoots
+}
+
+func objectKey(namesapce, name string) string {
+	return fmt.Sprintf("%s%c%s", namesapce, '/', name)
+}
```
rfranzke: Isn't the `helper` package a better place for such things?
Also, API-wise I feel it's a little inconsistent that a `ShootItems` structure has a `Union` function that doesn't take another `ShootItems` but a `ShootList`.

pkg/apis/core/v1beta1/types_utils.go
```diff
@@ -70,3 +72,29 @@ const (
...
+	shoots := make([]Shoot, 0, len(unionedShoots))
+	for _, v := range unionedShoots {
+		shoots = append(shoots, v)
+	}
+
+	return shoots
+}
+
+func objectKey(namesapce, name string) string {
+	return fmt.Sprintf("%s%c%s", namesapce, '/', name)
```
rfranzke:
```suggestion
	return fmt.Sprintf("%s/%s", namespace, name)
```

rfranzke (commented):
Thank you, generally PR looks good now, just a minor structuring comment.

pkg/apis/core/v1beta1/types_utils.go
swilen-iwanow: ShootItems could be type alias of []Shoot instead of new type, wdyt? That way we can 
`var shootItems ShootItems`
`shootItems = shootList.Items`

swilen-iwanow:
Oops I am sorry maybe I shouldn't have rebased.

rfranzke (approved):
/lgtm

PR #2459, opened 2020-06-16
Increases generic worker actuator reliability

danielfoehrKn:

/area usability
/kind enhancement
/area quality
/area robustness
/kind bug
/priority normal

Increases the reliability of the generic worker actuator to detect rolling updates and populates rolling updates  in a condition in the status of the Worker CRD.

This condition is set to `True` as long as there is a rolling update in progress
 - until all updated machines joined the cluster (`numUnavailable == 0`) 
 - until the old machine is deleted in the case of a rolling update with maxUnavailability = 0 (`numUpdated == numberOfAwakeMachines`) 

Adds checks to verify that the required machine sets have been created by the MCM. This should prevent stale cache issues.

**Restart stuck Machine controller manager**

Restarts the machine controller manager if it failed to create the proper machine sets after unsuccessfully waiting for the machine deployments to become ready.

**Bugfix**

Also fixes a stale cache bug that leads to not waiting for the rolling update to complete. The check if the rolling update is complete is sometimes done on an outdated machine deployment. 

Fixes #

The topic of this PR is increasing the reliability of the generic worker actuator. 
Because the introduced features use shared coding and are thematically coherent, I propose having them in the same PR. If that is unfeasible, let me know.

Please note that the cluster autoscaler scale down behaviour has not changed (scaled down during creation and rolling updates).
In the future, the cluster autoscaler might not be required to be scaled down any more.
See [this issue](https://github.com/gardener/autoscaler/issues/48) and the corresponding issue on the [MCM](https://github.com/gardener/machine-controller-manager/issues/472)

**Release note**:

```improvement operator
 Introduces a `RollingUpdate` condition in the generic worker actuator (condition.Type `RollingUpdate`) . Gardener provider extensions write this condition to the Worker CRD.
```
```improvement operator
The generic worker actuator more reliably waits for rolling updates to finish.  Waits until all updated machines joined the cluster and until old machines are deleted. Also  fixes a stale cache bug that leads to not waiting for the rolling update to complete.
```

```improvement operator
 The generic worker actuator detects and restarts 'stuck' machine controller manager pods.
```

### DIFF
--- a/extensions/pkg/controller/worker/genericactuator/actuator.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator.go
@@ -16,22 +16,27 @@ package genericactuator
 
 import (
 	"context"
+	"fmt"
 	"strings"
+	"time"
 
 	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
 	"github.com/gardener/gardener/extensions/pkg/controller/worker"
+	workerhelper "github.com/gardener/gardener/extensions/pkg/controller/worker/helper"
 	"github.com/gardener/gardener/extensions/pkg/util"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	gardenerkubernetes "github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/controllerutils"
 	"github.com/gardener/gardener/pkg/utils/flow"
 	"github.com/gardener/gardener/pkg/utils/imagevector"
+
 	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -156,6 +161,7 @@ func (a *genericActuator) cleanupMachineClasses(ctx context.Context, namespace s
 		}
 
 		if !wantedMachineDeployments.HasClass(accessor.GetName()) {
+			a.logger.V(6).Info("Deleting machine class", "name", accessor.GetName(), "namespace", accessor.GetNamespace())
 			if err := a.client.Delete(ctx, machineClass); err != nil {
 				return err
 			}
@@ -279,6 +285,58 @@ func (a *genericActuator) shallowDeleteAllObjects(ctx context.Context, namespace
 		}
 		return nil
 	})
+}
+
+// IsMachineControllerStuck determines if the machine controller pod is stuck.
+func (a *genericActuator) IsMachineControllerStuck(ctx context.Context, worker *extensionsv1alpha1.Worker) (bool, *string, error) {
+	machineDeployments := &machinev1alpha1.MachineDeploymentList{}
+	if err := a.client.List(ctx, machineDeployments, client.InNamespace(worker.Namespace)); err != nil {
+		return false, nil, err
+	}
+
+	machineSets := &machinev1alpha1.MachineSetList{}
+	if err := a.client.List(ctx, machineSets, client.InNamespace(worker.Namespace)); err != nil {
+		return false, nil, err
+	}
+
+	isStuck, msg := isMachineControllerStuck(machineSets.Items, machineDeployments.Items)
+	return isStuck, msg, nil
+}
+
+const (
+	// stuckMCMThreshold defines that the machine deployment set has to be created more than
+	// this duration ago to be considered for the check whether a machine controller manager pod is stuck
+	stuckMCMThreshold = 2 * time.Minute
+	// mcmFinalizer is the finalizer used by the machine controller manager
+	// not imported from the MCM to reduce dependencies
+	mcmFinalizer = "machine.sapcloud.io/machine-controller-manager"
+)
+
+// isMachineControllerStuck determines if the machine controller pod is stuck.
+// A pod is assumed to be stuck if
+//  - a machine deployment exists that does not have a machine set with the correct machine class
+//  - the machine set does not have a status that indicates (attempted) machine creation
+func isMachineControllerStuck(machineSets []machinev1alpha1.MachineSet, machineDeployments []machinev1alpha1.MachineDeployment) (bool, *string) {
+	// map the owner reference to the existing machine sets
+	ownerReferenceToMachineSet := workerhelper.BuildOwnerToMachineSetsMap(machineSets)
+
+	for _, machineDeployment := range machineDeployments {
+		if !controllerutils.HasFinalizer(&machineDeployment, mcmFinalizer) {
+			continue
+		}
+
+		// do not consider machine deployments that have just recently been created
+		if time.Now().UTC().Sub(machineDeployment.ObjectMeta.CreationTimestamp.UTC()) < stuckMCMThreshold {
+			continue
+		}
+
+		machineSet := workerhelper.GetMachineSetWithMachineClass(machineDeployment.Name, machineDeployment.Spec.Template.Spec.Class.Name, ownerReferenceToMachineSet)
+		if machineSet == nil {
+			msg := fmt.Sprintf("missing machine set for machine deployment (%s/%s)", machineDeployment.Namespace, machineDeployment.Name)
+			return true, &msg
+		}
+	}
+	return false, nil
 }
 
 // CleanupLeakedClusterRoles cleans up leaked ClusterRoles from the system that were created earlier without
--- a/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
@@ -22,8 +22,10 @@ import (
 	"github.com/gardener/gardener/extensions/pkg/controller"
 	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
 	workerhealthcheck "github.com/gardener/gardener/extensions/pkg/controller/healthcheck/worker"
-	"github.com/gardener/gardener/extensions/pkg/controller/worker"
+	extensionsworker "github.com/gardener/gardener/extensions/pkg/controller/worker"
+	workerhelper "github.com/gardener/gardener/extensions/pkg/controller/worker/helper"
 	"github.com/gardener/gardener/extensions/pkg/util"
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
@@ -38,6 +40,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -76,42 +79,21 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 		return errors.Wrapf(err, "failed to generate the machine deployments")
 	}
 
-	// Get list of existing machine class names and list of used machine class secrets.
+	var clusterAutoscalerUsed = extensionsv1alpha1helper.ClusterAutoscalerRequired(worker.Spec.Pools)
+
+	// When the Shoot is hibernated we want to remove the cluster autoscaler so that it does not interfer
+	// with Gardeners modifications on the machine deployment's replicas fields.
+	isHibernated := controller.IsHibernated(cluster)
+	if clusterAutoscalerUsed && isHibernated {
+		if err = a.scaleClusterAutoscaler(ctx, worker, 0); err != nil {
+			return err
+		}
+	}
+
+	// Get list of existing machine class names
 	existingMachineClassNames, err := a.listMachineClassNames(ctx, worker.Namespace, workerDelegate.MachineClassList())
 	if err != nil {
 		return err
-	}
-
-	// During the time a rolling update happens we do not want the cluster autoscaler to interfer, hence it
-	// is removed for now.
-	var (
-		clusterAutoscalerUsed = extensionsv1alpha1helper.ClusterAutoscalerRequired(worker.Spec.Pools)
-		rollingUpdate         = false
-	)
-
-	if clusterAutoscalerUsed {
-		// Check whether new machine classes have been computed (resulting in a rolling update of the nodes).
-		for _, machineDeployment := range wantedMachineDeployments {
-			if !existingMachineClassNames.Has(machineDeployment.ClassName) {
-				rollingUpdate = true
-				break
-			}
-		}
-
-		// When the Shoot gets hibernated we want to remove the cluster auto scaler so that it does not interfer
-		// with Gardeners modifications on the machine deployment's replicas fields.
-		if controller.IsHibernated(cluster) || rollingUpdate {
-			deployment := &appsv1.Deployment{}
-			if err := a.client.Get(ctx, kutil.Key(worker.Namespace, v1beta1constants.DeploymentNameClusterAutoscaler), deployment); err != nil {
-				if !apierrors.IsNotFound(err) {
-					return err
-				}
-			} else {
-				if err := util.ScaleDeployment(ctx, a.client, deployment, 0); err != nil {
-					return err
-				}
-			}
-		}
 	}
 
 	// Deploy generated machine classes.
@@ -135,6 +117,11 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 		return err
 	}
 
+	existingMachineDeploymentNames := sets.String{}
+	for _, deployment := range existingMachineDeployments.Items {
+		existingMachineDeploymentNames.Insert(deployment.Name)
+	}
+
 	// Generate machine deployment configuration based on previously computed list of deployments and deploy them.
 	a.logger.Info("Deploying the machine deployments", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
 	if err := a.deployMachineDeployments(ctx, cluster, worker, existingMachineDeployments, wantedMachineDeployments, workerDelegate.MachineClassKind(), clusterAutoscalerUsed); err != nil {
@@ -142,7 +129,28 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 	}
 
 	// Wait until all generated machine deployments are healthy/available.
-	if err := a.waitUntilWantedMachineDeploymentsAvailable(ctx, cluster, worker, wantedMachineDeployments); err != nil {
+	if err := a.waitUntilWantedMachineDeploymentsAvailable(ctx, cluster, worker, existingMachineDeploymentNames, existingMachineClassNames, wantedMachineDeployments, clusterAutoscalerUsed); err != nil {
+		// check if the machine controller manager is stuck
+		isStuck, msg, err2 := a.IsMachineControllerStuck(ctx, worker)
+		if err2 != nil {
+			a.logger.Error(err2, "failed to check if the machine controller manager pod is stuck after unsuccessfully waiting for all machine deployments to be ready.", "namespace", worker.Namespace)
+			// continue in order to return `err` and determine error codes
+		}
+
+		if isStuck {
+			podList := corev1.PodList{}
+			if err2 := a.client.List(ctx, &podList, client.InNamespace(worker.Namespace), client.MatchingLabels{"role": "machine-controller-manager"}); err2 != nil {
+				return errors.Wrapf(err2, "failed to list machine controller manager pods for worker (%s/%s)", worker.Namespace, worker.Name)
+			}
+
+			for _, pod := range podList.Items {
+				if err2 := a.client.Delete(ctx, &pod); err2 != nil {
+					return errors.Wrapf(err2, "failed to delete stuck machine controller manager pod for worker (%s/%s)", worker.Namespace, worker.Name)
+				}
+			}
+			a.logger.Info("Successfully deleted stuck machine controller manager pod", "reason", msg, "worker namespace", worker.Namespace, "worker name", worker.Name)
+		}
+
 		return gardencorev1beta1helper.DetermineError(err, fmt.Sprintf("Failed while waiting for all machine deployments to be ready: '%s'", err.Error()))
 	}
 
@@ -172,7 +180,7 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 	}
 
 	// Scale down machine-controller-manager if shoot is hibernated.
-	if controller.IsHibernated(cluster) {
+	if isHibernated {
 		deployment := &appsv1.Deployment{}
 		if err := a.client.Get(ctx, kutil.Key(worker.Namespace, a.mcmName), deployment); err != nil {
 			return err
@@ -182,27 +190,34 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 		}
 	}
 
-	if rollingUpdate {
-		deployment := &appsv1.Deployment{}
-		if err := a.client.Get(ctx, kutil.Key(worker.Namespace, v1beta1constants.DeploymentNameClusterAutoscaler), deployment); err != nil {
-			if !apierrors.IsNotFound(err) {
-				return err
-			}
-		} else {
-			if err := util.ScaleDeployment(ctx, a.client, deployment, 1); err != nil {
-				return err
-			}
+	if clusterAutoscalerUsed && !isHibernated {
+		if err = a.scaleClusterAutoscaler(ctx, worker, 1); err != nil {
+			return err
 		}
 	}
 
-	if err := a.updateWorkerStatusMachineDeployments(ctx, worker, wantedMachineDeployments); err != nil {
+	if err := a.updateWorkerStatusMachineDeployments(ctx, worker, wantedMachineDeployments, false); err != nil {
 		return errors.Wrapf(err, "failed to update the machine deployments in worker status")
 	}
 
 	return nil
 }
 
-func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, existingMachineDeployments *machinev1alpha1.MachineDeploymentList, wantedMachineDeployments worker.MachineDeployments, classKind string, clusterAutoscalerUsed bool) error {
+func (a *genericActuator) scaleClusterAutoscaler(ctx context.Context, worker *extensionsv1alpha1.Worker, replicas int32) error {
+	deployment := &appsv1.Deployment{}
+	if err := a.client.Get(ctx, kutil.Key(worker.Namespace, v1beta1constants.DeploymentNameClusterAutoscaler), deployment); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+	} else {
+		if err := util.ScaleDeployment(ctx, a.client, deployment, replicas); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, existingMachineDeployments *machinev1alpha1.MachineDeploymentList, wantedMachineDeployments extensionsworker.MachineDeployments, classKind string, clusterAutoscalerUsed bool) error {
 	for _, deployment := range wantedMachineDeployments {
 		var (
 			labels                    = map[string]string{"name": deployment.Name}
@@ -309,55 +324,113 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster 
 
 // waitUntilWantedMachineDeploymentsAvailable waits until all the desired <machineDeployments> were marked as healthy /
 // available by the machine-controller-manager. It polls the status every 5 seconds.
-func (a *genericActuator) waitUntilWantedMachineDeploymentsAvailable(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, wantedMachineDeployments worker.MachineDeployments) error {
-	return retryutils.UntilTimeout(ctx, 5*time.Second, 5*time.Minute, func(ctx context.Context) (bool, error) {
-		var numHealthyDeployments, numUpdated, numAvailable, numDesired, numberOfAwakeMachines int32
+func (a *genericActuator) waitUntilWantedMachineDeploymentsAvailable(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, alreadyExistingMachineDeploymentNames sets.String, alreadyExistingMachineClassNames sets.String, wantedMachineDeployments extensionsworker.MachineDeployments, clusterAutoscalerUsed bool) error {
+	autoscalerIsScaledDown := false
+	workerStatusUpdatedForRollingUpdate := false
 
-		// Get the list of all existing machine deployments
-		existingMachineDeployments := &machinev1alpha1.MachineDeploymentList{}
-		if err := a.client.List(ctx, existingMachineDeployments, client.InNamespace(worker.Namespace)); err != nil {
+	return retryutils.UntilTimeout(ctx, 5*time.Second, 5*time.Minute, func(ctx context.Context) (bool, error) {
+		var numHealthyDeployments, numUpdated, numAvailable, numUnavailable, numDesired, numberOfAwakeMachines int32
+
+		// Get the list of all machine deployments
+		machineDeployments := &machinev1alpha1.MachineDeploymentList{}
+		if err := a.client.List(ctx, machineDeployments, client.InNamespace(worker.Namespace)); err != nil {
 			return retryutils.SevereError(err)
 		}
 
+		// Get the list of all machine sets
+		machineSets := &machinev1alpha1.MachineSetList{}
+		if err := a.client.List(ctx, machineSets, client.InNamespace(worker.Namespace)); err != nil {
+			return retryutils.SevereError(err)
+		}
+
+		// map the owner reference to the machine sets
+		ownerReferenceToMachineSet := workerhelper.BuildOwnerToMachineSetsMap(machineSets.Items)
+
 		// Collect the numbers of available and desired replicas.
-		for _, existingMachineDeployment := range existingMachineDeployments.Items {
+		for _, deployment := range machineDeployments.Items {
+			wantedDeployment := wantedMachineDeployments.FindByName(deployment.Name)
+
 			// Filter out all machine deployments that are not desired (any more).
-			if !wantedMachineDeployments.HasDeployment(existingMachineDeployment.Name) {
+			if wantedDeployment == nil {
 				continue
+			}
+
+			machineSets := ownerReferenceToMachineSet[deployment.Name]
+
+			// use `wanted deployment` for these checks, as the existing deployments can be based on an outdated cache
+			alreadyExistingMachineDeployment := alreadyExistingMachineDeploymentNames.Has(wantedDeployment.Name)
+			newMachineClass := !alreadyExistingMachineClassNames.Has(wantedDeployment.ClassName)
+
+			if alreadyExistingMachineDeployment && newMachineClass {
+				a.logger.V(6).Info(fmt.Sprintf("Machine deployment %s is performing a rolling update", deployment.Name))
+				// Already existing machine deployments with a rolling update should have > 1 machine sets
+				if len(machineSets) <= 1 {
+					return retryutils.MinorError(fmt.Errorf("waiting for the MachineControllerManager to create the machine sets for the machine deployment (%s/%s)...", deployment.Namespace, deployment.Name))
+				}
+			}
+
+			// make sure that the machine set with the correct machine class for the machine deployment is deployed already
+			if machineSet := workerhelper.GetMachineSetWithMachineClass(wantedDeployment.Name, wantedDeployment.ClassName, ownerReferenceToMachineSet); machineSet == nil {
+				return retryutils.MinorError(fmt.Errorf("waiting for the machine controller manager to create the updated machine set for the machine deployment (%s/%s)...", deployment.Namespace, deployment.Name))
 			}
 
 			// If the shoot get hibernated we want to wait until all wanted machine deployments have been deleted
 			// entirely.
+			numberOfAwakeMachines += deployment.Status.Replicas
 			if controller.IsHibernated(cluster) {
-				numberOfAwakeMachines += existingMachineDeployment.Status.Replicas
 				continue
 			}
 
 			// If the Shoot is not hibernated we want to wait until all wanted machine deployments have as many
 			// available replicas as desired (specified in the .spec.replicas). However, if we see any error in the
 			// status of the deployment then we return it.
-			for _, failedMachine := range existingMachineDeployment.Status.FailedMachines {
+			for _, failedMachine := range deployment.Status.FailedMachines {
 				return retryutils.SevereError(fmt.Errorf("machine %s failed: %s", failedMachine.Name, failedMachine.LastOperation.Description))
 			}
 
 			// If the Shoot is not hibernated we want to wait until all wanted machine deployments have as many
 			// available replicas as desired (specified in the .spec.replicas).
-			if workerhealthcheck.CheckMachineDeployment(&existingMachineDeployment) == nil {
+			if workerhealthcheck.CheckMachineDeployment(&deployment) == nil {
 				numHealthyDeployments++
 			}
-			numDesired += existingMachineDeployment.Spec.Replicas
-			numUpdated += existingMachineDeployment.Status.UpdatedReplicas
-			numAvailable += existingMachineDeployment.Status.AvailableReplicas
+			numDesired += deployment.Spec.Replicas
+			numUpdated += deployment.Status.UpdatedReplicas
+			numAvailable += deployment.Status.AvailableReplicas
+			numUnavailable += deployment.Status.UnavailableReplicas
 		}
 
 		var msg string
-
 		switch {
 		case !controller.IsHibernated(cluster):
-			if numUpdated >= numDesired && int(numHealthyDeployments) == len(wantedMachineDeployments) {
+			// numUpdated == numberOfAwakeMachines waits until the old machine is deleted in the case of a rolling update with maxUnavailability = 0
+			// numUnavailable == 0 makes sure that every machine joined the cluster (during creation & in the case of a rolling update with maxUnavailability > 0)
+			if numUnavailable == 0 && numUpdated == numberOfAwakeMachines && int(numHealthyDeployments) == len(wantedMachineDeployments) {
 				return retryutils.Ok()
 			}
-			msg = fmt.Sprintf("Waiting until all desired machines are available (%d/%d machine objects available, %d/%d machinedeployments available)...", numAvailable, numDesired, numHealthyDeployments, len(wantedMachineDeployments))
+
+			// scale down cluster autoscaler during creation or rolling update
+			if clusterAutoscalerUsed && !autoscalerIsScaledDown {
+				if err := a.scaleClusterAutoscaler(ctx, worker, 0); err != nil {
+					return retryutils.SevereError(err)
+				}
+				a.logger.V(6).Info("Scaled down the cluster autoscaler", "namespace", worker.Namespace)
+				autoscalerIsScaledDown = true
+			}
+
+			// update worker status with condition that indicates an ongoing rolling update operation
+			if !workerStatusUpdatedForRollingUpdate {
+				if err := a.updateWorkerStatusMachineDeployments(ctx, worker, extensionsworker.MachineDeployments{}, true); err != nil {
+					return retryutils.SevereError(errors.Wrapf(err, "failed to update the machine status rolling update condition"))
+				}
+				workerStatusUpdatedForRollingUpdate = true
+			}
+
+			if numUnavailable == 0 && numAvailable == numDesired && numUpdated < numberOfAwakeMachines {
+				msg = fmt.Sprintf("Waiting until all old machines are drained and terminated. Waiting for %d machine(s)  ...", numberOfAwakeMachines-numUpdated)
+				break
+			}
+
+			msg = fmt.Sprintf("Waiting until machines are available (%d/%d desired machine(s) available, %d machine(s) pending, %d/%d machinedeployments available)...", numAvailable, numDesired, numUnavailable, numHealthyDeployments, len(wantedMachineDeployments))
 		default:
 			if numberOfAwakeMachines == 0 {
 				return retryutils.Ok()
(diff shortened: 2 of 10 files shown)


pkg/apis/extensions/v1alpha1/types_worker.go
```diff
@@ -175,3 +177,5 @@ type MachineDeployment struct {
 	// Maximum is the maximum number for this machine deployment.
 	Maximum int32 `json:"maximum"`
 }
+
+var RollingUpdate gardencorev1beta1.ConditionType = "RollingUpdate"
```
rfranzke: Can this be a const? Doc string missing. Should we name it `ConditionRollingUpdate` similar to other condition type constants?

extensions/pkg/controller/worker/genericactuator/actuator_test.go
```diff
@@ -276,4 +279,32 @@ var _ = Describe("Actuator", func() {
...
+			Reason: ReasonRollingUpdateProgressing,
+		}
+		conditionNoRollingUpdate = gardencorev1beta1.Condition{
+			Type:   extensionsv1alpha1.RollingUpdate,
+			Status: gardencorev1beta1.ConditionTrue,
+			Reason: ReasonRollingUpdateProgressing,
+		}
+	)
+
+	DescribeTable("#ConstructInternalClusterDomain", func(conditions []gardencorev1beta1.Condition, rollingUpdate bool, expectedConditionStatus gardencorev1beta1.ConditionStatus, expectedConditionReason string) {
```
rfranzke:
```suggestion
	DescribeTable("#getRollingUpdateCondition", func(conditions []gardencorev1beta1.Condition, rollingUpdate bool, expectedConditionStatus gardencorev1beta1.ConditionStatus, expectedConditionReason string) {
```

pkg/apis/extensions/v1alpha1/types_worker.go
danielfoehrKn: Sure will add the doc strings. 
Where do you got the naming from - see e.g [here](https://github.com/gardener/gardener/blob/fc8758d8ebafd9c17b0eeca723495d8796a27e52/pkg/apis/core/v1beta1/types_seed.go#L262) or [here](https://github.com/gardener/gardener/blob/fc8758d8ebafd9c17b0eeca723495d8796a27e52/pkg/apis/core/v1beta1/types_shoot.go#L963-L962).

rfranzke: Hm, then maybe let's prefix with the resource kind, i.e. `WorkerRollingUpdate`

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -84,23 +86,15 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 
 	// During the time a rolling update happens we do not want the cluster autoscaler to interfer, hence it
 	// is removed for now.
-	var (
-		clusterAutoscalerUsed = extensionsv1alpha1helper.ClusterAutoscalerRequired(worker.Spec.Pools)
-		rollingUpdate         = false
-	)
+	var clusterAutoscalerUsed = extensionsv1alpha1helper.ClusterAutoscalerRequired(worker.Spec.Pools)
+	rollingUpdateInProgress := a.isRollingUpdateInProgress(wantedMachineDeployments, existingMachineClassNames)
```
prashanth26:
```suggestion
	var (
		clusterAutoscalerUsed = extensionsv1alpha1helper.ClusterAutoscalerRequired(worker.Spec.Pools)
		rollingUpdateInProgress = a.isRollingUpdateInProgress(wantedMachineDeployments, existingMachineClassNames
	)
```
Minor suggestion keeping uniformity in mind.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -348,16 +357,18 @@ func (a *genericActuator) waitUntilWantedMachineDeploymentsAvailable(ctx context
...
 		}
 
 		var msg string
-
 		switch {
 		case !controller.IsHibernated(cluster):
-			if numUpdated >= numDesired && int(numHealthyDeployments) == len(wantedMachineDeployments) {
+			// numUpdated == numberOfAwakeMachines waits until the old machine is deleted in the case of a rolling update with maxUnavailability = 0
+			// numUnavailable == 0 makes sure that every machine joined the cluster (during creation & in the case of a rolling update with maxUnavailability > 0)
+			if numUnavailable == 0 && numUpdated == numberOfAwakeMachines && int(numHealthyDeployments) == len(wantedMachineDeployments) {
```
prashanth26, line 407: There still could be one condition that hasn't been captured. The terminating machines (could be due to larger drain timeouts and PDBs) from the old machineSets. I think it might be okay to ignore if you think it's not too important as they are only terminating machines and new machines are already created. 

To capture something like that, we might need to check all machines belonging to the old machineSet and make sure there sum is 0.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -195,14 +195,23 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
-	if err := a.updateWorkerStatusMachineDeployments(ctx, worker, wantedMachineDeployments); err != nil {
+	if err := a.updateWorkerStatusMachineDeployments(ctx, worker, wantedMachineDeployments, false); err != nil {
 		return errors.Wrapf(err, "failed to update the machine deployments in worker status")
 	}
 
 	return nil
 }
 
-func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, existingMachineDeployments *machinev1alpha1.MachineDeploymentList, wantedMachineDeployments worker.MachineDeployments, classKind string, clusterAutoscalerUsed bool) error {
+func (a *genericActuator) isRollingUpdateInProgress(wantedMachineDeployments extensionsworker.MachineDeployments, existingMachineClassNames sets.String) bool {
```
prashanth26: Shouldn't this function capture these conditions? - https://github.com/gardener/gardener/pull/2459/files#diff-ce16933f7218046c410bbaed72f771b7R368.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
rfranzke: Hm, why was this changed? IIUC it's not related to the `RollingUpdate` condition. Is this something that we have to improve, i.e., was the previous check not good enough? cc @prashanth26 

```golang
			if numUpdated >= numDesired && int(numHealthyDeployments) == len(wantedMachineDeployments) {	
```

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -405,12 +416,47 @@ func (a *genericActuator) updateWorkerStatusMachineDeployments(ctx context.Conte
...
+
+		worker.Status.Conditions = gardencorev1beta1helper.MergeConditions(worker.Status.Conditions, rollingUpdateCondition)
 		return nil
 	})
 }
 
+const ReasonRollingUpdateProgressing = "RollingUpdateProgressing"
+const ReasonNoRollingUpdate = "NoRollingUpdate"
+
+func getRollingUpdateCondition(conditions []gardencorev1beta1.Condition, isRollingUpdate bool) (gardencorev1beta1.Condition, error) {
```
rfranzke:
```suggestion
func buildRollingUpdateCondition(conditions []gardencorev1beta1.Condition, isRollingUpdate bool) (gardencorev1beta1.Condition, error) {
```

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -405,12 +416,47 @@ func (a *genericActuator) updateWorkerStatusMachineDeployments(ctx context.Conte
...
+			worker.Status.MachineDeployments = statusMachineDeployments
+		}
+
+		worker.Status.Conditions = gardencorev1beta1helper.MergeConditions(worker.Status.Conditions, rollingUpdateCondition)
 		return nil
 	})
 }
 
+const ReasonRollingUpdateProgressing = "RollingUpdateProgressing"
+const ReasonNoRollingUpdate = "NoRollingUpdate"
```
rfranzke: doc strings for exported variables + optionally const block

rfranzke (commented):
Generally, the PR looks OK. Earlier, I was thinking about a separate controller that is watching `MachineDeployment`s`/`MachineSet`s and then decides if a rolling update is ongoing. But as a rolling update can/should only be triggered by the reconciliation loop, it actually should also be fine to put it there.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: > This changes the logic that waits for the rolling update to finish.
Currently when triggering a rolling update (e.g volume resize) with maxUnavailability = 0, it does not wait for the new machine to join the cluster. This is a bug in my opinion (see Release notes).

In certain cases it did not wait for the Rolling update to complete. So the RollingUpdate condition was removed too early.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: > This changes the logic that waits for the rolling update to finish.
Currently when triggering a rolling update (e.g volume resize) with maxUnavailability = 0, it does not wait for the new machine to join the cluster. This is a bug in my opinion (see Release notes).

In certain cases it did not wait for the Rolling update to complete. So the RollingUpdate condition was removed too early.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: You are right. I did not want to change this check until I find a case, when comparing the MachineClasses is not sufficient. 

I found one: 

1. Volume Resize triggers rolling update
2. Creates new Machine Class
3. MachineDeployment references new MachineClass -> Creates new MachineSet
4. Situation 
  -  MachineClass has deletion timestamp set after waitUntilWantedMachineDeploymentsAvailable() completes, though Machineset cannot be deleted yet / takes more time (e.g PDB, ...)
 - waitUntilWantedMachineDeploymentsAvailable() did not complete yet because rollout not finished yet
4. Another rolling update is triggered before the old Machineset with the old MachineClass is deleted
  - Rolling update with a worker spec that produces the name of an already existing Machine class ([name generated based on hash](https://github.com/gardener/gardener-extension-provider-aws/blob/c6b1bd0e839b5dfe7519edf194bf88d0be8ae5e4/pkg/controller/worker/machines.go#L179-L178))
5. MachineClass referenced by the wanted deployment already exists -> rolling update not detected.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: @rfranzke you are right, it is not strictly related to putting a condition into the Worker status. But there is little use to such a condition, if the check for a rolling update itself does not work as expected.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: The check for the rolling update needs therefore (as you suggested), to be the same as the check when the rolling update is complete. 
Hence I ll move the check to the function waitUntilWantedMachineDeploymentsAvailable()

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
prashanth26: Yes, I think having `numUnavailable == 0` should improve capturing this rolling update for longer.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
prashanth26: Yes, makes sense.

danielfoehrKn: Actually the check for a rolling update needs to enhanced with a check for valid machine sets.  The cache of the client requesting the machine deployments can be outdated.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: > There still could be one condition that hasn't been captured. The terminating machines (could be due to larger drain timeouts and PDBs) from the old machineSets. 

This is catched by `numUnavailable == 0`

Also added checks prevent issues with a stale cache.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
prashanth26: But this cache should be eventually consistent right? So at worse, it would only be a few seconds later, right?

danielfoehrKn: yes, but then the check that should wait for the rolling update to finish is already done. So it does not wait.
That leads to a Shoot beeing 100 % reconciled even though the rolling update is still in progress.

extensions/pkg/controller/worker/genericactuator/actuator.go
```diff
@@ -281,6 +286,48 @@ func (a *genericActuator) shallowDeleteAllObjects(ctx context.Context, namespace
...
+	machineSets := &machinev1alpha1.MachineSetList{}
+	if err := a.client.List(ctx, machineSets, client.InNamespace(worker.Namespace)); err != nil {
+		return false, nil, err
+	}
+
+	isStuck, msg := isMachineControllerStuck(machineSets.Items, machineDeployments.Items)
+	return isStuck, msg, nil
+}
+
+func isMachineControllerStuck(machineSets []machinev1alpha1.MachineSet, machineDeployments []machinev1alpha1.MachineDeployment) (bool, *string) {
```
rfranzke, line 319: Probably it makes sense to have some grace period of ~1-2m or so compared to the machine deployments/sets creation timestamps.

extensions/pkg/controller/worker/helper/helper.go
```diff
@@ -0,0 +1,79 @@
...
+package helper
+
+import (
+	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
+)
+
+const (
+	nameLabel = "name"
+	// MachineSetKind is the kind of the owner reference of a machine set
+	MachineSetKind = "`MachineSet`"
```
rfranzke:
```suggestion
	MachineSetKind = "MachineSet"
```

extensions/pkg/controller/worker/helper/worker_helper_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestWorkerHelper(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Worker Helper Test Suite")
```
rfranzke:
```suggestion
	RunSpecs(t, "Extensions Controller Worker Helper Test Suite")
```

extensions/pkg/controller/worker/helper/helper.go
```diff
@@ -0,0 +1,79 @@
...
+				ownerToMachines[machineDeploymentName] = append(ownerToMachines[machineDeploymentName], machines[index])
+			}
+		}
+	}
+	return ownerToMachines
+}
+
+// BuildOwnerToMachineSetMap builds a map from a slice of machinev1alpha1.MachineSet, that maps the owner reference
+// to a slice of maMachineSets with the same owner reference
+func BuildOwnerToMachineSetMap(machineSets []machinev1alpha1.MachineSet) map[string][]machinev1alpha1.MachineSet {
```
rfranzke:
```suggestion
func BuildOwnerToMachineSetsMap(machineSets []machinev1alpha1.MachineSet) map[string][]machinev1alpha1.MachineSet {
```

extensions/pkg/controller/worker/machines.go
```diff
@@ -81,6 +82,17 @@ func (m MachineDeployments) HasDeployment(name string) bool {
 	return false
 }
 
+// GetDeploymentForName finds the deployment with the <name> from the <machineDeployments>
+// returns the machine deployment or nil
+func (m MachineDeployments) GetDeploymentForName(name string) *MachineDeployment {
```
rfranzke:
```suggestion
func (m MachineDeployments) FindByName(name string) *MachineDeployment {
```

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -135,14 +117,40 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
 		return errors.Wrapf(err, "failed to generate the machine deployment config")
 	}
 
 	// Wait until all generated machine deployments are healthy/available.
-	if err := a.waitUntilWantedMachineDeploymentsAvailable(ctx, cluster, worker, wantedMachineDeployments); err != nil {
+	if err := a.waitUntilWantedMachineDeploymentsAvailable(ctx, cluster, worker, existingMachineDeploymentNames, existingMachineClassNames, wantedMachineDeployments, clusterAutoscalerUsed); err != nil {
+		// check if the machine controller manager is stuck
+		isStuck, msg, err2 := a.IsMachineControllerStuck(ctx, worker)
+		if err2 != nil {
+			a.logger.Info("failed to check if Machine Controller Manager is stuck after unsuccessfully waiting for all machine deployments to be ready.", "error", err2.Error())
```
rfranzke:
```suggestion
			a.logger.Error("failed to check if Machine Controller Manager is stuck after unsuccessfully waiting for all machine deployments to be ready.", "error", err2.Error())
```

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -182,27 +190,34 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
-	if err := a.updateWorkerStatusMachineDeployments(ctx, worker, wantedMachineDeployments); err != nil {
+	if err := a.updateWorkerStatusMachineDeployments(ctx, worker, wantedMachineDeployments, false); err != nil {
 		return errors.Wrapf(err, "failed to update the machine deployments in worker status")
 	}
 
 	return nil
 }
 
-func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, existingMachineDeployments *machinev1alpha1.MachineDeploymentList, wantedMachineDeployments worker.MachineDeployments, classKind string, clusterAutoscalerUsed bool) error {
+func (a *genericActuator) scaleClusterAutoscaler(ctx context.Context, worker *extensionsv1alpha1.Worker, replicas int32) error {
```
rfranzke, line 206: Hm, ideally, the generic worker actuator does not scale the cluster-autoscaler anymore. It shouldn't know anything about it. Instead, it should simply set the `RollingUpdate` condition to `True` and then a controller in g/g can decide whether to scale down the CA. WDYT?
However, I see that it's difficult to do this in a backwards compatible way. So let's keep this code as is, but maybe you can add a comment that this code can be removed in a future version when such just mentioned controller in g/g is implemented.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -309,55 +324,115 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
...
-		var numHealthyDeployments, numUpdated, numAvailable, numDesired, numberOfAwakeMachines int32
+func (a *genericActuator) waitUntilWantedMachineDeploymentsAvailable(ctx context.Context, cluster *controller.Cluster, worker *extensionsv1alpha1.Worker, alreadyExistingMachineDeploymentNames sets.String, alreadyExistingMachineClassNames sets.String, wantedMachineDeployments extensionsworker.MachineDeployments, clusterAutoscalerUsed bool) error {
+	rollingUpdate := false
+	autoscalerIsScaledDown := false
+	workerStatusUpdatedForRollingUpdate := false
 
-		// Get the list of all existing machine deployments
-		existingMachineDeployments := &machinev1alpha1.MachineDeploymentList{}
-		if err := a.client.List(ctx, existingMachineDeployments, client.InNamespace(worker.Namespace)); err != nil {
+	return retryutils.UntilTimeout(ctx, 5*time.Second, 3*time.Minute, func(ctx context.Context) (bool, error) {
```
rfranzke: Why is it lowered to `3m`?

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -309,55 +324,115 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
...
-			if !wantedMachineDeployments.HasDeployment(existingMachineDeployment.Name) {
+			if wantedDeployment == nil {
 				continue
 			}
 
+			machineSets := ownerReferenceToMachineSet[deployment.Name]
+
+			// use `wanted deployment` for these checks, as the existing deployments can be based on an outdated cache
+			alreadyExistingMachineDeployment := alreadyExistingMachineDeploymentNames.Has(wantedDeployment.Name)
+			newMachineClass := !alreadyExistingMachineClassNames.Has(wantedDeployment.ClassName)
```
rfranzke, line 362: Why not checking the number of corresponding machine sets? Yes, it also works with checking for classes, but this might break in the future in case we add a "trigger-rolling-update-on-demand" functionality to MCM (no immediate plans, just thinking about it as it might be helpful in some cases). If there is no strong reason against it I'd prefer to use the no. of machine sets indicator.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -309,55 +324,115 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
...
+
+			if alreadyExistingMachineDeployment && newMachineClass {
+				rollingUpdate = true
+				a.logger.V(6).Info(fmt.Sprintf("Machine deployment %s is performing a rolling update", deployment.Name))
+				// Already existing machine deployments with a rolling update should have > 1 machine sets
+				if len(machineSets) <= 1 {
+					return retryutils.MinorError(errors.New(fmt.Sprintf("Waiting for the MachineControllerManager to create the machine sets for the machine deployment (%s/%s)...", deployment.Namespace, deployment.Name)))
+				}
```
rfranzke: Why do we need this check? Doesn't it not matter at all?

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -309,55 +324,115 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
...
+			}
+
+			// make sure that the machine set with the correct machine class for the machine deployment is already deployed
+			if machineSet := workerhelper.GetMachineSetWithMachineClass(wantedDeployment.Name, wantedDeployment.ClassName, ownerReferenceToMachineSet); machineSet == nil {
+				return retryutils.MinorError(errors.New(fmt.Sprintf("Waiting for the machine controller manager to create the updated machine set for the machine deployment (%s/%s)...", deployment.Namespace, deployment.Name)))
+			}
```
rfranzke: same question here

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
rfranzke: @prashanth26 can you please check this code again?

rfranzke (commented):
Generally, I find it a little hard to review this PR. It incorporates a lot of changes, especially in the already large `actuator_reconcile.go`. Anyways let's keep it like this for now, but maybe we can split it into smaller PRs in the future in case it touches such complex code.

extensions/pkg/controller/worker/genericactuator/actuator.go
danielfoehrKn: It is only called after it has already unsuccessfully waited (3/5 minutes) for the machinedeployments to become ready. After this I think there is no need for a grace period any more?

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: Yes that what I was referring to in the description. I agree with what you are saying, that was the idea.

> Please note that the cluster autoscaler scale down behaviour has not changed (scaled down during creation and rolling updates).
In the future, the cluster autoscaler might not be required to be scaled down any more.
See this issue and the corresponding issue on the MCM

In this PR I just did not want to change the behaviour when the scale down happens. This should be done when the external controller is ready.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: The idea was to lower the amount of time that has to pass until it is checked if the MC is stuck.
The operation will be retried anyhow due to the erroneous reconciliation.

However no strong opinion here.

rfranzke: If it's lowered then the `Worker`'s last operation is faster set to `Error`. I would keep it at `5m`. I think MCM being stuck is a very rare case anyways.

extensions/pkg/controller/worker/genericactuator/actuator.go
rfranzke: Hm yes, but wouldn't it be more convenient for the caller of this function to not care about these "grace periods" on his own? You are anyways reading the machine deployments/machine sets, so you can add the grace period behaviour this function so that callers aren't bothered with it.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: > "trigger-rolling-update-on-demand" 

can you elaborate on this one? I am not sure what you exactly mean.

Reasons for using the MachineClass
- ability to detect a rolling update vs a create operation (you cannot infer that from the machine set alone?)
- the cache of the client that retrieves the machine sets is many times outdated. Alternative would be to use a direct client, but I would prefer not to.
- MachineClass is something that the extension itself creates, so it does not need to indirectly rely on the MCM.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: It is a sanity check.  I observed we need it when operating on a stale cache.
We want to avoid making the decision wether the rolling update already completed based on an outdated cache. This was being done before and lead to not waiting for the update to complete.

The good thing is that we know what the MCM should create, so we can check for these things and can easily detect abnormalities. So we can garuantee to make decisions based on current information.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: Another sanity check. Yes the NUMBER of machine sets might be two, but this does not garuantee that the required machine set with the correct class is present.
There could be two MachineSets, but those could be left-overs (e.g from the previous reconciliation & not garbage collected), or the  MCM is stuck.

If we do not detect this, this could lead to issues as we only check the status of the machine deployment (which could have a status indication the machine sets are fine, even though the correct machine set is not even created yet.)

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: @prashanth26

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: @prashanth26

extensions/pkg/controller/worker/genericactuator/actuator.go
danielfoehrKn: Mhm ok, to me that sounds like additional complexity as this check currently never fails. But I get where you coming from.

Maybe someone else has another opinion here. E.g. @prashanth26

rfranzke: Why additional complexity? You can simply exit early/not consider machine deployments/sets that are too young. IMO, it's now more complex to demand from every caller to consider this grace period on its own.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
rfranzke: > can you elaborate on this one? I am not sure what you exactly mean.

In the future, we might introduce an operation annotation (or similar mechanism) in the MCM/Gardener to allow end-users triggering a rolling update of their worker machines without generating a new machine class.

> ability to detect a rolling update vs a create operation (you cannot infer that from the machine set alone?)

Hm, is this distinction important? What is it used for?

> the cache of the client that retrieves the machine sets is many times outdated. Alternative would be to use a direct client, but I would prefer not to.

I don't get it. Why isn't the cache of the machine classes outdated either?

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
rfranzke: ok, I see, thanks.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
rfranzke: Thanks

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: I can also provide a list of the test cases I used to verify this PR if that helps @prashanth26 .

prashanth26:
Hey Guys, 

I will address the comments on the PR by EOD.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: > Hm, is this distinction important? What is it used for? 

It is used so that the RollingUpdate condition is only true if there is a rolling update, not a creation of a pool.

> I don't get it. Why isn't the cache of the machine classes outdated either?

Because we are using the [already existing machine class names](https://github.com/gardener/gardener/blob/c5432f84de767d46d624d5126f45a63df02b3e31/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go#L94-L93) that we determined before creating the machinedeployments.

> In the future, we might introduce an operation annotation (or similar mechanism) in the MCM/Gardener to allow end-users triggering a rolling update of their worker machines without generating a new machine class.

Mhm, then I guess we need a direct client to avoid cache issues.

rfranzke: > It is used so that the RollingUpdate condition is only true if there is a rolling update, not a creation of a pool.

OK, but can't we consider a creation also like a rolling update for the sake of simplicity? I think the boundary conditions are the same (CA should be disabled, etc.), so this way we don't need special code.

> Mhm, then I guess we need a direct client to avoid cache issues.

Well, I'm okay with keeping it as is for now until @tim-ebert has also enhanced the extensions library with the new clientmap concepts. Just saying that checking the machine sets would be more robust.

danielfoehrKn: > OK, but can't we consider a creation also like a rolling update for the sake of simplicity? I think the boundary conditions are the same (CA should be disabled, etc.), so this way we don't need special code 

Maybe then we would be better off labeling the condition something like  "UpdateOperation" not "RollingUpdate"?
So any time either a create or rolling update is happening we set the condition to true.

And if in the future we would need to make a distinction between which deployments are actually in creation and which are in rolling update, we could still introducing a condition in the `worker.machineDeployments`.

rfranzke: My personal preference is to keep the "RollingUpdate" name, I think it's precise and easy to understand. Yes, we don't remove old nodes in creation because there are none, but IMO it doesn't matter.

danielfoehrKn: Why does it make sense to label something "RollingUpdate" if in fact it indicates RollingUpdates + Creations. 
Also what purpose does that serve? 
I feel like this is exactly not precise and confuses external components. 

>  Yes, we don't remove old nodes in creation because there are none

There are also X amount of machine created in parallel unlike in a rolling update.

rfranzke: "RollingUpdate" is a well established term and indicates that the nodes are being updated/reconciled. In this case it doesn't matter if it's a creation or a replacement of existing nodes. A more precise name would be something like "NodesReconciliation", but I prefer "RollingUpdate" because it's well-known and used in the community. 
Components that are evaluating this information anyways don't care about the characters inside this string. They just want to know that something is going on with the nodes, e.g. to down-scale the CA. As you explained, there is no use-case where a distinction between create/replacement is required.

extensions/pkg/controller/worker/genericactuator/actuator.go
```diff
@@ -281,6 +286,48 @@ func (a *genericActuator) shallowDeleteAllObjects(ctx context.Context, namespace
...
+	// map the owner reference to the existing machine sets
+	ownerReferenceToMachineSet := workerhelper.BuildOwnerToMachineSetsMap(machineSets)
+
+	for _, machineDeployment := range machineDeployments {
+		if !controllerutils.HasFinalizer(&machineDeployment, "machine.sapcloud.io/machine-controller-manager") {
```
prashanth26: Minor suggestion - `machine.sapcloud.io/machine-controller-manager` can be made a string const?

extensions/pkg/controller/worker/genericactuator/actuator.go
prashanth26: `MachineControllerStuck` is this something that was observed earlier? If yes, how often was it?  Because if we have something like this, we have a serious problem at the MCM and need to fix it.  

If it's something we haven't observed, I am not sure how helpful only this method is. But, I guess, it's okay to have. But not really sure if this is a reliable way for detecting all conditions during reconciliation. For eg: with support for scale to zero, the worker-pool might be created and the replicas could still be `0` right? Correct me if I am wrong here. 

A better way to check if MCM is stuck would be to check on the liveliness probe on the MCM? and implement this/a better logic there?

> Hm yes, but wouldn't it be more convenient for the caller of this function to not care about these "grace periods" on his own? You are anyways reading the machine deployments/machine sets, so you can add the grace period behavior this function so that callers aren't bothered with it.

If we do plan to have this logic, I would agree with @rfranzke on having the timeouts/grace-period inside the function for the caller to not care about it.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
prashanth26: Yes with `numUnavailable == 0` it looks better. Thanks. 

> I can also provide a list of the test cases I used to verify this PR if that helps @prashanth26 .

Do you this handy?

extensions/pkg/controller/worker/genericactuator/actuator.go
danielfoehrKn: > If we do plan to have this logic, I would agree with @rfranzke on having the timeouts/grace-period inside the function for the caller to not care about it. 

Alright, will do.

> MachineControllerStuck is this something that was observed earlier? If yes, how often was it? Because if we have something like this, we have a serious problem at the MCM and need to fix it.

Observed couple of times (I also have seen this personally sometimes), thats why I tackled it as part of my quality issues.
But I agree, that should be fixed in the MCM. It is just very difficult to reproduce due to its sporadic nature.

> with support for scale to zero, the worker-pool might be created and the replicas could still be 0 right?

Makes sense. I will remove the check for MachineSets having a status.
So the only check would be if the necessary MachineSet with the right MachineClass got created.
Any other ideas?

>  check on the liveliness probe on the MCM

The MCM pod is already deployed with a liveness probe that is performed by the kubelet. Is that what you had in mind? 

Or are you suggesting that there should be a probe build into the provider extension? 
I am not sure if that is the right place for it.

prashanth26: > Observed a couple of times (I also have seen this personally sometimes), thats why I tackled it as part of my quality issues. But I agree, that should be fixed in the MCM. It is just very difficult to reproduce due to its sporadic nature.

Okay. This is scaring me now. Please ping me if and when you see this ever again.

> Makes sense. I will remove the check for MachineSets having a status.
So the only check would be if the necessary MachineSet with the right MachineClass got created.
Any other ideas?

Yes, this should be enough for now. We can improve the liveliness checks on MCM if we see this reoccurring. 

> The MCM pod is already deployed with a liveness probe that is performed by the kubelet. Is that what you had in mind?

Yeah the ones probed by kubelet. But maybe improve the tasks to be performed by MCM during a liveliness check.

prashanth26:
Hi @danielfoehrKn ,

I have left a few comments yesterday and a new one today. I hope you are looking into a fix for it.  The reason i mentioned it here is because i didn't recieve notifications for the comments i left today here - https://github.com/gardener/gardener/pull/2459/files#r445513774

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: - Worker Pool deletion
	- wait until unwanted machines have been deleted
- Worker Pool creation
	-  does not set rolling update condition to true

**Causing a rolling update (new MachineClass)** -> e.g Volume resize
- Max unavailability = 0 
	- detects rolling update, waits for machines to join, then waits for old machines to be deleted (observe in logs)
	- With PDB that does not allow deletion
		- Wait for deletion of machine (force delete)
	- with same pre-existing MachineClass (e.g due to failed reconciliation / crash)
		- because we are using the MachineClass as an indicator, this is not properly detected as a rolling update
		- should nontheless wait for the update to be ready
- Creation of worker in multiple avail. zones
	- increasing the worker count only scales up an already existing MachineSet. Should not be a rolling update
- Max Unavailability = 1
	- Wait for the new machine to join
- MCM is stuck simulation
	- simulate MachineDeployment that has no corresponding MachineSet created yet. Simulated with a MachineDeployment that the MCM ignores - which has no finalizer. Temporarely disabled the finalizer check in the extension provider.
	- should detect a stuck MCM and should restart the MCM pod.

extensions/pkg/controller/worker/genericactuator/actuator.go
danielfoehrKn: > Okay. This is scaring me now. Please ping me if and when you see this ever again.

Sure will do. 

> We can improve the liveliness checks on MCM if we see this reoccurring. 

ok makes sense.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: Alright, I ll call it "RollingUpdate", I guess I am unaware of such a definition. Thanks for the review.

danielfoehrKn:
@prashanth26, I answered your review comments.
Thanks a lot for the review so far. 
I ll update the PR tomorrow.

danielfoehrKn:
Updated the PR PTAL @prashanth26

extensions/pkg/controller/worker/genericactuator/actuator.go
prashanth26: These changes look good to me.  Thanks.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
prashanth26: I think scenarios are exhaustive enough for now. looks good to me.

danielfoehrKn: I know it is quite some work, but I would be glad if someone would again test those (& possibly other scenarios that I did not think of). 
 ll do it again once all comments are resolved.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
danielfoehrKn: I will switch to MachineSets, once Tim Eberts has also enhanced the extensions library with the new clientmap concepts. Shall I open another PR for this or shall this PR stay open in the meanwhile?
cc @tim-ebert

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
prashanth26: Hi @danielfoehrKn ,

I can try to do it. However, I might be OOO until Tuesday. Would only be able to test the changes on Tuesday.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
timebertt: I'm not sure, when the ClientMap will make it into the extensions library. So probably you wouldn't want to wait for it, I guess.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -135,14 +117,40 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
+		// check if the machine controller manager is stuck
+		isStuck, msg, err2 := a.IsMachineControllerStuck(ctx, worker)
+		if err2 != nil {
+			a.logger.Error(err2, "failed to check if the Machine Controller Manager is stuck after unsuccessfully waiting for all machine deployments to be ready.", "namespace", worker.Namespace)
+			// continue in order to return `err` and determine error codes
+		}
+
+		if isStuck {
+			podList := corev1.PodList{}
+			if err2 := a.client.List(ctx, &podList, client.InNamespace(worker.Namespace), client.MatchingLabels(map[string]string{"role": "machine-controller-manager"})); client.IgnoreNotFound(err2) != nil {
```
ialidzhikov: There is no need to check for NotFound error as `List` invocation will return empty list, but not NotFound err.

ialidzhikov: You can use `client.MatchingLabels{"role": "machine-controller-manager"}` instead of  `client.MatchingLabels(map[string]string{"role": "machine-controller-manager"})`

```suggestion
			if err2 := a.client.List(ctx, &podList, client.InNamespace(worker.Namespace), client.MatchingLabels{"role": "machine-controller-manager"}); err2 != nil {
```

extensions/pkg/controller/worker/helper/helper.go
```diff
@@ -0,0 +1,79 @@
...
+			if machineDeploymentName, ok := machine.Labels[nameLabel]; ok {
+				ownerToMachines[machineDeploymentName] = append(ownerToMachines[machineDeploymentName], machines[index])
+			}
+		}
+	}
+	return ownerToMachines
+}
+
+// BuildOwnerToMachineSetsMap builds a map from a slice of machinev1alpha1.MachineSet, that maps the owner reference
+// to a slice of maMachineSets with the same owner reference
```
ialidzhikov:
```suggestion
// to a slice of MachineSets with the same owner reference
```

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -135,14 +117,40 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
+				return errors.Wrapf(err2, "failed to restart stuck machine controller manager for worker (%s/%s)", worker.Namespace, worker.Name)
+			}
+
+			for _, pod := range podList.Items {
+				if err2 := a.client.Delete(ctx, &pod); err2 != nil {
+					return errors.Wrapf(err2, "failed to restart stuck machine controller manager for worker (%s/%s)", worker.Namespace, worker.Name)
```
ialidzhikov:
```suggestion
					return errors.Wrapf(err2, "failed to delete stuck machine controller manager pod for worker (%s/%s)", worker.Namespace, worker.Name)
```

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -135,14 +117,40 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
+				}
+			}
+			a.logger.Info("Successfully restarted stuck machine controller manager pod", "reason", msg, "worker namespace", worker.Namespace, "worker name", worker.Name)
```
ialidzhikov:
```suggestion
			a.logger.Info("Successfully deleted stuck machine controller manager pod", "reason", msg, "worker namespace", worker.Namespace, "worker name", worker.Name)
```

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -135,14 +117,40 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
+		isStuck, msg, err2 := a.IsMachineControllerStuck(ctx, worker)
+		if err2 != nil {
+			a.logger.Error(err2, "failed to check if the Machine Controller Manager is stuck after unsuccessfully waiting for all machine deployments to be ready.", "namespace", worker.Namespace)
+			// continue in order to return `err` and determine error codes
+		}
+
+		if isStuck {
+			podList := corev1.PodList{}
+			if err2 := a.client.List(ctx, &podList, client.InNamespace(worker.Namespace), client.MatchingLabels(map[string]string{"role": "machine-controller-manager"})); client.IgnoreNotFound(err2) != nil {
+				return errors.Wrapf(err2, "failed to restart stuck machine controller manager for worker (%s/%s)", worker.Namespace, worker.Name)
```
ialidzhikov:
```suggestion
				return errors.Wrapf(err2, "failed to list machine controller manager pods for worker (%s/%s)", worker.Namespace, worker.Name)
```

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -135,14 +117,40 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
...
 		return errors.Wrapf(err, "failed to generate the machine deployment config")
 	}
 
 	// Wait until all generated machine deployments are healthy/available.
-	if err := a.waitUntilWantedMachineDeploymentsAvailable(ctx, cluster, worker, wantedMachineDeployments); err != nil {
+	if err := a.waitUntilWantedMachineDeploymentsAvailable(ctx, cluster, worker, existingMachineDeploymentNames, existingMachineClassNames, wantedMachineDeployments, clusterAutoscalerUsed); err != nil {
+		// check if the machine controller manager is stuck
+		isStuck, msg, err2 := a.IsMachineControllerStuck(ctx, worker)
+		if err2 != nil {
+			a.logger.Error(err2, "failed to check if the Machine Controller Manager is stuck after unsuccessfully waiting for all machine deployments to be ready.", "namespace", worker.Namespace)
```
ialidzhikov: The other usages are lowercase.

```suggestion
			a.logger.Error(err2, "failed to check if the machine controller manager is stuck after unsuccessfully waiting for all machine deployments to be ready.", "namespace", worker.Namespace)
```

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -309,55 +324,113 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
...
+				a.logger.V(6).Info(fmt.Sprintf("Machine deployment %s is performing a rolling update", deployment.Name))
+				// Already existing machine deployments with a rolling update should have > 1 machine sets
+				if len(machineSets) <= 1 {
+					return retryutils.MinorError(errors.New(fmt.Sprintf("Waiting for the MachineControllerManager to create the machine sets for the machine deployment (%s/%s)...", deployment.Namespace, deployment.Name)))
+				}
+			}
+
+			// make sure that the machine set with the correct machine class for the machine deployment is deployed already
+			if machineSet := workerhelper.GetMachineSetWithMachineClass(wantedDeployment.Name, wantedDeployment.ClassName, ownerReferenceToMachineSet); machineSet == nil {
+				return retryutils.MinorError(errors.New(fmt.Sprintf("Waiting for the machine controller manager to create the updated machine set for the machine deployment (%s/%s)...", deployment.Namespace, deployment.Name)))
```
ialidzhikov:
```suggestion
				return retryutils.MinorError(fmt.Errorf("Waiting for the machine controller manager to create the updated machine set for the machine deployment (%s/%s)...", deployment.Namespace, deployment.Name))
```

ialidzhikov (commented):
Minor nits from my side.

prashanth26:
Hi guys,

I am running a few tests to check this PR today. Please hold merging until then. 

Thanks.

ialidzhikov:
@danielfoehrKn , it looks like `concourse-ci/verify` failed because of a flaky test which was recently fixed in the master branch with https://github.com/gardener/gardener/pull/2532. If you want, you could rebase to get this change.

danielfoehrKn:
@ialidzhikov I am expecting more comments from @prashanth26  who is currently testing. Then I ll update the PR.

prashanth26:
I just tested out the PR for some shoot/worker-pool creation/update/deletion scenarios. It seems to be working well. Nice job with the PR. However, please address comments by other reviewers.

/lgtm

prashanth26 (dismissed):

prashanth26:
Also i see an `hold` until 1.7 from @tim-ebert .

/hold

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
prashanth26: I tested some scenarios, seems to be working well. Looks good to me.

danielfoehrKn: Thanks a lot for testing!

danielfoehrKn:
@hardikdr @prashanth26  any additional comments? Otherwise I would like to get this PR in soon to have it thoroughly tested and included in the next release.

hardikdr (approved):

prashanth26 (approved):
/lgtm

timuthy (approved):

PR #2460, opened 2020-06-16
Make Terraform output compatible to any type

timuthy:

/area robustness
/kind enhancement
/priority normal

This PR changes the Terraformer state handling, so that it can deal with output formats other than strings. Before the output was casted to a string which does not work if the Terraform state has the type "Number" in the `outputs` block.

```
        "countUpdateDomains": {
          "value": 5,
          "type": "number"
        },
```

This is a prerequisite for https://github.com/gardener/gardener-extension-provider-azure/issues/109.

**Release note**:

```noteworthy developer
The Terraformer can now deal with output types other than String in the Terraform state.
```


timebertt (approved):
/lgtm

rfranzke (approved):
/lgtm

PR #2464, opened 2020-06-16
Fix shoot maintenance defaulting

rfranzke:

/area usability
/kind bug
/priority normal

Creating a shoot with `.spec.maintenance=nil` was rejected with the following error message:

```
The Shoot "foo" is invalid: spec.maintenance: Required value: maintenance information is required
```

However, it was supposed to be defaulted. This PR fixes this behaviour and correctly defaults the maintenance section if it's not provided by the end-user:

```yaml
spec:
  maintenance:
    autoUpdate:
      kubernetesVersion: true
      machineImageVersion: true
    timeWindow:
      begin: 080000+0000
      end: 090000+0000
```

**Release note**:

```improvement user
The `.spec.maintenance` settings are now correctly defaulted when a `Shoot` is being created without any such configuration.
```


timuthy (approved):
/lgtm

timebertt (approved):
/lgtm

danielfoehrKn (approved):

PR #2465, opened 2020-06-16
Guestbook test is now compatible with shoots having allowPrivilegedContainers=false

dguendisch:
/area testing
/kind enhancement
/priority normal

Up to now the guestbook tests always required a shoot with `spec.kubernetes.allowPrivilegedContainers: true`, hence they couldn't be used to test clusters without that privilege.
This PR makes the respective deployments (redis and guestbook) compatible with shoots using 
`spec.kubernetes.allowPrivilegedContainers: false` so that the same tests can be used to test both flavors.

### DIFF
--- a/test/framework/applications/guestbooktest.go
+++ b/test/framework/applications/guestbooktest.go
@@ -125,8 +125,6 @@ func (t *GuestBookTest) DeployGuestBookApp(ctx context.Context) {
 	shoot := t.framework.Shoot
 	if !shoot.Spec.Addons.NginxIngress.Enabled {
 		ginkgo.Fail("The test requires .spec.addons.nginxIngress.enabled to be true")
-	} else if shoot.Spec.Kubernetes.AllowPrivilegedContainers == nil || !*shoot.Spec.Kubernetes.AllowPrivilegedContainers {
-		ginkgo.Fail("The test requires .spec.kubernetes.allowPrivilegedContainers to be true")
 	}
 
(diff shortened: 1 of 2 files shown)


schrodit (dismissed):
lgtm

rfranzke (dismissed):
/lgtm

test/framework/applications/guestbooktest.go
```diff
@@ -135,6 +133,12 @@ func (t *GuestBookTest) DeployGuestBookApp(ctx context.Context) {
 		"cluster": map[string]interface{}{
 			"enabled": false,
 		},
+		"podSecurityPolicy": map[string]interface{}{
```
ialidzhikov, line 136: Isn't podSecurityPolicy introduced as of v9.5.0 of the redis chart - https://github.com/helm/charts/pull/18425/files? 
It looks like we are using v7.0.0 of the chart - https://github.com/gardener/gardener/blob/dba80d637da3da853e4cf3128fee9a989344cfd3/test/framework/framework_test.go#L61.

schrodit: you are referring to just the test for the test framework. The actual guestbook test uses 10.2.1 https://github.com/gardener/gardener/blob/master/test/framework/applications/guestbooktest.go#L49

ialidzhikov: You are right, my bad. :(

ialidzhikov (approved):
/lgtm

PR #2466, opened 2020-06-16
Add node anti-affinities and PDBs for CoreDNS

zanetworker:
/area networking 
/kind enhancement

Currently there is a possibility that CoreDNS pods would reside on the same node. This can be problematic during machine OS upgrades. 

**Release note**:

```improvement operator
CoreDNS pods are now protected by a PDB during machine upgrades and should reside on different nodes for HA.
```


rfranzke:
Thanks! Should we also define a PDB?

rfranzke (approved):
/lgtm

timuthy (approved):
/lgtm

timebertt (approved):
Nice!
/lgtm

PR #2468, opened 2020-06-17
Improve Metrics-Server scraping

### LINKED ISSUE #2455: metrics-server unable to scrape kubelet

/area metering
/kind bug
/priority normal

**What happened**:
In our setup with openstack the metrics-server runs into the following error:
```
E0615 12:41:12.158755       1 manager.go:111] unable to fully collect metrics: unable to fully scrape metrics from source kubelet_summary:shoot--core--ab7vxtoulk-worker-gnvyr-z1-5c4dc9599b-kdq4j: unable t 
 o fetch metrics from Kubelet shoot--core--ab7vxtoulk-worker-gnvyr-z1-5c4dc9599b-kdq4j (shoot--core--ab7vxtoulk-worker-gnvyr-z1-5c4dc9599b-kdq4j): Get https://shoot--core--ab7vxtoulk-worker-gnvyr-z1-5c4dc9 
 599b-kdq4j:10250/stats/summary?only_cpu_and_memory=true: dial tcp: lookup shoot--core--ab7vxtoulk-worker-gnvyr-z1-5c4dc9599b-kdq4j on 100.64.0.10:53: no such host
```
The default setting for the kubelet-preferred-address in the master branch from the matrics server is documented as `[Hostname,InternalDNS,InternalIP,ExternalDNS,ExternalIP]`. In the tags/v0.3.6 there is no documentation about the default.

In our setup is the hostname not resolvable, so the metrics-server should fallback to the InternalIP. 

We try following settings in the metrics-server chart (/charts/shoot-core/components/charts/metrics-server/templates/deployment.yaml)

Set the kubelet-preferred-address (see [README.md from metrics-server](https://github.com/kubernetes-sigs/metrics-server/blob/master/README.md) direct to InternalIP
```yaml
...
      containers:
      - name: metrics-server
        image: {{ index .Values.images "metrics-server" }}
        imagePullPolicy: Always
        command:
        - /metrics-server
        - --profiling=false
        - --kubelet-preferred-address-types=InternalIP
        # nobody user only can write in home folder
        - --cert-dir=/home/certdir
        - --secure-port=8443
...
```
-> with this setting the metric-server works like expected

Set the kubelet-preferred-address (see [README.md from metrics-server](https://github.com/kubernetes-sigs/metrics-server/blob/master/README.md) to the "default" order ([Hostname,InternalDNS,InternalIP,ExternalDNS,ExternalIP]).
```yaml
...
      containers:
      - name: metrics-server
        image: {{ index .Values.images "metrics-server" }}
        imagePullPolicy: Always
        command:
        - /metrics-server
        - --profiling=false
        - --kubelet-preferred-address-types=[Hostname,InternalDNS,InternalIP,ExternalDNS,ExternalIP]
        # nobody user only can write in home folder
        - --cert-dir=/home/certdir
        - --secure-port=8443
...
```
-> that works, this should be the default but without this setting the metrics-server isn`t able to scrape the kubelet.

The metrics-server is able to scrape the kubelet.

The hostname from the worker nodes is unresolvable in the kubernetes cluster. 

**Environment**:

- Gardenlet version: 1.6.1
- Kubernetes version (use `kubectl version`): 1.18.0
- Cloud provider or hardware configuration: OpenStack
- Metrics-Server: 0.3.6

timuthy:

/area monitoring
/area robustness
/kind enhancement
/priority normal

This PR configures a list of preferable address types for the Metrics-Server. Please look at the linked issue for more information.

Fixes #2455

@danielfoehrKn would be nice if you could double check the integration test since you have experience in this field.

/cc @jia-jerry

**Release note**:

```improvement operator
Runtime metrics for Pods and Nodes are now also available in environments which don't support a domain name resolution for worker nodes.
```

### DIFF
--- a/test/integration/shoots/applications/metrics.go
+++ b/test/integration/shoots/applications/metrics.go
@@ -1,0 +1,117 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+	Overview
+		- Tests that runtime metrics are available
+
+	AfterSuite
+		- Cleanup Workload in Shoot
+
+	Test: Create arbitrary deployment
+	Expected Output
+		- Metrics are available
+ **/
+
+package applications
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+	"github.com/gardener/gardener/pkg/utils/retry"
+	"github.com/gardener/gardener/test/framework"
+
+	"github.com/gardener/gardener/test/framework/resources/templates"
+	"github.com/onsi/ginkgo"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metricsv1beta1 "k8s.io/metrics/pkg/apis/metrics/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	metricsTestTimeout = 10 * time.Minute
+)
+
+var _ = ginkgo.Describe("Shoot application metrics testing", func() {
+
+	f := framework.NewShootFramework(&framework.ShootConfig{
+		CreateTestNamespace: true,
+	})
+
+	var (
+		name = "metrics-test"
+	)
+
+	f.Beta().CIt("should read runtime metrics", func(ctx context.Context) {
+		templateParams := map[string]string{
+			"name":      name,
+			"namespace": f.Namespace,
+		}
+		err := f.RenderAndDeployTemplate(ctx, f.ShootClient, templates.SimpleLoadDeploymentName, templateParams)
+		framework.ExpectNoError(err)
+
+		err = f.WaitUntilDeploymentIsReady(ctx, name, f.Namespace, f.ShootClient)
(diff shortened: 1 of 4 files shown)


test/integration/shoots/applications/metrics.go
```diff
@@ -0,0 +1,130 @@
...
+		var writer bytes.Buffer
+		err = testTmpl.Execute(&writer, map[string]string{
+			"name":      name,
+			"namespace": f.Namespace,
+		})
+		framework.ExpectNoError(err)
+
+		manifestReader := kubernetes.NewManifestReader(writer.Bytes())
+		err = f.ShootClient.Applier().ApplyManifest(ctx, manifestReader, kubernetes.DefaultMergeFuncs)
+		framework.ExpectNoError(err)
```
schrodit: you could use this helper function to deploy and render the template https://github.com/gardener/gardener/blob/master/test/framework/template.go#L28 
Example: https://github.com/gardener/gardener/blob/master/test/framework/applications/guestbooktest.go#L169

```suggestion
		templateParams := map[string]string{
			"name":      name,
			"namespace": f.Namespace,
		}
		err = f.RenderAndDeployTemplate(ctx, f.ShootClient, templates.SimpleLoadDeploymentName, templateParams )
		framework.ExpectNoError(err)
```

schrodit (changes requested):

test/integration/shoots/applications/metrics.go
```diff
@@ -0,0 +1,130 @@
...
+
+		if len(pods.Items) == 0 {
+			ginkgo.Fail("at least one pod is needed")
+		}
+		podName := pods.Items[0].Name
+
+		ginkgo.By("Check runtime metrics")
+		framework.ExpectNoError(
+			retry.Until(ctx, 30*time.Second, func(ctx context.Context) (bool, error) {
```
rfranzke, line 86: Hm, maybe we should have a higher timeout here?

timuthy: It's the retry waiting time, not a timeout. Do you still want to increase it?

rfranzke: Hm, I can't see it in the code. For how long is this retried in case it fails?

timuthy: The timeout is configured [here](https://github.com/gardener/gardener/blob/8539fa54c882921a9b7f62bcfb947dc9516001aa/test/integration/shoots/applications/metrics.go#L124) and is set to 10 minutes. The context handed over by the test framework is closed then and `retry.Until` won't continue.

test/integration/shoots/applications/metrics.go
timuthy: Thank @schrodit, great advice. I changed the implementation for the [network test](https://github.com/gardener/gardener/blob/3576ef7931632d4a0f66ee46f942aaa3dee9bdbd/test/integration/shoots/applications/networking.go) as well. PTAL.

schrodit (approved):
also thanks for updating the network test

rfranzke (approved):

PR #2442, opened 2020-06-09
Restructure botanist deploy network CRD & add missing tests

danielfoehrKn:
**How to categorize this PR?**

/area quality
/kind test
/priority normal

**What this PR does / why we need it**:

Restructuring the botanist Network CRD deployment  during the Shoot reconciliation to be a testable component.
Also introducing missing tests .

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement operator

```

PR #2471, opened 2020-06-18
Add Cluster Identity to Garden, Seed and Shoot clusters 

### LINKED ISSUE #2054: Control plane migration: shoot specific owner ids for DNS entries

swilen-iwanow:

/area control-plane
/kind enhancement
/priority normal

Adds a `ConfigMap` named `cluster-identity` which holds the identity as data.
The cluster identity for the:
 - Garden cluster is deployed by the Garden API Server in the `kube-system` namespace on start.
 - Seed cluster is deployed in the `kube-system` namespace, if it doesn't already exist in case of Shooted Seed, and is set 
to the `seed.status.clusterIdentity`
 - Shoot cluster is deployed in the `kube-system` namespace in the Shoot and written to `shoot.status.clusterIdentity`.

All identities are in the form described in the issue.

Fixes #2054

**Release note**:

```noteworthy operator
Clusters in the Gardener topology (Garden, Seed and Shoot), now have ConfigMaps deployed in the `kube-system` namespace, that hold the cluster identity. Cluster identities for the Shoot and Seed are also visible in their `status.clusterIdentity` fields.
```
```action operator
Gardener API Server has a new mandatory flag for defining desired cluster identity of the Garden cluster that is `--cluster-identity`. In case gardener is operated by helm - charts for the identity of the clusters are also provided without default values, meaning that they should be filled by the Gardener operators/admin.  
```


charts/gardener/controlplane/charts/application/templates/configmap-cluster-idenetity.yaml
```diff
@@ -0,0 +1,7 @@
+kind: ConfigMap
```
rfranzke: typo in file name. But isn't this something that should rather be a CLI parameter to the gardener-apiserver, and it deploys it at start-up? What if people deploy g/g without Helm? This cluster identity will be an essential part for Gardener, so you need to have it.

charts/seed-bootstrap/templates/configmap-cluster-idenetity.yaml
```diff
@@ -0,0 +1,8 @@
+kind: ConfigMap
```
rfranzke: typo in filename

rfranzke: What if such configmap already exists (e.g., in case the seed is a garden)?

charts/shoot-core/components/charts/cluster-identity/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
rfranzke: Why we have duplicated the same chart three times? Probably it makes sense to have it only once somewhere in `charts`. It's always the same.

pkg/operation/types.go
```diff
@@ -72,6 +72,7 @@ type Operation struct {
 	APIServerAddress          string
 	APIServerHealthCheckToken string
 	SeedNamespaceObject       *corev1.Namespace
+	ClusterIdentity           *string
```
rfranzke: Why is this a pointer?

pkg/operation/seed/seed.go
```diff
@@ -870,3 +876,22 @@ func GetWildcardCertificate(ctx context.Context, c client.Client) (*corev1.Secre
 	}
 	return nil, nil
 }
+
+// determineClusterIdentity determines the identity of a cluster, in cases where the identity was
+// created manually or the Seed was created as Shoot, and later registered as Seed and already has
+// an identity, it should not be changed.
+func determineClusterIdentity(k8sSeedClient kubernetes.Interface) (string, error) {
```
```suggestion
func determineClusterIdentity(ctx context.Context, c client.Client) (string, error) {
```

pkg/operation/seed/seed.go
```diff
@@ -870,3 +876,22 @@ func GetWildcardCertificate(ctx context.Context, c client.Client) (*corev1.Secre
...
+	clusterIdentity := &corev1.ConfigMap{}
+	if err := k8sSeedClient.Client().Get(context.TODO(), kutil.Key(v1beta1constants.GardenNamespace, v1beta1constants.ClusterIdentity), clusterIdentity); err != nil {
```
```suggestion
	if err := c.Get(ctx, kutil.Key(v1beta1constants.GardenNamespace, v1beta1constants.ClusterIdentity), clusterIdentity); err != nil {
```

pkg/operation/seed/seed.go
```diff
@@ -870,3 +876,22 @@ func GetWildcardCertificate(ctx context.Context, c client.Client) (*corev1.Secre
...
+		if !apierrors.IsNotFound(err) {
+			return "", err
+		}
+
+		gardenNamespace := &corev1.Namespace{}
+		if err := k8sSeedClient.Client().Get(context.TODO(), kutil.Key(v1beta1constants.GardenNamespace), gardenNamespace); err != nil {
```
```suggestion
		if err := c.Get(ctx, kutil.Key(v1beta1constants.GardenNamespace), gardenNamespace); err != nil {
```

rfranzke (changes requested):
Does it make sense to store the cluster identities in some immutable status fields in the `Shoot` and `Seed` to prevent the I/O for the Get calls in every reconciliation?

/assign @mandelsoft @MartinWeindel 
/invite @mandelsoft @MartinWeindel 
/needs second-opinion

pkg/operation/types.go
swilen-iwanow: I prefer to be nil if no one set it, instead of "".

charts/seed-bootstrap/templates/configmap-cluster-idenetity.yaml
swilen-iwanow: It will reuse the value from the existing configmap.  https://github.com/gardener/gardener/pull/2471/files#diff-5ca37df1e48dc263572bf7822db58a8aR883

charts/gardener/controlplane/charts/application/templates/configmap-cluster-idenetity.yaml
swilen-iwanow: I was also thinking to add it as part of the API Server start up, but was not sure if it is its responsibility. Should I add it as a PostStartHook and remove it from the charts?

pkg/operation/types.go
rfranzke: But it will be always set in the shoot reconciliation, so how is it different from fields like `APIServerAddress` or `APIServerHealthCheckToken`

charts/gardener/controlplane/charts/application/templates/configmap-cluster-idenetity.yaml
rfranzke: What were your thoughts that made you unsure about the responsibility?

swilen-iwanow: Was wondering if the API Server should depend on a flag that it has no use of or has nothing  to do with the API Server as a server, but a Gardener instance as a whole. After that we had the meeting and we imply to provide it as a chart. Anyway, I am going to add it as part of it.

pkg/operation/types.go
swilen-iwanow: You are right. I (🤡) was thinking for cases where others try to use it before it is set, and because it is not part of the `Botanist` init to panic if they try to dereference it or at least suggest that this value might not be set .

charts/gardener/controlplane/charts/application/templates/configmap-cluster-idenetity.yaml
rfranzke: Hm, yes, but that's the best way to ensure this identity exists. If it must be added as part of the deployment then it might likely be missing and will lead to complexity and follow-up issues.

pkg/operation/types.go
rfranzke: Generally, this might make sense, but I don't think this use-case exists. The whole package is not build in a way that "others" could or should use it.  :-(

vendor/k8s.io/apiserver/pkg/server/options/cluster_identity.go
```diff
@@ -0,0 +1,48 @@
+/*
```
swilen-iwanow, line 1: Just realised these are in the vendor lol

hack/local-development/start-apiserver
```diff
@@ -24,15 +24,18 @@ tls_dir="$__tmp_tls_dir"
 trap cleanup_kubeconfig EXIT
 trap cleanup_tls_dir EXIT
 
+# TODO: Make cluster identity unique for every
```
danielfoehrKn: Left over TODO? This would be quite important for the local dev scenario.

hack/local-development/start-apiserver
```diff
@@ -41,37 +44,37 @@ case $(k8s_env) in
         kubectl -n garden port-forward service/etcd 32379:2379 > /dev/null &
         trap "kill -9 $! > /dev/null" EXIT
         GO111MODULE=on go run \
-          -mod=vendor \
```
danielfoehrKn: Is there a reason for changing the intendations?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -101,6 +101,11 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 			Name: "Deploying Shoot namespace in Seed",
 			Fn:   flow.TaskFn(botanist.DeployNamespace).RetryUntilTimeout(defaultInterval, defaultTimeout),
 		})
+		ensureShootClusterIdentity = g.Add(flow.Task{
+			Name:         "Ensuring Shoot cluster identity in Shoot's namespace in the Seed",
+			Fn:           flow.TaskFn(botanist.EnsureClusterIdentity).RetryUntilTimeout(defaultInterval, defaultTimeout),
```
danielfoehrKn, line 107: Gardener gradually moves towards testable components in the botanist e.g. for the [Network CRD deployment](https://github.com/gardener/gardener/pull/2442). I think new reconciliation steps should be implemented with components and unit tests.
I know this will result in higher initial effort, but it makes sense in the long run.

pkg/operation/botanist/controlplane.go
```diff
@@ -116,6 +116,54 @@ func (b *Botanist) deleteNamespace(ctx context.Context, name string) error {
 	return err
 }
 
+// EnsureClusterIdentity ensures that Shoot cluster-identity ConfigMap exists and stores its data in the operation
```
danielfoehrKn: Can you add to the description that this updates the Shoot.Status in the garden cluster as well as the identity-config map in the seed cluster?

danielfoehrKn (changes requested):
Can we also have documentation for the specific identities created in the context of the control plane migration? [The content here](https://github.com/gardener/gardener/issues/2054#issuecomment-601776007) would be a start.
For now a document under development would do I guess. Or update the GEP with the latest changes. 
Later we def. need a document under docs/concept and docs/usage aswell.

hack/local-development/start-apiserver
swilen-iwanow: well actually I added the ${USER} in the end do you think it is enough?

charts/gardener/controlplane/values.yaml
```diff
@@ -158,6 +158,7 @@ global:
           verbs: ["get", "list", "watch"]
         - level: Metadata
       webhook: {}
+    clusterIdentity: garden-cluster-identity
```
rfranzke: Can you move it up/below? Currently, it's in the `webhook` section (the commented fields below belong to `webhook`).

pkg/apis/core/validation/seed.go
```diff
@@ -164,7 +164,16 @@ func ValidateSeedSpecUpdate(newSeedSpec, oldSeedSpec *core.SeedSpec, fldPath *fi
...
 func ValidateSeedStatusUpdate(newSeed, oldSeed *core.Seed) field.ErrorList {
-	allErrs := field.ErrorList{}
+	var (
+		allErrs   = field.ErrorList{}
+		fldPath   = field.NewPath("status")
+		oldStatus = oldSeed.Status
+		newStatus = newSeed.Status
+	)
+
+	if oldStatus.ClusterIdentity != nil && newStatus.ClusterIdentity != oldStatus.ClusterIdentity {
```
rfranzke:
```suggestion
	if oldStatus.ClusterIdentity != nil && !apiequality.Semantic.DeepEqual(oldStatus.ClusterIdentity, newStatus.ClusterIdentity) {
```

rfranzke: unit test missing

pkg/apis/core/validation/shoot.go
```diff
@@ -250,6 +250,9 @@ func ValidateShootStatusUpdate(newStatus, oldStatus core.ShootStatus) field.Erro
 		allErrs = append(allErrs, apivalidation.ValidateImmutableField(newStatus.TechnicalID, oldStatus.TechnicalID, fldPath.Child("technicalID"))...)
 	}
 
+	if oldStatus.ClusterIdentity != nil && newStatus.ClusterIdentity != oldStatus.ClusterIdentity {
```
```suggestion
	if oldStatus.ClusterIdentity != nil && !apiequality.Semantic.DeepEqual(oldStatus.ClusterIdentity, newStatus.ClusterIdentity) {
```

rfranzke: unit test missing

pkg/apiserver/apiserver.go
```diff
@@ -74,3 +77,22 @@ func (c completedConfig) New() (*GardenerServer, error) {
 
 	return s, nil
 }
+
+type ExtraOptions struct {
```
rfranzke, line 82: doc string

pkg/apiserver/apiserver.go
```diff
@@ -74,3 +77,22 @@ func (c completedConfig) New() (*GardenerServer, error) {
...
+	if len(o.ClusterIdentity) == 0 {
+		allErrors = append(allErrors, fmt.Errorf("--cluster-identity must be specified"))
+	}
+
+	return allErrors
+}
+
+// AddFlags adds flags related to cluster identity to the options
+func (o *ExtraOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&o.ClusterIdentity, "cluster-identity", o.ClusterIdentity, "This flag is used for specifying the identiy of the Garden cluster")
```
rfranzke: typo

pkg/operation/botanist/clusteridentity/clusteridentity.go
```diff
@@ -0,0 +1,124 @@
...
+	name            string
+	uid             string
+
+	seedClient   client.Client
+	gardenClient client.Client
+
+	logger *logrus.Entry
+}
+
+func New(
```
rfranzke, line 45: doc string

pkg/operation/botanist/clusteridentity/clusteridentity.go
```diff
@@ -0,0 +1,124 @@
...
+		shoot := &gardencorev1beta1.Shoot{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      c.name,
+				Namespace: c.namespace,
+			},
+		}
+
+		c.clusterIdentity = &clusterIdentity
+		patch := []byte(fmt.Sprintf(`{"status":{"clusterIdentity": "%s"}`, clusterIdentity))
+		if err := c.gardenClient.Status().Patch(ctx, shoot, client.RawPatch(types.StrategicMergePatchType, patch)); err != nil {
```
rfranzke, line 76: Any specific reason why you use `PATCH` here?

pkg/operation/botanist/controlplane.go
```diff
@@ -116,6 +116,32 @@ func (b *Botanist) deleteNamespace(ctx context.Context, name string) error {
...
 }
 
+// EnsureClusterIdentity ensures that Shoot cluster-identity ConfigMap exists and stores its data
+// in the operation. Updates shoot.status.clusterIdentity if it doesn't exist already.
+func (b *Botanist) EnsureClusterIdentity(ctx context.Context) error {
+	return b.Shoot.Components.ClusterIdentity.Deploy(ctx)
+}
+
+// determineShootClusterIdentity determines the identity of a Shoot cluster, in cases that the identity exists it returns its value, else computes the identity and returns it.
+func (b *Botanist) determineShootClusterIdentity(ctx context.Context) (string, error) {
```
rfranzke: This looks like the same code as in `pkg/operation/botanist/clusteridentity/clusteridentity.go`.

rfranzke (changes requested):

charts/shoot-core/components/charts/cluster-identity/Chart.yaml
rfranzke: This comment is not yet resolved. We are still having the chart multiple times.

pkg/operation/botanist/clusteridentity/clusteridentity.go
swilen-iwanow: To avoid passing the Shoot as dependency, conflict during update or GET for the latest Shoot object.

charts/shoot-core/components/charts/cluster-identity/Chart.yaml
swilen-iwanow: The chart for the Shoot cluster identity is part of the "shoot-core" managed resources. Should I create new managed resource or change the logic how the shoot-core is calculated?

rfranzke: No, you could simply symlink the chart into `shoot-core` to avoid duplication.

pkg/operation/botanist/clusteridentity/clusteridentity.go
```diff
@@ -0,0 +1,124 @@
...
+	return err
+}
+
+// Destroy returns nil
+func (c *shootClusterIdentity) Destroy(ctx context.Context) error {
+	return nil
+}
+
+// determineShootClusterIdentity determines the identity of a Shoot cluster, in cases that the identity exists it returns its value, else computes the identity and returns it.
+func (c *shootClusterIdentity) determineShootClusterIdentity(ctx context.Context) (string, error) {
```
rfranzke: Shouldn't this return the `.status.clusterIdentity` of the `Shoot` if set instead of recomputing it?

charts/shoot-core/components/charts/cluster-identity/Chart.yaml
swilen-iwanow: Ah yes, that will work, thanks!

pkg/operation/botanist/clusteridentity/clusteridentity.go
rfranzke: Hm, ok, fine. OTOH, you probably could put this whole block into an `if` block that is only executed when the `.status.clusterIdentity` of the shoot is not set yet. Otherwise, nothing needs to be done and you can exit early?

pkg/operation/botanist/clusteridentity/clusteridentity.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	shootClusterIdentity := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      v1beta1constants.ClusterIdentity,
+			Namespace: c.seedNamespace,
+		},
+	}
+
+	_, err := controllerutil.CreateOrUpdate(ctx, c.seedClient, shootClusterIdentity, func() error {
```
rfranzke: Why do we need this configmap in the seed? Isn't the `.status.clusterIdentity` in the `Shoot` enough?

pkg/operation/botanist/clusteridentity/clusteridentity.go
swilen-iwanow: Yes, this method is called if the status.clusterIdentity is missing. This is checked in  `Deploy` where c.clusterIdentity is the .status.clusterIdentity

pkg/operation/botanist/clusteridentity/clusteridentity.go
swilen-iwanow: To be honest I am not sure, when we spoke I was left with the impression we want to have a copy, but I am not certain.

pkg/operation/botanist/controlplane.go
swilen-iwanow: It is a leftover, I will delete it

pkg/operation/botanist/clusteridentity/clusteridentity.go
rfranzke: I see, thank you.

pkg/operation/botanist/clusteridentity/clusteridentity.go
rfranzke: @mandelsoft @MartinWeindel did this "request" come from your side?

hack/local-development/start-apiserver
danielfoehrKn: Is ${USER}  truly uniqueue?

swilen-iwanow: Certainly not, but how unique do we need it to be. I guess in a single Gardener instance it would be enough. Would you like to suggest something else to be used instead.

pkg/operation/botanist/clusteridentity/clusteridentity_test.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"context"
+
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	gardenv1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	"github.com/gardener/gardener/pkg/operation/botanist/clusteridentity"
+	"github.com/gardener/gardener/pkg/operation/botanist/component"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+
+	//. "github.com/gardener/gardener/test/gomega"
```
rfranzke: ?

rfranzke (commented):
Thanks, looks quite good now. I haven't yet played with it locally. @MartinWeindel @mandelsoft please check

pkg/operation/botanist/clusteridentity/clusteridentity.go
rfranzke: I think the configmap is in the `kube-system` namespace of every cluster in order to be accessible for other components.

swilen-iwanow: I removed this, if the colleagues actually need it, I will add it again.

rfranzke: Hm, so the cluster identity configmap does only exist in kube-system in garden and shoot, but not in seed?
If this is true then I think we should also deploy it for the seed, to be consistent across our cluster topologys

swilen-iwanow: No, I am talking about the copy of the `Shoot` cluster identity in the `Seed` ( in the Shoot's namespace)

rfranzke: OK, thanks, then it should be all good! :)

ialidzhikov:
Could you please check `make check`?

/needs rebase

rfranzke (commented):
Thanks, I don't have further comments for now. Waiting for reviews from @danielfoehrKn @mandelsoft @MartinWeindel before I play with it locally...

hack/local-development/start-apiserver
danielfoehrKn: I guess you are right, it only has to be unique within one gardener installation.

danielfoehrKn (commented):
/lgtm have not tested it locally though

timuthy (dismissed):
Nice PR!
/lgtm

rfranzke (dismissed):

pkg/operation/botanist/clusteridentity/clusteridentity.go
```diff
@@ -0,0 +1,97 @@
...
+
+// Destroy returns nil
+func (c *shootClusterIdentity) Destroy(ctx context.Context) error {
+	return nil
+}
+
+// computeShootClusterIdentity computes the identity of a Shoot cluster and returns it.
+func (c *shootClusterIdentity) computeShootClusterIdentity(ctx context.Context) (string, error) {
+	gardenClusterIdentity := &corev1.ConfigMap{}
+	if err := c.gardenClient.Get(ctx, kutil.Key(metav1.NamespaceSystem, v1beta1constants.ClusterIdentity), gardenClusterIdentity); err != nil {
```
mandelsoft: Is the garden identity read for every shoot identity creation? It is immutable and should be constant for the whole lifetime of the gardenlet.

mandelsoft: It is validated, that the identity is not changed in shoot and seed objects, but what about the configmap(s) in the kube-system namespace. Here it should also not be possible to change the `cluster-identity` field.

May be I missed it, but is the shoot identity part of the cluster info object in the shoot namespace of the seed?

MartinWeindel (dismissed):
lgtm

pkg/operation/botanist/clusteridentity/clusteridentity.go
rfranzke: > Is the garden identity read for every shoot identity creation? It is immutable and should be constant for the whole lifetime of the gardenlet.

Agreed, it could be read once when the gardenlet starts.

> May be I missed it, but is the shoot identity part of the cluster info object in the shoot namespace of the seed?

I don't think so, no. The identities are only stored in the kube-system namespaces. Probably it should be also stored in a `.status.identity` field in the shoot status.

> It is validated, that the identity is not changed in shoot and seed objects, but what about the configmap(s) in the kube-system namespace. Here it should also not be possible to change the cluster-identity field.

Hm, good point. How to validate it? The only way is to have webhooks, but this is also again a lot of effort. I suggest to put the shoot identity in the new `.status.identity` (immutable) field and use it as the source of truth.

swilen-iwanow: Hello @mandelsoft 

> Is the garden identity read for every shoot identity creation? It is immutable and should be constant for the whole lifetime of the gardenlet.

Yes I agree. 

> May be I missed it, but is the shoot identity part of the cluster info object in the shoot namespace of the seed?

It is part of the `Shoot.Status.clusterIdentity` and it is in the `Cluster` resource in the `Seed`, although for the `Cluster` nothing extra was implemented, this is done by synching it in the end of the `Shoot` reconciliation. Wouldn't this be enough?

>> It is validated, that the identity is not changed in shoot and seed objects, but what about the configmap(s) in the kube-system namespace. Here it should also not be possible to change the cluster-identity field.

> Hm, good point. How to validate it? The only way is to have webhooks, but this is also again a lot of effort. I suggest to put the shoot identity in the new .status.identity (immutable) field and use it as the source of truth.

The way it works now for the Seed is, the main source of truth is the `status.clusterIdentity`, but if it is missing we lookup for a `ConfigMap`, if  a ConfigMap doesn't exist as well, we compute the clusterIdentity. This is done only for the Seed. For the Shoot we are using only the clusterIdentity field in the status and if it is missing we compute it.

rfranzke: Sounds great, thanks Svilen.

pkg/operation/seed/seed.go
```diff
@@ -728,6 +741,7 @@ func BootstrapCluster(k8sSeedClient kubernetes.Interface, seed *Seed, secrets ma
 		"ingress": map[string]interface{}{
 			"basicAuthSecret": monitoringBasicAuth,
 		},
+		"cluster-identity": map[string]interface{}{"clusterIdentity": &seed.Info.Status.ClusterIdentity},
```
stoyanr, line 750:
```suggestion
		"clusterIdentity": &seed.Info.Status.ClusterIdentity,
```
I don't see `cluster-identity` in the seed charts above, or am I missing something?

pkg/operation/seed/seed.go
```diff
@@ -643,6 +643,19 @@ func BootstrapCluster(k8sSeedClient kubernetes.Interface, seed *Seed, secrets ma
...
 
+	if seed.Info.Status.ClusterIdentity == nil {
+		seedClusterIdentity, err := determineClusterIdentity(context.TODO(), k8sSeedClient.Client())
+		if err != nil {
+			return err
+		}
+
+		seed.Info.Status.ClusterIdentity = &seedClusterIdentity
+		if err := k8sGardenClient.Client().Status().Update(context.TODO(), seed.Info); err != nil {
+
```
stoyanr:
```suggestion
```
Remove this extra line

pkg/operation/seed/seed.go
```diff
@@ -849,3 +863,22 @@ func GetWildcardCertificate(ctx context.Context, c client.Client) (*corev1.Secre
...
+	if err := c.Get(ctx, kutil.Key(metav1.NamespaceSystem, v1beta1constants.ClusterIdentity), clusterIdentity); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return "", err
+		}
+
+		gardenNamespace := &corev1.Namespace{}
+		if err := c.Get(ctx, kutil.Key(v1beta1constants.GardenNamespace), gardenNamespace); err != nil {
+			return "", err
+		}
+		return string(gardenNamespace.UID), nil
```
stoyanr, line 887: I don't understand why do we set the Seed cluster identity to the garden namespace UID if the configmap is not found? Shouldn't we generate it in a manner similar to the shoot cluster identity? Or vice versa, why do we bother generating the Shoot cluster identity in the way above instead of just using an UID there as well (see my comment above)? Why the difference between Seed and Shoot?

pkg/operation/botanist/clusteridentity/clusteridentity.go
```diff
@@ -0,0 +1,97 @@
...
+		return "", err
+	}
+
+	clusterIdentity := fmt.Sprintf("%s-%s-%s", c.seedNamespace, c.uid, gardenClusterIdentity.Data[v1beta1constants.ClusterIdentity])
```
stoyanr: Why do we bother generating the Shoot cluster identity in this ways instead of just using an UID (perhaps the `c.uid` part), similarly to the Seed?

stoyanr (commented):
Looks great, I only have a small questions and 2 minor corrections (one of them looks like it might be a bug though)

pkg/apis/core/v1alpha1/types_common.go
```diff
@@ -108,6 +108,8 @@ type Gardener struct {
 	Name string `json:"name" protobuf:"bytes,2,opt,name=name"`
 	// Version is the version of the Gardener which last acted on a resource.
 	Version string `json:"version" protobuf:"bytes,3,opt,name=version"`
+	// ClusterIdentity is the cluster identity of the Gardener cluster
+	ClusterIdentity string `json:"clusterIdentity" protobuf:"bytes,4,opt,name=clusterIdentity"`
```
rfranzke: Why is it added here? It's even not added as an optional field but as required field which violates the API conventions. Now it will be written to every shoot's `.status.gardener`, so there will be `.status.gardener.clusterIdentity` and `.status.clusterIdentity` which is highly confusing.
I suggest to remove this field from the API and simply handle it internally in the Gardenlet code.

pkg/operation/botanist/clusteridentity/clusteridentity.go
swilen-iwanow: @mandelsoft point here was that in the IaaS provider for example, the user will only see an UID that will not give any hint where this is comming from

pkg/apis/core/v1alpha1/types_common.go
swilen-iwanow: Okay, will change it

pkg/operation/seed/seed.go
stoyanr: ok, forget about this, this is ok, there seems to be a minor issue in `values.yaml` of the `seed-bootstrap` chart that confused me

charts/seed-bootstrap/values.yaml
```diff
@@ -137,6 +137,8 @@ ingress:
   # admin : admin base64 encoded
   basicAuthSecret: YWRtaW46JGFwcjEkSWRSaVM5c3MkR3U1MHMxaGUwL2Z6Tzh2elE4S1BEMQ==
 
+clusterIdentity: seed-cluster-identity
```
stoyanr:
```suggestion
cluster-identity:
  clusterIdentity: seed-cluster-identity
```

pkg/operation/seed/seed.go
rfranzke: Hm, good point. Doesn't this also mean that every seed will have the same identity (=uid of garden NS)?

swilen-iwanow: No, we are using the Garden ns in the Seed

swilen-iwanow: As the proposal goes when cluster identity is missing (configMap) we generate it using the `Namespace` UID. Maybe i have to change it to use the `kube-system` namespace instead of the garden namespace, because it could be missing in the Seed.

stoyanr: Yes, I also think `kube-system` is better, since Seeds (unlike Shoots) don't have own UIDs. How about including things like the seed name, and perhaps the garden identity in the Seed identity as well, similarly to the Shoots?

cmd/gardenlet/app/gardenlet.go
```diff
@@ -339,8 +343,19 @@ func NewGardenlet(ctx context.Context, cfg *config.GardenletConfiguration) (*Gar
 		return nil, err
 	}
 
+	gardenClusterIdentity := &corev1.ConfigMap{}
+	if err := k8sGardenClient.Client().Get(ctx, kutil.Key(metav1.NamespaceSystem, v1beta1constants.ClusterIdentity), gardenClusterIdentity); err != nil {
```
stoyanr: I am getting an error when trying to run `gardenlet` caused by this line:

```
panic: the cache is not started, can not read objects

goroutine 1 [running]:
main.main()
	/Users/i500152/go/src/github.com/gardener/gardener/cmd/gardenlet/main.go:57 +0x2ef
exit status 2
make: *** [start-gardenlet] Error 1
```

I wrapped the error below to make sure that the error is produced exactly by this line here. This should be fixed.

cmd/gardenlet/app/gardenlet.go
```diff
@@ -339,8 +343,19 @@ func NewGardenlet(ctx context.Context, cfg *config.GardenletConfiguration) (*Gar
...
+		return nil, err
+	}
+
+	clusterIdentity, ok := gardenClusterIdentity.Data[v1beta1constants.ClusterIdentity]
+	if !ok {
+		return nil, fmt.Errorf("unable to get Gardener`s cluster identity: %v", err)
```
stoyanr:
```suggestion
		return nil, errors.New("unable to get Gardener`s cluster identity")
```
It makes no sense to wrap this error here, if you want it wrapped you'd better do it where it actually occurred above.

cmd/gardenlet/app/gardenlet.go
```diff
@@ -339,8 +343,19 @@ func NewGardenlet(ctx context.Context, cfg *config.GardenletConfiguration) (*Gar
 		return nil, err
 	}
 
+	gardenClusterIdentity := &corev1.ConfigMap{}
+	if err := k8sGardenClient.Client().Get(ctx, kutil.Key(metav1.NamespaceSystem, v1beta1constants.ClusterIdentity), gardenClusterIdentity); err != nil {
+		return nil, err
```
stoyanr:
```suggestion
		return nil, fmt.Errorf("unable to get Gardener`s cluster identity: %v", err)
```

stoyanr (changes requested):
I am not able to start gardenlet locally, I suspect there is a bug somewhere, see the code comments below.

cmd/gardenlet/app/gardenlet.go
swilen-iwanow: As we talked, the problem is that I am using the Client() instead of DirectClient(). Didn't catch this because i have not enabled this feature gate.

cmd/gardenlet/app/gardenlet.go
swilen-iwanow: Done.

cmd/gardenlet/app/gardenlet.go
swilen-iwanow: Done.

stoyanr (dismissed):
lgtm, and I didn't encounter any more errors during testing

swilen-iwanow:
Rebased, PTAL.

timuthy:
Let me finally check.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -102,6 +102,11 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 			Name: "Deploying Shoot namespace in Seed",
 			Fn:   flow.TaskFn(botanist.DeployNamespace).RetryUntilTimeout(defaultInterval, defaultTimeout),
 		})
+		ensureShootClusterIdentity = g.Add(flow.Task{
+			Name:         "Ensuring Shoot cluster identity in Shoot's namespace in the Seed",
```
timuthy:
```suggestion
			Name:         "Ensuring Shoot cluster identity",
```

pkg/operation/botanist/clusteridentity/clusteridentity.go
```diff
@@ -0,0 +1,80 @@
...
+		clusterIdentity := fmt.Sprintf("%s-%s-%s", c.seedNamespace, c.uid, c.gardenClusterIdentity)
+		shoot := &gardencorev1beta1.Shoot{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      c.name,
+				Namespace: c.namespace,
+			},
+		}
+
+		patch := []byte(fmt.Sprintf(`{"status": {"clusterIdentity": "%s"}}`, clusterIdentity))
+		return c.gardenClient.Status().Patch(ctx, shoot, client.RawPatch(types.StrategicMergePatchType, patch))
```
timuthy: W.r.t. this PR https://github.com/gardener/gardener/pull/2522 it might be a good idea to update the shoot as well as the cluster resource, so that extensions in the seed can use the cluster-identity.

timuthy (changes requested):
Just two comments. Looks good and works fine otherwise.

timuthy (approved):
/lgtm
I tested it successfully with my local setup as well. Nice work!

danielfoehrKn (approved):

timuthy:
@swilen-iwanow before we release the next Gardener version with the change can you please add an "action required" block to the release notes, that makes Gardener operators change the default identity value in the Helm chart as a best practice? I'm afraid that many Gardener installations will get the id `garden-cluster-identity` otherwise.

swilen-iwanow:
Is that ok? They only have to set the flag before running the API Server

timebertt:
Well, it is good to make operators aware of the new flag before running `gardener-apiserver` but I would also add a sentence, that the value should be adapted here:
https://github.com/gardener/gardener/blob/2e993388559b10582f1fc00623cb371b5262df65/charts/gardener/controlplane/values.yaml#L175

I assume, most operators use the helm charts to deploy the gardener components and they would use `garden-cluster-identity` if they don't change it.
Maybe a good option would be to comment out the the default values in `values.yaml` and add `required` to the usage here:
https://github.com/gardener/gardener/blob/2e993388559b10582f1fc00623cb371b5262df65/charts/gardener/controlplane/charts/runtime/templates/apiserver/deployment.yaml#L174-L176

But then I think, `hack/check-charts.sh` would break, so we should probably add some kind of `values-test.yaml` file.
WDYT @swilen-iwanow @timuthy?

PR #2474, opened 2020-06-18
Adapt client usage in integration tests

timebertt:

/area testing
/kind enhancement
/priority normal

This PR adapts the usage of kubernetes clients in the integration test framework and all integration tests to the changes made in #2449.
Basically it just replaces all occurences of `kubernetes.Interface.Client()` with `kubernetes.Interface.DirectClient()`. This way, the integration tests explicitly do not rely on cached clients and always use direct clients when talking to clusters. This is done because the effort of implementing a clean and proper way of using cached clients in our integration tests seems to be huge and also not worth it, as we wouldn't profit too much from cached clients there.

/invite @gardener/test-infra-maintainers 

Part of #2414

✅ Depends on #2449, will rebase and drop the commits from #2449 once it is merged.
Please select the last commit to review the changes of this PR.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/test/framework/shootframework.go
+++ b/test/framework/shootframework.go
@@ -77,11 +77,11 @@ func NewShootFramework(cfg *ShootConfig) *ShootFramework {
 		Config:            cfg,
 	}
 
-	ginkgo.BeforeEach(func() {
+	CBeforeEach(func(ctx context.Context) {
 		f.CommonFramework.BeforeEach()
 		f.GardenerFramework.BeforeEach()
-		f.BeforeEach()
-	})
+		f.BeforeEach(ctx)
+	}, 8*time.Minute)
 	CAfterEach(f.AfterEach, 10*time.Minute)
 	return f
 }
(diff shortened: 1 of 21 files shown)


test/framework/shootframework.go
```diff
@@ -77,11 +77,11 @@ func NewShootFramework(cfg *ShootConfig) *ShootFramework {
...
 	}
 
-	ginkgo.BeforeEach(func() {
+	CBeforeEach(func(ctx context.Context) {
 		f.CommonFramework.BeforeEach()
 		f.GardenerFramework.BeforeEach()
-		f.BeforeEach()
-	})
+		f.BeforeEach(ctx)
+	}, 2*time.Minute)
```
dguendisch: Not sure `2mins` is OK here. As far as I can see, the old `ctx` had no explicit timeout; during `BeforeEach` it calls `f.AddShoot` which creates a seed client and e.g. has an own retry timeout of already `5mins` (should the seed be down temporarily or whatever).

dguendisch (changes requested):
One comment, rests looks good!

test/framework/shootframework.go
timebertt: Agreed. I increased it to `8m`.

dguendisch (approved):
lgtm, thanks

rfranzke (approved):

PR #2477, opened 2020-06-19
Remove hostPort usage for konnectivity-tunnel pods

zanetworker:

/area networking
/kind bug
/priority normal

This PR removes the usage of hostPorts on the konnectivity tunnel pods. This results in blocked deployment when scheduled on the same node. 

**Release note**:

```noteworthy operator
A bug which blocked APIserver deployments on the same node (due to hostPort usage) is now fixed. 
```


rfranzke (approved):
/lgtm

danielfoehrKn (approved):

PR #2478, opened 2020-06-19
Support vertical pod autoscaler for shoots

### LINKED ISSUE #1584: Offer VPA deployment next to shoot control plane
We should optionally offer to deploy and configure the VPA for shoot clusters. We can run the required deployments in the shoot namespace in the seed and wire them accordingly so that the webhooks can seed-internally communicate with the shoot's API server.

End-users could then get a VPA for their clusters out of the box, controlled and managed by Gardener. A new `.spec.kubernetes.verticalPodAutoscaler` flag would be needed in order to control whether it's enabled and to offer potential configuration options.

cc @vasu1124

rfranzke:

/area auto-scaling
/kind enhancement
/priority normal

This PR allows to enable the vertical-pod-autoscaler for shoot clusters.

Fixes #1584

/invite @ggaurav10 @amshuman-kr @wyb1 @zanetworker 

Due to https://github.com/gardener/gardener-resource-manager/issues/58 the `VPA` objects for the system components are currently not deployed. However, we should already discuss whether it makes sense to introduce them for the various components (e.g., CoreDNS now has both VPA and HPA).

Unfortunately, the VPA does not support the use-case of running out-of-cluster out-of-the-box. Consequently, we are "faking" an in-cluster config that actually points to the shoot instead of the seed where VPA runs.

The use-case for shooted seeds is not supported yet. I'll follow-up on it next week.

**Release note**:

```noteworthy user
The upstream community [`vertical-pod-autoscaler` component](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) is now supported for shoot clusters by setting `.spec.kubernetes.verticalPodAutoscaler.enabled=true` (disabled by default). More information can be found [in this document](https://github.com/gardener/gardener/blob/master/docs/usage/shoot_autoscaling.md#vertical-pod-auto-scaling).
```


zanetworker:
Nice! will take a look early next week.

zanetworker:
I have some general comments: 

- Based on [the google documentation](https://github.com/kubernetes/autoscaler/pull/3028/files#diff-db1705b74413022ac43703c7ad20007bR159) its not recommended to run both VPA and VPA on same pods. So if we are to choose for CoreDNS pods as an example, I would prefer to have HPA enabled as with VPA we are always capped by the physical node resources. On the otherhand for smaller cluster with high pod density we could still cap the node resources (by creating more DNS pods on the same node), this however is less likely to happen than the large cluster case.  @ggaurav10 @amshuman-kr can you also emphasize what are the side effects of running VPA and HPA together? 

- Regarding the recommender proportional scaling issue, we can always set a `maximumAllowed` to be on the safe side. For example: 
 
```yaml
spec:
  resourcePolicy:
    containerPolicies:
    - containerName: "kube-proxy"
      minAllowed:
        cpu: "100m"
        memory: "1024Mi"
      maxAllowed:
        cpu: "1200m"
        memory: "3200Mi"
``` 
- Finally, just for reference, there is also an [Open PR](https://github.com/kubernetes/autoscaler/pull/3028/files#diff-db1705b74413022ac43703c7ad20007bR159) that would allow controlling whether the limits are also scaled with VPA or not.

rfranzke:
Thanks, will adapt accordingly.
Regarding the maximum allowed: I don't know why we should set such configuration. If we do then we are back to square one with the same situation as of today: Until a certain limit we can scale, but that's it. The goal of this PR is to get rid of this limitation, i.e., to scale without limitations.

zanetworker:
> Regarding the maximum allowed: I don't know why we should set such configuration

the idea is to avoid unrealistic limits thats might be set by the VPA, so we give room for the VPA to do its job but safeguard against unreaslitic values that might break our clusters.

rfranzke:
Hm, ok, I can't overlook this now. We don't have such upper bound limits for the many VPA objects that we deploy to the seeds, so I would not add them now for the shoots. I'm also not sure how such values could happen/break our clusters. I suggest to follow up with the VPA experts (@wyb1 @ggaurav10 @amshuman-kr) in a separate thread/issue if you think this is a problem/something that should be considered. Then it can be added consistently to allow VPA objects in both seed/shoot.

zanetworker:
> I'm also not sure how such values could happen/break our clusters

- for unrealistic requests, we could block other pods from being scheduled on the node due to resources being reserved. 
- for limits, if they are too large, its like having no limits which again could consume resources if not all the other pods had requests defined. 

> I suggest to follow up with the VPA experts (@wyb1 @ggaurav10 @amshuman-kr) in a separate thread/issue if you think this is a problem/something that should be considered. 

Agreed.

rfranzke:
I've pushed another commit that allows to disable the VPA for seed by leveraging the seed settings. For shooted seeds that enable the VPA in `.spec.kubernetes.verticalPodAutoscaler` the `.spec.settings.verticalPodAutoscaler.enabled` will be `false` when the gardenlet registers the seed.

ggaurav10:
From online documentations, blogs, etc., in general, it doesn't seem to be a good idea to deploy HPA and VPA together.

Regarding setting "sane `maxAllowed`" for critical components such as coredns - 
What happens when those limits are reached - we would see workloads failing, and would eventually raise those limits (as we have done in the past). We would need to do this every time those limits are reached, except for those rare cases when the workload itself is misbehaving. IMHO, we need not set `maxAllowed` for such components. We can revisit this in case we see more number of misbehaving workloads putting unnecessary pressure on coredns.
wdyt?

Edit: vpa for coredns is removed in the 2nd commit. So these points have become moot for now :)

pkg/apis/core/types_shoot.go
```diff
@@ -303,6 +305,30 @@ type ClusterAutoscaler struct {
 	ScanInterval *metav1.Duration
 }
 
+// VerticalPodAutoscaler contains the configuration flags for the Kubernetes vertical pod autoscaler.
+type VerticalPodAutoscaler struct {
```
wyb1, line 309: Should there be a struct for each vpa component (recommender and updater)? This would make it more clear which component is being affected by the setting.

wyb1 (commented):
Nice PR! Overall looks good and I will check it again once the VPA objects are deployed in the shoot since it is currently disabled. 
I don't think we need to deploy the `vpa-exporter` right now since Prometheus does not scrape it for metrics and we won't have a dashboard for this anyway. The `vpa-exporter` will just consume unnecessary resources. Optionally, I could configure Prometheus to scrape the exporter and create a dashboard as a follow up to this PR.

> for unrealistic requests, we could block other pods from being scheduled on the node due to resources being reserved.

Correct, but VPA has been recommending reasonable requests and we should expect this behavior to continue.

> for limits, if they are too large, its like having no limits which again could consume resources if not all the other pods had requests defined.

Since VPA will scale the limit proportionally, we should make sure that the limits are not excessively higher than the requests otherwise like you said it is basically like having no limits. Unfortunately, I'm not sure what concrete effect this would have on our clusters.

rfranzke:
Thanks @wyb1 for your review!

> I don't think we need to deploy the vpa-exporter right now since Prometheus does not scrape it for metrics and we won't have a dashboard for this anyway. The vpa-exporter will just consume unnecessary resources. Optionally, I could configure Prometheus to scrape the exporter and create a dashboard as a follow up to this PR.

OK, both is fine for me. Just tell me what you prefer. If you want to work on a follow-up PR for prometheus then I keep it, otherwise I disable it.

> Since VPA will scale the limit proportionally, we should make sure that the limits are not excessively higher than the requests otherwise like you said it is basically like having no limits. Unfortunately, I'm not sure what concrete effect this would have on our clusters.

Do you think the adaptations I've made are reasonable? As "rule of thumb" I've set the limits to 4*requests for the system components put under VPA.

pkg/apis/core/types_shoot.go
rfranzke: Hm, no strong opinion. I'm fine with both, slightly prefer the current approach, mainly because it doesn't require the end-user to understand the architecture of VPA. He just needs to lookup the meaning of the flag, but he doesn't need to know which of the four components is responsible for executing it.
What do other folks think?

ggaurav10: Same. No strong opinions. Current approach does have an advantage that end-user need not bother which component is responsible for what, as mentioned by Rafael..
As long as the role of each parameter is documented properly, end-user need not care IMO.

wyb1: Then let's keep it as is. The names are enough and it is documented.

wyb1:
> OK, both is fine for me. Just tell me what you prefer. If you want to work on a follow-up PR for prometheus then I keep it, otherwise I disable it.

Let's remove the vpa-exporter from the shoot control planes. The metrics from kube state metrics should be enough to create a dashboard. I'll create a PR for this once this is merged.

> Do you think the adaptations I've made are reasonable? As "rule of thumb" I've set the limits to 4*requests for the system components put under VPA.

The limits look reasonable 👍

rfranzke:
Thanks @wyb1. I've made the changes and removed `vpa-exporter` from the shoot control plane. PTAL.

wyb1 (dismissed):
/lgtm

ggaurav10 (dismissed):

rfranzke:
Thanks @wyb1 and @ggaurav10 for your reviews.
I've rebased to resolve the conflicting files caused by other PRs.

charts/seed-bootstrap/charts/vpa/charts/application/templates/clusterrole-actor.yaml
```diff
@@ -0,0 +1,55 @@
+{{- if or .Values.recommender.enabled .Values.updater.enabled }}
```
ggaurav10, line 1: Somehow missed this change in my last review because in some of the first few round of reviews/commits, this check was not there.
Curious as to why do we need to check if recommender, updater and admission-controller are enabled separately. is it ok to merge them under single umbrella of `vpa.enabled`, or can there be a scenario when we deploy them partially?

rfranzke: @wyb1 suggested to not deploy the vpa-exporter when VPA is enabled for shoots. That's why I also made is configurable for the other components (to be consistent), although, it's currently only used to enable/disable vpa-exproter.

ggaurav10: ok. thanks for explaining. :)

ggaurav10 (dismissed):

pkg/operation/botanist/addons.go
```diff
@@ -316,6 +303,7 @@ func (b *Botanist) generateCoreAddonsChart() (*chartrenderer.RenderedChart, erro
 		global          = map[string]interface{}{
 			"kubernetesVersion": b.Shoot.Info.Spec.Kubernetes.Version,
 			"podNetwork":        b.Shoot.Networks.Pods.String(),
+			"vpaEnabled":        false, // TODO: b.Shoot.WantsVerticalPodAutoscaler, once https://github.com/gardener/gardener-resource-manager/issues/58 is resolved
```
wyb1: I see that the issue is closed. Will this TODO be removed in this PR or later?

rfranzke: Yep, thanks for reminding, will do it in this PR

rfranzke: done, ptal

wyb1 (approved):
/lgtm

PR #2480, opened 2020-06-19
[#2462] Add flag AllowSystemComponents to worker options

### LINKED ISSUE #2462: Specify NodeSelector/Affinities for Shoot core components
/area control-plane
/kind enhancement
/priority normal

I'd like to be able to specify `nodeSelector`s or `affinities` for the core component that Gardener deploys into each seed so that I can control on which worker pools those components run.

**Why is this needed**:
This is needed as some workerpools might have special hardware or might have very much resources which are reserved for workload.

By letting the core components run wherever they are scheduled they might end up on hardware which is very expensive and should only be used by the intended workload.

Allowing to specify `nodeSelector`s or `affinities` would also make it possible to create a dedicated workerpool for core components.

jannickfahlbusch:
/area control-plane
/kind enhancement
/kind api-change
/priority normal

This allows to restrict the core components from being able to be scheduled to every worker pool.

I've added the object `systemComponents` on worker pools which contains a field `allow`. This field defaults to true.

 Each node will receive a label `worker.gardener.cloud/system-components` with the specified value.

The core components received a `nodeSelector` for `worker.gardener.cloud/system-components=true`

Fixes #2462 

Currently it would be possible to restrict the core-components solely to a worker-group which is scaled to zero. I think this makes no sense. What do you think?

**Release note**:
```improvement user
It is now possible to restrict core components from running on a worker pool by specifying `systemComponents.allow: false` in the pool definition.

System components deployed by extensions (Such as typha from calico) need a separate adaptation and are not covered by this change.
```


jannickfahlbusch:
Somehow `make generate` does not generate the DeepCopy functions. Did I forget something?

pkg/apis/core/v1beta1/defaults.go
```diff
@@ -237,6 +237,9 @@ func SetDefaults_Worker(obj *Worker) {
 	if obj.MaxUnavailable == nil {
 		obj.MaxUnavailable = &DefaultWorkerMaxUnavailable
 	}
+	if obj.AllowSystemComponents == nil {
```
vpnachev: Same defaulting needs to be done for v1alpha1 version, as well.

pkg/apis/core/validation/shoot.go
```diff
@@ -1065,16 +1065,22 @@ func ValidateWorkers(workers []core.Worker, fldPath *field.Path) field.ErrorList
...
+		atLeastOnePoolWithCompatibleTaints        = len(workers) == 0
+		atLeastOnePoolWithAllowedSystemComponents = false
 	)
 
 	for i, worker := range workers {
 		if worker.Minimum != 0 && worker.Maximum != 0 {
 			atLeastOneActivePool = true
 		}
 
+		// ToDo: Should this also validate that the pool allowing System components is not scaled to zero?
```
vpnachev: Yes, the maximum must be >=1.
As far as I know, scale-to-zero is not implemented for all providers [1], so from my point of view the check should for `minimum>=1`.

[1] https://github.com/gardener/autoscaler/issues/27

pkg/apis/core/validation/shoot_test.go
```diff
@@ -15,6 +15,7 @@
 package validation_test
 
 import (
+	"github.com/gardener/gardener/pkg/apis/core/v1beta1"
```
vpnachev: move this import in the second group of packages

vpnachev (changes requested):
Thanks, looks very nice.

vpnachev:
> Somehow `make generate` does not generate the DeepCopy functions. Did I forget something?

If you have not, run `make install-requirements` and then re-run `make generate`.

pkg/apis/core/validation/shoot.go
jannickfahlbusch: Done

jannickfahlbusch:
Figured it out:

Since the repo is a Go module I assumed that I can place the repo outside of my `GOPATH`. When run outside of the `GOPATH`, `make generate` finishes without any error but does not generate additional sources.

jannickfahlbusch:
At the moment the following components would respect the setting:
* CoreDNS
* Metrics server
* Blackbox exporter
* VPN

Did I miss anything?

vpnachev:
Yeah, code generation require the repo to be checked out under $GOPATH even though we are using go modules, see https://github.com/kubernetes/kubernetes/issues/86753 for more info.

>Did I miss anything?

I think no. You would like to open a separate PR to https://github.com/gardener/gardener-extension-networking-calico/, but the change there will be incompatible with the currently released gardener versions, so I would wait or add some additional logic to set the `nodeSelector` only if gardener version support it.

/reviewed ok-to-test

timebertt:
Thanks @jannickfahlbusch for contributing this feature, really appreciate it!


> At the moment the following components would respect the setting:
> 
> * CoreDNS
> * Metrics server
> * Blackbox exporter
> * VPN
> 
> Did I miss anything?

You can probably add it to:
- nginx-ingress-controller
- kubernetes-dashboard

/status author-action

timebertt:
/needs review

I will try to take a look today :)
/assign

example/90-shoot.yaml
```diff
@@ -55,6 +55,7 @@ spec:
     # kubeletDataVolumeName: kubelet-dir
     # providerConfig:
     #   <some-provider-specific-worker-config>
+    # allowSystemComponents: true
```
rfranzke: API-wise, such a structure would be more prepared in case we want to introduce further configuration settings for the system components in the future:

```suggestion
    # systemComponents:
    #   allowed: true
```

WDYT?

pkg/apis/core/validation/shoot.go
```diff
@@ -1065,14 +1065,19 @@ func ValidateWorkers(workers []core.Worker, fldPath *field.Path) field.ErrorList
...
+		atLeastOneActivePool                      = false
+		atLeastOnePoolWithCompatibleTaints        = len(workers) == 0
+		atLeastOnePoolWithAllowedSystemComponents = false
 	)
 
 	for i, worker := range workers {
 		if worker.Minimum != 0 && worker.Maximum != 0 {
 			atLeastOneActivePool = true
+
+			if *worker.AllowSystemComponents {
```
```suggestion
			if worker.AllowSystemComponents != nil && *worker.AllowSystemComponents {
```

pkg/apis/extensions/v1alpha1/types_worker.go
```diff
@@ -128,6 +128,8 @@ type WorkerPool struct {
 	// Zones contains information about availability zones for this worker pool.
 	// +optional
 	Zones []string `json:"zones,omitempty"`
+	// AllowSystemComponents determines whether the pool should be allowed to host system components or not (defaults to true)
```
rfranzke:
```suggestion
	// AllowSystemComponents determines whether the pool should be allowed to host system components or not (defaults to true)
	// +optional
```

pkg/apis/extensions/v1alpha1/types_worker.go
```diff
@@ -128,6 +128,8 @@ type WorkerPool struct {
...
+	AllowSystemComponents *bool
```
rfranzke:
```suggestion
	AllowSystemComponents *bool `json:"allowSystemComponents,omitempty"`
```

Though, same suggestion like above.

pkg/operation/botanist/worker.go
```diff
@@ -91,6 +92,8 @@ func (b *Botanist) DeployWorker(ctx context.Context) error {
 			workerPool.Labels["node.kubernetes.io/role"] = "node"
 		}
 
+		workerPool.Labels[v1beta1constants.LabelWorkerPoolAllowSystemComponents] = strconv.FormatBool(*workerPool.AllowSystemComponents)
```
rfranzke: nil pointer check for missing

rfranzke (changes requested):
Nice PR, thank you!
Can you also enhance https://github.com/gardener/gardener/blob/master/docs/extensions/worker.md with a comment about the new field that is being introduced?

charts/shoot-core/components/charts/coredns/templates/coredns-deployment.yaml
```diff
@@ -39,6 +39,8 @@ spec:
                       values:
                         - kube-dns
       priorityClassName: system-cluster-critical
+      nodeSelector:
+        system-components.gardener.cloud/allowed: true
```
timuthy:
```suggestion
        worker.gardener.cloud/system-components: true
```
"allowed" is functionally more related to taints/tolerations, i.e. if you create a system component pod w/o a label selector, nodes with `system-components.gardener.cloud/allowed: false` will also be considered.

pkg/apis/core/v1alpha1/constants/types_constants.go
```diff
@@ -209,6 +209,9 @@ const (
 	// "core.gardener.cloud/apiserver-exposure" to indicate that it's responsible for apiserver exposure (via SNI).
 	LabelAPIServerExposureGardenerManaged = "gardener-managed"
 
+	// LabelWorkerPoolAllowSystemComponents is a constant that indicates whether the worker pool should host system components
+	LabelWorkerPoolAllowSystemComponents = "system-components.gardener.cloud/allowed"
```
timuthy:
```suggestion
	LabelWorkerPoolAllowSystemComponents = "worker.gardener.cloud/system-components"
```

timuthy (changes requested):

pkg/apis/core/v1alpha1/defaults_test.go
```diff
@@ -338,4 +338,20 @@ var _ = Describe("Defaults", func() {
...
+		BeforeEach(func() {
+			obj = &Worker{}
+		})
+
+		It("should set the allowSystemComponents field", func() {
+			obj.AllowSystemComponents = nil
+
+			SetDefaults_Worker(obj)
+
+			Expect(obj.AllowSystemComponents).To(Equal(&DefaultWorkerAllowSystemComponents))
```
timebertt:
```suggestion
			Expect(obj.AllowSystemComponents).To(PointTo(BeTrue()))
```
Same for the other API version.

pkg/apis/core/validation/shoot.go
```diff
@@ -1065,14 +1065,19 @@ func ValidateWorkers(workers []core.Worker, fldPath *field.Path) field.ErrorList
...
+				atLeastOnePoolWithAllowedSystemComponents = true
```
timebertt: Hmm, should we also check, if there are no or only compatible taints on the worker pool? Otherwise we can't be sure, that the system components are indeed able to run on a worker pool.
So maybe you can just move this check down to line 1091.

pkg/operation/botanist/worker.go
```diff
@@ -133,6 +136,7 @@ func (b *Botanist) DeployWorker(ctx context.Context) error {
 			DataVolumes:           dataVolumes,
 			KubeletDataVolumeName: workerPool.KubeletDataVolumeName,
 			Zones:                 workerPool.Zones,
+			AllowSystemComponents: workerPool.AllowSystemComponents,
```
timebertt: I think, we don't need an additional field on the Worker. The node labels are already added via `Worker.spec.pools[].labels` (as you already adapted in line 95). That way, the worker controller doesn't need to act upon this setting and doesn't even need to know about it...

timebertt (commented):
I don't have to add much to the previous reviews, only some minor things.
Otherwise, looks quite good already!

example/90-shoot.yaml
jannickfahlbusch: Sound reasonable to me, I'll change it

pkg/apis/extensions/v1alpha1/types_worker.go
jannickfahlbusch: Removed due to the comment from @tim-ebert below: https://github.com/gardener/gardener/pull/2480#discussion_r443415041

pkg/apis/extensions/v1alpha1/types_worker.go
jannickfahlbusch: Removed due to the comment from @tim-ebert below: https://github.com/gardener/gardener/pull/2480#discussion_r443415041

jannickfahlbusch:
> You can probably add it to:
>
> * nginx-ingress-controller
> * kubernetes-dashboard

I'd like to be able to do this. However I wouldn't consider the plugins as system components. 

How do you see that?

example/90-shoot.yaml
jannickfahlbusch: Done

pkg/apis/core/validation/shoot.go
jannickfahlbusch: Done

pkg/operation/botanist/worker.go
jannickfahlbusch: Done

charts/shoot-core/components/charts/coredns/templates/coredns-deployment.yaml
jannickfahlbusch: Done

pkg/apis/core/validation/shoot.go
jannickfahlbusch: Done

pkg/operation/botanist/worker.go
jannickfahlbusch: Removed

pkg/apis/core/v1alpha1/defaults_test.go
jannickfahlbusch: Changed

timebertt:
> > You can probably add it to:
> > 
> > * nginx-ingress-controller
> > * kubernetes-dashboard
> 
> I'd like to be able to do this. However I wouldn't consider the plugins as system components.
> 
> How do you see that?

Yes, I somewhat agree, but I think, it would be fine to apply the same mechanisms to the addons (if we want via a separate label/field in the shoot spec), but I don't have a strong opinion here.
@gardener/gardener-maintainers any opinions?

pkg/apis/core/types_shoot.go
```diff
@@ -708,6 +708,12 @@ type Worker struct {
 	Zones []string
 }
 
+// SystemComponents contains configuration for system components related to this worker pool
```
timebertt:
```suggestion
// WorkerSystemComponents contains configuration for system components related to this worker pool
```

rfranzke:
For the sake of simplicity, I would consider them as "system components" as well (although, they aren't, of course). The rationale behind is mainly that our path forward with the addons is not entirely clear, and adding a new field to the API now that we have to deprecate in the future again is unnecessary work.

jannickfahlbusch:
Sounds good, I'll add the node selectors there as well :)

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -858,9 +858,16 @@ type Worker struct {
...
 	// +optional
 	Zones []string `json:"zones,omitempty" protobuf:"bytes,17,rep,name=zones"`
-	// AllowSystemComponents determines whether the pool should be allowed to host system components or not (defaults to true)
+	// SystemComponents contains configuration for system components related to this worker pool
 	// +optional
-	AllowSystemComponents *bool `json:"allowSystemComponents,omitempty" protobuf:"bytes,18,opt,name=allowSystemComponents"`
+	SystemComponents *WorkerSystemComponents `json:"systemComponents,omitempty" protobuf:"bytes,18,opt,name=systemComponents"`
+}
+
+// SystemComponents contains configuration for system components related to this worker pool
```
timebertt:
```suggestion
// WorkerSystemComponents contains configuration for system components related to this worker pool
```

rfranzke (commented):
Nice PR! :) No further comments from my side as of now...
Could you squash the commits?

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -855,9 +855,16 @@ type Worker struct {
...
 	// +optional
 	Zones []string `json:"zones,omitempty" protobuf:"bytes,17,rep,name=zones"`
-	// AllowSystemComponents determines whether the pool should be allowed to host system components or not (defaults to true)
+	// SystemComponents contains configuration for system components related to this worker pool
 	// +optional
-	AllowSystemComponents *bool `json:"allowSystemComponents,omitempty" protobuf:"bytes,18,opt,name=allowSystemComponents"`
+	SystemComponents *WorkerSystemComponents `json:"systemComponents,omitempty" protobuf:"bytes,18,opt,name=systemComponents"`
+}
+
+// SystemComponents contains configuration for system components related to this worker pool
```
timebertt:
```suggestion
// WorkerSystemComponents contains configuration for system components related to this worker pool
```

pkg/apis/core/v1alpha1/defaults.go
```diff
@@ -237,8 +237,11 @@ func SetDefaults_Worker(obj *Worker) {
 	if obj.MaxUnavailable == nil {
 		obj.MaxUnavailable = &DefaultWorkerMaxUnavailable
 	}
-	if obj.AllowSystemComponents == nil {
-		obj.AllowSystemComponents = &DefaultWorkerAllowSystemComponents
+	if obj.SystemComponents == nil {
+		obj.SystemComponents = &WorkerSystemComponents{}
+	}
+	if obj.SystemComponents.Allow == nil {
+		obj.SystemComponents.Allow = &DefaultWorkerSystemComponentsAllow
```
timebertt: Hmm, I'm wondering if we should actually default this value in the Shoot object or rather leave it empty if unset and use the `DefaultWorkerSystemComponentsAllow` constant in the shoot controller as a "behavioural" default (set the node label to `true` if the flag in the ShootSpec is not set).
But I am not sure, if this would still be conformant with k8s API conventions.

pkg/apis/core/validation/shoot.go
```diff
@@ -1101,6 +1097,12 @@ func ValidateWorkers(workers []core.Worker, fldPath *field.Path) field.ErrorList
 				atLeastOnePoolWithCompatibleTaints = true
 			}
 		}
+
+		if atLeastOneActivePool && atLeastOnePoolWithCompatibleTaints {
+			if worker.SystemComponents != nil && worker.SystemComponents.Allow != nil && *worker.SystemComponents.Allow {
+				atLeastOnePoolWithAllowedSystemComponents = true
+			}
+		}
```
timebertt: Hmmm, I think this will allow Shoots with a worker section like this, although we will not be able to schedule system components on either worker pools:
```
workers:
- name: untainted-but-without-system-components
  systemComponents:
    allow: false
- name: tainted-but-system-components-allowed
  taints:
  - key: Foo
     effect: NoSchedule
```
Correct?
I think, we should check here, if the **current** `worker` has no or only compatible taints `systemComponents.allowed=true`.

jannickfahlbusch: Done

pkg/apis/core/v1alpha1/defaults.go
jannickfahlbusch: I'm not sure. 

I'd like to not have magic where defaulting is happening but not visible on the Shoot CR.
If you prefer it the other way around I can change the behavior

pkg/apis/core/validation/shoot_test.go
```diff
@@ -134,6 +143,7 @@ var _ = Describe("Shoot Validation Tests", func() {
 				Maximum:        0,
 				MaxSurge:       &maxSurge,
 				MaxUnavailable: &maxUnavailable,
+				//SystemComponents: systemComponents,
```
danielfoehrKn:
```suggestion
```

danielfoehrKn (commented):
/lgtm

pkg/apis/core/validation/shoot_test.go
jannickfahlbusch: Done

pkg/apis/core/v1alpha1/defaults.go
vpnachev: The magic can happen with a wrapping function, something like https://github.com/gardener/gardener/blob/7afb12735b14f49144b3a0558eca2b1ceeaba6b0/pkg/apis/core/v1beta1/helper/helper.go#L592-L595

for example:
```go
func SystemComponentsAllowed(worker *gardencorev1beta1.Worker) bool {
	return worker.SystemComponents == nil || worker.SystemComponents.Allow == nil || *worker.SystemComponents.Allow
}
```

vpnachev (dismissed):
/lgtm

pkg/apis/core/v1alpha1/defaults.go
jannickfahlbusch: I've changed the behavior in https://github.com/gardener/gardener/pull/2480/commits/8c17a3de043c1e988f916f4ef0e6b5e4993228af.
Please take a look. If it looks good, I'll squash everything together and submit this for a final review

pkg/apis/core/v1alpha1/constants/types_constants.go
```diff
@@ -209,6 +209,9 @@ const (
 	// "core.gardener.cloud/apiserver-exposure" to indicate that it's responsible for apiserver exposure (via SNI).
 	LabelAPIServerExposureGardenerManaged = "gardener-managed"
 
+	// LabelWorkerPoolSystemComponentsAllow is a constant that indicates whether the worker pool should host system components
+	LabelWorkerPoolSystemComponentsAllow = "worker.gardener.cloud/system-components"
```
timuthy:
```suggestion
	LabelWorkerPoolSystemComponents = "worker.gardener.cloud/system-components"
```
Please also adjust doc string.

pkg/apis/core/v1beta1/constants/types_constants.go
```diff
@@ -282,6 +282,9 @@ const (
 	// LabelWorkerPool is a deprecated constant for a label that indicates the worker pool the node belongs to
 	LabelWorkerPoolDeprecated = "worker.garden.sapcloud.io/group"
 
+	// LabelWorkerPoolSystemComponentsAllow is a constant that indicates whether the worker pool should host system components
+	LabelWorkerPoolSystemComponentsAllow = "worker.gardener.cloud/system-components"
```
timuthy:
```suggestion
	LabelWorkerPoolSystemComponents = "worker.gardener.cloud/system-components"
```
Please also adjust doc string.

pkg/apis/core/helper/helpers.go
```diff
@@ -285,3 +285,8 @@ func FindVersionsWithSameMajorMinor(versions []core.ExpirableVersion, version se
 func HibernationIsEnabled(shoot *core.Shoot) bool {
 	return shoot.Spec.Hibernation != nil && shoot.Spec.Hibernation.Enabled != nil && *shoot.Spec.Hibernation.Enabled
 }
+
+// SystemComponentsAllowed cheks if the given worker allows system components to be scheduled onto it
```
rfranzke:
```suggestion
// SystemComponentsAllowed checks if the given worker allows system components to be scheduled onto it
```

pkg/apis/core/types_shoot.go
```diff
@@ -706,6 +708,12 @@ type Worker struct {
 	Zones []string
 }
 
+// WorkerSystemComponents contains configuration for system components related to this worker pool
+type WorkerSystemComponents struct {
+	// Allow determines whether the pool should be allowed to host system components or not (defaults to true)
+	Allow *bool
```
rfranzke: Why is this a pointer/an optional field? `WorkerSystemComponents` itself is already optional, so the `Allow` could be a regular field inside it. cc @gardener/gardener-maintainers wdyt?

pkg/apis/core/v1alpha1/helper/helper.go
```diff
@@ -587,6 +587,11 @@ func ShootUsesUnmanagedDNS(shoot *gardencorev1alpha1.Shoot) bool {
 	return shoot.Spec.DNS != nil && len(shoot.Spec.DNS.Providers) > 0 && shoot.Spec.DNS.Providers[0].Type != nil && *shoot.Spec.DNS.Providers[0].Type == "unmanaged"
 }
 
+// SystemComponentsAllowed cheks if the given worker allows system components to be scheduled onto it
```
rfranzke:
```suggestion
// SystemComponentsAllowed checks if the given worker allows system components to be scheduled onto it
```

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -674,6 +674,11 @@ func GetLatestQualifyingShootMachineImage(image gardencorev1beta1.MachineImage,
 	return true, &gardencorev1beta1.ShootMachineImage{Name: image.Name, Version: &latestImageVersion.Version}, nil
 }
 
+// SystemComponentsAllowed cheks if the given worker allows system components to be scheduled onto it
```
rfranzke:
```suggestion
// SystemComponentsAllowed checks if the given worker allows system components to be scheduled onto it
```

pkg/apis/core/v1alpha1/defaults.go
rfranzke: @tim-ebert why do you suggest this? I don't think we have been consistent here in the past, but generally it's better to be explicit if we already have a dedicated field in the API. Yes, we haven't streamlined this yet for other fields, but for new fields we tried to default it, or? What do other folks think? @timuthy @mvladev @ialidzhikov @danielfoehrKn

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -948,6 +958,8 @@ var (
 	DefaultWorkerMaxSurge = intstr.FromInt(1)
 	// DefaultWorkerMaxUnavailable is the default value for Worker MaxUnavailable.
 	DefaultWorkerMaxUnavailable = intstr.FromInt(0)
+	// DefaultWorkerSystemComponentsAllow is the default value for Worker AllowSystemComponents
+	DefaultWorkerSystemComponentsAllow = true
```
timuthy, line 962: Still needed?

timuthy (changes requested):
Just a few nits. Looks good otherwise.

pkg/apis/core/types_shoot.go
timuthy: +1
No necessity to use a pointer.

timuthy:
@jannickfahlbusch concerning the defaulting discussion:
We @gardener/gardener-maintainers had an internal discussion and agreed that we'd prefer an explicit behavior, also because the [K8s API conventions encourage defaulting](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#defaulting).
Can you add the field defaulting functionality again, please? Sorry for the back and forth. We'll also document this best practice to make contributors' lives easier in the future.

pkg/apis/core/helper/helpers_test.go
```diff
@@ -15,10 +15,9 @@
 package helper_test
 
 import (
+	"github.com/Masterminds/semver"
```
vpnachev: this package should stay in the second group of imports

pkg/apis/core/types_shoot.go
jannickfahlbusch: Removed

pkg/apis/core/v1alpha1/defaults.go
jannickfahlbusch: I've readded the defaulting with https://github.com/gardener/gardener/pull/2480/commits/c159d9da2f77cbaf1a321b71eeeb546a5f76730d but left the helper functions in as I made use of them in the botanist

jannickfahlbusch:
Resolved all requested changes and squashed the commits together.

Please take a look again :)

rfranzke (dismissed):
/lgtm
Thank you very much!
I think it makes sense to add to the release note that system components deployed by extensions (e.g., calico deploys typha and other components) need a separate adaptation and are not covered by this PR.

jannickfahlbusch:
I've adapted the release note.

Do you have a list of components that would need adoption to the feature?
So far I'm only aware of [gardener-extension-networking-calico](https://github.com/gardener/gardener-extension-networking-calico)

rfranzke:
Thanks @jannickfahlbusch, great. If I remember correctly then the calico extension is the only one that deploys non-`DaemonSet` components into the shoot. @zanetworker @DockToFuture what about cilium?

Now that we're talking about it: It probably still makes sense to enhance the https://github.com/gardener/gardener/blob/master/docs/extensions/conventions.md#how-to-create-resources-in-the-shoot-cluster documentation so that extension developers are aware of the fact that they should respect the node labels when deploying components into the shoot.

jannickfahlbusch:
I've added a sentence for the extension documentation and also opened a change for calico: https://github.com/gardener/gardener-extension-networking-calico/pull/39

zanetworker:
@rfranzke Cilium does as well namely: 
- cilium-operator
- hubble

rfranzke (dismissed):

timuthy (dismissed):
/lgtm

vpnachev (dismissed):
/lgtm

jannickfahlbusch:
I had to push again as I forgot to format a test

vpnachev:
The CI validation is failing with:
```
> Check
Executing golangci-lint
Executing go vet
Executing gofmt
Unformatted files detected:
pkg/apis/core/validation/shoot_test.go
```

@jannickfahlbusch  You could run `make check` locally to verify the files are will formatted. 

PS. I was too slow :smile:

vpnachev (approved):
/lgtm

vpnachev:
/hold

There are some issues and resource manager cannot apply the deployments
```
"Errors occurred during applying: [error during apply of object \"apps/v1/Deployment/kube-system/vpn-shoot\": unable to convert unstructured object to apps/v1, Kind=Deployment: unrecognized type: string error during apply of object \"apps/v1/Deployment/kube-system/coredns\": unable to convert unstructured object to apps/v1, Kind=Deployment: unrecognized type: string error during apply of object \"apps/v1/Deployment/kube-system/metrics-server\": unable to convert unstructured object to apps/v1, Kind=Deployment: unrecognized type: string]",
```

jannickfahlbusch:
> resource manager cannot apply the deployments

How can I debug this to submit a fix?

Edit: I suspect that setting `worker.gardener.cloud/system-components: true` is an issue as `true` evaluates to a Boolean true in YAML.

charts/shoot-addons/charts/kubernetes-dashboard/templates/deployment-dashboard.yaml
```diff
@@ -25,6 +25,8 @@ spec:
         origin: gardener
         k8s-app: kubernetes-dashboard
     spec:
+      nodeSelector:
+        worker.gardener.cloud/system-components: true
```
vpnachev: The value of the selector must be escaped 
```suggestion
        worker.gardener.cloud/system-components: "true"
```

If you try to apply this manifest, on minukube for example or any other k8s cluster,
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: vpn-shoot
  name: vpn-shoot
  namespace: kube-system
spec:
  replicas: 1
  revisionHistoryLimit: 0
  selector:
    matchLabels:
      app: vpn-shoot
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      labels:
        app: vpn-shoot
    spec:
      containers:
      - image: eu.gcr.io/gardener-project/gardener/vpn-shoot:0.19.0
        imagePullPolicy: IfNotPresent
        name: vpn-shoot
      nodeSelector:
        worker.gardener.cloud/system-components: true
      restartPolicy: Always

```

you will get this error
```
Error from server (BadRequest): error when creating "vpn.yaml": Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.NodeSelector: ReadString: expects " or n, but found t, error found in #10 byte of ...|ponents":true},"rest|..., bigger context ...|ctor":{"worker.gardener.cloud/system-components":true},"restartPolicy":"Always"}}}}

```

vpnachev (changes requested):
The value of the label selector needs to be escaped.

charts/shoot-addons/charts/kubernetes-dashboard/templates/deployment-dashboard.yaml
jannickfahlbusch: Fixed

vpnachev (approved):
/lgtm
/reviewed ok-to-test

rfranzke (approved):

timebertt (approved):
Hmm, I was too late... 😄 
But nevertheless, PR looks great!
Thanks a lot for your efforts!

PR #2483, opened 2020-06-20
use legacy iptables for kube-proxy ~1.16.11

vpnachev:

/area networking
/kind enhancement
/priority normal

use legacy iptables for kube-proxy ~1.16.11

Mitigates https://github.com/kubernetes/kubernetes/issues/92275

**Release note**:

```improvement operator
`kube-proxy` for k8s versions ~1.16.11 now explicitly use the legacy iptables.
```


ialidzhikov:
Hmm, I am not sure that we have to workaround the hyperkube kube-proxy issue for v1.16.11 for now. Even we have to (hyperkube kube-proxy 1.16.11 stays broken), then we can use the non-hyperkube kube-proxy image.

vpnachev:
Makes sense. Also, there is some activity to fix the issue upstream, so let's wait some time and then we can check again whether we need to do something on our side.

/hold

rfranzke:
There is already a PR https://github.com/kubernetes/kubernetes/pull/92354 that should fix this problem. Do you want still want to have this workaround or should we simply not support the v1.16.11?

vpnachev:
Let's keep this PR open till 1.16.12 is released and if the regression is fixed, we will simply not enable 1.16.11.

vpnachev:
/close

I will re-open it if it is needed.

PR #2484, opened 2020-06-22
Update konnectivity tunnel to v0.0.10

zanetworker:

/area networking
/kind enhancement
/priority normal

This PR updates the konnectivity-tunnel to v0.0.10, additionally missing logic to handle correct deletion of seed konnectivity-tunnel secrets is added in this PR. 

**Release note**:

```improvement operator
Konnectivity tunnel is now updated to v0.0.10. 
```

### DIFF
--- a/pkg/operation/botanist/secrets.go
+++ b/pkg/operation/botanist/secrets.go
@@ -811,13 +811,18 @@ func (b *Botanist) DeploySecrets(ctx context.Context) error {
 	}
 
 	if b.Shoot.KonnectivityTunnelEnabled {
-		err = b.cleanupVPNSecrets(ctx)
-	} else {
-		err = b.deployOpenVPNTLSAuthSecret(ctx, existingSecretsMap)
-	}
+		if err := b.cleanupTunnelSecrets(ctx, "vpn-seed", "vpn-seed-tlsauth", "vpn-shoot"); err != nil {
+			return err
+		}
 
-	if err != nil {
-		return err
+	} else {
+		if err = b.cleanupTunnelSecrets(ctx, common.KonnectivityServerKubeconfig, common.KonnectivityServerCertName); err != nil {
+			return err
+		}
+
+		if err = b.deployOpenVPNTLSAuthSecret(ctx, existingSecretsMap); err != nil {
+			return err
+		}
 	}
 
 	wantedSecretsList, err := b.generateWantedSecrets(basicAuthAPIServer, staticToken, certificateAuthorities)
@@ -1016,9 +1021,8 @@ func (b *Botanist) generateBasicAuthAPIServer(ctx context.Context, existingSecre
 }
 
 func (b *Botanist) generateStaticToken(ctx context.Context, existingSecretsMap map[string]*corev1.Secret) (*secrets.StaticToken, error) {
-	staticTokenConfig := &secrets.StaticTokenSecretConfig{
-		Name: common.StaticTokenSecretName,
-		Tokens: map[string]secrets.TokenConfig{
+	var (
+		staticTokenConfigTokens = map[string]secrets.TokenConfig{
 			common.KubecfgUsername: {
 				Username: common.KubecfgUsername,
 				UserID:   common.KubecfgUsername,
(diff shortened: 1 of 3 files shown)


pkg/utils/secrets/static_token.go
```diff
@@ -114,6 +114,26 @@ func (s *StaticTokenSecretConfig) AppendStaticToken(staticToken *StaticToken) (*
...
+	}
+
+	// remove outdated tokens from the list of tokens
+	for _, outdatedTokenKey := range outdatedTokenKeys {
+		outdatedTokenIndex := staticTokenMap[outdatedTokenKey]
+
+		staticToken.Tokens[outdatedTokenIndex] = staticToken.Tokens[len(staticToken.Tokens)-1]
+		staticToken.Tokens = staticToken.Tokens[:len(staticToken.Tokens)-1]
+
+		return staticToken
```
rfranzke: Why you return here?

pkg/utils/secrets/static_token.go
```diff
@@ -114,6 +114,26 @@ func (s *StaticTokenSecretConfig) AppendStaticToken(staticToken *StaticToken) (*
 	return staticToken, nil
 }
 
+// RemoveFromStaticToken removes outdated tokens from the existing static tokens list.
+func RemoveFromStaticToken(staticToken *StaticToken, outdatedTokenKeys []string) *StaticToken {
```
rfranzke:
```suggestion
func (s *StaticToken) RemoveTokens(outdatedTokenUsernames ...string) *StaticToken {
```

rfranzke: Please try to not make the assumption that a token key is always the username. This might not necessarily be the case.

pkg/operation/botanist/secrets.go
```diff
@@ -1215,10 +1227,9 @@ func (b *Botanist) SyncShootCredentialsToGarden(ctx context.Context) error {
 	return nil
 }
 
-func (b *Botanist) cleanupVPNSecrets(ctx context.Context) error {
+func (b *Botanist) clueanupTunnelSecrets(ctx context.Context, secretsToDelete ...string) error {
```
rfranzke:
```suggestion
func (b *Botanist) cleanupTunnelSecrets(ctx context.Context, secretNames ...string) error {
```

pkg/utils/secrets/static_token.go
zanetworker: I saw that we have userID and userName always set alike so used the name, anyways I changed it.

rfranzke: Thanks. FYI: I've ran into problems as part of https://github.com/gardener/gardener/pull/2478 and needed to refactor this section because this assumption broke in my case.

zanetworker: so would the userID be enough from you PoV?

rfranzke: I think the username or userid should be the "id" of such an entry, yes.

ialidzhikov:
Could you please check `make check`?

pkg/utils/secrets/static_token.go
```diff
@@ -95,23 +95,40 @@ func (s *StaticTokenSecretConfig) GenerateStaticToken() (*StaticToken, error) {
 	}, nil
 }
 
-// GenerateStaticToken computes a random token of length 64.
-func (s *StaticTokenSecretConfig) AppendStaticToken(staticToken *StaticToken) (*StaticToken, error) {
+// AppendTokens appends required tokens to the existing static tokens list.
+func (b *StaticToken) AppendTokens(s *StaticTokenSecretConfig) (*StaticToken, error) {
```
rfranzke: As you are not passing a `StaticToken` but a `StaticTokenSecretConfig` I would rename it:

```suggestion
func (b *StaticToken) GenerateAndAppend(s *StaticTokenSecretConfig) (*StaticToken, error) {
```

pkg/utils/secrets/static_token.go
```diff
@@ -95,23 +95,40 @@ func (s *StaticTokenSecretConfig) GenerateStaticToken() (*StaticToken, error) {
...
 			Token:    token,
 		})
 	}
 
-	return staticToken, nil
+	return b, nil
+}
+
+// RemoveTokens removes outdated tokens from the existing static tokens list.
+func (b *StaticToken) RemoveTokens(outdatedTokenKeys []string) {
```
rfranzke:
```suggestion
func (b *StaticToken) RemoveTokens(userIDs ...string) {
```

pkg/utils/secrets/static_token.go
```diff
@@ -95,23 +95,40 @@ func (s *StaticTokenSecretConfig) GenerateStaticToken() (*StaticToken, error) {
...
+		staticTokenMap[token.UserID] = tokenIndex
+	}
+
+	// remove outdated tokens from the list of tokens
+	for _, outdatedTokenKey := range outdatedTokenKeys {
+		outdatedTokenIndex := staticTokenMap[outdatedTokenKey]
+
+		b.Tokens[outdatedTokenIndex] = b.Tokens[len(b.Tokens)-1]
+		b.Tokens = b.Tokens[:len(b.Tokens)-1]
+	}
```
rfranzke:
```suggestion
  userIDsSet := sets.NewString(userIDs...)

	for i := len(b.Tokens) - 1; i >= 0; i-- {
			if userIDsSet.Has(b.Token[i].UserID) {
				b.Tokens = append(b.Tokens[:i], b.Tokens[i+1:]...)
			}
	}
```

rfranzke (approved):
/lgtm

PR #2500, opened 2020-06-24
Set effective version when building images

timebertt:

/area dev-productivity
/kind enhancement
/priority normal

With this PR, docker images built by `make docker-images` are tagged and build with a version including the commit hash (exactly like the images built by the pipeline).
This allows developers to build prerelease images with a correct version inside to evaluate a new release. E.g. the gardenlet uses the `pkg/version.gitVersion` to report its own version and to the deploy the same gardenlet image version into shooted Seeds.

Fixes #

**Release note**:

```improvement developer
Docker images built by `make docker-images` are now tagged and build with the commit hash appended to the version.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,11 +18,15 @@ CONROLLER_MANAGER_IMAGE_REPOSITORY := $(REGISTRY)/controller-manager
 SCHEDULER_IMAGE_REPOSITORY         := $(REGISTRY)/scheduler
 SEED_ADMISSION_IMAGE_REPOSITORY    := $(REGISTRY)/seed-admission-controller
 GARDENLET_IMAGE_REPOSITORY         := $(REGISTRY)/gardenlet
-IMAGE_TAG                          := $(shell cat VERSION)
-PUSH_LATEST_TAG                    := true
-
+PUSH_LATEST_TAG                    := false
+VERSION                            := $(shell cat VERSION)
+EFFECTIVE_VERSION                  := $(VERSION)-$(shell git rev-parse HEAD)
 REPO_ROOT                          := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 LOCAL_GARDEN_LABEL                 := local-garden
+
+ifneq ($(strip $(shell git status --porcelain 2>/dev/null)),)
+	EFFECTIVE_VERSION := $(EFFECTIVE_VERSION)-dirty
+endif
 
 #########################################
 # Rules for local development scenarios #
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -18,8 +18,9 @@ CONROLLER_MANAGER_IMAGE_REPOSITORY := $(REGISTRY)/controller-manager
 SCHEDULER_IMAGE_REPOSITORY         := $(REGISTRY)/scheduler
 SEED_ADMISSION_IMAGE_REPOSITORY    := $(REGISTRY)/seed-admission-controller
 GARDENLET_IMAGE_REPOSITORY         := $(REGISTRY)/gardenlet
-IMAGE_TAG                          := $(shell cat VERSION)
-PUSH_LATEST_TAG                    := true
+PUSH_LATEST_TAG                    := false
+VERSION                            := $(shell cat VERSION)
+EFFECTIVE_VERSION                  := $(VERSION)-$(shell git rev-parse HEAD)
```
timuthy, line 23: I'm thinking about whether it makes sense checking for a dirty git state and assign it `$(VERSION)-$(shell git rev-parse HEAD)-dirty`, for example. We should avoid that someone accidentally overwrites an already pushed image that is at the moment being tested. WDYT?

vpnachev (commented):
Codewise looks fine.
I have some concerns about this change. 
`git rev-parse HEAD` is the commit hash of the last commit which might be something already merged in the master branch, but not your local changes. With this said, before building the image, the changes needs to be committed to get their own commit hash. For example, if I am deploying from master and you build and push local image based on same commit from master, I will use your local changes instead of the current master.

timebertt:
Thanks for your comments, that's a valid concern 👍 
I updated the Makefile to append `-dirty` to the `EFFECTIVE_VERSION`, if someone is building from a dirty tree.

hack/get-git-dirty-postfix.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
```
rfranzke: Why do we need a dedicated script for this while

timebertt: Hmm, I'm no expert in Makefiles. I didn't work, when I included it in the Makefile.

timebertt:
Found a way to calculate the `-dirty` postfix in the `Makefile`.
@timuthy @rfranzke @vpnachev PTAL :)

rfranzke (approved):
/lgtm

timuthy (approved):
Nice, thanks for implementing the "dirty" use-case 😄 
/lgtm

PR #2509, opened 2020-06-25
Use DirectClient when removing finalizers

timebertt:

/area quality
/kind bug
/priority normal

When removing finalizers, we might run into the same race condition like for other updates that happen concurrently, that the cache is not synced fast enough.
With this PR we always use the `DirectClient` for ensuring, that the finalizer is removed properly.

Fixes #

**Release note**:

```improvement operator
NONE
```


rfranzke (dismissed):
/lgtm

timebertt:
Fixed failing tests.

rfranzke (approved):

danielfoehrKn (approved):

PR #2518, opened 2020-06-26
Fix replica computation for kube-controller-manager

rfranzke:

/area control-plane
/kind technical-debt
/priority normal

This PR fixes the computation of the replica count for KCM which is under control by DWD. Earlier, the logic had the implicit assumption that DWD will scale up KCM for newly created shoots. However, this is a dependency which we don't want to have.

/invite @dguendisch @schrodit @ggaurav10 @amshuman-kr 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -21,7 +21,6 @@ import (
 	"path/filepath"
 	"time"
 
-	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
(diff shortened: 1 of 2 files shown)


dguendisch (dismissed):
Thank you!

pkg/operation/botanist/controlplane.go
```diff
@@ -1133,14 +1133,28 @@ func (b *Botanist) DeployKubeControllerManager(ctx context.Context) error {
...
+	if b.Shoot.Info.Status.LastOperation != nil && b.Shoot.Info.Status.LastOperation.Type == gardencorev1beta1.LastOperationTypeCreate {
+		if b.Shoot.HibernationEnabled {
+			// shoot is being created with .spec.hibernation.enabled=true, don't deploy KCM at all
+			defaultValues["replicas"] = 0
+		} else {
+			// shoot is being created with .spec.hibernation.enabled=false, deploy KCM
+			defaultValues["replicas"] = 1
+		}
+	} else {
+		if b.Shoot.HibernationEnabled && b.Shoot.Info.Status.IsHibernated {
```
ggaurav10: Should it be `==` instead of `&&`?

ggaurav10 (changes requested):

vpnachev (dismissed):
/lgtm

pkg/operation/botanist/controlplane.go
rfranzke: Yes, nice catch, thanks. PTAL

timebertt (approved):
/lgtm

ggaurav10 (approved):

schrodit (approved):

PR #2522, opened 2020-06-29
Update Cluster extension resource when Shoot is updated

rfranzke:

/area usability
/kind enhancement
/priority normal

Some parts of the shoot reconciliation flow are updating the `Shoot` resource in the garden cluster. Unfortunately, the `Cluster` resource in the seed cluster was not updated with these changes. Hence, it was having an outdated state until the shoot reconciliation started again from beginning.

This PR improves this situation by introducing a helper function that updates both the `Shoot` and the `Cluster`. This way subsequent steps get to the up-to-date version of the `Shoot` resource.

/cc @majst01 @Gerrit91 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/botanist.go
+++ b/pkg/operation/botanist/botanist.go
@@ -21,12 +21,15 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/client-go/util/retry"
+
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	"github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
 	"github.com/gardener/gardener/pkg/operation"
 	"github.com/gardener/gardener/pkg/operation/common"
 	shootpkg "github.com/gardener/gardener/pkg/operation/shoot"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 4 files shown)


pkg/operation/botanist/botanist.go
```diff
@@ -165,6 +166,22 @@ func (b *Botanist) CreateETCDSnapshot(ctx context.Context) error {
 	return err
 }
 
+// UpdateShootAndCluster updates the given `core.gardener.cloud/v1beta1.Shoot` resource in the garden cluster after
+// applying the given transform function to it. It will also update the `shoot` field in the
+// extensions.gardener.cloud/v1alpha1.Cluster` resource in the seed cluster with the updated shoot information.
+func (b *Botanist) UpdateShootAndCluster(ctx context.Context, shoot *gardencorev1beta1.Shoot, transform func() error) error {
+	if _, err := controllerutil.CreateOrUpdate(ctx, b.K8sGardenClient.Client(), shoot, transform); err != nil {
```
timebertt: Shouldn't this be updated with a "retry on conflict"?
Also, we should probably use the direct client for the Shoot update, as we are likely to run into conflicts otherwise, as we are constantly updating the status in the shoot flow.
```suggestion
	if err := kutil.TryUpdate(ctx, retry.DefaultRetry, b.K8sGardenClient.DirectClient(), shoot, transform); err != nil {
```

timebertt (commented):
Looks good, nit only :)

pkg/operation/botanist/botanist.go
rfranzke: done

timebertt (approved):
Thanks!
/lgtm

danielfoehrKn (approved):

PR #2164, opened 2020-04-08
Fix prometheus-config-reloader failing to start

mvladev:
**What this PR does / why we need it**:

Due to [CVE-2019-5736](https://nvd.nist.gov/vuln/detail/CVE-2019-5736) runC is now initially loaded into memory when the container starts.

After some time VPA recommends less memory (2,3Mb) than the size of of runC binary (about 10Mb).

This results in an error when trying to start (due to VPA action / pod deletion) the container:

```
failed to write 2485760 to memory.limit_in_bytes in /sys/fs/cgroup/memory/kubepods/prometheus-config-reloader/memory.limit_in_bytes: device or resource busy
```

see:
https://github.com/lxc/lxc/commit/6400238d08cdf1ca20d49bafb85f4e224348bf9d
https://github.com/helm/charts/issues/11447#issuecomment-464716379

**Which issue(s) this PR fixes**:
Fixes #2163 

**Special notes for your reviewer**:

**Release note**:

```improvement operator
VPA recommendations are now disabled for Prometheus's prometheus-config-reloader container.
```

/cc @amshuman-kr 
/cc @dansible

PR #2528, opened 2020-06-29
Add node-local-dns feature

zanetworker:

/area networking
/kind enhancement
/priority normal

This PR adds the `node-local-cache` component for optimizing DNS perfomance as a `Gardenlet` feature-gate. 

**Release note**:

```improvement operator
Node-local DNS is now supported and can be enabled with the `NodeLocalDNS` feature gate in the gardenlet's component configuration. More information can be found in the documentation (https://github.com/gardener/gardener/blob/master/docs/usage/node-local-dns.md).
```
```noteworthy user
When the `NodeLocalDNS` feature gate is enabled and a migration from IPTables to IPVS is performed then newer pods will be configured to use the node-local DNS while older pods will still use the non-cached CoreDNS server in the cluster. To enable older pods to use the node-local you have to restart the pods.
```


charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet-config-file
```diff
@@ -19,7 +19,11 @@ cgroupsPerQOS: true
 cgroupDriver: cgroupfs
 clusterDomain: {{ required "kubernetes.domain is required" .Values.kubernetes.domain }}
 clusterDNS:
+{{- if and .Values.kubernetes.ipvsEnabled .Values.kubernetes.nodeLocalDNSEnabled}}
+- {{ required "kubernetes.nodeLocalDNS is required" .Values.kubernetes.nodeLocalDNS }}
+{{- else }}
 - {{ required "kubernetes.clusterDNS is required" .Values.kubernetes.clusterDNS }}
+{{- end }}
```
rfranzke: The `ipvsEnabled `, `nodeLocalDNSEnabled`, and `nodeLocalDNS` variables are not needed. You can simply pass the node-local DNS cache IP as part of the `clusterDNS` variable and get the same result.

charts/shoot-core/components/charts/nodelocal-dns/templates/daemonset.yaml
```diff
@@ -0,0 +1,98 @@
+---
+apiVersion: apps/v1
```
rfranzke:
```suggestion
apiVersion: {{ include "daemonsetversion" . }}
```

charts/shoot-core/components/charts/nodelocal-dns/templates/daemonset.yaml
```diff
@@ -0,0 +1,98 @@
...
+          operator: Exists
+        - effect: NoExecute
+          operator: Exists
+        - effect: NoSchedule
+          operator: Exists
+      containers:
+        - name: node-cache
+          image: {{ index .Values.images "node-local-dns" }}
+          resources:
+            requests:
```
rfranzke, line 44: Can you add limits?

charts/shoot-core/components/charts/nodelocal-dns/templates/daemonset.yaml
```diff
@@ -0,0 +1,98 @@
...
+      maxUnavailable: 10%
+  selector:
+    matchLabels:
+      k8s-app: node-local-dns
+  template:
+    metadata:
+      labels:
+        k8s-app: node-local-dns
+        networking.gardener.cloud/to-dns: allowed
+      annotations:
```
rfranzke, line 25: Can you add this same comment here?
https://github.com/gardener/gardener/blob/master/charts/shoot-core/components/charts/coredns/templates/coredns-deployment.yaml#L27

charts/shoot-core/components/charts/nodelocal-dns/templates/service.yaml
```diff
@@ -0,0 +1,20 @@
+---
```
rfranzke, line 1: Hm, what's the purpose of this service? It points to CoreDNS, or? How is it different from https://github.com/gardener/gardener/blob/master/charts/shoot-core/components/charts/coredns/templates/coredns-service.yaml?

charts/shoot-core/components/charts/nodelocal-dns/templates/daemonset.yaml
rfranzke: Can you add a VPA if it's enabled?

docs/usage/node-local-dns.md
```diff
@@ -0,0 +1,32 @@
+# NodeLocalDNS Configuration
+
+This is a short guide describing how to enabled DNS caching on the shoot cluster nodes.
```
rfranzke:
```suggestion
This is a short guide describing how to enable DNS caching on the shoot cluster nodes.
```

docs/usage/node-local-dns.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Background
+
+Currently in Gardener we are using CoreDNS as a deployment that is auto-scaled horizontally to cover for  QPS-intensive applications. However, doing so does not seem to be enough to completely circumvent DNS bottlenecks such as:
```
rfranzke:
```suggestion
Currently in Gardener we are using CoreDNS as a deployment that is auto-scaled horizontally to cover for QPS-intensive applications. However, doing so does not seem to be enough to completely circumvent DNS bottlenecks such as:
```

docs/usage/node-local-dns.md
```diff
@@ -0,0 +1,32 @@
+# NodeLocalDNS Configuration
```
rfranzke, line 1: Can you add the document to the index please (docs/README.md)

pkg/operation/botanist/addons.go
```diff
@@ -379,6 +383,19 @@ func (b *Botanist) generateCoreAddonsChart() (*chartrenderer.RenderedChart, erro
 		return nil, err
 	}
 
+	// The node-local-dns interface cannot bind the kube-dns cluster IP since the interface
+	// used for IPVS load-balancing already uses this address.
+	if b.Shoot.IPVSEnabled() {
+		nodeLocalDNSConfig["clusterDNS"] = b.Shoot.Networks.CoreDNS.String()
+	} else {
+		nodeLocalDNSConfig["dnsServer"] = b.Shoot.Networks.CoreDNS.String()
+	}
```
rfranzke, line 392: See comment above, there is not even a difference with these values

rfranzke (dismissed):

charts/shoot-core/components/charts/nodelocal-dns/templates/service.yaml
zanetworker: this is to avoid DNS loops, the service IP needs to be different.

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet-config-file
zanetworker: Not sure if the order is preserved, according to the documentation of the [cluster-dns](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet)  flag:
 
> There is no guarantee as to which DNS server may be contacted for name resolution

rfranzke: Hm, which order? In both cases if the `if-else` block you only add exactly one IP. My point was that it doesn't help to have multiple variables here. Just keep `clusterDNS` and set it accordingly when rendering the Helm chart.

pkg/operation/botanist/addons.go
zanetworker: Which comment, this is not the same for the kubelet, this is for the chart. 

The have the same value, but there place in the chart is different, clusterDNS is not set for Kube-proxy and is replaced by the node-local-dns logic `__PILLAR__CLUSTER__DNS__`.

rfranzke: Ah, yep, right, thanks for pointing it out

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet-config-file
zanetworker: > Hm, which order?

I thought you mean remove the conditions and add them all to the list (maybe I misunderstood). 

then the logic of which value will be computed in the code with a condition right?

rfranzke: Yes, probably. The benefit would be that you can more easily write unit tests for the function that computes the chart values. OTOH, I agree that this is still not optimal because even if the Helm chart values are correctly computed/tested the Helm chart itself must still be correct as well, and this is not that straight-forward to test. :(
Anyway, for this particular code snippet I think it's more clear to have only one variable that is filled with the correct value by the caller.

charts/shoot-core/components/charts/nodelocal-dns/templates/daemonset.yaml
```diff
@@ -0,0 +1,102 @@
...
+      containers:
+        - name: node-cache
+          image: {{ index .Values.images "node-local-dns" }}
+          resources:
+            requests:
+              cpu: 25m
+              memory: 5Mi
+            limits:
+              cpu: 100m
+              memory: 25Mi
```
rfranzke:
```suggestion
              memory: 25Mi
            limits:
              cpu: 100m
              memory: 100Mi
```

ref https://github.com/gardener/gardener/pull/2164

charts/shoot-core/components/charts/nodelocal-dns/templates/daemonset.yaml
```diff
@@ -0,0 +1,102 @@
...
+          args:
+          - -localip
+          {{- if .Values.dnsServer }}
+          - "{{ .Values.nodeLocal }},{{ .Values.dnsServer }}"
+          {{- else }}
+          - "{{ .Values.nodeLocal }}"
+          {{- end }}
```
rfranzke, line 56: You could do the same here as well like suggested in https://github.com/gardener/gardener/pull/2528/files#r447442439

zanetworker: we already need these variables for the configMap, so it's just plug and play here. Parsing comma separating string in go then rendering seems off a bit to me here IMO.

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -126,16 +127,25 @@ func (b *Botanist) generateDownloaderConfig(machineImageName string) map[string]
...
+	var (
+		kubernetesConfig = map[string]interface{}{
+			"clusterDNS": b.Shoot.Networks.CoreDNS.String(),
+			"domain":     gardencorev1beta1.DefaultDomain,
+			"version":    b.Shoot.Info.Spec.Kubernetes.Version,
+		}
+	)
+
+	if b.Shoot.NodeLocalDNSEnabled && b.Shoot.IPVSEnabled() {
+		kubernetesConfig["clusterDNS"] = net.ParseIP("169.254.20.10").String()
```
danielfoehrKn: > If using kube-proxy in IPVS mode, --cluster-dns flag to kubelet needs to be modified to use <node-local-address> that NodeLocal DNSCache is listening on. Otherwise, there is no need to modify the value of the --cluster-dns flag, since NodeLocal DNSCache listens on both the kube-dns service IP as well as <node-local-address>

Does that set the the kubelet flag `--cluster-dns` pointing to the <node-local-address> that NodeLocal DNSCache pod is listening on?

danielfoehrKn: Can you please add a comment above or link [the relevant documentation](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/). Otherwise I find it hard to understand. 

Can we somehow make clear  in the comment that
- IPtables -> node-local-dns interface binds kube-dns service cluster ip & listens/binds  <node-local-address>
- IPVS -> node-local-dns pods listens/binds only on <node-local-address>.  + [the reason you already mentioned](https://github.com/gardener/gardener/pull/2528/files#r447438938) above. Then also state that --cluster-dns flag to kubelet needs to be modified to use <node-local-address> where the NodeLocal DNSCache is listening on.

danielfoehrKn (commented):
Nice feature. 

 I am not entirely convinced though, that this feature should only be hidden behind a feature gate. 

To me this is like an optional addon.
I think it would be useful to be able to configure it on the Shoot level. At least there is another system daemon installed and we cannot know if user workload makes heavy use of DNS.

zanetworker:
@danielfoehrKn  The intention is not to make it an add-on but rather a core component as it is defined at the moment. The reason it's behind a feature-gate is for evaluation purposes, not more (i.e., its not meant to be optional or at least that was not the intention I meant to convey).

pkg/operation/botanist/operatingsystemconfig.go
zanetworker: > Does that set the the kubelet flag --cluster-dns pointing to the that NodeLocal DNSCache pod is listening on?

yes. 



> Can you please add a comment above or link [the relevant documentation](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/). Otherwise I find it hard to understand.
> 
> Can we somehow make clear in the comment that
> 
> * IPtables -> node-local-dns interface binds kube-dns service cluster ip & listens/binds
> * IPVS -> node-local-dns pods listens/binds only on .  + [the reason you already mentioned](https://github.com/gardener/gardener/pull/2528/files#r447438938) above. Then also state that --cluster-dns flag to kubelet needs to be modified to use  where the NodeLocal DNSCache is listening on.

Okay will add the link to the usage documenation which should cover your the IPTables and IPVS comment ... Would that be enough from your perspective?

zanetworker: https://github.com/gardener/gardener/pull/2528/files#diff-9c519add19895c6c2c8598c45aa2cf78R32

danielfoehrKn:
I will test it out tomorrow and then will be able to give feedback (was on vacation).

docs/usage/node-local-dns.md
```diff
@@ -0,0 +1,32 @@
...
+
+![node-local-dns-architecture](node-local-dns.png)
+
+## Configuring NodeLocalDNS
+
+All that needs to be done to enable the usage of the `node-local-dns` feature is to set the corresponding feature-gate to true for the
+`Gardenlet`:
+
+```yaml
+ featureGates:
```
danielfoehrKn, line 28: Can you add the feature gate also in the example configuration of the gardenlet?

danielfoehrKn (commented):
I tested the feature flag for a Shoot with IP Tables
- The Shoot reconciled successfully and the node-local-dns daemon set got applied 
- The node-local-dns pods seem to be operating like they should. The dummy network interface `nodelocaldns` got created with the 2 IP adresses (Shoot is using IPTables) by the daemon set pods.
- DNS resolution works inside the container

I will still have to test ipVs and migration ipTables -> ipVS

zanetworker:
@danielfoehrKn as discussed offline, I still need to think about the best options to switch from IPTables to IPVS for a running cluster with minimal downtime. Thank you for testing anyways, I will put in draft till I figure out the best way to do that.

zanetworker:
@danielfoehrKn I added a release-note for the IPTables-IPVS migration case.

danielfoehrKn:
/needs/rebase

pkg/operation/botanist/addons.go
```diff
@@ -379,6 +383,20 @@ func (b *Botanist) generateCoreAddonsChart() (*chartrenderer.RenderedChart, erro
 		return nil, err
 	}
 
+	// The node-local-dns interface cannot bind the kube-dns cluster IP since the interface
+	// used for IPVS load-balancing already uses this address.
+	if b.Shoot.IPVSEnabled() {
+
```
danielfoehrKn:
```suggestion
```

docs/README.md
```diff
@@ -39,6 +39,7 @@
 * [Troubleshooting guide](usage/trouble_shooting_guide.md)
 * [`ManagedIstio` feature](usage/istio.md)
 * [Trusted TLS certificate for shoot control planes](usage/trusted-tls-for-control-planes.md)
+* [`NodeLocalDNS` feature ](usage/node-local-dns.md)
```
timebertt: Can you also add this link to your release note?
```suggestion
* [`NodeLocalDNS` feature](usage/node-local-dns.md)
```

charts/shoot-core/components/charts/nodelocal-dns/templates/daemonset.yaml
```diff
@@ -0,0 +1,102 @@
...
+          securityContext:
+            privileged: true
+          ports:
+            - containerPort: 53
+              name: dns
+              protocol: UDP
+            - containerPort: 53
+              name: dns-tcp
+              protocol: TCP
+            - containerPort: 9253
```
timebertt:
```suggestion
            - containerPort: {{ .Values.prometheus.port }}
```

charts/shoot-core/components/charts/nodelocal-dns/templates/psp/nodelocal-dns-psp.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: {{ include "podsecuritypolicyversion" .}}
+kind: PodSecurityPolicy
+metadata:
+  name: gardener.kube-system.node-local-dns
+  labels:
+    app: node-local-dns
+spec:
+  privileged: true
+  hostNetwork: true
+  volumes:
```
timebertt: I think the PSP needs some changes for unprivileged clusters:
```
  Warning  FailedCreate  13s (x3 over 16s)  daemonset-controller  Error creating: pods "node-local-dns-" is forbidden: unable to validate against any pod security policy: [spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[3]: Invalid value: "secret": secret volumes are not allowed to be used spec.containers[0].hostPort: Invalid value: 53: Host port 53 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 53: Host port 53 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 9253: Host port 9253 is not allowed to be used. Allowed ports: [] spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.containers[0].hostPort: Invalid value: 53: Host port 53 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 53: Host port 53 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 9253: Host port 9253 is not allowed to be used. Allowed ports: []]
  Warning  FailedCreate  5s (x5 over 15s)   daemonset-controller  Error creating: pods "node-local-dns-" is forbidden: unable to validate against any pod security policy: [spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.containers[0].hostPort: Invalid value: 53: Host port 53 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 53: Host port 53 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 9253: Host port 9253 is not allowed to be used. Allowed ports: [] spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[3]: Invalid value: "secret": secret volumes are not allowed to be used spec.containers[0].hostPort: Invalid value: 53: Host port 53 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 53: Host port 53 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 9253: Host port 9253 is not allowed to be used. Allowed ports: []]
```

charts/shoot-core/components/charts/nodelocal-dns/templates/service.yaml
```diff
@@ -0,0 +1,20 @@
...
+  name: kube-dns-upstream
+  namespace: kube-system
+  labels:
+    k8s-app: kube-dns-upstream
+spec:
+  ports:
+    - name: dns
+      port: 53
+      protocol: UDP
+      targetPort: 8053
```
timebertt, line 14: Would it make sense to use a named port here?
```suggestion
      targetPort: dns-udp
```

charts/shoot-core/components/values.yaml
```diff
@@ -75,3 +75,8 @@ vertical-pod-autoscaler:
   enabled: false
 cluster-identity:
   clusterIdentity: shoot-namesapce-shoot-ui-garden-cluster-identity
+node-local-dns:
+  enabled: false
+  podAnnotations: {}
+  images:
+    nodelocal-agent: image-repository:image-tag
```
timebertt:
```suggestion
    node-local-dns: image-repository:image-tag
```

docs/usage/node-local-dns.md
```diff
@@ -0,0 +1,32 @@
...
+This is a short guide describing how to enable DNS caching on the shoot cluster nodes.
+
+## Background
+
+Currently in Gardener we are using CoreDNS as a deployment that is auto-scaled horizontally to cover for QPS-intensive applications. However, doing so does not seem to be enough to completely circumvent DNS bottlenecks such as:
+
+- Cloud provider limits for DNS lookups.
+- Unreliable UDP connections that forces a period of timeout in case packets are dropped.
+- Unnecessary node hopping since CoreDNS is not deployed on all nodes, and as a result DNS queries end-up traversing multiple nodes before reaching the destination server.
+- Inefficient load-balancing of Kube-proxy services (e.g., round-robin might not be enough when using IPTables mode)
```
timebertt: Do you mean load-balancing of the kube-dns service?

docs/usage/node-local-dns.md
```diff
@@ -0,0 +1,32 @@
...
+- and more ...
+
+To workaround the issues described above, `node-local-dns` was introduced. The architecture is described below. The idea is simple:
+
+- For new queries, the connection is upgraded from UDP to TCP and forwarded towards the cluster IP for the original CoreDNS server.
```
timebertt, line 17: Just a question for my own understanding:
Is the IP connection indeed upgraded to TCP or is the caching CoreDNS rather talking via TCP to the upstream kube-dns service and then returning the response to the Pod?

pkg/operation/botanist/addons.go
timebertt:
```suggestion
```

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -126,16 +126,25 @@ func (b *Botanist) generateDownloaderConfig(machineImageName string) map[string]
...
 func (b *Botanist) generateOriginalConfig() (map[string]interface{}, error) {
+	var (
+		kubernetesConfig = map[string]interface{}{
+			"clusterDNS": b.Shoot.Networks.CoreDNS.String(),
+			"domain":     gardencorev1beta1.DefaultDomain,
+			"version":    b.Shoot.Info.Spec.Kubernetes.Version,
+		}
+	)
+
+	if b.Shoot.NodeLocalDNSEnabled && b.Shoot.IPVSEnabled() {
```
danielfoehrKn, line 139:
```suggestion
// if IPVS is enabled, instruct the kubelet to create pods resolving DNS to the `nodelocaldns` network interface link-local ip address
// for more information checkout the [usage documentation](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/)
	if b.Shoot.NodeLocalDNSEnabled && b.Shoot.IPVSEnabled() {
```

charts/seed-monitoring/charts/grafana/dashboards/owners/coredns-node-local-dns-dashboard.json
```diff
@@ -0,0 +1,2025 @@
+{
```
timebertt: Would it make sense to not enable this dashboard if the feature gate is disabled?

charts/shoot-core/components/charts/nodelocal-dns/templates/daemonset.yaml
```diff
@@ -0,0 +1,102 @@
...
+      k8s-app: node-local-dns
+  template:
+    metadata:
+      labels:
+        k8s-app: node-local-dns
+        networking.gardener.cloud/to-dns: allowed
+        # we won't be using the checksum of the configmap since coreDNS provides the "reload" plugins that does the reload if config changes.
+      annotations:
+        prometheus.io/port: "{{ .Values.prometheus.port }}"
+        prometheus.io/scrape: "{{ .Values.prometheus.scrape }}"
```
timebertt, line 27: Hmm, I don't see any node-local-dns scrape targets in prometheus after enabling the feature gate.

timebertt (changes requested):
Thanks for the PR!

I tested the PR and also tried the migration from IPTables to IPVS. After the migration I started a new Pod and it still had the `kube-dns` clusterIP in its `resolv.conf`.
```
/ # cat /etc/resolv.conf
nameserver 100.64.0.10
```

Is this correct? I thought, in the IPVS case Pods would need to talk to `169.254.20.10` in order to leverage the node local dns cache.

charts/shoot-core/components/charts/nodelocal-dns/templates/service.yaml
zanetworker: > tested the PR and also tried the migration from IPTables to IPVS. After the migration I started a new Pod and it still had the `kube-dns` clusterIP in its `resolv.conf`.
> 
> ```
> / # cat /etc/resolv.conf
> nameserver 100.64.0.10
> ```
> 
> Is this correct? I thought, in the IPVS case Pods would need to talk to `169.254.20.10` in order to leverage the node local dns cache.


Please make sure that the cluster is fully reconciled, I just tested it and indeed the pod got the cache IP.

zanetworker:
> tested the PR and also tried the migration from IPTables to IPVS. After the migration I started a new Pod and it still had the `kube-dns` clusterIP in its `resolv.conf`.
> 
> ```
> / # cat /etc/resolv.conf
> nameserver 100.64.0.10
> ```
> 
> Is this correct? I thought, in the IPVS case Pods would need to talk to `169.254.20.10` in order to leverage the node local dns cache.


Please make sure that the cluster is fully reconciled, I just tested it and indeed the pod got the cache IP.

charts/shoot-core/components/charts/nodelocal-dns/templates/service.yaml
zanetworker: I remember there were issues previously using named-ports for services, so I would vote for sticking with numerics for service target ports. We are also not using named target ports for services elsewhere AFAIK.

docs/usage/node-local-dns.md
zanetworker: No I meant kube-proxy (maybe I should have added IPTables as well) but maybe it is not so intuitive here so I will change it.

docs/usage/node-local-dns.md
zanetworker: That would be the connection from the node-local-agent to CoreDNS. The response is returned to the agent, then cached and provided to the requester.

charts/seed-monitoring/charts/grafana/dashboards/owners/coredns-node-local-dns-dashboard.json
zanetworker: Makes sense yes, but as I said, we don't yet have the logic to do that in place for the rest of the dashboards. So optionally, I can remove this dashboard this PR and later add it once we have the logic.

timebertt:
> Please make sure that the cluster is fully reconciled, I just tested it and indeed the pod got the cache IP.

Hmm, the cluster was fully reconciled and even now (after `~1h`) the pod still gets the `x.x.0.10` as the `nameserver`.
```
/ # cat /etc/resolv.conf
nameserver 100.64.0.10
```

charts/shoot-core/components/charts/nodelocal-dns/templates/service.yaml
timebertt: Ok, fine for me 👍

docs/usage/node-local-dns.md
timebertt: Got it, thanks for clarifying!

charts/seed-monitoring/charts/grafana/dashboards/owners/coredns-node-local-dns-dashboard.json
timebertt: Hmm, can't you do it similar to what you did with the VPN dashboard here:
https://github.com/gardener/gardener/blob/94eb4ab4fe4f05eeed58ab9aabf4235c26e586c0/charts/seed-monitoring/charts/grafana/templates/grafana-dashboards-configmap.yaml#L18-L20

danielfoehrKn (changes requested):
I am missing some documentation for the migration scenarios. Could you add that for

**IPVS -> IPTables**
 - Existing pods  continue to leverage the node-local dns cache

IPTables -> IPVS
- Existing pods **will stop using the node-local dns cache** (they used it before)
- Only new pods will use the cache

In addition to that I did some testing had issues when creating Shoots using IPVS and also migrating from IPtables -> IPVS.
I am not sure if these failures are related to the node-local-dns feature, but I can also not exclude it.
In the migration case for instance the existing node became `NotReady` after calico-node failed. 
We can get in touch to look over these issues if you like.

timebertt:
> > Please make sure that the cluster is fully reconciled, I just tested it and indeed the pod got the cache IP.
> 
> Hmm, the cluster was fully reconciled and even now (after `~1h`) the pod still gets the `x.x.0.10` as the `nameserver`.
> 
> ```
> / # cat /etc/resolv.conf
> nameserver 100.64.0.10
> ```

After some debugging, I figured out, that this was a bug caused by the `CachedRuntimeClients` feature gate (fixed in #2591).
Therefore, please ignore my comment on that problem, it was not caused by this PR.

zanetworker:
@danielfoehrKn we can have a look at your case together yes.

charts/seed-monitoring/charts/grafana/dashboards/owners/coredns-node-local-dns-dashboard.json
zanetworker: I removed monitoring from this PR.

charts/shoot-core/components/charts/nodelocal-dns/templates/daemonset.yaml
zanetworker: I removed monitoring from this PR.

docs/README.md
timebertt: Can you also add this link to your release note?

timebertt (dismissed):
Nice, thank you @zanetworker!
Looks completely good now :)
/lgtm

rfranzke:
@danielfoehrKn will take another look
/remove needs/rebase
/needs second-opinion

timebertt (approved):
/lgtm
I had no further issues while testing.

danielfoehrKn:
I can confirm that creation of Shoot clusters with IPVS and IPtables works - and that the node local dns looks good.
However migration from IPtables -> IPVS with a Cluster with Kubernetes version 1.18.4 (or 18.1) & calico does not work in my setup.
I think what happens is that calico-node quite predictably crashes - I will have to investigate further.
```
 MemoryPressure                Unknown   Tue, 21 Jul 2020 23:59:25 +0200   Wed, 22 Jul 2020 00:01:37 +0200   NodeStatusUnknown               Kubelet stopped posting node status.
  DiskPressure                  Unknown   Tue, 21 Jul 2020 23:59:25 +0200   Wed, 22 Jul 2020 00:01:37 +0200   NodeStatusUnknown               Kubelet stopped posting node status.
  PIDPressure                   Unknown   Tue, 21 Jul 2020 23:59:25 +0200   Wed, 22 Jul 2020 00:01:37 +0200   NodeStatusUnknown               Kubelet stopped posting node status.
  Ready                         Unknown   Tue, 21 Jul 2020 23:59:25 +0200   Wed, 22 Jul 2020 00:01:37 +0200   NodeStatusUnknown               Kubelet stopped posting node status.
```

I tried twice to reproduce the issue using the exact same cluster but without node-local-dns and then the migration worked.  

 I will tomorrow morning 
- get the calico-node logs from the crashed node & kubelet 
- try to test the migration also using Cilium as a networking extension .

danielfoehrKn:
debugged with @zanetworker and it seems not to be related to the node-local-dns feature. 

Migration from IPTables to IPVS with gardenlinux OS works fine. 

On Ubuntu version 18.4.20200228 the kubelet cannot talk to the API server any more, node is replaced and the new node is working just fine. Will need to investigate more.

/lgtm

danielfoehrKn (approved):

timebertt:
@zanetworker @danielfoehrKn 
We can get this PR in now, right?

PR #2530, opened 2020-06-30
Fix panic on initializing Shoot clients

### LINKED ISSUE #2526: Gardenlet panics when deleting Shoot

/area quality
/kind bug
/priority blocker

**What happened**:
#2449 seems to have introduced a regression: when the gardenlet tries to cleanup webhooks in the shoot with the kube-apiserver deployment scaled down to 0, it panics here:
https://github.com/gardener/gardener/blob/f120f2c74a503ab4a388aa055cd08d4081eaf5ed/pkg/operation/botanist/cleanup.go#L183

This is probably because the Shoot clients are not initialized anymore, if the kube-apiserver is scaled down:
https://github.com/gardener/gardener/blob/f120f2c74a503ab4a388aa055cd08d4081eaf5ed/pkg/operation/operation.go#L293-L300

gardenlet not to panic

**Environment**:

- Gardener version: `v1.7.0-dev-e47e4c057e49b2245529ac3a7bffc943973f40d1`
- Cloud provider or hardware configuration: azure (but probably also applies to others)
- Others:

timebertt:

/area quality
/kind bug
/priority blocker

This PR fixes a regression/panic introduced by #2449.
Basically the motivation in the PR was to ensure, the shoot clients are not initialized when its API server is not running at the moment (mainly to avoid, that the care controller would create new clientsets when the shoot is hibernated or being deleted). However, if the API server is scaled down for whatever reason, when the gardenlet tries to initialize clients and the reconcile/delete flow, it should not skip initializing the clients as this will cause it to panic later on in the flow when trying to use the client.

Fixes #2526

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/operation.go
+++ b/pkg/operation/operation.go
@@ -289,13 +289,15 @@ func (o *Operation) InitializeShootClients() error {
 		return nil
 	}
 
-	apiServerRunning, err := o.IsAPIServerRunning()
-	if err != nil {
-		return err
-	}
-	// Don't initialize clients for Shoots, that have are currently hibernated or in the process of being created or deleted.
-	if !apiServerRunning {
-		return nil
+	if o.Shoot.HibernationEnabled {
+		// Don't initialize clients for Shoots, that are currently hibernated and their API server is not running
+		apiServerRunning, err := o.IsAPIServerRunning()
+		if err != nil {
+			return err
+		}
+		if !apiServerRunning {
+			return nil
+		}
 	}
 
 	shootClient, err := o.ClientMap.GetClient(context.TODO(), keys.ForShoot(o.Shoot.Info))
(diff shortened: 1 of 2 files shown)


pkg/operation/operation.go
```diff
@@ -289,13 +289,15 @@ func (o *Operation) InitializeShootClients() error {
...
 	}
 
-	apiServerRunning, err := o.IsAPIServerRunning()
-	if err != nil {
-		return err
-	}
-	// Don't initialize clients for Shoots, that have are currently hibernated or in the process of being created or deleted.
-	if !apiServerRunning {
-		return nil
+	if o.Shoot.HibernationEnabled {
```
rfranzke, line 292: Hm, what about the case when a shoot is being woken up? Then `o.Shoot.HibernationEnabled=false`. Wouldn't it be safer to not have this condition?

timebertt: Actually, I just reverted this func to how it was before my PR (plus/minus renaming). The removal of this condition was actually the problem, that caused the regresssion.
If the shoot is woken up, the condition is false, so it skips checking if the API server is running and directly tries to get a client for the shoot, which is what we want in the reconcile flow. In the care controller, we don't want this, so we do the check for a running API server there.

rfranzke: OK, thanks for the explanation.

rfranzke (approved):
/lgtm

PR #2533, opened 2020-06-30
Pardon conditions when operation is processing and no last errors

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal

Conditions are now not only pardoned for `Create`/`Delete` operations but also for processing `Reconcile` operations in case there aren't any last errors.

/cc @tim-ebert 

**Release note**:

```improvement operator
Conditions are now not only pardoned for `Create`/`Delete` operations but also for processing `Reconcile` operations in case there aren't any last errors.
```


danielfoehrKn (approved):
/lgtm

PR #2534, opened 2020-06-30
Default machine image without version to latest non-preview version

danielfoehrKn:

/area operations
/kind bug
/priority normal

Properly default machine image without version to latest non-preview version.
Otherwise, Shoots that are created without a machine image or without a machine image version will get defaulted to a preview versions. 
Preview versions should always be opt-in, as they are not deemed "stable" by the Gardener Operator.

Fixes #

**Release note**:

```improvement operator
Fixes a bug that could lead to defaulting a  machine image of a Shoot to a preview version.
```

### DIFF
--- a/pkg/apis/core/helper/helpers.go
+++ b/pkg/apis/core/helper/helpers.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 
 	"github.com/gardener/gardener/pkg/apis/core"
-	versionutils "github.com/gardener/gardener/pkg/utils/version"
 
 	"github.com/Masterminds/semver"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


rfranzke (approved):
/lgtm

danielfoehrKn:
@tim-ebert, maybe for release 1.7?

pkg/apis/core/helper/helpers.go
```diff
@@ -57,58 +56,49 @@ func QuotaScope(scopeRef corev1.ObjectReference) (string, error) {
...
-	for _, imageVersion := range image.Versions {
-		v, err := semver.NewVersion(imageVersion.Version)
+	for _, version := range versions {
+		v, err := semver.NewVersion(version.Version)
 		if err != nil {
-			return core.ExpirableVersion{}, fmt.Errorf("error while parsing machine image version '%s' of machine image '%s': version not valid: %s", imageVersion.Version, image.Name, err.Error())
+			return core.ExpirableVersion{}, fmt.Errorf("error while parsing expirable version '%s': %s", version.Version, err.Error())
+		}
+
+		if filterPreviewVersions && version.Classification != nil && *version.Classification == core.ClassificationPreview {
```
ialidzhikov, line 87: Does it still make sense to default to preview version if the CloudProfile contains only preview versions? Currently it will deny if there are only preview versions in the CloudProfile.

ialidzhikov (commented):
Nice catch, only 1 question inline.

pkg/apis/core/helper/helpers.go
rfranzke: Hm, probably not, or? This would be quite inconsistent.

ialidzhikov: Ok, agreed.

ialidzhikov (approved):
/lgtm

PR #2535, opened 2020-06-30
Set shoot conditions to progressing after reconciliation

### LINKED ISSUE #2387: Add grace period for conditions after successful Shoot reconciliation

/area ops-productivity
/area quality
/kind enhancement
/priority normal

I see a lot of Shoot clusters, which finish to be created/woken up successfully (`lastOperation.state=Succeeded`) but need some more time to get completely healthy. E.g. `elasticsearch-logging` takes some time to completely start up (including volume provisioning) and therefore causes the `ControlPlaneHealthy` condition to be false for a short period of time after successful Shoot creation. Another example are the calico components, which take some time to start up and therefore sometimes cause either `SystemComponentsHealthy` or `EveryNodeReady` to be false.

I propose to add a grace period (of about `5m`, but configurable) after successful creation, in which failing health checks result in conditions with status `Progressing` instead of `False`.

**Why is this needed**:
These cases can happen quite often but don't indicate any problematic error. Pardoning those health checks for a short period of time after creation, can help gardener operators to filter out those non-problematic cases and focus on more relevant issues.

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal

Failed shoot conditions are now set to `Progressing` status for the configured `conditionThresholds` time after a successful shoot reconciliation. This is to prevent false negative status reports shortly after reconciliations.

Fixes #2387

* /invite @tim-ebert 
* I delete some dead code along the way (/cc @wyb1)

**Release note**:

```improvement operator
Failed shoot conditions are set to `Progressing` status for the configured `conditionThresholds` time after a successful shoot reconciliation. This is to prevent false negative status reports shortly after reconciliations.
```

### DIFF
--- a/pkg/operation/botanist/health_check.go
+++ b/pkg/operation/botanist/health_check.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"sync"
 	"time"
 
@@ -37,7 +36,6 @@ import (
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	"github.com/gardener/gardener/pkg/utils/kubernetes/health"
 
-	prometheusmodel "github.com/prometheus/common/model"
 	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


timebertt (approved):
Very nice! Thanks :)
/lgtm

PR #2536, opened 2020-07-01
Rely only on deployed tunnel pod in care controller

rfranzke:

/area ops-productivity
/kind bug
/priority normal

Disabling the `KonnectivityTunnel` feature gate leads to this error message currently:

```
Tunnel connection has not been established (could not forward to vpn-shoot pod: port forward connection could not be established within five seconds)
```

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/health_check.go
+++ b/pkg/operation/botanist/health_check.go
@@ -618,18 +618,13 @@ func (b *Botanist) checkSystemComponents(
 	if err := b.K8sShootClient.Client().List(ctx, podsList, client.InNamespace(metav1.NamespaceSystem), client.MatchingLabels{"type": "tunnel"}); err != nil {
 		return nil, err
 	}
-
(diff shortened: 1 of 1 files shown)


zanetworker:
the check you removed makes sure that the feature is enabled, otherwise it does not make sense to do the health-check. Does it rather mean that the resource manager did not finish the deletion yet?

rfranzke:
When shoots are reconciled only once per 24h and you disable konnectivity then it will only be replaced by the VPN after <= 24h. The health check, however, does already check for the VPN which produces the error messages pasted in the PR description.

zanetworker:
I see, thanks for explaining.

zanetworker (dismissed):
/lgtm

timebertt (dismissed):
/lgtm

timebertt (approved):
/lgtm

Issue #1240, opened 2019-07-23
Redeploy components instead of simply scaling them up when deleting hibernated shoot


rfranzke:
**What would you like to be added**:
From #1194:

> I think the thing we should rather do is to re-apply the expected components instead of scaling them imperatively - I'll take that one.

**Why is this needed**:
When we introduce new components we should ensure that they are available in case we delete a hibernated shoot.

PR #2540, opened 2020-07-01
Don't delete tunnel secrets on deletion

### LINKED ISSUE #2458: Failed to delete hibernated cluster w/ KAS network proxy

/area control-plane
/kind bug
/priority normal

**What happened**:
A 1.18.x shoot which had been created and hibernated could not be deleted after the `KonnectivityTunnel` feature gate was activated.

The Kube-Apiserver hung in the init state because of the following events:

```
  Warning  FailedMount  1s (x6 over 17s)  kubelet, ip-10-242-20-136.eu-west-1.compute.internal  MountVolume.SetUp failed for volume "vpn-seed-tlsauth" : secret "vpn-seed-tlsauth" not found
  Warning  FailedMount  1s (x6 over 17s)  kubelet, ip-10-242-20-136.eu-west-1.compute.internal  MountVolume.SetUp failed for volume "vpn-seed" : secret "vpn-seed" not found
```

The shoot is deleted successfully.

1. Disable `KonnectivityTunnel` (https://github.com/gardener/gardener/pull/2251)
1. Create 1.18.x shoot
1. Hibernate shoot
1. Activate `KonnectivityTunnel`
1. Reconcile shoot
1. Delete shoot

**Environment**:

- Gardener version: v1.6.1

timebertt:

/area networking
/kind bug
/priority normal

If the `KonnectivityTunnel` was enabled while a Shoot was hibernated, it can't be deleted without being reconciled first.

Fixes #2458 

/invite @zanetworker 
@rfranzke and I decided to not redeploy the API server on deletion, as we are not sure about all the consequences and rather just not delete the vpn secrets on deletion, so the API server can get ready again.

**Release note**:

```improvement operator
An issue has been fixed, which caused the deletion of hibernated Shoots to be blocked if the `KonnectivityTunnel` has been enabled while the Shoot was hibernated.
```

### DIFF
--- a/pkg/operation/botanist/secrets.go
+++ b/pkg/operation/botanist/secrets.go
@@ -90,13 +90,15 @@ func (b *Botanist) GenerateAndSaveSecrets(ctx context.Context) error {
 		}
 	}
 
-	if b.Shoot.KonnectivityTunnelEnabled {
-		if err := b.cleanupTunnelSecrets(ctx, &gardenerResourceDataList, "vpn-seed", "vpn-seed-tlsauth", "vpn-shoot"); err != nil {
-			return err
-		}
-	} else {
-		if err := b.cleanupTunnelSecrets(ctx, &gardenerResourceDataList, common.KonnectivityServerKubeconfig, common.KonnectivityServerCertName); err != nil {
-			return err
+	if b.Shoot.Info.DeletionTimestamp != nil {
+		if b.Shoot.KonnectivityTunnelEnabled {
+			if err := b.cleanupTunnelSecrets(ctx, &gardenerResourceDataList, "vpn-seed", "vpn-seed-tlsauth", "vpn-shoot"); err != nil {
+				return err
+			}
+		} else {
+			if err := b.cleanupTunnelSecrets(ctx, &gardenerResourceDataList, common.KonnectivityServerKubeconfig, common.KonnectivityServerCertName); err != nil {
+				return err
+			}
 		}
 	}
 
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -1365,7 +1365,7 @@ func (b *Botanist) setAPIServerServiceClusterIP(clusterIP string) {
 			},
 			Name:                     v1beta1constants.DeploymentNameKubeAPIServer,
 			IstioIngressNamespace:    common.IstioIngressGatewayNamespace,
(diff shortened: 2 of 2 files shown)


zanetworker (dismissed):
> @rfranzke and I decided to not redeploy the API server on deletion, as we are not sure about all the consequences and rather just not delete the VPN secrets on deletion, so the API server can get ready again. 

That's another way to do it but we anyway need to redeploy instead of scale for #1240 right? 

But regardless it fixes the issue at hand, so /lgtm

rfranzke:
@zanetworker yes, but we can't overlook the implications/side-effects this has now in a short period of time. Given that @tim-ebert is currently working on a new release candidate we preferred to discuss #1240 after the release was made. I can't remember anymore exactly why we had postponed #1240 but it's already open a long time and probably there was a reason.

timebertt:
> That's another way to do it but we anyway need to redeploy instead of scale for #1240 right?

Yes, exactly. As we two already discussed, we also have to think about all the other components we should redeploy.
So let's take some time somewhen after the release to figure out a proper way to solve the #1240 in a general sense.

rfranzke (dismissed):

rfranzke (approved):

pkg/operation/botanist/secrets.go
```diff
@@ -90,13 +90,15 @@ func (b *Botanist) GenerateAndSaveSecrets(ctx context.Context) error {
...
 	}
 
-	if b.Shoot.KonnectivityTunnelEnabled {
-		if err := b.cleanupTunnelSecrets(ctx, &gardenerResourceDataList, "vpn-seed", "vpn-seed-tlsauth", "vpn-shoot"); err != nil {
-			return err
-		}
-	} else {
-		if err := b.cleanupTunnelSecrets(ctx, &gardenerResourceDataList, common.KonnectivityServerKubeconfig, common.KonnectivityServerCertName); err != nil {
-			return err
+	if b.Shoot.Info.DeletionTimestamp != nil {
```
timuthy, line 93: Shouldn't it be
```
b.Shoot.Info.DeletionTimestamp == nil
```
?
/cc @tim-ebert

timebertt: Indeed. Nice catch!

timuthy: Are you going to fix it in another PR?

timebertt: Yep, done: #2562

PR #2541, opened 2020-07-01
 Update the helm install script version

ialidzhikov:

/priority normal

It looks like the v2.13.1 get script is broken and fails with:

```
Downloading https://kubernetes-helm.storage.googleapis.com/helm-og:image-linux-amd64.tar.gz
SHA sum of /tmp/helm-installer-9e9VWc/helm-og:image-linux-amd64.tar.gz does not match. Aborting.
Failed to install helm with the arguments provided: --version v2.13.1
Accepted cli arguments are:
	[--help|-h ] ->> prints this help
	[--version|-v <desired_version>] . When not defined it defaults to latest
	e.g. --version v2.4.0  or -v latest
	[--no-sudo]  ->> install without sudo
	For support, go to https://github.com/helm/helm.
make: *** [Makefile:139: install-requirements] Error 1
```

It seems that the tag computation is broken for some reason

```
++ cat /var/folders/p9/6dr9snnx6ws1dxkj4b66st6r0000gn/T/helm-installer-G3nyPR/helm-og:image-darwin-amd64.tar.gz.sha256
+ local 'expected_sum=<?xml version='\''1.0'\'' encoding='\''UTF-8'\''?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: kubernetes-helm/helm-og:image-darwin-amd64.tar.gz.sha256</Details></Error>'
+ '[' 6255bd212013e0898aba6ac0fda15e35d50bf430c9018401316f31ddde32e9d9 '!=' '<?xml version='\''1.0'\'' encoding='\''UTF-8'\''?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: kubernetes-helm/helm-og:image-darwin-amd64.tar.gz.sha256</Details></Error>' ']'
+ echo 'SHA sum of /var/folders/p9/6dr9snnx6ws1dxkj4b66st6r0000gn/T/helm-installer-G3nyPR/helm-og:image-darwin-amd64.tar.gz does not match. Aborting.'
SHA sum of /var/folders/p9/6dr9snnx6ws1dxkj4b66st6r0000gn/T/helm-installer-G3nyPR/helm-og:image-darwin-amd64.tar.gz does not match. Aborting.
```

This PR adapts to use the get script from the v2.16.9 branch.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/hack/install-requirements.sh
+++ b/hack/install-requirements.sh
@@ -20,7 +20,7 @@ echo "> Installing requirements"
 
 export GO111MODULE=on
 curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.27.0
-curl -s "https://raw.githubusercontent.com/helm/helm/v2.13.1/scripts/get" | bash -s -- --version 'v2.13.1'
+curl -s "https://raw.githubusercontent.com/helm/helm/v2.16.9/scripts/get" | bash -s -- --version 'v2.13.1'
 
 if [[ "$(uname -s)" == *"Darwin"* ]]; then
   cat <<EOM


hack/install-requirements.sh
```diff
@@ -20,7 +20,7 @@ echo "> Installing requirements"
 
 export GO111MODULE=on
 curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.27.0
-curl -s "https://raw.githubusercontent.com/helm/helm/v2.13.1/scripts/get" | bash -s -- --version 'v2.13.1'
+curl -s "https://raw.githubusercontent.com/helm/helm/master/scripts/get" | bash -s -- --version 'v2.13.1'
```
dguendisch: Wouldn't it be better to simply clone that script instead of relying what is currently in the master branch?

ialidzhikov: Hmm, I have seen several comments that recommend the usage of the install script from the master - ref https://github.com/helm/helm/issues/5486#issuecomment-474970334. I would rather rely on it from the master instead of double maintenance on our side, or?

dguendisch: I think the comment refers more to the fact that patches/fixes went obviously into master and not necessarily justified a new helm release (from their side).
Anyway, was just thinking out loud because of the usual security concerns with executing code from such "unstable" URLs (a one time "hack" of the helm repo and a one time sneak in of malicious code is enough to be run inside our CI/CD, even if helm maintainers would notice and take down a few hours later ...).
If there's at least a `helm/v3.x.y` version of the script fixing that issue, I'd prefer it, even more I'd prefer a copy of that script.
But we can defer this discussion as well because your current PR fixes the current CI/CD issues.

MartinWeindel:
```
curl -s "https://raw.githubusercontent.com/helm/helm/v3.2.4/scripts/get" | bash -s -- --version 'v2.13.1'
```
should also work

ialidzhikov: I updated to `v2.16.9`.

rfranzke (approved):

hack/install-requirements.sh
dguendisch: Thank you!

dguendisch (approved):

timebertt (approved):
/lgtm

PR #2544, opened 2020-07-01
Refresh clients on secret update

timebertt:

/area quality
/kind enhancement
/priority blocker

With this PR the ClientMaps check if the config used for constructing a ClientSet has changed since creation (e.g. kubeconfig secret was updated). If that's the case, the ClientSet is refreshed (invalidated and recreated).
This PR also fixes limiting the refresh checks directly after a ClientSet has been created.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/client/kubernetes/clientmap/internal/generic_clientmap.go
+++ b/pkg/client/kubernetes/clientmap/internal/generic_clientmap.go
@@ -32,7 +32,13 @@ var _ clientmap.ClientMap = &GenericClientMap{}
 
 const (
 	waitForCacheSyncTimeout = 5 * time.Minute
-	maxRefreshInterval      = 5 * time.Second
+)
+
+var (
+	// MaxRefreshInterval is the maximum rate at which the version and hash of a single ClientSet are checked, to
+	// decide whether the ClientSet should be refreshed. Also, the GenericClientMap waits at least MaxRefreshInterval
+	// after creating a new ClientSet before checking if it should be refreshed.
+	MaxRefreshInterval = 5 * time.Second
 )
 
 // GenericClientMap is a generic implementation of clientmap.ClientMap, which can be used by specific ClientMap
@@ -58,6 +64,7 @@ type clientMapEntry struct {
 	clientSet kubernetes.Interface
 	synced    bool
 	cancel    context.CancelFunc
+	hash      string
 
 	// refreshLimiter limits the attempts to refresh the entry due to an outdated server version.
 	refreshLimiter *rate.Limiter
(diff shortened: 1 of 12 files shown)


pkg/client/kubernetes/clientmap/internal/generic_clientmap.go
```diff
@@ -32,7 +32,10 @@ var _ clientmap.ClientMap = &GenericClientMap{}
 
 const (
 	waitForCacheSyncTimeout = 5 * time.Minute
-	maxRefreshInterval      = 5 * time.Second
+)
+
+var (
+	MaxRefreshInterval = 5 * time.Second
```
rfranzke, line 41: doc string for exported variable

timebertt: Done

rfranzke (dismissed):
/lgtm

pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
```diff
@@ -120,6 +121,34 @@ func (f *ShootClientSetFactory) NewClientSet(ctx context.Context, k clientmap.Cl
...
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to get seed client: %w", err)
+	}
+
+	project, err := ProjectForNamespaceWithClient(ctx, gardenClient.Client(), shoot.Namespace)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to get Project for Shoot %q: %w", key.Key(), err)
+	}
+
+	return shootpkg.ComputeTechnicalID(project.Name, shoot), seedClient, nil
```
ialidzhikov, line 153: The technicalID should be also present in the Shoot status. Would it be possible to get it from there instead of computing it again?

danielfoehrKn:
The PR looks good. Though in my opinion  and already discussed with @tim-ebert it is not good that we 
-  cannot refresh the garden client when the underlying secret changed. 
-  have Informers running with the old client and no way to change their client

This is a problem for the certificate rotation - we currently need to kill the gardenlet pod in order to use the new certificate.  

Edit: I linked [an Issue](https://github.com/gardener/gardener/issues/2546)

rfranzke:
@danielfoehrKn IMO this is no problem. The certificate rotation happens once per year, so let's live with the restart of the gardenlet pod in favor of less complex code.

pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
timebertt: `ComputeTechnicalID` takes a look at the `technicalID` in the shoot status and falls back to the computation if it is not set yet.
This is just how it was done before in `InitializeShootClients`, I just moved the code.
DYT we should only rely on the status?

ialidzhikov: You are right, I  missed that part in `ComputeTechnicalID`. Probably we can skip the `ProjectForNamespaceWithClient` invocation if the shoot already has the technicalID set.

danielfoehrKn:
@rfranzke for now thats fine with me. Though we might want to have an open conversation about the topic. I linked [an Issue](https://github.com/gardener/gardener/issues/2546).

danielfoehrKn (dismissed):

pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
timebertt: Well, good point. Done 👍

timebertt:
Addressed @ialidzhikov's comments, adapted the shoot clientmap test and added the missing test cases for `CalculateClientSetHash`.
PTAL :)

rfranzke (approved):

ialidzhikov (approved):
/lgtm

PR #2547, opened 2020-07-02
Set HibernationPossible=True if Shoot already hibernated

timebertt:

/area usability
/kind bug
/priority normal

With #2449 the `HibernationPossible` constraint is set to `False` if the Shoot is already hibernated. This PR reverts this change.
However, the constraint is still set to `False` if the control plane is not running at the moment, as we cannot be sure, that the shoot can be hibernated properly in that case.

Also see https://github.com/gardener/dashboard/pull/724

**Release note**:

```improvement operator
NONE
```


rfranzke (approved):
/lgtm

danielfoehrKn (approved):

PR #2548, opened 2020-07-02
Refactor infrastructure extension management with DeployWaiter concept

rfranzke:

/area testing
/kind enhancement
/priority normal

Similar to #2442, this PR is now refactoring the management of the `Infrastructure` extension resource as part of the `Botanist` structure in a way to leverage the new `DeployWaiter` interface introduced with #1980.

It also fixes a wrongly called method in the `WaitCleanup` function of the `Network` extension management.

/invite @mvladev 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_delete.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_delete.go
@@ -314,12 +314,12 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation) *gardencorev1bet
 		})
 		destroyNetwork = g.Add(flow.Task{
 			Name:         "Destroying shoot network plugin",
-			Fn:           flow.TaskFn(botanist.Shoot.Components.Network.Destroy).RetryUntilTimeout(defaultInterval, defaultTimeout),
+			Fn:           flow.TaskFn(botanist.Shoot.Components.Extensions.Network.Destroy).RetryUntilTimeout(defaultInterval, defaultTimeout),
 			Dependencies: flow.NewTaskIDs(cleanShootNamespaces),
 		})
 		waitUntilNetworkIsDestroyed = g.Add(flow.Task{
 			Name:         "Waiting until shoot network plugin has been destroyed",
-			Fn:           botanist.Shoot.Components.Network.WaitCleanup,
+			Fn:           botanist.Shoot.Components.Extensions.Network.WaitCleanup,
 			Dependencies: flow.NewTaskIDs(destroyNetwork),
 		})
 		destroyWorker = g.Add(flow.Task{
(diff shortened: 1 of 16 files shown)


pkg/operation/botanist/infrastructure.go
```diff
@@ -16,147 +16,81 @@ package botanist
...
-		requestInfrastructureOperation = creationPhase || shootIsWakingUp || restorePhase || controllerutils.HasTask(b.Shoot.Info.Annotations, common.ShootTaskDeployInfrastructure)
-		infrastructure                 = &extensionsv1alpha1.Infrastructure{
+	b.Shoot.Components.Extensions.Infrastructure.SetSSHPublicKey(b.Secrets[v1beta1constants.SecretNameSSHKeyPair].Data[secrets.DataKeySSHAuthorizedKeys])
+
+	if err := b.Shoot.Components.Extensions.Infrastructure.Deploy(ctx); err != nil {
+		return err
+	}
+
+	if b.isRestorePhase() {
+		return b.restoreExtensionObject(ctx, b.K8sSeedClient.Client(), &extensionsv1alpha1.Infrastructure{
```
swilen-iwanow, line 66: I think this similarly to the Network will throw a nil pointer, because inside the restore we are trying to access some fields from the Infrastructure obj. May be we can temporally add  `GET` of the Infra object, until @plkokanov 
 change is merged.

rfranzke: @swilen-iwanow hm, how does it work for `Network`s currently? There is no `GET` call either, or am I missing something?
https://github.com/gardener/gardener/blob/5773c96710adabe61574f39428ab2a6d43645a1f/pkg/operation/botanist/network.go#L55-L62

swilen-iwanow: No. My point was that with the change of the `Network`s to use a component approach and passing an empty object to the `restoreExtensionObject`, instead of the created object like it used to be, the restore scenario for the `Network`s is not working anymore. With the Infra it will be the same. What I'm saying is that we can add a quick fix by getting the object inside the `restoreExtensionObject` func, which could be done in a separate PR.

rfranzke: OK, understood, thanks. Can you open such a PR for the `Network`? I'm happy to incorporate the changes in this PR as well. Or, you wait until this PR is merged and do it together at once for both `Network` and `Infrastructure`.

swilen-iwanow: I can wait until this PR is merged. If #2511 is merged as well I will just implement the new interfaces that are supposed to fix this

pkg/operation/botanist/extensions/infrastructure/infrastructure.go
```diff
@@ -0,0 +1,224 @@
...
+	return i.providerStatus
+}
+
+// NodesCIDR returns the generated nodes CIDR of the provider.
+func (i *infrastructure) NodesCIDR() *string {
+	return i.nodesCIDR
+}
+
+// emptyInfrastructure returns an empty Infrastructure resource used for deletion.
+func (i *infrastructure) emptyInfrastructure() *extensionsv1alpha1.Infrastructure {
```
ialidzhikov: I guess it is unused.
```suggestion
```

pkg/operation/botanist/extensions/infrastructure/infrastructure_test.go
```diff
@@ -0,0 +1,304 @@
...
+					v1beta1constants.GardenerOperation: v1beta1constants.GardenerOperationWaitForState,
+					v1beta1constants.GardenerTimestamp: now.UTC().String(),
+				}
+			}),
+		)
+	})
+
+	Describe("#Wait", func() {
+		It("should return error when it's not found", func() {
+			c.EXPECT().Get(gomock.Any(), kutil.Key(namespace, name), &extensionsv1alpha1.Infrastructure{}).DoAndReturn(func(_ context.Context, _ client.ObjectKey, obj *extensionsv1alpha1.Infrastructure) error {
```
ialidzhikov: You could simply use `.Return` in this case.

```suggestion
			c.EXPECT().
				Get(gomock.Any(), kutil.Key(namespace, name), &extensionsv1alpha1.Infrastructure{}).
				Return(apierrors.NewNotFound(schema.GroupResource{}, name)).AnyTimes()
```

pkg/operation/botanist/extensions/infrastructure/infrastructure_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+			Expect(deployWaiter.Wait(ctx)).To(Succeed())
+			Expect(deployWaiter.ProviderStatus()).To(Equal(providerStatus))
+			Expect(deployWaiter.NodesCIDR()).To(PointTo(Equal(nodesCIDR)))
+		})
+	})
+
+	Describe("#Destroy", func() {
+		It("should not return error when it's not found", func() {
+			c.EXPECT().Get(gomock.Any(), kutil.Key(namespace, name), infra).DoAndReturn(func(_ context.Context, _ client.ObjectKey, obj *extensionsv1alpha1.Infrastructure) error {
```
ialidzhikov:
```suggestion
			c.EXPECT().
				Get(gomock.Any(), kutil.Key(namespace, name), infra).
				Return(apierrors.NewNotFound(schema.GroupResource{}, name))
```

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -460,7 +459,7 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation) *gardencorev1bet
 
 		destroyInternalDomainDNSRecord = g.Add(flow.Task{
 			Name:         "Destroying internal DNS entry",
-			Fn:           flow.TaskFn(component.OpWaiter(botanist.Shoot.Components.DNS.InternalEntry).Destroy),
+			Fn:           botanist.Shoot.Components.Extensions.DNS.InternalEntry.Destroy,
```
ialidzhikov: Hm, but this will only delete the DNSEntry but won't wait until it is removed from store. Is it this change on purpose? I guess it is better to fail with `failed to wait until ... is deleted` instead of `failed to wait until Shoot namespace in the seed is deleted`.

rfranzke: Hm, agreed, I wasn't aware of these implicit consequences. I'll change it back... thanks for finding it.

ialidzhikov (approved):
/lgtm

PR #2549, opened 2020-07-02
Fix health conditions if API server is not running

timebertt:

/area quality
/kind bug
/priority blocker

With #2449 all Shoot health conditions were set to true, if the API server is not running.
This PR fixes this issue and sets the same condition status as if the API server couldn't be reached.

Fixes #

**Release note**:

```improvement operator
NONE
```


rfranzke (approved):
/lgtm

PR #2550, opened 2020-07-02
Add option to specify shoot annotations for create shoot test

dguendisch:
/area testing
/kind enhancement
/priority normal

Adds an option `-annotations foo=bar,a=b` to the shoot creation test. The provided annotations are added to the test shoot when it is created.
This is needed to enable testing of shoot behaviour that is dependent on some annotations (i.e. the upcoming remedy controller).

### DIFF
--- a/test/framework/shoot_utils.go
+++ b/test/framework/shoot_utils.go
@@ -167,6 +167,27 @@ func CreateShootTestArtifacts(cfg *ShootCreationConfig, projectNamespace string,
 	return shoot.Name, shoot, nil
 }
 
+func parseAnnotationCfg(cfg string) (map[string]string, error) {
+	if !StringSet(cfg) {
+		return nil, nil
+	}
+	result := make(map[string]string)
+	annotations := strings.Split(cfg, ",")
+	for _, annotation := range annotations {
+		annotation = strings.TrimSpace(annotation)
+		if !StringSet(annotation) {
+			continue
+		}
+		keyValue := strings.Split(annotation, "=")
+		if len(keyValue) != 2 {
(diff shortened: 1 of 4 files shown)


ialidzhikov (dismissed):
/lgtm

schrodit (commented):
please also add the annotation in the docs, otherwise lgtm
https://github.com/gardener/gardener/blob/master/docs/testing/integration_tests.md#shoot-creation-test

test/framework/shoot_utils.go
```diff
@@ -160,13 +160,50 @@ func CreateShootTestArtifacts(cfg *ShootCreationConfig, projectNamespace string,
 		return "", nil, err
 	}
 
+	if err := setShootAnnotations(shoot, cfg); err != nil {
```
timebertt: This would be a great fit for `setShootMetadata`, right? Do you want to call it from there?

timebertt (commented):
Nit only, otherwise /lgtm

Will you then also use this flag to add the workflow name to TM Shoots? That would be really great to be able to jump from (failing) Shoot to argo workflow/TM run :)

schrodit:
> Will you then also use this flag to add the workflow name to TM Shoots? That would be really great to be able to jump from (failing) Shoot to argo workflow/TM run :)

Yes, this is one use case:)

test/framework/shoot_utils.go
dguendisch: makes sense

dguendisch:
addressed comments

timebertt (approved):
Thanks for addressing!
/lgtm

PR #2552, opened 2020-07-03
Avoid running into timeouts when dumping shoot state

timebertt:
Co-authored-by: Tim Schrodi <tim.schrodi@sap.com>

/area testing
/kind enhancement
/priority normal

We see some integration test runs, where we run into timeouts when dumping the shoot's state, because the API server is not running.
With this PR we check if the API server is running before dumping, so we can avoid running into timeouts, so that the Seed state will still be dumped.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/test/framework/shoot_utils.go
+++ b/test/framework/shoot_utils.go
@@ -99,7 +99,8 @@ func (f *ShootFramework) DumpState(ctx context.Context) {
 			f.Logger.Fatalf("Cannot decode shoot %s: %s", f.Shoot.GetName(), err)
 		}
 
-		if f.ShootClient != nil {
+		isRunning, err := f.IsAPIServerRunning(ctx)
+		if f.ShootClient != nil && isRunning && err == nil {
 			ctxIdentifier := fmt.Sprintf("[SHOOT %s]", f.Shoot.Name)
 			f.Logger.Info(ctxIdentifier)
 			if err := f.DumpDefaultResourcesInAllNamespaces(ctx, ctxIdentifier, f.ShootClient); err != nil {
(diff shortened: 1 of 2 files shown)


test/framework/shootframework.go
```diff
@@ -322,3 +325,19 @@ func (f *ShootFramework) WaitForShootCondition(ctx context.Context, interval, ti
 		return retry.MinorError(fmt.Errorf("shoot %q does not yet have expected condition", shoot.Name))
 	})
 }
+
+func (f *ShootFramework) IsAPIServerRunning(ctx context.Context) (bool, error) {
```
rfranzke, line 331: doc string

rfranzke (approved):

test/framework/shoot_utils.go
```diff
@@ -108,6 +109,12 @@ func (f *ShootFramework) DumpState(ctx context.Context) {
 			if err := f.dumpNodes(ctx, ctxIdentifier, f.ShootClient); err != nil {
 				f.Logger.Errorf("unable to dump information of nodes from shoot %s: %s", f.Shoot.Name, err.Error())
 			}
+		} else {
+			errMsg := ""
+			if err != nil {
+				errMsg = ": " + err.Error()
+			}
+			f.Logger.Errorf("unable to dump resources from shoot %s: API server is currently not running%s", f.Shoot.Name, errMsg)
```
dguendisch, line 117:
```suggestion
			f.Logger.Errorf("unable to dump resources from shoot %s: API server is currently not running %s", f.Shoot.Name, errMsg)
```

dguendisch (changes requested):
Good idea, thanks for spotting. Only a nit comment.

test/framework/shoot_utils.go
timebertt: Hmm, `errMsg` is either empty or filled with `: <some error message>`...

dguendisch: oops, you're right, thanks!

dguendisch (approved):

PR #2555, opened 2020-07-03
Upgrade resource-manager and enable always-update for Shoots

### LINKED ISSUE #134: Missing the mutating webhook leads to broken nginx ingress controller
/area control-plane
/platform aws

**What happened**:
1. Created a cluster with enabled nginx-ingress addon
2. I assume the webhook gardener-extension-provider-aws was down/unreachable in that moment
3. The nginx `ConfigMap` in the shoot's `kube-system` ns was created with `use-proxy-protocol: "false"` (obviously the default from https://github.com/gardener/gardener/blob/master/charts/shoot-addons/charts/nginx-ingress/values.yaml#L15)
4. However the LBs in AWS are configured with proxy protocol enabled
5. As a result, all requests via LB to the nginx ingress respond with `HTTP 400` as nginx didn't understand the proxy protocol.

Nginx ingress to work.

I think part of the problem is the gardener-extension-provider-aws webhook's `failurePolicy: Ignore`.
Maybe the webhook configuration could be more selective so that `failurePolicy: Fail` could be used instead.

timebertt:

/area quality
/kind bug
/priority normal

Upgrade resource-manager and enable the new `--always-update` flag for the resource-manager in the Shoot ControlPlane.

Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/134

**Release note**:

``` improvement operator github.com/gardener/gardener-resource-manager #71 @tim-ebert
Missing RBAC rules for updating and patching secrets were added.
```

``` improvement developer github.com/gardener/gardener-resource-manager #70 @rfranzke
The new `--always-update` command line parameter (default: `false`) allows to configure whether to always send a `PUT` request for managed resources regardless of whether their desired state differs from their actual state.
```

``` improvement operator github.com/gardener/gardener-resource-manager #68 @tim-ebert
A bug has been fixed, that caused new Services to get assigned a different ClusterIP than specified.
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -24,7 +24,7 @@ images:
 - name: gardener-resource-manager
   sourceRepository: github.com/gardener/gardener-resource-manager
   repository: eu.gcr.io/gardener-project/gardener/gardener-resource-manager
-  tag: "v0.13.1"
+  tag: "v0.14.0"
 - name: dependency-watchdog
   sourceRepository: github.com/gardener/dependency-watchdog
   repository: eu.gcr.io/gardener-project/gardener/dependency-watchdog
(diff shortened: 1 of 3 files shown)


rfranzke (approved):
/lgtm

PR #2562, opened 2020-07-07
Fix tunnel secret handling on Shoot deletion

timebertt:

/area quality
/kind bug
/priority normal

Fix tunnel secret handling: only cleanup the tunnel secrets, when the Shoot is not yet marked for deletion.

Fixes #

/cc @timuthy 
Thanks for catching!

**Release note**:

```improvement operator
Fixed a bug, that caused the tunnel secrets not to be deleted, in case the used tunnel has changed.
```


timuthy (approved):
/lgtm
Thanks for fixing!

PR #2563, opened 2020-07-07
Allow configuring machine controller settings per workerpool.

### LINKED ISSUE #1843: Expose maxEvictRetries and machineDrainTimeout to shoot-resource from MCM.
Currently, MCM exposes `maxEvictRetries` and `machineDrainTimeout` as part of the flags, [refer](https://github.com/gardener/machine-controller-manager/blob/master/cmd/machine-controller-manager/app/options/options.go#L106-L107).  It would be nice to have this exposed as part of the shoot-resource, maybe under `worker` section or separate MCM configuration. 

**Why is this needed**:
At the moment, we set one default value for all the shoots via MCM, this fits for few doesn't for many. Based on the nature of the workload and machines, I think it makes sense to allow setting up the different drain-timeout and eviction-retries per shoot.
- We have also seen a request from a consumer to expose the same.

PS: There is also a partially related issue on MCM for the same: [#372](https://github.com/gardener/machine-controller-manager/issues/372)

hardikdr:

/area usability
/kind api-change
/priority normal

**What this PR does / why we need it**:  This PR allows configuring the machine-controller knobs via shoot-resource. Following configurations are exposed as part of this PR:
- DrainTimeout, HealthTimeout, CreationTimeout, MaxEvictRetries, NodeConditions. 

Fixes #1843 

**Release note**:

```improvement operator
Shoot resource now allows configuring following machine-controller parameters: DrainTimeout, HealthTimeout, CreationTimeout, MaxEvictRetries, NodeConditions.
```


hardikdr:
/hold 

https://github.com/gardener/machine-controller-manager/pull/478 needs to be merged,  released, and vendored in g/g.

hardikdr:
Open for early review, though I'd prefer to keep the hold till MCM is released with the related changes.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -262,6 +263,12 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
 			},
 		}
 
+		// Find the workerpool related to current machine deployment.
+		workerPool := workerPoolForMachineDeployment(worker, deployment)
+		if workerPool == nil {
+			a.logger.Error(fmt.Errorf("Error findig worker pool for machine deployment"), "Worker pool settings will not be applied on related machine deployment.", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
```
prashanth26:
```suggestion
			a.logger.Error(fmt.Errorf("Error finding worker pool for machine deployment"), "Worker pool settings will not be applied on related machine deployment.", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
```

prashanth26:
@hardikdr - I think this PR also would require this change to avoid rollingUpdates - https://github.com/gardener/gardener/pull/2579 once the MCM image is updated.

hardikdr:
@tim-ebert I have rebased the PR.

Also, the publish and verify steps are running since quite long,.

timebertt:
> Also, the publish and verify steps are running since quite long,.

Actually, they weren't running at all. 
I triggered them for you :)

hardikdr:
Thanks @tim-ebert . 

Also please ignore the failure of the `verify` step, for now, we need to release and vendor MCM for it to pass. It would be nice if rest of the PR could be reviewed till then. 
- We'll try to release MCM sooner, possibly tomorrow.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -309,6 +316,13 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
 								Taints: deployment.Taints,
 							},
 						},
+						MachineControllerConfig: &machinev1alpha1.MachineControllerConfig{
+							MachineDrainTimeout:    workerPool.Settings.MachineDrainTimeout,
+							MachineCreationTimeout: workerPool.Settings.MachineCreationTimeout,
+							MachineHealthTimeout:   workerPool.Settings.MachineHealthTimeout,
+							MaxEvictRetries:        workerPool.Settings.MaxEvictRetries,
+							NodeConditions:         workerPool.Settings.NodeConditions,
```
danielfoehrKn: Nil pointer danger? workerPool can be nil and also Settings can be nil

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -277,6 +278,12 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
 			},
 		}
 
+		// Find the workerpool related to current machine deployment.
+		workerPool := workerPoolForMachineDeployment(worker, deployment)
+		if workerPool == nil {
+			a.logger.Error(fmt.Errorf("Error finding worker pool for machine deployment"), "Worker pool settings will not be applied on related machine deployment.", "worker", fmt.Sprintf("%s/%s", worker.Namespace, worker.Name))
```
danielfoehrKn: I guess its the "fault" of the  extension provider if machine deployments are generated that do not include the worker pool name, but I have troubles with just logging and continuing here. 
Can we have at least an event on the worker resource here?
Possibly it makes even sense to fail the reconciliation as otherwise the user thinks that all settings have been successfully applied - but in fact they are not.

extensions/pkg/controller/worker/genericactuator/interface.go
```diff
@@ -41,6 +41,7 @@ type WorkerDelegate interface {
 	DeployMachineClasses(context.Context) error
 
 	// GenerateMachineDeployments generates the configuration for the desired machine deployments.
+	// Workerpool name should be substring of the machine deployment name.
```
danielfoehrKn: I do not think this is stable?. 
What happens if 
- workerName: worker-12
- workerName: worker-1

The machine deployment for `worker-12` would match both the worker pool `worker-1` and `worker-12`.
When looping over an unsorted machineDeployment slice this could lead to changes to the MachineDeployment specifications leading to an unwanted rolling update.

danielfoehrKn (changes requested):

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
hardikdr: Yes, good catch, seemingly it is one. Out of curiosity, I tested a couple of times now- removing settings, but it doesn't throw the nil-pointer, couldn't understand why. I also found that, I had mistakenly missed the `omitempty [here](https://github.com/gardener/gardener/blob/42cf17c3d3269b329894d30141f30b32f5202d22/pkg/apis/core/v1alpha1/types_shoot.go#L923). I fixed, that locally, and tried still couldn't produce the nil-pointer. 

Anyways, I will try to align the code to check if corresponding fields[drainTimeout, etc] are set and then only use here. Thanks for pointing out.

extensions/pkg/controller/worker/genericactuator/interface.go
hardikdr: Thanks for pointing, I was in the dilemma, but I just realized,  should have a worker-pool name as part of its labels, I'll give it a try.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
hardikdr: Yes, agreed. It makes sense to fail the reconciliation, to avoid confusion. I'll make that change.

extensions/pkg/controller/worker/genericactuator/interface.go
danielfoehrKn: Yes, the machine deployment should have a label with the worker pool name I think.

prashanth26: >The machine deployment for worker-12 would match both the worker pool worker-1 and worker-12.
When looping over an unsorted machineDeployment slice this could lead to changes to the MachineDeployment specifications leading to an unwanted rolling update.

Nice catch @danielfoehrKn . 

Yes, I think the label way would be preferred.

example/90-shoot.yaml
```diff
@@ -105,6 +105,12 @@ spec:
     # zones: # optional, only relevant if the provider supports availability zones
     # - europe-central-1a
     # - europe-central-1b
+    # settings: # optional, it allows to configure the machine-controller.
```
rfranzke: Wouldn't `machineControllerManager` or `managementSettings` or similar be better here? `settings` alone is quite unspecific and vague (most other fields are "settings" for the worker pool itself, but this section here is about the mgmt of them)

extensions/pkg/controller/worker/genericactuator/interface.go
rfranzke: Alternatively, you can let the extension controllers provide the to-be-used settings. The same is done for labels/taints/annotations, see https://github.com/gardener/gardener-extension-provider-aws/blob/master/pkg/controller/worker/machines.go#L190-L192.
It has the advantage that you don't need any code here in the `genericactuator` (and no label matching or similar), but the "disadvantage" (not sure that it really is one) that the worker controllers have to be adapted. However, there it's easier because they are looping over the worker pools already.

pkg/apis/core/types_shoot.go
```diff
@@ -734,6 +734,17 @@ type Worker struct {
...
+	Settings *Settings
+}
+
+// Settings contains configurations for different worker-pools.
+type Settings struct {
+	MachineDrainTimeout    *metav1.Duration
+	MachineHealthTimeout   *metav1.Duration
+	MachineCreationTimeout *metav1.Duration
+	MaxEvictRetries        *int32
+	NodeConditions         *string
```
rfranzke: Please add doc strings.

pkg/apis/core/types_shoot.go
```diff
@@ -734,6 +734,17 @@ type Worker struct {
 	// Zones is a list of availability zones that are used to evenly distribute this worker pool. Optional
 	// as not every provider may support availability zones.
 	Zones []string
+	// Settings contains configurations for different worker-pools. Eg. MachineDrainTimeout, MachineHealthTimeout.
+	Settings *Settings
+}
+
+// Settings contains configurations for different worker-pools.
```
rfranzke: Please refine

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -918,6 +918,28 @@ type Worker struct {
 	// SystemComponents contains configuration for system components related to this worker pool
 	// +optional
 	SystemComponents *WorkerSystemComponents `json:"systemComponents,omitempty" protobuf:"bytes,18,opt,name=systemComponents"`
+	// Settings contains configurations for different worker-pools. Eg. MachineDrainTimeout, MachineHealthTimeout.
+	// +optional
+	Settings *Settings `protobuf:"bytes,19,opt,name=settings"`
+}
+
+// Settings contains configurations for different worker-pools.
```
rfranzke: Please refine

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -915,6 +915,28 @@ type Worker struct {
 	// SystemComponents contains configuration for system components related to this worker pool
 	// +optional
 	SystemComponents *WorkerSystemComponents `json:"systemComponents,omitempty" protobuf:"bytes,18,opt,name=systemComponents"`
+	// Settings contains configurations for different worker-pools. Eg. MachineDrainTimeout, MachineHealthTimeout.
+	// +optional
+	Settings *Settings `json:"settings,omitempty" protobuf:"bytes,19,name=settings"`
+}
+
+// Settings contains configurations for different worker-pools.
```
rfranzke: Please refine

pkg/apis/extensions/v1alpha1/types_worker.go
```diff
@@ -130,6 +130,28 @@ type WorkerPool struct {
 	// Zones contains information about availability zones for this worker pool.
 	// +optional
 	Zones []string `json:"zones,omitempty"`
+	// Settings contains configurations for different worker-pools. Eg. MachineDrainTimeout, MachineHealthTimeout.
+	// +optional
+	Settings *Settings `json:"settings,omitempty"`
```
rfranzke: Why don't you simply reuse the `gardencorev1beta1` struct here?

pkg/operation/botanist/worker.go
```diff
@@ -119,6 +119,15 @@ func (b *Botanist) DeployWorker(ctx context.Context) error {
...
 		}
 
+		settings := &extensionsv1alpha1.Settings{}
+		if workerPool.Settings != nil {
+			settings.MachineDrainTimeout = workerPool.Settings.MachineDrainTimeout
+			settings.MachineHealthTimeout = workerPool.Settings.MachineHealthTimeout
+			settings.MachineCreationTimeout = workerPool.Settings.MachineCreationTimeout
+			settings.MaxEvictRetries = workerPool.Settings.MaxEvictRetries
+			settings.NodeConditions = workerPool.Settings.NodeConditions
+		}
```
rfranzke: See comment above; if you would reuse `gardencorev1beta1` struct you can simply pass the `workerPool.Settings` in line 151

rfranzke (changes requested):

extensions/pkg/controller/worker/genericactuator/interface.go
hardikdr: Sure, I am fine either way. Agree that if we want to be consistent, we can simply move the logic to the providers, and not touch the generic actuator. I'll remove the code changes from here then.

hardikdr: I removed the changes from the genericactuator, and will follow-up on respected providers to adopt it. Ref: https://github.com/gardener/gardener/pull/2563/commits/ff6ab8d206cec92c0fba7529c268b33c85a853e7

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
hardikdr: This code-path is removed based on this comment: https://github.com/gardener/gardener/pull/2563#discussion_r457007752 .

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
hardikdr: This code-path is removed based on this comment: https://github.com/gardener/gardener/pull/2563#discussion_r457007752 .

example/90-shoot.yaml
hardikdr: Hm, agree with the reasoning. I'll change it to `machineControllerManager` then.

hardikdr: Though wondering if it would be end-user friendly, open for other suggestions as well meanwhile :)

pkg/apis/extensions/v1alpha1/types_worker.go
hardikdr: Yes, that's a good idea, I'll reuse the struct here.

hardikdr: done. ref: https://github.com/gardener/gardener/commit/9aacdeb0348a4d06173bb54d71af652807da0846#diff-5865efa9645384006d2ea212587c1ce4R135

pkg/operation/botanist/worker.go
hardikdr: Yes, that's a great suggestion, thanks. done - https://github.com/gardener/gardener/commit/9aacdeb0348a4d06173bb54d71af652807da0846#diff-873d79d4ef156d5ab5a81e3dce9a0a41R142

pkg/apis/core/v1beta1/types_shoot.go
hardikdr: done - https://github.com/gardener/gardener/commit/9aacdeb0348a4d06173bb54d71af652807da0846#diff-9a667891729b65e0ebbf3feef920fcd9R761 . Please suggest, if anything specific need to be added.

pkg/apis/core/v1alpha1/types_shoot.go
hardikdr: done - https://github.com/gardener/gardener/commit/9aacdeb0348a4d06173bb54d71af652807da0846#diff-9a667891729b65e0ebbf3feef920fcd9R761 . Please suggest, if anything specific need to be added.

pkg/apis/core/types_shoot.go
hardikdr: done - https://github.com/gardener/gardener/commit/9aacdeb0348a4d06173bb54d71af652807da0846#diff-9a667891729b65e0ebbf3feef920fcd9R761 . Please suggest, if anything specific need to be added.

pkg/apis/core/types_shoot.go
hardikdr: done - https://github.com/gardener/gardener/commit/9aacdeb0348a4d06173bb54d71af652807da0846#diff-9a667891729b65e0ebbf3feef920fcd9R762 . Thanks for pointing out. :)

example/90-shoot.yaml
hardikdr: I have renamed it to `machineControllerManager`, and the struct to `MachineControllerManagerSettings` just to be clearer. ref: https://github.com/gardener/gardener/commit/9aacdeb0348a4d06173bb54d71af652807da0846#diff-9a667891729b65e0ebbf3feef920fcd9R761

hardikdr:
I also just realized, I need to adapt the validation, I'll do it soon.

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -944,6 +944,28 @@ type Worker struct {
...
+	MachineHealthTimeout *metav1.Duration `json:"machineHealthTimeout,omitempty" protobuf:"bytes,2,name=machineHealthTimeout"`
+	// MachineCreationTimeout is the period after which creation of the machine is declared failed.
+	// +optional
+	MachineCreationTimeout *metav1.Duration `json:"machineCreationTimeout,omitempty" protobuf:"bytes,3,name=machineCreationTimeout"`
+	// MaxEvictRetries are the number of eviction retries on a pod after which drain is declared failed, and forceful deletion is triggered.
+	// +optional
+	MaxEvictRetries *int32 `json:"maxEvictRetries,omitempty" protobuf:"bytes,4,name=maxEvictRetries"`
+	// NodeConditions are the set of conditions if set to true for the period of MachineHealthTimeout, machine will be declared failed.
+	// +optional
+	NodeConditions *string `json:"nodeConditions,omitempty" protobuf:"bytes,5,name=nodeConditions"`
```
rfranzke: Wouldn't `[]string` be better API-wise/more convenient for end-users?

rfranzke (changes requested):
Nice PR :)

* As you already suggested yourself, please add some validation code + tests for the new API fields.
* You could also extend https://github.com/gardener/gardener/blob/master/docs/extensions/worker.md with information about these new fields so that extension developers know what to do.
* As you now plan to do it similar like for labels, annotations, and taints, please adapt https://github.com/gardener/gardener/blob/master/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go#L280-L317. 

Other than that no further comments. Happy to take a look at the PRs for the adaptation of the worker extensions.

pkg/apis/core/v1alpha1/types_shoot.go
hardikdr: I guess, []string would make sense. I kept it as a `string` here, as this is also a string in MCM being a command-line flag[as well], to avoid conversion in-between. I'd be happy to make it []string, if we agree.

hardikdr: done, converted to `[]string` - https://github.com/gardener/gardener/pull/2563/commits/74a036f812bbd8c04d94186c86c8a3a4ab136787#diff-9a667891729b65e0ebbf3feef920fcd9R772

hardikdr:
>As you now plan to do it similar like for labels, annotations, and taints, please adapt https://github.com/gardener/gardener/blob/master/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go#L280-L317.

Done, adapted the configurations in generic actuator: https://github.com/gardener/gardener/pull/2563/commits/74a036f812bbd8c04d94186c86c8a3a4ab136787#diff-ce16933f7218046c410bbaed72f771b7R312 .

hardikdr:
>You could also extend https://github.com/gardener/gardener/blob/master/docs/extensions/worker.md with information about these new fields so that extension developers know what to do.

Done, mentioned it briefly - https://github.com/gardener/gardener/pull/2563/commits/cc262f2df2d46626fe41660300030a74871cf532#diff-0c824cfb79147ede22dc6ef963f86201R99

>As you already suggested yourself, please add some validation code + tests for the new API fields.

I later realized we want to allow users to have any time-duration/values for these settings. Don't know what kind of validations would make sense, open for suggestions.

extensions/pkg/controller/worker/machines.go
```diff
@@ -42,17 +42,18 @@ func init() {
...
+	SecretName           string
+	Minimum              int32
+	Maximum              int32
+	MaxSurge             intstr.IntOrString
+	MaxUnavailable       intstr.IntOrString
+	Labels               map[string]string
+	Annotations          map[string]string
+	Taints               []corev1.Taint
+	State                *MachineDeploymentState
+	MachineConfiguration machinev1alpha1.MachineConfiguration
```
rfranzke:
```suggestion
	MachineConfiguration *machinev1alpha1.MachineConfiguration
```

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -309,6 +309,7 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, cluster
 								Taints: deployment.Taints,
 							},
 						},
+						MachineConfiguration: &deployment.MachineConfiguration,
```
rfranzke:
```suggestion
						MachineConfiguration: deployment.MachineConfiguration,
```

rfranzke (approved):
/lgtm

rfranzke:
@danielfoehrKn pls review again as your earlier review is blocking the merge

danielfoehrKn (approved):

PR #2564, opened 2020-07-07
Fix restore of extension objects during migration.

swilen-iwanow:

/area control-plane-migration
/kind bug
/priority normal

With "componentization" the botanist, restoration of extension objects during migration was broken. 
Previously the extension resources were created by botanist and the botanist had the latest object state, that was passed to the restore function. Now an empty objects are created and passed to the restoreExtensionObject ((botanist.Botanist).DeployNetwork) func that causes a nil pointer dereference error.
 
Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/migration.go
+++ b/pkg/operation/botanist/migration.go
@@ -95,6 +95,10 @@ func (b *Botanist) applyFuncToAllExtensionCRs(ctx context.Context, applyFunc fun
 }
 
 func (b *Botanist) restoreExtensionObject(ctx context.Context, client client.Client, extensionObj extensionsv1alpha1.Object, resourceKind string) error {
+	if err := client.Get(ctx, kutil.KeyFromObject(extensionObj), extensionObj); err != nil {
+		return err
+	}
+
 	if err := b.restoreExtensionObjectState(ctx, client, extensionObj, resourceKind, extensionObj.GetName(), extensionObj.GetExtensionSpec().GetExtensionPurpose()); err != nil {
 		return err
 	}
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -525,7 +525,7 @@ func (b *Botanist) DeployControlPlane(ctx context.Context) error {
 	}
 
 	if restorePhase {
-		return b.restoreExtensionObject(ctx, b.K8sSeedClient.Client(), cp, extensionsv1alpha1.ControlPlaneResource)
+		return b.restoreExtensionObject(ctx, b.K8sSeedClient.DirectClient(), cp, extensionsv1alpha1.ControlPlaneResource)
 	}
 
 	return nil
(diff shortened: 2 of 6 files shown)


swilen-iwanow:
@plkokanov

pkg/operation/botanist/migration.go
```diff
@@ -95,6 +95,10 @@ func (b *Botanist) applyFuncToAllExtensionCRs(ctx context.Context, applyFunc fun
 }
 
 func (b *Botanist) restoreExtensionObject(ctx context.Context, client client.Client, extensionObj extensionsv1alpha1.Object, resourceKind string) error {
```
timuthy, line 97: I wonder if the caller is supposed to pass the direct client if we rely on the lines below or if the cache is still "fast" enough. /cc @tim-ebert

stoyanr: @timuthy What do you mean by "fast" enough? If the cache is started at the point of the call, then "fastness" shouldn't matter, or do I misunderstand how it works? And vice versa, if it's not started, then it has to be the direct client.

stoyanr (dismissed):

pkg/operation/botanist/migration.go
timuthy: > If the cache is started at the point of the call, then "fastness" shouldn't matter, or do I misunderstand how it works?

We cannot assume that the cache is started at the point of the call. @tim-ebert introduced the cached client w/ the client map (https://github.com/gardener/gardener/pull/2449). With that, a backing cache might have already been started, especially if it's a seed client.

Let's take the network resource as an example:
- The network component uses the direct client https://github.com/gardener/gardener/blob/da23019a0af1921f6c5f83a35b24891ed9cbc163/pkg/operation/botanist/botanist.go#L88 to deploy the network resource to the seed cluster.
- The restore operation uses the cached seed client from the `Botanist/Operation`  https://github.com/gardener/gardener/blob/da23019a0af1921f6c5f83a35b24891ed9cbc163/pkg/operation/botanist/network.go#L56
- And now with this PR we try to get the network resource from the cache but the resource itself was previously created/updated with the direct client. So I doubt that the changes are reflected reflected in the cache and the [get call](https://github.com/gardener/gardener/blob/6d73f70f088a34ee289bd753c69fd98290b7aac4/pkg/operation/botanist/migration.go#L98) will either result in an 404 error or the restore function tries to update an already outdated network resource (-> resource version conflict).

stoyanr: @timuthy If you are right, this makes me a bit nervous. How is one supposed to use these clients correctly? All this seems quite counterintuitive to me, and errors are likely to continue creeping in. @tim-ebert Could you please clarify?

stoyanr: @timuthy You are indeed right, we were able to reproduce issues with resource version conflicts of extension resources while testing. The issues disappeared once we replaced `Client()` with `DirectClient()`. Again, some advice how to avoid this confusion in the future would be great.

timebertt: Yes, @timuthy is right and as you have already figured out, we should indeed use the `DirectClient` for such functions that are critical to the shoot flow or at least send updates that often occur simultaneously to others (e.g. update shoot status progress).

We are still gaining some first experience with the cached client and its also still behind a feature gate. So there are no clear guidelines/rules yet when to use which client and so on. But I guess, what I mentioned before is a good rule of thumb for the time being.

But as you said, this is not really intuitive and also not clear to developers, so I still want to add a bit of documentation on this topic (like already mentioned a few weeks back in #2414), but I just haven't been able to do it yet. Sorry for that.

timuthy (approved):
/lgtm

timebertt (approved):
/lgtm

PR #2566, opened 2020-07-07
Update npd to version v0.8.2

majst01:
update node-problem-detector to v0.8.2 to fix missing hard drive failure detection

/area robustness
/area monitoring
/priority normal

@see https://github.com/kubernetes/node-problem-detector/releases for release notes.

### DIFF
--- a/charts/shoot-core/components/charts/node-problem-detector/templates/daemonset.yaml
+++ b/charts/shoot-core/components/charts/node-problem-detector/templates/daemonset.yaml
@@ -60,6 +60,9 @@ spec:
             - name: localtime
               mountPath: /etc/localtime
               readOnly: true
+            - name: kmsg
+              mountPath: /dev/kmsg
+              readOnly: true
             - name: custom-config
               mountPath: /custom-config
               readOnly: true
@@ -98,6 +101,9 @@ spec:
           hostPath:
             path: /etc/localtime
             type: "FileOrCreate"
+        - name: kmsg
+          hostPath:
+            path: /dev/kmsg
         - name: custom-config
           configMap:
             name: node-problem-detector-custom-config
--- a/charts/shoot-core/components/charts/node-problem-detector/values.yaml
+++ b/charts/shoot-core/components/charts/node-problem-detector/values.yaml
@@ -45,7 +45,7 @@ hostpath:
 
 # image:
 #   repository: k8s.gcr.io/node-problem-detector
-#   tag: v0.7.1
+#   tag: v0.8.2
 #   pullPolicy: IfNotPresent
 
 images:
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -135,7 +135,7 @@ images:
 - name: node-problem-detector
   sourceRepository: github.com/kubernetes/node-problem-detector
   repository: k8s.gcr.io/node-problem-detector
-  tag: "v0.7.1"
+  tag: "v0.8.2"
 
 # Shoot optional addons
 - name: kubernetes-dashboard
--- a/charts/shoot-core/components/charts/node-problem-detector/Chart.yaml
+++ b/charts/shoot-core/components/charts/node-problem-detector/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: node-problem-detector
 version: "1.6.1"
-appVersion: v0.7.1
+appVersion: v0.8.2
 home: https://github.com/kubernetes/node-problem-detector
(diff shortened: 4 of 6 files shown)


hardikdr:
/lgtm from changes pov.
Haven't been able to test the happy-path though, can do it later this week, or else please update if already done.

majst01:
@hardikdr we have tested it here by overriding it with the gardenlet image-vector, looks good so far.

ialidzhikov:
Afaik the npd chart was copied from helm/charts. So there are no changes in the chart itself between this version range?

majst01:
@ialidzhikov thanks for the hint, in fact there is one more volume `/dev/kmsg` and NODE_NAME env, added these from upstream

charts/shoot-core/components/charts/node-problem-detector/templates/daemonset.yaml
```diff
@@ -98,6 +101,9 @@ spec:
           hostPath:
             path: /etc/localtime
             type: "FileOrCreate"
+        - name: kmsg
```
timebertt, line 104: I think, the PSP has to be adapted to this change, right?
https://github.com/gardener/gardener/blob/78ef5841ad5a0d9ba405a5ccb66b0e9869a9f390/charts/shoot-core/components/charts/node-problem-detector/templates/psp.yaml#L25-L27

majst01: huh, thanks for pointing out

charts/shoot-core/components/charts/node-problem-detector/values.yaml
```diff
@@ -86,6 +86,10 @@ affinity: {}
 nodeSelector: {}
 
 env:
+  - name: NODE_NAME
+    valueFrom:
+      fieldRef:
+        fieldPath: spec.nodeName
```
timebertt: We already have the env var here:
https://github.com/gardener/gardener/blob/2e993388559b10582f1fc00623cb371b5262df65/charts/shoot-core/components/charts/node-problem-detector/templates/daemonset.yaml#L49-L53
Why is this added a second time in the values?

This way, the env var will be added twice 😄 
```yaml
        env:
        - name: NODE_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
        - name: NODE_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
```

timebertt (changes requested):
Thanks for the PR, only one nit.
Tested and validated successfully, that NPD is running fine on an unprivileged cluster. :)

Can you please squash your commits then to get the PR ready for merging?

timebertt:
Actually, it would be really nice to have some kind of integration test for this kind of stuff like the NPD.
The test could provoke issues on the Nodes, that the NPD should detect and care controller should pick up and reflect in the Shoot status.
Ref #1564

timuthy:
> Actually, it would be really nice to have some kind of integration test for this kind of stuff like the NPD.
The test could provoke issues on the Nodes, that the NPD should detect and care controller should pick up and reflect in the Shoot status.

+1 but let's do it in a separate PR to not block this one from getting merged.

timebertt:
> > Actually, it would be really nice to have some kind of integration test for this kind of stuff like the NPD.
> > The test could provoke issues on the Nodes, that the NPD should detect and care controller should pick up and reflect in the Shoot status.
> 
> +1 but let's do it in a separate PR to not block this one from getting merged.

Sure, didn't want to imply that we should do it in this PR. Just a side-note :)

majst01:
I am to stupid to squash, even googling makes me sad :-(

timebertt (dismissed):
Thanks!
/lgtm now

timebertt:
> I am to stupid to squash, even googling makes me sad :-(

Don't be sad :)
The easy option would be:
1. `git reset --soft master`
2. `git commit`
3. `git push -f fork`

Or you can go for:
1. `git rebase -i master`
2. Edit the opened file use `pick` for the first commit (the original one) and replace all `pick`s of consecutive commits with `squash`/`s`.
3. Do a final edit of the commit message
4. `git push -f fork`

Does this help?

majst01:
@tim-ebert Thanks! this was the best explanation so far.

timebertt (approved):
Nice! :)
/lgtm
/ok-to-test

majst01:
Thanks @tim-ebert

PR #2567, opened 2020-07-07
Increase default QPS for Seed Clients

timebertt:

/area quality
/kind enhancement
/priority normal

This PR increases the default QPS value for Seed Clients.
Bevor my refactoring in #2449, every Shoot operation created its own Seed client, so the configured QPS setting in the GardenletConfiguration was actually applied only to each individual Seed client (per Shoot operation). That in turn means, that in sum the maximum QPS was actually not respected when talking to the Seed's API server.

This is a problem when many Shoot's are created in parallel and the gardenlet sends a lot of requests to the Seed's API server.

/invite @ialidzhikov 
/cc @zanetworker 

**Release note**:

```improvement operator
The default QPS setting for Seed Clients in the gardenlet has been increased to adapt to the actual amount concurrent of API calls.
```

### DIFF
--- a/charts/gardener/gardenlet/values.yaml
+++ b/charts/gardener/gardenlet/values.yaml
@@ -54,8 +54,8 @@ global:
       seedClientConnection:
         acceptContentTypes: application/json
         contentType: application/json
-        qps: 25
-        burst: 50
+        qps: 100
+        burst: 130
       # kubeconfig: |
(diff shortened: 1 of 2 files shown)


ialidzhikov (approved):
/lgtm

timuthy (approved):
/lgtm

PR #2568, opened 2020-07-07
Use DirectClient in BB/BE reconciler

timebertt:

/area quality
/kind bug
/priority normal

If the `CachedRuntimeClients` feature gate is enabled, the backup entry/bucket reconciler in the gardenlet might not execute any reconciliation the first time the backup bucket/entry is deployed.
That's because we are using an Informer created by the SharedInformerFactory [here](https://github.com/tim-ebert/gardener/blob/379bf51c9251722cae7ca23c33417ea4e3c3c9df/pkg/gardenlet/controller/backupbucket/backup_bucket.go#L76-L83) for receiving watch events, but we are actually using a cached runtime client for retrieving the object [in the reconciler](https://github.com/tim-ebert/gardener/blob/379bf51c9251722cae7ca23c33417ea4e3c3c9df/pkg/gardenlet/controller/backupbucket/backup_bucket_reconciler.go#L73) which uses a separate watch. This in turn means, that the cached runtime client might not have received the add event for a backup bucket/entry, while the Informer already has received an event and enqueued the object for reconciliation. The cached runtime client the gets a `NotFound` error and the reconciler exits early. 

Fixes #

In general, we should always use the same informers for registering event handlers which we also use in our controllers to retrieve the object in the beginning of the reconciliation. For example an Informer from the `SharedInformerFactory` and the respective `Lister` or an Informer from the controller-runtime cache and the cached runtime client.
I will do this refactoring in a separate PR, as it requires to not invalidate the Garden ClientSet in any case (also for version changes) which I will have to tackle first.
/cc @timuthy 

**Release note**:

```improvement operator
A bug has been fixed, that caused that newly created `BackupBuckets`/`BackupEntries` where not reconciled by the `gardenlet` immediately when the `CachedRuntimeClients` feature gate was enabled.
```

### DIFF
--- a/pkg/gardenlet/controller/backupbucket/backup_bucket_reconciler.go
+++ b/pkg/gardenlet/controller/backupbucket/backup_bucket_reconciler.go
@@ -70,7 +70,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 
 	bb := &gardencorev1beta1.BackupBucket{}
-	if err := gardenClient.Client().Get(r.ctx, request.NamespacedName, bb); err != nil {
+	if err := gardenClient.DirectClient().Get(r.ctx, request.NamespacedName, bb); err != nil {
 		if apierrors.IsNotFound(err) {
 			r.logger.Debugf("[BACKUPBUCKET RECONCILE] %s - skipping because BackupBucket has been deleted", request.NamespacedName)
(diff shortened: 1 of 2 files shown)


timebertt:
Probably we should cherry-pick this to `release-v1.7`.

timuthy (approved):
/lgtm

PR #2570, opened 2020-07-08
Fix machine image defaulting

### LINKED ISSUE #2569: Machine Image downgraded during apply

/area quality
/kind bug
/priority normal

**What happened**:
We have some Shoots that are applied without a machine image version and were previously defaulted to a preview version. This was fixed in #2534.

But now that the Shoots are applied again without a machine image version, they are defaulted to the non-preview version which means a machine image version downgrade, which should never happen.

Shoot versions not to be downgraded for existing cluster, even though a new apply doesn't specify the version.

- Create a new Shoot with a preview machine image version
- Apply the Shoot without any version
- See the machine image version get downgraded to a non-preview version.

/cc @danielfoehrKn 

**Environment**:

- Gardener version: v1.7.0
- Others:

timebertt:

/area quality
/kind bug
/priority normal

This PR fixes the machine image defaulting in the case, that the Shoot is updated with a machine image name set but version unset.
I also restructured the Shoot validator test along the way (first commit) to make it a bit more easy to overlook.

Fixes #2569

/invite @ialidzhikov 
/cc @vpnachev @danielfoehrKn 

**Release note**:

```improvement operator
A bug was fixed that caused the machine image version to be overwritten in case a Shoot was updated with a specified image name but with an unspecified image version.
```

### DIFF
--- a/plugin/pkg/shoot/validator/admission.go
+++ b/plugin/pkg/shoot/validator/admission.go
@@ -772,9 +772,26 @@ func ensureMachineImage(oldWorkers []core.Worker, worker core.Worker, images []c
 	// This should only happen in the maintenance time window of shoots and is performed by the
 	// shoot maintenance controller.
 
-	name := worker.Name
-	if oldWorker := helper.FindWorkerByName(oldWorkers, name); worker.Machine.Image == nil && oldWorker != nil && oldWorker.Machine.Image != nil {
-		return oldWorker.Machine.Image, nil
+	oldWorker := helper.FindWorkerByName(oldWorkers, worker.Name)
+	if oldWorker != nil && oldWorker.Machine.Image != nil {
+		// worker is already existing -> keep the machine image if name/version is unspecified
+		if worker.Machine.Image == nil {
+			// machine image completely unspecified in new worker -> keep the old one
+			return oldWorker.Machine.Image, nil
+		}
+
+		if oldWorker.Machine.Image.Name == worker.Machine.Image.Name {
+			// image name was not changed -> keep version from the new worker if specified, otherwise use the old worker image version
+			if len(worker.Machine.Image.Version) != 0 {
+				return worker.Machine.Image, nil
+			}
+			return oldWorker.Machine.Image, nil
+		} else {
+			// image name was changed -> keep version from new worker if specified, otherwise default the image version
+			if len(worker.Machine.Image.Version) != 0 {
+				return worker.Machine.Image, nil
+			}
+		}
 	}
 
 	imageName := ""
--- a/plugin/pkg/shoot/validator/admission_test.go
+++ b/plugin/pkg/shoot/validator/admission_test.go
@@ -686,46 +686,59 @@ var _ = Describe("validator", func() {
 			})
 		})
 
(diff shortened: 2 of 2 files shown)


plugin/pkg/shoot/validator/admission.go
```diff
@@ -772,9 +772,26 @@ func ensureMachineImage(oldWorkers []core.Worker, worker core.Worker, images []c
...
+	oldWorker := helper.FindWorkerByName(oldWorkers, worker.Name)
+	if oldWorker != nil && oldWorker.Machine.Image != nil {
+		// worker is already existing -> keep the machine image if name/version is unspecified
+		if worker.Machine.Image == nil {
+			// machine image completely unspecified in new shoot -> keep the old one
+			return oldWorker.Machine.Image, nil
+		}
+
+		if oldWorker.Machine.Image.Name == worker.Machine.Image.Name {
+			// image name was not changed -> keep version from new shoot if specified, otherwise use
```
ialidzhikov:
```suggestion
			// image name was not changed -> keep version from the new worker if specified, otherwise use the old worker image version
```

plugin/pkg/shoot/validator/admission.go
```diff
@@ -772,9 +772,26 @@ func ensureMachineImage(oldWorkers []core.Worker, worker core.Worker, images []c
...
 	// shoot maintenance controller.
 
-	name := worker.Name
-	if oldWorker := helper.FindWorkerByName(oldWorkers, name); worker.Machine.Image == nil && oldWorker != nil && oldWorker.Machine.Image != nil {
-		return oldWorker.Machine.Image, nil
+	oldWorker := helper.FindWorkerByName(oldWorkers, worker.Name)
+	if oldWorker != nil && oldWorker.Machine.Image != nil {
+		// worker is already existing -> keep the machine image if name/version is unspecified
+		if worker.Machine.Image == nil {
+			// machine image completely unspecified in new shoot -> keep the old one
```
ialidzhikov:
```suggestion
			// machine image completely unspecified in new worker -> keep the old one
```

plugin/pkg/shoot/validator/admission.go
```diff
@@ -772,9 +772,26 @@ func ensureMachineImage(oldWorkers []core.Worker, worker core.Worker, images []c
...
+		}
+
+		if oldWorker.Machine.Image.Name == worker.Machine.Image.Name {
+			// image name was not changed -> keep version from new shoot if specified, otherwise use
+			if len(worker.Machine.Image.Version) != 0 {
+				return worker.Machine.Image, nil
+			}
+			return oldWorker.Machine.Image, nil
+		} else {
+			// image name was changed -> keep version from new shoot if specified, otherwise default the image version
```
ialidzhikov:
```suggestion
			// image name was changed -> keep version from new worker if specified, otherwise default the image version
```

ialidzhikov (commented):
Thank you for this PR!

timebertt:
@ialidzhikov thanks, addressed your comment! PTAL :)

ialidzhikov (approved):
/lgtm

PR #2574, opened 2020-07-10
Support resource reservation in kubelet

### LINKED ISSUE #2559: Resource reservation for kube and system daemons

/area robustness
/kind enhancement
/priority normal

[Kubelet configuration](https://godoc.org/k8s.io/kubelet/config/v1beta1) can be used to [reserve resources for kube and system daemons](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/). Reserved resources are deducted from the node's capacity. Node allocatable resources can also be enforced by kubelet on three cgroups: system (eg. `system.slice`), kube (eg. `runtime.slice`) and pods (`kubepods`). If resources are enforced, relevant cgroup controllers are configured (`cpu.shares`, `memory.limit_in_bytes`, ...)

The shoot worker kubelet config should allow configuration of the following attributes: `SystemReserved`, `KubeReserved`. 

**Why is this needed**:
Currently, gardener does not support resource reservation, and the following [default configuration](https://github.com/gardener/gardener/blob/master/charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet-config-file) is used:

```
enforceNodeAllocatable:
- pods
kubeReserved:
  cpu: 80m
  memory: 1Gi
```

This configuration deducts `80m` and `1Gi` from node's capacity, and enforces the allocatable resources on the `kubepods` cgroup. For example, on a 8 CPU, 32 GB node, the `kubepods` cgroup would be configured with 8110 `cpu.shares` and 31769579520 `memory.limit_in_bytes`.  

In gardener nodes, kubelet and docker run as systemd services under the `system.slice` cgroup, which is configured with default values (eg. 1024 `cpu.shares`). The existing configuration can lead to starvation of kube components on nodes with high pod density and burstable, resource-intensive workloads: pods are not limited from consuming most of the node's resources and 
appropriate resources cannot be assigned to the kube/system cgroups.

guydc:
/area robustness
/kind enhancement
/priority normal

Exposes kubelet configuration for resource reservation: https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/. Defaults remain the same. 

Fixes https://github.com/gardener/gardener/issues/2559

**Release note:**

```improvement user
Users can now specify values for `systemReserved` and `kubeReserved` in the kubelet configuration as documented here: https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/
```


pkg/apis/core/types_shoot.go
```diff
@@ -600,6 +615,30 @@ type KubeletConfigEvictionSoftGracePeriod struct {
...
+type KubeletConfigSystemReserved struct {
+	// CPU is the system-reserved cpu.
+	CPU *resource.Quantity
+	// Memory is the system-reserved memory.
+	Memory *resource.Quantity
+	// EphemeralStorage is the system-reserved ephemeral-storage.
+	EphemeralStorage *resource.Quantity
+	// PID is the system-reserved process-ids.
+	PID *int64
+}
```
mvladev: Can we make those things the same as the `kubelet`[config](https://github.com/kubernetes/kubernetes/blob/9016740a6ffe91bb29824f80c34087b993903bd6/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go#L618-L645) (being `map[string]string`)?

guydc: Well, this implementation is consistent with the existing gardener examples for [kubelet params](https://github.com/gardener/gardener/blob/master/pkg/apis/core/types_shoot.go#L562)  which are also modeled as `map[string]string` in [kubelet apis](https://github.com/kubernetes/kubernetes/blob/9016740a6ffe91bb29824f80c34087b993903bd6/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go#L486). It also makes validation convenient.

I'm willing to make the change, but can you explain why you prefer this approach? Do you have future plans to embed the k8s kubeletconfig type in the worker spec? If so, do you think that validation is not required here, and it's fine to just let kubelet validate the config?

danielfoehrKn: Agree with @guydaichs , convenient validation is the reason why we are using concrete types in the other kubelet configurations. 
I would opt for keeping it the way it is. But lets wait for @mvladev response

pkg/apis/core/validation/shoot.go
```diff
@@ -1029,6 +1038,88 @@ func validateKubeletConfigEvictionSoftGracePeriod(eviction *core.KubeletConfigEv
 	return allErrs
 }
 
+// Reserved PIDs have to be non-negative
+const ReservedPIDsMinimum int64 = 0
+
+func validateKubeletConfigKubeReserved(reserved *core.KubeletConfigKubeReserved, fldPath *field.Path) field.ErrorList {
```
danielfoehrKn: WDYT about introducing checks based on the machine type? 
Enforcing some sane defaults, so that Shoots cannot be created which nodes are garuanteed to not be healthy.
For instance reserving more kubelet, system CPU & Memory than available according to the machine type.

Basically making it harder for the Shoot creator to "shoot" himself in the foot.

danielfoehrKn (commented):
Looks good the PR, thanks! Having a remark already.

pkg/apis/core/types_shoot.go
```diff
@@ -556,6 +556,21 @@ type KubeletConfig struct {
...
+	//   memory:	1Gi
+	//   cpu:		80m
+	KubeReserved *KubeletConfigKubeReserved
+	// SystemReserved is the configuration for resources reserved for system components.
+	SystemReserved *KubeletConfigSystemReserved
+	// EnforceNodeAllocatable is an array of cgroups whose resource would be enforced by kubelet.
+	// Default:
+	//   - pods
+	EnforceNodeAllocatable []string
+	// KubeReservedCgroup is the name of the cgroup that kubelet will enforce kube-reserved resource on.
```
timuthy: // The configured cgroup is not automatically created and thus must exist on the worker node before using this option.

pkg/apis/core/types_shoot.go
```diff
@@ -556,6 +556,21 @@ type KubeletConfig struct {
...
+	KubeReservedCgroup *string
+	// SystemReservedCgroup is the name of the cgroup that kubelet will enforce system-reserved resource on.
```
timuthy: // The configured cgroup is not automatically created and thus must exist on the worker node before using this option.

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -698,6 +698,26 @@ type KubeletConfig struct {
...
+	KubeReserved *KubeletConfigKubeReserved `json:"kubeReserved,omitempty" protobuf:"bytes,14,opt,name=kubeReserved"`
+	// SystemReserved is the configuration for resources reserved for system components.
+	// +optional
+	SystemReserved *KubeletConfigSystemReserved `json:"systemReserved,omitempty" protobuf:"bytes,15,opt,name=systemReserved"`
+	// EnforceNodeAllocatable is an array of cgroups whose resource would be enforced by kubelet.
+	// +optional
+	// Default:
+	//   - pods
+	EnforceNodeAllocatable []string `json:"enforceNodeAllocatable,omitempty" protobuf:"bytes,16,opt,name=enforceNodeAllocatable"`
+	// KubeReservedCgroup is the name of the cgroup that kubelet will enforce kube-reserved resource on.
```
timuthy: // The configured cgroup is not automatically created and thus must exist on the worker node before using this option.

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -698,6 +698,26 @@ type KubeletConfig struct {
...
+	// +optional
+	KubeReservedCgroup *string `json:"kubeReservedCgroup,omitempty" protobuf:"bytes,17,opt,name=kubeReservedCgroup"`
+	// SystemReservedCgroup is the name of the cgroup that kubelet will enforce system-reserved resource on.
```
timuthy: // The configured cgroup is not automatically created and thus must exist on the worker node before using this option.

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -695,6 +695,26 @@ type KubeletConfig struct {
...
+	KubeReserved *KubeletConfigKubeReserved `json:"kubeReserved,omitempty" protobuf:"bytes,14,opt,name=kubeReserved"`
+	// SystemReserved is the configuration for resources reserved for system components.
+	// +optional
+	SystemReserved *KubeletConfigSystemReserved `json:"systemReserved,omitempty" protobuf:"bytes,15,opt,name=systemReserved"`
+	// EnforceNodeAllocatable is an array of cgroups whose resource would be enforced by kubelet.
+	// +optional
+	// Default:
+	//   - pods
+	EnforceNodeAllocatable []string `json:"enforceNodeAllocatable,omitempty" protobuf:"bytes,16,opt,name=enforceNodeAllocatable"`
+	// KubeReservedCgroup is the name of the cgroup that kubelet will enforce kube-reserved resource on.
```
timuthy: // The configured cgroup is not automatically created and thus must exist on the worker node before using this option.

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -695,6 +695,26 @@ type KubeletConfig struct {
...
+	// +optional
+	KubeReservedCgroup *string `json:"kubeReservedCgroup,omitempty" protobuf:"bytes,17,opt,name=kubeReservedCgroup"`
+	// SystemReservedCgroup is the name of the cgroup that kubelet will enforce system-reserved resource on.
```
timuthy: // The configured cgroup is not automatically created and thus must exist on the worker node before using this option.

pkg/apis/core/validation/shoot.go
```diff
@@ -1029,6 +1038,88 @@ func validateKubeletConfigEvictionSoftGracePeriod(eviction *core.KubeletConfigEv
...
+		enforcePods   bool
+		enforceKube   bool
+		enforceSystem bool
+	)
+	if len(nodeAllocatable) > 0 {
+		for idx := range nodeAllocatable {
+			switch allocatable := nodeAllocatable[idx]; allocatable {
+			case PodsAllocatableName:
+				if enforcePods {
+					allErrs = append(allErrs, field.Invalid(fldPath.Index(idx), allocatable, "Enforced Node Allocatable should not be repeated"))
```
```suggestion
					allErrs = append(allErrs, field.Duplicate(fldPath.Index(idx), PodsAllocatableName))
```
Also for the duplicate value checks below.

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -754,6 +774,38 @@ type KubeletConfigEvictionSoftGracePeriod struct {
 	NodeFSInodesFree *metav1.Duration `json:"nodeFSInodesFree,omitempty" protobuf:"bytes,5,opt,name=nodeFSInodesFree"`
 }
 
+// KubeletConfigKubeReserved contains the kube-reserved resources configuration
+type KubeletConfigKubeReserved struct {
+	// CPU is the kube-reserved cpu.
+	// +optional
+	CPU *resource.Quantity `json:"cpu,omitempty" protobuf:"bytes,1,opt,name=cpu"`
```
timuthy: Can you default this value to make it more explicit, i.e. `80m`?

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -754,6 +774,38 @@ type KubeletConfigEvictionSoftGracePeriod struct {
...
+	// Memory is the kube-reserved memory.
+	// +optional
+	Memory *resource.Quantity `json:"memory,omitempty" protobuf:"bytes,2,opt,name=memory"`
```
timuthy: Can you default this value to make it more explicit, i.e. `1Gi`?

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -695,6 +695,26 @@ type KubeletConfig struct {
...
+	//   cpu:		80m
+	KubeReserved *KubeletConfigKubeReserved `json:"kubeReserved,omitempty" protobuf:"bytes,14,opt,name=kubeReserved"`
+	// SystemReserved is the configuration for resources reserved for system components.
+	// +optional
+	SystemReserved *KubeletConfigSystemReserved `json:"systemReserved,omitempty" protobuf:"bytes,15,opt,name=systemReserved"`
+	// EnforceNodeAllocatable is an array of cgroups whose resource would be enforced by kubelet.
+	// +optional
+	// Default:
+	//   - pods
+	EnforceNodeAllocatable []string `json:"enforceNodeAllocatable,omitempty" protobuf:"bytes,16,opt,name=enforceNodeAllocatable"`
```
timuthy: Can you default this value to make it more explicit, i.e. `- pods`?

pkg/apis/core/validation/shoot.go
```diff
@@ -1029,6 +1038,88 @@ func validateKubeletConfigEvictionSoftGracePeriod(eviction *core.KubeletConfigEv
...
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("pid"), *reserved.PID, fmt.Sprintf("PID must be atleast %d", ReservedPIDsMinimum)))
+	}
+	return allErrs
+}
+
+const KubeReservedAllocatableName string = "kube-reserved"
+const SystemReservedAllocatableName string = "system-reserved"
+const PodsAllocatableName string = "pods"
+const KubeReservedCgroupName string = "KubeReservedCgroupName"
+const systemReservedCgroupName string = "SystemReservedCgroupName"
```
```suggestion
const (
  kubeReservedAllocatableName string = "kube-reserved"
  systemReservedAllocatableName string = "system-reserved"
  podsAllocatableName string = "pods"
  kubeReservedCgroupName string = "KubeReservedCgroupName"
  systemReservedCgroupName string = "SystemReservedCgroupName"
)
```

timuthy (changes requested):
Nice PR! Just a few comments and suggestions. Thanks for contributing this feature.

charts/seed-operatingsystemconfig/original/values.yaml
```diff
@@ -59,6 +59,22 @@ worker:
...
+    # systemReserved:
+      # cpu: 80m
+      # memory: 1Gi
+      # pid: 1000
+      # ephemeral-storage: 10Gi
+    enforceNodeAllocatable:
+      - pods
+      # - kube-reserved
+      # - system-reserved
+    # kubeReservedCgroup: runtime.slice
```
danielfoehrKn: Gardener does not (yet?) create the cGroups according to the [recommendation made here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/node-allocatable.md#recommended-cgroups-setup) which would expect a cGroup named "runtime.slice". Instead all the systemd services installed by gardener are currently under system.slice cGroup.

Having this as a default comment here is a problem I think, because when enabled might lead to the kubelet to fail (when allocatable is also enabled for kube-reserved).

example/90-shoot.yaml
```diff
@@ -102,6 +102,22 @@ spec:
     #       nodeFSInodesFree: 0Mi
     #     featureGates:
     #       SomeKubernetesFeature: true
+    #     enforceNodeAllocatable:
+    #       - pods
+    #       - kube-reserved
+    #       - system-reserved
+    #     kubeReservedCgroup: runtime.slice
```
danielfoehrKn: See comment above regarding `runtime.slice` cGroup.

pkg/apis/core/validation/shoot.go
danielfoehrKn: As I have also written to you in pn, I think we should be clear which cases of misconfiguration we can be verified by Gardener and which not. And lets also determine if it even makes sense to have validation if there are so many cases that we cannot validate anyways. That I think is also the reason why other Kubernetes Services do not allow configuring these options. 
- cGroup names cannot be validated by Gardener upfront (user might have prepared the Os image in a way that Gardener does not know about). However we could say that the kubelet options `kubeReservedCgroup` and `systemReservedCgroup`  are not exposed and are always the Gardener created cGroups.
- `Kube-reserved` depends on the load on the node (workload dependent) and `system-reserved` depends on the OS. Here we can only validate against edge cases (request more memory than available on the node). Is that validation then useful?

I guess it also depends on the use case that you have in mind. 

Inviting @timuthy @rfranzke  @tim-ebert

charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet-config-file
```diff
@@ -28,7 +28,7 @@ enableControllerAttachDetach: true
 enableDebuggingHandlers: true
 enableServer: true
 enforceNodeAllocatable:
-- pods
+{{ toYaml .Values.worker.kubelet.enforceNodeAllocatable}}
```
timebertt: Nit
```suggestion
{{ toYaml .Values.worker.kubelet.enforceNodeAllocatable }}
```

pkg/apis/core/types_shoot.go
timebertt: I would agree to go with the typed APIs as @guydaichs proposed. 
Though, would it make sense to use the same struct type for `KubeReserved` and `SystemReserved`? I see `KubeletConfigKubeReserved` and `KubeletConfigSystemReserved` are basically the same.

pkg/apis/core/validation/shoot_test.go
```diff
@@ -2758,6 +2758,104 @@ var _ = Describe("Shoot Validation Tests", func() {
 			)),
 		)
 
+		validReservedPID := int64(100)
+		invalidReservedPID := int64(-100)
```
timebertt: Can you move these vars to a `var( )` block (existing or new one in an own `ginkgo.Context`)?

pkg/apis/core/validation/shoot_test.go
```diff
@@ -2758,6 +2758,104 @@ var _ = Describe("Shoot Validation Tests", func() {
...
+				"Field": Equal(field.NewPath("systemReserved.pid").String()),
+			})))),
+		)
+
+		kubeReservedCgroupName := "kubeReservedCgroup"
+		systemReservedCgroupName := "systemReservedCgroup"
+		kubeReservedAllocatableName := "kube-reserved"
+		systemReservedAllocatableName := "system-reserved"
+		podsAllocatableName := "pods"
+		invalidAllocatableName := "not-reserved"
```
timebertt: Same here

pkg/operation/botanist/operatingsystemconfig.go
```diff
@@ -315,6 +352,15 @@ func (b *Botanist) deployOperatingSystemConfigsForWorker(ctx context.Context, ma
 		if failSwapOn := kubeletConfig.FailSwapOn; failSwapOn != nil {
 			kubelet["failSwapOn"] = *failSwapOn
 		}
+		if enforceNodeAllocatable := kubeletConfig.EnforceNodeAllocatable; enforceNodeAllocatable != nil {
```
timebertt:
```suggestion
		if enforceNodeAllocatable := kubeletConfig.EnforceNodeAllocatable; len(enforceNodeAllocatable) > 0 {
```

pkg/apis/core/validation/shoot.go
```diff
@@ -1029,6 +1038,88 @@ func validateKubeletConfigEvictionSoftGracePeriod(eviction *core.KubeletConfigEv
...
+func validateKubeletConfigEnforceNodeAllocatable(nodeAllocatable []string, kubeReservedCgroup, systemReservedCgroup *string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	var (
+		enforcePods   bool
+		enforceKube   bool
+		enforceSystem bool
+	)
+	if len(nodeAllocatable) > 0 {
+		for idx := range nodeAllocatable {
+			switch allocatable := nodeAllocatable[idx]; allocatable {
```
timebertt:
```suggestion
		for idx, allocatable := range nodeAllocatable {
			switch allocatable {
```

timebertt: Anyways, I would probably rather go with a predefined `sets.String` of allowed Node Allocatables (`pods`, `kube-reserved`, `system-reserved`) and build up a set of specified allocatables and check for duplication instead of using a switch, as you basically do the same checks in every `case`. But that's just personal preference, no strong opinion on that one.

pkg/apis/core/types_shoot.go
```diff
@@ -556,6 +556,21 @@ type KubeletConfig struct {
 	ImagePullProgressDeadline *metav1.Duration
 	// FailSwapOn makes the Kubelet fail to start if swap is enabled on the node. (default true).
 	FailSwapOn *bool
+	// KubeReserved is the configuration for resources reserved for kubernetes node components.
+	// Default:
+	//   memory:	1Gi
+	//   cpu:		80m
```
timebertt:
```suggestion
	// KubeReserved is the configuration for resources reserved for kubernetes node components (mainly kubelet and container runtime).
	// Default: cpu=80m,memory=1Gi
```

pkg/apis/core/types_shoot.go
```diff
@@ -556,6 +556,21 @@ type KubeletConfig struct {
...
+	KubeReserved *KubeletConfigKubeReserved
+	// SystemReserved is the configuration for resources reserved for system components.
```
timebertt:
```suggestion
	// SystemReserved is the configuration for resources reserved for system processes not managed by kubernetes (e.g. journald).
```

timebertt (commented):
Thanks for this contribution, PR is already looking quite good!
Only have some minor comments to add to the previous reviews.

Regarding the topic with `{kube,system}ReservedCgroup`:
What do you (and the other reviewers) think about adding a check to the [cloud config script](https://github.com/gardener/gardener/blob/master/charts/shoot-cloud-config/templates/scripts/_cloud-config-script.sh) if the cgroup already exists and create it otherwise?

guydc:
Following this conversation and some offline discussion with @timuthy  and @danielfoehrKn, I intend to limit this PR only to the exposure of `KubeReserved` and `SystemReserved`, without the enforcement-related parameters (`SystemReservedCgroup`, `KubeReservedCgroup`, `EnforceNodeAllocatable`). I leave the node cgroup-management and resource enforcement topics to a future PR. 

The reason for this is:
- The `system.slice` cgroup cannot be used for enforcement of system-reserved resources due to the `cpuset` controller not being configured. This is a [k8s prerequisite](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/cgroup_manager_linux.go#L278).
- The [k8s-recommended](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/node-allocatable.md#recommended-cgroups-setup)  `podruntime` cgroup does not exist on gardener nodes, and docker, containerd and kubelet are not placed in this cgroup. So, there is no proper group that can be used for enforcement of kube-reserved.
- In general, end-users are not in control of the cgroup hierarchy on gardener nodes, so it doesn't make much sense to let users control the names of the enforced cgroups. 
- Changing `SystemReservedCgroup`, `KubeReservedCgroup` [requires a reboot](https://github.com/kubernetes/kubelet/blob/v0.18.5/config/v1beta1/types.go#L732). Enforcement of `KubeReserved`, `SystemReserved` cgroups can fail with kubelet panic, due to cgroup resizing limitations. So, changes to these parameters would require node-rolling. 


Regarding some of the other suggestions: 
- Only obvious validations can be made wrt. resources. For example, reserved resources should not exceed 100% of the node's capacity. Such validations have questionable value, so I personally prefer to not add them. 
- In the reduced scope of this PR, only the `kubepods` cgroup can be resized by changing shoot configuration. [Kubelet is tolerant](https://github.com/kubernetes/kubernetes/blob/v1.18.5/pkg/kubelet/cm/node_container_manager_linux.go#L97) to update failures for the `kubepods` cgroup, and updates are retried. A successful update is indicated by kubelet events and changes to the node allocatable values, failures are also recorded as events. In this context, I think that node-rolling is not necessary in this case.

@timuthy, @danielfoehrKn , @tim-ebert - WDYT ?

danielfoehrKn:
@guydaichs 
> Only obvious validations can be made wrt. resources. For example, reserved resources should not exceed 100% of the node's capacity. Such validations have questionable value, so I personally prefer to not add them.

Ok, I think that makes sense from my point of view.

> In the reduced scope of this PR, only the kubepods cgroup can be resized by changing shoot configuration.  Kubelet is tolerant to update failures for the kubepods cgroup.

Agreed, thanks for paying attention to the details here.

> I intend to limit this PR only to the exposure of KubeReserved and SystemReserved, without the enforcement-related parameters

Agree

Talked to @tim-ebert @timuthy @vpnachev an we generally agreed on the following:

1. **cGroup names** (`kubeReservedCgroup`, `systemReservedCgroup`)
> In general, end-users are not in control of the cgroup hierarchy on gardener nodes, so it doesn't make much sense to let users control the names of the enforced cgroups

Yes, we would prefer to not have the cGroup names configurable in the Shoot.
However Gardener should create the [recommended-cGroup-setup](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/node-allocatable.md#recommended-cgroups-setup) and internally the kubelet should be configured accordingly. 
@guydaichs as far as I know you are already working on that - thanks!

2. **Enforcement of  `NodeAllocatable`  - `KubeReserved`, `SystemReserved` cgroups**

For example:
```
enforceNodeAllocatable:
    - kube-reserved 
    - system-reserved
kubeReserved: 
     cpu: 100m
     memory: 1Gi 
     ephemeralStorage: 100Mi
     pid: 100
```

As you said, changing these parameters on a running worker node can lead to a kubelet panic (Node Not ready) and is therefore not safe. We prefer having such critical configurations deployed in a rolling update fashion of the worker pool (contrary to in-place like today).
This requires some changes also in the extension library that calculates the `MachineClasses`.

pkg/apis/core/validation/shoot.go
rfranzke: Hm, I tend to not over-complicate this PR for now. It's an advanced setting (like many other fields that we already have) and having such sophisticated validation will make it harder for us at the moment. I appreciate your thoughts about it, but I'd probably start small and unblock users that need to configure these settings. If we see that it's misused then we can revisit the validation topic. WDYT?

rfranzke (dismissed):
/lgtm

pkg/apis/core/types_shoot.go
```diff
@@ -556,6 +556,15 @@ type KubeletConfig struct {
 	ImagePullProgressDeadline *metav1.Duration
 	// FailSwapOn makes the Kubelet fail to start if swap is enabled on the node. (default true).
 	FailSwapOn *bool
+	// KubeReserved is the configuration for resources reserved for kubernetes node components (mainly kubelet and container runtime).
+	// When updating these values, be aware that cgroup resizes may not succeed on active worker nodes. Look for the NodeAllocatableEnforced event to determine if the configuration was applied.
+	// Default:
+	//   memory:	1Gi
+	//   cpu:		80m
```
timebertt:
```suggestion
	// Default: cpu=80m,memory=1Gi
```

pkg/apis/core/types_shoot.go
```diff
@@ -600,6 +609,18 @@ type KubeletConfigEvictionSoftGracePeriod struct {
...
+// KubeletConfigReserved contains reserved resources for daemons
+type KubeletConfigReserved struct {
+	// CPU is the reserved cpu.
+	CPU *resource.Quantity
+	// Memory is the reserved memory.
+	Memory *resource.Quantity
+	// EphemeralStorage is the reserved ephemeral-storage.
+	EphemeralStorage *resource.Quantity
+	// PID is the reserved process-ids.
+	PID *int64
```
timebertt: The kubelet seems to handle `pid` as a `resource.Quantity` just like the other resources (ref: https://github.com/kubernetes/kubernetes/blob/43fbe17dc682b703b0b663935a797244378f0aaf/cmd/kubelet/app/server.go#L1250-L1252)
I think, it would make sense to do the same in our API.
```suggestion
	PID *resource.Quantity
```

pkg/apis/core/types_shoot.go
```diff
@@ -600,6 +609,18 @@ type KubeletConfigEvictionSoftGracePeriod struct {
...
 
+// KubeletConfigReserved contains reserved resources for daemons
+type KubeletConfigReserved struct {
+	// CPU is the reserved cpu.
+	CPU *resource.Quantity
+	// Memory is the reserved memory.
+	Memory *resource.Quantity
+	// EphemeralStorage is the reserved ephemeral-storage.
+	EphemeralStorage *resource.Quantity
+	// PID is the reserved process-ids.
```
timebertt, line 618: To make use of the `pid` setting, the `SupportNodePidsLimit` has to be enabled (beta since 1.15, so only for <1.15). Can you add a small comment for that?

Ref: https://github.com/kubernetes/kubernetes/blob/43fbe17dc682b703b0b663935a797244378f0aaf/cmd/kubelet/app/server.go#L1250-L1252

pkg/apis/core/validation/shoot.go
```diff
@@ -1029,6 +1035,26 @@ func validateKubeletConfigEvictionSoftGracePeriod(eviction *core.KubeletConfigEv
...
+		allErrs = append(allErrs, validateResourceQuantityValue("cpu", *reserved.CPU, fldPath.Child("cpu"))...)
+	}
+	if reserved.Memory != nil {
+		allErrs = append(allErrs, validateResourceQuantityValue("memory", *reserved.Memory, fldPath.Child("memory"))...)
+	}
+	if reserved.EphemeralStorage != nil {
+		allErrs = append(allErrs, validateResourceQuantityValue("ephemeralStorage", *reserved.EphemeralStorage, fldPath.Child("ephemeralStorage"))...)
+	}
+	if *reserved.PID < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("pid"), *reserved.PID, fmt.Sprintf("PID must be atleast %d", ReservedPIDsMinimum)))
```
timebertt: See comment in `pkg/apis/core/types_shoot.go`.
If `pid` is also a `resource.Quantity` you can use `validateResourceQuantityValue` here as well.

timebertt (changes requested):
Thanks, looks quite good now.
Only some minor change requests from my side :)

timuthy (dismissed):
/lgtm

pkg/apis/core/validation/shoot.go
```diff
@@ -1029,6 +1035,26 @@ func validateKubeletConfigEvictionSoftGracePeriod(eviction *core.KubeletConfigEv
 	return allErrs
 }
 
+// Reserved PIDs have to be non-negative
+const ReservedPIDsMinimum int64 = 0
+
```
timebertt: `ReservedPIDsMinimum` is now unused.
```suggestion
```

timebertt (commented):
Thanks!
Only one nit left :)

timebertt (approved):
Nice, thank you very much for your work!
/lgtm

timuthy (approved):

PR #2576, opened 2020-07-10
[GEP-7] Add DNSOwner to Shoot operations to support migration scenario

### LINKED ISSUE #2335: Add management for DNSOwners and DNSEntries during Shoot reconciliation, migration and restoration

swilen-iwanow:

/area control-plane-migration
/kind enhancement
/priority normal

Adds DNSOwner as part of the Shoot's control plane. The DNSOwner is responsible of holding the information of the owner of DNSEntries. DNSEntries will point to this DNSOwner by their `spec.owner` field. We need it to enable migration scenario in which DNS entries are not deleted from the DNS provider during migration (_we target only change of the IP, without destroying and then recreating the DNS_). Crucial part of the lifecycle of the DNS is the order by they are deployed/desroyed depending on the desired result. This order was defined by the dns-controller. 

Fixes #2335 

To test this you can run a reconciliation of a Shoot and check if the owner of the DNS Record has changed to the cluster-identity of the Shoot. To test it for the migration scenario you will have to disable the validation of the API Server for the `Shoots.Spec.seedName` that doesn't allow change. You can do this by commenting the code inside `pkg/apis/core/validation/shoot.go` lines 169-172

```go
if oldSpec.SeedName != nil {
		// allow initial seed assignment
		allErrs = append(allErrs, apivalidation.ValidateImmutableField(newSpec.SeedName, oldSpec.SeedName, fldPath.Child("seedName"))...)
	}

```
Then you can change the `shoot.spec.seedName` to point to another seed (that has the same cloud provider). This will trigger the migration. You then monitor the DNS record in the provider, that only has to change the IP without it being destroyed. Hopefully everything goes fine and you have healthy migrated cluster. 

P.S: Use the latest extension controllers

**Release note**:

```action operator
The shoot reconciliation does now create `DNSOwner` objects in the seed cluster to manage the ownership of the DNS entries. You should run at least [`v0.7.16`](https://github.com/gardener/external-dns-management/releases/tag/v0.7.16) of the [external-dns-management](https://github.com/gardener/external-dns-management/) extension.
```

### DIFF
--- a/charts/seed-dns/owner/templates/dnsowner.yaml
+++ b/charts/seed-dns/owner/templates/dnsowner.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: dns.gardener.cloud/v1alpha1
+kind: DNSOwner
+metadata:
+  name: {{ .Release.Namespace }}-{{ .Values.name }}
+spec:
+  ownerId: {{ .Values.ownerID }}
+  active: {{ .Values.active }}
--- a/pkg/gardenlet/controller/shoot/shoot_control_delete.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_delete.go
@@ -28,7 +28,6 @@ import (
 	"github.com/gardener/gardener/pkg/controllerutils"
 	"github.com/gardener/gardener/pkg/operation"
 	botanistpkg "github.com/gardener/gardener/pkg/operation/botanist"
-	"github.com/gardener/gardener/pkg/operation/botanist/component"
 	"github.com/gardener/gardener/pkg/utils/errors"
 	"github.com/gardener/gardener/pkg/utils/flow"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
@@ -428,8 +427,8 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation) *gardencorev1bet
 		})
 
 		destroyNginxIngressDNSRecord = g.Add(flow.Task{
-			Name:         "Destroying ingress DNS record",
-			Fn:           flow.TaskFn(component.OpDestroyAndWait(botanist.Shoot.Components.Extensions.DNS.NginxEntry).Destroy),
+			Name:         "Destroying nginx ingress DNS record",
+			Fn:           flow.TaskFn(botanist.DestroyIngressDNSRecord),
 			Dependencies: flow.NewTaskIDs(syncPointCleaned),
 		})
 		destroyInfrastructure = g.Add(flow.Task{
@@ -443,8 +442,8 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation) *gardencorev1bet
 			Dependencies: flow.NewTaskIDs(destroyInfrastructure),
 		})
 		destroyExternalDomainDNSRecord = g.Add(flow.Task{
-			Name:         "Destroying external DNS entry",
-			Fn:           flow.TaskFn(component.OpWaiter(botanist.Shoot.Components.Extensions.DNS.ExternalEntry).Destroy),
+			Name:         "Destroying external domain DNS record",
+			Fn:           flow.TaskFn(botanist.DestroyExternalDNS),
 			Dependencies: flow.NewTaskIDs(syncPointCleaned),
 		})
 
@@ -459,12 +458,12 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation) *gardencorev1bet
 		)
 
 		destroyInternalDomainDNSRecord = g.Add(flow.Task{
-			Name:         "Destroying internal DNS entry",
-			Fn:           flow.TaskFn(component.OpWaiter(botanist.Shoot.Components.Extensions.DNS.InternalEntry).Destroy),
+			Name:         "Destroying internal domain DNS record",
+			Fn:           flow.TaskFn(botanist.DestroyInternalDNS),
 			Dependencies: flow.NewTaskIDs(syncPoint),
 		})
 		deleteDNSProviders = g.Add(flow.Task{
-			Name:         "Deleting DNS providers",
+			Name:         "Deleting additional DNS providers",
 			Fn:           flow.TaskFn(botanist.DeleteDNSProviders),
 			Dependencies: flow.NewTaskIDs(destroyInternalDomainDNSRecord),
 		})
--- a/pkg/operation/botanist/botanist.go
+++ b/pkg/operation/botanist/botanist.go
@@ -77,9 +77,12 @@ func New(o *operation.Operation) (*Botanist, error) {
 	}
 
 	o.Shoot.Components.Extensions.DNS.ExternalProvider = b.DefaultExternalDNSProvider(b.K8sSeedClient.DirectClient())
+	o.Shoot.Components.Extensions.DNS.ExternalOwner = b.DefaultExternalDNSOwner(b.K8sSeedClient.DirectClient())
 	o.Shoot.Components.Extensions.DNS.ExternalEntry = b.DefaultExternalDNSEntry(b.K8sSeedClient.DirectClient())
 	o.Shoot.Components.Extensions.DNS.InternalProvider = b.DefaultInternalDNSProvider(b.K8sSeedClient.DirectClient())
+	o.Shoot.Components.Extensions.DNS.InternalOwner = b.DefaultInternalDNSOwner(b.K8sSeedClient.DirectClient())
 	o.Shoot.Components.Extensions.DNS.InternalEntry = b.DefaultInternalDNSEntry(b.K8sSeedClient.DirectClient())
+	o.Shoot.Components.Extensions.DNS.NginxOwner = b.DefaultNginxIngressDNSOwner(b.K8sSeedClient.DirectClient())
 	o.Shoot.Components.Extensions.DNS.NginxEntry = b.DefaultNginxIngressDNSEntry(b.K8sSeedClient.DirectClient())
 	o.Shoot.Components.Extensions.DNS.AdditionalProviders, err = b.AdditionalDNSProviders(context.TODO(), b.K8sGardenClient.Client(), b.K8sSeedClient.DirectClient())
 	if err != nil {
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -1346,7 +1346,7 @@ func (b *Botanist) DefaultKubeAPIServerService() component.DeployWaiter {
 		b.K8sSeedClient.DirectClient(),
 		nil,
 		b.setAPIServerServiceClusterIP,
-		b.setAPIServerAddress,
+		func(address string) { b.setAPIServerAddress(address, b.K8sSeedClient.DirectClient()) },
 	)
 }
 
@@ -1395,37 +1395,63 @@ func (b *Botanist) setAPIServerServiceClusterIP(clusterIP string) {
 }
 
 // setAPIServerAddress sets the IP address of the API server's LoadBalancer.
-func (b *Botanist) setAPIServerAddress(address string) {
+func (b *Botanist) setAPIServerAddress(address string, seedClient client.Client) {
 	b.Operation.APIServerAddress = address
 
 	if b.NeedsInternalDNS() {
+		ownerID := *b.Shoot.Info.Status.ClusterIdentity + "-" + DNSInternalName
+		b.Shoot.Components.Extensions.DNS.InternalOwner = dns.NewDNSOwner(
+			&dns.OwnerValues{
+				Name:    DNSInternalName,
+				Active:  true,
+				OwnerID: ownerID,
+			},
+			b.Shoot.SeedNamespace,
+			b.K8sSeedClient.ChartApplier(),
+			b.ChartsRootPath,
+			seedClient,
+		)
 		b.Shoot.Components.Extensions.DNS.InternalEntry = dns.NewDNSEntry(
 			&dns.EntryValues{
 				Name:    DNSInternalName,
 				DNSName: common.GetAPIServerDomain(b.Shoot.InternalClusterDomain),
 				Targets: []string{b.APIServerAddress},
+				OwnerID: ownerID,
 			},
 			b.Shoot.SeedNamespace,
 			b.K8sSeedClient.ChartApplier(),
 			b.ChartsRootPath,
 			b.Logger,
-			b.K8sSeedClient.DirectClient(),
+			seedClient,
 			nil,
 		)
 	}
 
 	if b.NeedsExternalDNS() {
+		ownerID := *b.Shoot.Info.Status.ClusterIdentity + "-" + DNSExternalName
+		b.Shoot.Components.Extensions.DNS.ExternalOwner = dns.NewDNSOwner(
+			&dns.OwnerValues{
+				Name:    DNSExternalName,
+				Active:  true,
+				OwnerID: ownerID,
+			},
+			b.Shoot.SeedNamespace,
+			b.K8sSeedClient.ChartApplier(),
+			b.ChartsRootPath,
+			seedClient,
+		)
 		b.Shoot.Components.Extensions.DNS.ExternalEntry = dns.NewDNSEntry(
 			&dns.EntryValues{
 				Name:    DNSExternalName,
 				DNSName: common.GetAPIServerDomain(*b.Shoot.ExternalClusterDomain),
 				Targets: []string{b.APIServerAddress},
+				OwnerID: ownerID,
 			},
 			b.Shoot.SeedNamespace,
 			b.K8sSeedClient.ChartApplier(),
 			b.ChartsRootPath,
 			b.Logger,
-			b.K8sSeedClient.DirectClient(),
+			seedClient,
 			nil,
 		)
 	}
--- a/pkg/operation/botanist/dns.go
+++ b/pkg/operation/botanist/dns.go
@@ -57,46 +57,54 @@ func GenerateDNSProviderName(secretName, providerType string) string {
 	}
 }
 
-// DeployExternalDNS deploys the external DNSProvider and DNSEntry.
+// DeployExternalDNS deploys the external DNSOwner, DNSProvider, and DNSEntry resources.
 func (b *Botanist) DeployExternalDNS(ctx context.Context) error {
-	return b.ExternalDNS().Deploy(ctx)
-}
-
-// ExternalDNS returns the external DNSProvider which deploys / destroys
-// the external DNSProvider and DNSEntry.
-func (b *Botanist) ExternalDNS() component.Deployer {
 	if b.NeedsExternalDNS() {
+		if b.isRestorePhase() {
+			return dnsRestoreDeployer{
+				provider: b.Shoot.Components.Extensions.DNS.ExternalProvider,
+				entry:    b.Shoot.Components.Extensions.DNS.ExternalEntry,
+				owner:    b.Shoot.Components.Extensions.DNS.ExternalOwner,
+			}.Deploy(ctx)
+		}
+
 		return component.OpWaiter(
+			b.Shoot.Components.Extensions.DNS.ExternalOwner,
 			b.Shoot.Components.Extensions.DNS.ExternalProvider,
 			b.Shoot.Components.Extensions.DNS.ExternalEntry,
-		)
+		).Deploy(ctx)
 	}
 
 	return component.OpWaiter(
 		b.Shoot.Components.Extensions.DNS.ExternalEntry,
 		b.Shoot.Components.Extensions.DNS.ExternalProvider,
-	)
+		b.Shoot.Components.Extensions.DNS.ExternalOwner,
+	).Deploy(ctx)
 }
 
-// DeployInternalDNS deploys the internal DNSProvider and DNSEntry.
+// DeployInternalDNS deploys the internal DNSOwner, DNSProvider, and DNSEntry resources.
 func (b *Botanist) DeployInternalDNS(ctx context.Context) error {
-	return b.InternalDNS().Deploy(ctx)
-}
-
-// InternalDNS returns the internal DNSProvider which deploys / destroys
-// the internal DNSProvider and DNSEntry.
-func (b *Botanist) InternalDNS() component.Deployer {
 	if b.NeedsInternalDNS() {
+		if b.isRestorePhase() {
+			return dnsRestoreDeployer{
+				provider: b.Shoot.Components.Extensions.DNS.InternalProvider,
+				entry:    b.Shoot.Components.Extensions.DNS.InternalEntry,
+				owner:    b.Shoot.Components.Extensions.DNS.InternalOwner,
+			}.Deploy(ctx)
+		}
+
 		return component.OpWaiter(
+			b.Shoot.Components.Extensions.DNS.InternalOwner,
 			b.Shoot.Components.Extensions.DNS.InternalProvider,
 			b.Shoot.Components.Extensions.DNS.InternalEntry,
-		)
+		).Deploy(ctx)
 	}
 
 	return component.OpWaiter(
 		b.Shoot.Components.Extensions.DNS.InternalEntry,
 		b.Shoot.Components.Extensions.DNS.InternalProvider,
-	)
+		b.Shoot.Components.Extensions.DNS.InternalOwner,
+	).Deploy(ctx)
 }
 
 // DefaultExternalDNSProvider returns the external DNSProvider if external DNS is
@@ -156,6 +164,19 @@ func (b *Botanist) DefaultExternalDNSEntry(seedClient client.Client) component.D
 	))
 }
 
+// DefaultExternalDNSOwner returns DeployWaiter which removes the external DNSOwner.
+func (b *Botanist) DefaultExternalDNSOwner(seedClient client.Client) component.DeployWaiter {
+	return component.OpDestroy(dns.NewDNSOwner(
+		&dns.OwnerValues{
+			Name: DNSExternalName,
+		},
+		b.Shoot.SeedNamespace,
+		b.K8sSeedClient.ChartApplier(),
+		b.ChartsRootPath,
+		seedClient,
+	))
+}
+
 // DefaultInternalDNSProvider returns the internal DNSProvider if internal DNS is
 // enabled and if not, DeployWaiter which removes the internal DNSProvider.
 func (b *Botanist) DefaultInternalDNSProvider(seedClient client.Client) component.DeployWaiter {
@@ -209,6 +230,19 @@ func (b *Botanist) DefaultInternalDNSEntry(seedClient client.Client) component.D
 		b.Logger,
 		seedClient,
 		nil,
+	))
+}
+
+// DefaultInternalDNSOwner returns a DeployWaiter which removes the internal DNSOwner.
+func (b *Botanist) DefaultInternalDNSOwner(seedClient client.Client) component.DeployWaiter {
+	return component.OpDestroy(dns.NewDNSOwner(
+		&dns.OwnerValues{
+			Name: DNSInternalName,
+		},
+		b.Shoot.SeedNamespace,
+		b.K8sSeedClient.ChartApplier(),
+		b.ChartsRootPath,
+		seedClient,
 	))
 }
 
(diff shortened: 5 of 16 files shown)


danielfoehrKn:
Will take a look next week.
@mandelsoft can maybe check from the DNS side.

plkokanov:
I'll take a look at it today.

pkg/operation/botanist/botanist.go
```diff
@@ -77,11 +77,14 @@ func New(o *operation.Operation) (*Botanist, error) {
 	}
 
 	o.Shoot.Components.Extensions.DNS.ExternalProvider = b.DefaultExternalDNSProvider(b.K8sSeedClient.DirectClient())
+	o.Shoot.Components.Extensions.DNS.ExternalOwner = b.DefaultExternalDNSOwner(b.K8sSeedClient.DirectClient())
 	o.Shoot.Components.Extensions.DNS.ExternalEntry = b.DefaultExternalDNSEntry(b.K8sSeedClient.DirectClient())
 	o.Shoot.Components.Extensions.DNS.InternalProvider = b.DefaultInternalDNSProvider(b.K8sSeedClient.DirectClient())
 	o.Shoot.Components.Extensions.DNS.InternalEntry = b.DefaultInternalDNSEntry(b.K8sSeedClient.DirectClient())
 	o.Shoot.Components.Extensions.DNS.NginxEntry = b.DefaultNginxIngressDNSEntry(b.K8sSeedClient.DirectClient())
+	o.Shoot.Components.Extensions.DNS.InternalOwner = b.DefaultInternalDNSOwner(b.K8sSeedClient.DirectClient())
```
plkokanov: minor nit: can you move the `InternalOwner` definition above the `NginxEntry`

pkg/operation/botanist/controlplane.go
```diff
@@ -1400,6 +1400,7 @@ func (b *Botanist) setAPIServerAddress(address string) {
 				Name:    DNSInternalName,
 				DNSName: common.GetAPIServerDomain(b.Shoot.InternalClusterDomain),
 				Targets: []string{b.APIServerAddress},
+				OwnerID: *b.Shoot.Info.Status.ClusterIdentity,
```
plkokanov: Can the `ClusterIdentity` be nil here?

pkg/operation/botanist/dns.go
```diff
@@ -212,6 +262,35 @@ func (b *Botanist) DefaultInternalDNSEntry(seedClient client.Client) component.D
 	))
 }
 
+// DefaultInternalDNSOwner returns DeployWaiter which removes the internal DNSOwner.
```
plkokanov:
```suggestion
// DefaultInternalDNSOwner returns the internal DNSOwner if internal DNS is
// enabled and if not, DeployWaiter which removes the internal DNSOwner
```

pkg/operation/botanist/extensions/dns/dnsowner.go
```diff
@@ -0,0 +1,89 @@
...
+
+// WaitCleanup implements Waiter
+func (d *dnsOwner) WaitCleanup(ctx context.Context) error {
+	return kutil.WaitUntilResourceDeleted(ctx, d.client, d.owner(), 5*time.Second)
+}
+
+// Wait implements Waiter, not applicable for the DNSOwner
+func (d *dnsOwner) Wait(ctx context.Context) error { return nil }
+
+// entry returns an empty DNSEntry used for deletion.
```
plkokanov:
```suggestion
// owner returns an empty DNSOwner used for deletion.
```

plkokanov (changes requested):

pkg/operation/botanist/botanist.go
```diff
@@ -77,11 +77,14 @@ func New(o *operation.Operation) (*Botanist, error) {
...
 	o.Shoot.Components.Extensions.DNS.AdditionalProviders, err = b.AdditionalDNSProviders(context.TODO(), b.K8sGardenClient.Client(), b.K8sSeedClient.DirectClient())
+
```
rfranzke:
```suggestion
```

pkg/operation/botanist/controlplane.go
rfranzke: I don't think so

pkg/operation/botanist/dns.go
```diff
@@ -369,3 +448,23 @@ func (b *Botanist) DeleteDNSProviders(ctx context.Context) error {
...
+
+// MigrateExternalDNS destroys the external DNSEntry, DNSOwner and DNSProvider resources
+// from the cluster, without removing it from the DNS provider
+func (b *Botanist) MigrateExternalDNS(ctx context.Context) error {
+	return component.OpDestroy(
+		b.Shoot.Components.Extensions.DNS.ExternalOwner,
+		b.Shoot.Components.Extensions.DNS.ExternalProvider,
+		b.Shoot.Components.Extensions.DNS.ExternalEntry,
+	).Destroy(ctx)
+}
```
rfranzke: What does "without removing it from the DNS provider" mean? How does this work?

pkg/operation/botanist/extensions/dns/dnsowner.go
```diff
@@ -0,0 +1,89 @@
...
+type dnsOwner struct {
+	values         *OwnerValues
+	shootNamespace string
+	kubernetes.ChartApplier
+	chartPath string
+	logger    *logrus.Entry
+	client    crclient.Client
+}
+
+//Deploy implements Deployer and creates DNSOwner for the provided values
```
rfranzke:
```suggestion
// Deploy implements Deployer and creates DNSOwner for the provided values
```

pkg/operation/botanist/extensions/dns/dnsowner.go
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke, line 1: Does it make sense to add some unit tests for this file?

rfranzke (commented):
@mandelsoft @MartinWeindel please thoroughly check the changes. Today, we don't have `DNSOwner` objects and are using a different `ownerID` for the existing DNS entries.

pkg/operation/botanist/dns.go
swilen-iwanow: Means that, it is not deleted in route53 for example and during the restoration, only the IP is updated. As @mandelsoft said the order of removing the DNS resources (Owner, Provider and Entry) is crucial for the dns controller, to achieve migration.

rfranzke: OK, I see, thank you!

plkokanov:
> @mandelsoft @MartinWeindel please thoroughly check the changes. Today, we don't have DNSOwner objects and are using a different ownerID for the existing DNS entries.

The reconciliation automatically changes the owner IDs from the old to the new ones and vice versa.

pkg/operation/botanist/controlplane.go
swilen-iwanow: We have a step in the beginning of the reconciliation (before the namespace creation even), that sets the cluster identity and reloads the latest object into the botanist. The only way it can be nil is if some other step clears it.

charts/seed-dns/owner/templates/dnsowner.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: dns.gardener.cloud/v1alpha1
+kind: DNSOwner
+metadata:
+  name: {{ .Values.name }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  {{- if .Values.ownerID }}
+  ownerId: {{ .Values.ownerID }}
+  {{- end }}
```
MartinWeindel: Field `ownerId` is required.
Either `.Values.ownerID` is set or it makes no sense to deploy a DNSOwner object at all.

MartinWeindel (changes requested):
There is a minor inconsistency in the dnsowner.yaml

swilen-iwanow:
Thanks for the reviews. I've addressed the comments. Can you take another look :)

MartinWeindel (dismissed):
lgtm

plkokanov:
lgtm

plkokanov (dismissed):

pkg/operation/botanist/dns.go
```diff
@@ -156,6 +176,36 @@ func (b *Botanist) DefaultExternalDNSEntry(seedClient client.Client) component.D
...
 
+// DefaultExternalDNSOwner returns the external DNSOwner if external DNS is
+// enabled and if not, DeployWaiter which removes the external DNSOwner
+func (b *Botanist) DefaultExternalDNSOwner(seedClient client.Client) component.DeployWaiter {
+	if b.NeedsExternalDNS() {
+		return dns.NewDNSOwner(
+			&dns.OwnerValues{
+				Name:    DNSExternalName,
+				Active:  true,
+				OwnerID: *b.Shoot.Info.Status.ClusterIdentity,
```
danielfoehrKn: This creates a nil pointer panic in the gardenlet. 
Both the care controller and the shoot reconciliation create a new operation.
If the care controller is "quicker"(  e.g Seed is not yet ready etc. - see operation is created [here](https://github.com/gardener/gardener/blob/c0c5a694ae28d90f06c080008f5464d161fb30d3/pkg/gardenlet/controller/shoot/shoot_care_control.go#L175)), compared to the shoot reconciliation (creates the cluster-identity [here](https://github.com/gardener/gardener/blob/f55437b175a3868f220c26514948be6248a425f3/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go#L107-L106)) , the cluster identity in the Shoot status is nil.

Logs
```
E0724 11:20:24.895323   29917 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 568 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x26aa720, 0x3ad6090)
	/Users/d060239/go/src/github.com/gardener/gardener/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0xa3
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/Users/d060239/go/src/github.com/gardener/gardener/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x82
panic(0x26aa720, 0x3ad6090)
	/usr/local/Cellar/go/1.14/libexec/src/runtime/panic.go:967 +0x15d
github.com/gardener/gardener/pkg/operation/botanist.(*Botanist).DefaultExternalDNSOwner(0xc001846060, 0x2c30260, 0xc0002645a0, 0x2c1a980, 0xc001e0c720)
	/Users/d060239/go/src/github.com/gardener/gardener/pkg/operation/botanist/dns.go:187 +0xf0
github.com/gardener/gardener/pkg/operation/botanist.New(0xc000e8d680, 0x0, 0x0, 0x290c758)
	/Users/d060239/go/src/github.com/gardener/gardener/pkg/operation/botanist/botanist.go:80 +0x3c0
github.com/gardener/gardener/pkg/gardenlet/controller/shoot.(*defaultCareControl).Care(0xc0006ad020, 0xc0003a6380, 0xc00126a100, 0x14, 0xc00017dfc0, 0x1)
	/Users/d060239/go/src/github.com/gardener/gardener/pkg/gardenlet/controller/shoot/shoot_care_control.go:175 +0x7b7
github.com/gardener/gardener/pkg/gardenlet/controller/shoot.(*Controller).reconcileShootCareKey(0xc00082a6e0, 0xc00126a100, 0x14, 0x1, 0xc00126d610)
	/Users/d060239/go/src/github.com/gardener/gardener/pkg/gardenlet/controller/shoot/shoot_care_control.go:80 +0x1c4
github.com/gardener/gardener/pkg/controllerutils.DeprecatedCreateWorker.func1(0xc00126a0e0, 0xa, 0xc00126a0eb, 0x9, 0x0, 0x0, 0xc00126d5e0, 0x0)
	/Users/d060239/go/src/github.com/gardener/gardener/pkg/controllerutils/worker.go:46 +0x20a
sigs.k8s.io/controller-runtime/pkg/reconcile.Func.Reconcile(0xc0006a8090, 0xc00126a0e0, 0xa, 0xc00126a0eb, 0x9, 0x9, 0x0, 0x0, 0x0)
	/Users/d060239/go/src/github.com/gardener/gardener/vendor/sigs.k8s.io/controller-runtime/pkg/reconcile/reconcile.go:93 +0x4e
github.com/gardener/gardener/pkg/controllerutils.worker.func1.1(0x2c35840, 0xc000cda300, 0x2bcb5a0, 0xc0006a8090, 0x290145e, 0xa, 0x0)
	/Users/d060239/go/src/github.com/gardener/gardener/pkg/controllerutils/worker.go:99 +0x124
github.com/gardener/gardener/pkg/controllerutils.worker.func1()
	/Users/d060239/go/src/github.com/gardener/gardener/pkg/controllerutils/worker.go:116 +0x7e
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1(0xc000a401c0)
	/Users/d060239/go/src/github.com/gardener/gardener/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:152 +0x5f
k8s.io/apimachinery/pkg/util/wait.JitterUntil(0xc000a401c0, 0x3b9aca00, 0x0, 0x1, 0xc0007d4240)
	/Users/d060239/go/src/github.com/gardener/gardener/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:153 +0xf8
k8s.io/apimachinery/pkg/util/wait.Until(...)
	/Users/d060239/go/src/github.com/gardener/gardener/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88
github.com/gardener/gardener/pkg/controllerutils.CreateWorker.func1(0x2c35840, 0xc000cda300, 0x290145e, 0xa, 0x2bcb5a0, 0xc0006a8090, 0x2c1a600, 0xc00099c600, 0xc000051860, 0xc00068eda0)
	/Users/d060239/go/src/github.com/gardener/gardener/pkg/controllerutils/worker.go:57 +0xb4
created by github.com/gardener/gardener/pkg/controllerutils.CreateWorker
	/Users/d060239/go/src/github.com/gardener/gardener/pkg/controllerutils/worker.go:56 +0xec
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x246e210]
```

danielfoehrKn (dismissed):

plkokanov:
@MartinWeindel @mandelsoft just noticed that the CRD for the DNSOwner resources actually defines them as cluster scoped resources. I guess the current PR assumes that they are namespaced as they will be created with the same name for all shoots and that should be changed to include the name of the shoot in the DNSOwner name similar to the `backupentry` cluster resource.

Additionally thanks to @stoyanr for testing it out and finding that the DNS entries in the dns provider get recreated. 
Both the deletion and creation happen on the side of the dns-controller running in the destination seed. Which means that the migration flow is ok, but the restoration is not.
@mandelsoft I remembered that we talked that something like this could happen if the DNSOwner gets reconciled before the DNSEntry is known to the dns controller. Just for information, should this race condition be fixed in 0.7.13 (which we use for the migration tests)

Notice that currently the order of deployment for dns components during the `Restore` phase is the following:
```
return component.OpWaiter(
        b.Shoot.Components.Extensions.DNS.InternalOwner,
        b.Shoot.Components.Extensions.DNS.InternalProvider,
        b.Shoot.Components.Extensions.DNS.InternalEntry,
)
```

Generally I think it should be switched to this:
```
return component.OpWaiter(
        b.Shoot.Components.Extensions.DNS.InternalProvider,
        b.Shoot.Components.Extensions.DNS.InternalEntry,
        b.Shoot.Components.Extensions.DNS.InternalOwner,
)
```
but it causes the DNSEntry reconciliation to fail with the following errors: `No responsible provider found` and `unknown owner id '<owner-id>'`

One more potential problem is that in the PR currently we deploy two DNSOwners with the same ownerIDs (one for the internal and one for the external DNS) which could cause problems if the internal DNSOwner gets deployed before the external DNSEntry and the external DNSOwner, then the external DNSEntry will get reconciled and recreated in the dns provider (or vice versa).

So after playing a bit with it here's what i suggest has to be done during the restoration phase:
1. Deploy the internal and external DNSProviders
2. Deploy the internal and external DNSEntries first and wait for them to get their state updated to `Invalid` or `Error` (the state check might not be necessary but will have to test it later on).
3. Deploy the DNSOwner.
4. Wait for the internal and external DNSEntries to become ready.

Sorry for noticing this so late, but I never checked the logs of both dns-controllers during testing and logs in the cloud console :(  so again thanks @stoyanr for taking a look.

pkg/operation/botanist/dns.go
```diff
@@ -66,7 +66,16 @@ func (b *Botanist) DeployExternalDNS(ctx context.Context) error {
 // the external DNSProvider and DNSEntry.
 func (b *Botanist) ExternalDNS() component.Deployer {
 	if b.NeedsExternalDNS() {
+		if b.isRestorePhase() {
```
stoyanr, line 63: The code in this `if` is exactly the same as the one after it. Unless the order of deployment is changed (see other comments), this `if` is redundant and should be removed.

pkg/operation/botanist/dns.go
```diff
@@ -87,7 +97,16 @@ func (b *Botanist) DeployInternalDNS(ctx context.Context) error {
 // the internal DNSProvider and DNSEntry.
 func (b *Botanist) InternalDNS() component.Deployer {
 	if b.NeedsInternalDNS() {
+		if b.isRestorePhase() {
```
stoyanr, line 88: The code in this if is exactly the same as the one after it. Unless the order of deployment is changed (see other comments), this if is redundant and should be removed.

pkg/operation/botanist/dns.go
stoyanr: I was able to reproduce this issue as well. I think it's reproduced very reliably if you create a new shoot instead of migrating an existing one. We brainstormed it a bit with @plkokanov and concluded that in order to fix it, the DNS owner should be inititalized later on, after creating the API server load balancer, similarly to the DNS entry. I will provide a fix shortly.

stoyanr (changes requested):

mandelsoft:
The cluster scope for the DNSOwner objects is indeed not nice for namespace local scenarios.
The problem here is, that the identifier stored in the DNS systems must be unique. There are no namespaces. This was
the reason for using cluster -scoped  Owner objects. There must be a possibility to declare a global ownership for the cluster. If the owner objects were namespaced the declared identifiers would have to be implicitly extended by the namespace to achieve unique identifiers for owner objects with the same identifier in different namespaces.  But this would not allow for global responsibilities.
Another option would be to use namespaced owner objects to *enable* an identifier for a namespace. But this also does not solve the global responsibility and has other strange consequences contradicting with the global meaning of identifiers in the DNS system.

In any case it would be a very complex handling in the DNS controller that is hardly to explain. 

Therefore it indeed seems to be the most simple and straight forward solution to stick to the cluster-scope and use unique names for the shoot specific owner objects. (The namespace could be prepended to the name.) There should not be a permission problem, because all those objects are solely managed by the gardener.

stoyanr:
I fixed the nil pointer issue and addressed the above comments:
* The nil pointer error mentioned above is now fixed.
* `DNSOwner` is now correctly treated as a cluster resource.
* There are now 3 DNS owners, each with a unique ID - one for the internal domain, external domain, and nginx ingress DNS entries.
* The migration and restoration flows for all 3 DNS entries / owners have been revised to follow the recommendation from @plkokanov  above.

Most importantly, I tested the migration a few times and verified that all DNS entries are not recreated during the restoration. The internal and external DNS entries are only updated (since the IP changes), the nginx ingress one is not changed at all.

@plkokanov @mandelsoft @MartinWeindel @danielfoehrKn please take another look.

pkg/operation/botanist/extensions/dns/dnsowner.go
```diff
@@ -0,0 +1,89 @@
...
+	"time"
+
+	dnsv1alpha1 "github.com/gardener/external-dns-management/pkg/apis/dns/v1alpha1"
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/operation/botanist/component"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
```
ialidzhikov: It is already imported in the line above.

```suggestion
```

pkg/operation/botanist/extensions/dns/dnsowner.go
```diff
@@ -0,0 +1,89 @@
...
+		client:         client,
+	}
+}
+
+type dnsOwner struct {
+	values         *OwnerValues
+	shootNamespace string
+	kubernetes.ChartApplier
+	chartPath string
+	logger    *logrus.Entry
```
ialidzhikov: The logger is currently unused.

```suggestion
```

pkg/operation/botanist/extensions/dns/dnsowner_test.go
```diff
@@ -0,0 +1,142 @@
...
+	"k8s.io/apimachinery/pkg/version"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+var _ = Describe("#DNSOwner", func() {
+	const (
+		deployNS     = "test-chart-namespace"
+		secretName   = "extensions-dns-test-deploy"
```
ialidzhikov: It is currently unused.

```suggestion
```

ialidzhikov (commented):
Nits

pkg/operation/botanist/extensions/dns/dnsowner_test.go
```diff
@@ -0,0 +1,137 @@
...
+		vals = &OwnerValues{
+			Name:    "test-deploy",
+			Active:  true,
+			OwnerID: ownerID,
+		}
+
+		expectedDNSOwner = &dnsv1alpha1.DNSOwner{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      deployNS + "-" + dnsOwnerName,
+				Namespace: "default",
```
ialidzhikov: `DNSOwner` is currently cluster-scoped.

```suggestion
```

pkg/operation/botanist/extensions/dns/dnsowner_test.go
```diff
@@ -0,0 +1,137 @@
...
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
+	Describe("#Deploy", func() {
+		It("should create correct DNSOwner", func() {
+			Expect(defaultDepWaiter.Deploy(ctx)).ToNot(HaveOccurred())
+
+			actualDNSOwner := &dnsv1alpha1.DNSOwner{}
+			err := c.Get(ctx, client.ObjectKey{Name: deployNS + "-" + dnsOwnerName, Namespace: "default"}, actualDNSOwner)
```
ialidzhikov: `DNSOwner` is currently cluster-scoped.

```suggestion
			err := c.Get(ctx, client.ObjectKey{Name: deployNS + "-" + dnsOwnerName}, actualDNSOwner)
```

ialidzhikov (commented):
After Shoot deletion, the corresponding dnsowners are not deleted (they didn't receive a deletionTimestamp at all):
```
$ k get dnsowner
NAME                                 OWNERID                                                                                          AGE
shoot--foooooo--baaaaaaar-external   shoot--foooooo--baaaaaaar-b5c2177a-23ae-442a-a1fb-60e742bd74db-gardener-local-foooooo-external   71m
shoot--foooooo--baaaaaaar-internal   shoot--foooooo--baaaaaaar-b5c2177a-23ae-442a-a1fb-60e742bd74db-gardener-local-foooooo-internal   71m
```

Is that expected?

pkg/operation/botanist/extensions/dns/dnsowner_test.go
stoyanr: Yes, I know. However, the `fake` framework used by this test doesn't seem to support cluster scoped resources. If you omit the namespace, it will be created with a `default` namespace anyway, and subsequent retrieval with an empty namespace would fail. This is why I decided to keep the namespace and set it explicitly to default - I simply couldn't find another way to make this test pass. If you have an idea, let me know.

pkg/operation/botanist/extensions/dns/dnsowner_test.go
stoyanr: Yes, I know, but this is the only way I could find to make this test pass, see the explanation above.

stoyanr:
@ialidzhikov No, it's not expected, they should be deleted as well. Thanks for noticing this. I will fix this now and test some more reconciliation scenarios.

stoyanr:
@ialidzhikov The `DNSOwner` resources are properly deleted now when a shoot is deleted. I also tested:
* Enabling and disabling the ngnix ingress addon for a shoot
* Hibernating and waking up a shoot (with and without the nginx ingress addon)
* Migrating a shoot and then deleting it
* Creating and deleting a shoot w/o DNS

charts/seed-dns/owner/templates/dnsowner.yaml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: dns.gardener.cloud/v1alpha1
+kind: DNSOwner
+metadata:
+  name: {{ .Release.Namespace }}-{{ .Values.name }}
+spec:
+   ownerId: {{ .Values.ownerID }}
+   active: {{ .Values.active }}
```
rfranzke:
```suggestion
  ownerId: {{ .Values.ownerID }}
  active: {{ .Values.active }}
```

rfranzke:
PR looks great, no further comments, thanks for your extensive tests

stoyanr (dismissed):

ialidzhikov:
> @ialidzhikov The `DNSOwner` resources are properly deleted now when a shoot is deleted. I also tested:
> 
> * Enabling and disabling the ngnix ingress addon for a shoot
> * Hibernating and waking up a shoot (with and without the nginx ingress addon)
> * Migrating a shoot and then deleting it
> * Creating and deleting a shoot w/o DNS

Thank you for testing thoroughly. In the force push I don't see a unit test for this case - is it possible in general to add a unit test that would catch this?
As per https://github.com/gardener/documentation/blob/master/CONTRIBUTING.md#pull-request-checklist, could you please squash to a single commit and probably add @swilen-iwanow as [co-author](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors)?

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -459,12 +458,12 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation) *gardencorev1bet
...
 		destroyInternalDomainDNSRecord = g.Add(flow.Task{
-			Name:         "Destroying internal DNS entry",
-			Fn:           flow.TaskFn(component.OpWaiter(botanist.Shoot.Components.Extensions.DNS.InternalEntry).Destroy),
+			Name:         "Destroying internal domain DNS record",
+			Fn:           flow.TaskFn(botanist.DestroyInternalDNS),
 			Dependencies: flow.NewTaskIDs(syncPoint),
 		})
 		deleteDNSProviders = g.Add(flow.Task{
-			Name:         "Deleting DNS providers",
+			Name:         "Deleting additional DNS providers",
```
plkokanov, line 466: Is this step seems unnecessary considering the providers are deleted in DestroyInternalDNS and DestroyExternalDNS functions?

plkokanov (commented):
Looks good on my side :).

Just the additional unit tests as @ialidzhikov suggested.

stoyanr:
@ialidzhikov @plkokanov I added more unit tests for methods such as `DefaultNginxIngressDNSOwner`, `DefaultExternalDNSOwner`, and `DefaultInternalDNSOwner`, and `setAPIServerAddress`, and updated existing unit tests for `SetNginxIngressAddress`. However, I didn't add unit tests for methods that actually deploy, destroy, or migrate entries and owners. Such tests didn't exists and for a good reason:
* They add very little value, as the methods themselves perform `Deploy` of entries and owners created by the above methods. The unit tests for the above methods however also deploy them, so the duplication between tests would be quite significant.
* Besides deploying, these methods are also waiting for entries to become ready - and this is something that would be complicated to do in a unit tests, where there is no controller reconciling these entries. It could be done, but having in mind the previous considerations, I believe it's not worth doing it.

ghost:
Thank you @stoyanr for updating the PR. I will start a build. Once started, the build URL will be posted here.

ghost:
A build of this pull request has started. You can check on its progress here: https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-master/jobs/master-pull-request-job/builds/165

pkg/operation/botanist/extensions/dns/dnsowner_test.go
stoyanr: In fact, I found a way to avoid this by properly configuring the RESTMapper used by the test. Check the latest version of this PR.

pkg/operation/botanist/extensions/dns/dnsowner_test.go
stoyanr: In fact, I found a way to avoid this by properly configuring the RESTMapper used by the test. Check the latest version of this PR.

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -428,8 +427,8 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation) *gardencorev1bet
 		})
 
 		destroyNginxIngressDNSRecord = g.Add(flow.Task{
-			Name:         "Destroying ingress DNS record",
-			Fn:           flow.TaskFn(component.OpDestroyAndWait(botanist.Shoot.Components.Extensions.DNS.NginxEntry).Destroy),
+			Name:         "Destroying nginx ingress DNS record",
+			Fn:           flow.TaskFn(botanist.DestroyIngressDNSRecord),
```
ialidzhikov, line 431: Hmm, why `component.OpDestroyAndWait` is now switched to only `component.OpDestroy` (in `DestroyIngressDNSRecord` func)? I guess it now won't wait until the resources are deleted.

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -443,8 +442,8 @@ func (c *Controller) runDeleteShootFlow(o *operation.Operation) *gardencorev1bet
 			Dependencies: flow.NewTaskIDs(destroyInfrastructure),
 		})
 		destroyExternalDomainDNSRecord = g.Add(flow.Task{
-			Name:         "Destroying external DNS entry",
-			Fn:           flow.TaskFn(component.OpWaiter(botanist.Shoot.Components.Extensions.DNS.ExternalEntry).Destroy),
+			Name:         "Destroying external domain DNS record",
+			Fn:           flow.TaskFn(botanist.DestroyExternalDNS),
```
ialidzhikov, line 446: Same as above.

pkg/operation/botanist/extensions/dns/dnsowner.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"context"
+	"path/filepath"
+	"time"
+
+	dnsv1alpha1 "github.com/gardener/external-dns-management/pkg/apis/dns/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/gardener/gardener/pkg/client/kubernetes"
```
ialidzhikov, line 22: FYI, the convention at least in g/g is to have:
- 1st group - the golang imports
- 2nd group - the g/g imports
- 3rd group - non-g/g imports

ialidzhikov (commented):
@stoyanr , thank you for the tests.

pkg/gardenlet/controller/shoot/shoot_control_delete.go
ialidzhikov: @mvladev , for 1 more time I want to share that for me the "component" semantic is too messy - you have `Wait` vs `WaitForCleanup`; you have `Deploy` but with `invert=true` which does `Destroy`...

stoyanr: Fixed

pkg/gardenlet/controller/shoot/shoot_control_delete.go
stoyanr: Fixed

pkg/operation/botanist/extensions/dns/dnsowner.go
stoyanr: Fixed, thanks, I didn't know there is a convention, and my IDE seems to have its own opinion. Is the convention documented somewhere?

stoyanr:
@ialidzhikov Thanks for your comments, I addressed them now.

ialidzhikov (approved):
/lgtm

ialidzhikov:
ping @MartinWeindel @mandelsoft @danielfoehrKn

rfranzke (approved):
/lgtm

mandelsoft (approved):

PR #2577, opened 2020-07-10
Reserve excess capacity for etcd-main.

amshuman-kr:

/area control-plane
/area auto-scaling
/kind enhancement
/priority normal

If [gardener/kupid](https://github.com/gardener/kupid) is used as a gardener extension to schedule `etcd-main` workloads onto dedicated worker nodes, the existing reserve excess capacity pods will not reserve excess capacity on the dedicated worker nodes.

This PR introduces reserve excess capacity pods that will get scheduled on the same set of nodes as `etcd-main` workloads are likely to be scheduled.

Fixes #

1. For the reserve excess capacity pods, I have used the pod template with the same `selector` and `labels` as those of `etcd-main` to make sure they get injected with the same `nodeAffinity` and `tolerations` as `etcd-main`. Is this OK or should I introduce some specific label(s) for dedicated workload (e.g. `dedicated: etcd`)? Then I will need to modify the labels for `etcd-main` as well.
1. I have used the minimum values for `requests` assuming the minimum values were used to calculate the existing reserve capacity. Hope this is OK.
1. I have not reduced the `requests` for the common reserve excess capacity pods to subtract the `requests` for `etcd-main`. Should I?

**Release note**:

```improvement operator
Reserve excess capacity on the nodes where etcd-main workloads are likely to be scheduled. This helps if etcd-main workloads are scheduled on dedicated worker nodes in the seed cluster.
```

### DIFF
--- a/charts/seed-bootstrap/templates/reserve-excess-capacity/deployment-etcd.yaml
+++ b/charts/seed-bootstrap/templates/reserve-excess-capacity/deployment-etcd.yaml
@@ -1,0 +1,36 @@
+{{- if .Values.reserveExcessCapacity }}
+apiVersion: {{ include "deploymentversion" . }}
+kind: Deployment
+metadata:
+  name: reserve-excess-capacity-etcd
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: etcd-statefulset
+    role: main
+spec:
+  revisionHistoryLimit: 0
+  replicas: {{ index .Values.replicas "reserve-excess-capacity" }}
+  selector:
+    matchLabels:
+      app: etcd-statefulset
+      role: main
+  template:
+    metadata:
+      labels:
+        app: etcd-statefulset
+        role: main
+    spec:
+      terminationGracePeriodSeconds: 5
+      containers:
+      - name: pause-container
+        image: {{ index .Values.global.images "pause-container" }}
+        imagePullPolicy: IfNotPresent
+        resources:
+          requests:
+            cpu: 300m
+            memory: 1Gi
+          limits:
+            cpu: 900m
+            memory: 3Gi
+      priorityClassName: gardener-reserve-excess-capacity
+{{- end }}


vpnachev (commented):
Why should gardener take care of something that is configured outside or managed by other component?
It would be better whatever is creating or using the additional worker group to take care about the reserve excess capacity.

IMO, it would be better if kupid deploys the additional reserve excess capacity pods based on its configurations. In this way you can have as much as needed worker pools with configurable labels/tolerations/whatever.

/needs second-opinion

amshuman-kr:
> Why should gardener take care of something that is configured outside or managed by other component?

A very good point. But then this argument applies to regular reserve excess capacity equally well. Why should gardener take care of reserve excess capacity in general? It is something that should be the responsibility of whatever provisions the seeds.

> IMO, it would be better if kupid deploys the additional reserve excess capacity pods based on its configurations. In this way you can have as much as needed worker pools with configurable labels/tolerations/whatever.

kupid will also be a misfit for this purpose because kupid also doesn't provision the seeds and worker nodes. Ideally, this (reserve excess capacity in general) is best done in landscape setup (or in some subcomponent of landscape setup) wherever the seeds are actually provisioned.

vpnachev:
> A very good point. But then this argument applies to regular reserve excess capacity equally well. 

Not exactly, gardener simply deploys the reserve excess capacity deployment as a regular seed bootstrap and does not care about how the backing k8s cluster is configured. It does not take care about any specific configs like node labels or taints.

> Why should gardener take care of reserve excess capacity in general? It is something that should be the responsibility of whatever provisions the seeds.

Well, with the shooted seed approach gardener is not only provisioning seeds, but also takes care for the backing cluster. A general solution would be to deploy reserve excess capacity per worker pool for the shooted seeds, but for this case all of the taints should be explicitly listed in the shoot manifest.

The current proposal is too tightly coupled to a specific issue, that is under investigation in a single seed of a single landscape. Will we have to maintain in-code configs for any such attempt to solve some issue? I believe not.

amshuman-kr:
> A general solution would be to deploy reserve excess capacity per worker pool for the shooted seeds, but for this case all of the taints should be explicitly listed in the shoot manifest.

Agreed. Whatever component that prepares the shoot manifest for the seeds clusters should be in charge of what kind of reserve excess capacity is deployed in them.

> Not exactly, gardener simply deploys the reserve excess capacity deployment as a regular seed bootstrap and does not care about how the backing k8s cluster is configured. It does not take care about any specific configs like node labels or taints.

Of course. And in doing so, gardener makes at least the following two implicit assumptions which I am arguing are potentially unnecessary.
1. The seed clusters auto-scale.
1. All the worker nodes in the seed clusters are fungible.

> Well, with the shooted seed approach gardener is not only provisioning seeds, but also takes care for the backing cluster.

This has nothing to do with reserving excess capacity. Gardener might well be configured to take care of shooted seeds with cluser-autoscaling disabled, in which case, reserving excess capacity would be wasteful. I know that reserving excess capacity might be disabled by a flag but that actually argues my point that gardener shouldn't get into reserving excess capacity in seed clusters.

> The current proposal is too tightly coupled to a specific issue, that is under investigation in a single seed of a single landscape. Will we have to maintain in-code configs for any such attempt to solve some issue? I believe not.

Agreed. I am _not_ arguing for the current PR in which I limited myself to working with what exists.

I agree that a more general solution is desirable. Since such a general solution doesn't yet exist, I am proposing that such a general solution should take the complete problem of reserving excess capacity in seed clusters out of gardener, instead of merely taking out the partial problem of reserving excess capacity on dedicated nodes in the case of seed clusters with non-fungible worker nodes.

I am happy to pick up proposing/implementing such a general solution but I don't see any reason for a partial general solution for only reserving excess capacity on dedicated worker nodes.

timebertt:
I agree with @vpnachev, that if Gardener administrators configure something like kupid to move etcd to a separate worker pool from the outside, then capacity reservation should also be done from the outside.
In the simplest case, such a deployment would be configurable in the extension chart, so the administrator can configure the capacity reservation directly via the `ControllerRegistration`.

amshuman-kr:
The way I am thinking of the general solution, reserve excess capacity (general or per worker pool) can be deployed as a gardener extension. So, there will be no need for any changes in gardener. Hence, closing this PR.

PR #2580, opened 2020-07-13
Require Cluster identity values in helm charts

swilen-iwanow:

/area control-plane
/kind enhancement
/priority normal

Fixes #

@timuthy @tim-ebert, wdyt? I ran the hack/check-charts.sh, no errors. 
```
$ hack/check-charts.sh 
> Check Helm charts
All checks successful
```

**Release note**:

```improvement operator

```

### DIFF
--- a/charts/cluster-identity/templates/configmap-cluster-identity.yaml
+++ b/charts/cluster-identity/templates/configmap-cluster-identity.yaml
@@ -5,4 +5,4 @@ metadata:
   name: cluster-identity
   namespace: kube-system
 data:
-  cluster-identity: "{{ .Values.clusterIdentity }}"
+  cluster-identity: "{{ required ".Values.clusterIdentity is required" .Values.clusterIdentity }}"
--- a/charts/gardener/controlplane/charts/runtime/templates/apiserver/deployment.yaml
+++ b/charts/gardener/controlplane/charts/runtime/templates/apiserver/deployment.yaml
@@ -171,9 +171,7 @@ spec:
         - --audit-webhook-version={{ .Values.global.apiserver.audit.webhook.version }}
         {{- end }}
         - --authorization-always-allow-paths=/healthz
-        {{- if .Values.global.apiserver.clusterIdentity }}
-        - --cluster-identity={{ .Values.global.apiserver.clusterIdentity }}
-        {{- end }}
+        - --cluster-identity={{ required ".Values.global.apiserver.clusterIdentity is required" .Values.global.apiserver.clusterIdentity }}
         {{- if .Values.global.apiserver.disableAdmissionPlugins }}
         - --disable-admission-plugins={{ .Values.global.apiserver.disableAdmissionPlugins | join "," }}
         {{- end }}
--- a/charts/gardener/controlplane/values-test.yaml
+++ b/charts/gardener/controlplane/values-test.yaml
@@ -1,0 +1,3 @@
+global:
+  apiserver:
+    clusterIdentity: garden-cluster-identity
--- a/charts/seed-bootstrap/values-test.yaml
+++ b/charts/seed-bootstrap/values-test.yaml
@@ -1,0 +1,2 @@
+cluster-identity:
+  clusterIdentity: seed-cluster-identity
--- a/charts/shoot-core/components/values-test.yaml
+++ b/charts/shoot-core/components/values-test.yaml
@@ -1,0 +1,2 @@
+cluster-identity:
+  clusterIdentity: shoot-namesapce-shoot-ui-garden-cluster-identity
--- a/hack/check-charts.sh
+++ b/hack/check-charts.sh
@@ -26,10 +26,9 @@ if [[ -d "$1" ]]; then
     echo "$BROKEN_SYMLINKS"
     exit 1
   fi
-
   echo "Checking whether all charts can be rendered"
   for chart_dir in $(find charts -type d -exec test -f '{}'/Chart.yaml \;  -print -prune | sort); do
-    [ -f "$chart_dir/values-test.yaml" ] && values_files="-f $chart_dir/values-test.yaml"
+    [ -f "$chart_dir/values-test.yaml" ] && values_files="-f $chart_dir/values-test.yaml" || unset values_files
     helm template $values_files "$chart_dir" 1> /dev/null
   done
 fi
--- a/charts/cluster-identity/values-test.yaml
+++ b/charts/cluster-identity/values-test.yaml
@@ -1,0 +1,1 @@
+ clusterIdentity: shoot-namesapce-shoot-ui-garden-cluster-identity
--- a/charts/cluster-identity/values.yaml
+++ b/charts/cluster-identity/values.yaml
@@ -1,1 +1,1 @@
-clusterIdentity: shoot-namesapce-shoot-ui-garden-cluster-identity
+# clusterIdentity: shoot-namesapce-shoot-ui-garden-cluster-identity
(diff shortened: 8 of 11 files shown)


timuthy:
Yes, makes sense. Thanks for taking care! :)

charts/gardener/controlplane/charts/runtime/templates/apiserver/deployment.yaml
```diff
@@ -171,9 +171,7 @@ spec:
         - --audit-webhook-version={{ .Values.global.apiserver.audit.webhook.version }}
         {{- end }}
         - --authorization-always-allow-paths=/healthz
-        {{- if .Values.global.apiserver.clusterIdentity }}
-        - --cluster-identity={{ .Values.global.apiserver.clusterIdentity }}
-        {{- end }}
+        - --cluster-identity={{ required ".Values.global.apiserver.clusterIdentity" .Values.global.apiserver.clusterIdentity }}
```
timebertt:
```suggestion
        - --cluster-identity={{ required ".Values.global.apiserver.clusterIdentity is required" .Values.global.apiserver.clusterIdentity }}
```

charts/cluster-identity/templates/configmap-cluster-identity.yaml
```diff
@@ -5,4 +5,4 @@ metadata:
   name: cluster-identity
   namespace: kube-system
 data:
-  cluster-identity: "{{ .Values.clusterIdentity }}"
+  cluster-identity: "{{ required ".Values.clusterIdentity is required" .Values.clusterIdentity }}"
```
timebertt, line 8: The default for `.Values.clusterIdentity` is still provided here https://github.com/gardener/gardener/blob/9c50c7113a2b801594933d28757b0b1927eaffd0/charts/seed-bootstrap/values.yaml#L141
and here
https://github.com/gardener/gardener/blob/9c50c7113a2b801594933d28757b0b1927eaffd0/charts/shoot-core/components/values.yaml#L77

Can you comment those out, too?

timebertt (changes requested):
/assign

Thanks for taking care!
A few changes would be required from my POV.

timebertt:
> ```
> $ hack/check-charts.sh 
> > Check Helm charts
> All checks successful
> ```

Well, `hack/check-charts.sh` executes the checks for the first argument you pass ([ref](https://github.com/gardener/gardener/blob/a4d793a2c8559450a162d50475d546a2b628e776/hack/check-charts.sh#L31)).
So this will always succeed.

You have to run `hack/check-charts.sh ./charts` or `make check`.

As expected, the check will fail because of the required value.
```
$ ./hack/check-charts.sh ./charts
> Check Helm charts
Checking for chart symlink errors
Checking whether all charts can be rendered
Error: render error in "cluster-identity/templates/configmap-cluster-identity.yaml": template: cluster-identity/templates/configmap-cluster-identity.yaml:8:24: executing "cluster-identity/templates/configmap-cluster-identity.yaml" at <required ".Values.clusterIdentity is required" .Values.clusterIdentity>: error calling required: .Values.clusterIdentity is required
```

I will open a separate PR allowing to specify a `values-test.yaml` for this check.
I also found another issue with `check-charts.sh` which I want to address also.

swilen-iwanow:
> > ```
> > $ hack/check-charts.sh 
> > > Check Helm charts
> > All checks successful
> > ```
> 
> Well, `hack/check-charts.sh` executes the checks for the first argument you pass ([ref](https://github.com/gardener/gardener/blob/a4d793a2c8559450a162d50475d546a2b628e776/hack/check-charts.sh#L31)).
> So this will always succeed.
> 
> You have to run `hack/check-charts.sh ./charts` or `make check`.
> 
> As expected, the check will fail because of the required value.
> 
> ```
> $ ./hack/check-charts.sh ./charts
> > Check Helm charts
> Checking for chart symlink errors
> Checking whether all charts can be rendered
> Error: render error in "cluster-identity/templates/configmap-cluster-identity.yaml": template: cluster-identity/templates/configmap-cluster-identity.yaml:8:24: executing "cluster-identity/templates/configmap-cluster-identity.yaml" at <required ".Values.clusterIdentity is required" .Values.clusterIdentity>: error calling required: .Values.clusterIdentity is required
> ```
> 
> I will open a separate PR allowing to specify a `values-test.yaml` for this check.
> I also found another issue with `check-charts.sh` which I want to address also.

Yes. I saw this. I will wait for the PR mentioned. Will fix the other things also. Thank you!

timebertt:
#2584 is merged, please rebase and add the `values-test.yaml` for the respective charts.
/unhold
/needs rebase
/status author-action

charts/gardener/controlplane/values-test.yaml
```diff
@@ -0,0 +1,326 @@
+global:
```
vpnachev, line 1: Instead of coping the entire file, you can set just the required parameters
```yaml
global:
  # Gardener API server configuration values
  apiserver:
    clusterIdentity: garden-cluster-identity
```

charts/seed-bootstrap/values-test.yaml
```diff
@@ -0,0 +1,225 @@
+cloudProvider: aws
```
vpnachev: Here it is the same, set just the required parameters.

charts/shoot-core/components/values-test.yaml
```diff
@@ -0,0 +1,77 @@
+global:
```
vpnachev: Also, here - set only the required parameters.

hack/check-charts.sh
```diff
@@ -26,10 +26,10 @@ if [[ -d "$1" ]]; then
     echo "$BROKEN_SYMLINKS"
     exit 1
   fi
-
+
```
vpnachev:
```suggestion

```

vpnachev (changes requested):
To avoid double maintenance, in `values-test.yaml` files set only the required parameters.

swilen-iwanow:
> To avoid double maintenance, in `values-test.yaml` files set only the required parameters.

@vpnachev If we don't have all values inside the values-test.yaml, those chart folders with the test file wouldn't try to render non required values, that are inside the values.yaml, ain't that the idea behind the check-charts.sh?

vpnachev:
Helm is always considering the available `values.yaml` file (even it it is not explicitly specified with `-f`), so the charts are actually rendered with values from both `values.yaml` and `values-test.yaml`, as the later just overwrite some of the values. You can see more details here https://helm.sh/docs/chart_template_guide/values_files/

swilen-iwanow:
> Helm is always considering the available `values.yaml` file (even it it is not explicitly specified with `-f`), so the charts are actually rendered with values from both `values.yaml` and `values-test.yaml`, as the later just overwrite some of the values. You can see more details here https://helm.sh/docs/chart_template_guide/values_files/

I see, thank you!

hack/check-charts.sh
```diff
@@ -26,10 +26,10 @@ if [[ -d "$1" ]]; then
     echo "$BROKEN_SYMLINKS"
     exit 1
   fi
-
+  
   echo "Checking whether all charts can be rendered"
   for chart_dir in $(find charts -type d -exec test -f '{}'/Chart.yaml \;  -print -prune | sort); do
-    [ -f "$chart_dir/values-test.yaml" ] && values_files="-f $chart_dir/values-test.yaml"
+    [ -f "$chart_dir/values-test.yaml" ] && values_files="-f $chart_dir/values-test.yaml" || unset values_files
```
timebertt, line 31: Nice catch! :)

vpnachev (approved):
/lgtm

timebertt (approved):
Thanks!
/lgtm

PR #2581, opened 2020-07-13
Don't invalidate clients on version change

timebertt:

/area quality
/kind enhancement
/priority normal

Previously, the generic ClientMap invalidated ClientSets if the server version of the targeted cluster changed.
Now, it just refreshes the internals (saved server version and chart renderer) when the version changes, so the caches/informers keep running.

Now, the garden and seed clients are never invalidated and kept running during the whole lifetime of the gardenlet instance, unless the gardenlet is responsible for multiple Seeds, which is not recommended for production use anyways (in that case the Seed clients may be invalidated/refreshed when the kubeconfig secret changes).
Therefore, it is now possible to use the informers retrieved from the controller-runtime cache to construct controllers the same way like with informers created by a `SharedInformerFactory`. I will open a second PR which makes use of this in the BackupBucket/BackupEntry controller as a simple example for that.

Part of #2414 

/cc @timuthy 

**Release note**:

```noteworthy developer
You can now use the controller-runtime cache of our ClientSets (`kubernetes.Interface.Cache()`) to obtain informers for arbitrary API objects (via `GetInformer/GetInformerForKind`) to construct controllers.
```

### DIFF
--- a/pkg/client/kubernetes/client.go
+++ b/pkg/client/kubernetes/client.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/gardener/gardener/pkg/chartrenderer"
 	gardencoreclientset "github.com/gardener/gardener/pkg/client/core/clientset/versioned"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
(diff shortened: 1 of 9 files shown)


timuthy (approved):
/lgtm

PR #2582, opened 2020-07-13
Fix flapping Plant region

timebertt:

/area quality
/kind bug
/priority normal

Previously, the Plant controller falsely tried to obtain the region of the Plant cluster from the zone Node label, although this is supposed to be a fallback to the region label.
As maps are unsorted in go, the plant controller sometimes used the zone and sometimes the region for reporting the Plant's CloudInfo.
This PR fixes this behaviour and adds an additional cross-check to the Plant integration test.

Fixes #

**Release note**:

```improvement operator
A bug has been fixed, which caused the discovered Plant region to alternate between region and zone of the Nodes.
```

### DIFF
--- a/pkg/controllermanager/controller/plant/plant_utils.go
+++ b/pkg/controllermanager/controller/plant/plant_utils.go
@@ -79,15 +79,20 @@ func getCloudProviderForNode(providerID string) string {
 	return provider[0]
 }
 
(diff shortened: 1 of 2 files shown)


timuthy (approved):
/lgtm
Nice catch!

schrodit (approved):

PR #2583, opened 2020-07-13
remove the random order of fluent-bit extension parsers and filter

vlvasilev:

/area logging
/kind bug
/priority normal

When `CachedRuntimeClients` feature gate is activated, every time fluent-bit config map is computed it end up with different order of its extension parser and filter which triggers fluent-bit DaemonSet reconciliation and all existing pod are killed and replaces with new ones

Fixes #

**Release note**:

```improvement operator
An issue causing fluent-bit Pods to be restarted because of a new ConfigMap checksum when the CachedRuntimeClients feature gate is enabled, is now fixed.
```

### DIFF
--- a/pkg/operation/seed/seed.go
+++ b/pkg/operation/seed/seed.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


ialidzhikov:
Ref https://github.com/gardener/gardener/pull/2583#issuecomment-657637421

/status author-action

vpnachev:
```bash
$ git commit --amend --reset-author
```

ialidzhikov (approved):
/lgtm

PR #2586, opened 2020-07-14
Switch BB/BE controller to controller-runtime informers

timebertt:

/area quality
/kind enhancement
/priority normal
/size s

This PR replaces the Informers created by the `SharedInformerFactory` in the BackupBucket and BackupEntry controllers by informers from the controller-runtime cache. 
This way, we rely on the same watch for those objects, which enables us to use the cached client in the respective reconcilers. This is analogous to the way we do it in many other controllers with client-go-style Informers and Listers.

This can be seen as a simple example on how to construct/refactor controllers in gardener by only relying on controller-runtime clients and informers.

Ref #2581 
Ref #2568

/cc @timuthy 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/controllerutils/seedfilter.go
+++ b/pkg/controllerutils/seedfilter.go
@@ -15,11 +15,14 @@
 package controllerutils
 
 import (
+	"context"
+
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	gardencorelisters "github.com/gardener/gardener/pkg/client/core/listers/core/v1beta1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // LabelsMatchFor checks whether the given label selector matches for the given set of labels.
@@ -79,6 +82,17 @@ func SeedLabelsMatch(seedLister gardencorelisters.SeedLister, seedName string, l
 	return LabelsMatchFor(seed.Labels, labelSelector)
 }
 
+// seedLabelsMatchWithClient fetches the given seed by its name from the client and then checks whether the given
+// label selector matches the seed labels.
+func seedLabelsMatchWithClient(ctx context.Context, c client.Client, seedName string, labelSelector *metav1.LabelSelector) bool {
+	seed := &gardencorev1beta1.Seed{}
+	if err := c.Get(ctx, client.ObjectKey{Name: seedName}, seed); err != nil {
+		return false
+	}
+
+	return LabelsMatchFor(seed.Labels, labelSelector)
+}
+
 // ControllerInstallationFilterFunc returns a filtering func for the seeds and the given label selector.
 func ControllerInstallationFilterFunc(seedName string, seedLister gardencorelisters.SeedLister, labelSelector *metav1.LabelSelector) func(obj interface{}) bool {
 	return func(obj interface{}) bool {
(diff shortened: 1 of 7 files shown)


timuthy:
@tim-ebert I suppose you're going to adapt the PR to the outcome of the discussion in #2588? I'll take a look again later then.

timebertt:
> @tim-ebert I suppose you're going to adapt the PR to the outcome of the discussion in #2588? I'll take a look again later then.

Yes, will do :)

timebertt:
@timuthy I adapted the PR according to our discussion in https://github.com/gardener/gardener/pull/2588#discussion_r455624584.

Now the informers are retrieved in the `new` funcs and errors get propagated up to `gardenlet.go`. EventHandlers are still added in the `run` funcs, but they don't return errors so that's nice.
I have to admit, I like it much better now :)

pkg/gardenlet/controller/backupbucket/backup_bucket.go
```diff
@@ -56,45 +60,40 @@ type Controller struct {
...
-	backupBucketInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
-		FilterFunc: controllerutils.BackupBucketFilterFunc(confighelper.SeedNameFromSeedConfig(config.SeedConfig), seedLister, config.SeedSelector),
-		Handler: cache.ResourceEventHandlerFuncs{
-			AddFunc:    backupBucketController.backupBucketAdd,
-			UpdateFunc: backupBucketController.backupBucketUpdate,
-			DeleteFunc: backupBucketController.backupBucketDelete,
-		},
-	})
-
-	backupBucketController.backupBucketSynced = backupBucketInformer.Informer().HasSynced
```
timuthy, line 85: Can you revert the removal of the hasSycned handling? IIUC, we agreed on checking for the synced state since it is part of the informer interface and the fact that we don't need it at the moment, because the Controller-Runtime never returns an unsynced informer, is only an implementation detail.

pkg/controllerutils/seedfilter.go
```diff
@@ -79,6 +82,17 @@ func SeedLabelsMatch(seedLister gardencorelisters.SeedLister, seedName string, l
 	return LabelsMatchFor(seed.Labels, labelSelector)
 }
 
+// SeedLabelsMatchWithClient fetches the given seed by its name from the client and then checks whether the given
+// label selector matches the seed labels.
+func SeedLabelsMatchWithClient(ctx context.Context, c client.Client, seedName string, labelSelector *metav1.LabelSelector) bool {
```
timuthy:
```suggestion
func seedLabelsMatchWithClient(ctx context.Context, c client.Client, seedName string, labelSelector *metav1.LabelSelector) bool {
```

pkg/gardenlet/controller/backupentry/backup_entry.go
```diff
@@ -51,45 +55,40 @@ type Controller struct {
...
-	backupEntryInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
-		FilterFunc: controllerutils.BackupEntryFilterFunc(confighelper.SeedNameFromSeedConfig(config.SeedConfig), seedLister, config.SeedSelector),
-		Handler: cache.ResourceEventHandlerFuncs{
-			AddFunc:    backupEntryController.backupEntryAdd,
-			UpdateFunc: backupEntryController.backupEntryUpdate,
-			DeleteFunc: backupEntryController.backupEntryDelete,
-		},
-	})
-
-	backupEntryController.backupEntrySynced = backupEntryInformer.Informer().HasSynced
```
timuthy, line 80: See previous comment about hasSynced.

timuthy (changes requested):
Looks good in general, just a few suggestions.

pkg/gardenlet/controller/backupbucket/backup_bucket.go
timebertt: Yes, we agreed on that one. I just forgot to adapt it here.
Done now :)

timuthy (approved):
/lgtm
Thanks for addressing the comments so quickly.

rfranzke (approved):

PR #2587, opened 2020-07-14
Ensure Seed client is invalidated after deletion

timebertt:

/area quality
/kind bug
/priority normal

This PR ensures the seed clients are invalidated after the seed deletion.
This wasn't always the case before, because the seed lease controller always requeues the Seed also after successful deletion and could therefore create a new clientset again, even though the seed might already have been completely deleted.

Fixes #

**Release note**:

```improvement operator
A bug has been fixed, which caused Seed clients not to be invalidated properly on Seed deletion.
```

### DIFF
--- a/pkg/gardenlet/controller/seed/seed_control.go
+++ b/pkg/gardenlet/controller/seed/seed_control.go
@@ -91,6 +91,10 @@ func (c *Controller) reconcileSeedKey(key string) error {
 	seed, err := c.seedLister.Get(name)
 	if apierrors.IsNotFound(err) {
 		logger.Logger.Debugf("[SEED RECONCILE] %s - skipping because Seed has been deleted", key)
+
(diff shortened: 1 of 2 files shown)


timuthy:
> This wasn't always the case before, because the seed lease controller always requeues the Seed also after successful deletion and could therefore create a new clientset again, even though the seed might already have been completely deleted.

Is this because of the stale `seedlister` cache?

Tbh, I'm a bit concerned by this distributed invalidation handling. I'm currently working on a PR and it struck me because the invalidation happend too early :(

timuthy:
@tim-ebert shall we merge this change and think about a general solution later or what would you prefer?

timebertt:
> shall we merge this change and think about a general solution later

Yes, I think that makes sense.
Let's try to figure out a proper way to do it later.

timuthy (approved):
/lgtm

Issue #2489, opened 2020-06-23
Update K8s dependencies to 1.18.x


timuthy:
**How to categorize this issue?**

/area robustness
/kind enhancement
/priority normal

**What would you like to be added**:
Update the `vendor` sources in g/g to K8s 1.18.x.

timuthy:
To track progress: https://github.com/timuthy/gardener/tree/maintenance.vendor

timuthy:
Update: As soon as #2588 has been merged, we proceed with the actual update work.

phil9909:
FYI: This blocks us from updating [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/) from `0.5.x` to `0.6.x`.

We import `github.com/gardener/gardener/pkg/apis/core/v1beta1` in our go-code to dynamically create shoots.

timebertt:
Why is this blocking you?
If you are only importing the API packages, you should be good to go for using different versions of k/k or controller-runtime. Right?

phil9909:
This is what I assumed, but seems not to be true:

```
go build -o bin/manager main.go
# github.com/gardener/gardener/pkg/apis/core/v1beta1
../../../go/pkg/mod/github.com/gardener/gardener@v1.8.0/pkg/apis/core/v1beta1/conversions.go:84:15: scheme.AddConversionFuncs undefined (type *runtime.Scheme has no field or method AddConversionFuncs)
Makefile:57: recipe for target 'build' failed
make: *** [build] Error 2

```

`conversions.go` is apparently broken with k8s 1.18 dependencies

timebertt:
Ah ok, makes sense.
Thanks for explaining!

franpog859:
Hello! May I ask you when this update will be done? I currently use these versions of k8s dependencies in my component:
```
require (
	github.com/gardener/gardener v1.8.1
	...
	k8s.io/api v0.18.8
	k8s.io/apimachinery v0.18.8
	k8s.io/client-go v0.18.8
	sigs.k8s.io/controller-runtime v0.6.2
```
and the old Gardener's dependencies seem not to allow me to build it at all 😞
```
# github.com/gardener/gardener/pkg/apis/core/v1beta1
/Users/franpog859/go/pkg/mod/github.com/gardener/gardener@v1.8.1/pkg/apis/core/v1beta1/conversions.go:84:15: scheme.AddConversionFuncs undefined (type *runtime.Scheme has no field or method AddConversionFuncs)
```

[...]

PR #2588, opened 2020-07-14
Switch fed seed controllers to cached informers

timuthy:

/area control-plane
/area dev-productivity
/kind enhancement
/priority normal

This PR makes the federated seed controllers use the [Controller-Runtime](https://github.com/kubernetes-sigs/controller-runtime) managed and cached informers. It's a further and consequent step towards the Controller-Runtime and allows to get rid of any third party generated client code (e.g. from external-dns-management) which turned out to be a blocker for #2489.

I tested several scenarios, e.g. shoot creation/deletion, adding/removing workers with "new" OS type, seed creation/deletion, provoking errors in the extension controllers, etc.. It'd be nice if you could double check the state reconciliation as well (/cc @plkokanov @swilen-iwanow).

**Release note**:

```improvement operator
NONE
```


swilen-iwanow:
I can check it out : )

pkg/gardenlet/controller/federatedseed/extensions/artifact.go
```diff
@@ -47,27 +47,54 @@ type artifact struct {
...
-		containerRuntimeInformer      = extensionsInformers.Extensions().V1alpha1().ContainerRuntimes()
-		controlPlaneInformer          = extensionsInformers.Extensions().V1alpha1().ControlPlanes()
-		extensionInformer             = extensionsInformers.Extensions().V1alpha1().Extensions()
-		infrastructureInformer        = extensionsInformers.Extensions().V1alpha1().Infrastructures()
-		networkInformer               = extensionsInformers.Extensions().V1alpha1().Networks()
-		operatingSystemConfigInformer = extensionsInformers.Extensions().V1alpha1().OperatingSystemConfigs()
-		workerInformer                = extensionsInformers.Extensions().V1alpha1().Workers()
-	)
+func (c *controllerArtifacts) initialize(seedClient kubernetes.Interface) error {
+	dnsProviderInformer, err := seedClient.Cache().GetInformer(&dnsv1alpha1.DNSProvider{})
```
timebertt: WDYT about getting the respective Informers by using the `newObjectFunc` inside `registerExtensionControllerArtifacts` instead of creating them explicitly here?

Something like this:
```
	informer, err := cache.GetInformer(artifact.newObjFunc())
	if err != nil {
		return err
	}
	artifact.informer = informer
```

pkg/gardenlet/controller/federatedseed/seed_control.go
```diff
@@ -166,15 +162,15 @@ func (c *Controller) createReconcileSeedRequestFunc(ctx context.Context) reconci
 	return func(req reconcile.Request) (reconcile.Result, error) {
 		seed, err := c.seedLister.Get(req.Name)
 		if apierrors.IsNotFound(err) {
-			logger.Logger.Infof("Skipping - Seed %s was not found", req.Name)
+			logger.Logger.Infof("Skipping federated seed setup - Seed %s was not found", req.Name)
```
timebertt: Can you also switch the seed informer and lister in the top-level controller to the controller-runtime informer and client?

https://github.com/gardener/gardener/blob/6e528916105623fee733d2b3c043385375a08761/pkg/gardenlet/controller/federatedseed/seed_control.go#L84-L91
and
https://github.com/gardener/gardener/blob/6e528916105623fee733d2b3c043385375a08761/pkg/gardenlet/controller/federatedseed/seed_control.go#L163

pkg/gardenlet/controller/federatedseed/networkpolicy/namespace_control.go
```diff
@@ -107,15 +106,13 @@ func (r *namespaceReconciler) Reconcile(request reconcile.Request) (reconcile.Re
 	}
```
timebertt: Can you check that if statement here?
I think its superfluous, as `err` is always `nil` here.

pkg/gardenlet/controller/federatedseed/networkpolicy/controller.go
```diff
@@ -64,16 +66,18 @@ type Controller struct {
 }
 
 // NewController instantiates a new controller.
-func NewController(ctx context.Context, seedClient kubernetes.Interface, seedDefaultNamespaceKubeInformer kubeinformers.SharedInformerFactory, seedKubeInformerFactory kubeinformers.SharedInformerFactory, seedLogger *logrus.Entry, recorder record.EventRecorder, seedName string) *Controller {
+func NewController(ctx context.Context, seedClient kubernetes.Interface, seedDefaultNamespaceKubeInformer kubeinformers.SharedInformerFactory, seedLogger *logrus.Entry, recorder record.EventRecorder, seedName string) (*Controller, error) {
+	networkPoliciesInformer, err := seedClient.Cache().GetInformer(&networkingv1.NetworkPolicy{})
+	if err != nil {
+		return nil, err
+	}
```
timebertt, line 71: What would be your preference on this: initialize/get informers in the controller constructor (`NewController` here) or doing it in when starting the controller in `Run` along with adding the event handlers?
In https://github.com/gardener/gardener/pull/2586/files#diff-8047290eb35b00e8dbcbaec8e091f641R82-R94 I chose the second option, but I am open to do it otherwise.
Do we want to agree on one of the options as a guideline?

timebertt (commented):
Very nice, I like it.

Only have some nit/general comments.

pkg/gardenlet/controller/federatedseed/networkpolicy/controller.go
timuthy: I don't have a preference and can adapt the PR to do it in the `Run` function. One advantage is that the error handling can be omitted in `NewController `.

timebertt: Well, that's a good point. Then let's do it in the `Run` function.
Can you then also move this part to the `Run` function?
https://github.com/gardener/gardener/blob/6e528916105623fee733d2b3c043385375a08761/pkg/gardenlet/controller/federatedseed/extensions/extensions.go#L78-L80

timuthy: While adapting these changes I realized that getting the informers during `run` may not be the best idea in general. 

Some `run` funcs (see [here](https://github.com/gardener/gardener/blob/c0c5a694ae28d90f06c080008f5464d161fb30d3/pkg/gardenlet/controller/factory.go#L174-L179)) are already called in their own Goroutine and handling such "setup" errors like getting an informer would require the implementation of an error channel. While I'm not inherently against error channels, I think it's not necessary for these cases because the error could have already been checked during the construction in a `new` func or similar.

timebertt: I see, that's why I used `panic` for error handling during controller start in #2586 
If the controllers are not able to start, because they can't get the necessary informers, we can't do anything anyways even with proper error handling. So I think `panic` is fine for that case.
For the federated Seed controllers OTOH, `panic` is of course no solution. But I think, we already had error handling for the `Run` funcs before, right?
I am also open to get informers it in the `new` funcs, although I think that handling the errors there, will basically lead to the same result for the "normal" controllers (-> let gardenlet exit with error code).
WDYT?

timuthy: > So I think panic is fine for that case.

We should only use a panic (if possible) when we are in a general intialization, i.e. before controllers or futher long running Goroutines are started. The panic can happen when other controllers already run and thus they are terminated ungracefully. Of course, (most) actions are idempotent but still, not very nice.

> But I think, we already had error handling for the Run funcs before, right?

Not really, because it was not necessary, except when we waited for caches to sync ([example](https://github.com/gardener/gardener/blob/f120f2c74a503ab4a388aa055cd08d4081eaf5ed/pkg/gardenlet/controller/backupbucket/backup_bucket.go#L94-L97)) and this case was more or less disregarded because it was quite unlikely to happen. I just saw that we have this panic usage [here](https://github.com/gardener/gardener/blob/c0c5a694ae28d90f06c080008f5464d161fb30d3/pkg/gardenlet/controller/factory.go#L93) as well. But for the reason I explained above, I'm not really convinced yet that this is the way to go.

If we already have the chance to detect runtime errors early during initialization why not use this advantage and exit  before the controllers might go ahead and alter the system?

timebertt: > The panic can happen when other controllers already run and thus they are terminated ungracefully.

Yes, you're right. Just thought about it again and it may indeed take some time for the calls to `GetInformer` to return an error (waiting for the cache might timeout or whatever) and then other controllers might already run. Then it would be better of course to have proper error handling. I just thought, that the errors might arise directly, but that's not the case.
So agreed :)

Let's move the informer creation/retrieval to the `new` funcs then.
And how would you like to handle waiting for the caches then? This should not happen in the `new` func, I think, because then the different controllers can't wait for the respective caches to sync in parallel and it is rather a serial and blocking initialization.
So if we want to handle that gracefully, we would have to implement proper error handling for the `run` funcs also, right?

timuthy: > And how would you like to handle waiting for the caches then?

I wouldn't touch that now and leave it the way it is done in the `run` functions today. I've never seen this causing us any trouble.

timebertt: Ok, agreed. 👍

pkg/gardenlet/controller/federatedseed/extensions/artifact.go
timuthy: Good idea. I refactored the artifact handling a bit because the state and controllerinstallation information was mixed up too much, imo.

timuthy:
Thanks for your review @tim-ebert. I addressed the comments as discussed and also simplified the [artifact](https://github.com/gardener/gardener/blob/ea1778e29570afb3681cd01bb86eadb14d94ac9b/pkg/gardenlet/controller/federatedseed/extensions/artifact.go) handling slightly.

swilen-iwanow:
Very nice, I like how you did the init function of the informers 🤩! Tested the shootstate controller with this PR and it seems to work as it used to. 🚀

pkg/gardenlet/controller/factory.go
```diff
@@ -157,9 +157,11 @@ func (f *GardenletControllerFactory) Run(ctx context.Context) {
 		controllerInstallationController = controllerinstallationcontroller.NewController(f.clientMap, f.k8sGardenCoreInformers, f.cfg, f.recorder, gardenNamespace)
 		seedController                   = seedcontroller.NewSeedController(f.clientMap, f.k8sGardenCoreInformers, f.k8sInformers, f.healthManager, secrets, imageVector, componentImageVectors, f.identity, f.cfg, f.recorder)
 		shootController                  = shootcontroller.NewShootController(f.clientMap, f.k8sGardenCoreInformers, f.cfg, f.identity, f.gardenClusterIdentity, secrets, imageVector, f.recorder)
-		federatedSeedController          = federatedseedcontroller.NewFederatedSeedController(f.clientMap, f.k8sGardenCoreInformers, f.cfg, f.recorder)
 	)
 
+	federatedSeedController, err := federatedseedcontroller.NewFederatedSeedController(ctx, f.clientMap, f.cfg, f.recorder)
+	runtime.Must(err)
```
timebertt: I have added an error return parameter to this func in #2586.
Can you rebase and return an error also here at this line?

pkg/gardenlet/controller/federatedseed/seed_control.go
```diff
@@ -134,6 +130,17 @@ func (c *Controller) seedDelete(obj interface{}) {
...
+			UpdateFunc: c.seedUpdate,
+			DeleteFunc: c.seedDelete,
+		},
+	})
+
+	c.seedHasSynced = c.seedInformer.HasSynced
+
 	var waitGroup sync.WaitGroup
 
 	if !cache.WaitForCacheSync(ctx.Done(), c.seedHasSynced) {
```
timebertt: Can you remove the `seedHasSynced` member?
```suggestion
	var waitGroup sync.WaitGroup

	if !cache.WaitForCacheSync(ctx.Done(), c.seedInformer.HasSynced) {
```

pkg/gardenlet/controller/federatedseed/seed_control.go
```diff
@@ -162,19 +169,19 @@ func (c *Controller) Run(ctx context.Context, workers int) {
 	waitGroup.Wait()
 }
 
-func (c *Controller) createReconcileSeedRequestFunc(ctx context.Context) reconcile.Func {
+func (c *Controller) createReconcileSeedRequestFunc(ctx context.Context, gardenClient client.Client) reconcile.Func {
```
timebertt: Hmm, `Controller` already has a `gardenClient` and also a `clientMap`. Why are you passing it here again?

timebertt (commented):
Looks great, only a few minor comments :)
/needs rebase

timuthy:
Thanks for your review @tim-ebert. PTAL.

timebertt (approved):
/lgtm

PR #2591, opened 2020-07-15
Use DirectClient to wait for OSC

timebertt:

/area quality
/kind bug
/priority normal

When adapting the gardenlet's seed client usage in f120f2c74a503ab4a388aa055cd08d4081eaf5ed, I missed one usage of `common.WaitUntilExtensionCRReady` in `applyAndWaitForShootOperatingSystemConfig`, which should have also been switched to the `DirectClient`.
Therefore, the `cloud-config` script in the Shoot was not properly updated after some change in the worker config or similar in case the `CachedRuntimeClients` feature gate was activated. In that case, only after a second reconciliation the secret was updated correctly.
This PR fixes this bug.

Fixes #

/cc @zanetworker 

**Release note**:

```improvement operator
A bug has been fixed, which caused the `cloud-config` secret in the Shoot to not get updated correctly after a change to the worker config in case the `CachedRuntimeClients` feature gate was activated.
```

### DIFF
--- a/pkg/operation/botanist/operatingsystemconfig.go
+++ b/pkg/operation/botanist/operatingsystemconfig.go
@@ -366,7 +366,7 @@ func (b *Botanist) applyAndWaitForShootOperatingSystemConfig(ctx context.Context
 
 	if err := common.WaitUntilExtensionCRReady(
 		ctx,
-		b.K8sSeedClient.Client(),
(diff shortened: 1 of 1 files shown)


zanetworker (approved):
/lgtm

timuthy (approved):

danielfoehrKn (approved):

PR #2597, opened 2020-07-17
Use DirectClient to wait for deletion of Worker/ControlPlane

timebertt:

/area quality
/kind cleanup
/priority normal

I have found a few more places, where we should rather rely on the direct client when waiting (e.g. for deletion of Worker/ControlPlane) that haven't been adapted yet.
I haven't encountered any issues with these, but I would like to avoid any in some edge cases we haven't maybe hit yet.

Fixes #

**Release note**:

```improvement operator
NONE
```


ialidzhikov (approved):
/lgtm

PR #2605, opened 2020-07-19
Nit: Improve package name

ialidzhikov:

/area usability
/priority normal

Ref https://blog.golang.org/package-names. 

> By convention, the last element of the package path is the package name:

Currently when the package name != last element of package path, it can lead to collisions with other packages and this requires explicitly setting import identifier.

**Release note**:

```improvement operator
NONE
```


rfranzke (approved):
/lgtm

timebertt (approved):

PR #2606, opened 2020-07-20
Scale down VPA on shoot hibernation

rfranzke:

/area auto-scaling
/kind bug
/priority normal

The VPA components for the shoots weren't scaled down on hibernation. This PR fixes the behaviour.
Also, a nil pointer exception that occurred when a new shoot was created with VPA=enabled has been fixed.

Thanks @wyb1 for noticing!

**Release note**:

```improvement operator
The VPA for shoots running in the seed is now correctly scaled down when the shoot is being hibernated.
```


wyb1 (approved):
Thanks! /lgtm

timuthy (approved):
/lgtm

PR #2532, opened 2020-06-30
Fix flaky tests

rfranzke:
**How to categorize this PR?**

/area testing
/kind technical-debt
/priority normal

**What this PR does / why we need it**:
This PR fixes flaky tests in the `healtzh` and `shootsecrets` packages.

Ref https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-master/jobs/master-head-update-job/builds/283 and https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-master/jobs/master-head-update-job/builds/279

**Release note**:

```improvement operator
NONE
```

PR #2609, opened 2020-07-21
Fix health manager

timebertt:

/area quality
/area robustness
/kind bug
/priority normal

The periodic health manager seemed to be unstable under CPU pressure/throttling (e.g. in our CI pipelines https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-master/jobs/master-pull-request-job/builds/17).
While I couldn't get the failing unit tests to break on my machine, it was easy to reproduce while running in a docker container with restricted CPU shares.

This PR makes the health manager(s) more robust for concurrent access by accurately locking access to the managers' state members.

Ref #2532 

**Release note**:

```improvement operator
The gardenlet's `/healthz` endpoint has been improved to be more stable under certain circumstances like CPU throttling.
```

### DIFF
--- a/pkg/healthz/default_test.go
+++ b/pkg/healthz/default_test.go
@@ -49,12 +49,29 @@ var _ = Describe("Default", func() {
 
 		Describe("#Set", func() {
 			It("should correctly set the status to true", func() {
+				d.Start()
 				d.Set(true)
 				Expect(d.health).To(BeTrue())
 			})
 
 			It("should correctly set the status to false", func() {
+				d.Start()
 				d.Set(false)
+				Expect(d.health).To(BeFalse())
+			})
+
+			It("should not set the status to true (HealthManager not started)", func() {
+				d.Set(true)
+				Expect(d.health).To(BeFalse())
+			})
+
+			It("should not set the status to true (HealthManager already stopped)", func() {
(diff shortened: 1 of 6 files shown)


pkg/healthz/default_test.go
```diff
@@ -49,25 +49,42 @@ var _ = Describe("Default", func() {
...
+				d.Set(true)
+				Expect(d.health).To(BeFalse())
+			})
 		})
 
 		Describe("#Get", func() {
 			It("should get the correct status (true)", func() {
 				d.health = true
-				Expect(d.health).To(BeTrue())
+				Expect(d.Get()).To(BeTrue())
```
rfranzke, line 82: ... thanks :)

rfranzke (approved):
/lgtm

Thank you!

vpnachev (approved):
/lgtm

I am just curious why sometimes `.health` is used, and sometimes `.Get()` in the tests?

rfranzke:
@vpnachev In the tests for the `Get()` function the `Get()` function is now used correctly. In all other tests the `.health` is read directly to decouple these tests from `Get()` so that they don't fail in case `Get()` has a bug.

I think we can go ahead with the /merge.

Issue #2572, opened 2020-07-08
seed deletion fails due to provider controllerinstallation not being deleted


Diaphteiros:
**How to categorize this issue?**

/area control-plane
/kind bug
/priority normal

**What happened**:
Tried to delete a seed. Seed is now stuck in deletion.
Logs show that the provider controllerinstallation still exists, blocking seed deletion. The controllerinstallation doesn't have a deletion timestamp. Deleting it seems to resolve the problem.

**What you expected to happen**:
Extensions are cleaned up and seed is deleted.

**How to reproduce it (as minimally and precisely as possible)**:
Deploy Gardener into a cluster (with virtual cluster setup) and register that cluster as a seed by deploying the gardenlet into it. Then delete the seed with `kubectl delete seed <my-seed>`

**Anything else we need to know?**:
There are no relevant events on the seed resource.

Logs from the AWS provider extension pod:
```
{"level":"info","ts":"2020-07-08T07:08:57.531Z","logger":"controller-runtime.controller","msg":"Starting Controller","controller":"worker_state_controller"}
{"level":"info","ts":"2020-07-08T07:08:57.531Z","logger":"controller-runtime.controller","msg":"Starting workers","controller":"infrastructure_controller","worker count":5}
{"level":"info","ts":"2020-07-08T07:08:57.531Z","logger":"controller-runtime.controller","msg":"Starting workers","controller":"worker_state_controller","worker count":5}
{"level":"info","ts":"2020-07-08T07:08:57.531Z","logger":"controller-runtime.controller","msg":"Starting workers","controller":"healthcheck_controller-Worker-extensions.gardener.cloud-v1alpha1-okEu9lfQqV","worker count":5}
{"level":"info","ts":"2020-07-08T07:22:58.041Z","logger":"backupbucket_controller","msg":"Starting the deletion of backupbucket","backupbucket":"e56460ce-9656-44a8-9ccf-8ff5077c2451"}
{"level":"info","ts":"2020-07-08T07:22:58.609Z","logger":"backupbucket_controller","msg":"Successfully deleted backupbucket","backupbucket":"e56460ce-9656-44a8-9ccf-8ff5077c2451"}
{"level":"info","ts":"2020-07-08T07:22:58.619Z","logger":"backupbucket_controller","msg":"Removing finalizer.","backupbucket":"e56460ce-9656-44a8-9ccf-8ff5077c2451"}
[...]

PR #2610, opened 2020-07-21
Fix ControllerInstallation required calculation

### LINKED ISSUE #2519: Last responsible extension controller remains on seed after getting obsolete

/area quality
/area cost
/kind bug
/priority normal

**What happened**:
Gardener offers on-demand `ControllerRegistration`s which are only deployed to the seed cluster if a related extension Kind/Type is found on the seed. Likewise, once extensions are gone, Gardener checks if the extension controller is still required and deletes it if necessary.

Unfortunately, this hehavior doesn't work for the very last extension Kind/Type that is deleted from the seed, i.e. the infrastructure resource.

The provider extension to be deleted.

- Create a cluster on a seed w/o backup configuration
- Delete the cluster again
- `ControllerInstallation` for the provider extension is not gone

You can find the related exit condition here: https://github.com/gardener/gardener/blob/f120f2c74a503ab4a388aa055cd08d4081eaf5ed/pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go#L174-L177

The comment correctly says that it cannot decide if the extension is required or not but unfortunately it misses the following case:
- The controller registration is responsible for resources which are **not required** at all by the shoot, e.g. because you disabled the backup feature (`controllerRegistration.Spec.Resources`)
```
resources:
  - kind: BackupBucket     <--- not required in this scenario
    primary: true
    type: gcp
  - kind: BackupEntry        <--- not required in this scenario
    primary: true
    type: gcp
  - kind: ControlPlane
    primary: true
    type: gcp
  - kind: Infrastructure
    primary: true
    type: gcp
  - kind: Worker
    primary: true
    type: gcp
```
- `c.kindToRequiredTypes` has no required types **after** the `Infrastructure` extension is deleted, hence we never enter this block https://github.com/gardener/gardener/blob/f120f2c74a503ab4a388aa055cd08d4081eaf5ed/pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go#L166-L169
- Eventually the early exit is taken because `required == nil`

/cc @rfranzke

**Environment**:

- Gardener version: `master` with #2517

timuthy:

/area robustness
/kind bug
/priority normal

This PR fixes the calculation of ControllerInstallations. Earlier ControllerInstallations were still considered as required even though no corresponding extension resources or Shoot clusters existed. Please look at #2519 for more information.

We also added the validation for known extension kinds in ControllerRegistrations along the way.

Thanks to @rfranzke for collaborating on this PR.

Fixes #2519 #2572

**Release note**:

```improvement operator
An issue has been fixed which prevented not required ControllerInstallations from being deleted. As a side effect, it also blocked the deletion of Seed resources.
```
```improvement operator
Gardener now validates the extension kinds configured in `.spec.resources[].kind` in ControllerRegistrations. 
```

### DIFF
--- a/pkg/apis/extensions/v1alpha1/types.go
+++ b/pkg/apis/extensions/v1alpha1/types.go
@@ -16,8 +16,11 @@ package v1alpha1
 
 import (
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+
+	dnsv1alpha1 "github.com/gardener/external-dns-management/pkg/apis/dns/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // Status is the status of an Object.
@@ -68,3 +71,17 @@ type Object interface {
 	// GetExtensionStatus retrieves the object's status.
 	GetExtensionStatus() Status
 }
+
+// ExtensionKinds contains all supported extension kinds.
+var ExtensionKinds = sets.NewString(
+	BackupBucketResource,
+	BackupEntryResource,
+	ContainerRuntimeResource,
+	ControlPlaneResource,
+	dnsv1alpha1.DNSProviderKind,
+	ExtensionResource,
+	InfrastructureResource,
+	NetworkResource,
+	OperatingSystemConfigResource,
+	WorkerResource,
+)
--- a/pkg/apis/core/validation/controllerregistration.go
+++ b/pkg/apis/core/validation/controllerregistration.go
@@ -61,6 +61,11 @@ func ValidateControllerRegistrationSpec(spec *core.ControllerRegistrationSpec, f
 		if len(resource.Kind) == 0 {
 			allErrs = append(allErrs, field.Required(idxPath.Child("kind"), "field is required"))
 		}
+
+		if !extensionsv1alpha1.ExtensionKinds.Has(resource.Kind) {
+			allErrs = append(allErrs, field.NotSupported(idxPath.Child("kind"), resource.Kind, extensionsv1alpha1.ExtensionKinds.UnsortedList()))
+		}
+
 		if len(resource.Type) == 0 {
 			allErrs = append(allErrs, field.Required(idxPath.Child("type"), "field is required"))
 		}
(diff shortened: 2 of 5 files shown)


rfranzke (dismissed):
/lgtm
/needs second-opinion

pkg/apis/extensions/v1alpha1/types.go
```diff
@@ -68,3 +69,16 @@ type Object interface {
 	// GetExtensionStatus retrieves the object's status.
 	GetExtensionStatus() Status
 }
+
+// ExtensionKinds contains all supported extension kinds.
+var ExtensionKinds = sets.NewString(
```
vpnachev, line 76: Is the `DNSProvider` still working?

rfranzke: Oh, indeed, that's a nice catch, thanks @vpnachev.
@timuthy we should add `DNSProvider`, `DNSEntry`, and `DNSOwner` to this list as well.
/needs changes
/status author-action

timuthy: Thanks a lot for noticing this.

pkg/apis/extensions/v1alpha1/types.go
```diff
@@ -68,3 +71,17 @@ type Object interface {
...
+	BackupBucketResource,
+	BackupEntryResource,
+	ContainerRuntimeResource,
+	ControlPlaneResource,
+	dnsv1alpha1.DNSProviderKind,
```
rfranzke, line 81: What about `DNSEntry` and `DNSOwner`?

rfranzke (changes requested):

pkg/apis/extensions/v1alpha1/types.go
timuthy: Are these really extension Kinds we use for registration? Because we don't consider them [here](https://github.com/gardener/gardener/blob/8359e1ae278c1f121d1d3db69a96c1fa6e0ce4a5/pkg/operation/shoot/shoot.go#L465-L481) either.

rfranzke: Yes, thanks, you're right!

rfranzke (approved):
/lgtm

vpnachev (approved):
/lgtm

PR #2611, opened 2020-07-21
Custom 'uam' verb for project member user management

### LINKED ISSUE #2592: New project role for project members administration
/kind enhancement
/priority normal

New project role for project members administration
**Why is this needed**:
The members of the Gardner projects can be managed via the Gardner Dashboard and also via the Gardner API by any project member with the Gardener role “admin”. The problem is that when the Gardner API is used for an integration with an external UAM solution, project admins can still override the permissions configured by the external system. In order to resolve this problem the access for managing the project members have to be segregated using a separate role. This way the project owner can decide to whom to grant the UAM access via Gardener Dashboard/API and when the project is integrated with an external UAM solution, it will be possible to keep this access only for the project owner for emergency use cases.

Proposals for the name of the new role: “uam”  (from User Access Management)

**Proposal for the implementation:**
1)	Create a new Gardner project member role and segregate the access for managing project members only in this role.
2)	Make sure that this role will be always assigned to the project owner, just like the owner is always assigned with the admin role.
3)	For all existing projects assign the new role to all project members who have the admin role. The project owner will be able to remove it from the members that don’t need it later.
4)	Remove the permissions for the project members management from the admin role

rfranzke:

/area user-management
/kind enhancement
/priority normal

This PR introduces the new `uam` custom verb and `uam` project role (next to `admin` and `viewer`). The new behaviour is that only users bound to the `uam` verb are allowed to add/modify/remove human users from the `.spec.members` list of `Project`s. Service account users can still be managed via `update`/`patch`.

For existing `Project`s, the `uam` role will be automatically granted to all `admin` members by GCM before starting the `Project` controller. The owner can remove this role from the members as desired.
For new `Project`s only the owner will be bound to the `uam` verb by default, i.e., no `admin` member is allowed to manage human members by default.

Fixes #2592

* I fixed a bug in the `customauthorizer` admission plugin for the `modify-spec-tolerations-whitelist` verb along the way.
* /cc @donistz 
* /cc @gardener/dashboard-maintainers 

**Release note**:

```noteworthy user
A new `uam` role was introduced for `Project`s (next to `owner`, `admin`, and `viewer`). Members with this role will be bound to the `uam` custom RBAC verb for the respective `Project`. Only users bound to this verb are now allowed to add/modify/remove human users or groups from the `.spec.members[]` list of the `Project`. Please find more information [here](https://github.com/gardener/gardener/blob/master/docs/usage/projects.md).
```

### DIFF
--- a/docs/usage/projects.md
+++ b/docs/usage/projects.md
@@ -1,0 +1,88 @@
+# Projects
+
+The Gardener API server supports a cluster-scoped `Project` resource which is used to group usage of Gardener.
+For example, each development team has its own project to manage its own shoot clusters.
+
+Each `Project` is backed by a Kubernetes `Namespace` that contains the actual related Kubernetes resources like `Secret`s or `Shoot`s.
+
+**Example resource:**
+
+```yaml
+apiVersion: core.gardener.cloud/v1beta1
+kind: Project
+metadata:
+  name: dev
+spec:
+  namespace: garden-dev
+  description: "This is my first project"
+  purpose: "Experimenting with Gardener"
+  owner:
+    apiGroup: rbac.authorization.k8s.io
+    kind: User
+    name: john.doe@example.com
+  members:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: User
+    name: alice.doe@example.com
+    role: admin
+  # roles:
+  # - viewer 
+  # - uam
+  # - extension:foo
+  - apiGroup: rbac.authorization.k8s.io
+    kind: User
+    name: bob.doe@example.com
+    role: viewer
+# tolerations:
+#   defaults:
+#   - key: <some-key>
+#   whitelist:
+#   - key: <some-key>
+```
+
+The `.spec.namespace` field is optional and will be initialized if it's unset.
+The name of the resulting namespace will be generated and look like `garden-dev-5anj3`, i.e., it has a random suffix.
+It's also possible to adopt existing namespaces by labeling them `gardener.cloud/role=project` and `project.gardener.cloud/name=dev` beforehand (otherwise, they cannot be adopted). 
+
+The `spec.description` and `.spec.purpose` fields can be used to describe to fellow team members and Gardener operators what this project is used for.
+
+Each project has one dedicated owner, configured in `.spec.owner` using the `rbac.authorization.k8s.io/v1.Subject` type.
+The owner is the main contact person for Gardener operators.
+Please note that the `.spec.owner` field is deprecated and will be removed in future API versions in favor of the `owner` role, see below.
+
+The list of members (again a list in `.spec.members[]` using the `rbac.authorization.k8s.io/v1.Subject` type) contains all the people that are associated with the project in any way.
+Each project member must have at least one role (currently described in `.spec.members[].role`, additional roles can be added to `.spec.members[].roles[]`). The following roles exist:
+
+* `admin`: This allows to fully manage resources inside the project (e.g., secrets, shoots, configmaps, and similar).
+* `uam`: This allows to add/modify/remove human users or groups to/from the project member list. Technical users (service accounts) can be managed by all admins.
+* `viewer`: This allows to read all resources inside the project except secrets.
+* `owner`: This combines the `admin` and `uam` roles.
+* Extension roles (prefixed with `extension:`): Please refer to [this document](../extensions/project-roles.md).
+
+The [project controller](../concepts/controller-manager.md#project-controller) inside the Gardener Controller Manager is managing RBAC resources that grant the described privileges to the respective members.
+
+There are two central `ClusterRole`s `gardener.cloud:system:project-member` and `gardener.cloud:system:project-viewer` that grant the permissions for namespaced resources (e.g., `Secret`s, `Shoot`s, etc.).
+Via referring `RoleBinding`s created in the respective namespace the project members get bound to these `ClusterRole`s and, thus, the needed permissions.
+There are also project-specific `ClusterRole`s granting the permissions for cluster-scoped resources, e.g. the `Namespace` or `Project` itself.  
+For each role, the following `ClusterRole`s, `ClusterRoleBinding`s, and `RoleBinding`s are created:
+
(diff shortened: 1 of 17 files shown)


donistz:
Absolutely excellent! :)

docs/usage/projects.md
```diff
@@ -0,0 +1,90 @@
...
+```yaml
+apiVersion: core.gardener.cloud/v1beta1
+kind: Project
+metadata:
+  name: dev
+spec:
+  # The `spec.namespace` field is optional and will be initialized if unset - the resulting
+  # namespace will be generated and look like "garden-dev-<random-chars>", e.g. "garden-dev-5z43z".
+  # If the namespace is set then the namespace must be labelled with `gardener.cloud/role: project`
+  # and `project.gardener.cloud/name: <project-name>` (<project-name>=dev in this case).
```
vpnachev: As it is described very well below, you can remove these lines.

docs/usage/projects.md
```diff
@@ -0,0 +1,90 @@
...
+#   - key: <some-key>
+```
+
+The `.spec.namespace` field is optional and will be initialized if it's unset.
+The name of the resulting namespace will be generated and look like `garden-dev-5anj3`, i.e., it has a random suffix.
+It's also possible to adopt existing namespaces by labeling them `gardener.cloud/role=project` and `project.gardener.cloud/name=dev` beforehand (otherwise, they cannot be adopted). 
+
+The `spec.description` and `.spec.purpose` fields can be used to describe to fellow team members and Gardener operators what this project is used for.
+
+Each project has one dedicated owner, configured via a RBAC subject in `.spec.owner`.
```
vpnachev: >  configured via a RBAC subject in `.spec.owner`.

I think this is not necessary.

vpnachev: Or at least it is not very clear to me why it is mentioned.

pkg/controllermanager/controller/project/project.go
```diff
@@ -145,6 +159,52 @@ func (c *Controller) Run(ctx context.Context, workers int) {
...
+					panic(err)
+				}
+
+				var (
+					projectCopy = project.DeepCopy()
+					memberNames []string
+				)
+
+				for i, member := range projectCopy.Spec.Members {
+					if (member.Role == gardencorev1beta1.ProjectMemberAdmin || utils.ValueExists(gardencorev1beta1.ProjectMemberAdmin, member.Roles)) && !utils.ValueExists(gardencorev1beta1.ProjectMemberUserAccessManager, member.Roles) {
```
vpnachev, line 189: This means that even service accounts will get the uam role.
If I am a human and somebody with `uam` role disallow me to manage human users, I can use one of the existing service accounts and escalate my privileges.

It will be similar to the current behavior, but maybe we would like to change this with the `uam` role?

vpnachev (commented):
I have skimmed through it and have some comments.

pkg/controllermanager/controller/project/project.go
donistz: This is for backwards compatibility for existing resources (users, Service Accounts). The owner/admin will decide from whom to remove the uam role once we release the new change. For the new projects the uam role will not go together with the admin role. It will be assigned explicitly.

docs/usage/projects.md
rfranzke: Hm, why? I'm explaining the various fields in the `spec`.

vpnachev: The first time I was reading the sentence I thought that the owner is configured by `(Cluster)RoleBinding` and should be done outside of the `Project` resource.
A bit later I find out that you want to say that `RBAC subject` is actually the type of the field, e.g. the golang type.

Now, I know It is correct, but for detailed explanation of the field the [API reference](https://gardener.cloud/api-reference/core/#core.gardener.cloud/v1beta1.Project) should be used.

rfranzke: OK, I have rephrased it. PTAL :)

mvladev:
I would suggest not to use `uam` but `manage` or something similar. `uam` is simply not a ... verb.

rfranzke:
@mvladev What about `manage-members`? cc @donistz

mvladev:
In kubernetes the custom `escalate` [verb is used](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#privilege-escalation-prevention-and-bootstrapping). It's meaning is `escalate privileges`. In this context `manage` would already imply `manage-members`

rfranzke:
Hm, not sure, we already have the `modify-spec-tolerations-whitelist` custom verb, so `manage` sounds too general?

donistz:
“uam” is User Access Management – This is mange but for concrete resources

donistz:
> In kubernetes the custom `escalate` [verb is used](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#privilege-escalation-prevention-and-bootstrapping). It's meaning is `escalate privileges`. In this context `manage` would already imply `manage-members`

"escalate" is a verb that can be easily linked only to permissions management while "manage" can be for anything

rfranzke:
Agreed to @donistz. If you really want a linguistic verb, @mvladev, then please propose one that is more accurate than manage (still, what about `manage-members`). Personally, I have no issues with keeping it `uam` as it's still precise (though, no linguistic verb 😉).

vpnachev (dismissed):
/lgtm

Regarding verb name, I can agree with Martin that `uam` does not sounds great and I have some proposals:
- delegate
- charge
- empower
- accredit

rfranzke:
Thanks for the proposals, but can we agree on something that indicates that it's about users/members?

donistz:
Alternative verbs: 
*authorize
*permit
*admit

rfranzke:
@mvladev @vpnachev Do you agree with `manage-members` or do you want to suggest something different?

mvladev:
`manage-members` is okay.

rfranzke:
Done, I renamed the verb to `manage-members`. PTAL :)

vpnachev (approved):
/lgtm

rfranzke:
If there are no objections, I'd plan to merge the PR EOB today.

danielfoehrKn (commented):
/lgtm

PR #2612, opened 2020-07-21
Ignore ctrlreg in deletion when computing required ctrlinsts

vpnachev:

/area quality
/kind bug
/priority normal

Ignore ControllerRegistration in deletion when computing required ControllerInstalations.

To properly apply https://github.com/gardener/gardener-extension-os-suse-chost/pull/20 on existing landscapes, it is necessary to manually delete the existing controller installations. However, on large landscapes GCM creates new controller installations before the last old one to be deleted. With this change, once the deletion of the controller registration is triggered, no new installations will be made out of it.

Fixes #

**Release note**:

```improvement operator
`ControllerInstallation`s are no longer created for `ControllerRegistration`s that are in deletion.
```

### DIFF
--- a/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
+++ b/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
@@ -387,6 +387,15 @@ func deployNeededInstallations(
 	registrationNameToInstallationName map[string]string,
 ) error {
 	for _, registrationName := range wantedControllerRegistrations.UnsortedList() {
+		// Sometimes an operator needs to migrate to a new controller registration that supports the required
+		// kind and types, but it is required to offboard the old extension. Thus, the operator marks the old
+		// controller registration for deletion and manually delete its controller installation.
+		// In parallel, Gardener should not create new controller installations for the deleted controller registation.
+		if controllerRegistrations[registrationName].obj.DeletionTimestamp != nil {
+			logger.Infof("Do not create or update ControllerInstallation for %q which is in deletion", registrationName)
+			continue
+		}
+
 		logger.Infof("Deploying wanted ControllerInstallation for %q", registrationName)
 
 		if err := deployNeededInstallation(ctx, c, seed, controllerRegistrations[registrationName].obj, registrationNameToInstallationName[registrationName]); err != nil {
@@ -404,7 +413,6 @@ func deployNeededInstallation(
 	controllerRegistration *gardencorev1beta1.ControllerRegistration,
 	controllerInstallationName string,
 ) error {
-
 	installationSpec := gardencorev1beta1.ControllerInstallationSpec{
 		SeedRef: corev1.ObjectReference{
 			Name:            seed.Name,
--- a/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control_test.go
+++ b/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control_test.go
@@ -137,6 +137,7 @@ var _ = Describe("ControllerRegistrationSeedControl", func() {
 
 		alwaysPolicy   = gardencorev1beta1.ControllerDeploymentPolicyAlways
 		onDemandPolicy = gardencorev1beta1.ControllerDeploymentPolicyOnDemand
+		now            = metav1.Now()
 
 		type1  = "type1"
 		type2  = "type2"
@@ -790,11 +791,12 @@ var _ = Describe("ControllerRegistrationSeedControl", func() {
 
 			It("should correctly deploy needed controller installations", func() {
 				var (
-					wantedControllerRegistrations      = sets.NewString(controllerRegistration2.Name, controllerRegistration3.Name)
+					wantedControllerRegistrations      = sets.NewString(controllerRegistration2.Name, controllerRegistration3.Name, controllerRegistration4.Name)
 					registrationNameToInstallationName = map[string]string{
 						controllerRegistration1.Name: controllerInstallation1.Name,
 						controllerRegistration2.Name: controllerInstallation2.Name,
 						controllerRegistration3.Name: controllerInstallation3.Name,
+						controllerRegistration4.Name: "",
 					}
 				)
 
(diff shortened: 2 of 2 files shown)


rfranzke (changes requested):
Thanks, the change makes sense 👍 Can you add a new `ControllerRegistration` with a `deletionTimestamp` and support for resource kind/type that are used by one of the shoots [here](https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control_test.go#L342-L466)? That'd probably be enough to test this case (expectation would be that the computed list of required controller registrations doesn't change).

vpnachev:
I have amend it to make it easier for testing.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control_test.go
```diff
@@ -644,6 +659,35 @@ var _ = Describe("ControllerRegistrationSeedControl", func() {
 			)))
 		})
 
+		It("should correctly compute the result for a controller registration with deletion timestamp", func() {
+			reigstration := append(controllerRegistrationList, controllerRegistration8)
```
timebertt:
```suggestion
			registration := append(controllerRegistrationList, controllerRegistration8)
```

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -291,7 +293,8 @@ func computeControllerRegistrationMaps(
 	controllerRegistrationList []*gardencorev1beta1.ControllerRegistration,
 ) map[string]controllerRegistration {
 	var out = make(map[string]controllerRegistration)
-	for _, cr := range controllerRegistrationList {
+
+	for _, cr := range getActiveControllerRegistrations(controllerRegistrationList) {
```
timebertt: Hmm, if you filter out the "to-be-deleted" registrations here, the controller will always exit with error in line 165 if there is a registration with deletion timestamp but shoots/installations needing that registration.

For example, I tested this by
1. adding a networking registration of type `foo`
2. adding a shoot using this network type
3. deleting the registration

gcm will log:
```
time="2020-07-23T08:17:10+02:00" level=info msg="[CONTROLLERINSTALLATION SEED] Reconciling gcp" controllerregistration-seed=staging-gcp
time="2020-07-23T08:17:10+02:00" level=info msg="Error syncing ControllerRegistration-Seed gcp: need to install an extension controller for \"Network/foo\" but no appropriate ControllerRegistration found"
```

Also, if the shoot is deleted again and the registration is not needed anymore but already marked for deletion, the controller will not remove the installations anymore, because it will fail in line 371 (`ControllerRegistration "foo" does not exist`).

Can you fix those cases?

timebertt (dismissed):
Thanks for the PR, makes sense!

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control_test.go
```diff
@@ -705,6 +749,13 @@ var _ = Describe("ControllerRegistrationSeedControl", func() {
 
 			Expect(registrations).To(Equal(controllerRegistrations))
 		})
+
+		It("should ignore controller reigstration with deletion timestamp", func() {
```
rfranzke:
```suggestion
		It("should ignore controller registration with deletion timestamp", func() {
```

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
vpnachev: Nice finding, thanks!

I was testing only the case with `os-suse-*` where there was multiple extensions handling the same resources.
I have changed it now to silently not create new resources, but it is still updating existing controller installations.

vpnachev:
An admission plugin has been implemented, in this way it is guaranteed that creation of a controller installation for controller registration in deletion will not succeed.

The change in the controller is re-worked again and it now silently skips the creation of controller installation, however it still can update existing installations. I tried to implement some unit test about this change, something similar to https://github.com/gardener/gardener/blob/f2f47cc860e20523d279945d721e8ce506ba086d/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control_test.go#L772-L823 but I failed to understand how these tests works and whether it is possible such a test to be implemented.

rfranzke:
> An admission plugin has been implemented, in this way it is guaranteed that creation of a controller installation for controller registration in deletion will not succeed.

Hm, I can't follow - why is such an admission plugin needed? I mean, what's the reason this is so important that we have to forbid it API-wise? Can't we save this code for good? The only code that creates `ControllerRegistration` in Gardener is the controller you already changed, and if a user creates a `ControllerInstallation` manually then why should we forbid this?

> but I failed to understand how these tests works and whether it is possible such a test to be implemented.

What exactly didn't you understand? If I see it correctly, the only thing you have to do is to add a `ControllerRegistration` with a `deletionTimestamp` to [this list](https://github.com/gardener/gardener/blob/f2f47cc860e20523d279945d721e8ce506ba086d/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control_test.go#L793). As it's not expected that it's going to be created you don't need `EXPECT()` calls for the `k8sClient`. 
As a second test case, you could also add a `ControllerRegistration` with a `deletionTimestamp` which already has an existing `ControllerInstallation`. In this case you would expect that an `Update()` call is sent.

vpnachev:
> ... and if a user creates a `ControllerInstallation` manually then why should we forbid this?

I have no strong opinion, it was implemented just as a second prevention mechanism. Basically, applying the same approach by k8s community with namespaces - if it is in deletion, no new resources can be created in the namespace.

> What exactly didn't you understand?

Exactly these `.EXPECT().` calls are not clear to me what are they doing and how are they supposed to be used.
Looking through the generated code also does not help. I tried to find some documentation or guides or to look through https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/client but failed here as well. Could you point me to some guides/documentation about this framework?

What I really tried to do was to call `deployNeededInstallations` with a `ControllerRegistration` with deletionTimestamp set and then list the `ControllerInstallations` with the provided fake client expecting empty list as result - this failed on multiple places, so I have decided that I am doing it completely wrong.

rfranzke:
> I have no strong opinion, it was implemented just as a second prevention mechanism. Basically, applying the same approach by k8s community with namespaces - if it is in deletion, no new resources can be created in the namespace.

Not sure if this comparison is fair. Personally, I don't see the necessity to have this admission plugin (because, even if the `ControllerRegistration` is being deleted you might need to install it to some seed for whatever reason; i.e., I don't think we need to forbid it API-wise). Other opinions maybe? @gardener/gardener-maintainers 

> Could you point me to some guides/documentation about this framework?

Hm, I'm confused again, we are using [gomock](https://github.com/golang/mock) since years not only for Gardener but also heavily in the extensions. Do you mean something different?

vpnachev:
Ok, I was thinking that this is coming from k8s.io, while it is actually from gomock - I will spend some time with this package.

Together with @ialidzhikov we have taken a look through the change and he helped me a lot to understand more about the tests, especially for mocks and fake clients, and to implement one (thanks for that, @ialidzhikov !). 

In parallel, there is a [validation](https://github.com/gardener/gardener/blob/master/pkg/apis/core/validation/controllerregistration.go#L122) that disallow `ControllerRigistration` to be updated once it is in deletion, thus we have decided to also not update existing installations because it should not possible due to this validation.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -387,6 +387,15 @@ func deployNeededInstallations(
 	registrationNameToInstallationName map[string]string,
 ) error {
 	for _, registrationName := range wantedControllerRegistrations.UnsortedList() {
+		// Sometimes an operator needs to migrate to a new controller registration that supports the required
+		// kind and types, but it is required to offboard the old extension. Thus, the operator marks the old
+		// controller registration for deletion and manually delete its controller installation.
+		// In parallel, Gardener should not create new controller installations for the deleted controller registation.
+		if controllerRegistrations[registrationName].obj.DeletionTimestamp != nil {
+			logger.Infof("Do not create or update ControllerInstallation for %q which is in deletion", registrationName)
+			return nil
```
rfranzke:
```suggestion
			continue
```

?

ialidzhikov: Nice catch

plugin/pkg/controllerinstallation/admission.go
```diff
@@ -0,0 +1,150 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
ialidzhikov: I also vote to do not introduce this admission plugin. The change in the controllerregistration control to do not create new controllerinstallations when the ctrlreg has a `deletionTimestamp` should be enough.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
vpnachev: Indeed, thanks!

vpnachev:
The admission plugin has been removed.

rfranzke (approved):
/lgtm

plugin/pkg/controllerinstallation/admission.go
vpnachev: Removed

ialidzhikov (approved):
/lgtm

PR #2614, opened 2020-07-22
Fix flaky infrastructure test

timebertt:

/area testing
/kind bug
/priority normal
/size m

We have seen some flaky unit tests in the past few weeks, which were all time sensitive unit tests and seem to get quite unstable under CPU pressure/throttling (e.g. in our pipelines: https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-master/jobs/master-pull-request-job/builds/50).
One of these occurrences has already been fixed in #2609, this PR fixes the infrastructure component test.
It adds a fake implementation for `retry.{Until,UntilTimeout}`, which doesn't wait between retries to make tests quicker and more stable (because they are not time-sensitive anymore).

See https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-master/jobs/master-pull-request-job/builds/50

We have some other unit tests and components, that either don't correctly test the retries and/or still rely on proper timing.
We should fix them and replace the real `retry` funcs with the fake implementation added in this PR.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/extensions/infrastructure/infrastructure.go
+++ b/pkg/operation/botanist/extensions/infrastructure/infrastructure.go
@@ -77,17 +77,14 @@ func New(
 	logger *logrus.Entry,
 	client client.Client,
 	values *Values,
-	waitInterval time.Duration,
-	waitSevereThreshold time.Duration,
-	waitTimeout time.Duration,
 ) shoot.Infrastructure {
 	return &infrastructure{
 		client:              client,
 		logger:              logger,
 		values:              values,
-		waitInterval:        waitInterval,
-		waitSevereThreshold: waitSevereThreshold,
-		waitTimeout:         waitTimeout,
+		waitInterval:        DefaultInterval,
+		waitSevereThreshold: DefaultSevereThreshold,
+		waitTimeout:         DefaultTimeout,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


rfranzke (dismissed):
Very nice, thanks for picking this up!

/lgtm

pkg/operation/common/extensions.go
```diff
@@ -335,10 +344,15 @@ func WaitUntilExtensionCRDeleted(
 	name string,
 	interval time.Duration,
 	timeout time.Duration,
+	waiter retry.Ops,
```
timuthy: Now we pass `interval` + `timeout` + `waiter` separately even though they belong together. Additionally, I think the parameter count and passing `nil` so many times gets a little awkward. I thought about two possible ways:
1. Passing a `WaitFn` instead of `interval` + `timeout` + `waiter`. Helper functions could help to create a proper `WaitFn`, i.e. one for tests and one for prod execution.
2. Controlling the `waiter` via a global variable and exchanging it before executing the tests. 

If the "default waiter" is anyway problematic for the test execution, it shouldn't harm to exchange it globally.

WDYT?

timebertt: Well, that's a valid point. I like the second approach as it is more lightweight and doesn't introduce more complexity and helper/wrapper functions.

Then let's simply exchange `retry.{Until,UntilTimeout}` with the fake implementations in the tests where we need it (these are anyways aliases already). Would that sound good to you?

timuthy: Sure :)

rfranzke (approved):

timuthy (approved):
/lgtm
Thanks for implementing the suggestion!

PR #2615, opened 2020-07-22
Remove log-exclusion annotation for non-{testing,hibernated} shoots

### LINKED ISSUE #2593: "fluentbit.io/exclude=true" is not removed after .spec.purpose change

/area logging
/kind bug
/priority normal

**What happened**:
With #2320  gardener-seed-admission-controller adds the `fluentbit.io/exclude=true` annotation to the controlplane Pods when the spec.purpose=testing. When the purpose is changed, the exclude annotation stays in the controlplane Pods and this prevents fluent-bit to collect logs from them.

1. Create Shoot with `.spec.purpose=testing` and ensure that controlplane Pods have the annotation `fluentbit.io/exclude: "true"`

2. Change the purpose to `evaluation`

3. Ensure that logging components are deployed as part of the controlplane.

4. Ensure that the `fluentbit.io/exclude: "true"` is not removed from controlplane Pods.

**Environment**:

- Gardener version: v1.7.2
- Others:

rfranzke:

/area logging
/kind bug
/priority normal

When a shoot's purpose is changed from `testing` to something else or when a shoot is woken up then the Gardener Seed Admission Controller does now remove the fluentbit log exclusion annotation from pods again.
Earlier, these annotations were simply kept and only a pod restart was fixing it (i.e., the annotation was not added again). This may have led to prolonged times in which no logs were collected.

Fixes #2593

/invite @vlvasilev 

**Release note**:

```improvement operator
A bug was fixed that prevented the log collection for control plane components that belong to shoots whose purpose was changed from `testing` to something else or for those that were woken up.
```

### DIFF
--- a/pkg/seedadmission/pods.go
+++ b/pkg/seedadmission/pods.go
@@ -19,15 +19,15 @@ import (
 	"fmt"
 	"strings"
 
-	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	"github.com/gardener/gardener/pkg/extensions"
 	gardenlogger "github.com/gardener/gardener/pkg/logger"
 
 	"github.com/sirupsen/logrus"
 	"gomodules.xyz/jsonpatch/v2"
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


timuthy (dismissed):
/lgtm

pkg/seedadmission/pods.go
```diff
@@ -66,67 +66,110 @@ func MutatePod(ctx context.Context, c client.Client, logger *logrus.Logger, requ
...
-		Value:     annotationsToAdd,
-	})
-	return patch, nil
+	logger.Debugf("Skipping mutation for %s/%s due to policy check", acc.GetNamespace(), acc.GetName())
+	return nil, nil
 }
 
-func mutationRequired(ctx context.Context, c client.Client, namespace string) (bool, error) {
+func extractShoot(ctx context.Context, c client.Client, namespace string) (*gardencorev1beta1.Shoot, error) {
 	cluster, err := extensionscontroller.GetCluster(ctx, c, namespace)
```
timebertt: I know, this was already here. But didn't you say, we want to avoid references from g/pkg to g/extensions?
Can you fix that here?

pkg/seedadmission/pods.go
```diff
@@ -66,67 +66,110 @@ func MutatePod(ctx context.Context, c client.Client, logger *logrus.Logger, requ
...
+
+	if shoot.Spec.Purpose != nil && *shoot.Spec.Purpose == gardencorev1beta1.ShootPurposeTesting {
+		return true
+	}
+
+	return false
+}
+
+func computeAddOrReplacePatch(existingAnnotations, desiredAnnotations map[string]string) []jsonpatch.JsonPatchOperation {
+	if existingAnnotations == nil {
```
timebertt:
```suggestion
	if len(existingAnnotations) == 0 {
```

timebertt (commented):
Only minor comments.
Rest looks good :)

pkg/seedadmission/pods.go
rfranzke: What do you suggest? If we move the `cluster.go` file from `extensionscontroller` package to somewhere else we have to adapt all extension controllers. I don't want to copy the logic of getting the shoot out of a cluster.

timebertt: What about moving the logic to somewhere in `g/pkg` and make `extensionscontroller.GetCluster` an alias to that function?
Then we don't need to copy the logic, don't need to adapt the extensions but can still use it in `g/pkg` without importing `g/extensions`.

This would work, but I would also be fine with it, if we agree on having these imports (and probably also adding more in the future). I guess, we anyways don't want to split the whole thing up again 😄 
Just wanted to double-check and have no strong opinion on this topic to be honest.

rfranzke: Hm yes, we don't plan to split it again. The problem is that g/gext was developed side-by-side to g/g and is now just "copied" in g/g again. Over time, I would like to see the `g/extensions` folder disappear/to be merged into `g/pkg`.

Many things in `g/extensions` can be merged with `g/pkg` in some way. A good example is the `g/extensions/gardener/chartrenderer`. It's also strange that we have `g/extensions/log` and `g/pkg/logger`. Eventually, this should be harmonized.

To get started, I've moved `g/extensions/controller/cluster.go` to `g/pkg/extensions/cluster.go` and made type/func aliases (just to not make this PR too large; the `g/extensions/controller/cluster.go` should be deleted in a subsequent PR and the usages should be adapted to `g/pkg/extensions`.

PTAL! :)

timebertt: Yes. There are some duplications and other stuff we have to harmonize.
Thanks for adapting and starting with this change! :)

timebertt (approved):
/lgtm now!

timuthy (approved):

PR #2617, opened 2020-07-22
Prevent panic in maintenance controller when shoot has no annotations

rfranzke:

/area quality
/kind bug
/priority normal

If a `Shoot` has no annotations we have to initialize it first.

/invite @tim-ebert @vpnachev 

**Release note**:

```improvement operator
A bug has been fixed that could cause the gardener-controller-manager to panic if it tries to maintain a `Shoot` whose `.metadata.annotations=nil`.
```


vpnachev (commented):
Thanks.
Could you add a small unit test so that `.Maintain()`  should not fail when annotations is nil

timebertt (commented):
Thanks, can you apply a similar fix to this func, too?
https://github.com/gardener/gardener/blob/72f050a2ffba06dc0add36526eb69438efb280a5/pkg/operation/botanist/metadata.go#L38-L42

rfranzke:
@vpnachev There is no unit test for `Maintain` yet, so writing one is more effort and out of scope of this PR. Or did I miss the place where you suggest this case to be inserted?
@tim-ebert done

timebertt (approved):
Thanks!
/lgtm

vpnachev (approved):
/lgtm

>@vpnachev There is no unit test for Maintain yet, so writing one is more effort and out of scope of this PR. Or did I miss the place where you suggest this case to be inserted?

I have not checked in details the existing tests. So we can do it with another PR.

danielfoehrKn (commented):
/lgtm

PR #2619, opened 2020-07-23
Only remove tasks if generation wasn't changed mid-air

### LINKED ISSUE #2618: Shoot tasks are removed too early

/area usability
/kind bug
/priority normal

**What happened**:
Certain tasks in the shoot reconciliation flow are controlled via the `shoot.gardener.cloud/tasks` annotation in `Shoot`s. A good example is `deployInfrastructure`.
When the `.spec.provider.infrastructureConfig` is changed then the GAPI adds this task annotation to the shoot. This task annotation is removed at the end of a successful reconciliation operation by the Gardenlet.

When the `.spec.provider.infrastructureConfig` is now changed while the shoot reconciliation is processing but already started/passed the `deployInfrastructure` task in its flow, then the `.metadata.generation` is increased which will trigger a subsequent reconciliation. As the `deployInfrastructure` task is still present in the annotation nothing else happens. After the first reconciliation the gardenlet removes the `deployInfrastructure` task. Now the subsequent reconciliation starts WITHOUT this task annotation. This will lead to the next reconciliation being stuck.

The subsequent shoot reconciliation triggers the `deployInfrastructure` task as well.

1. Create a shoot.
1. Change the `.spec.provider.infrastructureConfig`.
1. Wait until the gardenlet has started the `deployInfrastructure` task.
1. Change the `.spec.provider.infrastructureConfig` again.
1. See the subsequent reconciliation failing with `time="2020-07-23T09:09:54+02:00" level=error msg="Infrastructure shoot--dev--foo/foo did not get ready yet" error="observed generation outdated (2/3)" operation=reconcile shoot=garden-dev/foo`

Thanks @grolu for spotting this!

**Environment**:

- Gardener version: `v1.7.2`

rfranzke:

/area usability
/kind bug
/priority normal

This PR only removes the shoot tasks if the generation wasn't changed mid-air. See #2618 for rationale.

Fixes #2618 

Thanks @timuthy @grolu for collaborating here! :)

**Release note**:

```improvement user
A bug has been fixed that could lead to stuck shoot reconciliations in case the `.spec.provider.infrastructureConfig` was changed multiple times while another shoot reconciliation operation was still in progress.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -90,6 +90,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		allowBackup                    = o.Seed.Info.Spec.Backup != nil
 		staticNodesCIDR                = o.Shoot.Info.Spec.Networking.Nodes != nil
 		useSNI                         = botanist.APIServerSNIEnabled()
+		generation                     = o.Shoot.Info.Generation
 		allTasks                       = controllerutils.GetTasks(o.Shoot.Info.Annotations)
 		requestControlPlanePodsRestart = controllerutils.HasTask(o.Shoot.Info.Annotations, common.ShootTaskRestartControlPlanePods)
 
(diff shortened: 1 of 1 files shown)


timuthy (approved):
/lgtm

vpnachev (approved):
/lgtm

PR #2625, opened 2020-07-24
added machineclasses CRD for out-of-tree machine controllers

MartinWeindel:

/area control-plane
/kind enhancement
/priority normal

The CRD of the generic machineclass used by all out-of-tree machine controllers is added to the well-known list of machine CRDs. This avoids that every infrastructure provider using OOT MCM has to deploy it explicitly.

Fixes #

**Release note**:

```improvement operator
added machineclasses CRD for out-of-tree machine controllers
```

### DIFF
--- a/extensions/pkg/controller/worker/machine_crds.go
+++ b/extensions/pkg/controller/worker/machine_crds.go
@@ -166,6 +166,31 @@ func init() {
 				},
 			},
 		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "machineclasses.machine.sapcloud.io",
+			},
+			Spec: apiextensionsv1beta1.CustomResourceDefinitionSpec{
+				Group:   machineGroup,
+				Version: machineVersion,
+				Scope:   apiextensionsv1beta1.NamespaceScoped,
+				Names: apiextensionsv1beta1.CustomResourceDefinitionNames{
+					Kind:       "MachineClass",
+					Plural:     "machineclasses",
+					Singular:   "machineclass",
+					ShortNames: []string{"machcls"},
+				},
+				AdditionalPrinterColumns: []apiextensionsv1beta1.CustomResourceColumnDefinition{
+					{
+						Name:        "Provider",
+						Type:        "string",
+						JSONPath:    ".provider",
+						Description: "Name of the provider which acts on the machine class object",
+					},
+					agePrinterColumn,
+				},
+			},
+		},
 	}
 
 	type machineClass struct {


MartinWeindel:
cc @prashanth26

prashanth26:
Hi @MartinWeindel ,

Thanks for the quick change. Also this might not be needed right? - https://github.com/gardener/gardener/blob/f12bb20a834bd14221f8d9a8f36dd70de7d2aa47/extensions/pkg/controller/worker/machine_crds.go#L196-L333. If it's not required, we could remove this as well?

MartinWeindel:
> Hi @MartinWeindel ,
> 
> Thanks for the quick change. Also this might not be needed right? -
> 
> https://github.com/gardener/gardener/blob/f12bb20a834bd14221f8d9a8f36dd70de7d2aa47/extensions/pkg/controller/worker/machine_crds.go#L196-L333
> 
> . If it's not required, we could remove this as well?

Wouldn't it be better to keep the old machine classes as long as there are still providers with in-tree machine controllers?

prashanth26:
> Wouldn't it be better to keep the old machine classes as long as there are still providers with in-tree machine controllers?

Sorry, you are right. I mistook this change to be on the `provider-vSphere` repo, it's actually on the `gardener/extensions`. Yes, this old code should exist here. Probably in future, we can get rid of the old machine classes once migrations are done.

extensions/pkg/controller/worker/machine_crds.go
```diff
@@ -166,6 +166,31 @@ func init() {
...
+					Plural:     "machineclasses",
+					Singular:   "machineclass",
+					ShortNames: []string{"machcls"},
+				},
+				AdditionalPrinterColumns: []apiextensionsv1beta1.CustomResourceColumnDefinition{
+					{
+						Name:        "Provider",
+						Type:        "string",
+						JSONPath:    ".provider",
+						Description: "Provider is the combination of name and location of cloud-specific drivers",
```
prashanth26:
```suggestion
						Description: "Name of the provider which acts on the machine class object",
```
Location no longer applies as the gRPC mechanism is replaced by the controller based driver. Hence just, the provider name should suffice.

rfranzke (dismissed):
/lgtm

hardikdr (dismissed):
Looks good. Thanks.

MartinWeindel:
updated description text for provider field as suggested by @prashanth26

rfranzke (approved):

prashanth26 (approved):
/lgtm

PR #2635, opened 2020-07-29
Use cached client in Shoot care controller, also fix panic

timebertt:

/area quality
/area cost
/kind enhancement
/priority normal
/size m

This PR optimizes the shoot care controller in the following ways:
- make use of cached Seed clients for listing the ControlPlanes' deployments/statefulsets
- fix a race condition that could lead to a panic when deleting a hibernated shoot (when initializing the shoot clients, `IsAPIServerRunning` might return false and not initialize `b.K8sShootClient`, while shortly after that when executing the health checks it might return true (control plane woken up in the meantime) and try to use `b.K8sShootClient` which is `nil`
- use contexts everywhere (with timeout of `config.Controllers.ShootCare.SyncPeriod.Duration`)
- delete some dead code (/cc @wyb1)

Ref #2414 

/cc @rfranzke 

**Release note**:

```improvement operator
The Shoot care controller has been optimized to leverage the cached Seed client instead of talking directly to the API server.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_care_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_care_control.go
@@ -30,7 +30,7 @@ import (
 	"github.com/gardener/gardener/pkg/gardenlet/apis/config"
 	confighelper "github.com/gardener/gardener/pkg/gardenlet/apis/config/helper"
 	"github.com/gardener/gardener/pkg/logger"
-	operation "github.com/gardener/gardener/pkg/operation"
+	"github.com/gardener/gardener/pkg/operation"
 	botanistpkg "github.com/gardener/gardener/pkg/operation/botanist"
 	"github.com/gardener/gardener/pkg/utils/flow"
 	"github.com/gardener/gardener/pkg/utils/imagevector"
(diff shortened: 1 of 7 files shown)


pkg/gardenlet/controller/shoot/shoot_care_control.go
```diff
@@ -222,7 +238,10 @@ func (c *defaultCareControl) Care(shootObj *gardencorev1beta1.Shoot, key string)
 		// Fetch seed conditions of shoot is a seed
 		func(ctx context.Context) error {
 			seedConditions, err = retrieveSeedConditions(ctx, botanist)
-			return err
+			if err != nil {
+				botanist.Logger.Errorf("Error retrieving seed conditions: %+v", err)
+			}
+			return nil
```
rfranzke, line 247: Why is the error no longer returned?

pkg/operation/botanist/constraints_check.go
```diff
@@ -35,31 +35,25 @@ func shootControlPlaneNotRunningConstraint(condition gardencorev1beta1.Condition
...
-	// don't check constraints if API server has already been deleted or has not been created yet
-	if !apiServerRunning {
+	apiServerRunning, err := initializeShootClients()
+	if err == nil && !apiServerRunning {
+		// don't check constraints if API server has already been deleted or has not been created yet
 		return shootControlPlaneNotRunningConstraint(hibernationConstraint)
 	}
 
-	if err := initializeShootClients(); err != nil {
+	if err != nil {
```
rfranzke: Any reason to not first handle the error case?

pkg/gardenlet/controller/shoot/shoot_care_control.go
```diff
@@ -116,26 +116,41 @@ func (c *defaultCareControl) conditionThresholdsToProgressingMapping() map[garde
...
+				return
+			}
+
+			err = b.InitializeShootClients(ctx)
+
+			// double-check if b.K8sShootClient is unset/uninitialized, otherwise we cannot execute
+			// health and constraint checks and garbage collection
+			if b.K8sShootClient == nil {
+				apiServerRunning = false
+			}
```
rfranzke: Why is this check needed? Probably you can clarify it in the comment.

rfranzke (commented):
Looks good generally, just a few questions.

pkg/gardenlet/controller/shoot/shoot_care_control.go
timebertt: The error returned was also not handled before, IISIC.
I thought about improving this, but I was wondering if the care controller should really error in case it can't retrieve the seed conditions. In that case, the health check results (which might have been executed successfully) would not be transported to the Shoot conditions. 
My thought now was, it makes sense to rather "loose" the seed conditions (and just log an error) than to not update the shoot health conditions with the latest results.
What's your opinion on that?

pkg/operation/botanist/constraints_check.go
timebertt: Well, no :)
I transposed and now it's even cleaner, I think.

pkg/gardenlet/controller/shoot/shoot_care_control.go
timebertt: I added some sentences to the comments. I hope this is more clear now. 
Can you check again?

pkg/gardenlet/controller/shoot/shoot_care_control.go
rfranzke: Hm, yes, makes sense. Thank you!

rfranzke (approved):
/lgtm

pkg/gardenlet/controller/shoot/shoot_care_control.go
rfranzke: Yep, thanks!

timuthy:
/assign
I'm taking a second look

timuthy (approved):
Nice PR!
/lgtm

PR #2637, opened 2020-07-29
Disable cached runtime-client for Shoots

timebertt:

/area cost
/kind enhancement
/priority normal

After observing the `CachedRuntimeClients` for some weeks, we see an increased memory footprint of the gardenlet.
This was expected because the gardenlet now caches even more API objects. Nevertheless, we might not want to enable the cache for Shoot client's as there might be hundreds of Shoots on some Seeds in big gardener installations which might increase the memory footprint too much and we don't really benefit so much from it, as the communication to the Shoot API servers is anyways happening in-cluster.

That's why this PR disables the cached controller-runtime clients for all Shoot ClientSet's, so that `Client()` will be equivalent to `DirectcClient()` for those.

Ref #2414 

/cc @rfranzke 

**Release note**:

```improvement operator
The cached controller-runtime clients have been disabled for Shoot clients to decrease the gardenlet's memory footprint in case the `CachedRuntimeClients` feature gate is enabled.
```

### DIFF
--- a/pkg/client/kubernetes/client.go
+++ b/pkg/client/kubernetes/client.go
@@ -251,7 +251,7 @@ func newClientSet(conf *config) (Interface, error) {
 	}
 
 	var runtimeClient client.Client
-	if UseCachedRuntimeClients {
+	if UseCachedRuntimeClients && !conf.disableCachedClient {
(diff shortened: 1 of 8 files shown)


rfranzke (approved):
/lgtm
Very nice PR!

PR #2639, opened 2020-07-29
Prevent seed-boostrap to delete Shoot VPA RBAC

ialidzhikov:

/area auto-scaling
/kind bug
/priority normal

Currently for a Shoot cluster which is registered as Seed, the seed controller wrongly deletes VPA RBAC from the cluster when the Shoot has`spec.kubernetes.verticalPodAutoscaler.enabled=true` and the corresponding Seed has 
 `spec.settings.verticalPodAutoscaler.enabled=false`.
The Shoot cluster SystemComponentsHealthy is constantly flapping (caused by the 2 control loops acting on the same resource - seed controller deleting it and gardener-resource-manager creating it):
```yaml
    - type: SystemComponentsHealthy
      status: 'False'
      lastTransitionTime: '2020-07-29T11:55:26Z'
      lastUpdateTime: '2020-07-29T11:57:26Z'
      reason: ClusterRoleMissing
      message: 'Required ClusterRole "system:evictioner" in namespace "" is missing.'
```

This PR adds the cluster type (shoot or seed) as part of the VPA RBAC name and in this way prevents the issue.

**Release note**:

```improvement operator
An issue causing the seed controller to delete VPA RBAC for ShootedSeed is now fixed.
```

### DIFF
--- a/charts/seed-bootstrap/charts/vpa/charts/application/templates/clusterrole-actor.yaml
+++ b/charts/seed-bootstrap/charts/vpa/charts/application/templates/clusterrole-actor.yaml
@@ -3,7 +3,7 @@
 apiVersion: {{ include "rbacversion" . }}
 kind: ClusterRole
 metadata:
-  name: system:vpa-actor
+  name: gardener.cloud:vpa:{{ .Values.clusterType }}:actor
   labels:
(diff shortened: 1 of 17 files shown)


rfranzke (approved):
/lgtm

PR #2640, opened 2020-07-30
Allow latest machine image version to have an expiration date

ialidzhikov:

/area usability
/kind task
/priority normal

Currently the gardener-apiserver does not allow setting expiration date for the latest machine image version. This of course makes sense semantically. But there are also cases where this restriction is not handy - for example when deprecating and removing support for machine image (machine image reached EoL and needs to be removed; machine image "renamed" and available under new name). This PR allows setting expiration date to the latest machine image version to improve the process of deprecation and removal of machine image - when the latest version expires, no new Shoots creation will be allows with that version.

With this PR when the latest machine image expires, the maintenance controller creates event like:
```
  Warning  MaintenanceError   2s                    gardener-controller-manager  Could not maintain machine image version: no latest qualifying Shoot machine image could be determined for machine image "foo". Either the machine image is reaching end of life and migration to another machine image is required or there is a misconfiguration in the CloudProfile. If it is the latter, make sure the machine image in the CloudProfile has at least one version that is not expired, not in preview and greater or equal to the current Shoot image version "1.0.0"
```

**Release note**:

```improvement operator
It is now allowed to set an expiration date for the latest machine image version in the `CloudProfile`s.
```

### DIFF
--- a/docs/usage/shoot_versions.md
+++ b/docs/usage/shoot_versions.md
@@ -97,6 +97,10 @@ The Gardener API server enforces the following requirements for versions:
 **Adding a version** 
 - A version must not have an expiration date in the past.
 - There can be only one `supported` version per minor version.
+- The latest Kubernetes version cannot have an expiration date.
+- The latest version for a machine image can have an expiration date. [*]
+
+<sub>[*] Useful for cases in which support for given machine image needs to be deprecated and removed (for example the machine image reaches end of life).</sub>
 
 ## Forceful migration of expired versions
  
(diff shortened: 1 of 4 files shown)


ialidzhikov:
@vpnachev , let me know whether this PR would help with the deprecation and removal process for machine images

/invite @vpnachev @danielfoehrKn

rfranzke (approved):
/lgtm
/needs second-opinion

vpnachev (approved):
/lgtm

timebertt (approved):
/lgtm

PR #2643, opened 2020-07-31
Secret generation waits for kube-apiserver LB if shoot DNS disabled

rfranzke:

/area networking
/kind bug
/priority normal

With #2359 the `DeploySecrets` function is no longer generating the secret, but the `GenerateAndSaveSecrets` function does. Hence, this task needs to depend on the "wait for kube-apiserver service to be ready task" in case the shoot DNS is disabled for the particular seed.

Thanks @mkoynov for reporting the bug!

**Release note**:

```improvement user
A bug has been fixed which prevented the server certificate of the shoot kube-apiservers that run on a seed with disabled shoot DNS from being generated correctly.
```


vpnachev (approved):
/lgtm

PR #2648, opened 2020-08-02
Increase fluent-bit resources

Kristian-ZH:

  /area logging
  /kind enhancement

Increase `fluent-bit` resources because of OOM issue on Dev

Fixes #

@vlvasilev @ialidzhikov @vpnachev 

**Release note**:

```improvement operator
None
```

### DIFF
--- a/charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-daemonset.yaml
+++ b/charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-daemonset.yaml
@@ -35,11 +35,11 @@ spec:
               fieldPath: spec.nodeName
         resources:
           limits:
-            cpu: 100m
-            memory: 100Mi
+            cpu: 300m
+            memory: 400Mi
           requests:
-            cpu: 20m
-            memory: 40Mi
+            cpu: 200m
+            memory: 200Mi
         volumeMounts:
         - name: config
           mountPath: /fluent-bit/etc


rfranzke:
Wouldn't defining a VPA object make more sense?

Kristian-ZH:
> Wouldn't defining a VPA object make more sense?

I was thinking on it too but I am not sure how exactly the VPA is working (fluent-bit has in-memory buffer with logs and if it is getting scaled very often it will be a problem) so maybe we will discuss it with @vlvasilev next week cuz this week we are both in vacation :D

rfranzke:
OK, thanks, then we start with this PR/the static limit increase and continue to work on a dynamic solution as soon as you're back. Enjoy your vacation!

/lgtm

charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-daemonset.yaml
```diff
@@ -35,11 +35,11 @@ spec:
...
           limits:
-            cpu: 100m
-            memory: 100Mi
+            cpu: 300m
+            memory: 400Mi
           requests:
-            cpu: 20m
-            memory: 40Mi
+            cpu: 200m
+            memory: 200Mi
```
vpnachev, line 42: I am fine with the increased limits, but why are you increasing the requests to so high values?

rfranzke: Hm, indeed, it seems not necessary. The old request values were quite OK

rfranzke:
As @Kristian-ZH is on vacation I'll 
/close this PR and open a new one that will also fix https://github.com/gardener/gardener/issues/2650

PR #2649, opened 2020-08-03
Add Event Controller

### LINKED ISSUE #2390: Event logs for maintenance and hibernation operations

/area operations
/area usability
/kind enhancement
/priority normal

Add an "event-log" of performed maintenance operations as well as hibernation / wake-up operations.

Below I listed three options that I could think of how to realize this. 

Option (1)

Add an events (or otherwise named field) into the Shoot.Status containing arbitrary Shoot events identified by the type field.

``` yaml
status:
  conditions: <>
  hibernated: true
  events:
    - type: MachineImageVersionMaintenance
      timestamp: "2020-05-25T14:47:54Z"
      status: "Successful"
      message: "Updated image of worker-pool 'coreos-1' from 'coreos' version 'xy' to version 'abc'."
      reason: "AutoUpdate of MachineImage configured"
    - type: MachineImageVersionMaintenance
      timestamp: "2020-05-25T14:47:53Z"
      status: "Successful"
      message: "Updated image of worker-pool 'coreos-2' from 'coreos' version 'xy' to version 'abc'."
      reason: "Machine image version is expired"
    - type: KubernetesVersionMaintenance
      timestamp: "2020-05-25T14:47:52Z"
      status: "Successful"
      message: "Updated Kubernetes version '0.0.1' to version '0.0.5'. This is an increase in the patch level."
      reason: "AutoUpdate of Kubernetes version configured"
    - type: KubernetesVersionMaintenance
      timestamp: "2020-05-25T14:47:53Z"
      status: "Successful"
    - type: KubernetesVersionMaintenance
      timestamp: "2020-05-24T14:47:53Z"
      status: "Successful"
    - type: Hibernation
      timestamp: "2020-05-24T14:47:53Z"
    - type: WakeUp
      timestamp: "2020-05-26T14:47:53Z"
```

Option (2)
To have a separation between Maintenance and Hibernation, this would be another option.
This would also group events that happened during the same maintenace time window.
``` yaml
status:
  conditions: <>
  hibernated: true
  maintenance:
    - lastSuccessfulMaintenance: "2020-05-25T14:47:53Z"
      events:
      - timestamp: "2020-05-25T14:47:53Z"
        status: "Successful"
        messages:
          - message: Updated image of worker-pool 'coreos-1' from 'coreos' version 'xy' to version 'abc'.
            reason: AutoUpdate of MachineImage configured
            type: MachineImageVersion
          - message: Updated Kubernetes version '0.0.1' to version '0.0.5'. This is an increase in the patch level.
            reason: AutoUpdate of Kubernetes version configured.
            type: KubernetesVersion
  hibernation:
    - timestamp: "2020-05-25T14:47:53Z"
      type: "Hibernation"
    - timestamp: "2020-05-26T14:47:53Z"
      type: "WakeUp"
```

Another option (3) would be to have a dedicated resource in the core.gardener-cloud API group 

``` yaml
kind: ShootOperation
metadata:
  name: <shoot-name>
  namespace: shoot--dev--<>
  ownerReferences:
    - kind: Shoot
      name: <shoot-name>
spec:
  type: Maintenance / Hibernation
status:
  lastOperation: "2020-05-21T00:00:12Z"
  events: # only contain actual changes
    - type: MachineImageVersionMaintenance
      timestamp: "2020-05-25T14:47:54Z"
      status: "Successful"
      message: "Updated image of worker-pool 'coreos-1' from 'coreos' version 'xy' to version 'abc'."
      reason: "AutoUpdate of MachineImage configured"
    - type: MachineImageVersionMaintenance
      timestamp: "2020-05-25T14:47:53Z"
      status: "Successful"
      message: "Updated image of worker-pool 'coreos-2' from 'coreos' version 'xy' to version 'abc'."
      reason: "Machine image version is expired"
    - type: KubernetesVersionMaintenance
      timestamp: "2020-05-25T14:47:52Z"
      status: "Successful"
      message: "Updated Kubernetes version '0.0.1' to version '0.0.5'. This is an increase in the patch level."
      reason: "AutoUpdate of Kubernetes version configured"
```

I tend to favour option (3) to not pollute the already quite large Shoot resource with Hibernation and Maintenance event logs. @vpnachev also suggested to restrict the TTL of an event to 14 days.
The same TTL could also be done if we store the event log in the external resource.

Another issue that I see is how to deal with failed Shoot operations e.g a Shoot using an expired Kubernetes version cannot be updated because of a `CloudProfile` misconfiguration.
It would be handy if the controller could persist this information without writing a new event for each re-tried operation. There are several ways to achieve this.

**Why is this needed**:

There is no place to properly store a log of automatically performed maintenance operations.
For instance for Maintenance, the Gardener Controller manager just writes events on the Shoot resource stating the reason for a performed update operation.
The problem is that Kubernetes events are Garbage collected after a certain amout of time ([configurable in the Kube-Apiserver](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/). 
Usually this is somewhere between 1 - 24 hours.

This is problematic because
- Users want to be able to see which maintenance operation Gardener automatically performed, when, why and if it was successful. 
- In case of unwanted update behaviour / problems, operators cannot easily determine what was the reason for the update operation.

In addition Gardener does currently only persist the current and desired state of the Hibernation.
Shoot owners might also want to see a log of the manually & automatically triggered Hibernations.

Integrators like the gardener Dashboard could read this information and display an operations log for both the Hibernation and the Maintenance operations.

BeckerMax:
/area usability
/kind enhancement
/priority normal

Add event controller

The goal of the event controller is to persist Shoot-related events longer for debugging reasons.

The controller deletes all non-Shoot events after the configured `ttlNonShootEvents`.
Shoot-related events are not touched by the controller and only eventually get deleted after the configured `ttl` in the kube-api-server configuration.

Considered alternatives to this controller were:
1. export relevant events from etcd to another persistent store.
2. extend the global `ttl` in the api-server for all events

While option 1. would be preferable it imposes dificulties as with the Control Plane Migration the shoot CP(with the events) can move between Seed worker nodes.

See also issue #2390 

Also:
* Added hibernation events and more fine-grained maintenance events for shoots.
* Replaced some Shoot related events with logs to decrease the volume of shoot events written.

Fixes #2390 

Very open for any suggestion as well regarding code/tests style. 

**Release note**:
```noteworthy operator
A new controller for `Event`s related to `Shoot` objects has been added to the Gardener Controller Manager (disabled by default). It can be used to extend the live-span of events regarding shoot clusters (the live-span of all other events can be configured separately). Please find more information in [this document](https://github.com/gardener/gardener/blob/master/docs/concepts/controller-manager.md#event-controller).
```


charts/gardener/controlplane/charts/runtime/templates/controller-manager/configmap-componentconfig.yaml
```diff
@@ -73,6 +73,15 @@ data:
         shootMonitorPeriod: {{ .Values.global.controller.config.controllers.seed.shootMonitorPeriod }}
         {{- end }}
       {{- end }}
+      {{- if .Values.global.controller.config.controllers.event }}
+      event:
+        {{- if .Values.global.controller.config.controllers.event.concurrentSyncs }}
+        concurrentSyncs: {{ .Values.global.controller.config.controllers.seed.concurrentSyncs }}
```
rfranzke:
```suggestion
        concurrentSyncs: {{ .Values.global.controller.config.controllers.event.concurrentSyncs }}
```

charts/gardener/controlplane/charts/runtime/templates/controller-manager/configmap-componentconfig.yaml
```diff
@@ -73,6 +73,15 @@ data:
...
+        {{- end }}
+        {{- if .Values.global.controller.config.controllers.event.ttlShortLivedEvents }}
+        ttlShortLivedEvents: {{ .Values.global.controller.config.controllers.seed.ttlShortLivedEvents }}
```
rfranzke:
```suggestion
        ttlShortLivedEvents: {{ .Values.global.controller.config.controllers.event.ttlShortLivedEvents }}
```

pkg/controllermanager/apis/config/v1alpha1/defaults.go
```diff
@@ -147,3 +147,13 @@ func SetDefaults_LeaderElectionConfiguration(obj *LeaderElectionConfiguration) {
 		obj.LockObjectName = ControllerManagerDefaultLockObjectName
 	}
 }
+
+// SetDefaults_EventControllerConfiguration sets defaults for the EventControllerConfiguration.
+func SetDefaults_EventControllerConfiguration(obj *EventControllerConfiguration) {
```
rfranzke, line 152: Can you add a small unit test? https://github.com/gardener/gardener/blob/master/pkg/controllermanager/apis/config/v1alpha1/defaults_test.go

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -71,7 +73,8 @@ type ControllerManagerControllerConfiguration struct {
 	// Seed defines the configuration of the Seed lifecycle controller.
 	// +optional
 	Seed *SeedControllerConfiguration `json:"seed,omitempty"`
-	// ShootMaintenance defines the configuration of the ShootMaintenance controller.
+	// Event defines the configuration of the EventControllerConfiguration controller.
+	// +optional
```
rfranzke: This seems misordered - doc strings don't match to the attributes anymore

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -95,6 +98,18 @@ type ControllerRegistrationControllerConfiguration struct {
 	ConcurrentSyncs int `json:"concurrentSyncs"`
 }
 
+// EventControllerConfiguration defines the configuration of the
+// EventControllerConfiguration controller.
+type EventControllerConfiguration struct {
+	// ConcurrentSyncs is the number of workers used for the controller to work on
+	// events.
+	// +optional
+	ConcurrentSyncs int `json:"concurrentSyncs,omitempty"`
```
```suggestion
	ConcurrentSyncs *int `json:"concurrentSyncs,omitempty"`
```

Optional fields are pointer types

charts/gardener/controlplane/values.yaml
```diff
@@ -202,6 +202,9 @@ global:
         qps: 100
         burst: 130
       controllers:
+#        event:
+#          concurrentSyncs: 5
+#          ttlShortLivedEvents: 4h
```
rfranzke:
```suggestion
#       event:
#         concurrentSyncs: 5
#         ttlShortLivedEvents: 4h
```

docs/concepts/controller-manager.md
```diff
@@ -54,6 +54,22 @@ The component configuration of the Gardener Controller Manager offers to configu
...
 * `staleExpirationTimeDays`: Expiration time after which stale `Project`s are finally auto-deleted (after `.status.staleSinceTimestamp`). If this value is changed and an auto-delete timestamp got already assigned to the projects then the new value will only take effect if it's increased. Hence, decreasing the `staleExpirationTimeDays` will not decrease already assigned auto-delete timestamps.
 
+### Event Controller
+
+With the Gardener Event Controller you can prolong the lifespan of shoot cluster maintenance and hibernation events.
+This is an optional controller which will become active once you provide the below mentioned configuration.
+
+All events in K8s are deleted after a configurable [kube-apiserver argument](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) called `--event-ttl` (defaulting to 1 hour).
```
```suggestion
All events in K8s are deleted after a configurable time-to-live (controlled via a [kube-apiserver argument](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) called `--event-ttl` (defaulting to 1 hour)).
```

docs/concepts/controller-manager.md
```diff
@@ -54,6 +54,22 @@ The component configuration of the Gardener Controller Manager offers to configu
...
+The need to prolong the time-to-live for certain events frequently arises when debugging life systems.
```
```suggestion
The need to prolong the time-to-live for certain events frequently arises when debugging live systems.
```

or "production" :)

docs/concepts/controller-manager.md
```diff
@@ -54,6 +54,22 @@ The component configuration of the Gardener Controller Manager offers to configu
...
+This is especially true for events impacting the lifecycle of customers clusters directly, namely maintenance and hibernation events concerning shoot clusters.
+This controller leaves the above mentioned events untouched while deleting all other events after a configured time.
+In order to activate it provide the following configuration:
+
+* `ttlShortLivedEvents`: When an event reaches this time-to-live it gets deleted unless it is a maintenance or hibernation event.
+
+> :warning: In addition, you need to configure the `--event-ttl` for the kubeapi-server to define an upper-limit of how long maintenance and hibernation events should be stored.
```
```suggestion
> :warning: In addition, you need to configure the `--event-ttl` for the kube-apiserver to define an upper-limit of how long maintenance and hibernation events should be stored.
```

docs/concepts/controller-manager.md
```diff
@@ -54,6 +54,22 @@ The component configuration of the Gardener Controller Manager offers to configu
...
+With the Gardener Event Controller you can prolong the lifespan of shoot cluster maintenance and hibernation events.
+This is an optional controller which will become active once you provide the below mentioned configuration.
+
+All events in K8s are deleted after a configurable [kube-apiserver argument](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) called `--event-ttl` (defaulting to 1 hour).
+The need to prolong the time-to-live for certain events frequently arises when debugging life systems.
+This is especially true for events impacting the lifecycle of customers clusters directly, namely maintenance and hibernation events concerning shoot clusters.
+This controller leaves the above mentioned events untouched while deleting all other events after a configured time.
+In order to activate it provide the following configuration:
+
+* `ttlShortLivedEvents`: When an event reaches this time-to-live it gets deleted unless it is a maintenance or hibernation event.
```
rfranzke: Hmm, why is it limited to hibernation/maintenance events (instead of simply capturing/keeping all shoot-related events like reconciliations and potential future other events)?

example/20-componentconfig-gardener-controller-manager.yaml
```diff
@@ -29,6 +29,9 @@ controllers:
     staleGracePeriodDays: 14
     staleExpirationTimeDays: 90
     staleSyncPeriod: 12h
+  event:
+    concurrentSyncs: 5
+    ttlShortLivedEvents: 4h
```
rfranzke: Why is it `4h` here while to default to `1h` in the `config/v1alpha1/defaults.go`?

pkg/controllermanager/controller/event/event.go
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/controllermanager/controller/event/event.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sync"
+	"time"
+)
+
+type Controller struct {
```
rfranzke, line 37: doc string for exported struct missing

pkg/controllermanager/controller/event/event_control_test.go
```diff
@@ -0,0 +1,215 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/controllermanager/controller/factory.go
```diff
@@ -17,6 +17,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	eventcontroller "github.com/gardener/gardener/pkg/controllermanager/controller/event"
```
rfranzke:
```suggestion

	eventcontroller "github.com/gardener/gardener/pkg/controllermanager/controller/event"
```

pkg/controllermanager/controller/shoot/shoot_hibernation_types.go
```diff
@@ -123,9 +126,16 @@ func (h *hibernationJob) Run() {
 		return
 	}
 	h.logger.Debugf("Successfully set hibernation.enabled to %t", h.enabled)
+	if h.enabled {
+		msg := "Hibernating cluster"
```
rfranzke:
```suggestion
		msg := "Hibernating cluster due to schedule"
```

pkg/controllermanager/controller/shoot/shoot_hibernation_types.go
```diff
@@ -123,9 +126,16 @@ func (h *hibernationJob) Run() {
...
+		h.recorder.Eventf(h.target, corev1.EventTypeNormal, gardencorev1beta1.ShootEventHibernationEnabled, "%s", msg)
+	} else {
+		msg := "Waking up cluster from hibernation"
```
rfranzke:
```suggestion
		msg := "Waking up cluster due to schedule"
```

We should make clear that these events do not capture the hibernation/wake-ups triggered by the user itself but only those triggered automatically due to a configured hibernation schedule.

pkg/controllermanager/controller/event/event_control.go
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/controllermanager/controller/event/event_control.go
```diff
@@ -0,0 +1,93 @@
...
+	"fmt"
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/keys"
+	"github.com/gardener/gardener/pkg/logger"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"time"
```
rfranzke: Usually, we group the imports in three blocks: First block contains golang core package (`context`, etc.), second block contains `github.com/gardener/gardener` packages, third block contains third-party packages.

pkg/controllermanager/controller/event/event_control.go
```diff
@@ -0,0 +1,93 @@
...
+		if err != nil {
+			log.Debug("Failed to delete event due to %w", err)
+		}
+		return reconcile.Result{}, err
+	}
+}
+
+func longLivedEvent(event *corev1.Event) bool {
+	if event.InvolvedObject.Kind == "Shoot" {
+		if longLivedEventReasons.Has(event.Reason) {
```
rfranzke: As questioned above: Should we consider all events for a `Shoot` as "long-lived" without the above whitelist?

pkg/controllermanager/controller/event/event_control.go
```diff
@@ -0,0 +1,93 @@
...
+	timeUntilDeletion := deleteAt.Sub(nowFunc())
+	if timeUntilDeletion > 0 {
+		return reconcile.Result{RequeueAfter: timeUntilDeletion}, nil
+	} else {
+		err := gardenClient.Client().Delete(ctx, event)
+		if err != nil {
+			log.Debug("Failed to delete event due to %w", err)
+		}
+		return reconcile.Result{}, err
+	}
```
rfranzke:
```suggestion
	}

	err := gardenClient.Client().Delete(ctx, event)
	if err != nil {
		log.Debug("Failed to delete event due to %w", err)
	}
	return reconcile.Result{}, err
```

rfranzke (changes requested):
/assign
Thanks for the PR, pretty well readable!

rfranzke:
Can you please sign the CLA and check `make verify`?

example/20-componentconfig-gardener-controller-manager.yaml
BeckerMax: 1h is the default ttl if none is configured in the kube-api-server.
4h is what we currently have configured in the landscape-setup https://github.wdf.sap.corp/kubernetes/landscape-setup/pull/1466/files
What should we chose? `ttlShortLivedEvents` should replace the ttl currently configured in the kube-api-server. Should I assume the k8s default of 1h our 4h? I lean towards 1h and would update this in all places. WDYT?

docs/concepts/controller-manager.md
BeckerMax: I think we are afraid that events could `pollute` etcd so it runs out of space. What if someone introduces a very frequently recorded event that is now kept for 14days?

rfranzke: End-users are not allowed to create events, so they can only come from operators/components. I'm thinking about the use-case where an extension controller wants to submit events, too. With the current whitelist this would not be easily possible (because it would be garbage-collected too fast), and the extensions would have to implement a similar event controller to keep their events.
So, I don't think there is a risk of pollution at this point in time as event creation is not exposed to end-users.

example/20-componentconfig-gardener-controller-manager.yaml
rfranzke: I'm fine with `4h`, but it should be the same in both places (defaults in the config API and defaults in the example files/Helm chart values).

docs/concepts/controller-manager.md
BeckerMax: As I discussed this also with @timebertt WDYT?

timebertt: Good point. I think, it makes sense to keep all events referring to `Shoot` objects instead of only ones with whitelisted `Reasons`.

docs/concepts/controller-manager.md
```diff
@@ -54,6 +54,22 @@ The component configuration of the Gardener Controller Manager offers to configu
...
+All events in K8s are deleted after a configurable time-to-live (controlled via a [kube-apiserver argument](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) called `--event-ttl` (defaulting to 1 hour)).
+The need to prolong the time-to-live for certain events frequently arises when debugging live systems.
+This is especially true for events impacting the lifecycle of customers clusters directly, namely maintenance and hibernation events concerning shoot clusters.
+This controller leaves the above mentioned events untouched while deleting all other events after a configured time.
+In order to activate it provide the following configuration:
+
+* `ttlShortLivedEvents`: When an event reaches this time-to-live it gets deleted unless it is a maintenance or hibernation event.
+
+> :warning: In addition, you need to configure the `--event-ttl` for the kube-apiserver to define an upper-limit of how long maintenance and hibernation events should be stored.
+The `--event-ttl` should be larger than the `ttl-short-lived-events` or this controller will have no effect.
```
timebertt:
```suggestion
The `--event-ttl` should be larger than the `ttlShortLivedEvents` or this controller will have no effect.
```

pkg/controllermanager/apis/config/types.go
```diff
@@ -95,6 +97,16 @@ type PlantControllerConfiguration struct {
 	SyncPeriod metav1.Duration
 }
 
+// EventControllerConfiguration defines the configuration of the
+// EventControllerConfiguration controller.
```
timebertt:
```suggestion
// Event controller.
```

pkg/controllermanager/apis/config/types.go
```diff
@@ -51,6 +51,8 @@ type ControllerManagerControllerConfiguration struct {
 	CloudProfile *CloudProfileControllerConfiguration
 	// ControllerRegistration defines the configuration of the ControllerRegistration controller.
 	ControllerRegistration *ControllerRegistrationControllerConfiguration
+	// Event defines the configuration of the EventControllerConfiguration controller.
```
timebertt: Can you please cross-check the different API versions and use the same doc strings in both?
```suggestion
	// Event defines the configuration of the Event controller.
```

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -95,6 +98,18 @@ type ControllerRegistrationControllerConfiguration struct {
 	ConcurrentSyncs int `json:"concurrentSyncs"`
 }
 
+// EventControllerConfiguration defines the configuration of the
+// EventControllerConfiguration controller.
```
timebertt:
```suggestion
// event controller.
```

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -95,6 +98,18 @@ type ControllerRegistrationControllerConfiguration struct {
...
+type EventControllerConfiguration struct {
+	// ConcurrentSyncs is the number of workers used for the controller to work on
+	// events.
+	// +optional
+	ConcurrentSyncs *int `json:"concurrentSyncs,omitempty"`
+	// TTLShortLivedEvents is the time-to-live for all non shoot related events
```
timebertt:
```suggestion
	// TTLShortLivedEvents is the time-to-live for all non-shoot related events (defaults to `1h`).
```

pkg/controllermanager/controller/shoot/shoot_hibernation_control_test.go
```diff
@@ -244,7 +247,7 @@ var _ = Describe("Shoot Hibernation", func() {
 							Hibernation: &gardencorev1beta1.Hibernation{},
 						},
 					}
-					job = NewHibernationJob(clientMap, log, &shoot, enabled)
+					job = NewHibernationJob(clientMap, log, recorder, &shoot, enabled)
```
timebertt: Can you use a mock recorder (`pkg/mock/client-go/tools/record/mocks.go`) in this test and `EXPECT()` a call to `Eventf` with the correct type? (maybe it makes sense to have a case for `Hibernated` and on for `WokenUp`)

docs/concepts/controller-manager.md
BeckerMax: I looked at common events with ` kubectl get events --all-namespaces --field-selector "involvedObject.kind==Shoot" --sort-by=reason`. Overwhelmingly they are `SchedulingSuccessful`, `ScheduledNextSync`, `Reconciling` and `Reconciled` and `ReoncileError`. Then some are `deleting` and `deleted`.
1. I do not have live data but it seems these events can add up quickly if we save all of them of the last for example 14 days.
2. Do these really need to be events? Is ScheduledNextSync important enough to be recorded as an event?

example/20-componentconfig-gardener-controller-manager.yaml
BeckerMax: As discussed going to default it to 1h. But making it consistent at all places

pkg/controllermanager/apis/config/v1alpha1/defaults.go
BeckerMax: added

docs/concepts/controller-manager.md
BeckerMax: We also removed this event because it did not seem to be helpful: https://github.com/gardener/gardener/pull/2649/files#diff-a6ee7351b6be637d35413a1a0f49bedcL226

rfranzke: Thanks for checking this. I think it would be helpful to know when a reconciliation/deletion operations ran, so we can keep `Reconciling` an `Deleting`, for example, and remove all others. Together with the to-be-introduced operation ID (that should become part of the event message as well) we can later easily find the correct logs, so it's not necessary to send events for the result of the operation (e.g., `ReconcileError` or `DeletionSuccesful`). At least we could start like this and gain some experience. If we see that we the data volume is not growing to large we can still get those events back.

pkg/controllermanager/controller/event/event.go
```diff
@@ -0,0 +1,120 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package event
+
+import (
+	"context"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
```
timebertt, line 22: Please rebase onto master and run `make format` to sort the imports :)

pkg/controllermanager/controller/event/event_control.go
```diff
@@ -0,0 +1,95 @@
...
+
+	err = gardenClient.Client().Delete(ctx, event)
+	if err != nil {
+		log.Debug("Failed to delete event due to %w", err)
+	}
+	return reconcile.Result{}, err
+}
+
+func longLivedEvent(event *corev1.Event) bool {
+	if event.InvolvedObject.Kind == "Shoot" {
```
timebertt, line 81: Can you also check the API Group?
```suggestion
	if gv, err := schema.ParseGroupVersion(event.InvolvedObject.APIVersion); err != nil || gv.Group != gardencorev1beta1.GroupName {
		return false
	}
	if event.InvolvedObject.Kind == "Shoot" {
```

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -56,6 +56,9 @@ type ControllerManagerControllerConfiguration struct {
 	// ControllerRegistration defines the configuration of the ControllerRegistration controller.
 	// +optional
 	ControllerRegistration *ControllerRegistrationControllerConfiguration `json:"controllerRegistration,omitempty"`
+	// Event defines the configuration of the Event controller.
```
timebertt:
```suggestion
	// Event defines the configuration of the Event controller. If unset, the event controller will be disabled.
```

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -95,6 +98,17 @@ type ControllerRegistrationControllerConfiguration struct {
 	ConcurrentSyncs int `json:"concurrentSyncs"`
 }
 
+// EventControllerConfiguration defines the configuration of the Event controller
+type EventControllerConfiguration struct {
+	// ConcurrentSyncs is the number of workers used for the controller to work on
+	// events.
```
timebertt, line 104:
```suggestion
	// events (defaults to 5).
```

pkg/controllermanager/controller/event/event.go
```diff
@@ -0,0 +1,120 @@
...
+type Controller struct {
+	clientMap              clientmap.ClientMap
+	cfg                    *config.ControllerManagerConfiguration
+	eventQueue             workqueue.RateLimitingInterface
+	numberOfRunningWorkers int
+	workerCh               chan int
+}
+
+// NewController instantiates a new event controller
+func NewController(clientMap clientmap.ClientMap, cfg *config.ControllerManagerConfiguration) *Controller {
```
timebertt: I know, the other controllers also do it this way. But WDYT about passing only the required `EventControllerConfiguration`? (only a suggestion, no strong opinion)
```suggestion
func NewController(clientMap clientmap.ClientMap, cfg *config.EventControllerConfiguration) *Controller {
```

docs/concepts/controller-manager.md
BeckerMax: Now all Shoot-related events will be persisted. I changed the code and documentation.

pkg/controllermanager/controller/event/event_control.go
```diff
@@ -0,0 +1,76 @@
...
+	event := &corev1.Event{}
+	if err := gardenClient.Client().Get(ctx, req.NamespacedName, event); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Debug("Skipping because Event has been deleted")
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	if event.InvolvedObject.Kind == "Shoot" {
```
timebertt: Can you also check, that the API group is `core.gardener.cloud`?

pkg/controllermanager/controller/event/event_control.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (c *Controller) reconcileEvent(req reconcile.Request) (reconcile.Result, error) {
+	ctx := context.TODO()
+
+	gardenClient, err := c.clientMap.GetClient(ctx, keys.ForGarden())
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get garden client: %w", err)
+	}
+
+	log := logger.NewFieldLogger(logger.Logger, "event", req.Name)
```
timebertt:
```suggestion
	log := logger.NewFieldLogger(logger.Logger, "event", req.Namespace + "/" + req.Name)
```

pkg/controllermanager/controller/event/event_control.go
```diff
@@ -0,0 +1,76 @@
...
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	if event.InvolvedObject.Kind == "Shoot" {
+		return reconcile.Result{}, nil
+	}
+
+	deleteAt := event.FirstTimestamp.Add(c.cfg.Controllers.Event.TTLNonShootEvents.Duration)
```
timebertt: Shouldn't we rather check `event.LastTimestamp` here? Otherwise a reoccurring event might be deleted too early, correct?

pkg/controllermanager/controller/event/event_control.go
```diff
@@ -0,0 +1,76 @@
...
+	timeUntilDeletion := deleteAt.Sub(nowFunc())
+	if timeUntilDeletion > 0 {
+		return reconcile.Result{RequeueAfter: timeUntilDeletion}, nil
+	}
+
+	err = gardenClient.Client().Delete(ctx, event)
+	if err != nil {
+		log.Debug("Failed to delete event due to %w", err)
```
timebertt: Should we rather return the error here? It will then be logged only once in all cases, now it is logged twice in case  log level is set to `debug`.

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -129,7 +128,7 @@ func (c *defaultControl) ScheduleShoot(ctx context.Context, obj *gardencorev1bet
 	// If no Seed is referenced, we try to determine an adequate one.
 	seed, err := determineSeed(shoot, c.seedLister, c.shootLister, c.cloudProfileLister, c.config.Schedulers.Shoot.Strategy)
 	if err != nil {
-		c.reportFailedScheduling(shoot, err)
+		schedulerLogger.Errorf("Scheduling Failed: Failed to determine Seed for Shoot %+v: %v", shoot, err)
```
timebertt: I think, `SchedulingFailed` should definitely be reported, as the event is the only interface through which the user can know, why his Shoot is not being created. (kube-scheduler also reports a similar event when it fails to schedule Pods)
Logging it makes sense, but let's keep the event.

pkg/controllermanager/controller/event/event_control_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
+	Context("Shoot Events", func() {
+
+		It("should ignore them", func() {
+			mockClientGet(k8sGardenRuntimeClient, request.NamespacedName, shootEvent)
+			k8sGardenRuntimeClient.EXPECT().Delete(gomock.Any(), gomock.Any()).Times(0)
```
timebertt: Can you remove this line? It is confusing with `Times(0)` :D 
```suggestion
```

pkg/controllermanager/controller/event/event_control_test.go
```diff
@@ -0,0 +1,201 @@
...
+			expected := reconcile.Result{}
+			actual, err := eventController.reconcileEvent(request)
+			Expect(err).To(BeNil())
```
timebertt: Please use `HaveOccurred` for matching errors. It also checks, that `actual` is really an `error` and not just `nil`. This avoids making incorrect assertions :)
```suggestion
			Expect(err).NotTo(HaveOccurred())
```

timebertt (changes requested):
Very nice PR!
I really have only minor comments :)

docs/concepts/controller-manager.md
BeckerMax: Was changed in the newest version to ttlNonShootEvents and corrected

pkg/controllermanager/controller/event/event.go
BeckerMax: nice! Thanks for the tip!

docs/concepts/controller-manager.md
BeckerMax: I tried to adapt the above suggestions regarding replacing events with logs as best as I could in the current commit.

pkg/controllermanager/controller/shoot/shoot_hibernation_control_test.go
BeckerMax: I did that and introduced a second test. This made the test much longer and a bit more clunky. But it tests both cases now including events. WDYT?

pkg/controllermanager/controller/event/event.go
BeckerMax: Great Idea 🤩

docs/concepts/controller-manager.md
BeckerMax: Could you recheck if the events are now fired correctly / logs are written as expected? I left `shoot_control_migrate.go` untouched as it is a new component and we can gather some experience with that first. Also CP migration should not happen frequently I assume.

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -1071,13 +1071,14 @@ var (
...
 const (
-	// ShootEventMaintenanceDone indicates that a maintenance operation has been performed.
-	ShootEventMaintenanceDone = "MaintenanceDone"
-	// ShootEventMaintenanceError indicates that a maintenance operation has failed.
-	ShootEventMaintenanceError = "MaintenanceError"
-
-	// ShootEventSchedulingSuccessful indicates that a scheduling decision was taken successfully.
-	ShootEventSchedulingSuccessful = "SchedulingSuccessful"
+	// ShootEventImageMaintenance indicates that a maintenance operation regarding the image version has been performed.
+	ShootEventImageUpdate = "MachineImageVersionMaintenance"
```
timebertt: A small suggestion: Can you rename this const to `ShootEventImageVersionUpdate` or the other k8s event to make the naming consistent. You could also include Maintenance in the const name instead of Update, but as you wish :)

pkg/gardenlet/controller/shoot/controllerinstallation_control.go
```diff
@@ -94,7 +94,8 @@ func (c *Controller) reconcileControllerInstallationKey(key string) error {
 		}
 
 		c.getShootQueue(shoot).Add(key)
-		c.recorder.Eventf(shoot, corev1.EventTypeNormal, "ExtensionUpdated", "Marked shoot for enqueueing because dependent extension installation was updated: %s", controllerInstallation.Name)
+		shootLogger := logger.NewShootLogger(logger.Logger, shoot.Name, shoot.Namespace)
+		shootLogger.Info("[SHOOT CONTROLLERINSTALLATION] Marked shoot for enqueueing because dependent extension installation was updated", controllerInstallation.Name)
```
timebertt:
```suggestion
		shootLogger.Infof("[SHOOT CONTROLLERINSTALLATION] Marked shoot for enqueueing because dependent extension installation %q was updated", controllerInstallation.Name)
```

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -484,8 +482,7 @@ func (c *Controller) reconcileShoot(logger *logrus.Entry, shoot *gardencorev1bet
 	}
 
 	durationUntilNextSync := c.durationUntilNextShootSync(o.Shoot.Info)
-	message := fmt.Sprintf("Scheduled next queuing time for Shoot in %s (%s)", durationUntilNextSync, time.Now().UTC().Add(durationUntilNextSync))
-	c.recorder.Event(o.Shoot.Info, corev1.EventTypeNormal, "ScheduledNextSync", message)
+	logger.Infof("Scheduled next queuing time in %s (%s) for Shoot %+v", durationUntilNextSync, time.Now().UTC().Add(durationUntilNextSync), o.Shoot.Info)
```
timebertt: The logger already has fields identifying the Shoot :)
```suggestion
	logger.Infof("Scheduled next queuing time in %s (%s)", durationUntilNextSync, time.Now().UTC().Add(durationUntilNextSync))
```

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -505,12 +502,13 @@ func (c *Controller) durationUntilNextShootSync(shoot *gardencorev1beta1.Shoot)
 }
 
 func (c *Controller) updateShootStatusAndRequeueOnSyncError(g gardencore.Interface, shoot *gardencorev1beta1.Shoot, err error) (reconcile.Result, error) {
-	message := fmt.Sprintf("Shoot cannot be synced with Seed: %v", err)
-	c.recorder.Event(shoot, corev1.EventTypeNormal, gardencorev1beta1.EventOperationPending, message)
+	msg := fmt.Sprintf("Shoot cannot be synced with Seed: %v", err)
+	shootLogger := logger.NewShootLogger(logger.Logger, shoot.ObjectMeta.Name, shoot.ObjectMeta.Namespace)
+	shootLogger.Infof("[SYNC SHOOT] %s: %v", msg, err)
```
timebertt: `msg` already includes `err` and I think, this deserves `Error` log level ;)
```suggestion
	shootLogger.Error(err)
```

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -151,12 +151,11 @@ func (c *defaultControl) ScheduleShoot(ctx context.Context, obj *gardencorev1bet
 		if _, ok := err.(*common.AlreadyScheduledError); ok {
 			return nil
 		}
-		c.reportFailedScheduling(shoot, err)
+		schedulerLogger.Errorf("Failed Scheduling Shoot %+v onto Seed %s: %v", shoot, seed.Name, err)
```
timebertt: If you use `%+v` with `shoot` this will probably be a very lengthy log message, right?
```suggestion
		schedulerLogger.Errorf("Failed Scheduling Shoot %q onto Seed %q: %v", shoot.Namespace+"/"+shoot.Name, seed.Name, err)
```

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -505,12 +502,13 @@ func (c *Controller) durationUntilNextShootSync(shoot *gardencorev1beta1.Shoot)
 }
 
 func (c *Controller) updateShootStatusAndRequeueOnSyncError(g gardencore.Interface, shoot *gardencorev1beta1.Shoot, err error) (reconcile.Result, error) {
-	message := fmt.Sprintf("Shoot cannot be synced with Seed: %v", err)
-	c.recorder.Event(shoot, corev1.EventTypeNormal, gardencorev1beta1.EventOperationPending, message)
+	msg := fmt.Sprintf("Shoot cannot be synced with Seed: %v", err)
+	shootLogger := logger.NewShootLogger(logger.Logger, shoot.ObjectMeta.Name, shoot.ObjectMeta.Namespace)
```
timebertt: Can you pass the logger from the calling code to this function instead of creating a new one?

pkg/controllermanager/controller/event/event_control_test.go
```diff
@@ -0,0 +1,208 @@
...
+	})
+
+	Context("Non-Shoot events", func() {
+
+		Context("ttl is not yet reached", func() {
+
+			BeforeEach(func() {
+				nowFunc = func() time.Time {
+					return time.Date(0, 0, 0, 0, 0, 0, 0, time.UTC)
+				}
```
timebertt, line 116: Can you ensure, that `nowFunc` is reset after each test? As you are mocking it for every case in the moment, it doesn't really matter, but it is best practice to reset faked/mocked variables after tests to avoid executing incorrect tests when new cases are added in the future.

You can use our helper func `test.WithVar[s]` for that (see [here](https://github.com/gardener/gardener/blob/056f925a09162fb90fcce938b04c0397250bf25f/pkg/operation/botanist/extensions/infrastructure/infrastructure_test.go#L107-L110) for an example).

pkg/controllermanager/controller/shoot/shoot_hibernation_control_test.go
```diff
@@ -259,6 +278,28 @@ var _ = Describe("Shoot Hibernation", func() {
...
+					shootIface.EXPECT().Get(name, metav1.GetOptions{}).Return(&shoot, nil),
+
+					c.EXPECT().CoreV1beta1().Return(gardenIface),
+					gardenIface.EXPECT().Shoots(namespace).Return(shootIface),
+					shootIface.EXPECT().Update(gomock.AssignableToTypeOf(&gardencorev1beta1.Shoot{})).Do(func(actual *gardencorev1beta1.Shoot) {
+						Expect(actual.Spec.Hibernation).To(Equal(&gardencorev1beta1.Hibernation{
+							Enabled: &enabled,
+						}))
+					}),
+					recorder.EXPECT().Eventf(&shoot, corev1.EventTypeNormal, gardencorev1beta1.ShootEventHibernationDisabled, "%s", "Waking up cluster due to schedule").Times(1),
```
timebertt, line 302: Just FYI, you don't have to specify `.Times(1)`. If no count is specified, it's assumed to be `1`.
Not saying, that you should change it, it's ok to be explicit. Just to let you know :)

timebertt (dismissed):
Looks great! Thanks for including my feedback.
I only have minor comments mainly.

pkg/controllermanager/apis/config/types.go
```diff
@@ -95,6 +97,15 @@ type PlantControllerConfiguration struct {
 	SyncPeriod metav1.Duration
 }
 
+// EventControllerConfiguration defines the configuration of the Event controller.
+type EventControllerConfiguration struct {
+	// ConcurrentSyncs is the number of workers used for the controller to work on
+	// events (defaults to 5).
+	ConcurrentSyncs *int
```
timuthy:
```suggestion
	ConcurrentSyncs int
```

Imo, we should be consistent with the other configuration structs. It's not a problem to introduce it as a required field either because the `EventControllerConfiguration` struct itself is optional in `ControllerManagerControllerConfiguration`.

/cc @rfranzke

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -95,6 +98,17 @@ type ControllerRegistrationControllerConfiguration struct {
 	ConcurrentSyncs int `json:"concurrentSyncs"`
 }
 
+// EventControllerConfiguration defines the configuration of the Event controller.
+type EventControllerConfiguration struct {
+	// ConcurrentSyncs is the number of workers used for the controller to work on
+	// events (defaults to 5).
+	// +optional
+	ConcurrentSyncs *int `json:"concurrentSyncs,omitempty"`
```
timuthy:
```suggestion
	ConcurrentSyncs int `json:"concurrentSyncs"`
```
Same as previous comment.

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -151,12 +151,11 @@ func (c *defaultControl) ScheduleShoot(ctx context.Context, obj *gardencorev1bet
 		if _, ok := err.(*common.AlreadyScheduledError); ok {
 			return nil
 		}
-		c.reportFailedScheduling(shoot, err)
```
timuthy: I'm concerned that we remove this user feedback. Errors can occur during the seed assignment, for example if [DNS is misconfigured](https://github.com/gardener/gardener/blob/056f925a09162fb90fcce938b04c0397250bf25f/plugin/pkg/shoot/dns/admission.go#L207) and with this change the user won't notice such issues.

timuthy (changes requested):
Nice contribution!

pkg/scheduler/controller/shoot/scheduler_control.go
rfranzke: Hm, good point. The problem is that these events easily pile up due to the retry of the scheduler. This may lead to an extensive volume of events. Probably it makes sense to also delete the events of type `Warning` in the event controller if they are older than a certain time?
Or, if there are "too many"' (whatever that is) other `Warning` events with a newer timestamps already, then the event controller could clean up the older ones.
/cc @timebertt

timebertt: Well, I think these events would then not be stored separately in etcd, but rather the eventCount would be increased and the lastTimestamp refreshed, right?

rfranzke: Is it? Then it's all fine :)

timebertt: I would think so: https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/383-new-event-api-ga-graduation
@BeckerMax can you confirm?

danielfoehrKn:
From my perspective the code looks good after addressing the issues of the other reviewers.
Will test it out after the suggestions are applied.

pkg/scheduler/controller/shoot/scheduler_control.go
BeckerMax: Yes, in this case the events are grouped together to a single object. From https://github.com/kubernetes/community/blob/master/contributors/design-proposals/instrumentation/events-redesign.md :
```
Deduplication logic groups together Event which have the same:

- Source Component and Host
- InvolvedObject Kind, Namespace, Name, API version and UID,
- Type (works as Severity)
- Reason
```

pkg/controllermanager/apis/config/types.go
BeckerMax: I see the advantage of consistency. But for me this is a number that should have a reasonable default. The user can hardly guess which number is appropriate here - so the average user should stick to the default and adapt if running into performance problems. That said, I will adapt it now to your suggestion but would still like to hear your opinion.

pkg/controllermanager/controller/event/event_control_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+	Context("Shoot Events", func() {
+
+		It("should ignore them", func() {
+			mockClientGet(k8sGardenRuntimeClient, request.NamespacedName, shootEvent)
+			expected := reconcile.Result{}
+			actual, err := eventController.reconcileEvent(request)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(actual).To(Equal(expected))
+		})
```
timebertt, line 106: Can you add a test case, that checks, that it is still ignored after `ttl`? Does that make sense?

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -156,7 +156,6 @@ func (c *defaultControl) ScheduleShoot(ctx context.Context, obj *gardencorev1bet
 	}
 
 	schedulerLogger.Infof("Shoot '%s' (Cloud Profile '%s', Region '%s') successfully scheduled to seed '%s' using SeedDeterminationStrategy '%s'", shoot.Name, shoot.Spec.CloudProfileName, shoot.Spec.Region, seed.Name, c.config.Schedulers.Shoot.Strategy)
-	c.reportSuccessfulScheduling(shoot, seed.Name)
```
timebertt, line 159: Why did you remove this event? I think it makes sense to keep it, just like the `Scheduled` event from the `kube-scheduler`.

timebertt (commented):
Looks great now FMPOV!
Just 2 questions :)

pkg/controllermanager/controller/event/event_control_test.go
BeckerMax: The only way that could happen in practice is if somebody configures a `ttlNonShootEvents` of `0s` or maybe a very small number. 
As it is not a reasonable case to happen I think we do not need to test this.

pkg/scheduler/controller/shoot/scheduler_control.go
BeckerMax: I removed it as we are limited in the amount of Shoot events we can persist. I think this is a difference to the kube-scheduler which assumes that events are not kept for weeks in etcd.
Initially, in response to a comment by @rfranzke we decided that we remove events signalling results (like scheduled successful or failed). Then we decided that we do still need errors as user feedback in the form of events.
Regarding `SuccessfulScheduling: is this an important user feedback that should be kept for weeks? How often does this event get fired? Depending on this information I would keep or delete it, wdyt?

rfranzke: This event gets only fired once after initial creation at the moment. I have no issues with keeping it, but I'm also fine with not keeping it ;-) Your call..

BeckerMax: As it is analogous to the kubescheduler and only fired once at creation I added the event again.

rfranzke (commented):
Thanks, I'm okay with the PR and don't have further suggestions. The only thing I'd kindly ask you to add is a little bit of documentation here: https://github.com/gardener/gardener/blob/master/docs/concepts/controller-manager.md#control-loops
You can introduce a new section `Event` controller and explain what it does, how to configure it, etc. (does not need to be too long/artificially long, but detailed enough so that an external user understands it).
Thank you very much, and sorry for not mentioning this earlier.

/needs documentation
/status author-action

BeckerMax:
@rfranzke I already had the documentation included. But now I also added the `ConcurrentSyncs` flag as it became a required value.

docs/concepts/controller-manager.md
```diff
@@ -54,6 +54,22 @@ The component configuration of the Gardener Controller Manager offers to configu
...
+With the Gardener Event Controller you can prolong the lifespan of events related to Shoot clusters.
+This is an optional controller which will become active once you provide the below mentioned configuration.
+
+All events in K8s are deleted after a configurable time-to-live (controlled via a [kube-apiserver argument](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) called `--event-ttl` (defaulting to 1 hour)).
+The need to prolong the time-to-live for Shoot cluster events frequently arises when debugging customer issues on live systems.
+This controller leaves events involving Shoots untouched while deleting all other events after a configured time.
+In order to activate it, provide the following configuration:
+
+* `concurrentSyncs`: The amount of goroutines scheduled for reconciling events.
+* `ttlNonShootEvents`: When an event reaches this time-to-live it gets deleted unless it is a Shoot-related event.
```
rfranzke: You could mention that this field is defaulted to 1h (similar to the apiserver flag)

rfranzke (dismissed):
/lgtm

docs/concepts/controller-manager.md
BeckerMax: Done

rfranzke (approved):

pkg/controllermanager/controller/event/event_control_test.go
timebertt: Ok, makes sense

timebertt (approved):
Thanks for adapting all the feedback!
/lgtm
🚀

timuthy (approved):
/lgtm

PR #2653, opened 2020-08-03
[ci:component:github.com/gardener/gardener-resource-manager:v0.14.0->v0.15.0]

ghost:
*Release Notes*:
``` noteworthy developer github.com/gardener/gardener-resource-manager #69 @rfranzke
Resources annotated with `resources.gardener.cloud/delete-on-invalid-update=true` will now be deleted in case the Gardener-Resource-Manager fails to update them and receives an `422 Unprocessable Entity` error. This error is usually sent by the Kubernetes API server in case its static validation fails.
```


timebertt (approved):

rfranzke (approved):
/area usability
/kind enhancement
/size xs
/topology seed shoot
/lgtm

PR #2655, opened 2020-08-03
Use goimports for formatting

timebertt:

/area dev-productivity
/kind task
/priority normal

Following up on discussions like https://github.com/gardener/gardener/pull/2336#discussion_r426385866, this PR switches formatting to goimports, in order to get rid of `please sort the imports` comments on PRs.

Unfortunately, goimports also does not correctly group all third-party/local imports into one import group according to https://github.com/golang/go/wiki/CodeReviewComments#imports.
But as discussed in the linked thread, it probably makes sense to stick to official tools like goimports and agree on the formatting enforced by it, so we can just omit PR comments like `please sort the imports` if goimports accepts the proposed sorting.

Fixes #

**Release note**:

```improvement development
goimports is now used to format go source files.
```


go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/go-openapi/spec v0.19.3
 	github.com/gobuffalo/packr v1.30.1
 	github.com/gogo/protobuf v1.3.1
-	github.com/golang/mock v1.4.3
+	github.com/golang/mock v1.4.4-0.20200731163441-8734ec565a4d
```
dguendisch, line 20: By intent?

dguendisch (changes requested):

go.mod
timebertt: Yes, it is done to include golang/mock#458, which is currently only on master and not yet released.
(see the commit message of the first commit 164b9c5a0f6ead02118148683902f1b1fd4c74ec)

dguendisch: I see, thanks!

dguendisch (approved):

timuthy (approved):
Nice! /lgtm

schrodit (approved):

rfranzke (approved):
I still don't like goimports, but I agree it's necessary to improve our dev productivity... thanks for pushing this forward!
/lgtm

timebertt:
> I still don't like goimports, but I agree it's necessary to improve our dev productivity...

I don't like it either. Totally agreed 😄

PR #2663, opened 2020-08-04
Add utility function to refresh shoot webhook netpols

rfranzke:

/area networking
/kind bug
/priority normal

A utility function that reconciles the shoot webhook configurations has been added to the generic `ControlPlane` actuator package. It is recommended for all extensions that have shoot webhooks to call this function before starting the control loops to ensure that the webhook configurations are updated in case the ports change.

/cc @timebertt @stoyanr 

**Release note**:

```noteworthy developer
A utility function that reconciles the shoot webhook configurations has been added to the generic `ControlPlane` actuator package. It is recommended for all extensions that have shoot webhooks to call this function before starting the control loops to ensure that the webhook configurations are updated in case the ports change.
```

### DIFF
--- a/extensions/pkg/controller/controlplane/genericactuator/actuator.go
+++ b/extensions/pkg/controller/controlplane/genericactuator/actuator.go
@@ -27,6 +27,7 @@ import (
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	gardenerkubernetes "github.com/gardener/gardener/pkg/client/kubernetes"
 	"github.com/gardener/gardener/pkg/utils/chart"
+	"github.com/gardener/gardener/pkg/utils/flow"
 	"github.com/gardener/gardener/pkg/utils/imagevector"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	"github.com/gardener/gardener/pkg/utils/managedresources"
@@ -38,6 +39,8 @@ import (
 	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


extensions/pkg/controller/controlplane/genericactuator/actuator.go
```diff
@@ -525,3 +506,69 @@ func (a *actuator) Migrate(
...
+	}); err != nil {
+		return err
+	}
+
+	fns := make([]flow.TaskFn, 0, len(namespaceList.Items))
+
+	for _, namespace := range namespaceList.Items {
+		var (
+			networkPolicy = extensionswebhookshoot.GetNetworkPolicyMeta(namespace.Name, providerName)
+			namespace     = networkPolicy.Namespace
```
vpnachev: This variable is already set and it has the same value, so I think it is not necessary to redefine it here, wdyt?

rfranzke: Yeah, I wasn't entirely sure whether it's safe to use the iteration variable when the functions are added, so I declared these variables.
/invite @timuthy

rfranzke: I remember we had some issues with that approach, e.g., [here](https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/containerruntime.go#L47)

vpnachev: In this case I would just rename it to not be the same as the loop variable.

vpnachev: I had to refresh, my bad, sorry.

vpnachev (dismissed):
/lgtm

extensions/pkg/controller/controlplane/genericactuator/actuator.go
```diff
@@ -525,3 +506,65 @@ func (a *actuator) Migrate(
...
+		return err
+	}
+
+	fns := make([]flow.TaskFn, 0, len(namespaceList.Items))
+
+	for _, namespace := range namespaceList.Items {
+		networkPolicy := extensionswebhookshoot.GetNetworkPolicyMeta(namespace.Name, providerName)
+
+		fns = append(fns, func(ctx context.Context) error {
+			if err := c.Get(ctx, kutil.Key(namespace.Name, networkPolicy.Name), &networkingv1.NetworkPolicy{}); err != nil {
```
timuthy: I still think the ref to the `namespace` loop var causes issues. See this example https://play.golang.org/p/_ehBa__iRC_e.

timuthy (changes requested):

extensions/pkg/controller/controlplane/genericactuator/actuator.go
rfranzke: Yep, thanks Tim, I've adapted the code and also implemented a unit test for the function. Please check.

timuthy (approved):
/lgtm

extensions/pkg/controller/controlplane/genericactuator/actuator.go
```diff
@@ -525,3 +506,69 @@ func (a *actuator) Migrate(
...
+			networkPolicyName = networkPolicy.Name
+		)
+
+		fns = append(fns, func(ctx context.Context) error {
+			if err := c.Get(ctx, kutil.Key(namespaceName, networkPolicyName), &networkingv1.NetworkPolicy{}); err != nil {
+				if !apierrors.IsNotFound(err) {
+					return err
+				}
+				return nil
+			}
```
stoyanr, line 568: This code skips the webhook reconciliation if the networkpolicy was not found. Why is this needed / when would this happen? Isn't it better to reconcile anyway, just in case?

stoyanr (commented):
Just a small question from my side.

extensions/pkg/controller/controlplane/genericactuator/actuator.go
rfranzke: For example, when a shoot is already in the process of being deleted and the webhook configuration has already been cleaned up (then we don't want to recreate it unnecessarily).

vpnachev (approved):
/lgtm

PR #2660, opened 2020-08-04
Change default webhook server port to 10250

stoyanr:
**How to categorize this PR?**
/area security
/kind enhancement
/priority normal

**What this PR does / why we need it**:
Changes the default webhook server port for `gardener-seed-admission-controller` and `vpa-admission-controller` from 443 to 10250. This is done in order to avoid requesting the capabilities required to use a system port (443), that on some providers (e.g. OpenShift) would require special security configuration. The port 10250 is chosen because it's already open on GKE, and using it avoids special configuration steps on GKE.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:
```improvement operator
The default webhook server port for `gardener-seed-admission-controller` and `vpa-admission-controller` is now changed to 10250.
```

PR #2664, opened 2020-08-04
Fix VPA webhook port configuration

rfranzke:

/area networking
/kind bug
/priority normal

With #2660 the VPA port configuration was changed, however, there are some left overs. The webhook port (= service port) is still 443, and the network policy for the kube-apiserver needs to be updated to the correct new 10250 server port.

Thanks @timebertt for helping me with this!
/cc @stoyanr

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
+++ b/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
@@ -47,7 +47,10 @@ spec:
         {{- if .Values.admissionController.registerByURL }}
         - --register-by-url=true
         - --webhook-address=https://vpa-webhook.{{.Release.Namespace}}
-        - --webhook-port={{ .Values.admissionController.port }}
+        - --webhook-port={{ .Values.admissionController.servicePort }} # the port is only respected if register-by-url is true, that's why it's in this if-block
+                                                                       # if it's false it will not set the port during registration, i.e., it will be defaulted to 443,
+                                                                       # so the servicePort has to be 443 in this case
+                                                                       # see https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/admission-controller/config.go#L70-L74
         {{- end }}
         image: {{ index .Values.global.images "vpa-admission-controller" }}
         imagePullPolicy: IfNotPresent
--- a/charts/seed-controlplane/charts/kube-apiserver/templates/networkpolicy-allow-kube-apiserver-policy.yaml
+++ b/charts/seed-controlplane/charts/kube-apiserver/templates/networkpolicy-allow-kube-apiserver-policy.yaml
@@ -42,7 +42,7 @@ spec:
           app: vpa-admission-controller
     ports:
     - protocol: TCP
-      port: 443
+      port: 10250
   {{- end }}
   ingress:
     # Allow connection from everything which needs to talk to the API server
--- a/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-exporter.yaml
+++ b/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-exporter.yaml
@@ -35,6 +35,7 @@ spec:
       containers:
       - command:
         - /usr/local/bin/vpa-exporter
+        - --port={{ .Values.exporter.port }}
         image: {{ index .Values.global.images "vpa-exporter" }}
         imagePullPolicy: IfNotPresent
         name: exporter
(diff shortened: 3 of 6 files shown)


vpnachev (approved):
/lgtm

timuthy (approved):

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
```diff
@@ -47,7 +47,10 @@ spec:
         {{- if .Values.admissionController.registerByURL }}
         - --register-by-url=true
         - --webhook-address=https://vpa-webhook.{{.Release.Namespace}}
-        - --webhook-port={{ .Values.admissionController.port }}
+        - --webhook-port={{ .Values.admissionController.servicePort }} # the port is only respected if register-by-url is true, that's why it's in this if-block
+                                                                       # if it's false it will not set the port during registration, i.e., it will be defaulted to 443,
+                                                                       # so the servicePort has to be 443 in this case
+                                                                       # see https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/admission-controller/config.go#L70-L74
```
stoyanr, line 53: I don't understand the reasoning behind this change. There is `{{- if .Values.admissionController.registerByURL }}`, so the important thing is what happens if `register-by-url` is true. In this case, the Webhook URL is passed directly to the API server (there is no service), and therefore the port above must be the port on which the webhook server listens to, that is 10250 by default, not 443. What am I missing?

charts/seed-controlplane/charts/kube-apiserver/templates/networkpolicy-allow-kube-apiserver-policy.yaml
```diff
@@ -42,7 +42,7 @@ spec:
           app: vpa-admission-controller
     ports:
     - protocol: TCP
-      port: 443
+      port: 10250
```
stoyanr, line 45: Perhaps this port should be configurable via `values.yaml`. Also, looking at the description of this policy it seems wrong / insufficient, that's why I didn't recognize it should be changed. Perhaps the description should be updated as well?

stoyanr (commented):
There is a question / comment from my side.

charts/seed-controlplane/charts/kube-apiserver/templates/networkpolicy-allow-kube-apiserver-policy.yaml
rfranzke: Yes, agreed. I didn't make it configurable yet because the Go code is not explicitly setting the port value when rendering the VPA chart, so I did it similarly here (to only have a minimal change to get the issue fixed).

Generally, I don't like that we are relying on the default values in the Helm charts on too many places (which is also the reason why it this issue appeared in the first place). Ideally, we define the constants/values in Go and pass them for rendering - everywhere.
This requires some time-consuming refactoring, though... but I really would like to change that. The testability as it's now is not very good.

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
rfranzke: You can check the comment here: https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/admission-controller/main.go#L67

The self-registration process is constructing the final URL out of the values passed by `--webhook-address` and `--webhook-port`, see https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/admission-controller/main.go#L133.

Earlier, the `--webhook-port` was set to the server port (10250) which was wrong - now it's the service port (443)

stoyanr: OK, I got it, so here if `.Values.admissionController.registerByURL` we are still passing the *service* address and port. Why do we ever need to do this? I thought this option would be useful only if there is no service or for some reason we don't want to use it?

charts/seed-controlplane/charts/kube-apiserver/templates/networkpolicy-allow-kube-apiserver-policy.yaml
stoyanr: OK, thanks for the explanation!

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
stoyanr: I answered my own question, never mind.

PR #2641, opened 2020-07-31
Use `host.docker.internal` in local-garden

timebertt:
**How to categorize this PR?**

/area dev-productivity
/kind enhancement
/priority normal

**What this PR does / why we need it**:
This PR replaces `docker.for.mac.localhost` by `host.docker.internal` in our scripts for spinning up the local nodeless garden.
This way, connection from the docker VM to the host is unified for Mac and Windows users.
It also adds `host.docker.internal` to the local garden API server cert, so you can talk to it from within a docker container without TLS verification errors.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```action developer
The `local-garden`'s kube-apiserver cert has been updated to include `host.docker.internal` as an alternative DNS Name, so clients running in docker containers can successfully validate the TLS cert when talking to the local garden. If you have copied the `local-garden` kubeconfig to somewhere else, please update your copy with the newly generate one.
```

PR #2665, opened 2020-08-04
Allow extensions to overwrite service port for webhook registration

rfranzke:

/area networking
/kind enhancement
/priority normal

The code was implicitly assuming that service port and server port are the same. However, some extensions might want to expose their server on a different port than the one they use to expose the service.
This PR allows to specify the service port via the new `--webhook-config-service-port` command line parameter (optional). If it's not present, the old behaviour is preserved (i.e., the service port is defaulted to the server port).

/cc @timebertt @stoyanr 

**Release note**:

```improvement developer
Extension developers who want to use different ports for their provider extension webhook server and the corresponding server port can now specify the service port with the new `--webhook-config-service-port` command line flag. If it's not present the service port is defaulted to the webhook server port (i.e., old behaviour is preserved).
```
```action developer
Extension developers that use the webhook framework with the `url` mode do now have to explicitly specify the port as part of their supplied `--webhook-config-url` flag (earlier, the webhook server port was implicitly used).
```

### DIFF
--- a/extensions/pkg/webhook/cmd/options.go
+++ b/extensions/pkg/webhook/cmd/options.go
@@ -32,6 +32,9 @@ const (
 	ModeFlag = "webhook-config-mode"
 	// URLFlag is the name of the command line flag to specify the URL that is used to register the webhooks in Kubernetes.
 	URLFlag = "webhook-config-url"
+	// ServicePortFlag is the name of the command line flag to specify the service port that exposes the webhook server.
+	// If not specified it will fallback to the webhook server port.
+	ServicePortFlag = "webhook-config-service-port"
 	// NamespaceFlag is the name of the command line flag to specify the webhook config namespace for 'service' mode.
 	NamespaceFlag = "webhook-config-namespace"
 )
@@ -42,6 +45,8 @@ type ServerOptions struct {
 	Mode string
 	// URL is the URl that is used to register the webhooks in Kubernetes.
 	URL string
+	// ServicePort is the service port that exposes the webhook server.
+	ServicePort int
 	// Namespace is the webhook config namespace for 'service' mode.
 	Namespace string
 
@@ -50,10 +55,12 @@ type ServerOptions struct {
 
 // ServerConfig is a completed webhook server configuration.
 type ServerConfig struct {
-	// Mode is the URl that is used to register the webhooks in Kubernetes.
+	// Mode is the webhook client config mode (service or url).
 	Mode string
-	// URL is the URl that is used to register the webhooks in Kubernetes.
+	// URL is the URL that is used to register the webhooks in Kubernetes.
 	URL string
+	// ServicePort is the service port that exposes the webhook server.
+	ServicePort int
 	// Namespace is the webhook config namespace for 'service' mode.
 	Namespace string
 }
@@ -61,9 +68,10 @@ type ServerConfig struct {
 // Complete implements Completer.Complete.
 func (w *ServerOptions) Complete() error {
 	w.config = &ServerConfig{
-		Mode:      w.Mode,
-		URL:       w.URL,
-		Namespace: w.Namespace,
+		Mode:        w.Mode,
+		URL:         w.URL,
+		ServicePort: w.ServicePort,
+		Namespace:   w.Namespace,
 	}
 
 	if len(w.Mode) == 0 {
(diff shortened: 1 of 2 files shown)


vpnachev (dismissed):
/lgtm

timuthy (dismissed):

extensions/pkg/webhook/cmd/options.go
```diff
@@ -215,9 +224,13 @@ func (c *AddToManagerConfig) AddToManager(mgr manager.Manager) ([]admissionregis
 	if err != nil {
 		return nil, nil, errors.Wrapf(err, "could not create webhooks")
 	}
-
 	webhookServer := mgr.GetWebhookServer()
 
+	port := webhookServer.Port
+	if c.Server.ServicePort > 0 {
```
stoyanr: Shouldn't we check the mode being `service` as well?

stoyanr (commented):
Small comment from my side.

extensions/pkg/webhook/cmd/options.go
rfranzke: The port is relevant for both modes, see https://github.com/gardener/gardener/blob/master/extensions/pkg/webhook/registration.go#L122-L127

stoyanr: Yes, but which port? If the mode is URL and the URL is not a service URL, this should be the webhook server port, not the service port, correct? E.g., we should still check for `service` or `url-service`?

rfranzke: Yes, good point, thanks. I have removed the port usage for the registration when the mode is URL - it doesn't make sense, and introducing additional CLI flags would overcomplicate it, I guess. I'll add a release note that the port must be part of the URL when the URL mode is used, thanks.

extensions/pkg/webhook/cmd/options.go
```diff
@@ -215,9 +224,13 @@ func (c *AddToManagerConfig) AddToManager(mgr manager.Manager) ([]admissionregis
 	if err != nil {
 		return nil, nil, errors.Wrapf(err, "could not create webhooks")
 	}
-
 	webhookServer := mgr.GetWebhookServer()
 
+	servicePort := webhookServer.Port
+	if c.Server.Mode == extensionswebhook.ModeService && c.Server.ServicePort > 0 {
```
stoyanr:
```suggestion
	if (c.Server.Mode == extensionswebhook.ModeService || c.Server.Mode == extensionswebhook.ModeURLWithServiceName) && c.Server.ServicePort > 0 {
```

stoyanr (changes requested):

extensions/pkg/webhook/cmd/options.go
stoyanr: Hmm, as you wish, though I am not particularly in favor of this. For mode `url`, you would need to pass the port twice - once in the URL itself (`webhook-config-url`), and another time as `webhook-config-server-port` for the port the webhook server to listen on, then you could do a mistake and pass a different port.

stoyanr: Also, I think this would then require a change in the `Makefile` files in most of the extensions, see e.g. https://github.com/gardener/gardener-extension-provider-azure/blob/255a61a772cd9ef489129d7794e3f5d656647e47/Makefile#L27-L52

extensions/pkg/webhook/cmd/options.go
rfranzke: nice catch, thank you!

extensions/pkg/webhook/cmd/options.go
rfranzke: Hm, if the default HTTPS port is used (443) which is the case for all extensions then no change is needed, or?

stoyanr: But it's not the case, you can see `--webhook-config-server-port=8443` in the above link.

rfranzke: OK, yes, true. Well, then that's the price that this must be added. I will take care when I revendor the extensions. The hostname (`docker.for.mac.localhost` should be anyways replaced by `host.docker.internal`, ref #2641). OK?

stoyanr: OK, thanks

stoyanr (approved):

vpnachev (approved):
/lgtm

PR #2668, opened 2020-08-05
Make watch cache sizes configurable

timebertt:

/area networking
/area usability
/area cost
/kind enhancement
/priority normal

This PR enables configuring the kube-apiserver's `--default-watch-cache-size` and `--watch-cache-sizes` flags via the Shoot API.

Configuring these flags might be useful for large-scale Shoot clusters with a lot of parallel update requests and a lot of watching controllers (e.g. large shooted Seed clusters).
When the API server's watch cache's capacity is too small to cope with the amount of update requests and watchers for a particular resource, it might happen that controller watches are permanently stopped with `too old resource version` errors. This causes (the default client-go implementation of) controller watches to relist all object of the watched resource, which can result in very high network traffic (e.g. when watching ConfigMaps on Seed clusters).

Part of #1953 

Starting from kubernetes v1.19, the API server's watch cache size is adapted dynamically and setting the watch cache size flags will have no effect, except when setting it to 0 (which disables the watch cache) (ref https://github.com/kubernetes/kubernetes/pull/91818).

/invite @rfranzke 
/cc @vlerenc 

**Release note**:

```noteworthy user
It is now possible to configure the kube-apiserver's `--default-watch-cache-size` and `--watch-cache-sizes` flags via the Shoot API (`spec.kubernetes.kubeAPIServer.watchCacheSizes`). Please see [this document](https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml) for an example and consult the [kube-apiserver command-line reference](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) in case you plan on configuring it for your Shoot.
```


pkg/apis/core/types_shoot.go
```diff
@@ -428,6 +432,30 @@ type AdmissionPlugin struct {
...
+	// (flag `--watch-cache-sizes`).
+	// See: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/
+	Resources []ResourceWatchCacheSize
+}
+
+// ResourceWatchCacheSize contains configuration of the API server's watch cache size for one specific resource.
+type ResourceWatchCacheSize struct {
+	// APIGroup is the API group of the resource for which the watch cache size should be configured.
+	// An unset value is used to specify the legacy core API (e.g. for `secrets`).
+	APIGroup string
```
rfranzke:
```suggestion
	APIGroup *string
```

It is optional, right?

pkg/apis/core/v1alpha1/types_shoot.go
```diff
@@ -547,6 +552,33 @@ type AdmissionPlugin struct {
...
+	// +optional
+	Resources []ResourceWatchCacheSize `json:"resources,omitempty" protobuf:"bytes,2,rep,name=resources"`
+}
+
+// ResourceWatchCacheSize contains configuration of the API server's watch cache size for one specific resource.
+type ResourceWatchCacheSize struct {
+	// APIGroup is the API group of the resource for which the watch cache size should be configured.
+	// An unset value is used to specify the legacy core API (e.g. for `secrets`).
+	// +optional
+	APIGroup string `json:"apiGroup,omitempty" protobuf:"bytes,1,opt,name=apiGroup"`
```
rfranzke:
```suggestion
	APIGroup *string `json:"apiGroup,omitempty" protobuf:"bytes,1,opt,name=apiGroup"`
```

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -544,6 +549,33 @@ type AdmissionPlugin struct {
...
+	// +optional
+	Resources []ResourceWatchCacheSize `json:"resources,omitempty" protobuf:"bytes,2,rep,name=resources"`
+}
+
+// ResourceWatchCacheSize contains configuration of the API server's watch cache size for one specific resource.
+type ResourceWatchCacheSize struct {
+	// APIGroup is the API group of the resource for which the watch cache size should be configured.
+	// An unset value is used to specify the legacy core API (e.g. for `secrets`).
+	// +optional
+	APIGroup string `json:"apiGroup,omitempty" protobuf:"bytes,1,opt,name=apiGroup"`
```
rfranzke:
```suggestion
	APIGroup *string `json:"apiGroup,omitempty" protobuf:"bytes,1,opt,name=apiGroup"`
```

charts/seed-controlplane/charts/kube-apiserver/values-test.yaml
```diff
@@ -0,0 +1,8 @@
+watchCacheSizes:
```
rfranzke: Left over?

pkg/apis/core/validation/shoot.go
```diff
@@ -635,6 +637,27 @@ func validateNetworking(networking core.Networking, fldPath *field.Path) field.E
 	return allErrs
 }
 
+// ValidateWatchCacheSizes validates the given WatchCacheSizes fields.
+func ValidateWatchCacheSizes(sizes *core.WatchCacheSizes, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if sizes != nil {
+		if defaultSize := sizes.Default; defaultSize != nil && *defaultSize < 0 {
```
rfranzke: You could use `apivalidation.ValidateNonnegativeField`.

pkg/apis/core/validation/shoot.go
```diff
@@ -635,6 +637,27 @@ func validateNetworking(networking core.Networking, fldPath *field.Path) field.E
...
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("default"), *defaultSize, "must not be negative"))
+		}
+
+		for idx, resourceWatchCacheSize := range sizes.Resources {
+			idxPath := fldPath.Child("resources").Index(idx)
+			if len(resourceWatchCacheSize.Resource) == 0 {
+				allErrs = append(allErrs, field.Required(idxPath.Child("resource"), "must not be empty"))
+			}
+			if resourceWatchCacheSize.CacheSize < 0 {
+				allErrs = append(allErrs, field.Invalid(idxPath.Child("size"), resourceWatchCacheSize.CacheSize, "must not be negative"))
```
rfranzke: same as above

rfranzke (changes requested):
nits

charts/seed-controlplane/charts/kube-apiserver/values-test.yaml
timebertt: Nope. I thought, it would be good to test, that the chart can be correctly rendered, if the values are added (default values are commented out).
But as you said, we should generally think about a proper way to handle default values in our charts. So I don't mind removing it in this PR :)

pkg/apis/core/types_shoot.go
timebertt: Yes, it's optional. And it even works that way, as empty strings are omitted on `omitempty`.
Do I still have to use a pointer for optional strings according to the API conventions?

rfranzke: I would vote for 'yes', if there is no strong argument against it. :)

timebertt: Ah, found it: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required
Will adapt :)

charts/seed-controlplane/charts/kube-apiserver/values-test.yaml
rfranzke: Hm, yeah, right. I would probably do that separately, i.e., we should comment all Helm chart default values and set them explicitly in code. Separate topic, though 😄

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -457,6 +457,11 @@ type KubeAPIServerConfig struct {
 	// of the kube-apiserver.
 	// +optional
 	ServiceAccountConfig *ServiceAccountConfig `json:"serviceAccountConfig,omitempty" protobuf:"bytes,8,opt,name=serviceAccountConfig"`
+	// WatchCacheSizes contains configuration of the API server's watch cache sizes.
+	// Starting from kubernetes v1.19, the API server's watch cache size is adapted dynamically and setting the watch
+	// cache size flags will have no effect, except when setting it to 0 (which disables the watch cache).
+	// +optional
+	WatchCacheSizes *WatchCacheSizes `json:"watchCacheSizes,omitempty" protobuf:"bytes,9,opt,name=watchCacheSizes"`
```
zanetworker, line 468: do you think it makes sense to add user documentation for this?

pkg/apis/core/v1beta1/types_shoot.go
```diff
@@ -544,6 +549,33 @@ type AdmissionPlugin struct {
...
+type WatchCacheSizes struct {
+	// Default configures the default watch cache size of the kube-apiserver
+	// (flag `--default-watch-cache-size`, defaults to 100).
+	// See: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/
+	// +optional
+	Default *int32 `json:"default,omitempty" protobuf:"varint,1,opt,name=default"`
+	// Resources configures the watch cache size of the kube-apiserver per resource
+	// (flag `--watch-cache-sizes`).
+	// See: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/
+	// +optional
```
zanetworker, line 566: While its nice that we expose it, the semantics here might be vague. Like when should I as a user decided to set this optional field?

pkg/apis/core/v1beta1/types_shoot.go
timebertt: Hmm, TBH, I think the comments here (which will also be transported to the API reference) are enough documentation on our side. I see, the topic is rarely documented on k/k side, but I would rather see it there...

pkg/apis/core/v1beta1/types_shoot.go
rfranzke: Hm, if you want to configure the `--watch-cache-sizes` flag, then you set this list (as stated in the doc string). Why is it vague from your point of view/what could be improved?

zanetworker: Point is, why would I want to configure `--watch-cache-sizes`, how will the user see the need to configure this field?

rfranzke: Hm, okay, I see what you are meaning. Doesn't this rather belong in k/k documentation? We also don't explain why/when a user should set the OIDC or the feature gates or similar.

zanetworker: OIDC is a feature and it's optional (not affecting performance or costs). This feature / flag here is managed, either we un-expose  it (because only us understand the problem it tries to solve) or hide it (only operators set it when they see symptoms of cache size issues), or provide documentation on what to do when there are 1) unexpected high costs, 2) performance degradation or error logs that are okay (according to the community https://github.com/kubernetes/kubernetes/issues/85723#issuecomment-559583963) but not okay when they happen too often because they signal an underlying problem.

rfranzke: I agree, but again, I see this documentation rather in k/k as the kube-apiserver exposes this flag. We just "tunnel it through". I'm happy to refer to such documentation from here, but I don't think we should maintain it ourselves.

zanetworker: The issue does not usually happen (its mainly our design and choice of GWs that led it to happen), so kinda felt that we are partially responsible to document the semantics. But sure, was just voicing it out for discussion :)

pkg/apis/core/types_shoot.go
timebertt: Done

charts/seed-controlplane/charts/kube-apiserver/values-test.yaml
timebertt: Done, removed the file.

pkg/apis/core/validation/shoot.go
timebertt: Done

rfranzke (dismissed):
/lgtm

pkg/apis/core/v1beta1/types_shoot.go
timebertt: I don't quite get what you mean. This PR is just exposing the two API server flags to make it configurable by users, nothing more.
So this PR alone does not tackle a specific issue, but rather augments the configuration options.
We will of course be using this option to tackle the issue we have seen on big Seeds, but still endusers can run into similar scalability issues, when running large clusters, so it makes sense to have it configurable.

zanetworker: > but still endusers can run into similar scalability issues, when running large clusters, so it makes sense to have it configurable. 

This is what I mean, documenting when do they need to use these flags. Lets take this offline if you still don't see my point. I resolved the comment for now.

timebertt: So you are basically asking for a few sentences like in the PR description to be added to the doc string, right?

> Configuring these flags might be useful for large-scale Shoot clusters with a lot of parallel update requests and a lot of watching controllers (e.g. large shooted Seed clusters).
When the API server's watch cache's capacity is too small to cope with the amount of update requests and watchers for a particular resource, it might happen that controller watches are permanently stopped with too old resource version messages.

If that's what you meant, I can do so :)

charts/seed-controlplane/charts/kube-apiserver/templates/_helpers.tpl
```diff
@@ -120,3 +120,20 @@ rules:
 {{- end }}
 {{- end -}}
 {{- end -}}
+
+{{- define "kube-apiserver.watchCacheSizes" -}}
+{{- with .Values.watchCacheSizes }}
```
ialidzhikov, line 125: Wouldn't be easier to read and understand with usage of `if statements` for emptiness instead of `with`. Currently `with` sets the value of `watchCacheSizes` to dot (`.`) and I was wondering from where comes `.resources` and `.default`. WDYT?

```suggestion
{{- if .Values.watchCacheSizes }}
# do something
{{- end }}
{{- if .Values.watchCacheSizes.resources }}
# do something
{{- end }}
```

timebertt: I did it like you suggested before and I wasn't happy with the long value references so I switched to `with` to make it more readable. But no strong opinion :)

pkg/apis/core/v1beta1/types_shoot.go
zanetworker: the doc string would be good enough for now I guess :)

timebertt: Great, done :)

ialidzhikov (dismissed):
/lgtm

zanetworker (dismissed):
/lgtm

rfranzke (dismissed):

timebertt:
Fixed `check-generate`.

vpnachev (approved):
/lgtm

PR #2669, opened 2020-08-05
Clean-up DNS provider migration

timuthy:

/area robustness
/kind cleanup
/priority normal

This PR removes migration logic which was once introduced by https://github.com/gardener/gardener/pull/1959 to migrate primary DNS providers. Furthermore, functionless DNS providers (non-primary provider w/o a secret ref) are now removed and logged by the DNS admission plugin, not by the shoot controller anymore. 

We once put this to the controller to prevent spec changes for shoot resources that would have happened outside of the maintenance time window (due to regular `shoot.gardener.cloud/status` labelling). But since we have this migration logic for a while we can be sure, that each shoot was at least reconciled once with one of the previously released g/g versions.

**Release note**:

```improvement operator
NONE
```


rfranzke (approved):
/lgtm

PR #2678, opened 2020-08-07
Fix dependencies in shoot flow

timuthy:

/area robustness
/kind bug
/priority critical

This PR fixes a nil pointer dereference (due to race condition) which happens during the creation of new shoot clusters.

**Release note**:

```improvement operator
An issue has been fixed which caused the Gardenlet to exit ungracefully due to the missing shoot cluster identify in the `.status`.
```


rfranzke (approved):
/lgtm

PR #2679, opened 2020-08-07
Creates new objects every time artifact.newFunc is called

plkokanov:

/area control-plane
/kind bug
/priority blocker

This PR makes it so that `artifact.newFunc` will return a new object every time it is called instead of reusing the same object per extension resource kind or controllerinstallation kind.

None

**Release note**:

```improvement operator
Objects into which  `controllerinstallations` and `extensions` are retrieved during reconciliation by the ControllerInstallation controller and ShootState Sync controller are now created for each reconciliation, instead of the same object being reused multiple times per resource kind.
```


timuthy (approved):
/lgtm

Thank you for taking care!

rfranzke (approved):
/lgtm

plkokanov:
Created 3 clusters locally successfully without additional data in the shootstates.
We can add an additional test for the planned cp-migration integration tests to check if shootstates get generated and populated properly.

rfranzke:
Yes, please. Actually, this test should be independent of the control plane migration but be a dedicated integration test for the shoot state controller.

PR #2694, opened 2020-08-10
[GEP-7] Don't add entries with empty data in ShootState

plkokanov:

/area control-plane-migration
/kind enhancement
/priority normal

Previously the ShootState contained entries which don't actually have any data e.g.: 
```
- data:
       certificate: null
       privateKey: null
   name: kubecfg
   type: certificate
```

This PR makes it so that when trying to extract infodata from existing secrets and the secrets do not contain any data which has to be saved, these empty entries are not added to the ShootState.

Fixes #

**Release note**:

```improvement operator
Entries with empty data are no longer added to the ShootState.
```

### DIFF
--- a/pkg/operation/shootsecrets/secrets_manager.go
+++ b/pkg/operation/shootsecrets/secrets_manager.go
@@ -271,12 +271,6 @@ func (s *SecretsManager) generateInfoDataAndUpdateResourceList(secretConfig secr
 	if err != nil {
 		return err
 	}
-
-	// No data was generated, means that nothing should be saved in the ShootState
-	if data == nil {
-		return nil
-	}
-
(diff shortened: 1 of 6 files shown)


rfranzke (approved):
/lgtm
/needs second-opinion

swilen-iwanow:
Should we take care of the empty entries already in the ShootState somehow. I think the pr doesn't remove them, right?

plkokanov:
The empty entries currently don't cause any errors. 
Furthermore, I would have to hardcode that the kubecfg entry has to be deleted if it's certificates are null, or implement a function that checks if infodata should be removed depending on the current infodata implementation.

swilen-iwanow (approved):
the leftover fields in ShootState are my only concern, otherwise LGTM

PR #2695, opened 2020-08-10
Improve condition when VPA for shooted seeds is disabled

### LINKED ISSUE #2683: VPA deleted too early by seed controller when a shooted seeds enables VPA feature in `.spec.kubernetes`

/area auto-scaling
/kind bug
/priority critical
/exp intermediate
/platform all
/topology seed

**What happened**:
When a shooted seed sets `.spec.kubernetes.verticalPodAutoscaler.enabled=true` then it might happen that this change is    only rolled out after 24h, i.e., in the maintenance time window (if `.spec.maintenance.confineSpecUpdates=true`). However, the seed controller already cleans up the VPA deployment in the `garden` namespace, leading to the fact that the seed might not have a VPA for <= 24h.

VPA in the `garden` namespace should only be deleted if the new VPA deployment has already been rolled out next to the deployment.

See above.

- Gardener version: v1.8

rfranzke:

/area quality
/kind bug
/priority critical
/size xs
/topology garden seed
/platform all

With this PR, the shooted seed controller will only disable the VPA in the seed if it already has been deployed next to the shoot control plane. As an example, assuming a shooted seed with `.spec.maintenance.confineSpecUpdates=true` and `spec.kubernetes.verticalPodAutoscaler.enabled=true`. When the new gardenlet gets deployed, then

* it enqueues the shooted seed and deploys itself into the cluster, however, it keeps `vpaEnabled=true` as the VPA deployment was not yet found next to the control plane
* if the shooted is reconciled in its maintenance time window then it will wait until the reconciliation finishes and requeue the shooted seed with `10s` delay
* after the reconciliation finished the VPA deployment next to the control plane will be found, i.e., the `vpaEnabled` will be set to `false`.

Fixes #2683

**Release note**:

```improvement operator
A bug has been fixed that might have caused a `24h` absence of the VPA for shooted seeds.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/seed_registration_control.go
+++ b/pkg/gardenlet/controller/shoot/seed_registration_control.go
@@ -149,16 +149,21 @@ func (c *defaultSeedRegistrationControl) Reconcile(shootObj *gardencorev1beta1.S
 	}
 
 	if shoot.DeletionTimestamp == nil && shootedSeedConfig != nil {
-		if shoot.Status.LastOperation == nil || shoot.Status.LastOperation.State != gardencorev1beta1.LastOperationStateSucceeded {
+		if shoot.Generation != shoot.Status.ObservedGeneration || shoot.Status.LastOperation == nil || shoot.Status.LastOperation.State != gardencorev1beta1.LastOperationStateSucceeded {
 			shootLogger.Infof("[SHOOTED SEED REGISTRATION] Waiting for shoot %s to be reconciled before registering it as seed", shoot.Name)
 			return reconcile.Result{
 				RequeueAfter: 10 * time.Second,
 			}, nil
 		}
 
+		seedClient, err := c.clientMap.GetClient(ctx, keys.ForSeedWithName(*shootObj.Spec.SeedName))
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to get seed client: %w", err)
+		}
+
 		if shootedSeedConfig.NoGardenlet {
 			shootLogger.Infof("[SHOOTED SEED REGISTRATION] Registering %s as seed as configuration says that no gardenlet is desired", shoot.Name)
-			if err := registerAsSeed(ctx, gardenClient.Client(), shoot, shootedSeedConfig); err != nil {
+			if err := registerAsSeed(ctx, gardenClient.Client(), seedClient.Client(), shoot, shootedSeedConfig); err != nil {
 				message := fmt.Sprintf("Could not register shoot %q as seed: %+v", shoot.Name, err)
 				shootLogger.Errorf(message)
 				c.recorder.Event(shoot, corev1.EventTypeWarning, "SeedRegistration", message)
@@ -171,7 +176,7 @@ func (c *defaultSeedRegistrationControl) Reconcile(shootObj *gardencorev1beta1.S
 			}
 
 			shootLogger.Infof("[SHOOTED SEED REGISTRATION] Deploying gardenlet into %s which will register shoot as seed", shoot.Name)
-			if err := deployGardenlet(ctx, gardenClient, shootedSeedClient, shoot, shootedSeedConfig, c.imageVector, c.config); err != nil {
+			if err := deployGardenlet(ctx, gardenClient, seedClient, shootedSeedClient, shoot, shootedSeedConfig, c.imageVector, c.config); err != nil {
 				message := fmt.Sprintf("Could not deploy Gardenlet into shoot %q: %+v", shoot.Name, err)
 				shootLogger.Errorf(message)
 				c.recorder.Event(shoot, corev1.EventTypeWarning, "GardenletDeployment", message)
@@ -287,19 +292,19 @@ func applySeedSecret(ctx context.Context, c client.Client, shoot *gardencorev1be
 	return err
 }
 
-func prepareSeedConfig(ctx context.Context, c client.Client, shoot *gardencorev1beta1.Shoot, shootedSeedConfig *gardencorev1beta1helper.ShootedSeed, secretRef *corev1.SecretReference) (*gardencorev1beta1.SeedSpec, error) {
-	shootSecret, err := getShootSecret(ctx, c, shoot)
+func prepareSeedConfig(ctx context.Context, gardenClient client.Client, seedClient client.Client, shoot *gardencorev1beta1.Shoot, shootedSeedConfig *gardencorev1beta1helper.ShootedSeed, secretRef *corev1.SecretReference) (*gardencorev1beta1.SeedSpec, error) {
+	shootSecret, err := getShootSecret(ctx, gardenClient, shoot)
 	if err != nil {
 		return nil, err
 	}
 
-	backupProfile, err := applySeedBackupConfig(ctx, c, shoot, shootSecret, shootedSeedConfig)
+	backupProfile, err := applySeedBackupConfig(ctx, gardenClient, shoot, shootSecret, shootedSeedConfig)
 	if err != nil {
 		return nil, err
 	}
 
 	if secretRef != nil {
-		if err := applySeedSecret(ctx, c, shoot, shootSecret, secretRef.Name, secretRef.Namespace); err != nil {
+		if err := applySeedSecret(ctx, gardenClient, shoot, shootSecret, secretRef.Name, secretRef.Namespace); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/gardenlet/controller/shoot/seed_registration_control.go
```diff
@@ -730,3 +740,23 @@ func checkSeedAssociations(ctx context.Context, c client.Client, seedName string
 
 	return nil
 }
+
+func mustEnableVPA(ctx context.Context, c client.Client, shoot *gardencorev1beta1.Shoot) (bool, error) {
+	if !gardencorev1beta1helper.ShootWantsVerticalPodAutoscaler(shoot) {
```
vpnachev: If the VPA was just disabled but the shoot is still not reconciled, then both shoot and seed controllers will maintain the VPA till the next shoot reconciliation. Will be there any conflicts or a single cluster can be running with two separate VPA deployments?

pkg/gardenlet/controller/shoot/seed_registration_control.go
```diff
@@ -730,3 +740,23 @@ func checkSeedAssociations(ctx context.Context, c client.Client, seedName string
...
+		// Shoot does not not configure the VPA in the `.spec.kubernetes` section, so we need to enable the VPA for this
```
timebertt:
```suggestion
		// Shoot does not configure the VPA in the `.spec.kubernetes` section, so we need to enable the VPA for this
```

pkg/gardenlet/controller/shoot/seed_registration_control.go
timebertt: Good catch! There will most likely be conflicts.

rfranzke: Hm, which kind of conflicts? Any suggestions how to resolve them?

pkg/gardenlet/controller/shoot/seed_registration_control.go
```diff
@@ -149,16 +149,21 @@ func (c *defaultSeedRegistrationControl) Reconcile(shootObj *gardencorev1beta1.S
...
 	if shoot.DeletionTimestamp == nil && shootedSeedConfig != nil {
-		if shoot.Status.LastOperation == nil || shoot.Status.LastOperation.State != gardencorev1beta1.LastOperationStateSucceeded {
+		if shoot.Generation != shoot.Status.ObservedGeneration || shoot.Status.LastOperation == nil || shoot.Status.LastOperation.State != gardencorev1beta1.LastOperationStateSucceeded {
 			shootLogger.Infof("[SHOOTED SEED REGISTRATION] Waiting for shoot %s to be reconciled before registering it as seed", shoot.Name)
 			return reconcile.Result{
 				RequeueAfter: 10 * time.Second,
 			}, nil
 		}
 
+		seedClient, err := c.clientMap.GetClient(ctx, keys.ForSeedWithName(*shootObj.Spec.SeedName))
```
timebertt, line 159: Hmm, I'm wondering if `shootObj.Spec.SeedName` can be `nil` here.
This is implicitly covered by `shoot.Generation != shoot.Status.ObservedGeneration`, right?

pkg/gardenlet/controller/shoot/seed_registration_control.go
vpnachev: It there is no vpa deployment in the shoot namespace in the seed, then the seed controller should deploy it in the garden namespace inside the shoot. Pretty much similar to the case when VPA on the shoot has been just enabled, but waiting next reconciliation to deploy.

timebertt: > Hm, which kind of conflicts?

Conflicts as in multiple `vpa-updater` instances simultaneously trying to evict Pods for example.

rfranzke: Hm, is this a problem? The VPA admission controller is still updating the limits, so are these serious conflicts?

timebertt: I would expect that this will not go well. There will even be multiple webhooks, right?
So I think, we should avoid this if we can.
If you remove the check with `gardencorev1beta1helper.ShootWantsVerticalPodAutoscaler` and only rely on the check if the VPA deployment is found, then it would be sufficient, I think.

timebertt: Ah, there won't be multiple webhooks, the second admission controller will overwrite it probably. So forget that part...

rfranzke: I had a quick call with @vpnachev. So indeed, we agreed that this situation should be avoided (especially in the reverse case where shoot VPA is disabled it might happen that the seed does not have any VPA at all for a period of time).

The very easiest solution would be to put the VPA deployment task at the very end of the shoot reconciliation flow. This way, after the flow finishes, the `Seed` object is updated with `10s` and either `.spec.settings.verticalPodAutoscaler.enabled` will be `true` or `false`, but the maximum delay would be `10s`, as said.

A more complex solution requires some kind of synchronization between the seed controller and the shooted seed controller. For example, the `DeployVPA` function could annotate the `Seed` object (in case the shoot is a shooted seed) with information about the VPA deployment status, and then the seed controller can, based on this annotation, determine whether to delete/deploy the VPA.

Opinions?

vpnachev: I vote for the current solution and open an enhancement issue with the above proposal.

timuthy: +1 for option 1.

rfranzke: TBH, I would prefer fixing it right away as we are on it now. I'd also wait for option 1

vpnachev: Sorry, yes option 1 makes sense to be implemented now.
I have no strong opinion for the rest.

rfranzke: OK, I've made the changes, please check again :)

pkg/gardenlet/controller/shoot/seed_registration_control.go
rfranzke: Probably it cannot happen because a) the events [are filtered](https://github.com/gardener/gardener/blob/master/pkg/gardenlet/controller/shoot/shoot.go#L132) to match the gardenlet's seed config, (i.e., when the `Shoot` has no `.spec.seedName` it will never reach this code path), and b) the `.status.lastOperation` will only be set if the `Shoot` is being reconciled (which only happens when it's assigned to a `Seed`).

timebertt: Ah, right. Great! :)

vpnachev (dismissed):
/lgtm

pkg/gardenlet/controller/shoot/seed_registration_control.go
timebertt: As discussed on the phone. I think, we should remove this check and only use the check below (`vpa-admission-controller` present?) as an indicator for whether to enable the VPA in the Seed.
This way, we can ensure, that the seed registration controller doesn't enable the Seed VPA after the Shoot VPA was disabled but not yet deleted.

timebertt (changes requested):

pkg/gardenlet/controller/shoot/seed_registration_control.go
timebertt: You could probably also leverage the Cluster resource here, just like in https://github.com/gardener/gardener/pull/2698#discussion_r468030823.

rfranzke: I've removed the check for now, ptal

timebertt (approved):
Thanks!
/lgtm

vpnachev (approved):
/lgtm

pkg/gardenlet/controller/shoot/seed_registration_control.go
```diff
@@ -730,3 +740,17 @@ func checkSeedAssociations(ctx context.Context, c client.Client, seedName string
...
+		if apierrors.IsNotFound(err) {
+			// VPA deployment in shoot namespace was not found, so we have to enable the VPA for this seed until it's
+			// being deployed.
+			return true, nil
+		}
+		return false, err
+	}
+
+	// VPA deployment in shoot namespace was found, so we don't need to enable the VPA for this seed.
+	return false, nil
```
timuthy, line 755: Shouldn't we check for a `deletionTimestamp` in case a deployment was found? Maybe the deployment is not gone, yet.

danielfoehrKn (approved):
/lgtm

pkg/gardenlet/controller/shoot/seed_registration_control.go
rfranzke: We can do that, but we don't do it on other places because the `Deployment` is not backed by a finalizer, i.e., it will be immediately removed from the system once you delete it (at least how we do it), even if there is still the replicaset/pods.

timuthy: Okay, valid point. Thanks!

PR #2698, opened 2020-08-10
Add VPA control plane health check

### LINKED ISSUE #2685: Health status of VPA for shoots is not visible in the `ControlPlaneHealthy` condition

/area ops-productivity
/kind bug
/priority normal
/platform all
/exp intermediate
/topology garden seed

**What happened**:
When the VPA is enabled for shoots (`.spec.kubernetes.verticalPodAutoscaler.enabled=true`) then its health is not visible in the `ControlPlaneHealthy` condition, i.e., the shoot care controller does not check it. This leads to the fact that a broken VPA is hardly detected.

The `ControlPlaneHealthy` condition for shoots should also reflect the health of the VPA.

Create a shoot with VPA enabled, break it, and see the `ControlPlaneHealthy` condition to be unchanged.

- Gardener version: v1.8

timuthy:

/area control-plane
/kind enhancement
/priority normal

This PR enables health checks for VPA deployments belonging to a shoot cluster. It is visible as part of `.status.conditions[type=ControlPlaneHealthy]`.

Fixes #2685

An issues with the VPA deployment in the seed bootstrap was fixed along the way. Due to the `runtime` and `application` split in https://github.com/gardener/gardener/pull/2478, configured `values` in the seed bootstrap were not respected anymore. 

**Release note**:

```improvement operator
Control plane health checks have been added for VPA components of shoot clusters. These are executed regularly as soon as the shoot uses the Vertical-Pod-Autoscaling feature (https://github.com/gardener/gardener/blob/master/docs/usage/shoot_autoscaling.md#vertical-pod-auto-scaling).
```

### DIFF
--- a/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
+++ b/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
@@ -7,7 +7,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     app: vpa-admission-controller
-{{ toYaml .Values.labels | indent 4 }}
+{{ toYaml .Values.deploymentLabels | indent 4 }}
 spec:
   replicas: {{ .Values.admissionController.replicas }}
   revisionHistoryLimit: 0
--- a/pkg/apis/core/v1beta1/constants/types_constants.go
+++ b/pkg/apis/core/v1beta1/constants/types_constants.go
@@ -78,6 +78,15 @@ const (
 	// the kube-state-metrics-seed pod.
 	DeploymentNameKubeStateMetricsSeed = "kube-state-metrics-seed"
 
+	// DeploymentNameVPAAdmissionController is a constant for the name of the VPA admission controller deployment.
+	DeploymentNameVPAAdmissionController = "vpa-admission-controller"
+	// DeploymentNameVPAExporter is a constant for the name of the VPA exporter deployment.
+	DeploymentNameVPAExporter = "vpa-exporter"
+	// DeploymentNameVPARecommender is a constant for the name of the VPA recommender deployment.
+	DeploymentNameVPARecommender = "vpa-recommender"
+	// DeploymentNameVPAUpdater is a constant for the name of the VPA updater deployment.
+	DeploymentNameVPAUpdater = "vpa-updater"
+
 	// StatefulSetNameAlertManager is a constant for the name of a Kubernetes stateful set object that contains
 	// the alertmanager pod.
 	StatefulSetNameAlertManager = "alertmanager"
--- a/pkg/apis/core/v1beta1/constants/utils.go
+++ b/pkg/apis/core/v1beta1/constants/utils.go
@@ -34,3 +34,12 @@ func getDeprecatedAnnotation(annotations map[string]string, annotationKey, depre
 
 	return val, ok
 }
+
+// GetShootVPADeploymentNames returns the names of all VPA related deployments related to shoot clusters.
+func GetShootVPADeploymentNames() []string {
+	return []string{
+		DeploymentNameVPAAdmissionController,
+		DeploymentNameVPARecommender,
+		DeploymentNameVPAUpdater,
+	}
+}
--- a/pkg/operation/botanist/health_check.go
+++ b/pkg/operation/botanist/health_check.go
@@ -265,6 +265,12 @@ func computeRequiredControlPlaneDeployments(
 		}
 	}
 
+	if gardencorev1beta1helper.ShootWantsVerticalPodAutoscaler(shoot) {
+		for _, vpaDeployment := range v1beta1constants.GetShootVPADeploymentNames() {
+			requiredControlPlaneDeployments.Insert(vpaDeployment)
+		}
+	}
+
 	return requiredControlPlaneDeployments, nil
 }
 
--- a/pkg/operation/botanist/health_check.go
+++ b/pkg/operation/botanist/health_check.go
@@ -26,6 +26,7 @@ import (
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	gardenerextensions "github.com/gardener/gardener/pkg/extensions"
 	"github.com/gardener/gardener/pkg/features"
 	gardenletfeatures "github.com/gardener/gardener/pkg/gardenlet/features"
 	"github.com/gardener/gardener/pkg/logger"
@@ -40,6 +41,7 @@ import (
 	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -481,6 +483,7 @@ func (b *HealthChecker) CheckExtensionCondition(condition gardencorev1beta1.Cond
 
 // checkControlPlane checks whether the control plane of the Shoot cluster is healthy.
 func (b *Botanist) checkControlPlane(
+	ctx context.Context,
 	checker *HealthChecker,
 	condition gardencorev1beta1.Condition,
 	seedDeploymentLister kutil.DeploymentLister,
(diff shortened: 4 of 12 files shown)


pkg/operation/seed/seed.go
```diff
@@ -627,9 +627,11 @@ func BootstrapCluster(k8sGardenClient, k8sSeedClient kubernetes.Interface, seed
...
 		"vpa": map[string]interface{}{
 			"enabled": vpaEnabled,
-			"admissionController": map[string]interface{}{
-				"podAnnotations": map[string]interface{}{
-					"checksum/secret-vpa-tls-certs": utils.ComputeSHA256Hex(jsonString),
+			"runtime": map[string]interface{}{
+				"admissionController": map[string]interface{}{
+					"podAnnotations": map[string]interface{}{
+						"checksum/secret-vpa-tls-certs": utils.ComputeSHA256Hex(jsonString),
+					},
```
rfranzke: Wow, nice catch, thank you, Tim!

pkg/apis/core/v1beta1/constants/types_constants.go
```diff
@@ -78,6 +78,15 @@ const (
 	// the kube-state-metrics-seed pod.
 	DeploymentNameKubeStateMetricsSeed = "kube-state-metrics-seed"
 
+	// DeploymentNameVPAAdmissionController is a constant for the name of the VPA admission controller deployment.
+	DeploymentNameVPAAdmissionController = "vpa-admission-controller"
+	// DeploymentNameVPAAdmissionController is a constant for the name of the VPA exporter deployment.
+	DeploymentNameVPAExporter = "vpa-exporter"
+	// DeploymentNameVPAAdmissionController is a constant for the name of the VPA recommender deployment.
+	DeploymentNameVPARecommender = "vpa-recommender"
+	// DeploymentNameVPAAdmissionController is a constant for the name of the VPA updater deployment.
```
rfranzke: Nit: Fix these doc strings, please

pkg/apis/core/v1beta1/constants/utils.go
```diff
@@ -34,3 +34,12 @@ func getDeprecatedAnnotation(annotations map[string]string, annotationKey, depre
 
 	return val, ok
 }
+
+// GetShootVPADeploymentNames returns the names of all VPA related deployments related to shoot clusters.
+func GetShootVPADeploymentNames() []string {
```
rfranzke, line 39: Hm, why is this a function? Can't it be a statically defined `[]string` var?

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     app: vpa-admission-controller
-{{ toYaml .Values.labels | indent 4 }}
+{{ toYaml .Values.deploymentLabels | indent 4 }}
```
rfranzke, line 10: Shouldn't these labels get merged together with `.Values.labels`? If I see it correctly, the deployment will now only be labelled with the `role=controlplane` labels while it had earlier

```yaml
app: vpa-admission-controller
garden.sapcloud.io/role: vpa
```

pkg/operation/botanist/health_check.go
```diff
@@ -265,6 +265,12 @@ func computeRequiredControlPlaneDeployments(
 		}
 	}
 
+	if gardencorev1beta1helper.ShootWantsVerticalPodAutoscaler(shoot) {
```
rfranzke, line 270: Can you add a small unit test to reflect this branch's logic?

rfranzke (changes requested):

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
timuthy: Good point. My thought of replacing it was the following:

Previously:
```
app: vpa-admission-controller
garden.sapcloud.io/role: vpa
```

With this PR:
```
app: vpa-admission-controller
gardener.cloud/role: seed
```

If we merged the label then we would have this result which is somehow confusing because `garden.sapcloud.io/role` is the old label key with a different value as the one for `gardener.cloud/role`.
```
app: vpa-admission-controller
garden.sapcloud.io/role: vpa
gardener.cloud/role: seed
```

I thought, replacing the label is not really critical because we haven't used any selectors for `garden.sapcloud.io/role: vpa` so far. What do you think?

rfranzke: Sure, I don't mind changing the labels if we are sure that we don't break other functionality that depend on these labels to be present. I remember we had a label selector for cleaning up the VPA earlier when it was backed by a feature gate, but I don't find this code anymore, so probably you're right and it's fine. Thank you.

pkg/apis/core/v1beta1/constants/utils.go
timuthy: > Can't it be a statically defined []string var?

Yes, it can. OTOH, it is supposed to be constant and a public `[]string` var can be mutated from any place which isn't possible if you use a getter `func`. I'm not strongly voting against the var but still wanted to let you know my thought. WDYT?

rfranzke: Hm, yes, that's a valid argument, of course. Fine then, thanks.

timuthy:
Thanks for your review @rfranzke. PTAL.

rfranzke (dismissed):

/lgtm

pkg/operation/botanist/health_check.go
ialidzhikov, line 270: This will actually cause failing healthcheck for `< 24h` for Shoots with `.spec.maintenance.confineSpecUpdateRollout=true` that now set `.spec.kubernetes.verticalPodAutoscaler.enabled=true`, right?

rfranzke: Nice catch! ❤️  Indeed, we should probably check here whether there is a `Deployment` for e.g. the `vpa-admission-controller` at all before adding the component names to the required names (yes, this makes the "deployment exists" check somewhat obsolete).

timuthy: I think we have the same problem for the cluster autoscaler, haven't we?

ialidzhikov: Yep, you are right. The same applies also for the cluster-autoscaler deployment health check.

vpnachev: This health check could have been executed only if `generation == observedGeneration`, but with #2233 the `generation` is not bumped when the spec is changed, thus this comparison cannot be used.

The other option is to extend the shoot status with a field that reflects that actual deployment status of the VPA.

rfranzke: Would it be an option to leverage the shoot spec of the `Cluster` resource to determine whether to perform these checks or not?

vpnachev: Very good idea.

timuthy: Yep, nice idea 👍

timuthy: After having another look at the code, I've spotted more areas where we need to rely on the cluster resource. I tend to open a separate and dedicated PR if you don't mind?

ialidzhikov: Okay, I think I am fine as it is currently. Thank you for checking it.

timuthy:
The health checks were align with the confine spec rollout feature in e2f1e84, at least for the VPA and cluster autoscaler checks. I'll open an issue in order to check and adjust the decision making for other components as well, e.g. Alertmanager.

pkg/operation/botanist/health_check.go
```diff
@@ -489,7 +492,23 @@ func (b *Botanist) checkControlPlane(
...
 	extensionConditions []ExtensionCondition,
 ) (*gardencorev1beta1.Condition, error) {
-	if exitCondition, err := checker.CheckControlPlane(b.Shoot.Info, b.Shoot.SeedNamespace, condition, seedDeploymentLister, seedEtcdLister, seedWorkerLister); err != nil || exitCondition != nil {
+	cluster, err := gardenerextensions.GetCluster(ctx, b.K8sSeedClient.Client(), b.Shoot.SeedNamespace)
+	if err != nil {
+		msg := "failed to start control plane health check"
+		if apierrors.IsNotFound(err) {
+			msg = "control plane is not yet ready"
+		}
+		b.Logger.Errorf(msg+": %v", err)
```
```suggestion
		b.Logger.Errorf("%s: %v", msg, err)
```

rfranzke (dismissed):
/lgtm

rfranzke (approved):

pkg/apis/core/v1beta1/constants/utils.go
```diff
@@ -34,3 +34,12 @@ func getDeprecatedAnnotation(annotations map[string]string, annotationKey, depre
...
+	return []string{
+		DeploymentNameVPAAdmissionController,
+		DeploymentNameVPARecommender,
+		DeploymentNameVPAUpdater,
+	}
```
vpnachev, line 44: Why `DeploymentNameVPAExporter` is missing?

vpnachev (commented):
Overall lgtm, I have just one question inline.

pkg/apis/core/v1beta1/constants/utils.go
rfranzke: It's not deployed for the shoot VPA, only when the VPA is running in the `garden` namespace (on purpose), cc @wyb1

ialidzhikov (approved):
/lgtm

PR #2700, opened 2020-08-10
Don't execute shoot health checks simultaneously

### LINKED ISSUE #2689: shoot care health checks sometimes broken

/area ops-productivity auto-scaling
/kind enhancement
/priority normal
/exp expert
/platform all
/topology seed

I don't remember if it only happened when the `CachedRuntimeClient` feature gate was enabled, but I observed that the shoot care health checks break when a v1.8 gardenlet first starts up. See this:

shoot-care-context-deadline-exceeded

After a minute or so it self-heals again, i.e., the condition statuses can be correctly computed and are also reported correctly. However, we should avoid this flickering to prevent alerts and end-user confusion.

See also the following logs of gardenlet:

```
Could not initialize Shoot client for health check: error creating new ClientSet for key "garden-foo/bar": failed to get Shoot object "garden-foo/bar": rate: Wait(n=1) would exceed context deadline
Could not initialize Shoot client for health check: error creating new ClientSet for key "garden-baz/foo": failed to get Shoot object "garden-baz/foo": context deadline exceeded
```

/cc @timebertt 

**Why is this needed**:
Reduce ops effort and improve overall quality/robustness, smoother roll outs.

timebertt:

/area quality
/kind enhancement
/priority normal

This PR adds some random delay when requeueing existing objects in the Shoot care controller and BackupBucket controller to avoid reconciling/checking all of the existing objects at the same time after startup, which can lead to the gardenlet hitting client side rate limits pretty fast.

Fixes #2689 

**Release note**:

```improvement operator
The gardenlet now spreads Shoot health checks to avoid running into rate limits directly after startup.
```

### DIFF
--- a/pkg/gardenlet/controller/backupentry/backup_entry_controller.go
+++ b/pkg/gardenlet/controller/backupentry/backup_entry_controller.go
@@ -16,9 +16,11 @@ package backupentry
 
 import (
 	"fmt"
+	"time"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	"github.com/gardener/gardener/pkg/logger"
+	"github.com/gardener/gardener/pkg/utils"
 
 	"k8s.io/client-go/tools/cache"
 )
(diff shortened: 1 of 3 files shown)


rfranzke (approved):
/lgtm

timuthy (approved):
Nice!

PR #2704, opened 2020-08-11
Improve error message when zones for region without zones are configured

rfranzke:

/area usability
/kind enhancement
/priority normal

This PR improves the error message when a zone is configured for a region that does not support availability zones. Instead of telling the user that the region as a whole is not allowed he will be pointed to the fact that the region does not support availability zones and that he should not configure them.

/cc @antheajung

**Release note**:

```improvement operator
NONE
```


timuthy (approved):
/lgtm

PR #2709, opened 2020-08-11
Fix kube-apiserver certificate when DNS is disabled

stoyanr:
/area quality
/kind bug
/priority normal

When DNS is disabled, the `kube-apiserver` certificate should contain the LoadBalancer address of the `kube-apiserver` service. This is broken with the current flow, I was able to reproduce it a few times in our OpenShift test cluster. 

Fixes #

The fix introduces a dependency from the `loadSecrets` task to the `waitUntilKubeAPIServerServiceIsReady` in the reconciliation flow when DNS is disabled (or rather, moves that dependency from `generateSecrets` to `loadSecrets`). I tested the fix in our OpenShift test cluster and I wasn't able to reproduce the issue with it.

**Release note**:
```improvement operator
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -131,20 +131,20 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 			Dependencies: flow.NewTaskIDs(deployKubeAPIServerService),
 		})
 		loadSecrets = g.Add(flow.Task{
-			Name:         "Loading existing secrets into ShootState",
-			Fn:           flow.TaskFn(botanist.LoadExistingSecretsIntoShootState),
-			Dependencies: flow.NewTaskIDs(deployNamespace, ensureShootStateExists),
-		})
-		generateSecrets = g.Add(flow.Task{
-			Name: "Generating secrets and saving them into ShootState",
-			Fn:   flow.TaskFn(botanist.GenerateAndSaveSecrets),
+			Name: "Loading existing secrets into ShootState",
+			Fn:   flow.TaskFn(botanist.LoadExistingSecretsIntoShootState),
 			Dependencies: func() flow.TaskIDs {
-				taskIDs := flow.NewTaskIDs(deployNamespace, loadSecrets, ensureShootStateExists)
+				taskIDs := flow.NewTaskIDs(deployNamespace, ensureShootStateExists)
 				if !dnsEnabled && !o.Shoot.HibernationEnabled {
 					taskIDs.Insert(waitUntilKubeAPIServerServiceIsReady)
 				}
 				return taskIDs
 			}(),
+		})
+		generateSecrets = g.Add(flow.Task{
+			Name:         "Generating secrets and saving them into ShootState",
+			Fn:           flow.TaskFn(botanist.GenerateAndSaveSecrets),
+			Dependencies: flow.NewTaskIDs(deployNamespace, loadSecrets, ensureShootStateExists),
 		})
 		deploySecrets = g.Add(flow.Task{
 			Name:         "Deploying Shoot certificates / keys",


rfranzke:
Can you explain how this PR is related to https://github.com/gardener/gardener/pull/2643? Did I miss a case in #2643?

stoyanr:
@rfranzke I think the explanation is basically the same that you wrote in that PR. The `LoadExistingSecretsIntoShootState` actually generates the secrets (the call to the `secretsManager` `Load` method actually calls the secrets generator). Regarding how exactly this leads to the issue, @plkokanov could provide more information. What I am sure is that I reproduced the issue a few times (by deleting the secret from both the shoot namespace and the shoot state, and then regenerating), and with the proposed fix everything worked fine.

rfranzke (commented):
OK, I've checked this a little bit, and I've to admit that I'm somewhat confused. 
My high-level understanding of these three tasks is:

* `loadSecrets` reads existing secrets in the shoot namespace in the seed and puts them into the `ShootState` resource
* `generateSecrets` reads the `ShootState` resource, generates the missing secrets and stores them into the `ShootState` resource as well
* `deploySecrets` reads the `ShootState` resource and creates the `Secret` objects in the shoot namespace in the seed

IIUC, the `loadSecrets` step was only needed for migration purposes and can be removed again (???)

@stoyanr I don't see the `LoadExistingSecretsIntoShootState` function generating new secrets(/certificates) anywhere - do I miss something? I wonder how this change can fix the issue and why #2643 was not enough to fix the problem entirely.

> The LoadExistingSecretsIntoShootState actually generates the secrets (the call to the secretsManager Load method actually calls the secrets generator).

Where do you see this? 

/assign @plkokanov @swilen-iwanow 
/under-investigation

I'm generally OK with this change if it is necessary, but I want to understand what I'm missing/why this PR helps fixing this problem.

plkokanov:
The loadSecrets step will call [wanted_secrets.go:110](https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/wanted_secrets.go#L110)

I've been trying to see whether this somehow affects the DNSNames for the kube apiserver secret config, but have yet to find out how.

rfranzke:
/status author-action
/hold

@stoyanr We have checked the code some more and can't find the place where the `loadSecrets` task would generate the server certificate for the KAPI of the shoot.
Can you share more about
* how you produce this issue with the current `master`
* why you think that these changes fix the problem?

timebertt (dismissed):
/lgtm

stoyanr:
@rfranzke I was able to reproduce the issue several times on Monday but not today. It's not reproducible every time, since it's basically a race condition between `waitUntilKubeAPIServerServiceIsReady` and `loadSecrets`.

How to reproduce:
* Stop gardenlet
* Delete the kube-apiserver secret from shoot state and from the shoot namespace
* Start gardenlet again, get the certificate from the secret, and decode it with `openssl`

plkokanov:
Just a note, I edited the reconcile flow to always execute the loadSecrets step before waiting for the api service loadbalancer to be ready (just for testing purposes). Here are the entries in the api-server cert for subject alternative names
```
DNS:kube-apiserver, DNS:kube-apiserver.shoot--mig--migration1, DNS:kube-apiserver.shoot--mig--migration1.svc, DNS:api.migration1.mig.internal.dev.k8s.ondemand.com, DNS:kubernetes, DNS:kubernetes.default, DNS:kubernetes.default.svc, DNS:kubernetes.default.svc.cluster.local, DNS:ab33251787fa1404e803de99c4f523c8-1366641310.eu-west-1.elb.amazonaws.com, IP Address:127.0.0.1, IP Address:100.64.0.1
```
I will run it a couple more times and move the loadSecrets step further back (before the steps that creates the loadbalancer).

stoyanr:
@plkokanov I did the same and still wasn't able to reproduce it ... Which means that even if there is an issue, this PR isn't likely to fix anything. 

I still can't explain what I was able to observe on Monday though. I am quite sure I did the above steps while using the `v1.8.0` gardenlet at this time. I could have made a mistake during the test, or perhaps the actual cause of the issue is a different one.

If you don't mind, let's keep the PR open for a while (I will convert it to draft), in case the issue is reproduced once again. If we don't see it within the next 1-2 weeks, I will close the PR.

rfranzke:
If there are no objections and updates until tomorrow EOB (CEST) I would go ahead closing this PR.

PR #2696, opened 2020-08-10
Add VPA definitions for VPA components

rfranzke:
**How to categorize this PR?**

/area auto-scaling
/kind enhancement
/priority normal
/exp intermediate
/topology seed
/platform all

**What this PR does / why we need it**:
This PR adds missing VPA definitions for the VPA components themselves so that they get auto-scaled as well.

**Which issue(s) this PR fixes**:
Fixes #2684

**Special notes for your reviewer**:
I removed the `enableVPA` chart value as it was always set to `true`. Also, I reduced the requests/limits according to observations on larger landscapes.
/invite @wyb1 @amshuman-kr @ggaurav10 

**Release note**:

```improvement operator
The VPA components in the seed are now vertically auto-scaled as well.
```

Issue #1825, opened 2020-01-20
Use gardener-resource-manager to deploy logging/monitoring stack and seed bootstrapping components


rfranzke:
**What would you like to be added**:
We could use the gardener-resource-manager in the `garden` namespace of seeds to deploy the logging/monitoring stacks for shoots. Currently, we have several `Delete*` functions that are enlisting the resources of each chart which makes the whole monitoring and logging stack code hard to maintain and error-prone.

**Why is this needed**:
This will reduce a lot of code in our end, especially the `Delete*` functions for AlertManager, the whole logging or monitoring stack as the GRM would take over the job.

rfranzke:
/remove lifecycle/rotten
@tim-ebert will also work on using the GRM for the seed bootstrapping.

vpnachev:
cc @Kristian-ZH @vlvasilev

rfranzke:
Let's
/close
this one in favor of #2754.

Issue #1633, opened 2019-11-19
Seed deletion leaves workload in garden namespace


timebertt:
**What happened**:
If you delete a Seed object, the `garden` namespace and some workload in it will not be removed from the cluster.

After Seed deletion:
```
$ k -n garden get pod
NAME                                         READY   STATUS              RESTARTS   AGE
aggregate-prometheus-0                       0/2     ContainerCreating   0          1s
daily-curator-1574143500-q4ffd               0/1     Completed           0          95m
elasticsearch-logging-0                      2/2     Running             0          18h
fluent-bit-j5wkl                             1/1     Running             0          18h
fluent-bit-xsflg                             1/1     Running             0          18h
fluentd-es-0                                 1/1     Running             2          18h
gardener-resource-manager-847485d776-8rffj   1/1     Running             0          10s
grafana-6cdcc5cf68-nnmt9                     1/1     Running             0          18h
hourly-curator-1574146800-ht4jl              0/1     Completed           0          40m
hvpa-controller-6c5fccf85-v8pkb              1/1     Running             0          18h
kibana-logging-74f4bf4474-8n4t5              2/2     Running             0          18h
prometheus-0                                 0/2     ContainerCreating   0          1s
reserve-excess-capacity-5cf46749-488cg       1/1     Running             0          18h
reserve-excess-capacity-5cf46749-75f72       1/1     Running             0          18h
reserve-excess-capacity-5cf46749-cnndr       1/1     Running             0          18h
reserve-excess-capacity-5cf46749-gqznh       1/1     Running             0          18h
reserve-excess-capacity-5cf46749-hqbwl       1/1     Running             0          18h
reserve-excess-capacity-5cf46749-khr92       1/1     Running             0          18h
reserve-excess-capacity-5cf46749-lm8l7       1/1     Running             0          18h
reserve-excess-capacity-5cf46749-twr2f       1/1     Running             0          18h
vpa-exporter-5c9cfddf47-wvpq8                1/1     Running             0          10s
vpa-recommender-7f9fbbfdbb-pb4t9             1/1     Running             0          18h
vpa-updater-84d885687-dstz8                  1/1     Running             0          18h
```

**What you expected to happen**:
Gardener should clean up all the added resources after the Seed was deregistered.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- Gardener version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
[...]

PR #2715, opened 2020-08-11
Fix VPA deletion in both seed and shoot

rfranzke:

/area quality
/kind bug
/priority normal

This PR fixes two bugs:
1. Deletion of VPA in seed did not clean up the newly introduced `VerticalPodAutoscaler` objects (#2696)
1. Deletion of VPA in shoot did clean up the `CustomResourceDefintion`s (which is especially bad if the shoot is used as seed cluster) - this was desired behaviour but comes with the downside that the (potentially many) `VerticalPodAutoscaler` objects get lost (FYI: I'm adding the new `resources.gardener.cloud/keep-object` annotation also for the CRDs in the seed even if they are not managed by the GRM yet (though, they will be managed by it in future, @timebertt is already working on it as part of #1633 and #1825)).

**Release note**:

```improvement user
When the VPA for shoots is disabled then the `CustomResourceDefinition`s are no longer deleted (they will remain in the system, together with all the `VerticalPodAutoscaler` objects - if you don't need them anymore you can remove them with `kubectl delete crd <crd-names>`).
```
``` noteworthy developer github.com/gardener/gardener-resource-manager #73 @rfranzke
The new `resources.gardener.cloud/keep-object` annotation can be used on resources managed by `ManagedResource` objects in order to keep them in the system in case they get removed from the `ManagedResource` or the `ManagedResource` itself is being deleted.
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -24,7 +24,7 @@ images:
 - name: gardener-resource-manager
   sourceRepository: github.com/gardener/gardener-resource-manager
   repository: eu.gcr.io/gardener-project/gardener/gardener-resource-manager
-  tag: "v0.16.0"
+  tag: "v0.17.0"
 - name: dependency-watchdog
   sourceRepository: github.com/gardener/dependency-watchdog
   repository: eu.gcr.io/gardener-project/gardener/dependency-watchdog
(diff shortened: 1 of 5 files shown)


ialidzhikov (approved):
/lgtm

PR #2718, opened 2020-08-12
Revert labels for seed VPA deployment

timuthy:

/area control-plane
/kind cleanup
/priority normal

Parts of #2698 changed the role label for seed VPA deployments from `garden.sapcloud.io/role: vpa` -> `gardener.cloud/role: seed`. However, we don't use the `seed` value for seed deployments so far and thus only having it for VPA makes this handling even more inconsistent. Additionally, the labels of the deployment diverged unnecessarily from the ones of the pod(s).

This PR reverts to the previously used label.

An improvement for the `ControlPlaneHealthy` condition has been implemented along the way.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/health_check.go
+++ b/pkg/operation/botanist/health_check.go
@@ -507,18 +507,18 @@ func (b *Botanist) checkControlPlane(
 ) (*gardencorev1beta1.Condition, error) {
 	cluster, err := gardenerextensions.GetCluster(ctx, b.K8sSeedClient.Client(), b.Shoot.SeedNamespace)
 	if err != nil {
-		msg := "failed to start control plane health check"
 		if apierrors.IsNotFound(err) {
-			msg = "control plane is not yet ready"
+			c := checker.FailedCondition(condition, "ControlPlaneNotReady", "Control plane is not yet ready because of missing cluster resource.")
+			return &c, nil
 		}
-		b.Logger.Errorf("%s: %v", msg, err)
-		return nil, errors.New(msg)
+		b.Logger.Errorf("Failed to execute control plane health checks: %v", err)
+		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


pkg/operation/botanist/health_check.go
```diff
@@ -507,18 +507,18 @@ func (b *Botanist) checkControlPlane(
...
-		msg := "failed to start control plane health check"
 		if apierrors.IsNotFound(err) {
-			msg = "control plane is not yet ready"
+			c := checker.FailedCondition(condition, "ControlPlaneNotReady", "Control plane is not yet ready because of missing cluster resource.")
+			return &c, nil
 		}
-		b.Logger.Errorf("%s: %v", msg, err)
-		return nil, errors.New(msg)
+		b.Logger.Errorf("Failed to execute control plane health checks: %v", err)
+		return nil, err
```
rfranzke, line 515: Shouldn't we also report a failed condition here? Now, the error is only visible in the logs, or? same below

rfranzke (commented):
looks good, just one question

pkg/operation/botanist/health_check.go
timuthy: Good question! 
My thought was that this is rather an "unspecified" runtime error. The error message will be propagated and reported as `UNKOWN` in the `ControlPlaneHealthy` condition (see [here](https://github.com/gardener/gardener/blob/e34a58af4a6d4edbfb0ec017901559405ee1a7d7/pkg/operation/botanist/health_check.go#L737)). Before this PR, the same was done in case we didn't find the cluster resource which wasn't nice because at least for new shoots, we expect that the cluster resource is missing.

rfranzke: OK, sounds good, thanks

rfranzke (approved):
/lgtm

PR #2720, opened 2020-08-12
Dedicated network policy for kube-apiserver-to-vpa communication

rfranzke:

/area networking
/kind enhancement
/priority normal

This PR adds a dedicated `NetworkPolicy` for the communication from the kube-apiserver to the VPA admission controller.
Earlier, this communication was allowed by the `allow-kube-apiserver` policy which is reconciled as part of the `DeployKubeAPIServer` step. However, as the `DeployVerticalPodAutoscaler` step only comes at the very end of the flow it's way better to let it manage the necessity resources altogether (this is not yet strictly pervaded, though, this is where we should get to).

/invite @timuthy 
thanks for noticing!

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/networkpolicy.yaml
+++ b/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/networkpolicy.yaml
@@ -1,0 +1,28 @@
+{{- if eq .Values.clusterType "shoot" }}
+apiVersion: {{ include "networkpolicyversion" . }}
+kind: NetworkPolicy
+metadata:
+  annotations:
+    gardener.cloud/description: |
+      Allows Egress from pods shoot's kube-apiserver to talk to the
+      VPA admission controller.
+  name: allow-kube-apiserver-to-vpa-admission-controller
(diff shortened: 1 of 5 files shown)


timuthy (approved):
Thanks for taking care!!!

timuthy:
I just noticed that this make the annotation [networkpolicy/vpa-enabled](https://github.com/gardener/gardener/blob/a2765d2aa5423b5c6e5ab13b42342ad3078e74af/pkg/operation/botanist/controlplane.go#L928) obsolete since the KAS pods are restarted before the policy is created.
We are also missing the deletion of the policy if VPA is disabled, right?

timuthy (changes requested):
See comment above.

rfranzke:
Thanks @timuthy, right, I have removed the annotation (it seems to not be necessary anyways as the kube-apiserver always creates a new connection to the VPA admission controller). I've also added the `NetworkPolicy` to the list of resources that must be deleted when VPA gets destroyed.

timuthy (approved):
/lgtm

PR #2723, opened 2020-08-12
Increase resource requests for fluent-bit

vpnachev:

/area logging
/kind enhancement
/priority normal

Increase resource requests for fluent-bit

On large landscapes with large seed nodes the fluent bit is handling a lot more logs. With this change its requests are closer to the real usage so it can have more guaranteed resources.

Fixes #

/invite @vlvasilev @Kristian-ZH 
/assign @vlvasilev @Kristian-ZH 

**Release note**:

```improvement operator
The resource requests for the `fluent-bit` DaemonSet running in the seed clusters have been increased.
```

### DIFF
--- a/charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-daemonset.yaml
+++ b/charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-daemonset.yaml
@@ -38,8 +38,8 @@ spec:
             cpu: 300m
             memory: 400Mi
           requests:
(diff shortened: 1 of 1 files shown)


Kristian-ZH (dismissed):
On my setup it is even 200m CPU and 200Mi memory but it is also OK

vlvasilev (dismissed):
/lgtm

rfranzke (dismissed):

rfranzke (approved):

PR #2731, opened 2020-08-14
Refactor kube-scheduler deployment

rfranzke:

/kind enhancement
/priority normal
/topology seed
/platform all
/exp intermediate

This PR refactors the deployment for the kube-scheduler and adopts the `DeployWaiter` interface for components. Consequently, it eliminates the Helm chart and adds full unit tests, instead. This improves several aspects and the overall code quality.

Part of #2754

I haven't touched the `kube-scheduler` related system resources (https://github.com/gardener/gardener/tree/master/charts/shoot-core/components/charts/kube-scheduler) but could follow up in a subsequent PR with it.

**Release note**:

```improvement operator
NONE
```


pkg/operation/botanist/controlplane/kube_scheduler.go
```diff
@@ -0,0 +1,472 @@
...
+		deployment.Spec.Selector = &metav1.LabelSelector{MatchLabels: labels}
+		deployment.Spec.Template = corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Annotations: map[string]string{
+					"checksum/configmap-componentconfig":           componentConfigChecksum,
+					"checksum/secret-" + k.secrets.Kubeconfig.Name: k.secrets.Kubeconfig.Checksum,
+					"checksum/secret-" + k.secrets.Server.Name:     k.secrets.Server.Checksum,
+				},
+				Labels: utils.MergeStringMaps(labelsWithControlPlaneRole, map[string]string{
+					v1beta1constants.LabelNetworkPolicyToDNS:            v1beta1constants.LabelNetworkPolicyAllowed,
```
timuthy: Do you also want to add v1beta1constants.LabelPodMaintenanceRestart: "true"? I already wanted to do it in another PR but then I saw this one 😄

rfranzke: Sure, I can do it in a follow-up PR once this one is merged (to not mix a refactoring with a new feature) 👍

timuthy: Okay, thank you :)

ialidzhikov:
To be honest, I am with mixed feelings wrt helm vs go struct approach. Some cons against the go struct approach that I see at first sight:
- I like that currently with helm each manifest is in its own separate file. I usually search for given deployment or other manifest and I can find it fast when I know the naming convention of the files. With the go struct approach all objs are `CreateOrUpdate`'d in single `Deploy` func which makese harder to maintain I think.
- the go struct approach seems too verbose. I see that multuple computations are extracted into smaller funcs (for example `computeCommandLineFlags`) which leads to going up and down in the source - for example need to check utility func at L371, then go back to L130 where this partial spec is stored in variable and then go L199 to check how this variable is used.
- `computeVersionConstraints` func alternative in helm comes for free
- with helm we don't need to vendor `k8s.io/kube-scheduler`
- go struct way does not scale and it is not maintainable for huge charts

Probably you can share what are the pros? Probably easier testing? But it should be possible to achieve that also with helm + chartRenderer and chartApplier, right?

/cc-ing also  @gardener/gardener-maintainers for feedback

rfranzke:
> With the go struct approach all objs are CreateOrUpdate'd in single Deploy func which makese harder to maintain I think.

Just to understand it correctly, your point here is that the deployment is not in its own file, right? If it's only that, I'm open to move the kube-scheduler related things into its own package and have a dedicated file per resource, just like in the Helm chart.

> the go struct approach seems too verbose. 

Yes, there is some boilerplate, agreed.

>  I see that multuple computations are extracted into smaller funcs (for example computeCommandLineFlags) which leads to going up and down in the source - for example need to check utility func at L371, then go back to L130 where this partial spec is stored in variable and then go L199 to check how this variable is used.

I don't think this is a valid argument. You'll find lots of `{{ template }}` in the Helm charts that have more complex logic as well, which are later ` {{ include }}`d at the right place in the actual manifest. That's exactly the same, we just haven't followed this approach extensively everywhere (and, hence, have only a few `{{ if-then-else }}` expressions which make the impression that it's "easier to read").

> computeVersionConstraints func alternative in helm comes for free

Yes, but to the price that errors are not handled properly and might cause side-effects if certain conditions don't hold true. That's exactly what we are trying to improve now (although, this particular error handling of the version comparisons is not the main motivation, of course).

> with helm we don't need to vendor k8s.io/kube-scheduler

Well, so what? We only vendor it for the APIs of the configuration. I don't see any negative consequences doing so. What is your point here?

> go struct way does not scale and it is not maintainable for huge charts

For some things it might make sense to keep them (CRDs, for example), sure. I don't suggest to get rid of the Helm chart/chart applier/etc. code in its entirety.

As discussed in today's meeting, the main motivation for this PR was to improve the code coverage wrt unit tests. Using the already established testing frameworks with Ginkgo/Gomega and Gomock was the most straight-forward way without the necessity of introducing new concepts/complexities.
The current structure with the charts, their (default) values, and the corresponding Go implementations is too fragile and error-prone. Too many parts are distributed, there is no consistent data flow, and both cause a high risk of breaking something as we don't have the tests.

I'm not completely against the Helm charts and the chart applier. An alternative approach would be to explicitly provide the important values in the Go code for rendering the chart (like deployment name, labels, container name, ports, etc.) and don't "hard-code" them in the chart itself. Then we can have unit tests to check if the correct Helm values are computed by Go, but we'll also need other unit tests for the Helm chart itself to check if the values are used in the correct way (e.g., that the container name lands in `.spec.template.spec.containers[].name`), and so on. What will remain with this approach is the distribution of the code (you won't have one place in the tree where everything related to kube-scheduler is, you'll still have the chart and the Go code). How would the chart then look like in the end? It won't be that "readable" as it's toady, that's for sure. In that shape, is it still valuable enough to justify the complexity?
Also, we would need tests for the various `if-then-else` constructs in the Helm charts. And how would such a test look like? Rendering it and then comparing the result against a pre-defined YAML file with the expected output? Sure, can be done, but to be honest, isn't this more complicated and doesn't solve our problems today? I think, already today have charts where we have differences for the various Kubernetes versions, so we have `if-then-else` constructs in the chart, but then we also have them in the Go code to provide the correct values. This is totally unnecessary and hard to read/understand.

Do you have other ideas/proposals? I'm not insisting on this approach, there might be better ways, but I like that it solves the problems of testability, code distribution, and easier data flow (this part is not yet consistent, of course, but will improve over time). As @timuthy suggested, we can put the whole kube-scheduler related code into its own package, this will probably help a little bit more. However, I'm open to better approaches and very curious of what @timebertt will present end of the week.

mvladev:
I think that this PR is setting things in the right direction. I would like to have a separate controller for each component (`kube-scheduler`, `kube-apiserver` and so) and this is an intermediate solution until then.

It's a little more verbose, but it can be made many times more testable than the current solution. The boilerplate code could be refactored later on as common code is extracted into packages / structs.

pkg/operation/botanist/controlplane.go
```diff
@@ -1302,43 +1304,31 @@ func (b *Botanist) DeployKubeControllerManager(ctx context.Context) error {
...
-		if schedulerConfig.KubeMaxPDVols != nil {
-			defaultValues["kubeMaxPDVols"] = *schedulerConfig.KubeMaxPDVols
-		}
-	}
+	return kubescheduler.New(
+		b.K8sSeedClient.DirectClient(),
+		b.Shoot.SeedNamespace,
+		b.Shoot.Info.Spec.Kubernetes.Version,
+		image.String(),
+		int32(b.Shoot.GetReplicas(1)),
```
timebertt: Can you change the return type of `GetReplicas` to `int32`?
```suggestion
		b.Shoot.GetReplicas(1),
```

pkg/operation/botanist/controlplane.go
```diff
@@ -1302,43 +1304,31 @@ func (b *Botanist) DeployKubeControllerManager(ctx context.Context) error {
...
 
-	values, err := b.InjectSeedShootImages(defaultValues, common.KubeSchedulerImageName)
-	if err != nil {
-		return err
-	}
+// DeployKubeScheduler deploys the Kubernetes scehduler.
+func (b *Botanist) DeployKubeScheduler(ctx context.Context) error {
+	b.Shoot.Components.ControlPlane.KubeScheduler.SetSecrets(kubescheduler.Secrets{
+		Kubeconfig: kubescheduler.Secret{Name: "kube-scheduler", Checksum: b.CheckSums[v1beta1constants.DeploymentNameKubeScheduler]},
+		Server:     kubescheduler.Secret{Name: "kube-scheduler-server", Checksum: b.CheckSums[common.KubeSchedulerServerName]},
```
timebertt: Can we have a separate constant for the name of the kubeconfig secret to avoid confusion/mistakes?
Maybe it makes sense to have those constants in the kubescheduler package or should we have them in the common package?

Also, can you please pass the respective constants to the `Name` field, as it will be used as the `SecretRef` in the Volume?
```suggestion
		Kubeconfig: kubescheduler.Secret{Name: v1beta1constants.SecretNameKubeScheduler Checksum: b.CheckSums[v1beta1constants.SecretNameKubeScheduler]},
		Server:     kubescheduler.Secret{Name: common.KubeSchedulerServerName, Checksum: b.CheckSums[common.KubeSchedulerServerName]},
```

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -0,0 +1,473 @@
...
+)
+
+const (
+	// LabelRole is a constant for the value of a label with key 'role' whose value is 'scheduler'.
+	LabelRole = "scheduler"
+	// DataKeyComponentConfig is a constant for the key of the data map in a ConfigMap whose value is the
+	// component configuration of the kube-scheduler.
+	DataKeyComponentConfig = "config.yaml"
+	// PortNameMetrics is a constant for the name of the metrics port of the kube-scheduler.
+	PortNameMetrics = "metrics"
```
timebertt, line 55: Just a question: Am I assuming correctly that this constant should be used by the prometheus component later on to compute the scrape config for the kube scheduler? Is this the concept for how we want to improve the data flow between components?

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -0,0 +1,473 @@
...
+	componentConfigYAML, err := NewConfigCodec().Encode(componentConfig)
+	if err != nil {
+		return "", "", err
+	}
+	componentConfigYAML = strings.Replace(componentConfigYAML, kubeschedulerv1alpha1.SchemeGroupVersion.String(), apiVersion, -1)
+
+	return componentConfigYAML, utils.ComputeChecksum(componentConfig), nil
+}
+
+func (k *kubeScheduler) computeCommandLineFlags(port int32) []string {
```
timebertt: Nit, this doesn't only compute flags, but rather the whole command 😉 
```suggestion
func (k *kubeScheduler) computeCommand(port int32) []string {
```

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -0,0 +1,473 @@
...
+								Name:      configMap.Name,
+								MountPath: volumeMountPathConfig,
+							},
+						},
+					},
+				},
+				DNSPolicy:                     corev1.DNSClusterFirst,
+				RestartPolicy:                 corev1.RestartPolicyAlways,
+				SchedulerName:                 corev1.DefaultSchedulerName,
+				TerminationGracePeriodSeconds: pointer.Int64Ptr(30),
```
timebertt: Can we remove those defaults?

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -0,0 +1,473 @@
...
+		apiVersion = "componentconfig/v1alpha1"
+	}
+
+	componentConfig := &kubeschedulerv1alpha1.KubeSchedulerConfiguration{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: apiVersion,
+			Kind:       "KubeSchedulerConfiguration",
+		},
+		ClientConnection: componentbaseconfigv1alpha1.ClientConnectionConfiguration{
+			Kubeconfig: volumeMountPathKubeconfig + "/" + secrets.DataKeyKubeconfig,
```
timebertt: Just a suggestion :)
Can be applied to all the other paths (in the command line flags) too, if you want.
```suggestion
			Kubeconfig: filepath.Join(volumeMountPathKubeconfig, secrets.DataKeyKubeconfig),
```

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -0,0 +1,473 @@
...
+	if k.versionConstraints.k8sEqual110 {
+		command = append(command, kutil.FeatureGatesToCommandLineParameter(map[string]bool{"PodPriority": true}))
+	}
+
+	command = append(command, "--v=2")
+	return command
+}
+
+func (k *kubeScheduler) computeVersionConstraints() (*kubeSchedulerVersionConstraints, error) {
+	k8sEqual110, err := versionutils.CompareVersions(k.version, "~", "1.10")
```
timebertt: I'm thinking about how this computation and the `kubeSchedulerVersionConstraints` construct can be simplified to reduce the amount of explicit error handling. Also this will parse the Shoot version for every constraint and each constraint for every shoot being reconciled...

WDYT about:
- "pre-computing" the semver constraints in an `init` func (panicking on error, as this is a programmer level error, maybe with `utilruntime.Must`)
- and parsing the shoots version to a `*semver.Version` when constructing the `Shoot` operation object?

Constraints can then be checked by just calling the `Check` func (which doesn't return an error).
This would slim down this component and the `New` func also wouldn't need to return an error. (Also we would have implicit test coverage for the constraints 😉 )

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
```diff
@@ -0,0 +1,472 @@
...
+	apiVersion, checksum := "componentconfig/v1alpha1", "3370189444db371972ebc8add5b7c76944c9f2912735ed59b7660427b52373c6"
+
+	if k8sVersionGreaterEqual118, _ := versionutils.CompareVersions(version, ">=", "1.18"); k8sVersionGreaterEqual118 {
+		apiVersion, checksum = "kubescheduler.config.k8s.io/v1alpha2", "54b447618b8ce997b6e0d537be011ce18ff5046172980bf5efb7df7ceb6e97a9"
+	} else if k8sVersionGreaterEqual112, _ := versionutils.CompareVersions(version, ">=", "1.12"); k8sVersionGreaterEqual112 {
+		apiVersion, checksum = "kubescheduler.config.k8s.io/v1alpha1", "5995cdb0cc43de90d58325b7f11962c5b5f5d284bfbd3ee8736e0e97be8fbb0f"
+	}
+
+	return "algorithmSource: {}\napiVersion: " + apiVersion + "\nbindTimeoutSeconds: null\nclientConnection:\n  acceptContentTypes: \"\"\n  burst: 0\n  contentType: \"\"\n  kubeconfig: /var/lib/kube-scheduler/kubeconfig\n  qps: 0\nkind: KubeSchedulerConfiguration\nleaderElection:\n  leaderElect: true\n  leaseDuration: 0s\n  lockObjectName: \"\"\n  lockObjectNamespace: \"\"\n  renewDeadline: 0s\n  resourceLock: \"\"\n  resourceName: \"\"\n  resourceNamespace: \"\"\n  retryPeriod: 0s\npodInitialBackoffSeconds: null\npodMaxBackoffSeconds: null\n", checksum
```
timebertt: Can you put that into a multi-line string to make it easier to read?

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
```diff
@@ -0,0 +1,472 @@
...
+}
+
+func portForKubernetesVersion(version string) int32 {
+	if k8sVersionGreateEqual113, _ := versionutils.CompareVersions(version, ">=", "1.13"); k8sVersionGreateEqual113 {
```
timebertt:
```suggestion
	if k8sVersionGreaterEqual113, _ := versionutils.CompareVersions(version, ">=", "1.13"); k8sVersionGreaterEqual113 {
```

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
```diff
@@ -0,0 +1,472 @@
...
+									Ports: []corev1.ContainerPort{
+										{
+											Name:          "metrics",
+											ContainerPort: portForKubernetesVersion(version),
+											Protocol:      corev1.ProtocolTCP,
+										},
+									},
+									Env:                      env,
+									TerminationMessagePath:   "/dev/termination-log",
+									TerminationMessagePolicy: corev1.TerminationMessageReadFile,
```
timebertt: Any specific reason, why you use a literal for `TerminationMessagePath` and a constant for `TerminationMessagePolicy`?

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
```diff
@@ -0,0 +1,472 @@
...
+										{
+											Name:      secretNameKubeconfig,
+											MountPath: "/var/lib/kube-scheduler",
+										},
+										{
+											Name:      secretNameServer,
+											MountPath: "/var/lib/kube-scheduler-server",
+										},
+										{
+											Name:      "kube-scheduler-config",
```
timebertt:
```suggestion
											Name:      configMapName,
```

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
```diff
@@ -0,0 +1,472 @@
...
+											SecretName: secretNameServer,
+										},
+									},
+								},
+								{
+									Name: "kube-scheduler-config",
+									VolumeSource: corev1.VolumeSource{
+										ConfigMap: &corev1.ConfigMapVolumeSource{
+											LocalObjectReference: corev1.LocalObjectReference{
+												Name: "kube-scheduler-config",
```
timebertt:
```suggestion
									Name: configMapName,
									VolumeSource: corev1.VolumeSource{
										ConfigMap: &corev1.ConfigMapVolumeSource{
											LocalObjectReference: corev1.LocalObjectReference{
												Name: configMapName,
```

pkg/operation/botanist/controlplane.go
```diff
@@ -1302,43 +1304,31 @@ func (b *Botanist) DeployKubeControllerManager(ctx context.Context) error {
...
 
-	if schedulerConfig := b.Shoot.Info.Spec.Kubernetes.KubeScheduler; schedulerConfig != nil {
-		defaultValues["featureGates"] = schedulerConfig.FeatureGates
-
-		if schedulerConfig.KubeMaxPDVols != nil {
-			defaultValues["kubeMaxPDVols"] = *schedulerConfig.KubeMaxPDVols
-		}
-	}
+	return kubescheduler.New(
+		b.K8sSeedClient.DirectClient(),
```
timebertt: I think the cached client should be enough for this case. Have you tested this and encountered any issues with it?
```suggestion
		b.K8sSeedClient.Client(),
```

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
```diff
@@ -0,0 +1,472 @@
...
+					Expect(err).NotTo(HaveOccurred())
+
+					kubeScheduler.SetSecrets(Secrets{
+						Kubeconfig: Secret{Name: secretNameKubeconfig, Checksum: secretChecksumKubeconfig},
+						Server:     Secret{Name: secretNameServer, Checksum: secretChecksumServer},
+					})
+
+					gomock.InOrder(
+						c.EXPECT().Get(ctx, kutil.Key(namespace, configMapName), gomock.AssignableToTypeOf(&corev1.ConfigMap{})),
+						c.EXPECT().Update(ctx, configMapFor(version)),
```
timebertt: Can you use the `DeepDerivativeEqual` to get the enhanced YAML diff?
```suggestion
						c.EXPECT().Update(ctx, gomock.AssignableToTypeOf(&corev1.ConfigMap{})).
							Do(func(ctx context.Context, obj runtime.Object, opts ...client.UpdateOption) {
								Expect(obj).To(DeepDerivativeEqual(configMapFor(version)))
							}),
```

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -0,0 +1,473 @@
...
+			Value: *k.config.KubeMaxPDVols,
+		}}
+	}
+	return nil
+}
+
+func (k *kubeScheduler) computeComponentConfig() (string, string, error) {
+	var apiVersion string
+	if k.versionConstraints.k8sGreaterEqual118 {
+		apiVersion = "kubescheduler.config.k8s.io/v1alpha2"
```
timebertt, line 336: On 1.18 the scheduler crashes with the following error:
```
: v1alpha2.KubeSchedulerConfiguration.ClientConnection: v1alpha1.ClientConnectionConfiguration.BindTimeoutSeconds: TypeMeta: APIVersion: ReadObject: found unknown field: algorithmSource, error found in #10 byte of ...|thmSource":{},"apiVe|..., bigger context ...|{"algorithmSource":{},"apiVersion":"kubescheduler.config.k8s.io/v1al|...
```

The reason seems to be, that `algorithmSource` was removed in [`kubescheduler.config.k8s.io/v1alpha2.KubeSchedulerConfiguration`](https://github.com/kubernetes/kube-scheduler/blob/da23134c2b6a206bfa461e2779b32c7c93d19708/config/v1alpha2/types.go#L40).

Is there a possibility to not marshal the fields, that we don't explicitly set? (I see a lot of fields in the config API don't have `omitempty` tags). Maybe we should go back to a go template for the config 🤔

timebertt (changes requested):
Thanks for the PR! Looks good in general, I somehow like the go struct approach, now that I have spent some time with it in the review :)

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -0,0 +1,473 @@
...
+			Kubeconfig: volumeMountPathKubeconfig + "/" + secrets.DataKeyKubeconfig,
+		},
+		LeaderElection: kubeschedulerv1alpha1.KubeSchedulerLeaderElectionConfiguration{
+			LeaderElectionConfiguration: componentbaseconfigv1alpha1.LeaderElectionConfiguration{
+				LeaderElect: pointer.BoolPtr(true),
+			},
+		},
+	}
+
+	componentConfigYAML, err := NewConfigCodec().Encode(componentConfig)
```
stoyanr: I would inject the codec, instead of creating it here. This would allow for it to be easily mocked away in tests, to be replaced by a different implementation, etc.

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
rfranzke: Yes, I am working on a subsequent PR that improves those points, right.

stoyanr (commented):
Just a small comment regarding the codec creation and usage that @timebertt brought to my attention.

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
rfranzke: Wouldn't this break if you develop/build Gardener on e.g. Windows?

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
rfranzke: I don't use the constants in the tests on purpose to detect potential issues when they get changed.

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
rfranzke: see https://github.com/gardener/gardener/pull/2731#discussion_r476267068

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
rfranzke: Yes, I think it makes sense to go back to a template for the configuration. It is already cumbersome to overwrite those API versions, and it will become more cumbersome when the structure differs between the version (e.g., for the `algorithmSource` field).

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
rfranzke: Thanks, I think we will go back to a template for the configuration (at least for the kube-scheduler), see https://github.com/gardener/gardener/pull/2731#discussion_r475690713.

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
timebertt: Ok perfect, makes sense to me.

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
timebertt: Oh well, I missed that. Just forget my suggestion, then 😄

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
timebertt: That makes sense in general. But, `configMapName` is not a constant from the production code, it's only a variable in your test. I think, there's no problem with reusing that :)

rfranzke: Ah, got it, yep, thanks

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
```diff
@@ -0,0 +1,482 @@
...
+				Data:       map[string]string{"config.yaml": componentConfigYAML},
+			}
+		}
+		vpa = &autoscalingv1beta2.VerticalPodAutoscaler{
+			ObjectMeta: metav1.ObjectMeta{Name: vpaName, Namespace: namespace},
+			Spec: autoscalingv1beta2.VerticalPodAutoscalerSpec{
+				TargetRef: &autoscalingv1.CrossVersionObjectReference{
+					APIVersion: "apps/v1",
+					Kind:       "Deployment",
+					Name:       "kube-scheduler",
```
timebertt: Nit
```suggestion
					Name:       deploymentName,
```

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
```diff
@@ -0,0 +1,482 @@
...
+											SecretName: secretNameServer,
+										},
+									},
+								},
+								{
+									Name: configMapName,
+									VolumeSource: corev1.VolumeSource{
+										ConfigMap: &corev1.ConfigMapVolumeSource{
+											LocalObjectReference: corev1.LocalObjectReference{
+												Name: "kube-scheduler-config",
```
timebertt:
```suggestion
												Name: configMapName,
```

timebertt (commented):
Thanks for incorporating my feedback!
Looks great now, tested it again and works with 1.17 and 1.18, now.

Only two small suggestions to make the test more consistent and easier to adapt in the future, but no strong opinion there.

timebertt (approved):
Thanks! Good to go from my POV 🚀 
/lgtm
/approve

PR #2732, opened 2020-08-16
Prevent deletion of seed cluster which is still in use

vpnachev:

/area robustness
/kind enhancement
/priority normal

Prevent deletion of seed cluster which is still in use by:
1. Admission plugin that disallow the `delete` call to seeds that are still used by shoots or backupbuckets
1. Admission plugin that disallow removal of `use-as-seed` annotation from the shooted seed that is still in use.

Fixes #

**Release note**:

```improvement operator
It is not possible a seed cluster to be deleted, directly or via removal of the `shoot.gardener.cloud/use-as-seed` annotation on the shoot, if the seed is still used by a `BackupBucket` or it is hosting the control plane of a shoot cluster.
```

### DIFF
--- a/cmd/gardener-apiserver/app/gardener_apiserver.go
+++ b/cmd/gardener-apiserver/app/gardener_apiserver.go
@@ -42,6 +42,7 @@ import (
 	"github.com/gardener/gardener/plugin/pkg/global/extensionvalidation"
 	"github.com/gardener/gardener/plugin/pkg/global/resourcereferencemanager"
 	plantvalidator "github.com/gardener/gardener/plugin/pkg/plant"
+	seedvalidator "github.com/gardener/gardener/plugin/pkg/seed/validator"
 	shootdns "github.com/gardener/gardener/plugin/pkg/shoot/dns"
 	"github.com/gardener/gardener/plugin/pkg/shoot/oidc/clusteropenidconnectpreset"
 	"github.com/gardener/gardener/plugin/pkg/shoot/oidc/openidconnectpreset"
@@ -164,6 +165,7 @@ func (o *Options) complete() error {
 	shootquotavalidator.Register(o.Recommended.Admission.Plugins)
 	shootdns.Register(o.Recommended.Admission.Plugins)
 	shootvalidator.Register(o.Recommended.Admission.Plugins)
+	seedvalidator.Register(o.Recommended.Admission.Plugins)
 	controllerregistrationresources.Register(o.Recommended.Admission.Plugins)
 	plantvalidator.Register(o.Recommended.Admission.Plugins)
 	openidconnectpreset.Register(o.Recommended.Admission.Plugins)
@@ -178,6 +180,7 @@ func (o *Options) complete() error {
 		shootdns.PluginName,
 		shootquotavalidator.PluginName,
 		shootvalidator.PluginName,
+		seedvalidator.PluginName,
 		controllerregistrationresources.PluginName,
 		plantvalidator.PluginName,
 		deletionconfirmation.PluginName,
--- a/plugin/pkg/seed/validator/admission.go
+++ b/plugin/pkg/seed/validator/admission.go
@@ -1,0 +1,152 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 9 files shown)


plugin/pkg/utils/miscellaneous.go
```diff
@@ -45,3 +45,26 @@ func GetProject(namespace string, projectLister corelisters.ProjectLister) (*cor
...
+
+// IsSeedUsedByBackupBucket checks whether there is a backupbucket refering the provided seed name
+func IsSeedUsedByBackupBucket(seedName string, backupbuckets []*core.BackupBucket) bool {
+	for _, backupbucket := range backupbuckets {
+		if backupbucket.Spec.SeedName != nil && *backupbucket.Spec.SeedName == seedName {
+			return true
+		}
+	}
+	return false
+}
```
rfranzke, line 70: Can you add unit tests for these functions?

plugin/pkg/shoot/validator/admission.go
```diff
@@ -47,6 +47,9 @@ import (
 const (
 	// PluginName is the name of this admission plugin.
 	PluginName = "ShootValidator"
+
+	useAsSeedAnnotation = "shoot.gardener.cloud/use-as-seed"
```
rfranzke: Can you rather use https://github.com/gardener/gardener/blob/d2b1097a54781271f67c91c91f0f14ea6e969f94/pkg/apis/core/v1beta1/constants/types_constants.go#L234?

plugin/pkg/shoot/validator/admission.go
```diff
@@ -47,6 +47,9 @@ import (
...
+	gardenNamespace     = "garden"
```
rfranzke: Can you rather use https://github.com/gardener/gardener/blob/d2b1097a54781271f67c91c91f0f14ea6e969f94/pkg/apis/core/v1beta1/constants/types_constants.go#L230?

plugin/pkg/shoot/validator/admission.go
```diff
@@ -885,3 +906,47 @@ func ensureMachineImage(oldWorkers []core.Worker, worker core.Worker, images []c
...
+		return nil
+	}
+
+	oldVal, oldOk := oldShoot.Annotations[useAsSeedAnnotation]
+	if !oldOk {
+		return nil
+	}
+	if len(oldVal) == 0 {
+		return nil
+	}
```
```suggestion
	if !oldOk || len(oldVal) == 0 {
		return nil
	}
```

plugin/pkg/shoot/validator/admission.go
```diff
@@ -885,3 +906,47 @@ func ensureMachineImage(oldWorkers []core.Worker, worker core.Worker, images []c
...
+
+	val, ok := shoot.Annotations[useAsSeedAnnotation]
+	if !ok || len(val) == 0 {
```
```suggestion
	if ok {
	  return nil
	}
```

Eliminates one level of nesting

cmd/gardener-apiserver/app/gardener_apiserver.go
```diff
@@ -164,6 +165,7 @@ func (o *Options) complete() error {
 	shootquotavalidator.Register(o.Recommended.Admission.Plugins)
 	shootdns.Register(o.Recommended.Admission.Plugins)
 	shootvalidator.Register(o.Recommended.Admission.Plugins)
+	seedvalidator.Register(o.Recommended.Admission.Plugins)
```
rfranzke, line 168: Please add docs to https://github.com/gardener/gardener/blob/master/docs/concepts/apiserver_admission_plugins.md

plugin/pkg/seed/validator/admission.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	// Ignore updates to status or other subresources
+	if a.GetSubresource() != "" {
+		return nil
+	}
+
+	seedName := a.GetName()
+	if len(seedName) == 0 {
+		return apierrors.NewInternalError(errors.New("empty Seed name"))
```
rfranzke: How can this happen? Is this for the `DELETECOLLECTION` case?

plugin/pkg/seed/validator/admission.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	// Exit early if shoot spec hasn't changed
```
rfranzke: ???

plugin/pkg/seed/validator/admission.go
```diff
@@ -0,0 +1,160 @@
...
+	if a.GetOperation() != admission.Delete {
```
rfranzke: Why do we have to verify this again? The admission plugin is only registered for `Delete`

rfranzke (changes requested):
/assign
/needs changes documentation

plugin/pkg/shoot/validator/admission.go
vpnachev: I have done it initially so, but It looks very strange to me to import something versioned in this package.
If you are fine, I can change it again.

rfranzke: I think we should clean this up eventually (the constants suggest to be versioned due the directory structure, but in fact we don't have (and don't plan to have) multiple versions of them.

rfranzke (approved):
/lgtm

PR #2735, opened 2020-08-17
Remove gardener-seed-admission mutatingwebhookconfiguration

### LINKED ISSUE #2593: "fluentbit.io/exclude=true" is not removed after .spec.purpose change

vlvasilev:

/area logging
/kind cleanup
/priority normal

With this PR we remove the existing gardener-seed-admission Mutatingwebhookconfiguration and remove the mutating pod part of the seed-admission-controller. Thus, the annotation `"fluentbit.io/exclude": "true"` will not be set to any pod.

Fixes [#2593](https://github.com/gardener/gardener/issues/2593)

**Release note**:

```improvement operator
Remove gardener-seed-admission Mutationgwebhookconfiguration and the mutating pod functionality of the seed-admission-controller
```


charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-configmap.yaml
```diff
@@ -282,9 +282,9 @@ data:
         Kube_URL            https://kubernetes.default.svc.cluster.local:443
         Buffer_Size         1M
         Labels              On
-        Annotations         Off
+        Annotations         On
         tls.verify          Off
-        K8S-Logging.Exclude On
+        K8S-Logging.Exclude Off
```
timuthy, line 287: Do you think it makes sense to change these settings in a second step, a few weeks later? There might still be plenty of pods with the exclude annotation.

pkg/operation/seed/seed.go
```diff
@@ -341,6 +342,11 @@ func BootstrapCluster(k8sGardenClient, k8sSeedClient kubernetes.Interface, seed
 		Tag:        &tag,
 	}
 
+	// Delete the mutatingwebhookconfigoration
```
timuthy, line 345: Please add a TODO which says that this logic can be removed in a future Gardener release.

pkg/operation/seed/seed.go
```diff
@@ -769,3 +775,14 @@ func determineClusterIdentity(ctx context.Context, c client.Client) (string, err
...
 }
+
+func deleteMutatingWebHookConfiguration(ctx context.Context, k8sSeedClient client.Client) error {
+	a := &admissionregistrationv1.MutatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "gardener-seed-admission-controller",
+		},
+	}
+
+	err := k8sSeedClient.Delete(context.TODO(), a)
```
timuthy:
```suggestion
	err := k8sSeedClient.Delete(ctx, a)
```

timuthy (changes requested):
Looks good in general, only some questions or minor comments.

charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-configmap.yaml
vpnachev: Regarding `K8S-Logging.Exclude` I think it should be done now, in this way logs for still annotated pods will be collected.
Regarding `Annotations` - why should it be enabled?

timuthy: > I think it should be done now, in this way logs for still annotated pods will be collected.

Can you explain why you want to do this? Tbh, I don't know the details but I was thinking that it might not be a good idea to suddenly record logs of pods which were previously excluded due to `"fluentbit.io/exclude": "true"` and rather give it some time to reduce the odds such pods still exist in the control plane.

vlvasilev: The new version of fluent-bit will ignore those pods. And if a shoot purpose is changed from testing to production, this will prevent the bug of not removing the annotations. The new version of fluent-bit plugin watch Cluster resourses from witch deduct the shoot purpose.

pkg/operation/seed/seed.go
```diff
@@ -769,3 +775,14 @@ func determineClusterIdentity(ctx context.Context, c client.Client) (string, err
 	}
 	return clusterIdentity.Data[v1beta1constants.ClusterIdentity], nil
 }
+
+func deleteMutatingWebHookConfiguration(ctx context.Context, k8sSeedClient client.Client) error {
+	a := &admissionregistrationv1.MutatingWebhookConfiguration{
```
```suggestion
	a := &admissionregistrationv1beta1.MutatingWebhookConfiguration{
```

Use `v1beta1` in case the seed version is too old for `v1`

rfranzke (changes requested):

charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-configmap.yaml
timuthy: Okay got it, thanks for the explanation.

rfranzke (approved):
/lgtm

timuthy (approved):
/lgtm

PR #2739, opened 2020-08-17
Refactor Node Bootstrap

### LINKED ISSUE #2719: Refactor Node Bootstrap for worker nodes with CRI

/area os
/area security
/kind enhancement
/priority normal

Remove the usage of the `docker` CLI during the Node bootstrap but use the CRI instead.
For worker pools that use CRI, this remove the dependency to `docker` altogether.

**CRI enabled worker pools currently require docker**:

Shoot worker pools can specify to use a CRI based container manager such as containerd instead of the default docker container runtime.

As a prerequisite to enabling CRI, the OS image (e.g AWS AMI) needs to have both docker and the CRI container manager preinstalled.

Docker is required because Gardener is using it for several fundamental tasks during the node bootstrap process
 - [Pull the Hyperkube OCI image](https://github.com/gardener/gardener/blob/baa5c6489dfb720597975e2c909af7c7ce595b7a/charts/shoot-cloud-config/templates/scripts/_cloud-config-script.sh#L21)
 - [Extract the kubectl from the OCI image](https://github.com/gardener/gardener/blob/cc33eb85338dfdfc42052ebf5d71cbc9f28657a8/charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet-monitor.service#L15-L14)
 - [Extract kubelet from the OCI image](https://github.com/gardener/gardener/blob/51634b550b0819af24d2dd1a08d5a15288b48d46/charts/seed-operatingsystemconfig/original/templates/kubelet/_kubelet.service#L25) 
 
 As a consequence, OS-Extensions such as [os-gardenlinux](https://github.com/gardener/gardener-extension-os-gardenlinux), always [have to enable the docker systemd service](https://github.com/gardener/gardener-extension-os-gardenlinux/blob/9dd9612d1514308f5a7f3c11d115f5bee9bd6891/pkg/generator/templates/cloud-init.gardenlinux.template#L46)

**Why is this needed**:

 - Always requiring Docker increases the OS image size: always required to contain docker packages for bootstrapping purposes rather than only containing the actual container runtime used to run the container.
  - Images with both docker and the CRI can still be offered for convencience reasons 
- Unwanted resource consumption by running `dockerd` daemon process on the host. 
  ```
  root       13320  0.1  1.7 804580 95004 ?        Ssl  Aug10   3:17 /usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock
  ```
  - with a non-representative memory overhead of ~800 MB
   ```
   pmap 13320 | tail -n 1
   total           804584K

   ```
  - The docker daemon could be removed after the initial bootstrapping, but this is not implemented and also requires additional effort. It is better to get rid of it in the first place.

- Getting rid of the `dockerd` daemon process (any process for that sake) running with root permissions decreases [potential security vulnerabilities](https://www.cvedetails.com/vulnerability-list/vendor_id-13534/product_id-28125/Docker-Docker.html).

**Proposal details**

Kubernetes has since the [introduction of the Container Runtime Interface](https://kubernetes.io/blog/2016/12/container-runtime-interface-cri-in-kubernetes/) (CRI)  moved to CRI as the standard way of container managers to integrate with the kubelet.
The CRI is used by the Kubelet to interact with each container manager in the same fashion - no matter the underlying implementation.

The proposal is to replace the `docker` functionality used in the node bootstrapping process, with calls to the CRI interface implemented usually via a Unix Domain Socket (UDS). 
Docker would not be required any more to bootstrap Shoot worker nodes with CRI enabled. 
To be clear, the current bootstrapping process for worker pools that use the `docker runtime` will still uses the `docker` CLI for bootstrapping.

Gardener can leverage this concept to create a bootstrapping process that works for all CRI enabled container managers alike. 
An OS preinstalled client can interact with the CRI implementation at a predefined socket (e.g. `/run/containerd/containerd.sock` for containerd) to fulfil the steps required for the bootstrapping process.
These are essentially the same as previously done with the `docker` CLI (see above) - only doing the same with the CRI instead.
As a client implementation, the tool [crictl](https://github.com/kubernetes-sigs/cri-tools) can be used. 
This tool is a Kubernetes incubator project, regularly maintained for each Kubernetes / CRI release and seems to be adopted also by the major cloud providers.

danielfoehrKn:

/area os
/kind enhancement
/priority normal

Refactoring node bootstrap to remove docker CLI dependency when bootstrapping nodes with CRI enabled (only containerd atm.).
Relying on `crictl` tool that is either preinstalled on the OS, or if not available, being downloaded.

`crictl` can also be used by an operator during debugging on `docker` nodes, because the kubelet contains the docker CRI shim and also creates the socket `unix:///var/run/dockershim.sock`. 

Fixes #2719

**Special notes for your reviewer**

Adjustment in  Gardenlinux OS (required as Gardenlinux has a read-only file system). 
 - Please see [this PR](https://github.com/gardenlinux/gardenlinux/pull/123)

Other OS (tested with Ubuntu and more to follow) can just install crictl via a systemd service (added by gardener core) if not already present on the machine.
For private cloud scenarios, the binary has to be already installed (like also for packages such as jq, docker, socat, ...)  under `usr/local/bin/crictl`.

Os extensions do not have to enable systemd docker.service any more when bootstrapping with CRI.
- See [PR for OS-Ubuntu](https://github.com/gardener/gardener-extension-os-ubuntu/pull/19)
- See [PR for OS-gardenlinux](https://github.com/gardener/gardener-extension-os-gardenlinux/pull/12)

This PR is the first step to moving CRI specific configurations out of the gardener repository and also out of the os-extensions into their dedicated container runtime extensions (similar to `runtime-gvisor`) . I will open an extra issue describing the details.

**Release note**:

```improvement operator
Refactored Node Bootstrap to not rely on the docker CLI for bootstrapping CRI enabled nodes (e.g with contianerd) - but use `crictl` instead. `Crictl` will be downloaded if not preinstalled on the OS under `usr/local/bin/crictl`. 
```
```improvement operator
Fixes a bug that caused docker to preload the `hyperkube` image when CRI containerd is enabled.
```


rfranzke:
Can we put this behind a feature gate to not break existing OS extensions?

danielfoehrKn:
> Can we put this behind a feature gate to not break existing OS extensions? 

These changes should work without changes to the OS-Extensions as docker is always enabled previously and crictl is downloaded by a systemd service from the Gardener core.

rfranzke:
OK, thanks, the first part sounds good. About the second part:

> and crictl is downloaded by a systemd service from the Gardener core.

What does this mean? Where is it downloaded from? Shouldn't we "ship" it similar to how we bring kubectl/kubelet on the nodes?

timebertt:
> What does this mean? Where is it downloaded from?

AFAIS it is downloaded from GitHub releases.
I also have some concerns about this part. Isn't this the same problem as with kubelet and kubectl, which we are circumventing by extracting them from the container images?

Can we use curl/wget to talk to the cri socket to start and extract binaries (maybe crictl at first and then use it for the rest)?

danielfoehrKn:
> What does this mean? Where is it downloaded from? 

The binary is downloaded from public internet if it is not available on the node under `/usr/local/bin/crictl` yet. 
This would be the same approach as for packages like `jq` , `socat` , `docker` you name it (downloading from package repository). 
For private cloud / scenarios with no internet access the binaries & packages have to pre-installed on the OS.

The problem with shipping `crictl` via container image is that we rely on `crictl` to download and extract images in the first place.

I think the  bottom line is if we want to have a bootstrap process that works independent of the backing container runtime via CRI, both the container runtime and `crictl` needs to be preinstalled for private cloud use-cases.
For public cloud having `crictl`preinstalled is optional (will be the case for gardenlinux but not for e.g Ubuntu).


EDIT: an alternative would be to let an extension define the  command that initially pulls crictl.  This could be done by using its native tools like `ctr` for containerd or the docker cli. But then again we are loosing the decoupling. 
Though this would work for containerd and docker, already CRI-O would cause problems, as podman would have to then be preinstalled on the node for image pulling ....

danielfoehrKn:
> curl/wget to talk to the cri socket to start and extract binaries (maybe crictl at first and then use it for the rest)?

Though not using curl, that is exactly what `crictl` does. So duplicating that functionality with custom code does not make too much sense imo.

danielfoehrKn:
Tested today that spinning up clusters with gardenlinux [with these changes](https://github.com/gardenlinux/gardenlinux/pull/123) and Ubuntu works for both docker and containerd. 

You are welcome to check the code already.

I ll do more testing 
 - Existing Shoots should be unaffected  by these changes (check for gardenlinux & ubuntu os/runtime combinations)
 - Should work without changes to the OS extensions

Known issue
 - Shoots using Gardenlinux and containerd CRI will not work any more with these changes. This is because old versions of Gardenlinux do not include `crictl` and cannot be installed due to the read-only file system
  - either force Shoots to use a newer OS version via image expiration
  - or work on a workaround / migration (prefferred). Possibly by using docker native CLI in case docker.service is enabled and docker cli is installed on the host already.

danielfoehrKn:
Closing this for now as discussed with @rfranzke and @timebertt.

Although we agreed that this is a topic that should receive some attention in the future, priorities are different atm.
Will reopen at a more suitable time.

Possible writing an update to the GEP10 (additional container runtimes) as well.

PR #2740, opened 2020-08-18
Increase concurrent syncs for KCM controllers

rfranzke:

/area scalability
/kind enhancement
/priority normal
/topology seed
/platform all
/exp beginner

Increase concurrent controller syncs on Shoot's kube-controller-manager to allow faster processing for larger clusters.

**Release note**:

```improvement user
The number of concurrent controller syncs of the kube-controller-manager of Shoot clusters has been increased to allow faster processing of events.
```


timebertt (approved):
> `/area scalability`

❤️ 

Nice!
/lgtm

Do you already have some idea/feeling on how much this will increase the resource footprint?

rfranzke:
> Do you already have some idea/feeling on how much this will increase the resource footprint?

No, but I'd assume that for small clusters there's not much change as the additional go routines are not actively doing something and just waiting for new tasks in their event queues. For larger clusters it might lead to some CPU/memory increase, but here I'd assume that this won't be too much as the reconciliations are quite fast, usually.

timuthy (approved):
/lgtm

PR #2743, opened 2020-08-18
Add monitoring for API Server Watches

wyb1:

/area monitoring
/kind enhancement
/priority normal

Add metric shoot:apiserver_watch_duration:quantile. Add dashboard to show length of API Server watches based on resource. These metrics can be helpful in identifying resources with troublesome WATCHES. WATCHES should last a long time so if a resource has many WATCHES that are very short this could point to an issue.

Fixes #

**Release note**:

```improvement operator
Add monitoring for API Server Watches
```

### DIFF
--- a/charts/seed-monitoring/charts/core/charts/prometheus/rules/kube-apiserver.rules.yaml
+++ b/charts/seed-monitoring/charts/core/charts/prometheus/rules/kube-apiserver.rules.yaml
@@ -49,6 +49,31 @@ groups:
     expr: histogram_quantile(0.5, rate(apiserver_request_duration_seconds_bucket[5m]))
     labels:
       quantile: "0.5"
+  # TODO replace with better metrics in the future (wyb1)
+  - record: shoot:apiserver_watch_duration:quantile
+    expr: histogram_quantile(0.2, sum(rate(apiserver_request_duration_seconds_bucket{verb="WATCH",resource=~"configmaps|deployments|secrets|daemonsets|services|nodes|pods|namespaces|endpoints|statefulsets|clusterroles|roles"}[5m])) by (le,scope,resource))
+    labels:
+      quantile: "0.2"
+  - record: shoot:apiserver_watch_duration:quantile
+    expr: histogram_quantile(0.5, sum(rate(apiserver_request_duration_seconds_bucket{verb="WATCH",resource=~"configmaps|deployments|secrets|daemonsets|services|nodes|pods|namespaces|endpoints|statefulsets|clusterroles|roles"}[5m])) by (le,scope,resource))
+    labels:
+      quantile: "0.5"
+  - record: shoot:apiserver_watch_duration:quantile
+    expr: histogram_quantile(0.9, sum(rate(apiserver_request_duration_seconds_bucket{verb="WATCH",resource=~"configmaps|deployments|secrets|daemonsets|services|nodes|pods|namespaces|endpoints|statefulsets|clusterroles|roles"}[5m])) by (le,scope,resource))
+    labels:
+      quantile: "0.9"
+  - record: shoot:apiserver_watch_duration:quantile
+    expr: histogram_quantile(0.2, sum(rate(apiserver_request_duration_seconds_bucket{verb="WATCH",group=~".+garden.+"}[5m])) by (le,scope,resource))
+    labels:
+      quantile: "0.2"
+  - record: shoot:apiserver_watch_duration:quantile
+    expr: histogram_quantile(0.5, sum(rate(apiserver_request_duration_seconds_bucket{verb="WATCH",group=~".+garden.+"}[5m])) by (le,scope,resource))
+    labels:
+      quantile: "0.5"
+  - record: shoot:apiserver_watch_duration:quantile
+    expr: histogram_quantile(0.9, sum(rate(apiserver_request_duration_seconds_bucket{verb="WATCH",group=~".+garden.+"}[5m])) by (le,scope,resource))
+    labels:
+      quantile: "0.9"
   - alert: KubeApiServerTooManyOpenFileDescriptors
     expr: 100 * process_open_fds{job="kube-apiserver"} / process_max_fds > 50
     for: 30m
--- a/charts/seed-monitoring/charts/grafana/dashboards/operators/kubernetes-api-server-watches-dashboard.json
+++ b/charts/seed-monitoring/charts/grafana/dashboards/operators/kubernetes-api-server-watches-dashboard.json
@@ -1,0 +1,429 @@
+{
+  "annotations": {
+    "list": []
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 1,
+  "iteration": 1598001198472,
+  "links": [],
+  "panels": [
+    {
+      "datasource": null,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 4,
+      "panels": [],
+      "repeat": "Scope",
+      "scopedVars": {
+        "Scope": {
+          "selected": false,
+          "text": "cluster",
+          "value": "cluster"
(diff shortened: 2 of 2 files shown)


timebertt:
Thank you very much for the PR!
I will check it out in detail tomorrow :)

charts/seed-monitoring/charts/core/charts/prometheus/rules/kube-apiserver.rules.yaml
```diff
@@ -49,6 +49,19 @@ groups:
     expr: histogram_quantile(0.5, rate(apiserver_request_duration_seconds_bucket[5m]))
     labels:
       quantile: "0.5"
+  # TODO remove with better metrics in the future (wyb1)
+  - record: shoot:apiserver_watch_duration:quantile
+    expr: histogram_quantile(0.2, sum(rate(apiserver_request_duration_seconds_bucket{verb="WATCH",resource=~"configmaps|deployments|secrets|daemonsets|services|nodes|pods|namespaces|endpoints|events|statefulsets|clusterroles|roles"}[5m])) by (le,scope,resource))
```
timebertt: In general the list of resources looks good. Still, two remarks:
- I think we can remove `events`, we have hardly any watches on events and it shouldn't be a problem if those watches don't last very long.
- I'm wondering if we should also collect metrics for our extension resources on the Seeds. If we have problems with short watches there, this will likely result in failures. As we are anyways in the evaluation phase for those metrics, can we add them and see what we will record over the next few weeks? My proposal would be to add a second recording rule with something like `group=~".+garden.+"`, if that's possible.

WDYT?

charts/seed-monitoring/charts/grafana/dashboards/operators/apiserver-watches-dashboard.json
```diff
@@ -0,0 +1,391 @@
+{
```
timebertt: Does it make sense to rename this file to `kubernetes-api-server-watches-dashboard.json` to keep it consistent with the other dashboard files?

charts/seed-monitoring/charts/grafana/dashboards/operators/apiserver-watches-dashboard.json
```diff
@@ -0,0 +1,391 @@
...
+      "5m",
+      "15m",
+      "30m",
+      "1h",
+      "2h",
+      "1d"
+    ]
+  },
+  "timezone": "",
+  "title": "API Server Watches",
```
timebertt:
```suggestion
  "title": "Kubernetes API Server Watches",
```

charts/seed-monitoring/charts/grafana/dashboards/operators/apiserver-watches-dashboard.json
```diff
@@ -0,0 +1,391 @@
...
+      "linewidth": 1,
+      "nullPointMode": "null",
+      "options": {
+        "dataLinks": []
+      },
+      "percentage": false,
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "repeat": "Quantile",
```
timebertt: IIUC, this configures to have one panel per selected quantile, right?
I think for `scope` it makes kind of sense, but it would be nice to be able to see the different quantiles in one panel. (maybe also for `scope`)
WDYT?

charts/seed-monitoring/charts/grafana/dashboards/operators/apiserver-watches-dashboard.json
```diff
@@ -0,0 +1,391 @@
...
+      "dashes": false,
+      "datasource": null,
+      "description": "$Quantile of watches on this resource are less than or equal to the value.",
+      "fieldConfig": {
+        "defaults": {
+          "custom": {}
+        },
+        "overrides": []
+      },
+      "fill": 1,
```
timebertt: I think `fill=0` makes this dashboard easier on the eyes 😉 
```suggestion
      "fill": 0,
```

charts/seed-monitoring/charts/core/charts/prometheus/rules/kube-apiserver.rules.yaml
timebertt: In smaller clusters, we might not have enough data for the rate of `5m` to compute a value. Should we increase to `10m`?
```suggestion
    expr: histogram_quantile(0.2, sum(rate(apiserver_request_duration_seconds_bucket{verb="WATCH",resource=~"configmaps|deployments|secrets|daemonsets|services|nodes|pods|namespaces|endpoints|events|statefulsets|clusterroles|roles"}[10m])) by (le,scope,resource))
```

timebertt (commented):
Very nice PR!
/invite @istvanballok

charts/seed-monitoring/charts/core/charts/prometheus/rules/kube-apiserver.rules.yaml
wyb1: I can change it to `10m`. I assumed this is relevant for our large seeds so `5m` would be sufficient.

wyb1: I will remove `events`. Regarding the extensions, do you mean that we should have the same metrics again, but it is a subset containing only requests from the group `.+garden.+`?

charts/seed-monitoring/charts/grafana/dashboards/operators/apiserver-watches-dashboard.json
wyb1: I can remove the repeat for `Quantile`, but I would suggest that we keep the repeat for `scope`. At some point the panel will become too crowded and I think separating it by scope is fine. There is a shared tooltip that makes it easy to compare different panels.

timebertt: Yes, sounds good 👍

charts/seed-monitoring/charts/core/charts/prometheus/rules/kube-apiserver.rules.yaml
timebertt: Well, you're right. `5m` should be sufficient. If I think about it: if we have too few values for the rate computation, then we probably also don't run into issues with the watch duration 😄 
So, let's keep `5m`...

timebertt: Yes, basically I mean the same metric name with the same source, only with a selection for gardener extension/gardener-resource-manager/dns resources.

charts/seed-monitoring/charts/grafana/dashboards/operators/apiserver-watches-dashboard.json
```diff
@@ -0,0 +1,391 @@
...
+      "title": "Watch Durations for Scope: $Scope",
+      "type": "row"
+    },
+    {
+      "aliasColors": {},
+      "bars": false,
+      "dashLength": 10,
+      "dashes": false,
+      "datasource": null,
+      "description": "$Quantile of watches on this resource are less than or equal to the value.",
```
timebertt: I think it might be irritating to people, that in the normal/healthy case all watch duration metrics are at `60s`. Can you add a note here, that this is the maximum value recorded by the API server and `60s` means actually `[60s,+Inf)`?

wyb1:
Thanks for the review @timebertt! I've addressed the comments. PTAL

timebertt (dismissed):
Thanks! Great work, I like it :)
/lgtm

istvanballok (approved):
/lgtm
Kudos! 👍

wyb1:
Added thresholds to the panels to add visual clues for which values are acceptable.

timebertt (approved):
Good idea!
/lgtm

rfranzke (approved):

Issue #2457, opened 2020-06-16
deploy logging stack as early as possible in shoot reconciliation flow


vlerenc:
**How to categorize this issue?**
/area control-plane logging ops-productivity
/kind enhancement
/priority normal

**What would you like to be added**:
I checked a cluster that couldn't be deleted because of infra dependencies:
```
$ k get pods
NAME                                         READY   STATUS              RESTARTS   AGE
daily-curator-1592268360-fgw25               0/1     Completed           0          137m
elasticsearch-logging-0                      2/2     Running             0          18d
etcd-events-0                                2/2     Running             0          18d
etcd-main-0                                  2/2     Running             0          18d
gardener-resource-manager-549878d987-mkskc   1/1     Running             0          11h
hourly-curator-1592275260-cvzb6              0/1     Completed           0          22m
kibana-logging-5f44699c95-txjzs              2/2     Running             0          18d
kube-controller-manager-9f655d794-lgkpg      0/1     ContainerCreating   0          2d16h
kube-scheduler-76f7bff879-h5hcw              0/1     CrashLoopBackOff    748        39h
term-6224465904394368752                     1/1     Running             0          8m12s
```

What I noticed was:
1. There was no API server anymore, but dependent components. This is more cosmetic as it creates somewhat needless additional confusion to ops/support staff. The expectation is that the components get deleted in reverse order of creation.
2. I couldn't find the TF pod logs in the logging stack, but maybe I was looking wrong? The only thing I saw were scheduler logs. Where are the TF pod logs? Dropped after the TF pod issue (constant retries that flooded fluentd that then dropped all else)?

**Why is this needed**:
(1) is about less noise/avoidable errors and (2) about finding the logs for what the cluster operation error refers to.

ialidzhikov:
> 2. I couldn't find the TF pod logs in the logging stack, but maybe I was looking wrong? The only thing I saw were scheduler logs. Where are the TF pod logs? Dropped after the TF pod issue (constant retries that flooded fluentd that then dropped all else)?

Actually the provider extensions log the full content of the logs of a terraform Pod after its execution. See https://github.com/gardener/gardener/blob/a3c0c1f06848b44e8492319c83e51b8f99cd3901/extensions/pkg/terraformer/terraformer.go#L204-L212

[...]

PR #2750, opened 2020-08-19
Deploy logging stack earlier in the reconciliation flow

Kristian-ZH:

/area logging
/kind enhancement
/priority normal

When Loki pod is missing, fluent-bit keeps logs(trying to send it) for 3.5min. After that the log will be dropped.
With this change there is bigger chance to catch logs from the shoot creation

https://github.com/gardener/gardener/issues/2457

**Release note**:

```improvement operator
Deploy logging stack earlier in the reconciliation flow.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -103,6 +103,11 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 			Name: "Deploying Shoot namespace in Seed",
 			Fn:   flow.TaskFn(botanist.DeployNamespace).RetryUntilTimeout(defaultInterval, defaultTimeout),
 		})
+		deploySeedLogging = g.Add(flow.Task{
+			Name:         "Deploying shoot logging stack in Seed",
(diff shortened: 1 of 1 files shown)


pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -248,6 +248,11 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 			Fn:           flow.TaskFn(botanist.WaitUntilKubeAPIServerReady).SkipIf(o.Shoot.HibernationEnabled),
 			Dependencies: flow.NewTaskIDs(deployKubeAPIServer),
 		})
+		deploySeedLogging = g.Add(flow.Task{
```
rfranzke: Is something speaking against deploying it right after the creation of the shoot namespace and the generation of the secrets? How is it dependent on the kube-apiserver?

Kristian-ZH: Hmmm you are correct. Loki depends on the seed's Apiserver because it fetches the logs metadata from it, but does not depend on the shoot apiserver

rfranzke: 👍 Then let's deploy it as early as possible, i.e., after the secrets are generated and ready

rfranzke (approved):
/lgtm

Issue #2747, opened 2020-08-18
Gardenlet times out talking to the Seed API server


timebertt:
**How to categorize this issue?**

/area quality performance robustness scalability
/kind bug
/priority normal

**What happened**:
Over the past few weeks we have observed, that the gardenlet is often timing out talking to the Seed API server during Shoot reconciliation but also while renewing the leader lease, ending in quite frequent logs including `context deadline exceeded`.
We have seen this on two of our shooted Seeds (Azure in the US, Control plane running in the EU), which have a quite dynamic load (many Shoots creating/deleting in parallel).

The problematic steps in the Shoot reconciliation/deletion that fail are mostly those, that use the `defaultTimeout` of 30s and touch a lot of resources (for example `DeleteSeedMonitoring`).

**What you expected to happen**:
The gardenlet should be able to cope with many Shoots being reconciled in parallel.

**How to reproduce it (as minimally and precisely as possible)**:
What actually limits the effect of this issue is to set the `shoot.concurrentSyncs` value of the gardenlet configuration to a low value (e.g. 20). With this, the `context deadline exceeded` errors cease to pop up in the logs.

**Anything else we need to know?**:

**Environment**:

- Gardener version: >= v1.7.0
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration: azure (?)
- Others:

/cc @timuthy @dguendisch @zanetworker @rfranzke

rfranzke:
Can you please share more information which parts you already investigated (networking, etc.)?

timebertt:
While opening this issue, I stumbled upon the wrong API groups in `DeleteSeedMonitoring`, e.g. `RoleBinding` is not in `core/v1`:
https://github.com/gardener/gardener/blob/1208bc82aa0c6910b858a00993ac0b2f1588c87c/pkg/operation/botanist/monitoring.go#L423

There are 5 of these in that function (see https://github.com/gardener/gardener/compare/master...timebertt:fix/api-groups) and in each iteration the dynamic rest mapper rediscover all API resources, because it gets a `NoKindMatchError`. As the dynamic rest mapper locks on rediscovery, one call will block all other Shoot reconciliations/deletions trying to rediscover some resource in the seed.
In some cases (e.g. the gardenlet running far away from its control plane like in our case), a full discovery may take about 2 seconds, so `5*2s + lock contention time` might easily exceed the default timeout of 30s.

/status under-investigation

Stuff we should check/do next:
[...]

PR #2752, opened 2020-08-19
Fix wrong API groups

timebertt:

/area quality
/kind bug
/priority critical

This PR fixes the wrong API groups used in the `DeleteSeedMonitoring` which caused a lot of rediscoveries on the REST mapper.
To make it more safe, I switched from unstructured objects to the respective types (also did it for the other places in the botanist package where we used unstructured objects).

Part of #2747 

**Release note**:

```improvement operator
A bug has been fixed that caused the REST Mapper to rediscover the available API resources very often.
```

### DIFF
--- a/pkg/operation/botanist/monitoring.go
+++ b/pkg/operation/botanist/monitoring.go
@@ -29,11 +29,15 @@ import (
 	"github.com/gardener/gardener/pkg/utils"
 	"github.com/gardener/gardener/pkg/utils/secrets"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	networkingv1 "k8s.io/api/networking/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime"
+	autoscalingv1beta2 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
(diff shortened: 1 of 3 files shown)


rfranzke (dismissed):
/lgtm

pkg/operation/botanist/monitoring.go
```diff
@@ -413,47 +418,149 @@ func (b *Botanist) DeleteSeedMonitoring(ctx context.Context) error {
...
+		&corev1.ServiceAccount{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: b.Shoot.SeedNamespace,
+				Name:      "prometheus",
+			},
+		},
+		&corev1.Service{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: b.Shoot.SeedNamespace,
+				Name:      "prometheus",
```
ialidzhikov: I know it is not introduced with this PR, but there is no Service named `prometheus` in the shoot ns. We could remove it - 1 call less. :)

pkg/operation/botanist/monitoring.go
```diff
@@ -413,47 +418,149 @@ func (b *Botanist) DeleteSeedMonitoring(ctx context.Context) error {
...
+				Name:      "blackbox-exporter-config-prometheus",
+			},
+		},
+		&corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: b.Shoot.SeedNamespace,
+				Name:      "prometheus-basic-auth",
+			},
+		},
+		&extensionsv1beta1.Ingress{
```
ialidzhikov, line 507: I know it is not introduced with this PR but why the same resource is deleted 2 times from different apiVersion. Ingress will be no longer served from `extensions/v1beta1` starting v1.22 according to the current plans and this is far in the future. I don't see why we need deletion also from `networking.k8s.io/v1beta1`.

pkg/operation/botanist/monitoring.go
timebertt: Interesting, but we specify
https://github.com/gardener/gardener/blob/e8fc9090204e32dbeab38b20539431e4418916a5/charts/seed-monitoring/charts/core/charts/prometheus/templates/prometheus.yaml#L47
in the Prometheus StatefulSet.... 🤔 
Why isn't there a service called `prometheus`? (cc @wyb1)

pkg/operation/botanist/monitoring.go
timebertt: Good point, removed `networkingv1beta1.Ingress`.

pkg/operation/botanist/monitoring.go
rfranzke: Independent of the question why `.spec.serviceName` is configured this way, we definitely don't have a `Service` with name `prometheus`, so it makes sense to remove this object as @ialidzhikov suggested

timebertt: Yep. Done, removed.

ialidzhikov (approved):
/lgtm

Issue #2175, opened 2020-04-13
Graduate logging feature to beta 


vpnachev:
**What would you like to be added**:
The logging stack running in the seed clusters responsible for:
1. logging of the the control plane components
1. seed system components like `gardenlet` and gardener extensions

was initially contributed as an alpha feature an year and a half ago https://github.com/gardener/gardener/pull/176. Since then this components is running without major issues and I propose to graduate it to `beta`, which means:
- [ ] The feature gate is enabled by default, [ref](https://github.com/gardener/gardener/blob/master/pkg/gardenlet/features/features.go#L28).
- [ ] Deploy the logging component as early as possible in the shoot [creation/reconciliation  flow](https://github.com/gardener/gardener/blob/master/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go#L96). This is useful when creation is failing and persisted logs are needed for analysis.
- [ ] Use OIDC for authentication instead of Basic Authentication.

**Why is this needed**:

vpnachev:
@mvladev for the  task
```
Use OIDC for authentication instead of Basic Authentication.
```
I think it can be handled independently of this one. There is already an enhancement proposal #999 about it. What do you think?

vpnachev:
With https://github.com/gardener/gardener/issues/918 the logging stack will be completely redesigned, thus it does not make sense to graduate the existing solution to `beta`.

PR #2755, opened 2020-08-19
Promote HVPA feature gates to beta

### LINKED ISSUE #2724: HVPA feature promotion

/area auto-scaling usability
/kind enhancement
/priority normal

Let's promote the [`HVPA` feature gates](https://github.com/gardener/gardener/blob/master/pkg/gardenlet/features/features.go#L29-L30) in gardenlet to `beta` and enable them by default. After a couple of release we should then promote them to `GA` and eventually remove the feature gate.

**Why is this needed**:
Gardener is pretty much relying on HVPA for auto-scaling the etcds and kube-apiservers of shoot clusters. The past releases have shown that it's working quite well and stable enough to plan for its promotion, so that we can eventually get rid of the feature gate again.

/assign @ggaurav10 @amshuman-kr

rfranzke:

/area auto-scaling
/kind enhancement
/priority normal

With this PR we promote the HVPA feature gates to beta as the features are quite stable meanwhile.

Fixes #2724

/invite @ggaurav10 @amshuman-kr 

**Release note**:

```noteworthy operator
The `HVPA` and `HVPAForShootedSeed` feature gates have been promoted and are enabled by default. You can still disable them explicitly if you prefer so.
```

### DIFF
--- a/pkg/gardenlet/features/features.go
+++ b/pkg/gardenlet/features/features.go
@@ -26,8 +26,8 @@ var (
 	FeatureGate  = featuregate.NewFeatureGate()
 	featureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 		features.Logging:              {Default: false, PreRelease: featuregate.Alpha},
-		features.HVPA:                 {Default: false, PreRelease: featuregate.Alpha},
-		features.HVPAForShootedSeed:   {Default: false, PreRelease: featuregate.Alpha},
+		features.HVPA:                 {Default: true, PreRelease: featuregate.Beta},
+		features.HVPAForShootedSeed:   {Default: true, PreRelease: featuregate.Beta},
 		features.ManagedIstio:         {Default: false, PreRelease: featuregate.Alpha},
 		features.KonnectivityTunnel:   {Default: false, PreRelease: featuregate.Alpha},
 		features.APIServerSNI:         {Default: false, PreRelease: featuregate.Alpha},
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -33,12 +33,12 @@ const (
 
 	// HVPA enables simultaneous horizontal and vertical scaling in Seed Clusters.
 	// owner @ggaurav10, @amshuman-kr
-	// alpha: v0.31.0
+	// alpha: v0.31.0, beta: v1.9.0
 	HVPA featuregate.Feature = "HVPA"
 
 	// HVPAForShootedSeed enables simultaneous horizontal and vertical scaling in shooted seed Clusters.
 	// owner @ggaurav10, @amshuman-kr
-	// alpha: v0.32.0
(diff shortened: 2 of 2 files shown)


ggaurav10 (approved):

pkg/gardenlet/features/features.go
```diff
@@ -26,8 +26,8 @@ var (
 	FeatureGate  = featuregate.NewFeatureGate()
 	featureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 		features.Logging:              {Default: false, PreRelease: featuregate.Alpha},
-		features.HVPA:                 {Default: false, PreRelease: featuregate.Alpha},
-		features.HVPAForShootedSeed:   {Default: false, PreRelease: featuregate.Alpha},
+		features.HVPA:                 {Default: true, PreRelease: featuregate.Beta},
```
ialidzhikov, line 29: Isn't there already a redesign work for hvpa - ref https://github.com/gardener/hvpa-controller/pull/76? Do we want to enable it despite the fact that it will be changed/redesigned?

rfranzke: @ggaurav10 @amshuman-kr can you comment on that, please?

amshuman-kr: Yes, [gardener/hvpa-controller#76](https://github.com/gardener/hvpa-controller/pull/76) is a redesigned implementation for HVPA based on the new [design proposal](https://github.com/gardener/hvpa-controller/blob/master/docs/proposals/HVPA-EP-0001/HVPA-EP-001.md) to be able to control horizontal scaling as well as vertical scaling (currently only the vertical scaling is controlled).

The design has been reviewed and the implementation is under review. I have a few suggestions for the implementation, so it will take some time for the implementation to be merged. Also, we need to run a round of performance test to verify the changes are working fine at scale. So, it might be a few weeks before the redesign is released.

AFAIK, the HVPA feature gates are already enabled in our own landscapes. So, if the concern is that redesign might cause instability, then the feature gates themselves will not help (at least not for our own landscapes). We will need other mechanisms to mitigate (good test coverage, performance testing, phased rollout of the new version etc.)

So, I am fine promoting the feature gate. But no strong opinion. I am fine holding the promotion too.

amshuman-kr (approved):
The PR itself looks good apart from the main question of [holding the promotion until the HVPA redesign](https://github.com/gardener/gardener/pull/2755#discussion_r473627835).

pkg/gardenlet/features/features.go
vpnachev: Then I suggest to keep it alpha, as we have done with the logging https://github.com/gardener/gardener/issues/2175#issuecomment-633915482

amshuman-kr: Ah, there is precedence. Makes sense 👍

PR #2756, opened 2020-08-19
Auto-restart scheduler in shoot maintenance

### LINKED ISSUE #2722: Auto-restart kube-scheduler in maintenance time window

/area robustness quality
/kind enhancement
/priority normal
/topology seed
/exp beginner
/platform all

With #2098 we have added the auto-restarts of the KCM (and other components with the respective labels) in the shoot's maintenance time windows. The rationale behind it was to circumvent stale cache/stuck controller behaviours. However, we have not enabled this for the `kube-scheduler` which is no different than KCM and could have the same problems.
Hence, let's also do the same for the `kube-scheduler` as we have seen the mentioned issues with it meanwhile as well.

/cc @hardikdr who might have more inputs/comments.

**Why is this needed**:
Robustness, less ops effort, better user experience.

rfranzke:

/area ops-productivity robustness
/kind enhancement
/priority normal

The `kube-scheduler` is now auto-restarted in the shoot maintenance time window, similar to other controllers.

Fixes #2722

As #2731 is still in discussion, I'm making this change now separately and rebase the PR later if necessary.

**Release note**:

```improvement operator
The `kube-scheduler` is now auto-restarted in the shoot maintenance time window, similar to other controllers.
```

### DIFF
--- a/charts/seed-controlplane/charts/kube-scheduler/templates/kube-scheduler.yaml
+++ b/charts/seed-controlplane/charts/kube-scheduler/templates/kube-scheduler.yaml
@@ -47,6 +47,9 @@ spec:
         networking.gardener.cloud/to-dns: allowed
         networking.gardener.cloud/to-shoot-apiserver: allowed
         networking.gardener.cloud/from-prometheus: allowed
+{{- if .Values.podLabels }}
+{{ toYaml .Values.podLabels | indent 8 }}
+{{- end }}
     spec:
       containers:
       - name: kube-scheduler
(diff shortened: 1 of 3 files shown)


timuthy (approved):
/lgtm

danielfoehrKn (approved):

ialidzhikov:
We initially stared with restart for cloud-controller-manager, then we did the same for kcm and mcm, now we do it also the kube-scheduler in this PR. This sounds like we bury/workaround issues? Are there occurrences of kube-scheduler Pod which "hangs" for some reason? If yes, shouldn't we rather try to understand the root cause?

rfranzke:
@ialidzhikov I don't think there are clear steps to reproduce the problems that we see (luckily very rarely). If you can help here it's highly appreciated to tackle the root cause in the first place, sure, but until then this one-time auto-restart in the maintenance time window is a simply thing to improve the ops experience.

PR #2757, opened 2020-08-19
Remove shoot controller tasks immediately after they are completed

### LINKED ISSUE #2742: Update shoot.gardener.cloud/tasks once the step is completed

/area robustness
/kind enhancement
/priority normal

Currently `shoot.gardener.cloud/tasks` annotations helps keeping track of "tasks" that need to be performed during the reconcile of a Shoot for example (`shoot.gardener.cloud/tasks='deployInfrastructure,restartControlPlanePods'`. Currently the annotation is updated at the end of the reconciliation flow. Which means that for example if Shoot is stuck for hours on waiting until worker is ready, then each reconcile/retry will execute the tasks again and again (for example `deployInfrastructure`, `restartControlPlanePods`). In some cases this can be undesired as for example restarting controlplane pods contributes to the cloud provider rate limit.
Generally we can remove the task, once the step associated with this task is completed successfully.

**Why is this needed**:
To prevent given task to be executed "n" times.

rfranzke:

/area robustness networking
/kind enhancement
/priority normal

Earlier, the shoot tasks were only removed at the end of a successful reconciliation. Now they are removed immediately once the respective task completed successfully. This prevents that they are executed too often again and again in case a subsequent task fails and the whole shoot reconciliation must be retried.

Fixes #2742

/cc @vlerenc 
/invite @ialidzhikov 

**Release note**:

```improvement operator
The shoot task annotation is now updated as soon as the respective task has completed successfully to prevent recurring executions in case the whole shoot reconciliation flow fails.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -91,7 +91,6 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		staticNodesCIDR                = o.Shoot.Info.Spec.Networking.Nodes != nil
 		useSNI                         = botanist.APIServerSNIEnabled()
 		generation                     = o.Shoot.Info.Generation
-		allTasks                       = controllerutils.GetTasks(o.Shoot.Info.Annotations)
 		requestControlPlanePodsRestart = controllerutils.HasTask(o.Shoot.Info.Annotations, common.ShootTaskRestartControlPlanePods)
 
 		g                      = flow.NewGraph("Shoot cluster reconciliation")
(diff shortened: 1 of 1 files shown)


danielfoehrKn (dismissed):
Looks good to me

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -446,28 +455,31 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
...
+			err = fmt.Errorf("failed to invalidate shoot client: %w", err)
+			return gardencorev1beta1helper.NewWrappedLastErrors(gardencorev1beta1helper.FormatLastErrDescription(err), err)
+		}
+	}
+
+	o.Logger.Infof("Successfully reconciled Shoot %q", o.Shoot.Info.Name)
+	return nil
+}
+
+func (c *Controller) removeTaskAnnotation(o *operation.Operation, generation int64, tasksToRemove ...string) error {
```
ialidzhikov: This func currently is not using anything from the receiver.

```suggestion
func removeTaskAnnotation(o *operation.Operation, generation int64, tasksToRemove ...string) error {
```

ialidzhikov (commented):
Minor nit

timuthy (approved):

timebertt (approved):

PR #2759, opened 2020-08-20
Report failed machines for hibernated clusters

### LINKED ISSUE #2691: Generic worker actuator doesn't check machine status when a shoot is hibernated

/area ops-productivity
/kind bug
/priority normal
/platform all
/exp intermediate
/topology seed

**What happened**:
When a shoot is hibernated then we don't check the error messages in the machines or machine deployments. For example, when the user changes the AWS credentials without updating the secret in Gardener and then tries to hibernate the shoot then it will never succeed, though, we don't show why. Let's improve this 

Generic worker actuator should clearly state what is wrong.

See above.

- Gardener version: v1.8

timuthy:

/area ops-productivity
/kind bug
/priority normal

This PR fixes an issues which prevented machine error from being reported if the shoot is being hibernated. More details can be found in the linked issue.

I tested the change/scenario with a re-vendored GCP provider extension successfully.

Fixes #2691

**Release note**:

```improvement operator
The generic worker actuator now also reports failed machines from the corresponding machine deployment in case the shoot cluster is being hibernated. Earlier scale down issues during hibernation were not reported to users, e.g. if something was wrong with the configured cloud provider account and thus the machine deletion was denied.
```

### DIFF
--- a/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
@@ -360,8 +360,8 @@ func (a *genericActuator) waitUntilWantedMachineDeploymentsAvailable(ctx context
 			// We want to wait until all wanted machine deployments have as many
 			// available replicas as desired (specified in the .spec.replicas).
 			// However, if we see any error in the status of the deployment then we return it.
-			for _, failedMachine := range deployment.Status.FailedMachines {
-				return retryutils.SevereError(fmt.Errorf("machine %s failed: %s", failedMachine.Name, failedMachine.LastOperation.Description))
+			if machineErrs := workerhelper.ReportFailedMachines(deployment.Status); machineErrs != nil {
+				return retryutils.SevereError(machineErrs)
 			}
 
 			numberOfAwakeMachines += deployment.Status.Replicas
--- a/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
@@ -357,9 +357,16 @@ func (a *genericActuator) waitUntilWantedMachineDeploymentsAvailable(ctx context
 				}
 			}
 
-			// If the shoot get hibernated we want to wait until all wanted machine deployments have been deleted
-			// entirely.
+			// We want to wait until all wanted machine deployments have as many
+			// available replicas as desired (specified in the .spec.replicas).
+			// However, if we see any error in the status of the deployment then we return it.
+			for _, failedMachine := range deployment.Status.FailedMachines {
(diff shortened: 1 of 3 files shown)


extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -357,9 +357,16 @@ func (a *genericActuator) waitUntilWantedMachineDeploymentsAvailable(ctx context
 				}
 			}
 
-			// If the shoot get hibernated we want to wait until all wanted machine deployments have been deleted
-			// entirely.
+			// We want to wait until all wanted machine deployments have as many
+			// available replicas as desired (specified in the .spec.replicas).
+			// However, if we see any error in the status of the deployment then we return it.
+			for _, failedMachine := range deployment.Status.FailedMachines {
```
danielfoehrKn: While you are already at it: if there are multiple failed machines, can we return a summary/or the number of all failed machines here. Otherwise only one failed machine is reported.

rfranzke (dismissed):
/lgtm

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
timuthy: Here my suggestion 296c56125224d7ff4d59fa544ab1e3e0407306b1.

timuthy:
I pushed another commit (296c56125224d7ff4d59fa544ab1e3e0407306b1) to address @danielfoehrKn's comment and group all failed machines of a deployment.

extensions/pkg/controller/worker/helper/helper.go
```diff
@@ -77,3 +82,29 @@ func GetMachineSetWithMachineClass(machineDeploymentName, machineClassName strin
...
+		}
+		descriptionPerFailedMachines[machine.LastOperation.Description] += fmt.Sprintf(", %q", machine.Name)
+	}
+
+	allErrs := &multierror.Error{
+		ErrorFormat: utilerrors.NewErrorFormatFuncWithPrefix("machine(s) failed"),
+	}
+	for description, names := range descriptionPerFailedMachines {
+		allErrs = multierror.Append(allErrs, fmt.Errorf("%s: %s", names, description))
+	}
```
rfranzke: What about replacing `map[string]string` with `map[string][]string` and `append`ing the machine names, so that you can later `strings.Join` them with `,` as delimiter? This way you also wouldn't need the check in line 95 whether the description is already known.

timuthy: Sure, no strong opinion.

rfranzke (approved):

danielfoehrKn (approved):
Thanks for improving the error messages!

PR #2760, opened 2020-08-20
Nit: Use logrus.FieldLogger instead of logrus.Entry

ialidzhikov:

/area dev-productivity
/priority normal

`logrus.FieldLogger` is interface that generalizes the `logrus.Entry` and `logrus.Logger` types. We should not restrict utility funcs (for example `WaitUntilExtensionCRReady`) to accept only `logrus.Entry` and it should be also possible passing `logrus.Logger`.

**Release note**:

```improvement operator
NONE
```


timebertt (approved):
Thanks!
/lgtm

rfranzke (approved):
/lgtm

PR #2761, opened 2020-08-20
Deny shoots referencing absent DNS provider secrets

### LINKED ISSUE #2736: Validate DNS provider secret refs

/area control-plane
/kind enhancement
/priority normal

If a shoot configures DNS providers `.spec.dns.providers`, Gardener should check if the referenced secret via `SecretName` exists in the project namespace and deny or admit the request accordingly.

**Why is this needed**:
Today, Gardener will only fail during the reconciliation if a provider secret does not exists and with the above mentioned admission the user will get the feedback sooner.

timuthy:

/area robustness
/kind enhancement
/kind api-change
/priority normal

This PR makes the `ResourceReferenceManager` admission plugin check if referenced DNS provider secrets (`shoot.spec.dns.providers[*].secretName`) exist in the project namespace.

Fixes #2736

**Release note**:

```noteworthy user
Gardener now checks if referenced DNS provider secrets (.spec.dns.providers[*].secretName) exist in the project namespace during shoot creation and update requests. Requests will be denied if the referenced secret is not available.
```

### DIFF
--- a/plugin/pkg/global/resourcereferencemanager/admission.go
+++ b/plugin/pkg/global/resourcereferencemanager/admission.go
@@ -581,6 +581,17 @@ func (r *ReferenceManager) ensureShootReferences(ctx context.Context, attributes
 		}
 	}
 
+	if shoot.Spec.DNS != nil && shoot.DeletionTimestamp == nil {
+		for _, dnsProvider := range shoot.Spec.DNS.Providers {
+			if dnsProvider.SecretName == nil {
+				continue
+			}
+			if err := r.lookupSecret(shoot.Namespace, *dnsProvider.SecretName); err != nil {
+				return fmt.Errorf("failed to reference DNS provider secret %v", err)
+			}
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -581,6 +581,17 @@ func (r *ReferenceManager) ensureShootReferences(ctx context.Context, attributes
 		}
 	}
 
+	if shoot.Spec.DNS != nil {
```
rfranzke: Can you only do the check if the shoot has no deletion timestamp? Otherwise, we might end up with stuck shoots as the `spec` is immutable for shoots in deletion.

timuthy: Nice catch.

rfranzke (dismissed):
/lgtm

timuthy:
Sorry @rfranzke, I accidentally pushed to the wrong branch which took away your approval.

rfranzke (approved):

timebertt (approved):
/lgtm

PR #2764, opened 2020-08-20
Forward user information to GAPI in local setup

rfranzke:
See https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/#kubernetes-apiserver-client-authentication

/area dev-productivity
/kind bug
/priority normal

Configure the kube-apiserver correctly to forward the user information to the Gardener API server (otherwise, it will always use the proxy client certificate for authenticating with GAPI).

/cc @timuthy 

**Release note**:

```improvement developer
The kube-apiserver used for the local garden development does now properly forward the client information to the gardener-apiserver.
```


timuthy (approved):
/lgtm
Awesome. Thank you!

Issue #2754, opened 2020-08-19
☂️-Issue for elimination of Helm charts in favor of Golang implementation


rfranzke:
**How to categorize this issue?**

/area dev-productivity robustness quality testing
/kind enhancement
/priority 3

**What would you like to be added**:
This is an umbrella issue for improving our existing code, mainly in the [`operation`](https://github.com/gardener/gardener/tree/master/pkg/operation) and [`botanist`](https://github.com/gardener/gardener/tree/master/pkg/operation/botanist) packages. 
We lack unit and/or integration tests at too many places and should improve this. One important pain point is that the data flow is too distributed, many values are duplicated in Helm charts (e.g., labels, ports, names, ...) which makes it very hard to introduce a change as this might break unexpectedly.

One proposal is to leverage the recently introduced component interfaces (#1980) and eliminate the Helm charts in favor of Golang structs (see #2731). Another proposals is (at the point of time writing this issue) currently evaluated by @timebertt and discussed in two days.

Once agreed upon the preferred approach how to proceed, we can replicate it for the majority of the task functions in the [shoot reconciliation and deletion flows](https://github.com/gardener/gardener/tree/master/pkg/gardenlet/controller/shoot).

Agreements:
- With the node-local-dns refactoring, we experienced that it can be cumbersome to compare for equality the serialized obj in from the ManagedResource secrets. The order of annotations in the serialized obj was changing based on the ConfigMap data. While other components like vpnshoot, coredns rather have static ConfigMaps/Secrets and the annotations keys and the order of annotations is not changing that often. That's why for node-local-dns we decided to compare the obj (the DaemonSet) using golang struct. If a component hits similar issue to the node-local-dns, it can use compare the obj using golang struct (first deserialize the obj, then compare it to the expected one).

# Tasks

The following Helm charts still need to be refactored:

- `shoot-core/components`:
  - [x] https://github.com/gardener/gardener/pull/7547
  - [x] https://github.com/gardener/gardener/pull/7401
  - [x] https://github.com/gardener/gardener/pull/6134
  - [x] [`monitoring/blackbox-exporter`](https://github.com/gardener/gardener/tree/master/charts/shoot-core/components/charts/monitoring/charts/blackbox-exporter) https://github.com/gardener/gardener/pull/7978
[...]

PR #2766, opened 2020-08-21
Refactor kube-scheduler deployment but keep charts

timebertt:

/kind enhancement
/priority normal
/topology seed
/platform all
/exp intermediate

This PR is a sketch of a possible approach to refactor our current handling of values for helm charts and to introduce tests for them.
It is based on #2731 which proposes a different approach for tackling these two issues.

Ref #2754

This PR is not finished, some parts of the code would need to be cleaned if we want to take this approach.
It rather aims at getting some feeling of how this idea would look like in code as a basis for discussion and decision and not at proposing a perfect and clean refactoring.

**Release note**:

```improvement operator
NONE
```


timebertt:
We evaluated both proposed approaches and weighted the pros and cons.
At the end we decided in favor of the approach implemented in #2731.
/close

PR #2770, opened 2020-08-21
Fix tryUpdate utility functions to retry on conflict as expected

### LINKED ISSUE #2454: TryUpdate and TryUpdateStatus not working as expected when conflict occurs.

/area robustness
/kind bug
/priority normal

**What happened**:

Using the `TryUpdateStatus` function from `pkg/utils/kubernetes/update.go` discovered
an issue that was preventing the resource to be updated as expected. The issue is preventing `status.state` from update if a conflict occurs, when using `tryUpdate()` . For example:
For the first call of `tryUpdateStatus` we have the obj `worker.status.state=nil`.
And in tryUpdate we get the object from the server, transform it so `worker.status.state=something`, we check the object for changes before and after the transform and if it has changed we try to update. If a conflict occurs we fail, but our in memory object has `worker.status.state=something`.
So on the next iteration we get the object from API Server, which still has `worker.status.state=nil` in the server, but the object in memory is with `worker.status.state=something` (for some reason client.Get() merges it). On the next step we transform it which doesn't change anything and test if the transformed object is different from the one received from the server. They are now equal so we don't update and simply return.

`status.state` to be updated as expected.

Concurrently update an object using the `TryUpdate*` functions from `pkg/utils/kubernetes` package.

**Environment**:

- Gardener version: master

swilen-iwanow:

/area quality
/kind bug
/priority normal

Adds runtime.Scheme to tryUpdate functions API, in order to get the actual object from the server.
Fixes #2454

**Release note**:

```improvement operator
An issue causing TryUpdate and TryUpdateStatus funcs to do not retry on conflict in a corner case is now resolved.
```

### DIFF
--- a/extensions/pkg/controller/utils.go
+++ b/extensions/pkg/controller/utils.go
@@ -224,6 +224,7 @@ func tryUpdate(ctx context.Context, backoff wait.Backoff, c client.Client, obj r
 		return err
 	}
 
+	resetCopy := obj.DeepCopyObject()
 	return exponentialBackoff(ctx, backoff, func() (bool, error) {
 		if err := c.Get(ctx, key, obj); err != nil {
 			return false, err
@@ -240,6 +241,7 @@ func tryUpdate(ctx context.Context, backoff wait.Backoff, c client.Client, obj r
 
 		if err := updateFunc(ctx, obj); err != nil {
 			if apierrors.IsConflict(err) {
+				reflect.ValueOf(obj).Elem().Set(reflect.ValueOf(resetCopy).Elem())
 				return false, nil
 			}
 			return false, err
--- a/pkg/utils/kubernetes/update.go
+++ b/pkg/utils/kubernetes/update.go
@@ -43,6 +43,7 @@ func tryUpdate(ctx context.Context, backoff wait.Backoff, c client.Client, obj r
 		return err
 	}
 
+	resetCopy := obj.DeepCopyObject()
 	return exponentialBackoff(ctx, backoff, func() (bool, error) {
 		if err := c.Get(ctx, key, obj); err != nil {
 			return false, err
@@ -59,6 +60,7 @@ func tryUpdate(ctx context.Context, backoff wait.Backoff, c client.Client, obj r
 
 		if err := updateFunc(ctx, obj); err != nil {
 			if apierrors.IsConflict(err) {
+				reflect.ValueOf(obj).Elem().Set(reflect.ValueOf(resetCopy).Elem())
 				return false, nil
 			}
 			return false, err
--- a/pkg/utils/kubernetes/update.go
+++ b/pkg/utils/kubernetes/update.go
@@ -60,6 +60,11 @@ func tryUpdate(ctx context.Context, backoff wait.Backoff, c client.Client, obj r
 
 		if err := updateFunc(ctx, obj); err != nil {
 			if apierrors.IsConflict(err) {
+				// In case of a conflict we are resetting the obj to its original version, as it was
+				// passed to the function, to ensure that, on the next iteration the
+				// equality check of the obj recieved from the server and the object after
+				// its transformation would be valid. Otherwise the obj would be with mutated
+				// fields in result of the transform function from previous iteration.
 				reflect.ValueOf(obj).Elem().Set(reflect.ValueOf(resetCopy).Elem())
 				return false, nil
 			}
--- a/pkg/utils/kubernetes/update_test.go
+++ b/pkg/utils/kubernetes/update_test.go
@@ -1,0 +1,93 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kubernetes
+
+import (
+	"context"
+
+	"github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+
+	. "github.com/onsi/ginkgo"
+
+	. "github.com/onsi/gomega"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+var _ = Describe("#tryUpdate", func() {
+	It("should set state to obj, when conflict occurs", func() {
+		s := runtime.NewScheme()
+		Expect(extensionsv1alpha1.AddToScheme(s)).NotTo(HaveOccurred())
+		objInFakeClient := newInfraObj()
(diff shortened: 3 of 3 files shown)


swilen-iwanow:
@timebertt I am not very happy with the result, because now scheme has to be provided on a lot of levels, but couldn't think of a better solution. Anyway I ran some tests in isolation and it seems to work, now I have to run some manual tests to check for regressions.

swilen-iwanow:
cc @plkokanov @rfranzke

plkokanov:
@swilen-iwanow can you write a unit test for the TryUpdate methods which covers both the issue that is fixed with this PR and the normal workflow? 
To not pass the scheme as a function parameter you can add an init() method which combines all the schemes and use that in tryUpdate. Not sure how much I like this approach tho.

We could also probably write some code using reflection to copy the fields from the object that is retrieved from the server to the object which is used in `transform()`. Then we need to make sure that the two objects are of the same type.

I'll think a bit more on it.

swilen-iwanow:
I was thinking about the same approach, even hardcoding the schemes inside the function. Main concern was that it is a utility package and that way it is useless for client code outside gardener. If it is being used by extension controllers for example

pkg/utils/kubernetes/update.go
```diff
@@ -20,31 +20,50 @@ import (
...
+func TryUpdateStatus(ctx context.Context, backoff wait.Backoff, c client.Client, obj runtime.Object, scheme *runtime.Scheme, transform func() error) error {
+	return tryUpdate(ctx, backoff, c, obj, scheme, c.Status().Update, transform)
 }
 
-func tryUpdate(ctx context.Context, backoff wait.Backoff, c client.Client, obj runtime.Object, updateFunc func(context.Context, runtime.Object, ...client.UpdateOption) error, transform func() error) error {
+func tryUpdate(ctx context.Context, backoff wait.Backoff, c client.Client, obj runtime.Object, scheme *runtime.Scheme, updateFunc func(context.Context, runtime.Object, ...client.UpdateOption) error, transform func() error) error {
 	key, err := client.ObjectKeyFromObject(obj)
 	if err != nil {
 		return err
 	}
```
timuthy, line 44: What do you think about if we introduce something like this here:
```
resetCopy := obj.DeepCopyObject()

objGetter := func() runtime.Object {
  return resetCopy.DeepCopyObject()
}
```

The `objGetter` function can then be called in the `exponentialBackoff` part to get the object state before transformation in every iteration. I know that this is not the same as having a "fresh" object but it should still solve the issue explained in #2454.

timuthy (commented):
Thanks for the PR. As you said, passing the scheme every time seems inconvenient. I posted another suggestion that can solve the tryUpdate problem and I'm curious about your opinion.

pkg/utils/kubernetes/update.go
swilen-iwanow: I am not sure I understand what you suggest. But assuming you meant to do obj = objGetter() we are going to lose the pointer to the object from the caller function and it wouldn't be affected, right?

timuthy: Yes you are right, I missed that one. The idea was to "reset" the object if a conflict occurred.

```
reflect.ValueOf(obj).Elem().Set(reflect.ValueOf(resetCopy).Elem())
```

timebertt (commented):
Sorry for the delay.

I must admit, adding a scheme to the `TryUpdate` functions seems not to be a great option indeed, as we need to pass it now in every call and cannot properly default it.

I had a look again at the issue and it states:

> On the next step we transform it which doesn't change anything and test if the transformed object is different from the one received from the server. They are now equal so we don't update and simply return.

So I'm wondering if the problem is rather with the `reflect.DeepEqual` check. Maybe we should either
- save a copy of the passed object for comparison after mutation at the very beginning of the function (before we `Get` it from the API server). This way, we can properly detect the case you mentioned in the issue, but would send an update call if the mutation was already done (might happen very rarely and wouldn't hurt)
or
- add an option to the function to specify, that the object should always be updated regardless of the equality check
or
- use `client.Patch` just like in https://github.com/gardener/gardener/pull/2798 (probably the best option anyways for updating large objects, e.g. with large state date)

swilen-iwanow:
```use client.Patch just like in #2798 (probably the best option anyways for updating large objects, e.g. with large state date) ```
Do you mean patching even if the object is not actually different, I don't think this is a good idea.

**Comment by timebertt** [comment]
No, we don't have to send the patch every time even it does not differ from the current state.

File: pkg/utils/kubernetes/update.go (hunk 1 above)
Comment (swilen-iwanow) [reply pkg/utils/kubernetes/update.go:44]: I tried it like you proposed and it seems that it is working fine. @timebertt  are you ok with such solution

Comment (timuthy) [reply pkg/utils/kubernetes/update.go:44]: Nice, thanks for trying it out 👍

Comment (timebertt) [reply pkg/utils/kubernetes/update.go:44]: Yes, I'm fine with it. Seems like a good solution, thanks 👍 
I would still try to go for using Patch requests wherever possible, but that is a separate refactoring, which we don't have to do right now. :)

**Comment by timebertt** [comment]
I think, this PR is
/ready
now, right?

File: extensions/pkg/controller/utils.go [hunk 2]
```diff
@@ -240,6 +241,7 @@ func tryUpdate(ctx context.Context, backoff wait.Backoff, c client.Client, obj r
 
 		if err := updateFunc(ctx, obj); err != nil {
 			if apierrors.IsConflict(err) {
+				reflect.ValueOf(obj).Elem().Set(reflect.ValueOf(resetCopy).Elem())
```
timebertt, line 244: I discussed this approach with @rfranzke and we were wondering how this even works on a conflict.
Because now, you copy back the `resetCopy` to the `obj` and with it also the outdated `resourceVersion`.
Have you tested this?
Can you add a unit test checking the exact behaviour that you are trying to fix?

If this approach is not working, I would propose to go for the scheme approach described in https://github.com/gardener/gardener/issues/2454#issuecomment-656096285.
Maybe we can think of a way to make the scheme optional, so we don't have to adapt existing calls, where we don't see any issues with it.

timebertt (dismissed):

extensions/pkg/controller/utils.go
plkokanov: I was wondering the same thing today.
But after this "copy", we again `Get` the object from the server into `obj` which should update the resourceVersion. Or am I missing something.
Also /thumbsup for the unit test.

timuthy: > But after this "copy", we again Get the object from the server into obj which should update the resourceVersion. Or am I missing something.

I think this has ever happened since we introduced this tryUpdate procedure, hasn't it? In case of a conflict we always tried to refetch the object and thus overwrite the necessary fields of the very same referenced object.

swilen-iwanow: Yes, I used this https://gist.github.com/swilen-iwanow/9dee546b30cfc41e509b79a38692f511 to test it. 
Concurrently updating the status and verifying that the state is there, after a conflict has occurred. 

Reseting the object help us to start from the beginning and clear the transformation we did to the object which later messes the equality check.

Prior to this in case of a conflict we fetch the object but don't clear the transformation of the object so it is a "dirty" version of it because of the previous iteration. And after we transform the refetched "dirty" object and compare it to the new transformed they are equal and we return.

timebertt: > In case of a conflict we always tried to refetch the object and thus overwrite the necessary fields of the very same referenced object.

Ah, of course. We must have missed that.
Nevertheless, @rfranzke had some concerns regarding the reflection stuff. Can we neglect them or should we go with the safer approach (with schemes)?

> Yes, I used this https://gist.github.com/swilen-iwanow/9dee546b30cfc41e509b79a38692f511 to test it.
Concurrently updating the status and verifying that the state is there, after a conflict has occurred.

Perfect. Can you transform that into a unit test and add it to this PR?

rfranzke: @timebertt If the solution as it is currently works, should we then rather go for it instead of changing the functions signature with the scheme? The main argument would be that it's easier to use (although I still don't like reflection in general, but that's just a personal preference, so if it works as desired with this one additional line I can live with it 😉)

timebertt: Sounds good to me :)
Then let's go for this approach 🚀

timuthy: The only think I'd love to see @swilen-iwanow is a short comment about why this is necessary, since we discussed so much in this thread 😄

ialidzhikov (approved):
@swilen-iwanow , can you add a release note?

/lgtm

rfranzke (approved):

PR #2771, opened 2020-08-24
Add shoot reference protection for DNS secrets

### LINKED ISSUE #2738: Protect DNS provider secrets referenced by shoots

/area control-plane
/kind enhancement
/priority normal

It should be forbidden to delete DNS provider secrets in the project namespace if they are still referenced by any shoot cluster in the project.

timuthy:

/area robustness
/kind enhancement
/priority normal

This PR adds a new controller to GCM which scans shoot objects for referenced DNS provider secrets in order to protect them from deletion requests as long as they are actively being referenced. More information can be found in the contained documentation.

Fixes #2738

**Release note**:

```noteworthy user
Gardener now protects DNS provider secrets from deletion requests as long as they are still being referenced by one or multiple shoot clusters. More information can be found [in the documentation](https://github.com/gardener/gardener/blob/v1.12.0/docs/concepts/controller-manager.md#shoot-reference-controller).
```


pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -80,6 +80,9 @@ type ControllerManagerControllerConfiguration struct {
 	ShootQuota ShootQuotaControllerConfiguration `json:"shootQuota"`
 	// ShootHibernation defines the configuration of the ShootHibernation controller.
 	ShootHibernation ShootHibernationControllerConfiguration `json:"shootHibernation"`
+	// ShootReference defines the configuration of the ShootReference controller.
```
timebertt: Can you add a comment stating that this is defaulted, when not specified? So that operators know, that the controller will still be enabled, although it was omitted from the config. E.g.:
```suggestion
	// ShootReference defines the configuration of the ShootReference controller. If not specified, it will be defaulted with `concurrentSyncs=5`.
```

pkg/controllermanager/apis/config/v1alpha1/defaults.go
```diff
@@ -98,6 +98,12 @@ func SetDefaults_ControllerManagerConfiguration(obj *ControllerManagerConfigurat
 		}
 	}
 
+	if obj.Controllers.ShootReference == nil {
+		obj.Controllers.ShootReference = &ShootReferenceControllerConfiguration{
+			ConcurrentSyncs: 5,
+		}
```
timebertt, line 104: Can you add a test for this part [here](https://github.com/gardener/gardener/blob/aca4fd6bc7011f6d25a40ba8d6f29a0b6bc2443a/pkg/controllermanager/apis/config/v1alpha1/defaults_test.go#L37)?

pkg/controllermanager/controller/shoot/shoot.go
```diff
@@ -158,20 +165,26 @@ func (c *Controller) Run(ctx context.Context, shootMaintenanceWorkers, shootQuot
 		controllerutils.DeprecatedCreateWorker(ctx, c.configMapQueue, "ConfigMap", c.reconcileConfigMapKey, &waitGroup, c.workerCh)
 	}
 
+	for i := 0; i < shootReferenceWorkers; i++ {
+		controllerutils.CreateWorker(ctx, c.shootReferenceQueue, "ShootReference", NewShootReferenceReconciler(c.clientMap, logger.Logger), &waitGroup, c.workerCh)
```
timebertt: Why did you decide to create a new reconciler instance for each worker?

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -0,0 +1,213 @@
...
+		logger:    logger,
+	}
+}
+
+type shootReferenceReconciler struct {
+	clientMap clientmap.ClientMap
+	logger    logrus.FieldLogger
+}
+
+// Reconcile checks the shoot in the given request for references to further objects in order to protect the from
```
timebertt:
```suggestion
// Reconcile checks the shoot in the given request for references to further objects in order to protect them from
```

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -0,0 +1,213 @@
...
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, nil
+}
+
+func (r *shootReferenceReconciler) reconcileShootReferences(ctx context.Context, shoot *gardencorev1beta1.Shoot) error {
+	gardenClient, err := r.clientMap.GetClient(ctx, keys.ForGarden())
+	if err != nil {
+		return fmt.Errorf("failed to get garden client: %w", err)
+	}
```
timebertt: Can you pass the client instead of `Get`ting it again?
```suggestion
func (r *shootReferenceReconciler) reconcileShootReferences(ctx context.Context, c client.Client, shoot *gardencorev1beta1.Shoot) error {
```

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -0,0 +1,213 @@
...
+	return nil
+}
+
+func getUnreferencedSecretNames(ctx context.Context, gardenClient client.Client, shoot *gardencorev1beta1.Shoot) ([]string, error) {
+	namespace := shoot.Namespace
+
+	secrets := &corev1.SecretList{}
+	if err := gardenClient.List(ctx, secrets, client.InNamespace(namespace)); err != nil {
+		return nil, err
+	}
```
timebertt: As discussed on the phone, I am slightly worried about the performance aspect of looping over all Secrets in the Shoot namespace for every update of a given Shoot.
So let's
- use two separate queues/reconcilers for Shoots and Secrets to not fan-out compute effort on each Shoot update and rather enqueue referenced secrets
- only enqueue Shoots, if the relevant spec section changed (for now `spec.dns.providers`)

timebertt (changes requested):
Thanks for the PR!
Some smaller suggestions here and a short note about what we discussed on the phone.
/assign

pkg/controllermanager/controller/shoot/shoot_reference_control.go
timuthy: I had a second thought about this and think the solution with a separate queue/reconciler is still elegant but also intricate since the "protection mechanism" is distributed. When the GCM restarts we also have to iterate over every secret in the cluster of all namespaces which may be a large amount, too.

What we can alternately do, is to stick to the logic of this PR but instead of just adding a finalizer to secrets we also label them at the same time. We can then make the reconciler request only the labelled secrets in the project namespace, instead of requesting all. WDYT?

timebertt: That's a valid concern, got it.
What about only enqueuing Secrets if they have a deletion timestamp (for both add and update)? Then we wouldn't need to iterate over all secrets on startup and only act on the ones we care about (deleting secrets without finalizers will only cause a delete event and no update event, IIRC).
We will need to iterate over all Shoots on startup anyways and we can then enqueue the referenced Secrets into the Secret queue and the secret reconciler will then take over setting/removing the finalizer. WDYT?

timuthy: Any ad-hoc concerns about the labelling approach?

timebertt: One concern would be, that users might remove the label (e.g. by using `k replace` instead of `k apply`).
If we rely on that label being present for releasing the secret, we might miss some secrets we should release.

timuthy:
Several updates and changes since the last review round:
- Gardener managed secrets in the Garden cluster (SSH Keypair, Kubeconfig, Monitoring) are now labelled with `gardener.cloud/role`.
- The shoot reference controller disregards secrets with `gardener.cloud/role` to decrease the amount of iterated secrets and in favor of adding another secret controller like proposed from @timebertt (discussed in internal meeting).
- Finalizers are now patched instead of updated.
- In case of deletion, finalizers are only removed if shoot the shoot was successfully deleted by Gardener (no `gardener` finalizer).
- The Gardener client is passed to shoot related controller of `GCM` in favor of the whole ClientMap which reduces error handling.
- `CreateWorker` now injects the passed context and further objects into the `reconcile.Reconciler` implementation.

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -0,0 +1,250 @@
...
+	if err := r.releaseUnreferencedSecrets(shoot); err != nil {
+		return err
+	}
+
+	// Remove finalizer from shoot in case it's being deleted and not handled by Gardener any more.
+	if shoot.DeletionTimestamp != nil && !controllerutils.HasFinalizer(shoot, gardencorev1beta1.GardenerName) {
+		return controllerutils.PatchRemoveFinalizers(r.ctx, r.gardenClient, shoot, FinalizerName)
+	}
+
+	// Add finalizer to user managed and referenced secrets.
```
danielfoehrKn:
```suggestion
	// Add finalizer to referenced secrets that are not managed by Gardener.
```

To make it more clear that finalizers are added to referenced secrets if they are not managed by Gardener. 
First, I thought that the finalizers are added to two different set of secrets (1) user managed 2) referenced secrets).

danielfoehrKn (commented):
> CreateWorker now injects the passed context and further objects into the reconcile.Reconciler implementation

Could you explain a bit more why the client has to be injected and not just handed over like for the [maintenance control](https://github.com/gardener/gardener/blob/65977627c1bf12dbc32dec1a7754ff4024de650b/pkg/controllermanager/controller/shoot/shoot_maintenance_control.go#L131) ? 
Is that to be consistent with the injection of the context injection or is there more to it?

timuthy:
> Is that to be consistent with the injection of the context injection or is there more to it?

The idea was to align it more with the reconcilers we wrote for the extensions and the Controller-Runtime in general. Handing it over instead would have worked as well.

pkg/controllermanager/controller/factory.go
```diff
@@ -127,7 +127,7 @@ func (f *GardenControllerFactory) Run(ctx context.Context) {
 		projectController                = projectcontroller.NewProjectController(f.clientMap, f.k8sGardenCoreInformers, f.k8sInformers, f.cfg, f.recorder)
 		secretBindingController          = secretbindingcontroller.NewSecretBindingController(f.clientMap, f.k8sGardenCoreInformers, f.k8sInformers, f.recorder)
 		seedController                   = seedcontroller.NewSeedController(f.clientMap, f.k8sGardenCoreInformers, f.k8sInformers, f.cfg, f.recorder)
-		shootController                  = shootcontroller.NewShootController(f.clientMap, f.k8sGardenCoreInformers, f.k8sInformers, f.cfg, f.recorder)
+		shootController                  = shootcontroller.NewShootController(k8sGardenClient, f.k8sGardenCoreInformers, f.k8sInformers, f.cfg, f.recorder)
```
timebertt: Why did you change this?
The client map is designed to be queried at the begin of every operation where clients are needed. So if the reason for it is only to reduce explicit error handling, I would argue that we should rather keep it consistent across all controllers.

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -0,0 +1,250 @@
...
+	return nil
+}
+
+// InjectClient implements `inject.Stoppable`.
+func (r *shootReferenceReconciler) InjectStopChannel(stopCh <-chan struct{}) error {
+	r.ctx = contextutil.FromStopChannel(stopCh)
+	return nil
+}
+
+// Reconcile checks the shoot in the given request for references to further objects in order to protect them the from
```
timebertt:
```suggestion
// Reconcile checks the shoot in the given request for references to further objects in order to protect them from
```

pkg/controllerutils/finalizers.go
```diff
@@ -30,6 +30,28 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// PatchFinalizers adds the given finalizers to the object via a patch request.
+func PatchFinalizers(ctx context.Context, c client.Client, obj kutil.Object, finalizers ...string) error {
+	beforePatch := obj.DeepCopyObject()
+
+	f := sets.NewString(obj.GetFinalizers()...)
+	f.Insert(finalizers...)
+	obj.SetFinalizers(f.UnsortedList())
```
timebertt: I think, we should be careful with the order of the finalizers here, when sending patch requests.
Can you use `controllerutil.AddFinalizer` here?

pkg/controllerutils/finalizers.go
```diff
@@ -30,6 +30,28 @@ import (
...
+
+	return c.Patch(ctx, obj, client.MergeFrom(beforePatch))
```
timebertt: Should we rather use a locked patch request?
```suggestion
	return c.Patch(ctx, obj, client.MergeFromWithOptions(beforePatch, client.MergeFromWithOptimisticLock{}))
```

pkg/controllerutils/finalizers.go
```diff
@@ -30,6 +30,28 @@ import (
...
+
+// PatchRemoveFinalizers removes the given finalizers from the object via a patch request.
+func PatchRemoveFinalizers(ctx context.Context, c client.Client, obj kutil.Object, finalizers ...string) error {
+	beforePatch := obj.DeepCopyObject()
+
+	f := sets.NewString(obj.GetFinalizers()...)
+	f.Delete(finalizers...)
+	obj.SetFinalizers(f.UnsortedList())
+
+	return c.Patch(ctx, obj, client.MergeFrom(beforePatch))
```
timebertt: Same as above

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -0,0 +1,250 @@
...
+func (r *shootReferenceReconciler) getUnreferencedSecrets(shoot *gardencorev1beta1.Shoot) ([]*corev1.Secret, error) {
+	namespace := shoot.Namespace
+
+	secrets := &corev1.SecretList{}
+	if err := r.gardenClient.List(r.ctx, secrets, client.InNamespace(namespace), UserManagedSelector); err != nil {
+		return nil, err
+	}
+
+	shoots := &gardencorev1beta1.ShootList{}
+	if err := r.gardenClient.List(r.ctx, shoots, client.InNamespace(namespace)); err != nil {
```
timebertt: I must say, I'm still a bit worried about doing all of these calls for all Shoots on startup (can be a lot of calls, and might lead to hitting some rate limits again, if the cached clients are not enabled).
Can you use a lister here?

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -0,0 +1,250 @@
...
+			if err := r.gardenClient.Get(ctx, kutil.Key(s.Namespace, name), secret); err != nil {
+				return err
+			}
+
+			// Don't handle Gardener managed secrets.
+			if _, ok := secret.Labels[v1beta1constants.GardenRole]; ok {
+				return nil
+			}
+
+			added = true
```
timebertt: This can easily lead to a data race. Can you use `Once` or some other thread-safe mechanism?
The race detector should be able to catch this, so can you add a test with multiple referenced secrets as well?

timebertt (dismissed):
Left you some more comments inline :)

pkg/controllerutils/finalizers.go
timuthy: Why is the order relevant?

pkg/controllerutils/finalizers.go
timuthy: What's the benefit?

pkg/controllermanager/controller/factory.go
timuthy: Imo, it's confusing that the GCM deals with the client map at many places although the only client it can really get is the garden client, for other clients it will always error. Passing the garden client directly not only reduces the error handling but also simplifies the usage within the controller functions. I'd additionally vote to do the same for alle GCM controllers.

timebertt: Well, the Plant controller also uses the client map for retrieving a Plant client. Also the client map is responsible for refreshing the client's server version (used in some charts to check which API versions should be used).
IISIC, gcm doesn't depend on this in its charts, but that's an easy point to miss.

pkg/controllerutils/finalizers.go
timebertt: It also sends the `resourceVersion` with the request, just like with an update request, which will be denied by the API server, if the `resourceVersion` was outdated.
I was wondering, if we should do it when sending critical patches (like updating the finalizers), but I am also not so sure about it myself.

pkg/controllerutils/finalizers.go
timebertt: I think, `json-patch`es on arrays work with indices. That means, that switching the order of the finalizers, might lead to patches that don't result in the desired operation.

timuthy: I double checked and actually I think we can use patch safely in combination with `client.MergeFromWithOptimisticLock{}` because the element order during runtime is safe with `f.UnsortedList()`. WDYT? Falling back to update calls is still an option, I still don't want to miss a chance to pave the path towards patch request.

timebertt: I'm also still in favor of using Patch requests here. 
Why do you think, the element order is safe at runtime? I would expect, that when inserting the slice elements into a map (like in `sets.NewString`) and then range over it to insert them into a slice again (like in `Set.UnsortedList`) the order will not always be the same (as maps are unsorted).
So I guess, the safer option wrt to order would be doing something like in the c-r helper function (`controllerutil.AddFinalizer`).

timuthy: Okay makes sense. Thanks for explaining.

timuthy:
Thanks for your valuable feedback. I addressed all your comments.

cmd/gardener-controller-manager/app/gardener_controller_manager.go
```diff
@@ -352,7 +356,7 @@ func (g *Gardener) Run(ctx context.Context, cancel context.CancelFunc) error {
 		g.LeaderElection.Callbacks = leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(_ context.Context) {
 				g.Logger.Info("Acquired leadership, starting controllers.")
-				run(ctx)
+				utilruntime.Must(run(ctx))
```
timebertt: Any specific reason why you panic on error here and not log the error just like in the gardenlet?

timebertt (commented):
Thanks for addressing my feedback!
Looks great now, only one question :)

cmd/gardener-controller-manager/app/gardener_controller_manager.go
timuthy: No specific reason, I changed it to log.

timebertt (dismissed):
Thanks!
/lgtm

danielfoehrKn (dismissed):
/lgtm thanks!

rfranzke (commented):
No comments, just one question: Will all of this work if the new Gardener version (with this change) is deployed but there are still shoots which weren't reconciled yet (i.e., shoots whose ssh-keypair/kubeconfig secrets don't have the role labels yet)?

timuthy:
> No comments, just one question: Will all of this work if the new Gardener version (with this change) is deployed but there are still shoots which weren't reconciled yet (i.e., shoots whose ssh-keypair/kubeconfig secrets don't have the role labels yet)?

Good question @rfranzke! I'll work but of course the reconciler then needs to cache and iterate over a lot more secrets in the beginning. Are you proposing to split the PR up and release it in two iterations?

rfranzke:
I'm not proposing anything yet :) Just wanted to raise this question. If you think this is not a problem then let's get it in as is.

timuthy:
Well, thanks @rfranzke! To be on the safe side, I opened #2844 which should then be released first after this PR gets merged.

timuthy:
/do-not-merge and wait for #2844 to be released.

pkg/utils/kubernetes/object.go
```diff
@@ -18,17 +18,10 @@ import (
 	"context"
 
 	"k8s.io/apimachinery/pkg/api/meta"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// Object is wrapper interface combining runtime.Object and metav1.Object interfaces together.
-type Object interface {
```
rfranzke, line 27: Why was this removed? If I remember correctly, at least some extensions might use it.

timuthy: I removed it in favor of the interface from the Controller-Runtime, see [example](https://github.com/gardener/gardener/pull/2771/files#diff-9d372c6b0cd8943fb3fc9cbaaebdbc2dR36).

rfranzke: OK, thanks :)

rfranzke (approved):
/lgtm

timebertt (approved):
Thanks!
/lgtm

PR #2772, opened 2020-08-24
Cleanup leader election clients/recorders

timebertt:

/area quality
/kind cleanup
/priority normal
/size xs

This PR cleans up two thing related to leader election in our components:
- gcm uses a separate client for leader election again according to best practice
- gardenlet reports leader election events to the Seed cluster instead of the garden cluster (which is confusing/incorrect IMO)

Fixes #

**Release note**:

```improvement operator
The gardenlet now reports leader election events to the Seed cluster instead of the Garden cluster.
```

### DIFF
--- a/cmd/gardener-controller-manager/app/gardener_controller_manager.go
+++ b/cmd/gardener-controller-manager/app/gardener_controller_manager.go
@@ -47,6 +47,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/informers"
 	kubeinformers "k8s.io/client-go/informers"
+	kubernetesclientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/leaderelection"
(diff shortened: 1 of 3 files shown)


rfranzke:
What's the motivation for 

> g-scheduler doesn't create an additional client for leader election

timebertt:
> What's the motivation for
> 
> > g-scheduler doesn't create an additional client for leader election

No specific motivation. I just saw, that we also use the same client in gcm, so I changed it along the way.

rfranzke:
Interesting, I missed that during the client map refactoring. Earlier, we already had dedicated clients (that's why it's also the same for the kube-scheduler), see https://github.com/gardener/gardener/blob/v1.6.6/cmd/gardener-controller-manager/app/gardener_controller_manager.go#L256-L259

AFAIK that's a best practice in Kubernetes controllers to prevent the leader election contributing to the client-go's client-side rate limiting settings.

timebertt:
> AFAIK that's a best practice in Kubernetes controllers to prevent the leader election contributing to the client-go's client-side rate limiting settings.

Makes sense to me. Then I would suggest to revert this change and have separate clients in all components

rfranzke:
👍 Thank you

timebertt:
> Then I would suggest to revert this change and have separate clients in all components

Done. Added back a separate client for leader election in the gcm.
PTAL :)

rfranzke (approved):
/lgtm

Thank you!

PR #2775, opened 2020-08-24
use ext4 filesystem with project quota support

### LINKED ISSUE #2774: use ext4 instead of xfs in kubelet data volume

/area os
/area security
/kind enhancement
/priority normal

Currently, the Kubelet Data Volume is formatted with the `xfs` filesystem. The motivation for using xfs was project quota support which can be used for efficient [ephemeral storage quota management](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20180906-quotas-for-ephemeral-storage.md), as described here #1893. Gardenlinux currently does not support `xfs`. So, this feature is not usable in the the standard gardener cluster setup. 

Seeing as project quota support was also added to the `ext4` filesystem, it makes sense to update this feature and use the more common `ext4` filesystem instead, to ensure better portability. 

**Why is this needed**:
Enables encryption of kubelet-managed state (eg. ephemeral storage) for providers that do not support encryption of root volumes (such as AliCloud) in gardenlinux-based nodes.

guydc:

/area os
/kind enhancement
/priority normal

Fixes #2774 

**Release note**:

```improvement user
Workers that configure a kubelet data volume can now use gardenlinux OS
```


rfranzke (approved):
/lgtm

Issue #1650, opened 2019-11-23
Quota for user facing resources


timuthy:
**What would you like to be added**:
Gardener should have configurable, project scoped quotas for resources that end users (specifically project members) are permitted to create:

```
- apiGroups:
  - ""
  resources:
  - secrets
  - configmaps
  - serviceaccounts

- apiGroups:
  - garden.sapcloud.io
  - core.gardener.cloud
  resources:
  - shoots
  - secretbindings
  - quotas
  - plants

- apiGroups:
  - settings.gardener.cloud
  resources:
  - openidconnectpresets

- apiGroups:
  - garden.sapcloud.io
  - core.gardener.cloud
  resources:
  - projects
```

The list above contains all possible resources, although some of them might be negligible or aren't critical for the system stability (tbd).

A quota could look like:

```
apiVersion: core.gardener.cloud/v1alpha1
kind: ResourceQuota
metadata:
  name: standard-quota
  namespace: garden-project
spec:
- apiGroups:
  - garden.sapcloud.io
  - core.gardener.cloud
  resources:
  - shoots
  - secretbindings
  - quotas
  - plants
  limits:
    maxObjects: 200
    maxSizePerObject: 1m 
```

Unfortunately, k8s native `ResourceQuotas` don't cover all necessary API objects and uses cases:
https://kubernetes.io/docs/tasks/administer-cluster/quota-api-object/

**Why is this needed**:
Quotas are one countermeasure to prevent the Gardener- and Kube-Apiserver, it's backing etcds as well as involving controllers from being flooded.

/cc @petersutter @ThormaehlenFred

swilen-iwanow:
I also think, that it is reasonable gardener admins to be able to specify a quota for creating projects. Otherwise the project API could be abused.

rfranzke:
It might be interesting to look into whether `ResourceQuota`s are possible to be reused for object counts of our Gardener-managed resources (https://kubernetes.io/docs/concepts/policy/resource-quotas/#object-count-quota only states "custom resources"). Is it possible?

For the object size there is https://github.com/kubernetes/kubernetes/pull/83261. We could revendor `k8s.io/apiserver` to `v0.16.7` to get this `3MB` limit: https://github.com/kubernetes/apiserver/blob/v0.16.7/pkg/server/config.go#L298-L311 (cc @ialidzhikov).

timuthy:
[...]

PR #2389, opened 2020-05-26
Nit: Specify sideEffects for validate-kubeconfig-secrets webhook

ialidzhikov:
**How to categorize this PR?**

/area usability
/kind enhancement
/priority normal

**What this PR does / why we need it**:
Currently trying kubectl apply with `--dry-run=server` and a cloud provider secret:
```
$ k apply -f cloudprovider-secret.yaml --dry-run=server
Error from server (BadRequest): error when creating "cloudprovider-secret.yaml": admission webhook "validate-kubeconfig-secrets.gardener.cloud" does not support dry run
```

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
Ref https://github.com/gardener/gardener-extension-provider-aws/pull/89

**Release note**:

```improvement operator
NONE
```

PR #2781, opened 2020-08-26
Add resource size validator

timuthy:

/area security
/kind enhancement
/priority normal

This PR adds the Resource Size Validator which is a webook for GCM and can be used to restrict the size of resources applied by end users. Please consult the enclosed documentation `docs/concepts/controller-manager.md` for more details.

Fixes parts of #1650

**Release note**:

```noteworthy operator
The Gardener Controller Manager is now equipped with a validation handler which checks incoming resource requests against configured quota configurations. It especially enables operators to restrict the maximum size of a single resource (e.g. shoot, plant, secret, ...) users apply to the Garden cluster and is at the same time a measure against DoS attacks. Please consult the documentation `docs/concepts/controller-manager.md#Resource-Size-Validator` for more details.
```


docs/concepts/controller-manager.md
```diff
@@ -72,4 +72,54 @@ The `--event-ttl` should be larger than the `ttlNonShootEvents` or this controll
 
 ## Webhooks
 
-ℹ️ This document is incomplete and under construction.
+While the Gardener API server works with [admission plugins](./apiserver_admission_plugins.md) to validate and mutate resources belonging to Gardener related API groups, e.g. `core.gardener.cloud`, the same is needed for resources with arbitrary API groups as well.
```
```suggestion
While the Gardener API server works with [admission plugins](./apiserver_admission_plugins.md) to validate and mutate resources belonging to Gardener related API groups, e.g. `core.gardener.cloud`, the same is needed for resources belonging to non-Gardener API groups as well.
```

docs/concepts/controller-manager.md
```diff
@@ -72,4 +72,54 @@ The `--event-ttl` should be larger than the `ttlNonShootEvents` or this controll
...
+### Kubeconfig Secret Validator
+
+[Malicious Kubeconfigs](https://github.com/kubernetes/kubectl/issues/697) applied by end users may cause a leakage of sensitive data.
+This handler checks if the incoming request contains a Kubernetes secret with a `.data.kubeconfig` field and denies the request if the Kubeconfig structure violates Gardener's security standards.
+
+### Namespace Validator
+
+Namespaces are the backing entities of Gardener projects in which shoot clusters objects reside.
+This validation handler protects active namespaces against premature deletion requests.
+Therefore, it denies deletion requests if a namespace still contains shoot clusters or if it belongs to a non-deleting Gardener project (project w/o `.meta.deletionTimestamp`).
```
```suggestion
Therefore, it denies deletion requests if a namespace still contains shoot clusters or if it belongs to a non-deleting Gardener project (w/o `.metadata.deletionTimestamp`).
```

docs/concepts/controller-manager.md
```diff
@@ -72,4 +72,54 @@ The `--event-ttl` should be larger than the `ttlNonShootEvents` or this controll
...
+
+### Resource Size Validator
+
+Since users directly apply Kubernetes native objects to the Garden cluster, it also involves the risk of being susceptible to DoS attacks because these resources are again read by further controllers in the Gardener landscape.
```
```suggestion
Since users directly apply Kubernetes native objects to the Garden cluster, it also involves the risk of being vulnerable to DoS attacks because these resources are read by controllers in the Gardener landscape.
```

docs/concepts/controller-manager.md
```diff
@@ -72,4 +72,54 @@ The `--event-ttl` should be larger than the `ttlNonShootEvents` or this controll
...
+    unrestrictedUsers:
+    - groupName: "gardener.cloud:system:seeds"
+      userNames: ["*"]
+    operationMode: block #log
+```
+
+With the configuration above, the Resource Size Validator denies requests for shoots and plants with Gardener's core API group which exceed a size of 100 kB. The same is done for Kubernetes secrets.
+
+As this feature is meant to protect the system from malicious requests sent by users, it is recommended to exclude trusted groups and users from the size restriction via `resourceAdmissionConfiguration.unrestrictedUsers`.
+For example, the backing user for the Gardenlet should always be capable of changing the shoot resource instead of being the one that locks itself out due to size restrictions.
```
```suggestion
For example, the backing user for the Gardenlet should always be capable of changing the shoot resource instead of being blocked due to size restrictions.
```

docs/concepts/controller-manager.md
```diff
@@ -72,4 +72,54 @@ The `--event-ttl` should be larger than the `ttlNonShootEvents` or this controll
...
+This is because the Gardenlet itself occasionally changes the shoot specification, labels or annotations and might violate against the quota if the existing resource is already close to the quota boundary.
```
```suggestion
This is because the Gardenlet itself occasionally changes the shoot specification, labels or annotations, and might violate the quota if the existing resource is already close to the quota boundary.
```

docs/concepts/controller-manager.md
```diff
@@ -72,4 +72,54 @@ The `--event-ttl` should be larger than the `ttlNonShootEvents` or this controll
...
+Also, operators are supposed to be trusted users and subjecting them to a size limitation can inhibit important operational tasks.
+
+Size limitations depend on the individual Gardener setup and choosing the wrong values can affect the availability of your Gardener service.
+`resourceAdmissionConfiguration.operationMode` allows to control if a violating request is actually denied (default) or only logged.
+It's recommended to start with `log`, check the logs for exceeding requests, adjust the limits if necessary and finally switch to `block`.
```
rfranzke: Nice documentation!

hack/local-development/dev-setup-register-gardener
```diff
@@ -252,5 +252,70 @@ $CONTROLLER_MANAGER_PORT_STRING
...
+- name: validate-resource-size-gardener.gardener.cloud
```
rfranzke, line 285: Why do we have this "twice", and two times with different client config? Can't it be combined into one block?

pkg/controllermanager/apis/config/types.go
```diff
@@ -202,6 +203,44 @@ type ServerConfiguration struct {
 	HTTP Server
 	// HTTPS is the configuration for the HTTPS server.
 	HTTPS HTTPSServer
+	// ResourceAdmissionConfiguration is the configuration for the resource admission.
+	ResourceAdmissionConfiguration *ResourceAdmissionConfiguration
+}
+
+// ResourceAdmissionConfiguration contains settings about arbitrary kinds and the size each resource should have at most.
+type ResourceAdmissionConfiguration struct {
+	Limits []ResourceLimit
```
rfranzke: doc string missing

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -221,6 +222,49 @@ type ServerConfiguration struct {
 	HTTP Server `json:"http"`
 	// HTTPS is the configuration for the HTTPS server.
 	HTTPS HTTPSServer `json:"https"`
+	// ResourceAdmissionConfiguration is the configuration for the resource admission.
```
```suggestion
	// ResourceAdmissionConfiguration is the configuration for the resource admission.
	// +optional
```

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -221,6 +222,49 @@ type ServerConfiguration struct {
...
+	ResourceAdmissionConfiguration *ResourceAdmissionConfiguration `json:"resourceAdmissionConfiguration,omitempty"`
+}
+
+// ResourceAdmissionConfiguration contains settings about arbitrary kinds and the size each resource should have at most.
+type ResourceAdmissionConfiguration struct {
+	Limits []ResourceLimit `json:"limits,omitempty"`
```
rfranzke: doc string missing

pkg/controllermanager/apis/config/validation/ControllerManagerConfiguration.go
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: Can you pease rename the filename to all lowercase letters?

pkg/controllermanager/server/handlers/webhooks/validate_resource_size.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	if isUnrestrictedUser(request.UserInfo, h.config.UnrestrictedUsers) {
+		return admissionResponse(true, "")
+	}
+
+	requested := &request.Resource
+	if request.RequestResource != nil {
+		requested = request.RequestResource
+	}
```
rfranzke: Can you add an inline comment what is going on here/why this is needed?

pkg/controllermanager/server/handlers/webhooks/validate_resource_size.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+		}
+		if groupFound && confighelper.UserNameMatches(userConfig, userInfo.Username) {
+			return true
+		}
+	}
+	return false
+}
+
+func matchingLimit(limits []apisconfig.ResourceLimit, gvr *metav1.GroupVersionResource) *resource.Quantity {
```
rfranzke:
```suggestion
func findLimitForGVR(limits []apisconfig.ResourceLimit, gvr *metav1.GroupVersionResource) *resource.Quantity {
```

rfranzke (changes requested):
/assign
Love it - very clean, easy-to-follow and easy-to-read PR, thanks a lot.

charts/gardener/controlplane/charts/application/templates/validatingwebhook-controller-manager.yaml
```diff
@@ -53,5 +53,34 @@ webhooks:
...
+    url: https://gardener-controller-manager.garden/webhooks/validate-resource-size
+    {{- else }}
+    service:
+      namespace: garden
+      name: gardener-controller-manager
+      path: /webhooks/validate-resource-size
+    {{- end }}
+    caBundle: {{ required ".Values.global.controller.config.server.https.tls.caBundle is required" (b64enc .Values.global.controller.config.server.https.tls.caBundle) }}
+{{- end }}
+sideEffects: None
```
ialidzhikov:
```suggestion
  sideEffects: None
```

charts/gardener/controlplane/charts/application/templates/validatingwebhook-controller-manager.yaml
```diff
@@ -53,5 +53,34 @@ webhooks:
       path: /webhooks/validate-kubeconfig-secrets
     {{- end }}
     caBundle: {{ required ".Values.global.controller.config.server.https.tls.caBundle is required" (b64enc .Values.global.controller.config.server.https.tls.caBundle) }}
-  sideEffects: None
```
ialidzhikov: We could keep this line. Ref https://github.com/gardener/gardener/pull/2389

docs/concepts/controller-manager.md
```diff
@@ -72,4 +72,54 @@ The `--event-ttl` should be larger than the `ttlNonShootEvents` or this controll
...
+### Resource Size Validator
+
+Since users directly apply Kubernetes native objects to the Garden cluster, it also involves the risk of being susceptible to DoS attacks because these resources are again read by further controllers in the Gardener landscape.
+One example is the creation of shoot clusters with large annotation values (up to 256 kB per value) which can cause severe out-of-memory issues for the Gardenlet component.
+[Vertical autoscaling](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) can help to mitigate such situations, but we cannot expect to scale infinitely and thus need means to block the attack itself.
+
+The Resource Size Validator checks arbitrary incoming admission requests against a configured maximum size for the resource's group-version-kind combination and denies the request if the contained object exceeds the quota.
+
+Example for Gardener Controller Manager configuration:
+```
```
ialidzhikov:
```suggestion
```yaml
```

pkg/controllermanager/apis/config/types.go
```diff
@@ -202,6 +203,44 @@ type ServerConfiguration struct {
...
+
+	// UnrestrictedUsers contains configuration for users and group which aren't subjected to any resource size limit.
+	UnrestrictedUsers []UserConfig
+	// Mode specifies the mode the webhooks operates in, e.g. "block", "audit"
```
ialidzhikov:
```suggestion
	// OperationMode specifies the mode the webhooks operates in. Allowed values are "block" and "audit". Defaults to "block".
```

pkg/controllermanager/apis/config/validation/ControllerManagerConfiguration.go
```diff
@@ -0,0 +1,95 @@
...
+	for i, userConfig := range config.UnrestrictedUsers {
+		fld := fldPath.Child("unrestrictedUsers").Index(i)
+		if userConfig.GroupName == "" {
+			allErrs = append(allErrs, field.Invalid(fld.Child("groupName"), userConfig.GroupName, "group name must not be empty"))
+		}
+		hasUsers := false
+		for j, userName := range userConfig.UserNames {
+			hasUsers = true
+			if userName == "" {
+				allErrs = append(allErrs, field.Invalid(fld.Child("userNames").Index(j), userName, "group name must not be empty"))
```
ialidzhikov:
```suggestion
				allErrs = append(allErrs, field.Invalid(fld.Child("userNames").Index(j), userName, "user name must not be empty"))
```

docs/concepts/controller-manager.md
```diff
@@ -72,4 +72,54 @@ The `--event-ttl` should be larger than the `ttlNonShootEvents` or this controll
...
+### Namespace Validator
+
+Namespaces are the backing entities of Gardener projects in which shoot clusters objects reside.
+This validation handler protects active namespaces against premature deletion requests.
+Therefore, it denies deletion requests if a namespace still contains shoot clusters or if it belongs to a non-deleting Gardener project (project w/o `.meta.deletionTimestamp`).
+
+### Resource Size Validator
+
+Since users directly apply Kubernetes native objects to the Garden cluster, it also involves the risk of being susceptible to DoS attacks because these resources are again read by further controllers in the Gardener landscape.
+One example is the creation of shoot clusters with large annotation values (up to 256 kB per value) which can cause severe out-of-memory issues for the Gardenlet component.
```
ialidzhikov: Correct me if I am wrong but the resource size webhook now uses `ioutil.ReadAll` to read the request body. So if you can cause OOM issues to gardenlet with similar requests, then now you would be able to cause them to the gardener-controller-manager. Then webhook itself has `failurePolicy=Fail`. If gardener-controller-manager is running OOM, then gardenlet or any other component cannot create/update Shoots which means that you end up with almost similar lockdown. I think it makes sense to check also that case.

hack/local-development/dev-setup-register-gardener
timuthy: One client config is for the KAS and the other one for GAS which needs to be different because it usually runs as a local process during development. Do you see another way to implement this?

docs/concepts/controller-manager.md
timuthy: We don't cover the full story, I agree with you. We shouldn't simply call `decodeAdmissionRequest` but rather have a general object size restriction which we can check for violation first. What do you think about `1.5 Mi` which is in most of the cases also the limit of etcd values?
In addition, it's probably reasonable to move out the webhooks from GCM in the future.

hack/local-development/dev-setup-register-gardener
rfranzke: OK, got it, thank you

docs/concepts/controller-manager.md
timuthy: I added a general limit of `3 Mi` per request which is the same request size KAS uses.

timuthy:
Thanks for your rewiews @rfranzke @ialidzhikov. I already addressed your comments but as discussed in our internal meeting, I'll continue to move out the webhooks to a dedicated component so that we have:
- multiple replicas for this dedicated component
- a clear separation

timuthy:
PR has been rebased on #2832. Please consider a98e6ee for review.
/hold

rfranzke:
Please update the PR now that #2832 is merged :)

charts/gardener/controlplane/values.yaml
```diff
@@ -225,6 +225,16 @@ global:
               -----BEGIN RSA PRIVATE KEY-----
               ...
               -----END RSA PRIVATE KEY-----
+    # resourceAdmissionConfiguration:
+    #   limits:
+    #   - apiGroups: ["core.gardener.cloud"]
+    #     apiVersions: ["*"]
+    #     resources: ["shoots", "plants"]
+    #     size: 100 Ki
```
rfranzke:
```suggestion
    #     size: 100Ki
```

docs/concepts/admission-controller.md
```diff
@@ -12,4 +12,42 @@ This handler checks if the incoming request contains a Kubernetes secret with a
 
 Namespaces are the backing entities of Gardener projects in which shoot clusters objects reside.
 This validation handler protects active namespaces against premature deletion requests.
-Therefore, it denies deletion requests if a namespace still contains shoot clusters or if it belongs to a non-deleting Gardener project (w/o `.metadata.deletionTimestamp`).
\ No newline at end of file
+Therefore, it denies deletion requests if a namespace still contains shoot clusters or if it belongs to a non-deleting Gardener project (w/o `.metadata.deletionTimestamp`).
+
+### Resource Size Validator
+
+Since users directly apply Kubernetes native objects to the Garden cluster, it also involves the risk of being vulnerable to DoS attacks because these resources are read by controllers in the Gardener landscape.
```
rfranzke:
```suggestion
Since users directly apply Kubernetes native objects to the Garden cluster, it also involves the risk of being vulnerable to DoS attacks because these resources are continuously watched and read by controllers.
```

docs/concepts/admission-controller.md
```diff
@@ -12,4 +12,42 @@ This handler checks if the incoming request contains a Kubernetes secret with a
...
+One example is the creation of shoot clusters with large annotation values (up to 256 kB per value) which can cause severe out-of-memory issues for the Gardenlet component.
```
rfranzke:
```suggestion
One example is the creation of `Shoot` resources with large annotation values (up to 256 kB per value) which can cause severe out-of-memory issues for the Gardenlet component.
```

docs/concepts/admission-controller.md
```diff
@@ -12,4 +12,42 @@ This handler checks if the incoming request contains a Kubernetes secret with a
...
+[Vertical autoscaling](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) can help to mitigate such situations, but we cannot expect to scale infinitely and thus need means to block the attack itself.
```
rfranzke:
```suggestion
[Vertical autoscaling](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) can help to mitigate such situations, but we cannot expect to scale infinitely, and thus need means to block the attack itself.
```

docs/concepts/admission-controller.md
```diff
@@ -12,4 +12,42 @@ This handler checks if the incoming request contains a Kubernetes secret with a
...
+    limits:
+    - apiGroups: ["core.gardener.cloud"]
+      apiVersions: ["*"]
+      resources: ["shoots", "plants"]
+      size: 100k
+    - apiGroups: [""]
+      apiVersions: ["v1"]
+      resources: ["secrets"]
+      size: 100k
+    unrestrictedUsers:
```
rfranzke: I'm wondering whether it makes sense to use the `[]rbacv1.Subject` type here. WDYT?

pkg/admissioncontroller/apis/config/helper/helper_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestHelper(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Gardener Controller Manager API Helper Suite")
```
rfranzke:
```suggestion
	RunSpecs(t, "Gardener Admission controller API Helper Suite")
```

pkg/admissioncontroller/apis/config/v1alpha1/types.go
```diff
@@ -37,6 +38,50 @@ type AdmissionControllerConfiguration struct {
...
 	HTTPS HTTPSServer `json:"https"`
+	// ResourceAdmissionConfiguration is the configuration for the resource admission.
+	// +optional
+	ResourceAdmissionConfiguration *ResourceAdmissionConfiguration `json:"resourceAdmissionConfiguration,omitempty"`
+}
+
+// ResourceAdmissionConfiguration contains settings about arbitrary kinds and the size each resource should have at most.
+type ResourceAdmissionConfiguration struct {
+	// Limits contains configuration for resources which are subjected to size limitations.
+	Limits []ResourceLimit `json:"limits,omitempty"`
```
rfranzke: Why `omitempty`?

pkg/admissioncontroller/apis/config/v1alpha1/types.go
```diff
@@ -37,6 +38,50 @@ type AdmissionControllerConfiguration struct {
...
+	OperationMode *ResourceAdmissionWebhookMode `json:"operationMode,omitempty"`
+}
+
+// ResourceAdmissionWebhookMode is an alias type for the resource admission webhook mode.
+type ResourceAdmissionWebhookMode string
+
+// UserConfig contain user und group information of unrestricted users.
+type UserConfig struct {
+	// GroupName specifies the name of the group. Wildcard "*" matches all groups.
+	GroupName string `json:"groupName,omitempty"`
```
rfranzke: Why `omitempty`?

pkg/admissioncontroller/apis/config/validation/controllerManagerConfiguration.go
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: please fix the filename

pkg/admissioncontroller/apis/config/validation/controllerManagerConfiguration_test.go
```diff
@@ -0,0 +1,165 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: please fix the filename

pkg/admissioncontroller/apis/config/validation/controllerManagerConfiguration_test.go
```diff
@@ -0,0 +1,165 @@
...
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
+	gomegatypes "github.com/onsi/gomega/types"
+	"k8s.io/apimachinery/pkg/api/resource"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+)
+
+var _ = Describe("#ValidateAdmissionControllerConfiguration", func() {
+
```
rfranzke:
```suggestion
```

pkg/admissioncontroller/apis/config/validation/validation_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"testing"
+)
+
+func TestValidation(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Gardener Controller Manager Config API Validation Suite")
```
rfranzke:
```suggestion
	RunSpecs(t, "Gardener Admission Controller Config API Validation Suite")
```

pkg/admissioncontroller/server/handlers/webhooks/webhooks_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestWebhooks(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Gardener Controller Manager Webhooks Suite")
```
rfranzke:
```suggestion
	RunSpecs(t, "Gardener Admission Controller Webhooks Suite")
```

rfranzke (changes requested):

docs/concepts/admission-controller.md
timuthy: Yes, good idea. I switched to it in 87272f1165ddfde470844ce7934a026bf6af1033.

timuthy:
/unhold
I addressed the feedback in 87272f1. PTAL.

docs/concepts/admission-controller.md
```diff
@@ -12,4 +12,51 @@ This handler checks if the incoming request contains a Kubernetes secret with a
...
+    operationMode: block #log
+```
+
+With the configuration above, the Resource Size Validator denies requests for shoots and plants with Gardener's core API group which exceed a size of 100 kB. The same is done for Kubernetes secrets.
+
+As this feature is meant to protect the system from malicious requests sent by users, it is recommended to exclude trusted groups, users or service accounts from the size restriction via `resourceAdmissionConfiguration.unrestrictedSubjects`.
+For example, the backing user for the Gardenlet should always be capable of changing the shoot resource instead of being blocked due to size restrictions.
+This is because the Gardenlet itself occasionally changes the shoot specification, labels or annotations, and might violate the quota if the existing resource is already close to the quota boundary.
+Also, operators are supposed to be trusted users and subjecting them to a size limitation can inhibit important operational tasks.
+Kinds `ServiceAccount` and `Group` additionally allow wildcards ("*") for their names.
```
rfranzke: Why only those two kinds and not `User` as well?

pkg/admissioncontroller/apis/config/validation/admissionControllerConfiguration.go
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: fix file name, ref https://stackoverflow.com/questions/25161774/what-are-conventions-for-filenames-in-go

pkg/admissioncontroller/apis/config/validation/admissionControllerConfiguration_test.go
```diff
@@ -0,0 +1,186 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: fix file name, ref https://stackoverflow.com/questions/25161774/what-are-conventions-for-filenames-in-go

rfranzke (changes requested):

rfranzke (dismissed):
/lgtm

timuthy:
Thanks @rfranzke, I force pushed again to squash the commits.

pkg/admissioncontroller/server/handlers/webhooks/validate_resource_size.go
```diff
@@ -0,0 +1,171 @@
...
+	respond(w, reviewResponse)
+}
+
+// admit does only allow the request if the object in the request does not exceed a configured limit.
+func (h *objectSizeHandler) admit(request *admissionv1beta1.AdmissionRequest, logEntry logrus.FieldLogger) *admissionv1beta1.AdmissionResponse {
+	if request.SubResource != "" {
+		return admissionResponse(true, "")
+	}
+
+	if isUnrestrictedUser(request.UserInfo, h.config.UnrestrictedSubjects) {
```
ialidzhikov, line 99: This would panic with nil pointer if `resourceAdmissionConfiguration` is not configured in the component config. We could fail early at component start that `resourceAdmissionConfiguration` is required or we can add nil check.

pkg/admissioncontroller/apis/config/types.go
```diff
@@ -62,3 +94,10 @@ type TLSServer struct {
 	// ServerKeyPath is the path to the private key file.
 	ServerKeyPath string
 }
+
+var (
```
ialidzhikov:
```suggestion
const (
```

pkg/admissioncontroller/server/handlers/webhooks/utils.go
```diff
@@ -56,3 +61,43 @@ func respond(w http.ResponseWriter, response *admissionv1beta1.AdmissionResponse
...
+			return err
+		}
+		if lr.N <= 0 {
+			return apierrors.NewRequestEntityTooLargeError(fmt.Sprintf("limit is %d", limit))
+		}
+		body = data
+	}
+
+	// Verify that the correct content-type header has been sent.
+	if contentType := r.Header.Get("Content-Type"); contentType != wantedContentType {
```
ialidzhikov, line 88: Hmm, does that mean that the webhook will fail if someone is trying to create a Shoot with `application/vnd.kubernetes.protobuf`?

pkg/admissioncontroller/server/handlers/webhooks/utils_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	runtimeserializer "k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/apimachinery/pkg/runtime/serializer/versioning"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	kubernetesscheme "k8s.io/client-go/kubernetes/scheme"
+)
+
+const (
+	noSubResource = ""
+	emptyMessage  = ""
```
ialidzhikov, line 47: They are currently only used in `pkg/admissioncontroller/server/handlers/webhooks/validate_resource_size.go`. You could move them there.

timuthy: I added it here as a preparation for adding further tests for the namespace and kubeconfig secret webhook.

pkg/admissioncontroller/server/handlers/webhooks/utils.go
timuthy: I don't think so because the API server sends the request to webhook (https://github.com/kubernetes/kubernetes/blob/e36e68f5f6964a3e612b36bfea0bb17c8f05e083/staging/src/k8s.io/apiserver/pkg/util/webhook/client.go#L150), not the client.

pkg/admissioncontroller/server/handlers/webhooks/validate_resource_size.go
timuthy: Nice catch! I now default the configuration if it's not set.

timuthy:
Thanks @ialidzhikov for the feedback. PTAL.

ialidzhikov (approved):
/lgtm

PR #2782, opened 2020-08-26
Fix context cancel for garbage collection in care controller

danielfoehrKn:

/area quality
/kind bug
/priority normal

Fixes Garbage collection problems in the Shoot care controller such as

```
Could not initialise Shoot client for garbage collection of shoot garden-dev/aws-glinux2: Get \"...../deployments/kube-apiserver\": context canceled
```

Introduces a dedicated  context for the garbage collection in the care controller. 

Currently the context to initialise the shoot clients gets canceled after the health checks are finished. The garbage collection uses the very same context underneath when initialising its clients.

So if the health checks complete before the garbage collection starts, garbage collection fails to create its clients due to an already canceled context.  
This happens often for hibernated clusters where the health checks complete right away.

Fixes #

This is I think a different issue than reported in #2689
**Release note**:

```improvement operator
Fixes a bug that could cause Shoot garbage collection to fail in certain cases.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_care_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_care_control.go
@@ -235,9 +235,6 @@ func (c *defaultCareControl) Care(shootObj *gardencorev1beta1.Shoot, key string)
 
 	initializeShootClients := shootClientInitializer(ctx, botanist)
 
-	// Trigger garbage collection
-	go garbageCollection(ctx, initializeShootClients, botanist)
-
 	_ = flow.Parallel(
 		// Trigger health check
 		func(ctx context.Context) error {
@@ -264,6 +261,12 @@ func (c *defaultCareControl) Care(shootObj *gardencorev1beta1.Shoot, key string)
 		// Trigger constraint checks
 		func(ctx context.Context) error {
 			constraintHibernationPossible = botanist.ConstraintsChecks(ctx, initializeShootClients, constraintHibernationPossible)
+			return nil
+		},
+		// Trigger garbage collection
+		func(ctx context.Context) error {
+			garbageCollection(ctx, initializeShootClients, botanist)
+			// errors during garbage collection are only being logged and do not cause the care operation to fail
 			return nil
 		},
 	)(ctx)


timuthy:
I'm concerned about what we do here in general because we start a go routine w/o waiting for it to exit and with this change we sure get rid of the error message but foster this zombie routine even more, don't you agree?

timebertt:
I agree with @timuthy. This change does not solve anything from my perspective and rather fosters new problems.

But I agree, that we should fix the issue, that the context used for garbage collection is cancelled when the health checks return, but the GC has not finished yet. I think, this should be fixed by calling the garbage collection in the same manner as the health checks, i.e. via `flow.Parallel`.

danielfoehrKn:
> I'm concerned about what we do here in general because we start a go routine w/o waiting for it to exit and with this change we sure get rid of the error message but foster this zombie routine even more, don't you agree?

The starting of the go routine was already done before? The only thing that changes is, that the garbage collection go routine has a separate context (with also the same timeout).
So I do not see why this change is making it worse?

> This change does not solve anything from my perspective and rather fosters new problems

Which ones that not already exist?

> I think, this should be fixed by calling the garbage collection in the same manner as the health checks, i.e. via flow.Parallel

Do you mean in the same flow as the health checks? This way we would couple the garbage collection to the health checks. In theory I have no problem with that besides that the health conditions might be update a bit later sometimes (or never if the garbage collection fails).

timebertt:
> Do you mean in the same flow as the health checks? This way we would couple the garbage collection to the health checks.

Yes exactly. I should have already done this when I introduced the context back then...


> the health conditions might be update a bit later sometimes (or never if the garbage collection fails).

I'm not sure, if I get that point. Can you elaborate?

timuthy:
> So I do not see why this change is making it worse?

Earlier the context was eliminated after the `Care()` function returned, now it continues to run in the background (until timeout), so in the meantime we start another `Care()` operation and run another zombie goroutine and so on and so on.

danielfoehrKn:
OK thanks guys makes sense.

danielfoehrKn:
I draft this PR because I have no time next week to work on it. Will reopen later.

timebertt (approved):
Thanks, this makes much more sense now!
/lgtm

timuthy (approved):
Thanks

PR #2783, opened 2020-08-27
Move kube-scheduler related resources into kubescheduler pkg

rfranzke:

/area dev-productivity testing quality
/kind enhancement
/priority normal

After #2731, there were still resources/code related to kube-scheduler scattered in other packages/places. With this PR, we move them into the `kubescheduler` package as well. This includes

* Resources created in the shoot (RBAC roles for Kubernetes 1.13)
* Prometheus scrape configuration
* AlertManager alerting rules
* fluent-bit parsers and filters

Please note that the only things it does not include are the Grafana dashboards. @wyb1 might have some ideas how to achieve the same for them in the future.

Part of #2754

/assign @timebertt @wyb1 @Kristian-ZH @vlvasilev 

**Release note**:

```improvement operator
NONE
```


charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-configmap.yaml
```diff
@@ -364,7 +358,7 @@ data:
         LabelMapPath /fluent-bit/etc/kubernetes_label_map.json
         DynamicHostPath {"kubernetes": {"namespace_name": "namespace"}}
         DynamicHostPrefix http://loki.
-        DynamicHostSulfix .svc:3100/loki/api/v1/push
+        DynamicHostSuffix .svc:3100/loki/api/v1/push
```
Kristian-ZH: This will not work with the current fluent-bit version.
Was a typo which this huge PR of @vlvasilev is fixing: https://github.com/gardener/logging/pull/64 :D

wyb1 (commented):
Prometheus cannot start due to invalid config.
```yaml
    - job_name: kube-scheduler
      scheme: https
      tls_config:
        insecure_skip_verify: true
        cert_file: /etc/prometheus/seed/prometheus.crt
        key_file: /etc/prometheus/seed/prometheus.key
      honor_labels: false
      kubernetes_sd_configs:
      - role: endpoints
        namespaces:
          names: [shoot--myns--test-shoot]
      relabel_configs:
      - source_labels:
        - __meta_kubernetes_service_name
        - __meta_kubernetes_endpoint_port_name
        action: keep
        regex: kube-scheduler;metrics
      - action: labelmap
        regex: __meta_kubernetes_service_label_(.+)
      - source_labels: [ __meta_kubernetes_pod_name ]
        target_label: pod
      metric_relabel_configs:
      - scheduler_binding_latency_microseconds_bucket
      - scheduler_e2e_scheduling_latency_microseconds_bucket
      - scheduler_scheduling_algorithm_latency_microseconds_bucket
      - rest_client_requests_total
      - process_max_fds
      - process_open_fds
```
This is the rendered config but the `metric_relabel_configs` content is invalid. It should look like this:
```yaml
      metric_relabel_configs:
      - source_labels: [ __name__ ]
        regex: ^(scheduler_binding_latency_microseconds_bucket|scheduler_scheduling_algorithm_latency_microseconds_bucket|scheduler_scheduling_algorithm_latency_microseconds_bucket|rest_client_requests_total|process_max_fds|loki_ingester_chunk_bounds_hours_count|process_open_fds)$

File: charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-configmap.yaml (hunk 1 above)
Comment (rfranzke) [reply charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-configmap.yaml]: hehe, ok, I'll revert it. It was a best guess anyways :)

**Comment by rfranzke** [comment]
@wyb1 Thanks, strange, I haven't experienced it. I will take a look.

**Comment by rfranzke** [comment]
@wyb1 fixed, PTAL

File: charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-configmap.yaml (hunk 1 above)
Comment (rfranzke) [reply charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-configmap.yaml]: I reverted it, PTAL

**Review by vlvasilev — DISMISSED** [review]
/lgtm

File: test/gomega/matchers.go [hunk 2]
```diff
@@ -26,13 +26,24 @@ import (
 func DeepDerivativeEqual(expected interface{}) types.GomegaMatcher {
 	// if CharactersAroundMismatchToInclude is to small, then format.MessageWithDiff will be unable to output our
```
timebertt:
```suggestion
	// if CharactersAroundMismatchToInclude is too small, then format.MessageWithDiff will be unable to output our
```

test/gomega/matchers.go
```diff
@@ -26,13 +26,24 @@ import (
...
 	}
 
-	return &deepDerivativeMatcher{
-		expected: expected,
+	return newDeepDerivativeMatcher(expected)
+}
+
+// DeepEqual is similar to DeepEqual except that unset fields in actual are
+// ignored (not compared). This allows us to focus on the fields that matter to
+// the semantic comparison.
```
timebertt: Copy+paste mistake?

test/gomega/matchers.go
```diff
@@ -26,13 +26,24 @@ import (
 func DeepDerivativeEqual(expected interface{}) types.GomegaMatcher {
```
timebertt: Site note: We should probably combine all the custom matchers, that are commonly used and currently scattered across g/g, in a single package. I think `pkg/utils/test` would be more suitable than `test/gomega`.
But let's do this in a separate PR.

charts/seed-monitoring/charts/core/charts/prometheus/rules-tests/kube-scheduler.rules.test.yaml
```diff
@@ -1,23 +0,0 @@
-rule_files:
-- ../rules/kube-scheduler.rules.yaml
-
-evaluation_interval: 30s
-
-tests:
```
timebertt: These tests don't have a replacement, right?
Are we planning on having such a replacement for those tests? Is it worth the effort?
/cc @wyb1 @istvanballok

charts/seed-monitoring/charts/core/charts/prometheus/values.yaml
```diff
@@ -31,6 +31,10 @@ replicas: 1
 apiserverServiceIP: 100.10.10.10
 port: 9090
 
+components:
```
timebertt: These values are unused, IISC.
You simply append the scheduler configs to the same string builders as the extension configs in `pkg/operation/botanist/monitoring.go`.

wyb1 (dismissed):
/lgtm

wyb1 (commented):
There is still something wrong with the prometheus config. I am trying to figure out what it is

pkg/operation/botanist/controlplane/kubescheduler/monitoring.go
```diff
@@ -0,0 +1,136 @@
...
+    - __meta_kubernetes_endpoint_port_name
+    action: keep
+    regex: ` + ServiceName + `;` + PortNameMetrics + `
+  - action: labelmap
+    regex: __meta_kubernetes_service_label_(.+)
+  - source_labels: [ __meta_kubernetes_pod_name ]
+    target_label: pod
+  metric_relabel_configs:
+  - source_labels: [ __name__ ]
+    regex: ^(` + monitoringMetricSchedulerBindingLatencyMicrosecondsBucket + `|` + monitoringMetricSchedulerE2eSchedulingLatencyMicrosecondsBucket + `|` + monitoringMetricSchedulerSchedulingAlgorithmLatencyMicrosecondsBucket + `|` + monitoringMetricRestClientRequestsTotal + `|` + monitoringMetricProcessMaxFds + `|` + monitoringMetricProcessOpenFds + `)$`
```
wyb1:
```suggestion
    action: keep
    regex: ^(` + monitoringMetricSchedulerBindingLatencyMicrosecondsBucket + `|` + monitoringMetricSchedulerE2eSchedulingLatencyMicrosecondsBucket + `|` + monitoringMetricSchedulerSchedulingAlgorithmLatencyMicrosecondsBucket + `|` + monitoringMetricRestClientRequestsTotal + `|` + monitoringMetricProcessMaxFds + `|` + monitoringMetricProcessOpenFds + `)$`
```
Default action is replace which causes this to be invalid.

pkg/operation/botanist/monitoring.go
```diff
@@ -60,7 +60,19 @@ func (b *Botanist) DeploySeedMonitoring(ctx context.Context) error {
 		usersDashboards     = strings.Builder{}
 	)
 
-	// Find extensions provider-specific monitoring configuration
+	// Fetch component-specific monitoring configuration
+	for _, componentFn := range []func() (string, string, error){
+		b.Shoot.Components.ControlPlane.KubeScheduler.MonitoringConfiguration,
```
timebertt: Can we rather have a clear interface for components, that want to contribute to the monitoring configuration?

Maybe the interface could then have 4 functions similar to the keys of the configmaps that extensions use to contribute config:
```go
type MonitoringComponent interface {
     ScrapeConfig() (string, error)
     AlertConfig() (string, error)
     DashboardsOperators() (string, error)
     DashboardsUsers() (string, error)
}
```
Or something similar. Maybe some return types could rather be `map[string]string` so we can then use `utils.MergeStringMaps()` instead of writing YAML formatted strings to string builders.

Maybe this also makes sense for the Logging part.
WDYT?

timebertt (changes requested):
I haven't look through all changes, but I have some early comments for you :)

charts/seed-monitoring/charts/core/charts/prometheus/rules-tests/kube-scheduler.rules.test.yaml
wyb1: The tests are useful to verify that the alerts can actually fire and to help understand the alert. Is there a way we can keep the tests?

rfranzke:
@timebertt @wyb1 please take another look, I've incorporated your suggestions
Thanks @timebertt for the collaboration!

/unassign
/remove status/author-action
/needs review

pkg/operation/botanist/controlplane/kubescheduler/monitoring.go
```diff
@@ -0,0 +1,143 @@
...
+    action: keep
+    regex: ` + ServiceName + `;` + PortNameMetrics + `
+  - action: labelmap
+    regex: __meta_kubernetes_service_label_(.+)
+  - source_labels: [ __meta_kubernetes_pod_name ]
+    target_label: pod
+  metric_relabel_configs:
+  - source_labels: [ __name__ ]
+    action: keep
+    regex: ^(` + monitoringMetricSchedulerBindingLatencyMicrosecondsBucket + `|` + monitoringMetricSchedulerE2eSchedulingLatencyMicrosecondsBucket + `|` + monitoringMetricSchedulerSchedulingAlgorithmLatencyMicrosecondsBucket + `|` + monitoringMetricRestClientRequestsTotal + `|` + monitoringMetricProcessMaxFds + `|` + monitoringMetricProcessOpenFds + `)$`
```
wyb1: wouldn't it be better to have a slice of whitelisted metrics and then join them with `|`?
```suggestion
    regex: ^(` + strings.Join(whitelistedMetrics, "|") + `)$`
```

rfranzke: Yes, makes sense. I moved this out from the `const` block to the `var` block and implemented your suggestion.

pkg/operation/botanist/addons.go
```diff
@@ -287,38 +249,26 @@ func (b *Botanist) deployCloudConfigExecutionManagedResource(ctx context.Context
...
-	if err := b.deleteStaleSecretsMatchLabel(ctx, secretLabels, wantedSecretNames); err != nil {
-		return err
-	}
 
-	// TODO: remove this in the future
-	return b.deleteLegacyCloudConfigSecret(ctx)
+	return b.deleteStaleSecretsMatchLabel(ctx, secretLabels, wantedSecretNames)
 }
 
 // deleteStaleSecretsMatchLabel deletes the stale secrets that match the labels but are not used anymore
```
timebertt, line 274: Can you remove the `deleteLegacyCloudConfigSecret` func as well, as you removed the last usage?

pkg/operation/botanist/component/interfaces.go
```diff
@@ -47,12 +47,31 @@ type MigrateWaiter interface {
...
+// MonitoringComponent exposes configuration for Prometheus as well as the AlertManager.
+type MonitoringComponent interface {
+	// ScrapeConfig returns the scrape configuration for Prometheus.
+	ScrapeConfig() (string, error)
+	// AlertingRules returns the alerting rules for AlertManager.
+	AlertingRules() (map[string]string, error)
+}
+
+// LoggingConfiguration is a function alias for returning logging parsers and filters.
+type LoggingConfiguration func() (string, string, error)
```
timebertt, line 59: Why did you chose a function type alias over an interface?
Doing it via an interface would be consistent to `MonitoringComponent`, allow us to add more functions / logging config options to components later on and would allow explicitly declaring an implementation of the logging options on a component/component interface (kind of like a marker interface). WDYT?

pkg/operation/botanist/component/interfaces.go
```diff
@@ -47,12 +47,31 @@ type MigrateWaiter interface {
 	WaitMigrate(ctx context.Context) error
 }
 
+// MonitoringComponent exposes configuration for Prometheus as well as the AlertManager.
+type MonitoringComponent interface {
+	// ScrapeConfig returns the scrape configuration for Prometheus.
+	ScrapeConfig() (string, error)
+	// AlertingRules returns the alerting rules for AlertManager.
```
timebertt: Can you add an explanation for what is expected from the returned map?
```suggestion
	// AlertingRules returns the alerting rules configs for AlertManager (mapping file name to rule config).
```

pkg/operation/botanist/component/interfaces.go
```diff
@@ -47,12 +47,31 @@ type MigrateWaiter interface {
...
+
 // DeployWaiter controls and waits for life-cycle operations of a component.
 type DeployWaiter interface {
 	Deployer
 	Waiter
 }
 
+// ObservableDeployWaiter controls and waits for life-cycle operations of a component. It also exposes monitoring
+// configuration.
+type ObservableDeployWaiter interface {
```
timebertt: What's the benefit of this compound interface?
IMHO, I doesn't make sense to have interfaces for those combinations. Usually, a consumer of any component interface won't care, whether a component implements both the deployment and monitoring config functions. A single consumer will most probably only care about one aspect: either to deploy/destroy it, getting monitoring config from it or logging config, etc. but not all at once. I am right?

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -66,9 +78,9 @@ leaderElection:
   leaderElect: true`
 )
 
-// KubeScheduler contains function for a kube-scheduler deployer.
+// KubeScheduler contains functions for a kube-scheduler deployer.
 type KubeScheduler interface {
-	component.DeployWaiter
+	component.ObservableDeployWaiter
```
timebertt: See the comment above, I would declare implementation of both interfaces separately here, something like this:
```suggestion
	component.DeployWaiter
	component.MonitoringComponent
```

wyb1 (dismissed):
/lgtm

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -299,13 +311,79 @@ func (k *kubeScheduler) emptyVPA() *autoscalingv1beta2.VerticalPodAutoscaler {
...
+		return common.DeployManagedResource(ctx, k.client, managedResourceName, k.namespace, false, k.computeShootResourcesData())
+	}
+
+	for _, obj := range []runtime.Object{k.emptyManagedResource(), k.emptyManagedResourceSecret()} {
+		if err := k.client.Delete(ctx, obj); client.IgnoreNotFound(err) != nil {
+			return err
+		}
+	}
+
+	return nil
```
timebertt, line 340: I would suggest to use the `common.NewManagedResource*` funcs here instead of  `emptyManagedResource*`. This way, we won't have duplication on how the secret name is computed.

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -299,13 +311,79 @@ func (k *kubeScheduler) emptyVPA() *autoscalingv1beta2.VerticalPodAutoscaler {
...
+				Name: "system:controller:kube-scheduler",
+			},
+			RoleRef: rbacv1.RoleRef{
+				APIGroup: rbacv1.GroupName,
+				Kind:     "ClusterRole",
+				Name:     "system:auth-delegator",
+			},
+			Subjects: subjects,
+		}
+		clusterRoleBindingYAML, _ = runtime.Encode(codec, clusterRoleBinding)
```
timebertt, line 364: Why is the returned error ignored here?

pkg/operation/botanist/controlplane/kubescheduler/monitoring.go
```diff
@@ -0,0 +1,155 @@
...
+	monitoringMetricProcessOpenFds                                        = "process_open_fds"
+
+	monitoringAlertingRules = `groups:
+- name: kube-scheduler.rules
+  rules:
+  - alert: KubeSchedulerDown
+    expr: absent(up{job="` + monitoringPrometheusJobName + `"} == 1)
+    for: 15m
+    labels:
+      service: ` + ServiceName + `
```
timebertt: Nit: I would rather use the deployment name for this label than the service name.
```suggestion
      service: ` + v1beta1constants.DeploymentNameKubeScheduler + `
```

pkg/operation/botanist/controlplane/kubescheduler/monitoring.go
```diff
@@ -0,0 +1,155 @@
...
+	"text/template"
+
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+)
+
+const (
+	monitoringPrometheusJobName = "kube-scheduler"
+
+	monitoringMetricSchedulerBindingLatencyMicrosecondsBucket             = "scheduler_binding_latency_microseconds_bucket"
+	monitoringMetricSchedulerE2eSchedulingLatencyMicrosecondsBucket       = "scheduler_e2e_scheduling_latency_microseconds_bucket"
```
timebertt: Nit
```suggestion
	monitoringMetricSchedulerE2ESchedulingLatencyMicrosecondsBucket       = "scheduler_e2e_scheduling_latency_microseconds_bucket"
```

pkg/operation/botanist/controlplane/kubescheduler/monitoring.go
```diff
@@ -0,0 +1,155 @@
...
+    labels:
+      quantile: "0.9"
+  - record: cluster:scheduler_binding_latency_seconds:quantile
+    expr: histogram_quantile(0.5, sum(` + monitoringMetricSchedulerBindingLatencyMicrosecondsBucket + `) BY (le, cluster)) / 1e+06
+    labels:
+      quantile: "0.5"`
+)
+
+var (
+	monitoringsMetrics = []string{
```
timebertt: Only a suggestion.
```suggestion
	monitoringAllowedMetrics = []string{
```

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
```diff
@@ -332,6 +387,39 @@ var _ = Describe("KubeScheduler", func() {
 				Expect(kubeScheduler.Deploy(ctx)).To(MatchError(fakeErr))
 			})
 
+			It("should fail because the managed resource cannot be deleted", func() {
```
timebertt, line 390: Can you also add tests for the other cases (mr/secret not found and mr/secret can't be created on k8s 1.13)?

pkg/operation/botanist/controlplane/kubescheduler/monitoring.go
```diff
@@ -0,0 +1,155 @@
...
+    labels:
+      quantile: "0.99"
+  - record: cluster:scheduler_binding_latency_seconds:quantile
+    expr: histogram_quantile(0.9, sum(` + monitoringMetricSchedulerBindingLatencyMicrosecondsBucket + `) BY (le, cluster)) / 1e+06
+    labels:
+      quantile: "0.9"
+  - record: cluster:scheduler_binding_latency_seconds:quantile
+    expr: histogram_quantile(0.5, sum(` + monitoringMetricSchedulerBindingLatencyMicrosecondsBucket + `) BY (le, cluster)) / 1e+06
+    labels:
+      quantile: "0.5"`
```
timebertt: Can we have a newline at the end of every long string?
```suggestion
      quantile: "0.5"
`
```

pkg/operation/botanist/monitoring.go
```diff
@@ -60,7 +61,26 @@ func (b *Botanist) DeploySeedMonitoring(ctx context.Context) error {
...
+			return err
+		}
+		scrapeConfigs.WriteString(fmt.Sprintln(scrapeConfig))
+
+		alertingRule, err := component.AlertingRules()
+		if err != nil {
+			return err
+		}
+		for filename, rule := range alertingRule {
+			alertingRules.WriteString(fmt.Sprintf("%s: |\n  %s\n", filename, strings.ReplaceAll(rule, "\n", "\n  ")))
```
timebertt: Now, alerting rules and scrape config of the scheduler are passed to values under `extensions` which seems a bit off. 
Can you use separate values for that? I would also vote for using `map[string]string` for the components' `alertingRules` and merging maps instead of appending YAML-formatted strings. `map[string]string` can be passed to the values directly.

pkg/operation/botanist/monitoring.go
```diff
@@ -60,7 +61,26 @@ func (b *Botanist) DeploySeedMonitoring(ctx context.Context) error {
...
-	// Find extensions provider-specific monitoring configuration
+	// Fetch component-specific monitoring configuration
+	for _, component := range []component.MonitoringComponent{
+		b.Shoot.Components.ControlPlane.KubeScheduler,
+	} {
+		scrapeConfig, err := component.ScrapeConfig()
+		if err != nil {
+			return err
+		}
+		scrapeConfigs.WriteString(fmt.Sprintln(scrapeConfig))
```
timebertt: A second thought about `scrapeConfigs`:
Maybe it would be better to use `[]string` for collecting the scrape configs, so implementors of `MonitoringComponent` don't have to know about the correct YAML-indent which must be used. Instead we should iterate over the collected scrape configs in the template and add the array item marker (`- `) and the right indentation, so it can be handled centrally, instead of handling it in all components.

pkg/operation/common/managedresources_test.go
```diff
@@ -0,0 +1,106 @@
...
+		ctrl = gomock.NewController(GinkgoT())
+		c = mockclient.NewMockClient(ctrl)
+	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
+	Describe("#DeployManagedResource", func() {
+		It("should return the error of the secret reconciliation", func() {
```
timebertt, line 57: Can you also add tests for the case, that the secret/mr does not already exist?

pkg/operation/seed/seed.go
```diff
@@ -359,14 +360,27 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+		for _, componentFn := range []component.LoggingConfiguration{
+			kubescheduler.LoggingConfiguration,
+		} {
+			parser, filter, err := componentFn()
+			if err != nil {
+				return err
+			}
+
+			filters.WriteString(fmt.Sprintln(filter))
+			parsers.WriteString(fmt.Sprintln(parser))
```
timebertt, line 374: Same here as for the monitoring scrape/alert config. Can you move the stuff from the core components to values separate from the `extension` values?

timebertt (commented):
Great work!
I have some more comments (mainly nits and small suggestions). All in all, it looks quite good :)

pkg/operation/botanist/component/interfaces.go
rfranzke: The reason is that the logging stack is deployed as part of the seed controller's reconciliation loop while the kube-scheduler component is designed to be instantiated for a shoot. It felt wrong to create a `kubeScheduler` structure in the seed controller, hence, I decided for the type alias.

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
rfranzke: I'll call the `common.NewManagedResourceSecret` function inside the `emptyManagedResourceSecret` function to get the name without re-computation.

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
rfranzke: Because all the content of `clusterRoleBinding` is static/not dynamic/not dependent on any user input.

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
rfranzke: What do you mean exactly? Isn't this covered by the `success tests for various kubernetes versions` table?

pkg/operation/botanist/monitoring.go
rfranzke: > Now, alerting rules and scrape config of the scheduler are passed to values under extensions which seems a bit off.

I don't see any benefit of duplicating this part in the Prometheus Helm chart. From the chart's API point of view these things are just additional configurations, no matter where they are coming from. I can rename the value key from `extensions` to `additional{...}` to make that more clear in the chart itself, but I don't see any benefit of duplicating the rendering logic.

> I would also vote for using map[string]string for the components' alertingRules and merging maps instead of appending YAML-formatted strings. map[string]string can be passed to the values directly.

Can we do this as part of the Prometheus deployment refactoring? This PR is about the kube-scheduler and I don't want to mix too many things at once. I agree that the string formatting is odd, but we might have to keep it anyway (at least for a while) due to the current monitoring/logging extension contract. I'd vote for cleaning this up in a dedicated improvement track. For newly introduced components that implement the `MonitoringComponent` interface it is easy/clear enough, or?

pkg/operation/common/managedresources_test.go
rfranzke: OK, now I get what you have meant in your other comment above. I haven't added these "cases" intentionally as they don't belong to these functions here, IMO. They belong to unit tests for `CreateOrUpdate`, but these unit tests are about higher-level functions that call `CreateOrUpdate` internally. If we always multiply them then we will blow up our tests artificially without a lot of benefit, or? I can add them for sure, but where do you suggest to draw the line?

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
timebertt: Sorry, I forgot to delete this comment after I found the part in the table... 🙈

rfranzke: OK, I've added the tests for the "mr/secret can't be created on k8s 1.13".
About the first part see https://github.com/gardener/gardener/pull/2783#discussion_r484456098

pkg/operation/botanist/component/interfaces.go
timebertt: Hmm, missed that. Yes, that makes sense.

pkg/operation/common/managedresources_test.go
timebertt: Good point, agreed. What is tested now, is probably already enough.

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
timebertt: Well, maybe not dependent on user input, that's right. But when someone wants to change the schemes/gvks that are part of `kubernetes.ShootSerializer`, I might be easy to miss side-effects in dependent packages like this one, if we ignore the returned error silently.

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -299,13 +311,80 @@ func (k *kubeScheduler) emptyVPA() *autoscalingv1beta2.VerticalPodAutoscaler {
...
 func (k *kubeScheduler) emptyDeployment() *appsv1.Deployment {
 	return &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: v1beta1constants.DeploymentNameKubeScheduler, Namespace: k.namespace}}
 }
 
+func (k *kubeScheduler) emptyManagedResource() *resourcesv1alpha1.ManagedResource {
+	return &resourcesv1alpha1.ManagedResource{ObjectMeta: metav1.ObjectMeta{Name: managedResourceName, Namespace: k.namespace}}
+}
+
+func (k *kubeScheduler) emptyManagedResourceSecret() *corev1.Secret {
+	secretName, _ := common.NewManagedResourceSecret(k.client, managedResourceName, k.namespace, nil)
```
timebertt: Can you rather extract the name computation logic and use a dedicated function for that?

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
timebertt: Nevermind, we now have unit tests for that 😄

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
```diff
@@ -332,6 +387,88 @@ var _ = Describe("KubeScheduler", func() {
...
+					Expect(err).NotTo(HaveOccurred())
+
+					kubeScheduler = New(c, namespace, semverVersion, image, replicas, nil)
+					kubeScheduler.SetSecrets(Secrets{
+						Kubeconfig: Secret{Name: secretNameKubeconfig, Checksum: secretChecksumKubeconfig},
+						Server:     Secret{Name: secretNameServer, Checksum: secretChecksumServer},
+					})
+				})
+
+				It("should fail because the managed resource secret cannot be created (k8s 1.13 only)", func() {
```
timebertt: The test description is kind of confusing.
```suggestion
				It("should fail because the managed resource secret cannot be updated (k8s 1.13 only)", func() {
```

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler_test.go
```diff
@@ -332,6 +387,88 @@ var _ = Describe("KubeScheduler", func() {
...
+						c.EXPECT().Get(ctx, kutil.Key(namespace, deploymentName), gomock.AssignableToTypeOf(&appsv1.Deployment{})),
+						c.EXPECT().Update(ctx, gomock.AssignableToTypeOf(&appsv1.Deployment{})),
+						c.EXPECT().Get(ctx, kutil.Key(namespace, managedResourceSecretName), gomock.AssignableToTypeOf(&corev1.Secret{})),
+						c.EXPECT().Update(ctx, gomock.AssignableToTypeOf(&corev1.Secret{})).Return(fakeErr),
+					)
+
+					Expect(kubeScheduler.Deploy(ctx)).To(MatchError(fakeErr))
+				})
+
+				It("should fail because the managed resource cannot be created (k8s 1.13 only)", func() {
```
timebertt: Same here.
```suggestion
				It("should fail because the managed resource cannot be updated (k8s 1.13 only)", func() {
```

pkg/operation/botanist/monitoring.go
```diff
@@ -60,7 +61,28 @@ func (b *Botanist) DeploySeedMonitoring(ctx context.Context) error {
...
+	// Fetch component-specific monitoring configuration
+	for _, component := range []component.MonitoringComponent{
+		b.Shoot.Components.ControlPlane.KubeScheduler,
+	} {
+		scrapeConfig, err := component.ScrapeConfig()
+		if err != nil {
+			return err
+		}
+		for _, config := range scrapeConfig {
+			scrapeConfigs.WriteString(fmt.Sprintf("- %s", indent(config, 2)))
```
timebertt:
```suggestion
			scrapeConfigs.WriteString(fmt.Sprintf("- %s\n", indent(config, 2)))
```

pkg/operation/botanist/monitoring.go
timebertt: Yes, agreed. Let's do it in the prometheus component refactoring.

pkg/operation/botanist/component/interfaces.go
```diff
@@ -47,6 +47,17 @@ type MigrateWaiter interface {
 	WaitMigrate(ctx context.Context) error
 }
 
+// MonitoringComponent exposes configuration for Prometheus as well as the AlertManager.
+type MonitoringComponent interface {
+	// ScrapeConfig returns the scrape configuration for Prometheus.
+	ScrapeConfig() ([]string, error)
```
timebertt: Shouldn't this rather be named `ScrapeConfigs`?
```suggestion
	ScrapeConfigs() ([]string, error)
```

timebertt (commented):
Sorry, one more nit.
Other than that, it looks great now :)

timebertt (approved):
Thanks a lot for addressing all my feedback :) 
Looks great!
/lgtm

PR #2788, opened 2020-08-27
Fix for missing  DNSOwner CRD when shootDNS is disabled

### LINKED ISSUE #2785: Fail to create Shoot when seed.ShootDNS is disabled

/area control-plane
/kind bug
/priority normal

**What happened**:
Tried to create a Shoot cluster without setting up internal and external domain secrets and Seed's `ShootDNS` disabled. Shoot creation failed because gardener tried to operate on `DNSOwner` CRD which was not present on the seed cluster.

In detail, the DNSOwner CRD is currently only present as part of the `external-dns-management` extension. When Gardener calculates the dependencies for the seed cluster because DNS operations are disabled, it will not factor in the `external-dns` extension as necessary. 

During the shoot reconciliation loop gardener will try to delete `DNSOwner` artifacts and fail due to the CRD not installed on the cluster

Shoot creation to proceed successfully. 

Create a shoot with seed's shootDNS disabled. The seed cluster must not have the `DNSOwner` CRD already installed

**Environment**:

- Gardener version: master
- Kubernetes version (use `kubectl version`): 1.18.8
- Cloud provider or hardware configuration: Azure
- Others:

kon-angelo:

/area control-plane
/kind bug
/priority normal

With this PR the CRD `DNSOwner` will be deployed in all seed clusters regardless of the DNS settings (similarly to `DNSProvider` and `DNSEntry`). With this PR, the shoot reconciliation loop can be completed even when the `external-dns-management` extension is not deployed on the seed cluster. Previously the reconciliation loop would fail due to the missing CRD in the seed.

Fixes #2785 

The CRD is created based on the [external-dns repo](https://github.com/gardener/external-dns-management/blob/master/charts/external-dns-management/templates/crds.yaml#L223-L257) with added validation (based on [this struct](https://github.com/gardener/external-dns-management/blob/master/pkg/apis/dns/v1alpha1/dnsowner.go).

Due the added validation there is a discrepancy between the original source in the external-dns repo, which can be resolved either by removing the validation or adding validation to external-dns repo's CRD.  

**Release note**:

```improvement operator

```

### DIFF
--- a/charts/seed-bootstrap/templates/extensions/crd-dnsowner.yaml
+++ b/charts/seed-bootstrap/templates/extensions/crd-dnsowner.yaml
@@ -1,0 +1,67 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: dnsowners.dns.gardener.cloud
+  labels:
+    gardener.cloud/deletion-protected: "true"
+spec:
+  conversion:
+    strategy: None
+  group: dns.gardener.cloud
+  names:
+    kind: DNSOwner
+    listKind: DNSOwnerList
+    plural: dnsowners
+    shortNames:
+      - dnso
+    singular: dnsowner
+  scope: Cluster
+  subresources:
+    status: {}
+  version: v1alpha1
+  versions:
+    - name: v1alpha1
+      served: true
+      storage: true
+  additionalPrinterColumns:
+    - JSONPath: .spec.ownerId
+      description: Owner Id
+      name: OWNERID
+      type: string
+    - JSONPath: .metadata.creationTimestamp
+      name: AGE
+      type: date
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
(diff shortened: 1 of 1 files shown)


timuthy (changes requested):
Thanks for the PR but I don't think this is a proper fix because seed cluster now need to have this CRD even though it disables DNS. We should rather let the destroy functions of DNS tolerate the `meta.IsNoMatchError`s.

timuthy (approved):
Okay, I just saw we do the same for `DNSEntry` and `DNSProvider` which I didn't expect. Then it should be good to do it like this for `DNSOwner`, too.

kon-angelo:
> Okay, I just saw we do the same for DNSEntry and DNSProvider which I didn't expect. Then it should be good to do it like this for DNSOwner, too.

@timuthy I was going to write this. These three entities go hand-in-hand and should be handled similarly. I also discussed it with @stoyanr and we decided to add the CRD instead of handling the error.

I added the CC'ed people from the issue in case we have more opinions. 

/cc @swilen-iwanow @MartinWeindel @mandelsoft @stoyanr

rfranzke (approved):

MartinWeindel (approved):
lgtm

PR #2789, opened 2020-08-28
Add mount for KCM required for certain OS

danielfoehrKn:

/area os
/kind enhancement
/kind bug
/priority normal

Adds a host path volume to the KCM that is required for certain OS.

The KCM discovers the Root CAs available on the node via [predefined locations](https://github.com/golang/go/blob/1bb247a469e306c57a5e0eaba788efb8b3b1acef/src/crypto/x509/root_linux.go#L7-L15).

```
	"/etc/ssl/certs/ca-certificates.crt",                // Debian/Ubuntu/Gentoo etc.
	"/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora/RHEL 6
	"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
	"/etc/pki/tls/cacert.pem",                           // OpenELEC
	"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
	"/etc/ssl/cert.pem",                                 // Alpine Linux
```

This is why the kcm mounts the `etc/ssl` directory.

For certain OS (e.g coreos 2512.3.0), the `etc/ssl/certs` contains only symlinks to the actual `ca-certificates.crt` file.
```
lrwxrwxrwx. 1 root root    54 May 22 20:41  ca-certificates.crt -> ../../../usr/share/ca-certificates/ca-certificates.crt
```
 
However, the KCM cannot access `ca-certificates/ca-certificates.crt` , because it does not mount the correct directory.

Fixes #

Please also see [related issue on kubeadm for coreos](https://github.com/kubernetes/kubeadm/issues/671) and [this](https://github.com/kubernetes/kubernetes/pull/59122).

We can also think about mounting the other directories like `"/etc/pki/tls/certs/`. 
But I would like to talk about and first go ahead with this issue.
**Release note**:

```improvement operator
KCM mounts `/usr/share/ca-certificates` required for certain OS (e.g CoreOS 2512.3.0),
```

### DIFF
--- a/charts/seed-controlplane/charts/kube-controller-manager/templates/kube-controller-manager.yaml
+++ b/charts/seed-controlplane/charts/kube-controller-manager/templates/kube-controller-manager.yaml
@@ -130,6 +130,9 @@ spec:
           mountPath: /var/lib/kube-controller-manager
         - name: kube-controller-manager-server
           mountPath: /var/lib/kube-controller-manager-server
+        - name: usr-share-cacerts
+          mountPath: /usr/share/ca-certificates
+          readOnly: true
       dnsPolicy: ClusterFirst
       restartPolicy: Always
       schedulerName: default-scheduler
@@ -147,3 +150,7 @@ spec:
       - name: kube-controller-manager-server
         secret:
           secretName: kube-controller-manager-server
+      - hostPath:
+          path: /usr/share/ca-certificates
+          type: "DirectoryOrCreate"
+        name: usr-share-cacerts


rfranzke (changes requested):
This should not be mounted here but via the webhooks as it's also done for `/etc/ssl`, see https://github.com/gardener/gardener-extension-provider-aws/blob/master/pkg/webhook/controlplane/ensurer.go#L274-L298

danielfoehrKn:
Closing this in favour [aws extension pr](https://github.com/gardener/gardener-extension-provider-aws/pull/173).

PR #2791, opened 2020-08-28
Kube apiserver mount ssl directories instead of files

danielfoehrKn:

/area control-plane
/priority normal

The current apiserver tries to mount certain files such as `/etc/ssl/ca-bundle.pem`.
If these files do not exist, they will be created as a **directory**  (also see [docs](https://v1-15.docs.kubernetes.io/docs/concepts/storage/volumes/#hostpath)).

We have seen confusing error messages in the KCM such as 

```
error listing AWS instances: “RequestError: send request failed\ncaused by: Post https://ec2.us-gov-west-1.amazonaws.com/: x509: certificate signed by unknown authority: /etc/ssl/certs/ca-certificates.crt is a directory 
```
Possibly the erroneous file `/etc/ssl/ca-bundle.pem` was only accessed due to missing `/etc/ssl/certs/ca-certificates.crt` (as such it might be [related to this PR](https://github.com/gardener/gardener/pull/2789)).

Nevertheless I think these directories (with filenames) should not be created because this might lead to confusion and confusing error messages.

Therefore I see two options
 - change the current file mounts to have [hostpath type `FileOrCreate`](https://v1-15.docs.kubernetes.io/docs/concepts/storage/volumes/#hostpath)  so that no directories are created, but files
 - only mount directories with  [hostpath type `DirectoryOrCreate `](https://v1-15.docs.kubernetes.io/docs/concepts/storage/volumes/#hostpath). I prefer having an empty directory created if it does not exist, over a dummy file. 

Fixes #

**Release note**:

```improvement operator
Shoot Kube apiserver mounts ssl directories instead of files to avoid creation of empty files. 
This includes directory `/etc/ssl/`.
```
```improvement operator
Please update your provider extension to version >= 1.16.0. Previous provider extensions  remove the `/etc/ssl/` directory mount from the kube-apiserver deployment via webhook . This is a problem when the Shoot Kubernetes version > 1.17 and CSI is enabled, as the Shoot API Server could require the Root CAs in  `/etc/ssl/` to make requests to OIDC providers or webhook endpoints.
The Gardenlet now always adds the `/etc/ssl/` mount - a provider extension with a previous version would remove the `/etc/ssl/` mount from the deployment if the Shoot Kubernetes version > 1.17 and CSI is enabled.
```

### DIFF
--- a/charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
+++ b/charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
@@ -258,26 +258,17 @@ spec:
           readOnly: true
         {{- end }}
         {{- if semverCompare ">= 1.17" .Values.kubernetesVersion }}
-        # All those locations are taken from
+        # locations are taken from
         # https://github.com/golang/go/blob/1bb247a469e306c57a5e0eaba788efb8b3b1acef/src/crypto/x509/root_linux.go#L7-L15
-        # we cannot be sure on which Seed Cluster Node OS is running so, it's safer to mount them all
-        - name: debian-family-cabundle
-          mountPath: /etc/ssl/certs/ca-certificates.crt
-          readOnly: true
-        - name: fedora-rhel6-cabundle
-          mountPath: /etc/pki/tls/certs/ca-bundle.crt
-          readOnly: true
-        - name: opensuse-cabundle
-          mountPath: /etc/ssl/ca-bundle.pem
-          readOnly: true
-        - name: openelec-cabundle
-          mountPath: /etc/pki/tls/cacert.pem
+        # we cannot be sure on which Node OS the Seed Cluster is running so, it's safer to mount them all
+        - name: fedora-rhel6-openelec-cabundle
+          mountPath: /etc/pki/tls
           readOnly: true
         - name: centos-rhel7-cabundle
-          mountPath: /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+          mountPath: /etc/pki/ca-trust/extracted/pem
           readOnly: true
-        - name:  alpine-linux-cabundle
-          mountPath: /etc/ssl/cert.pem
+        - name: fedora-rhel6-alpine-openelec-cabundle
+          mountPath: /etc/ssl
           readOnly: true
         {{- end }}
       {{- if .Values.konnectivityTunnel.enabled }}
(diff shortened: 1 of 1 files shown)


rfranzke:
ping @danielfoehrKn @mvladev

danielfoehrKn:
@rfranzke as to your comment why the `/etc/ssl` directory is not explicitly mounted: this directory is added by the [provider webhooks anyways](https://github.com/gardener/gardener-extension-provider-aws/blob/ce90d1fc72e4a8278f36c14426c705cc35b0c38a/pkg/webhook/controlplane/ensurer.go#L299) in case CSI is not enabled & the upstream hyperkube image is not used. 
Thus I think adding the mount here would be a duplication right?

rfranzke:
We cannot make any assumptions in g/g wrt what extensions or third-party webhooks will do. If g/g needs these mounts then we should add them here.

danielfoehrKn:
Syncing with @mvladev 
- we always need to mount these directories because the API server can make requests to OIDC or webhooks
- I will add the /etc/ssl directory to this PR

The problem is, that the provider-extensions [remove the /etc/ssl directory](https://github.com/gardener/gardener-extension-provider-aws/blob/ce90d1fc72e4a8278f36c14426c705cc35b0c38a/pkg/webhook/controlplane/ensurer.go#L294) when CSI is enabled. Thats why adding it here currently has no effect - they ll be removed by the provider webhooks.

I propose not touching the /etc/ssl mount in the provider extensions.

rfranzke:
Makes sense. We should only touch the `/etc/ssl` mounts for KCM then. Will you open PRs in the provider extensions?

danielfoehrKn:
Updated and tested with Kubernetes 1.16, 1.17 & 1.18 together with the aws-provider extension.

rfranzke (approved):
/lgtm

mvladev (approved):
/lgtm

rfranzke:
Btw, what about `/usr/share/ca-certificates`?

danielfoehrKn:
@rfranzke you are right, I meant to add it. Thanks for noticing it.

danielfoehrKn:
Removing release notes that specify incompatible changes due to #2852 .

> Please update your provider extension to version >= 1.16.0. Previous provider extensions  remove the `/etc/ssl/` directory mount from the kube-apiserver deployment via webhook when the Shoot Kubernetes version > 1.17 and CSI is enabled. This is a problem, as the Shoot API Server could require the Root CAs in  `/etc/ssl/` to make requests to OIDC providers or webhook endpoints.
The Gardenlet now always adds the `/etc/ssl/` mount - a provider extension with a previous version would remove the `/etc/ssl/` mount from the deployment if the Shoot Kubernetes version > 1.17 and CSI is enabled.

vpnachev:
```
Please update your provider extension to version >= 1.16.0....
```

This release note is wrong now when each provider extension is following its own release lifecycle,e.g:
- gcp 1.11.0 : https://github.com/gardener/gardener-extension-provider-gcp/releases
- alicloud 1.17.0: https://github.com/gardener/gardener-extension-provider-alicloud/releases/.

There might be even other extension not hosted under https://github.com/gardener org.

rfranzke:
@danielfoehrKn Please make this API server mount configurable (via a feature gate) to not introduce incompatibility problems.

PR #2792, opened 2020-08-28
Remove kube-state-metrics-seed from control planes

wyb1:

/area monitoring
/kind cleanup
/priority normal

Each control plane has 2 kube state metrics deployments. One for the shoot and one for the components running in the seed. Since we have a kube state metrics running in the garden ns we can configure prometheus to use that ksm and remove the one running in the control plane.

Fixes #

**Release note**:

```improvement operator
NONE
```


charts/seed-monitoring/charts/core/charts/prometheus/templates/allow-prometheus-network-policy.yaml
```diff
@@ -17,6 +17,12 @@ spec:
     - podSelector:
         matchLabels:
           networking.gardener.cloud/from-prometheus: allowed
+    - podSelector:
+        matchLabels:
+          networking.gardener.cloud/from-prometheus: allowed
+      namespaceSelector:
+        matchLabels:
+          role: garden
```
timebertt: Does it make sense to make this pod selector more specific to only allow talking to the kube-state-metrics deployment (i.e. using `component: kube-state-metrics`)? We could also use a dedicated label for the direction `shoot prometheus -> seed component` to distinguish from `seed prometheus -> seed component` (e.g. `networking.gardener.cloud/from-shoot-prometheus`).
We probably don't want to allow the Shoot prometheus' (I don't remember the correct plural 😉) to talk to all seed components that want to be scraped by the seed prometheus.

charts/seed-monitoring/charts/core/charts/prometheus/templates/config.yaml
```diff
@@ -274,6 +277,30 @@ data:
       - target_label: instance
         replacement: kube-state-metrics
       metric_relabel_configs:
+      # We want to drop all metrics that
+      # do not belong to the shoot
+
+      # Add a temp label to all ksm metrics
+      - target_label: __tmp_keep
```
timebertt: Although I kind of understand what's happening here, I would like to think about an alternative to make it easier to understand.
What about adding a second scrape config for the seed ksm, which keeps only metrics for the current Shoot namespace, (and rewrites the job name to `kube-state-metrics` at the end, if needed)?

timebertt (changes requested):
Thanks for the PR! Makes sense to me :)

Though, what I'm slightly worried about, is the performance of the Shoot prometheis (?).
I would imagine, that scraping the global ksm might take a long time in large Seeds. Have you seen any increases in scrape times in your tests? Can you rule out this concern somehow?

wyb1:
Thanks for the review @timebertt!
> I would imagine, that scraping the global ksm might take a long time in large Seeds. Have you seen any increases in scrape times in your tests? Can you rule out this concern somehow?

That is a valid point. I'm not sure what the performance impact on ksm is if 100+ prometheus are scraping it every minute. If it is an issue we could horizontally scale ksm. Prometheus is already configured to scrape the service since it doesn't matter which replica is scraped so this would distribute the load across several replicas.

charts/seed-monitoring/charts/core/charts/prometheus/templates/config.yaml
wyb1: I discussed with @istvanballok and we have a few ideas on how we can simplify the config. Unfortunately it is a bit difficult since it is not possible to do negations with this implementation of the regex engine.

charts/seed-monitoring/charts/core/charts/prometheus/templates/allow-prometheus-network-policy.yaml
wyb1: Makes sense. I can make the label more specific to ksm.

Regarding the dedicated label, this would make sense if we start scraping multiple components in the garden ns, but this is not (yet) the case. If there is a use case where we need to scrape more central components, then I would suggest we add a dedicated label.

timebertt:
> I'm not sure what the performance impact on ksm is if 100+ prometheus are scraping it every minute.

Well, that's also a concern, but one we could tackle by scaling ksm, you're right.

But actually, I was rather asking about the opposite direction 😄
To phrase it differently: What's the impact on the Shoot prometheus' side, if it has to scrape the global ksm which very many metrics (for all namespaces)?

wyb1:
> What's the impact on the Shoot prometheus' side, if it has to scrape the global ksm which very many metrics (for all namespaces)?

Filtering the metrics from the global ksm has close to zero performance impact on the shoot prometheus. In the end it should be the same amount of metrics anyway so there should also be no impact on the required disk space. My only concern is the ksm being scraped by 100+ shoot prometheus', but this can be solved through horizontal scaling

timebertt:
> so there should also be no impact on the required disk space

Yep, sure. What about the performance implications while scraping the global ksm? Have you seen any increases in scrape times in your tests?

I don't try to make a point here. Just want to understand the implications beforehand :)

wyb1:
> Yep, sure. What about the performance implications while scraping the global ksm? Have you seen any increases in scrape times in your tests?

Sorry I wasn't clear 😄 there should be no implications regarding scraping the global ksm or on disk space. I will double check later and provide some data

timebertt:
> there should be no implications regarding scraping the global ksm or on disk space. I will double check later and provide some data

Ok, great. Thanks!

wyb1:
Regarding the performance, I tested the scrape duration. Once with the old ksm setup and then with the ksm in the garden ns. Here are the results.
ksm in shoot ns:
Screen Shot 2020-09-01 at 09 30 20

ksm in garden ns:
Screen Shot 2020-09-01 at 09 46 01
(Values range from 0.01s to 0.04s)

Scraping the ksm in the garden ns seems to take a bit longer but this is still insignificant IMO.

We should observe how this behaves in larger seeds.

timebertt:
> (Values range from 0.01s to 0.04s)
> 
> Scraping the ksm in the garden ns seems to take a bit longer but this is still insignificant IMO.

This sounds great. I didn't expect the implications to be that little :) 


> We should observe how this behaves in larger seeds.

Sure. But I see, we already have a dashboard panel for the scrape duration, so we can easily observe this. Great :)

wyb1:
I changed the config to use a whitelist approach instead of blacklisting.

I've applied the changes PTAL @timebertt

istvanballok (dismissed):
/lgtm

charts/seed-monitoring/charts/core/charts/prometheus/templates/config.yaml
```diff
@@ -274,6 +277,29 @@ data:
...
+        replacement: false
+      # Keep all shoot metrics
+      - source_labels: [type]
+        regex: shoot
+        target_label: __tmp_keep
+        replacement: true
+      # we want to keep all metrics belonging to the shoot
+      - source_labels: [namespace]
```
timebertt, line 306: Hmm, the Shoot ksm also collects metrics for `nodes`:
https://github.com/gardener/gardener/blob/3b4ae6caf2fb60731d5481e63b2d8fe577760b90/charts/seed-bootstrap/charts/kube-state-metrics/templates/deployment.yaml#L40
This will be filtered out now because they probably don't have a namespace, right?
If we need those metrics, would it make sense to only collect them once in the seed ksm?

charts/seed-monitoring/charts/core/charts/prometheus/templates/config.yaml
timebertt: I assume, having a separate scrape config for the Seed ksm is not possible?

charts/seed-monitoring/charts/core/charts/prometheus/templates/config.yaml
```diff
@@ -274,6 +277,29 @@ data:
       - target_label: instance
         replacement: kube-state-metrics
       metric_relabel_configs:
+      # We want to drop all metrics that
+      # do not belong to the shoot
+
+      # Add a temp label to all ksm metrics
+      - target_label: __tmp_keep
+        replacement: false
```
timebertt: Can you quote the temp values to avoid confusion with bools?
```suggestion
        replacement: "false"
```

timebertt (commented):
Looked through it all, now.
Looks good, just a few comments/questions.

charts/seed-monitoring/charts/core/charts/prometheus/templates/config.yaml
wyb1: it is possible but there are alerts and dashboards that reference the the job `kube-state-metrics`. This is why I wanted to keep it as one job. If you prefer I can change it to 2 jobs

charts/seed-monitoring/charts/core/charts/prometheus/templates/config.yaml
wyb1: That is correct but we still need those because the aggregate prometheus collects these for seed monitoring (indirectly via the `prometheus-0` in the garden ns).

charts/seed-monitoring/charts/core/charts/prometheus/templates/config.yaml
timebertt: Hmm, ok. I would prefer to split it up into 2 jobs. But if that's too much effort, we can leave it like this (although, I would say that 2 jobs are the cleaner way).
You can choose :)

charts/seed-monitoring/charts/core/charts/prometheus/templates/config.yaml
timebertt: Sorry, I got this mixed up. I looked at the seed ksm, not the shoot ksm 🙈
So this is all fine... :)

timebertt (dismissed):
Thanks! Looks great now!
/lgtm

wyb1:
@timebertt Thanks for the review! Rebased changes

timebertt (dismissed):
/remove needs/rebase
/lgtm

timebertt:
@wyb1 Can you check the failing tests?

wyb1:
I fixed the tests relevant to this PR but there are still some unit tests failing that also failed on master

timebertt:
Thanks, the failing tests on master should be fixed with https://github.com/gardener/gardener/pull/2808

timebertt (dismissed):

timebertt:
/needs rebase
to get in the fix from #2808

wyb1:
rebased

timebertt (dismissed):
Thanks!
/lgtm

pkg/operation/botanist/monitoring.go
```diff
@@ -448,6 +434,38 @@ func (b *Botanist) DeleteSeedMonitoring(ctx context.Context) error {
...
+		}
+	}
+
+	return nil
+}
+
+// DeleteSeedMonitoring will delete the monitoring stack from the Seed cluster to avoid phantom alerts
+// during the deletion process. More precisely, the Alertmanager and Prometheus StatefulSets will be
+// deleted.
+func (b *Botanist) DeleteSeedMonitoring(ctx context.Context) error {
```
rfranzke, line 449: Shouldn't this function call `DeleteKubeStateMetricsSeed`?

wyb1: `DeleteKubeStateMetricsSeed` is called in the `DeploySeedMonitoring` function to remove any existing `kube-state-metrics-seed` deployments running in the control plane. Why should `DeleteSeedMonitoring` call this function?

rfranzke: If an existing shoot already has a deletion timestamp and the new Gardener version with this change is deployed then `DeploySeedMonitoring` won't be called, but only `DeleteSeedMonitoring`. In this case the `kube-state-metrics-seed` won't be deleted, or?

wyb1: Good catch, I will call the function in `DeleteSeedMonitoring`

pkg/operation/botanist/monitoring.go
```diff
@@ -448,6 +434,43 @@ func (b *Botanist) DeleteSeedMonitoring(ctx context.Context) error {
...
+		return err
+	}
+
+	for _, obj := range []runtime.Object{
+		&autoscalingv1beta2.VerticalPodAutoscaler{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: b.Shoot.SeedNamespace,
+				Name:      "kube-state-metrics-seed-vpa",
+			},
+		},
```
timebertt: I guess, this one is not needed then, if you call `DeleteKubeStateMetricsSeed` beforehand.

timebertt (approved):
Thanks! Let's get it in this time 😉 
/lgtm

rfranzke (approved):

PR #2798, opened 2020-08-31
[GEP-7] Use patch to update the ShootState and remove LoadSecrets step

plkokanov:

/area control-plane-migration
/kind cleanup
/priority normal

This PR makes it so that the resources in the `ShootState.Spec.Gardener` list are updated with a PATCH request and if there were no changes to the resources the PATCH request is skipped.
Previously `CreateOrUpdate` was used, however there is no need to check if the ShootState is created (this happens in a previous step). Additionally there was a caching issue because `CreateOrUpdate` was called with a reference to the cached `Client()` instead of using `DirectClient()`

Part of #1631 

Note that I do not use `Get()` to retrieve the ShootState resource before using `Patch()`. This basically skips a call to the server but relies on the fact that no-one will change the `ShootState.Spec.Gardener` list on the server during reconciliation without also updating the `ShootState` field in the `Operation` struct. 
I am open to implementing it with Get() -> Patch() or Get() -> Update() in a `retry.RetryOnConflict()` call if it seems better.

**Release note**:

```improvement operator
Fixed a caching issue when updating the ShootState resource. Additionally PATCH is now used to update the `ShootState.Spec.Gardener` field.
```
```improvement operator
The step which loads existing secrets into the ShootState has been removed as it is not needed anymore.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -134,16 +134,11 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 			Fn:           flow.TaskFn(botanist.Shoot.Components.ControlPlane.KubeAPIServerService.Wait).SkipIf(o.Shoot.HibernationEnabled && !useSNI),
 			Dependencies: flow.NewTaskIDs(deployKubeAPIServerService),
 		})
-		loadSecrets = g.Add(flow.Task{
-			Name:         "Loading existing secrets into ShootState",
-			Fn:           flow.TaskFn(botanist.LoadExistingSecretsIntoShootState),
-			Dependencies: flow.NewTaskIDs(deployNamespace, ensureShootStateExists),
-		})
 		generateSecrets = g.Add(flow.Task{
 			Name: "Generating secrets and saving them into ShootState",
 			Fn:   flow.TaskFn(botanist.GenerateAndSaveSecrets),
 			Dependencies: func() flow.TaskIDs {
-				taskIDs := flow.NewTaskIDs(deployNamespace, loadSecrets, ensureShootStateExists)
+				taskIDs := flow.NewTaskIDs(deployNamespace, ensureShootStateExists)
 				if !dnsEnabled && !o.Shoot.HibernationEnabled {
 					taskIDs.Insert(waitUntilKubeAPIServerServiceIsReady)
 				}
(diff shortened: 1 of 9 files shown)


timebertt (commented):
Thanks for the PR. Switching to patches seems to make a lot of sense.

> I am open to implementing it with Get() -> Patch() or Get() -> Update() in a `retry.RetryOnConflict()` call if it seems better.

I don't think, getting the current object with `Get` before `Patch`ing should be necessary, as you just want to ensure, that the current state is properly reflected in the ShootState.

timuthy (dismissed):
/lgtm

pkg/operation/operation.go
```diff
@@ -436,7 +436,7 @@ func (o *Operation) EnsureShootStateExists(ctx context.Context) error {
 	blockOwnerDeletion := false
 	ownerReference.BlockOwnerDeletion = &blockOwnerDeletion
 
-	_, err := controllerutil.CreateOrUpdate(ctx, o.K8sGardenClient.Client(), shootState, func() error {
+	_, err := controllerutil.CreateOrUpdate(ctx, o.K8sGardenClient.DirectClient(), shootState, func() error {
```
timebertt, line 439: Have you seen any issues here with the cached client?
If yes, I would probably also go for a patch here instead of using the `DirectClient`.

pkg/operation/operation_test.go
```diff
@@ -140,4 +142,43 @@ var _ = Describe("operation", func() {
...
+			o                      *Operation
+			ctrl                   *gomock.Controller
+			gardenClient           *mock.MockInterface
+			k8sGardenRuntimeClient *mockclient.MockClient
+		)
+
+		BeforeEach(func() {
+			ctrl = gomock.NewController(GinkgoT())
+			gardenClient = mock.NewMockInterface(ctrl)
+			k8sGardenRuntimeClient = mockclient.NewMockClient(ctrl)
```
timebertt: The clientset (`kubernetes.Interface`) is just a container for all the different clients we use. If you just need it to pass down a runtime client in your test, you can use the fake clientset, which makes it easier to read/understand, e.g.:
https://github.com/gardener/gardener/blob/a39cd072a8c6eabab4c408a3e18c0b90b920e851/pkg/controllermanager/controller/controllerregistration/controllerregistration_control_test.go#L288-L289

timebertt (commented):
Only some minor comments about the client usage from my side.
I haven't looked into the whole stuff with the secrets in detail, though...

pkg/operation/operation.go
plkokanov: Not in this function directly, however we ran into a case where the `Get()` inside `CreateOrUpdate()` would return `NotFound` error when the object actually exists in the cluster.
The first time EnsureShootStateExists is executed it should create the ShootState object. Generally, this means that it should be ok to use the cached client, but after gardenlet restarts we might encounter the error I mentioned above.

pkg/operation/operation_test.go
plkokanov: Yes was wondering if I should directly use the fake client, but copy pasted from the tests above :/, I'll see if I can improve the test a bit later today.

pkg/operation/operation.go
timebertt: Hmm, ok. Thanks for explaining.

pkg/operation/operation_test.go
timebertt: It was just suggestion. But if you want to adapt, perfect :)

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -222,12 +217,12 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		generateEncryptionConfigurationMetaData = g.Add(flow.Task{
 			Name:         "Generating etcd encryption configuration",
 			Fn:           flow.TaskFn(botanist.GenerateEncryptionConfiguration).DoIf(enableEtcdEncryption),
-			Dependencies: flow.NewTaskIDs(deployNamespace, ensureShootStateExists, loadSecrets, generateSecrets),
+			Dependencies: flow.NewTaskIDs(deployNamespace, ensureShootStateExists),
```
rfranzke, line 220: Why was `generateSecrets` removed?

plkokanov: Don't need it for this step. `generateSecrets` is needed for the step which persists the ETCDEncrytionConfiguration as that step is the one reading/writing from/to the ShootState.
GenerateEncryptionConfiguration uses a variable to store only the generated configuration.

rfranzke: OK, thanks!

rfranzke (dismissed):

plkokanov:
Just a note, rebased and updated the test case with @timebertt's suggestion.

timebertt (commented):
Thanks!
/lgtm

rfranzke (approved):

PR #2802, opened 2020-09-01
VPA admission controller shouldn't self register mutatingWebhook

### LINKED ISSUE #2779: Mutating webhook doesn't abide by the failurePolicy of Ignore and blocks further pod creation

/area control-plane
/kind bug
/priority normal

**What happened**:
The issue happened during upgrade of Gardener, which had enhancement to move VPA to shoot control plane in the shoot namespace. During the upgrade there is a short period of time when there is no VPA pod backing the mutating webhook which results in error. Although the failurePolicy for the mutating webhook vpa-webhook-config  is set to “Ignore” but during the upgrade each restart of the vpa-admission pod resets this timeout value for mutating webhook back to 30 seconds which doesn’t honor the failurePolicy “Ignore” (a limitation from upstream Kubernetes
) and hence blocks any further pod creation. 

The mutating webhookconfiguration of failurePolicy: Ignore should be respected and a failure should not block further pod creation. 

Sequence of events  per my understanding, but the actual flow during upgrade might be different so please correct. 
Edit the mutating webhookconfiguration vpa-webhook-config  in the seed  to 10s. 
Restart the vpa-admission pod 
Check the timeout it should be set back to 30s which doesn't follow the failurePolicy of Ignore. 

A possible  mitigation discussed could be to disable the self registration of the mutating webhook and let Gardener deploy it.. In this way we will have full control of the timeout and it can be properly set to something different to 30 seconds.

**Environment**:

- Gardener version: 0.85.0
- Kubernetes version (use `kubectl version`): Kubernetes 1.16 
- Others:

ggaurav10:

/area control-plane
/area auto-scaling
/kind task
/priority normal

If VPA admission controller self registers mutatingWebhook, default timeout of 30s is used which can result in API call failures. This PR disables self registration of vpa webhook, and enables Gardener to manage it instead.

Fixes #2779

**Release note**:

```improvement operator
Disable self-registration of vpa-webhook by vpa-admission-controller. Now Gardener manages the vpa-webhook
```

### DIFF
--- a/charts/seed-bootstrap/charts/vpa/charts/application/templates/vpa-webhook-config.yaml
+++ b/charts/seed-bootstrap/charts/vpa/charts/application/templates/vpa-webhook-config.yaml
@@ -1,0 +1,51 @@
+{{- if .Values.admissionController.enabled }}
+{{- if not .Values.admissionController.registerWebhook }}
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: vpa-webhook-config
+webhooks:
+- admissionReviewVersions:
+  - v1beta1
+  clientConfig:
+    caBundle: {{ .Values.admissionController.caCert }}
+{{- if .Values.admissionController.registerByURL }}
+    url: https://vpa-webhook.{{.Values.admissionController.controlNamespace}}:{{.Values.admissionController.servicePort}} # the port is only respected if register-by-url is true, that's why it's in this if-block
+                                                                                                                          # if it's false it will not set the port during registration, i.e., it will be defaulted to 443,
+                                                                                                                          # so the servicePort has to be 443 in this case
+                                                                                                                          # see https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/admission-controller/config.go#L70-L74
+{{- else }}
+    service:
+      name: vpa-webhook
+      namespace: {{ .Values.admissionController.controlNamespace }}
+      port: 443
+{{- end }}
+  failurePolicy: Ignore
+  matchPolicy: Exact
+  name: vpa.k8s.io
+  reinvocationPolicy: Never
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    resources:
+    - pods
+    scope: '*'
+  - apiGroups:
+    - autoscaling.k8s.io
+    apiVersions:
+    - '*'
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - verticalpodautoscalers
+    scope: '*'
+  sideEffects: None
+  timeoutSeconds: 10
+  {{- end }}
+  {{- end }}
--- a/charts/seed-bootstrap/charts/vpa/charts/application/templates/vpa-webhook-config.yaml
+++ b/charts/seed-bootstrap/charts/vpa/charts/application/templates/vpa-webhook-config.yaml
@@ -1,7 +1,6 @@
 {{- if .Values.admissionController.enabled }}
-{{- if not .Values.admissionController.registerWebhook }}
 ---
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: {{ include "webhookadmissionregistration" . }}
 kind: MutatingWebhookConfiguration
 metadata:
   name: vpa-webhook-config
@@ -10,7 +9,7 @@ webhooks:
   - v1beta1
   clientConfig:
     caBundle: {{ .Values.admissionController.caCert }}
-{{- if .Values.admissionController.registerByURL }}
+{{- if eq .Values.clusterType "shoot" }}
     url: https://vpa-webhook.{{.Values.admissionController.controlNamespace}}:{{.Values.admissionController.servicePort}} # the port is only respected if register-by-url is true, that's why it's in this if-block
                                                                                                                           # if it's false it will not set the port during registration, i.e., it will be defaulted to 443,
                                                                                                                           # so the servicePort has to be 443 in this case
@@ -47,5 +46,4 @@ webhooks:
     scope: '*'
   sideEffects: None
   timeoutSeconds: 10
-  {{- end }}
   {{- end }}
--- a/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
+++ b/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
@@ -49,15 +49,7 @@ spec:
         - --tls-private-key=/etc/tls-certs/tls.key
         - --address=:8944
         - --port={{ .Values.admissionController.port }}
-        - --register-webhook={{.Values.admissionController.registerWebhook}}
-        {{- if .Values.admissionController.registerByURL }}
-        - --register-by-url=true
-        - --webhook-address=https://vpa-webhook.{{.Release.Namespace}}
-        - --webhook-port={{ .Values.admissionController.servicePort }} # the port is only respected if register-by-url is true, that's why it's in this if-block
-                                                                       # if it's false it will not set the port during registration, i.e., it will be defaulted to 443,
-                                                                       # so the servicePort has to be 443 in this case
-                                                                       # see https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/admission-controller/config.go#L70-L74
-        {{- end }}
+        - --register-webhook=false
         image: {{ index .Values.global.images "vpa-admission-controller" }}
         imagePullPolicy: IfNotPresent
         env:
--- a/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
+++ b/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
@@ -49,6 +49,7 @@ spec:
         - --tls-private-key=/etc/tls-certs/tls.key
         - --address=:8944
         - --port={{ .Values.admissionController.port }}
+        - --register-webhook={{.Values.admissionController.registerWebhook}}
         {{- if .Values.admissionController.registerByURL }}
         - --register-by-url=true
         - --webhook-address=https://vpa-webhook.{{.Release.Namespace}}
--- a/charts/seed-bootstrap/charts/vpa/values.yaml
+++ b/charts/seed-bootstrap/charts/vpa/values.yaml
@@ -16,6 +16,9 @@ admissionController:
   podAnnotations: {}
   podLabels: {}
   enableServiceAccount: true
+  registerWebhook: false
+  caCert: abcd
+  controlNamespace: abcd
   registerByURL: false
   port: 10250
   servicePort: 443
--- a/charts/seed-bootstrap/charts/vpa/values.yaml
+++ b/charts/seed-bootstrap/charts/vpa/values.yaml
@@ -16,10 +16,8 @@ admissionController:
   podAnnotations: {}
   podLabels: {}
   enableServiceAccount: true
-  registerWebhook: false
   caCert: abcd
   controlNamespace: abcd
-  registerByURL: false
   port: 10250
   servicePort: 443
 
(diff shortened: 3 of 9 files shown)


ggaurav10:
/hold WIP
shoot's VPA needs to be deployed in shoot

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/vpa-webhook-config.yaml
```diff
@@ -0,0 +1,51 @@
+{{- if .Values.admissionController.enabled }}
+{{- if not .Values.admissionController.registerWebhook }}
+---
+apiVersion: admissionregistration.k8s.io/v1
```
vpnachev: This API is available only in k8s >=1.16?
Please use https://github.com/gardener/gardener/blob/master/charts/utils-templates/templates/_versions.tpl

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
```diff
@@ -49,6 +49,9 @@ spec:
         - --tls-private-key=/etc/tls-certs/tls.key
         - --address=:8944
         - --port={{ .Values.admissionController.port }}
+        {{- if not .Values.admissionController.registerWebhook }}
+        - --register-webhook=false
+        {{- end }}
```
vpnachev:
```suggestion
        - --register-webhook= {{ .Values.admissionController.registerWebhook }}
```

vpnachev (dismissed):
I have just quickly looked through the change and have some proposals.

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/vpa-webhook-config.yaml
```diff
@@ -0,0 +1,51 @@
+{{- if .Values.admissionController.enabled }}
```
rfranzke: This file should go to the `application` directory

pkg/operation/botanist/controlplane.go
```diff
@@ -234,6 +235,7 @@ func (b *Botanist) DeployVerticalPodAutoscaler(ctx context.Context) error {
 			"podLabels": utils.MergeMaps(podLabels, map[string]interface{}{
 				v1beta1constants.LabelNetworkPolicyFromShootAPIServer: "allowed",
 			}),
+			"caCert":               b.Secrets["vpa-tls-certs"].Data[secretsutils.DataKeyCertificateCA],
 			"enableServiceAccount": false,
 			"registerByURL":        true,
```
rfranzke: Why is this kept `true`?

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/vpa-webhook-config.yaml
rfranzke: +1, use `v1beta1` please

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/vpa-webhook-config.yaml
```diff
@@ -0,0 +1,51 @@
...
+kind: MutatingWebhookConfiguration
+metadata:
+  name: vpa-webhook-config
+webhooks:
+- admissionReviewVersions:
+  - v1beta1
+  clientConfig:
+    caBundle: {{ .Values.admissionController.caCert }}
+{{- if .Values.admissionController.registerByURL }}
+    url: https://vpa-webhook.{{.Release.Namespace}}:{{.Values.admissionController.servicePort}} # the port is only respected if register-by-url is true, that's why it's in this if-block
```
rfranzke: I guess `{{ .Release.Namespace }}` is wrong here if this webhook config shall be deployed into the shoot

charts/seed-bootstrap/charts/vpa/charts/application/templates/vpa-webhook-config.yaml
```diff
@@ -0,0 +1,51 @@
+{{- if .Values.admissionController.enabled }}
+{{- if not .Values.admissionController.registerWebhook }}
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
```
rfranzke:
```suggestion
apiVersion: {{ include "webhookadmissionregistration" . }}
```

charts/seed-bootstrap/charts/vpa/values.yaml
```diff
@@ -16,6 +16,9 @@ admissionController:
   podAnnotations: {}
   podLabels: {}
   enableServiceAccount: true
+  registerWebhook: false
+  caCert: abcd
```
rfranzke, line 19: To make clear it's an optional value:

```suggestion
# caCert: abcd
```

charts/seed-bootstrap/charts/vpa/values.yaml
```diff
@@ -16,6 +16,9 @@ admissionController:
...
+  controlNamespace: abcd
   registerByURL: false
```
rfranzke:
```suggestion
```

Can this one be removed now? We always want to create the webhook config ourselves, right?

charts/seed-bootstrap/charts/vpa/charts/application/templates/vpa-webhook-config.yaml
```diff
@@ -0,0 +1,51 @@
+{{- if .Values.admissionController.enabled }}
+{{- if not .Values.admissionController.registerWebhook }}
```
rfranzke:
```suggestion
```

charts/seed-bootstrap/charts/vpa/charts/application/templates/vpa-webhook-config.yaml
```diff
@@ -0,0 +1,51 @@
...
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: vpa-webhook-config
+webhooks:
+- admissionReviewVersions:
+  - v1beta1
+  clientConfig:
+    caBundle: {{ .Values.admissionController.caCert }}
+{{- if .Values.admissionController.registerByURL }}
```
rfranzke:
```suggestion
{{- if eq .Values.clusterType "shoot" }}
```

charts/seed-bootstrap/charts/vpa/charts/application/templates/vpa-webhook-config.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - '*'
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - verticalpodautoscalers
+    scope: '*'
+  sideEffects: None
+  timeoutSeconds: 10
+  {{- end }}
```
rfranzke:
```suggestion
```

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
```diff
@@ -49,6 +49,7 @@ spec:
         - --tls-private-key=/etc/tls-certs/tls.key
         - --address=:8944
         - --port={{ .Values.admissionController.port }}
+        - --register-webhook={{.Values.admissionController.registerWebhook}}
```
rfranzke:
```suggestion
        - --register-webhook=false
```

charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
```diff
@@ -49,6 +49,7 @@ spec:
...
         {{- if .Values.admissionController.registerByURL }}
         - --register-by-url=true
         - --webhook-address=https://vpa-webhook.{{.Release.Namespace}}
```
rfranzke: This whole section can now be removed, right?

pkg/operation/botanist/addons.go
```diff
@@ -404,11 +404,15 @@ func (b *Botanist) generateCoreAddonsChart() (*chartrenderer.RenderedChart, erro
...
-			"clusterType":         "shoot",
-			"admissionController": map[string]interface{}{"enableServiceAccount": false},
-			"exporter":            map[string]interface{}{"enableServiceAccount": false},
-			"recommender":         map[string]interface{}{"enableServiceAccount": false},
-			"updater":             map[string]interface{}{"enableServiceAccount": false},
+			"clusterType": "shoot",
+			"admissionController": map[string]interface{}{
+				"enableServiceAccount": false,
+				"controlNamespace":     b.Shoot.SeedNamespace,
+				"registerByURL":        true,
```
```suggestion
```

pkg/operation/botanist/addons.go
```diff
@@ -430,6 +434,10 @@ func (b *Botanist) generateCoreAddonsChart() (*chartrenderer.RenderedChart, erro
 		nodeNetwork = b.Shoot.GetNodeNetwork()
 	)
 
+	if _, ok := b.Secrets["vpa-tls-certs"]; ok {
+		verticalPodAutoscaler["admissionController"].(map[string]interface{})["caCert"] = b.Secrets["vpa-tls-certs"].Data[secrets.DataKeyCertificateCA]
+	}
```
rfranzke: Why don't you simply move that up to the `admissionController` section?

pkg/operation/common/utils.go
```diff
@@ -390,6 +391,10 @@ func DeleteVpa(ctx context.Context, c client.Client, namespace string, isShoot b
 			&corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: "vpa-recommender", Namespace: namespace}},
 			&corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: "vpa-updater", Namespace: namespace}},
 		)
+
+		resources = append(resources,
+			&admissionregistrationv1beta1.MutatingWebhookConfiguration{ObjectMeta: metav1.ObjectMeta{Name: "vpa-webhook-config"}},
```
rfranzke, line 393: Move it up to line 342/343.

pkg/operation/seed/seed.go
```diff
@@ -640,6 +640,12 @@ func BootstrapCluster(k8sGardenClient, k8sSeedClient kubernetes.Interface, seed
 					},
 				},
 			},
+			"application": map[string]interface{}{
+				"admissionController": map[string]interface{}{
+					"controlNamespace": v1beta1constants.GardenNamespace,
+					"caCert":           deployedSecretsMap["vpa-tls-certs"].Data[secretsutils.DataKeyCertificateCA],
```
rfranzke: Can you make the `vpa-tls-certs` string a constant somewhere and use it consistently at the various places?

rfranzke (changes requested):

pkg/operation/botanist/addons.go
ggaurav10: if the vpa is not enabled for the shoot, the secret `vpa-tls-certs` is not deployed. So adding caCert will require this check to prevent null dereference.

pkg/operation/common/utils.go
ggaurav10: the resources here are deleted using seed client. and if we are deleting VPA for a shoot, then we wouldn't want to delete this resource in the seed.

pkg/operation/botanist/addons.go
rfranzke: I see, makes sense.. let's keep it like this then. Something we should get rid when we refactor this deployment as part of #2754

pkg/operation/common/utils.go
rfranzke: I see, then move it 392/393 instead of having another `append` call :)

charts/seed-bootstrap/charts/vpa/values.yaml
```diff
@@ -16,7 +16,8 @@ admissionController:
   podAnnotations: {}
   podLabels: {}
   enableServiceAccount: true
-  registerByURL: false
+  # caCert: abcd
```
rfranzke: Sorry, I just saw that it's no optional value. Let's revert it to:

```suggestion
  caCert: abcd
```

rfranzke (commented):
Changes look good now, I need to play with it locally. Will come back. :)

rfranzke (approved):
/lgtm

PR #2806, opened 2020-09-02
Support management of external machine dependencies

dkistner:
/kind enhancement
/priority normal

Extend the worker controller in the Gardener extension library to allow provider extensions to manage external machine dependencies within the worker control loops.

Examples for external machine dependencies could be Azure AvailabilitySet or OpenStack ServerGroups which are created for the machines of a workerpool. Those external dependencies are bound to the lifecycle of the workerpool e.g. they need to be created when a new workerpool is configured, the need to be deleted when the workerpool is gone or they need to be replaced when workerpool configuration has changed.

This PR extend the `WorkerDelegate` interface with hooks to deploy and cleanup those external dependencies in the worker controller. The hooks need to be implemented by the provider extensions. The hooks are called within the `Worker` reconciliation and deletion flow.

This is still work in progress. Opened to gather feedback.

**Release note**:

```action developer
Machine dependency hook methods `DeployMachineDependencies` and `DeployMachineDependencies` need to be implemented in the worker controllers of the provider extensions.
```

```action developer
The provider extension need to transform the existing `GetMachineImages()` in the implementation of the `WorkerDelegate` interface to the new `UpdateMachineImageStatus()` method. The provider extensions need now to update the provider status on their own in the new `UpdateMachineImageStatus()` method instead of returning it.
```

### DIFF
--- a/extensions/pkg/controller/worker/genericactuator/actuator_delete.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator_delete.go
@@ -99,6 +99,11 @@ func (a *genericActuator) Delete(ctx context.Context, worker *extensionsv1alpha1
 		return errors.Wrapf(err, "failed deleting machine-controller-manager")
 	}
 
+	// Cleanup machine dependencies.
+	if err := workerDelegate.CleanupMachineDependencies(ctx); err != nil {
+		return errors.Wrap(err, "failed to cleanup machine dependencies")
+	}
+
 	return nil
 }
 
--- a/extensions/pkg/controller/worker/genericactuator/interface.go
+++ b/extensions/pkg/controller/worker/genericactuator/interface.go
@@ -43,10 +43,17 @@ type WorkerDelegate interface {
 	// GenerateMachineDeployments generates the configuration for the desired machine deployments.
 	GenerateMachineDeployments(context.Context) (worker.MachineDeployments, error)
 
-	// GetMachineImages returns the list of used machine images for this `Worker` resource. It will be stored in the
-	// `.status.providerStatus` field of the `Worker` resource such that the controller can look up its provider-specific
-	// machine image information in case the required version has been removed from its componentconfig.
-	GetMachineImages(context.Context) (runtime.Object, error)
+	// UpdateMachineImagesStatus will store a list of machine images used by the
+	// machines associated with this Worker resource in its provider status.
+	// The controller can look up its provider-specific machine image information
+	// in case the required version has been removed from the `CloudProfile`.
+	UpdateMachineImagesStatus(context.Context) error
+
+	// DeployMachineDependencies is a hook to create external machine dependencies.
+	DeployMachineDependencies(context.Context) error
+
+	// CleanupMachineDependencies is a hook to cleanup external machine dependencies.
+	CleanupMachineDependencies(context.Context) error
 }
 
 // DelegateFactory acts upon Worker resources.
--- a/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
@@ -27,7 +27,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -71,6 +70,11 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 		return 1, nil
 	}
 
+	// Deploy machine dependencies.
+	if err := workerDelegate.DeployMachineDependencies(ctx); err != nil {
+		return errors.Wrap(err, "failed to deploy machine dependencies")
+	}
+
 	// Deploy the machine-controller-manager into the cluster.
 	if err := a.deployMachineControllerManager(ctx, logger, worker, cluster, workerDelegate, replicaFunc); err != nil {
 		return err
@@ -106,14 +110,9 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 		return errors.Wrapf(err, "failed to deploy the machine classes")
 	}
 
-	// Store machine image information in worker provider status.
-	logger.Info("Storing machine images in worker status")
-	machineImages, err := workerDelegate.GetMachineImages(ctx)
-	if err != nil {
-		return errors.Wrapf(err, "failed to get the machine images")
-	}
-	if err := a.updateWorkerStatusMachineImages(ctx, worker, machineImages); err != nil {
-		return errors.Wrapf(err, "failed to update the machine images in worker status")
+	// Update the machine images in the worker provider status.
+	if err := workerDelegate.UpdateMachineImagesStatus(ctx); err != nil {
+		return errors.Wrapf(err, "failed to update the machine image status")
 	}
 
 	// Get the list of all existing machine deployments.
@@ -205,6 +204,11 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 
 	if err := a.updateWorkerStatusMachineDeployments(ctx, worker, wantedMachineDeployments, false); err != nil {
 		return errors.Wrapf(err, "failed to update the machine deployments in worker status")
+	}
+
+	// Cleanup machine dependencies.
+	if err := workerDelegate.CleanupMachineDependencies(ctx); err != nil {
+		return errors.Wrap(err, "failed to cleanup machine dependencies")
 	}
 
 	return nil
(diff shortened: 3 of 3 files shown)


extensions/pkg/controller/worker/genericactuator/actuator.go
```diff
@@ -383,3 +384,38 @@ func CleanupLeakedClusterRoles(ctx context.Context, c client.Client, providerNam
 
 	return flow.Parallel(fns...)(ctx)
 }
+
+// runMachineDependencyOperation is a wrapper to run a machine dependency operation
+// and to persist the returned status object in .status.providerStatus of the Worker resource.
+func (a *genericActuator) runMachineDependencyOperation(ctx context.Context, worker *extensionsv1alpha1.Worker, operationFn func(context.Context) (runtime.Object, error)) error {
+	// Run the machine dependency operation.
+	workerProviderStatus, err := operationFn(ctx)
+
```
rfranzke:
```suggestion
```

rfranzke (commented):
Generally I'm fine with this approach, but I have doubts about the provider status updates. Without this PR, the provider status is updated with the machine images after the machine classes are deployed. Now, with this PR, there are two more places where the provider status is overwritten with other stuff. How is this supposed to work in the end (to not lose information)? I think this part should be improved.

extensions/pkg/controller/worker/genericactuator/actuator.go
dkistner: Hmm I guess the issue is that `machine images` and now also `machine dependencies` are stored in the provider status together. The provider status in the worker status is of type `*runtime.RawExtension` and I guess therefore the contents of the provider status are a blackbox for the `genericactuator`, right?
If yes, we can't really distinguish the contents in the provider status and it is in the responsibility of the provider extensions.

So if we want to separate the things, to make updates for `machine images` and `machine dependencies` independently, then we could consider to have dedicated fields of type `*runtime.RawExtension` in the worker status for the `machine images` and `machine dependencies`. The results returned by `GetMachineImages` would be stored in `.status.machineImages` and the results returned by `DeployMachineDependencies` and `DeployMachineDependencies` would be stored in `.status.machineDependencies`.

So not really have another idea to prevent information lost if an extension does not behave correctly and return an invalid provider status.

rfranzke: No, I’m against multiple fields. We should not let implementation details drive the API design.
What about introducing a `UpdateProviderStatus` method in the `WorkerDelegate` interface which is called once after the machine classes/machine deployments/machine dependencies were deployed (but before we wait for the machine deployments to be available)? The concrete implementations in the extensions can then construct this status as necessary.

dkistner: Hmm sounds promising. Let me check. 👍

dkistner: I played a little bit around with an `UpdateProviderStatus` of the `WorkerDelegate` interface.
I guess on issue is that we would need to call the `UpdateProviderStatus` method multiple times.

In the reconciliation flow we would need to call it after `DeployMachineDependencies` + `GetMachineImages` and once more after `CleanupMachineDependencies`. In the deletion flow just once after `CleanupMachineDependencies`.


The methods `DeployMachineDependencies/CleanupMachineDependencies` and `GetMachineImages` would return different substatus of the worker provider status. Then we could extend the `WorkerDelegate` interface with method like this:
```
UpdateProviderStatus(runtime.Object, runtime.Object) error
```
So the first `runtime.Object` parameter is for the machine image substatus and the second one for the machine dependencies substatus.

I guess this would work, but it would bring some assumption into the `genericactuator` how the worker provider status should be structured. I'm not sure if we want that. IMHO the worker provider status should be a blackbox for the `genericactuator` and only the respective provider-extension should decide how the status is structured.
But in this case the provider-extension need to make sure that the status is generated correctly and that no existing fields will be overridden.

So what is your option on that?

rfranzke: I agree to your statement that we don't want to bring these assumptions about the provider status structure into the generic actuator. 
I would rather propose to replace the `GetMachineImages()` function in the `WorkerDelegate` interface with `UpdateMachineImagesStatus()`, and introduce `UpdateMachineDependenciesStatus()`. The actual implementation of the `WorkerDelegate` can then update the `.status` of the `Worker` itself.

WDYT?

dkistner: Yes, I agree. The provider extension only should be responsible for the worker provider status.

So lets replace the `GetMachineImages(context.Context) (runtime.Object, error)` method with a `UpdateMachineImageStatus(context.Context) error` method. The difference here is that `UpdateMachineImageStatus` need to update the worker provider status itself.

I would like to keep the `DeployMachineDependencies(context.Context) error /CleanupMachineDependencies(context.Context) error` methods (without returning a `runtime.Object` as they also update the worker provider status on their own) instead of one `UpdateMachineDependenciesStatus(context.Context) error` . They are called at different times in the flow and we need to do different things in those functions. We could probably also do everything in one method but this would make the implementations more complex.

WDYT? If you agree I would update the PR accordingly.

rfranzke: Yes, I think that sounds reasonable :)

dkistner:
@rfranzke I have updated the PR accordingly.

rfranzke:
@dkistner can you please sign the CLA?

rfranzke (dismissed):
/lgtm
/invite @ialidzhikov @prashanth26

prashanth26:
Hi colleagues,

I will take a look at this PR. However, I am OOO today. Might need time over the weekend to look into it. Just heads up. 
/assign

extensions/pkg/controller/worker/genericactuator/interface.go
```diff
@@ -43,10 +43,17 @@ type WorkerDelegate interface {
...
 	GenerateMachineDeployments(context.Context) (worker.MachineDeployments, error)
 
-	// GetMachineImages returns the list of used machine images for this `Worker` resource. It will be stored in the
-	// `.status.providerStatus` field of the `Worker` resource such that the controller can look up its provider-specific
-	// machine image information in case the required version has been removed from its componentconfig.
-	GetMachineImages(context.Context) (runtime.Object, error)
+	// UpdateMachineImagesStatus will store a list of used in machine images for
+	// this Worker resource in its provider status. The controller can look up its
+	// provider-specific machine image information in case the required version
+	// has been removed from its componentconfig.
```
ialidzhikov: Nit

```suggestion
	// has been removed from the `CloudProfile`.
```

Since https://github.com/gardener-attic/gardener-extensions/pull/455 the machine images are not maintained in the componentconfig of the extension controller.

ialidzhikov (dismissed):
/lgtm

dkistner:
Ok lets wait for @prashanth26 review. After addressing requested changes I will rebase and squash the commits.

extensions/pkg/controller/worker/genericactuator/actuator_delete.go
```diff
@@ -99,6 +99,11 @@ func (a *genericActuator) Delete(ctx context.Context, worker *extensionsv1alpha1
 		return errors.Wrapf(err, "failed deleting machine-controller-manager")
 	}
 
+	// Cleanup machine dependencies.
```
prashanth26, line 102: Does it make sense to redeploy `DeployMachineDependecies` (https://github.com/gardener/gardener/blob/11052936434e07373e7774be18ea98830575cdc2/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go#L73-L76) during deletion as well? 

We do similar operations for the deletion of MachineClasses here - https://github.com/gardener/gardener/blob/11052936434e07373e7774be18ea98830575cdc2/extensions/pkg/controller/worker/genericactuator/actuator_delete.go#L60-L65.  

I am just wondering if something like this would help to update the objects before the actual cleanup in a deletion loop. In the case of machine classes, we would be updating the cloud credentials on the secrets. I am only partially able to imagine the flow (i might be missing use cases) in the machine dependencies case, just wondering if something like this makes sense for machine dependencies as well? @dkistner you would have a better idea at this?

rfranzke: Probably it doesn't make sense. The reason for redeploying the machine classes is to update the machine class secrets. But the machine dependencies can be seen as infrastructure, and we also don't redeploy the infrastructure before deletion.

prashanth26: Okay yes. Even I am unable to come up with scenarios that would help to implement it this way. I agree with you, I am going to resolve this for now.

prashanth26 (dismissed):

rfranzke:
@dkistner anything open for this PR or can it be merged now?

dkistner:
No seems now to be ready. I have squashed the commits and rebased. Let's wait for the test and checks and then we can bring it in.

extensions/pkg/controller/worker/genericactuator/interface.go
```diff
@@ -43,10 +43,17 @@ type WorkerDelegate interface {
 	// GenerateMachineDeployments generates the configuration for the desired machine deployments.
 	GenerateMachineDeployments(context.Context) (worker.MachineDeployments, error)
 
-	// GetMachineImages returns the list of used machine images for this `Worker` resource. It will be stored in the
-	// `.status.providerStatus` field of the `Worker` resource such that the controller can look up its provider-specific
-	// machine image information in case the required version has been removed from its componentconfig.
-	GetMachineImages(context.Context) (runtime.Object, error)
+	// UpdateMachineImagesStatus will store a list of used in machine images for
```
danielfoehrKn:
```suggestion
	// UpdateMachineImagesStatus will store a list of machine images used by the machines associated with
```

dkistner: done

rfranzke (approved):
/remove status/author-action

prashanth26 (approved):

PR #2809, opened 2020-09-03
Enable race detector, switch to go test

timebertt:

/area quality robustness dev-productivity testing
/kind enhancement
/priority normal

This PR does 3 things around unit testing:
- `make test` now executes `go test` instead of `ginkgo` to make use of cached test results ([ref](https://golang.org/pkg/cmd/go/internal/test/))
  - `make test-cov` (executed in pipelines) still uses `ginkgo` which produces a more helpful output in case you want to reproduce pipeline test runs/failures (only my opinion, can be discussed)
- enables the go race detector to find at least obvious race conditions in unit tests ([ref](https://blog.golang.org/race-detector))
- makes the period health manager test race-free

Fixes #

**Release note**:

```noteworthy developer
`make test` now makes use of cached test results to reduce turn-around times when writing unit tests.
```
```improvement developer
The go race detector was enabled for unit tests.
```
```action developer
`./hack/test.sh` now executes tests via `go test` instead of `ginkgo`. Please adapt your extensions' `Makefile`s, if you use the vendored hack scripts.
```

### DIFF
--- a/hack/test.sh
+++ b/hack/test.sh
@@ -17,4 +17,4 @@ set -e
 
 echo "> Test"
 
-GO111MODULE=on ginkgo -mod=vendor $@
+GO111MODULE=on go test -race -mod=vendor $@ | grep -v 'no test files'
--- a/Makefile
+++ b/Makefile
@@ -165,12 +165,12 @@ format:
 
 .PHONY: test
 test:
-	@./hack/test.sh -r ./cmd/... ./extensions/... ./pkg/... ./plugin/...
+	@./hack/test.sh ./cmd/... ./extensions/... ./pkg/... ./plugin/...
 	$(MAKE) test-prometheus
 
 .PHONY: test-cov
 test-cov:
-	@./hack/test-cover.sh -r ./cmd/... ./extensions/... ./pkg/... ./plugin/...
+	@./hack/test-cover.sh ./cmd/... ./extensions/... ./pkg/... ./plugin/...
 	$(MAKE) test-prometheus
 
 .PHONY: test-cov-clean
--- a/hack/test-cover.sh
+++ b/hack/test-cover.sh
@@ -17,7 +17,7 @@ set -e
 
 echo "> Test Cover"
 
-"$(dirname $0)"/test.sh -cover $@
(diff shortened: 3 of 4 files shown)


rfranzke:
As these scripts are also used by the extensions, can you please add a release note that the invocations (in the `Makefile`s) might need to be adapted?

hack/test.sh
```diff
@@ -17,4 +17,4 @@ set -e
 
 echo "> Test"
 
-GO111MODULE=on ginkgo -mod=vendor $@
+GO111MODULE=on go test -race -mod=vendor $@
```
rfranzke: What about `--skipPackages`? See https://github.com/gardener/gardener-extension-provider-aws/blob/master/Makefile#L131
Is the recommended way to not enlist the `./test/...` argument anymore now?

timebertt: Nice catch! 
I checked all our extensions and it seems like, there is anyways no test under `./test` that is run on `make test`, so we can just leave it out.

If there is the necessity for it, we can use `$(go list ./test/... | grep -v directoriesToSkip)` to skip some directories, although I would argue that the cleaner way would be to separate unit/e2e tests that can be run with a simple `make test` from integration tests that can only be run in a specific environment (e.g. test machinery or by invocation with credentials and so on) and skip disabled/inactive tests in the test code via `ginkgo.Skip()`.

rfranzke: 👍 sounds good

timebertt:
> As these scripts are also used by the extensions, can you please add a release note that the invocations (in the `Makefile`s) might need to be adapted?

Good point. Done, PTAL :)

rfranzke (dismissed):
/lgtm

hack/test.sh
timuthy:
```suggestion
GO111MODULE=on go test -race -mod=vendor $@  | grep -v 'no test files'
```
Helps to get rid of the `[no test files]` noise.

timebertt: Good idea, done 👍

timuthy (approved):

rfranzke (approved):

PR #2816, opened 2020-09-04
Reduce unnecessary reconciles in ControllerInstallationRequired controller

timebertt:

/area quality scalability
/kind enhancement
/priority normal

Before, on startup of the gardenlet the ControllerInstallationRequired reconcile function was executed for every extension object in the Seed, although the function actually doesn't care about the enqueued object and rather does the same computation over and over again for each extension object.
This resulted in a lot of unnecessary API calls in large Seeds and might have caused the gardenlet to run into client-side rate limits when talking to the garden cluster, as the number of API calls were growing with the number of Shoots.

Now the ControllerInstallation artifacts enqueue an empty `reconcile.Request`, which make the reconcile functions only run once for each extension kind on startup instead of for each extension object.

Part of #2689

**Release note**:

```improvement operator
The gardenlet's `ControllerInstallationRequired` controller was optimized to execute less API calls to improve stability on startup of the gardenlet.
```

### DIFF
--- a/pkg/gardenlet/controller/federatedseed/extensions/artifact.go
+++ b/pkg/gardenlet/controller/federatedseed/extensions/artifact.go
@@ -181,9 +181,9 @@ func newControllerInstallationArtifact(gvk schema.GroupVersionKind, newObjFunc f
 
 	a.addEventHandlerFn = func() {
 		a.informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-			AddFunc:    createEnqueueFunc(a.queue),
-			UpdateFunc: createEnqueueOnUpdateFunc(a.queue, a.predicate),
-			DeleteFunc: createEnqueueFunc(a.queue),
+			AddFunc:    createEnqueueEmptyRequestFunc(a.queue),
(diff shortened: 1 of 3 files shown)


timuthy:
Can you elaborate how not enqueuing the actual object saves API calls if we still use the [informer](https://github.com/gardener/gardener/blob/e03a995c79fcaae59697269ef9473c7d70359234/pkg/gardenlet/controller/federatedseed/extensions/artifact.go#L142. ) to list and watch objects?

rfranzke (approved):
/lgtm
/assign @timuthy
/needs second-opinion

timebertt:
> Can you elaborate how not enqueuing the actual object saves API calls if we still use the [informer](https://github.com/gardener/gardener/blob/e03a995c79fcaae59697269ef9473c7d70359234/pkg/gardenlet/controller/federatedseed/extensions/artifact.go#L142.) to list and watch objects?

It saves API calls, because on startup the queue will then only contain one item per extension kind (instead of one item per extension object), so we will only execute reconciliation once per kind instead of per object. So the saved API calls are not on the informer side, but rather in the reconciliation func.

timuthy (approved):
/lgtm
Thanks for the explanation.

PR #2825, opened 2020-09-08
Remove GOMAXPROCS assignment

timuthy:

/area robustness
/kind cleanup
/priority normal

This PR removes the GOMAXPROCS assignment because since Go 1.5 it is set to the number of logical cores already by default, see [here](https://github.com/kubernetes/kubernetes/pull/63481). 

**Release note**:

```improvement operator
NONE
```


vpnachev (approved):
/lgtm

rfranzke (approved):

PR #2826, opened 2020-09-08
Clean up list call with deprecated purpose label

ialidzhikov:

/area dev-productivity
/kind cleanup
/priority normal

Clean up the list call for machine secrets with the deprecated label. It was introduced in https://github.com/gardener-attic/gardener-extensions/pull/608 to allow smooth transition. We can now clean it up. For some Shoots hibernated from a long period of time (~1 year) there still might be a machine class secret with the label  `garden.sapcloud.io/purpose` but it should already have a deletionTimestamp and should be deleted when machine-controller-manager is scaled up again (on wake up or deletion).

Contributes to https://github.com/gardener/gardener/issues/1649

**Release note**:

```improvement operator
The generic worker actuator does no longer consider machine class secrets with the label `garden.sapcloud.io/purpose`.
```


rfranzke (approved):
/lgtm

timuthy (approved):

PR #2828, opened 2020-09-08
Fix make test-cov

timebertt:

/area dev-productivity testing
/kind test
/priority normal

This PR
- fixes `make test-cov`, which currently yields an error like this one:
```
Test Suite Passed
awk: illegal field $(github.com/gardener/gardener/some/file.go:41.3,41.83), name "1"
 input record number 1, file
 source line number 1
make: *** [Makefile:154: test-cov] Error 2
```
- un-excludes `/test/` from coverage calculation
- removes duplicate invocation of `test-prometheus` on `make verify-extended`

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,4 +185,4 @@ test-prometheus:
 verify: check format test
 
 .PHONY: verify-extended
-verify-extended: install-requirements check-generate check format test-cov test-cov-clean test-prometheus
+verify-extended: install-requirements check-generate check format test-cov test-cov-clean
--- a/hack/test-cover.sh
+++ b/hack/test-cover.sh
@@ -24,8 +24,8 @@ COVERPROFILE_TMP="$(dirname $0)/../test.coverprofile.tmp"
 COVERPROFILE_HTML="$(dirname $0)/../test.coverage.html"
 
 echo "mode: set" > "$COVERPROFILE_TMP"
-find . -name "*.coverprofile" -type f | xargs cat | grep -v mode: | sort -r | awk '{if($$1 != last) {print $$0;last=$$1}}' >> "$COVERPROFILE_TMP"
-cat "$COVERPROFILE_TMP" | grep -vE "\.pb\.go|\/test\/|zz_generated" > "$COVERPROFILE"
+find . -name "*.coverprofile" -type f | xargs cat | grep -v mode: | sort -r | awk '{if($1 != last) {print $0;last=$1}}' >> "$COVERPROFILE_TMP"
+cat "$COVERPROFILE_TMP" | grep -vE "\.pb\.go|zz_generated" > "$COVERPROFILE"
 rm -rf "$COVERPROFILE_TMP"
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -171,7 +171,6 @@ test:
 .PHONY: test-cov
 test-cov:
 	@./hack/test-cover.sh ./cmd/... ./extensions/... ./pkg/... ./plugin/...
-	$(MAKE) test-prometheus
```
rfranzke: Can we rather exclude `test-prometheus` from the `verify-extended` rule so that `test-cov` stays similar to `test`?

hack/test-cover.sh
```diff
@@ -24,8 +24,8 @@ COVERPROFILE_TMP="$(dirname $0)/../test.coverprofile.tmp"
 COVERPROFILE_HTML="$(dirname $0)/../test.coverage.html"
 
 echo "mode: set" > "$COVERPROFILE_TMP"
-find . -name "*.coverprofile" -type f | xargs cat | grep -v mode: | sort -r | awk '{if($$1 != last) {print $$0;last=$$1}}' >> "$COVERPROFILE_TMP"
-cat "$COVERPROFILE_TMP" | grep -vE "\.pb\.go|\/test\/|zz_generated" > "$COVERPROFILE"
+find . -name "*.coverprofile" -type f | xargs cat | grep -v mode: | sort -r | awk '{if($1 != last) {print $0;last=$1}}' >> "$COVERPROFILE_TMP"
```
rfranzke, line 27: Why do we need this change/how was this working before?

hack/test-cover.sh
```diff
@@ -24,8 +24,8 @@ COVERPROFILE_TMP="$(dirname $0)/../test.coverprofile.tmp"
...
+cat "$COVERPROFILE_TMP" | grep -vE "\.pb\.go|zz_generated" > "$COVERPROFILE"
```
rfranzke, line 28: Why do we include `test` now?

hack/test-cover.sh
timebertt: I guess, it wasn't working before. 😄 
It seems, that this script was copied from `Makefile` to the shell script in https://github.com/gardener/gardener/commit/08c1bfd897eabab227a2d92a4d663d6a71d3ab35#diff-b67911656ef5d18c4ae36cb6741b7965L188 but the `$$`s were only needed for make to escape the special character `$`. 
However in the pipeline (on linux?), it seemed to work. But not on my Mac...

hack/test-cover.sh
timebertt: I assume, we don't want to exclude packages that contain `test` in their path, e.g. `/pkg/utils/test`.
`/test` is anyways not run in `make {test,test-cov}`, so I think, it isn't needed anymore.

rfranzke (approved):
/lgtm

Issue #2751, opened 2020-08-19
☂️-Issue for unit/integration tests in the extensions library


rfranzke:
**How to categorize this issue?**

/area dev-productivity robustness quality testing
/kind enhancement
/priority normal

**What would you like to be added**:
This is an umbrella issue for improving our unit test coverage in our extensions library ([`extensions` directory](https://github.com/gardener/gardener/tree/master/extensions)).
We lack unit and/or integration tests at too many places and should improve this. One important pain point is that we sometimes detect issues in the library only after revendoring it in an actual extension implementation (e.g., provider-aws). This results in additional work and effort that could be avoided by having good tests in the first place.

This issue lists the potential areas where we should improve:
- [x] ~~Unit tests for the generic worker actuator (https://github.com/gardener/gardener/tree/master/extensions/pkg/controller/worker/genericactuator)~~ (we don't want to invest here because we would rather like to rework/redesign the worker controller)
- [x] Dummy implementation of an extension controller (e.g., in `g/g/extensions/test`) which tests the basic controller flows and APIs of the extension library, potentially by leveraging https://sigs.k8s.io/controller-runtime/pkg/envtest
- ... (to be extended)

timebertt:
In addition to the mentioned tests we could
- [x] make use of [go-apidiff](https://github.com/joelanford/go-apidiff) to easily detect breaking changes in signatures of the extensions library (see https://github.com/kubernetes-sigs/controller-runtime/blob/master/hack/apidiff.sh for an example usage) -> tracked in https://github.com/gardener/gardener/issues/5301

timebertt:
We should definitely
- [x] add missing unit tests for the predicates provided by the extensions library (e.g. [`AddTypePredicate`](https://github.com/gardener/gardener/blob/092922b5765ad5e84cc93a40c87c5a80e5841023/extensions/pkg/predicate/predicate.go#L164) is untested but used in all extension controllers) ->tracked in https://github.com/gardener/gardener/pull/6249

timebertt:
We need integration tests for
- [ ] the extension webhooks (start a manager, register webhooks using the extensions library, do some dummy mutations, assert that they are applied, ...)

Currently there's no way of telling if a change to the webhook library code will break something there (ref https://github.com/gardener/gardener/pull/3393#issuecomment-760014129).

acumino:
> In addition to the mentioned tests we could
> 
[...]

PR #2831, opened 2020-09-09
Add Testrun for extensions controllers integration test

rfranzke:

/kind test
/priority normal

In order to enable the TestMachinery bot to execute the extensions controllers integration test (introduced in #2830) we have to define the `Testrun` in the `master` branch. This means that this PR must be merged before we can trigger `/test-single extensions-controllers` on #2830.

Part of #2751 

**Release note**:

```improvement developer
NONE
```

### DIFF
--- a/.ci/testruns/extensions-controllers.yaml
+++ b/.ci/testruns/extensions-controllers.yaml
@@ -1,0 +1,11 @@
+apiVersion: testmachinery.sapcloud.io/v1beta1
+kind: Testrun
+metadata:
+  generateName: tm-gardener-extensions-controllers-
+  namespace: default
+spec:
+  ttlSecondsAfterFinished: 172800 # 2 days
+  testflow:
+  - name: extensions-controllers
+    definition:
+      name: extensions-controllers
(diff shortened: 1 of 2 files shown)


.ci/testruns/extensions-controllers.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: testmachinery.sapcloud.io/v1beta1
+kind: Testrun
+metadata:
+  generateName: tm-gardener-extensions-controllers-
+  namespace: default
+spec:
+  ttlSecondsAfterFinished: 3600 # 1 hour
```
schrodit: This is the time to live of the testrun resource which means that the CR will be deleted from the testmachinery cluster after 1 hour.
The testmachinery dashboard uses this resource to show the test results which means that after this 1 hour you are unable to view the test results via the testmachinery dashboard.
Therefore, I would extend the ttl so that developers are able to see the test results for more than 1 hour after the test has finished.

rfranzke: I see, thanks, done.

schrodit (approved):
/lgtm

Issue #2829, opened 2020-09-09
Support HTTPS for /healthz, /readyz, /metrics and /debug endpoints


timuthy:
**How to categorize this issue?**

/area security
/kind enhancement

**What would you like to be added**:
[Similar to Gardenlet](https://github.com/gardener/gardener/blob/74fbee9bb67270899b0afd8d9cc1bd806bdf80c1/cmd/gardenlet/app/gardenlet.go#L440-L448), controller-manager, admission-controller, resource-manager, scheduler and seed-admission-controller should serve `/healthz`, `/readyz` and `/metrics` endpoints via HTTPS instead of HTTP. 
Additionally, `/debug` should also be served via HTTPS if enabled (ref https://github.com/gardener/gardener/issues/4567).

rfranzke:
Should we close this in favor of #4251? cc @timebertt

timuthy:
Hm, I don't think that one comes with the other. The scheduler for instance was changed to use the Controller-Runtime but the endpoints are still served via http only. Even on the contrary, with the Controller-Runtime serving these endpoints with https is not supported yet (https://github.com/kubernetes-sigs/controller-runtime/issues/979). Hence, we should only close it if we decide that this is not too relevant.

timebertt:
Once we refactor gardenlet to a native controller-runtime component (https://github.com/gardener/gardener/issues/4251), i.e. use a c-r manager to expose metrics and health endpoints, we will also be back to plain http for `/healthz`, `/readyz`, `/metrics` and `/debug` there as well.

Do we consider this to be important?
If so, we could
- a) tackle https://github.com/kubernetes-sigs/controller-runtime/issues/1007 and contribute the required mechanisms to c-r
  - in that case, we probably would want to allow reusing the same listener (port, certs, etc.) for webhook server, health and metrics endpoints as our components currently use one port for everything
- b) use [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) as a sidecar container for our components like described in https://github.com/kubernetes-sigs/controller-runtime/pull/993#issuecomment-645045329
  - in addition to TLS, this would allow configuring RBAC for the endpoints (e.g. like it is the case for API servers)

If not and we want to go with HTTP for the endpoints, we need to
- c) adapt seed-prometheus scrape config to allow scraping with `http` scheme:  https://github.com/gardener/gardener/pull/4706#discussion_r713648371

[...]

PR #2832, opened 2020-09-09
Move GCM webhooks to dedicated Admission Controller

timuthy:

/area scalability
/kind enhancement
/priority normal

This PR moves Gardener-Controller-Manager related webhooks to a dedicated component/binary called `Gardener-Admission-Controller`. Introducing a dedicated webhook server allows us to efficiently run multiple replicas in the garden cluster and also follows a separation of concerns.

Prerequisite for #2781

- I didn't remove the HTTPS settings for GCM because in the near future remaining endpoints should be protected by TLS (#2829)
- The namespace deletion validator now works with the cached runtime-client instead of client + project informer + shoot  informer.

**Release note**:

```noteworthy operator
Gardener validating webhooks have been moved from the `Gardener-Controller-Manager` to a dedicated component `Gardener-Admission-Controller`. Therefore, new values have been added to the Gardener Helm chart (`charts/gardener/controlplane`). Please consult the documentation (`docs/concepts/admission-controller.md`) for more information about the `Gardener-Admission-Controller`.
```
```noteworthy developer
Gardener validating webhooks have been moved from the `Gardener-Controller-Manager` to a dedicated component `Gardener-Admission-Controller`. Please run `make dev-setup` to create the necessary configuration for this component for local development use-cases. `make start-admission-controller` starts the new component locally.
```


charts/gardener/controlplane/charts/runtime/templates/admission-controller/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+        release: "{{ .Release.Name }}"
+        heritage: "{{ .Release.Service }}"
+        {{- if .Values.global.admission.podLabels }}
+{{ toYaml .Values.global.admission.podLabels | indent 8 }}
+        {{- end }}
+    spec:
+      {{- if not .Values.global.admission.kubeconfig }}
+      serviceAccountName: {{ required ".Values.global.admission.serviceAccountName is required" .Values.global.admission.serviceAccountName }}
+      {{- end }}
+      {{- if gt (int .Values.global.admission.replicaCountreplicaCount) 1 }}
```
timebertt:
```suggestion
      {{- if gt (int .Values.global.admission.replicaCount) 1 }}
```

charts/gardener/controlplane/charts/runtime/templates/admission-controller/secret-cert.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: gardener-admission-controller-cert
+  namespace: garden
+  labels:
+    app: gardener
+    role: admission-controller
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+type: Opaque
```
timebertt, line 13: Would it make sense, to use `kubernetes.io/tls` as the secret type here?
I see, we also don't use it in the other charts...
```suggestion
type: kubernetes.io/tls
```

charts/gardener/controlplane/charts/runtime/templates/admission-controller/serviceaccount.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if and .Values.global.admission.enabled (not .Values.global.admission.kubeconfig) }}
+apiVersion: v1
```
timebertt, line 2: Does the admission-controller need any RBAC rules?

pkg/admissioncontroller/apis/config/types.go
```diff
@@ -0,0 +1,64 @@
...
+type ServerConfiguration struct {
+	// HTTPS is the configuration for the HTTPS server.
+	HTTPS HTTPSServer
+}
+
+// Server contains information for HTTP(S) server configuration.
+type Server struct {
+	// BindAddress is the IP address on which to listen for the specified port.
+	BindAddress string
+	// Port is the port on which to serve unsecured, unauthenticated access.
```
timebertt: Hmm, `unsecured, unauthenticated access` is not true here / might be confusing, right?

cmd/gardener-admission-controller/app/gardener_admission_controller.go
```diff
@@ -0,0 +1,221 @@
+package app
```
timebertt, line 15: License header missing

cmd/gardener-admission-controller/main.go
```diff
@@ -0,0 +1,21 @@
+package main
```
timebertt, line 15: License header missing

cmd/gardener-admission-controller/main.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/gardener/gardener/cmd/utils"
+
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+)
+
+func main() {
+	ctx := utils.ContextFromStopChannel(signals.SetupSignalHandler())
+	command := app.NewCommandStartGardenerAdmissionController(ctx)
+
+	if err := command.Execute(); err != nil {
```
timebertt: Can you pass down the context via `ExecuteContext`? (You can then get it in `RunE` via `cmd.Context()`)
```suggestion
	command := app.NewCommandStartGardenerAdmissionController()

	if err := command.ExecuteContext(ctx); err != nil {
```

cmd/gardener-admission-controller/app/gardener_admission_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		Use:   "gardener-admission-controller",
+		Short: "Launch the Gardener admission controller",
+		Long:  `The Gardener admission controller serves a validation webhook endpoint for resources in the garden cluster.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := opts.configFileSpecified(); err != nil {
+				panic(err)
+			}
+			if err := opts.validate(args); err != nil {
+				panic(err)
+			}
```
timebertt: Can you return the errors instead of panicking?

cmd/gardener-admission-controller/app/gardener_admission_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+)
+
+// options has all the context and parameters needed to run a Gardener admission controller.
+type options struct {
+	// ConfigFile is the location of the Gardener controller manager's configuration file.
+	ConfigFile string
```
timebertt: I think, this field can be unexported.
```suggestion
	configFile string
```

cmd/gardener-admission-controller/app/gardener_admission_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	data, err := ioutil.ReadFile(file)
+	if err != nil {
+		return nil, err
+	}
+	return o.decodeConfig(data)
+}
+
+func (o *options) configFileSpecified() error {
+	if len(o.ConfigFile) == 0 {
+		return fmt.Errorf("missing Gardener controller manager config file")
```
timebertt:
```suggestion
		return fmt.Errorf("missing Gardener admission controller config file")
```

cmd/gardener-admission-controller/app/gardener_admission_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	ClientMap clientmap.ClientMap
+}
+
+func (o *options) run(ctx context.Context) error {
+	c, err := o.loadConfigFromFile(o.ConfigFile)
+	if err != nil {
+		return err
+	}
+
+	if c, err = o.applyDefaults(c); err != nil {
```
timebertt: I am a bit confused by the version handling here...
Why are you first decoding the config in the internal version, then converting it back to the external one, default it and then convert it to the internal version again? Is this somehow required?

cmd/gardener-admission-controller/app/gardener_admission_controller.go
```diff
@@ -0,0 +1,221 @@
...
+
+	// Initialize logger
+	logger := logger.NewLogger(cfg.LogLevel)
+	logger.Info("Starting Gardener admission controller...")
+
+	// Prepare a Kubernetes client object for the Garden cluster which contains all the Clientsets
+	// that can be used to access the Kubernetes API.
+	if kubeconfig := os.Getenv("KUBECONFIG"); kubeconfig != "" {
+		cfg.GardenClientConnection.Kubeconfig = kubeconfig
+	}
```
timebertt, line 165: Shouldn't the env var only be used as a fallback, if the config doesn't provide a kubeconfig, instead of overwriting the config?

pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion.go
```diff
@@ -41,19 +40,16 @@ import (
...
-
 	codecs serializer.CodecFactory
 }
 
 // NewValidateNamespaceDeletionHandler creates a new handler for validating namespace deletions.
-func NewValidateNamespaceDeletionHandler(k8sGardenClient kubernetes.Interface, projectLister gardencorelisters.ProjectLister, shootLister gardencorelisters.ShootLister) http.HandlerFunc {
+func NewValidateNamespaceDeletionHandler(k8sGardenClient kubernetes.Interface) http.HandlerFunc {
 	scheme := runtime.NewScheme()
 	utilruntime.Must(corev1.AddToScheme(scheme))
 	utilruntime.Must(admissionregistrationv1beta1.AddToScheme(scheme))
```
timebertt: Do we still need `admissionregistrationv1beta1`?

cmd/gardener-admission-controller/app/gardener_admission_controller.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// Run starts the Gardener admission controller.
+func (a *AdmissionController) Run(ctx context.Context) error {
+	k8sGardenClient, err := a.ClientMap.GetClient(ctx, keys.ForGarden())
+	if err != nil {
+		return err
+	}
+
+	if err := a.ClientMap.Start(ctx.Done()); err != nil {
```
timebertt, line 194: I think, one problem that we might hit here, is that we now don't wait for the needed informers to sync before we start the webhook server.
One option would be to retrieve an informer for every resource we want to talk to and then use the `HasSynced` funcs to wait for the caches to sync.

WDYT about constructing the webhook handlers before starting the server, then passing the client to some kind of `Initialize` func there and let the handlers retrieve the needed informers and necessary `HasSynced` funcs?

Something like this would help to make sure, the webhook can directly serve requests off the cache, when it starts accepting traffic, and not to introduce to much unneeded latency, when calling the gardener api.

timebertt (changes requested):
Nice PR, @timuthy! Very clean refactoring. Thanks!
I only have minor comments and nit suggestions :)

Dockerfile
```diff
@@ -55,6 +55,15 @@ WORKDIR /
 
 ENTRYPOINT ["/gardenlet"]
 
+############# admission-controller #############
```
rfranzke, line 58: Just a reminder that we will need to adapt the `meta/ci` ref

docs/README.md
```diff
@@ -9,6 +9,7 @@
 ## Concepts
 
 * Components
+    * [Gardener Admission Controller](concepts/admission-controller.md)
```
rfranzke: Can you move it after the scheduler but before the seed admission controller?

docs/concepts/admission-controller.md
```diff
@@ -0,0 +1,15 @@
+# Gardener Admission Controller
+
+While the Gardener API server works with [admission plugins](./apiserver_admission_plugins.md) to validate and mutate resources belonging to Gardener related API groups, e.g. `core.gardener.cloud`, the same is needed for resources belonging to non-Gardener API groups as well.
```
rfranzke:
```suggestion
While the Gardener API server works with [admission plugins](./apiserver_admission_plugins.md) to validate and mutate resources belonging to Gardener related API groups, e.g. `core.gardener.cloud`, the same is needed for resources belonging to non-Gardener API groups as well, e.g. `Secret`s in the `core` API group.
```

example/20-componentconfig-gardener-admission-controller.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: AdmissionControllerConfiguration
+gardenClientConnection:
+  acceptContentTypes: application/json
+  contentType: application/json
+  qps: 100
+  burst: 130
+server:
+  https:
+    bindAddress: 0.0.0.0
+    port: 2719
```
rfranzke:
```suggestion
    port: 2721
```

hack/update-codegen.sh
```diff
@@ -86,6 +86,25 @@ bash "${PROJECT_ROOT}"/vendor/k8s.io/code-generator/generate-internal-groups.sh
...
+  "config:v1alpha1" \
+  -h "${PROJECT_ROOT}/hack/LICENSE_BOILERPLATE.txt"
+
+bash "${PROJECT_ROOT}"/vendor/k8s.io/code-generator/generate-internal-groups.sh \
+  conversion \
+  github.com/gardener/gardener/pkg/client/admissioncontrollerconfig \
+  github.com/gardener/gardener/pkg/admissioncontroller/apis \
+  github.com/gardener/gardener/pkg/admissioncontroller/apis \
+  "config:v1alpha1" \
+  --extra-peer-dirs=github.com/gardener/gardener/pkg/controllermanager/apis/config,github.com/gardener/gardener/pkg/controllermanager/apis/config/v1alpha1,k8s.io/apimachinery/pkg/apis/meta/v1,k8s.io/apimachinery/pkg/conversion,k8s.io/apimachinery/pkg/runtime,k8s.io/component-base/config,k8s.io/component-base/config/v1alpha1 \
```
rfranzke:
```suggestion
  --extra-peer-dirs=github.com/gardener/gardener/pkg/admissioncontroller/apis/config,github.com/gardener/gardener/pkg/admissioncontroller/apis/config/v1alpha1,k8s.io/apimachinery/pkg/apis/meta/v1,k8s.io/apimachinery/pkg/conversion,k8s.io/apimachinery/pkg/runtime,k8s.io/component-base/config,k8s.io/component-base/config/v1alpha1 \
```

pkg/admissioncontroller/apis/config/v1alpha1/dafaults_test.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: fix typo in file name

pkg/admissioncontroller/apis/config/v1alpha1/dafaults_test.go
```diff
@@ -0,0 +1,40 @@
...
+		BeforeEach(func() {
+			obj = &AdmissionControllerConfiguration{}
+		})
+
+		It("should correctly default the controller manager configuration", func() {
+			SetDefaults_AdmissionControllerConfiguration(obj)
+
+			Expect(obj.LogLevel).To(Equal("info"))
+			Expect(obj.Server.HTTPS.BindAddress).To(Equal("0.0.0.0"))
+			Expect(obj.Server.HTTPS.Port).To(Equal(2719))
```
rfranzke:
```suggestion
			Expect(obj.Server.HTTPS.Port).To(Equal(2721))
```

pkg/admissioncontroller/apis/config/v1alpha1/defaults.go
```diff
@@ -0,0 +1,36 @@
...
+// SetDefaults_AdmissionControllerConfiguration sets defaults for the configuration of the Gardener admission controller.
+func SetDefaults_AdmissionControllerConfiguration(obj *AdmissionControllerConfiguration) {
+	if len(obj.LogLevel) == 0 {
+		obj.LogLevel = "info"
+	}
+	if len(obj.Server.HTTPS.BindAddress) == 0 {
+		obj.Server.HTTPS.BindAddress = "0.0.0.0"
+	}
+	if obj.Server.HTTPS.Port == 0 {
+		obj.Server.HTTPS.Port = 2719
```
rfranzke:
```suggestion
		obj.Server.HTTPS.Port = 2721
```

rfranzke (changes requested):

charts/gardener/controlplane/charts/runtime/templates/admission-controller/secret-cert.yaml
timuthy: Tbh, neither am I sure about the benefits or consequences nor have I found something during a quick research except that is has some benefits for `kubectl`. To be sure, I'd like to stick to `generic` for now since we use it in other charts as well if you agree?

cmd/gardener-admission-controller/app/gardener_admission_controller.go
timuthy: This is the way you apply the defaults registered in the scheme for this kind. We did the same before in GCM, so I didn't question this handling.

cmd/gardener-admission-controller/app/gardener_admission_controller.go
timuthy: We do this for all of our components this way. Any objection to change it generally @rfranzke ?

cmd/gardener-admission-controller/app/gardener_admission_controller.go
timuthy: I was a bit more pragmatic and hope you are fine with the changes.

Dockerfile
timuthy: Thanks, will do as soon as this PR is merged.

charts/gardener/controlplane/charts/runtime/templates/admission-controller/secret-cert.yaml
timebertt: Yep, agreed.

cmd/gardener-admission-controller/app/gardener_admission_controller.go
rfranzke: Let's simply keep it as it is/also was in the past to not break existing usages. It's not an important decision here, either you go left or right, and we decided to go left. :)

rfranzke (commented):
No more comments from my side, I'll try to play with it locally later today.

cmd/gardener-admission-controller/app/gardener_admission_controller.go
timebertt: Hmm, ok. I would have to test that, but I would definitely expect some latency on startup in larger landscapes because the handler has to load all Shoots and all Projects. So, I am still a bit concerned about this.

Another option to mitigate this situation and still keep it pragmatic in code, would be to add a `startupProbe` (or `livenessProbe` with some `initialDelaySeconds`).
Still, this would only be based on heuristics and not really detect any long running initial syncs...
/cc @rfranzke

charts/gardener/controlplane/charts/runtime/templates/admission-controller/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+        imagePullPolicy: {{ .Values.global.admission.image.pullPolicy }}
+        command:
+        - /gardener-admission-controller
+        - --config=/etc/gardener-admission-controller/config/config.yaml
+        {{- if .Values.global.admission.resources }}
+        resources:
+{{ toYaml .Values.global.admission.resources | indent 10 }}
+        {{- end }}
+        terminationMessagePath: /dev/termination-log
+        terminationMessagePolicy: File
```
timebertt, line 83: Can you add some health checks? I think we should at least add a TCP liveness and readiness probe.

cmd/gardener-admission-controller/app/gardener_admission_controller.go
timebertt: Sorry, I have just now seen that you added the `HasSynced` funcs, that's great :)

pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion.go
```diff
@@ -41,20 +42,35 @@ import (
...
-
 	codecs serializer.CodecFactory
 }
 
 // NewValidateNamespaceDeletionHandler creates a new handler for validating namespace deletions.
-func NewValidateNamespaceDeletionHandler(k8sGardenClient kubernetes.Interface, projectLister gardencorelisters.ProjectLister, shootLister gardencorelisters.ShootLister) http.HandlerFunc {
+func NewValidateNamespaceDeletionHandler(ctx context.Context, k8sGardenClient kubernetes.Interface) (http.HandlerFunc, error) {
+	// Initialize caches here to ensure http requests can be served quicker with pre-syncronized caches.
+	var hasSyncFuncs []cache.InformerSynced
+	informer, err := k8sGardenClient.Cache().GetInformerForKind(ctx, corev1.SchemeGroupVersion.WithKind("secret"))
```
timebertt: I think `Kind` should always be uppercase, right?
Anyways, can you use `GetInformer` with an API object to get type safety?
```suggestion
	informer, err := k8sGardenClient.Cache().GetInformer(ctx, &corev1.Secret{})
```

pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion.go
```diff
@@ -41,20 +42,35 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	hasSyncFuncs = append(hasSyncFuncs, informer.HasSynced)
+	informer, err = k8sGardenClient.Cache().GetInformerForKind(ctx, corev1.SchemeGroupVersion.WithKind("shoot"))
```
timebertt: `Shoot` is not in `corev1` 😄 
```suggestion
	informer, err = k8sGardenClient.Cache().GetInformer(ctx, &gardencorev1beta1.Shoot{})
```

pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion.go
```diff
@@ -41,20 +42,35 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	hasSyncFuncs = append(hasSyncFuncs, informer.HasSynced)
+
+	for _, hasSync := range hasSyncFuncs {
+		if !hasSync() {
+			return nil, errors.New("timed out waiting for caches to sync")
+		}
+	}
```
timebertt: I think the `HasSynced` funcs themselves don't wait, that's what `cache.WaitForCacheSync` is for, right?
```suggestion
	if !cache.WaitForCacheSync(ctx.Done(), hasSyncFuncs...) {
		return nil, errors.New("timed out waiting for caches to sync")
	}
```

timebertt (dismissed):
Thanks for addressing my feedback.
A few more comments inline :)

pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion.go
timuthy: Yes, you are right. I pushed it too early with the outstanding comments yesterday. I think we should also combine it with a readiness probe.

charts/gardener/controlplane/charts/runtime/templates/admission-controller/deployment.yaml
timuthy: See discussion about health checks [here](https://github.com/gardener/gardener-extension-provider-gcp/pull/38#discussion_r387523877) and why we did not add them yet to any webhook component. Readiness is fine since we now wait for the cache sync.

timuthy:
@rfranzke @timebertt let me address the last comments before you play around with it. I'll let you know asap.

timuthy:
Thanks for you feedback which is now incorporated in the PR:
- Add `readyz` and `healthz` endpoints as well as readiness and liveness checks.
- Fixed wait for cache sync and informer retrieval bug.

charts/gardener/controlplane/charts/runtime/templates/admission-controller/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+        {{- if .Values.global.admission.resources }}
+        resources:
+{{ toYaml .Values.global.admission.resources | indent 10 }}
+        {{- end }}
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: {{ required ".Values.global.admission.config.server.https.port is required" .Values.global.admission.config.server.https.port }}
+            scheme: HTTPS
+          initialDelaySeconds: 300
```
rfranzke, line 73: Any reason for 5m initial delay?

pkg/server/handler.go
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import "net/http"
+
+// StaticOKHandler returns a http handler which returns 200/ok on every call.
+func StaticOKHandler() http.Handler {
```
rfranzke: Can't you use https://github.com/gardener/gardener/blob/master/pkg/healthz/default.go like in https://github.com/gardener/gardener/blob/master/cmd/gardener-controller-manager/app/gardener_controller_manager.go#L346?

timebertt:
No additional comments from my side.
Looks great :)

charts/gardener/controlplane/charts/runtime/templates/admission-controller/deployment.yaml
timuthy: Yes, it's the timeout for the cache sync wait. Otherwise we might already contact the `healthz` endpoint even though the initialization hasn't finished yet.

pkg/server/handler.go
timuthy: Yes, I can use it as an alternative. @timebertt and I agreed that for now a "static" handler is sufficient but we might want to improve it in the future. So I decided to add a simple static handler for that in order to make it obvious to developer that no health checks at all are executed. WDYT?

rfranzke: Maybe I'm missing something, but I don't see a difference in the static handler compared to this default healthz checker, or?

timuthy: I'm thinking that we don't need no health checker with all its functionalities (start, set, stop, ...) if there are no real health checks happening. But I'm also happy to change it if it's beneficial.

timuthy: I've changed it in dfc526810fe87b11ea8a0b9f0593841558b71091. Your call :)

rfranzke (approved):
/lgtm

rfranzke:
Thanks @timuthy! Kindly update the pipeline definitions now that it's merged :)

PR #2834, opened 2020-09-09
Do not wait for MachineDeployment deletion when mcm is scaled down

### LINKED ISSUE #2784: Cannot remove worker pool when Shoot is hibernated

/area quality ops-productivity
/kind bug
/priority normal

**What happened**:
Currently Shoot reconciliation gets stuck when a worker pool of hibernated Shoot cluster is removed.

Shoot reconciliation to do not stuck.

1. Create Shoot with 2 worker pools

2. Hibernate the Shoot

3. After successful hibernation, delete one of the worker pools

4. Ensure that Shoot reconciliation hangs forever at step `Waiting until shoot worker nodes have been reconciled`

There is no machine-controller-manager Pods in the control plane.
```
$ k -n shoot--foo--bar get po
No resources found in shoot--foo--bar namespace.
```

But the provider extension waits forever unwanted machine deployments to be deleted:

```
{"level":"info","ts":"2020-08-27T06:50:58.463Z","logger":"aws-worker-actuator.worker-actuator","msg":"Waiting until unwanted machine deployment is deleted: shoot--foo--bar-cpu-worker2-z1 ...","worker":"shoot--foo--bar/bar"}

{"level":"info","ts":"2020-08-27T06:51:03.463Z","logger":"aws-worker-actuator.worker-actuator","msg":"Waiting until unwanted machine deployment is deleted: shoot--foo--bar-cpu-worker2-z1 ...","worker":"shoot--foo--bar/bar"}
```

**Environment**:

- Gardener version: v1.9.0
- Others:

ialidzhikov:

/area quality ops-productivity
/kind task
/priority normal

Currently the generic worker actuator waits until unwanted MachineDeployments are removed from the store (for example worker removal). This PR adds a check to perform such wait only when the machine-controller-manager is scaled up.
When Shoot is hibernated and a worker is removed, then the corresponding MachineDeployment will receive a deletionTimestamp but its finalizer will be released on next machine-controller-manager scale up (Shoot wake up or Shoot deletion).

Fixes #2784 

**Release note**:

```improvement operator
The generic worker actuator does now wait for undesired MachineDeployment deletion only when the machine-controller-manager is scaled up.
```

### DIFF
--- a/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
@@ -173,9 +173,16 @@ func (a *genericActuator) Reconcile(ctx context.Context, worker *extensionsv1alp
 		return errors.Wrapf(err, "failed to cleanup the orphaned machine class secrets")
 	}
 
-	// Wait until all unwanted machine deployments are deleted from the system.
-	if err := a.waitUntilUnwantedMachineDeploymentsDeleted(ctx, logger, worker, wantedMachineDeployments); err != nil {
-		return errors.Wrapf(err, "error while waiting for all undesired machine deployments to be deleted")
+	replicas, err := replicaFunc()
+	if err != nil {
+		return errors.Wrapf(err, "failed to get machine-controller-manager replicas")
+	}
+
+	if replicas > 0 {
+		// Wait until all unwanted machine deployments are deleted from the system.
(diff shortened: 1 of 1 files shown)


vpnachev (approved):
/lgtm
/needs/second-opinion

prashanth26 (approved):
Interesting approach. I didn't think of this way while trying to figure out the solution. It makes sense here. 

/lgtm

rfranzke (approved):

PR #2838, opened 2020-09-10
resize prometheus pvc size to 20Gi

knightXun:
resize prometheus pvc size to 20Gi, because the minimum pvc size of alicloud is 20Gi.

### DIFF
--- a/charts/seed-bootstrap/values.yaml
+++ b/charts/seed-bootstrap/values.yaml
@@ -2,7 +2,7 @@ cloudProvider: aws
 
 prometheus:
   port: 9090
-  storage: 10Gi
+  storage: 20Gi
 
 aggregatePrometheus:
   port: 9090


gardener-robot-ci-1:
Thank you @knightXun for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

rfranzke:
Can you rather configure your alicloud `Seed` object appropriately (https://github.com/gardener/gardener/blob/master/example/50-seed.yaml#L59)? If you set it to `20Gi` then this value will be taken into account. This way we keep the gardener/gardener source code free from provider-specifics.
/hold
/status author-action

rfranzke:
@knightXun does it work? :) Does it make sense to close the PR then?

PR #2840, opened 2020-09-11
Switch off gardener-apiserver's vertical scale down by hvpa

ggaurav10:

/area control-plane
/area auto-scaling
/priority normal

Switch off gardener-apiserver's vertical scale down by hvpa

Fixes #

**Release note**:

```improvement operator
Switch off gardener-apiserver's vertical scale down by hvpa. Since VPA doesn't seem to be handling the resource usage spikes well, it can result in repeated oomkills. Turning off scale down should reduce such occurancces.
```

### DIFF
--- a/charts/gardener/controlplane/charts/runtime/templates/apiserver/hvpa.yaml
+++ b/charts/gardener/controlplane/charts/runtime/templates/apiserver/hvpa.yaml
@@ -58,7 +58,7 @@ spec:
 {{- end }}
     scaleDown:
       updatePolicy:
-        updateMode: "Auto"
+        updateMode: "Off"
 {{- if .Values.global.apiserver.hvpa.vpaScaleDownStabilization }}
 {{ toYaml .Values.global.apiserver.hvpa.vpaScaleDownStabilization | indent 6 }}
 {{- end }}


rfranzke (approved):
/lgtm
/invite @amshuman-kr

danielfoehrKn:
Does this mean that the Gardener API server is never automatically scaled down?
I suppose the current scale down behaviour is causing issues.

Can you maybe explain a little bit why this is done and also write it in the release notes?

amshuman-kr (approved):

amshuman-kr:
> Does this mean that the Gardener API server is never automatically scaled down?

Only vertical scaling down is switched off. HPA scaling down might still happen.

> I suppose the current scale down behaviour is causing issues.

Yes. The VPA recommender is not coping well with spikes in load and scaling down potentially makes this worse. Not scaling down can potentially give the pods a bit more room during spikes in load (if spikes are smaller than past spikes).

> Can you maybe explain a little bit why this is done and also write it in the release notes?

@ggaurav10 Could you please update the release note with a bit more explanation/rationale.

danielfoehrKn (approved):

timebertt:
> Only vertical scaling down is switched off. HPA scaling down might still happen.

So, this means, if somehow g-api was scaled up to like 8Gi or even more, it will keep running with such high requests?

Can't we rather somehow stabilize the scale down behaviour, if it was recently scaled up / is being rolled / was OOM killed?

> The VPA recommender is not coping well with spikes in load and scaling down potentially makes this worse.

I thought, VPA would watch events to detect OOMs and add artificial memory bumps to its metrics model. Isn’t this the case? Or why doesn't it work for our case?

charts/gardener/controlplane/charts/runtime/templates/apiserver/hvpa.yaml
```diff
@@ -58,7 +58,7 @@ spec:
 {{- end }}
     scaleDown:
       updatePolicy:
-        updateMode: "Auto"
+        updateMode: "Off"
```
timebertt, line 61: Can we please make this configurable instead of turning it off for all people, although they might not see problems with that in their landscapes?

timebertt (changes requested):

PR #2847, opened 2020-09-14
Enabling SNI feature gate is now graceful

### LINKED ISSUE #22: [BUG] DWD cannot talk to the shoot API servers anymore when APIServerSNI feature gate is enabled in g/g
**Describe the bug:**
When the `APIServerSNI` feature gate is enabled in g/g then the DWD cannot talk to the kube-apiserver's anymore although they are perfectly reachable:

```
I0624 07:14:37.040182       1 prober.go:308] shoot-kube-apiserver/shoot--dev--foo/external: probe failed with error: Get https://api.foo.dev.internal.example.com/version?timeout=10s: context deadline exceeded. Will retry...
I0624 07:14:37.040244       1 prober.go:320] shoot-kube-apiserver/shoot--dev--foo/external: probe result: &scaler.probeResult{lastError:(*url.Error)(0xc0007e0540), resultRun:3}
```

This leads to the problem that the KCMs are being scaled down.
This problem resolves on its own when you restart DWD or after some time ~15-20min.

**Expected behaviour:**
DWD should continue to reach the kube-apiservers

/assign @amshuman-kr @mvladev

mvladev:

/area control-plane
/kind enhancement
/priority normal

With this change when SNI was disabled and then enabled during the reconciliation the following happens:

- kube-apiserver service is annotated with the istio's `exportTo` annotation to add it to the service mesh
- kube-apiserver service is kept of type `LoadBalancer` to keep traffic in
the cluster running as DNS cache will force traffic to go to the current
kube-apiserver Loadbalancer IP.
- istio resources are added so traffic from the service mesh can be send
to the istio ingress gateway (for newer clients)
- DNS entries are changed to point to the istio IGW Loadbalancer
- at the end of the reconciliation when the tunnel connection is working,
update the kube-apiserver to type of `ClusterIP`.

Fixes gardener/dependency-watchdog#22

I would recommend to move the logic of kube-apiserver service management + SNI in a a separate controller. This would allow to update the DNSEntries only when the DNS's TTL has passed and ensure with 100% that end-users experience only a connection close and are able to to quickly reconnect to the istio IGW.

**Release note**:

```improvement operator
Enabling SNI feature gate now changes the type of kube-apiserver Service to `ClusterIP` only after `DNSEntries` are updated to point to the istio ingress gateway.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -91,6 +91,7 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		staticNodesCIDR                = o.Shoot.Info.Spec.Networking.Nodes != nil
 		useSNI                         = botanist.APIServerSNIEnabled()
 		generation                     = o.Shoot.Info.Generation
+		sniPhase                       = botanist.Shoot.Components.ControlPlane.KubeAPIServerSNIPhase
 		requestControlPlanePodsRestart = controllerutils.HasTask(o.Shoot.Info.Annotations, common.ShootTaskRestartControlPlanePods)
 
 		g                      = flow.NewGraph("Shoot cluster reconciliation")
@@ -119,7 +120,9 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 		})
 		deployKubeAPIServerService = g.Add(flow.Task{
 			Name: "Deploying Kubernetes API server service in the Seed cluster",
-			Fn: flow.TaskFn(botanist.Shoot.Components.ControlPlane.KubeAPIServerService.Deploy).
+			Fn: flow.TaskFn(func(ctx context.Context) error {
+				return botanist.DeployKubeAPIService(ctx, sniPhase)
+			}).
 				RetryUntilTimeout(defaultInterval, defaultTimeout).
 				SkipIf(o.Shoot.HibernationEnabled && !useSNI),
 			Dependencies: flow.NewTaskIDs(deployNamespace, ensureShootClusterIdentity),
@@ -346,6 +349,23 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 			Name:         "Waiting until the Kubernetes API server can connect to the Shoot workers",
 			Fn:           flow.TaskFn(botanist.WaitUntilTunnelConnectionExists).SkipIf(o.Shoot.HibernationEnabled),
 			Dependencies: flow.NewTaskIDs(deployManagedResources, waitUntilNetworkIsReady, waitUntilWorkerReady),
+		})
+		_ = g.Add(flow.Task{
+			Name: "Finishing Kubernetes API server service SNI transition",
+			Fn: flow.TaskFn(func(ctx context.Context) error {
+				return botanist.DeployKubeAPIService(ctx, sniPhase.Done())
+			}).
+				RetryUntilTimeout(defaultInterval, defaultTimeout).
+				SkipIf(o.Shoot.HibernationEnabled).
+				DoIf(sniPhase == component.PhaseEnabling || sniPhase == component.PhaseDisabling),
+			Dependencies: flow.NewTaskIDs(waitUntilTunnelConnectionExists),
+		})
+		_ = g.Add(flow.Task{
+			Name: "Deleting SNI resources if SNI is disabled",
+			Fn: flow.TaskFn(botanist.Shoot.Components.ControlPlane.KubeAPIServerSNI.Destroy).
+				RetryUntilTimeout(defaultInterval, defaultTimeout).
+				DoIf(sniPhase.Done() == component.PhaseDisabled),
+			Dependencies: flow.NewTaskIDs(waitUntilTunnelConnectionExists),
 		})
 		deploySeedMonitoring = g.Add(flow.Task{
 			Name:         "Deploying Shoot monitoring stack in Seed",
--- a/pkg/operation/botanist/component/phase.go
+++ b/pkg/operation/botanist/component/phase.go
@@ -1,0 +1,48 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package component
+
+// Phase is the phase of a component.
+type Phase int
+
+const (
+	// PhaseUnknown is in an unknown component phase.
+	PhaseUnknown Phase = iota
+	// PhaseEnabled is when a component was enabled before and it's still active.
+	PhaseEnabled
+	// PhaseDisabled is when a component was disabled before and it's still disabled.
+	PhaseDisabled
+	// PhaseEnabled is when a component was disabled before, but it's being activated.
+	PhaseEnabling
+	// PhaseDisabling is when a component was enabled before, but it's being disabled.
+	PhaseDisabling
+)
+
+// Done returns a completed phase. e.g.
+// Enabling -> Enabled
+// Disabling -> Disabled
+// otherwise returns the same phase.
+func (s Phase) Done() Phase {
+	switch s {
+	case PhaseEnabling:
+		return PhaseEnabled
+	case PhaseDisabling:
+		return PhaseDisabled
+	case PhaseEnabled, PhaseDisabled, PhaseUnknown:
+		return s
+	default:
+		return PhaseUnknown
+	}
+}
--- a/pkg/operation/botanist/component/phases_test.go
+++ b/pkg/operation/botanist/component/phases_test.go
@@ -1,0 +1,36 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package component_test
+
+import (
+	"github.com/gardener/gardener/pkg/operation/botanist/component"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Phases' Done", func() {
+	DescribeTable("correct phase for", func(old, expected component.Phase) {
+		Expect(old.Done()).To(Equal(expected))
+	},
+		Entry("unknown is not changed", component.PhaseUnknown, component.PhaseUnknown),
+		Entry("not phase a is always unknown", component.Phase(1234), component.PhaseUnknown),
+		Entry("enabled is not changed", component.PhaseEnabled, component.PhaseEnabled),
+		Entry("disabled is not changed", component.PhaseDisabled, component.PhaseDisabled),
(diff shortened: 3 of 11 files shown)


pkg/operation/botanist/controlplane.go
```diff
@@ -1356,6 +1355,43 @@ func (b *Botanist) DefaultKubeAPIServerService() component.DeployWaiter {
 	)
 }
 
+// SNIState returns true if kube-apiserver service exists
```
timuthy:
```suggestion
// SNIPhase returns true if kube-apiserver service exists
```

pkg/operation/botanist/component/types.go
```diff
@@ -0,0 +1,49 @@
...
+	// Component was enabled before and it's still active.
+	PhaseEnabled Phase = "Enabled"
+	// Component was disabled before and it's still disabled.
+	PhaseDisabled Phase = "Disabled"
+	// Component was disabled before, but it's being activated.
+	PhaseEnabling Phase = "Enabling"
+	// Component was enabled before, but it's being disabled.
+	PhaseDisabling Phase = "Disabling"
+	// Component is in unknown phase.
+	PhaseUnknown Phase = "Unknown"
```
rfranzke: Fix doc strings to start with name of the constant

pkg/operation/botanist/component/types.go
```diff
@@ -0,0 +1,49 @@
...
+
+package component
+
+type Phase string
+
+// Done returns a completed phase. e.g.
+// Enabling -> Enabled
+// Disabling -> Disabled
+// otherwise returns the same phase.
+func (s Phase) Done() Phase {
```
rfranzke: Can you add a small unit test?

pkg/operation/botanist/component/types.go
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: Should the file be renamed to `phase.go`?

pkg/operation/botanist/controlplane/kube_apiserver_service.go
```diff
@@ -29,6 +29,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
```
rfranzke: one line above imports the same package, remove this one and just use `client.<...>`

pkg/operation/botanist/controlplane/kube_apiserver_service.go
```diff
@@ -39,23 +40,26 @@ const (
...
 	chartsRootPath string,
 	logger logrus.FieldLogger,
-	client client.Client,
+	client crclient.Client,
 	waiter retry.Ops,
 	clusterIPFunc func(clusterIP string),
 	ingressFunc func(ingressIP string),
 
 ) component.DeployWaiter {
+	loadBalancerServiceKey := crclient.ObjectKey{}
```
rfranzke:
```suggestion
	var loadBalancerServiceKey client.ObjectKey
```

pkg/operation/botanist/controlplane.go
```diff
@@ -1332,20 +1332,19 @@ func (b *Botanist) DeployKubeScheduler(ctx context.Context) error {
...
-	var sniService *client.ObjectKey
-
-	if b.APIServerSNIEnabled() {
-		sniService = &client.ObjectKey{Name: common.IstioIngressGatewayServiceName, Namespace: common.IstioIngressGatewayNamespace}
-	}
+func (b *Botanist) DefaultKubeAPIServerService(sniPhase component.Phase) component.DeployWaiter {
+	return b.kubeAPIServiceService(sniPhase)
+}
 
+func (b *Botanist) kubeAPIServiceService(sniPhase component.Phase) component.DeployWaiter {
```
rfranzke, line 1326: What's the purpose of this function? The only thing `DefaultKubeAPIServerService` is calling is this one, so it can probably be moved into `DefaultKubeAPIServerService` again?

pkg/operation/botanist/controlplane.go
```diff
@@ -1356,6 +1355,43 @@ func (b *Botanist) DefaultKubeAPIServerService() component.DeployWaiter {
...
+		ctx,
+		client.ObjectKey{Name: v1beta1constants.DeploymentNameKubeAPIServer, Namespace: b.Shoot.SeedNamespace},
+		svc,
+	) != nil {
+		if sniEnabled {
+			return component.PhaseEnabled
+		}
+
+		return component.PhaseDisabled
+	}
```
rfranzke: Doesn't it make more sense to handle the error properly here instead of making some assumptions?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -347,6 +350,23 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 			Fn:           flow.TaskFn(botanist.WaitUntilTunnelConnectionExists).SkipIf(o.Shoot.HibernationEnabled),
 			Dependencies: flow.NewTaskIDs(deployManagedResources, waitUntilNetworkIsReady, waitUntilWorkerReady),
 		})
+		_ = g.Add(flow.Task{
+			Name: "Migrating Kubernetes API server to ClusterIP service in the Seed cluster",
```
```suggestion
			Name: "Migrating Kubernetes API server service to ClusterIP service in the Seed cluster if necessary",
```

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -347,6 +350,23 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
...
+			Fn: flow.TaskFn(func(ctx context.Context) error {
+				return botanist.DeployKubeAPIService(ctx, sniPhase.Done())
+			}).
+				RetryUntilTimeout(defaultInterval, defaultTimeout).
+				SkipIf(o.Shoot.HibernationEnabled).
+				DoIf(sniPhase == component.PhaseEnabling || sniPhase == component.PhaseDisabling),
```
rfranzke, line 360: If `sniPhase == component.PhaseDisabling` then we would want to have a load balancer service, right?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -347,6 +350,23 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
...
+			Dependencies: flow.NewTaskIDs(waitUntilTunnelConnectionExists),
+		})
+		_ = g.Add(flow.Task{
+			Name: "Deleting SNI resources",
```
```suggestion
			Name: "Deleting SNI resources if SNI is disabled",
```

rfranzke (changes requested):
Nice PR, looking forward to it if it will fix the problem (still need to play with it), having some upfront suggestions/questions

pkg/operation/shoot/types.go
```diff
@@ -90,9 +90,10 @@ type Components struct {
 
 // ControlPlane contains references to K8S control plane components.
 type ControlPlane struct {
-	KubeAPIServerService component.DeployWaiter
-	KubeAPIServerSNI     component.DeployWaiter
-	KubeScheduler        kubescheduler.KubeScheduler
+	KubeAPIServerService      component.DeployWaiter
+	KubeAPIServerServicePhase component.Phase
```
timuthy:
```suggestion
	KubeAPIServerSNIPhase component.Phase
```
since it's only SNI related

timuthy (dismissed):
I wonder if we can somehow circumvent the TTL issue w/o a separate controller. WDYT if we migrated to SNI (and back) in two phases:

1. Reconciliation: Turn on SNI but leave LB service.
2. Reconciliation: Switch to `clusterIP`.

This is supposed to give the system and all related components more time to refresh the DNS records in their cache.

pkg/operation/botanist/controlplane.go
```diff
@@ -1356,6 +1355,43 @@ func (b *Botanist) DefaultKubeAPIServerService() component.DeployWaiter {
...
+	case svc.Spec.Type == corev1.ServiceTypeClusterIP && sniEnabled:
+		return component.PhaseEnabled
+	case svc.Spec.Type == corev1.ServiceTypeClusterIP && !sniEnabled:
+		return component.PhaseDisabling
+	default:
+		return component.PhaseDisabled
+	}
+}
+
+// DefaultKubeAPIServerService returns a deployer for kube-apiserver service.
```
danielfoehrKn:
```suggestion
// DeployKubeAPIService returns a deployer for kube-apiserver service.
```

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
danielfoehrKn: The comment is only valid if sniPhase=enabling . If sniPhase=disabling then the svc type is changed to LoadBalancer. 
Can we find a more generic description here?

pkg/operation/botanist/controlplane/kube_apiserver_service.go
```diff
@@ -125,16 +149,13 @@ func (d *kubeAPIService) Wait(ctx context.Context) error {
...
 	return d.waiter.Until(ctx, 5*time.Second, func(ctx context.Context) (done bool, err error) {
-		loadBalancerIngress, err := kutil.GetLoadBalancerIngress(ctx, d.client, svcNS, svcName)
+		loadBalancerIngress, err := kutil.GetLoadBalancerIngress(
+			ctx,
+			d.client,
+			d.loadBalancerServicekey.Namespace,
+			d.loadBalancerServicekey.Name,
+		)
 		if err != nil {
 			d.logger.Info("Waiting until the API Server LoadBalancer deployed in the Seed cluster is ready...")
```
danielfoehrKn: It can either be the Shoot API server Load balancer or the Istio Ingress gateway LB.
It would make it clearer if the correct svc is stated.

pkg/operation/botanist/controlplane/kube_apiserver_service.go
```diff
@@ -125,16 +149,13 @@ func (d *kubeAPIService) Wait(ctx context.Context) error {
...
-	svcName := d.service.Name
-
-	if d.values.EnableSNI {
-		svcNS = d.sniService.Namespace
-		svcName = d.sniService.Name
-	}
-
 	return d.waiter.Until(ctx, 5*time.Second, func(ctx context.Context) (done bool, err error) {
-		loadBalancerIngress, err := kutil.GetLoadBalancerIngress(ctx, d.client, svcNS, svcName)
+		loadBalancerIngress, err := kutil.GetLoadBalancerIngress(
```
danielfoehrKn, line 153: Can you add a comment that it is either the Shoot API server service or the istio ingress gateway service - based on the SNI phase (enabling, disabling, ...) ? 
For me that is not obvious right away due to the changes you made.

danielfoehrKn (commented):
Looking good.

I feel like while on the one side the component interfaces add to the testability, they decrease the code readability a lot. Many interfaces and implementations.  
Not sure how you fell about it.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
mvladev: The wording of the task was wrong. It should be 

```
Finishing Kubernetes API server service SNI transition
```


The idea is that kube-apiserver's service should be moved to the next phase.

pkg/operation/botanist/controlplane.go
mvladev: It's called by `DeployKubeAPIService` as well

pkg/operation/botanist/component/phase.go
```diff
@@ -0,0 +1,47 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package component
+
+type Phase int
```
rfranzke, line 18: doc string missing

pkg/operation/botanist/component/phases_test.go
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

rfranzke (commented):
only nit comments

mvladev:
Rebased

rfranzke (approved):
/lgtm

I think it's reasonable to start with this approach and get some experience. Given that the kube-apiserver pod(s) of the "migrating" shoot must be restarted anyways (as they `--advertise-address` flag is changed) and the used DNS TTL is `120`s the migration usually works quite smoothly and without interrupts  (at least in my local tests). I propose to get some real-world experience with it for now. If we see issues I'd probably agree to @timuthy and split the migration over two distinct reconciliation flows (https://github.com/gardener/gardener/pull/2847#pullrequestreview-488355548).

PR #2850, opened 2020-09-14
konnectivity server now logs to stdout

vpnachev:

/area control-plane networking ops-productivity
/kind enhancement
/priority normal

Konnectivity ~agent and~ server logs to stdout. Currently these components are dumping their logs in files in the pod ephemeral filesystem, but the container images comes without shell and basic tools which means that the logs cannot be consumed. Switching to stdout allows plain `kubectl logs <pod-name>` to be used, also the logs from the seed component will be preserved in the Loki instance (cc @Kristian-ZH @vlvasilev new log parser might be needed for this container).

Fixes #

**Release note**:

```improvement operator
The `konnectivity-server` is now logging to the stdout.
```

### DIFF
--- a/charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
+++ b/charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
@@ -281,9 +281,7 @@ spec:
           - /proxy-server
         args:
         - --uds-name=/etc/srv/kubernetes/konnectivity-server/konnectivity-server.socket
-        - --log-file=/var/log/konnectivity-server/info.log
-        - --logtostderr=false
-        - --log-file-max-size=12
+        - --logtostderr=true
         - --cluster-cert=/certs/konnectivity-server/konnectivity-server.crt
         - --cluster-key=/certs/konnectivity-server/konnectivity-server.key
         - --agent-namespace={{ .Values.konnectivityTunnel.agentNamespace }}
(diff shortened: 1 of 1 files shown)


vpnachev:
/hold

It turns out that the proxy agent is logging with rate of 700 log messages per second.
There is a upstream improvement PR https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/112 (cc @mvladev ).

vpnachev:
The proxy-server is looking much better - with a single node shoot cluster the log message rate is 3 messages per second.
Maybe I will split the PR change only the server settings.

vpnachev:
/unassign
/needs/review

charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
```diff
@@ -281,9 +281,7 @@ spec:
           - /proxy-server
         args:
         - --uds-name=/etc/srv/kubernetes/konnectivity-server/konnectivity-server.socket
-        - --log-file=/var/log/konnectivity-server/info.log
```
rfranzke, line 284: The `/var/log/konnectivity-server` mount can be removed now right?

vpnachev: Nice catch, thanks!

rfranzke (approved):
/lgtm

PR #2858, opened 2020-09-16
Refactor cluster-autoscaler deployment

rfranzke:

/kind enhancement
/priority normal
/topology seed
/platform all
/exp intermediate

This PR refactors the deployment for the cluster-autoscaler and adopts the DeployWaiter interface for components. Consequently, it eliminates the Helm chart and adds full unit tests, instead. This improves several aspects and the overall code quality.

Part of #2754

**Release note**:

```improvement operator
NONE
```


pkg/operation/botanist/controlplane/clusterautoscaler/cluster_autoscaler.go
```diff
@@ -0,0 +1,497 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	// ServiceName is the name of the service of the cluster-autoscaler.
+	ServiceName = "cluster-autoscaler"
+	// SecretName is a constant for the secret name for the cluster-autoscaler's kubeconfig secret.
+	SecretName = "cluster-autoscaler"
+	// UserName is the name that should be used for the secret that the cluster-autoscaler uses to
+	// authenticate itself with the kube-apiserver (e.g., the common name in its client certificate).
```
danielfoehrKn, line 55:
```suggestion
	// authenticate itself with the kube-apiserver (e.g. the common name in its client certificate).
```

pkg/operation/botanist/controlplane/clusterautoscaler/cluster_autoscaler.go
```diff
@@ -0,0 +1,497 @@
...
+}
+
+// Secrets is collection of secrets for the cluster-autoscaler.
+type Secrets struct {
+	// Kubeconfig is a secret which can be used by the cluster-autoscaler to communicate to the kube-apiserver.
+	Kubeconfig component.Secret
+}
+
+// BootstrapSeed deploys the RBAC configuration for the control cluster.
+func BootstrapSeed(ctx context.Context, c client.Client, namespace, _ string) error {
```
danielfoehrKn: Maybe it would make sense to put this function into its own file e.g `cluster_autoscaler_seed_bootstrap.go`?.
I just did not realise at first that this is only used during the Seed bootstrap, while the other contents for the file are used during the Shoot reconciliation.

danielfoehrKn (dismissed):
Very nice PR - it is also well readable by splitting it into logging/ monitoring etc.
Tried it out seems to be working fine at the first glance. 

Did not see many issue.

pkg/operation/botanist/controlplane/clusterautoscaler/cluster_autoscaler.go
rfranzke: I'm usually following American style, i.e., I put commas after "i.e." and "e.g." :)

pkg/operation/botanist/controlplane/clusterautoscaler/cluster_autoscaler.go
rfranzke: Yep, makes sense, will do! Thanks.

pkg/operation/botanist/controlplane/clusterautoscaler/monitoring.go
```diff
@@ -0,0 +1,101 @@
...
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+)
+
+const (
+	monitoringPrometheusJobName = "cluster-autoscaler"
+
+	monitoringMetricProcessMaxFds  = "process_max_fds"
+	monitoringMetricProcessOpenFds = "process_open_fds"
```
danielfoehrKn, line 31: I just spotted that these constants are duplicated in the kube-scheduler package ( and I am about to use them aswell for the kcm).
I think it would make sense to put them into a common package.

danielfoehrKn (commented):
Thanks for addressing the comments.

pkg/operation/botanist/controlplane/clusterautoscaler/monitoring.go
rfranzke: @wyb1 does this make sense? I would guess these metrics are component-specific and they are just "by accident" the same?

pkg/utils/kubernetes/object.go
```diff
@@ -34,3 +37,21 @@ func ObjectName(obj runtime.Object) string {
...
+		if err := DeleteObject(ctx, c, obj); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// DeleteObject deletes a Kubernetes object. It ignores 'not found' and 'no match' errors.
+func DeleteObject(ctx context.Context, c client.Client, object runtime.Object) error {
+	if err := c.Delete(ctx, object); client.IgnoreNotFound(err) != nil && !meta.IsNoMatchError(err) {
```
ialidzhikov, line 53: I think it makes sense to have this behaviour configurable. We could modify this func to also accept errors to tolerate (`[]TolerateErrorFunc`). Not every caller would want to ignore exactly these 2 errors. WDYT?

ialidzhikov (commented):
Looks good!

pkg/utils/kubernetes/object.go
rfranzke: Well, I would consider this somewhat "over-engineered" for this simple utility function. I would propose that in such a situation the user can simply call `sigs.k8s.io/controller-runtime/client.Delete()` function directly and handle the errors he wants to handle specially directly in this code?

ialidzhikov (dismissed):
/lgtm

rfranzke:
Rebased to `master` - @wyb1 do you have any feedback?

danielfoehrKn (approved):
/lgtm

pkg/operation/botanist/controlplane/clusterautoscaler/monitoring.go
wyb1: These metrics are exposed by most exporters so we can add them to a common package as long as we can decide to add them for the component (or not).

rfranzke: Hm, ok, thanks, I see. I would still prefer to keep them here for now as I wouldn't know where to put these (which "common" package) - for now it should be more clear as it is.

PR #2860, opened 2020-09-16
Fix a bug where admission plugins of one shoot can be applied to all shoots on the same seed

vpnachev:

/area control-plane
/kind bug
/priority critical

Fix a bug where custom admission plugins of one shoot can be applied to all shoots on the same seed.

The issue is that every call of `GetAdmissionPluginsForVersion` returns always the same slice. If a shoot is customizing a default admission plugin it will overwrite the shared slice here https://github.com/gardener/gardener/blob/95e8072608731a52f74fd46027f37c389aba4d1c/pkg/operation/botanist/controlplane.go#L1104 and with the next reconciliation other shoots reconciled by the same gardenlet are getting the same custom config even though they have not customized it.

Fixes #

Thanks you @BeckerMax @ialidzhikov @schrodit @dansible (I hope do not miss anybody) for the help debugging this issue.

**Release note**:

```improvement operator
An issue where the custom admission plugin settings might of one shoot cluster can be applied to other shoots reconciled by the same seed cluster has been fixed.
```


ialidzhikov (approved):
/lgtm

PR #2862, opened 2020-09-17
Defrag shooted seed etcd-main every day.

amshuman-kr:

/area control-plane
/kind post-mortem
/priority critical

Defrag shooted seed etcd-main every day. This is to temporarily mitigate the high load on seed apiserver from recent changes.

A temporary mitigation for -> https://github.tools.sap/kubernetes-canary/issues-canary/issues/155#issuecomment-119374

This is only a temporary mitigation until the issue with high load into the seed apiserver is addressed.

Less than `24h` schedule is not going to help unless we also chaning `auto-compaction-retention` which is currently set to [`24h`](https://github.com/gardener/etcd-druid/blob/master/charts/etcd/templates/etcd-bootstrap-configmap.yaml#L130) which is not exposed as a field in the `Etcd` CRD of druid. @shreyas-s-rao Does it make sense to already introduce the `autoCompactionRetention` as a field in the the `Etcd` CRD and create a patch release (without including the volume deletion changes)? I think we need to be prepared for any issues on azure seeds. Most of them have relatively high DB sizes.

/cc @shreyas-s-rao  @rfranzke @vlerenc 

**Release note**:

```improvement operator
Defrag shooted seed etcd-main every day. This is to temporarily mitigate the high load on seed apiserver from recent changes.
```

### DIFF
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -1510,7 +1510,7 @@ func (b *Botanist) DeployETCD(ctx context.Context) error {
 			foundStatefulset = false
 		}
 
-		defragmentSchedule, err := DetermineDefragmentSchedule(b.Shoot.Info, etcd)
+		defragmentSchedule, err := DetermineDefragmentSchedule(b.Shoot.Info, etcd, b.ShootedSeed, role)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


shreyas-s-rao:
Thanks for the PR @amshuman-kr . Regarding auto compaction, right now we make use of etcd's `auto-compaction` policy to do this, rather than a scheduled action like defragmentation which is done by backup sidecar. So even if we make the `auto-compaction` time configurable in the charts, we still won't be able to make sure compaction and defragmentation are synchronized, unless we add this logic within sidecar. Would that make sense?

Nonetheless, for the moment, we can expose the `auto-compaction-retention` field in `Etcd` spec atleast to enable defragmentation to be less than `24h`.

shreyas-s-rao (approved):
/lgtm

rfranzke (approved):
/lgtm

amshuman-kr:
> right now we make use of etcd's auto-compaction policy to do this, rather than a scheduled action like defragmentation which is done by backup sidecar. 

Yes. I was not suggesting to change that. We configure etcd to `auto-compact` with `24h` retention. I was merely suggesting that we allow the retention period to be configured via the `Etcd` CRD so that we can configure it to be less than `24h` if needed.

> So even if we make the auto-compaction time configurable in the charts, we still won't be able to make sure compaction and defragmentation are synchronized, unless we add this logic within sidecar. Would that make sense?

No. I think adding additional compaction functionality to the sidecar would be too much effort for what can already be done well enough by configuring etcd flexibly. I.e. we can take care to configure the compaction retention period to be sufficiently less than the defragmentation period to be sure that defragmentation will reclaim good amount of space.

amshuman-kr:
@rfranzke This change will be rolled out to the shooted seed etcds during their maintenance window, right? Just to be sure it won't roll the etcd-main of all shooted seeds at the same time.

rfranzke:
This depends on the gardenlet configuration, but if the `reconcileInMaintenanceTimeWindowOnly` flag is set to `true` then the answer would be yes.

Issue #2677, opened 2020-08-07
Logging stabilization


Kristian-ZH:
This issue describes the necessary tasks for the [Logging stack](https://github.com/gardener/logging/pull/58) stabilization.
The following things have to be done:
  - [ ] Observe Logging stack behaviour across landscapes.
  - [ ] Merge logging documentations [this](https://github.com/gardener/logging/pull/58) and [this](https://github.com/gardener/gardener/pull/2601). 
  - [ ] Merge [gardenctl](https://github.com/gardener/gardenctl/pull/213) related changes.
  - [ ] Add `fluent-bit` filters and parsers for all of the extensions.
  - [ ] Make fluent-bit aware of `testing` and `hibernated` clusters (Currently if there is a log messages just before the hibernation or messages from `tf-pods` the `fluent-bit` tries to flush them into non existing Loki, because it is already deleted ).

rfranzke:
/area logging

Just saw the following issue:

```
$ kg logs fluent-bit-trpr9
[proxy] error opening plugin /fluent-bit/plugins/out_loki.so: "/fluent-bit/plugins/out_loki.so: cannot open shared object file: No such file or directory"
```

Kristian-ZH:
> /area logging
> 
> Just saw the following issue:
> 
> ```
> $ kg logs fluent-bit-trpr9
> [proxy] error opening plugin /fluent-bit/plugins/out_loki.so: "/fluent-bit/plugins/out_loki.so: cannot open shared object file: No such file or directory"
> ```

@vlvasilev

vpnachev:
Remove the admission that is annotating pods with `fluentbit.io/exclude` as we now have a controller inside the `fluent-bit` dropping the unneeded logs.

rfranzke:
Given #2865, let's add the following item as well:
- [ ] Rework logging dashboards to make it easy to expose additional control plane component logs (e.g., cluster-autoscaler)

PR #2865, opened 2020-09-17
[WIP]expose the cluster-autoscaler logs to the user

vlvasilev:

/area logging
/kind enhancement
/priority normal

With this PR we expose the Cluster-autoscaler logs to the user.

Fixes #

**Release note**:

```improvement operator
Expose cluster-autoscaler logs to the user
```

### DIFF
--- a/charts/seed-monitoring/charts/grafana/dashboards/owners/shoot-vpa-dashboard.json
+++ b/charts/seed-monitoring/charts/grafana/dashboards/owners/shoot-vpa-dashboard.json
@@ -267,6 +267,32 @@
         "align": false,
         "alignLevel": null
       }
+    },
+    {
+      "datasource": "loki",
+      "gridPos": {
+        "h": 11,
+        "w": 24,
+        "x": 0,
+        "y": 13
+      },
+      "id": 8,
+      "options": {
+        "showLabels": false,
+        "showTime": true,
+        "sortOrder": "Descending",
+        "wrapLogMessage": false
+      },
+      "targets": [
+        {
+          "expr": "{container_name=\"cluster-autoscaler\"} |~ \"$search\"",
+          "refId": "A"
+        }
+      ],
+      "timeFrom": null,
+      "timeShift": null,
+      "title": "cluster-autoscaler logs",
+      "type": "logs"
     }
   ],
   "refresh": false,


charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-configmap.yaml
```diff
@@ -171,6 +171,11 @@ data:
         Parser              kubeapiserverParser
         Reserve_Data        True
 
+    [FILTER]
```
rfranzke: Hm, we don't use this construct (`Record type user`) for other component logs that we expose to end-users (controller-manager, scheduler) - how does it work for them?

vlvasilev: Hm. you are right, as I see now we use only predefined dashboards.

vlvasilev:
Now I have add new dashboard but I didn't test it, because my local setup is broken and I have to leave for the moment.
When I test it I will remove [WIP] from the title.
Sorry for the inconvenience.

vpnachev:
It looks very weird to have the logs of the cluster autoscaler in the VPA dashboard.
It would be better to have dedicated dashboard for the cluster autoscaler with some metrics and show the logs there (cc @wyb1 @istvanballok @prashanth26 @hardikdr ).

Another option is to simply move all logging related panes to the `Kubernetes Control Plane Status` dashboard - wdyt?

Kristian-ZH:
I Do not think that adding this Panel to the `Kubernetes Dashboard Status` is a good idea because this Dashboard is really huge and uncomfortable for using.

We also can not make separate Dashboard Like `Pod Logs` which includes all Pods available for the users because when there is a select field, the selected option is applied to the URL and can be modified after that

So the best option for me is to create a separate Dashboard for each of the components that includes a basic metrics for it and its logs. Also we should add a search field, which the users will be able to use for searching a certain message in their logs

rfranzke:
What's the plan with this PR? Does it make sense to keep it open, or will you rather rework the dashboards etc. in a larger PR?

Kristian-ZH:
I think that reworking the dashboards in a larger PR is a better option

rfranzke:
Makes sense. Then let's
/close
this one in favor of https://github.com/gardener/gardener/issues/2677#issuecomment-696577387

PR #2866, opened 2020-09-18
Update istio to 1.7.1

mvladev:

/area networking
/area control-plane
/kind enhancement
/priority normal

- Istio Ingress Gateway listens on port 9443 for TLS traffic and no longer requires any privileges and capabilities
- Istiod now listens on 10250 port for admission configuration and no longer requires any privileges and capabilities

n/a

[Istio now requires K8S 1.16+](https://istio.io/latest/news/releases/1.7.x/announcing-1.7/upgrade-notes/#require-kubernetes-1-16)

**Release note**:

```action operator
Kubernetes 1.16+ is required for Seed clusters which have `ManagedIstio` feature gate.
```

```improvement operator
Istio is updated to 1.7.1
```


rfranzke (dismissed):
/lgtm

rfranzke:
Can you check the failing test?

mvladev:
@rfranzke I forgot to push and was wondering why the tests were failing after succeeding locally.

rfranzke:
OK thanks.
Regarding  

> Kubernetes 1.16+ is required for Seed clusters which have `ManagedIstio` feature gate.

Is it really required, i.e., can't we stick to the `v1beta1` CRD API for now? If not, I'm okay with the change as the feature gate is still alpha, but it would be nicer if we don't introduce this restriction now. WDYT?

mvladev:
The istio's [support matrix](https://docs.google.com/document/d/14Xfe9V0vv8d7W75c21d-cY2XyK6xvRH0_Kg8j_zfh4Y) offers lots of info on why this is needed. Using `v1beta1` CRDs would work, but we have to maintain that by our self. Also It would limit us when reporting issues as it's very likely that the Istio team would simply say "unsupported K8S version".

rfranzke (approved):
Alright, let's go ahead then
/lgtm

PR #2876, opened 2020-09-22
Vendor github.com/gardener/controller-runtime@v0.6.3-gardener.1

### LINKED ISSUE #2675: Dynamic RESTMapper limited

/area control-plane
/area scalability
/kind bug
/priority normal

**What happened**:
If the `APIServerSNI` feature gate is disabled and Istio is not installed on the Seed, it can happen, that under high load, the Shoot reconciliation will fail, because the dynamic REST mapper of the Seed client is rate limited:
```
could not delete object of kind \“EnvoyFilter\” \“istio-ingress/shoot--foo--bar\“: too many API calls to the RESTMapper within a timeframe, could not delete object of kind \“Gateway\” \“shoot--foo--bar/kube-apiserver\“: too many API calls to the RESTMapper within a timeframe, could not delete object of kind \“VirtualService\” \“shoot--foo--bar/kube-apiserver\“: too many API calls to the RESTMapper within a timeframe
```

The problem is, that the dynamic RESTMapper returns a different error than the expected `NoKindMatchError` in that case (ref https://github.com/kubernetes-sigs/controller-runtime/issues/1120) and therefore, the `TolerateErrorFunc` can't catch it:
https://github.com/gardener/gardener/blob/514f00cbd32cc3022032376f48ebaf0f3ad0ed3e/pkg/operation/botanist/controlplane/kube_apiserver_sni.go#L76-L85

The same problem can hit us in the seed bootstrap here:
https://github.com/gardener/gardener/blob/514f00cbd32cc3022032376f48ebaf0f3ad0ed3e/pkg/operation/seed/istio/proxy_protocol.go#L50-L58
e.g:
```
seed is not yet ready: condition \"Bootstrapped\" has invalid status False (expected True) due to BootstrappingFailed: failed to delete manifests: 1 error occurred: could not delete object of kind \"VirtualService\" \"istio-ingress/proxy-protocol-blackhole\": too many API calls to the RESTMapper within a timeframe"
```

The Shoot reconciliation/Seed bootstrapping should not block in that case and should just ignore the error, because it means, that the Istio stuff is anyways not installed.

Simultaneously create hundreds of Shoots? 😁 🤔 

**Environment**:

- Gardener version: v1.8.0
- Others:

### LINKED ISSUE #2747: Gardenlet times out talking to the Seed API server

rfranzke:

/area scalability robustness
/kind bug
/priority normal

With this PR we vendor a forked `sigs.k8s.io/controller-runtime` repository which includes the not-yet-merged improvement from @timebertt (https://github.com/kubernetes-sigs/controller-runtime/pull/1151).

We also pass a custom rate limiter that allows one request every 5s as suggested in https://github.com/gardener/gardener/issues/2747#issuecomment-688064801.

Fixes #2747
Fixes #2675

/invite @timebertt 
/assign @timebertt 

**Release note**:

```improvement operator
NONE
```


pkg/client/kubernetes/runtime_client.go
```diff
@@ -93,7 +94,11 @@ func newRuntimeClientWithCache(config *rest.Config, options client.Options, cach
 func setClientOptionsDefaults(config *rest.Config, options *client.Options) error {
 	if options.Mapper == nil {
 		// default the client's REST mapper to a dynamic REST mapper (automatically rediscovers resources on NoMatchErrors)
-		mapper, err := apiutil.NewDynamicRESTMapper(config, apiutil.WithLazyDiscovery)
+		mapper, err := apiutil.NewDynamicRESTMapper(
+			config,
+			apiutil.WithLazyDiscovery,
+			apiutil.WithLimiter(rate.NewLimiter(0.2, 1)), // one request every 5s
```
timebertt:
```suggestion
			apiutil.WithLimiter(rate.NewLimiter(rate.Every(5*time.Second), 1)),
```

timebertt (commented):
Thanks for taking care!
This will also fix #2675.

Only one nit suggestion, otherwise /lgtm.

timebertt (approved):
Thanks!
/lgtm

PR #2881, opened 2020-09-23
Refactoring kube controller manager deployment for Shoot clusters.

danielfoehrKn:

/area control-plane
/kind technical-debt
/priority normal

Refactoring kube controller manager to use the component interface and use the go client instead of the helm chart.

Part of #2754 

Similar to the refactoring of the `kube-scheduler`,  this introduces a new loki parser  `kubeControllerManagerParser` that is used instead of previously defined `kubeapiserverParser`. 
The configuration is the same though. @Kristian-ZH 

**Release note**:

```improvement operator

```


pkg/operation/botanist/controlplane/kubecontrollermanager/logging.go
```diff
@@ -0,0 +1,42 @@
...
+	loggingParser     = `[PARSER]
+    Name        ` + loggingParserName + `
+    Format      regex
+    Regex       ^(?<severity>\w)(?<time>\d{4} [^\s]*)\s+(?<pid>\d+)\s+(?<source>[^ \]]+)\] (?<log>.*)$
+    Time_Key    time
+    Time_Format %m%d %H:%M:%S.%L
+`
+	loggingFilter = `[FILTER]
+    Name                parser
+    Match               kubernetes.*` + v1beta1constants.DeploymentNameKubeControllerManager + `*` + v1beta1constants.DeploymentNameKubeControllerManager + `*
```
rfranzke: Please define a constant for the `containerName`

```suggestion
    Match               kubernetes.*` + v1beta1constants.DeploymentNameKubeControllerManager + `*` + containerName + `*
```

pkg/operation/botanist/controlplane/kubecontrollermanager/monitoring.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+var (
+	monitoringAllowedMetrics = []string{
+		monitoringMetricRestClientRequestsTotal,
+		monitoringMetricProcessMaxFds,
+		monitoringMetricProcessOpenFds,
+	}
+
+	monitoringScrapeConfigTmpl = `job_name: ` + monitoringPrometheusJobName + `
```
rfranzke, line 57: Can you add a `// TODO` comment to remove the hard-coded prometheus paths once the prometheus component is refactored?

pkg/operation/botanist/controlplane/kubecontrollermanager/monitoring_alertingrules_kube_controller_manager.test.yaml
```diff
@@ -1,5 +1,5 @@
 rule_files:
```
rfranzke: rename the file to `monitoring_alertingrules.test.yaml`

pkg/operation/botanist/controlplane/kubecontrollermanager/monitoring_test.go
```diff
@@ -0,0 +1,112 @@
...
+    for: 15m
+    labels:
+      service: kube-controller-manager
+      severity: critical
+      type: seed
+      visibility: all
+    annotations:
+      description: Deployments and replication controllers are not making progress.
+      summary: Kube Controller Manager is down.`
+)
```
rfranzke, line 109: Can you move that down (below the test)?

pkg/operation/botanist/controlplane/kubecontrollermanager/monitoring_test.go
```diff
@@ -0,0 +1,112 @@
...
+var _ = Describe("Monitoring", func() {
+	DescribeTable("success tests for scrape config various kubernetes versions",
+		func(version, expectedScrapeConfig string) {
+			semverVersion, err := semver.NewVersion(version)
+			Expect(err).NotTo(HaveOccurred())
+			kubeControllerManager := New(nil, "", semverVersion, "", nil, nil, nil)
+
+			scrapeConfigs, err := kubeControllerManager.ScrapeConfigs()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(scrapeConfigs).To(ConsistOf(Equal(expectedScrapeConfig)))
```
rfranzke:
```suggestion
			test.ScapeConfigs(kubeControllerManager, expectedScrapeConfig)
```

pkg/operation/botanist/controlplane.go
```diff
@@ -1229,67 +1231,60 @@ func IsValidAuditPolicyVersion(shootVersion string, schemaVersion *schema.GroupV
...
-			defaultValues["nodeCIDRMaskSize"] = *controllerManagerConfig.NodeCIDRMaskSize
-		}
-	}
-
-	values, err := b.InjectSeedShootImages(defaultValues, common.KubeControllerManagerImageName)
-	if err != nil {
-		return err
-	}
+	b.Shoot.Components.ControlPlane.KubeControllerManager.SetSecrets(kubecontrollermanager.Secrets{
+		Ca:                kubecontrollermanager.Secret{Name: v1beta1constants.SecretNameCACluster, Checksum: b.CheckSums[constants.SecretNameCACluster]},
```
```suggestion
		Ca:                kubecontrollermanager.Secret{Name: v1beta1constants.SecretNameCACluster, Checksum: b.CheckSums[v1beta1constants.SecretNameCACluster]},
```

pkg/operation/botanist/controlplane.go
```diff
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/gardener/gardener/pkg/apis/core/v1alpha1/constants"
```
```suggestion
```

rfranzke (changes requested):
First quick scan, will continue asap

danielfoehrKn:
Will address the comments when you completed your first review cycle. Thanks!

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+	volumeMountPathCA = "/srv/kubernetes/ca"
+	// volumeMountPathCA is the volume mount path for the service account key that is a PEM-encoded private RSA or ECDSA key used to sign service account tokens.
+	volumeMountPathServiceAccountKey = "/srv/kubernetes/service-account-key"
+	// volumeMountPathCA is the volume mount path for the kubeconfig which can be used by the kube-controller-manager to communicate with the kube-apiserver.
+	volumeMountPathKubeconfig = "/var/lib/kube-controller-manager"
+	// volumeMountPathCA is the volume mount path for the x509 TLS server certificate and key for the HTTPS server inside the kube-controller-manager (which is used for metrics and health checks).
+	volumeMountPathServer = "/var/lib/kube-controller-manager-server"
+
+	// ShootManagedResourceName is the name of the managed resource that contains the resources to be deployed into the Shoot cluster.
+	ShootManagedResourceName = "shoot-core-kube-controller-manager"
```
rfranzke: Why is this exported?

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+type Secrets struct {
+	// Kubeconfig is a secret that contains a kubeconfig which can be used by the kube-controller-manager to communicate with the kube-apiserver.
+	Kubeconfig Secret
+	// Server is a secret containing a x509 TLS server certificate and key for the HTTPS server inside the kube-controller-manager (which is used for metrics and health checks).
+	Server Secret
+	// Ca is a secret containing a root CA x509 certificate and key that is used for the flags.
+	// --cluster-signing-cert-file
+	// --cluster-signing-key-file
+	// --root-ca-file
+	Ca Secret
```
rfranzke:
```suggestion
	CA Secret
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+	utilruntime.Must(err)
+}
+
+// Secrets is collection of secrets for the kube-controller-manager.
+type Secrets struct {
+	// Kubeconfig is a secret that contains a kubeconfig which can be used by the kube-controller-manager to communicate with the kube-apiserver.
+	Kubeconfig Secret
+	// Server is a secret containing a x509 TLS server certificate and key for the HTTPS server inside the kube-controller-manager (which is used for metrics and health checks).
+	Server Secret
+	// Ca is a secret containing a root CA x509 certificate and key that is used for the flags.
```
rfranzke:
```suggestion
	// CA is a secret containing a root CA x509 certificate and key that is used for the flags.
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+	// --cluster-signing-cert-file
+	// --cluster-signing-key-file
+	// --root-ca-file
+	Ca Secret
+	// ServiceAccountKey is a secret containing a PEM-encoded private RSA or ECDSA key used to sign service account tokens.
+	// used for the flag: --service-account-private-key-file
+	ServiceAccountKey Secret
+}
+
```
rfranzke:
```suggestion
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+	versionConstraintK8sGreaterEqual116, err = semver.NewConstraint(">= 1.16")
+	utilruntime.Must(err)
+	versionConstraintK8sGreaterEqual117, err = semver.NewConstraint(">= 1.17")
+	utilruntime.Must(err)
+}
+
+// Secrets is collection of secrets for the kube-controller-manager.
+type Secrets struct {
+	// Kubeconfig is a secret that contains a kubeconfig which can be used by the kube-controller-manager to communicate with the kube-apiserver.
+	Kubeconfig Secret
```
rfranzke:
```suggestion
	Kubeconfig component.Secret
```

(same for other attributes)

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+		}
+	}
+	return horizontalPodAutoscalerConfig
+}
+
+var (
+	versionConstraintK8sEqual113        *semver.Constraints
+	versionConstraintK8sGreaterEqual112 *semver.Constraints
+	versionConstraintK8sSmaller112      *semver.Constraints
+	versionConstraintK8sGreaterEqual113 *semver.Constraints
```
rfranzke, line 496:
```suggestion
	versionConstraintK8sSmaller112      *semver.Constraints
	versionConstraintK8sGreaterEqual112 *semver.Constraints
	versionConstraintK8sEqual113        *semver.Constraints
	versionConstraintK8sGreaterEqual113 *semver.Constraints
```

nit :)

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+			Kind:       "Deployment",
+			Name:       v1beta1constants.DeploymentNameKubeControllerManager,
+		}
+		vpa.Spec.UpdatePolicy = &autoscalingv1beta2.PodUpdatePolicy{
+			UpdateMode: &vpaUpdateMode,
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
```
rfranzke: Please move the VPA creation after the creation of the `Deployment`

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+
+	if _, err := controllerutil.CreateOrUpdate(ctx, k.client, service, func() error {
+		service.Labels = labels
+		service.Spec.Selector = labels
+		service.Spec.Type = corev1.ServiceTypeClusterIP
```
rfranzke, line 156:
```suggestion
		service.Spec.Type = corev1.ServiceTypeClusterIP
		service.Spec.ClusterIP = corev1.ClusterIPNone
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+					v1beta1constants.LabelPodMaintenanceRestart:         "true",
+					v1beta1constants.LabelNetworkPolicyToDNS:            v1beta1constants.LabelNetworkPolicyAllowed,
+					v1beta1constants.LabelNetworkPolicyToShootAPIServer: v1beta1constants.LabelNetworkPolicyAllowed,
+					v1beta1constants.LabelNetworkPolicyFromPrometheus:   v1beta1constants.LabelNetworkPolicyAllowed,
+				}),
+			},
+			Spec: corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:            v1beta1constants.DeploymentNameKubeControllerManager,
```
rfranzke: please define a constant for the containerName, see https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go#L61

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+func (k *kubeControllerManager) emptyService() *corev1.Service {
+	return &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: ServiceName, Namespace: k.namespace}}
+}
+
+func (k *kubeControllerManager) emptyDeployment() *appsv1.Deployment {
+	return &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: v1beta1constants.DeploymentNameKubeControllerManager, Namespace: k.namespace}}
+}
+
+func (k *kubeControllerManager) emptyManagedResource() *resourcesv1alpha1.ManagedResource {
+	return &resourcesv1alpha1.ManagedResource{ObjectMeta: metav1.ObjectMeta{Name: ShootManagedResourceName, Namespace: k.namespace}}
```
rfranzke:
```suggestion
	return &resourcesv1alpha1.ManagedResource{ObjectMeta: metav1.ObjectMeta{Name: ShootManagedResourceName, Namespace: k.namespace}}
```
```suggestion
	return &resourcesv1alpha1.ManagedResource{ObjectMeta: metav1.ObjectMeta{Name: managedResourceName, Namespace: k.namespace}}
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+}
+
+func (k *kubeControllerManager) emptyManagedResourceSecret() *corev1.Secret {
+	return &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: common.ManagedResourceSecretName(ShootManagedResourceName), Namespace: k.namespace}}
```
rfranzke:
```suggestion
	return &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: common.ManagedResourceSecretName(managedResourceName), Namespace: k.namespace}}
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+	if versionConstraintK8sGreaterEqual113.Check(k.version) {
+		return corev1.URISchemeHTTPS
+	}
+	return corev1.URISchemeHTTP
+}
+
+func (k *kubeControllerManager) computeCommand(port int32) []string {
+	var (
+		command                              []string
+		defaultHorizontalPodAutoscalerConfig = k.getHorizontalPodAutoscalerConfig()
```
rfranzke, line 361: What about 
```suggestion
		hpaConfig = k.getHorizontalPodAutoscalerConfig()
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+		DownscaleStabilization:  &metav1.Duration{Duration: gardencorev1beta1.DefaultDownscaleStabilization},
+		InitialReadinessDelay:   &metav1.Duration{Duration: gardencorev1beta1.DefaultInitialReadinessDelay},
+		SyncPeriod:              &metav1.Duration{Duration: gardencorev1beta1.DefaultHPASyncPeriod},
+		Tolerance:               &defaultHPATolerance,
+		UpscaleDelay:            &metav1.Duration{Duration: gardencorev1beta1.DefaultHPAUpscaleDelay},
+	}
+
+	if k.config.HorizontalPodAutoscalerConfig != nil {
+		if k.config.HorizontalPodAutoscalerConfig.CPUInitializationPeriod != nil {
+			horizontalPodAutoscalerConfig.CPUInitializationPeriod = k.config.HorizontalPodAutoscalerConfig.CPUInitializationPeriod
```
rfranzke:
```suggestion
		if v := k.config.HorizontalPodAutoscalerConfig.CPUInitializationPeriod; v != nil {
			horizontalPodAutoscalerConfig.CPUInitializationPeriod = v
```

This pattern makes it a bit shorter

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+
+	if k.config.NodeCIDRMaskSize != nil {
+		command = append(command, fmt.Sprintf("--node-cidr-mask-size=%d", *k.config.NodeCIDRMaskSize))
+	}
+
+	command = append(command,
+		fmt.Sprintf("--cluster-cidr=%s", k.podNetwork.String()),
+		fmt.Sprintf("--cluster-name=%s", k.namespace),
+		"--cluster-signing-cert-file=/srv/kubernetes/ca/ca.crt",
+		"--cluster-signing-key-file=/srv/kubernetes/ca/ca.key",
```
rfranzke:
```suggestion
		fmt.Sprintf("--cluster-signing-cert-file=%s/%s", volumeMountPathCA, secrets.DataKeyCACertificate"),
		fmt.Sprintf("--cluster-signing-key-file=%s/%s", volumeMountPathCA, secrets.DataKeyCAPrivateKey),
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+		command = append(command,
+			fmt.Sprintf("--horizontal-pod-autoscaler-downscale-delay=%s", defaultHorizontalPodAutoscalerConfig.DownscaleDelay.Duration.String()),
+			fmt.Sprintf("--horizontal-pod-autoscaler-upscale-delay=%s", defaultHorizontalPodAutoscalerConfig.UpscaleDelay.Duration.String()),
+		)
+	}
+
+	command = append(command,
+		fmt.Sprintf("--horizontal-pod-autoscaler-sync-period=%s", defaultHorizontalPodAutoscalerConfig.SyncPeriod.Duration.String()),
+		fmt.Sprintf("--horizontal-pod-autoscaler-tolerance=%v", *defaultHorizontalPodAutoscalerConfig.Tolerance),
+		"--kubeconfig=/var/lib/kube-controller-manager/kubeconfig",
```
rfranzke:
```suggestion
		fmt.Sprintf("--kubeconfig=%s/%s", volumeMountPathKubeconfig, secrets.DataKeyKubeconfig),
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+		"--leader-elect=true",
+		"--node-monitor-grace-period=120s",
+		"--pod-eviction-timeout=2m0s",
+		"--root-ca-file=/srv/kubernetes/ca/ca.crt",
```
rfranzke:
```suggestion
		fmt.Sprintf("--root-ca-file=%s/%s", volumeMountPathCA, secrets.DataKeyCACertificate),
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,529 @@
...
+		"--service-account-private-key-file=/srv/kubernetes/service-account-key/id_rsa",
```
rfranzke:
```suggestion
		fmt.Sprintf("--service-account-private-key-file=%s/%s", volumeMountPathServiceAccountKey, secrets.DataKeyRSAPrivateKey),
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
```diff
@@ -0,0 +1,802 @@
+package kubecontrollermanager_test
```
rfranzke, line 15: license header missing

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
```diff
@@ -0,0 +1,802 @@
...
+
+	. "github.com/onsi/ginkgo/extensions/table"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	autoscalingv1beta2 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
+	"k8s.io/utils/pointer"
+
+	mockclient "github.com/gardener/gardener/pkg/mock/controller-runtime/client"
```
rfranzke: imports sorting if possible :)

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
```diff
@@ -0,0 +1,802 @@
...
+						c.EXPECT().Update(ctx, gomock.AssignableToTypeOf(&corev1.Secret{})),
+						c.EXPECT().Get(ctx, kutil.Key(namespace, ShootManagedResourceName), gomock.AssignableToTypeOf(&resourcesv1alpha1.ManagedResource{})),
+						c.EXPECT().Update(ctx, gomock.AssignableToTypeOf(&resourcesv1alpha1.ManagedResource{})).Return(fakeErr),
+					)
+
+					Expect(kubeControllerManager.Deploy(ctx)).To(MatchError(fakeErr))
+				})
+			})
+		})
+		Context("Tests expecting success", func() {
```
rfranzke, line 295:
```suggestion

		Context("Tests expecting success", func() {
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
```diff
@@ -0,0 +1,802 @@
...
+func getHorizontalPodAutoscalerConfig(config *gardencorev1beta1.HorizontalPodAutoscalerConfig) *gardencorev1beta1.HorizontalPodAutoscalerConfig {
+	defaultHPATolerance := gardencorev1beta1.DefaultHPATolerance
+	horizontalPodAutoscalerConfig := gardencorev1beta1.HorizontalPodAutoscalerConfig{
+		CPUInitializationPeriod: &metav1.Duration{Duration: gardencorev1beta1.DefaultCPUInitializationPeriod},
+		DownscaleDelay:          &metav1.Duration{Duration: gardencorev1beta1.DefaultHPADownscaleDelay},
+		DownscaleStabilization:  &metav1.Duration{Duration: gardencorev1beta1.DefaultDownscaleStabilization},
+		InitialReadinessDelay:   &metav1.Duration{Duration: gardencorev1beta1.DefaultInitialReadinessDelay},
+		SyncPeriod:              &metav1.Duration{Duration: gardencorev1beta1.DefaultHPASyncPeriod},
+		Tolerance:               &defaultHPATolerance,
+		UpscaleDelay:            &metav1.Duration{Duration: gardencorev1beta1.DefaultHPAUpscaleDelay},
```
rfranzke: don't use the constants here but the plain values

rfranzke (changes requested):
Nice PR, thank you very much.
Additionally to the inline comments:

1. You can remove https://github.com/gardener/gardener/blob/master/charts/shoot-core/components/values.yaml#L21-L22
1. You can remove https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/addons.go#L453
1. Use the `ServiceName` constant here: https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/wanted_secrets.go#L125
1. Use the `SecretName` constant here: https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/wanted_secrets.go#L197
1. Use the `SecretNameServer` constant here: https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/wanted_secrets.go#L216 (and delete the other constant from the `common` package)
1. Does it make sense to move [this code](https://github.com/gardener/gardener/blob/master/pkg/gardenlet/controller/shoot/shoot_control_delete.go#L154-L164) into the new component package as well and call it?
1. Does it make sense to add a `WakeUp` function in the component for [this code](https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/controlplane.go#L298-L360) and call it?
1. [This code](https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/waiter.go#L193-L292) should definitely be moved into the component, and if possible, you could add a unit test for it as well?

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
danielfoehrKn: So which importsorting tool are we using now ? I removed my own

danielfoehrKn:
Working on this asap.

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
danielfoehrKn: Used in the kube controller manager test

danielfoehrKn:
Updated the PR. Thanks for the review!
> Does it make sense to move this code into the new component package as well and call it?

I feel like that there is already a bunch of stuff in the new component and this code is only related to the deletion flow - I would vote for keeping it there.

> Does it make sense to add a WakeUp function in the component for this code and call it?

The only thing it does is scaling up the kcm deployment. Not sure if that extra encapsulation would add value.

> This code should definitely be moved into the component, and if possible, you could add a unit test for it as well?

Done.
The function WaitForControllersToBeActive() does not fail if the kube controller manager deployment cannot be found.Is that desired?  
I added a unit test for it but I am not sure.

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
rfranzke: There is no need to expose a constant just for a unit test. Please make it unexported.

rfranzke:
> I feel like that there is already a bunch of stuff in the new component and this code is only related to the deletion flow - I would vote for keeping it there.

Yeah, agreed, it's pretty generic and the relevant deployment name is already exported in a constant.

> The only thing it does is scaling up the kcm deployment. Not sure if that extra encapsulation would add value.

+1, agreed.

> The function WaitForControllersToBeActive() does not fail if the kube controller manager deployment cannot be found.Is that desired?
I added a unit test for it but I am not sure.

Hm, no, probably this case should be added. Nice catch, thanks.

/needs rebase

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter.go
```diff
@@ -0,0 +1,139 @@
...
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	"github.com/gardener/gardener/pkg/operation/common"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+	"github.com/gardener/gardener/pkg/utils/retry"
+)
+
+func (k *kubeControllerManager) Wait(_ context.Context) error        { return nil }
+func (k *kubeControllerManager) WaitCleanup(_ context.Context) error { return nil }
+
+func (k *kubeControllerManager) WaitForControllersToBeActive(ctx context.Context) error {
```
rfranzke:
```suggestion
func (k *kubeControllerManager) WaitForControllerToBeActive(ctx context.Context) error {
```

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter_test.go
```diff
@@ -0,0 +1,230 @@
+package kubecontrollermanager
```
rfranzke, line 15: license header missing

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter.go
```diff
@@ -0,0 +1,139 @@
...
+			return fmt.Errorf("kube controller manager deployment not found: %v", err)
+		}
+		return err
+	}
+
+	controllers = append(controllers, controllerInfo{
+		name: v1beta1constants.DeploymentNameKubeControllerManager,
+		labels: map[string]string{
+			"app":  "kubernetes",
+			"role": "controller-manager",
```
rfranzke:
```suggestion
			v1beta1constants.LabelApp:  v1beta1constants.LabelKubernetes,
			v1beta1constants.LabelRole: labelRole,
```

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter.go
```diff
@@ -0,0 +1,139 @@
...
+func (k *kubeControllerManager) Wait(_ context.Context) error        { return nil }
+func (k *kubeControllerManager) WaitCleanup(_ context.Context) error { return nil }
+
+func (k *kubeControllerManager) WaitForControllersToBeActive(ctx context.Context) error {
+	type controllerInfo struct {
+		name   string
+		labels map[string]string
+	}
+
+	type checkOutput struct {
```
rfranzke: Can this coding related to parallelism be removed now? There is only one KCM pod for a cluster, so there is no need to parallelise anything.

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,541 @@
...
+	ServiceAccountKey component.Secret
+}
+
+// Secret is a structure that contains information about a Kubernetes secret which is managed externally.
+type Secret struct {
+	// Name is the name of the Kubernetes secret object.
+	Name string
+	// Checksum is the checksum of the secret's data.
+	Checksum string
+}
```
rfranzke: This can be removed now, or?

rfranzke (changes requested):

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
```diff
@@ -0,0 +1,828 @@
...
+									"role": "controller-manager",
+								},
+							},
+							Template: corev1.PodTemplateSpec{
+								ObjectMeta: metav1.ObjectMeta{
+									Annotations: map[string]string{
+										"checksum/secret-" + "ca":                             secretChecksumCA,
+										"checksum/secret-" + "service-account-key":            secretChecksumServiceAccountKey,
+										"checksum/secret-" + "kube-controller-manager":        secretChecksumKubeconfig,
+										"checksum/secret-" + "kube-controller-manager-server": secretChecksumServer,
```
ialidzhikov:
```suggestion
										"checksum/secret-ca":                             secretChecksumCA,
										"checksum/secret-service-account-key":            secretChecksumServiceAccountKey,
										"checksum/secret-kube-controller-manager":        secretChecksumKubeconfig,
										"checksum/secret-kube-controller-manager-server": secretChecksumServer,
```

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter_test.go
```diff
@@ -0,0 +1,244 @@
...
+						},
+					}
+					return nil
+				}),
+			)
+
+			err := kubeControllerManager.WaitForControllerToBeActive(ctx)
+			Expect(err).To(HaveOccurred())
+			s := err.Error()
+			Expect(s).To(Equal("retry failed with max attempts reached, last error: controller kube-controller-manager is not active"))
```
ialidzhikov:
```suggestion
			Expect(err.Error()).To(Equal("retry failed with max attempts reached, last error: controller kube-controller-manager is not active"))
```

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
```diff
@@ -0,0 +1,828 @@
...
+}
+
+func probeSchemeForKubernetesVersion(version string) corev1.URIScheme {
+	if k8sVersionGreaterEqual113, _ := versionutils.CompareVersions(version, ">=", "1.13"); k8sVersionGreaterEqual113 {
+		return corev1.URISchemeHTTPS
+	}
+	return corev1.URISchemeHTTP
+}
+
+func commandForKubernetesVersion(version string, port int32, nodeCIDRMaskSize *int32, clusterName string, serviceNetwork, podNetwork *net.IPNet, horizontalPodAutoscalerConfig *gardencorev1beta1.HorizontalPodAutoscalerConfig, featureGateFlags string) []string {
```
ialidzhikov, line 695: Isn't this the `computeCommand` func that exists in `pkg/operation/botanist/controlplane/kube_controller_manager.go`. Can we somehow avoid duplicating the utility functions in the tests? Imo it does not bring much value as `computeCommand` and `portForKubernetesVersion` funcs will be still untested.

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,533 @@
...
+		version:        version,
+		image:          image,
+		config:         config,
+		podNetwork:     podNetwork,
+		serviceNetwork: serviceNetwork,
+	}
+}
+
+type kubeControllerManager struct {
+	log            *logrus.Entry
```
ialidzhikov:
```suggestion
	log            logrus.FieldLogger
```

Can you use logrus. FieldLogger as it is the interface that generalise `logrus.Entry` and `logrus.Logger` types. In this way we would allow `logrus.Entry` or `logrus.Logger` to be passed without forcing the caller to convert from the one to another. See https://github.com/gardener/gardener/pull/2760

pkg/operation/botanist/controlplane/kubecontrollermanager/shoot_resources.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"context"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/authentication/user"
+
+	"github.com/gardener/gardener/pkg/client/kubernetes"
```
ialidzhikov: You could move the gardener/gardener imports as second import group.

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
ialidzhikov, line 30: You could move the gardener/gardener imports as second import group.

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter.go
```diff
@@ -0,0 +1,89 @@
...
+	return retry.UntilTimeout(ctx, pollInterval, pollTimeout, func(ctx context.Context) (done bool, err error) {
+		podList := &corev1.PodList{}
+		err = k.seedClient.List(ctx, podList,
+			client.InNamespace(k.namespace),
+			client.MatchingLabels(map[string]string{
+				v1beta1constants.LabelApp:  v1beta1constants.LabelKubernetes,
+				v1beta1constants.LabelRole: labelRole,
+			}))
+		if err != nil {
+			return retry.SevereError(fmt.Errorf("could not check whether controller %s is active: %+v", v1beta1constants.DeploymentNameKubeControllerManager, err))
```
ialidzhikov:
```suggestion
			return retry.SevereError(fmt.Errorf("could not check whether controller %s is active: %w", v1beta1constants.DeploymentNameKubeControllerManager, err))
```

ialidzhikov (commented):
`make start-gardenlet` panics with nil pointer:

```
E1007 15:02:48.552534    8862 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 605 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x288d3e0, 0x3aee5e0)
	/Users/me/go/src/github.com/gardener/gardener/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0xa6
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/Users/me/go/src/github.com/gardener/gardener/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x89
panic(0x288d3e0, 0x3aee5e0)
	/usr/local/opt/go/libexec/src/runtime/panic.go:969 +0x175
github.com/gardener/gardener/pkg/operation/botanist.(*Botanist).DefaultKubeControllerManager(0xc0004e1680, 0x2daf5a0, 0xc000caa480, 0x0, 0x0)
	/Users/me/go/src/github.com/gardener/gardener/pkg/operation/botanist/controlplane.go:1262 +0x2af
```

danielfoehrKn:
Will work on this next week. Thanks for the review @ialidzhikov

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
danielfoehrKn: > Isn't this the computeCommand func that exists in pkg/operation/botanist/controlplane/kube_controller_manager.go. Can we somehow avoid duplicating the utility functions in the tests? 

It is but without using constants. This way the unit tests will fail if the command computation has been changed incorrectly. 
You can also check the clusterautoscaler and the kubescheduler tests.

charts/seed-controlplane/charts/kube-controller-manager/templates/kube-controller-manager.yaml
```diff
@@ -1,149 +0,0 @@
...
-        networking.gardener.cloud/to-dns: allowed
-        networking.gardener.cloud/to-shoot-apiserver: allowed
-        networking.gardener.cloud/from-prometheus: allowed
-{{- if .Values.podLabels }}
-{{ toYaml .Values.podLabels | indent 8 }}
-{{- end }}
-    spec:
-      containers:
-      - name: kube-controller-manager
-        image: {{ index .Values.images "kube-controller-manager" }}
```
BeckerMax, line 36: Can we also delete the entry in of `kube-controller-manager` in charts/images.yaml ?

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -274,7 +274,7 @@ func (c *Controller) runDeleteShootFlow(ctx context.Context, o *operation.Operat
 		})
 		waitForControllersToBeActive = g.Add(flow.Task{
 			Name:         "Waiting until kube-controller-manager is active",
-			Fn:           flow.TaskFn(botanist.WaitForControllersToBeActive).DoIf(cleanupShootResources).RetryUntilTimeout(defaultInterval, defaultTimeout),
+			Fn:           flow.TaskFn(botanist.WaitForKubeControllerManagerToBeActive).DoIf(cleanupShootResources).RetryUntilTimeout(defaultInterval, defaultTimeout),
```
BeckerMax, line 388: Like mentioned below: I feel this should be multiple commits (or even PRs). This change is separate from the helm chart refactorings.

pkg/operation/botanist/controlplane.go
```diff
@@ -1248,67 +1249,60 @@ func IsValidAuditPolicyVersion(shootVersion string, schemaVersion *schema.GroupV
...
 			}
-			defaultValues["replicas"] = replicaCount
+			replicaCount = replicas
 		} else {
 			// shoot is being reconciled with .spec.hibernation.enabled!=.status.isHibernated, so deploy KCM. in case the
 			// shoot is being hibernated then it will be scaled down to zero later after all machines are gone
-			defaultValues["replicas"] = 1
+			replicaCount = 1
 		}
 	}
```
BeckerMax: Extract to separate function for example 'wantedReplicaCount' returning the replicaCount

pkg/operation/botanist/controlplane/clusterautoscaler/monitoring_test.go
```diff
@@ -26,7 +26,7 @@ var _ = Describe("Monitoring", func() {
 
 	Describe("#ScapeConfig", func() {
 		It("should successfully test the scape configuration", func() {
```
BeckerMax: scape -> scrape

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,536 @@
...
+
+const (
+	// ServiceName is the name of the service of the kube-controller-manager.
+	ServiceName = "kube-controller-manager"
+	// kubeControllerManagerContainerName is the name of the kube-controller-manager container
+	kubeControllerManagerContainerName = v1beta1constants.DeploymentNameKubeControllerManager
+
+	// SecretName is a constant for the secret name for the kube-controller-manager's kubeconfig secret.
+	SecretName = "kube-controller-manager"
+	// SecretNameServer is the name of the kube-controller-managerr server certificate secret.
```
BeckerMax: managerr -> manager

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,536 @@
...
+	SecretNameServer = "kube-controller-manager-server"
+
+	// labelRole is a constant for the value of a label with key 'role' whose value is 'controller-manager'.
```
BeckerMax: Feels a bit strange to write the value of the const in the commentary. If you would ever change the value the comment could be forgotten to be changed.

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,536 @@
...
+	volumeMountPathKubeconfig = "/var/lib/kube-controller-manager"
+	// volumeMountPathServer is the volume mount path for the x509 TLS server certificate and key for the HTTPS server inside the kube-controller-manager (which is used for metrics and health checks).
+	volumeMountPathServer = "/var/lib/kube-controller-manager-server"
+
+	// managedResourceName is the name of the managed resource that contains the resources to be deployed into the Shoot cluster.
+	managedResourceName = "shoot-core-kube-controller-manager"
+)
+
+// KubeControllerManager contains functions for a kube-controller-manager deployer.
+type KubeControllerManager interface {
```
BeckerMax, line 77: I am curious why in g/g we are declaring the interface right next to the implementation. Is this a kubernetes thing? Because in Golang I think you generally declare the Interface on the Consumer side. Advantages are that you can define the Interface for the consumer as narrow as possible. Also I am not sure if an interface in this case is even necessary. Why not reference the concrete type? https://github.com/golang/go/wiki/CodeReviewComments#interfaces

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
```diff
@@ -0,0 +1,536 @@
...
+	if k.secrets.ServiceAccountKey.Name == "" || k.secrets.ServiceAccountKey.Checksum == "" {
+		return fmt.Errorf("missing ServiceAccountKey secret information")
+	}
+
+	var (
+		vpa        = k.emptyVPA()
+		service    = k.emptyService()
+		deployment = k.emptyDeployment()
+
+		labels = map[string]string{
```
BeckerMax: Because we saw that in the past already : https://github.com/gardener/gardener/pull/2860/files
Do you have an idea how we could make the usage of maps more safe?
If anywhere in this large function something accidentally adds a field to the map it will have side-effects on all usages. 
As there are no constant maps: I guess we could declare the map at every place we use it instead of sharing it (would cost performance probably?). WDYT?

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
```diff
@@ -0,0 +1,824 @@
...
+				BeforeEach(func() {
+					kubeControllerManager.SetSecrets(Secrets{
+						Kubeconfig:        component.Secret{Name: "kube-controller-manager", Checksum: secretChecksumKubeconfig},
+						Server:            component.Secret{Name: "kube-controller-manager-server", Checksum: secretChecksumServer},
+						CA:                component.Secret{Name: "ca", Checksum: secretChecksumCA},
+						ServiceAccountKey: component.Secret{Name: "service-account-key", Checksum: secretChecksumServiceAccountKey},
+					})
+				})
+
+				It("should fail because the service cannot be created", func() {
```
BeckerMax: should fail **when** the service cannot be created

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
```diff
@@ -0,0 +1,824 @@
...
+						c.EXPECT().Get(ctx, kutil.Key(namespace, vpaName), gomock.AssignableToTypeOf(&autoscalingv1beta2.VerticalPodAutoscaler{})),
+						c.EXPECT().Update(ctx, gomock.AssignableToTypeOf(&autoscalingv1beta2.VerticalPodAutoscaler{})).Do(func(ctx context.Context, obj runtime.Object, opts ...client.UpdateOption) {
+							Expect(obj).To(DeepEqual(vpa))
+						}),
+					)
+
+					is112Version, _ := versionutils.CompareVersions(version, "~", "1.12")
+					is113Version, _ := versionutils.CompareVersions(version, "~", "1.13")
+
+					if k8sVersion := is112Version || is113Version; k8sVersion {
```
BeckerMax: if is112Version || is113Version {
}

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
```diff
@@ -0,0 +1,824 @@
...
+							c.EXPECT().Update(ctx, gomock.AssignableToTypeOf(&corev1.Secret{})).Do(func(ctx context.Context, obj runtime.Object, opts ...client.UpdateOption) {
+								Expect(obj).To(DeepEqual(managedResourceSecret))
+							}),
+							c.EXPECT().Get(ctx, kutil.Key(namespace, managedResourceName), gomock.AssignableToTypeOf(&resourcesv1alpha1.ManagedResource{})),
+							c.EXPECT().Update(ctx, gomock.AssignableToTypeOf(&resourcesv1alpha1.ManagedResource{})).Do(func(ctx context.Context, obj runtime.Object, opts ...client.UpdateOption) {
+								Expect(obj).To(DeepEqual(managedResource))
+							}),
+						)
+					} else {
+						gomock.InOrder(
```
BeckerMax, line 604: Out of of curiosity:
Why is a Delete expected in the deploy flow when the version is larger than 1.13?

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
```diff
@@ -0,0 +1,824 @@
...
+		It("should return nil as it's not implemented as of now", func() {
+			Expect(kubeControllerManager.WaitCleanup(ctx)).To(Succeed())
+		})
+	})
+})
+
+// Utility functions
+
+func portForKubernetesVersion(version string) int32 {
+	if k8sVersionGreaterEqual113, _ := versionutils.CompareVersions(version, ">=", "1.13"); k8sVersionGreaterEqual113 {
```
BeckerMax, line 682: versionutils.CompareVersions(version, ">=", "1.13")  is already readable enough. I don't think you need the variable `k8sVersionGreaterEqual113 `

pkg/operation/botanist/waiter.go
```diff
@@ -158,6 +156,13 @@ func (b *Botanist) WaitUntilKubeAPIServerReady(ctx context.Context) error {
 	})
 }
 
+// WaitForKubeControllerManagerToBeActive waits for the kube controller manager of a Shoot cluster has acquired leader election, thus is active.
+func (b *Botanist) WaitForKubeControllerManagerToBeActive(ctx context.Context) error {
```
BeckerMax, line 202: How is this change related to the rest of the changes. I think it should at least be a different commit to the helm chart refactorings

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter.go
```diff
@@ -0,0 +1,89 @@
...
+			return retry.MinorError(fmt.Errorf("controller %s is not active", v1beta1constants.DeploymentNameKubeControllerManager))
+		}
+
+		// Check if the controller is active by reading its leader election record.
+		leaderElectionRecord, err := common.ReadLeaderElectionRecord(ctx, k.shootClient, resourcelock.EndpointsResourceLock, metav1.NamespaceSystem, v1beta1constants.DeploymentNameKubeControllerManager)
+		if err != nil {
+			return retry.SevereError(fmt.Errorf("could not check whether controller %s is active: %w", v1beta1constants.DeploymentNameKubeControllerManager, err))
+		}
+
+		if delta := metav1.Now().UTC().Sub(leaderElectionRecord.RenewTime.Time.UTC()); delta <= pollInterval-time.Second {
```
BeckerMax: Can you help me understanding this by explaining what this line does?

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter_test.go
```diff
@@ -0,0 +1,244 @@
...
+			gomock.InOrder(
+				seedClient.EXPECT().Get(ctx, kutil.Key(namespace, "kube-controller-manager"), gomock.AssignableToTypeOf(&appsv1.Deployment{})),
+				seedClient.EXPECT().List(gomock.Any(), gomock.AssignableToTypeOf(&corev1.PodList{}), listOptions).Return(fakeErr),
+			)
+
+			err := kubeControllerManager.WaitForControllerToBeActive(ctx)
+			Expect(err).To(MatchError(fmt.Sprintf("could not check whether controller kube-controller-manager is active: %s", errorMsg)))
+		})
+
+		It("should fail if there is more than one kube controller manager pod", func() {
```
BeckerMax, line 130: Maybe add a test if there are no pods returned (but deployment exists)?

BeckerMax (changes requested):

BeckerMax:
In general I think this PR adds a lot of value and readability.
What I did not review in detail was if every flag was correctly translated from helm chart to golang. Also the waiter logic is new to me and I can not review if this correctly used. Maybe someone else with more experience can have a look at that again.
In general it would have helped me a lot as a new reviewer if you would have split it up into multiple commits which I can separately look at. Also it feels that some parts could would have been more readable if refactored into a function. I left some comments regarding this. Thanks!

charts/seed-controlplane/charts/kube-controller-manager/templates/kube-controller-manager.yaml
rfranzke: No, we still need the image in the image vector (which is the `charts/images.yaml` file).

pkg/gardenlet/controller/shoot/shoot_control_delete.go
rfranzke: I have no strong opinion about multiple commits, but please keep it in this PR.

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
rfranzke: What about replacing it with a function that returns the map and calling it below? (If you agree, this should also be done for `kubescheduler`/`clusterautoscaler` packages).

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
rfranzke: I guess there is no strong reason for it why we do it this way. We could also define the interfaces in the `botanist` package (consumer), but I don't see a clear benefit now from doing it this way. Do you? For now, it probably "cleaner" to have it in the same package, not sure. What do other folks think?

> Also I am not sure if an interface in this case is even necessary. Why not reference the concrete type?

In the future we might have mocks for it when we test higher-level functions.

pkg/gardenlet/controller/shoot/shoot_control_delete.go
danielfoehrKn: @BeckerMax okay thanks for the review Max. Is it okay if I keep it in one commit for this PR as most people already review it?. I ll try to find more reasonable commit sizes for the upcoming kube-apiserver refactoring.

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
danielfoehrKn: Thanks @BeckerMax good suggestion. I replaced it with calls to a function that return a new map.

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
danielfoehrKn: Because below that version there was a managed resource deployed - when upgrading the K8s version, it should be deleted.

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager_test.go
danielfoehrKn: I only need it because versionutils.CompareVersions. returns two arguments :)

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter.go
danielfoehrKn: I did not create this LoC but merely moved the code block.
With the help of @rfranzke : added a comment explaining it.

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter_test.go
danielfoehrKn: That is a very good idea. This will prevent successful reconciliations if the watchdog downscaled the KCM pod.

danielfoehrKn: Added a test that should fail if no kube controller manager pod can be found

pkg/operation/botanist/waiter.go
danielfoehrKn: see comment above, will do on next refactoring. Thanks

pkg/gardenlet/controller/shoot/shoot_control_delete.go
BeckerMax: Sure, it is fine to keep it in one commit here.
What I wanted to raise is the general feedback that structuring in commits helps reviewers who are not familiar with the code base to understand the changes. Also reverts (if needed) are easier.

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
BeckerMax: @rfranzke WDYT of the comments here: https://github.com/golang/go/wiki/CodeReviewComments#interfaces

> Do not define interfaces on the implementor side of an API "for mocking"; instead, design the API so that it can be tested using the public API of the real implementation.

Seems to be the common practice in go:
https://stackoverflow.com/questions/55105509/golang-interfaces-mocking

rfranzke: I don't agree to this statement in general. Mocks will make testing higher-level function (like the shoot flow) much easier. If you don't mock the KCM then you have to duplicate its behaviour in the shoot flow test again (although you already have tests in the KCM package).

BeckerMax: But I agree with you that it is convenient for mocks.

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter.go
BeckerMax: Thanks that makes it clearer. I don't yet understand how the pollInterval is related to that though.

WDYT about explaining it in code instead?
```
  lastRenew:= metav1.Now().UTC().Sub(leaderElectionRecord.RenewTime.Time.UTC()
  defaultRenewalTime := 2*time.Second
  if lastRenew <= defaultRenewalTime * 2 {
    return retry.Ok()
  }
```
cc @rfranzke

pkg/operation/botanist/controlplane/kubecontrollermanager/shoot_resources.go
```diff
@@ -0,0 +1,86 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/authentication/user"
+)
+
+func (k *kubeControllerManager) deployShootManagedResource(ctx context.Context) error {
+	if versionConstraintK8sGreaterEqual112.Check(k.version) && versionConstraintK8sSmaller114.Check(k.version) {
+		return common.DeployManagedResource(ctx, k.seedClient, managedResourceName, k.namespace, false, k.computeShootResourcesData())
```
rfranzke:
```suggestion
		return common.DeployManagedResourceForShoot(ctx, k.seedClient, managedResourceName, k.namespace, false, k.computeShootResourcesData())
```

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
```diff
@@ -126,13 +126,6 @@ func (k *kubeScheduler) Deploy(ctx context.Context) error {
 		service    = k.emptyService()
 		deployment = k.emptyDeployment()
 
-		labels = map[string]string{
```
rfranzke, line 129: Can you also adapt it for the `clusterautoscaler` pkg?

rfranzke (changes requested):
only nits now

pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go
danielfoehrKn: yes, that was done already.

pkg/operation/seed/seed.go
```diff
@@ -345,6 +346,7 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
 		componentsFunctions := []component.LoggingConfiguration{
 			clusterautoscaler.LoggingConfiguration,
 			kubescheduler.LoggingConfiguration,
+			kubecontrollermanager.LoggingConfiguration,
 		}
 		userAllowedComponents := []string{v1beta1constants.DeploymentNameKubeAPIServer, v1beta1constants.DeploymentNameKubeControllerManager}
```
```suggestion
		userAllowedComponents := []string{v1beta1constants.DeploymentNameKubeAPIServer}
```

rfranzke (changes requested):
Thanks, one last nit

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter.go
danielfoehrKn: Looks good to me - added the code in the way suggested.

pkg/operation/seed/seed.go
danielfoehrKn: thanks for spotting.

rfranzke (dismissed):
/lgtm

pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go
BeckerMax: @danielfoehrKn do you also do that for the kubescheduler/clusterautoscaler packages?

pkg/operation/botanist/controlplane/kubecontrollermanager/waiter.go
BeckerMax: I would delete the comment then as it is now redundant 😉

BeckerMax (approved):

rfranzke (approved):

PR #2888, opened 2020-09-24
Shoot conditions `lastUpdateTime` is no longer updated continuously

rfranzke:

/area scalability robustness
/kind enhancement
/priority normal

The shoot's `.status.conditions[].lastUpdateTime` is no longer continuously updated to prevent spamming etcd with too many unnecessary updates.

If the gardenlet does not renew its lease anymore then the GCM will, after a certain threshold, set the shoot conditions to `Unknown` to indicate that the displayed information is most likely outdated.

**Release note**:

```improvement operator
The shoot's `.status.conditions[].lastUpdateTime` is no longer continuously updated. Note that if the gardenlet does not renew its lease anymore then the gardener-controller-manager will, after a certain threshold, set the `Shoot`'s conditions to `Unknown` to indicate that the displayed information is most likely outdated.
```


vlerenc:
Thanks @rfranzke. When we merge it/deploy it to dev, we should check whether it indeed behaves as expected and the no longer updated `.status.conditions[].lastUpdateTime` will not cause trouble in the components.

WDYT if we no longer update `.status.conditions[].lastUpdateTime` for shoots:
* @grolu for the dashboard
* @wyb1 @mliepold for the availability monitoring

petersutter:
we have no logic that depends on `.status.conditions[].lastUpdateTime`. 

This would be the only effect in case it is not set (Last Updated will show nothing), however we have a PR ready to remove this field https://github.com/gardener/dashboard/pull/810

Screenshot 2020-09-24 at 15 55 52

mvladev:
Btw there is already a condition builder which checks if condition really needs to be updated e.g. https://github.com/gardener/gardener/blob/2fb3b07c5c741b432889ce155a89a82cd88532b1/pkg/gardenlet/controller/seed/seed_lease_control.go#L101-L120

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -92,13 +76,16 @@ func UpdatedCondition(condition gardencorev1beta1.Condition, status gardencorev1
...
-		LastUpdateTime:     Now(),
+		LastUpdateTime:     condition.LastUpdateTime,
 		Codes:              codes,
 	}
 
 	if condition.Status != status {
-		newCondition.LastTransitionTime = Now()
+		now := Now()
+		newCondition.LastTransitionTime = now
+		newCondition.LastUpdateTime = now
```
mvladev: The LastUpdate and LastTransitionTime should be updated only under specific conditions e.g. update to status / reason or message:

https://github.com/gardener/gardener/blob/24334336a6f069b00ad5c7b839fac980aa9ea5f8/pkg/apis/core/v1beta1/helper/condition_builder.go#L100-L155

mvladev (changes requested):

amshuman-kr (dismissed):
I am not aware of all the details of how these conditions and constraints are used. But looks good in principle :-)

petersutter (dismissed):

rfranzke:
> Btw there is already a condition builder which checks if condition really needs to be updated e.g.

Yes, I know, actually I started using it, but then I faced that the `botanist/health_check.go` would be adapted as well, resulting in a larger refactoring. I have a WIP branch for it and will continue separately.

pkg/apis/core/v1beta1/helper/helper.go
rfranzke: done

pkg/utils/kubernetes/health/health.go
```diff
@@ -342,10 +342,10 @@ func CheckAPIServerAvailability(ctx context.Context, condition gardencorev1beta1
 		} else {
 			body = string(bodyRaw)
 		}
-		message := fmt.Sprintf("API server /healthz endpoint endpoint check returned a non ok status code %d. %s (%s)", statusCode, responseDurationText, body)
+		message := fmt.Sprintf("API server /healthz endpoint endpoint check returned a non ok status code %d. (%s)", statusCode, body)
```
vpnachev:
```suggestion
		message := fmt.Sprintf("API server /healthz endpoint check returned a non ok status code %d. (%s)", statusCode, body)
```

vpnachev (dismissed):
/lgtm

vpnachev (approved):
/lgtm

mvladev (approved):
/lgtm

PR #2898, opened 2020-09-25
Redeploy Backup{Bucket,Entry} secrets in deletion flow

rfranzke:

/area usability
/kind bug
/priority normal

With this PR we are redeploying the secrets for `Backup{Bucket,Entry}` resource in their deletion flow (in case they have changed in the meantime and the resources shall be deleted).
As part of the PR I have also factored out the other small steps in the `deploy()` functions into dedicated flow tasks.

/cc @RaphaelVogel - thanks for reporting

**Release note**:

```improvement operator
A bug has been fixed that prevented `BackupBucket` and `BackupEntry` resources from being deleted despite the respective infrastructure provider credentials have been updated to the correct data.
```

### DIFF
--- a/pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go
+++ b/pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go
@@ -73,16 +73,21 @@ func (a *actuator) Reconcile(ctx context.Context) error {
 	var (
 		g = flow.NewGraph("Backup Bucket Reconciliation")
 
-		deployBackupBucketExtension = g.Add(flow.Task{
-			Name: "Deploying backup bucket extension resource",
-			Fn:   flow.TaskFn(a.deployBackupBucketExtension).RetryUntilTimeout(defaultInterval, defaultTimeout),
+		deployBackupBucketSecret = g.Add(flow.Task{
+			Name: "Deploying backup bucket secret to seed",
+			Fn:   flow.TaskFn(a.deployBackupBucketExtensionSecret).RetryUntilTimeout(defaultInterval, defaultTimeout),
 		})
-
+		deployBackupBucketExtension = g.Add(flow.Task{
+			Name:         "Deploying backup bucket extension resource",
+			Fn:           flow.TaskFn(a.deployBackupBucketExtension).RetryUntilTimeout(defaultInterval, defaultTimeout),
+			Dependencies: flow.NewTaskIDs(deployBackupBucketSecret),
+		})
 		_ = g.Add(flow.Task{
 			Name:         "Waiting until backup bucket is reconciled",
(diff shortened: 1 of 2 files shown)


pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go
```diff
@@ -95,18 +100,35 @@ func (a *actuator) Reconcile(ctx context.Context) error {
...
+			Name: "Destroying generated backup bucket secret in garden cluster",
+			Fn:   a.deleteGeneratedBackupBucketSecretInGarden,
+		})
+		deployBackupBucketSecret = g.Add(flow.Task{
+			Name: "Deploying backup bucket secret to seed",
+			Fn:   flow.TaskFn(a.deployBackupBucketExtensionSecret).RetryUntilTimeout(defaultInterval, defaultTimeout),
+		})
+		deleteBackupBucketExtensionSecret = g.Add(flow.Task{
+			Name:         "Deleting backup bucket secret in seed",
+			Fn:           flow.TaskFn(a.deleteBackupBucketExtensionSecret).RetryUntilTimeout(defaultInterval, defaultTimeout),
```
ialidzhikov: Hm, in this step don't we delete the just redeployed secret? If yes, I am wondering how the redeploying the secret helps as we delete it right away.

pkg/gardenlet/controller/backupentry/backup_entry_actuator.go
```diff
@@ -97,13 +108,28 @@ func (a *actuator) Delete(ctx context.Context) error {
...
+			Fn:   a.waitUntilBackupBucketReconciled,
+		})
+		deployBackupEntrySecret = g.Add(flow.Task{
+			Name:         "Deploying backup entry secret to seed",
+			Fn:           flow.TaskFn(a.deployBackupEntrySecret).RetryUntilTimeout(defaultInterval, defaultTimeout),
+			Dependencies: flow.NewTaskIDs(waitUntilBackupBucketReconciled),
+		})
+		deleteBackupEntryExtensionSecret = g.Add(flow.Task{
+			Name:         "Deleting backup entry secret in seed",
+			Fn:           flow.TaskFn(a.deleteBackupEntryExtensionSecret).RetryUntilTimeout(defaultInterval, defaultTimeout),
```
ialidzhikov: The same question also here.

pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go
rfranzke: The redeployment updates the credentials in case they were changed. This step here sets the deletion timestamp (finalizer protects it from being removed). I didn't change the order, but I agree that it's confusing. I will move it down after the deletion of the bucket resource to make it more clear.

ialidzhikov (approved):
/lgtm

PR #2877, opened 2020-09-22
Fix SNI phase during cluster creation

danielfoehrKn:
**How to categorize this PR?**

/area control-plane
/kind bug
/priority normal

**What this PR does / why we need it**:

Currently cluster creation fails because the botanist cannot be created as the `kube-apiserver` service in the Seed does not exist yet.

 PR integration tests would have prevent that issue.
**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement operator

```

PR #2901, opened 2020-09-25
Fix kube-apiserver service type when SNI is enabled

mvladev:

/area control-plane
/kind regression
/priority normal

kube-apiserver service is now properly created with type `ClusterIP` when SNI is enabled and it doesn't exist.

Fixes # n/a

Unfortunately due to the fix in #2877 a bug was introduced - `svc` is never `nil` and therefore the kube-apiserver service was always created with type `LoadBalancer`.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -1359,12 +1359,6 @@ func (b *Botanist) SNIPhase(ctx context.Context) (component.Phase, error) {
 	}
 
 	switch {
-	case svc == nil && sniEnabled:
-		// initial cluster creation with SNI enabled (enabling only relevant for migration)
-		return component.PhaseEnabled, nil
-	case svc == nil && !sniEnabled:
-		// initial cluster creation without SNI enabled
-		return component.PhaseDisabled, nil
 	case svc.Spec.Type == corev1.ServiceTypeLoadBalancer && sniEnabled:
 		return component.PhaseEnabling, nil
 	case svc.Spec.Type == corev1.ServiceTypeClusterIP && sniEnabled:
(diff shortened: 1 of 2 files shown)


ialidzhikov (dismissed):
/lgtm
/invite @danielfoehrKn

pkg/operation/botanist/controlplane.go
```diff
@@ -1359,19 +1359,16 @@ func (b *Botanist) SNIPhase(ctx context.Context) (component.Phase, error) {
...
-		// initial cluster creation without SNI enabled
-		return component.PhaseDisabled, nil
 	case svc.Spec.Type == corev1.ServiceTypeLoadBalancer && sniEnabled:
 		return component.PhaseEnabling, nil
 	case svc.Spec.Type == corev1.ServiceTypeClusterIP && sniEnabled:
 		return component.PhaseEnabled, nil
 	case svc.Spec.Type == corev1.ServiceTypeClusterIP && !sniEnabled:
 		return component.PhaseDisabling, nil
 	default:
+		if sniEnabled {
```
danielfoehrKn, line 1369:
```suggestion
              // initial cluster creation with SNI enabled (enabling only relevant for migration)
		if sniEnabled {
```

pkg/operation/botanist/controlplane.go
```diff
@@ -1359,19 +1359,16 @@ func (b *Botanist) SNIPhase(ctx context.Context) (component.Phase, error) {
...
+			return component.PhaseEnabled, nil
+		}
 		return component.PhaseDisabled, nil
```
danielfoehrKn, line 1374:
```suggestion
              // initial cluster creation with SNI disabled
		return component.PhaseDisabled, nil
```

danielfoehrKn (commented):
Thanks.

mvladev:
Updated the PR with @danielfoehrKn comments

ialidzhikov (approved):
/lgtm

PR #2903, opened 2020-09-25
fix secrets.Interface not matching implementing struct secrets.Secrets

kon-angelo:

/area control-plane
/kind api-change
/priority normal

The implementation of `pkg/utils/secrets/secrets.go::Secrets` changed. As a result it no longer implements the interface ` pkg/utils/secrets/secrets.go::Interface` which results in compilation failing in the extensions when gardener is revendored there.

**Release note**:

```improvement operator
Fix `secrets.Interface` not matching the implementing struct `Secrets`
```


rfranzke (commented):
Can you please check the failing `verify` step?

ialidzhikov (approved):
Thanks for catching this!

/lgtm

PR #2910, opened 2020-09-28
Don't unnecessarily retrieve Seed secret

timebertt:

/area cost quality
/kind cleanup
/priority normal

Before this PR, the Seed kubeconfig secret was retrieved for each Shoot reconcile/care operation, although it is not needed anymore because the Seed client is already present in the ClientMap.
This PR removes all related code to avoid a lot of unnecessary API calls.

Fixes #

**Release note**:

```improvement operator
NONE
```


rfranzke (approved):
/lgtm

Thanks @vpnachev for providing the valuable data that pointed us to this no longer required code.

ialidzhikov (approved):
/lgtm

PR #2848, opened 2020-09-14
Set minimum resources for nginx VPA

vpnachev:
**How to categorize this PR?**

/area scalability
/kind enhancement
/priority normal

**What this PR does / why we need it**:
Set minimum resources for nginx VPA.

In some cases the VPA scale to very low values the CPU requests and limits for the nginx controller:
```bash
$ k get pod -l app=nginx-ingress,component=controller -o json | jq '.items[].spec.containers[].resources' -cr;  k top pod -l app=nginx-ingress,component=controller
{"limits":{"cpu":"92m","memory":"3254996328"},"requests":{"cpu":"23m","memory":"813749082"}}
NAME                                               CPU(cores)   MEMORY(bytes)
addons-nginx-ingress-controller-5865b5dbf4-c2kq9   19m          635Mi
```

But there are spike loads in the controller that the VPA cannot react fast enough and then it ends in the `CrashLoopBackOff`. In this example the resources seems fine, but when a spike occurs and something like 400m CPU are needed, it will start failing.
```bash
$ k get pod -l app=nginx-ingress,component=controller -o json | jq '.items[].spec.containers[].resources' -cr;  k top pod -l {"limits":{"cpu":"92m","memory":"992613920"},"requests":{"cpu":"23m","memory":"248153480"}}
NAME                                               CPU(cores)   MEMORY(bytes)
addons-nginx-ingress-controller-6c87957d9c-gz6qb   16m          172Mi
```
**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
/invite @amshuman-kr @wyb1 

**Release note**:

```improvement operator
The VPA for the nginx addon is now set to guarantee at least 100m cpu and 128Mi memory for the controller pod.
```

PR #2174, opened 2020-04-13
Revert "Optimize apiserver HVPA minAllowed config."

mvladev:
Reverts gardener/gardener#2056

To sum up the problem:

1. API server Pod/container is started with `100m` CPU requests + limits.
1. API server starts to do post-start hooks (fixing bootstrap ClusterRole / ClusterRoleBinding).
1. Previous operation is throttled due to low CPU.
1. Liveness checks fail because 2. is not finished.
1. Kubelet restarts the API server container which is still doing 2. (`Container kube-apiserver failed liveness probe, will be restarted`)
1. Go to 1.

Deadlock. See bellow for full details:

API servers fail to start and getting constantly restarted, because their healthcheck is failing with:

```
[+]ping ok
[+]etcd ok
[+]poststarthook/generic-apiserver-start-informers ok
[+]poststarthook/start-apiextensions-informers ok
[+]poststarthook/start-apiextensions-controllers ok
[+]poststarthook/bootstrap-controller ok
[-]poststarthook/rbac/bootstrap-roles failed: reason withheld
[+]poststarthook/ca-registration ok
[+]poststarthook/start-kube-apiserver-informers ok
[+]poststarthook/start-kube-aggregator-informers ok
[+]poststarthook/apiservice-registration-controller ok
[+]poststarthook/apiservice-status-available-controller ok
[+]poststarthook/apiservice-openapi-controller ok
[+]poststarthook/kube-apiserver-autoregistration ok
[+]autoregister-completion ok
healthz check failed
```

It takese quite a while for the API server to sync its RBAC roles, because it was CPU throttled - the initial CPU usage is quite high when the API server starts, but after that it goes down significantly.  Those poststart hooks are quite CPU intensive.

So this is a `kube-apiserver` which I started with `100` CPU requests and limits and I've modified it to have `120` seconds delay before `liveness` probe starts.

In this case we get:

```
  Normal   Scheduled  2m56s                 default-scheduler    Successfully assigned some-namespace/kube-apiserver-79b4b67ddb-7sjvg to someworker
  Normal   Pulled     2m54s                 kubelet, someworker  Container image "alpine-iptables:3.10.3" already present on machine
  Normal   Created    2m54s                 kubelet, someworker  Created container set-iptable-rules
  Normal   Started    2m54s                 kubelet, someworker  Started container set-iptable-rules
  Normal   Pulled     2m53s                 kubelet, someworker  Container image "hyperkube:v1.16.8" already present on machine
[...]

PR #2911, opened 2020-09-29
Set vpa minAllowed resource policy for gardener components

BeckerMax:
In some cases the vpa can downscale components below the bare minimum the process needs to function. This PR prevents the vpa from doing that for the gardener components.

Solves the same problem as:
https://github.com/gardener/gardener/pull/2848

/area robustness
/kind enhancement
/priority normal

### DIFF
--- a/charts/gardener/controlplane/charts/runtime/templates/admission-controller/vpa.yaml
+++ b/charts/gardener/controlplane/charts/runtime/templates/admission-controller/vpa.yaml
@@ -11,4 +11,10 @@ spec:
     name: gardener-admission-controller
   updatePolicy:
     updateMode: Auto
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        minAllowed:
+          cpu: 20m
+          memory: 25Mi
 {{- end }}
--- a/charts/gardener/controlplane/charts/runtime/templates/apiserver/vpa.yaml
+++ b/charts/gardener/controlplane/charts/runtime/templates/apiserver/vpa.yaml
@@ -18,5 +18,11 @@ spec:
     name: gardener-apiserver
   updatePolicy:
     updateMode: Auto
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        minAllowed:
+          cpu: 50m
+          memory: 256Mi
 {{- end }}
 {{- end }}
--- a/charts/gardener/controlplane/charts/runtime/templates/scheduler/vpa.yaml
+++ b/charts/gardener/controlplane/charts/runtime/templates/scheduler/vpa.yaml
@@ -11,4 +11,10 @@ spec:
     name: gardener-scheduler
   updatePolicy:
     updateMode: Auto
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        minAllowed:
+          cpu: 10m
+          memory: 25Mi
 {{- end }}
--- a/charts/gardener/gardenlet/charts/runtime/templates/vpa.yaml
+++ b/charts/gardener/gardenlet/charts/runtime/templates/vpa.yaml
@@ -11,4 +11,10 @@ spec:
     name: gardenlet
   updatePolicy:
     updateMode: Auto
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        minAllowed:
+          cpu: 50m
+          memory: 200Mi
 {{- end }}
--- a/charts/gardener/controlplane/charts/runtime/templates/controller-manager/vpa.yaml
+++ b/charts/gardener/controlplane/charts/runtime/templates/controller-manager/vpa.yaml
@@ -11,4 +11,10 @@ spec:
     name: gardener-controller-manager
   updatePolicy:
     updateMode: Auto
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        minAllowed:
+          cpu: 30m
+          memory: 300Mi
 {{- end }}


charts/gardener/controlplane/charts/runtime/templates/admission-controller/vpa.yaml
```diff
@@ -11,4 +11,11 @@ spec:
     name: gardener-admission-controller
   updatePolicy:
     updateMode: Auto
+  {{- if .Values.global.admission.resources.requests }}
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        minAllowed:
+  {{ toYaml .Values.global.admission.resources.requests | indent 10 }}
```
rfranzke:
```suggestion
{{ toYaml .Values.global.admission.resources.requests | indent 10 }}
```

charts/gardener/gardenlet/charts/runtime/templates/vpa.yaml
```diff
@@ -11,4 +11,11 @@ spec:
     name: gardenlet
   updatePolicy:
     updateMode: Auto
+  {{- if .Values.global.gardenlet.resources.requests }}
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        minAllowed:
+  {{ toYaml .Values.global.gardenlet.resources.requests | indent 10 }}
```
rfranzke:
```suggestion
{{ toYaml .Values.global.gardenlet.resources.requests | indent 10 }}
```

charts/gardener/controlplane/charts/runtime/templates/admission-controller/vpa.yaml
```diff
@@ -11,4 +11,11 @@ spec:
     name: gardener-admission-controller
   updatePolicy:
     updateMode: Auto
+  {{- if .Values.global.admission.resources.requests }}
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        minAllowed:
+{{ toYaml .Values.global.admission.resources.requests | indent 10 }}
```
timebertt: Can we use separate values for `minAllowed`? I think, you wouldn't use the same values for requests and lower bounds in some cases and the double meaning is confusing.
I would imagine the values structure to look something like this:
```yaml
resources:
  # unchanged ...
  requests:
    cpu: 50m
    memory: 50Mi
  limits:
    cpu: 300m
    memory: 256Mi
vpa:
  enabled: true # breaking change, but I think it makes sense
  minAllowed:
    cpu: 10m
    memory: 25Mi
```

charts/gardener/gardenlet/charts/runtime/templates/vpa.yaml
```diff
@@ -11,4 +11,11 @@ spec:
     name: gardenlet
   updatePolicy:
     updateMode: Auto
+  {{- if .Values.global.gardenlet.resources.requests }}
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        minAllowed:
+{{ toYaml .Values.global.gardenlet.resources.requests | indent 10 }}
```
timebertt: This is only applied for manually deployed gardenlets. Did you also observe problems with the vpa for gardenlets deployed by gardener?

timebertt (changes requested):
Setting `minAllowed` probably makes sense, I think.

> Hard coded minAllowed values would not help us here as the load varies so much from landscape to landscape

I'm curious, why `minAllowed` would have to be configured differently on larger gardener installations, as it's only the lower bound. Can you elaborate, why an operator would like to do so?

BeckerMax:
Thanks @timebertt for the comments. I think this issue needs some clarification:

Generally there seems to be two different problems we want to solve:
1. On very dynamic landscapes with high Shoot churn rates we see problems where the vpa can not react fast enough to load spikes. This could lead to the components under vpa reaching resource limits.
2. In some cases the vpa can downscale components below the bare minimum the process needs to function.

After further discussion we realized that `minAllowed` is made for problem **2.**.  Therefore, it should also not be landscape dependent.
This PR tried to solve problem 1. by configuring a high minAllowed to always have spare resources on the pods. But this kind of defeats the purpose of the vpa all together. Therefore the 1. problem can only be solved by the vpa itself (for example by increasing the steps memory/cpu is increased in very dynamic landscapes).

charts/gardener/controlplane/charts/runtime/templates/admission-controller/vpa.yaml
BeckerMax: Yes that would make more sense. But generally I would move away from landscape dependent configuration for `minAllowed` as discussed in the comment below.

BeckerMax:
I guess it makes still sense to configure minAllowed to protect the gardener components from the second problem (vpa can downscale components below the bare minimum the process needs to function). Though that could only happen on very quiet landscapes.

charts/gardener/gardenlet/charts/runtime/templates/vpa.yaml
BeckerMax: > This is only applied for manually deployed gardenlets

This change would be applied to all Gardenlets as the Gardenlet in the Soil uses its own helm chart to deploy the Gardenlets in the Seed.
cc @rfranzke

charts/gardener/controlplane/charts/runtime/templates/admission-controller/vpa.yaml
BeckerMax: @timebertt  As I now use hard-coded min values can you recheck the PR? thanks

charts/gardener/controlplane/charts/runtime/templates/apiserver/vpa.yaml
```diff
@@ -18,5 +18,11 @@ spec:
     name: gardener-apiserver
   updatePolicy:
     updateMode: Auto
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        minAllowed:
```
amshuman-kr, line 24: Don't we need a bit more minimum resources for for the gardener-apiserver?

amshuman-kr (commented):
Overall LGTM

charts/gardener/controlplane/charts/runtime/templates/scheduler/vpa.yaml
```diff
@@ -11,4 +11,10 @@ spec:
     name: gardener-scheduler
   updatePolicy:
     updateMode: Auto
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        minAllowed:
+          cpu: 10m
+          memory: 20Mi
```
rfranzke:
```suggestion
          memory: 25Mi
```

ref https://github.com/gardener/gardener/pull/2164

rfranzke (commented):
Should we set `50m` minimum CPU for all Gardener components to accommodate with restarts (even when the CPU usage goes down afterwards)?

charts/gardener/controlplane/charts/runtime/templates/apiserver/vpa.yaml
BeckerMax: For these values I checked the consumption on a relatively quiet landscape. These would be barely be sufficient. But I can gladly increase them to have a bit of room 50m / 256Mi.

charts/gardener/controlplane/charts/runtime/templates/scheduler/vpa.yaml
BeckerMax: You mean to take into account that after a restart the cpu usage will go up for a short period?

rfranzke: Yes. If the VPA scales down the scheduler to `10m` CPU limits then the scheduler might not be able to start anymore (that's what we've seen for some extension controllers).

BeckerMax: makes sense to me.

charts/gardener/controlplane/charts/runtime/templates/apiserver/vpa.yaml
BeckerMax: Increased it to 50/256 with latest commit.

amshuman-kr: For the regular `kube-apiserver` containers, `200Mi` was too less on a restart (https://github.com/gardener/gardener/pull/2174). I am not sure about the `gardener-apiserver`, though.

rfranzke: We could start with this value for now and improve it over the time based on our observations how well it works out.

charts/gardener/controlplane/charts/runtime/templates/scheduler/vpa.yaml
BeckerMax: I checked grafana again and especially after new instances got created. Seems like the currently configured CPU is enough for that  (and we do not need 50). I would suggest to start like this and adapt on the way if necessary. WDYT?

rfranzke (approved):
/lgtm

timebertt (approved):
Thanks for doing the background research!
/lgtm

PR #2916, opened 2020-09-29
Extension controllers determine codes on error

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal

Earlier, we were expecting that the various actuators return the error with the codes already. This required to adapt multiple places in the actuator's reconciliation/deletion function to properly determine them. With this PR, we do it at a single place which is more convenient, less error-prone and more robust.

**Release note**:

```improvement operator
NONE
```


ialidzhikov:
Rebase needed.

ialidzhikov (approved):
/lgtm

timebertt (approved):
/lgtm

PR #2919, opened 2020-09-29
Store coordination.k8s.io/leases in the etcd-events

vpnachev:

/area control-plane performance storage
/kind enhancement
/priority normal

Store coordination.k8s.io/leases in the etcd-events to reduce load on etcd-main. 

Fixes #

**Release note**:

```improvement operator
The `coordination.k8s.io/leases` resources are now stored in the `etcd-events`. With this change, the db of `etcd-main` will be less often updated, hence less snapshots will be taken.
```

### DIFF
--- a/charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
+++ b/charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
@@ -123,7 +123,7 @@ spec:
         - --etcd-certfile=/srv/kubernetes/etcd/client/tls.crt
         - --etcd-keyfile=/srv/kubernetes/etcd/client/tls.key
         - --etcd-servers=https://etcd-main-client:{{ .Values.etcdServicePort }}
-        - --etcd-servers-overrides=/events#https://etcd-events-client:{{ .Values.etcdServicePort }}
+        - --etcd-servers-overrides=/events#https://etcd-events-client:{{ .Values.etcdServicePort }},coordination.k8s.io/leases#https://etcd-events-client:{{ .Values.etcdServicePort }}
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm
/second-opinion
/invite @mvladev

timebertt:
What will happen to leases that are already stored in etcd-main? I guess, they won't get cleaned up, right?

vpnachev:
> What will happen to leases that are already stored in etcd-main? I guess, they won't get cleaned up, right?

Yes, they will be orphaned there. One option that comes into my mind is to run a clean-up command against etcd-main
```bash
ETCDCTL_API=3 etcdctl --endpoints=etcd-main-client:2379 del --prefix /registry/leases
```

However, this cannot be done from this repository now, at least not so easy.

rfranzke:
Well, what about simply keeping them orphaned for existing clusters? They are small anyways, and they don't cause any trouble.

ialidzhikov:
One minor nit - the name etcd-events will be misleading as the etcd will be no longer used only for events.

timebertt (approved):
Thanks!
/lgtm

mvladev:
The problem is that it causes perceived data-loss. It might cause problems for end-users using the lease objects for some reasons other than healtchecks.

rfranzke:
Hm, that's a valid point (although, I'm not sure how likely this is). I can image three ways to proceed now:
1. Only store the `Lease`s in etcd-events for shooted seeds
1. Do it for all shoots, but put it behind a feature gate
1. Make it configurable per shoot

Opinions?

rfranzke:
Shall we close this PR for now until we clarified how to proceed?

vpnachev:
Yes.
/close

Issue #2915, opened 2020-09-29
[Flaky test] Extensions Controller BackupBucket Test Suite


timebertt:
**How to categorize this issue?**

/area testing
/kind flake
/priority critical

**What happened**:

`Extensions Controller BackupBucket Test Suite` (BackupBucket controller integration test) is flaky with a high rate:
- https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-master/jobs/master-pull-request-job/builds/9
- https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-master/jobs/master-pull-request-job/builds/10
- https://concourse.ci.gardener.cloud/builds/4168

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

Not sure yet. The flake seems to be related to cpu pressure on the CI machines, as it fails at different steps of the test with `context deadline exceeded` errors or errors indicating, that the reconciler couldn't act.

I tried to reproduce it locally, no luck so far.

**Anything else we need to know?**:

**Environment**:

- Gardener version: v1.11.0-dev
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

timebertt:
After half a day trying to reproduce it, I'm stuck with this.
I tried reproducing it
- in a cpu restricted docker container
- with [stress](https://godoc.org/golang.org/x/tools/cmd/stress)
- with stress in a cpu restricted docker container
- with different combinations of timeout settings, resource limitations, parallel executions...

But so far, no luck. With heavy cpu limits / high parallelism, the test times out starting the kube-apiserver:
```
Running Suite: Extensions Controller BackupBucket Test Suite
============================================================
Random Seed: 1601388879
Will run 2 of 2 specs

STEP: starting test environment
2020-09-29T14:14:39.507Z	DEBUG	controller-runtime.test-env	starting control plane	{"api server flags": []}
2020-09-29T14:15:02.091Z	ERROR	controller-runtime.test-env	unable to start the controlplane	{"tries": 0, "error": "timeout waiting for process kube-apiserver to start"}
github.com/go-logr/zapr.(*zapLogger).Error
	/Users/timebertt/go/src/github.com/gardener/gardener/vendor/github.com/go-logr/zapr/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/envtest.(*Environment).startControlPlane
[...]

PR #2925, opened 2020-09-30
Gardenlet does not fail liveness probe when configured seed is deleted

danielfoehrKn:

/area usability
/priority normal

Currently the Gardenlet does fail its liveness probe when its configured seed  (Component config config.SeedConfig.name) is deleted.
This is problematic as this leads to the termination of the gardenlet pod. 
When the Gardenlet starts up, it automatically re-registers the Seed.

This cycle is causing problems to delete the Seed.

Please consider that this introduces an additional request against the Garden cluster.

**Additionally**:
 I wanted to bring up again the topic of cleanup after Seed deletion.
Currently we do not cleanup the Garden namespace of deleted Seed clusters.
The argument is that we do not know what is deployed in the garden namespace. 
This is true for regular Seed clusters. 
Though for shooted seed clusters I would argue that we can safely delete the garden ns.
This allows to safely add and remove the shooted-seed annotation without having a gardenlet deployment (& other components) lingering in the garden namespace after Seed deletion. 
Advantages
 - no manual cleaning required
-  If that idling gardenlet  pod is restarted due to any reason (e.g **failing liveness probes**) it re-registers the unwanted Seed.

Fixes #

- uses the regular client that could have a cache. How quickly is the client cache synced? 
Problem could be that the cache still contains the already deleted seed but the lease is already gone -> the health check fails.
- does not terminate when connection to the Garden cluster is not possible. This is to avoid failing livenessProbes due to intermitted connection failures. Restarts are costly in terms of API requests against the Garden API server and can have unwanted side effects such as unwanted Seed registrations. Would appreciate opinions here.

**How to reproduce**:
 - Create a shooted seed (with no Shoots on them) 
 - Delete the seed resource
 - Observe that the Gardenlet in the Seed accepts the Seed deletion
 - Observe that the Gardenlet pod in the Seed is killed due to failing liveness probe
 - Observe that the seed is re-registered

**Release note**:

```noteworthy operator
The gardenlet does not fail the liveness probe when the seed resource configured in the Gardenlet's config has been deleted. 
```

### DIFF
--- a/pkg/gardenlet/controller/seed/seed.go
+++ b/pkg/gardenlet/controller/seed/seed.go
@@ -20,7 +20,16 @@ import (
 	"sync"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	kubeinformers "k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
@@ -37,15 +46,6 @@ import (
 	"github.com/gardener/gardener/pkg/logger"
 	"github.com/gardener/gardener/pkg/utils"
 	"github.com/gardener/gardener/pkg/utils/imagevector"
-
-	"github.com/prometheus/client_golang/prometheus"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	kubeinformers "k8s.io/client-go/informers"
-	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/tools/record"
-	"k8s.io/client-go/util/workqueue"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 // Controller controls Seeds.
@@ -242,16 +242,26 @@ func (c *Controller) startHealthManagement(ctx context.Context) {
 		select {
 		case <-ctx.Done():
 			return
-		default:
-			time.Sleep(LeaseResyncGracePeriodSeconds / 2 * time.Second)
+		case <-time.After(LeaseResyncGracePeriodSeconds / 2 * time.Second):
 
-			health := true
+			isHealthy := true
 
-			if seedName == "" {
+			if len(seedName) > 0 {
+				if _, err := c.k8sGardenCoreInformers.Core().V1beta1().Seeds().Lister().Get(seedName); err != nil {
+					if apierrors.IsNotFound(err) {
+						// the Seed configured for the Gardenlet does not exist.
+						// Do not expect an existing lease
+						expectedHealthReports = 0
+					} else {
+						logger.Logger.Errorf("error when getting the seed %q for health management: %+v", seedName, err)
+						isHealthy = false
+					}
+				}
+			} else {
 				seedList, err := c.k8sGardenCoreInformers.Core().V1beta1().Seeds().Lister().List(seedLabelSelector)
 				if err != nil {
 					logger.Logger.Errorf("error while listing seeds for health management: %+v", err)
-					health = false
+					isHealthy = false
 				}
 				expectedHealthReports = len(seedList)
 			}
@@ -259,11 +269,11 @@ func (c *Controller) startHealthManagement(ctx context.Context) {
 			c.lock.Lock()
 
 			if len(c.leaseMap) != expectedHealthReports {
-				health = false
+				isHealthy = false
 			} else {
 				for _, status := range c.leaseMap {
 					if !status {
-						health = false
+						isHealthy = false
 						break
 					}
 				}
@@ -271,7 +281,7 @@ func (c *Controller) startHealthManagement(ctx context.Context) {
 
 			c.leaseMap = make(map[string]bool)
 			c.lock.Unlock()
-			c.healthManager.Set(health)
+			c.healthManager.Set(isHealthy)
 		}
 	}
 }


rfranzke:
Hm, not sure about this PR. What is the kubelet doing when you delete the `Node` and it restarts? Wouldn't the "right way" be to deprovision the gardenlet if the seed is being deleted? Also, this PR doesn't solve the problem but just works around it  (e.g., when the gardenlet pod restarts for whatever other reason (eviction, VPA, ...), it registers the seed again).
Can you share more steps for reproduction?

> Please consider that this introduces an additional request against the Garden cluster.

How often is this executed?

> I wanted to bring up again the topic of cleanup after Seed deletion.
Currently we do not cleanup the Garden namespace of deleted Seed clusters.
The argument is that we do not know what is deployed in the garden namespace.
This is true for regular Seed clusters.
Though for shooted seed clusters I would argue that we can safely delete the garden ns.

Yes, the plan was to tackle this with the refactorings planned by @timebertt to use `ManagedResource`s for the bootstrapping components in the seed. Generally, I think we can always delete the garden namespace (even for non-shooted seeds), because it's considered Gardener-system-relevant and should be owned by Gardener itself.

danielfoehrKn:
> What is the kubelet doing when you delete the Node and it restarts

Why would the kubelet restart? When deleting the Node, the cloud controller deletes the VM from the CloudProvider. 
So unlike for the gardenlet, the kubelet will always get killed eventually during Node deletion - it cannot try  to register the Node again.

> Wouldn't the "right way" be to deprovision the gardenlet if the seed is being deleted

Yes I think so. I also added the `topic of cleanup after Seed deletion` above. 
But just letting the health checks fail and causing the process to be killed is not the right way in my opinion.
I would have to check, but I do not think that the kubelet just terminates itself in such cases. 
I think that the kubelet is build to be used in a systemd service with `restart=always` policy.

Regarding `topic of cleanup after Seed deletion`:  we do not know where the Gardenlet runs so deletion vai `garden` ns in the Seed cluster does not 100 % work. Gardenlet can run anywhere not only in cluster in the garden ns.
This is why the Gardenlet should not be killed and consequently try to re-register the seed.

> I think we can always delete the garden namespace (even for non-shooted seeds), because it's considered Gardener-system-relevant and should be owned by Gardener itself.

What about when registering the Garden cluster as Seed - then ns deletion does not work.

timebertt:
@danielfoehrKn What's the plan with this PR?

danielfoehrKn:
> What's the plan with this PR?

Waiting for more opinions - I think I stated why I think this PR makes sense but we are not on the same page yet.

timebertt:
Some thoughts from my side:

> this PR doesn't solve the problem but just works around it (e.g., when the gardenlet pod restarts for whatever other reason (eviction, VPA, ...), it registers the seed again).

I agree with @rfranzke. The liveness probe killing the gardenlet is only part of the problem and we can't solve it by removing all causes of termination (e.g. eviction during node roll, VPA and so on).
The solution (FMPOV) should rather be:
- either the gardenlet should be automatically removed once the Seed has been deleted
  I guess, we can/should only do this for shooted seeds. If you have deployed the gardenlet manually you will have to clean it up manually, too.
- or if the gardenlet keeps running / is restarted after seed deletion, it should not (be able to) re-register itself
  Though, I am wondering, why this isn't already the case, as bootstrap-tokens for the gardenlet bootstrap are only valid for 24h and should get cleaned up by kcm after expiry (given that the `tokencleaner` controller is enabled)
Probably these options are not exclusive and it makes sense to implement them both.

WDYT?

rfranzke:
> either the gardenlet should be automatically removed once the Seed has been deleted
I guess, we can/should only do this for shooted seeds. If you have deployed the gardenlet manually you will have to clean it up manually, too.

Isn't this implemented like this already? If not, what's missing?

> or if the gardenlet keeps running / is restarted after seed deletion, it should not (be able to) re-register itself
Though, I am wondering, why this isn't already the case, as bootstrap-tokens for the gardenlet bootstrap are only valid for 24h and should get cleaned up by kcm after expiry (given that the tokencleaner controller is enabled)
Probably these options are not exclusive and it makes sense to implement them both.

The bootstrap token is only used once to generate a real kubeconfig (similar to TLS bootstrapping of the kubelet, see https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/). From then on it doesn't use the bootstrap token again. It can use its real kubeconfig until the client certificate expires (usually after `1y`).
Anyways, how would we track that a seed was deleted and that a particular gardenlet is not supposed to re-register it? In some cases we might even want to have this behaviour. Again, I think it makes sense to stay close to the kubelet implementation here and rather follow the first proposal.

danielfoehrKn:
> this PR doesn't solve the problem but just works around it (e.g., when the gardenlet pod restarts for whatever other reason (eviction, VPA, ...), it registers the seed again).

Yes we cannot prevent the Gardenlet from restarting / getting killed in all cases.
Same for the kubelet.
 
I think what I am proposing here is not a workaround, but rather a fix for the most common problem when deleting a Seed that is manually registered (not a Shooted Seed).

> Again, I think it makes sense to stay close to the kubelet implementation here and rather follow the first proposal.

I also checked the kubelet behaviour if the Node object is deleted.

Observations
- the running kubelet does not self-terminate or attempts to re-register the Node. 
- Only when deleting the kubelet process, the node is re-registered.

Otherwise in my POV they would have the same problem as we have with the Gardenlet. It would be hard to remove a Node manually as it would right away register itself again.

This supports my point of view to not terminate the Gardenlet pod.

pkg/gardenlet/controller/seed/seed.go
```diff
@@ -246,33 +248,53 @@ func (c *Controller) startHealthManagement(ctx context.Context) {
...
+			if len(seedName) > 0 {
+				gardenClient, err := c.clientMap.GetClient(ctx, keys.ForGarden())
+				if err != nil {
+					logger.Logger.Errorf("error when getting the garden client for health management: %+v", err)
+					// gardenlet should not fail the liveness probe and be terminate due to a connection timeout
+					// this could have consequences such as re-registering an already deleted seed
+					return
+				}
+				seed := &gardencorev1beta1.Seed{}
+				if err := gardenClient.Client().Get(ctx, kutil.Key(seedName), seed); err != nil {
```
rfranzke: As discussed in person, please exchange this line by using the `c.k8sGardenCoreInformers`.

rfranzke (changes requested):
I thought about it again and I also don't have a better idea now to get rid of the problem. If the liveness probe fails then the gardenlet is restarted immediately and re-registers it. You must be "fast" and delete the gardenlet shortly after the `Seed` was deleted, but this is no good solution.
Hence, let's proceed for now as you suggest.

rfranzke (dismissed):
/lgtm

danielfoehrKn:
There are races in the extension library tests where this PR does not change any code..

rfranzke:
Yes, see #2915.
Anyways, you need to
/rebase

timebertt:
> > either the gardenlet should be automatically removed once the Seed has been deleted
> > I guess, we can/should only do this for shooted seeds. If you have deployed the gardenlet manually you will have to clean it up manually, too.
> 
> Isn't this implemented like this already? If not, what's missing?

Ah, yes. It is done, when you remove the `use-as-seed` annotation or delete the Shoot itself but not when you delete the Seed, mixed that up.

> > or if the gardenlet keeps running / is restarted after seed deletion, it should not (be able to) re-register itself
> > Though, I am wondering, why this isn't already the case, as bootstrap-tokens for the gardenlet bootstrap are only valid for 24h and should get cleaned up by kcm after expiry (given that the tokencleaner controller is enabled)
> > Probably these options are not exclusive and it makes sense to implement them both.
> 
> The bootstrap token is only used once to generate a real kubeconfig (similar to TLS bootstrapping of the kubelet, see https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/). From then on it doesn't use the bootstrap token again. It can use its real kubeconfig until the client certificate expires (usually after `1y`).

Hmm, yes. I thought, it would only use the bootstrap kubeconfig / service account to register itself, but that isn't the case.

timebertt:
> I also don't have a better idea now to get rid of the problem

What about deleting the garden kubeconfig as part of the Seed deletion flow? Then the gardenlet won't be able to re-register again.

> I think what I am proposing here is not a workaround, but rather a fix for the most common problem when deleting a Seed that is manually registered (not a Shooted Seed).

Yeah, got it. But what do you think about the proposal above? It wouldn't only solve the most common problem but also eviction, VPA and so on. right?

I would prefer this approach, as I think having a failing liveness/readiness probe, once the Seed has been deleted, is somewhat conceptually correct.

rfranzke:
@timebertt this approach does not always work either as it is not mandatory to use a bootstrap token out of which the gardenlet self-generates a certificate. You can also deploy it with a precreated kubeconfig that you mount as secret into it, and then there is nothing to delete.

> I would prefer this approach, as I think having a failing liveness/readiness probe, once the Seed has been deleted, is somewhat conceptually correct.

I kind of have the same feeling, but OTOH, I still don't see a better solution for now. :(

/cc @danielfoehrKn

timebertt:
> @timebertt this approach does not always work either as it is not mandatory to use a bootstrap token out of which the gardenlet self-generates a certificate. You can also deploy it with a precreated kubeconfig that you mount as secret into it, and then there is nothing to delete.

Yeah, right 😞 

> I think having a failing liveness/readiness probe, once the Seed has been deleted, is somewhat conceptually correct.

I also thought about this again in some more detail. And now I would say, that indeed the readiness probe should fail for a Gardenlet with deleted Seed. But the liveness probe actually should only check if the container itself is still alive, which it would be in this case, it just doesn't act on anything, because it should not do so. But in fact it is alive and functioning and isn't stuck in any way.
So reiterating over it, I think the conceptually correct behaviour would be, that the readiness probe fails (like it is today), but the liveness probe succeeds (like you propose in this PR).
So let's go ahead with this change :)

timebertt:
Still, it doesn't solve the problem with VPA and so on (which we probably can neglect for the time being, as the liveness probes are the most common problem, as you mentioned).

Nevertheless, one last idea about how we could detect, if the Seed has already been registered once:
On startup we could check, if the cluster has the `cluster-identity` ConfigMap and block the Seed registration if that's the case.
This would cover the different kubeconfig options (static, bootstrap token, serviceaccount token), would be easy to detect and wouldn't be cleaned up when the garden namespace is deleted.
Yes, we will have to move the cluster-identity out of the seed-bootstrap chart, once we clean this up, but I would argue, that this is anyways the correct way to deploy it.

WDYT? (Doesn't have to be tackled in this PR)

danielfoehrKn:
Just talked to @timebertt - the problem with checking for the existing of the `cluster-identity` config map is, that every Shoot cluster already has a `cluster-identity` config map. 
Whenever we deploy a Gardenlet on a Shoot cluster, the gardenlet would not registering it as a Seed because the config map already exists.
Consequently this breaks the shooted seed use case.

Is there would be a fail-safe way to distinguish a Seed cluster identity vs a Shoot cluster identity? @swilen-iwanow 
Existing identity config maps are not overridden -[ Shooted seeds have Shoot cluster-identities](https://github.com/gardener/gardener/blob/f762495175c09ab12f13ea3f3c714ddd897c38ad/pkg/operation/seed/seed.go#L828).
For regular Seeds (not shooted seeds) the Gardenlet would not register again if the Seed resource does not exist yet.
For Shooted Seeds it would always try to register (no change).

danielfoehrKn:
In any case, I think preventing the re-registering of a Seed resource is a different issue than preventing to fail the liveness probe.
It could be discussed in an issue and handled with another PR.
WDYT

rfranzke:
> In any case, I think preventing the re-registering of a Seed resource is a different issue than preventing to fail the liveness probe.
It could be discussed in an issue and handled with another PR.

+1

danielfoehrKn:
Will work on this either today or tomorrow.

rfranzke:
ping @danielfoehrKn

danielfoehrKn:
I retested the PR. It is ready from my side. Thanks for your patience.

pkg/gardenlet/controller/seed/seed.go
```diff
@@ -245,33 +246,46 @@ func (c *Controller) startHealthManagement(ctx context.Context) {
...
 
-			if seedName == "" {
+			if len(seedName) > 0 {
+				if _, err := c.k8sGardenCoreInformers.Core().V1beta1().Seeds().Lister().Get(seedName); err != nil {
+					if !apierrors.IsNotFound(err) {
+						logger.Logger.Errorf("error when getting the seed %q for health management: %+v", seedName, err)
+						// gardenlet should not fail the liveness probe and be terminate due to a connection timeout
+						// this could have consequences such as re-registering an already deleted seed or unnecessary Gardenlet restarts
+						// leading to many network requests to fill the caches.
+						return
```
rfranzke: Hm, in this case, an unexpected error has occurred, so shouldn't the health be set to `false` here?

timebertt:
@danielfoehrKn can you add a release note?
/kind enhancement

pkg/gardenlet/controller/seed/seed.go
danielfoehrKn: Yes, thanks for paying attention. Otherwise, the health management for Gardenlets configured for one seed would not make sense because it never fails.

pkg/gardenlet/controller/seed/seed.go
```diff
@@ -245,33 +246,46 @@ func (c *Controller) startHealthManagement(ctx context.Context) {
 		default:
 			time.Sleep(LeaseResyncGracePeriodSeconds / 2 * time.Second)
```
timebertt: As you're already on it, can you change this to a non-blocking case?
```suggestion
		case <-time.After(LeaseResyncGracePeriodSeconds / 2 * time.Second):
```

rfranzke (dismissed):
/lgtm

rfranzke (dismissed):

rfranzke:
@danielfoehrKn can you please rebase to `master` to fix the `verify` step?

pkg/gardenlet/controller/seed/seed.go
```diff
@@ -224,54 +226,76 @@ func (c *Controller) Run(ctx context.Context, workers int) {
...
-
-			if seedName == "" {
-				seedList, err := c.k8sGardenCoreInformers.Core().V1beta1().Seeds().Lister().List(seedLabelSelector)
+		case <-time.After(LeaseResyncGracePeriodSeconds / 2 * time.Second):
+
+			isHealthy := true
+
+			if len(seedName) > 0 {
+				seed := &gardencorev1beta1.Seed{}
+				if err := gardenClient.DirectClient().Get(ctx, kutil.Key(seedName), seed); err != nil {
```
rfranzke: Maybe I missed this earlier, but can we please continue to use the cache to avoid so many "real" calls?

danielfoehrKn:
Thanks for the reviews.

I found another issue that I missed during my tests. 
When using the lister to query the seeds, interrupted network connections are not detected as getting the seeds from the lister continuous to work (tested it for > 5 min). This is true whether the Gardenlet uses cachedRuntimeClients or not.
Same when using the cached runtime client without the lister.

Therefore I changed it to use the direct client. Otherwise, I do not see the point in having this code at all, as it will not throw an error.
I propose to decrease the ping interval from 10s -> 20s to save network traffic.

PTAL if that looks reasonable
@rfranzke @timebertt

pkg/gardenlet/controller/seed/seed.go
danielfoehrKn: I just wrote a comment, too fast :)

danielfoehrKn:
Talked to @rfranzke and decided to only include the fix that the Gardenlet does not fail the liveness probe when configured seed is deleted.
Will open a dedicated PR for the other changes.

rfranzke (approved):
/lgtm

timebertt:
> @danielfoehrKn can you add a release note?

ping

timebertt (approved):
/lgtm

PR #2927, opened 2020-09-30
Fix min/maxReplica setting for Seed vpa

BeckerMax:
This change fixes a bug that when hvpa is enabled the values
apiServer.autoscaler.minReplicas and maxReplicas are not respected
for shooted Seeds.

/area auto-scaling
/area control-plane

/kind bug
/priority normal

Again respecting min and maxReplicas for Seed API Server. Currently, system critical components like the Seed API server can be scaled down to 1 instance as minReplicas is not respected

### DIFF
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -995,15 +995,19 @@ func (b *Botanist) DeployKubeAPIServer(ctx context.Context) error {
 		foundDeployment = false
 	}
 
-	if b.ShootedSeed != nil && !hvpaEnabled {
+	if b.ShootedSeed != nil {
 		var (
 			apiServer  = b.ShootedSeed.APIServer
 			autoscaler = apiServer.Autoscaler
 		)
-		defaultValues["replicas"] = *apiServer.Replicas
 		minReplicas = *autoscaler.MinReplicas
 		maxReplicas = autoscaler.MaxReplicas
+	}
 
+	if b.ShootedSeed != nil && !hvpaEnabled {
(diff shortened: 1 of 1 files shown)


pkg/operation/botanist/controlplane.go
```diff
@@ -995,15 +995,19 @@ func (b *Botanist) DeployKubeAPIServer(ctx context.Context) error {
...
 		)
-		defaultValues["replicas"] = *apiServer.Replicas
 		minReplicas = *autoscaler.MinReplicas
 		maxReplicas = autoscaler.MaxReplicas
+	}
 
+	if b.ShootedSeed != nil && !hvpaEnabled {
+		apiServer := b.ShootedSeed.APIServer
+
+		defaultValues["replicas"] = *apiServer.Replicas
```
rfranzke, line 1010:
```suggestion

		if !hvpaEnabled {
			defaultValues["replicas"] = *apiServer.Replicas
```

You can have the `if b.ShootedSeed != nil` only once

ggaurav10 (approved):

pkg/operation/botanist/controlplane.go
BeckerMax: No, then the else below changes its meaning. The else also needs to respect the case when ShootedSeed == nil (and hvpa whatever).

BeckerMax: The else on && connected ifs are super confusing I think btw.

rfranzke (approved):
/lgtm

PR #2931, opened 2020-09-30
improve Seed deletion with configured Backup Bucket

danielfoehrKn:

/area usability
/priority normal

Currently an admission plugin blocks seed deletion if there is still a backup Bucket referencing it.
When manually deleting it, it will be recreated during the next Seed reconcile.
You have to be "quick" to delete the Seed resource before another reconciliation kicks in that creates the BB again.
This in combination with #2925 makes it hard to delete a Seed cluster.

- Allow setting a deletion timestamp on seeds with referencing BB (@vpnachev )
- Do not delete the BB during the Seed reconciliation
- Do not remove finalizer during Seed reconcile if still referenced by Shoot or BB

The same goal is reached: do not allow Seed deletion if still referenced by Shoot or BB.
However by setting the deletion timestamp on the Seed
 - the controllerinstallations are removed
 - the BB is not deployed again

**How to reproduce**:
 - Create a Seed with backup enable that has no deployed Shoots
 - Delete the Bucket manually
 - Observe that the Backup is recreated 
 - Seed cannot be deleted again (circular)

Fixes #

@vpnachev could you check if that makes sense to you ? 
**Release note**:

```noteworthy operator
Deletion of a Seed is now possible with an existing Backup Bucket (but having no Shoots deployed!). The Bucket is deleted automatically during the Seed reconciliation flow. 
```

### DIFF
--- a/plugin/pkg/seed/validator/admission.go
+++ b/plugin/pkg/seed/validator/admission.go
@@ -46,10 +46,9 @@ func Register(plugins *admission.Plugins) {
 // ValidateSeed contains listers and and admission handler.
 type ValidateSeed struct {
 	*admission.Handler
-	seedLister         corelisters.SeedLister
-	shootLister        corelisters.ShootLister
-	backupBucketLister corelisters.BackupBucketLister
-	readyFunc          admission.ReadyFunc
+	seedLister  corelisters.SeedLister
+	shootLister corelisters.ShootLister
+	readyFunc   admission.ReadyFunc
 }
 
 var (
@@ -80,7 +79,6 @@ func (v *ValidateSeed) SetInternalCoreInformerFactory(f coreinformers.SharedInfo
 	v.shootLister = shootInformer.Lister()
 
 	backupBucketInformer := f.Core().InternalVersion().BackupBuckets()
-	v.backupBucketLister = backupBucketInformer.Lister()
 
 	readyFuncs = append(readyFuncs, seedInformer.Informer().HasSynced, shootInformer.Informer().HasSynced, backupBucketInformer.Informer().HasSynced)
 }
@@ -92,9 +90,6 @@ func (v *ValidateSeed) ValidateInitialization() error {
 	}
 	if v.shootLister == nil {
 		return errors.New("missing shoot lister")
-	}
-	if v.backupBucketLister == nil {
-		return errors.New("missing backupbucket lister")
 	}
 	return nil
 }
@@ -135,17 +130,8 @@ func (v *ValidateSeed) Validate(ctx context.Context, a admission.Attributes, o a
 		return apierrors.NewInternalError(err)
 	}
 
-	backupbuckets, err := v.backupBucketLister.List(labels.Everything())
-	if err != nil {
-		return apierrors.NewInternalError(err)
-	}
-
 	if admissionutils.IsSeedUsedByShoot(seedName, shoots) {
 		return admission.NewForbidden(a, fmt.Errorf("cannot delete seed '%s' which is still used by shoot(s)", seedName))
-	}
-
-	if admissionutils.IsSeedUsedByBackupBucket(seedName, backupbuckets) {
-		return admission.NewForbidden(a, fmt.Errorf("cannot delete seed '%s' which is still used by backupbucket(s)", seedName))
 	}
 
 	return nil
--- a/plugin/pkg/seed/validator/admission_test.go
+++ b/plugin/pkg/seed/validator/admission_test.go
@@ -98,15 +98,14 @@ var _ = Describe("validator", func() {
 				Expect(err).To(BeForbiddenError())
 			})
 
-			It("should disallow seed deletion because it is still referenced by a backupbucket", func() {
+			It("should allow seed deletion even though it is still referenced by a backupbucket (will be cleaned up during Seed reconciliation)", func() {
 				backupBucket.Spec.SeedName = &seedName
 				Expect(coreInformerFactory.Core().InternalVersion().BackupBuckets().Informer().GetStore().Add(&backupBucket)).To(Succeed())
 				attrs := admission.NewAttributesRecord(&seed, nil, core.Kind("Seed").WithVersion("version"), "", seed.Name, core.Resource("seeds").WithVersion("version"), "", admission.Delete, &metav1.DeleteOptions{}, false, nil)
 
 				err := admissionHandler.Validate(context.TODO(), attrs, nil)
 
-				Expect(err).To(HaveOccurred())
-				Expect(err).To(BeForbiddenError())
+				Expect(err).ToNot(HaveOccurred())
 			})
 
 			It("should disallow seed deletion because shoot migration is yet not finished", func() {
--- a/plugin/pkg/shoot/validator/admission_test.go
+++ b/plugin/pkg/shoot/validator/admission_test.go
@@ -543,7 +543,7 @@ var _ = Describe("validator", func() {
 				Expect(err).To(BeForbiddenError())
 			})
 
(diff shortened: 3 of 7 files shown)


pkg/gardenlet/controller/seed/seed_control.go
```diff
@@ -184,13 +184,6 @@ func (c *defaultControl) ReconcileSeed(obj *gardencorev1beta1.Seed, key string)
 			return nil
 		}
 
-		if seed.Spec.Backup != nil {
-			if err := deleteBackupBucketInGarden(ctx, gardenClient.Client(), seed); err != nil {
-				seedLogger.Error(err.Error())
-				return err
-			}
-		}
```
rfranzke: I don't understand why the `BackupBucket` is no longer deleted here when the seed's deletion timestamp is set. Why was this code removed / how is the resource deleted now?

danielfoehrKn: BB deletion  has to be done manually. 

Before Seed deletion was blocked by the admission plugin if a BB existed. So this code did not make sense in my opinion as there can never be a BB . 

With this PR the Seed deletion (setting deletion timestamp) is not blocked any more when there is a BB (Seed has a deletion timestamp). The Seed finalizer though is not removed when there is still an existing BB.
This is why this code has to be removed, otherwise when deleting a Seed with BB, the BB will be deleted automatically.

rfranzke: There should be nothing that needs to be done "manually" in Gardener, at least not on purpose. I agree that it should be allowed to set the deletion timestamp on the `Seed` even if there is still a `BackupBucket`, but I still don't get why we have to remove the automatic `BackupBucket` deletion in the seed controller once the object has a deletion timestamp.

danielfoehrKn: I think the reasoning was to protect existing Shoot backups from accidental and indirect deletion via the Seed. Right @vpnachev ? 
Is there any value in retaining Backup entries of already deleted Shoots?

rfranzke: I can't follow. My question was:

> but I still don't get why we have to remove the automatic BackupBucket deletion in the seed controller once the `Seed` object has a deletion timestamp.

Again: Why did you remove this code?

> Is there any value in retaining Backup entries of already deleted Shoots?

Not sure how this is related now, but yes, sometimes operators want to keep the shoot's etcd backups even after the shoot has been deleted (for some grace period, maybe 2 weeks or so).

danielfoehrKn: > Again: Why did you remove this code?

Otherwise this PR introduces a  change of behaviour towards BB deletion.
If this code is removed, a BB is deleted automatically when a Seed has a deletion timestamp.
Previously you had to manually delete the BB before the Seed deletion was allowed by the AdmissionPlugin.

danielfoehrKn: Now that I established why I deleted the code, the other question is, if we want that manual BB deletion step in the first place. 
And because @vpnachev introduced the admission plugin validation, I wanted to ask him what his original reason was.

rfranzke: > If this code is removed, a BB is deleted automatically when a Seed has a deletion timestamp.
Previously you had to manually delete the BB before the Seed deletion was allowed by the AdmissionPlugin.

Well, this is how it was before @vpnachev introduced the validation. IMO, this is completely fine. I suggest to add this code again and only remove the validation.

> if we want that manual BB deletion step in the first place.

See above: If we can avoid then we don't want any manual step at all anywhere.

danielfoehrKn: > See above: If we can avoid then we don't want any manual step at all anywhere.

I think that depends on the following:
> Is there any value in retaining Backup entries (&Bucket) of already deleted Shoots?

Yes automation is great, but  before I remove it, I want to understand why @vpnachev introduced the safeguard that the BB has to be manually deleted. 
We could also automate all Shoot cluster deletion when the Seed resource is deleted (without checking for existing Shoot clusters) - but we choose to not do it for safety reasons right?

vpnachev: Excuse me for the delay.

To be honest, after some time I think we should remove the validation for `BackupBuckets` and let the seed controller delete them.

danielfoehrKn: Alright, thanks. 

> remove the validation for BackupBuckets and let the seed controller delete them.

done. PTAL

rfranzke (dismissed):
/lgtm

rfranzke:
Can you please
/rebase

vpnachev (dismissed):
/lgtm

plugin/pkg/seed/validator/admission.go
```diff
@@ -135,18 +135,9 @@ func (v *ValidateSeed) Validate(ctx context.Context, a admission.Attributes, o a
 		return apierrors.NewInternalError(err)
 	}
 
-	backupbuckets, err := v.backupBucketLister.List(labels.Everything())
```
vpnachev, line 138: Please, remove the backup bucket lister related code from lines 51, 82, 96.

plugin/pkg/seed/validator/admission.go
```diff
@@ -135,18 +135,9 @@ func (v *ValidateSeed) Validate(ctx context.Context, a admission.Attributes, o a
...
-	if err != nil {
-		return apierrors.NewInternalError(err)
-	}
-
 	if admissionutils.IsSeedUsedByShoot(seedName, shoots) {
 		return admission.NewForbidden(a, fmt.Errorf("cannot delete seed '%s' which is still used by shoot(s)", seedName))
 	}
 
-	if admissionutils.IsSeedUsedByBackupBucket(seedName, backupbuckets) {
```
vpnachev, line 147: You can completely remove this method - e.g. delete https://github.com/gardener/gardener/blob/e0a0b192fbdc1a3c4329f4a702d4694d9c3d5437/plugin/pkg/utils/miscellaneous.go#L62-L70.

vpnachev (commented):
The documentation [here](https://github.com/gardener/gardener/blob/master/docs/concepts/apiserver_admission_plugins.md#seedvalidator) needs to be updated, too.

You can also remove the backup validation for shoot seeds.

plugin/pkg/seed/validator/admission_test.go
```diff
@@ -98,17 +89,6 @@ var _ = Describe("validator", func() {
 				Expect(err).To(BeForbiddenError())
 			})
 
-			It("should disallow seed deletion because it is still referenced by a backupbucket", func() {
```
timebertt, line 101: Can you keep this test case and rather invert it? E.g.:
`should allow seed deletion if it is referenced by BackupBucket (will get cleaned up after Seed deletion)`

Though, it wouldn't really test concrete part of the implementation now, I think it still makes sense to have such tests, that will pop up when someone in the future might want to change the logic. Then they at least have to change the test, but it wouldn't go unnoticed, that there was a concrete reason for doing it the way it is currently implemented.

plugin/pkg/shoot/validator/admission_test.go
```diff
@@ -542,54 +542,6 @@ var _ = Describe("validator", func() {
 				Expect(err).To(HaveOccurred())
 				Expect(err).To(BeForbiddenError())
 			})
-
-			It("should reject removal of the annotation when the respective seed is used by backupbucket", func() {
```
timebertt, line 546: Same here basically.
WDYT?

timebertt (commented):
Two minor suggestions related to the tests, rest lgtm :)

rfranzke (approved):
/lgtm

PR #2936, opened 2020-10-01
Clean up obsolete defaulting logic

timuthy:

/area robustness
/kind cleanup
/priority normal

This PR removes the manual invocation of defaulting functions for several config APIs since these are implicitly called by [Decode](https://github.com/kubernetes/kubernetes/blob/1b4fa94c2131dd0b11b470a280b967b0b10ffe2d/staging/src/k8s.io/apimachinery/pkg/runtime/interfaces.go#L78-L79).

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/cmd/gardener-controller-manager/app/gardener_controller_manager.go
+++ b/cmd/gardener-controller-manager/app/gardener_controller_manager.go
@@ -23,7 +23,6 @@ import (
 	"os"
 
 	"github.com/gardener/gardener/cmd/utils"
-	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	gardencoreinformers "github.com/gardener/gardener/pkg/client/core/informers/externalversions"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
 	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
@@ -47,7 +46,6 @@ import (
 	"k8s.io/client-go/informers"
 	kubeinformers "k8s.io/client-go/informers"
 	kubernetesclientset "k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/client-go/tools/record"
 )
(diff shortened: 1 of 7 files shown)


rfranzke (dismissed):
/lgtm

cmd/gardener-controller-manager/app/gardener_controller_manager.go
```diff
@@ -127,22 +127,6 @@ func (o *Options) validate(args []string) error {
 	return nil
 }
 
-func (o *Options) applyDefaults(in *config.ControllerManagerConfiguration) (*config.ControllerManagerConfiguration, error) {
```
timebertt, line 130: Not related to this PR, but can you check line 84, though?
It seems wrong to add the gardener API types to the kubernetes `scheme.Scheme` 😄 
```go
if err := gardencorev1beta1.AddToScheme(scheme.Scheme); err != nil {
```

pkg/scheduler/apis/config/loader/loader.go
```diff
@@ -24,27 +24,13 @@ import (
...
 
-var (
-	Codec  runtime.Codec
-	Scheme *runtime.Scheme
-)
+var Scheme *runtime.Scheme
 
 func init() {
 	Scheme = scheme.Scheme
 	install.Install(Scheme)
```
timebertt: This also installs the config API types into the kubernetes `scheme.Scheme`. Can you fix that?

cmd/gardenlet/app/gardenlet.go
```diff
@@ -138,34 +138,13 @@ func (o *Options) validate(args []string) error {
 	return nil
 }
 
-func (o *Options) applyDefaults(in *config.GardenletConfiguration) (*config.GardenletConfiguration, error) {
```
timebertt, line 141: Same in this file in L99:
```go
if err := gardencorev1beta1.AddToScheme(o.scheme); err != nil {
```
Can you fix that?

timebertt (commented):
Thanks for cleaning up!
I have some comments, though they are not directly related to your changes.
I just noticed these things while checking your PR, so maybe you can also clean them up as you are already on it :) 

Anyways, your changes lgtm.

timuthy:
Thanks for you reviews. I checked and implemented the suggestions from @timebertt and also tested them manually to be sure. PTAL.

timebertt (approved):
Thanks for taking care!
/lgtm

PR #2937, opened 2020-10-01
Fix premature deletion of vpa-webhook-config for ShootedSeed

ialidzhikov:

/area auto-scaling
/kind bug
/priority critical

https://github.com/gardener/gardener/pull/2886 is basically fixing naming collision for VPA webhook between seed boostrap and shoot control reconcile/ (via ManagedResource). But there is also one more case affecting ShootedSeed.
Currently for a ShootedSeed with `spec.kubernetes.verticalPodAutoscaler.enabled=true`, the seed bootstrap will prematurely delete the stale `vpa-webhook-config` which will leave the ShootedSeed without any VPA webhook until upcoming reconciliation in the maintenance time window which will create `vpa-webhook-config-shoot`.
Such situation without any VPA webhook leads to Pods eviction at each `~2m` with reason `EvictedByVPA`.

This PR also cleans up the `vpa-webhook-config` webhook from Shoots as `vpa-webhook-config-shoot` should be the acting one.

**Release note**:

```improvement operator
An issue causing premature deletion of the stale VPA webhook before the creation of the new VPA webhook for ShootedSeeds is now fixed.  
```

```improvement operator
An issue causing the stale VPA webhook to do not be cleaned up is now fixed.
```

### DIFF
--- a/pkg/operation/botanist/addons.go
+++ b/pkg/operation/botanist/addons.go
@@ -34,6 +34,7 @@ import (
 	"github.com/gardener/gardener/pkg/utils/secrets"
 	versionutils "github.com/gardener/gardener/pkg/utils/version"
 
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


rfranzke (approved):
/lgtm

Issue #2822, opened 2020-09-08
 ☂️ -Issue for graduating CachedRuntimeClients to beta


timuthy:
**How to categorize this issue?**

/area scalability
/kind enhancement

**What would you like to be added**:
Since the introduction of the cached controller-runtime clients #2449 we should strive for graduating this feature to beta so that we can make use of all the delivered features in development. We broke down the graduation process to the following tasks after we tested the feature on different internal landscapes:

Required for graduation:
- [x] Get rid of the `DirectClient` and replace its usages by Patch requests and other proper mechanisms (see https://github.com/gardener/gardener/issues/2822#issuecomment-702132185) ➡️ preparations: see linked PRs below, removal in https://github.com/gardener/gardener/pull/4225
- [x] Improve g-api stability during rolling updates ➡️ #2814
- [x] Distribute rollout of Gardenlets for shooted seeds ➡️ #2687
- [x] Refactor existing controllers to get and use informers from the controller-runtime cache instead of relying on SharedInformerFactories (i.e. get rid of duplicated watches)
  - gardener-scheduler ➡️  #3746
  - gardener-controller-manager ➡️ #3768, #3982 (and similar)
  - gardenlet ➡️ #4503, #4510
- [x] Replace `CreateOrUpdate` with `GetAndCreateOr*Patch` ➡️ #4227
- [x] ~Replace `CreateOrUpdate` in [gardener-resource-manager library](https://github.com/gardener/gardener/blob/e8b66dccd11adb73e284cbf4a3ac36246eeb0b88/vendor/github.com/gardener/gardener-resource-manager/pkg/manager/managedresources.go#L106)~
- [x] Eliminate wrongful usages of `retry.RetryOnConflict` ➡️ #4246
- [x] Eliminate wrongful usages of `Try{Update,Patch}` in gardener components ➡️ #4378
- [x] add some developer documentation on client usage (e.g. caveats of cached clients, conflict management / optimistic locking, different patch types, ...) ➡️ #4425
- [x] use strategic merge patch for patching finalizers of built-in or gardener objects (provided by gardener-apiserver) in order to run without optimistic locking (make use of `merge` and `$deleteFromPrimitiveList` directive on finalizers array (see [doc](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md#merge-directive))) to reduce the amount of conflicts (especially on start of reconciliations) ➡️ #4696
[...]

PR #2940, opened 2020-10-02
Specifically cached client

timebertt:

/area cost networking
/kind enhancement
/priority normal

This PR add an implementation of the `client.{Reader,Client}` interfaces that can disable the cache or enable the cache only for a given set of GroupKinds.

The problem with the cached controller-runtime client is, that one `Get` or `List` call triggers the client to start a watch for the respective object and therefore caches all object of that kind. This leads to an unnecessarily increased memory footprint for controllers that start caches, although they don't need to.
This implementation is useful for extension controllers that e.g. create ManagedResources and corresponding Secrets but don't want to cache all Secrets in the Seed cluster (which might be a lot).
For an example usage see https://github.com/gardener/gardener-extension-networking-calico/pull/55.

This PR also enables configuring these cache settings for our clientsets that we use in g/g, if we see the necessity to do so in the future.

Related to #2822 and #1953 

/cc @timuthy @MartinWeindel 

**Release note**:

```noteworthy developer
The newly added `client.Client` implementation under `pkg/client/kubernetes/utils` can be used to enable/disable the controller-runtime cache only for a given set of object kinds.
- e.g. you can pass `utils.NewClientFuncWithDisabledCacheFor(&corev1.Secret{})` to `manager.Options.NewClient` to use a client, that always reads Secrets directly from the API server instead of caching all Secrets in the cluster.
```


pkg/client/kubernetes/options.go
```diff
@@ -18,16 +18,21 @@ import (
...
 
+// Config carries options for new ClientSets.
 type Config struct {
 	clientOptions       client.Options
 	restConfig          *rest.Config
 	cacheResync         *time.Duration
 	disableCachedClient bool
+
+	readSpecifiedFromCache    bool
+	specificallyCachedObjects []runtime.Object
```
timuthy, line 39: Can you use a separate struct for these options?
I think it makes it easier to see that these two fields are always used together and you can explicitly check if the options is set via `!= nil` [here](https://github.com/gardener/gardener/pull/2940/files#diff-12b0ad2256a3d5aa1a1937a7e752c733R265).

pkg/client/kubernetes/client.go
```diff
@@ -261,9 +262,16 @@ func newClientSet(conf *Config) (Interface, error) {
 
 	var runtimeClient client.Client
 	if UseCachedRuntimeClients && !conf.disableCachedClient {
-		runtimeClient, err = newRuntimeClientWithCache(conf.restConfig, conf.clientOptions, runtimeCache)
-		if err != nil {
-			return nil, err
+		if len(conf.specificallyCachedObjects) > 0 {
+			runtimeClient, err = utils.NewClientWithSpecificallyCachedReader(runtimeCache, conf.restConfig, conf.clientOptions, conf.readSpecifiedFromCache, conf.specificallyCachedObjects...)
```
timuthy: Btw. what is the reason we create an entirely new client object for the cached client instead of just creating the cache and use it together with the `directClient` in `client.DelegatingClient`? It hasn't changed with this PR but still wanted to ask for the motivation.

timuthy (commented):
Nice PR. I don't have much to add except the nits I described in the comments.

pkg/client/kubernetes/options.go
timebertt: Good suggestion. Done 👍

pkg/client/kubernetes/client.go
timebertt: TBH: I just implemented it the same way, c-r does it ([ref](https://github.com/kubernetes-sigs/controller-runtime/blob/ac380d61764a160b32946e606b0c9ecd2834e3e8/pkg/manager/manager.go#L374)) and haven't given this much thought.

Maybe it would make sense to reuse the direct client, but there might have been a concrete reason, why c-r implemented it like this.
Also I wouldn't be comfortable doing this change right now actually, because it will foster lock contention in the single client even more, which might be really bad for large landscapes (see similar issues with the RESTMapper). I still want to open an issue in c-r about the lock contention and do some benchmarks/profiling to have some numbers/proof. This will hopefully give me more confidence about making a decision here and I can reiterate then :)

timuthy: Okay, got it. Thanks for the explanation.

timuthy (approved):
/lgtm

rfranzke (approved):
/lgtm

Very nice PR!

PR #2945, opened 2020-10-06
Rework logging dashboards

Kristian-ZH:

/area logging
/kind enhancement
/priority normal

New logging dashboard is added prefixed with `Logging ....` for each exposed component.
Logging panels from `Kubernetes Control Plane Status` are removed

Fixes # https://github.com/gardener/logging/issues/72

@wyb1 @vlvasilev @rfranzke 

@g-pavlov Can you please take a look on the documentation changes?

**Release note**:

```improvement user
New logging dashboards are added, prefixed with `Logging` for each exposed component.
Logging panels in `Kubernetes Controlplane Status` are removed
```

### DIFF
--- a/charts/seed-monitoring/charts/grafana/values.yaml
+++ b/charts/seed-monitoring/charts/grafana/values.yaml
@@ -20,3 +20,21 @@ extensions:
 
 konnectivityTunnel:
   enabled: false
+
+exposedComponents:
+- dashboardName: "Kube Apiserver"
+  jobName: "kube-apiserver"
+  podPrefix: "kube-apiserver"
+  fileName: "kube-apiserver.json"
+- dashboardName: "Kube Controller Manager"
+  jobName: "kube-controller-manager"
+  podPrefix: "kube-controller-manager"
+  fileName: "kube-controller-manager.json"
+- dashboardName: "Kube Scheduler"
+  jobName: "kube-scheduler"
+  podPrefix: "kube-scheduler"
+  fileName: "kube-scheduler.json"
+- dashboardName: "Cluster Autoscaler"
+  jobName: "cluster-autoscaler"
+  podPrefix: "cluster-autoscaler"
+  fileName: "cluster-autoscaler.json"
--- a/charts/seed-monitoring/charts/grafana/dashboards/owners/kubernetes-control-plane-status-dashboard.json
+++ b/charts/seed-monitoring/charts/grafana/dashboards/owners/kubernetes-control-plane-status-dashboard.json
@@ -1249,32 +1249,6 @@
       }
     },
     {
-      "datasource": "loki",
-      "gridPos": {
-        "h": 11,
-        "w": 24,
-        "x": 0,
-        "y": 37
-      },
-      "id": 52,
-      "options": {
-        "showLabels": false,
-        "showTime": true,
-        "sortOrder": "Descending",
-        "wrapLogMessage": false
-      },
-      "targets": [
-        {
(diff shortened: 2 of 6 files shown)


charts/seed-monitoring/charts/grafana/values.yaml
```diff
@@ -19,4 +19,26 @@ extensions:
   dashboards: ""
 
 konnectivityTunnel:
-  enabled: false
\ No newline at end of file
+  enabled: false
+
+exposedComponents:
+- dashboardName: "Logging Kube Apiserver"
```
wyb1: Can we remove the logging prefix from the dashboards? I don't think the dashboard is only useful for logging. WDYT?

charts/seed-monitoring/charts/grafana/values.yaml
```diff
@@ -19,4 +19,26 @@ extensions:
...
+  fileName: "logging-kube-apiserver.json"
+- dashboardName: "Logging Kube Controller Manager"
+  jobName: "kube-controller-manager"
+  podPrefix: "kube-controller-manager"
+  fileName: "logging-kube-controller-manager.json"
+- dashboardName: "Logging Kube Scheduler"
+  jobName: "kube-scheduler"
+  podPrefix: "kube-scheduler"
+  fileName: "logging-kube-scheduler.json"
+- dashboardName: "Logging Cloud Controller Manager"
```
wyb1: Cloud controller dashboard should be added via the extensions. See the [provider-aws](https://github.com/gardener/gardener-extension-provider-aws/blob/master/charts/internal/seed-controlplane/charts/cloud-controller-manager/ccm-monitoring-dashboard.json) for an example

charts/seed-monitoring/charts/grafana/values.yaml
Kristian-ZH: I did this prefix to make it easier for users to find the dashboards in which there are logs.
But okay, I will describe the dashboards in the logging documentation. Also will add them in the release note so i think it is okay to remove the prefix

charts/seed-monitoring/charts/grafana/values.yaml
Kristian-ZH: Okey, then I can just add a logging panels CCM dashboard of all providers?

charts/seed-monitoring/charts/grafana/values.yaml
wyb1: You can also add a `logging` tag to the dashboard

charts/seed-monitoring/charts/grafana/values.yaml
Kristian-ZH: Done :)

charts/seed-monitoring/charts/grafana/values.yaml
Kristian-ZH: Done :)

wyb1 (approved):
I didn't see any logs in the dashboards, but I also just created a shoot to test this. If this is expected then /lgtm

Kristian-ZH:
> I didn't see any logs in the dashboards, but I also just created a shoot to test this. If this is expected then /lgtm

If the Grafana is opened via the Ingress, it should contains logs.
port-forwarding does not work because Loki expects header with `orgID`

wyb1 (approved):

vlvasilev (approved):
/lgtm

vpnachev (approved):

PR #2954, opened 2020-10-08
Do not deploy control plane exposure when hibernated

mvladev:

/area control-plane
/kind bug
/priority normal

Control plane exposure should only be created when the cluster is not hibernated or is not using SNI.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -256,18 +256,24 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 			Dependencies: flow.NewTaskIDs(deployKubeAPIServer),
 		})
 		deployControlPlaneExposure = g.Add(flow.Task{
-			Name:         "Deploying shoot control plane exposure components",
-			Fn:           flow.TaskFn(botanist.DeployControlPlaneExposure).RetryUntilTimeout(defaultInterval, defaultTimeout).SkipIf(useSNI),
+			Name: "Deploying shoot control plane exposure components",
+			Fn: flow.TaskFn(botanist.DeployControlPlaneExposure).RetryUntilTimeout(defaultInterval, defaultTimeout).
+				SkipIf(useSNI).
+				SkipIf(o.Shoot.HibernationEnabled),
 			Dependencies: flow.NewTaskIDs(deployReferencedResources, waitUntilKubeAPIServerIsReady),
 		})
 		waitUntilControlPlaneExposureReady = g.Add(flow.Task{
-			Name:         "Waiting until Shoot control plane exposure has been reconciled",
-			Fn:           flow.TaskFn(botanist.WaitUntilControlPlaneExposureReady).SkipIf(useSNI),
+			Name: "Waiting until Shoot control plane exposure has been reconciled",
+			Fn: flow.TaskFn(botanist.WaitUntilControlPlaneExposureReady).
+				SkipIf(useSNI).
+				SkipIf(o.Shoot.HibernationEnabled),
 			Dependencies: flow.NewTaskIDs(deployControlPlaneExposure),
 		})
 		destroyControlPlaneExposure = g.Add(flow.Task{
-			Name:         "Destroying shoot control plane exposure",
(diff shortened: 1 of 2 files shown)


ialidzhikov:
Well, I am not able to reproduce the issue to judge for myself whether this PR helps or not. Can you share steps to reproduce?

/invite @timuthy

timuthy:
The general problem with excluding steps during hibernation is that when you delete a hibernated cluster and you changed the deployment in the meantime, the resource (here cp exposure) is not up-to-date. That struck us in the past especially if the clusters were hibernated for a longer period of time, i.e. you never really know when migration code is supposed to be removed (/cc @ialidzhikov). Also see https://github.com/gardener/gardener/issues/1240.

mvladev:
> you changed the deployment in the meantime, the resource (here cp exposure) is not up-to-date.

This is another problem which should be fixed in the deletion flow - you must recreate/update all resources which are relevant to the cluster deletion. Kube apiserver / service / kubecontroller manager and so must all be up.

timuthy:
Can you still explain why this PR is of kind bug and what the bug is actually about, as @ialidzhikov already asked?

mvladev:
I'm not able to reproduce it, but I've seen the following error:

>  Normal   Hibernated      11h                   gardener-controller-manager  Hibernating cluster due to schedule
>  Normal   Reconciling     11h (x6 over 12h)     gardenlet                    Reconciling Shoot cluster state
>  Normal   Reconciled      11h (x5 over 12h)     gardenlet                    Reconciled Shoot cluster state
>  Warning  ReconcileError  11h                   gardenlet                    Flow "Shoot cluster reconciliation" encountered task errors: [task "Deploying internal domain DNS record" failed: Failed to create "internal" DNS record: "retry failed with context deadline exceeded, last error: dnsentries.dns.gardener.cloud \"internal\" not found" (status=Deleting, message=entry is scheduled to be deleted) task "Deploying external domain DNS record" failed: Failed to create "external" DNS record: "retry failed with context deadline exceeded, last error: dnsentries.dns.gardener.cloud \"external\" not found" (status=Deleting, message=entry is scheduled to be deleted) task "Waiting until Shoot control plane exposure has been reconciled" failed: Error while waiting for ControlPlane shoot--foo/foo-exposure to become ready: extension encountered error during reconciliation: Error reconciling controlplane: could not get kube-apiserver service load balancer address: Service "kube-apiserver" not found]

Which I don't know how it would happen in a normal reconcilation, but the `Waiting until Shoot control plane exposure has been reconciled` should not be executed if the cluster is hibernated or SNI is enabled.

timuthy (changes requested):
After iterating over the changes again, I reluctantly want to merge those changes for the following reason:
- By design, we've never deleted extension resources when the cluster was hibernated. Instead, we apply it again and the extension controllers are supposed to deal with the hibernation state of the cluster. This PR changes the behavior profoundly (even if it sounds reasonable) and thus can have unforeseeable side effects because we don't control every extension.
- >Which I don't know how it would happen in a normal reconcilation, but the Waiting until Shoot control plane exposure has been reconciled should not be executed if the cluster is hibernated or SNI is enabled.

Extension resources are supposed to be deployed **on every reconciliation** as there might be migration logic in a Gardener release and as said, it's the extension's tasks to deal with the hibernation state. Otherwise you could argue to not reconcile at all while the cluster is hibernated.

timuthy:
Any further opinions @mvladev @gardener/gardener-maintainers?

timebertt:
Not really, I don't fully get the problem with keeping the way it is atm and tend to agree with @timuthy.

rfranzke:
I agree with the colleagues, the main argument is

> By design, we've never deleted extension resources when the cluster was hibernated. Instead, we apply it again and the extension controllers are supposed to deal with the hibernation state of the cluster. This PR changes the behavior profoundly (even if it sounds reasonable) and thus can have unforeseeable side effects because we don't control every extension.

As we aren't even sure if this PR helps with the above seen problem and we don't have any reproduction steps, I'd suggest to close the PR for now and revisit it when we have more details/information on it?

rfranzke:
OK, if nobody objects until EOB today, let's close the PR then.

PR #2955, opened 2020-10-08
Remove deprecated seed taints

### LINKED ISSUE #2953: Seed .spec.settings.scheduling.visibility is flapping when it is set to false

/area quality
/kind bug
/priority normal

**What happened**:
Currently `.spec.settings.scheduling.visibility` is flapping between true and false when ShootedSeed `shoot.gardener.cloud/use-as-seed` annotation contains the `unvisible` identifier.

1. Add `unvisible` to the `shoot.gardener.cloud/use-as-seed` annotation

2. Ensure that the corresponding gardenlet seedConfig in the gardenlet configmap looks like:

```yaml
  spec:
    settings:
      scheduling:
        visible: false
```

Note that there are no taints.

3. Ensure that with Seed registration the Seed is set to visible=false and the invisible taint is added:

```yaml
  spec:
    settings:
      scheduling:
        visible: false
    taints:
    - key: seed.gardener.cloud/invisible
```

4. Ensure that with next Seed registration, visible is set to true and invisible taint is removed.

```yaml
  spec:
    settings:
      scheduling:
        visible: true
```

Note that the invisible taint is not present in the seedConfig in the gardenlet configmap.
This will cause gardenlet to send Update with the missing invisible Seed. Ref https://github.com/gardener/gardener/blob/v1.10.2/pkg/gardenlet/controller/shoot/seed_registration_control.go#L369-L400

The migrateSettings logic on apiserver side will thread the missing invisible taint as taint removal and will set the .spec.settings.visible=true. Ref https://github.com/gardener/gardener/blob/v1.10.2/pkg/registry/core/seed/strategy.go#L156-L173

**Environment**:

- Gardener version: master
- Others:

vpnachev:

/area ops-productivity
/kind api-change cleanup
/priority normal

Remove the seed taints that were deprecated with #2315.

Before this change, every time when the following sample seed spec is applied with `kubectl replace`
```yaml
kind: Seed
metadata:
  name: aws
spec:
  settings:
    excessCapacityReservation:
      enabled: true
    scheduling:
      visible: false
    shootDNS:
      enabled: true
    verticalPodAutoscaler:
      enabled: true
  taints:
  - key: seed.gardener.cloud/protected
```
the `spec.settings.scheduling.visible` is alternatively changing its value between `false` and `true`. 
This happening because when the taint `seed.gardener.cloud/invisible` is not provided, it is ambiguous to understand whether the seed should become visible or it has never been invisible. I have tried to implement the guide from [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility) and to sync both fields, but, it turns out that removing the taint cannot change `spec.settings.scheduling.visible` from `false` to `true`, but it will be re-added during the defaulting. Thus I have decided to completely remove these deprecated taints.

Fixes #2953 

**Release note**:

```action operator
The seed taints `seed.gardener.cloud/{disable-capacity-reservation,disable-dns,invisible}` are no longer respected and are disallowed in gardener version v1.12 (will be enabled again in the next minor version without any special semantic like previously). They are automatically removed from the seed resources , please use the respective seed.spec.settings fields from now on.
```

```improvement operator
Fix a bug which was causing the `seed.spec.settings` fields to be flapping between `true` and `false` when the settings and the taints in the sent request are not synced.
```


pkg/apis/core/v1alpha1/defaults.go
```diff
@@ -103,38 +103,34 @@ func SetDefaults_Seed(obj *Seed) {
...
+		"seed.gardener.cloud/invisible",
+	}
+	// remove taints
+	for _, taint := range taintsToRemove {
+		for i := len(obj.Spec.Taints) - 1; i >= 0; i-- {
+			if obj.Spec.Taints[i].Key == taint {
+				obj.Spec.Taints = append(obj.Spec.Taints[:i], obj.Spec.Taints[i+1:]...)
+			}
+		}
+	}
```
vpnachev: I am not fully confident that I have to forcefully remove the taints from the seed resource. Keeping them should not be a problem now when we no longer sync them with spec.settings fields. So any feedback will be appreciated.

example/50-seed.yaml
```diff
@@ -50,8 +50,8 @@ spec:
 # taints:
 # - key: seed.gardener.cloud/protected # only shoots in the `garden` namespace can use this seed
 # - key: <some-key>
-# DEPRECATED: The below keys are deprecated in favor of the `.spec.settings` section. Please don't use them anymore and
-# switch to the new seed settings.
+# DEPRECATED: The below keys are deprecated in favor of the `.spec.settings` section. They are bo longer respected and
```
ialidzhikov: Imo we can remove the corresponding taints also from the example. The doc already mentions about the removed taints and points to the version in which they are removed.
`seed.gardener.cloud/protected` is still respected and still present, so we need to keep it.

pkg/apis/core/v1alpha1/defaults.go
ialidzhikov: Well, I am fine with both options. If we pick this approach to remove them during defaulting, then I would suggest to add a TODO that reminds us to remove this clenaup code in a future version.
@gardener/gardener-maintainers , WDYT?

example/50-seed.yaml
vpnachev: done

pkg/apis/core/v1alpha1/defaults.go
timuthy: I'm also fine with removing the taints here.

timuthy:
I'm fine with the removal in general and don't have much comments unless the example in the description doesn't work for me to reproduce the issue but rather the following:

```
spec:
  settings:
    excessCapacityReservation:
      enabled: true
    scheduling:
      visible: true
    shootDNS:
      enabled: true
    verticalPodAutoscaler:
      enabled: true
  taints:
  - key: seed.gardener.cloud/invisible
```

vpnachev:
> unless the example in the description doesn't work for me

Do you use `kubectl apply` or `kubectl replace` to update the seed?

pkg/apis/core/v1alpha1/defaults.go
vpnachev: After some time I think we need the above clean up for one minor release to prevent orphaning these 3 taints, otherwise shoots without respective toleration cannot be scheduled on the seed. Of course, we can skip the clean up but then a Gardener operator might need to manually clean up them.

timuthy:
> > unless the example in the description doesn't work for me
> 
> Do you use `kubectl apply` or `kubectl replace` to update the seed?

I used `replace`.

vpnachev:
And you send the replace command at least 2 times?

timuthy:
> And you send the replace command at least 2 times?

Yep. First created it via `kubectl create -f` and then executed multiple times `kubectl replace -f`.

ialidzhikov (approved):
/lgtm

PR #2957, opened 2020-10-09
Add unit tests for admission handlers

### LINKED ISSUE #2854: Add unit tests for Gardener Admission Controller

/area quality
/kind task
/priority normal

Add unit tests for the namespace deletion and kubeconfig check:
https://github.com/gardener/gardener/tree/master/pkg/admissioncontroller/server/handlers/webhooks

**Why is this needed**:
These code branches are untested at the moment.

timuthy:

/area quality
/kind cleanup
/kind enhancement
/priority normal

This PR refactors the Gardener admission webhooks and adds unit tests for the different handlers.

Fixes #2854

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/admissioncontroller/server/handlers/webhooks/utils.go
+++ b/pkg/admissioncontroller/server/handlers/webhooks/utils.go
@@ -29,6 +29,13 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const (
+	// maxRequestBody is the maximum size of the `AdmissionReview` body.
+	// Same, fixed value from API server for general safety to reduce the odds of OOM issues while reading the body.
+	// https://github.com/kubernetes/kubernetes/blob/d8eac8df28e6b50cd0f5380e23fc57daaf92972e/staging/src/k8s.io/apiserver/pkg/server/config.go#L322
+	maxRequestBody = 3 * 1024 * 1024
+)
+
 func admissionResponse(allowed bool, msg string) *admissionv1beta1.AdmissionResponse {
 	response := &admissionv1beta1.AdmissionResponse{
 		Allowed: allowed,
--- a/pkg/admissioncontroller/server/handlers/webhooks/utils_test.go
+++ b/pkg/admissioncontroller/server/handlers/webhooks/utils_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"reflect"
 
 	. "github.com/gardener/gardener/pkg/admissioncontroller/server/handlers/webhooks"
 	core "github.com/gardener/gardener/pkg/apis/core/install"
@@ -53,7 +54,7 @@ var _ = Describe("Utils tests", func() {
 			scheme  = runtime.NewScheme()
 			decoder = serializer.NewCodecFactory(scheme).UniversalDecoder()
 
-			sizeRequest = int64(336)
+			sizeRequest = int64(342)
 
 			request = func() *http.Request {
 				return createHTTPRequest(&corev1.Secret{
@@ -61,7 +62,7 @@ var _ = Describe("Utils tests", func() {
 						Kind:       "Secret",
 						APIVersion: corev1.SchemeGroupVersion.String(),
 					},
-				}, scheme, authenticationv1.UserInfo{})
+				}, scheme, authenticationv1.UserInfo{}, admissionv1beta1.Create)
 			}
 		)
 
@@ -100,10 +101,10 @@ var _ = Describe("Utils tests", func() {
 	})
 })
 
-func createHTTPRequest(obj runtime.Object, scheme *runtime.Scheme, user authenticationv1.UserInfo) *http.Request {
+func createHTTPRequest(obj runtime.Object, scheme *runtime.Scheme, user authenticationv1.UserInfo, op admissionv1beta1.Operation) *http.Request {
 	admissionReview := &admissionv1beta1.AdmissionReview{}
 
-	if obj != nil {
+	if obj != nil && !reflect.ValueOf(obj).IsNil() {
 		writer := bytes.NewBuffer(nil)
 		jsonSerializer := runtimeserializer.NewSerializerWithOptions(
 			runtimeserializer.DefaultMetaFactory,
@@ -128,7 +129,15 @@ func createHTTPRequest(obj runtime.Object, scheme *runtime.Scheme, user authenti
 			Version:  gvr.Version,
 			Resource: gvr.Resource,
 		}
+
+		var name string
+		if meta, ok := obj.(metav1.Object); ok {
+			name = meta.GetName()
+		}
+
 		admissionReview.Request = &admissionv1beta1.AdmissionRequest{
+			Name:            name,
+			Operation:       op,
 			Resource:        v1Gvr,
 			RequestResource: &v1Gvr,
 			UserInfo:        user,
--- a/pkg/admissioncontroller/server/handlers/webhooks/validate_kubeconfig_secrets_test.go
+++ b/pkg/admissioncontroller/server/handlers/webhooks/validate_kubeconfig_secrets_test.go
@@ -1,0 +1,179 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package webhooks_test
+
+import (
+	"net/http"
+	"net/http/httptest"
+
+	. "github.com/gardener/gardener/pkg/admissioncontroller/server/handlers/webhooks"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var _ = Describe("ValidateKubeconfigSecret", func() {
+	var (
+		empty = func() runtime.Object {
+			return nil
+		}
+
+		configMap = func() runtime.Object {
+			return &corev1.ConfigMap{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "ConfigMap",
+					APIVersion: "v1",
+				},
+			}
+		}
+
+		secretTypeMeta = metav1.TypeMeta{
+			Kind:       "Secret",
+			APIVersion: "v1",
+		}
+
+		noKubeconfigSecret = func() runtime.Object {
+			return &corev1.Secret{
+				TypeMeta: secretTypeMeta,
+				Data: map[string][]byte{
+					"foo": {},
+				},
+			}
+		}
+
+		validKubeconfig = `
+---
+apiVersion: v1
+kind: Config
+current-context: local-garden
+clusters:
+- name: local-garden
+  cluster:
+    certificate-authority-data: Z2FyZGVuZXIK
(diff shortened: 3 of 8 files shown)


pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion_test.go
```diff
@@ -0,0 +1,258 @@
...
+						p := project(namespace.Name, "project2")
+						t := metav1.Now()
+						p.SetDeletionTimestamp(&t)
+						return p
+					}(),
+				).
+				withNamespace(*namespace).
+				withShoots(client.InNamespace(namespace.Name), shoot("namespace1", "shoot1")),
+			false, "there are still Shoots"),
+		Entry("should not pass because namespace is used by namespace", namespace,
```
ialidzhikov:
```suggestion
		Entry("should not pass because namespace is used by project", namespace,
```

pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion_test.go
```diff
@@ -0,0 +1,258 @@
...
+			newMockClient().
+				withProjects(
+					project("namespace1", "project1"),
+					project(namespace.Name, "project2"),
+				).
+				withNamespace(*namespace),
+			false, "you must delete the corresponding project \"project2\""),
+		Entry("should pass because namespace is used by namespace but not found by client", namespace,
```
ialidzhikov:
```suggestion
		Entry("should pass because namespace is used by project but not found by client", namespace,
```

pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion_test.go
```diff
@@ -0,0 +1,258 @@
...
+						p := project(namespace.Name, "project2")
+						t := metav1.Now()
+						p.SetDeletionTimestamp(&t)
+						return p
+					}(),
+				).
+				withNamespace(*namespace).
+				withShoots(client.InNamespace(namespace.Name)),
+			true, emptyMessage),
+		Entry("should not pass because namespace is used by namespace w/ deletionTimestamp but has shoots", namespace,
```
ialidzhikov:
```suggestion
		Entry("should not pass because namespace is used by project w/ deletionTimestamp but has shoots", namespace,
```

pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion_test.go
```diff
@@ -0,0 +1,258 @@
...
+		Entry("should pass because no project or shoots available", namespace,
+			newMockClient().withProjects(),
+			true, emptyMessage),
+		Entry("should pass because namespace is not project related", namespace,
+			newMockClient().
+				withProjects(
+					project("namespace1", "project1"),
+					project("namespace2", "project2"),
+				), true, emptyMessage),
+		Entry("should pass because namespace is used by namespace w/ deletionTimestamp", namespace,
```
ialidzhikov:
```suggestion
		Entry("should pass because namespace is used by project w/ deletionTimestamp", namespace,
```

ialidzhikov (commented):
Minor nits

pkg/admissioncontroller/server/handlers/webhooks/utils.go
```diff
@@ -29,6 +29,13 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const (
+	// maxRequestBody is the maximum size of the `AdmissionReview` body.
+	// Same, fixed value from API server for general safety to reduce the odds of OOM issues while reading the body.
+	// https://github.com/kubernetes/kubernetes/blob/d8eac8df28e6b50cd0f5380e23fc57daaf92972e/staging/src/k8s.io/apiserver/pkg/server/config.go#L322
+	maxRequestBody = 3 * 1024 * 1024
```
timebertt, line 36: Hmm, I know this doesn't change with this PR, but still one question:
I think we are not storing gardener resources in protobuf in etcd. Should we be more careful than k/apiserver here or do you think, we can neglect this?

pkg/admissioncontroller/server/handlers/webhooks/utils_test.go
```diff
@@ -53,15 +53,15 @@ var _ = Describe("Utils tests", func() {
 			scheme  = runtime.NewScheme()
 			decoder = serializer.NewCodecFactory(scheme).UniversalDecoder()
 
-			sizeRequest = int64(336)
+			sizeRequest = int64(342)
```
timebertt, line 57: One side question about `#DecodeAdmissionRequest - should not succeed because of wrong content type`:
What does actually happen, when someone talks to the API server using protobuf? Are webhook requests always send in JSON from the API server?

pkg/admissioncontroller/server/handlers/webhooks/validate_kubeconfig_secrets_test.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+		}
+	)
+
+	DescribeTable("Validate Kubeconfig",
+		func(objFn func() runtime.Object, subresource string, op admissionv1beta1.Operation, expectedAllowed bool, expectedMsg string) {
+			scheme := runtime.NewScheme()
+			Expect(corev1.AddToScheme(scheme)).To(Succeed())
+
+			user := func() authenticationv1.UserInfo {
```
timebertt, line 140: Why do you use a func that is invoked a few lines below here instead of a literal?

pkg/admissioncontroller/server/handlers/webhooks/validate_kubeconfig_secrets_test.go
```diff
@@ -0,0 +1,161 @@
...
+					Groups:   []string{"group"},
+				}
+			}
+			response := httptest.NewRecorder()
+			validator := NewValidateKubeconfigSecretsHandler()
+
+			request := createHTTPRequest(objFn(), scheme, user(), op)
+			if subresource != "" {
+				request = createHTTPRequestForSubresource(subresource)
+			}
```
timebertt: Is this still needed? I think it's not executed anywhere.

pkg/admissioncontroller/server/handlers/webhooks/validate_kubeconfig_secrets_test.go
```diff
@@ -0,0 +1,161 @@
...
+			Expect(admissionReview.Response.Allowed).To(Equal(expectedAllowed))
+			if expectedMsg != "" {
+				Expect(admissionReview.Response.Result).To(
+					PointTo(MatchFields(IgnoreExtras, Fields{
+						"Message": ContainSubstring(expectedMsg),
+					})))
+			}
+
+		},
+		Entry("request is empty (create)", empty, noSubResource, admissionv1beta1.Create, false, "missing admission request"),
```
timebertt: Can you add a test case for passing an object other than a Secret?

pkg/admissioncontroller/server/handlers/webhooks/validate_kubeconfig_secrets_test.go
```diff
@@ -0,0 +1,161 @@
...
+		Entry("should pass because no Kubeconfig is found (create)", noKubeconfigSecret, noSubResource, admissionv1beta1.Create, true, emptyMessage),
+		Entry("should pass because Kubeconfig is valid (create)", validKubeconfigSecret, noSubResource, admissionv1beta1.Create, true, emptyMessage),
+		Entry("should fail because Kubeconfig is invalid (create)", invalidKubeconfigSecret, noSubResource, admissionv1beta1.Create, false, "exec configurations are not supported"),
```
timebertt: Can you also add a test case for passing a Secret with `kubeconfig` set to some random string (or just not a valid kubeconfig yaml)?

pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion_test.go
```diff
@@ -0,0 +1,258 @@
...
+			Expect(admissionReview.Response.Allowed).To(Equal(expectedAllowed))
+			if expectedMsg != "" {
+				Expect(admissionReview.Response.Result).To(
+					PointTo(MatchFields(IgnoreExtras, Fields{
+						"Message": ContainSubstring(expectedMsg),
+					})))
+			}
+
+		},
+		Entry("should pass because no project or shoots available", namespace,
```
timebertt, line 103: Can you add test cases for 
- a different operation than delete
- an empty admission request

just like in the `ValidateKubeconfigSecret` test?

pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion_test.go
```diff
@@ -0,0 +1,258 @@
...
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "foo",
+			},
+		}
+	)
+
+	DescribeTable("Namespace deletion admission",
+		func(namespace *corev1.Namespace, cl client.Client, expectedAllowed bool, expectedMsg string) {
+			ctrl := gomock.NewController(GinkgoT())
```
timebertt: I find it a bit confusing, that you create a lot of mocks here, for which calls are never verified. So I assume you only mock the clients in order to have some preexisting objects.

If that's correct and you don't need to check, that explicit calls are made, I would replace the `mockClient` by [fake clients](https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/client/fake) and the mock cache and `syncedInformer` by [fake informers](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/informertest/fake_cache.go).
We shouldn't need to reimplement the stuff already available upstream, if it's sufficient for our use cases.

Another option without mocking or faking anything would be to use a local control plane (envtest).

WDYT?

timebertt (commented):
Nice PR, thank you very much!
I have mostly nit comments, but also one general comment on the mocks.

pkg/admissioncontroller/server/handlers/webhooks/utils.go
timuthy: The problem in general is, that you can't be sure about the `--max-request-bytes` configuration of etcd backing the GAPI server. That's the main reason why for the moment I decided to choose the same value as K8s and gain some experience with it. In the worst case, GAPI returns an error because the object cannot be saved in etcd, so even if this value was too high, it wouldn't really harm or cause service outages.

pkg/admissioncontroller/server/handlers/webhooks/utils_test.go
timuthy: > Are webhook requests always send in JSON from the API server?

Concerning my research https://github.com/gardener/gardener/pull/2781#discussion_r488542877, yes.

pkg/admissioncontroller/server/handlers/webhooks/utils.go
timebertt: Ok, that sounds good. Thanks for explaining.

pkg/admissioncontroller/server/handlers/webhooks/utils_test.go
timebertt: Ah, great. Thanks.

pkg/admissioncontroller/server/handlers/webhooks/validate_namespace_deletion_test.go
timuthy: I was switching to the fakeClient when I noticed a mock is more appropriate here even if the primary goal is not to verify the calls. With mocks it's easily possible to specify or configure the behavior, especially for errors apart from `NotFound`. But as you said, not verifying the calls was awkward, so I'm now calling `ctrl.Finish()` at the end of every test case.

ialidzhikov (approved):
/lgtm

pkg/admissioncontroller/server/handlers/webhooks/utils_test.go
```diff
@@ -100,10 +101,10 @@ var _ = Describe("Utils tests", func() {
 	})
 })
 
-func createHTTPRequest(obj runtime.Object, scheme *runtime.Scheme, user authenticationv1.UserInfo) *http.Request {
+func createHTTPRequest(obj runtime.Object, scheme *runtime.Scheme, user authenticationv1.UserInfo, op admissionv1beta1.Operation) *http.Request {
 	admissionReview := &admissionv1beta1.AdmissionReview{}
 
-	if obj != nil {
+	if obj != nil && !reflect.ValueOf(obj).IsNil() {
```
timebertt, line 107: What's the reason for this change?

timebertt (approved):
Thanks!
/lgtm

pkg/admissioncontroller/server/handlers/webhooks/utils_test.go
timuthy: This change enables the following cases at the same time:
```
createHTTPRequest(nil, ...
```

and

```
var obj runtime.Object
obj = (*corev1.ConfigMap)(nil)
createHTTPRequest(obj, ...
```

timebertt: Got it, thanks :)

PR #2963, opened 2020-10-10
Delete VolumeAttachments on hibernation

### LINKED ISSUE #172: VolumeAttachments are orphaned on hibernation

/area storage
/kind bug
/priority normal
/platform gcp

**What happened**:
VolumeAttachments are orphaned after hibernation and wake up. Such orphan VolumeAttachments block PV deletion afterwards.

PV to do not hang in Terminating because of orphan VolumeAttachments.

1. Create a Shoot

2. Create sample PVC

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ebs-claim
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 4Gi
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: centos
        command: ["/bin/sh"]
        args: ["-c", "while true; do echo $(date -u) >> /data/out.txt; sleep 5; done"]
        volumeMounts:
        - name: persistent-storage
          mountPath: /data
      volumes:
      - name: persistent-storage
        persistentVolumeClaim:
          claimName: ebs-claim
```

3. Ensure that there is one VolumeAttachment as expected

```
$ k get volumeattachments.storage.k8s.io
NAME                                                                   ATTACHER                PV                                         NODE                                                     ATTACHED   AGE
csi-428c30ed273c01d110380beb8bfac4052e7b3567300c9835421df82a0a99ddf4   pd.csi.storage.gke.io   pv--09fdeb10-ab27-441c-9daf-413238e2491b   shoot--foooooo--gcp-local-cpu-worker-z1-786584bb-gm8kb   true.      90s
```

4. Hibernate

5. Wake up

6. Ensure that the VolumeAttachment is not cleaned up after wake up

```
$ k get volumeattachments.storage.k8s.io
NAME                                                                   ATTACHER                PV                                         NODE                                                     ATTACHED   AGE
csi-3a2f98271fdce6a780cb93777e5c5042b64e0782855355fd1cd9a270231857d8   pd.csi.storage.gke.io   pv--09fdeb10-ab27-441c-9daf-413238e2491b   shoot--foooooo--gcp-local-cpu-worker-z1-786584bb-r4gls   true       89s
csi-428c30ed273c01d110380beb8bfac4052e7b3567300c9835421df82a0a99ddf4   pd.csi.storage.gke.io   pv--09fdeb10-ab27-441c-9daf-413238e2491b   shoot--foooooo--gcp-local-cpu-worker-z1-786584bb-gm8kb   true       9m25s
```

7. Delete the sample resources from step 2

8. Ensure that PV hangs in Terminating

```
$ k get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS        CLAIM               STORAGECLASS   REASON   AGE
pv--09fdeb10-ab27-441c-9daf-413238e2491b   4Gi        RWO            Delete           Terminating   default/ebs-claim   default                 23m
```

**Environment**:

- Gardener version (if relevant): v1.10
- Extension version: v1.11.0
- Kubernetes version (use `kubectl version`): v1.18.8
- Others:

### LINKED ISSUE #2545: PVs cannot be destroyed after waking up one hibernated shoots

/area storage
/kind bug
/priority normal

**What happened**:
`PVs` cannot be destroyed for waking up shoot clusters. The status of `PVs` are always in `Terminating`.
When executing `kubectl get volumeattachments.storage.k8s.io`, I noticed `PVs` are still binding to the old nodes (which already destroyed during hibernating).
`PVs` can be destroyed successfully.
1. Deploy one `deployment` or `statefulset` with `PVC`.
2. Hibernate the shoot.
3. Wake up the shoot and delete the `deployment` or `statefulset`, and `PVC`, but the `PV` cannot be destroyed.
4. Execute `kubectl get volumeattachments.storage.k8s.io`, can see the `PV` is still binding to the old node.

The issue seems related to CSI plugins. Tested on AWS and Alicloud, and both providers have the issue.

**Environment**:

- Gardener version: v1.6.5
- Kubernetes version (use `kubectl version`): 1.18.4
- Cloud provider or hardware configuration: alicloud-provider, aws-provider
- Others:

### LINKED ISSUE #2958: Guestbook test on openstack leak volumes from redis StatefulSet

/area testing
/kind bug
/priority normal

**What happened**:
In regular tests I noticed that a lot of volumes leak in our openstack landscape. Most of the leaked volumes are 8GB in size and contain some redis data, so I assume they are from the guestbook test with deploys a redis `sts`.

No leaking.

Currently I only observe it after the fact (after the resp. test cluster was already deleted). Haven't tried to reproduce yet.

We are currently testing k8s 1.19.2 and 1.18.9 clusters, unfortunately from the leaked volumes I cannot tell which clusters the volumes originated from.

**Environment**:

- Gardener version:
- Kubernetes version (use `kubectl version`): 1.19.2 and 1.18.9
- Cloud provider or hardware configuration: openstack
- Others:

ialidzhikov:

/area storage
/kind bug
/priority normal

Delete VolumeAttachments on hibernation.

Fixes #2545
Fixes https://github.com/gardener/gardener-extension-provider-gcp/issues/172
Fixes #2958

**Release note**:

```noteworthy operator
gardenlet is now deleting all `VolumeAttachment`s on shoot hibernation. As during hibernation machine-controller-manager performs a "force" deletion of machines and does not wait for volumes to detach, kube-controller-manager is not able to delete the corresponding `VolumeAttachment`s (and also the external-attacher to notice this deletion and remove its finalizer from the VolumeAttachment). Deleting `VolumeAttachment`s on hibernation should prevent `VolumeAttachment`s to be orphaned. Currently in the upstream kube-controller-manager, there is no garbage collection for `VolumeAttachment`s (see [kubernetes/kubernetes#77324](https://github.com/kubernetes/kubernetes/issues/77324)).
```

### DIFF
--- a/pkg/operation/botanist/volumeattachments.go
+++ b/pkg/operation/botanist/volumeattachments.go
@@ -1,0 +1,52 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package botanist
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/gardener/gardener/pkg/utils/retry"
+
+	"github.com/sirupsen/logrus"
+	storagev1beta1 "k8s.io/api/storage/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// DeleteVolumeAttachments deletes all VolumeAttachments.
+func DeleteVolumeAttachments(ctx context.Context, c client.Client) error {
+	return c.DeleteAllOf(
+		ctx,
+		&storagev1beta1.VolumeAttachment{},
+	)
+}
+
+// WaitUntilVolumeAttachmentsDeleted waits until no VolumeAttachments exist anymore.
+func WaitUntilVolumeAttachmentsDeleted(ctx context.Context, c client.Client, log *logrus.Entry) error {
+	return retry.Until(ctx, 5*time.Second, func(ctx context.Context) (done bool, err error) {
+		vaList := &storagev1beta1.VolumeAttachmentList{}
+		if err := c.List(ctx, vaList); err != nil {
+			return retry.SevereError(err)
+		}
+
+		if len(vaList.Items) == 0 {
+			return retry.Ok()
+		}
+
+		log.Infof("Waiting until all VolumeAttachments have been deleted in the shoot cluster...")
+		return retry.MinorError(fmt.Errorf("not all VolumeAttachments have been deleted in the shoot cluster"))
+	})
+}
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -412,6 +412,19 @@ func (b *Botanist) HibernateControlPlane(ctx context.Context) error {
 		if err := b.WaitUntilEndpointsDoNotContainPodIPs(ctxWithTimeOut); err != nil {
 			return err
 		}
+
+		// TODO: remove this mitigation once there is a garbage collection for VolumeAttachments (ref https://github.com/kubernetes/kubernetes/issues/77324)
+		// Currently on hibernation Machines are forecefully deleted and machine-controller-manager does not wait volumes to be detached.
+		// In this case kube-controller-manager cannot delete the corresponding VolumeAttachment objects and they are orphaned.
+		// Such orphaned VolumeAttachments then prevent/block PV deletion. For more details see https://github.com/gardener/gardener-extension-provider-gcp/issues/172.
+		// As the Nodes are already deleted, we can delete all VolumeAttachments.
+		if err := DeleteVolumeAttachments(ctxWithTimeOut, b.K8sShootClient.Client()); err != nil {
+			return err
+		}
+
+		if err := WaitUntilVolumeAttachmentsDeleted(ctxWithTimeOut, b.K8sShootClient.Client(), b.Logger); err != nil {
+			return err
+		}
 	}
 
 	// invalidate shoot client here before scaling down API server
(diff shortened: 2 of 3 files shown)


vpnachev (dismissed):
/lgtm

pkg/operation/botanist/volumeattachments.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+
+		if len(vaList.Items) == 0 {
+			return retry.Ok()
+		}
+
+		b.Logger.Infof("Waiting until all VolumeAttachments have been deleted in the shoot cluster...")
+		return retry.MinorError(fmt.Errorf("not all VolumeAttachments have been deleted in the shoot cluster"))
+	})
+}
```
timuthy: Can you rather pass the client instead of using the pointer receiver and write some unit tests?

timuthy (changes requested):
Are there any known issues/disadvantages using this approach? Can we be sure that the external-attacher removes the finalizer even though the machine is gone?

ialidzhikov:
> Are there any known issues/disadvantages using this approach?

Well, I went through the CSI KEP, performed local testing, deployed this change on our dev env and haven't observed issues so far. I known that this does not look safe but we should do something to prevent volume leak on some infrastructures (see  #2958). 

> Can we be sure that the external-attacher removes the finalizer even though the machine is gone?

I hit no issues wrt to that so far. After the machine deletion on cloud provider side, volumes transition to available as they are detached on cloud provider side during the deletion (at least this is the case for aws and I would expect the same for other other providers in mcm). And external-attacher should not fail to detach a Volume which is already available.

ialidzhikov:
> Can we be sure that the external-attacher removes the finalizer even though the machine is gone?

I did a testing and I see that external-attacher is always calling `ControllerUnpublishVolume` of the csi-driver. The good thing is that the csi-driver handle this case.

Example logs of aws-ebs-csi-driver:

```
I1014 08:06:57.029891       1 controller.go:297] ControllerUnpublishVolume: called with args {VolumeId:vol-08ec22df22d730a9f NodeId:i-049b9e089d66121b2 Secrets:map[] XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}
W1014 08:06:57.113975       1 cloud.go:404] DetachDisk called on non-attached volume: vol-08ec22df22d731234
```

Ref https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/v0.7.0/pkg/cloud/cloud.go#L390-L427


I tested locally on aws, gcp, openstack, alicloud (for azure the CSI is postponed to k/k@v1.20) and the clean up of VolumeAttachments on hibernation worked well (-> the PV was able to be successfully deleted after wake up).

timuthy (approved):
/lgtm
Thank you also for the thorough research!

PR #2967, opened 2020-10-13
Shoot deletion from hibernation improvements

### LINKED ISSUE #1240: Redeploy components instead of simply scaling them up when deleting hibernated shoot

mvladev:

/area control-plane
/kind technical-debt
/priority normal

- Now all control plane-related steps from the standard reconciliation are executed
- Remove `WakeUpControlPlane` method

`WakeUpControlPlane` was attempting to do everything related to waking up a control plane of the cluster, but it was not doing things in parallel or any errors would cause the entire function to be executed from the start. There was quite a lot of duplication and missing steps which would result in a situations where the deletion would fail e.g. missing `DeployReferencedResources` step.

Fixes #1240

A similar `WakeUpKubeAPIServer` method exists which would be fixed in the same manner in another PR. 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_delete.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_delete.go
@@ -32,6 +32,7 @@ import (
 	"github.com/gardener/gardener/pkg/utils/flow"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	retryutils "github.com/gardener/gardener/pkg/utils/retry"
+	versionutils "github.com/gardener/gardener/pkg/utils/version"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -50,8 +51,10 @@ func (c *Controller) runDeleteShootFlow(ctx context.Context, o *operation.Operat
 		shootNamespaceInDeletion             bool
 		kubeAPIServerDeploymentFound         = true
 		kubeControllerManagerDeploymentFound = true
+		kubeAPIServerDeploymentReplicas      int32
 		infrastructure                       *extensionsv1alpha1.Infrastructure
 		controlPlaneDeploymentNeeded         bool
+		enableEtcdEncryption                 bool
 		tasksWithErrors                      []string
 		err                                  error
 	)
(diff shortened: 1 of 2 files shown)


pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -237,20 +297,52 @@ func (c *Controller) runDeleteShootFlow(ctx context.Context, o *operation.Operat
...
+				deployETCD,
+				waitUntilEtcdReady,
+				waitUntilKubeAPIServerServiceIsReady,
+				waitUntilControlPlaneReady,
+				createOrUpdateETCDEncryptionConfiguration,
+			).InsertIf(!staticNodesCIDR),
+		})
+		_ = g.Add(flow.Task{
+			Name:         "Scale up Kubernetes API server",
+			Fn:           flow.TaskFn(botanist.ScaleKubeAPIServerToOne).RetryUntilTimeout(defaultInterval, defaultTimeout).DoIf(cleanupShootResources),
```
timuthy: This means that we downscale the KAS in case the shoot was not hibernated before which we shouldn't need to do, right? Also for the other up-scale calls, we should only invoke them if a hibernated cluster is deleted.

timuthy (changes requested):
Some tests with your changes went well for me, just one comment.

rfranzke:
ping @mvladev

pkg/gardenlet/controller/shoot/shoot_control_delete.go
```diff
@@ -237,20 +300,54 @@ func (c *Controller) runDeleteShootFlow(ctx context.Context, o *operation.Operat
...
+				waitUntilKubeAPIServerServiceIsReady,
+				waitUntilControlPlaneReady,
+				createOrUpdateETCDEncryptionConfiguration,
+			).InsertIf(!staticNodesCIDR),
+		})
+		_ = g.Add(flow.Task{
+			Name: "Scale up Kubernetes API server",
+			Fn: flow.TaskFn(botanist.ScaleKubeAPIServerToOne).
+				RetryUntilTimeout(defaultInterval, defaultTimeout).
+				DoIf(cleanupShootResources && kubeAPIServerDeploymentReplicas == 0),
```
rfranzke, line 329: cc @dguendisch  - I guess this also fixes the problem that we sporadically see in the integration tests (KAPI scaled down to 0)

rfranzke (dismissed):
/lgtm
/needs second-opinion
/invite @timuthy

timuthy (approved):
/lgtm

PR #2968, opened 2020-10-13
Use StrategicMergePatch to scale resources

mvladev:

/area control-plane
/kind cleanup
/priority normal

Strategic Merge Patch always updates the resource, regardless of current `resourceVersion` (the api call doesn't fail if the resource was updated). This is the desired behavior of the function.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/client/kubernetes/scaling.go
+++ b/pkg/client/kubernetes/scaling.go
@@ -16,48 +16,58 @@ package kubernetes
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


timuthy (approved):
Nice!
/lgtm

timebertt (approved):
Thank you, this makes perfect sense!
/lgtm

PR #2969, opened 2020-10-13
[GEP-7] Adds checks whether the target seed supports DNS before migration

plkokanov:

/area control-plane-migration
/kind enhancement
/priority normal

When the `Shoot.Spec.SeedName` field is updated to trigger migration, the current and new Seeds are checked whether they enable DNS or not. If either fo them (or both) does not enable DNS, the migration is forbidden.

Part of #1631 

None

**Release note**:

```improvement operator
Migrating a Shoot from a Seed which has DNS disabled to a Seed which has DNS enabled will now generate a default domain name for the Shoot, if the Shoot's DNS section is not marked as `unmanaged`.
```
```improvement operator
Attempting to migrating a Shoot with `spec.dns` != nil  to a Seed which has DNS disabled is now forbidden and will return an error.
```

### DIFF
--- a/plugin/pkg/shoot/dns/admission.go
+++ b/plugin/pkg/shoot/dns/admission.go
@@ -195,8 +195,20 @@ func (d *DNS) Admit(ctx context.Context, a admission.Attributes, o admission.Obj
 		if shoot.Spec.SeedName == nil {
 			return nil
 		}
-		if oldShoot.Spec.SeedName != nil && shoot.Spec.DNS != nil {
-			return checkFunctionlessDNSProviders(shoot.Spec.DNS)
+
+		if oldShoot.Spec.SeedName != nil {
+			if *oldShoot.Spec.SeedName != *shoot.Spec.SeedName {
+				if err := checkIfShootMigrationIsPossible(d.seedLister, oldShoot, shoot); err != nil {
+					return err
+				}
+				if shoot.Spec.DNS != nil {
+					return checkFunctionlessDNSProviders(shoot.Spec.DNS)
+				}
+				return nil
+			}
+			if shoot.Spec.DNS != nil {
+				return checkFunctionlessDNSProviders(shoot.Spec.DNS)
+			}
 		}
 	}
 
@@ -239,6 +251,19 @@ func checkFunctionlessDNSProviders(dns *core.DNS) error {
 	for _, provider := range dns.Providers {
 		if !gardenerutils.IsTrue(provider.Primary) && (provider.Type == nil || provider.SecretName == nil) {
 			return apierrors.NewBadRequest("non-primary DNS providers in .spec.dns.providers must specify a `type` and `secretName`")
+		}
+	}
+	return nil
+}
+
+func checkIfShootMigrationIsPossible(seedLister corelisters.SeedLister, oldShoot, newShoot *core.Shoot) error {
+	for _, seedName := range []string{*oldShoot.Spec.SeedName, *newShoot.Spec.SeedName} {
+		seedDNSDisabled, err := seedDisablesDNS(seedLister, seedName)
+		if err != nil {
+			return apierrors.NewBadRequest(fmt.Sprintf("could not get referenced seed: %+v", err.Error()))
+		}
+		if seedDNSDisabled {
+			return apierrors.NewBadRequest("source and destination seeds must enable DNS so that the shoot can be migrated")
 		}
 	}
 	return nil
--- a/plugin/pkg/shoot/dns/admission_test.go
+++ b/plugin/pkg/shoot/dns/admission_test.go
@@ -699,6 +699,22 @@ var _ = Describe("dns", func() {
 					Expect(*shoot.Spec.DNS.Domain).To(HaveSuffix(fmt.Sprintf(".%s.%s", projectName, domain)))
 				})
 
+				It("should pass because a default domain was re-assigned for the shoot (no domain)", func() {
+					oldShoot := shoot.DeepCopy()
+					shoot.Spec.DNS = nil
+
+					Expect(kubeInformerFactory.Core().V1().Secrets().Informer().GetStore().Add(&defaultDomainSecret)).To(Succeed())
+					Expect(coreInformerFactory.Core().InternalVersion().Projects().Informer().GetStore().Add(&project)).To(Succeed())
+					Expect(coreInformerFactory.Core().InternalVersion().Seeds().Informer().GetStore().Add(&seed)).To(Succeed())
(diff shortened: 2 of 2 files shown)


plugin/pkg/shoot/dns/admission.go
```diff
@@ -196,7 +196,8 @@ func (d *DNS) Admit(ctx context.Context, a admission.Attributes, o admission.Obj
 		if shoot.Spec.SeedName == nil {
 			return nil
 		}
-		if oldShoot.Spec.SeedName != nil {
+
+		if oldShoot.Spec.SeedName != nil && *oldShoot.Spec.SeedName == *shoot.Spec.SeedName {
```
rfranzke: /invite @timuthy

rfranzke (commented):
/lgtm

plugin/pkg/shoot/dns/admission.go
rfranzke: ping @timuthy

timuthy (commented):
The changes look good towards your description. 

> If the Shoot.Spec.DNS is nil and the new Seed supports DNS, then a default domain is generated for the Shoot (if the Shoot dns is not unmanaged)

However, I wonder if this is really the behavior we desire because it is irreversible. If a shoot from a non-DNS seed is migrated to a DNS seed, it'll never be able to run on a non-DNS seed again.  Opinions @plkokanov @rfranzke?

rfranzke:
Good point @timuthy. I don't have a clear opinion yet. @plkokanov @swilen-iwanow what are your thoughts?

swilen-iwanow:
Now when you point that out I would rather not go down that road, because it looks like a side effect of the migration

plkokanov:
If I change this, should we completely forbid migration between seeds with different DNS settings or add handling for shoots without DNS running on seeds that support DNS (currently the reconciliation flow breaks with a nil ptr exception if that happens).

timuthy:
> If I change this, should we completely forbid migration between seeds with different DNS settings or add handling for shoots without DNS running on seeds that support DNS (currently the reconciliation flow breaks with a nil ptr exception if that happens).

I vote for the the latter, i.e. let shoots w/o DNS run on DNS featured seeds.

plkokanov:
@mvladev reminded me last week that if the Shoot does not have DNS, when it is migrated the APIServer will be recreated and a new IP can be assigned for it. This could lead to the APIServer (and other dependant) certs to have to be regenerated and the nodes on th shoot cluster restarted. I think we currently do not have logic to regenerate certificates in this way and restarting the nodes can be a huge burden.
I guess in the tests I did locally for the initial implementation I somehow got lucky.

So what I can do for now is just forbid migration of shoots that do not have DNS config. @rfranzke @timuthy @swilen-iwanow wdyt?

rfranzke:
Yes, thanks, that's actually true, I forgot about it. I agree with forbidding migration of shoots scheduled to seeds with disabled DNS. If the shoot has an internal DNS record (but no external one) then it actually can be migrated (so just checking for shoots with `.spec.dns=nil` wouldn't be good enough).

rfranzke (dismissed):
/lgtm
/invite @timuthy
/remove status/author-action
/unassign @plkokanov

rfranzke:
ping @timuthy

plugin/pkg/shoot/dns/admission.go
```diff
@@ -245,6 +252,24 @@ func checkFunctionlessDNSProviders(dns *core.DNS) error {
...
+			return apierrors.NewBadRequest(fmt.Sprintf("could not get referenced seed: %+v", err.Error()))
+		}
+		if seedDNSDisabled {
+			return apierrors.NewBadRequest("source and destination seeds must enable DNS so that the shoot can be migrated")
+		}
+	}
+
+	if newShoot.Spec.DNS != nil {
+		return checkFunctionlessDNSProviders(newShoot.Spec.DNS)
+	}
```
timuthy: I actually would like to refrain from calling this here because the admission code is already pretty nested, and responsibility wise `checkIfShootMigrationIsPossible` shouldn't call any provider validation.

timuthy (changes requested):

plugin/pkg/shoot/dns/admission.go
plkokanov: Just so you know why I called it here: 
Generally, if the seed is updated nothing prevents the user to also update the DNS Providers.
So when the seed is updated I wanted to check if migration is possible, check if the providers are ok and then always return from the function so that we don't continue and generate a default domain if the DNS of the shoot happens to be nil.

timuthy: Yes, I'm not objecting the use case but I think you can also check the DNS providers first, no matter if shoot migration is initiated or not and call `checkIfShootMigrationIsPossible` if necessary.

plkokanov:
@timuthy could you have another look? Since we always have to return after checking the DNSProviders (otherwise the tests fail) I didn't move it at the top. Otherwise the code would look something like:
```
if oldShoot.Spec.SeedName != nil {
	if shoot.Spec.DNS != nil {
		if err := checkFunctionlessDNSProviders(shoot.Spec.DNS); err != nil {
			return err
		}
	}
	if *oldShoot.Spec.SeedName != *shoot.Spec.SeedName {
		return checkIfShootMigrationIsPossible(d.seedLister, oldShoot, shoot)
	}

	if shoot.Spec.DNS != nil {
		return nil
	}
}
```

timuthy (changes requested):
Thanks @plkokanov for addressing my comment. Please see my suggestion [here](https://github.com/timuthy/gardener/commit/5c2a4234b499c9c074e23c9abae44d2465c155b0) to improve nesting and get rid of the duplicated invocation.

plkokanov:
@timuthy adapted as you suggested. 
This is still a bit different than the original behaviour, as previously if the shoot.Spec.DNS == nil we would continue with the function. The tests do pass, but I'm not sure if there aren't any corner cases.

timuthy:
> @timuthy adapted as you suggested.
> This is still a bit different than the original behaviour, as previously if the shoot.Spec.DNS == nil we would continue with the function. The tests do pass, but I'm not sure if there aren't any corner cases.

As discussed, you are right with that difference and even though it might not make sense to re-assign a removed `dns` section during updates, it's a breaking change we shouldn't introduce. Sorry I didn't get your point earlier.

plkokanov:
Reverted to the previous change and added the additional test case as we discussed.

timuthy (approved):
/lgtm
Thanks @plkokanov

rfranzke (approved):
Thanks guys!

PR #2992, opened 2020-10-16
Add restrictions in user's Grafana for logs

Kristian-ZH:

/area logging
/kind enhancement
/priority normal

With this PR will be created two tenants in the Loki. `user` and `operator`. Two datasources will be created also for these two tenants. The users will no have permissions to access logs stored in the operator's tenant

https://github.com/gardener/logging/issues/72

@wyb1 @vlvasilev 

**Release note**:

```improvement operator
port-forwarding of the operator's and user's grafana will not work anymore for accessing logs.
The only way to access logs is by using the Ingress, because it attaches a new header with the tenantID in the request.
```

### DIFF
--- a/charts/seed-bootstrap/charts/fluent-bit/values.yaml
+++ b/charts/seed-bootstrap/charts/fluent-bit/values.yaml
@@ -27,3 +27,4 @@ lokiLabels:
   systemdLabels:
     hostname: "host_name"
     unit: "systemd_component"
+exposedComponentsTagPrefix: "user-exposed"
--- a/charts/seed-bootstrap/charts/fluent-bit/templates/_fluent-bit-config.tpl
+++ b/charts/seed-bootstrap/charts/fluent-bit/templates/_fluent-bit-config.tpl
@@ -110,13 +110,47 @@
         MinBackoff 30
         Buffer true
         BufferType dque
-        QueueDir  /fluent-bit/buffers
+        QueueDir  /fluent-bit/buffers/operator
         QueueSegmentSize 300
         QueueSync normal
-        QueueName gardener-kubernetes
+        QueueName gardener-kubernetes-operator
         FallbackToTagWhenMetadataIsMissing true
         TagKey tag
         DropLogEntryWithoutK8sMetadata true
+        TenantID operator
+    
+    [Output]
+        Name loki
+        Match {{ .Values.exposedComponentsTagPrefix }}.kubernetes.*
+        Url http://loki.garden.svc:3100/loki/api/v1/push
+        LogLevel info
+        BatchWait 40
+        BatchSize 30720
+        Labels {test="fluent-bit-go", lang="Golang"}
+        LineFormat json
+        ReplaceOutOfOrderTS true
+        DropSingleKey false
+        AutoKubernetesLabels true
+        LabelSelector gardener.cloud/role:shoot
+        RemoveKeys kubernetes,stream,type,time,tag
+        LabelMapPath /fluent-bit/etc/kubernetes_label_map.json
+        DynamicHostPath {"kubernetes": {"namespace_name": "namespace"}}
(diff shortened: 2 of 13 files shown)


charts/seed-bootstrap/charts/fluent-bit/values.yaml
```diff
@@ -17,3 +17,10 @@ ports:
 additionalFilters: ""
 additionalParsers: ""
 fluentBitConfigurationsOverwrites: {}
+exposedComponents:
+  tagPrefix: "user-exposed"
+  components:
+  - "kube-apiserver"
+  - "kube-controller-manager"
+  - "kube-scheduler"
+  - "cluster-autoscaler"
```
rfranzke: "Hard-coding" them here does not fit to #2754. I would suggest to enhance the logging contract in a meaningful way, see https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/component/interfaces.go#L60-L61 and https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/controlplane/kubescheduler/logging.go and https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/controlplane/clusterautoscaler/logging.go.

So far, it's only implemented for kube-scheduler and cluster-autoscaler. Hence, it's fine to keep kube-apiserver and kube-controller-manager for now (although, KCM is being refactored as part of #2881, cc @danielfoehrKn, and KAPI will follow soon).

rfranzke (changes requested):
Can you delete the `charts/seed-monitoring/tmp.txt` file, please?
/invite @vlvasilev

charts/seed-bootstrap/charts/fluent-bit/values.yaml
Kristian-ZH: Can you check now, please

pkg/operation/seed/logging.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/operation/seed/logging.go
```diff
@@ -0,0 +1,34 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package seed
+
+import "strings"
+
+var (
+	userAllowedComponets = []string{"kube-apiserver", "kube-controller-manager", "kube-scheduler", "cluster-autoscaler"}
```
rfranzke: typo
missing unit test

rfranzke (changes requested):
Hm, I'm confused now - that's not what we discussed in Slack.
My proposal was to extend the function type [here](https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/component/interfaces.go#L60-L61) to return not `(string, string, error)` but some struct `(LoggingConfig, error)` where the `LoggingConfig` has the following properties:

1. `filters` (string)
1. `parsers` (string)
1. `podName` (string)
1. `userExposed` (bool)

The `kubescheduler` and `clusterautoscaler` packages in `pkg/operation/botanist/controlplane/...` can then tell that they want to expose their logs to end-users. For the `kube-apiserver` and kube-controller-manager` we have to wait for @danielfoehrKn's refactoring, so they can remain "hard-coded" for now.

WDYT?

rfranzke (approved):
/lgtm

Thank you very much!

PR #2995, opened 2020-10-16
Extension enhancements

timuthy:

/area control-plane
/area cost
/area performance
/kind enhancement
/priority normal

This PR contains several usability, performance and cost improvements for the `/extensions` library, mainly used by Gardener extensions
- List all controller names that can be disabled via `--disable-controllers` in the cmd help.
- Only inject shoot client into health check if requested which mainly saves API discovery calls every check period.
- Patch conditions instead of updating them to save network traffic, especially if the extension holds `.status.state`.
- Patch results of health checks in bulk instead of updating every condition type individually.

/cc @danielfoehrKn 

**Release note**:

```improvement operator
Several improvements for extension health checks have been implemented which increase the runtime performance and reduce network traffic.
```
```improvement developer
Several improvements for extension health checks (lib `github.com/gardener/gardener/extensions`) have been implemented which increase the runtime performance and reduce network traffic. Please update your `github.com/gardener/gardener` to the latest version to profit from these changes.
```

### DIFF
--- a/extensions/pkg/controller/healthcheck/general/managed_resource.go
+++ b/extensions/pkg/controller/healthcheck/general/managed_resource.go
@@ -33,7 +33,6 @@ import (
 type ManagedResourceHealthChecker struct {
 	logger              logr.Logger
 	seedClient          client.Client
-	shootClient         client.Client
 	managedResourceName string
 }
 
@@ -47,11 +46,6 @@ func CheckManagedResource(managedResourceName string) healthcheck.HealthCheck {
 // InjectSeedClient injects the seed client
 func (healthChecker *ManagedResourceHealthChecker) InjectSeedClient(seedClient client.Client) {
 	healthChecker.seedClient = seedClient
-}
-
-// InjectShootClient injects the shoot client
-func (healthChecker *ManagedResourceHealthChecker) InjectShootClient(shootClient client.Client) {
-	healthChecker.shootClient = shootClient
 }
 
 // SetLoggerSuffix injects the logger
--- a/extensions/pkg/controller/utils.go
+++ b/extensions/pkg/controller/utils.go
@@ -206,6 +206,51 @@ func GetSecretByReference(ctx context.Context, c client.Client, ref *corev1.Secr
 	return secret, nil
 }
 
+// TryPatch tries to apply the given transformation function onto the given object, and to patch it afterwards with optimistic locking.
+// It retries the patch with an exponential backoff.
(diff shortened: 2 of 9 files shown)


extensions/pkg/controller/utils.go
```diff
@@ -206,6 +206,51 @@ func GetSecretByReference(ctx context.Context, c client.Client, ref *corev1.Secr
...
+// It retries the patch with an exponential backoff.
+func TryPatch(ctx context.Context, backoff wait.Backoff, c client.Client, obj runtime.Object, transform func() error) error {
+	return tryPatch(ctx, backoff, c, obj, c.Patch, transform)
+}
+
+// TryPatchStatus tries to apply the given transformation function onto the given object, and to patch its
+// status afterwards with optimistic locking. It retries the status patch with an exponential backoff.
+func TryPatchStatus(ctx context.Context, backoff wait.Backoff, c client.Client, obj runtime.Object, transform func() error) error {
+	return tryPatch(ctx, backoff, c, obj, c.Status().Patch, transform)
+}
```
rfranzke: Does it make sense to add some unit tests for those?

rfranzke (commented):
Generally /lgtm

extensions/pkg/controller/utils.go
timuthy: Yep, I'll add some.

extensions/pkg/controller/healthcheck/general/managed_resource.go
```diff
@@ -49,11 +48,6 @@ func (healthChecker *ManagedResourceHealthChecker) InjectSeedClient(seedClient c
 	healthChecker.seedClient = seedClient
 }
 
-// InjectShootClient injects the shoot client
-func (healthChecker *ManagedResourceHealthChecker) InjectShootClient(shootClient client.Client) {
```
danielfoehrKn, line 53: nice, so for managed resource health checks, a shoot client is not created any more?

danielfoehrKn (dismissed):
Looks good to me. Thanks!

extensions/pkg/controller/healthcheck/general/managed_resource.go
timuthy: Yes, a health checker now needs to have a `InjectShootClient` function to get the client.

extensions/pkg/controller/utils.go
timuthy: I added unit test, similar to `tryUpdate` and at the same time removed the code from the `/extensions` dir. Now only a func variable points to the implementation in `kutil`. I hope that's ok.

rfranzke (approved):
/lgtm

PR #2996, opened 2020-10-16
Integration test for the logging stack

vlvasilev:

/area logging
/kind test
/priority normal

This PR add an integration test for the logging by simulating seed environment with 100 cluster.
For the purpose is used a shoot on which is created namespace `garden` with the fluent-bit DaemonSet copied from the seed which holds the testing shoot control plane. Also one Loki StatefullSet is deployed with increased resources to handle the load from all of the namespaces. For the purpose of the simulation 100 Namespaces and Clusters resources are deployed in order to simulate multi shoot environment. In each namespace is deployed a logger application which logs 10000 logs for 20 seconds. All of the logs are sent to the single Loki instance and then the number of the logs (1 million) is verified.
This integration Test test the gardener fluent-bit plugin under high pressure.
   
Fixes #

**Release note**:

```improvement developer
Integration test for the logging is added simulating seed with 100 shoots
```


test/integration/shoots/logging/logging.go
```diff
@@ -0,0 +1,266 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

test/integration/shoots/logging/logging.go
```diff
@@ -0,0 +1,266 @@
...
+		framework.ExpectNoError(err)
+		//Get Loki Service
+		err = f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: lokiName}, lokiService)
+		framework.ExpectNoError(err)
+		//Get Loki Service
+		err = f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: lokiConfigMapName}, lokiConfMap)
+		framework.ExpectNoError(err)
+	}, initializationTimeout)
+
+	f.Beta().CIt("should get container logs from loki for all namespaces", func(ctx context.Context) {
```
schrodit: it deploys a lot of resources which may influence other tests.
Therefore, I would suggest to execute in serial.
```suggestion
	f.Beta().Serial().CIt("should get container logs from loki for all namespaces", func(ctx context.Context) {
```

test/integration/shoots/logging/logging.go
```diff
@@ -0,0 +1,266 @@
...
+		}))
+		for i := 0; i < numberOfSimulatedClusters; i++ {
+			shootNamespace := fmt.Sprintf("%s%v", simulatesShootNamespacePrefix, i)
+			ginkgo.By(fmt.Sprintf("Wait until logger application is ready in namespace %s", shootNamespace))
+			err := f.WaitUntilDeploymentsWithLabelsIsReady(ctx, loggerLabels, shootNamespace, f.ShootClient)
+			framework.ExpectNoError(err)
+		}
+
+		ginkgo.By("Verify loki received logger application logs for all namespaces")
+		err = WaitUntilLokiReceivesLogs(ctx, 5*time.Second, f, v1beta1constants.GardenNamespace, "pod_name", logger, logsCount*numberOfSimulatedClusters, f.ShootClient)
```
schrodit: is 5 seconds enough for loki to receive the logs?

schrodit (changes requested):
have you tested if 2 nodes are enough to handle the load.

cc @dguendisch

test/integration/shoots/logging/logging.go
vlvasilev: Sorry, I just used it for debug and forget to change it.

vlvasilev:
> have you tested if 2 nodes are enough to handle the load.

I was testing it on two nodes with 4CPU and 15GB RAM each.

dguendisch:
> I was testing it on two nodes with 4CPU and 15GB RAM each.

The regular TestMachinery shoots are typically 2 nodes with 2CPU and 8GB RAM each.
So if this test should be run somehow regularly and reusing the existing test shoots, can you check with these sizes as well?

Alternatively if the test is important enough we can spent a dedicated shoot for it, although I'm not sure this is satisfied in this case.

Nit: the PR title looks like copy-pasta :)

vlvasilev:
> The regular TestMachinery shoots are typically 2 nodes with 2CPU and 8GB RAM each.

@dguendisch The test pass well on a regular test shoot for about 21 minutes.
Is this time fair or we have to speed it up?

Kristian-ZH:
> > The regular TestMachinery shoots are typically 2 nodes with 2CPU and 8GB RAM each.
> 
> @dguendisch The test pass well on a regular test shoot for about 21 minutes.
> Is this time fair or we have to speed it up?

In my opinion, the time logs count should be decreased from 1m to 500k and then to set the timeout limit to 15min

Because ~21min means that the timeout should be 25min which is too much time for one test

dguendisch:
Current beta/serial testsuite runs on avg ~45mins. So with this test we'll add some 50% duration.
I can not fully judge whether this is really needed to stress test the fluent-bits, but if possible a speed up would be much welcomed. 20mins seems pretty much for one single test.

Besides that I also run the test and the nodes were not utilized too much (max 2GB RAM max 1CPU, so we have headroom as a few testclusters currently run with 2CPUs/4GB RAM).
I also noticed that the PVC for loki is some 30GB while at the end of the test only ~60MB were utilized.
Maybe you could downsize the PVC? (If yes, however keep in mind that you'd have to downsize differently for different cloudproviders as AliCloud has some min. volume size...)

Kristian-ZH:
> Current beta/serial testsuite runs on avg ~45mins. So with this test we'll add some 50% duration.
> I can not fully judge whether this is really needed to stress test the fluent-bits, but if possible a speed up would be much welcomed. 20mins seems pretty much for one single test.
> 
> Besides that I also run the test and the nodes were not utilized too much (max 2GB RAM max 1CPU, so we have headroom as a few testclusters currently run with 2CPUs/4GB RAM).
> I also noticed that the PVC for loki is some 30GB while at the end of the test only ~60MB were utilized.
> Maybe you could downsize the PVC? (If yes, however keep in mind that you'd have to downsize differently for different cloudproviders as AliCloud has some min. volume size...)

I think that we can entirely remove the PVC, /cc @vlvasilev

vlvasilev:
Yes, we can. The log retention is not necessary.

vlvasilev:
The Loki PVC is removed, the emitted logs are reduced to 500k. Now the test is 11 minutes long.

test/framework/shoot_utils.go
```diff
@@ -45,16 +45,16 @@ func (f *ShootFramework) ShootKubeconfigSecretName() string {
 	return fmt.Sprintf("%s.kubeconfig", f.Shoot.GetName())
 }
 
-// GetLokiLogs gets logs for <podName> from the loki instance in <lokiNamespace>
-func (f *ShootFramework) GetLokiLogs(ctx context.Context, lokiNamespace, podName string, client kubernetes.Interface) (*SearchResponse, error) {
+// GetLokiLogs gets logs for <key>, <value> from the loki instance in <lokiNamespace>
```
vpnachev:
```suggestion
// GetLokiLogs gets logs from the last 1 hour for <key>=<value> from the loki instance in <lokiNamespace>
```

test/integration/shoots/logging/logging.go
```diff
@@ -0,0 +1,279 @@
...
+		//Get the Loki ServiceAccount from the seed
+		err = f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: lokiName}, lokiServiceAccount)
+		framework.ExpectNoError(err)
+		//Get the Loki Service from the seed
+		err = f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: lokiName}, lokiService)
+		framework.ExpectNoError(err)
+		//Get the Loki ConfigMap from the seed
+		err = f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: lokiConfigMapName}, lokiConfMap)
+		framework.ExpectNoError(err)
+	}, initializationTimeout)
```
vpnachev: Remove the comments and do not assign the returned error to variable
```suggestion
	framework.CBeforeEach(func(ctx context.Context) {
		checkRequiredResources(ctx, f.SeedClient)
		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: fluentBitName}, fluentBit))
		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: fluentBitConfigMapName}, fluentBitConfMap))
		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: fluentBitName}, fluentBitService))
		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: fluentBitClusterRoleName}, fluentBitClusterRole))
		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: fluentBitClusterRoleName}, fluentBitClusterRoleBinding))
		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: fluentBitName}, fluentBitServiceAccount))
		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: "", Name: "clusters.extensions.gardener.cloud"}, clusterCRD))
		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: lokiName}, lokiSts))
		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: lokiName}, lokiServiceAccount))
		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: lokiName}, lokiService))
		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: lokiConfigMapName}, lokiConfMap))
	}, initializationTimeout)
```

test/integration/shoots/logging/logging.go
```diff
@@ -0,0 +1,279 @@
...
+
+	f.Beta().Serial().CIt("should get container logs from loki for all namespaces", func(ctx context.Context) {
+		ginkgo.By("Deploy the garden Namespace")
+		gardenNamespace = &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: v1beta1constants.GardenNamespace,
+			},
+		}
+		err := create(ctx, f.ShootClient.Client(), gardenNamespace)
+		framework.ExpectNoError(err)
```
vpnachev: What do you think about?
```suggestion
		framework.ExpectNoError(create(ctx, f.ShootClient.Client(), gardenNamespace))
```

dguendisch:
I retried, somehow the test got stuck after having found 485k entries:
```
STEP: Verify loki received logger application logs for all namespaces
time="2020-10-19T12:13:51+02:00" level=info msg="Waiting to receive 500000 logs, currently received 140575"
time="2020-10-19T12:14:54+02:00" level=info msg="Waiting to receive 500000 logs, currently received 232929"
time="2020-10-19T12:15:58+02:00" level=info msg="Waiting to receive 500000 logs, currently received 361179"
time="2020-10-19T12:17:02+02:00" level=info msg="Waiting to receive 500000 logs, currently received 474984"
time="2020-10-19T12:18:07+02:00" level=info msg="Waiting to receive 500000 logs, currently received 485000"
time="2020-10-19T12:19:12+02:00" level=info msg="Waiting to receive 500000 logs, currently received 485000"
time="2020-10-19T12:20:16+02:00" level=info msg="Waiting to receive 500000 logs, currently received 485000"
time="2020-10-19T12:21:21+02:00" level=info msg="Waiting to receive 500000 logs, currently received 485000"
```

Another question: is it required for the test to have 100 namespaces? If not I think from debugging, later instrospection and cleanup point-of-view, 1 ns for the log-generators would be preferable, wdyt?

And another nit: would be better if the ns would contain the testname somewhere, so that when we dump problems, we know that this is a leftover from this test here.

test/integration/shoots/logging/logging.go
```diff
@@ -0,0 +1,279 @@
...
+			ginkgo.By(fmt.Sprintf("Wait until logger application is ready in namespace %s", shootNamespace))
+			err := f.WaitUntilDeploymentsWithLabelsIsReady(ctx, loggerLabels, shootNamespace, f.ShootClient)
+			framework.ExpectNoError(err)
+		}
+
+		ginkgo.By("Verify loki received logger application logs for all namespaces")
+		err = WaitUntilLokiReceivesLogs(ctx, 1*time.Minute, f, v1beta1constants.GardenNamespace, "pod_name", logger, logsCount*numberOfSimulatedClusters, f.ShootClient)
+		framework.ExpectNoError(err)
+
+	}, getLogsFromLokiTimeout, framework.WithCAfterTest(func(ctx context.Context) {
```
vpnachev, line 208: I guess it would be more readable if the above function is not implemented in-place. It was very strange to me next argument here of something and started wondering what is happening.

test/integration/shoots/logging/utils.go
```diff
@@ -51,15 +62,26 @@ func checkRequiredResources(ctx context.Context, k8sSeedClient kubernetes.Interf
 	}
 }
 
-// WaitUntilLokiReceivesLogs waits until the loki instance in <lokiNamespace> receives <expected> logs from <podName>
-func WaitUntilLokiReceivesLogs(ctx context.Context, f *framework.ShootFramework, lokiNamespace, podName string, expected int, client kubernetes.Interface) error {
-	return retry.Until(ctx, 5*time.Second, func(ctx context.Context) (done bool, err error) {
-		search, err := f.GetLokiLogs(ctx, lokiNamespace, podName, client)
+// WaitUntilLokiReceivesLogs waits until the loki instance in <lokiNamespace> receives <expected> logs from <key>, <value>
```
vpnachev:
```suggestion
// WaitUntilLokiReceivesLogs waits until the loki instance in <lokiNamespace> receives <expected> logs for <key>=<value>
```

vpnachev (commented):
some minor nits

test/integration/shoots/logging/logging.go
vlvasilev: Just like the example above you gave me in CBeforeEach?

vlvasilev:
Sorry for the waiting but we are facing fluent-bit issue with this test. It seems that when fluent-bit is under pressure it silently drops logs. When we double the fluent-bit resources it handle the load but with the normal ones it could barely handle 500000 logs for 10 minutes. Normally on canary we found that fluent-bit handle no more than 12-13k logs per 10 minutes so we decided to  decrease the emitted log to 250000. Because we use cluster with only two nodes this guarantees that fluent-bit could handle 10x load, but from 100 test one fails. We must fine tune it until it gets 100/100 successful.

dguendisch:
@vlvasilev now that you mention tuning: the tests should better not assume certain sizing of the test clusters, imagine we change this sizing later on (for other reasons).
Or is the tuning you mentioned independent from the shoot cluster sizing?

vlvasilev:
@dguendisch It looks like there is a bug with the fluent-bit tail plugin. It skips some of the first logs of a random /var/log/container/ files silently.  It is not related to the fluent-bit CPU or memory.

test/integration/shoots/logging/logging.go
```diff
@@ -0,0 +1,244 @@
...
+		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: fluentBitConfigMapName}, fluentBitConfMap))
+		//Get the Fluent-Bit Service from the seed
+		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: fluentBitName}, fluentBitService))
+		//Get the Fluent-Bit ClusterRole from the seed
+		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: fluentBitClusterRoleName}, fluentBitClusterRole))
+		//Get the Fluent-Bit Rolebinding from the seed
+		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: fluentBitClusterRoleName}, fluentBitClusterRoleBinding))
+		//Get the Fluent-Bit ServiceAccount from the seed
+		framework.ExpectNoError(f.SeedClient.Client().Get(ctx, types.NamespacedName{Namespace: v1beta1constants.GardenNamespace, Name: fluentBitName}, fluentBitServiceAccount))
+		//Get the Fluent-Bit ServiceAccount from the seed
```
dguendisch:
```suggestion
		//Get the Fluent-Bit PriorityClass from the seed
```

dguendisch (changes requested):
Minor nit + my question from back then https://github.com/gardener/gardener/pull/2996#issuecomment-711998863 whether it wouldn't be better to have one or two ns instead of 100.
Otherwise looks good.

dguendisch:
One other comment, I noticed you added a delta of 100 so that the logs don't need to exactly match the expected 200k.
On differen runs I got
`Received 200094 logs. Expected 200000 logs with delta 100. Expected logs received.`
`Received 200005 logs. Expected 200000 logs with delta 100. Expected logs received.`
`Received 200100 logs. Expected 200000 logs with delta 100. Expected logs received.`
So one more line in the last run and the test would have failed. What are those extra lines?

vlvasilev:
@dguendisch . In this test we test manly the fluent-bit behavior, because only there we have a plugin written by us. The Loki is inflated with extra resources to handle load dedicated for 100 shoots, so its not as in the real landscape. Our custom fluent-bit-to-loki plugin decides where to send logs based on the namespace of the log entry. There could be many namespaces in one Seed so this could cause a synchronization issue, which we experienced with the first release when we tested it with only two namespaces.
All of the logs related to one Shoot is stored in a single `batch` and we use bulk push requests to Loki. So 100 pods in two namespaces or 2 pods in two namespaces wouldn't cause any difference.
About the extra lines. We noticed that when logs lines are missing their are only the first lines from a given logger. Which logger will lack log lines is not predictable, neither the amount of the missing logs. 
For example, if we have a `loggerA` and it have logged 50 lines for 1 second, the fluent-bit will start to process the logs from the 51-st log entry like `tail -f` does. This is from fluent-bit version 1.6.4, but we have experienced the same issue with 1.5.4, so we set the `read_from_head` option and even with these we have had the same issue.
At the next test run fluent-bit started to process  `loggerA` from the beginning. Totally random behavior.
In the test first we wait all of the fluent-bit pods to be ready and then we start to create the loggers.
We suggested that the fluent-bit takes time to start watching a file from its tail or head regardless the logger file(in the node) was created before or after the run of the fluent-bit pod.
So lets print one line and then wait for a given  period of time, before we start flooding the wanted amount of logs.
If the pods are 100 then we will have 100 extra line. But as I mention, some of those lines could be loosed.
And if we print 200000 logs we will wait in the best case for 200100 and in the worst case for 200000. This is the delta.
If we receive more than 200100 logs this will be a crucial error in the logging stack or in the test as no one have printed one extra log.
Even with this workaround we still experience a test failure due to missing logs  once per 100 times.

dguendisch:
Thanks for the clarification!

dguendisch (approved):
/lgtm

vpnachev (approved):
/lgtm

vpnachev:
Please, rebase to the latest master to get https://github.com/gardener/gardener/pull/3201/ which should fix the verify check.
/needs rebase

schrodit (approved):

PR #3002, opened 2020-10-19
EveryNodeReady shoot condition considers Kubelet version

### LINKED ISSUE #2990: Add alert for nodes that did not get kubelet patch update or show them Nodes status

/area monitoring 
/area ops-productivity
/kind enhancement
/priority normal

I would like to receive alert if the nodes of a shoot cluster are running on kubelet version different from the kubernetes version of the shoot. A corner case is to distinguish manually added nodes to the cluster. 

A query that can be used as a starting point
```
count(kube_node_info{type="shoot",kubelet_version!="<shoot-k8s-version>"}) > bool 0 == 1
```

Maybe an alert is not the right solution or in parallel we can report such machines via the `EveryNodeReady` for which I guess the MCM has to report the version mismatch.

cc @wyb1 @istvanballok @hardikdr @prashanth26 

**Why is this needed**:
Currently we lack automatic observability over such issues and we need to be notified by other means about issues lile #2981.

rfranzke:

/area ops-productivity monitoring
/kind enhancement
/priority normal

This PR adds an additional check that is executed as part of the status evaluation for the `EveryNodeReady` condition on shoots. Worker nodes matching the same major/minor Kubernetes version as defined in the shoot will be considered. Their kubelet version must match to the desired Kubernetes version of the shoot otherwise the condition's status will be `False`.

Fixes #2990

/invite @vpnachev 

**Release note**:

```noteworthy user
The `EveryNodeReady` condition on `Shoot` resources does now reflect kubelet version mismatches for shoot worker nodes (e.g., its status will be `False` if a kubelet wasn't updated after a patch version change, for example).
```

### DIFF
--- a/pkg/operation/botanist/health_check_test.go
+++ b/pkg/operation/botanist/health_check_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/gardener/gardener/pkg/operation/botanist"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 
+	"github.com/Masterminds/semver"
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
 	resourcesv1alpha1 "github.com/gardener/gardener-resource-manager/pkg/apis/resources/v1alpha1"
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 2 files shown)


pkg/operation/botanist/health_check_test.go
```diff
@@ -646,14 +653,50 @@ var _ = Describe("health check", func() {
...
+				newNode(nodeName, true, labels.Set{"worker.gardener.cloud/pool": workerPoolName1}, "v1.18.2"),
+			},
+			[]gardencorev1beta1.Worker{
+				{
+					Name:    workerPoolName1,
+					Maximum: 10,
+					Minimum: 1,
+				},
+			},
+			BeNil()),
```
danielfoehrKn, line 691: That should fail right?

Why would the condition be READY if the worker node still has a different minor version as the Shoot?

danielfoehrKn (changes requested):

pkg/operation/botanist/health_check_test.go
rfranzke: A rolling update might take time, so there is no reason to fail the condition. It's completely fine and normal to have two sets of nodes with different minor versions during such an upgrade.

vpnachev: If there is minor version mismatch, it should be detected by the worker controller.

vpnachev (approved):
/lgtm

danielfoehrKn (approved):

Issue #2745, opened 2020-08-18
NodeLocalDNS Feature lacks DNS network policy for control-plane components


zanetworker:
**How to categorize this issue?**

/area networking
/kind bug
/priority normal

**What happened**:
Without the `NodeLocalDNS` feature enabled,  a network policy to allow DNS traffic from the shoots control-plane components is defined as follows: 
```yaml 
spec:
  egress:
  - ports:
    - port: 53
      protocol: UDP
     ....
    to:
    - namespaceSelector:
        matchLabels:
          role: kube-system
      podSelector:
        matchExpressions:
        - key: k8s-app
          operator: In
          values:
          - kube-dns
          - coredns
```
This policy basically blocks DNS traffic destined anywhere else other than the CoreDNS pods in the `kube-system` namespace. 

With `NodeLocalDNS` we need to enable DNS traffic to the `node-cache` process running on the nodes. There are two ways to go about this: 
1. <s> Use a podSelector for `node-local-dns` pods -> that won't work because these pods are running on `hostNetwork`.</s> 
2. Use ipBlocks to the DNS service ClusterIP which is fixed, as well as the `node-local-dns` IP (in case IPVS is used `169.254.20.10`) -> this is the suggested fix. 

Suggested solution: 
```yaml
  egress:
  - ports:
    - port: 53
      protocol: UDP
    ....
    to:
    - ipBlock:
        cidr: 100.64.0.10/32
    - ipBlock:
        cidr: 169.254.20.10/32
```
 
**What you expected to happen**:
Shoot controlplanee traffic would be able to access DNS. 

**How to reproduce it (as minimally and precisely as possible)**:
Enable `NodeLocalDNS` on the seeds.

cc @ialidzhikov

zanetworker:
After looking into this a bit more, I noticed that the `allow-dns` policy is not the main cause. Removing all policies however resolves the problem. This will need to be debugged more, to do so a test-env with the seed having node-local-dns is required (otherwise more time on the dev landscape). Will try to create a nested local setup and troubleshoot this one more closely later.

rfranzke:
Any progress on this one?

HerrmannHinz:
@zanetworker how did you remove them? if i kill them they do come back after a few seconds.

PR #3003, opened 2020-10-19
Fix allow-to-dns network policy when nodeLocalDNS is enabled

mvladev:

/area networking
/kind bug
/priority normal

`gardener.cloud--allow-to-dns` network policy in the Shoot cluster now allows communication to the cluster ip of kube-dns

Before this change on a pod labeled with `networking.gardener.cloud/to-dns=allowed`:

```console
root@busybox:/# nslookup example.com.
;; connection timed out; no servers could be reached
```

After the change:

```console
root@busybox:/#  nslookup example.com.
Server:         10.252.0.10
Address:        10.252.0.10#53

Non-authoritative answer:
Name:   example.com
Address: 93.184.216.34
Name:   example.com
Address: 2606:2800:220:1:248:1893:25c8:1946
```

**Release note**:

```improvement operator
Fix a bug where `gardener.cloud--allow-to-dns`  network policy do not allow traffic to DNS when `NodeLocalDNS` feature gate is enabled.
```

### DIFF
--- a/charts/shoot-core/components/charts/network-policies/templates/allow-to-dns.yaml
+++ b/charts/shoot-core/components/charts/network-policies/templates/allow-to-dns.yaml
@@ -19,7 +19,11 @@ spec:
   - to:
     - podSelector:
         matchExpressions:
-        - {key: k8s-app, operator: In, values: [kube-dns,coredns]}
+        - {key: k8s-app, operator: In, values: [kube-dns]}
+    {{- if .Values.nodeLocalDNS.enabled }}
+    - ipBlock:
+        cidr: {{ .Values.nodeLocalDNS.kubeDNSClusterIP }}/32
+    {{- end }}
     ports:
     - protocol: UDP
       port: 53
--- a/pkg/operation/botanist/addons.go
+++ b/pkg/operation/botanist/addons.go
@@ -25,6 +25,7 @@ import (
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	"github.com/gardener/gardener/pkg/chartrenderer"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
+	netpol "github.com/gardener/gardener/pkg/operation/botanist/addons/networkpolicy"
 	"github.com/gardener/gardener/pkg/operation/botanist/component"
 	"github.com/gardener/gardener/pkg/operation/botanist/extensions/dns"
 	"github.com/gardener/gardener/pkg/operation/common"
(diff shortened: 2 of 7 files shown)


pkg/operation/botanist/addons.go
```diff
@@ -382,7 +382,12 @@ func (b *Botanist) generateCoreAddonsChart() (*chartrenderer.RenderedChart, erro
 		nodeExporterConfig        = map[string]interface{}{}
 		blackboxExporterConfig    = map[string]interface{}{}
 		nodeProblemDetectorConfig = map[string]interface{}{}
-		networkPolicyConfig       = map[string]interface{}{}
+		networkPolicyConfig       = map[string]interface{}{
+			"nodeLocalDNS": map[string]interface{}{
+				"enabled": b.Shoot.NodeLocalDNSEnabled,
+				"ip":      b.Shoot.Networks.CoreDNS.String(),
```
danielfoehrKn: Maybe rename it to make it more clear.
```suggestion
				"kubeDNSClusterIP":      b.Shoot.Networks.CoreDNS.String(),
```

danielfoehrKn (commented):
What about the service ip for  kube-dns-upstream svc?

mvladev:
> What about the service ip for kube-dns-upstream svc?

Which IP is that exactly? `b.Shoot.Networks.CoreDNS.String()` is exactly that?

charts/shoot-core/components/charts/network-policies/templates/allow-to-dns.yaml
```diff
@@ -20,6 +20,10 @@ spec:
     - podSelector:
         matchExpressions:
         - {key: k8s-app, operator: In, values: [kube-dns,coredns]}
+    {{- if .Values.nodeLocalDNS.enabled }}
+    - ipBlock:
+        cidr: {{ .Values.nodeLocalDNS.kubeDNSClusterIP }}/32
+    {{- end }}
```
timuthy, line 26: IIUC, this is only needed if Kube-Proxy runs in `IPTables` mode. Should we consider this case here, too?

timuthy: Btw. is [169.254.20.10](https://github.com/gardener/gardener/blob/71f3bf7e773ed767f26ba7d911b7e97c2f1f35e2/pkg/operation/botanist/operatingsystemconfig.go#L140) allowed by default?

mvladev: > Btw. is 169.254.20.10 allowed by default?

Yes

mvladev:
This is now enabled if IPTables proxy is used. Have another look @timuthy

timuthy (dismissed):
/lgtm
Thanks @mvladev!

rfranzke:
Does this PR fix https://github.com/gardener/gardener/issues/2745?

mvladev:
> Does this PR fix #2745?

It doesn't. This issue only targets shoot clusters and the `kube-system` namespace. 

I've just noticed that the IP should be `169.254.20.10` if IPVS is enabled.

rfranzke:
@mvladev what does you comment mean - do you have to update this PR, i.e., we should not merge it yet?

mvladev:
> @mvladev what does you comment mean - do you have to update this PR, i.e., we should not merge it yet?

It means two different bugs:

- first is that our Shoot clusters are affected if `NodeLocalDNS` feature gate is enabled. This PR fixes it.
- second one (#2745) affect us if we have a Seed cluster with `NodeLocalDNS` (not guaranteed to be created by Gardener ) which this PR doesn't fix. 

For #2745 a separate PR is needed.

rfranzke (approved):
/lgtm

wyb1 (approved):
/lgtm

timuthy (approved):
/lgtm

PR #3004, opened 2020-10-19
Update konnectivity server config

DockToFuture:

/area networking robustness
/kind enhancement
/priority normal

Set `delete-existing-uds-file=true`.

When the connectivity server container crashes and restarts it runs in an error if the UDS file already exists.
With this flag  set it cleans up the UDS file before listening.
See: #https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/69

```improvement operator
None
```


vpnachev (approved):
/lgtm

PR #3007, opened 2020-10-20
konnectivity-tunnel can be configured via annotation

### LINKED ISSUE #2982: Expose KonnectivityTunnel as feature gate in shoot spec

/kind enhancement
/priority normal

Expose `KonnectivityTunnel` as a feature gate in the shoot spec configuration.

**Why is this needed**:

The `KonnectivityTunnel` is another solution for the seed to shoot communication which has a lot of benefits over the `OpenVPN tunnel` solution, especially for private setups as is mentioned [here](https://github.com/gardener/gardener/blob/master/docs/usage/reverse-tunnel.md). At the moment the  `KonnectivityTunnel` can be enabled via featureGates in the [gardenlet](https://github.com/gardener/gardener/blob/master/example/20-componentconfig-gardenlet.yaml#L86) configuration.
Since this feature is not [stable](https://github.com/gardener/gardener/issues/2956) and we are depending on the upstream, we can't replace the `OpenVPN tunnel` solution with it. But it would be good if we can enable it in the shoot spec via a featureGate, to use it for testing or non-productive clusters. This would make setups in private zones much easier and it would not be needed to build provider specific solutions as for example mentioned [here](https://github.com/gardener/gardener-extension-provider-gcp/issues/56) for the gcp-provider.

/cc: @stiller-leser, @dkistner

DockToFuture:

/area networking
/kind enhancement
/priority normal

This PR makes it possible to enable the konnectivity-tunnel for a single shoot via "alpha.featuregates.shoot.gardener.cloud/konnectivity-tunnel: true" annotation on the shoot or to disable it via "alpha.featuregates.shoot.gardener.cloud/konnectivity-tunnel: false" annotation on the shoot. The shoot annotation will overwrite the config set by the gardenlet for the konnectivity-tunnel.

Fixes https://github.com/gardener/gardener/issues/2982

**Release note**:

```noteworthy user
The `KonnectivityTunnel` feature can now be configured for a single shoot cluster via the `alpha.featuregates.shoot.gardener.cloud/konnectivity-tunnel` annotation on the `Shoot` (`true` to enable it, `false` to disable it).
```

### DIFF
--- a/pkg/operation/shoot/shoot.go
+++ b/pkg/operation/shoot/shoot.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 	"time"
 
@@ -195,6 +196,9 @@ func (b *Builder) Build(ctx context.Context, c client.Client) (*Shoot, error) {
 	}
 
 	shoot.KonnectivityTunnelEnabled = gardenletfeatures.FeatureGate.Enabled(features.KonnectivityTunnel) && kubernetesVersionGeq118
+	if konnectivityTunnelEnabled, err := strconv.ParseBool(shoot.Info.Annotations[v1beta1constants.AnnotationShootKonnectivityTunnel]); err == nil && kubernetesVersionGeq118 {
+		shoot.KonnectivityTunnelEnabled = konnectivityTunnelEnabled
+	}
 
 	needsClusterAutoscaler, err := gardencorev1beta1helper.ShootWantsClusterAutoscaler(shootObject)
 	if err != nil {
--- a/docs/usage/reverse-tunnel.md
+++ b/docs/usage/reverse-tunnel.md
@@ -32,12 +32,15 @@ How it should look like at the end:
 
 ### How to Configure
 
-To enable the usage of the reverse tunnel feature, the Gardenlet `KonnectivityTunnel` feature-gate must be set to `true`.
+To enable the usage of the reverse tunnel feature, either the Gardenlet `KonnectivityTunnel` feature-gate must be set to `true` as shown below or the shoot must be annotated with `"alpha.featuregates.shoot.gardener.cloud/konnectivity-tunnel: true"`.
+
 ```yaml
 featureGates:
   KonnectivityTunnel: true
 ``` 
-Please refer to the examples [here](https://github.com/gardener/gardener/blob/master/example/20-componentconfig-gardenlet.yaml) for more information. 
+Please refer to the examples [here](https://github.com/gardener/gardener/blob/master/example/20-componentconfig-gardenlet.yaml) for more information.
+
+To disable the feature-gate the shoot must be annotated with `"alpha.featuregates.shoot.gardener.cloud/konnectivity-tunnel: false"`
 
 Once the feature-gate is enabled, a `proxy-server` sidecar will be deployed alongside the `kube-apiserver` which will receive and redirect connections 
 to the corresponding `proxy-agent` on the shoot cluster responsible for dataplane traffic routing. 
--- a/pkg/apis/core/v1alpha1/constants/types_constants.go
+++ b/pkg/apis/core/v1alpha1/constants/types_constants.go
@@ -225,6 +225,9 @@ const (
 	// AnnotationShootSkipCleanup is a key for an annotation on a Shoot resource that declares that the clean up steps should be skipped when the
 	// cluster is deleted. Concretely, this will skip everything except the deletion of (load balancer) services and persistent volume resources.
 	AnnotationShootSkipCleanup = "shoot.gardener.cloud/skip-cleanup"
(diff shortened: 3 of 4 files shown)


rfranzke (changes requested):
Can we rename the annotation to `alpha.shoot.gardener.cloud/konnectivity-tunnel` with possible values `enabled,disabled` as discussed in today's daily?
/cc @timuthy

mvladev:
I'll rather have some generic solution such as `featuregates.shoot.gardener.cloud/{feature-gate-name}: "true"` e.g. `featuregates.shoot.gardener.cloud/KonnectivityTunnel: "true"` so it can be applied to multiple features at the same time

mvladev:
Also you don't have to explicitly state `enabled`, `disabled`. [strconv.ParseBool](https://golang.org/pkg/strconv/#ParseBool ) should be sufficient.

rfranzke:
Not sure, we might not want to expose certain gardenlet feature gates in the shoot itself, and a general solution can anyways not be implemented because the respective annotations must be evaluated/checked individually in the code. Also, not all feature gates apply to shoots directly.
If it's only about the name (`alpha.shoot.gardener.cloud/konnectivity-tunnel`) vs. `alpha.featuregates.shoot.gardener.cloud/konnectivity-tunnel`) then sure, I'm not against including `featuregates.` in the key.

mvladev:
> the respective annotations must be evaluated/checked individually in the code.

We already do that for the features which are described in `pkg/features/features.go`.

> Also, not all feature gates apply to shoots directly.

Yes this is true. In some cases it can be possible to disable a feature and not the other way around. e.g. you cannot enable APIServerSNI if the Seed cluster doesn't support it, but you should be able to disable it.

rfranzke:
Hm, no, I disagree. I am against blindly exposing all gardenlet feature gates, usually only configurable by operators/administrators, to end-users. The `APIServerSNI` is actually a great example why we don't want to do it - disabling it per shoot imposes costs in the seed itself that are not necessarily redirected to the shoot owner. In this special case you might want to argue that it still might make sense because end-users might have outdated clients not supporting SNI (although, as learnt, this is highly unlikely anyways as it is supported since many years by most languages), though, I think it's too dangerous to allow end-users to tamper with all potential gardenlet feature gates.
In this special case with the `KonnectivityTunnel` it's anyways only a temporary solution. Eventually, we might want to entirely remove this gate plus the old VPN solution from the codebase, so it doesn't make any sense to enable/disable it in the shoot anymore (at some point it will be stable and the only/default solution).
Not sure what other @gardener/gardener-maintainers folks think, but I have strong doubts that the general exposure of these alpha features is a good idea as of now.

timuthy:
> Hm, no, I disagree. I am against blindly exposing all gardenlet feature gates, usually only configurable by operators/administrators, to end-users. The `APIServerSNI` is actually a great example why we don't want to do it - disabling it per shoot imposes costs in the seed itself that are not necessarily redirected to the shoot owner. In this special case you might want to argue that it still might make sense because end-users might have outdated clients not supporting SNI (although, as learnt, this is highly unlikely anyways as it is supported since many years by most languages), though, I think it's too dangerous to allow end-users to tamper with all potential gardenlet feature gates.
> In this special case with the `KonnectivityTunnel` it's anyways only a temporary solution. Eventually, we might want to entirely remove this gate plus the old VPN solution from the codebase, so it doesn't make any sense to enable/disable it in the shoot anymore (at some point it will be stable and the only/default solution).
> Not sure what other @gardener/gardener-maintainers folks think, but I have strong doubts that the general exposure of these alpha features is a good idea as of now.

+1
Let's decide individually which feature gates we want/need to support on a per shoot basis. This is also less error prone for the developer who develops the feature since she/he doesn't need to bother with this use-case.

rfranzke (dismissed):
/lgtm

pkg/operation/shoot/shoot.go
```diff
@@ -195,6 +196,9 @@ func (b *Builder) Build(ctx context.Context, c client.Client) (*Shoot, error) {
 	}
 
 	shoot.KonnectivityTunnelEnabled = gardenletfeatures.FeatureGate.Enabled(features.KonnectivityTunnel) && kubernetesVersionGeq118
+	if konnectivityTunnelEnabled, err := strconv.ParseBool(shoot.Info.Annotations["alpha.featuregates.shoot.gardener.cloud/konnectivity-tunnel"]); err == nil && kubernetesVersionGeq118 {
```
timuthy: Can you add `alpha.featuregates.shoot.gardener.cloud/konnectivity-tunnel` [here](https://github.com/gardener/gardener/blob/master/pkg/apis/core/v1beta1/constants/types_constants.go)?

timuthy (approved):
/lgtm
Thank you!

rfranzke (approved):

PR #3011, opened 2020-10-21
Increase `KubeEtcdFullBackupFailed` Alert to 2 days

wyb1:

/area monitoring
/kind cleanup
/priority normal

Increase time until `KubeEtcdFullBackupFailed` alert fires to 2 days. Previously this alert would fire after 1 day, but this causes many false positive alerts.

Fixes #

This change should be reverted once the cause of false positives is fixed.

**Release note**:

```improvement operator
Fire `KubeEtcdFullBackupFailed` alert after 2 days instead of 1 day
```

### DIFF
--- a/charts/seed-monitoring/charts/core/charts/prometheus/rules-tests/kube-etcd3.rules.test.yaml
+++ b/charts/seed-monitoring/charts/core/charts/prometheus/rules-tests/kube-etcd3.rules.test.yaml
@@ -32,9 +32,9 @@ tests:
     values: '1+0x62'
   # KubeEtcdFullBackupFailed
   - series: 'etcdbr_snapshot_latest_timestamp{job="kube-etcd3-backup-restore",role="main",kind="Full"}'
-    values: '0+0x2912'
+    values: '0+0x5790'
   - series: 'etcdbr_snapshot_required{job="kube-etcd3-backup-restore",role="main",kind="Full"}'
-    values: '1+0x2912'
+    values: '1+0x5790'
   # KubeEtcdRestorationFailed
   - series: 'etcdbr_restoration_duration_seconds_count{job="kube-etcd3-backup-restore",role="main",succeeded="false"}'
     values: '0+0x7 1 2 2'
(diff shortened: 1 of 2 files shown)


rfranzke:
Hm, wait a sec, can you share more information why we want/need this PR? Actually, we take a full snapshot once per day, so the old alert seems to be reasonable to me (without more background information).

wyb1:
> Actually, we take a full snapshot once per day, so the old alert seems to be reasonable to me (without more background information).

That is correct, but we get way too many alerts for `KubeEtcdFullBackupFailed`. Most (if not all) resolve after 1 day. This change would be temporary until the root cause is fixed. Right now these alerts provide very little value since almost all clusters with a hibernation schedule trigger this alert.

rfranzke:
Can we rather find out what's the root cause for it and estimate how long it would take to fix it before we work around it (which would anyway take some time to be rolled out, i.e., the number of false positive alerts won't decrease immediately anyways)?

wyb1:
Sure, I will close this PR for now

shreyas-s-rao:
@rfranzke The root cause for missing alternate full snapshots is basically that some shoots have their full snapshot schedule set within the hibernation period, so backup sidecar would have to take the full snapshot on shoot unhibernation. When sidecar starts, it checks whether there was a full snapshot in the last 24 hours - if yes, it skips full snapshot, or else it takes one. But sometimes it can so happen that there was a slight delay in the previous day's full snapshot (maybe etcd pod came up late for any number of reasons), because of which on next day's unhibernation, sidecar sees that there **was** a full snapshot in last 24 hours, and decides to skip it. After a few minutes, the alert fires promptly.

The fix for this is rather simple (but a temporary one of course, till we can add intelligence to sidecar to understand whether a daily full snapshot was taken based on the given cron schedule). The PR is already [here](https://github.com/gardener/etcd-backup-restore/pull/272), and will be part of next backup-restore release which will happen hopefully by early next week. /cc @wyb1

PR #3013, opened 2020-10-21
Condition transitions False->Progressing on reason/message change

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal

Failed conditions on `Shoot`s do now transition from `False` status to `Progressing` in case the reason or message changes (if thresholds are defined in the gardenlet component config only, otherwise `Progressing` is not used anyways).
This will improve ops productivity as a condition will no longer stay `False` for a long time although different errors are happening subsequently.

/cc @timuthy 

**Release note**:

```noteworthy user
Failed conditions on `Shoot`s do now transition from `False` status to `Progressing` in case the reason or message changes (if thresholds are defined in the gardenlet component config only, otherwise `Progressing` is not used anyways).
```

### DIFF
--- a/pkg/operation/botanist/health_check.go
+++ b/pkg/operation/botanist/health_check.go
@@ -383,7 +383,9 @@ func (b *HealthChecker) FailedCondition(condition gardencorev1beta1.Condition, r
 
 	case gardencorev1beta1.ConditionFalse:
 		threshold, ok := b.conditionThresholds[condition.Type]
-		if ok && b.lastOperation != nil && b.lastOperation.State == gardencorev1beta1.LastOperationStateSucceeded && Now().UTC().Sub(b.lastOperation.LastUpdateTime.UTC()) <= threshold {
+		if ok &&
+			((b.lastOperation != nil && b.lastOperation.State == gardencorev1beta1.LastOperationStateSucceeded && Now().UTC().Sub(b.lastOperation.LastUpdateTime.UTC()) <= threshold) ||
+				(reason != condition.Reason)) {
 			return gardencorev1beta1helper.UpdatedCondition(condition, gardencorev1beta1.ConditionProgressing, reason, message, codes...)
 		}
 	}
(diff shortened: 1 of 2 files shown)


timuthy (dismissed):
/lgtm

rfranzke:
Any further opinions?

pkg/operation/botanist/health_check.go
```diff
@@ -383,7 +383,9 @@ func (b *HealthChecker) FailedCondition(condition gardencorev1beta1.Condition, r
 
 	case gardencorev1beta1.ConditionFalse:
 		threshold, ok := b.conditionThresholds[condition.Type]
-		if ok && b.lastOperation != nil && b.lastOperation.State == gardencorev1beta1.LastOperationStateSucceeded && Now().UTC().Sub(b.lastOperation.LastUpdateTime.UTC()) <= threshold {
+		if ok &&
+			((b.lastOperation != nil && b.lastOperation.State == gardencorev1beta1.LastOperationStateSucceeded && Now().UTC().Sub(b.lastOperation.LastUpdateTime.UTC()) <= threshold) ||
+				(reason != condition.Reason || message != condition.Message)) {
```
timebertt: Hmm, what if the message permanently changes?
E.g. the worker controller adds the name of a failed machine to some of the error messages. If for some reason it continuously changes (bug in mcm/worker controller, operation is retried with a new machine or something similar), the error might not be surfaced, although the underlying error doesn't actually change.

timebertt (commented):
Hmm, TBH I'm not completely sure about the implications of this change.
I think, we need a bit more sophisticated logic here, unfortunately.
WDYT?

pkg/operation/botanist/health_check.go
rfranzke: Can you share such an error message?
I am open to revert it again if we see that it hides too many error that we would actually want to see, no problem.

> I think, we need a bit more sophisticated logic here, unfortunately.

For example?

rfranzke:
/unassign
/needs review

Incorporated feedback from @timebertt to only react on changes of the `reason` field (not `message`).

timebertt (approved):
Thanks
/lgtm

timuthy (approved):

PR #3014, opened 2020-10-21
Increase severe threshold timeout for etcd to 3m

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal

The etcd-druid waits for 1m for the statefulset backing etcd pods to be ready. Then it errors and writes this error into the `Etcd` resource's `.status.lastError` field. When Gardener detects this then it considers it as severe after `30s`. In the end, it might take some time (depending on the underlying infrastructure) to provision a volume, attach and mount it, and start etcd, so let's give the whole process a bit more time.

Ref #2397 

/cc @amshuman-kr @shreyas-s-rao @timuthy 

**Release note**:

```improvement operator
NONE
```


timuthy (approved):
/lgtm

amshuman-kr (approved):

PR #3016, opened 2020-10-21
Re-evaluate shoot status label if all seen errors cleaned

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal

With this PR we enhance the function that cleans the task IDs from the `Shoot`'s `.status.lastErrors[]` list (in case they become healthy). If no erroneous task remain in this list then we now re-evaluate the shoot status and, if necessary, update the status label.

This will improve ops productivity as the status is updated way faster now (without the necessity to wait for the next shoot care controller reconciliation loop (which also re-evaluates the status label, though, whose sync period is configurable and might be quite long).

**Release note**:

```improvement operator
NONE
```


timuthy (approved):
/lgtm

rfranzke:
Any further opinions?

timebertt (approved):
Makes perfect sense!
/lgtm

PR #3024, opened 2020-10-22
Add SeedChange feature gate for gardener-apiserver

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Adds `SeedChange` feature gate for `gardener-apiserver`. If set, this feature gate enables updating the `spec.seedName` field during shoot validation from a non-empty value in order to trigger shoot control plane migration.

Fixes #

Unlike other Gardener components, `gardener-apiserver` is not configured via a `componentconfig` but instead relies on passing the feature gates via the command line and uses the `	DefaultMutableFeatureGate` during startup. Therefore, I am simply adding the new feature gate to the `DefaultMutableFeatureGate` and making sure it's used instead of the (empty) `FeatureGate` when calling `o.Recommended.Admission.ApplyTo`.

**Release note**:

```noteworthy operator
`gardener-apiserver` now has a new feature gate `SeedChange`. If set, this feature gate enables updating the `spec.seedName` field during shoot validation from a non-empty value in order to trigger shoot control plane migration.
```

### DIFF
--- a/docs/deployment/feature_gates.md
+++ b/docs/deployment/feature_gates.md
@@ -26,6 +26,7 @@ The following tables are a summary of the feature gates that you can set on diff
 | ManagedIstio | `false` | `Alpha` | `1.5` | |
 | APIServerSNI | `false` | `Alpha` | `1.7` | |
 | MountHostCADirectories | `false` | `Alpha` | `1.11.0` | |
+| SeedChange | `false` | `Alpha` | `1.12.0` | |
 
 ## Using a feature
 
@@ -69,3 +70,4 @@ A *General Availability* (GA) feature is also referred to as a *stable* feature.
 * `ManagedIstio` enables a Gardener-tailored [Istio](https://istio.io) in each Seed cluster. Disable this feature if Istio is already installed in the cluster. Istio is not automatically removed if this feature is disabled. See the [detailed documentation](../usage/istio.md) for more information.
 * `APIServerSNI` enables only one LoadBalancer to be used for every Shoot cluster API server in a Seed. Enable this feature when `ManagedIstio` is enabled or Istio is manually deployed in Seed cluster. See [GEP-8](../proposals/08-shoot-apiserver-via-sni.md) for more details.
 * `MountHostCADirectories` enables mounting common CA certificate directories in the Shoot API server pod that might be required for webhooks or OIDC. 
+* `SeedChange` enables updating the `spec.seedName` field during shoot validation from a non-empty value in order to trigger shoot control plane migration.
--- a/pkg/apis/core/validation/shoot.go
+++ b/pkg/apis/core/validation/shoot.go
@@ -26,6 +26,7 @@ import (
 	"github.com/gardener/gardener/pkg/apis/core"
 	"github.com/gardener/gardener/pkg/apis/core/helper"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	"github.com/gardener/gardener/pkg/features"
 	"github.com/gardener/gardener/pkg/operation/botanist"
 	"github.com/gardener/gardener/pkg/utils"
 	cidrvalidation "github.com/gardener/gardener/pkg/utils/validation/cidr"
@@ -43,6 +44,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 )
 
 var (
@@ -148,7 +150,7 @@ func ValidateShootSpecUpdate(newSpec, oldSpec *core.ShootSpec, deletionTimestamp
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newSpec.Region, oldSpec.Region, fldPath.Child("region"))...)
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newSpec.CloudProfileName, oldSpec.CloudProfileName, fldPath.Child("cloudProfileName"))...)
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newSpec.SecretBindingName, oldSpec.SecretBindingName, fldPath.Child("secretBindingName"))...)
-	if oldSpec.SeedName != nil {
+	if oldSpec.SeedName != nil && !utilfeature.DefaultFeatureGate.Enabled(features.SeedChange) {
 		// allow initial seed assignment
 		allErrs = append(allErrs, apivalidation.ValidateImmutableField(newSpec.SeedName, oldSpec.SeedName, fldPath.Child("seedName"))...)
 	}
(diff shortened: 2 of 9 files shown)


pkg/apis/core/validation/shoot.go
```diff
@@ -148,7 +150,7 @@ func ValidateShootSpecUpdate(newSpec, oldSpec *core.ShootSpec, deletionTimestamp
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newSpec.Region, oldSpec.Region, fldPath.Child("region"))...)
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newSpec.CloudProfileName, oldSpec.CloudProfileName, fldPath.Child("cloudProfileName"))...)
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newSpec.SecretBindingName, oldSpec.SecretBindingName, fldPath.Child("secretBindingName"))...)
-	if oldSpec.SeedName != nil {
+	if oldSpec.SeedName != nil && !utilfeature.DefaultFeatureGate.Enabled(features.SeedChange) {
```
rfranzke, line 153: Can you adapt the test as well? :)

timebertt (commented):
Thanks!

Can you add the feature gate to [this doc](https://github.com/gardener/gardener/blob/master/docs/deployment/feature_gates.md)?

I think, with allowing the seed change, we should add some documentation for it (maybe transform [GEP-7](https://github.com/gardener/gardener/blob/master/docs/proposals/07-shoot-control-plane-migration.md) into a "real" doc under `/docs`). Doesn't have to be with this PR, though, but at least before the next release.

/needs documentation

pkg/apis/core/validation/shoot.go
stoyanr: Done. Also added an utility function `WithFeatureGate` to properly set / unset feature gates in tests since `featuregatetesting.SetFeatureGateDuringTest` from Kubernetes could not be used with Ginkgo tests.

stoyanr:
@timebertt 
> Can you add the feature gate to this doc?

Done

docs/deployment/feature_gates.md
```diff
@@ -26,6 +26,7 @@ The following tables are a summary of the feature gates that you can set on diff
 | ManagedIstio | `false` | `Alpha` | `1.5` | |
 | APIServerSNI | `false` | `Alpha` | `1.7` | |
 | MountHostCADirectories | `false` | `Alpha` | `1.11.0` | |
+| SeedChange | `false` | `Alpha` | `1.12.0` | |
```
timebertt, line 29: Can you add it to the `## List of feature gates`, too?

timebertt (commented):
Also, can you make the feature gate configurable via the `gardener/controlplane` chart?

docs/deployment/feature_gates.md
stoyanr: It is there already ...

docs/deployment/feature_gates.md
```diff
@@ -69,3 +70,4 @@ A *General Availability* (GA) feature is also referred to as a *stable* feature.
 * `ManagedIstio` enables a Gardener-tailored [Istio](https://istio.io) in each Seed cluster. Disable this feature if Istio is already installed in the cluster. Istio is not automatically removed if this feature is disabled. See the [detailed documentation](../usage/istio.md) for more information.
 * `APIServerSNI` enables only one LoadBalancer to be used for every Shoot cluster API server in a Seed. Enable this feature when `ManagedIstio` is enabled or Istio is manually deployed in Seed cluster. See [GEP-8](../proposals/08-shoot-apiserver-via-sni.md) for more details.
 * `MountHostCADirectories` enables mounting common CA certificate directories in the Shoot API server pod that might be required for webhooks or OIDC. 
+* SeedChange enables updating the `spec.seedName` field during shoot validation from a non-empty value in order to trigger shoot control plane migration.
```
timebertt:
```suggestion
* `SeedChange` enables updating the `spec.seedName` field during shoot validation from a non-empty value in order to trigger shoot control plane migration.
```

stoyanr:
@timebertt 

> Also, can you make the feature gate configurable via the gardener/controlplane chart?

It is already configurable, as there is already `values.apiserver.featureGates` that is used appropriately in the charts and is currently empty, see https://github.com/gardener/gardener/blob/9af2e8a04a08cb78c66bbdc0ab47af577c5b28d2/charts/gardener/controlplane/values.yaml#L79 and https://github.com/gardener/gardener/blob/9af2e8a04a08cb78c66bbdc0ab47af577c5b28d2/charts/gardener/controlplane/charts/runtime/templates/apiserver/deployment.yaml#L203. It should be empty since the feature gate is not enabled by default.
I couldn't find a place in this repo where these values are manipulated, so I think no further change is needed. Am I missing something?

docs/deployment/feature_gates.md
stoyanr: Corrected, thanks.

pkg/utils/test/test.go
```diff
@@ -121,3 +122,21 @@ func WithWd(path string) func() {
 		}
 	}
 }
+
+// WithFeatureGate sets the specified gate to the specified value, and returns a function that restores the original value.
+// Failures to set or restore cause the test to fail.
+// Example use:
+// defer WithFeatureGate(utilfeature.DefaultFeatureGate, features.<FeatureName>, true)()
```
timebertt: Nit, example code in doc strings should be indented:
```suggestion
//   defer WithFeatureGate(utilfeature.DefaultFeatureGate, features.<FeatureName>, true)()
```

timebertt (commented):
Thanks! One nit, otherwise lgtm.

timebertt:
> I couldn't find a place in this repo where these values are manipulated, so I think no further change is needed. Am I missing something?

Ah perfect, looks good. No further change should be needed.

pkg/utils/test/test.go
stoyanr: Done

timebertt (approved):

rfranzke (approved):
/lgtm

pkg/apis/core/validation/shoot_test.go
```diff
@@ -538,6 +543,18 @@ var _ = Describe("Shoot Validation Tests", func() {
 			)
 		})
 
+		It("should allow updating the seed if it has been set previously and the SeedChange feature gate is enabled", func() {
+			defer test.WithFeatureGate(utilfeature.DefaultFeatureGate, features.SeedChange, true)()
```
danielfoehrKn, line 547: Do I understand this correctly that test.WithFeatureGate() is executed right away (even though in the defer statement) - and returning a function that is then called by defer reverting the feature gate on the test?

I had to look more than twice to understand this. Why not just setting the concrete value before the test and then the opposite in the defer? 

I guess here it comes to elegance vs readability.

plkokanov: This is similar to the `WithVars` and other methods that we already have in the `utils/test` package.

stoyanr: @danielfoehrKn Yes, you understand it correctly. This is a "pattern" in Go, and `defer` is often used exactly in this way. The idea is that you achieve in a single line, and with a single function what you would otherwise achieve with 2 lines and 2 functions, and all passing of parameters that you would otherwise have to do yourself (e.g. the of the original value) happens automatically. Other functions in the `test` package also follow this pattern. In Kubernetes itself, there is a function `featuregatetesting.SetFeatureGateDuringTest` that also follows this pattern, unfortunately it can't be used directly with Ginkgo since it requires access to `testing.T`, so I basically re-implemented it here for Ginkgo.

PR #3025, opened 2020-10-22
Replace Loki VPA with HVPA

Kristian-ZH:

/area logging
/kind enhancement
/priority normal

Replace the VPA of the Loki with HVPA

Fixes #

/assign @amshuman-kr @ggaurav10 @rfranzke @vlvasilev 

**Release note**:

```improvement operator
Loki vpa is replaced with hvpa
```


charts/seed-bootstrap/charts/loki/templates/hvpa.yaml
```diff
@@ -0,0 +1,84 @@
...
+        resourcePolicy:
+          containerPolicies:
+            - containerName: loki
+              maxAllowed:
+                memory: {{ .Values.hvpa.maxAllowed.memory }}
+                cpu:  "{{ .Values.hvpa.maxAllowed.cpu }}"
+              minAllowed:
+                memory: {{ .Values.hvpa.minAllowed.memory }}
+                cpu: "{{ .Values.hvpa.minAllowed.cpu }}"
+            - containerName: backup-restore
```
rfranzke: ?

pkg/operation/botanist/logging.go
```diff
@@ -39,9 +50,19 @@ func (b *Botanist) DeploySeedLogging(ctx context.Context) error {
...
 	}
 
+	hvpaValues := make(map[string]interface{})
+	hvpaEnabled := gardenletfeatures.FeatureGate.Enabled(features.HVPA)
+	if b.ShootedSeed != nil {
+		hvpaEnabled = gardenletfeatures.FeatureGate.Enabled(features.HVPAForShootedSeed)
+	}
+
+	hvpaValues["enabled"] = hvpaEnabled
+	hvpaValues["maintenanceWindow"] = b.Shoot.Info.Spec.Maintenance.TimeWindow
```
rfranzke: I don't think we need the maintenance time window for the Loki scaling, it's not that critical like etcd. WDYT? cc @amshuman-kr

Kristian-ZH: I am okay to remove the check

amshuman-kr: Yes. Not required here.

Kristian-ZH: Okay, I will remove it :)

charts/seed-bootstrap/charts/loki/templates/hvpa.yaml
```diff
@@ -0,0 +1,82 @@
...
+    deploy: true
+    scaleUp:
+      updatePolicy:
+        updateMode: "Auto"
+{{- if .Values.hvpa.scaleUpStabilization }}
+{{ toYaml .Values.hvpa.scaleUpStabilization | indent 6 }}
+{{- end }}
+    scaleDown:
+      updatePolicy:
+        updateMode: "MaintenanceWindow"
```
amshuman-kr:
```suggestion
        updateMode: "Auto"
```
We decided to scale down ETCD only during maintenance window because of the disruption to the control-plane. It might be keep automatic scale down for loki. WDYT?

pkg/operation/botanist/logging.go
amshuman-kr: We can remove this as well as change scaleDown `updateMode` to `Auto` as mentioned above.

Kristian-ZH: It should be removed now

amshuman-kr (commented):
Thanks for the PR @Kristian-ZH! The changes look good. My understanding might not be up-to-date, does this change only work for the loki in the seed's `garden` namespace but also for the loki in the shoot namespaces?

Kristian-ZH:
> Thanks for the PR @Kristian-ZH! The changes look good. My understanding might not be up-to-date, does this change only work for the loki in the seed's `garden` namespace but also for the loki in the shoot namespaces?

Thanks for the comments and for the help!
The comments should be fixed now.

The HVPA will be applied in Shoot namespaces and the Garden namespace

amshuman-kr (dismissed):
LGTM

amshuman-kr:
> The HVPA will be applied in Shoot namespaces and the Garden namespace

Was confused by the name `DeploySeedLogging` 🙂

ggaurav10 (dismissed):

Kristian-ZH:
Changed the Gap values as @ggaurav10 recommended me

ggaurav10 (approved):

pkg/operation/botanist/logging.go
```diff
@@ -39,9 +50,18 @@ func (b *Botanist) DeploySeedLogging(ctx context.Context) error {
 		return err
 	}
 
+	hvpaValues := make(map[string]interface{})
+	hvpaEnabled := gardenletfeatures.FeatureGate.Enabled(features.HVPA)
+	if b.ShootedSeed != nil {
+		hvpaEnabled = gardenletfeatures.FeatureGate.Enabled(features.HVPAForShootedSeed)
```
rfranzke, line 56: Should we keep the VPA in case the HVPA feature gates are disabled? (same below)?

Kristian-ZH: In my opinion - No. With the current behaviour with these evictions, I think the Loki will work better without the VPA

rfranzke: OK, sure, probably this makes sense.

rfranzke (approved):
/lgtm

PR #3027, opened 2020-10-23
[GEP-7] Keep shoot managed resource objects during migration

stoyanr:

/area control-plane-migration
/kind bug
/priority normal

Ensures that shoot managed resource objects are kept during migration by the generic controlplane and worker actuators. This is important in order to ensure that objects such as storage classes, cluster roles and rolebindings, and CSI node daemon sets are not deleted from the shoot during the migration.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/extensions/pkg/controller/controlplane/genericactuator/actuator.go
+++ b/extensions/pkg/controller/controlplane/genericactuator/actuator.go
@@ -504,6 +504,14 @@ func (a *actuator) Migrate(
 	cp *extensionsv1alpha1.ControlPlane,
 	cluster *extensionscontroller.Cluster,
 ) error {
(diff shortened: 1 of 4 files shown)


rfranzke (commented):
Looks good, should we add a unit test for the `KeepManagedResourceObjects` function?

stoyanr:
@rfranzke 

> Looks good, should we add a unit test for the `KeepManagedResourceObjects` function?

Added such a test now, and while doing it found out that the initial implementation of the function could be improved - the initial `Get` was not needed since `TryUpdate` does it before updating.

rfranzke (approved):
/lgtm

rfranzke:
Great @stoyanr !

danielfoehrKn (approved):

PR #3028, opened 2020-10-23
Show Service events when LoadBalancer cannot be ensured

### LINKED ISSUE #2943: Show Service events when LoadBalancer cannot be ensured

/area ops-productivity
/kind enhancement
/priority normal

Currently when the vpn-shoot Service is not ensured, the reconciliation fails with generic error like:

```
Flow "Flow "Shoot cluster reconciliation" encountered task errors: [task "Waiting until the Kubernetes API server can connect to the Shoot workers" failed: retry failed with context deadline exceeded, last error: could not forward to vpn-shoot pod: port forward connection could not be established within five seconds]"]
```

We could add a step like `WaitUntilNginxIngressServiceIsReady` that waits until the vpn-shoot Service `.status.loadBalancer.ingress[]` has elements (when the KonnectivityTunnel feature gate is disabled).
Additionally we could also fetch events associated with the given Service and with the appropriate type and append the event(s) as part of error message.
In this way when the Service cannot be ensured because of quota issue/cloud provider credentials issue, then user would be able to see the reason.

ialidzhikov:

/area ops-productivity
/kind enhancement
/priority normal

Fixes #2943

**Release note**:

```noteworthy user
When a Shoot load balancer (vpn-shoot or addons-nginx-ingress-controller) cannot be ensured, gardenlet now fetches the involved object events (with type Warning) and adds them to the error message which is shown in the Shoot status. In this way users and operators will be able to identify better issues in which load balancer cannot be ensured because of invalid cloud provider credentials or another cloud provider issue.
```

### DIFF
--- a/pkg/operation/botanist/event.go
+++ b/pkg/operation/botanist/event.go
@@ -1,0 +1,35 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package botanist
+
+import corev1 "k8s.io/api/core/v1"
+
+// SortableEvents implements sort.Interface for []api.Event based on the Timestamp field.
+type SortableEvents []corev1.Event
+
+// Len implements sort.Interface.
+func (list SortableEvents) Len() int {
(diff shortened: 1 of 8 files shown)


pkg/operation/botanist/event.go
```diff
@@ -0,0 +1,32 @@
...
+
+func (list SortableEvents) Len() int {
+	return len(list)
+}
+
+func (list SortableEvents) Swap(i, j int) {
+	list[i], list[j] = list[j], list[i]
+}
+
+func (list SortableEvents) Less(i, j int) bool {
```
rfranzke: doc strings

pkg/operation/botanist/event_test.go
```diff
@@ -0,0 +1,70 @@
...
+					event3,
+				}
+			)
+
+			sort.Sort(botanist.SortableEvents(events))
+
+			Expect(events).To(Equal([]corev1.Event{event2, event3, event1}))
+		})
+	})
+
```
rfranzke:
```suggestion
```

rfranzke (commented):
Love it, thanks for this great PR!
Only nits
/lgtm

rfranzke (approved):
/lgtm

danielfoehrKn (approved):
nice - this will hopefully lower the burden on ops :)

PR #3029, opened 2020-10-23
Fix update of the Shoot hibernation status after successful reconciliation

### LINKED ISSUE #2960: Mid-air updates lead to shoot deletion failures

/area robustness
/kind bug
/priority normal

**What happened**:
A cluster which had been woken up from hibernation was deleted and but stuck in the deletion flow at the `Initializing connection to Shoot` step.

The cluster to be deleted successfully.

1. Hibernate existing or new cluster
2. Change `.spec` of shoot, e.g. a worker group
3. Wait for the reconciliation to beginn
4. Change `.spec.hibernation.enabled  = false` **during reconciliation**
5. Immediately delete shoot

After the first reconciliation you should notice that the shoot has `.status.hibernated: "false"` even though it's never woken up. As consequence, the deletion fails because Gardener never tries to wake up the control plane.

- Gardener version: v1.11.1

/cc @MartinWeindel

danielfoehrKn:

/area quality
/kind bug
/priority normal

Fix update of the Shoot hibernation status after successful reconciliation by deriving the Shoot hibernation status from the Cluster resource instead of the Shoot resource (where the `Spec.Hibernation` could have been changed during the reconciliation).
Fixes #2960

**Release note**:

```improvement operator
Fixes a bug in the Shoot reconciliation that could lead to failed Shoot deletions when the Hibernation spec has been changed during an active reconciliation.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control.go
@@ -28,6 +28,7 @@ import (
 	"github.com/gardener/gardener/pkg/client/kubernetes"
 	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/keys"
 	"github.com/gardener/gardener/pkg/controllerutils"
+	gardenerextensions "github.com/gardener/gardener/pkg/extensions"
 	"github.com/gardener/gardener/pkg/logger"
 	"github.com/gardener/gardener/pkg/operation"
 	"github.com/gardener/gardener/pkg/operation/common"
@@ -450,7 +451,7 @@ func (c *Controller) reconcileShoot(logger *logrus.Entry, shoot *gardencorev1bet
 	}
 
 	c.recorder.Event(o.Shoot.Info, corev1.EventTypeNormal, gardencorev1beta1.EventReconciled, "Reconciled Shoot cluster state")
-	o.Shoot.Info, err = c.updateShootStatusOperationSuccess(ctx, gardenClient.GardenCore(), o.Shoot.Info, o.Seed.Info.Name, operationType)
+	o.Shoot.Info, err = c.updateShootStatusOperationSuccess(ctx, gardenClient.GardenCore(), o.Shoot.Info, o.Shoot.SeedNamespace, o.Seed.Info.Name, operationType)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -546,10 +547,12 @@ func (c *Controller) updateShootStatusOperationStart(ctx context.Context, g gard
...
 
-func (c *Controller) updateShootStatusOperationSuccess(ctx context.Context, g gardencore.Interface, shoot *gardencorev1beta1.Shoot, seedName string, operationType gardencorev1beta1.LastOperationType) (*gardencorev1beta1.Shoot, error) {
+func (c *Controller) updateShootStatusOperationSuccess(ctx context.Context, g gardencore.Interface, shoot *gardencorev1beta1.Shoot, shootSeedNamespace string, seedName string, operationType gardencorev1beta1.LastOperationType) (*gardencorev1beta1.Shoot, error) {
 	var (
-		now         = metav1.NewTime(time.Now().UTC())
-		description string
+		now          = metav1.NewTime(time.Now().UTC())
+		description  string
+		err          error
+		isHibernated bool
```
rfranzke:
```suggestion
		isHibernated = gardencorev1beta1helper.HibernationIsEnabled(shoot)
```

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -560,10 +563,19 @@ func (c *Controller) updateShootStatusOperationSuccess(ctx context.Context, g ga
 		description = "Shoot cluster has been successfully deleted."
 	}
 
+	if len(shootSeedNamespace) > 0 {
+		isHibernated, err = c.isHibernationActive(ctx, shootSeedNamespace, seedName)
+		if err != nil {
+			return nil, fmt.Errorf("error updating Shoot (%s/%s) after successful reconciliation: %v", shoot.Namespace, shoot.Name, err)
+		}
+	} else {
+		isHibernated = gardencorev1beta1helper.HibernationIsEnabled(shoot)
```
rfranzke:
```suggestion
```

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -560,10 +563,19 @@ func (c *Controller) updateShootStatusOperationSuccess(ctx context.Context, g ga
 		description = "Shoot cluster has been successfully deleted."
 	}
 
+	if len(shootSeedNamespace) > 0 {
+		isHibernated, err = c.isHibernationActive(ctx, shootSeedNamespace, seedName)
+		if err != nil {
+			return nil, fmt.Errorf("error updating Shoot (%s/%s) after successful reconciliation: %v", shoot.Namespace, shoot.Name, err)
```
rfranzke:
```suggestion
			return nil, fmt.Errorf("error updating Shoot (%s/%s) after successful reconciliation when checking for active hiberation: %v", shoot.Namespace, shoot.Name, err)
```

rfranzke (changes requested):
Thanks, looks great, only nits

pkg/gardenlet/controller/shoot/shoot_control.go
danielfoehrKn: This unfortunately causes `gardencorev1beta1helper.HibernationIsEnabled(shoot)` to always be executed even though it is not required if `len(shootSeedNamespace) > 0`. Thats why it is in an if-else block.

rfranzke: Yes, but it's not a costly operation at all, so doing it here will save some nesting below (improving readability slightly). I don't insist on this, but I generally think it's a better readable style to do it like this in case the operation is cheap.

danielfoehrKn: ok sure not problem - its indeed only a minimal gain. I though about it as well - but I see it is an issue for you so I ll change it.

rfranzke (approved):
/lgtm
/second-opinion
/invite @timuthy

timuthy (approved):
/lgtm
Thanks for taking care!

PR #3030, opened 2020-10-23
Enhance error code mapping

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal

More error code mappings

**Release note**:

```improvement operator
NONE
```


timebertt (approved):

pkg/apis/core/v1beta1/helper/errors.go
```diff
@@ -53,9 +53,9 @@ var (
...
 	insufficientPrivilegesRegexp = regexp.MustCompile(`(?i)(AccessDenied|OperationNotAllowed)`)
-	dependenciesRegexp           = regexp.MustCompile(`(?i)(PendingVerification|Access Not Configured|accessNotConfigured|DependencyViolation|OptInRequired|DeleteConflict|Conflict|inactive billing state|ReadOnlyDisabledSubscription|is already being used|InUseSubnetCannotBeDeleted|VnetInUse|timeout while waiting for state to become|InvalidCidrBlock|already busy for|InsufficientFreeAddressesInSubnet|InternalServerError|RetryableError|Future#WaitForCompletion: context has been cancelled)`)
+	dependenciesRegexp           = regexp.MustCompile(`(?i)(PendingVerification|Access Not Configured|accessNotConfigured|DependencyViolation|OptInRequired|DeleteConflict|Conflict|inactive billing state|ReadOnlyDisabledSubscription|is already being used|InUseSubnetCannotBeDeleted|VnetInUse|timeout while waiting for state to become|InvalidCidrBlock|already busy for|InsufficientFreeAddressesInSubnet|InternalServerError|RetryableError|Future#WaitForCompletion: context has been cancelled|internalerror|internal server error)`)
```
dguendisch, line 56: quick question: don't all these "internal server errors" lead to false positives? They sound like you cannot derive anything from them, or is it just "experience" that they are more or less always returned on user-/dependency-errors?

rfranzke: They are internal errors on the infrastructure provider. We don't have a dedicated error code for them (yet?), and atm I don't want to include them, hence, we classify them as "INFRA_DEPENDENCIES" for now

dguendisch: Ah my bad sorry, I misinterpreted what the `dependenciesRegexp` is used for, sure it denotes `ErrorInfraDependencies`, so it's correct 👍

PR #3034, opened 2020-10-23
Adapt terraformer library

timebertt:

/area quality robustness
/kind enhancement
/priority normal

This PR changes the terraformer library in the following ways:
- minor cleanups (first commit)
- adapt the library to terraformer v2, ref https://github.com/gardener/terraformer/pull/48 (second commit)

terraformer v2 is still in the works and not yet released. But this change here is made in a backwards-compatible manner.
Controllers using it, have to set `UseV2(true)` in order to get the PodSpec compatible to v2, otherwise the old PodSpec is used.

Part of https://github.com/gardener/terraformer/issues/40

**Release note**:

```noteworthy developer
The [terraformer library](https://github.com/gardener/gardener/tree/master/extensions/pkg/terraformer) can now be used to deploy `terraformer@v2` Pods. To enable this, you have to set `UseV2(true)` on the `Terraformer` instance, otherwise the old `PodSpec`, that is compatible with v1, will be used.
```

### DIFF
--- a/extensions/pkg/terraformer/config.go
+++ b/extensions/pkg/terraformer/config.go
@@ -65,6 +65,18 @@ func (t *terraformer) SetDeadlinePod(d time.Duration) Terraformer {
 	return t
 }
 
+// UseV2 configures if it should use flags compatible with terraformer@v2.
+func (t *terraformer) UseV2(v2 bool) Terraformer {
+	t.useV2 = v2
+	return t
+}
+
+// SetLogLevel sets the log level of the Terraformer pod. It only takes effect when UseV2 is set to true.
+func (t *terraformer) SetLogLevel(level string) Terraformer {
+	t.logLevel = level
+	return t
+}
+
 // InitializerConfig is the configuration about the location and naming of the resources the
 // Terraformer expects.
 type InitializerConfig struct {
--- a/extensions/pkg/terraformer/config.go
+++ b/extensions/pkg/terraformer/config.go
@@ -21,12 +21,13 @@ import (
 	"strings"
 	"time"
 
-	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


rfranzke (dismissed):
/lgtm

rfranzke (dismissed):

extensions/pkg/terraformer/terraformer.go
```diff
@@ -21,19 +21,21 @@ import (
...
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
+	"github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/utils/retry"
```
ialidzhikov, line 38: Isn't the convention we follow about imports the following - group the imports from the given project as second import group and the third-party imports - as third import group?

extensions/pkg/terraformer/config.go
```diff
@@ -64,6 +65,18 @@ func (t *terraformer) SetDeadlinePod(d time.Duration) Terraformer {
 	return t
 }
 
+// UseV2 configures if it should use flags compatible with terraformer@v2.
+func (t *terraformer) UseV2(v2 bool) Terraformer {
+	t.useV2 = v2
+	return t
+}
+
+// SetLogLevel sets the log level of the Terraformer pod.
```
ialidzhikov:
```suggestion
// SetLogLevel sets the log level of the Terraformer pod. It takes effect only when `UseV2` is set to true.
```

ialidzhikov:
> - minor cleanups and add RestartPolicy=Never (wonder why we didn't have this before, was there some issue with it?) (first commit)

It is here https://github.com/gardener/gardener/blob/v1.11.3/extensions/pkg/terraformer/terraformer.go#L342.

extensions/pkg/terraformer/terraformer.go
timebertt: I guess, the agreement was that with #2655 we just accept everything, that is accepted by `goimports`, which should be the case here.

timebertt:
> > * minor cleanups and add RestartPolicy=Never (wonder why we didn't have this before, was there some issue with it?) (first commit)
> 
> It is here https://github.com/gardener/gardener/blob/v1.11.3/extensions/pkg/terraformer/terraformer.go#L342.

Oops. Missed that in the diff.

ialidzhikov (approved):
/lgtm

rfranzke (approved):

PR #3035, opened 2020-10-23
support providerConfig for shooted seeds

MartinWeindel:

/area control-plane
/kind enhancement
/priority normal

The `providerConfig` of the Seed resource can be specified in the `shoot.gardener.cloud/use-as-seed` annotation of the shoot to be used as seed.

Fixes #

**Release note**:

```noteworthy operator
It is now possible to submit provider configuration (for the `.spec.provider.providerConfig` field in `Seed` resources) when creating shooted seeds. Please consult [the documentation](https://github.com/gardener/gardener/blob/v1.12.0/docs/usage/shooted_seed.md) for more information
```

### DIFF
--- a/pkg/apis/core/v1beta1/helper/helper.go
+++ b/pkg/apis/core/v1beta1/helper/helper.go
@@ -20,6 +20,10 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/json"
+
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	versionutils "github.com/gardener/gardener/pkg/utils/version"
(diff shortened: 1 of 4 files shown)


rfranzke:
We definitely need to get rid of this annotation... 😉

rfranzke (dismissed):

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -414,6 +424,72 @@ func parseShootedSeedBackup(settings map[string]string) (*gardencorev1beta1.Seed
 	return backup, nil
 }
 
+func parseShootedSeedProviderConfig(settings map[string]string) (*runtime.RawExtension, error) {
```
timebertt, line 428: Can you reuse `unstructured.SetNestedField` and `json.Marshal` instead of reimplementing this stuff on our side?

I think this func should look something like this:
```go
func parseShootedSeedProviderConfig(settings map[string]string) (*runtime.RawExtension, error) {
	// reconstruct providerConfig structure
	providerConfig := map[string]interface{}{}

	for k, v := range settings {
		if strings.HasPrefix(k, "providerConfig.") {
			if err := unstructured.SetNestedField(providerConfig, v, strings.Split(k, ".")[1:]...); err != nil {
				return nil, err
			}
		}
	}

	jsonStr, err := json.Marshal(providerConfig)
	if err != nil {
		return nil, err
	}

	return &runtime.RawExtension{
		Raw: jsonStr,
	}, nil
}
```

timebertt (changes requested):

timebertt (approved):
/lgtm

rfranzke (approved):

PR #3036, opened 2020-10-23
Forbid functionless DNS providers

timuthy:

/area usability
/kind api-change
/kind cleanup
/priority normal

This PR forbids the creation of non-primary DNS providers that are functionless, i.e. w/o `.type` or `.secretName` specification. Earlier those providers were removed during create or update requests.

Functionless providers were allowed for backwards compatibility reasons and to detect such shoots, a warning was printed to the G-API Server log.

**Release note**:

```action user
DNS providers not used for the shoot's `.spec.dns.domain`, also known as additional, non-primary providers, must now be specified with `.type` and `.secretName`. Such providers were previously removed automatically from the shoot specification during create or update requests.
```

### DIFF
--- a/plugin/pkg/shoot/dns/admission.go
+++ b/plugin/pkg/shoot/dns/admission.go
@@ -196,9 +196,8 @@ func (d *DNS) Admit(ctx context.Context, a admission.Attributes, o admission.Obj
 		if shoot.Spec.SeedName == nil {
 			return nil
 		}
-		if oldShoot.Spec.SeedName != nil {
-			removeFunctionlessDNSProviders(shoot.Name, shoot.Namespace, shoot.Spec.DNS)
-			return nil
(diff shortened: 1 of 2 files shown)


plugin/pkg/shoot/dns/admission.go
```diff
@@ -225,35 +224,29 @@ func (d *DNS) Admit(ctx context.Context, a admission.Attributes, o admission.Obj
...
-
-		providers = append(providers, provider)
-		continue
 	}
-
-	dns.Providers = providers
+	return nil
 }
 
+// checkPrimaryDNSProvider checks if the shoot uses a default domains and returns an error
```
danielfoehrKn:
```suggestion
// checkPrimaryDNSProvider checks if the shoot uses the default domain and returns an error
```

danielfoehrKn (dismissed):
/lgtm

rfranzke (approved):
/lgtm

Issue #2942, opened 2020-10-05
Graduate APIServerSNI (+ ManagedIstio) feature gate


timuthy:
**How to categorize this issue?**

/area control-plane
/area cost
/area scalability
/kind enhancement
/priority normal

**What would you like to be added**:
The `APIServerSNI` and `ManagedIstio` (usually used together) should be graduated to **beta** after we have collected sufficient experience and test results on regularly tested and busy landscapes. Any potential issues and blockers graduating this feature can be referenced or documented here.

Steps:
- [x] Enable feature gates on development landscape 
- [x] Enable feature gates on landscape featuring extended automated tests
- [x] Enable feature gates on landscape with stakeholder workload (large landscape)
- [x] Enable Prometheus metrics and Grafana Dashboards for `ManagedIstio`
- [x] Add `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` environment variables to all Gardener components to avoid going to through the `apiserver-proxy`
- [x] Replace `nginx` with `envoy` to have the same proxy on both the client and server side. Enovy supports [upstream proxy protocol transport](https://www.envoyproxy.io/docs/envoy/v1.16.0/intro/arch_overview/other_features/ip_transparency.html?highlight=envoy%20filters%20listener%20proxy_protocol#proxy-protocol)
- [x] Add Prometheus scraping for `apiserver-proxy`
- [x] Expose `apiserver-proxy`'s `envoy` metrics` available for scraping.
- [x] Allow for end-users to specify the `istio-ingress-gateway` namespace and name.
- [x] Add `PodDisruptionBudget` for `istio-ingress-gateway` and `istiod`
- [x] Add a `MutatingWebhookConfiguration` for `Pods` in the Shoot cluster, adding `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT`. It'll run as a sidecar to the API Server (to not have any extra latency to all `CREATE` requests) with `FailurePolicy: Ignore` and very low `timeoutSeconds` - (1 or 2)
- [x] Set KEEPALIVE on both `apiserver-proxy` and `istio-ingressgateway`
- [x] ~Change the port of `kube-apiserver` gateways to `9443` to prevent issues if `istio-ingressgateway` service has port `443` is removed.~
- [x] ~Preserve original source IP~ GA feature

/cc @mvladev

**Why is this needed**:
Both feature gates are an essential milestone to host control planes in a cost efficient and scalable manner.

[...]

PR #3041, opened 2020-10-24
Replace nginx with envoy

mvladev:

/area networking
/kind enhancement
/priority normal

`nginx` is replaced with `envoy` for `apiserver-proxy`. This allows to have the same proxy for both the client and the server.

PROXY Protocol v2 is now used by `apiserver-proxy`. 

Part of #2942

**Release note**:

```improvement operator
`nginx` is replaced with `envoy` for `apiserver-proxy`. This allows to have the same proxy for both the client and the server.
```

```improvement operator
`apiserver-proxy` now uses `PROXY Protocol` v2 when talking to the upstream kube-apiserver.
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -246,9 +246,9 @@ images:
 
 # API Server SNI
 - name: apiserver-proxy
-  sourceRepository: github.com/nginxinc/docker-nginx
-  repository: nginx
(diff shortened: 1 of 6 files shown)


rfranzke (approved):
/lgtm

PR #3042, opened 2020-10-26
Default kubeReserved PID to 20k

BeckerMax:
Protect kubelet from being starved out of PIDs by workload.

/area availability
/area security
/priority normal

Currently a kubelet, the container runtime or other system daemons can be starved out of processes if a pod on a node creates excessive amount of processes.
Setting this flag will guarantee that all processes outside of pods will have 20k pids guaranteed.

I checked on large Shooted Seeds and PIDs used in the system.slice(includes container runtime and kubelet) were ~4k so I multiplied it by 5 to be on the safe side.

Verified that it works:

```
# Before:
$ cat /sys/fs/cgroup/pids/kubepods/pids.max
4194304

# After (20k less)
$ cat /sys/fs/cgroup/pids/kubepods/pids.max
4174304

# Total available pids:
cat /proc/sys/kernel/pid_max
4194304
```

This change does not require a restart of the nodes or the pods. Only the kubelet gets restarted.

I also removed some tests in the defaults_test that I did not think added meaningful test coverage.

**Release note**:
```noteworthy user
Defaults reserved PIDs for kubelet and container runtime to 20k.
```

### DIFF
--- a/hack/api-reference/core.md
+++ b/hack/api-reference/core.md
@@ -4263,7 +4263,7 @@ KubeletConfigReserved
 <em>(Optional)</em>
 <p>KubeReserved is the configuration for resources reserved for kubernetes node components (mainly kubelet and container runtime).
 When updating these values, be aware that cgroup resizes may not succeed on active worker nodes. Look for the NodeAllocatableEnforced event to determine if the configuration was applied.
-Default: cpu=80m,memory=1Gi</p>
+Default: cpu=80m,memory=1Gi,pid=20k</p>
 </td>
 </tr>
 <tr>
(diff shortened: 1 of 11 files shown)


rfranzke (dismissed):
/lgtm

ialidzhikov (commented):
Ref https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-master/jobs/master-pull-request-job/builds/288

```
> Generate / Vendor Check
>> make generate
 M hack/api-reference/core.md
 M pkg/apis/core/v1alpha1/generated.proto
 M pkg/apis/core/v1beta1/generated.proto
 M pkg/openapi/openapi_generated.go
make: *** [Makefile:160: check-generate] Error 1
```

timebertt (approved):
/lgtm

rfranzke (approved):

PR #3044, opened 2020-10-26
Expose apiserver envoy proxy metrics

mvladev:
On all Nodes `16910` (node-exporter port + 1) is now used for metrics scraping at `/stats/prometheus`

/area networking
/area monitoring
/kind enhancement
/priority normal

Allows for Prometheus to scrap this component (I'll do it in a separate PR).

Part of #2942

Envoy's admin interface can be used to make some configuration changes by just calling the endpoint. Normally in container it'll listen on `127.0.0.1`, however in this case it's listening on the host network. To avoid the security issue, the admin management listens on unix socket and then it's exposed via http listener.

**Release note**:

```action user
`apiserver-proxy` now listens on port `16910` on all Nodes if `APIServerSNI` feature gate is enabled. It causes a conflict with workloads with `hostNetwork: true` and listening on `0.0.0.0:16910`, so it's required to change your workload's bind port.
```

### DIFF
--- a/charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-daemonset.yaml
+++ b/charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-daemonset.yaml
@@ -26,6 +26,7 @@ spec:
         role: apiserver-proxy
         networking.gardener.cloud/to-dns: allowed
         networking.gardener.cloud/to-apiserver: allowed
+        networking.gardener.cloud/from-seed: allowed
     spec:
       serviceAccountName: apiserver-proxy
       priorityClassName: system-cluster-critical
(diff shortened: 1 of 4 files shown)


rfranzke (approved):
/lgtm

mvladev:
/hold

I need to test something.

mvladev:
I've increased the CPU and Memory limits with and with `400req/s` for `api/v1/serviceaccounts` we get a about `70m` cou and `30Mi` memory usage.

rfranzke (approved):
/lgtm

Thanks @mvladev

wyb1 (commented):
I would expect after port forwarding the `apiserver-proxy` on port 16910 and curling `localhost:16910/metrics` that I would get metrics, but instead I get nothing.

mvladev:
> I would expect after port forwarding the apiserver-proxy on port 16910 and curling localhost:16910/metrics that I would get metrics, but instead I get nothing.

They are exposed at `/stats/prometheus`. I can re-write the `:path`, but I would rather not. 

Should they really be at `/metrics`?

wyb1:
> They are exposed at /stats/prometheus.

Sorry, I missed that. /lgtm

PR #3053, opened 2020-10-28
Fix node-exporter configuration

wyb1:

/area monitoring
/kind bug
/priority normal

* Use default values for ignored fs types.
* Fix ignored mount points
* Simplify mountpoints

On an xfs filesystem the reported size of the root volume was incorrect. This
change fixes the issue.

Fixes #

**Release note**:

```improvement user
Node exporter properly reports filesystem size for operating systems that use an xfs filesystem
```


istvanballok (approved):
/lgtm

rfranzke (approved):
/lgtm

PR #3065, opened 2020-10-29
Support custom registry mirror

### LINKED ISSUE #3064: Support custom registry-mirror

/area cost
/kind enhancement
/priority normal

An ability to configure a custom registry mirror to use as pull through cache for images from DockerHub.

The docker daemon would need an environment variable like this: `DOCKER_OPTS=--registry-mirror <URL>`

And containerd would need this section in its configuration file:

```
[plugins.cri.registry.mirrors]
  [plugins.cri.registry.mirrors."docker.io"]
    endpoint = ["<uri>"]
```

**Why is this needed**:
Docker limits pull requests starting from 2nd November on: https://docs.docker.com/docker-hub/download-rate-limit/
This means that unauthenticated pulls are rate limited based on the source IP.

With `provider: metal` at metal-stack we hide all cluster traffic behind a single IP.
Therewith we will be hit by this rate limit with clusters > two nodes.

It`s also a cost saving for all providers when a "local" registry mirror is used to cache images.

mwindower:

/area cost
/kind enhancement
/priority normal

Fixes #3064

**Release note**:

```improvement operator

```

### DIFF
--- a/charts/seed-operatingsystemconfig/original/templates/docker/_docker-registry-mirror.dropin
+++ b/charts/seed-operatingsystemconfig/original/templates/docker/_docker-registry-mirror.dropin
@@ -1,0 +1,8 @@
+{{- define "docker-registry-mirror" -}}
+- name: docker.service
+  dropIns:
+  - name: 10-registry-mirror.conf
+      content: |
+      [Service]
+      Environment="DOCKER_OPTS=--registry-mirror {{  }}"
+{{- end -}}
--- a/charts/seed-operatingsystemconfig/original/templates/docker/_docker-registry-mirror.dropin
+++ b/charts/seed-operatingsystemconfig/original/templates/docker/_docker-registry-mirror.dropin
@@ -4,5 +4,5 @@
   - name: 10-registry-mirror.conf
       content: |
       [Service]
-      Environment="DOCKER_OPTS=--registry-mirror {{  }}"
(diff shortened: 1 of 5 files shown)


ghost:
Thank you @mwindower for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

timebertt:
As @danielfoehrKn already stated in this issue, the CloudProfile is not the right place to configure a docker registry mirror. Probably this should be rather handled by a container runtime extension controller.
Yes, we haven't yet externalized the default container runtime handling and "cleanly" implemented the extension contract for this, so this isn't possible yet.

As a short-term workaround, the following might be possible options for you:
- deploy `DaemonSets` to the shoots which manipulate the default docker and containerd config
- use a webhook for `OperatingSystemConfigs` that manipulates the unit / config files

rfranzke:
I agree with @timebertt. I would rather try to prevent adding fields into the already beta-promoted Gardener resources for things where we already know that they won't last/are not the right place. Would one of the proposed  workarounds be a solution for you?

mwindower:
Yes sure, we'll adapt our OS extension directly or use the webhook approach.
I'll close the PR, but it's nice to hear, that CRI configuration will be externalized in the future :-)

Issue #17, opened 2018-01-13
ETCD Backup & Restore


gardener-robot-ci-1:
[kubernetes-jenkins](https://github.com/kubernetes-jenkins):
_Wednesday Jul 19, 2017 at 18:44 GMT_
_Originally opened as https://git.removed/kubernetes-attic/garden-operator/issues/31_

----

# Story
* As provider I want etcd backups for my shoot clusters, so that I can restore them should they be lost.
* Later: As provider I want etcd backups to be restored automatically should the PV be definitely lost or the current etcd can't start from the current state.   

# Motivation
Business continuity.

# Acceptance Criteria
- [x] Regular etcd backups are taken by the garden operator for all shoot clusters and are made known to the Gardener UI via TPRs/CRDs
  - [x] AWS
  - [x] Azure
  - [ ] GCP
  - [ ] OpenStack

# Resources
* Take a look at the [CoreOS etcd operator](https://github.com/coreos/etcd-operator) if you can
* Maybe [Heptio Ark](https://blog.heptio.com/announcing-two-new-heptio-open-source-projects-heptio-ark-and-heptio-sonobuoy-7cef88a06f8) (disaster recovery) may be interesting to check out
* See also https://github.com/coreos/etcd/blob/master/Documentation/op-guide/recovery.md#restoring-a-cluster

# Open Questions
How frequent can we backup the data? Can we replicate the data on a lower level with a sidecar deployment or would we need an etcd cluster then? Is there something like point in time recovery for etcd?

# Definition of Done
- [ ] Knowledge is distributed: Have you spread your knowledge in pair programming/code review?
- [ ] Unit Tests are provided: Have you written automated unit tests or added manual NGPTT tickets?
- [ ] Integration Tests are provided: Have you written automated integration tests?
- [ ] Minimum API exposure: If you have added public API, was it really necessary/is it minimal?
- [ ] Operations guide: Have you updated the [operations guide](https://git.infra.removed/kubernetes/kube-docs/wiki/Operations%20Guide)?

:information_source: Migrated from Jira issue [KUBE-158](https://jtrack.removed/browse/KUBE-158)

vlerenc:
Replaced with https://github.com/gardener/etcd-backup-restore/issues/1.

PR #3066, opened 2020-10-29
[GEP-7] Update proposal for ETCD leader election

plkokanov:

/area control-plane-migration
/area documentation
/kind enhancement
/priority normal

Updates the proposal for ETCD based leader election.

Part of #1631 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/docs/proposals/07-shoot-control-plane-migration.md
+++ b/docs/proposals/07-shoot-control-plane-migration.md
@@ -202,43 +202,99 @@ The ShootState synchronization controller will become part of the gardenlet. It 
 1. The Shoot's namespace in __Source Seed__ is deleted.
 
 
-### Leader Election and Control Plane Termination
+### Migrating Backups
 
-During migration "split brain" scenario must be avoided. This means that a Shoot's control plane in the __Source Seed__ must be scaled down before it is scaled up in the __Destination Seed__.
+For security reasons a `Seed` should only have access to its own backup bucket, meaning that the `etcd-backup-restore` sidecar responsible for saving and restoring the ETCD data should only be able to access the backup bucket of the `Seed` it currently operates in. Therefore, during Control Plane Migration a mechanism is needed to copy the ETCD backup from the __Source Seed__'s backup bucket to that of the __Destination Seed__. To that end `etcd-backup-restore` is extended with a new subcommand - `migrate` which runs in the Seed and does the following:
+
+1. Creates a backup directory for the `Shoot` cluster in the __Destination Seed__'s backup bucket if it does not already exist.
+1. Waits for some time (until a preconfigured timeout) for the `etcd-backup-restore` sidecar in the __Source Seed__ to finish saving a final snapshot.
+1. Writes into a "header" file inside the `Shoot`'s backup directory in the __Source Seed__'s backup bucket that copying of data __from__ this directory is in process, the timestamp when the copy operation began, a timestamp indicating when migration was actually triggered and the name of the __Destination Seed__
+1. Writes into a "header" file inside the `Shoot`'s backup directory in the __Destination Seed__'s backup bucket that copying of data __to__ this directory is in process, the timestamp of when the copy operation began and a timestamp indicating when migration was actually triggered.
+1. When the `etcd-backup-restore` sidecar in the __Destination Seed__ tries to restore the data it checks the "header":
+  - If data is still being copied, it does nothing and dies.
+  - If the data has finished copying, it restores the ETCD from it.
+1. When the copy operation has finished successfully, the "header" files in both the source and destination backup directories are updated accordingly.
+1. Subsequent executions of the command (in case the operation is repeated as part of the Control Plane Migration flow) check that the data has already been copied and do nothing
+
+*Note:* the migration timestamp might be redundant as it will closely correspond to the timestamp indicating the start of the copy operation.
+*Note:* Any additional checks and steps that have to be taken by the `etcd-backup-restore` sidecar in the __Source Seed__ in case there are problems with the Seed are described in more detail in the __Leader Election__ section below.
+
+The `migrate` subcommand requires secrets for both the backup buckets of the __Source__ and __Destination__ Seeds. The latter is already taken care of by the Backup Entry controller. For the former - a temporary secret needs to be created and provided to the Seed cluster. The exact mechanism of how this will be done is to be discussed in more details in [Dedicated secrets per seed backup bucket](https://github.com/gardener-security/security-backlog/issues/17) and [Eliminate all trust credentials between hyperscalers and Gardener](https://github.com/gardener-security/security-backlog/issues/25)
+
+The ETCD Druid is extended so that it starts `etcd-backup-restore migrate` as a job when it is notified that Control Plane Migration has been triggered. This notification happens via a `gardener.cloud/operation=restore` annotation in the `etcd` resource and (if necessary) a few additional fields indicating the cluster owner and when migration was triggered:
+```
+spec:
+  ownerToken: <seed_name>
+  ownerTimestamp: "2018-11-18T15:14:57Z"
+```
+The ETCD Druid will automatically reconcile the `etcd` resource as normal after it has finished with the `etcd-backup-restore migrate` operation.
+### Leader Election
 
 **Note**: This section is still under discussion. The plan is to first implement the `ShootState` and modify the reconciliation flow and extensions accordingly. Additionally multiple scenarios need to be considered depending on the reachability to the Garden cluster from the __Source Seed__ components, the __Source Seed's__ API server from the Garden cluster and the __Source Seed's__ API server from the controllers running on the seed. **The initial implementation will only cover the case where everything is running**.
 
-Extension controllers do not need leader election functionality because they only reconcile the extension resources if the reconcile operation annotation is specified in the resource. Since this is set only during a reconciliation triggered by the Garden Controller Manager it cannot happen during migration.
+During migration "split brain" scenario must be avoided. This means that a Shoot's control plane in the __Source Seed__ must be scaled down before it is scaled up in the __Destination Seed__.
 
-For other controllers in the controlplane (e.g MCM, etcd backup restore, kube apiserver) leader election has to be implemented. We plan to introduce gardenlet soon therefore the garden cluster is expected to be reachable from all seeds.
+### Garden cluster and __Source Seed__ are healthy and there are no network problems
 
-Another flag might be needed to tell the gardenlet in the __Destination Seed__ that the control plane in the __Source Seed__ has been scaled down and the Shoot reconciliation can begin.
+If both the Garden cluster and __Source Seed__ cluster are healthy, the `gardenlet` after checking the `spec.seedName`, can directly scale down the Shoot's control plane as part of the migration flow.
 
-Horizontal Pod Autoscalers also need to be considered and removed.
+### If components in the __Source Seed__ cannot reliably read who the leader is from the Garden cluster
+We consider the following cases in which the `gardenlet` and the extension controllers in the __Source Seed__ cannot fetch information about who the leader is:
 
-#### Garden cluster and __Source Seed__ are healthy and there are no network problems
+To avoid possible reconciliations by Control Plane Components and Extension Controllers in both the __Destination Seed__ and __Source Seed__ we use the following leader election procedure (split into two parts - one for components in the `Shoot`'s control plane which depend on its APIServer and one for extension controllers which do not depend on it):
 
-If both the Garden cluster and __Source Seed__ cluster are healthy, the Garden Controller Manager (or gardenlet after checking the `spec.seedName`) can directly scale down the Shoot's control plane as part of the migration flow.
+#### ETCD Backup Based Leader Election
 
-#### If components in the __Source Seed__ cannot reliably read who the leader is from the Garden cluster
+When migration is triggered:
+1. The ETCD Druid in the __Destination Seed__ is informed about the ownership of the ETCD cluster and that Control Plane Migration has been triggered
+2. It triggeres `etcd-backup-restore migrate` which first checks to see whether a finished snapshot has been made by the `etcd-backup-restore` sidecar in the __Source Seed__ and then begins the copy operation. If no snapshot is created within a timeout it will copy the latest snapshot.
+4. The `etcd-backup-restore` sidecar initially (during startup) and continuously (when uploading backups) checks to see if a copy operation to a new bucket has started or not.
+    1. If it finds that a copy operation __from__ its bucket is in process or has been finished, it determines that it is in the __Source Seed__ and that control plane migration has been triggered.
+    2. If it finds that a copy operation __to__ its bucket is in process, it determines that it is in the __Destination Seed__ and that control plane migration has been triggered. Once it finds that the copy operation has completed successfully it restores the ETCD data.
+    3. If the ownership information matches and thare is no indication that a copy operation is in process, `etcd-backup-restore` continues to work exactly the way it does today.
+5. When the `etcd-backup-restore` sidecar in the `Shoot's` ETCD pod determines that migration is on and that it is on the __Source Seed__, it stops taking snapshots (full and delta snapshots) and terminates with an error
+    1. This should also cut-off the client requests because the `readinessProbe` of the main ETCD container probes the `etcd-backup-restore` sidecar.
+    1. Just to be sure, the health check in the backup-restore sidecar (used by the main ETCD container's `readinessProbe`) is enhanced to also compare the ownership information passed by `etcd-druid` with what is in the backup. This should ensure that if the main etcd container `readinessProbe` will continue to fail if the `etcd-backup-restore` sidecar comes back up.
+    1. Please note that the etcd container itself might not die here. But the client requests to the container will be cut off.
+5. When the `etcd-backup-restore` sidecar in the `Shoot's` ETCD pod determines that migration is on and that it is on the __Destination Seed__, it first waits for the migration process to finish
+    1. It then checks the available backups and restores the ETCD data from the snapshot with the latest timestamp (older than the migration timestamp).
+7. Once the ETCD container stops accepting the client requests in the __Source Seed__, all components which depend on it (e.g. the Kube APIServer, Machine Controller Manager and so on) will also stop working.
 
-Currently we have come up with two ideas to handle this case:
 
-**DNS leader election:** A DNS TXT entry with TTL=60s and value seed='__Source Seed__' is used. The record is created and maintained by the Gardener Controller Manager (by using the DNS Controller Manager and its DNSEntry resource). When a control plane migration is detected, the Gardener Controller Manager changes the value of the DNS Entry to seed='__Destination Seed__' and waits for 2*TTL + 1 = 121 seconds to ensure that the change is propagated to all controllers in the old seed. We rely on the fact that DNS is highly available (100% for AWS Route53) and that control plane components in the __Source Seed__ can see the changes.
+#### Stopping extension controllers
+An additional step needs to be taken to prevent the extension controllers from reconciling their resources if they were stuck in a crashloop backoff triggered by a previous reconciliation, which can happen in the following casess:
+
+1. `Gardenlet` cannot connect to the Garden APIServer
+
+    In this case the  `gardenlet` cannot fetch information about `Shoots` and therefore does not know if control plane migration has been triggered. The `gardenlet` will not trigger new reconciliations, however extension controllers can still be reconciling their resources if they are stuck in an error backoff from a previous reconciliation.
+
+2. `Gardenlet` cannot connect to the `Seed`'s APIServer.
+
+    In this case the `gardenlet` knows if migration has been triggered but it will not start shoot migration/reconciliation as it will first check the `Seed` conditions and try to update the `Cluster` resource both of will fail. Extension controllers can still have a connection to the `Seed`'s APIServer (if they are not running where the `gardenlet` is running - different node or cluster) which means that due to previously triggered reconciliations (and/or crashloop backoff) they could reconcile their resources.
+
+
+3. The `Seed` components (etcd-druid, extension controllers, etc) cannot connect to the `Seed`'s APIServer.
+
+    In this case the extension controllers will not be able to reconcile their resources as they cannot fetch them from the `Seed`'s APIServer. When the connection to the APIServer comes back, the controllers might be stuck in a backoff from a previous reconciliation or the resources could still be annotated with `gardener.cloud/operation=reconcile`. This could lead to a race condition depending on who has a chance to `update` or `get` the resources first - `gardenlet` or the `Seed`'s APIServer. If `gardenlet` manages to update the resource before they are read by the extension controllers, they will be properly updated with `gardener.cloud/operation=migrate`. Otherwise, they would be reconciled as normal.
+
+4. All of the above.
+
+A new `ClusterOwner` resource is introduced in the `Seed` cluster (this does not have to be a new resource - it could just be a `ConfigMap`). It is managed by the `gardenlet` and contains the following:
+```
+spec:
+  owner: <seed-name>
+  renewTimestamp: "2020-10-29T15:14:57Z"
+  leaseDurationSeconds: 120
+```
+Every `X` seconds `gardenlet` checks to see if it can connect to the Garden APIServer and if successfull updates the `ClusterOwner` resource and renews the timestamp. The extension controllers fetch this resource before each reconciliation and check if the lease has expired. If it hasn't, then migration has not been triggered and `gardenlet` can reliably tell who the leader is, therefore controllers reconcile as normal. If the lease has expired, then either control plane migration has been triggered or `gardenlet` cannot tell who the leader is.
+*Note:* this resource could aslo be managed by the ETCD Druid and updated depending on the health of the `etcd-backup-restore` sidecar as a more reliable source of when the migration was actually triggered as it fetches this information from a "header" file in the `Shoot`'s backup directory. This would require the `etcd` resource to be deployed as soon as possible in the __Destination Seed__ (before other extension resources) so that the `etcd-backup-restore migrate` operation can be triggered to update the "header" in the `Shoot`'s source backup directory which will cause the `etcd-backup-restore` sidecar in the __Source Seed__ to stop and the `ClusterOwner` resource to be updated before extension controllers in the __Destination Seed__ can start reconciling their resources.
+
+**Alternative: DNS leader election:**
+
+A DNS TXT entry with TTL=60s and value seed='__Source Seed__' is used. The record is created and maintained by the Gardener Controller Manager (by using the DNS Controller Manager and its DNSEntry resource). When a control plane migration is detected, the Gardener Controller Manager changes the value of the DNS Entry to seed='__Destination Seed__' and waits for 2*TTL + 1 = 121 seconds to ensure that the change is propagated to all controllers in the old seed. We rely on the fact that DNS is highly available (100% for AWS Route53) and that control plane components in the __Source Seed__ can see the changes.
 
 Control plane components have to be shut down even when there is no access to the __Source Seed's__ API server. To be able to do that a daemonset is deployed in each Seed cluster. When the daemonset in the __Source Seed__ sees that it is nolonger the leader (by checking the DNS record) and there is no connection to the __Source Seed's__ API server, the daemonset will kill the Shoot's control plane pods by directly talking to the Kubelet's API server. If the __Source Seed's__ API Server comes back up, then gardenlet should take care of scaling down the deployments and statefulsets in the Shoot's control plane. This could be problematic if the gardenlet is in a crashloop backoff or takes too much time to do the scaling.
 
 As an alternative to the daemonset, a sidecar container can be added to each control plane component. The sidecar checks the DNS Entry to see if it is still the leader. If it is not, it shuts down the entire pod. This way we do not run the risk of deploymnets and statefulsets recreating the control plane pods after the seed's apiserver comes back up.
 
 The problems with using DNS as leader election is caching. Additionally, not all DNS servers respect TTL settings.
-
-**Using timestamps in the ETCD backup entries:** Once a Shoot is successfully created on the __Source Seed__ a timestamp is saved in the Cluster Resource and/or saved in the ETCD backup restore sidecar (either as an environment variable or additional configuration). The timestamps must not be modified afterwards and is used when the backup restore container writes data to the backup entry of the Shoot in the following way:
-- If there is no timestamp in the backup entry, the current timestamp is uploaded.
-- If there is a timestamp in the backup entry it is compared to the backup restore container's timestamp:
-    1. If it is the same, nothing is done
-    2. If it is older, it is replaced with the timestamp of the current backuprestore container
-    3. If it is newer - the current backuprestore container does not have ownership of the backup entry
-
-When case 3 happens it means that the shoot has been migrated and the backuprestore container in the  __Destination Seed__ has started using the Shoot's backup entry. The backuprestore on the __Source Seed__ should be configured to shut itself and the etcd container down once it sees the newer timestamp. Shutting down the etcd will cause the kubernetes control plane components to go into crashloop backoff and the MCM will not be able to do anything as it will not be able to list nodes (this has to be verified with MCM).
-
-For this approach to work backups must be enabled for the Shoot which is migrated. Additionally, synchronization based on the timestamps in the backup entry depend on the frequency of backups made by the backuprestore container.


docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
+  migrationTimestamp: "2018-11-18T15:14:57Z"
+```
+2. The ETCD Druid adds this migration timestamp to the header file inside the `Shoot`'s backup parent directory together with the __Destination Seed's__ name to indicate who the master is.
+3. When the `Shoot`'s ETCD pod is launched in the __Destination Seed__ the backup-restore sidecar checks the available backups. It restores the ETCD data from the snapshot with the latest timestamp (older then the migration timestamp).
+4. The `Shoot`'s ETCD `backup-restore` sidecar in the __Source Seed__ checks if the `Seed` name in the backup parent directory header matches the current __Source Seed__.
+    - If it does not match, it does not take a snapshot and should kill the ETCD container.
+    - If it matches - migration has not been triggerd a snapshot can be made.
+5. Once the ETCD container dies, all components which depend on it (e.g. the Kube APIServer,  Machine Controller Manager and so on) will also stop working.
+
+However, additional work has to be done for other components which do not rely on the `Shoot`'s ETCD. Following is one proposal to handle this case:
```
rfranzke: Can you explain why such additional work is needed?

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
+
+The `Cluster` resource has an additional timestamp field which is updated either by the ETCD Druid or the `backup-restore` sidecar every `X` seconds. If the ETCD Druid is used to update the timestamp it must monitor the health of the `backup-restore` sidecar. However, if `backup-restore` is used - it must have knowledge of kubernetes.
```
rfranzke: The `Cluster` resource is quite large, and updating it periodically will stress the seed's etcd. Probably a `Lease` object is more applicable here?

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
+```
+spec:
+  leadershipTimestamp: "2020-10-29T15:14:57Z"
+```
+Before extension controllers reconcile their custom resources they check this timestamp in the `Cluster` resource to see whether the current `Seed` is the leader:
```
rfranzke: Can you explain why this is necessary? Extension controllers only reconcile when the gardenlet demands it (by annotating their custom resource with `gardener.cloud/operation=reconcile`). However, if the gardenlet already knows that it is not responsible then it won't annotate in the first place.
Okay, while writing this: It can happen that extension controllers are in an exponential backoff and continue to reconcile their resources (despite the fact that the migration was triggered in the mean time and their seed is no longer the leader).

Anyways, one step back. This whole proposals is about the scenario where the source seed is unavailable/broken/down for whatever reason. Most likely, the etcd/kube-apiserver of this source seed will be offline in this case. Hence, updating the `Cluster` (or any other resource) won't be possible in the first place, right?
OK, this might still be necessary when the source seed comes up again. 

Well, can you update this document with such details? It wasn't clear in the first place but only when I started writing about it.

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
+- If the timestamp has been updated in the past `X` seconds, then the controllers will reconcile their resources. The controllers can wait for atleast 2 * `X` seconds after the `Cluster` resource has been created before reconciling - to make sure there are no controllers reconciling in the __Source Seed__.
```
rfranzke:
```suggestion
- If the timestamp has been updated in the past `X` seconds, then the controllers will reconcile their resources. The controllers can wait for at least 2 * `X` seconds after the `Cluster` resource has been created before reconciling - to make sure there are no controllers reconciling in the __Source Seed__.
```

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
-- If there is no timestamp in the backup entry, the current timestamp is uploaded.
-- If there is a timestamp in the backup entry it is compared to the backup restore container's timestamp:
-    1. If it is the same, nothing is done
-    2. If it is older, it is replaced with the timestamp of the current backuprestore container
-    3. If it is newer - the current backuprestore container does not have ownership of the backup entry
+A DNS TXT entry with TTL=60s and value seed='__Source Seed__' is used. The record is created and maintained by the Gardener Controller Manager (by using the DNS Controller Manager and its DNSEntry resource). When a control plane migration is detected, the Gardener Controller Manager changes the value of the DNS Entry to seed='__Destination Seed__' and waits for 2*TTL + 1 = 121 seconds to ensure that the change is propagated to all controllers in the old seed. We rely on the fact that DNS is highly available (100% for AWS Route53) and that control plane components in the __Source Seed__ can see the changes.
```
rfranzke, line 294:
```suggestion
A DNS TXT entry with TTL=60s and value seed='__Source Seed__' is used. The record is created and maintained by the gardenlet (by using the DNS Controller Manager and its DNSEntry resource). When a control plane migration is detected, the Gardener Controller Manager changes the value of the DNS Entry to seed='__Destination Seed__' and waits for 2*TTL + 1 = 121 seconds to ensure that the change is propagated to all controllers in the old seed. We rely on the fact that DNS is highly available (100% for AWS Route53) and that control plane components in the __Source Seed__ can see the changes.
```

rfranzke (commented):
/invite @amshuman-kr @shreyas-s-rao @vlerenc

docs/proposals/07-shoot-control-plane-migration.md
plkokanov: Yep,I was thinking about having a separate resource to handle this a bit after I updated the GEP.

docs/proposals/07-shoot-control-plane-migration.md
plkokanov: Yep, those are exactly the scenarios I wanted to cover. I'll add them to the doc.

amshuman-kr:
@plkokanov Sorry for the delay. I and @shreyas-s-rao are a bit held up with the etcd-backup-restore and etcd-druid release. I will go through the proposal and give my comments by tomorrow.

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
+The leader election depends on two header files:
+1. Header file in the `Shoot`'s backup parent directory indicating who is the master and when the migration happened as a timestamp.
+2. Header file in each snapshot folder with timestamps.
 
-**DNS leader election:** A DNS TXT entry with TTL=60s and value seed='__Source Seed__' is used. The record is created and maintained by the Gardener Controller Manager (by using the DNS Controller Manager and its DNSEntry resource). When a control plane migration is detected, the Gardener Controller Manager changes the value of the DNS Entry to seed='__Destination Seed__' and waits for 2*TTL + 1 = 121 seconds to ensure that the change is propagated to all controllers in the old seed. We rely on the fact that DNS is highly available (100% for AWS Route53) and that control plane components in the __Source Seed__ can see the changes.
+When migration is triggered:
+1. The ETCD Druid in the __Destination Seed__ is notified of the start of the migration. This "notification" can happen via the `Cluster` resource with an additional field indicating when the migration was actually triggered as a timestamp.
```
amshuman-kr:
```suggestion
1. The ETCD Druid is informed about the ownership of the an ETCD cluster instance via additional fields (such as ownership token and timestamp) in the `Etcd` resource.
```
I would suggest to inform `etcd-druid` about this via the `Etcd` resource. We can enhance it to have the additional field(s) for this.

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
+```
+spec:
+  migrationTimestamp: "2018-11-18T15:14:57Z"
+```
+2. The ETCD Druid adds this migration timestamp to the header file inside the `Shoot`'s backup parent directory together with the __Destination Seed's__ name to indicate who the master is.
```
amshuman-kr:
```suggestion
2. The ETCD Druid passes on this ownership information to the `etcd-backup-restore`.
```

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
+1. Header file in the `Shoot`'s backup parent directory indicating who is the master and when the migration happened as a timestamp.
+2. Header file in each snapshot folder with timestamps.
 
-**DNS leader election:** A DNS TXT entry with TTL=60s and value seed='__Source Seed__' is used. The record is created and maintained by the Gardener Controller Manager (by using the DNS Controller Manager and its DNSEntry resource). When a control plane migration is detected, the Gardener Controller Manager changes the value of the DNS Entry to seed='__Destination Seed__' and waits for 2*TTL + 1 = 121 seconds to ensure that the change is propagated to all controllers in the old seed. We rely on the fact that DNS is highly available (100% for AWS Route53) and that control plane components in the __Source Seed__ can see the changes.
+When migration is triggered:
+1. The ETCD Druid in the __Destination Seed__ is notified of the start of the migration. This "notification" can happen via the `Cluster` resource with an additional field indicating when the migration was actually triggered as a timestamp.
+```
+spec:
+  migrationTimestamp: "2018-11-18T15:14:57Z"
```
amshuman-kr:
```suggestion
  ownerToken: <seed_name>
  ownerTimestamp: "2018-11-18T15:14:57Z"
```
I would suggest to include some owner/master information token along with the migration timestamp. We could simply pass the seed name as the owner token.

Also, I would suggest to change terminology to ownership information rather than migration information because it might have to be supplied for freshly created shoot clusters too (please see below).

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
+```
+2. The ETCD Druid adds this migration timestamp to the header file inside the `Shoot`'s backup parent directory together with the __Destination Seed's__ name to indicate who the master is.
+3. When the `Shoot`'s ETCD pod is launched in the __Destination Seed__ the backup-restore sidecar checks the available backups. It restores the ETCD data from the snapshot with the latest timestamp (older then the migration timestamp).
```
amshuman-kr:
```suggestion
3. During shoot control-plane migration, it is expected that the `Etcd` resource in the __Source Seed__ continues to contain the older ownership token (__Source Seed__ name) and timestamp as of when it was the owner but the `Etcd` resource in the __Destination Seed__ is updated with new ownership token (__Destination Seed__ name) and timestamp.
4. The `etcd-backup-restore` sidecar initially (during startup) and continually (during backup upload?) compares the ownership information that is passed by `etcd-druid` (according to what is maintained in the corresponding `Etcd` resource) with the ownership information maintained in the header file in the corresponding `Shoot`'s backup parent directory.
   1. If the ownership timestamp in the backup is more recent than and if the owner token in the backup is different from what has been passed by `etcd-druid`, it determines that migration has been triggered and that it is in the __Source Seed__ of the migration process.
   2. If the ownership timestamp in the backup is older than and if the owner token in the backup is different from what has been passed by `etcd-druid`, it determines that the migration has been triggered and that it is in the __Destination Seed__ of the migration.
   3. If the ownership timestamp in the backup header file is the same as what it has received from `etcd-druid` it determines that no migration has been triggered and it continues exactly the way it does as of today.
5. When the backup-restore sidecar in the `Shoot`s ETCD pod determines that migration is on and that it is on the __Destination Seed__ (see above), it first adds the new ownership information (passed by `etcd-druid`) to the header file inside the `Shoot`'s backup parent directory.
   1. It then checks the available backups and restores the ETCD data from the snapshot with the latest timestamp (older than the migration timestamp).
```
Regarding the term _adds_ above, we might have to decide if the _destination_ `etcd-druid` just overwrites the ownership information in the header or appends. But this can be left as a detail in `etcd-druid` implementation.

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
+4. The `Shoot`'s ETCD `backup-restore` sidecar in the __Source Seed__ checks if the `Seed` name in the backup parent directory header matches the current __Source Seed__.
```
amshuman-kr:
```suggestion
6. When the backup-restore sidecar in the `Shoot`s ETCD pod determines that migration is on and that it is on the __Source Seed__ (see above), it stops taking snapshots (full and delta snapshots) and terminates with an error.
   1. This should also cut-off the client requests because the `readinessProbe` of the main ETCD container probes the backup-restore sidecar.
   1. Just to be sure, the health check in the backup-restore sidecar (used by the main ETCD container's readinessProbe) is enhanced to also compare the ownership information passed by `etcd-druid` with what is in the backup. This should ensure that if the main etcd container readinessProbe will continue to fail if the backup-restore sidecar comes back up.
   1. Please note that the etcd container itself might not die here. But the client requests to the container will be cut off.
```

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
+    - If it does not match, it does not take a snapshot and should kill the ETCD container.
+    - If it matches - migration has not been triggerd a snapshot can be made.
+5. Once the ETCD container dies, all components which depend on it (e.g. the Kube APIServer,  Machine Controller Manager and so on) will also stop working.
```
amshuman-kr:
```suggestion
5. Once the ETCD container stops accepting the client requests in the __Source Seed__, all components which depend on it (e.g. the Kube APIServer,  Machine Controller Manager and so on) will also stop working.
```

docs/proposals/07-shoot-control-plane-migration.md
amshuman-kr: Can't the `Etcd` resource be used here? It isn't that large and is quite relevant to the topic.

amshuman-kr: > The `Cluster` resource has an additional timestamp field which is updated either by the ETCD Druid or the `backup-restore` sidecar every `X` seconds.

Are we proposing a kind of lease or health information via the `Cluster` resource? If so, why is this needed?

docs/proposals/07-shoot-control-plane-migration.md
amshuman-kr: I agree with the approach suggested by @rfranzke. Seems simpler.

amshuman-kr: Also, for the case where the source seed comes up again, can't we make the `gardenlet` update the ownership/migration information in the `Cluster` resouce spec (similar to the `Etcd` resource spec) which then can be used by the extensions. This will eliminate the need to lease/leadershipTimestamp from druid or the backup-restore sidecar and will also be a lot less stressful to the _Source Seed_ apiserver.

This approach is simpler and workable because in the migration scenario (even when the source seed is down) we only need leader _detection_ because leader _election_ is already implicitly done centrally in the garden cluster when the destination seed was chosen for the shoot and the migration was triggered. I think a true leader election based on leader timestamp or lease unnecessary (and sub-optimal) here.

docs/proposals/07-shoot-control-plane-migration.md
amshuman-kr, line 294: The same comments I made about leader election [above](https://github.com/gardener/gardener/pull/3066#discussion_r518661666) holds here in the alternative too.

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -221,24 +221,44 @@ Horizontal Pod Autoscalers also need to be considered and removed.
...
+Control plane components have to be shut down even when there is no access to the __Source Seed's__ API server. To be able to do that a daemonset is deployed in each Seed cluster. When the daemonset in the __Source Seed__ sees that it is nolonger the leader (by checking the DNS record) and there is no connection to the __Source Seed's__ API server, the daemonset will kill the Shoot's control plane pods by directly talking to the Kubelet's API server. If the __Source Seed's__ API Server comes back up, then gardenlet should take care of scaling down the deployments and statefulsets in the Shoot's control plane. This could be problematic if the gardenlet is in a crashloop backoff or takes too much time to do the scaling.
```
amshuman-kr, line 296: > When the daemonset in the __Source Seed__ sees that it is nolonger the leader (by checking the DNS record) and there is no connection to the __Source Seed's__ API server, the daemonset will kill the Shoot's control plane pods by directly talking to the Kubelet's API server.

Is this really viable?

amshuman-kr (commented):
@plkokanov Thanks a lot the proposal and it is quite close what we had discussed in the etcd team some time ago. I apologize for the delay in reviewing.

I have given below some suggestions to the flow as well as the spec changes to make the process clear for both migration and non-migration scenarios.

docs/proposals/07-shoot-control-plane-migration.md
plkokanov: The daemonset idea is no-longer needed I think. I left it here just for historical reasons. I'll indicate it a bit more.

docs/proposals/07-shoot-control-plane-migration.md
plkokanov: >Are we proposing a kind of lease or health information via the Cluster resource? If so, why is this needed?

This lease information is necessary for the extension controllers which means that they would have to read the `Etcd` resource. Additionally, I think that having a different resource might be better if in the future we decide to use some other form of leader election.

amshuman-kr: 👍 In the https://github.com/gardener/gardener/pull/3066#discussion_r518661666 below, I have argued against leader _election_ because technically we do not need it. We only need leader _detection_. But there I assumed that we will continue to use `Cluster` and `Etcd` resources to detect leaders. But with that approach, we can use some other resource, of course. Only not the `Lease` resource, though. Do let me know what you think.

docs/proposals/07-shoot-control-plane-migration.md
plkokanov: One problem here would be if the source seed's APIServer comes back up but there is still no connectivity to the garden cluster. Then the gardenlet wouldn't know who the leader is. 

We have to decide what extension controllers do in the following case:
If migration has not been triggered and the extension controllers have resources to reconcile due to exponential backoff, should they reconcile them when there is no connectivity to the garden cluster?

If we want the extension controllers to reconcile their resources then we should depend on a lease object or the etcd resource as that is the only reliable way for them to know that this seed is still master.

rfranzke:
It seems there is neither progress nor an anytime-soon plan to follow up with the implementation of this GEP (assuming it would be finished and merged). I suggest to close it for now and reopen/rework it once it gets more relevance and when there are actually plans to pursue the outlined improvements. Please correct me if I'm wrong.

plkokanov:
I was planning to update this by the end of the week. It kind of depends on the "Dedicated secrets per backup bucket issue" - [#17](https://github.com/gardener-security/security-backlog/issues/17) but I think we can move both topics in parallel. Then I should be free to start implementing this bad case scenario as soon as this gets accepted.

timebertt:
What's the status here? Please update the proposal or close it accordingly.
/author-action

plkokanov:
@rfranzke @timebertt @amshuman-kr I have updated the PR, but there are still a few topics that it would be good to discuss and align our opinions on (perhaps in a dedicated meeting)
1. How we are going to provision a secret for the backup bucket of the __Source Seed__ in the __Destination Seed__ - this will have to be aligned with [Dedicated secrets per seed backup bucket](https://github.com/gardener-security/security-backlog/issues/17)
2. Which controller we are going to use to trigger `etcd-backup-restore migrate` and how it will be notified of CP Migration - currently I have specified that this will be taken care of by the ETCD Druid as it already contains information about versions/charts of `etcd-backup-restore`, but when talking with @amshuman-kr we also considered that a separate controller (Backup Migration Controller) might be easier to do.
3. The exact "handshake" that will happen within "header" files in the Shoot's source and destination backup directories so that the `etcd-backup-restore migrate` and `etcd-backup-restore` sidecars in both Seeds can be synced properly.
4. Whether we will use an additional resource to stop extension controllers from reconciling in the __Source Seed__ in case a reconciliation was triggered just before a bad case scenario and/or the extension controllers are stuck in a crashloop backoff. This resource would only stop one additional reconciliation and currently the only problematic extension seems to be the `infrastructure` (maybe also the certificate extension? but I have not checked that one) - so this could also be handled by revoking the access of the infrastructure controller which might be a possibility with [Eliminate all trust credentials between hyperscalers and Gardener](https://github.com/gardener-security/security-backlog/issues/25)

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -202,43 +202,99 @@ The ShootState synchronization controller will become part of the gardenlet. It
...
 
 
-### Leader Election and Control Plane Termination
+### Migrating Backups
 
-During migration "split brain" scenario must be avoided. This means that a Shoot's control plane in the __Source Seed__ must be scaled down before it is scaled up in the __Destination Seed__.
+For security reasons a `Seed` should only have access to its own backup bucket, meaning that the `etcd-backup-restore` sidecar responsible for saving and restoring the ETCD data should only be able to access the backup bucket of the `Seed` it currently operates in. Therefore, during Control Plane Migration a mechanism is needed to copy the ETCD backup from the __Source Seed__'s backup bucket to that of the __Destination Seed__. To that end `etcd-backup-restore` is extended with a new subcommand - `migrate` which runs in the Seed and does the following:
+
+1. Creates a backup directory for the `Shoot` cluster in the __Destination Seed__'s backup bucket if it does not already exist.
+1. Waits for some time (until a preconfigured timeout) for the `etcd-backup-restore` sidecar in the __Source Seed__ to finish saving a final snapshot.
```
vlerenc, line 210: I guess: Only if a backup in the source seed is currently under way, else it can migrate right away. It would be to make this clear in this description.

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -202,43 +202,99 @@ The ShootState synchronization controller will become part of the gardenlet. It
...
+1. Writes into a "header" file inside the `Shoot`'s backup directory in the __Source Seed__'s backup bucket that copying of data __from__ this directory is in process, the timestamp when the copy operation began, a timestamp indicating when migration was actually triggered and the name of the __Destination Seed__
+1. Writes into a "header" file inside the `Shoot`'s backup directory in the __Destination Seed__'s backup bucket that copying of data __to__ this directory is in process, the timestamp of when the copy operation began and a timestamp indicating when migration was actually triggered.
+1. When the `etcd-backup-restore` sidecar in the __Destination Seed__ tries to restore the data it checks the "header":
+  - If data is still being copied, it does nothing and dies.
+  - If the data has finished copying, it restores the ETCD from it.
+1. When the copy operation has finished successfully, the "header" files in both the source and destination backup directories are updated accordingly.
+1. Subsequent executions of the command (in case the operation is repeated as part of the Control Plane Migration flow) check that the data has already been copied and do nothing
```
vlerenc, line 217: It's good that the operation is idempotent, so let's write that, at least in brackets.

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -202,43 +202,99 @@ The ShootState synchronization controller will become part of the gardenlet. It
...
+
+1. Creates a backup directory for the `Shoot` cluster in the __Destination Seed__'s backup bucket if it does not already exist.
+1. Waits for some time (until a preconfigured timeout) for the `etcd-backup-restore` sidecar in the __Source Seed__ to finish saving a final snapshot.
+1. Writes into a "header" file inside the `Shoot`'s backup directory in the __Source Seed__'s backup bucket that copying of data __from__ this directory is in process, the timestamp when the copy operation began, a timestamp indicating when migration was actually triggered and the name of the __Destination Seed__
+1. Writes into a "header" file inside the `Shoot`'s backup directory in the __Destination Seed__'s backup bucket that copying of data __to__ this directory is in process, the timestamp of when the copy operation began and a timestamp indicating when migration was actually triggered.
+1. When the `etcd-backup-restore` sidecar in the __Destination Seed__ tries to restore the data it checks the "header":
+  - If data is still being copied, it does nothing and dies.
```
vlerenc, line 214: The word "dies" is an odd choice and what happens here anyways exactly?

Before answering that question though, how/why is the new ETCD's `etcd-backup-restore` sidecar even running already? The migrate command (introduced in the beginning) indicates a functionality that must be separately invoked (I understand, it is not implicit, i.e. doesn't happen automatically when the dest seed `etcd-backup-restore` sidecar detects that its data must be fetched/migrated first from another seed). But if that's the case, i.e. its a separate step in the migration flow, why is the new ETCD/`etcd-backup-restore` sidecar already even running?

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -202,43 +202,99 @@ The ShootState synchronization controller will become part of the gardenlet. It
...
+1. Subsequent executions of the command (in case the operation is repeated as part of the Control Plane Migration flow) check that the data has already been copied and do nothing
+
+*Note:* the migration timestamp might be redundant as it will closely correspond to the timestamp indicating the start of the copy operation.
+*Note:* Any additional checks and steps that have to be taken by the `etcd-backup-restore` sidecar in the __Source Seed__ in case there are problems with the Seed are described in more detail in the __Leader Election__ section below.
+
+The `migrate` subcommand requires secrets for both the backup buckets of the __Source__ and __Destination__ Seeds. The latter is already taken care of by the Backup Entry controller. For the former - a temporary secret needs to be created and provided to the Seed cluster. The exact mechanism of how this will be done is to be discussed in more details in [Dedicated secrets per seed backup bucket](https://github.com/gardener-security/security-backlog/issues/17) and [Eliminate all trust credentials between hyperscalers and Gardener](https://github.com/gardener-security/security-backlog/issues/25)
```
vlerenc, line 222: Can we be more explicit here (not refer the security repo) as this is about the concrete proposal.

Also, why are you mentioning "Eliminate all trust credentials between hyperscalers and Gardener"? In order to reach the primary goal in your introduction, i.e. "seeds should only have access to their own backup bucket", we need dedicated roles and all of that happens on hyperscaler side. Are you suggesting to temporarily open the ACL for the dest seed role to the source backup bucket?

This is probably one of the most critical points in this proposal, so some more details would be helpful to understand the proposal better.

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -202,43 +202,99 @@ The ShootState synchronization controller will become part of the gardenlet. It
...
+
+The ETCD Druid is extended so that it starts `etcd-backup-restore migrate` as a job when it is notified that Control Plane Migration has been triggered. This notification happens via a `gardener.cloud/operation=restore` annotation in the `etcd` resource and (if necessary) a few additional fields indicating the cluster owner and when migration was triggered:
```
vlerenc, line 224: See above. If that's started as job, why is the ETCD with its sidecar already running?

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -202,43 +202,99 @@ The ShootState synchronization controller will become part of the gardenlet. It
...
+### Leader Election
 
 **Note**: This section is still under discussion. The plan is to first implement the `ShootState` and modify the reconciliation flow and extensions accordingly. Additionally multiple scenarios need to be considered depending on the reachability to the Garden cluster from the __Source Seed__ components, the __Source Seed's__ API server from the Garden cluster and the __Source Seed's__ API server from the controllers running on the seed. **The initial implementation will only cover the case where everything is running**.
 
-Extension controllers do not need leader election functionality because they only reconcile the extension resources if the reconcile operation annotation is specified in the resource. Since this is set only during a reconciliation triggered by the Garden Controller Manager it cannot happen during migration.
+During migration "split brain" scenario must be avoided. This means that a Shoot's control plane in the __Source Seed__ must be scaled down before it is scaled up in the __Destination Seed__.
```
vlerenc, line 235: Isn't that a general CPM consideration?

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -202,43 +202,99 @@ The ShootState synchronization controller will become part of the gardenlet. It
...
+    1. If it finds that a copy operation __from__ its bucket is in process or has been finished, it determines that it is in the __Source Seed__ and that control plane migration has been triggered.
+    2. If it finds that a copy operation __to__ its bucket is in process, it determines that it is in the __Destination Seed__ and that control plane migration has been triggered. Once it finds that the copy operation has completed successfully it restores the ETCD data.
+    3. If the ownership information matches and thare is no indication that a copy operation is in process, `etcd-backup-restore` continues to work exactly the way it does today.
+5. When the `etcd-backup-restore` sidecar in the `Shoot's` ETCD pod determines that migration is on and that it is on the __Source Seed__, it stops taking snapshots (full and delta snapshots) and terminates with an error
+    1. This should also cut-off the client requests because the `readinessProbe` of the main ETCD container probes the `etcd-backup-restore` sidecar.
```
vlerenc, line 256: Why is even the "cooperative hand-shake" that expects the source seed to still be active so unsafe? Why cannot the gardenlet, the druid, or the sidecar properly shut down the source ETCD main container to avoid data be written after final backup? I see multiple options:
- The `gardenlet` can easily scale down the API servers in a cooperative scenario, so ETCD keeps running, but no data will be written anymore (disadvantage: works only in cooperative mode)
- The `druid` could reschedule the ETCD pod with new migration-aware configuration (disadvantage: works only in cooperative mode + wastes multiple minutes until the new pod comes up again)
- The ETCD main pod is launched with `shareProcessNamespace: true` and the `etcd-backup-restore` sidecar gracefully terminates ETCD in the main/sibling container that retriggers the ETCD pod hand-shake and makes sure, ETCD remains offline while the final backup is taken
- The `etcd-backup-restore` sidecar fails its `readiness` check immediately and starts its final backup only after the `endpoint` for its ETCD vanished

Maybe there are more, but that we depend on "should also cut-off the client requests because the `readinessProbe` of the main ETCD container probes the `etcd-backup-restore` sidecar" eventually, i.e. after some time in which potentially more data is written, sounds a bit scary. The third and fourth approaches above (if workable) would even work in the not fully cooperative scenario where the source seed is almost completely cut off with the exception to the blob store, which is the least likely to be lost and if lost, then there is anyways no hope anymore to retrieve a fresh backup anymore.

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -202,43 +202,99 @@ The ShootState synchronization controller will become part of the gardenlet. It
...
+    1. Just to be sure, the health check in the backup-restore sidecar (used by the main ETCD container's `readinessProbe`) is enhanced to also compare the ownership information passed by `etcd-druid` with what is in the backup. This should ensure that if the main etcd container `readinessProbe` will continue to fail if the `etcd-backup-restore` sidecar comes back up.
```
vlerenc, line 257: See above. I do not understand the use of the word "should" and "just to be sure".

Let me use "should" myself now: We "should" have a rock solid mechanism in place, especially in the cooperative mode. If we cannot, let the reader now why. "Should" and "just to be sure" do not induce trust in such a critical operation.

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -202,43 +202,99 @@ The ShootState synchronization controller will become part of the gardenlet. It
...
+    1. Please note that the etcd container itself might not die here. But the client requests to the container will be cut off.
+5. When the `etcd-backup-restore` sidecar in the `Shoot's` ETCD pod determines that migration is on and that it is on the __Destination Seed__, it first waits for the migration process to finish
+    1. It then checks the available backups and restores the ETCD data from the snapshot with the latest timestamp (older than the migration timestamp).
+7. Once the ETCD container stops accepting the client requests in the __Source Seed__, all components which depend on it (e.g. the Kube APIServer, Machine Controller Manager and so on) will also stop working.
 
-#### If components in the __Source Seed__ cannot reliably read who the leader is from the Garden cluster
 
-Currently we have come up with two ideas to handle this case:
+#### Stopping extension controllers
```
vlerenc, line 264: I like a lot that we have covered so much ground here along different dimensions:
- Gardener and/or Seed API Server
- Gardenlet, extensions, from PoV of the different components
- Reachability not generally impaired (all-or-nothing), but also dependent on node or load/temporal 
- Reconciliation/operation still ongoing or not

What I am wondering however: wasn't there a good way to propagate what the `etcd-backup-restore` sidecar knows (when it sees that migration was triggered) to the rest of the bunch (`gardenlet` and by that also the extensions) or was that considered dirty?
A very long time ago, we thought, the safest way to coordinate a hand-shake is by revoking access (difficult) or DNS (TTLs and cache implementation may be tricky/unsafe, though) or a blob store and the one the sidecar has access to is ideal, because it's less likely to be impaired, it's used by the ETCD pod already and that's the most critical resource anyway that we want to inform of the migration (so that ETCD can shut down and the final backup can be taken).

docs/proposals/07-shoot-control-plane-migration.md
```diff
@@ -202,43 +202,99 @@ The ShootState synchronization controller will become part of the gardenlet. It
...
+
+As an alternative to the daemonset, a sidecar container can be added to each control plane component. The sidecar checks the DNS Entry to see if it is still the leader. If it is not, it shuts down the entire pod. This way we do not run the risk of deploymnets and statefulsets recreating the control plane pods after the seed's apiserver comes back up.
 
-For this approach to work backups must be enabled for the Shoot which is migrated. Additionally, synchronization based on the timestamps in the backup entry depend on the frequency of backups made by the backuprestore container.
+The problems with using DNS as leader election is caching. Additionally, not all DNS servers respect TTL settings.
```
vlerenc, line 300: Since this is a GEP, albeit in draft state, what's the verdict? Do we have one already?

vlerenc (changes requested):
See inline comments. One more thing about the title/content of the GEP: isn't the title misleading? It discusses, to the end, more general questions of CPM, but starts off with a specific topic, the data migration between seeds, which not at all reflects the title of the PR/GEP, so all three seem oddly out of whack?

plkokanov:
> See inline comments. One more thing about the title/content of the GEP: isn't the title misleading? It discusses, to the end, more general questions of CPM, but starts off with a specific topic, the data migration between seeds, which not at all reflects the title of the PR/GEP, so all three seem oddly out of whack?

You're right, this became a bit big and covers a few topics which definitely do not match with the title of the PR any more. I will most likely close this PR in favour of new ones and structure the content properly. 
I'd keep the backup migration inside this document though, and also mention it as a requirement at the start of the document.

docs/proposals/07-shoot-control-plane-migration.md
plkokanov: Do you mean whether we chose the DNS or ETCD (object storaged) based leader election? If so, we are going for etcd-based one for now so I can remove the DNS part.

docs/proposals/07-shoot-control-plane-migration.md
plkokanov: Yes indeed we discussed that. For that to work we would need something - ETCD Druid seems like the best fit, to monitor the health of the `etcd-backup-restore` sidecar, and then update a "lease" resource which the other controllers read. But we are more inclined to leaving the `gardenlet` juts checking if it can connect to the Garden APIServer and if it can - read the leader info from the `Shoot` spec and update this "lease".

Edit: an additional option would be that the ETCD Druid monitors the health and readiness of the `etcd-backup-restore` and adds the current status to the `etcd` resource together with a timestamp and a duration of when this timestamp expires and a new check should be done. The `gardenlet` reads this health information from the `etcd` resource and updates the "lease" resource mentioned above accordingly.

docs/proposals/07-shoot-control-plane-migration.md
plkokanov: The question here is how we invoke this functionality:
The current contract is that during Control Plane Migration extension resources in the Destination seed are annotated with `gardener.cloud/operation=restore`. We could reuse this to annotate the `etcd` resource and let the ETCD Druid start `etcd-backup-restore migrate` (although `restore` here seems a bit out of place as a term)

One way to do that is to add a new controller loop that watches just for this annotation and starts the `etcd-backup-restore migrate` independently of the controller loop that watches the `etcd` resource and starts the ETCD statefulsets and the `etcd-backup-restore` sidecar. So the synchronisation would have to happen either in the `etcd` resource - the second controller does not reconcile until the `restore` annotation is removed, or both controllers reconcile at the same time and synchronisation is done in the backup buckets.

Another way would be to do what we currently do for the `worker` controller as it supports Restore and Reconcile operations - once the `worker` resource is annotated with `gardener.cloud/operation=restore` a Restore operation is triggered which adds all machine resources and then automatically the `worker` is requeued for a Reconcile operation which installs the Machine Controller Manager and fixes the owner references between the machine resources. So for the ETCD Druid the `etcd` controller would first call its Restore function to launch `etcd-backup-restore migrate` and once that is done, requeue the `etcd` resource for a Reconcile operation in which the `etcd-backup-restore` sidecar is created.

I feel like using the Restore/Reconcile operations might be the better approach as it confirms to the already established contract for the extension controllers.

docs/proposals/07-shoot-control-plane-migration.md
plkokanov: Yep, this was only part of the Leader Election section and got moved a bit now that I edited the document. But I guess it makes sense to have it at the start.

plkokanov:
Closing this in favour of #4107

PR #3017, opened 2020-10-21
[GEP-13] Automated Seed Management

stoyanr:
**How to categorize this PR?**

/area documentation
/kind enhancement
/priority normal

**What this PR does / why we need it**:
Adds a GEP for Automated Seed Management.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement operator
NONE
```

PR #3067, opened 2020-10-29
[GEP-14] Advanced Scheduling

stoyanr:

/area documentation
/kind enhancement
/priority normal

Adds a GEP for Advanced Scheduling.

Fixes #

Based on #3017. The proposed features were initially proposed as part of #3017, but following the feedback are now moved into their own GEP. Although valid, they are considered lower priority than the features described in GEP-13. Therefore, this GEP is not expected to progress beyond draft state in the near future.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/docs/proposals/14-advanced-scheduling.md
+++ b/docs/proposals/14-advanced-scheduling.md
@@ -1,0 +1,79 @@
+# Advanced Scheduling
+
+This proposal describes certain advanced scheduling features, such as:
+
+* Scheduling shoots on seeds based on taints and tolerations.
+* Migrating shoots away from unhealthy seeds.
+* Descheduling and migrating shoots away from healthy but overloaded seeds.
+ 
+These features are considered "advanced", because they could be helpful in certain situations, but are not considered essential for regular operations. Implementing them would involve changes to various existing Gardener components, as well as perhaps introducing new ones. This document describes these features in more detail and proposes a design approach for some of them.
+
+In Gardener, scheduling shoots onto seeds is quite similar to scheduling pods onto nodes in Kubernetes. Therefore, a guiding principle behind the proposed design approaches is taking advantage of best practices and existing components already used in Kubernetes. 
+
+Some of these features depend on [GEP-7: Shoot Control Plane Migration](07-shoot-control-plane-migration.md), including avoiding "split brain" situations using [leader election](07-shoot-control-plane-migration.md#leader-election-and-control-plane-termination).
+
+**Note:** These features were initially proposed as part [GEP-13: Automated Seed Management](13-automated-seed-management.md), but were later moved into their own GEP. Although valid, they are considered lower priority than the features described in GEP-13. Therefore, this GEP is not expected to progress beyond draft state in the near future.
+
+## Taints and Tolerations for Seeds and Shoots
+
+This feature is very similar to using taints and tolerations for nodes and pods in Kubernetes, as described in [Taints and Tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/). *Taints* are applied to seeds, and allow a seed to repel a set of shoots. *Tolerations* are applied to shoots, and allow (but do not require) the shoots to schedule onto seeds with matching taints.
+
+Taints and tolerations work together to ensure that shoots are not scheduled onto inappropriate seeds. One or more taints are applied to a seed; this marks that the seed should not accept any shoots that do not tolerate the taints.
+
+This feature is a prerequisite for the remaining features proposed in this GEP, and is also valuable on its own, therefore it is described first.
+
+Taints on seeds and tolerations on shoots are already partially implemented in Gardener. They consist of a key and an (optional) value that should match for a toleration to tolerate a taint. One such taint is `gardener.cloud/protected`, which is used to mark a seed as protected, which means that it may only be used by shoots in the `garden` namespace. Taints and tolerations are also taken into account by the scheduler when scheduling shoots onto seeds. In order to schedule a shoot onto a seed that has taints, all of the seed taints must be tolerated by the shoot.
+
+This proposal adds to the existing implementation the concept of [taint effects](#taint-effects), as explained below.
+
+### Taint Effects
+
+In addition to a key and a value, a taint can have an *effect*. The possible taint effects are the same as in Kubernetes: `NoSchedule`, `PreferNoSchedule`, and `NoExecute`. Since multiple taints can be put on the same seed and multiple tolerations on the same shoot, only the taints that are not tolerated by a shoot have the indicated effects on it. In particular:
+
+* If there is at least one un-tolerated taint with effect `NoSchedule` then Gardener will not schedule the shoot onto that seed. Since this is what the scheduler does already if an effect is not specified, `NoSchedule` is the implicit effect of a taint with no effects specified.
+* If there is no un-tolerated taint with effect `NoSchedule` but there is at least one un-tolerated taint with effect `PreferNoSchedule` then Gardener will try to not schedule the shoot onto the seed.
+* If there is at least one un-tolerated taint with effect `NoExecute` then the shoot will be evicted from the seed (if it is already running on the seed), and will not be scheduled onto the seed (if it is not yet running on the seed).
+ 
+## Migrating Shoots Away from Unhealthy Seeds
+
+There are situations in which a seed can't function normally and should be either deleted or repaired "offline". This could be made easier by migrating shoots away from unhealthy seeds to healthy ones automatically or semi-automatically. This scenario is triggered by detecting an unhealthy seed. Once such a seed is detected, it should be drained of all its shoots, which means that they should be evicted from the seed and scheduled to other, healthy seeds.
+
+### Marking Seeds as "Unschedulable"
+
+The first step in migrating shoots away from an unhealthy seed is marking the seed as being "unschedulable" via a special taint, for example `seed.gardener.cloud/unschedulable`, with a `NoSchedule` effect. Initially, this could simply be done manually, or perhaps assisted by a special `gardenctl` command, for example `gardenctl taint <seed> `. In addition, the `gardenctl drain <seed>` command described below should add the taint to the specified seed prior to actually evicting all currently scheduled shoots.
+
+Marking a seed as "unschedulable" prevents any new shoots to be scheduled onto this seed, due to the `NoSchedule` effect of the taint. It is useful for evicting shoots away from unhealthy seeds, but not limited to this use case. Operators might want to mark some seeds as "unschedulable" also in cases where a seed temporarily cannot accept more shoots.
+
+Later, detecting unhealthy seeds and marking them as "unschedulable" could also be done automatically. This could be achieved by extending the `gardenlet` seed controller. In addition to setting certain seed conditions, the seed controller could also add the relevant taints with `NoSchedule` effects. Such taints could include for example:
+
+* `seed.gardener.cloud/not-ready`: The seed is `NotReady`. 
+* `seed.gardener.cloud/over-capacity`: The seed's capacity for shoots has been exceeded, see [Ensuring Seeds Capacity for Shoots Is Not Exceeded](13-automated-seed-management.md#ensuring-seeds-capacity-for-shoots-is-not-exceeded).
+
+**Note:** The above taints are only meant as examples. In particular, the `seed.gardener.cloud/not-ready` taint could be valuable if in some cases shoots could be asked to tolerate a "not-ready" taint, similarly to pods tolerating not-ready nodes. If this is not the case, then such a taint should probably not be introduced. 
+
+### Evicting Shoots from Drained Seeds
+
+The next step in migrating shoots away from an unhealthy seed is actually draining the seed, which involves evicting all shoots from this seed. This could be achieved in one of the following ways:
+
+* By manually marking all seed shoots as orphaned, by deleting the `seedName` field in the shoot specification.
+* By a special `gardenctl` command, for example `gardenctl drain <seed>`, that would automate the action described above. It should also add the `seed.gardener.cloud/unschedulable` taint to the specified seed.
+
+**Note:** The behavior described here is similar to the behavior of the `kubectl drain <node>` command. This command adds a taint with a `NoSchedule` effect to the node, but doesn't add any taints with a `NoExecute` effect. Instead, it simply forcibly evicts all pods from the node.
+
+### Rescheduling Shoots to Healthy Seeds 
+
+The next step in migrating shoots away from an unhealthy seed is rescheduling all orphaned shoots onto healthy seeds by `gardener-scheduler`. The scheduler reschedules shoots for which the `seedName` field has been unset, while also taking into accounts taints and tolerations, as well as the seed capacity for shoots, as described in [Taints and Tolerations for Seeds and Shoots](#taints-and-tolerations-for-seeds-and-shoots) and [Ensuring Seeds Capacity for Shoots Is Not Exceeded](13-automated-seed-management.md#ensuring-seeds-capacity-for-shoots-is-not-exceeded).
+
+To reschedule a shoot, the scheduler updates the `seedName` field in the shoot specification with the name of the new seed. 
+
+### Migrating the Shoots to the New Seeds
+
+The final step is actually migrating the shoots control planes from the old to the new seeds. The migration is triggered by the updates to the `seedName` field in the shoot specification by the scheduler described above.
+
+The migration process is described in more detail in [GEP-7: Shoot Control Plane Migration](07-shoot-control-plane-migration.md). 
+
+## Descheduling Shoots from Healthy but Overloaded Seeds
+
+In certain situations, for example if the seed capacity for shoots drops below its initial value for whatever reason, it might be required to deschedule shoots from an otherwise healthy but already overloaded seed. In this case, only a certain number of shoots should be descheduled from the overloaded seed and scheduled onto other seeds. 
+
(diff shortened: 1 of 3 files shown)


rfranzke (dismissed):
The proposal looks good to me, no comments at this point in time (implementation details can be discussed later at the respective PR, or better, at the respective (to-be-created) issue before the concrete implementation starts).

docs/proposals/14-advanced-scheduling.md
```diff
@@ -0,0 +1,77 @@
...
+
+Some of these features depend on [GEP-7: Shoot Control Plane Migration](07-shoot-control-plane-migration.md), including avoiding "split brain" situations using [leader election](07-shoot-control-plane-migration.md#leader-election-and-control-plane-termination).
+
+**Note:** These features were initially proposed as part [GEP-13: Automated Seed Management](13-automated-seed-management.md), but were later moved into their own GEP. Although valid, they are considered lower priority than the features described in GEP-13. Therefore, this GEP is not expected to progress beyond draft state in the near future.
+
+## Taints and Tolerations for Seeds and Shoots
+
+This feature is very similar to using taints and tolerations for nodes and pods in Kubernetes, as described in [Taints and Tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/). *Taints* are applied to seeds, and allow a seed to repel a set of shoots. *Tolerations* are applied to shoots, and allow (but do not require) the shoots to schedule onto seeds with matching taints.
```
timebertt, line 19: I think, we already have a subset of the proposal wrt taints and tolerations implemented.
Can you outline in the document, what is missing / what needs to be added/changed?

docs/proposals/14-advanced-scheduling.md
```diff
@@ -0,0 +1,77 @@
...
+ 
+## Migrating Shoots Away from Unhealthy Seeds
+
+There are situations in which a seed can't function normally and should be either deleted or repaired "offline". This could be made easier by migrating shoots away from unhealthy seeds to healthy ones automatically or semi-automatically. This scenario is triggered by detecting an unhealthy seed. Once such a seed is detected, it should be drained of all its shoots, which means that they should be evicted from the seed and scheduled to other, healthy seeds.
+
+### Marking Seeds as "Unschedulable"
+
+The first step in migrating shoots away from an unhealthy seed is marking the seed as being "unschedulable" via a special taint, for example `seed.gardener.cloud/unschedulable`, with a `NoSchedule` effect. Initially, this could simply be done manually, or perhaps assisted by a special `gardenctl` command, for example `gardenctl taint <seed> `. In addition, the `gardenctl drain <seed>` command described below should add the taint to the specified seed prior to actually evicting all currently scheduled shoots.
```
timebertt, line 43: Minor remark: I think marking Seeds as "Unschedulable" is actually only loosely coupled to the use case of evicting shoots away from unhealthy seeds. Operators might want to mark some seeds as "unschedulable" also in some cases, where a Seed temporarily can't accept more shoots.

docs/proposals/14-advanced-scheduling.md
```diff
@@ -0,0 +1,77 @@
...
+
+Marking a seed as "unschedulable" prevents any new shoots to be scheduled onto this seed, due to the `NoSchedule` effect of the taint.
+
+Later, detecting unhealthy seeds and marking them as "unschedulable" could also be done automatically. This could be achieved by extending the `gardenlet` seed controller to react on certain seed conditions by adding certain taints with `NoSchedule` effects. Such taints could include for example:
```
timebertt: The Seed controller in the gardenlet shouldn't try to detect those conditions, as it sets those conditions itself. And if something is broken and the gardenlet is stuck, it also won't mark the seeds as "unschedulable".

docs/proposals/14-advanced-scheduling.md
```diff
@@ -0,0 +1,77 @@
...
+
+* `seed.gardener.cloud/not-ready`: The seed is `NotReady`.
```
timebertt, line 49: I don't get the motivation for this additional taint, which is basically replicating a `False` value of the `Ready` condition.

docs/proposals/14-advanced-scheduling.md
```diff
@@ -0,0 +1,77 @@
...
+Later, detecting unhealthy seeds and marking them as "unschedulable" could also be done automatically. This could be achieved by extending the `gardenlet` seed controller to react on certain seed conditions by adding certain taints with `NoSchedule` effects. Such taints could include for example:
+
+* `seed.gardener.cloud/not-ready`: The seed is `NotReady`. 
+* `seed.gardener.cloud/over-capacity`: The seed's capacity for shoots has been exceeded, see [Ensuring Seeds Capacity for Shoots Is Not Exceeded](13-automated-seed-management.md#ensuring-seeds-capacity-for-shoots-is-not-exceeded).
+
+### Evicting Shoots from Drained Seeds
+
+The next step in migrating shoots away from an unhealthy seed is actually draining the seed, which involves evicting all shoots from this seed. This could be achieved in one of the following ways:
+
+* By manually marking all seed shoots as orphaned, either by adding an `orphan` annotation or simply be deleting the `seedName` field in the shoot specification.
```
timebertt: I think, unsetting `seedName` should be enough. We probably, don't need a second/separate indicator for shoots that should be moved/rescheduled.

docs/proposals/14-advanced-scheduling.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Scheduling Orphaned Shoots to Healthy Seeds 
+
+The next step in migrating shoots away from an unhealthy seed is rescheduling all orphaned shoots onto healthy seeds by `gardener-scheduler`. The scheduler should ensure, among other things, that:
+
+* A shoot with the `orphan` annotation or with an empty `seedName` field is rescheduled.
+* A shoot is never scheduled onto a seed tainted with a taint with a `NoSchedule` effect (unless the shoot tolerates the taint).
+* The new seed capacity for shoots is not exceeded by adding a new shoot, see [Ensuring Seeds Capacity for Shoots Is Not Exceeded](13-automated-seed-management.md#ensuring-seeds-capacity-for-shoots-is-not-exceeded). This could perhaps also be achieved by automatically tainting the seed with a `NoSchedule` taint once its capacity is exceeded by the `gardenlet`.
+
+To reschedule a shoot, the scheduler updates the `seedName` field in the shoot specification with the name of the new seed.
```
timebertt: I think, all of the requirements should already be fulfilled, if the previous sections are implemented.
So this section can probably be shortened/condensed to rather describe "Rescheduling Shoots" (scheduler acts also on existing shoots for which the `seedName` field has been unset).

docs/proposals/14-advanced-scheduling.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Migrating the Shoots to the New Seeds
+
+The final step is actually migrating the shoots control planes from the old to the new seeds. The migration is triggered by the updates to the `seedName` field in the shoot specification by the scheduler described above.
+
+The migration process is described in more detail in [GEP-7: Shoot Control Plane Migration](07-shoot-control-plane-migration.md). 
+
+## Descheduling Shoots from Healthy but Overloaded Seeds
+
+In certain situations, for example if the seed capacity for shoots drops below its initial value for whatever reason, it might be required to deschedule shoots from an otherwise healthy but already overloaded seed. In this case, only a certain number of shoots should be descheduled from the overloaded seed and scheduled onto other seeds.
```
timebertt, line 77: Hmm, I don't think, that there is some foreseeable need to automate a "partial Seed drain". Once the others sections are implemented, the operator is able to do this manually, if such a rare situation arises, where it is needed.

timebertt (changes requested):
Thanks for creating this GEP. I very much like, that this document now contains a set of more isolated topics, which can be discussed more in-depth and also independently from other topics around "automated Seed management". 👍 
I have added some feedback inline.

docs/proposals/14-advanced-scheduling.md
stoyanr: Yes, I will do it.

docs/proposals/14-advanced-scheduling.md
stoyanr: Yes, fully agreed, and it was in this sense it was mentioned in the document. Do you think I should change something to make it more explicit?

docs/proposals/14-advanced-scheduling.md
stoyanr: Is your remark more towards the wording, e.g. "react on certain conditions" should be rather "in addition to setting certain conditions", or is it an objection about whether `gardenlet` is the correct component to do it?

docs/proposals/14-advanced-scheduling.md
stoyanr: It is mostly meant as an example. Such a taint could be valuable if in some cases shoots could be asked to tolerate a "not-ready" taint, similarly to pods tolerating not-ready nodes (e.g. CCM pods). If this is not the case, then such a taint should probably not be introduced. Do you have better examples for taints that shoots might actually need to tolerate in certain cases?

docs/proposals/14-advanced-scheduling.md
stoyanr: Yes, I agree, I only described it like this it because I am not sure unsetting the `seedName` and then setting it to a different value would work as expected right now (trigger the migration). This should however be tested and fixed.

docs/proposals/14-advanced-scheduling.md
stoyanr: Yes, I will rewrite it as suggested.

docs/proposals/14-advanced-scheduling.md
stoyanr: Should I remove the entire section then? Or do you suggest to rework it in some way?

docs/proposals/14-advanced-scheduling.md
stoyanr: Added a note to the document to clarify this:

> **Note:** The above taints are only meant as examples. In particular, the `seed.gardener.cloud/not-ready` taint could be valuable if in some cases shoots could be asked to tolerate a "not-ready" taint, similarly to pods tolerating not-ready nodes. If this is not the case, then such a taint should probably not be introduced.

docs/proposals/14-advanced-scheduling.md
stoyanr: Assumed the former, and corrected the document accordingly:

> In addition to setting certain seed conditions, the seed controller could also add the relevant taints with `NoSchedule` effects.

docs/proposals/14-advanced-scheduling.md
stoyanr: Added a few more sentences to make this explicit:

> It is useful for evicting shoots away from unhealthy seeds, but not limited to this use case. Operators might want to mark some seeds as "unschedulable" also in cases where a seed temporarily cannot accept more shoots.

docs/proposals/14-advanced-scheduling.md
stoyanr: Added a paragraph to describe the existing implementation and what is being added.

docs/proposals/14-advanced-scheduling.md
stoyanr: Removed the mention of an `orphan` annotation.

docs/proposals/14-advanced-scheduling.md
stoyanr: Rewritten as suggested and shortened considerably.

docs/proposals/14-advanced-scheduling.md
stoyanr: Keeping the section for now, until further feedback. It's anyway just a mention of something that could be valuable, without going into any kind of detail.

stoyanr:
@timebertt Revised the document to reflect your feedback, please take a look if it's ok from your perspective.

timebertt:
> @timebertt Revised the document to reflect your feedback, please take a look if it's ok from your perspective.

Thanks, @stoyanr. Will do so soon :)

docs/proposals/14-advanced-scheduling.md
timebertt: Ok got it.

I have one use case in mind, that might be a bit more useful:
If you want to enable a new feature gate only on certain seeds, a new taint specific to the feature gate could be added in order to repel shoots with workload, that might be potentially incompatible with the feature gate change. Users could then specify, that the are actually ok with their Shoots being scheduled on this "experimental Seed".

If you want to go even further, we could possibly implement a Canary rollout feature with this, adding a new Seed, adding taints for the new feature gate, then gradually scheduling more and more shoots there. But that's way more complex and maybe just a nice dream 😄

docs/proposals/14-advanced-scheduling.md
```diff
@@ -0,0 +1,79 @@
...
+**Note:** The above taints are only meant as examples. In particular, the `seed.gardener.cloud/not-ready` taint could be valuable if in some cases shoots could be asked to tolerate a "not-ready" taint, similarly to pods tolerating not-ready nodes. If this is not the case, then such a taint should probably not be introduced. 
+
+### Evicting Shoots from Drained Seeds
+
+The next step in migrating shoots away from an unhealthy seed is actually draining the seed, which involves evicting all shoots from this seed. This could be achieved in one of the following ways:
+
+* By manually marking all seed shoots as orphaned, by deleting the `seedName` field in the shoot specification.
+* By a special `gardenctl` command, for example `gardenctl drain <seed>`, that would automate the action described above. It should also add the `seed.gardener.cloud/unschedulable` taint to the specified seed.
+
+**Note:** The behavior described here is similar to the behavior of the `kubectl drain <node>` command. This command adds a taint with a `NoSchedule` effect to the node, but doesn't add any taints with a `NoExecute` effect. Instead, it simply forcibly evicts all pods from the node.
```
timebertt, line 61:
```suggestion
**Note:** The behavior described here is similar to the behavior of the `kubectl drain <node>` command. This command adds a taint with a `NoSchedule` effect to the node, but doesn't add any taints with a `NoExecute` effect. Instead, it simply evicts all pods from the node one by one.
```

docs/proposals/14-advanced-scheduling.md
```diff
@@ -0,0 +1,79 @@
...
+* `seed.gardener.cloud/not-ready`: The seed is `NotReady`. 
+* `seed.gardener.cloud/over-capacity`: The seed's capacity for shoots has been exceeded, see [Ensuring Seeds Capacity for Shoots Is Not Exceeded](13-automated-seed-management.md#ensuring-seeds-capacity-for-shoots-is-not-exceeded).
+
+**Note:** The above taints are only meant as examples. In particular, the `seed.gardener.cloud/not-ready` taint could be valuable if in some cases shoots could be asked to tolerate a "not-ready" taint, similarly to pods tolerating not-ready nodes. If this is not the case, then such a taint should probably not be introduced. 
+
+### Evicting Shoots from Drained Seeds
+
+The next step in migrating shoots away from an unhealthy seed is actually draining the seed, which involves evicting all shoots from this seed. This could be achieved in one of the following ways:
+
+* By manually marking all seed shoots as orphaned, by deleting the `seedName` field in the shoot specification.
```
timebertt, line 58: We could think about doing this via a subresource (e.g. `shoots/evict`), just like the eviction subresource for pods. And then disallowing this change via a normal `UPDATE shoots` call.
The benefits would be:
- the action can be protected by RBAC and therefore restricted to operators and controllers
- the subresource could accept additional paramters (not sure, if there is a use case for such things upfront, maybe "graceful eviction"?)
- further usecases like `ShootDisruptionBudgets` could be implemented more easily and in a cleaner way

This is a bit theoretical, but I guess, it can potentially save us some headaches and give us more freedom in the future.
WDYT?

docs/proposals/14-advanced-scheduling.md
```diff
@@ -0,0 +1,79 @@
...
+
+**Note:** The behavior described here is similar to the behavior of the `kubectl drain <node>` command. This command adds a taint with a `NoSchedule` effect to the node, but doesn't add any taints with a `NoExecute` effect. Instead, it simply forcibly evicts all pods from the node.
+
+### Rescheduling Shoots to Healthy Seeds 
+
+The next step in migrating shoots away from an unhealthy seed is rescheduling all orphaned shoots onto healthy seeds by `gardener-scheduler`. The scheduler reschedules shoots for which the `seedName` field has been unset, while also taking into accounts taints and tolerations, as well as the seed capacity for shoots, as described in [Taints and Tolerations for Seeds and Shoots](#taints-and-tolerations-for-seeds-and-shoots) and [Ensuring Seeds Capacity for Shoots Is Not Exceeded](13-automated-seed-management.md#ensuring-seeds-capacity-for-shoots-is-not-exceeded).
+
+To reschedule a shoot, the scheduler updates the `seedName` field in the shoot specification with the name of the new seed.
```
timebertt, line 67:
```suggestion
To reschedule a shoot, the scheduler updates the `seedName` field in the shoot specification with the name of the new seed, just like for new Shoots.
Currently, there is no use case for differentiating between rescheduling an evicted Shoot and scheduling a new Shoot. If there will be a use case for this in the future, the scheduler can easily achieve this by checking for example the Shoot's `status.seedName` field.
```
Right?

timebertt (commented):
Looks quite good to me.
Only a few more minor suggestions.

docs/proposals/14-advanced-scheduling.md
timebertt: Yes, fine for me :)

timebertt:
@stoyanr Can you outline your plans regarding this GEP and topic in general?

rfranzke:
It seems there is neither progress nor an anytime-soon plan to follow up with the implementation of this GEP (assuming it would be finished and merged). I suggest to close it for now and reopen/rework it once it gets more relevance and when there are actually plans to pursue the outlined improvements. Please correct me if I'm wrong.

timebertt:
Closing as proposed above. Please feel free to reopen, if you want to continue/rework it already.
/close

PR #3068, opened 2020-10-29
Remove pod panels from node-details dashboard

### LINKED ISSUE #2988: need to monitor real time pod numbers

/area monitoring
/kind enhancement
/priority normal

in Grafana dashboard `Node/Worker Pool Overview`, need to have a pod number count for each workgrop. please do not average it;
better, we can select a k8s namespace in this dashboard, so we can only shows the pod count in the namespace.

**Why is this needed**:
currently there is a `Average # of Pods on each Node in Worker Group(s) All`, it always shows flat cruve, even there was a high workload in seveal hours.
we run Jenkins Agents in the kubernetes cluster, would need to understand how many workloads does jenkins bring to this k8s cluster in certain time slot.

wyb1:

/area monitoring
/kind cleanup
/priority normal

Remove pod panels from node-details dashboard

Fixes #2988

**Release note**:

```improvement operator
Remove pod panels from node-details dashboard. These panels only showed pods running in the `kube-system` namespace so the data is not very relevant and only leads to confusion.
```

```improvement operator
Remove pod panels from node-details dashboard. These panels only showed pods running in the `kube-system` namespace so the data is not very relevant and only leads to confusion.
```

### DIFF
--- a/charts/seed-monitoring/charts/grafana/dashboards/owners/node-details-dashboard.json
+++ b/charts/seed-monitoring/charts/grafana/dashboards/owners/node-details-dashboard.json
@@ -6,11 +6,13 @@
   "editable": true,
   "gnetId": 2115,
   "graphTooltip": 0,
-  "iteration": 1560430729381,
+  "id": 1,
+  "iteration": 1603989086481,
   "links": [],
   "panels": [
     {
       "collapsed": false,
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm

Issue #3019, opened 2020-10-21
Kubeconfig rotation does not work as expected


rfranzke:
**How to categorize this issue?**

/area usability
/kind bug
/priority critical

**What happened**:
After the rotation of the kubeconfig of a shoot the new kubeconfig is (sometimes?) not synced into the project namespace in the garden cluster, hence, the downloadable kubeconfig does not work and the end-user cannot interact with his cluster.

**What you expected to happen**:
The kubeconfig in the project namespace is up to date.

**How to reproduce it (as minimally and precisely as possible)**:
n/a yet, sorry

**Anything else we need to know?**:
I've seen occasions in which even the `kubecfg` secret in the shoot namespace in the seed was not correct, i.e., in those cases not even a second reconciliation helps, but only trying to rotate the credentials again.

**Environment**:

- Gardener version: v1.11.3

vpnachev:
After some investigation I found several weird things
- the related secrets in the shoot namespace in the seed are re-generated
- the shootstate in the garden is never updated and keeps the old static-token
- sometimes the kubeconfig synced to the garden clusters is containing the deleted token, sometimes it is properly updated
- not every trigger of kubeconfig rotation reproduces the issue. 

Just wild guessing, it might be related to https://github.com/gardener/gardener/pull/2798 and to the cached clients, but I have no proofs about it. I will keep looking into it trying to understand more what exactly is happening.

vpnachev:
I can confirm that the shoot state is never updated and this is always reproducible.
The reason seems to be the [reflect.DeepEqual](https://github.com/gardener/gardener/blob/2cfba7d68e6e8815f59b9620883afbe7f328c601/pkg/operation/botanist/secrets.go#L90) that does not detect differences correctly.

I have added some logs:
<details>


```diff
diff --git a/pkg/apis/core/v1alpha1/helper/shootstate_list.go b/pkg/apis/core/v1alpha1/helper/shootstate_list.go
index bc60271bf..dde5b29ec 100644
--- a/pkg/apis/core/v1alpha1/helper/shootstate_list.go
+++ b/pkg/apis/core/v1alpha1/helper/shootstate_list.go
@@ -85,6 +85,15 @@ func (g *GardenerResourceDataList) Get(name string) *gardencorev1alpha1.Gardener
 	return nil
 }
 
+// Keys returns the names of the items in the list
+func (g *GardenerResourceDataList) Keys() []string {
+	keys := []string{}
[...]

Issue #2965, opened 2020-10-12
Golang race detector finds races in Gardener Code


BeckerMax:
Running the [data race detector](https://golang.org/doc/articles/race_detector.html) on Garden components detects multiple races. This issue is meant for discussing how we proceed with the findings.

Background:
My goal was to find a test setup which detects issues coming from shared access to data structures like https://github.com/gardener/gardener/pull/2860. The issue #2860 was caused by shared access on the same slice. This would qualify as a race the detector could find if multiple workers would modify the [admission plugins object](https://github.com/vpnachev/gardener/blob/684ba8a6fea0221ba74583f5e9fd818fbb5f6a84/pkg/operation/botanist/controlplane.go#L1066) concurrently. My assumption is that this could occur if multiple Shoots with modifications to the admission plugins would be reconciled in parallel. 
In order to provoke the error I built the gardener binaries with the `--race` flag and ran them locally. Multiple races were detected unrelated to the above issue.

Open points:
- Analyze criticality of the detected races for the different gardener binaries.
- Possibly run the race detector (for a limited time) on dev. To be kept in mind: The cost of race detection varies by program, but for a typical program, memory usage may increase by 5-10x and execution time by 2-20x.

Tasks:
- [ ] Build a race detector enabled container images flavor and run them on dev (permanently and when we do release validation).

/area control-plane
/area testing
/area quality
/priority normal
/kind discussion

BeckerMax:
Next step is to run Gardener locally with the `-race` flag. Create a few Shoots and then look at the findings. Compile the findings of the race detector and see wether they are actionable. Post the findings into this issue.

acumino:
/reopen
/lifecycle/frozen

timebertt:
We discussed this issue again and had the following thoughts:

- running the race detector probably reveals some races out of our control, false positives or even ones that we are not interested in
- we would need some experience about how valuable running the race detector is
- it's unclear what's the behavior on a race (log or panic?) and how we would follow up on races found on landscapes or in CI
- we generally don't want to close it without action, so the next step is to run the race detector locally and clarify these questions

ialidzhikov:
/unassign
as I also don't have capacity to start looking into this issue (it is also unclear what we want to achieve with this issue)

rfranzke:
This issue is stale since > `2y` and never got much attraction. It seems it is not relevant at all. Let's
/close
it now since every attempt to work on this yielded in giving up eventually in favor of other issues with more priority.
[...]

PR #3069, opened 2020-10-29
Fix bug in secret generation caused by usage of shared memory

vpnachev:

/area usability
/kind bug
/priority normal

With this PR, when the kubeconfig rotation is triggered, the ShootState will be properly updated. Previously, it was containing the initial static token and never updated.

Part of #3019 https://github.com/gardener/gardener/issues/2965
I will check separately whether/how is this related with the sync of the kubeconfig from the seed to the garden cluster.

**Release note**:

```improvement operator
A bug that was preventing the `ShootState` resource to be updated with newly generated secrets is now fixed.
```

### DIFF
--- a/pkg/apis/core/v1alpha1/helper/shootstate_list.go
+++ b/pkg/apis/core/v1alpha1/helper/shootstate_list.go
@@ -36,8 +36,8 @@ func (e *ExtensionResourceStateList) Get(kind string, name, purpose *string) *ga
 
 // Delete removes an ExtensionResourceState from the list by kind, name and purpose
 func (e *ExtensionResourceStateList) Delete(kind string, name, purpose *string) {
-	for i, obj := range *e {
-		if matchesExtensionResourceState(&obj, kind, name, purpose) {
+	for i := len(*e) - 1; i >= 0; i-- {
+		if matchesExtensionResourceState(&(*e)[i], kind, name, purpose) {
 			*e = append((*e)[:i], (*e)[i+1:]...)
 			return
 		}
@@ -68,9 +68,10 @@ type GardenerResourceDataList []gardencorev1alpha1.GardenerResourceData
 
 // Delete deletes an item from the list
 func (g *GardenerResourceDataList) Delete(name string) {
-	for i, e := range *g {
-		if e.Name == name {
+	for i := len(*g) - 1; i >= 0; i-- {
+		if (*g)[i].Name == name {
 			*g = append((*g)[:i], (*g)[i+1:]...)
+			return
 		}
 	}
 }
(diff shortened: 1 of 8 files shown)


test/framework/cleanup.go
```diff
@@ -53,10 +53,9 @@ func AddCleanupAction(fn func()) CleanupActionHandle {
 func RemoveCleanupAction(p CleanupActionHandle) {
 	cleanupActionsLock.Lock()
 	defer cleanupActionsLock.Unlock()
-	for i, item := range cleanupHookList {
-		if item.actionHandle == p {
+	for i := len(cleanupHookList); i >= 0; i-- {
```
ialidzhikov: Ref https://github.com/gardener/gardener/blob/v1.11.2/test/framework/cleanup.go#L19

vpnachev: Reverted. Thanks for spotting it.

vpnachev:
Also a small integration test is added for kubeconfig rotation. I am open to contribute it via a separate PR, too.

/invite @schrodit @dguendisch

dguendisch (commented):
Thanks for the integration test, it looks good!

schrodit (dismissed):

schrodit:
>🔥 Oops, something went wrong @vpnachev

Currently the integration test cannot be executed for a PR.
We try to enable that as soon as the landscaper is ready

pkg/apis/core/v1alpha1/helper/shootstate_list.go
```diff
@@ -36,10 +36,9 @@ func (e *ExtensionResourceStateList) Get(kind string, name, purpose *string) *ga
 
 // Delete removes an ExtensionResourceState from the list by kind, name and purpose
 func (e *ExtensionResourceStateList) Delete(kind string, name, purpose *string) {
-	for i, obj := range *e {
-		if matchesExtensionResourceState(&obj, kind, name, purpose) {
+	for i := len(*e) - 1; i >= 0; i-- {
+		if matchesExtensionResourceState(&(*e)[i], kind, name, purpose) {
 			*e = append((*e)[:i], (*e)[i+1:]...)
-			return
```
plkokanov, line 42: The combinations of kind, name and purpose should be unique in the ExtensionResourceStateList, this is why the return was used here. Maybe we should edit the Comment a bit to indicate that and leave the return once a resource is found.

pkg/apis/core/v1alpha1/helper/shootstate_list.go
```diff
@@ -68,8 +67,8 @@ type GardenerResourceDataList []gardencorev1alpha1.GardenerResourceData
 
 // Delete deletes an item from the list
 func (g *GardenerResourceDataList) Delete(name string) {
-	for i, e := range *g {
-		if e.Name == name {
+	for i := len(*g) - 1; i >= 0; i-- {
+		if (*g)[i].Name == name {
 			*g = append((*g)[:i], (*g)[i+1:]...)
```
plkokanov, line 73: Same as the ExtensionResourceStateList -> the names should be unique. Not sure why return was skipped here on my part :/

plkokanov (changes requested):
Thanks a lot for the integration test and the fix. Just some additional notes.

pkg/apis/core/v1alpha1/helper/shootstate_list.go
vpnachev: There is one issue if the return is not removed - it will remove only the first occurrence of an Extension matching kind,name and purpose and keep the rest. Most probably it is used in a way that duplication do not happen, but I have not checked for that.

plkokanov: Yep, the list is meant to contain only uniquely identifiable items. We didn't chose a map because lists are better from API perspective. You can check that the `Upsert` method looks for an already existing item and changes that if it finds something, instead of inserting a duplicate.

vpnachev: I have re-added the return. Also added it in the other `Delete` methods.

plkokanov (approved):
/lgtm

ialidzhikov (approved):
/lgtm

PR #3071, opened 2020-10-30
Add reference protection for audit policy ConfigMaps

ialidzhikov:

/area quality
/kind enhancement
/priority normal

After https://github.com/gardener/gardener/pull/2967, the components are being redeployed in the deletion flow. Currently if Shoot with audit policy is deleted and afterwards the policy ConfigMap is also deleted, the Shoot deletion will be stuck with reason:

```yaml
  lastErrors:
    - description: >-
        task "Deploying Kubernetes API server" failed: retry failed with context
        deadline exceeded, last error: retrieving audit policy from the
        ConfigMap 'auditpolicy' failed with reason 'configmaps
        "auditpolicy" not found'
```

This PR adds a finalizer on audit policy ConfigMap until there is a Shoot that reference it.

Well, with this PR we have code duplication in the reference controller tests and controller itself. We should check if we can improve that and avoid the code duplication.

**Release note**:

```improvement user
gardener-controller-manager's Shoot reference controller now also handles audit policy ConfigMap references.
```

### DIFF
--- a/pkg/controllermanager/controller/shoot/shoot_reference_control.go
+++ b/pkg/controllermanager/controller/shoot/shoot_reference_control.go
@@ -69,7 +69,15 @@ func (c *Controller) shootReferenceUpdate(oldObj, newObj interface{}) {
 }
 
 func refChange(oldShoot, newShoot *gardencorev1beta1.Shoot) bool {
+	return shootDNSFieldChanged(oldShoot, newShoot) || shootKubeAPIServerAuditConfigFieldChanged(oldShoot, newShoot)
+}
+
+func shootDNSFieldChanged(oldShoot, newShoot *gardencorev1beta1.Shoot) bool {
 	return !apiequality.Semantic.Equalities.DeepEqual(oldShoot.Spec.DNS, newShoot.Spec.DNS)
+}
+
+func shootKubeAPIServerAuditConfigFieldChanged(oldShoot, newShoot *gardencorev1beta1.Shoot) bool {
+	return !apiequality.Semantic.Equalities.DeepEqual(oldShoot.Spec.Kubernetes.KubeAPIServer.AuditConfig, newShoot.Spec.Kubernetes.KubeAPIServer.AuditConfig)
 }
 
 // FinalizerName is the name of the finalizer used for the reference protection.
@@ -78,13 +86,17 @@ const FinalizerName = "gardener.cloud/reference-protection"
 // SecretLister fetches secret objects with the given options.
 type SecretLister func(ctx context.Context, secretList *corev1.SecretList, options ...client.ListOption) error
 
+// ConfigMapLister fetches configmap objects with the given options.
+type ConfigMapLister func(ctx context.Context, configMapList *corev1.ConfigMapList, options ...client.ListOption) error
+
 // NewShootReferenceReconciler creates a new instance of a reconciler which checks object references from shoot objects.
 // A special `userSecretLister` serves as an option to retrieve secret objects which are not gardener managed.
-func NewShootReferenceReconciler(l logrus.FieldLogger, clientMap clientmap.ClientMap, userSecretLister SecretLister) reconcile.Reconciler {
+func NewShootReferenceReconciler(l logrus.FieldLogger, clientMap clientmap.ClientMap, userSecretLister SecretLister, configMapLister ConfigMapLister) reconcile.Reconciler {
 	return &shootReferenceReconciler{
-		clientMap:    clientMap,
-		secretLister: userSecretLister,
-		logger:       l,
+		clientMap:       clientMap,
+		secretLister:    userSecretLister,
+		configMapLister: configMapLister,
+		logger:          l,
 	}
 }
 
@@ -92,8 +104,9 @@ type shootReferenceReconciler struct {
 	ctx context.Context
 
 	// secretLister is supposed to be the most performant option to retrieve secret objects in this controller.
-	secretLister SecretLister
-	clientMap    clientmap.ClientMap
+	secretLister    SecretLister
+	configMapLister ConfigMapLister
+	clientMap       clientmap.ClientMap
 
 	logger logrus.FieldLogger
 }
@@ -136,16 +149,28 @@ func (r *shootReferenceReconciler) reconcileShootReferences(c client.Client, sho
 		return err
 	}
 
+	// Iterate over all user configmaps in project namespace and check if they can be released.
+	if err := r.releaseUnreferencedConfigMaps(c, shoot); err != nil {
+		return err
+	}
+
 	// Remove finalizer from shoot in case it's being deleted and not handled by Gardener any more.
 	if shoot.DeletionTimestamp != nil && !controllerutils.HasFinalizer(shoot, gardencorev1beta1.GardenerName) {
 		return controllerutils.PatchRemoveFinalizers(r.ctx, c, shoot, FinalizerName)
 	}
 
 	// Add finalizer to referenced secrets that are not managed by Gardener.
-	needsFinalizer, err := r.handleReferencedSecrets(c, shoot)
+	addedFinalizerToSecret, err := r.handleReferencedSecrets(c, shoot)
 	if err != nil {
 		return err
 	}
+
+	addedFinalizerToConfigMap, err := r.handleReferencedConfigMap(c, shoot)
+	if err != nil {
+		return err
+	}
+
+	needsFinalizer := addedFinalizerToSecret || addedFinalizerToConfigMap
 
 	// Manage finalizers on shoot.
 	hasFinalizer := controllerutils.HasFinalizer(shoot, FinalizerName)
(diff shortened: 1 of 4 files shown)


pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -68,32 +69,41 @@ func (c *Controller) shootReferenceUpdate(oldObj, newObj interface{}) {
 	}
 }
 
-func refChange(oldShoot, newShoot *gardencorev1beta1.Shoot) bool {
```
timuthy, line 71: Can you keep `refChange` and call all the relevant functions here? This will prevent a more and more expanding OR chain [here](https://github.com/gardener/gardener/pull/3071/files#diff-a03ba9a488c2d4f8e80461ff9f802de0fb38c91fc4c695c21636b4b5767426ffR61).

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -68,32 +69,41 @@ func (c *Controller) shootReferenceUpdate(oldObj, newObj interface{}) {
...
+func shootDNSFieldChange(oldShoot, newShoot *gardencorev1beta1.Shoot) bool {
 	return !apiequality.Semantic.Equalities.DeepEqual(oldShoot.Spec.DNS, newShoot.Spec.DNS)
 }
 
+func shootKubeAPIServerFieldChange(oldShoot, newShoot *gardencorev1beta1.Shoot) bool {
+	return !apiequality.Semantic.Equalities.DeepEqual(oldShoot.Spec.Kubernetes.KubeAPIServer, newShoot.Spec.Kubernetes.KubeAPIServer)
```
timuthy: Shouldn't we check only for `AuditConfig` changes specifically?

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -136,17 +146,29 @@ func (r *shootReferenceReconciler) reconcileShootReferences(c client.Client, sho
...
 	}
 
 	// Add finalizer to referenced secrets that are not managed by Gardener.
-	needsFinalizer, err := r.handleReferencedSecrets(c, shoot)
+	addedFinalizerToSecret, err := r.handleReferencedSecrets(c, shoot)
 	if err != nil {
 		return err
 	}
 
+	addedFinalizerToConfigMap, err := r.handleReferencedConfigMap(c, shoot)
```
timuthy, line 168:
```suggestion
	addedFinalizerToConfigMap, err := r.handleReferencedConfigMaps(c, shoot)
```

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -252,6 +314,55 @@ func (r *shootReferenceReconciler) getUnreferencedSecrets(c client.Client, shoot
...
+		if s.Name == shoot.Name && shoot.DeletionTimestamp != nil && !controllerutils.HasFinalizer(&s, gardencorev1beta1.GardenerName) {
+			continue
+		}
+		apiServerConfig := s.Spec.Kubernetes.KubeAPIServer
+		if apiServerConfig != nil &&
+			apiServerConfig.AuditConfig != nil &&
+			apiServerConfig.AuditConfig.AuditPolicy != nil &&
+			apiServerConfig.AuditConfig.AuditPolicy.ConfigMapRef != nil {
+
+			configMapRef := apiServerConfig.AuditConfig.AuditPolicy.ConfigMapRef
```
timuthy: Can you extract this to a separate function since you can also re-use it in `handleReferencedConfigMap`?

timuthy (changes requested):
So far it looks good to me, only general comments about the code.

pkg/controllermanager/controller/shoot/shoot_reference_control.go
vpnachev: Actually, no. It adds finalizer only to the config map referenced by the current shoot.

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -252,6 +314,55 @@ func (r *shootReferenceReconciler) getUnreferencedSecrets(c client.Client, shoot
...
+
+	shoots := &gardencorev1beta1.ShootList{}
+	if err := c.List(r.ctx, shoots, client.InNamespace(namespace)); err != nil {
+		return nil, err
+	}
+
+	referencedConfigMaps := sets.NewString()
+	for _, s := range shoots.Items {
+		// Ignore own references if shoot is in deletion and references are not needed any more by Gardener.
+		if s.Name == shoot.Name && shoot.DeletionTimestamp != nil && !controllerutils.HasFinalizer(&s, gardencorev1beta1.GardenerName) {
```
vpnachev, line 337: I think we should not skip shoots in deletion.
As long as the shoot exists, the finalizer on the the config map should not be removed. I see it will be re-added a few moments later in `handleReferencedConfigMap`, but it may be long enough and the garbage collector will delete the configmap.

vpnachev (commented):
Thanks!
Looks very well. I have just one concert that shoots in deletion should not be ignored as referring shoots.

pkg/controllermanager/controller/shoot/shoot_reference_control.go
ialidzhikov: > As long as the shoot exists, the finalizer on the the config map should not be removed

It should be the case. Note also the `!controllerutils.HasFinalizer(&s, gardencorev1beta1.GardenerName)` part. It should release the referenced ConfigMap when the `gardener` finalizer is removed from the Shoot (-> the Shoot deletion completed).
Let me try once again this case.

vpnachev: Oh, I missed the check for the `gardener` finelizer. Everything is good.

vpnachev (approved):
/lgtm

pkg/controllermanager/controller/shoot/shoot_reference_control.go
```diff
@@ -69,31 +69,44 @@ func (c *Controller) shootReferenceUpdate(oldObj, newObj interface{}) {
...
 func refChange(oldShoot, newShoot *gardencorev1beta1.Shoot) bool {
+	return shootDNSFieldChanged(oldShoot, newShoot) || shootKubeAPIServerAuditConfigFieldChanged(oldShoot, newShoot)
+}
+
+func shootDNSFieldChanged(oldShoot, newShoot *gardencorev1beta1.Shoot) bool {
 	return !apiequality.Semantic.Equalities.DeepEqual(oldShoot.Spec.DNS, newShoot.Spec.DNS)
 }
 
+func shootKubeAPIServerAuditConfigFieldChanged(oldShoot, newShoot *gardencorev1beta1.Shoot) bool {
+	return !apiequality.Semantic.Equalities.DeepEqual(oldShoot.Spec.Kubernetes.KubeAPIServer.AuditConfig, newShoot.Spec.Kubernetes.KubeAPIServer.AuditConfig)
```
vpnachev, line 80: If `.Spec.Kubernetes.KubeAPIServer` is nil, this will panic with segmentation fault.
```suggestion
	return !apiequality.Semantic.Equalities.DeepEqual(getAuditPolicyConfigMapRef(oldShoot.Spec.Kubernetes.KubeAPIServer), getAuditPolicyConfigMapRef(newShoot.Spec.Kubernetes.KubeAPIServer))
```

pkg/controllermanager/controller/shoot/shoot_reference_control.go
ialidzhikov: I just confirmed one more time - the ConfigMap finalizer is removed after the Shoot deletion completes (its `gardener` finalizer is removed).

pkg/controllermanager/controller/shoot/shoot_reference_control.go
ialidzhikov: Hm, kubernetes.kubeAPIServer is always defaulted. Ref https://github.com/gardener/gardener/blob/v1.12.1/pkg/apis/core/v1beta1/defaults.go#L138-L155
I think it should not be a problem.

vpnachev: Currently it will not be a problem, but if the default is changed after some time (several years, or once Gardener support only k8s clusters >=1.16), this code might not get adapted and it will become a bug.

timuthy (approved):
/lgtm

pkg/controllermanager/controller/shoot/shoot_reference_control.go
ialidzhikov: I get your point but it is something that needs to be adapted if (or when) the defaulting logic is changed. In a similar way in the func above `shootDNSFieldChanged` we rely that `ShootDNS` admission plugin (something that you can disable with the `--disable-admission-plugins` flag of the gardener apiserver) will default the dns section if ti is not specified.

timuthy: +1
Reasonable point, but we rely on defaults at many points in g/g.

timebertt:
Good job guys, thanks a lot! 🚀

PR #3072, opened 2020-10-30
Add ResourceQuota for projects

### LINKED ISSUE #1650: Quota for user facing resources

timuthy:

/area security
/area robustness
/kind enhancement
/priority normal

This PR enhances the Project controller by the option to maintain a default `ResourceQuota` for projects. If specified, GCM creates or updates a `ResourceQuota` named `gardener` in the project namespace.

Fixes #1650

**Release note**:

```noteworthy operator
Operators can now define a default `ResourceQuota` resource which is automatically created in project related namespaces. Please consult the documentation (/docs/concepts/controller-manager.md) for more information.
```


pkg/controllermanager/apis/config/validation/validation.go
```diff
@@ -0,0 +1,53 @@
...
+	return allErrs
+}
+
+func validateControllerManagerControllerConfiguration(conf config.ControllerManagerControllerConfiguration, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	projectFldPath := fldPath.Child("project")
+	if conf.Project != nil {
+		allErrs = append(allErrs, validateProjectControllerConfiguration(conf.Project, projectFldPath)...)
+	} else {
+		allErrs = append(allErrs, field.Required(projectFldPath, "must provide a project controller configuration"))
```
rfranzke: Why is it required?

rfranzke (commented):
Great PR, one small question :)

docs/concepts/controller-manager.md
```diff
@@ -26,10 +26,35 @@ After the namespace was created/adopted the reconciler creates several `ClusterR
...
+       apiVersion: v1
+       kind: ResourceQuota
+       spec:
+         hard:
+           count/shoots.core.gardener.cloud: "100"
+           count/secretbindings.core.gardener.cloud: "10"
+           count/secrets: "800"
+```
+
+The Project controller takes the shown quota specification and creates a `ResourceQuota` with the name `gardener` in the project namespace.
```
timebertt: Can you add a sentence about the `projectSelector`?

example/20-componentconfig-gardener-controller-manager.yaml
```diff
@@ -31,6 +31,16 @@ controllers:
     staleGracePeriodDays: 14
     staleExpirationTimeDays: 90
     staleSyncPeriod: 12h
+  # quota:
+  #   projectSelector: {}
```
timebertt: If we allow the administrator to specify a `projectSelector` for the default quota, would it make sense to allow specify a set/array of default quotas for different label selectors?
Otherwise, if the `projectSelector` is only there to allow excluding some projects from the default quota: should we rather remove the `projectSelector` and use a well-defined label for excluding projects from the default quota?

pkg/controllermanager/apis/config/types.go
```diff
@@ -127,6 +130,16 @@ type ProjectControllerConfiguration struct {
...
 }
 
+// QuotaConfiguration defines quota configurations.
+type QuotaConfiguration struct {
+	// ProjectSelector is an optional setting to select the projects considered for quotas.
+	// Defaults to empty LabelSelector, which matches everything.
+	ProjectSelector *metav1.LabelSelector
+	// ResourceQuota is the quota specification used for the project set-up.
+	// Only v1.ResourceQuota resources are supported.
+	ResourceQuota runtime.Object
```
timebertt: Why is this a `runtime.Object`? Is it not possible to use `v1.ResourceQuota` directly?

pkg/controllermanager/apis/config/types.go
```diff
@@ -127,6 +130,16 @@ type ProjectControllerConfiguration struct {
 	StaleSyncPeriod *metav1.Duration
 }
 
+// QuotaConfiguration defines quota configurations.
+type QuotaConfiguration struct {
+	// ProjectSelector is an optional setting to select the projects considered for quotas.
+	// Defaults to empty LabelSelector, which matches everything.
```
timebertt:
```suggestion
	// Defaults to empty LabelSelector, which matches all projects.
```

timebertt (changes requested):
Generally, looks good. Just a few suggestions :)

pkg/controllermanager/apis/config/validation/validation.go
timuthy: I added it as required because `.Project` is used at some places w/o any `nil` check. It's anyway defaulted, so this check is also redundant. I'll remove it.

pkg/controllermanager/apis/config/types.go
timuthy: It's just used as proposed in the `runtime.RawExtension` [doc string](https://github.com/kubernetes/apimachinery/blob/3a3fc754224fdaf5be79aa6583c78d1f5e411930/pkg/runtime/types.go#L49).

timebertt: But `ResourceQuota` is not really an "extension object", we already know which object is supposed to be in that field, so do we actually need `runtime.RawExtension` here?

example/20-componentconfig-gardener-controller-manager.yaml
timuthy: > If we allow the administrator to specify a projectSelector for the default quota, would it make sense to allow specify a set/array of default quotas for different label selectors?

Okay, will do.

> Otherwise, if the projectSelector is only there to allow excluding some projects from the default quota: should we rather remove the projectSelector and use a well-defined label for excluding projects from the default quota?

You can use it for exclusion, but you are not bound to it. That's the nice and flexible part about label selectors and the main reason I introduced them here.

pkg/controllermanager/apis/config/types.go
timuthy: Maybe not an extension object but a place holder for any external version which makes the API independent from the implementation. Hence, we are open to support further quota GVKs in the future. I don't see a disadvantage as long as it is documented.

timebertt: Ah yep. That makes sense. Thanks for explaining!

timuthy:
Thanks for your comments @rfranzke and @timebertt. The PR is ready for another round of reviews. I especially added support for multiple quota configs in 3781169.

rfranzke (dismissed):
/lgtm

docs/concepts/controller-manager.md
```diff
@@ -26,10 +26,37 @@ After the namespace was created/adopted the reconciler creates several `ClusterR
 These RBAC resources are prefixed with `gardener.cloud:system:project{-member,-viewer}:<project-name>`.
 Gardener administrators and extension developers can define their own roles, see [this document](../extensions/project-roles.md) for more information.
 
+In addition, operators can configure the Project controller to maintain a default [ResourceQuota](https://kubernetes.io/docs/concepts/policy/resource-quotas/) for project namespaces.
+Quotas can especially limit the creation of user facing resources, e.g. `Shoots`, `SecretBindings`, `Secrets` and thus protect the Garden cluster from massive resource exhaustion but also enable operators to align quotas with respective enterprise policies.
+
+> :warning: **Gardener itself is not exempted from configured quotas**. For example, Gardener creates `Secrets` for every shoot cluster in the project namespace and at the same time increases the available quota count. Please mind this additional resource consumption. 
+
+The GCM configuration provides a template section `controllers.project.quota.resourceQuota` where such a ResourceQuota (see example below) can be deposited.
```
timebertt:
```suggestion
The GCM configuration provides a template section `controllers.project.quotas` where such a ResourceQuota (see example below) can be deposited.
```

pkg/controllermanager/apis/config/types.go
```diff
@@ -117,6 +118,8 @@ type ProjectControllerConfiguration struct {
 	// MinimumLifetimeDays is the number of days a `Project` may exist before it is being
 	// checked whether it is actively used or got stale.
 	MinimumLifetimeDays *int
+	// Quotas is the default configuration every projects is set up with if a quota is not already specified.
```
timebertt: Nit
```suggestion
	// Quotas is the default configuration every project is set up with if a quota is not already specified.
```

pkg/controllermanager/controller/project/project_control_reconcile.go
```diff
@@ -252,6 +273,67 @@ func (c *defaultControl) reconcileNamespaceForProject(ctx context.Context, garde
...
+
+	if _, err := controllerutil.CreateOrUpdate(ctx, c, projectResourceQuota, func() error {
+		projectResourceQuota.SetOwnerReferences(common.MergeOwnerReferences(projectResourceQuota.GetOwnerReferences(), *ownerReference))
+		projectResourceQuota.Spec.Hard = quotas
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	return nil
```
timebertt: Can you avoid the extra `Get` call in `CreateOrUpdate`?
WDYT about using `Patch` here, if the quota doesn't exist, yet?

timebertt (commented):
Very nice, Tim! Thanks for addressing my suggestions.
Only a few minor comments left :)

pkg/controllermanager/controller/project/project_control_reconcile.go
timuthy: Thanks @timebertt, definitely makes sense to remove the extra Get call. However, I'm sticking to `CreateOrUpdate` to keep things simple and since we are in the same cluster, I don't expect too many performance benefits with a `PATCH` request. Do you agree?

timebertt: > I don't expect too many performance benefits with a `PATCH` request.

Yes, I wouldn't expect that either. As you moved the quota merging into the `CreateOrUpdate`, this is already better than before with the extra `Get` call. Perfect.
Thanks!

timebertt (approved):
Thanks for addressing my comments again, looks great now!
/lgtm

rfranzke (approved):

PR #3074, opened 2020-10-30
Auto-delete (Cluster)RoleBindings on invalid updates

timebertt:

/area ops-productivity
/kind enhancement
/priority normal

With this PR we add `resources.gardener.cloud/delete-on-invalid-update=true` to all `(Cluster)RoleBindings` in the `shoot-core` and `shoot-addons` charts.
By this, gardener-resource-manager will delete rolebindings that were changed by the enduser and can't be updated (e.g. the `roleRef` field was changed by the enduser).
If endusers do such changes, operators will have to intervene and cleanup manually, which should not be necessary.

Fixes #

**Release note**:

```improvement operator
Gardener now deletes `(Cluster)RoleBindings` of system components or addons, that were changed to an invalid state by endusers to be able to reconcile them back to the desired state.
```

### DIFF
--- a/charts/seed-bootstrap/charts/vpa/charts/application/templates/clusterrolebinding-actor.yaml
+++ b/charts/seed-bootstrap/charts/vpa/charts/application/templates/clusterrolebinding-actor.yaml
@@ -5,6 +5,8 @@ metadata:
   name: gardener.cloud:vpa:{{ .Values.clusterType }}:actor
   labels:
   {{ toYaml .Values.labels | indent 4 }}
+  annotations:
+    resources.gardener.cloud/delete-on-invalid-update: "true"
 roleRef:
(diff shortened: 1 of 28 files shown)


vpnachev (approved):
/lgtm

timuthy:
I suspect that this would be also a necessary change for many Gardener extensions? At least if they deploy (Cluster)RoleBindings to the shoot.

timebertt:
> I suspect that this would be also a necessary change for many Gardener extensions? At least if they deploy (Cluster)RoleBindings to the shoot.

Hmm, it is not strictly necessary. But yes, the extensions can also be hit by such user updates of course.
I think, it makes sense to think about adding a setting to enable this for all resources contained in a `ManagedResource`.

PR #3075, opened 2020-10-30
Ensure seeds capacity for shoots is not exceeded

### LINKED ISSUE #2938: Max cap for shoots per seed in scheduler
We have seen seeds (`gardenlet`) go down after being overloaded with shoots, so it may be safer to have a max cap for shoots per seed until we have implemented the automated seed cluster management that would automatically add new seeds.

**Why is this needed**:
To prevent the `gardenlet` on a seed to go down.

/cc @danielfoehrKn

stoyanr:

/area robustness
/kind enhancement
/priority normal

Ensures seeds capacity for shoots is not exceeded, as described in GEP-13 (see #3017).

* Introduce `capacity` and `allocatable` resource lists in the seed's status, similarly to the Node status.
* Initialise these fields in `gardenlet` seed controller based on the new `resources.capacity` and `resources.reserved` fields in the `GardenletConfiguration`.
* Validate `resources.capacity` and `resources.reserved`.
* During scheduling, filter out candidates that don't have available capacity for shoots.

Fixes #2938 

Resource requests by shoots and more advanced scheduling logic are not yet introduced in this PR. They will be introduced in separate PRs.

**Release note**:

```noteworthy operator
The scheduler ensures that only candidate seeds with available capacity for shoots are considered during scheduling. The resource capacity and reservations can be configured in the gardenlet's component configuration. By default, the capacity for shoots in a seed is `200`.
```


charts/gardener/gardenlet/values.yaml
```diff
@@ -94,6 +94,9 @@ global:
         shootStateSync:
           concurrentSyncs: 5
           syncPeriod: 30s
+      resources:
+        capacity:
+          shoots: 200 # TODO: What is a good default value here?
```
rfranzke: What about ~275? We have good experience with these values for now. /cc @vpnachev @vlerenc

charts/gardener/gardenlet/values.yaml
```diff
@@ -94,6 +94,9 @@ global:
         shootStateSync:
           concurrentSyncs: 5
           syncPeriod: 30s
+      resources:
```
rfranzke, line 99: Why is this part of the gardenlet configuration itself? Shouldn't it be part of the `Seed`'s specification? This way, the gardenlet uses the same resource settings for all seeds it is reconciling (although, these seeds might indeed have different capacities).

pkg/gardenlet/apis/config/v1alpha1/types.go
```diff
@@ -45,6 +45,8 @@ type GardenletConfiguration struct {
 	// Controllers defines the configuration of the controllers.
 	// +optional
 	Controllers *GardenletControllerConfiguration `json:"controllers,omitempty"`
+	// Resources defines the total capacity for seed resources and the amount reserved for use by Gardener.
```
rfranzke, line 48:
```suggestion
	// Resources defines the total capacity for seed resources and the amount reserved for use by Gardener.
	// +optional
```

pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -299,6 +300,11 @@ func filterCandidates(shoot *gardencorev1beta1.Shoot, seedList []*gardencorev1be
 			continue
 		}
 
+		if allocatableShoots, ok := seed.Status.Allocatable[gardencorev1beta1.ResourceShoots]; ok && int64(seedUsage[seed.Name]) >= allocatableShoots.Value() {
```
rfranzke, line 303: Can you add a unit test case for this condition?

rfranzke (commented):
/assign

pkg/scheduler/controller/shoot/scheduler_control.go
stoyanr: Will try to do it, shouldn't be too hard although there are currently no unit tests for this package at all.

charts/gardener/gardenlet/values.yaml
stoyanr: This was discussed at length in the team meeting. The final decision was to put `capacity` and `allocatable` in the seed status, but not in the seed spec, following the `kubelet` model, as initially proposed. The new `ShootedSeed` resource would also contain the gardenlet configuration and there it's in the spec. See also the relevant sections in the GEP.

pkg/scheduler/controller/shoot/scheduler_control.go
rfranzke: Hm, there is https://github.com/gardener/gardener/blob/master/pkg/scheduler/controller/shoot/scheduler_test.go (which probably should be renamed to `scheduler_control_test.go`), or?

charts/gardener/gardenlet/values.yaml
rfranzke: Hm, okay, maybe I missed that because I could only join late. What were again the arguments for this decision? Just "because it's in Kubernetes/kubelet like this" is a weak one, I guess. kubelets cannot be responsible for multiple nodes, only one node, but the gardenlet can theoretically be responsible for the entire list of seeds registered in a Gardener landscape. I agree that this breaks also for other configuration options (e.g., you want to enable/disable feature gates on a per-seed basis), but at least for new fields we could try to move it to the right place.
Anyways, I won't insist here if the majority decided that this is the way to go, I just wanted to raise the topic for reconsideration.

pkg/scheduler/controller/shoot/scheduler_control.go
stoyanr: Sorry, I was simply too quick confused, looking in a different package. I will add the tests.

charts/gardener/gardenlet/values.yaml
stoyanr: I think the discussion revolved around `spec` (user updatable and causes reconciliation when changed) and `status` (not user updatable and ignored by controllers when updated by other controllers). If we consider custom resources with their own extensions / plugins in the future, we were inclined to see `capacity` more of a status thing. Also, `allocatable` is now calculated based on a `reserved` configuration, but in the future `gardenlet` could try to do something smarter to determine a correct value (again, similarly to `kubelet`). The argument that other configuration options aren't configurable per seed (as you also pointed out above) also came up. We also considered adding to `spec` **and** `status`, but eventually `ShootedSeed` was proposed as the way to have both - `capacity` / `allocatable` in the Seed status but the Gardenlet configuration as part of the ShootedSeed spec, together with other values that need to be customized per Seed, and this remained as the consensus.

rfranzke: OK, thank you!

charts/gardener/gardenlet/values.yaml
vpnachev: I would start with 250 for the default value or even less. But as this is configurable value, I have no strong opinion.

pkg/scheduler/controller/shoot/scheduler_control.go
stoyanr: Done

pkg/gardenlet/apis/config/v1alpha1/types.go
stoyanr: Fixed

charts/gardener/gardenlet/values.yaml
stoyanr: Changed to 250

stoyanr:
@rfranzke I addressed your comments, could you please take another look?

pkg/gardenlet/controller/shoot/seed_registration_control.go
```diff
@@ -676,6 +676,7 @@ func deployGardenlet(ctx context.Context, gardenClient, seedClient, shootedSeedC
 					"seedClientConnection":  externalConfig.SeedClientConnection.ClientConnectionConfiguration,
 					"shootClientConnection": externalConfig.ShootClientConnection,
 					"controllers":           externalConfig.Controllers,
+					"resources":             externalConfig.Resources,
```
rfranzke: Can you make this overwriteable from the shooted seed configuration annotation?

rfranzke (changes requested):
Minor enhancement suggestion, otherwise it looks good.

/needs rebase second-opinion
/invite @danielfoehrKn

pkg/gardenlet/controller/shoot/seed_registration_control.go
stoyanr: Done, including tests

rfranzke (dismissed):
/lgtm

rfranzke (approved):

pkg/gardenlet/controller/seed/seed_control.go
```diff
@@ -176,6 +177,21 @@ func (c *defaultControl) ReconcileSeed(obj *gardencorev1beta1.Seed, key string)
...
 
+	// Initialize capacity and allocatable
+	var capacity, allocatable corev1.ResourceList
+	if c.config.Resources != nil && len(c.config.Resources.Capacity) > 0 {
+		capacity = make(corev1.ResourceList, len(c.config.Resources.Capacity))
+		allocatable = make(corev1.ResourceList, len(c.config.Resources.Capacity))
+		for resourceName, quantity := range c.config.Resources.Capacity {
+			capacity[resourceName] = quantity
+			if reservedQuantity, ok := c.config.Resources.Reserved[resourceName]; ok {
+				allocatable[resourceName] = *resource.NewQuantity(quantity.Value()-reservedQuantity.Value(), quantity.Format)
```
timebertt, line 188: Can you use `Quantity.Sub` here to be on the safe side? Might not be relevant for integers like the number of shoots, but I would like to stick to the idiomatic use of `resource.Quantity`.
```suggestion
				allocatableQuantity := quantity.DeepCopy()
				allocatableQuantity.Sub(reservedQuantity)
				allocatable[resourceName] = allocatableQuantity
```

timebertt (commented):
Sorry for joining the review so late. I wasn't aware, that this PR would be merged before agreeing upon the GEP.

Nevertheless, I really like the PR, looks really good.
Now that this is already in, would it make sense to add a small doc about the new fields of the Seed and GardenletConfig and also the ShootedSeed flags?

I also have one minor code suggestion, maybe you can also address it in a follow up PR :)

stoyanr:
@timebertt Ok, I will do it. I assume the ShootedSeed flags should be added to `docs/usage/shooted_seed.md`, however I couldn't find any appropriate document for fields in the Seed `status`? Same for the `gardenlet` configuration, except `example/20-componentconfig-gardenlet.yaml`, where an example for `resources.capacity.shoots` is already added.

timebertt:
> I assume the ShootedSeed flags should be added to `docs/usage/shooted_seed.md`

Yes, that would be the right place.


> however I couldn't find any appropriate document for fields in the Seed `status`?

Hmm, yes. Probably a good fit for the capacity topic is `docs/concepts/scheduler.md`.

stoyanr:
@timebertt Done, see #3158

PR #3078, opened 2020-10-30
Ensure seed capacity for resources requested by shoots is not exceeded

### LINKED ISSUE #2938: Max cap for shoots per seed in scheduler

stoyanr:

/area robustness
/kind enhancement
/priority normal

Ensure seed capacity for resources requested by shoots is not exceeded, as described in GEP-13 (see #3017).

* Introduce resources requested by shoots via the new `resourceRequirements` field in the shoot spec.
* Validate `resourceRequirements.requests` (it is immutable).
* During scheduling, filter out candidates that don't have available capacity for any resource requested by the shoot.

Fixes #2938

* Based on #3075
* The new field in the shoot spec is named `resourceRequirements` rather than `resources` since there is already a `resources` field used for something else.
* A new `ResourceRequirements` Gardner type is introduced rather than using `corev1.ResourceRequirements`, since the scheduler is only interested in resource requests, but not limits. It is unclear whether resource limits could be useful in our case.

**Release note**:

```improvement operator
Shoots can requests resources via the new `resourceRequirements` field in the shoot spec. The scheduler ensures that only candidate seeds with available capacity for any resource requested by the shoot are considered during scheduling.
```


rfranzke:
Do we have concrete use-cases for allowing end-users to specify resource requirements for their shoots? The GEP states

> For example, if a shoot requests (via a `resourceRequirements` field) a certain number of persistent volumes, but the seed's allocatable amount of persistent volumes minus the number of persistent volumes already requested by other shoots scheduled onto that seed is lower, the shoot can't be scheduled onto that seed.

End-Users don't know/care what we do in the seed, i.e., they should not have the ability to configure anything like this on their shoots.

Still, I get the underlying motivation: Staying with the example of PVCs: a shoot needs a certain number of volumes in the seed, and we should ensure that it only gets scheduled when the max number of volumes the seed can handle is not exceeded, but isn't this rather something that is derived out of the concrete implementation how a shoot is brought up, hence, should not be configureable from outside (as it won't help anyways)?

stoyanr:
@rfranzke 

> Do we have concrete use-cases for allowing end-users to specify resource requirements for their shoots? 

I simply implemented partly the more advanced use case described in the GEP.

> End-Users don't know/care what we do in the seed, i.e., they should not have the ability to configure anything like this on their shoots.

I must admit I find this actually convincing. It's interesting that it's brought up here, while neither you nor anybody else objected in the GEP yet (neither in the comments, nor in the team discussion), even though this was there from the very beginning. Is it because it's easier to review code than documents? :-)

Could you please raise the same objection in the GEP? If no actual use-cases are brought up, then this should simply be dropped from the GEP I guess and this PR should be abandoned.

rfranzke:
> Is it because it's easier to review code than documents? :-)

Sure, definitely! Just kidding, I think this boils down to https://github.com/gardener/gardener/pull/3017#discussion_r515927004. The GEP covers a lot of topics, some more detailed than others, and it's somewhat hard to grasp the whole e2e story with the many details at once. As stated in this linked comment: I like that we can have an overview document which highlights all the use-cases and topics related to automatic seed management, but it will be much easier for everybody if these use-cases will be split into smaller sub-topics (that either deserve their own GEP with more details, or at least an issue with more details (depending on your preference) where we can discuss the details (ideally, before we start the implementation).

stoyanr:
Closing as discussed above.

PR #3079, opened 2020-10-30
[ci:component:github.com/gardener/autoscaler-v0.10.0:v0.10.0->v0.12.0]

gardener-robot-ci-2:
**Release Notes*:
``` improvement operator github.com/gardener/autoscaler #59 @hardikdr
Autoscaler avoids scaling down the machines which are already being terminated.
```

``` improvement operator github.com/gardener/autoscaler #58 @jsravn
MCM provider now also returns unregistered nodes to Autoscaler. This change enables autoscaler to pick up an alternate worker-pool if the chosen one can't be scaled-up.
```


ialidzhikov:
@hardikdr @AndreasBurger , could you please check why the ci/cd created this PR with no changed files?

/close

AndreasBurger:
This repository's images.yaml defines autoscaler twice (https://github.com/gardener/gardener/blob/master/charts/images.yaml#L75-L84) which the `component-descriptor` script translates to two dependencies for the autoscaler-component (one for `v0.10.0` and one for `v0.12.0`) since the repositories start with `eu.gcr.io/gardener-project`.
So this component now recieves update-PRs twice. This one was for the v0.10.0 autoscaler.

We've added an additional env-var (as of yet undocumented) `LOCAL_DEPENDENCY_NAME` that contains the local name of the component that is being updated to distinguish between components that are present multiple times. Unfortunately that name defaults to `<component name>-<component version>` which is unique, but also changes with every new version.

So we'd either need to change the component-descriptor-script to not add autoscaler twice or give the instances explicit names so we can properly handle them during upgrade-PR-creation. Let's talk about that on Monday.

prashanth26:
@AndreasBurger - We see this happens with [every CA release](https://github.com/gardener/gardener/pull/3272). Any ideas on how we could avoid this? Could modify something in the chart itself?

PR #3080, opened 2020-10-30
Move admission controller to separate Helm chart

### LINKED ISSUE #2983: Race condition for gardener/controlplane/charts/application chart

/area robustness
/kind bug
/priority normal

**What happened**:
The chart `gardener/charts/gardener/controlplane/charts/application` was used to deploy necessary assets to the virtual garden cluster. Sometimes though, the containing `ServiceAccounts` cannot be created by KCM because of the [ValidatingWebhookConfiguration](https://github.com/gardener/gardener/blob/b7449f036123a01398c3dd10726bde5d5f296538/charts/gardener/controlplane/charts/application/templates/validatingwebhook-admission-controller.yaml#L1) which is part of the very same chart.

The `ServiceAccounts` to be reconciled successfully, i.e. `Secrets` with service account tokens are created.

Deploy chart `gardener/charts/gardener/controlplane/charts/application` with `.Values.global.admission.enabled: true`.

As this is a race condition, the error is not reproducible every time.

/cc @ialidzhikov thanks for the analysis
/cc @Diaphteiros 

**Environment**:

- Gardener version: v1.11.x

timuthy:

/area delivery
/kind bug
/priority normal

This PR moves Gardener Admission Controller related assets to a separate chart to avoid race conditions by the `ValidatinWebhookConfiguration` described in issue #2983. The idea is that a service account is deployed and prepared by deploying the Gardener control plane chart before continuing with the deployment of the new admission chart.

/cc @Diaphteiros 

Fixes #2983

**Release note**:

```noteworthy operator
⚠️ The Gardener Admission Controller has moved from the control plane (/charts/gardener/controlplane) to a separate admission chart (/charts/gardener/admission). Operators are advised to deploy the admission chart always after the control plane chart, as stated in the documentation. Please adjust your Gardener deployments accordingly.
```


rfranzke (changes requested):
/assign

Please adapt

* https://github.com/gardener/gardener/blob/master/charts/gardener/controlplane/Chart.yaml#L2
* https://github.com/gardener/gardener/blob/master/charts/gardener/controlplane/local-values.yaml#L64-L146

Generally, I see that there is no other way / I don't have a better idea, but I don't like that there are still artefacts left in the `controlplane` chart that belong to the `admission` chart (service account creation).

The other approach (not splitting the chart) would be to adapt the `ValidatingWebhookConfiguration` for the `validate-kubeconfig-secrets.gardener.cloud` webhook to exclude the `garden` namespace (I guess we can safely assume that admins/operators don't try to inject malicious kubeconfigs into their own systems). WDYT?

timuthy:
> Generally, I see that there is no other way / I don't have a better idea, but I don't like that there are still artefacts left in the controlplane chart that belong to the admission chart (service account creation).

Yes, you have a point there and it bothers me as well, tbh.

> The other approach (not splitting the chart) would be to adapt the ValidatingWebhookConfiguration for the validate-kubeconfig-secrets.gardener.cloud webhook to exclude the garden namespace (I guess we can safely assume that admins/operators don't try to inject malicious kubeconfigs into their own systems). WDYT?

Sure, why not. Thanks for the suggestion! Let me close this one and continue with excluding the `garden` namespace.

PR #3082, opened 2020-10-31
Add apiserver-proxy webhook to all SNI enabled clusters

mvladev:

/area networking
/kind enhancement 
/priority normal

The API Server proxy webhook adds `KUBERNETES_SERVICE_HOST` environment variable to all SNI enabled Shoot clusters. It skips additional network
hop.

It can be disabled for individual Pods or entire namespaces.

See more details at https://github.com/gardener/apiserver-proxy

Part of #2942

~This a draft until gardener/apiserver-proxy#8 is merged.~

**Release note**:

```noteworthy operator
A new webhook `mutatingwebhookconfigurations.admissionregistration.k8s.io` is deployed for all `APIServerSNI`-enabled clusters. It's running as a sidecar to the KubeAPI Server.
```

```action user
A new webhook `mutatingwebhookconfigurations.admissionregistration.k8s.io` is deployed for all `APIServerSNI`- enabled clusters. It adds `KUBERNETES_SERVICE_HOST` environment variable pointing to the upstream Kube API Server. To disable this behavior:

- label your Pods with `apiserver-proxy.networking.gardener.cloud/inject: disable`
- or label the entire namespace with `apiserver-proxy.networking.gardener.cloud/inject: disable`
- or label your Shoot resource with `alpha.featuregates.shoot.gardener.cloud/apiserver-sni-pod-injector: disable` to disable it cluster-wide.
```

```action user
For `APIServerSNI`-enabled clusters, Pods talking to the Kube API Server need to be allowed to connect to  `coredns` running in `kube-system` namespace in order to resolve the hostname of the Kube API server. It also needs to have access to the IP from the `default/kubernetes` endpoint and the upstream IP of of the kube-apiserver. 

If the Pod:

- is not matched by any `NetworkPolicy` - no action is required. 
- is not injected with `KUBERNETES_SERVICE_HOST`, because the feature is disabled - no action is required.
- is matched by `NetworkPolicies` allowing ingress to `coredns` in `kube-system` and allows traffic top the `default/kubernetes` endpoint and the upstream upstream IP of of the kube-apiserver - no action is required.
- is matched by `NetworkPolicies` that do not allow access to `coredns` in `kube-system` and/or do not allows traffic top the `default/kubernetes` endpoint and/or the upstream upstream IP of of the kube-apiserver - a `NetworkPolicy` allowing such egress must be added e.g.:

  > apiVersion: networking.k8s.io/v1
  > kind: NetworkPolicy
  > metadata:
  >   name: allow-to-apiserver
  > spec:
  >   podSelector: {}
  >   egress:
  >   - to:
  >     - ipBlock:
  >         cidr: <IP from default/kubernetes endpoint>/32
  >     - ipBlock:
  >         cidr: <ip from apiserver FQDN e.g. nslookup api.foo.bar>/32
  >   - ports:
  >     - port: 8053
  >       protocol: UDP
  >     - port: 8053
  >       protocol: TCP
  >     to:
  >     - podSelector:
  >         matchExpressions:
  >         - key: k8s-app
  >           operator: In
  >           values:
  >           - kube-dns
  >       namespaceSelector:
  >         matchLabels:
  >           gardener.cloud/purpose: kube-system
  >   policyTypes:
  >   - Egress
  >   - Ingress
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -253,3 +253,7 @@ images:
   sourceRepository: github.com/gardener/apiserver-proxy
   repository: eu.gcr.io/gardener-project/gardener/apiserver-proxy
   tag: "v0.1.0"
+- name: apiserver-proxy-pod-webhook
+  sourceRepository: github.com/gardener/apiserver-proxy
+  repository: eu.gcr.io/gardener-project/gardener/apiserver-proxy-pod-webhook
+  tag: "v0.2.0"
--- a/charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
+++ b/charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
@@ -301,7 +301,7 @@ spec:
           readOnly: true
           {{- end }}
         {{- end }}
-      {{- if .Values.sni.enabled }}
+      {{- if .Values.sni.podMutatorEnabled }}
       - name: apiserver-proxy-pod-mutator
         image: {{ index .Values.images "apiserver-proxy-pod-webhook" }}
         args:
--- a/charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
+++ b/charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
@@ -301,6 +301,20 @@ spec:
           readOnly: true
           {{- end }}
         {{- end }}
+      {{- if .Values.sni.enabled }}
+      - name: apiserver-proxy-pod-mutator
+        image: {{ index .Values.images "apiserver-proxy-pod-webhook" }}
+        args:
+        - --apiserver-fqdn={{ .Values.sni.apiserverFQDN }}
+        - --host=localhost
+        - --cert-dir=/srv/kubernetes/apiserver
+        - --cert-name=kube-apiserver.crt
+        - --key-name=kube-apiserver.key
+        - --port=9443
+        volumeMounts:
+        - name: kube-apiserver
+          mountPath: /srv/kubernetes/apiserver
+      {{- end }}
       {{- if .Values.konnectivityTunnel.enabled }}
       - name: konnectivity-server
         image: {{ index .Values.images "konnectivity-server" }}
--- a/charts/seed-controlplane/charts/kube-apiserver/templates/hvpa.yaml
+++ b/charts/seed-controlplane/charts/kube-apiserver/templates/hvpa.yaml
@@ -76,6 +76,10 @@ spec:
               minAllowed:
                 memory: 400M
                 cpu: 300m
+            {{- if .Values.sni.podMutatorEnabled }}
+            - containerName: apiserver-proxy-pod-mutator
+              mode: "Off"
+            {{- end }}
             - containerName: vpn-seed
               mode: "Off"
   weightBasedScalingIntervals:
--- a/charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
+++ b/charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
@@ -1,0 +1,45 @@
+apiVersion: {{ include "webhookadmissionregistration" . }}
+kind: MutatingWebhookConfiguration
+metadata:
+  name: apiserver-proxy.networking.gardener.cloud
+  annotations:
+    networking.gardener.cloud/description: |-
+      This webhook adds `KUBERNETES_SERVICE_HOST` environment variable
+      to all containers and init containers matched by it.
+  labels:
+    gardener.cloud/role: system-component
+    origin: gardener
+webhooks:
+- admissionReviewVersions:
+  - v1beta1
+  clientConfig:
+    caBundle: {{ .Values.webhook.caBundle }}
+    url: https://127.0.0.1:9443/webhook/pod-apiserver-env
+  failurePolicy: Ignore
+  matchPolicy: Exact
+  name: apiserver-proxy.networking.gardener.cloud
+  namespaceSelector:
+    matchExpressions:
+    - key: apiserver-proxy.networking.gardener.cloud/inject
+      operator: NotIn
+      values:
+      - disable
+  objectSelector:
+    matchExpressions:
+    - key: apiserver-proxy.networking.gardener.cloud/inject
+      operator: NotIn
+      values:
+      - disable
+  reinvocationPolicy: Never
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    resources:
+    - pods
+    scope: '*'
+  sideEffects: None
+  timeoutSeconds: 2
--- a/charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
+++ b/charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.podMutatorEnabled }}
 apiVersion: {{ include "webhookadmissionregistration" . }}
 kind: MutatingWebhookConfiguration
 metadata:
@@ -43,3 +44,4 @@ webhooks:
     scope: '*'
   sideEffects: None
   timeoutSeconds: 2
+{{- end }}
--- a/docs/usage/apiserver-sni-injection.md
+++ b/docs/usage/apiserver-sni-injection.md
@@ -11,7 +11,7 @@ In some cases it's desirable to opt-out of Pod injection:
 - DNS is disabled on that individual Pod, but it still needs to talk to the kube-apiserver.
 - Want to test the `kube-proxy` and `kubelet` in-cluster discovery.
 
-### Opt-out of pod injection for specific pods.
+### Opt-out of pod injection for specific pods
 
 To opt out of the injection, the Pod should be labeled with `apiserver-proxy.networking.gardener.cloud/inject: disable` e.g.:
 
@@ -60,3 +60,24 @@ kubectl label namespace my-namespace apiserver-proxy.networking.gardener.cloud/i
 ```
 
 > NOTE: Please be aware that it's not possible to disable injection on namespace level and enable it for individual pods in it.
+
+### Opt-out of pod injection for the entire cluster
+
+If the injection is causing problems for different workloads and ignoring individual pods or namespaces is not possible, then the feature could be disabled for the entire cluster with the `alpha.featuregates.shoot.gardener.cloud/apiserver-sni-pod-injector` annotation with value `disable` on the `Shoot` resource itself:
+
+```yaml
+apiVersion: core.gardener.cloud/v1beta1
+kind: Shoot
+metadata:
+  annotations:
+    alpha.featuregates.shoot.gardener.cloud/apiserver-sni-pod-injector: 'disable'
+  name: my-cluster
+```
+
+or via `kubectl` for existing shoot cluster:
+
+```console
+kubectl label shoot my-cluster alpha.featuregates.shoot.gardener.cloud/apiserver-sni-pod-injector=disable
+```
+
+> NOTE: Please be aware that it's not possible to disable injection on cluster level and enable it for individual pods in it.
--- a/docs/usage/apiserver-sni-injection.md
+++ b/docs/usage/apiserver-sni-injection.md
@@ -1,0 +1,62 @@
+# APIServerSNI environment variable injection
+
+If the Gardener administrator has enabled `APIServerSNI` feature gate for a particular Seed cluster, then in each Shoot cluster's `kube-system` namespace a `DaemonSet` called `apiserver-proxy` is deployed. It routes traffic to the upstream Shoot Kube APIServer. See the [APIServer SNI GEP](../proposals/08-shoot-apiserver-via-sni.md) for more details.
+
+To skip this extra network hop, a [mutating webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) called `apiserver-proxy.networking.gardener.cloud` is deployed next to the API server in the Seed. It adds `KUBERNETES_SERVICE_HOST` environment variable to each container and init container that do not specify it. See the webhook [repository](https://github.com/gardener/apiserver-proxy/) for more information.
+
+## Opt-out of pod injection
+
+In some cases it's desirable to opt-out of Pod injection:
+
+- DNS is disabled on that individual Pod, but it still needs to talk to the kube-apiserver.
+- Want to test the `kube-proxy` and `kubelet` in-cluster discovery.
+
+### Opt-out of pod injection for specific pods.
+
+To opt out of the injection, the Pod should be labeled with `apiserver-proxy.networking.gardener.cloud/inject: disable` e.g.:
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: nginx
+  labels:
+    app: nginx
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: nginx
+  template:
+    metadata:
+      labels:
+        app: nginx
+        apiserver-proxy.networking.gardener.cloud/inject: disable
+    spec:
+      containers:
+      - name: nginx
+        image: nginx:1.14.2
+        ports:
+        - containerPort: 80
+```
+
+### Opt-out of pod injection on namespace level
+
+To opt out of the injection of **all** Pods in a namespace, you should label your namespace with `apiserver-proxy.networking.gardener.cloud/inject: disable` e.g.:
+
+```yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    apiserver-proxy.networking.gardener.cloud/inject: disable
+  name: my-namespace
+```
+
+or via `kubectl` for existing namespace:
+
+```console
+kubectl label namespace my-namespace apiserver-proxy.networking.gardener.cloud/inject=disable
+```
+
+> NOTE: Please be aware that it's not possible to disable injection on namespace level and enable it for individual pods in it.
--- a/pkg/gardenlet/controller/federatedseed/networkpolicy/hostnameresolver/resolver.go
+++ b/pkg/gardenlet/controller/federatedseed/networkpolicy/hostnameresolver/resolver.go
@@ -1,0 +1,190 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hostnameresolver
+
+import (
+	"context"
+	"net"
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+type resolver struct {
+	lock          sync.RWMutex
+	upstreamFQDN  string
+	upstreamPort  int32
+	refreshTicker *time.Ticker
+	onUpdate      func()
+	log           logrus.FieldLogger
+	addrs         []string
+	// used for testing
+	lookup lookup
+}
+
+type noOpResover struct{}
+
+type lookup interface {
+	LookupHost(ctx context.Context, host string) (addrs []string, err error)
+}
+
+// Provider allows to start and attach callbacks for a specific host
+// resolution updates.
+type Provider interface {
+	HasSynced() bool
+	Start(ctx context.Context)
+	WithCallback(onUpdate func())
+	HostResolver
+}
+
+// HostResolver is used for getting endpoint subsets with resolved IPs.
+type HostResolver interface {
+	Subset() []corev1.EndpointSubset
+}
+
+// NewProvider returns a Provider for a specific host and port with resync
+// indicating how often the hostname resolution is happening.
+func NewProvider(host string, port string, log logrus.FieldLogger, resync time.Duration) Provider {
+	return &resolver{
+		upstreamFQDN:  host,
+		upstreamPort:  intstr.Parse(port).IntVal,
+		refreshTicker: time.NewTicker(resync),
+		log:           log,
+		lookup:        net.DefaultResolver,
+	}
+}
+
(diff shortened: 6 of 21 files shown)


charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - key: apiserver-proxy.networking.gardener.cloud/inject
+      operator: NotIn
+      values:
+      - disable
+  objectSelector:
+    matchExpressions:
+    - key: apiserver-proxy.networking.gardener.cloud/inject
+      operator: NotIn
+      values:
+      - disable
```
rfranzke, line 33: I guess this breaks for shoots with an old Kubernetes version, right? Gardener is still capable of managing Kubernetes 1.10, so this whole manifest should be adapted to cater with all versions.

docs/usage/apiserver-sni-injection.md
```diff
@@ -0,0 +1,55 @@
+# APIServerSNI environment variable injection
+
+If the Gardener administrator has enabled `APIServerSNI` for a particular Seed cluster, then in each Shoot cluster's `kube-system` namespace a `DaemonSet` called `apiserver-proxy` is deployed. It routes traffic to the upstream Shoot Kube APIServer. See the [APIServer SNI GEP](../proposals/08-shoot-apiserver-via-sni.md) for more details.
+
+To skip this extra network hop, a [mutating webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) called `apiserver-proxy.networking.gardener.cloud` is deployed. It adds `KUBERNETES_SERVICE_HOST` environment variable to each container and init container that do not specify it. See the webhooks's [repository](https://github.com/gardener/apiserver-proxy/) for more information.
+
+## Opt-out of Pods injection
+
+To opt out of the injection, the Pod should be labeled with `apiserver-proxy.networking.gardener.cloud/inject: disable` e.g.:
```
rfranzke, line 16: Can you add a sentence in which situations/scenarios one might want to opt out?

docs/usage/apiserver-sni-injection.md
```diff
@@ -0,0 +1,55 @@
+# APIServerSNI environment variable injection
+
+If the Gardener administrator has enabled `APIServerSNI` for a particular Seed cluster, then in each Shoot cluster's `kube-system` namespace a `DaemonSet` called `apiserver-proxy` is deployed. It routes traffic to the upstream Shoot Kube APIServer. See the [APIServer SNI GEP](../proposals/08-shoot-apiserver-via-sni.md) for more details.
```
rfranzke:
```suggestion
If the Gardener administrator has enabled `APIServerSNI` feature gate for a particular Seed cluster, then in each Shoot cluster's `kube-system` namespace a `DaemonSet` called `apiserver-proxy` is deployed. It routes traffic to the upstream Shoot Kube APIServer. See the [APIServer SNI GEP](../proposals/08-shoot-apiserver-via-sni.md) for more details.
```

docs/usage/apiserver-sni-injection.md
```diff
@@ -0,0 +1,55 @@
+# APIServerSNI environment variable injection
```
rfranzke, line 1: Can you add the document to the [index](https://github.com/gardener/gardener/blob/master/docs/README.md)?

rfranzke (changes requested):
/assign

charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
mvladev: How so? This resource is created by Gardener resource manager. It'll send this field, but it'll be ignored by the API server.

rfranzke: Is it? OK, if so, fine. Just wanted to point that out :)

mvladev:
```
Request Body: {"THIS_FIELD_DOES_NOT_EXIST":"bar","apiVersion":"v1","kind":"ConfigMap","metadata":{"name":"foo","namespace":"default"}}
curl -k -v -XPOST  -H "Content-Type: application/json" -H "Accept: application/json" -H "User-Agent: kubectl/v1.18.6 (darwin/amd64) kubernetes/dff82dc" 'https://kubernetes.docker.internal:6443/api/v1/namespaces/default/configmaps'
POST https://kubernetes.docker.internal:6443/api/v1/namespaces/default/configmaps 201 Created in 8 milliseconds
Response Headers:
    Content-Length: 258
    Date: Mon, 02 Nov 2020 07:29:04 GMT
    Cache-Control: no-cache, private
    Content-Type: application/json
Response Body: {"kind":"ConfigMap","apiVersion":"v1","metadata":{"name":"foo","namespace":"default","selfLink":"/api/v1/namespaces/default/configmaps/foo","uid":"eea2f54b-6ac8-4e0b-b88c-05b5740ddce4","resourceVersion":"1887256","creationTimestamp":"2020-11-02T07:29:04Z"}}
configmap/foo created
```

The client is sending some `THIS_FIELD_DOES_NOT_EXIST` field which is not stored by the API server. The validation is only on the client side via OpenAPI.

mvladev:
@rfranzke the PR was updated to include the newly released webhook `v.0.2.0`. Have another look.

charts/images.yaml
```diff
@@ -253,3 +253,7 @@ images:
   sourceRepository: github.com/gardener/apiserver-proxy
   repository: eu.gcr.io/gardener-project/gardener/apiserver-proxy
   tag: "v0.1.0"
```
rfranzke, line 255: Can we update this to `v0.2.0` as well?
```suggestion
  tag: "v0.2.0"
```

charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: {{ include "webhookadmissionregistration" . }}
+kind: MutatingWebhookConfiguration
+metadata:
+  name: apiserver-proxy.networking.gardener.cloud
+  annotations:
+    networking.gardener.cloud/description: |-
```
rfranzke, line 7:
```suggestion
    gardener.cloud/description: |-
```

rfranzke (commented):
Very nice PR @mvladev

charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
mvladev: We already have the following:

```yaml
resources.gardener.cloud/description: |-
  DO NOT EDIT - This resource is managed by gardener-resource-manager.
```

it's better not to use the `gardener.cloud/description` as it might be used for other things in the future.

charts/images.yaml
mvladev: There are no functional changes between `0.1.0` and `0.2.0`. This will cause unneeded rollout of the daemonset.

rfranzke: OK

charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
rfranzke: Agreed

rfranzke (dismissed):
/lgtm

charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
```diff
@@ -295,6 +295,20 @@ spec:
...
+      - name: apiserver-proxy-pod-mutator
+        image: {{ index .Values.images "apiserver-proxy-pod-webhook" }}
+        args:
+        - --apiserver-fqdn={{ .Values.sni.apiserverFQDN }}
+        - --host=localhost
+        - --cert-dir=/srv/kubernetes/apiserver
+        - --cert-name=kube-apiserver.crt
+        - --key-name=kube-apiserver.key
+        - --port=9443
+        volumeMounts:
```
timebertt, line 314: Should we add resource requests and limits for this container?

charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
```diff
@@ -0,0 +1,45 @@
...
+      values:
+      - disable
+  reinvocationPolicy: Never
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
```
timebertt, line 41: I'm not so sure about this question, but I want to double-check: Should `UPDATE` be added here?
It's probably a super rare use case to change one of the few updatable fields in a Pod with some update mechanism other than patch. Nevertheless, this could cause the API server to return an error indicating, that `env` cannot be changed, although the user didn't change anything. So this could _potentially_ break some users, who do such weird stuff, right?

docs/usage/apiserver-sni-injection.md
```diff
@@ -0,0 +1,62 @@
...
+
+If the Gardener administrator has enabled `APIServerSNI` feature gate for a particular Seed cluster, then in each Shoot cluster's `kube-system` namespace a `DaemonSet` called `apiserver-proxy` is deployed. It routes traffic to the upstream Shoot Kube APIServer. See the [APIServer SNI GEP](../proposals/08-shoot-apiserver-via-sni.md) for more details.
+
+To skip this extra network hop, a [mutating webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) called `apiserver-proxy.networking.gardener.cloud` is deployed. It adds `KUBERNETES_SERVICE_HOST` environment variable to each container and init container that do not specify it. See the webhooks's [repository](https://github.com/gardener/apiserver-proxy/) for more information.
```
timebertt:
```suggestion
To skip this extra network hop, a [mutating webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) called `apiserver-proxy.networking.gardener.cloud` is deployed next to the API server in the Seed. It adds `KUBERNETES_SERVICE_HOST` environment variable to each container and init container that do not specify it. See the webhooks's [repository](https://github.com/gardener/apiserver-proxy/) for more information.
```

timebertt (commented):
I have only a few minor question, rest looks good.
Appreciate the doc for this functionality!

charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
mvladev: https://github.com/kubernetes/kubernetes/blob/a86afc12df023e8a45a1c3b6523aed38c3e6cd18/pkg/apis/core/validation/validation.go#L3927-L3930

You cannot update this field.

timebertt: Yes, exactly. And that's why an update request with the env var set will fail. 
If the webhook would also add it on updates, the field won't be changed and the request succeeds.
As said, this is a pretty rare use case. But nevertheless, should we consider it?

mvladev: > As said, this is a pretty rare use case. But nevertheless, should we consider it?

The request would be invalid in the first place (you'll need to get resource to perform such operation), so I guess not.

charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
mvladev: I'll rather wait and gather some metrics from various workloads on different clusters before adding such requests / limits.

mvladev:
I've updated the release notes - have a look  at them.

mvladev:
One side-effect which might cause problem is that Pods would now be required to have network connectivity to core-dns.

danielfoehrKn:
> One side-effect which might cause problem is that Pods would now be required to have network connectivity to core-dns.

Could not Gardener deploy the in the release note mentioned Network policy to ensure that every pod can resolve the _KUBERNETES_SERVICE_HOST_  ? Or is that a security concern?

danielfoehrKn:
Based on my limited understanding: -  I thought one  reason (besides adding the FQDN of the Shoot API server) for the apiserver envoy proxy deamon is to ensure that every request a pod makes against the Shoot API Server uses the PROXY protocol - this is what the envoy ingress gateway  in the Seed listens for?
 
However, I do not see this EnvoyFilter configuration any more but only regular TCP  -> does this explain why requests from the pod via TCP work?

rfranzke:
> One side-effect which might cause problem is that Pods would now be required to have network connectivity to core-dns.

Can you elaborate? Why would that be a problem?

mvladev:
> Could not Gardener deploy the in the release note mentioned Network policy to ensure that every pod can resolve the KUBERNETES_SERVICE_HOST ? 

Network policies are namespaced resource, so no.

> apiserver envoy proxy deamon is to ensure that every request a pod makes against the Shoot API Server uses the PROXY protocol - this is what the envoy ingress gateway in the Seed listens for?

Yes, the ingress gateway in the Seed listens for

- TLS on port 443 - this is where your `api.foo.bar` requests go to.
- PROXY Protocol on port 8443 - this is where the apiserver-proxy sends the requests to.

> However, I do not see this EnvoyFilter configuration any more but only regular TCP -> does this explain why requests from the pod via TCP work?

The request flow is the following:

```text
$ curl https://kubernetes.default.svc.cluster.local

Pod -- TLS over TCP --> API Server Proxy -- PROXY PROTOCOL --> Istio IngressGateway -- TLS over TCP --> kube-apiserver

$ curl https://api.foo.bar #(where Istio Ingress Gateway knows that clients requesting api.foo.bar must send traffic to that apiserver)

Pod -- TLS over TCP --> Istio IngressGateway -- TLS over TCP --> kube-apiserver
```

> Can you elaborate? Why would that be a problem?

Users might already have policies that allow traffic only to the `default/kubernetes` endpoint IP. This PR changes where you'll need access to DNS **and** the IP of the API server (nslookup api.foo.bar)

Okay so this is the _I want to talk only to API server_ NetworkPolicy. It allows access via `kubernetes.default.svc.cluster.local` and `api.foo.bar`

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: to-apiserver
spec:
  podSelector: {}
  egress:
  - to:
    - ipBlock:
        cidr: <IP from default/kubernetes endpoint>/32
    - ipBlock:
        cidr: <ip from apiserver FQDN e.g. nslookup api.foo.bar>/32
  - ports:
    - port: 8053
      protocol: UDP
    - port: 8053
      protocol: TCP
    to:
    - podSelector:
        matchExpressions:
        - key: k8s-app
          operator: In
          values:
          - kube-dns
      namespaceSelector:
        matchLabels:
          gardener.cloud/purpose: kube-system
  policyTypes:
  - Egress
  - Ingress
```

mvladev:
I just noticed that we probably have to fix https://github.com/gardener/gardener/blob/2fb3b07c5c741b432889ce155a89a82cd88532b1/pkg/gardenlet/controller/federatedseed/networkpolicy/controller.go as well to handle this change as well.

/hold

charts/shoot-core/components/charts/apiserver-proxy/templates/apiserver-proxy-pod-mutatingwebhook.yaml
timebertt: Yeah, that's also right. It was just a minor question anyways, it's not too relevant.
But I guess, it wouldn't hurt to add `UPDATE` as well (so the env vars will always be ensured, no matter if `CREATE` or `UPDATE`), would it?

pkg/operation/common/types.go
```diff
@@ -489,8 +489,11 @@ const (
 	// APIServerProxyImageName is the image of apiserver-proxy
 	APIServerProxyImageName = "apiserver-proxy"
 
-	// APIServerPorxySidecarImageName is the image of apiserver-proxy sidecar
-	APIServerPorxySidecarImageName = "apiserver-proxy-sidecar"
+	// APIServerProxySidecarImageName is the image of apiserver-proxy sidecar.
+	APIServerProxySidecarImageName = "apiserver-proxy-sidecar"
+
+	// APIServerProxyPodMutatorWebhookImageName is the image of piserver-proxy pod mutator webhook.
```
ialidzhikov:
```suggestion
	// APIServerProxyPodMutatorWebhookImageName is the image of apiserver-proxy pod mutator webhook.
```

rfranzke:
To proceed, the plan would be to still enable this behaviour by default, but to allow people to opt-out of it and live with the consequences until they adapt their workload to cater with it.

/status author-action

mvladev:
> to allow people to opt-out

The opt-out solution is already there - it can be disabled on namespace or pod level. The question is - is it sufficient?

rfranzke:
As discussed via Slack, a general opt-out via an annotation on the shoot (because the feature is still alpha) would probably be needed to not force people from handling each and every namespace/pod in their cluster separately.

mvladev:
I've updated the PR with fix for the federated network policy controller.

charts/seed-controlplane/charts/kube-apiserver/templates/hvpa.yaml
```diff
@@ -76,6 +76,10 @@ spec:
               minAllowed:
                 memory: 400M
                 cpu: 300m
+            {{- if .Values.sni.enabled }}
```
rfranzke:
```suggestion
            {{- if .Values.sni.podMutatorEnabled }}
```

rfranzke (commented):
looks great, no major comments

rfranzke (dismissed):
/lgtm

docs/usage/apiserver-sni-injection.md
```diff
@@ -0,0 +1,83 @@
...
+### Opt-out of pod injection for the entire cluster
+
+If the injection is causing problems for different workloads and ignoring individual pods or namespaces is not possible, then the feature could be disabled for the entire cluster with the `alpha.featuregates.shoot.gardener.cloud/apiserver-sni-pod-injector` annotation with value `disable` on the `Shoot` resource itself:
+
+```yaml
+apiVersion: core.gardener.cloud/v1beta1
+kind: Shoot
+metadata:
+  annotations:
+    alpha.featuregates.shoot.gardener.cloud/apiserver-sni-pod-injector: 'disable'
```
ialidzhikov, line 73: For example the already existsing annotation `alpha.featuregates.shoot.gardener.cloud/konnectivity-tunnel` accepts `true/false`, but this annotation accepts `disable`. Wouldn't it be better to have unified behaviour of similar annotations?

```suggestion
    alpha.featuregates.shoot.gardener.cloud/apiserver-sni-pod-injector: 'false'
```

pkg/gardenlet/controller/federatedseed/networkpolicy/hostnameresolver/resolver.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return len(l.addrs) > 0
+}
+
+// Start waits for stopCtx to be done and resolves the upstream
+// hostname every resync period.
+// Updates are send if returned hosts are changed.
+func (l *resolver) Start(stopCtx context.Context) {
+	updateFunc := func() {
+		arrds, err := l.lookup.LookupHost(stopCtx, l.upstreamFQDN)
```
ialidzhikov:
```suggestion
		addresses, err := l.lookup.LookupHost(stopCtx, l.upstreamFQDN)
```

pkg/gardenlet/controller/federatedseed/networkpolicy/hostnameresolver/resolver.go
```diff
@@ -0,0 +1,190 @@
...
+
+			l.log.WithField("resolvedIPs", l.addrs).Infoln("updated resolved addresses")
+
+			if l.onUpdate != nil {
+				l.onUpdate()
+			}
+		}
+	}
+
+	// start the update in the begginging
```
ialidzhikov:
```suggestion
	// start the update in the beginning
```

pkg/gardenlet/controller/federatedseed/networkpolicy/hostnameresolver/resolver_test.go
```diff
@@ -0,0 +1,177 @@
...
+		}, time.Millisecond*3, time.Millisecond).Should(ConsistOf(corev1.EndpointSubset{
+			Addresses: []corev1.EndpointAddress{
+				{IP: "1.2.3.4"}, {IP: "5.6.7.8"},
+			},
+			Ports: []corev1.EndpointPort{{Protocol: corev1.ProtocolTCP, Port: 1234}},
+		}))
+
+		f.setAddrs([]string{"5.6.7.8"})
+
+		Eventually(func() []corev1.EndpointSubset { //nolint:unlambda
```
ialidzhikov, line 166: Why `//nolint:unlambda` is required and what it does? `make check` is passing when it is not present.

pkg/gardenlet/controller/federatedseed/networkpolicy/hostnameresolver/resolver.go
```diff
@@ -0,0 +1,190 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+type resolver struct {
+	lock          sync.RWMutex
+	upstreamFQDN  string
+	upstreamPort  int32
+	refreshTicker *time.Ticker
+	onUpdate      func()
+	log           *logrus.Entry
```
ialidzhikov:
```suggestion
	log           logrus.FieldLogger
```

pkg/gardenlet/controller/federatedseed/networkpolicy/hostnameresolver/resolver_test.go
```diff
@@ -0,0 +1,177 @@
...
+		updateCalled = false
+		lock = sync.Mutex{}
+		f = &fakeLookup{
+			addrs: []string{"5.6.7.8", "1.2.3.4"}, // sorting check
+		}
+		r = &resolver{
+			lookup:        f,
+			upstreamPort:  1234,
+			refreshTicker: time.NewTicker(time.Millisecond * 3),
+			log:           logrus.NewEntry(logger.NewNopLogger()),
```
ialidzhikov:
```suggestion
			log:           logger.NewNopLogger(),
```

docs/usage/apiserver-sni-injection.md
mvladev: The connectivity tunnel can be enabled on demand. This is part of another feature and cannot be enabled on its own.

pkg/gardenlet/controller/federatedseed/networkpolicy/hostnameresolver/resolver_test.go
mvladev: This is a false-positive warning on the linter. For example visual studio code with gopls backend complains about it.

rfranzke (approved):

ialidzhikov (approved):
/lgtm

PR #3084, opened 2020-11-02
Enhance error code mapping

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal

Enhance error code mapping

**Release note**:

```improvement operator
NONE
```


pkg/apis/core/v1beta1/helper/errors.go
```diff
@@ -52,8 +52,8 @@ func (e *ErrorWithCodes) Error() string {
...
-	dependenciesRegexp           = regexp.MustCompile(`(?i)(PendingVerification|Access Not Configured|accessNotConfigured|DependencyViolation|OptInRequired|DeleteConflict|Conflict|inactive billing state|ReadOnlyDisabledSubscription|is already being used|InUseSubnetCannotBeDeleted|VnetInUse|timeout while waiting for state to become|InvalidCidrBlock|already busy for|InsufficientFreeAddressesInSubnet|InternalServerError|RetryableError|Future#WaitForCompletion: context has been cancelled|internalerror|internal server error)`)
+	insufficientPrivilegesRegexp = regexp.MustCompile(`(?i)(AccessDenied|OperationNotAllowed|Error 403)`)
+	dependenciesRegexp           = regexp.MustCompile(`(?i)(PendingVerification|Access Not Configured|accessNotConfigured|DependencyViolation|OptInRequired|DeleteConflict|Conflict|inactive billing state|ReadOnlyDisabledSubscription|is already being used|InUseSubnetCannotBeDeleted|VnetInUse|timeout while waiting for state to become|InvalidCidrBlock|already busy for|InsufficientFreeAddressesInSubnet|InternalServerError|RetryableError|Future#WaitForCompletion: context has been cancelled|internalerror|internal server error|PendingMachines|DrainingMachines)`)
```
timebertt: Are `PendingMachines|DrainingMachines` really dependencies gardener is waiting on?
E.g. persistent `DrainingMachines` is often caused by misconfigured PDBs, so should we categorize that as a configuration problem?
And should we really just ignore `PendingMachines`? This might still indicate an issue, the operator should maybe look into.

rfranzke: > Are PendingMachines|DrainingMachines really dependencies gardener is waiting on?
> E.g. persistent DrainingMachines is often caused by misconfigured PDBs, so should we categorize that as a configuration problem?

I was thinking about introducing a new error code, `ERR_SCALING` or something similar, but this comes with some additional efforts (dashboard adaptation, etc.), so I just added them to existing codes for now.

> And should we really just ignore PendingMachines? This might still indicate an issue, the operator should maybe look into.

I think that after a certain time, if the machine does not join the cluster, the MCM will put an error in it's status, and then this error will be visible. `PendingMachines` is only displayed if a new machine is in the process of being created and while waiting for it to join the cluster, but after the respective timeout it's removed again. /cc @hardikdr @prashanth26 

Anyways, I'll remove the two expressions from the PR for now

timebertt (approved):
/lgtm

pkg/apis/core/v1beta1/helper/errors.go
prashanth26: >I think that after a certain time, if the machine does not join the cluster, the MCM will put an error in it's status, and then this error will be visible. PendingMachines is only displayed if a new machine is in the process of being created and while waiting for it to join the cluster, but after the respective timeout it's removed again. /cc @hardikdr @prashanth26

Yes, the machineDeployment preserves any errors on machine objects even on being replaced by a new one. Until all the machineDeployments ready replicas equals the desired. So any such errors of a machine failing on creation/draining should be preserved even if machine objects are replaced.

PR #3085, opened 2020-11-02
Skip certain tasks in shoot deletion flow if namespace is terminating

rfranzke:

/kind bug
/priority normal

Certain tasks should be skipped when the namespace is already terminating (because no new resources can be created in such namespaces anyways).

On the way, I cleaned up some duplication and replaced usage of deprecated functions.

**Release note**:

```improvement operator
NONE
```


timebertt (approved):
Thanks for cleaning up!
/lgtm

PR #3091, opened 2020-11-02
Set DB sync to full mode

vlvasilev:

/area logging
/kind enhancement
/priority normal

With this PR we set the fluent-bit tail plugin DB sync to `full` mode. The default was `normal` where the offsets of the currently processing files is not immediately written to the DB. In result when fluent-bit pod is restarted the log processor start from the last written offset which is not up to date and causing repetitions of logs. When the log aggregator has logic to eliminate the log duplication(Elasticsearch) this is not an issue, but when it doesn't like Loki this cause many log duplication and `Entry Out OF Order` warnings. Setting DB sync to full prevents this of happening.

Also we increase the `chunk_target_size` to `1536000` bytes as is recommended by [Grafana](https://grafana.com/docs/loki/latest/best-practices/current-best-practices/#7-use-chunk_target_size)  

Fixes #

**Release note**:

```improvement operator
Fluent-bit tail plugin DB synchronization is set to FULL to avoid log duplication when fluent-bit pod is restarted.
```
```improvement operator
Loki chunk_target_size option is set to 1536000 bytes as recommended by Grafana 
```

### DIFF
--- a/charts/seed-bootstrap/charts/fluent-bit/templates/_fluent-bit-config.tpl
+++ b/charts/seed-bootstrap/charts/fluent-bit/templates/_fluent-bit-config.tpl
@@ -27,6 +27,7 @@
         Exclude_Path      *_garden_fluent-bit-*.log,*_garden_loki-*.log
         Parser            docker
         DB                /var/log/flb_kube.db
+        DB.sync           full
         read_from_head    true
         Skip_Long_Lines   On
         Mem_Buf_Limit     30MB
--- a/charts/seed-bootstrap/charts/loki/templates/loki-configmap.yaml
+++ b/charts/seed-bootstrap/charts/loki/templates/loki-configmap.yaml
@@ -9,6 +9,7 @@ data:
   loki.yaml: |-
     auth_enabled: {{ .Values.authEnabled }}
     ingester:
+      chunk_target_size: 1536000
       chunk_idle_period: 3m
       chunk_block_size: 262144
       chunk_retain_period: 3m


vpnachev (dismissed):
/lgtm

charts/seed-bootstrap/charts/fluent-bit/templates/_fluent-bit-config.tpl
```diff
@@ -27,6 +27,7 @@
         Exclude_Path      *_garden_fluent-bit-*.log,*_garden_loki-*.log
         Parser            docker
         DB                /var/log/flb_kube.db
+        DB.sync           full
```
vpnachev, line 30: > Note that full has a high I/O performance cost.

How will fluent bit perform with this change on a large seed cluster, for example with 100+ shoots?

JordanJordanov: The current integration test runs with 50 containers. A large seed cluster can contain 100 times more ( ~5000). 
So we will run additional (customized) tests to validate.

vlvasilev: On `dev` landscape `aws` seed with 67 shoot there is no difference.

rfranzke: @vlvasilev @JordanJordanov shall we then go ahead merging this PR or do you want to perform further tests?

JordanJordanov: We will perform more tests.
Actually we will enable 1st on dev, staging & CANARY via landscape.yaml and only then we will change the default value for all landscapes

rfranzke: I see. Shall we then close this PR for now and reopen again when it's relevant to not keep it open for a long time?

vpnachev: /close

@vlvasilev please re-open once you have validated this change from performance point of view.

vlvasilev:
After some period of monitoring the fluent-bit behavior on canary and live there were no any warning or errors. So I am reopening the PR

vpnachev (approved):
/lgtm

JordanJordanov (approved):

vpnachev (approved):
/lgtm

PR #3092, opened 2020-11-03
Add KEEPALIVE to apiserver-proxy

mvladev:

/area networking
/kind bug
/priority normal

Some upstream LoadBalancers have low idle connection timeouts and reset the connection. For AWS NLBs its 350sec (unconfigurable).

If a controller is watching the API server via the proxy, AWS NLB might simply silently terminate the connection if no data is send between the client and the API server.

n/a

**Release note**:

```improvement operator
`apiserver-proxy` now uses tcp keepalive every 55 seconds to prevent idle timeouts between it and the SNI LoadBalancer.
```


rfranzke (approved):
/lgtm

PR #1940, opened 2020-02-14
Adapt PSP RBAC naming and more constants

ialidzhikov:
**What this PR does / why we need it**:
Adapt PSP RBAC naming and more constants to the new API group.

**Which issue(s) this PR fixes**:
Ref gardener/gardener#1649

**Special notes for your reviewer**:
I think I will need 1 more PR for deprecated labels/annotations handling in the charts.

**Release note**:

```improvement operator
Two more annotations from to the old API group have now equivalents in the new API group as follows:

`shoot.garden.sapcloud.io/use-as-seed` -> `shoot.gardener.cloud/use-as-seed`
`shoot.garden.sapcloud.io/ignore-alerts` -> `shoot.gardener.cloud/ignore-alerts`
```

PR #3094, opened 2020-11-03
Clean up more deprecated labels and annotations

ialidzhikov:

/kind cleanup
/kind api-change
/priority normal

Ref https://github.com/gardener/gardener/issues/1649

Dependencies for this PR:
- [x] release of gardener/gardenectl that includes https://github.com/gardener/gardenctl/pull/412 - as prior to this PR gardenctl was using `project.garden.sapcloud.io/name` -> https://github.com/gardener/gardenctl/releases/tag/v0.23.1

**Release note**:

```action user
The project controller is now adapted to accept and maintain only the project namespace labels from the new API group - `gardener.cloud/role=project` and `project.gardener.cloud/name=<project-name>`. Until now the project controller was accepting and maintaining also the labels from the old API groups -  `garden.sapcloud.io/role=project` and `project.garden.sapcloud.io/name=<project-name>`. With this change, the project controller removes the namespace labels associated to the old API group. If you are still using these deprecated labels, you need to adapt your machinery.
```

```action user
The already deprecated annotations `shoot.garden.sapcloud.io/use-as-seed` and `shoot.garden.sapcloud.io/ignore-alerts` are no longer respected by the corresponding Gardener components. If you are still using these deprecated annotations, you need to adapt your machinery to use respectively `shoot.gardener.cloud/use-as-seed` and `shoot.gardener.cloud/ignore-alerts`. You need to be careful with use-as-seed annotation as if you don't adapt to the new annotation and update your Gardener version to the one that no longer respects the deprecated annotation, this will be handled as deletion of the ShootedSeed and will mark the Seed for deletion.
```


rfranzke:
Can you check the failing `verify`?

pkg/apis/core/v1alpha1/types_common.go
```diff
@@ -117,11 +117,6 @@ const (
 	// ExternalGardenerName is the value in a Kubernetes core resources `.metadata.finalizers[]` array on which the
 	// Gardener will react when performing a delete request on a resource.
 	ExternalGardenerName = "gardener.cloud/gardener"
-	// ExternalGardenerNameDeprecated is the value in a Kubernetes core resources `.metadata.finalizers[]` array on which the
-	// Gardener will react when performing a delete request on a resource.
-	//
-	// Deprecated: Use `ExternalGardenerName` instead.
-	ExternalGardenerNameDeprecated = "garden.sapcloud.io/gardener"
```
rfranzke, line 124: This finalizer might be part of some `Secret`s that exist since a long time - I don't see code that would handle/remove/replace them properly. WDYT?

ialidzhikov: There was a migration code included in https://github.com/gardener/gardener/pull/1940. I also see a small number of Secrets that have this finalizer - I believe these are cases where user created the secret with the deprecated finalizer or finalizer was not removed from the Secret for some reason. I will have a look later today..

ialidzhikov: I had a look and I can summarise in the following way:
 - I believe that most of the Secrets with that finalizer are created initially with the finalizer by the end user or the finalizer is re-added (for example with `kubectl apply`) again by the end users.
- There were several cases in which there was a Secret with this finalizer but no associated SecretBinding. I believe these Secrets are leftover from the issue resolved with https://github.com/gardener/gardener/pull/1927.
- There was a single case in which the namespace was gone but there was a Secret with that finalizer with the given namespace name. I recreated the ns to allow proper deletion for this Secret.

IMO we cannot do much about the cases above. As I wrote already above there was a migration login in #1940 (that was cleaned up in #2449) and I think I did the trick. However I cleaned up the finalizer from Secrets that match the 3 cases above. Let me know if you have further comments.

rfranzke: Alright, thanks for checking so thoroughly.

/lgtm

rfranzke (approved):

PR #3097, opened 2020-11-03
Clean up setting fake Shoot status in the Cluster

ialidzhikov:

/kind cleanup
/priority normal

I believe we can now cleanup this workaround introduced with https://github.com/gardener/gardener/pull/2217 to prevent the ShootNotFailed predicate in the extensions library from reacting false negatively. 

**Release note**:

```action developer
A temporary workaround during the Cluster resource sync to the Seed by setting a fake Shoot status to prevent ShootNotFailed predicate in the extensions library from reacting false negatively is now cleaned up. Before upgrading to this version of Gardener, make sure that all of the extensions in your environment that use the ShootNotFailed predicate vendor `github.com/gardener/gardener@v1.4.0` or above (that contains https://github.com/gardener/gardener/pull/2265).
```


rfranzke (approved):
/lgtm

PR #3098, opened 2020-11-03
Keep the Loki updates from the HVPA

Kristian-ZH:

/area logging
/kind bug
/priority normal

Keep the Loki updates from the HVPA 

Fixes #

@ggaurav10 

**Release note**:

```improvement operator
A bug has been fixed that prevented the Loki HVPA recommendations from not being reverted.
```

### DIFF
--- a/charts/seed-bootstrap/charts/loki/values.yaml
+++ b/charts/seed-bootstrap/charts/loki/values.yaml
@@ -32,22 +32,22 @@ livenessProbe:
   httpGet:
     path: /ready
     port: metrics
-  initialDelaySeconds: 45
+  initialDelaySeconds: 120
 
 readinessProbe:
   httpGet:
     path: /ready
     port: metrics
-  initialDelaySeconds: 45
+  initialDelaySeconds: 120
 
 replicas: 1
 
 resources:
   limits:
-    cpu: 100m
-    memory: 256Mi
+    cpu: 450m
+    memory: 512Mi
   requests:
-    cpu: 20m
+    cpu: 200m
     memory: 64Mi
 
 securityContext:
@@ -67,8 +67,8 @@ hvpa:
     cpu: 800m
     memory: 2G
   minAllowed:
-    cpu: 20m
-    memory: 200M
+    cpu: 200m
+    memory: 64M
   targetAverageUtilizationCpu: 80
   targetAverageUtilizationMemory: 80
   scaleUpStabilization:
@@ -91,8 +91,8 @@ hvpa:
         percentage: 80
   limitsRequestsGapScaleParams:
     cpu:
-      value: "200m"
+      value: "300m"
       percentage: 40
     memory:
-      value: "300M"
+      value: "400M"
       percentage: 40
--- a/pkg/operation/botanist/logging.go
+++ b/pkg/operation/botanist/logging.go
@@ -24,6 +24,7 @@ import (
 	"github.com/gardener/gardener/pkg/features"
 	gardenletfeatures "github.com/gardener/gardener/pkg/gardenlet/features"
 	"github.com/gardener/gardener/pkg/operation/common"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	autoscalingv1beta2 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -31,7 +32,6 @@ import (
 
 // DeploySeedLogging will install the Helm release "seed-bootstrap/charts/loki" in the Seed clusters.
 func (b *Botanist) DeploySeedLogging(ctx context.Context) error {
-
 	// TODO: remove in a future release
 	// Clean up the stale loki-vpa.
 	lokiVpa := &autoscalingv1beta2.VerticalPodAutoscaler{ObjectMeta: metav1.ObjectMeta{Name: "loki-vpa", Namespace: b.Shoot.SeedNamespace}}
@@ -50,6 +50,11 @@ func (b *Botanist) DeploySeedLogging(ctx context.Context) error {
 		return err
 	}
 
+	lokiValues := map[string]interface{}{
+		"global":   images,
+		"replicas": b.Shoot.GetReplicas(1),
+	}
+
 	hvpaValues := make(map[string]interface{})
 	hvpaEnabled := gardenletfeatures.FeatureGate.Enabled(features.HVPA)
 	if b.ShootedSeed != nil {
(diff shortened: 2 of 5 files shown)


pkg/operation/botanist/logging.go
```diff
@@ -56,13 +61,25 @@ func (b *Botanist) DeploySeedLogging(ctx context.Context) error {
 		hvpaEnabled = gardenletfeatures.FeatureGate.Enabled(features.HVPAForShootedSeed)
 	}
 
-	hvpaValues["enabled"] = hvpaEnabled
+	if hvpaEnabled {
+		foundLokiStatefulset := true
```
rfranzke:
```suggestion
		var foundLokiStatefulSet bool
```

pkg/operation/botanist/logging.go
```diff
@@ -32,6 +35,8 @@ import (
 // DeploySeedLogging will install the Helm release "seed-bootstrap/charts/loki" in the Seed clusters.
 func (b *Botanist) DeploySeedLogging(ctx context.Context) error {
 
+	lokiValues := make(map[string]interface{})
```
rfranzke:
```suggestion
	lokiValues := map[string]interface{}{
			"global":   images,
			"replicas": b.Shoot.GetReplicas(1),
	}
```

pkg/operation/botanist/logging.go
```diff
@@ -56,13 +61,25 @@ func (b *Botanist) DeploySeedLogging(ctx context.Context) error {
...
-		"hvpa":     hvpaValues,
+		if foundLokiStatefulset {
+			lokiValues["resources"] = lokiStatefulSet.Spec.Template.Spec.Containers[0].Resources.DeepCopy()
+		}
 	}
 
+	hvpaValues["enabled"] = hvpaEnabled
+
+	lokiValues["global"] = images
+	lokiValues["replicas"] = b.Shoot.GetReplicas(1)
```
rfranzke:
```suggestion
```

pkg/operation/seed/seed.go
```diff
@@ -326,23 +328,52 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+			lokiStatefulSet := &appsv1.StatefulSet{}
+			if err := k8sSeedClient.Client().Get(ctx, kutil.Key(v1beta1constants.GardenNamespace, "loki"), lokiStatefulSet); client.IgnoreNotFound(err) != nil {
+				return err
+			} else if apierrors.IsNotFound(err) {
+				foundLokiStatefulset = false
+			}
+
+			if foundLokiStatefulset && hvpaEnabled {
+				lokiValues["resources"] = lokiStatefulSet.Spec.Template.Spec.Containers[0].Resources.DeepCopy()
+			}
```
rfranzke: Code duplication. What about extraction this into a function and calling it on the two places that use it?

charts/seed-bootstrap/charts/loki/values.yaml
```diff
@@ -67,7 +67,7 @@ hvpa:
     cpu: 800m
     memory: 2G
   minAllowed:
-    cpu: 20m
+    cpu: 200m
```
ggaurav10, line 70: initial `request` is changed to `100m`. `minAllowed` is changed to `200m`, and `minChange` is min(`100m`, `80% of requests = 80m`).
VPA is immediately going to recommend `200m` because of `minAllowed`, and because it is matching `minChange`, loki is going to be immediately scaled to `200m`.
we should either increase `minChange` or set initial `request` to `200m`.

ggaurav10 (changes requested):

pkg/operation/common/utils.go
```diff
@@ -428,6 +428,19 @@ func DeleteLoggingStack(ctx context.Context, k8sClient client.Client, namespace
 	return nil
 }
 
+// GetLokiHvpaRecommendedResources  returns reccomended resources for Loki by the HVPA
+func GetLokiHvpaRecommendedResources(ctx context.Context, k8sClient client.Client) (*corev1.ResourceRequirements, error) {
+	lokiStatefulSet := &appsv1.StatefulSet{}
+	if err := k8sClient.Get(ctx, kutil.Key(v1beta1constants.GardenNamespace, "loki"), lokiStatefulSet); client.IgnoreNotFound(err) != nil {
+		return nil, err
+	} else if apierrors.IsNotFound(err) {
```
rfranzke: Shouldn't it be 
```suggestion
	} else if !apierrors.IsNotFound(err) {
```

pkg/operation/common/utils.go
```diff
@@ -428,6 +428,19 @@ func DeleteLoggingStack(ctx context.Context, k8sClient client.Client, namespace
 	return nil
 }
 
+// GetLokiHvpaRecommendedResources  returns reccomended resources for Loki by the HVPA
+func GetLokiHvpaRecommendedResources(ctx context.Context, k8sClient client.Client) (*corev1.ResourceRequirements, error) {
```
rfranzke: Can you implement a unit test for it? Also, maybe make it more general, i.e., change its signatur to
```golang
GetResourcesForContainerInStatefulSet(context.Context, client.Client, key client.ObjectKey, containerName string) (*corev1.ResourceRequirements, error)
```

Kristian-ZH: > Can you implement a unit test for it?

I am not sure how exactly can I mock the `k8sClient.Get(ctx, key, statefulSet)` function save in the given statefulSet, the wanted object

rfranzke: There are plenty of examples in other unit tests that you can take a look at and get inspiration from :)

Kristian-ZH: done

pkg/operation/common/utils_test.go
```diff
@@ -551,6 +553,107 @@ var _ = Describe("common", func() {
...
+					Name:      "contaienr-1",
+					Resources: *expectedResources,
+				},
+			}
+
+			c.EXPECT().Get(ctx, kutil.Key(testNamespace, testStatefulset), gomock.AssignableToTypeOf(&appsv1.StatefulSet{})).SetArg(2, *statefulSet).Return(nil)
+
+			rr, err := GetResourcesForContainerInStatefulSet(ctx, c, kutil.Key(testNamespace, testStatefulset))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(len(rr)).To(Equal(1))
```
rfranzke:
```suggestion
			Expect(rr).To(HaveLen(statefulSet.Spec.Template.Spec.Containers))
```

(same below)

pkg/operation/common/utils.go
```diff
@@ -428,6 +428,23 @@ func DeleteLoggingStack(ctx context.Context, k8sClient client.Client, namespace
 	return nil
 }
 
+// GetResourcesForContainerInStatefulSet  returns the current resources for container in the StatefulSet
+func GetResourcesForContainerInStatefulSet(ctx context.Context, k8sClient client.Client, key client.ObjectKey) ([]*corev1.ResourceRequirements, error) {
```
rfranzke: As the function name suggests, I rather meant that you provide the container name and return only the resources for this container. But actually it's better like this because you only need one `Get()` call even if you are interested in the resources of multiple containers in the same statefulset. I just suggest to rename it one more time:
```suggestion
func GetContainerResourcesInStatefulSet(ctx context.Context, k8sClient client.Client, key client.ObjectKey) ([]*corev1.ResourceRequirements, error) {
```

rfranzke (commented):
Looks good now, just a few nit suggestions!

rfranzke (dismissed):
/lgtm

charts/seed-bootstrap/charts/loki/values.yaml
```diff
@@ -32,22 +32,22 @@ livenessProbe:
...
 replicas: 1
 
 resources:
   limits:
-    cpu: 100m
+    cpu: 300m
     memory: 256Mi
   requests:
-    cpu: 20m
+    cpu: 200m
```
ggaurav10: difference between cpu requests and memory is 100m, and minChange is also 100m. If the pod goes into continuous restarts because of cpu throttling, VPA will not be able to recommend higher recommendations.
Should we increase `limits` to `450m` just to be on safe side?. Same for the memory as well.

ggaurav10 (changes requested):

charts/seed-bootstrap/charts/loki/values.yaml
Kristian-ZH: Did the change. Can you check please if it is better now?

ggaurav10 (approved):
lgtm

rfranzke (approved):

PR #3100, opened 2020-11-03
Change rolling strategy for vpn deployment

vpnachev:

/area ops-productivity
/kind enhancement
/priority normal

The default strategy is 
```yaml
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
```
and it can allow deletion of the old replica before the new one is created. This will result in network unavailability between the kube-apiserver and in-cluster services, so some mutating and validating webhooks can fail. It can become even worse if these webhooks prevent the creation of pods in the kube-system namespace. I know this change is not completely fixing this problem, but it is improving a bit the behavior. A better approach would be deploy the vpn with at least 2 replicas and PDB, but this will consume more resources in the end-user cluster, also I am not sure whether the vpn can really run with multiple replicas( @DockToFuture?).

Fixes #

**Release note**:

```improvement operator
The `vpn` deployment rolling strategy is improved so that the new pod is created before the old one is deleted.
```

### DIFF
--- a/charts/shoot-core/components/charts/vpn-shoot/templates/deployment.yaml
+++ b/charts/shoot-core/components/charts/vpn-shoot/templates/deployment.yaml
@@ -12,6 +12,11 @@ metadata:
 spec:
   revisionHistoryLimit: 0
   replicas: 1
+  strategy:
+    rollingUpdate:
+      maxSurge: 100%
+      maxUnavailable: 0%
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm

timuthy (approved):

PR #3106, opened 2020-11-04
Do not set Shoot status to Failed too early when operation cannot be initialized

### LINKED ISSUE #3018: Shoot's `.status.lastOperation` set to `Failed` too early

/area usability
/kind bug
/priority normal

**What happened**:
`gardener-apiserver` was put under load (and got scaled up very fast), however, until new apiserver pods were available, it sometimes returned

  ```text
  the server has received too many requests and has asked us to try again later (get controllerregistrations.core.gardener.cloud)
  ```

  The gardenlet sets the `.status.lastOperation.state` to `Failed` when trying to initialize a new operation without waiting for the configured retry period.

The `Failed` status is only set after the configured period in the gardenlet component config.

n/a yet, sorry

- Gardener version: v1.11.3
- Kubernetes version (use `kubectl version`): 1.17.8

ialidzhikov:

/area quality
/kind bug
/priority normal
Fixes #3018

**Release note**:

```improvement operator
gardenlet no longer sets the Shoots status to Failed too early when operation cannot be initialized
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control.go
@@ -307,17 +307,18 @@ func (c *Controller) deleteShoot(logger *logrus.Entry, shoot *gardencorev1beta1.
 		return reconcile.Result{}, nil
 	}
 
+	// Trigger regular shoot deletion flow.
+	c.recorder.Event(shoot, corev1.EventTypeNormal, gardencorev1beta1.EventDeleting, "Deleting Shoot cluster")
+	shootNamespace := shootpkg.ComputeTechnicalID(project.Name, shoot)
+	shoot, err = c.updateShootStatusOperationStart(ctx, gardenClient.GardenCore(), shoot, shootNamespace, operationType)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
 	o, operationErr := c.initializeOperation(ctx, logger, shoot, project, cloudProfile, seed)
 	if operationErr != nil {
 		_, updateErr := c.updateShootStatusOperationError(ctx, gardenClient.GardenCore(), shoot, fmt.Sprintf("Could not initialize a new operation for Shoot deletion: %s", operationErr.Error()), operationType, lastErrorsOperationInitializationFailure(shoot.Status.LastErrors, operationErr)...)
 		return reconcile.Result{}, utilerrors.WithSuppressed(operationErr, updateErr)
-	}
-
-	// Trigger regular shoot deletion flow.
-	c.recorder.Event(o.Shoot.Info, corev1.EventTypeNormal, gardencorev1beta1.EventDeleting, "Deleting Shoot cluster")
-	o.Shoot.Info, err = c.updateShootStatusOperationStart(ctx, gardenClient.GardenCore(), o.Shoot.Info, o.Shoot.SeedNamespace, operationType)
-	if err != nil {
-		return reconcile.Result{}, err
 	}
 
 	// At this point the deletion is allowed, hence, check if the seed is up-to-date, then sync the Cluster resource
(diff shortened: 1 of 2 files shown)


rfranzke (dismissed):
/lgtm
/invite @timuthy 
/needs second-opinion

timuthy (changes requested):
Hm, I don't see this https://github.com/gardener/gardener/issues/3018#issuecomment-721075772 / https://github.com/gardener/gardener/issues/3018#issuecomment-721094915 being handled in the PR, am I right? Let's suppose you create a new shoot with some arbitrary domain (`.spec.dns.domain: shoot.foo.bar`) but w/o a provider -> operation initialization will fail but Gardener will enqueue the shoot again because `HasNonRetryableErrorCode` is `false`.

ialidzhikov:
> Hm, I don't see this #3018 (comment) / #3018 (comment) being handled in the PR, am I right? Let's suppose you create a new shoot with some arbitrary domain (.spec.dns.domain: shoot.foo.bar) but w/o a provider -> operation initialization will fail but Gardener will enqueue the shoot again because HasNonRetryableErrorCode is false.

You are right, currently the initializeOperation errors caused by configuration problems are not handled with error code. I will have a look. Do you have in mind other cases (except wrongly configured `.spec.dns.domain`) that need such handling?

timuthy:
> You are right, currently the initializeOperation errors caused by configuration problems are not handled with error code. I will have a look. Do you have in mind other cases (except wrongly configured .spec.dns.domain) that need such handling?

Thanks @ialidzhikov. I think generally every case that leads to an error when initializing the operation. These are and will be most probably configuration problems from users.

ialidzhikov:
@timuthy , I checked now and it seems that we have in place the handling for incomplete dns config. Ref https://github.com/gardener/gardener/blob/v1.12.5/pkg/gardenlet/controller/shoot/shoot_control.go#L626-L638
Creation of such Shoot is flagged with ERR_CONFIGURATION_PROBLEM and set to Failed immediately:

```yaml
status:
  lastErrors:
  - codes:
    - ERR_CONFIGURATION_PROBLEM
    description: unable to figure out which secret should be used for dns
    taskID: initializeOperation
  lastOperation:
    description: 'Could not initialize a new operation for Shoot reconciliation: unable
      to figure out which secret should be used for dns'
    lastUpdateTime: "2020-11-06T13:36:03Z"
    progress: 0
    state: Failed
    type: Create
  observedGeneration: 1
```

> I think generally every case that leads to an error when initializing the operation.

This confused me a little bit. 😅   For example the error `cannot calculate required extensions for shoot my-shoot: the server has received too many requests and has asked us to try again later (get controllerregistrations.core.gardener.cloud)` can be also returned from the `initializeOperation` func and we rather want to retry on such error. My understanding is that we introduce errors like `IncompleteDNSConfigError` for known user configuration problems and add handling for them with error code. Otherwise on generic error we allow retry for `retryDuration`. That's why I also asked whether there are other cases that need such handling.

PS. Sorry, I wanted to add this comment to the PR, but wrongly added it to the issue previously.

timuthy:
I see, good point @ialidzhikov. Then we have to differentiate between retry-able errors and configuration errors here as well :(

ialidzhikov:
I think this is already in place. Do you see any error currently in `initializeOperation` func that needs to be a configuration error -> we won't retry such Shoot? It seems good to me.

timuthy:
IIUC, `updateShootStatusOperationStart` deletes the `initializeOperation` from the `lastErrors` [here](https://github.com/gardener/gardener/blob/c320f104698977e82ab22a63c968a66a99326ecd/pkg/gardenlet/controller/shoot/shoot_control.go#L539), so that the [error handling](https://github.com/gardener/gardener/blob/c320f104698977e82ab22a63c968a66a99326ecd/pkg/gardenlet/controller/shoot/shoot_control_migrate.go#L76) for `c.initializeOperation` won't have a non-retryable error. Can you double check?

ialidzhikov:
> IIUC, `updateShootStatusOperationStart` deletes the `initializeOperation` from the `lastErrors` [here](https://github.com/gardener/gardener/blob/c320f104698977e82ab22a63c968a66a99326ecd/pkg/gardenlet/controller/shoot/shoot_control.go#L539), so that the [error handling](https://github.com/gardener/gardener/blob/c320f104698977e82ab22a63c968a66a99326ecd/pkg/gardenlet/controller/shoot/shoot_control_migrate.go#L76) for `c.initializeOperation` won't have a non-retryable error. Can you double check?

In https://github.com/gardener/gardener/blob/559cd2126eac4e2e685550f21a2d3fd8f035e776/pkg/gardenlet/controller/shoot/shoot_control.go#L329 

hence in 

https://github.com/gardener/gardener/blob/559cd2126eac4e2e685550f21a2d3fd8f035e776/pkg/gardenlet/controller/shoot/shoot_control.go#L643-L655

we have the implicit check for non-retryable error and if we have such error we upsert it the the lastErrors. So in this case updateShootStatusOperationError is not called with empty lastErrors or lastErrors that do no contain the non-retryable error.

timuthy:
Thanks for the investigation @ialidzhikov. I agree, `lastErrorsOperationInitializationFailure` is the savior 👍

timuthy (approved):
/lgtm

PR #3114, opened 2020-11-05
Gardenlet does not use self-generated TLS cert during self-deployment

rfranzke:

/area security scalability
/kind bug
/priority normal

Without this PR, the gardenlet deploys further instances of itself with the same self-generated certificate. This is bad in two ways:
1. Security-wise, the same certificate is used multiple times.
1. Any re-deployment of the gardenlet leads to rollouts of all further instances because a new certificate was generated and needs to be populated.

/cc @prashanth26 

**Release note**:

```improvement operator
A bug has been fixed that caused the gardenlet to deploy further instances of itself with its own self-generated server certificate. It prevents undesired redeployments of these further instances.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/seed_registration_control.go
+++ b/pkg/gardenlet/controller/shoot/seed_registration_control.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


prashanth26 (approved):
/lgtm

timuthy (approved):
/lgtm
Nice!

timebertt (approved):
/lgtm

PR #3115, opened 2020-11-05
Ignore missing audit configmaps when deploying KAPI in shoot deletion

rfranzke:

/area usability open-source
/kind task
/priority normal

#2967 introduced an incompatible change because the shoot deletion flow is now trying to redeploy the kube-apiserver deployment. This fails when the end-user has already deleted the audit policy configmap. As Gardener did not protect this configmap with a finalizer earlier the configmap was removed from the system immediately. (This was improved with #3071, but still, Gardener v1.12 is introducing an incompatible change initially caused by #2967.
Let's ignore missing configmaps for the time being to give all end-users enough time to adapt their workflows.

**Release note**:

```improvement user
Missing audit policy `ConfigMap`s for `Shoot`s are now ignored when trying to redeploy the kube-apiserver in the shoot deletion flow.
```


timuthy (approved):
/lgtm

timebertt (approved):
Thanks for addressing my comment :)
/lgtm

PR #3117, opened 2020-11-05
Disable audit policy configmap reference protection by default

rfranzke:

/area open-source usability
/kind task
/priority normal

#3071 introduced reference protection for audit policy configmaps used by shoots. This protection is enabled by default and an incompatible behavioural change that breaks the workflows of some users.
Hence, let's disable it by default for now to allow everybody to adapt their workflows appropriately. We can revisit the topic and enable it by default again/remove the configuration option as whole after a couple release.

**Release note**:

```noteworthy operator
The audit policy config map reference protection controller [introduced with v1.12.0](https://github.com/gardener/gardener/pull/3071) is now disabled by default. You can explicitly enable it in the gardener-controller-manager's component configuration by setting `.controllers.shootReference.protectAuditPolicyConfigMaps=true`.
```


timuthy (approved):
/lgtm

timebertt (approved):
/lgtm

PR #3128, opened 2020-11-06
Add .mailmap aliases to identify authors uniquely

msohn:
See https://git-scm.com/docs/git-check-mailmap#_mapping_authors

/area open-source
/kind cleanup
/priority normal

Ensure authors are identified uniquely in git.

Fixes #

**Release note**:
```improvement operator
NONE
```

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,117 @@
+Adel Zaalouk <adel.zaalouk@sap.com>
+Adel Zaalouk <adel.zaalouk@sap.com> <adel.zalok.89@gmail.com>
+Amshuman K R <amshuman.rao.karaya@sap.com>
+Andreas Fritzler <andreas.fritzler@sap.com>
+Andreas Fritzler <andreas.fritzler@sap.com> <afritzler@users.noreply.github.com>
+Andreas Fritzler <andreas.fritzler@sap.com> <andreas.fritzler@gmail.com>
+Andreas Thaler <andreas.thaler01@sap.com>
+Avi Deitcher <avi@deitcher.net>
+Axel Christ <axel.christ@sap.com> <adracus@gmail.com>
+Aylei <rayingecho@gmail.com>
+Bekir Dogan <bdogan@ptc.com>
+Bekir Dogan <bdogan@ptc.com> <bekirdo@gmail.com>
+Bharath Suresh Kallur <bharath.suresh.kallur@sap.com>
+Bruno Klein <brunoklein99@gmail.com>
+Christian Berendt <berendt@23technologies.cloud>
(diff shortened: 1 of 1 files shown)


rfranzke:
What's the motivation for this, i.e., why do we want to maintain such file in our repo(s)? Besides the fact that it contains duplicates, it will come with some maintenance effort. Would we also put it into our 100+ other repos? Is there an automatism that would maintain it?

rfranzke:
ping @msohn

rfranzke:
@msohn, any statement on above questions?

msohn:
Run git shortlog -sne to see what this tries to fix. Apparently many authors use multiple identities (name, email address) in git commits they contribute to gardener. .mailmap is used by git shortlog [1] and git log --mailmap [2] to map multiple git identities of the same user to a single name and email address. You can also set log.mailmap=true in ~/.gitconfig to always use the mappings. I generated the draft of this .mailmap using [3].

I do not insist, so if you don't want to maintain such mapping simply close this PR.
If you are interested to maintain this I can try to generate this on all gardener repositories and create a merged .mailmap across all gardener repositories and we could check in the resulting .mailmap into all repositories.

[1] https://git-scm.com/docs/git-shortlog#_mapping_authors
[2] https://git-scm.com/docs/git-log#Documentation/git-log.txt---no-mailmap
[3] https://github.com/greenrd/genmailmap

rfranzke:
Thanks for the explanation. If we would take this serious then we would need automation that maintains these mailmap files across all repositories. Otherwise, the content might get outdated eventually. If we have such automation then I'm fine with it, but I'd refrain from having it if the content is just manually generated and maintained because this cannot scale. @gardener/gardener-maintainers

msohn:
I think this isn't fully automatable since some human needs to decide which identities belong to the same person which isn't always obvious

timebertt:
I would like to understand if this is just a cosmetic optimization for the case that someone runs `git log`.
Or is this used somewhere else (e.g. GitHub or some other processing)?

rfranzke:
@msohn can you clarify on @timebertt's question so that we can decide how to proceed, please?

vpnachev:
I do not like very much this mapping because it clearly exposes information that maybe not everybody would like to show publicly. Of course, the mails can be always retrieved from the commit metadata, except if the [private email](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address) feature is used. Maybe the CLA also does not allow private email?

rfranzke:
OK, let's
/close
the PR for now. It seems that the majority is against maintaining such a list.

PR #3131, opened 2020-11-09
Add managed ingress controller feature for Seeds

### LINKED ISSUE #201: Seed cluster ingresses needs domain records
When deploying shoot clusters, gardener deploys monitoring applications and creates ingresses for them:
```
$ gardener > kubectl get ing -n shoot-seed-msa-1
NAME           HOSTS                                           ADDRESS   PORTS     AGE
alertmanager   a.msa-1.seed.garden-1.az.dub.rd2.thingworx.io             80, 443   1h
grafana            g.msa-1.seed.garden-1.az.dub.rd2.thingworx.io             80, 443   1h
prometheus     p.msa-1.seed.garden-1.az.dub.rd2.thingworx.io             80, 443   1h
```

But these domains are not created, or there is no wildchar record is available.

I tried to use external-dns to create these records to no avail, because external-dns requires services, or ingresses to have external-dns.alpha.kubernetes.io/hostname= annotations, which are not available.

BeckerMax:
Adds nginx ingress controller to the Seed

/area networking usability
/kind enhancement
/priority normal
/topology seed

Adds the ability to include a current version of [nginx-ingress](https://github.com/kubernetes/ingress-nginx) to Seeds.
Configurable either in the Seed spec or via an annotation in Shooted seeds.
Migration for ShootedSeeds from nginx-addon to Seed Ingress Controller is supported.
Migration for ShootedSeeds back to nginx-addon is not supported and will result in a validation error. We want ShootedSeeds to use the new Ingress Controller instead of the nginx-addon.

Enabling managed nginx-ingress controller on the Seed will:
- Switch ingressClass of Ingresses for Seed and Shoot monitoring to the new IngressClass: `nginx-gardener`
- Apply charts of current version of [nginx-ingress](https://github.com/kubernetes/ingress-nginx)
- Add dns Extension to Seed if not there yet
- Add DNSProvider (constructed from the Shootmanifest in case of ShootedSeed) and DNSEntry

Fixes #201 

```feature operator
Gardener now offers to manage a dedicated ingress controller for seed clusters (earlier, this was a manual operator task when registering seeds). You can find more information about it [here](https://github.com/gardener/gardener/tree/master/docs/deployment/deploy_gardenlet_manually.md#kubernetes-cluster-that-should-be-registered-as-a-seed-cluster).
```

```breaking operator
If the `nginx-ingress` addon for a shoot used as seed is disabled then you can no longer enable it anymore. Instead, use the new managed ingress controller feature. You can find more information about it [here](https://github.com/gardener/gardener/tree/master/docs/deployment/deploy_gardenlet_manually.md#kubernetes-cluster-that-should-be-registered-as-a-seed-cluster). Existing shoots used as seeds with `.spec.addons.nginxIngress.enabled=true` will continue to work.
```


charts/seed-bootstrap/charts/nginx-ingress/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    app: nginx-ingress
+    chart: nginx-ingress-0.8.0
+    heritage: Tiller
+    release: addons
```
rfranzke: These labels can be removed from all resources inside this chart

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-service.yaml
```diff
@@ -0,0 +1,27 @@
...
+metadata:
+  labels:
+    app: nginx-ingress
+    component: "controller"
+    heritage: Tiller
+    release: addons
+  name: nginx-ingress-controller
+  namespace: garden
+  annotations:
+    dns.gardener.cloud/dnsnames: "{{ .Values.domain }}"
```
rfranzke: Remove this and create a `DNSEntry` instead

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-configmap.yaml
```diff
@@ -0,0 +1,14 @@
...
+  labels:
+    app: nginx-ingress
+    component: "controller"
+    heritage: Tiller
+    release: addons
+  name: nginx-ingress-controller
+  namespace: garden
+data:
+  server-name-hash-bucket-size: "256"
+  use-proxy-protocol: "false"
```
rfranzke: I guess this needs to be configurable somehow because this setting won't work on all infrastructures. Maybe we should introduce a provider config in the `Seed` resource where such things can be specified 👀

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-deployment.yaml
```diff
@@ -0,0 +1,115 @@
...
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
+          resources:
+            limits:
+              cpu: 1000m
+              memory: 3Gi
+            requests:
+              cpu: 100m
+              memory: 100Mi
```
rfranzke: Can we lower the limits and define a VPA for it?

docs/deployment/deploy_gardenlet_manually.md
```diff
@@ -26,40 +26,38 @@ Manually deploying a gardenlet is required in the following cases:
...
+```yaml
+spec:
+dns:
+	ingressDomain: dev.my-seed.example.com
+	nginxIngress:
+		enabled: true
+		provider: aws-route53
+		secretRef:
+			name: ingress-secret
+			namespace: garden
```
rfranzke: Can you use spaces?

example/50-seed.yaml
```diff
@@ -26,6 +26,12 @@ spec:
       namespace: garden
   dns:
     ingressDomain: dev.my-seed.example.com
+#   nginxIngress:
```
rfranzke: Let's rename this to `ingress` or `ingressController` so that we can switch internally if we want and do not hard-code the nginx-ingress.

docs/deployment/deploy_gardenlet_manually.md
```diff
@@ -26,40 +26,38 @@ Manually deploying a gardenlet is required in the following cases:
...
-  There must be a DNS record `*.ingress.<SEED-CLUSTER-DOMAIN>` where `<SEED-CLUSTER-DOMAIN>` is the value of the `.dns.ingressDomain` field of [a seed cluster resource](../../example/50-seed.yaml) (or configured using the `seedConfig{}` in the [gardenlet configuration](../../example/20-componentconfig-gardenlet.yaml)).
-  
-### Example `Nginx` Ingress Controller
+- Every Seed cluster needs an Ingress controller which distributes external requests to internal components like grafana and prometheus. Gardener supports to approaches to achieve this:
+    
+a. Gardener managed Ingress controller and DNS records. For this configure the following lines in your [Seed resource](../../example/50-seed.yaml):
+```yaml
+spec:
+dns:
+	ingressDomain: dev.my-seed.example.com
```
rfranzke:
```suggestion
	ingressDomain: ingress.dev.my-seed.example.com
```

pkg/apis/core/types_seed.go
```diff
@@ -108,7 +108,20 @@ type SeedBackup struct {
...
+	// If this is set spec.settings.nginxIngress can not be enabled and vise versa.
 	IngressDomain string
+	// NginxIngress controls settings for a gardener managed nginx-ingress controller
+	// +optional
+	NginxIngress *SeedSettingNginxIngress `json:"nginxIngress,omitempty" protobuf:"bytes,6,opt,name=nginxIngress"`
+}
+
+type SeedSettingNginxIngress struct {
+	// Enabled controls whether a gardener managed nginx-ingress is deployed. It is used to route external traffic to the Seed.
+	Enabled bool `json:"enabled" protobuf:"bytes,1,opt,name=enabled"`
```
rfranzke: Do we need this field? If `SeedSettingNginxIngress` is not `nil`, then it's enabled, right?

pkg/apis/core/types_seed.go
```diff
@@ -108,7 +108,20 @@ type SeedBackup struct {
 type SeedDNS struct {
 	// IngressDomain is the domain of the Seed cluster pointing to the ingress controller endpoint. It will be used
 	// to construct ingress URLs for system applications running in Shoot clusters.
+	// If this is set spec.settings.nginxIngress can not be enabled and vise versa.
 	IngressDomain string
+	// NginxIngress controls settings for a gardener managed nginx-ingress controller
+	// +optional
```
rfranzke: Remove all `// +optional` comments - they are not needed in the internal API version

pkg/apis/core/types_seed.go
```diff
@@ -108,7 +108,20 @@ type SeedBackup struct {
...
+	NginxIngress *SeedSettingNginxIngress `json:"nginxIngress,omitempty" protobuf:"bytes,6,opt,name=nginxIngress"`
```
rfranzke: Remove all `json` and `protobuf` tags - they are not needed in the internal API version.

pkg/apis/core/v1beta1/constants/types_constants.go
```diff
@@ -304,4 +304,7 @@ const (
 
 	// ClusterIdentity is a constant equal to the name and data key (that stores the identity) of the cluster-identity ConfigMap
 	ClusterIdentity = "cluster-identity"
+
+	// SeedNginxIngressClass defines the ingress class for the seed nginx ingress controller
+	SeedNginxIngressClass = "nginx-gardener"
```
rfranzke, line 321:
```suggestion
	SeedIngressClass = "nginx-gardener"
```

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -327,6 +328,12 @@ func parseShootedSeed(annotation string) (*ShootedSeed, error) {
 	}
 	shootedSeed.SeedProviderConfig = seedProviderConfig
 
+	nginxIngressSettings, err := parseNginxIngressSettings(settings)
```
rfranzke: Get rid of the nginx hard-codings

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go.orig
```diff
@@ -0,0 +1,599 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: Can you delete this file, please?

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -276,6 +279,23 @@ func computeKindTypesForShoots(
 	return wantedKindTypeCombinations
 }
 
+// computeKindTypesForSeed computes the list of wanted kind/type combinations for extension resources based on the
+// Seed configuration
+func computeKindTypesForSeed(
```
rfranzke, line 284: Please add a unit test for this function.

pkg/gardenlet/controller/shoot/seed_registration_control.go
```diff
@@ -339,6 +339,7 @@ func prepareSeedConfig(ctx context.Context, gardenClient client.Client, seedClie
 		},
 		DNS: gardencorev1beta1.SeedDNS{
 			IngressDomain: fmt.Sprintf("%s.%s", common.IngressPrefix, *(shoot.Spec.DNS.Domain)),
+			NginxIngress:  shootedSeedConfig.NginxIngressSettings,
```
rfranzke: Should we try to first take the settings from the shoot's `.spec.dns` section? Only if required fields are missing here (e.g., `.spec.dns.providers[] ` is empty`, then we need the additional information from the shooted seed configuration).

pkg/operation/botanist/monitoring.go
```diff
@@ -220,6 +220,7 @@ func (b *Botanist) DeploySeedMonitoring(ctx context.Context) error {
 		},
 		"prometheus":               prometheus,
 		"kube-state-metrics-shoot": kubeStateMetricsShoot,
+		"ingress-class":            getIngressClass(b.Seed.Info.Spec.DNS.NginxIngress),
```
rfranzke:
```suggestion
		"ingressClass":            getIngressClass(b.Seed.Info.Spec.DNS.NginxIngress),
```

pkg/controllermanager/controller/controllerregistration/seed_control.go
```diff
@@ -36,6 +36,7 @@ func (c *Controller) seedAdd(obj interface{}) {
 		return
 	}
 	c.seedQueue.Add(key)
+	c.controllerRegistrationSeedQueue.Add(key)
```
rfranzke: Should we only do that if the (nginx) ingress spec in the seed has changed?

pkg/operation/seed/seed.go
```diff
@@ -864,3 +894,90 @@ func deleteMutatingWebHookConfiguration(ctx context.Context, k8sSeedClient clien
...
+		if err := switchIngressClassToSeedNginx(ctx, kutil.Key(ns.Name, "grafana-users"), seedClient); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func switchIngressClassToSeedNginx(ctx context.Context, ingressKey types.NamespacedName, seedClient client.Client) error {
+	ingress := &extensionsv1beta1.Ingress{}
+	if err := seedClient.Get(ctx, ingressKey, ingress); err != nil {
```
rfranzke, line 1156: If we receive a "not found" error here then we shouldn't do anything. Some shoots might not have these ingresses, and that's perfectly fine for us.

pkg/operation/seed/seed.go
```diff
@@ -864,3 +894,90 @@ func deleteMutatingWebHookConfiguration(ctx context.Context, k8sSeedClient clien
...
+		return err
+	}
+	if _, err := controllerutil.CreateOrUpdate(ctx, seedClient, ingress, func() error {
```
rfranzke: Can you use `seedClient.Update` here?

pkg/operation/seed/seed.go
```diff
@@ -864,3 +894,90 @@ func deleteMutatingWebHookConfiguration(ctx context.Context, k8sSeedClient clien
...
+}
+
+func createDNSProvider(providerNS string, providerName string, dnsSpec gardencorev1beta1.SeedDNS, seedClient client.Client) error {
+	dnsProvider := &v1alpha1.DNSProvider{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: providerNS,
+			Name:      providerName,
+		},
+	}
+	if _, err := controllerutil.CreateOrUpdate(context.TODO(), seedClient, dnsProvider, func() error {
```
```suggestion
	if _, err := controllerutil.CreateOrUpdate(ctx, seedClient, dnsProvider, func() error {
```

rfranzke (changes requested):

charts/seed-bootstrap/charts/nginx-ingress/templates/serviceaccount.yaml
BeckerMax: Can I remove those three labels from all the charts? They seem to be obsolete.

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-service.yaml
BeckerMax: If I create a dns entry myself I need to specify a target ip. This ip is dynamically created with the LB service so I can't know it when the charts are applied. An annotation is the recommended way to create dns records for services according to: https://github.com/gardener/external-dns-management#automatic-creation-of-dns-entries-for-services-and-ingresses

charts/seed-bootstrap/charts/nginx-ingress/templates/serviceaccount.yaml
rfranzke: Generally yes, but the charts will be replaced as part of #2754 anyways, so no need to invest into it now.

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-service.yaml
rfranzke: In the shoot controller we create a load balancer service, wait for the IP to be populated into the status, and then create a `DNSEntry` with it. It is somewhat confusing if we follow a different approach now in the seed controller. I'm just trying to strive for consistency.

BeckerMax:
Current state after discussion with @rfranzke:
1. new proposed API:
```
spec:
  dns: // dnsProvider config. Similiar to how it is done in Shoot
    ingressDomain: foo.bar.com
    provider: aws-route53
    secretName: foobar
    ... (see '.spec.dns' in shoot resource)
  ingressController: // generic Ingress Controller configuration
    managed: true // flag to deactivate it but not lose the config
    name: nginx // currently only nginx available
    providerConfig: // Infrastructure specific setting
      some_nginx_setting: 4711
```
2. Shooted Seed passes the ProviderConfig, `name` and `managed` via annotations.
3. For ShootedSeeds `NginxIngress` is activated if NginxAddon is deactivated in the Seed, name and managed is set via annotation.

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-service.yaml
BeckerMax: Should we change it in the shoot controller then? It seems less work to do it with an annotation.

rfranzke: We cannot because we create a `DNSEntry` for the external record whose target is not a load balancer IP of `Service`

BeckerMax:
@rfranzke I made slight changes to the API we discussed, see: https://github.com/gardener/gardener/pull/3131/commits/17b467732d5958c9d168e2e0cf476276c5f26c7c
For the dnsProvider I did not follow the dnsProvider in the Shoot spec. I think it is a different use case in a Seed. We do neither have multiple domains nor do we have multiple providers and also do not plan to support that. Also I removed the `managed` or `enabled` flag from the IngressController as I think turning the controller off while keeping the configuration is not a strong enough argument to add a field into the API.
I think APIs should be minimal.
WDYT about it?

pkg/apis/core/v1alpha1/types_seed.go
```diff
@@ -62,23 +62,26 @@ type SeedSpec struct {
 	BlockCIDRs []string `json:"blockCIDRs,omitempty" protobuf:"bytes,2,rep,name=blockCIDRs"`
 	// DNS contains DNS-relevant information about this seed cluster.
 	DNS SeedDNS `json:"dns" protobuf:"bytes,3,opt,name=dns"`
+	// IngressController configures a Gardener managed Ingress Controller listening on the ingressDomain
+	// +optional
+	IngressController *IngressController `json:"ingressController,omitempty" protobuf:"bytes,4,opt,name=ingressController"`
```
rfranzke: Move it at the end of this struct so that the protobuf order stays the same.

pkg/apis/core/v1alpha1/types_seed.go
```diff
@@ -124,18 +127,26 @@ type SeedDNS struct {
...
-	// Provider describes the type of the dns-provider, for example `aws-route53`, see https://github.com/gardener/external-dns-management
-	Provider string `json:"provider" protobuf:"bytes,3,opt,name=provider"`
+// SeedDNSProvider configures a DNSProvider for Seeds
+type SeedDNSProvider struct {
+	// Type describes the type of the dns-provider, for example `aws-route53`
+	Type string `json:"type" protobuf:"bytes,1,opt,name=type"`
 	// SecretRef is a reference to a Secret object containing cloud provider credentials used for registering external domains.
-	SecretRef corev1.SecretReference `json:"secretRef" protobuf:"bytes,4,opt,name=secretRef"`
+	SecretRef corev1.SecretReference `json:"secretRef" protobuf:"bytes,2,opt,name=secretRef"`
+}
```
rfranzke: > For the dnsProvider I did not follow the dnsProvider in the Shoot spec. I think it is a different use case in a Seed. We do neither have multiple domains nor do we have multiple providers and also do not plan to support that. 

There is no concrete use-case at this point in time, yes. However, this section here is mostly used to configure the `dnsv1alpha1.DNSProvider` object. Hence, let's please add the `domains` and `zones` fields like in the shoot DNS section, see also https://github.com/gardener/gardener/blob/master/charts/seed-dns/provider/templates/dnsprovider.yaml#L34-L53.
It doesn't "hurt" and it may be desired by people who have multiple DNS domains and hosted zones in one of their accounts but only want to activate one (or a subset) of them for Gardener.

pkg/apis/core/v1alpha1/types_seed.go
```diff
@@ -124,18 +127,26 @@ type SeedDNS struct {
...
+
+// IngressController configures a Gardener managed Ingress Controller listening on the ingressDomain
+type IngressController struct {
+	// Kind defines which kind of IngressController to use, for example `nginx`
+	Kind string `json:"kind" protobuf:"bytes,1,opt,name=kind"`
+	// ProviderConfig specifies infrastructure specific configuration for the ingressController
+	// +optional
+	ProviderConfig *runtime.RawExtension `json:"providerConfig,omitempty" protobuf:"bytes,2,opt,name=providerConfig"`
```
rfranzke, line 166: We should add a `domain` field here now that we have a dedicated section for the ingress controller in our API. This allows us to deprecate `.spec.dns.ingressDomain` and remove it in the next API version.

rfranzke (commented):
>  Also I removed the managed or enabled flag from the IngressController as I think turning the controller off while keeping the configuration is not a strong enough argument to add a field into the API.
I think APIs should be minimal.

Sure, fine for me. Thanks.

pkg/apis/core/v1alpha1/types_seed.go
BeckerMax: As discussed this would make IngressController required so we did not go with it.

rfranzke: Hm no, we can still use the `.spec.dns.ingressDomain` if `ingressController` is `nil`. If `ingressController.domain` is set then the `.spec.dns.ingressDomain` must not be set.

BeckerMax: Anyways in the long run ingressController would become required then. Because we would deprecate `spec.dns.ingressdomain`. But the IngressController is optional.

rfranzke: I cannot follow, what are you suggesting?

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-service.yaml
BeckerMax: done in https://github.com/gardener/gardener/pull/3131/commits/62040659fa259ee0aa470b71e897eb4e994768e1

pkg/controllermanager/controller/controllerregistration/seed_control.go
BeckerMax: I addressed this in https://github.com/gardener/gardener/pull/3131/commits/16c8862d6d2fcb2bb40daf3e3b0681065318f8b1

Unsure if we want to keep it though.

pkg/apis/core/v1alpha1/types_seed.go
BeckerMax: Right now `.spec.dns.IngressDomain` is a required field.
You suggest to
A. Add a new field `ingressDomain` under `IngressController`
B. Deprecate `dns.ingressDomain` and eventually remove it.
But if we follow through on B then `IngressController` will hold a required field and therefore needs to be required as well.
I don't think we want IngressController to become required.
Therefore, I suggest to leave `ingressDomain` under `spec.dns` and to not move it to `IngressController`

rfranzke: > But if we follow through on B then IngressController will hold a required field and therefore needs to be required as well.

Hm, no, why? It's just that one of `.spec.dns.ingressDomain` OR `.spec.ingressController.domain` must be specified. This is a backwards-compatible change because
* `.spec.dns.ingressDomain` was required earlier and is now optional
* if `.spec.dns.ingressDomain` is not provided now then another field must be provided, but as `.spec.dns.ingressDomain` was required before this is okay.

BeckerMax: We agreed upon that moving ingressDomain under ingressController would lead to ingressController being required in a new API version. In this case we could rename it to ingress. This could look like:
```
spec:
  ingress: // Renamed from previously IngressController
    domain: // In favor of spec.dns.ingressDomain
    controller:
      kind: // like 'nginx-ingress'
      provider-config:    
```

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-configmap.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app: nginx-ingress
+    component: "controller"
+  name: nginx-ingress-controller
+  namespace: garden
+data:
+{{ .Values.providerConfig | indent 2 }}
```
rfranzke:
```suggestion
{{- if .Values.providerConfig }}
{{ .Values.providerConfig | indent 2 }}
{{- end }}
```

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-deployment.yaml
```diff
@@ -0,0 +1,113 @@
...
+                        - nginx-ingress
+                    - key: component
+                      operator: In
+                      values:
+                        - controller
+                    - key: release
+                      operator: In
+                      values:
+                        - addons
+      priorityClassName: system-cluster-critical
```
rfranzke: Hm, really? Should we remove that? The only consequence would be that logging/monitoring is not available, but "cluster critical" is definitely wrong. You can assign a dedicated priority > 0 instead.

docs/deployment/deploy_gardenlet_manually.md
```diff
@@ -26,40 +26,40 @@ Manually deploying a gardenlet is required in the following cases:
...
-    ingress resources for seed components like grafana and prometheus).
-      
-  - Handle ingress resources that are deployed 
-    during the `Seed` bootstrap by the gardenlet.
-      
-  >:warning:<br>  
-  There must be a DNS record `*.ingress.<SEED-CLUSTER-DOMAIN>` where `<SEED-CLUSTER-DOMAIN>` is the value of the `.dns.ingressDomain` field of [a seed cluster resource](../../example/50-seed.yaml) (or configured using the `seedConfig{}` in the [gardenlet configuration](../../example/20-componentconfig-gardenlet.yaml)).
-  
-### Example `Nginx` Ingress Controller
+- Every Seed cluster needs an Ingress controller which distributes external requests to internal components like grafana and prometheus. Gardener supports to approaches to achieve this:
```
rfranzke:
```suggestion
- Every Seed cluster needs an Ingress controller which distributes external requests to internal components like grafana and prometheus. Gardener supports two approaches to achieve this:
```

pkg/apis/core/types_seed.go
```diff
@@ -109,11 +111,33 @@ type SeedBackup struct {
...
+	// Zones contains information about which hosted zones shall be included/excluded for this provider.
+	Zones *DNSIncludeExclude
+}
+
+// IngressController configures a Gardener managed Ingress Controller listening on the ingressDomain
+type IngressController struct {
+	// Kind defines which kind of IngressController to use, for example `nginx`
+	Kind string
+	// ProviderConfig specifies infrastructure specific configuration for the ingressController
+	ProviderConfig *runtime.RawExtension
```
rfranzke, line 150: Domain field?

example/50-seed.yaml
```diff
@@ -25,7 +25,16 @@ spec:
...
+    ingressDomain: ingress.dev.my-seed.example.com
+#   provider:
+#     type: aws-route53
+#     secretRef:
+#       name: ingress-secret
+#       namespace: garden
+# ingressController:
+#   kind: nginx
+#   providerConfig:
+#     <some-provider-specific-config-for-the-ingressController>
```
rfranzke: domain field

pkg/apis/core/v1alpha1/types_seed.go
```diff
@@ -124,6 +127,30 @@ type SeedDNS struct {
...
 	// to construct ingress URLs for system applications running in Shoot clusters.
 	IngressDomain string `json:"ingressDomain" protobuf:"bytes,1,opt,name=ingressDomain"`
+	// Provider configures a DNSProvider
+	// +optional
+	Provider *SeedDNSProvider `json:"provider,omitempty" protobuf:"bytes,2,opt,name=provider"`
+}
+
+// SeedDNSProvider configures a DNSProvider
+type SeedDNSProvider struct {
+	// Domains contains information about which domains shall be included/excluded for this provider.
```
rfranzke: optional tags are missing here

pkg/apis/core/v1alpha1/types_seed.go
```diff
@@ -124,6 +127,30 @@ type SeedDNS struct {
...
+	Domains *DNSIncludeExclude `json:"domains,omitempty" protobuf:"bytes,1,opt,name=domains"`
+	// Type describes the type of the dns-provider, for example `aws-route53`
+	Type string `json:"type" protobuf:"bytes,2,opt,name=type"`
+	// SecretRef is a reference to a Secret object containing cloud provider credentials used for registering external domains.
+	SecretRef corev1.SecretReference `json:"secretRef" protobuf:"bytes,3,opt,name=secretRef"`
+	// Zones contains information about which hosted zones shall be included/excluded for this provider.
+	Zones *DNSIncludeExclude `json:"zones,omitempty" protobuf:"bytes,4,opt,name=zones"`
```
rfranzke, line 148: optional tags

pkg/apis/core/v1alpha1/types_seed.go
rfranzke: ping

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -408,6 +415,26 @@ func parseShootedSeedShootDefaults(settings map[string]string) (*gardencorev1bet
 	return shootNetworks, nil
 }
 
+func parseIngressSettings(settings map[string]string) (*gardencorev1beta1.IngressController, error) {
```
rfranzke: Can you add a unit test for this, please?

pkg/apis/core/v1beta1/types_seed.go
```diff
@@ -127,6 +130,30 @@ type SeedDNS struct {
...
 	// to construct ingress URLs for system applications running in Shoot clusters.
 	IngressDomain string `json:"ingressDomain" protobuf:"bytes,1,opt,name=ingressDomain"`
+	// Provider configures a DNSProvider
+	// +optional
+	Provider *SeedDNSProvider `json:"provider,omitempty" protobuf:"bytes,2,opt,name=provider"`
+}
+
+// SeedDNSProvider configures a DNSProvider for Seeds
+type SeedDNSProvider struct {
+	// Domains contains information about which domains shall be included/excluded for this provider.
```
rfranzke: optional tags

pkg/apis/core/v1beta1/types_seed.go
```diff
@@ -127,6 +130,30 @@ type SeedDNS struct {
...
+	Domains *DNSIncludeExclude `json:"domains,omitempty" protobuf:"bytes,1,opt,name=domains"`
+	// Type describes the type of the dns-provider, for example `aws-route53`
+	Type string `json:"type" protobuf:"bytes,2,opt,name=type"`
+	// SecretRef is a reference to a Secret object containing cloud provider credentials used for registering external domains.
+	SecretRef corev1.SecretReference `json:"secretRef" protobuf:"bytes,3,opt,name=secretRef"`
+	// Zones contains information about which hosted zones shall be included/excluded for this provider.
```
rfranzke, line 149: optional tags

pkg/apis/core/v1beta1/types_seed.go
```diff
@@ -127,6 +130,30 @@ type SeedDNS struct {
...
+	Zones *DNSIncludeExclude `json:"zones,omitempty" protobuf:"bytes,4,opt,name=zones"`
+}
+
+// IngressController configures a Gardener managed Ingress Controller listening on the ingressDomain
+type IngressController struct {
+	// Kind defines which kind of IngressController to use, for example `nginx`
+	Kind string `json:"kind" protobuf:"bytes,1,opt,name=kind"`
+	// ProviderConfig specifies infrastructure specific configuration for the ingressController
+	// +optional
+	ProviderConfig *runtime.RawExtension `json:"providerConfig,omitempty" protobuf:"bytes,2,opt,name=providerConfig"`
```
rfranzke, line 169: domain field

pkg/apis/core/validation/seed.go
```diff
@@ -137,6 +137,19 @@ func ValidateSeedSpec(seedSpec *core.SeedSpec, fldPath *field.Path) field.ErrorL
 		allErrs = append(allErrs, apivalidation.ValidateAnnotations(seedSpec.Settings.LoadBalancerServices.Annotations, fldPath.Child("settings", "loadBalancerServices", "annotations"))...)
 	}
 
+	if seedSpec.IngressController != nil {
+		if seedSpec.IngressController.Kind != "nginx" {
```
rfranzke: maybe we can have a constant for `nginx` in the API?

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -276,6 +279,23 @@ func computeKindTypesForShoots(
...
+	seed *gardencorev1beta1.Seed,
+) sets.String {
+	var (
+		wantedKindTypeCombinations = sets.NewString()
+		ingressController          = seed.Spec.IngressController
+	)
+
+	if ingressController != nil && ingressController.Kind == "nginx" {
```
rfranzke: use the to-be-introduced constant

pkg/controllermanager/controller/controllerregistration/seed_control.go
```diff
@@ -36,10 +37,29 @@ func (c *Controller) seedAdd(obj interface{}) {
...
+
+	if apiequality.Semantic.DeepEqual(oldObject.Spec.IngressController, newObject.Spec.IngressController) {
+		key, err := cache.MetaNamespaceKeyFunc(newObj)
+		if err != nil {
+			return
+		}
+		c.seedQueue.Add(key)
+	} else {
+		c.seedAdd(newObj)
+	}
```
rfranzke: maybe you can add a comment here why we have these two cases

pkg/operation/operation.go
```diff
@@ -435,6 +435,12 @@ func (o *Operation) ShootVersion() string {
 	return o.Shoot.Info.Spec.Kubernetes.Version
 }
 
+// NginxIngressAddonEnabled returns true if the nginx-ingress addon is enabled in the Shoot manifest and the seed-nginx ingress is disabled for shooted Seeds
+func (o *Operation) NginxIngressAddonEnabled() bool {
+	shootNginxEnabled := o.Shoot.Info.Spec.Addons != nil && o.Shoot.Info.Spec.Addons.NginxIngress != nil && o.Shoot.Info.Spec.Addons.NginxIngress.Enabled
+	return shootNginxEnabled
+}
```
rfranzke: Why was this moved from the `shootpkg` to this package?

pkg/operation/seed/seed.go
```diff
@@ -940,3 +1015,99 @@ func determineClusterIdentity(ctx context.Context, c client.Client) (string, err
...
+func getIngressClass(seedIngressEnabled bool) string {
+	if seedIngressEnabled {
+		return v1beta1constants.SeedIngressClass
+	}
+	return "nginx"
+}
+
+func migrateIngressClassForShootIngresses(ctx context.Context, seedClient client.Client) error {
+	shootNamespaces := &corev1.NamespaceList{}
+	if err := seedClient.List(ctx, shootNamespaces, client.MatchingLabels{"gardener.cloud/role": "shoot"}); err != nil {
```
rfranzke: Can you use the constants here instead of "hard-coding" the strings?

pkg/operation/seed/seed.go
```diff
@@ -940,3 +1015,99 @@ func determineClusterIdentity(ctx context.Context, c client.Client) (string, err
...
+		}
+		return err
+	}
+
+	ingress.Annotations["kubernetes.io/ingress.class"] = v1beta1constants.SeedIngressClass
+	if err := seedClient.Update(ctx, ingress); err != nil {
+		return err
+	}
+
+	return nil
```
```suggestion
	return seedClient.Update(ctx, ingress)
```

pkg/operation/seed/seed.go
```diff
@@ -821,11 +858,49 @@ func bootstrapComponents(c kubernetes.Interface, namespace string, imageVector i
...
 
 	return components, nil
 }
 
+func managedIngress(seed *Seed) bool {
+	if seed.Info.Spec.IngressController != nil && seed.Info.Spec.IngressController.Kind == "nginx" {
+		return true
+	}
+	return false
+}
```
```suggestion
	return seed.Info.Spec.IngressController != nil
}
```

pkg/operation/seed/seed.go
```diff
@@ -685,9 +694,32 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
 		}
 	}
 
+	if managedIngress(seed) {
+		seedClient := k8sSeedClient.Client()
+
+		// TODO(beckermax): Remove after next release - temporary migration logic.
```
rfranzke: Hm, why? We need this always, right? We would even need the inverted migration when somebody disabled the managed ingress controller again.

pkg/operation/seed/seed.go
```diff
@@ -685,9 +694,32 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+		if err := migrateIngressClassForShootIngresses(ctx, seedClient); err != nil {
+			return err
+		}
+
+		dnsProvider := seed.Info.Spec.DNS.Provider
+		if dnsProvider == nil {
+			return fmt.Errorf("managed Ingress is requested but dns Provider is missing configuration")
+		}
```
rfranzke: No strong opinion here, but this is/should already been validated in the API validation.

pkg/gardenlet/controller/shoot/seed_registration_control.go
```diff
@@ -366,6 +373,7 @@ func prepareSeedConfig(ctx context.Context, gardenClient client.Client, seedClie
 			BlockCIDRs:    shootedSeedConfig.BlockCIDRs,
 			ShootDefaults: shootedSeedConfig.ShootDefaults,
 		},
+		IngressController: shootedSeedConfig.IngressSettings,
```
rfranzke: This can be computed by us: If the shoot's nginx-ingress addon is disabled, then we enable this here. If the shoot's nginx-addon is enabled, then we disable this here.
Then you can overwrite these computations with what was provided the `shootedSeedConfig`.

rfranzke (dismissed):

pkg/operation/seed/seed.go
BeckerMax: That is true :(. And yes we need the inverse as well. Feels wrong to attempt to set the ingressClass for all the shoot Ingresses on every Seed reconcile. Most of the time it will be a waste. But I have no better idea.

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-deployment.yaml
BeckerMax: We could assign `gardener-shoot-controlplane` as the priorityClass, WDYT?
`// PriorityClassNameShootControlPlane is the name of a priority class for critical pods of a shoot control plane.`

charts/seed-bootstrap/charts/nginx-ingress/templates/default-backend-deployment.yaml
```diff
@@ -0,0 +1,45 @@
...
+    app: nginx-ingress
+  name: nginx-ingress-k8s-backend
+  namespace: garden
+spec:
+  revisionHistoryLimit: 0
+  replicas: 1
+  selector:
+    matchLabels:
+      app: nginx-ingress
+      release: addons
```
timuthy, line 14:
```suggestion
      release: addons
      component: "nginx-ingress-k8s-backend"
```

charts/seed-bootstrap/charts/nginx-ingress/values.yaml
```diff
@@ -0,0 +1,14 @@
+## nginx configuration
+## Ref: https://github.com/kubernetes/ingress/blob/master/controllers/nginx/configuration.md
+##
```
timuthy: Can you move this down to `config`?

charts/seed-monitoring/charts/alertmanager/templates/allow-alertmanager-network-policy.yaml
```diff
@@ -21,5 +21,12 @@ spec:
       namespaceSelector:
```
timuthy: Can this be removed when seed ingress is enabled?

timuthy:
/assign
I'm halfway through but still want to test and will come back tomorrow.

charts/seed-monitoring/charts/grafana/templates/grafana-network-policy.yaml
```diff
@@ -31,6 +31,13 @@ spec:
         matchLabels:
```
timuthy, line 30: Can this be removed when seed ingress is enabled?

pkg/apis/core/validation/seed.go
```diff
@@ -137,6 +143,48 @@ func ValidateSeedSpec(seedSpec *core.SeedSpec, fldPath *field.Path) field.ErrorL
 		allErrs = append(allErrs, apivalidation.ValidateAnnotations(seedSpec.Settings.LoadBalancerServices.Annotations, fldPath.Child("settings", "loadBalancerServices", "annotations"))...)
 	}
 
+	if seedSpec.DNS.IngressDomain != nil {
+		allErrs = append(allErrs, validateDNS1123Subdomain(*seedSpec.DNS.IngressDomain, fldPath.Child("dns", "ingressDomain"))...)
+	}
+
+	if seedSpec.DNS.IngressDomain == nil && seedSpec.Ingress == nil {
+		allErrs = append(allErrs, field.Invalid(fldPath, seedSpec, "Either specify spec.ingress or spec.dns.ingressDomain"))
```
timuthy:
```suggestion
	if seedSpec.DNS.IngressDomain == nil && (seedSpec.Ingress == nil || len(seedSpec.Ingress.Domain) == 0) {
		allErrs = append(allErrs, field.Invalid(fldPath, seedSpec, "either specify spec.ingress or spec.dns.ingressDomain"))
```

pkg/apis/core/validation/seed.go
```diff
@@ -137,6 +143,48 @@ func ValidateSeedSpec(seedSpec *core.SeedSpec, fldPath *field.Path) field.ErrorL
...
+	if seedSpec.Ingress != nil {
+		if seedSpec.DNS.IngressDomain != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("dns", "ingressDomain"), "",
+				"Either specify spec.ingress.domain or spec.dns.ingressDomain"),
+			)
+		} else {
+			if !availableIngressKinds.Has(seedSpec.Ingress.Controller.Kind) {
+				allErrs = append(allErrs, field.NotSupported(
+					fldPath.Child("ingress", "controller", "kind"),
+					"Unsupported kind selected as managed ingress",
```
timuthy:
```suggestion
					seedSpec.Ingress.Controller.Kind,
```

pkg/apis/core/validation/seed_test.go
```diff
@@ -450,6 +450,101 @@ var _ = Describe("Seed Validation Tests", func() {
...
+
+				errorList := ValidateSeed(seed)
+
+				Expect(errorList).To(ConsistOfFields(Fields{
+					"Type":  Equal(field.ErrorTypeNotSupported),
+					"Field": Equal("spec.ingress.controller.kind"),
+				}))
+			})
+
+			It("should fail if the ingress domain is invalid", func() {
```
timuthy, line 520: Can you add another test for `seed.Spec.Ingress.Domain = ""`?

pkg/controllermanager/controller/controllerregistration/seed_control.go
```diff
@@ -25,21 +25,36 @@ import (
...
 	"github.com/gardener/gardener/pkg/logger"
 
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/tools/cache"
 )
 
-func (c *Controller) seedAdd(obj interface{}) {
+func (c *Controller) seedAdd(obj interface{}, addToControllerRegistrationQueue bool) {
```
timuthy, line 34: Can you elaborate why we need any controller registration handling if we already have dedicated `defaultControllerRegistrationSeedControl` controller watching seed events ([ref](https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go))?

example/50-seed.yaml
```diff
@@ -25,7 +25,29 @@ spec:
...
+#       - my-other-custom-domain.com
+#       exclude:
+#       - yet-another-custom-domain.com
+#     zones:
+#       include:
+#       - zone-id-1
+#       exclude:
+#       - zone-id-2
+# ingress: # If set then `.spec.dns.ingressDomain` must be nil and `.spec.dns.provider` must be set.
+#   domain: ingress.dev.my-seed.example.com
```
timuthy, line 46: Do you think the API benefits from this new `domain` field? I tend to say that this makes the code and usability slightly confusing since now both fields are part of the API and we won't/can't remove them in the near future. Additionally, `ingressDomain` logically still fits under `dns`, so FMPOV no need to introduce this new field. WDYT?

timuthy (dismissed):

timebertt:
/rebase

Sorry, I won't make it to a review this week.
Please, don't wait for me, @timuthy and @rfranzke already had thorough looks 👀

example/50-seed.yaml
timuthy: As discussed, I'm resolving this item and we stick to both fields because `ingressDomain` is only configurable if a non-managed Ingress controller is used, and `ingress.domain` if the manage Ingress controller is used i.e., mutual exclusion.

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-deployment.yaml
```diff
@@ -0,0 +1,113 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: nginx-ingress
+    component: "controller"
+  name: nginx-ingress-controller
+  namespace: garden
+spec:
+  replicas: 3
```
danielfoehrKn, line 10: I am sure you thought about why you pick exactly 3 replicas and vpa (not hvpa)
What's the reason for that?

Is it to be able to safely roll the nginx deployment?

charts/seed-bootstrap/charts/nginx-ingress/templates/vpa.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: "autoscaling.k8s.io/v1beta2"
+kind: VerticalPodAutoscaler
```
danielfoehrKn, line 2: I am curious, what is the reason for using VPA and not HVPA?
To prevent accidental scaledowns of the replicas that can hurt the Seed?

pkg/apis/core/types_seed.go
```diff
@@ -109,11 +111,43 @@ type SeedBackup struct {
...
+	SecretRef corev1.SecretReference
+	// Domains contains information about which domains shall be included/excluded for this provider.
+	Domains *DNSIncludeExclude
+	// Zones contains information about which hosted zones shall be included/excluded for this provider.
+	Zones *DNSIncludeExclude
+}
+
+// Ingress configures the Ingress specific settings of the Seed cluster
+type Ingress struct {
+	// Domain specifies the IngressDomain of the Seed cluster pointing to the ingress controller endpoint. It will be used
```
danielfoehrKn:
```suggestion
	// Domain specifies the Ingress domain of the Seed cluster pointing to the ingress controller endpoint. It will be used
```
The word IngressDomain is from the SeedDNS section - maybe adjust the wording.

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-service.yaml
```diff
@@ -0,0 +1,23 @@
...
+      targetPort: 80
+    - name: https
+      port: 443
+      protocol: TCP
+      targetPort: 443
+  selector:
+    app: nginx-ingress
+    component: "controller"
+    release: addons
+  type: "LoadBalancer"
```
danielfoehrKn, line 23: Seeing this, we ll end up with two load balancers instead of only one when enabling the managed istio.
To me, it seems more elegant to just deploy the ingress resources in the Seed cluster to be used with the existing istio deployment.

Deploying a separate nginx only seems useful when managed istio is not enabled yet (though its in the process of graduation already).

Maybe I am missing something here - thanks for your opinion.

pkg/operation/seed/seed.go
```diff
@@ -761,6 +778,24 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+	// managed nginx-ingress handling
+	if err := handleIngressDNSEntry(ctx, k8sSeedClient, chartApplier, seed); err != nil {
+		return err
+	}
+
+	var ingressClass string
+	if managedIngress(seed) {
+		ingressClass = v1beta1constants.SeedNginxIngressClass
+	} else {
+		ingressClass = v1beta1constants.ShootNginxIngressClass
```
danielfoehrKn, line 793: I guess this is a fair assumption, as the grafana,.. ingresses were always deployed with nginx ingress class. Odd though, because I think we never officially required having nginx as the Seed's  ingress controller.

pkg/operation/seed/seed.go
```diff
@@ -761,6 +778,24 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+		if err := deleteIngressController(ctx, k8sSeedClient.Client()); err != nil {
+			return err
+		}
+	}
+	if err := migrateIngressClassForShootIngresses(ctx, k8sGardenClient.Client(), k8sSeedClient.Client(), seed, ingressClass); err != nil {
```
danielfoehrKn, line 798: I think we should think about using ingress classes instead of the deprecated ingress class annotation. See: https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class

pkg/operation/seed/seed.go
```diff
@@ -966,3 +1017,196 @@ func determineClusterIdentity(ctx context.Context, c client.Client) (string, err
...
+}
+
+func copySecretFromGardenerToSeed(ctx context.Context, gardenClient, seedClient client.Client, secretKey types.NamespacedName, targetSecret *corev1.Secret) error {
+	gardenSecret := &corev1.Secret{}
+	if err := gardenClient.Get(ctx, secretKey, gardenSecret); err != nil {
+		return err
+	}
+
+	if gardenSecret == nil {
+		return errors.New("error during seed bootstrap DNS provider setup: empty secret")
```
danielfoehrKn: Can we reformulate this to state that the secret ref in Seed    - SeedDNS.Provider.SecretRef does not reference a secret with valid dns provider credentials?

pkg/gardenlet/controller/shoot/seed_registration_control.go
```diff
@@ -391,12 +421,27 @@ func prepareSeedConfig(ctx context.Context, gardenClient client.Client, seedClie
...
+	}
+
+	if err := seedClient.Get(ctx, kutil.Key(shoot.Status.TechnicalID, common.ShootDNSIngressName), &dnsv1alpha1.DNSEntry{}); err != nil {
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}
+
+	return false, nil
```
danielfoehrKn, line 441: Can you help me to understand why this check is necessary? 
Also, how can there be a DNS Entry of the name "ingress" if the Shoot ingress addon is not activated?

danielfoehrKn (commented):
Very nice PR!

Is there a reason why we do not take advantage of an already existing istio installation of the Seed cluster (feature gate is soon to be graduated) ? 
Instead of deploying nginx we could simply reuse istio
 - use the same istio ingress gateway (only one LB) or deploy multiple Istio INgress gateways if desired
 - adjust the ingress resources deployed for grafana,...  (istio can understand K8s stock ingress resources)

Apologize for not having raised that earlier.

EDIT: has been answered in the comments below.

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-service.yaml
rfranzke: Good point, but out of scope of this PR. With this feature it will be possible to extend further managed ingress controller. For now, we only have nginx, but in the future we might want to add an istio ingress feature.

pkg/operation/seed/seed.go
rfranzke: Yes, we have, see https://github.com/gardener/gardener/blob/master/docs/deployment/deploy_gardenlet_manually.md#example-nginx-ingress-controller (the document should have been more clear that it has to be nginx because this is the hard-coded ingress class in gardener).

pkg/operation/seed/seed.go
rfranzke: Good point, but out of scope of this PR. This can be handled separately (you can open an issue if you want) but is not related to the managed ingress feature that @BeckerMax contributes here.

pkg/operation/seed/seed.go
danielfoehrKn: ok thanks

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-service.yaml
danielfoehrKn: Ok, sounds good to me

danielfoehrKn (changes requested):
When testing the PR (except shooted seed), there is one thing I noticed.

- Turning on the managed ingress for the Seed does create the ingress controller and the DNS entry for the Seed cluster, however does not migrate the ingress domain of the Shoot clusters (only the ingress class).  
This only happens during the reconciliation of the Shoot when deploying the monitoring stack.
I think we should document that behavior stating that Seed owners must keep <del> their own ingress controller running </del> the Seed's previous domain (spec.ingressDomain) alive until every Shoot has been reconciled once (after one day).

- Same thing happens when removing the managed ingress again and enabling the spec.ingressDomain. However, this time it will for sure cause broken monitoring ingresses. 
The ingresses still point to the DNS domain of the managed ingress that has already been destroyed.
They should be updated to point to the  Seed's spec.ingressDomain.

pkg/gardenlet/controller/shoot/seed_registration_control.go
BeckerMax: The check ensures that the migration goes smoothly from the `Shoot nginx-ingress addon` to the new `Seed ingress controller` (relevant only for ShootedSeeds).

We need the check as the ShootedSeed and the Shoots can be reconciled at different points in time (Seed -> immediately, Shoots -> maintenance window). In the check we make sure that we introduce the Ingress Controller in the Seed only **after** the shoots got reconciled. This ensures that the old dnsEntry got cleaned up already.

BeckerMax:
@danielfoehrKn Thanks for your review and the questions.

**Edit:** I understand now that you are talking about changing the ingressDomain in addition to activating the ingressController. I agree with you that this would potentially lead to unavailability of the Shoot monitoring until the Shoots get reconciled with the new ingressDomain. AFAIK this would also happen if you currently change the ingressDomain of a Seed. So I would argue that this is not related to this change. Would you agree with me here?
We can document what you suggested above but do we even want to support changing the ingressDomain in general? 
By the way: Keeping the old ingressDomain would not help the cause here as the ingressClass of the monitoring ingresses would already be changed to `nginx-gardener` and therefore would not work with the old custom ingressController anymore, or? Bottom line, I would say that domain migration is a non supported use case currently and would come with downtime.

**Original answer:**
Regarding case 1 (switch from custom ingress Controller to managed Seed Ingress):
Due to migration logic in the Seed bootstrap function the ingressClass for monitoring Ingresses in the shoot ns should get migrated immediately during Seed reconcilation, see: https://github.com/gardener/gardener/pull/3131/files#diff-8495b57b25d39744fe1b20b1e1ce695d130f730142af86ff9c5929afcdbabeb4R1125
So, i would not expect a downtime for the Shoot monitoring and no need for the custom ingress controllers to stay running. I would assume we rather want the Seed owner to disable the dns record pointing to their ingress controller so it does not interfere with our dnsentry using the same domain. I will test it again though to be sure the migration works as expected.

Case 2 (switch from managed Seed ingress to custom ingress Controller):
The ingressController and the DNS Entry will get deleted. The ingressClass will be migrated to `nginx` again directly (during the Seed reconcilation with the logic linked above). Then the shoot monitoring will stay unavailable until the Seed owner deploys a new IngressController managing the ingressClass `nginx`. Can you elaborate what you would expect instead?

pkg/controllermanager/controller/controllerregistration/seed_control.go
BeckerMax: Background:
When a dnsProvider is specified in the Seed resource we want the dns extension to be installed in the Seed (if not present yet). In order to detect whether a dnsProvider is needed we need to watch the Seed resource (spec.dns.provider).
In the code if a Seed resource is [updated](https://github.com/gardener/gardener/blob/52a23a2d77ca8dda65fa592752b91becfda8d79e/pkg/controllermanager/controller/controllerregistration/seed_control.go#L46) we check whether a the dnsProvider changed. If yes we add the Seed to the [controller-registration-seed-queue](https://github.com/gardener/gardener/blob/52a23a2d77ca8dda65fa592752b91becfda8d79e/pkg/controllermanager/controller/controllerregistration/seed_control.go#L42). In this queue it will check if a dns-extension is needed [here](https://github.com/gardener/gardener/pull/3131/files#diff-0d8996ff3546653f546482d70a198c7c4da27e0102f684a87d5e982df34896d4R163). This is also the Controller you mentioning above I assume.

We had a discussion whether we can implement it similar to how we watch backupBuckets [here](https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/controllerregistration/controllerregistration.go#L114) and watch for when a dnsProvider gets added. But that won't work as the backupBucket and the dnsProvider are different in nature. The backupbucket resource watched is gardencorev1beta1.BackupBucket which holds a reference to the Seed. There is no equivalent resource for the dnsProvider. Therefore, we can not watch the dnsProvider resource and need to watch the Seed resource as it is implemented right now. So if you have no further objections I would leave the implementation as it is right now.

charts/seed-bootstrap/charts/nginx-ingress/templates/controller-deployment.yaml
BeckerMax: Yes, 3 replicas are there to guarantee availability when the deployment gets rolled.

charts/seed-bootstrap/charts/nginx-ingress/templates/vpa.yaml
BeckerMax: In gardener we are using the hvpa in different ways.
Option 1) Disabling the hpa component and just use the vpa part of the hvpa as done for [Loki](https://github.com/gardener/gardener/blob/69cc831f7c7fdc91405e16fe792a2f5b433f23c6/charts/seed-bootstrap/charts/loki/templates/hvpa.yaml#L17).
-> I don't think we are going to see heavy resource spikes for the nginx-ingress-controller (which was meant to be better solved with the hvpa vpa in comparison to normal vpa). Especially as we have 3 replicas the normal vpa should be able to handle the scaling.
Option 2) Also use the hpa component of the hvpa.
-> I do not see us benefiting from hpa here. 3 replicas will be fine and we probably do not need more. We also do not want less to prevent downtimes. So just vanilla 3 replicas without hpa seems to be good enough.

Conclusion: Normal vpa with 3 replicas seems sufficient and as it is sufficient I prefer the solution with less complexity.

danielfoehrKn: thanks for the explanation!

pkg/gardenlet/controller/shoot/seed_registration_control.go
danielfoehrKn: So you detect if the Shoot has already been reconciled after removal of the Shoot nginx-ingress addon by checking if the ingress DNS entry is already removed?

Thanks, that makes sense!

Maybe you could add a small comment there.

BeckerMax: I added a comment

charts/seed-monitoring/charts/alertmanager/templates/allow-alertmanager-network-policy.yaml
BeckerMax: ~I now activate only the relevant network policy~

BeckerMax:
Thanks for the reviews. @timuthy @danielfoehrKn I addressed your comments and this PR would be ready to merge from my side. Can you have another look?
The problem raised by @danielfoehrKn that migrating the ingressDomain is not supported: This limitation exists independently of the Ingress PR. As there are multiple potential solutions for it I would suggest opening another issue and discussing it there so that it does not block this PR.

danielfoehrKn:
@BeckerMax thanks, I agree with your assessment.  I can open an issue for it. 

/lgtm

danielfoehrKn (dismissed):

charts/seed-monitoring/charts/alertmanager/templates/allow-alertmanager-network-policy.yaml
BeckerMax: To support migration cases we would need to change this in the code where we also migrate the [ingressClass](https://github.com/gardener/gardener/pull/3131/files#diff-8495b57b25d39744fe1b20b1e1ce695d130f730142af86ff9c5929afcdbabeb4R798). This seems like complicating the code without a real security benefit. We (@rfranzke) think this could be done in a separate refactoring of network policies where the netpol allowing traffic to/from prometheus should be created in the component (like nginx-ingress) instead of all network policies being defined in [prometheus](https://github.com/gardener/gardener/blob/master/charts/seed-monitoring/charts/core/charts/prometheus/templates/allow-prometheus-network-policy.yaml).

rfranzke (approved):
/lgtm

BeckerMax:
Related issue https://github.com/gardener/gardener/issues/3281
cc @danielfoehrKn

PR #3132, opened 2020-11-09
Allow project members to read ResourceQuota

timuthy:

/area usability
/kind enhancement
/priority normal

This PR enhances the `gardener.cloud:system:project-member-aggregation` and `gardener.cloud:system:project-viewer-aggregation` ClusterRoles by allowing read operations for ResourceQuotas.

**Release note**:

```improvement user
Project admins and viewers are now allowed to read `corev1.ResourceQuota` objects.
```

### DIFF
--- a/charts/gardener/controlplane/charts/application/templates/rbac-user.yaml
+++ b/charts/gardener/controlplane/charts/application/templates/rbac-user.yaml
@@ -176,6 +176,7 @@ rules:
   - ""
   resources:
   - events
+  - resourcequotas
   verbs:
   - get
   - list
@@ -260,6 +261,7 @@ rules:
   resources:
   - events
   - configmaps
+  - resourcequotas
   - serviceaccounts
   verbs:
   - get


charts/gardener/controlplane/charts/application/templates/rbac-user.yaml
```diff
@@ -180,6 +180,13 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - resourcequotas
+  verbs:
+  - get
+  - list
```
rfranzke:
```suggestion
  verbs:
  - get
  - list
  - watch
```

rfranzke (changes requested):

rfranzke:
> Project viewers are not permitted to read quotas, because I rather consider it an administrative task.

Why not? 👀 I don't get the difference.

timuthy:
> Why not? 👀 I don't get the difference.

Principle of least privilege. As a viewer you cannot influence the quota in any way by creating/deleting resources, i.e. I don't see a use case for that audience at the moment, do you?

rfranzke:
Hm, I cannot follow, sorry. As the name suggests, isn't the viewer role exactly used to "view" resources in the project namespace? The only exception are `Secret`s for obvious reasons. Viewers cannot influence shoots in any way either, still, they are allowed to see them. What's the difference with `ResourceQuota`s?

Despite from this theoretical point of view, I also see use-cases where end-users want to automate their own alerting or monitoring of such resource quotas. Why should they be forced to use a service account with `admin` role for that?

timuthy:
> Despite from this theoretical point of view, I also see use-cases where end-users want to automate their own alerting or monitoring of such resource quotas. Why should they be forced to use a service account with admin role for that?

Makes sense, thanks. I'll add it.

danielfoehrKn (approved):

rfranzke (approved):
/lgtm

PR #3099, opened 2020-11-03
Refactor etcd-druid deployment

rfranzke:
**How to categorize this PR?**

/area dev-productivity quality robustness
/kind technical-debt test
/priority normal
/topology seed
/platform all
/exp intermediate

**What this PR does / why we need it**:
This PR refactors the deployment for the etcd-druid and adopts the component concept. Consequently, it eliminates the Helm chart and adds full unit tests, instead. This improves several aspects and the overall code quality.

**Which issue(s) this PR fixes**:
Part of #2754 

/cc @shreyas-s-rao @amshuman-kr 

**Release note**:

```improvement operator
NONE
```

PR #3133, opened 2020-11-09
Properly de-bootstrap seed cluster on deletion (only for components)

### LINKED ISSUE #3056: Seed deletion stuck due to cluster-autoscaler managedresource

/area control-plane
/kind bug
/priority normal

**What happened**:
I tried to delete a seed, performing the following operations:
1. delete `seed` resource
2. delete resources created by `gardenlet` helm chart
3. delete `garden` namespace on seed cluster

Step 3 fails due to resources with finalizers existing in the namespace:
```
NAME                                                         NAMESPACE  AGE
secret/managedresource-cluster-autoscaler                    garden     83m
managedresource.resources.gardener.cloud/cluster-autoscaler  garden     83m
```

Step 1 should already delete the `ManagedResource` and `secret` resources so that step 3 doesn't fail.

- deploy `gardenlet` into cluster
- try to 'undo' that by performing the steps listed above

According to @rfranzke the reason is that the Gardener resource manager is removed in step 3 and thus can't handle the finalizers on the resources. The solution would be to already remove the resources when the `seed` resource is deleted.

**Environment**:

- Gardener version: `v1.12.0`
- Kubernetes version (use `kubectl version`): base cluster `1.18.9`, seed clusters `1.19.3`
- Cloud provider or hardware configuration: seen for azure and gcp seed, did not happen on aws (probably a race condition)

rfranzke:

/area usability
/kind bug
/priority normal

This PR implements proper methods to de-bootstrap a seed cluster.

Fixes #3056

The same will be adapted in #3099 

**Release note**:

```improvement operator
A bug has been fixed that resulted in a stuck `Seed` deletion due to orphaned `ManagedResource`s in the `garden` namespace.
```


danielfoehrKn (approved):

timuthy (approved):
/lgtm

PR #3136, opened 2020-11-09
Projects can be excluded from the stale checks

### LINKED ISSUE #2721: Project namespace annotation to prevent stale projects from being deleted

/area usability
/kind enhancement
/priority normal
/topology garden
/exp beginner
/platform all

We could introduce a new annotation on the `Namespace` that is backing a `Project` (e.g., `project.garden.sapcloud.io/skip-stale-check=true`) in order to prevent the stale project controller (ref #2446) from considering this `Project` for auto-deletion due to inactivity.
The annotation would be only considered if added to the `Namespace` to prevent end-users from adding it, i.e., it could be only added by operators/administrators.

**Why is this needed**:
Sometimes an operator might want to prevent `Project`s from being auto-deleted/garbage collected when they are not used for some time.

rfranzke:

/kind enhancement
/priority normal

Projects whose namespaces are annotated with `project.gardener.cloud/skip-stale-check=true` are now excluded from the stale checks.

Fixes #2721

**Release note**:

```noteworthy operator
It is now possible exclude specific `Project`s from the stale checks by annotating their related `Namespace`s with `project.gardener.cloud/skip-stale-check=true`.
```

### DIFF
--- a/pkg/operation/common/types.go
+++ b/pkg/operation/common/types.go
@@ -254,6 +254,11 @@ const (
 	// Deprecated: Use `ProjectName` instead.
 	ProjectNameDeprecated = "project.garden.sapcloud.io/name"
 
+	// ProjectSkipStaleCheck is the key of an annotation on a project namespace that marks the associated Project to be
+	// skipped by the stale project controller. If the project has already configured stale timestamps in its status
+	// then they will be reset.
+	ProjectSkipStaleCheck = "project.gardener.cloud/skip-stale-check"
+
 	// NamespaceProject is they key of an annotation on namespace whose value holds the project uid.
 	NamespaceProject = "namespace.gardener.cloud/project"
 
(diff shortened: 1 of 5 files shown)


pkg/operation/common/types.go
```diff
@@ -254,6 +254,11 @@ const (
 	// Deprecated: Use `ProjectName` instead.
 	ProjectNameDeprecated = "project.garden.sapcloud.io/name"
 
+	// ProjectSkipStaleCheck is the key of an annotation on a project namespace that makes the stale project controller
+	// excluding the project. If the project has already configured stale timestamps in its status then they will be
```
danielfoehrKn:
```suggestion
	// ProjectSkipStaleCheck is the key of an annotation on a project namespace that marks the associated Project to be skipped by the stale project controller.
	//  If the project has already configured stale timestamps in its status then they will be
```

timuthy (approved):
/lgtm

PR #3137, opened 2020-11-09
Merge duplicate project members

### LINKED ISSUE #2849: Validation should prevent duplicate ServiceAccounts in Project member list
Currently, it is possile to add the same service account twice in the member list:

```
  - apiGroup: rbac.authorization.k8s.io
    kind: User
    name: system:serviceaccount:garden-lukas:robot
    role: admin
  - kind: ServiceAccount
    name: robot
    namespace: garden-lukas
    role: uam
```

Gardener allows to configure the member like this and it seems to merge the roles.

Prevent to add the same service account multiple times.

**Why is this needed**:
The Dashboard would also need to merge the accounts as we do not want to present the same service account multiple times. Now, if the user wants to modify the (merged) service account, we do not know which of the entries should be edited. It would be easier and cleaner, if it would not be possible to add the same service account in the first place.

rfranzke:

/area usability 
/kind enhancement
/priority normal

With this PR we prevent duplicates in the `Project`'s `.spec.members[]` list. In order to be backwards compatible we are now merging such duplicates in before writing them to etcd.

Fixes #2849

After a couple of releases this merging code can be removed again.

**Release note**:

```action user
We are preparing a change that will lead to validation errors when the `Project` resource contains duplicates in the `.spec.members[]` list. For the time being, duplicates in this list are merged into a single member automatically by the Gardener API Server. In the future, this will no longer happen, instead, a validation error will be returned if a user sends a `Project` resource with duplicate members. Please adapt your API usage to not send any of such resources.
```

### DIFF
--- a/pkg/apis/core/validation/project.go
+++ b/pkg/apis/core/validation/project.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/validation/path"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/authentication/serviceaccount"
 )
 
 // ValidateProject validates a Project object.
@@ -72,8 +73,23 @@ func ValidateProjectSpec(projectSpec *core.ProjectSpec, fldPath *field.Path) fie
 	}
 	ownerFound := false
 
+	members := make(map[string]struct{}, len(projectSpec.Members))
+
 	for i, member := range projectSpec.Members {
 		idxPath := fldPath.Child("members").Index(i)
+
+		apiGroup, kind, namespace, name, err := ProjectMemberProperties(member)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(idxPath.Child("name"), member.Name, err.Error()))
+			continue
+		}
+		id := ProjectMemberId(apiGroup, kind, namespace, name)
+
+		if _, ok := members[id]; ok {
+			allErrs = append(allErrs, field.Duplicate(idxPath, member))
+		} else {
+			members[id] = struct{}{}
+		}
 
 		allErrs = append(allErrs, ValidateProjectMember(member, idxPath)...)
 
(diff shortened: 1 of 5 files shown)


pkg/apis/core/validation/project.go
```diff
@@ -243,3 +259,41 @@ func ValidateProjectStatusUpdate(newProject, oldProject *core.Project) field.Err
...
+	return apiGroup, kind, namespace, name, nil
+}
+
+// ProjectMemberId returns an internal ID for the project member.
+func ProjectMemberId(apiGroup, kind, namespace, name string) string {
+	return fmt.Sprintf("%s_%s_%s_%s", apiGroup, kind, namespace, name)
+}
+
+// ProjectMemberFromId returns the member properties from the id.
+func ProjectMemberFromId(id string) (string, string, string, string, error) {
```
ialidzhikov: This func is not used (not called)?

pkg/apis/core/validation/project.go
```diff
@@ -72,9 +73,24 @@ func ValidateProjectSpec(projectSpec *core.ProjectSpec, fldPath *field.Path) fie
...
 
+		apiGroup, kind, namespace, name, err := ProjectMemberProperties(member)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(idxPath.Child("name"), member.Name, err.Error()))
+			continue
+		}
+		id := ProjectMemberId(apiGroup, kind, namespace, name)
+
+		if _, ok := members[id]; ok {
+			allErrs = append(allErrs, field.Duplicate(idxPath, member))
```
ialidzhikov, line 89: Is this validation logic considered? I tried to update/create projects with duplicate members as described in #2849 but the duplicate members are merged and I don't receive any error about duplicated entries?

rfranzke: Yes, it is considered, but the gardener-apiserver auto-merges your duplicates before calling the validation. This is to ensure backwards-compatibility. In a future version, we remove this auto-merge code again, and then duplicates won't be accepted anymore.

pkg/apis/core/validation/project.go
rfranzke: Thanks, I've removed it.

pkg/registry/core/project/strategy.go
```diff
@@ -39,21 +40,78 @@ func (projectStrategy) NamespaceScoped() bool {
...
+					Name:      name,
+				},
+			})
+			oldMembersToRoles[id] = make(map[string]struct{})
+		}
+
+		for _, role := range member.Roles {
+			if _, ok := oldMembersToRoles[id][role]; !ok {
+				memberToNewRoles[id] = append(memberToNewRoles[id], role)
+			}
```
timebertt, line 98: Can you simplify the computation of the roles for the new members?
I think, it would be easier to read, if you would drop these lines and drop `memberToNewRoles` and rather loop over the keys of `oldMembersToRoles` for each member at the end.

timebertt (commented):
Looks good already.
Only one minor suggestion to improve readability.

pkg/apis/core/validation/project.go
ialidzhikov: Ah, yep, I got it.

pkg/registry/core/project/strategy.go
rfranzke: As discussed in person, this could potentially lead to a different order of the roles which breaks the API.

timebertt (approved):
/lgtm

ialidzhikov (approved):
/lgtm

PR #3141, opened 2020-11-09
Allow configuring kube-apiserver's max inflight requests for shoots

### LINKED ISSUE #3108: Expose max-requests-inflight parameter of kube-apiserver in Shoot

/area usability scalability
/kind enhancement
/priority normal

We should expose the `--max-requests-inflight` parameter on the `kube-apiserver` deployment in the `Shoot` resource so that end-users can change it in case their clusters' workloads require it to be higher/lower.

**Why is this needed**:
Mainly for scalability reasons.

rfranzke:

/area usability scalability
/kind enhancement
/priority normal

It is now possible to configure the `kube-apiserver`'s `--max-requests-inflight` and `--max-mutating-requests-inflight` flags via the `Shoot` spec.

Fixes #3108

* /cc @mvladev 
* The flags are present in all supported Kubernetes versions.

**Release note**:

```noteworthy user
It is now possible to configure the `kube-apiserver`'s `--max-requests-inflight` and `--max-mutating-requests-inflight` flags by setting the `.spec.kubernetes.kubeAPIServer.requests.max{Non}MutatingInflight` fields (default: `{400}`/`200`) in the `Shoot` specification.
```


pkg/apis/core/validation/shoot.go
```diff
@@ -571,7 +571,27 @@ func validateKubernetes(kubernetes core.Kubernetes, fldPath *field.Path) field.E
 			}
 		}
 
-		allErrs = append(allErrs, ValidateWatchCacheSizes(kubeAPIServer.WatchCacheSizes, fldPath.Child("watchCacheSizes"))...)
+		allErrs = append(allErrs, ValidateWatchCacheSizes(kubeAPIServer.WatchCacheSizes, fldPath.Child("kubeAPIServer", "watchCacheSizes"))...)
+
+		const maxMaxRequestsInflight = 800
```
mvladev: Don't set a limit - the API server can handle way above that number with better etcd tuning.

rfranzke: What about the concerns from @timuthy in https://github.com/gardener/gardener/issues/3108#issuecomment-721782228?

> the API server can handle way above that number with better etcd tuning.

Yes, but do we have this etcd tuning already? If we don't set a limit then we can expect people to set this to a very high number, and this will definitely result in scalability issues and ops effort. I would really like to limit this, at least at the beginning, to be on the safe side. It's already possible to set it twice as high as the default which should be a good first improvement. I'm open to further increase the limit in case we see that we can cater with it.

OK?

timuthy (dismissed):
/lgtm

danielfoehrKn (dismissed):
Worked. One more thing to keep in mind for the kube-apiserver refactoring :)

rfranzke:
@danielfoehrKn Yes, since you started there were a couple of changes already

pkg/apis/core/validation/shoot.go
mvladev: Here is what K8S is doing:

https://github.com/kubernetes/kubernetes/blob/c7911a384cbc11a4b5003da081b181d6b814d07e/test/kubemark/resources/start-kubemark-master.sh#L566-L570

For 1000 Nodes max requests in flight is 1500 and mutating 500
For 3000 Nodes max requests in flight is 3000 and mutating 1000

For 5000 Nodes I would guess that something like 5000 / 2000 could be used.

mvladev:
Wouldn't make more sense for the configuration to be like:

```yaml
spec:
  kubernetes:
     kubeAPIServer:
        requests:
          maxMutatingInFlight: 300
          maxNonMutatingInFlight: 400
          # minTimeout: 1000 # this doesn't exist yet but it can be added to configure the --min-request-timeout flag
```

This would allow for API expansion and focusing on requests.
I would use `maxNonMutatingInFlight` with emphasis on non-mutating as this is the behavior of the flag.

pkg/apis/core/validation/shoot.go
rfranzke: Thanks for the reference. I still suggest to start with above values, we can increase them later if needed.

rfranzke:
@mvladev I've made the API change as you proposed, PTAL.

timebertt (dismissed):
/lgtm

ialidzhikov (dismissed):
/lgtm

rfranzke:
@timebertt @ialidzhikov I had to rebase

ialidzhikov (approved):
/lgtm

timebertt (approved):

PR #3142, opened 2020-11-10
DNSEntry TTL is now configurable in gardenlet config

### LINKED ISSUE #2998: Make DNS Entry TTL configurable

/kind enhancement
/priority normal

The ttl of the DNS entry deployed by the gardenlet is currently set to [120](https://github.com/gardener/gardener/blob/master/charts/seed-dns/entry/values.yaml#L3).
Some dns providers enforce a minimum ttl of more than 120 e.g. 300.

To support such dns providers, it should be possible to make the ttl configurable.

rfranzke:

/area usability networking
/kind enhancement
/priority normal

It is now possible to configure the TTL used for `DNSEntry` objects in the shoot controller via the gardenlet's component config. It still defaults to `120s`.

Fixes #2998

**Release note**:

```noteworthy operator
It is now possible to configure the TTL used for `DNSEntry` objects in the shoot controller via the gardenlet's component config (`.controllers.shoot.dnsEntryTTLSeconds`, default: `120`).
```


pkg/gardenlet/apis/config/validation/validation.go
```diff
@@ -42,3 +51,37 @@ func ValidateGardenletConfiguration(cfg *config.GardenletConfiguration) field.Er
...
+		allErrs = append(allErrs, apivalidation.ValidateNonnegativeField(int64(cfg.RetryDuration.Duration), fldPath.Child("retryDuration"))...)
+	}
+
+	if cfg.SyncPeriod != nil {
+		allErrs = append(allErrs, apivalidation.ValidateNonnegativeField(int64(cfg.SyncPeriod.Duration), fldPath.Child("syncPeriod"))...)
+	}
+
+	if cfg.DNSEntryTTLSeconds != nil {
+		const (
+			dnsEntryTTLSecondsMin = 0
```
MartinWeindel: Recursive DNS servers, whether open recursives or local resolvers, typically do not acknowledge a TTL of less than 30 seconds. Therefore I would set the allowed minimum to 30.

pkg/gardenlet/apis/config/validation/validation.go
```diff
@@ -42,3 +51,37 @@ func ValidateGardenletConfiguration(cfg *config.GardenletConfiguration) field.Er
...
+			dnsEntryTTLSecondsMax = 3600
+		)
+
+		if *cfg.DNSEntryTTLSeconds < 0 || *cfg.DNSEntryTTLSeconds > 3600 {
```
MartinWeindel: use defined constants

MartinWeindel (changes requested):
There is a minor inconsistency with the validation

pkg/gardenlet/apis/config/types.go
```diff
@@ -200,6 +200,8 @@ type ShootControllerConfiguration struct {
 	RetryDuration *metav1.Duration
 	// SyncPeriod is the duration how often the existing resources are reconciled.
 	SyncPeriod *metav1.Duration
+	// DNSEntryTTLSeconds is the TTL in seconds that is being used for DNS entries when reconciling shoots.
+	DNSEntryTTLSeconds *int64
```
danielfoehrKn: Can you add the default value to the description?

pkg/gardenlet/apis/config/validation/validation.go
```diff
@@ -72,3 +81,37 @@ func validateSNI(sniPath *field.Path, sni *config.SNI) field.ErrorList {
...
+	}
+
+	if cfg.SyncPeriod != nil {
+		allErrs = append(allErrs, apivalidation.ValidateNonnegativeField(int64(cfg.SyncPeriod.Duration), fldPath.Child("syncPeriod"))...)
+	}
+
+	if cfg.DNSEntryTTLSeconds != nil {
+		const (
+			dnsEntryTTLSecondsMin = 30
+			dnsEntryTTLSecondsMax = 3600
```
danielfoehrKn: Where is this value coming from?

rfranzke: From space ;-) Jokes aside, there is no specific reason why I chose it. Probably it's even too large. Maybe we should set the maximum to something like `300` or `5m`. Otherwise, the control plane migration won't work smoothly, I assume. I've set to to `300` now. 
/cc @plkokanov @swilen-iwanow 

Hope that is still fine for your use-cases @schrodit @MartinWeindel

schrodit: 300 is fine for the OTC use case.
@vasu1124 do you know any other dns service that has a higher min ttl?

rfranzke: Let's go ahead with `300` then. we can increase it later if needed.

MartinWeindel: Sorry for my late answer, but I just remember Jerry Jia told me, that in the case of Alicloud at least 600 seconds are needed to avoid buying enterprise service for DNS.

rfranzke: OK, thanks. I'll change it to `600` then.

rfranzke: Wait a sec, actually it is already on `600`. please check.

MartinWeindel (approved):
/lgtm

timebertt (approved):
/lgtm

PR #3144, opened 2020-11-10
Reset shoot rate limiting on deletion

### LINKED ISSUE #3050: Faster requeueing when exponentially backoffed shoot shall be deleted

/area usability
/kind enhancement
/priority normal

When a shoot that fails to be created or reconciled for a long time then it is exponentially backoffed. When it is later deleted by the user, then it might still take the full backoff time (~16m) until the deletion flow starts (because we only use one queue and only differentiate which flow to start when we pick it up).
The same applies for other operation types like migration and restoration.

Let's enhance our event handlers in the shoot controller and enqueue shoots faster when the operation type changes.

/cc @timebertt @vlerenc @dguendisch 

**Why is this needed**:
Faster reaction time of Gardener, improved user experience

rfranzke:

/area usability
/kind enhancement
/priority normal

The gardenlet's shoot controller does now forget about the rate limiting when a shoot's deletion timestamp is set. This is to make starting deletion operations faster.

Fixes #3050

**Release note**:

```improvement user
The gardenlet's shoot controller does now forget about the rate limiting when a shoot's deletion timestamp is set. This is to make starting deletion operations faster.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control.go
@@ -52,11 +52,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-func (c *Controller) shootAdd(obj interface{}) {
+func (c *Controller) shootAdd(obj interface{}, resetRateLimiting bool) {
 	key, err := cache.MetaNamespaceKeyFunc(obj)
 	if err != nil {
 		logger.Logger.Errorf("Couldn't get key for object %+v: %v", obj, err)
 		return
+	}
+
+	if resetRateLimiting {
+		c.getShootQueue(obj).Forget(key)
 	}
 	c.getShootQueue(obj).Add(key)
 }
@@ -80,7 +84,11 @@ func (c *Controller) shootUpdate(oldObj, newObj interface{}) {
 		return
 	}
 
-	c.shootAdd(newObj)
+	// If the shoot's deletion timestamp is set then we want to forget about the potentially established exponential
+	// backoff and enqueue it faster.
+	resetRateLimiting := oldShoot.DeletionTimestamp == nil && newShoot.DeletionTimestamp != nil
+
+	c.shootAdd(newObj, resetRateLimiting)
 }
 
 func (c *Controller) shootDelete(obj interface{}) {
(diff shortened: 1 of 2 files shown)


pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -52,12 +52,16 @@ import (
...
-func (c *Controller) shootAdd(obj interface{}) {
+func (c *Controller) shootAdd(obj interface{}, resetRateLimiting bool) {
 	key, err := cache.MetaNamespaceKeyFunc(obj)
 	if err != nil {
 		logger.Logger.Errorf("Couldn't get key for object %+v: %v", obj, err)
 		return
 	}
+
+	if resetRateLimiting {
+		c.getShootQueue(obj).Forget(key)
```
timuthy, line 63: Is resetting really necessary? IIUC, `.Add(key)` adds the element immediately to the queue and `.AddRateLimited(key)` with a delay.

rfranzke: Hm, I haven't studied all the internals of these queues, but if this would be true then this PR would be unnecessary.
The doc string of `Forget` states:

```golang
// Forget indicates that an item is finished being retried.  Doesn't matter whether it's for perm failing
// or for success, we'll stop the rate limiter from tracking it.  This only clears the `rateLimiter`, you
// still have to call `Done` on the queue.
```

I'm not calling `Done()` because a worker might be in the process of reconciling the shoot (my understanding is that if I'd be calling `Done()` then it might happen that two workers can simultaneously process a shoot). `Forget()` is doing exactly what I want - clearing the rate limiter.

timuthy (approved):
Okay, thanks for explaining. I tried to reproduce it:

- `Add()` adds the element immediately to the queue w/o rate limits due to the object composition:

  -- [rateLimitingType](https://github.com/kubernetes/client-go/blob/bef66adadf9a6371b5c114db0b95c12412c2414b/util/workqueue/rate_limiting_queue.go#L52) with [AddRateLimited()](https://github.com/kubernetes/client-go/blob/bef66adadf9a6371b5c114db0b95c12412c2414b/util/workqueue/rate_limiting_queue.go#L59) and [Forget()](https://github.com/kubernetes/client-go/blob/bef66adadf9a6371b5c114db0b95c12412c2414b/util/workqueue/rate_limiting_queue.go#L67), --> no `Add()` override
  ---- [delayingType](https://github.com/kubernetes/client-go/blob/master/util/workqueue/delaying_queue.go#L73) with [AddAfter()](https://github.com/kubernetes/client-go/blob/bef66adadf9a6371b5c114db0b95c12412c2414b/util/workqueue/delaying_queue.go#L160), --> no `Add()` override
  ------ [Type](https://github.com/kubernetes/client-go/blob/bef66adadf9a6371b5c114db0b95c12412c2414b/util/workqueue/queue.go#L65) with [Add()](https://github.com/kubernetes/client-go/blob/bef66adadf9a6371b5c114db0b95c12412c2414b/util/workqueue/queue.go#L108)



- I suspect in #3050, the first delete operation failed again and thus it was added by `AddRateLimited()` which already got an effective high rate limit.

/lgtm

timebertt (approved):
/lgtm

PR #3129, opened 2020-11-09
Refactor etcd deployment

rfranzke:
**How to categorize this PR?**

/area dev-productivity quality robustness
/kind technical-debt test
/priority normal
/topology seed
/platform all
/exp intermediate

**What this PR does / why we need it**:
This PR refactors the deployment for the etcd and adopts the component concept. Consequently, it eliminates the Helm chart and adds full unit tests, instead. This improves several aspects and the overall code quality.

**Which issue(s) this PR fixes**:
Part of #2754 

/cc @shreyas-s-rao @amshuman-kr 
ℹ️ Similar refactoring for etcd-druid: #3099

**Release note**:

```improvement operator
NONE
```

PR #3150, opened 2020-11-11
Refactor metrics-server deployment

rfranzke:

/kind technical-debt test
/priority normal
/topology seed
/platform all
/exp intermediate

This PR refactors the deployment for the metrics-server and adopts the component concept. Consequently, it eliminates the Helm chart and adds full unit tests, instead. This improves several aspects and the overall code quality.

Part of #2754

This PR also introduces a registry helper for `ManagedResource`s and adapts other components to use it. I'll adapt the same for the etcd component (#3129) once the PR is merged in a follow-up work.

**Release note**:

```improvement operator
NONE
```


vpnachev (approved):
/lgtm

pkg/utils/managedresources/registry.go
```diff
@@ -0,0 +1,153 @@
...
+		codec:        codec.CodecForVersions(serializer, serializer, schema.GroupVersions(groupVersions), schema.GroupVersions(groupVersions)),
+		nameToObject: make(map[string]*object),
+	}
+}
+
+// Add adds the given object the registry. It computes a filename based on its type, namespace, and name. It serializes
+// the object to YAML and stores both representations (object and serialization) in the registry.
+func (r *Registry) Add(obj runtime.Object) error {
+	r.lock.Lock()
+	defer r.lock.Unlock()
```
timebertt, line 61: Is locking really necessary for the registry? I probably won't/shouldn't be used concurrently in multiple goroutines, right?
If we have several registries per ongoing shoot reconciliation each locking and unlocking with high frequency, I would imagine that it _could_ have some performance impacts.

pkg/utils/managedresources/registry.go
```diff
@@ -0,0 +1,153 @@
...
+	for name, object := range r.nameToObject {
+		out = append(out, fmt.Sprintf("* %s:\n%s", name, object.serialization))
+	}
+	return strings.Join(out, "\n\n")
+}
+
+func (r *Registry) objectName(obj runtime.Object) (string, error) {
+	var (
+		typeOf   = strings.Split(reflect.TypeOf(obj).String(), ".")
+		typeName = strings.ToLower(typeOf[len(typeOf)-1])
```
timebertt, line 139: Why did you choose to use reflection directly here, instead of reusing the scheme for checking the GVK of the object?
You can also use `apiutil.GVKForObject`, if convenience is a concern...

timebertt (commented):
I have two questions about this PR, although it's already merged :)

pkg/utils/managedresources/registry.go
rfranzke: Yeah, I wasn't even sure myself when I implemented it. I wanted to be on the safe side, but your comment confirms my gud feeling that it's most likely not necessary. I'll remove it, thanks!

pkg/utils/managedresources/registry.go
rfranzke: No specific reason, I simply wasn't thinking of this api utility function, thanks for mentioning it. I'll improve it.

PR #3168, opened 2020-11-16
Do not redeploy external DNS record for shoot deletion

### LINKED ISSUE #3159: Deletion deadlock with invalid primary domain

/area usability
/kind bug
/priority normal

**What happened**:

Gardener is unable to delete shoot clusters, with a invalid primary dns domain.

Gardener tries to redeploy dns entries before starting the deletion flow, which will result in a deadlock if the dns configuration is invalid.

```
   task "Deploying external domain DNS record" failed: Failed to create "external" DNS record: "DNS record \"external\" is not ready (status=Error, message=No responsible provider found)" (status=Error, message=No responsible provider found)
```

The only possible fix is to patch the etcd as the primary dns field is immutable.

Gardener should be able to delete cluster if a wrong dns was specified.

1. Create a shoot cluster with a invalid primary dns:
  ```
  dns:
    domain: invalid.exampl.com
    providers:
      - domains:
          include:
            -  invalid.example.com
        primary: true
        secretName: my-secret
        type: cloudflare-dns
  ```
2. Delete the cluster

**Environment**:

- Gardener version: 1.12.8

rfranzke:

/kind bug
/priority normal

The external DNS record is only used in the kubeconfig handed out to the end-user. Gardener itself and all system components use the internal DNS record (or in-cluster communication).
Hence, it's not necessary the redeploy the external DNS record.

Fixes #3159

/cc @mvladev 

**Release note**:

```improvement user
An issue has been fixed that can cause `Shoot`s from being stuck in deletion due to invalid `.spec.dns` configuration.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_delete.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_delete.go
@@ -256,11 +256,6 @@ func (c *Controller) runDeleteShootFlow(ctx context.Context, o *operation.Operat
 			Fn:           flow.TaskFn(botanist.DeployInternalDNS).DoIf(cleanupShootResources),
 			Dependencies: flow.NewTaskIDs(deployReferencedResources, waitUntilKubeAPIServerServiceIsReady),
 		})
-		deployExternalDomainDNSRecord = g.Add(flow.Task{
-			Name:         "Deploying external domain DNS record",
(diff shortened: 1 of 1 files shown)


timuthy (approved):

danielfoehrKn (approved):

vpnachev (approved):
/lgtm

PR #3173, opened 2020-11-16
Introduce MaintenancePreconditionsSatisfied shoot constraint

### LINKED ISSUE #3122: Detect webhooks that are potentially problematic for maintenance activities and place a constraint on shoots

/area ops-productivity
/kind enhancement
/priority normal

We should detect webhooks that are potentially problematic for maintenance activities and place a constraint on shoots of type `MaintenancePossible` to forbid maintenance similarly to how we currently forbid hibernation for similar reasons. The constraint could look like this:

```yaml
    constraints:
    - type: MaintenancePossible
      status: 'False'
      lastTransitionTime: '2020-11-05T02:43:34Z'
      lastUpdateTime: '2020-10-29T07:32:57Z'
      reason: ProblematicWebhooks
      message: >-
        Maintenance activities cannot be performed on this shoot because of 
        MutatingWebhookConfiguration "sen-webhooks-shoot": webhook 
        "vpn-service-webhook.sen.sap.com" with failurePolicy "Fail" will probably 
        cause the maintenance activities to fail.
```

Alternatively, we may refrain from forbidding maintenance, but still add a constraint and a visual cue in the dashboard based on this constraint, e.g. a red "Schedule Maintenance" button with a popup that displays the constraint message.

**Why is this needed**:
During DoD activities, we noticed errors where the shoot reconciliation during maintenance activities failed due to a user webhook modifying `kube-system` pods. If the webhook is not reachable from `kube-apiserver` for whatever reason (e.g. no running `vpn-shoot` pods, or webhook pods), calling the webhook may fail with `context deadline exceeded` even if the failure policy is set to `Fail`, or even if it's set to `Ignore` and the timeout is long enough (30s or more). This in turn prevents `kube-system` pods from being created, which may prevent `calico` from properly initializing and cause the nodes to fail to join the cluster due to kubelet reporting `NetworkPluginNotReady`.

The issue is partially mitigated by https://github.com/gardener/gardener/pull/3100, but not prevented entirely.

The solution proposed above was suggested in the DoD sync.

rfranzke:

/area usability
/kind enhancement
/priority normal

The `Shoot` now has a new constraint with type `MaintenancePreconditionsSatisfied` which indicates whether it's safe to maintain a shoot (see [this document](https://github.com/gardener/gardener/blob/master/docs/usage/shoot_maintenance.md) to get an overview what happens during maintenance). End-users should check this information to properly configure their clusters in order to avoid problems.

Fixes #3122

/cc @gardener/dashboard-maintainers 

**Release note**:

```noteworthy user
The `Shoot` now has a new constraint with type `MaintenancePreconditionsSatisfied` which indicates whether it's safe to maintain a shoot (see [this document](https://github.com/gardener/gardener/blob/master/docs/usage/shoot_maintenance.md) to get an overview what happens during maintenance). End-users should check this information to properly configure their clusters in order to avoid problems.
```

### DIFF
--- a/docs/usage/shoot_status.md
+++ b/docs/usage/shoot_status.md
@@ -50,11 +50,15 @@ Let's check the following example to get better understanding. Let's say that th
 
 Constraints represent conditions of a Shoot’s current state that constraint some operations on it.
 
-Currently there is only one Shoot constraint:
+Currently there are two constraints:
 
 - `HibernationPossible`
 
   This constraint indicates whether a Shoot is allowed to be hibernated. The rationale behind this constraint is that a Shoot can have `ValidatingWebhookConfiguration`s or `MutatingWebhookConfiguration`s with rules for CREATE Pods or Nodes and `failurePolicy=Fail`. Such webhooks are preventing wake up for a Shoot cluster as new Nodes cannot be created or new system component Pods cannot be created because the webhook is not running. To prevent such deadlock situation, the gardener-apiserver does not allow Shoot to be hibernated when the `HibernationPossible` has status `False`.
+
+- `MaintenancePreconditionsSatisfied`
+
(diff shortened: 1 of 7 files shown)


vpnachev (approved):
/lgtm
/reviewed ok-to-test

timebertt (approved):
Thanks!
/lgtm

PR #3176, opened 2020-11-16
Do not wait KCM if it is already gone

vpnachev:

/kind bug
/priority normal

Fix a bug where the gardenlet is stuck on waiting the KCM to be scaled up when it is already deleted.

Fixes #

**Release note**:

```improvement operator
A bug in shoot deletion flow that was stuck waiting the `kube-controller-manager` to be scaled up when it is already deleted is now fixed.
```


rfranzke (approved):
/lgtm

PR #3178, opened 2020-11-17
Prevent orphaned machine resources on worker deletion

### LINKED ISSUE #2453: Generic worker actuator leaves orphaned MCM resources when deleting shoots
**What happened**:
The generic worker actuator sometimes tends to leave orphan resources while shoot deletion.

These resources are not expected to be left out.

Not really sure.

**Anything else we need to know**:
The creation timestamps on the orphan resources were later than shoot deletion timestamps which is weird. We will need to look into this. 

**Environment**:
Seem on Azure and other infra as well.

rfranzke:

/area ops-productivity robustness quality
/kind bug
/priority normal

Most likely caused by stale caches, in rare situations it might happen that the generic worker actuator's deletion flow leaves orphaned machine classes and/or machine class secrets in the shoot namespace in the seed.
In order to prevent that, the flow is now improved as follows:

* Instead of first listing all classes/secrets/deployments we use a `DELETECOLLECTION` call and move the responsibility to the API server to list and delete individually.
* To be sure that we don't miss objects we use the API reader in the `waitUntilMachineResourcesDeleted` function instead of relying on the cache.

Extension controllers using the generic worker actuator have to implement a new function that returns an initialized machine class object (so far, only machine class list objects were required).

Fixes #2453

/invite @timebertt @ialidzhikov @prashanth26 

**Release note**:

```action developer
The generic worker actuator interface does now include a new function `MachineClass() runtime.Object` that needs to be implemented. It is similar to `MachineClassList() runtime.Object` with the difference that it does not return the list object but the machine class object itself.
```
```noteworthy operator
A bug has been fixed that might led to orphaned machine resources in the shoot namespace in the seed that are stuck with the machine-controller-manager finalizer.
```
```action developer
The `CleanupLeakedClusterRoles` function has been removed from the generic worker actuator package. You can find more information about it [here](https://github.com/gardener-attic/gardener-extensions/pull/378) and [here](https://github.com/gardener/gardener/issues/2144).
```

### DIFF
--- a/extensions/pkg/controller/worker/genericactuator/actuator_delete.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator_delete.go
@@ -23,6 +23,7 @@ import (
 	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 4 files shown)


extensions/pkg/controller/worker/genericactuator/actuator_delete.go
```diff
@@ -68,25 +69,19 @@ func (a *genericActuator) Delete(ctx context.Context, worker *extensionsv1alpha1
...
-	if err := a.cleanupMachineDeployments(ctx, logger, existingMachineDeployments, nil); err != nil {
-		return errors.Wrapf(err, "cleaning up machine deployments failed")
+	if err := a.client.DeleteAllOf(ctx, &machinev1alpha1.MachineDeployment{}, client.InNamespace(worker.Namespace)); err != nil {
+		return errors.Wrapf(err, "cleaning up all machine deployments failed")
 	}
 
 	// Delete all machine classes.
-	if err := a.cleanupMachineClasses(ctx, logger, worker.Namespace, workerDelegate.MachineClassList(), nil); err != nil {
-		return errors.Wrapf(err, "cleaning up machine classes failed")
+	if err := a.client.DeleteAllOf(ctx, workerDelegate.MachineClass(), client.InNamespace(worker.Namespace)); err != nil {
```
timebertt, line 81: One note:
`client.DeleteAllOf` should also work with the list type, see https://github.com/kubernetes-sigs/controller-runtime/blob/33f1d80ebeaa33c9e3d86c86c713ac5847dbe859/pkg/client/client_cache.go#L54-L57
Though, adding the `MachineClass` func and doing it this way is probably a bit cleaner...

extensions/pkg/controller/worker/genericactuator/actuator_delete.go
```diff
@@ -68,25 +69,19 @@ func (a *genericActuator) Delete(ctx context.Context, worker *extensionsv1alpha1
...
-	// Get the list of all existing machine deployments.
-	existingMachineDeployments := &machinev1alpha1.MachineDeploymentList{}
-	if err := a.client.List(ctx, existingMachineDeployments, client.InNamespace(worker.Namespace)); err != nil {
-		return err
-	}
-
 	// Delete all machine deployments.
-	if err := a.cleanupMachineDeployments(ctx, logger, existingMachineDeployments, nil); err != nil {
-		return errors.Wrapf(err, "cleaning up machine deployments failed")
+	if err := a.client.DeleteAllOf(ctx, &machinev1alpha1.MachineDeployment{}, client.InNamespace(worker.Namespace)); err != nil {
```
timebertt, line 75: Can you add back some logging for these steps?
```suggestion
	logger.Info("Deleting all MachineDeployments")
	if err := a.client.DeleteAllOf(ctx, &machinev1alpha1.MachineDeployment{}, client.InNamespace(worker.Namespace)); err != nil {
```

timebertt (changes requested):
Two minor comments, rest looks good.

extensions/pkg/controller/worker/genericactuator/actuator_delete.go
rfranzke: Hm, in my tests using the list type I got the following error:

```
{"level":"error","ts":"2020-11-17T17:45:40.281+0100","logger":"controller","msg":"Reconciler error","controller":"worker_controller","name":"worker","namespace":"shoot--foobar--aws2","error":"cleaning up all machine classes failed: object does not implement the Object interfaces","errorVerbose":"object does not implement the Object interfaces\n"}
```

timebertt: Hmm, ok. Then my assumption seems to be wrong...

timebertt (approved):
Thanks.
/lgtm

prashanth26 (approved):
/lgtm

PR #3183, opened 2020-11-18
Health check only considers metrics-server MR for gardener >= 1.13

rfranzke:

/area ops-productivity
/kind bug
/priority normal

With the metrics-server refactoring in #3150 a new `ManagedResource` was introduced in the shoot namespace in the seed. However, it should only be considered if the shoot was already handled by a Gardener version >= 1.13, and this check was missing.

/invite @timebertt 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_care_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_care_control.go
@@ -259,6 +259,7 @@ func (c *defaultCareControl) Care(shootObj *gardencorev1beta1.Shoot, key string)
 				initializeShootClients,
 				c.conditionThresholdsToProgressingMapping(),
 				c.config.Controllers.ShootCare.StaleExtensionHealthCheckThreshold,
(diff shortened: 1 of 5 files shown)


pkg/gardenlet/controller/shoot/shoot_care_control.go
```diff
@@ -177,6 +178,11 @@ func (c *defaultCareControl) Care(shootObj *gardencorev1beta1.Shoot, key string)
 		return fmt.Errorf("failed to get garden client: %w", err)
 	}
 
+	gardenerVersion, err := semver.NewVersion(shoot.Status.Gardener.Version)
```
timebertt: Can we add a `GardenerSemver` member or something similar to the `shoot.Shoot` type?
I guess, this has potential for reuse :)

timebertt (commented):
One nit :)

timebertt (approved):
Thanks!
/lgtm

PR #3184, opened 2020-11-18
Fix node local dns issue for control plane pods

### LINKED ISSUE #2745: NodeLocalDNS Feature lacks DNS network policy for control-plane components

DockToFuture:

/area networking
/kind bug
/priority normal

Traffic is now allowed to `cluster dns` and the `node local ipvs address` to resolve a dns resolution issue with the `NodeLocalDNS` feature for dns names in control plane pods.

Fixes #2745

**Release note**:

```improvement operator
Traffic is now allowed to `cluster dns` and the `node local ipvs address` to resolve a dns resolution issue with the `NodeLocalDNS` feature for dns names in control plane pods.
```

### DIFF
--- a/charts/global-network-policies/templates/allow-to-dns.yaml
+++ b/charts/global-network-policies/templates/allow-to-dns.yaml
@@ -15,6 +15,16 @@ spec:
       networking.gardener.cloud/to-dns: allowed
   egress:
   - to:
+    {{- if .Values.nodeLocalDNSEnabled }}
+    {{- if .Values.dnsServer }}
+    - ipBlock:
+        cidr: {{ .Values.dnsServer }}/32 # required for node local dns feature, allows egress traffic to kube-dns
+    {{- end }}
+    {{- if .Values.nodeLocalIPVSAddress }}
+    - ipBlock:
+        cidr: {{ .Values.nodeLocalIPVSAddress }}/32 # required for node local dns feature, allows egress traffic to node local dns cache
+    {{- end }}
+    {{- end }}
     - namespaceSelector:
         matchLabels:
           role: kube-system
--- a/charts/global-network-policies/values.yaml
+++ b/charts/global-network-policies/values.yaml
@@ -20,3 +20,7 @@ privateNetworks:
 - network: 100.64.0.0/10
   except:
   - 100.64.1.0/24
+
+dnsServer:  ""
+nodeLocalIPVSAddress: ""
+nodeLocalDNSEnabled: false
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -729,8 +729,11 @@ func getResourcesForAPIServer(nodeCount int32, scalingClass string) (string, str
 func (b *Botanist) deployNetworkPolicies(ctx context.Context, denyAll bool) error {
 	var (
 		globalNetworkPoliciesValues = map[string]interface{}{
-			"blockedAddresses": b.Seed.Info.Spec.Networks.BlockCIDRs,
-			"denyAll":          denyAll,
+			"blockedAddresses":     b.Seed.Info.Spec.Networks.BlockCIDRs,
+			"denyAll":              denyAll,
+			"dnsServer":            b.Shoot.Networks.CoreDNS.String(),
+			"nodeLocalIPVSAddress": NodeLocalIPVSAddress,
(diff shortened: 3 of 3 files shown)


charts/global-network-policies/templates/allow-to-dns.yaml
```diff
@@ -15,6 +15,10 @@ spec:
       networking.gardener.cloud/to-dns: allowed
   egress:
   - to:
+    - ipBlock:
+        cidr: 100.64.0.10/32 # required for node local dns
+    - ipBlock:
+        cidr: 169.254.20.10/32 # required for node local dns
```
rfranzke: Can you explain more about it, i.e., why are those particular two ranges required?

charts/global-network-policies/templates/allow-to-dns.yaml
```diff
@@ -15,6 +15,10 @@ spec:
       networking.gardener.cloud/to-dns: allowed
   egress:
   - to:
+    - ipBlock:
+        cidr: 100.64.0.10/32 # required for node local dns feature, allows egress traffic to kube-dns
```
rfranzke: What if my service CIDR is `10.20.0.0/16`, for example?

charts/global-network-policies/templates/allow-to-dns.yaml
```diff
@@ -15,6 +15,12 @@ spec:
       networking.gardener.cloud/to-dns: allowed
   egress:
   - to:
+    {{- if .Values.dnsServer }}
+    - ipBlock:
+        cidr: {{ .Values.dnsServer }}/32 # required for node local dns feature, allows egress traffic to kube-dns
+    {{- end }}
+    - ipBlock:
+        cidr: 169.254.20.10/32 # required for node local dns feature, allows egress traffic to node local dns cache
```
rfranzke: Can you inject this as Helm value and use https://github.com/gardener/gardener/blob/6b183a89697dd64b88bdd4e60cc04a299d40bd16/pkg/operation/botanist/types.go#L33 to not duplicate the CIDR here?

rfranzke (changes requested):
Thansk @DockToFuture, better now, one minor suggestion before merging

charts/global-network-policies/templates/allow-to-dns.yaml
```diff
@@ -15,6 +15,14 @@ spec:
...
   egress:
   - to:
+    {{- if .Values.dnsServer }}
+    - ipBlock:
+        cidr: {{ .Values.dnsServer }}/32 # required for node local dns feature, allows egress traffic to kube-dns
+    {{- end }}
+    {{- if .Values.nodeLocalIPVSAddress }}
+    - ipBlock:
+        cidr: {{ .Values.nodeLocalIPVSAddress }}/32 # required for node local dns feature, allows egress traffic to node local dns cache
+    {{- end }}
```
rfranzke, line 26: Thanks! Those two fields are now always added even if the feature gate is disabled. Should this be taken into account?

rfranzke (approved):
/lgtm

PR #3188, opened 2020-11-19
Extend fluent-bit dashboard with custom metrics

Kristian-ZH:

/area logging
/kind enhancement
/priority normal

Upgrade output plugin version to `v0.29.0`
Include in the fluent-bit dashboard new metrics from the custom plugin.
The new fluent-bit image also includes a fix for the OutOfOrder Loki behaviour

Fixes #

@wyb1 @vlvasilev 

**Release note**:

```improvement operator
Sort logs to fix out of order issue
```
```improvement operator
The output plugin exposes custom metrics
```
```improvement operator
Modified fluent-bit dashboard to include the new metrics
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -204,7 +204,7 @@ images:
 - name: fluent-bit-plugin-installer
   sourceRepository: github.com/gardener/logging
   repository: eu.gcr.io/gardener-project/gardener/fluent-bit-to-loki
-  tag: "v0.28.0"
+  tag: "v0.32.0"
 - name: loki
   sourceRepository: github.com/grafana/loki
   repository: grafana/loki
--- a/charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-daemonset.yaml
+++ b/charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-daemonset.yaml
@@ -54,7 +54,10 @@ spec:
             memory: 150Mi
         ports:
         - name: metrics
-          containerPort: 2020
+          containerPort: {{ .Values.ports.metrics }}
+          protocol: TCP
+        - name: metrics-plugin
+          containerPort: {{ .Values.ports.outputPluginMetrics }}
           protocol: TCP
         readinessProbe:
           httpGet:
--- a/charts/seed-bootstrap/templates/aggregate-prometheus/config.yaml
+++ b/charts/seed-bootstrap/templates/aggregate-prometheus/config.yaml
@@ -132,6 +132,26 @@ data:
       metric_relabel_configs:
 {{ include "prometheus.keep-metrics.metric-relabel-config" .Values.allowedMetrics.fluentbit| indent 6 }}
 
+    - job_name: fluent-bit-output-plugin
+      honor_labels: false
+      kubernetes_sd_configs:
+      - role: endpoints
+        namespaces:
+          names: [ garden ]
+      relabel_configs:
+      - source_labels:
+        - __meta_kubernetes_service_name
+        - __meta_kubernetes_endpoint_port_name
+        action: keep
+        regex: fluent-bit;metrics-plugin
+      # common metrics
+      - action: labelmap
+        regex: __meta_kubernetes_service_label_(.+)
+      - source_labels: [ __meta_kubernetes_pod_name ]
+        target_label: pod
+      metric_relabel_configs:
+{{ include "prometheus.keep-metrics.metric-relabel-config" .Values.allowedMetrics.fluentbitOutputPlugin | indent 6 }}
+
     - job_name: 'vpa-exporter'
       kubernetes_sd_configs:
       - role: endpoints
--- a/charts/seed-bootstrap/values.yaml
+++ b/charts/seed-bootstrap/values.yaml
@@ -111,6 +111,14 @@ allowedMetrics:
   - fluentbit_filter_add_records_total
   - fluentbit_filter_drop_records_total
 
+  fluentbitOutputPlugin:
+  - promtail_dropped_entries_total
+  - fluentbit_loki_gardener_errors_total
+  - fluentbit_loki_gardener_logs_without_metadata_total
+  - fluentbit_loki_gardener_incoming_logs_total
+  - fluentbit_loki_gardener_forwarded_logs_total
+  - fluentbit_loki_gardener_dropped_logs_total
+
   vpa:
   - vpa_status_recommendation
   - vpa_spec_container_resource_policy_allowed
@@ -153,6 +161,7 @@ allowedMetrics:
   - loki_panic_total
   - loki_logql_querystats_duplicates_total
   - loki_logql_querystats_ingester_sent_lines_total
+  - prometheus_target_scrapes_sample_out_of_order_total
 
   istiod:
   - galley_istio_authentication_meshpolicies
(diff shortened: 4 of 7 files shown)


vlvasilev (dismissed):
/lgtm

charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-daemonset.yaml
```diff
@@ -56,6 +56,9 @@ spec:
         - name: metrics
           containerPort: 2020
           protocol: TCP
+        - name: metrics-plugin
+          containerPort: 2021
```
rfranzke:
```suggestion
          containerPort: {{ .Values.ports.outputPluginMetrics }}
```

charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-daemonset.yaml
```diff
@@ -56,6 +56,9 @@ spec:
         - name: metrics
           containerPort: 2020
```
rfranzke:
```suggestion
          containerPort: {{ .Values.ports.metrics }}
```

Kristian-ZH:
Will convert this PR to draft because I would like to change a little bit the dashboard

Kristian-ZH:
ready

rfranzke (dismissed):
/lgtm
/needs second-opinion
/invite @wyb1 @istvanballok @vpnachev

vpnachev (dismissed):
/lgtm

vlvasilev:
There is bug with fluent-bit-to-loki 0.30 version where some times we close a wait group two times leading to panic.
Please do not merge with 0.30 version

charts/images.yaml
```diff
@@ -204,7 +204,7 @@ images:
 - name: fluent-bit-plugin-installer
   sourceRepository: github.com/gardener/logging
   repository: eu.gcr.io/gardener-project/gardener/fluent-bit-to-loki
-  tag: "v0.28.0"
+  tag: "v0.30.0"
```
vlvasilev: Bugged version of fluent-bit-to-loki plugin.

vlvasilev (changes requested):

charts/images.yaml
vpnachev: /hold

wyb1 (dismissed):
/lgtm

just out of curiosity when would these tables be different?

image

Kristian-ZH:
> /lgtm
> 
> just out of curiosity when would these tables be different?
> 
> image

Wrong configurations for the Output plugin for example. Between the two metrics there are logic for removing unnecessary keys, getting client for the correct Loki  and  forming the log line.

If there is difference between the two tables, it indicates an issue in the output plugin so we will create also alerts for these metrics

Kristian-ZH:
After that the log is processed to the Promtail client which sends it to the Loki. For this reason we monitor the Promtail dropped logs also

rfranzke:
What's the status of this PR, and in particular, can you share details about https://github.com/gardener/gardener/pull/3188#discussion_r529608425 and the plan forward?

Kristian-ZH:
> What's the status of this PR, and in particular, can you share details about [#3188 (comment)](https://github.com/gardener/gardener/pull/3188#discussion_r529608425) and the plan forward?



> What's the status of this PR, and in particular, can you share details about [#3188 (comment)](https://github.com/gardener/gardener/pull/3188#discussion_r529608425) and the plan forward?

We are waiting to release a new version of the logging where this bug is fixed. Now the release job is triggered.
The bug is that we were calling two sequential  times WaitGroup.Done().

rfranzke:
Thank you @Kristian-ZH 👍

rfranzke (dismissed):
/lgtm

vpnachev (dismissed):
/lgtm

vlvasilev (dismissed):
/lgtm

vpnachev:
@wyb1 @istvanballok could you take a look?

charts/seed-bootstrap/templates/aggregate-prometheus/config.yaml
```diff
@@ -132,6 +132,28 @@ data:
...
 
+    - job_name: fluent-bit-output-plugin
+      honor_labels: false
+      kubernetes_sd_configs:
+      - role: endpoints
+        namespaces:
+          names: [ garden ]
+      relabel_configs:
+      - target_label: __metrics_path__
+        replacement: /metrics
```
istvanballok: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config
`/metrics` is the default metrics path, so you might be able to omit these 2 lines

charts/seed-bootstrap/values.yaml
```diff
@@ -111,6 +111,15 @@ allowedMetrics:
   - fluentbit_filter_add_records_total
   - fluentbit_filter_drop_records_total
 
+  fluentbitOutputPlugin:
+  - promtail_dropped_entries_total
+  - output_plugin_errors
+  - output_plugin_no_metadata
+  - output_plugin_incoming_logs
+  - output_plugin_past_send_requests
+  - output_plugin_dropt_logs
```
istvanballok: I think it might be a bit more involved to fix this typo: `dropt`, as it is coming from the fluent-bit-to-loki plugin,
https://github.com/gardener/logging/blob/99b64c43b0a2e84e833b74d318cad8292431ad34/pkg/metrics/metrics.go#L54
but better now then later :) [drop -> dropped](https://www.oxfordlearnersdictionaries.com/definition/english/drop_1)

Kristian-ZH: Yes, I know already about this typo, but as you mentioned, it is not coming from this PR.

Kristian-ZH: So lets hold this PR until I release the logging repo with the renaming fix

istvanballok:
Metric names are strings that are typically used (=hard coded) in diverse locations (recording rules, alerting rules, promql queries in Grafana dashboards), with no "compiler" support to assist with renames - so we can only dream about the comfortable situation when golang developers rename a variable with their IDE, and it just works. Furthermore, the motivation for scraping metrics is to store them for a few weeks at least, to show how the values change over time, which makes renaming metrics even more challenging (e.g. the promql query would have to support both names during the transition period, as long as the old metric names are still kept around according to the retention policy - otherwise, the old metrics would be lost for the visualization).
Long story short, after chatting with @Kristian-ZH, we concluded that it might be worthwhile to fix the typos and adhere to the [naming conventions](https://prometheus.io/docs/practices/naming/#metric-and-label-naming) (e.g. _total suffix for counters) for the very first release of the custom metrics feature.

vpnachev (approved):
/lgtm
/invite @istvanballok

charts/seed-bootstrap/values.yaml
```diff
@@ -111,6 +111,14 @@ allowedMetrics:
   - fluentbit_filter_add_records_total
   - fluentbit_filter_drop_records_total
 
+  fluentbitOutputPlugin:
+  - promtail_dropped_entries_total
+  - fluentbit_loki_gardener_errors_total
+  - fluentbit_loki_gardener_logs_without_metadata_total
+  - fluentbit_loki_gardener_incoming_logs_total
+  - fluentbit_loki_gardener_forwarded_logs_total
+  - fluentbit_loki_gardener_dropped_logs_total
```
istvanballok, line 120: 👍

istvanballok (approved):
/lgtm

PR #3192, opened 2020-11-19
Fix test namespace dump accidentally reusing a potentially expired context

dguendisch:

/area testing
/kind bug
/priority normal

In case of errors during test namespace deletion some k8s resources of the namespace are dumped. This dumping however reused the given context which at this point in time might already have been expired.
Therefore a new context with 1min is created to allow resource dumping.

Fixes #

**Release note**:


rfranzke (approved):
/lgtm

schrodit (approved):

PR #3194, opened 2020-11-20
Editable dashboards by default

wyb1:

/area monitoring
/kind enhancement
/priority normal

All dashboards should now be editable by default (after logging in). Previously on some dashboards the operator would have to change the settings to make a dashboard editable.

Also some panels "connected" data when there was no data. This makes it difficult to identify when there is missing data (either prometheus was down or the target was down) which can be relevant for debugging. This has been changed for some panels.

Fixes #

**Release note**:

```improvement operator
NONE
```


rfranzke (approved):

istvanballok (approved):
/lgtm

PR #3196, opened 2020-11-20
Reconcile Infrastructure for hibernated clusters

timuthy:

/area robustness
/kind enhancement
/priority normal

https://github.com/gardener/gardener/commit/29b45ab868a5e5ba3c80f9da4c020241350292ee skipped infrastructure reconciliation during maintenance to save infrastructure calls as well as our previous logging stack couldn't deal with Terraformer logs while the logging endpoint was not available. We can now enable infrastructure reconciliation again for these clusters, since the current logging implementation just disregards component logs while the cluster is hibernated (/cc @vlvasilev). Furthermore, regular infra reconciliation is especially necessary for long-term hibernated clusters because only certain versions of Gardener and Extensions may contain important migration logic.

/cc @ialidzhikov @dkistner 

**Release note**:

```improvement operator
Gardener triggers an infrastructure reconciliation during maintenance also for hibernated clusters. This ensures that the infrastructure is always up-to-date, even for long-term hibernated clusters.
```

### DIFF
--- a/pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
+++ b/pkg/controllermanager/controller/shoot/shoot_maintenance_control.go
@@ -195,9 +195,7 @@ func (c *defaultMaintenanceControl) Maintain(shootObj *gardencorev1beta1.Shoot, 
 			metav1.SetMetaDataAnnotation(&s.ObjectMeta, v1beta1constants.GardenerOperation, common.ShootOperationRetry)
 		}
 
-		if !gardencorev1beta1helper.HibernationIsEnabled(s) {
-			controllerutils.AddTasks(s.Annotations, common.ShootTaskDeployInfrastructure)
-		}
+		controllerutils.AddTasks(s.Annotations, common.ShootTaskDeployInfrastructure)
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm

ialidzhikov:
Thanks for the PR! What about the logic in `pkg/operation/botanist/infrastructure.go` from https://github.com/gardener/gardener/commit/29b45ab868a5e5ba3c80f9da4c020241350292ee, can we revert it also?

rfranzke:
@ialidzhikov why? I guess it makes sense to request infrastructure reconciliation when a shoot is woken up, or?

ialidzhikov:
> @ialidzhikov why? I guess it makes sense to request infrastructure reconciliation when a shoot is woken up, or?

Yeah, I have no objections against it. Just noted that prior to 29b45ab we had only the maintenance controller that adds the deployInfrastructure task.

ialidzhikov (approved):
/lgtm

rfranzke:
Yes, I guess the PR back then included two improvements in the same commit. 👀

dkistner (approved):

PR #3204, opened 2020-11-21
Allow injecting arbitrary EnvVars into terraformer pods

timebertt:

/area security quality
/kind cleanup
/priority normal

Currently, the terraformer library offers only one mechanism for injecting environment variables into terraformer pods:
setting the plain values directly in the `PodSpec` via `Terraformer.SetVariablesEnvironment`.
This is not desirable or acceptable for some cases, especially from a security point-of-view, e.g. for setting terraform variables (i.e. infrastructure credentials) via env vars (`TF_VAR_*`) like most provider extensions hosted under the gardener org do.

This PR deprecates the old `Terraformer.SetVariablesEnvironment` function in favor of the newly added and more flexible `Terraformer.SetEnvVars` function.
The new function allows to inject arbitrary `corev1.EnvVars` into the `PodSpec` (e.g. with a `secretKeyRef`).

Fixes #

**Release note**:

```action developer
`Terraformer.SetVariablesEnvironment` has been deprecated in favor of `Terraformer.SetEnvVars`. Please adapt your usage of the terraformer library accordingly.
```

### DIFF
--- a/extensions/pkg/terraformer/config.go
+++ b/extensions/pkg/terraformer/config.go
@@ -16,8 +16,8 @@ package terraformer
 
 import (
 	"context"
-	"errors"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
@@ -42,8 +42,25 @@ const (
 )
 
 // SetVariablesEnvironment sets the provided <tfvarsEnvironment> on the Terraformer object.
+// Deprecated: use SetEnvVars instead
 func (t *terraformer) SetVariablesEnvironment(tfvarsEnvironment map[string]string) Terraformer {
-	t.variablesEnvironment = tfvarsEnvironment
+	var envVars []corev1.EnvVar
+	for k, v := range tfvarsEnvironment {
+		envVars = append(envVars, corev1.EnvVar{Name: k, Value: v})
+	}
+
+	// maps are unsorted, sort resulting slice after looping over map to avoid flickering results
+	sort.Slice(envVars, func(i, j int) bool {
+		return envVars[i].Name < envVars[j].Name
+	})
+
+	t.envVars = append(t.envVars, envVars...)
+	return t
+}
+
+// SetEnvVars sets the provided environment variables for the Terraformer Pod.
+func (t *terraformer) SetEnvVars(envVars ...corev1.EnvVar) Terraformer {
+	t.envVars = append(t.envVars, envVars...)
 	return t
 }
 
@@ -188,10 +205,6 @@ func (t *terraformer) prepare(ctx context.Context) (int, error) {
 	numberOfExistingResources, err := t.NumberOfResources(ctx)
 	if err != nil {
 		return -1, err
-	}
-
-	if t.variablesEnvironment == nil {
-		return -1, errors.New("no Terraform variables environment provided")
 	}
 
 	// Clean up possible existing pod artifacts from previous runs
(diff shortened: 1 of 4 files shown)


rfranzke (dismissed):
/lgtm

extensions/pkg/terraformer/config.go
```diff
@@ -42,8 +42,25 @@ const (
...
+// Deprecated: use SetEnvVars instead
 func (t *terraformer) SetVariablesEnvironment(tfvarsEnvironment map[string]string) Terraformer {
-	t.variablesEnvironment = tfvarsEnvironment
+	var envVars []corev1.EnvVar
+	for k, v := range tfvarsEnvironment {
+		envVars = append(envVars, corev1.EnvVar{Name: k, Value: v})
+	}
+
+	// maps are unsorted, sort resulting slice after looping over map to avoid flickering results
+	sort.SliceStable(envVars, func(i, j int) bool {
```
kon-angelo: Nitpick: I would think that you don't need stable sorting since you are sorting on map keys and those should be unique

extensions/pkg/terraformer/types.go
```diff
@@ -96,7 +97,9 @@ const (
 type Terraformer interface {
 	UseV2(bool) Terraformer
 	SetLogLevel(string) Terraformer
+	// Deprecated: use SetEnvVars instead
```
kon-angelo, line 100: I would like to highlight the fact that It seems the behaviour was changed from a "hard" set to an "update/append".  The same behaviour is used for the replacement `SetEnvVars`. The main API loss here is the ability to remove keys by setting a new map. This makes me wonder whether deprecation vs out-right removal from the interface is better.

Disclaimer: I haven't checked all usages of the (newly) deprecated function.

extensions/pkg/terraformer/config.go
timebertt: Well, that's a nice catch 😄
You're right, so I changed it to `sort.Slice`.

extensions/pkg/terraformer/types.go
timebertt: > behaviour was changed from a "hard" set to an "update/append" [...] vs out-right removal from the interface is better

Yes, that observation is correct. I mainly did it via `append`, so users can use both functions in parallel. Though, I don't know if there's even a use case for it. And I also only deprecated the old func, so that users of the library are not put under pressure to transition/migrate. In general, I like to allow such smooth transitions wherever possible and I don't see a problem with it here.

> The main API loss here is the ability to remove keys by setting a new map

Yes, that's also true. But I guess, there is no need for it.

Side-note: This comes with an `action` release note and when developers vendor such a new gardener release, they are responsible for checking if their code is still compatible (at best with a good suite of unit tests). So I don't see any problem with the changed behaviour, as the difference is quite subtle.

But if you see any problems with it, please bring them up :)

kon-angelo: My (somewhat theoretical) how-should-it-be-handled PoV here was that given the subtle change of behaviour in the deprecated function we could:
a) make the `envVar` translation but keep the hard-set for full backwards compatibility
b) outright delete the func to force an action during the revendoring. 

But in practice I haven't seen such a use in our repos, so no objections on my side. The `action` note (which admittedly I missed on my first read) does the job.

kon-angelo (approved):
/lgtm

rfranzke (approved):

PR #3216, opened 2020-11-24
Remove GardenerID and GardenerName checks from CheckSeed method

krgostev:
/area control-plane-migration
/kind enhancement
/priority normal

Remove GardenerID and GardenerName checks Seed is healthy as those values are defined for each gardenlet and will most likely be different when CP migration is triggered. Otherwise the error below will occur and the migration process will not start.
`Shoot cannot be synced with Seed: seed is not yet ready: observing Gardener version not up to date (&Gardener{ID:3180e2fe4ad646e011c35e7faf68a5f10e2837407feaa9b876d1e25257f86014,Name:gardenlet-7dfc58f57f-dfxdp,Version:v1.12.8,}/&Gardener{ID:543c33e44050deeeaa2175342764563559021950f1d567f6ee6c859b05342341,Name:gardenlet-5c84bbc686-d82p6,Version:v1.12.8,})`
Part of [Issue 1631](https://github.com/gardener/gardener/issues/1631)

**Release note**:
```
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control.go
@@ -205,7 +205,7 @@ func (c *Controller) reconcileShootRequest(req reconcile.Request) (reconcile.Res
 	log = log.WithField("operation", "reconcile")
 
 	if shouldPrepareShootForMigration(shoot) {
-		if err := c.isSeedAvailable(seed); err != nil {
+		if err := c.isSeedReadyForMigration(seed); err != nil {
 			return reconcile.Result{}, fmt.Errorf("target Seed is not available to host the Control Plane of Shoot %s: %v", shoot.GetName(), err)
 		}
 
@@ -365,12 +365,12 @@ func (c *Controller) finalizeShootDeletion(ctx context.Context, gardenClient kub
 	return reconcile.Result{}, c.removeFinalizerFrom(ctx, gardenClient, shoot)
 }
 
-func (c *Controller) isSeedAvailable(seed *gardencorev1beta1.Seed) error {
+func (c *Controller) isSeedReadyForMigration(seed *gardencorev1beta1.Seed) error {
 	if seed.DeletionTimestamp != nil {
 		return fmt.Errorf("seed is set for deletion")
 	}
 
-	return health.CheckSeed(seed, c.identity)
+	return health.CheckSeedForMigration(seed, c.identity)
 }
 
 func (c *Controller) reconcileShoot(logger *logrus.Entry, shoot *gardencorev1beta1.Shoot, project *gardencorev1beta1.Project, cloudProfile *gardencorev1beta1.CloudProfile, seed *gardencorev1beta1.Seed) (reconcile.Result, error) {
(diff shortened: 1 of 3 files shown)


pkg/utils/kubernetes/health/health.go
```diff
@@ -248,8 +247,8 @@ func CheckSeed(seed *gardencorev1beta1.Seed, identity *gardencorev1beta1.Gardene
 	if seed.Status.ObservedGeneration < seed.Generation {
 		return fmt.Errorf("observed generation outdated (%d/%d)", seed.Status.ObservedGeneration, seed.Generation)
 	}
-	if !apiequality.Semantic.DeepEqual(seed.Status.Gardener, identity) {
-		return fmt.Errorf("observing Gardener version not up to date (%v/%v)", seed.Status.Gardener, identity)
+	if seed.Status.Gardener.Version != identity.Version {
```
rfranzke, line 258: Instead of removing it always, should we rather have a special check for cp migration that only considers the version? Otherwise we miss cases in which the gardenlet pod is redeployed and needs to bootstrap the seed again.

stoyanr: Agreed, it must not be removed in principle. @kris94 In our initial discussion I suggested adding another function `CheckSeedForMigration` without this check, or adding a boolean parameter to this function e.g. `forMigration`, and disabling the check only when it's true.

stoyanr (changes requested):

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -370,7 +370,7 @@ func (c *Controller) isSeedAvailable(seed *gardencorev1beta1.Seed) error {
 		return fmt.Errorf("seed is set for deletion")
 	}
 
-	return health.CheckSeed(seed, c.identity)
+	return health.CheckSeedForMigration(seed, c.identity)
```
rfranzke, line 373: Can you rename the `isSeedAvailable` function to `isSeedReadyForMigration` or something similar? It wasn't clear from the name and I had to check where this thing is actually called.

rfranzke (changes requested):
Looks great, minor suggestion

rfranzke (dismissed):
/lgtm

pkg/utils/kubernetes/health/health_test.go
```diff
@@ -227,6 +227,59 @@ var _ = Describe("health", func() {
 		)
 	})
 
+	Describe("CheckSeedForMigration", func() {
+		DescribeTable("seeds",
+			func(seed *gardencorev1beta1.Seed, identity *gardencorev1beta1.Gardener, matcher types.GomegaMatcher) {
+				Expect(health.CheckSeedForMigration(seed, identity)).To(matcher)
+			},
+			Entry("healthy with mathing version", &gardencorev1beta1.Seed{
```
plkokanov:
```suggestion
			Entry("healthy with matching version", &gardencorev1beta1.Seed{

plkokanov (commented):
Just a really minor typo.
Otherwise /lgtm

rfranzke (approved):

PR #3219, opened 2020-11-24
Parse the time zones from logs and set the tail plugin to read from head

vlvasilev:

/area logging
/kind enhancement
/priority normal

With this PR we:
- parse the time zone of a log when reading it from the node /var/log/containers directory.
- when fluent-bit containers runs the tail plugin starts to read a file from the head(like it was prior fluent-bit 1.6).
- make the readiness and liveness probe fail after 30 seconds and the liveness probes starts after 90 seconds.
- `get`, `list` and `watch` for Pods are removed from the fluent-bit RBAC as no log needed.
 
Fixes #

**Release note**:

```other operator
Parse the time zone of a log when reading it from the node /var/log/containers directory.
```
```other operator
When fluent-bit containers runs the tail plugin starts to read a file from the head(like it was prior fluent-bit 1.6).
```
```other operator
Make the readiness and liveness probe fail after 30 seconds and the liveness probes starts after 90 seconds.
```
```other operator
`get`, `list` and `watch` for Pods are removed from the fluent-bit RBAC as no longer needed.
```

### DIFF
--- a/charts/seed-bootstrap/charts/fluent-bit/templates/_fluent-bit-config.tpl
+++ b/charts/seed-bootstrap/charts/fluent-bit/templates/_fluent-bit-config.tpl
@@ -27,6 +27,7 @@
         Exclude_Path      *_garden_fluent-bit-*.log,*_garden_loki-*.log
         Parser            docker
         DB                /var/log/flb_kube.db
+        read_from_head    true
         Skip_Long_Lines   On
         Mem_Buf_Limit     30MB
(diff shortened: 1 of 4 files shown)


rfranzke (approved):
/lgtm
/needs second-opinion
/invite @vpnachev @Kristian-ZH

Kristian-ZH (approved):
/lgtm

vpnachev (approved):

PR #3220, opened 2020-11-24
Switch to "gardener.cloud/role" label key for system components

ialidzhikov:

/kind cleanup
/priority normal

As the system components and addons are managed via ManagedResource, we can adapt the role label key on the system-components. The healthchecks are performed via the ManagedResoure. There is also the `shoot.gardener.cloud/no-cleanup` label that is injected via the ManagedResource that prevents system-components to be deleted as part of the deletion flow - see also https://github.com/gardener/gardener-extension-networking-calico/pull/7.

Ref #1649

**Release note**:

```noteworthy user
The system components that were previous specifying label `garden.sapcloud.io/role: (optional-addon|monitoring|system-component)` are now adapted to specify `gardener.cloud/role: (optional-addon|monitoring|system-component)`.
```

```noteworthy operator
The Shoot namespace in the Seed no longer specifies label `garden.sapcloud.io/role: shoot`.
```

### DIFF
--- a/charts/shoot-core/components/charts/vpn-shoot/templates/deployment.yaml
+++ b/charts/shoot-core/components/charts/vpn-shoot/templates/deployment.yaml
@@ -6,6 +6,7 @@ metadata:
   namespace: kube-system
   labels:
     gardener.cloud/role: system-component
+    # TODO (ialidzhikov): remove in a future version
     garden.sapcloud.io/role: system-component
     app: vpn-shoot
     origin: gardener
@@ -32,6 +33,7 @@ spec:
       labels:
         origin: gardener
         gardener.cloud/role: system-component
+        # TODO (ialidzhikov): remove in a future version
         garden.sapcloud.io/role: system-component
         app: vpn-shoot
         type: tunnel
--- a/pkg/operation/botanist/cleanup.go
+++ b/pkg/operation/botanist/cleanup.go
@@ -70,7 +70,7 @@ var (
 	GracePeriodFiveMinutes = utilclient.DeleteWith{client.GracePeriodSeconds(5 * 60)}
 
 	// NotSystemComponent is a requirement that something doesn't have the GardenRole GardenRoleSystemComponent.
-	NotSystemComponent = utils.MustNewRequirement(v1beta1constants.DeprecatedGardenRole, selection.NotEquals, v1beta1constants.GardenRoleSystemComponent)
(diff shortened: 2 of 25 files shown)


charts/shoot-core/components/charts/vpn-shoot/templates/deployment.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   namespace: kube-system
   labels:
     gardener.cloud/role: system-component
+    # TODO (ialidzhikov): remove in a future version
```
rfranzke, line 9: Why not now?

pkg/operation/botanist/cleanup.go
```diff
@@ -70,7 +70,7 @@ var (
 	GracePeriodFiveMinutes = utilclient.DeleteWith{client.GracePeriodSeconds(5 * 60)}
 
 	// NotSystemComponent is a requirement that something doesn't have the GardenRole GardenRoleSystemComponent.
-	NotSystemComponent = utils.MustNewRequirement(v1beta1constants.DeprecatedGardenRole, selection.NotEquals, v1beta1constants.GardenRoleSystemComponent)
```
rfranzke, line 73: I guess we should keep this for a couple of releases to give people enough time to reconcile all their clusters, right? Also, some shoots might already have a deletion timestamp but still the deprecated label.

ialidzhikov: As I tried to describe in the PR, the system components have also the no-cleanup label set via the ManagedResource and it should prevent them to be deleted during cluster deletion.

charts/shoot-core/components/charts/vpn-shoot/templates/deployment.yaml
ialidzhikov: Because there is currently a netpol that is using the deprecated label and this netpol is adapted to the new label key in this PR.

rfranzke: I see, thanks for the explanation!

pkg/operation/botanist/cleanup.go
rfranzke: OK, then we are good. Just wanted to raise and clarify this point again  to be sure we don't miss a case.

rfranzke (approved):
/lgtm

timebertt (approved):
Thanks!
/lgtm

PR #2807, opened 2020-09-02
Improve worker controller/actuator logs

timebertt:
**How to categorize this PR?**

/area logging quality
/kind enhancement
/priority normal

**What this PR does / why we need it**:
This PR improves/refactors the worker controller and generic actuator logs to
- add the `"worker"` log field only once per operation (makes it more consistent and available across all logs)
- add the `"operation"` log field to easily identify the flow the controller/actuator is currently executing (one of `reconcile, delete, migrate, restore`)
- generally add more logs on each step

The main motivation for this is to make the logs more useful when debugging issues like #2453.
It might be used as a reference for improving log handling in the other extension controllers as well.

**Which issue(s) this PR fixes**:
Ref https://github.com/gardener/gardener/issues/2453#issuecomment-683759031

**Special notes for your reviewer**:

**Release note**:

```improvement operator
Logs in the worker controller and generic actuator have been improved to make them more consistent and useful.
```

PR #3223, opened 2020-11-25
Cleanup terraformer library

timebertt:

/area logging quality
/kind cleanup
/priority normal

This PR cleans up several things related to the terraformer library:
- removes deprecated functions: `Terraformer.SetVariablesEnvironment` and `GenerateVariablesEnvironment` (⚠️ breaking change)
- cleans up `context.TODO`s and replace them by proper contexts passed by callers (⚠️ breaking change)
- improve the logs in the infrastructure controller/reconciler similar to https://github.com/gardener/gardener/pull/2807
- switch to `logr` in the terraformer, as all the extensions use it everywhere and using `logrus` only there makes the logs inconsistent and hard to read (⚠️ breaking change)

Fixes #

**Release note**:

```breaking dependency
The deprecated functions in the terraformer library (`SetVariablesEnvironment` and `GenerateVariablesEnvironment`) have been removed.
```
```breaking dependency
The `Terraformer` functions have been changed to allow passing proper contexts. Please adapt your usage accordingly.
```
```breaking dependency
The terraformer library was switched to `logr` instead of `logrus` in order to have more consistent and readable logging in the infrastructure controllers of provider extensions. Please adapt your usage accordingly.
```

### DIFF
--- a/extensions/pkg/terraformer/terraformer.go
+++ b/extensions/pkg/terraformer/terraformer.go
@@ -21,7 +21,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/sirupsen/logrus"
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -35,6 +35,7 @@ import (
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
+	kutils "github.com/gardener/gardener/pkg/utils/kubernetes"
 	"github.com/gardener/gardener/pkg/utils/retry"
 )
 
@@ -47,11 +48,11 @@ const (
 
 type factory struct{}
 
-func (factory) NewForConfig(logger logrus.FieldLogger, config *rest.Config, purpose, namespace, name, image string) (Terraformer, error) {
+func (factory) NewForConfig(logger logr.Logger, config *rest.Config, purpose, namespace, name, image string) (Terraformer, error) {
 	return NewForConfig(logger, config, purpose, namespace, name, image)
 }
 
-func (f factory) New(logger logrus.FieldLogger, client client.Client, coreV1Client corev1client.CoreV1Interface, purpose, namespace, name, image string) Terraformer {
+func (f factory) New(logger logr.Logger, client client.Client, coreV1Client corev1client.CoreV1Interface, purpose, namespace, name, image string) Terraformer {
 	return New(logger, client, coreV1Client, purpose, namespace, name, image)
 }
 
@@ -66,7 +67,7 @@ func DefaultFactory() Factory {
 
 // NewForConfig creates a new Terraformer and its dependencies from the given configuration.
 func NewForConfig(
-	logger logrus.FieldLogger,
+	logger logr.Logger,
 	config *rest.Config,
 	purpose,
 	namespace,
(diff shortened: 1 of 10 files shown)


extensions/pkg/terraformer/terraformer.go
```diff
@@ -178,50 +183,56 @@ func (t *terraformer) execute(ctx context.Context, command string) error {
...
 
 	if !skipApplyOrDestroyPod {
 		// Create Terraform Pod which executes the provided command
 		generateName := t.computePodGenerateName(command)
+
+		logger.Info("Deploying Terraformer Pod", "generateName", generateName)
 		pod, err := t.deployTerraformerPod(ctx, generateName, command)
 		if err != nil {
-			return fmt.Errorf("failed to deploy the Terraformer Pod with .meta.generateName '%s': %s", generateName, err.Error())
+			return fmt.Errorf("failed to deploy the Terraformer Pod with .meta.generateName %q: %w", generateName, err)
```
vpnachev, line 196: Why are you switching to `%w`? It is not looking very good to me.

```golang
package main

import (
	"fmt"
)

func main() {
	const name, id = "bueller", 17
	err := fmt.Errorf("user %q (id %d) not found", name, id)
	fmt.Printf("%w\n", err) // %!w(*errors.errorString=&{user "bueller" (id 17) not found})
	fmt.Printf("%v\n", err) // user "bueller" (id 17) not found
}
```
The above code snippet is taken from [here](https://golang.org/pkg/fmt/#Errorf)

extensions/pkg/terraformer/waiter.go
```diff
@@ -19,10 +19,12 @@ import (
 	"fmt"
 	"time"
 
-	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
-	"github.com/gardener/gardener/pkg/utils/retry"
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
```
vpnachev, line 25:
```suggestion
```

vpnachev (commented):
Looks very good to me. 
I have just one comment about formatting.

extensions/pkg/terraformer/terraformer.go
timebertt: `%w` is only working with `fmt.Errorf`: https://golang.org/pkg/fmt/#Errorf
That's why it's looking strange with `fmt.Printf` (behaves like `%+v`, I believe).

E.g.:
```go
const name, id = "bueller", 17
err := fmt.Errorf("user %q (id %d) not found", name, id)
fmt.Println(fmt.Errorf("wrapped error: %w", err)) // wrapped error: user "bueller" (id 17) not found
```

kon-angelo: Well the printing should be the same. Snippet for reference:

```go
package main

import (
	"fmt"
)

func main() {
	const name, id = "bueller", 17
	err := fmt.Errorf("user %q (id %d) not found", name, id)
	errOut1:= fmt.Errorf("%v",err)
	errOut2:= fmt.Errorf("%w",err)
	fmt.Printf("%v\n", errOut1) //prints user "bueller" (id 17) not found
	fmt.Printf("%v\n", errOut2) //user "bueller" (id 17) not found
}
```

@timebertt You should use `%w` only if you  want the callers to have access to the original error. If you think that the callers cannot use the internal error then wrap it with `%v`

```go
return fmt.Errorf("failed to deploy the Terraformer Pod with .meta.generateName %q: %v", generateName, err)
```

vpnachev: Thanks for the explanations :heart: .

timebertt: > @timebertt You should use `%w` only if you want the callers to have access to the original error. If you think that the callers cannot use the internal error then wrap it with `%v`

I guess, in general it makes sense to wrap errors, that will be propagated up to other components (as it's the case here).
They might decide to actually unwrap the errors at some point.

vpnachev (approved):
/lgtm
/needs second-opinion

kon-angelo (approved):

rfranzke (approved):
/lgtm

Issue #3023, opened 2020-10-22
Invalid credentials error on machineDeployment updates


prashanth26:
**What happened**:
Invalid Credentials error on rolled out machines

I am seeing this occurring when a machineDeployment has undergone a rolling update (with new machineset/machines) and is pointing to a new set of machineClass (and credentials) and one of the old machine pointing to the old machineClass (and old credentials) is stuck in the deletion and the credentials are outdated. Now with updating the credentials to a new one, only the new machineClass receives the update but the old machineClass doesn't. This seems to be a corner case that we need to improve handling here. 

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

- Create a machineDeployment with valid credentials.
- Now invalidate the credentials by removing access to it.
- Now trigger a rolling update in parallel by modifying the machineClass. 
- The update fails.
- Now update credentials to the new one on machineDeployment by pointing to new credentials. 
- Now the old machine is still stuck in deletion with old credentials. 

**Anything else we need to know**:
Probably occurs more with clusters in hibernation deletion. 

**Environment**:
Openstack

prashanth26:
Agreed. Even I was confused as to where to keep the issue. But I think it's better solved at g/g.

prashanth26:
I will take a look at this issue. However, will need a few days to start working on this as I want to finish the MCM OOT story for AWS. Will assign it to me when I start working on this.

timuthy:
@prashanth26 can you provide an update, is the issue already being worked on?

prashanth26:
Hi @timuthy ,

I wasn't getting any time to work on this. However, @vpnachev was kind enough to pick it instead of me for now. So I think he is working on this.

vpnachev:
All extension have been fixed, I am closing now this issue.
/close

PR #3224, opened 2020-11-25
Update cloud credentials even for unwanted machine deployments.

vpnachev:

/area robustness
/kind bug
/priority normal

The worker controller will now update the cloud credentials also for machine deployments that are not referenced by the worker resource, e.g. they are in deletion. 

Part of https://github.com/gardener/gardener/issues/3023

**Release note**:

```bugfix dependency
The generic worker actuator is now ensuring that all machine class secrets have up-to-date cloud credentials.
```

```breaking dependency
The `WorkerDelegate` must implement method `GetMachineControllerManagerCloudCredentials` returning map with cloud credential keys and values just like they are used by the machine-controller-manager.
```

### DIFF
--- a/extensions/pkg/controller/worker/genericactuator/actuator.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator.go
@@ -213,6 +213,27 @@ func (a *genericActuator) cleanupMachineClassSecrets(ctx context.Context, logger
 	return nil
 }
 
+// updateCloudCredentialsInAllMachineClassSecrets updates the cloud credentials
+// for all existing machine class secrets.
+func (a *genericActuator) updateCloudCredentialsInAllMachineClassSecrets(ctx context.Context, logger logr.Logger, cloudCredentials map[string][]byte, namespace string) error {
+	logger.Info("Updating cloud credentials for existing machine class secrets")
+	secretList, err := a.listMachineClassSecrets(ctx, namespace)
+	if err != nil {
+		return errors.Wrapf(err, "failed to list machine class secrets in namespace %s", namespace)
+	}
+
+	for _, secret := range secretList.Items {
+		secretCopy := secret.DeepCopy()
+		for key, value := range cloudCredentials {
+			secretCopy.Data[key] = value
+		}
+		if err := a.client.Patch(ctx, secretCopy, client.MergeFrom(&secret)); err != nil {
+			return errors.Wrapf(err, "failed to patch secret %s/%s with cloud credentials", namespace, secret.Name)
+		}
+	}
+	return nil
+}
+
 // shallowDeleteMachineClassSecrets deletes all unused machine class secrets (i.e., those which are not part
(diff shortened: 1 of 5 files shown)


vpnachev:
A sample implementation in a provider extension looks like https://github.com/gardener/gardener-extension-provider-aws/pull/223

rfranzke (commented):
Wouldn't a more "lightweight" and cleaner approach be if we split the `DeployMachineClasses` function into `DeployMachineClassSecrets` and `DeployMachineClasses`. Then `DeployMachineClassSecrets` can always crupdate all machine class secrets, including those still in the system but meanwhile potentially unwanted, and `DeployMachineClasses` can handle the machine classes without crupdating the secrets. This is more adaptation effort in the extensions, but probably has a cleaner flow and a cleaner `WorkerDelegate` interface. What is your opinion?

vpnachev:
Actually, I wanted to move the cloud credential into dedicated single secret and mount it into the machine-controller-manager deployment, just how it is done for the cloud-controller-manager, however MCM does not support such option at the moment.
At the moment, I do not see the value of splitting `DeployMachineClasses` into two separate functions.

I changed the PR to patch all machine class secrets, for those that are up-to-date, this should be an empty patch.

rfranzke:
In a discussion with @vpnachev and @timuthy we talked about an option for easier handling in the future: If the machine classes would support to read the infrastructure credentials from a different secret than the user-data then we don't need all this special logic. We could just reuse the already existing `cloudprovider` secret.

The MCM maintainers @hardikdr @prashanth26 @AxiomSamarth wanted to discuss internally.
/hold until the colleagues came back

Assuming that we agree on this, we'd like to merge this PR as is, despite the fact that it "uglyfies" the `WorkerDelegate` interface a little bit. However, it would be only a temporary step until MCM supports the single secret, and it results in only minimal changes in the actual provider extensions.

/assign @vpnachev @timuthy @hardikdr @prashanth26 @AxiomSamarth

rfranzke:
The MCM maintainers accepted the proposal, so we can
/unhold
this PR and proceed with it until the discussed enhancement (https://github.com/gardener/machine-controller-manager/pull/578) is merged and available for us and the provider extensions to use.

timuthy:
Frankly, I didn't expect an implementation in MCM this quick (thanks @rfranzke 🚀). Does it then still even make sense to proceed with this PR instead of waiting for the next MCM release?

rfranzke:
For backwards compatibility reasons we should proceed, I think. Some provider extension might want/need to stick to a certain MCM version for whatever reasons, so let's get this fix in or for a couple of releases. WDYT?

timuthy:
> For backwards compatibility reasons we should proceed, I think. Some provider extension might want/need to stick to a certain MCM version for whatever reasons, so let's get this fix in or for a couple of releases. WDYT?

Okay, makes sense. Thank you!

timuthy (approved):

rfranzke (approved):
/lgtm

PR #3226, opened 2020-11-25
Force gardenlet restart if SNI is enabled in the Seed cluster

mvladev:

/area networking
/kind bug
/priority normal

For seed clusters created from Shoot clusters, make sure to restart the `gardenlet` so the `KUBERNETES_SERVICE_HOST` environment variable is injected and federated network policy controller is working correctly.

Fixes #

**Release note**:

```noteworthy operator
`gardenlet` is now restarted if `APIServerSNI` is enabled on the Seed cluster.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/seed_registration_control.go
+++ b/pkg/gardenlet/controller/shoot/seed_registration_control.go
@@ -32,9 +32,11 @@ import (
 	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
 	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/keys"
 	"github.com/gardener/gardener/pkg/controllerutils"
+	"github.com/gardener/gardener/pkg/features"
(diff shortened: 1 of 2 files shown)


vpnachev:
https://github.com/gardener/gardener/blob/7060587f524b75a165e22d2c00563b1b3c01e69c/charts/gardener/gardenlet/charts/runtime/templates/deployment.yaml#L23 is already restarting gardenlet when the feature gate is enabled/disabled, isn't it?

rfranzke (approved):
/lgtm

rfranzke:
@vpnachev the problem is that the gardenlet must be restarted after the shoot(ed seed) it is getting installed to has been reconciled with gardener v1.13

only then the shoot(ed seed)'s apiserver has the injector sidecar container which would inject the needed `KUBERNETES_SERVICE_HOST` environment variable into the gardenlet pod

vpnachev (approved):
Ok, thanks for the explanation.
/lgtm

PR #3228, opened 2020-11-25
Add minAllowed VPA ResourcePolicy for Cluster Autoscaler

timuthy:

/area scalability
/area robustness
/kind enhancement
/priority normal

This PR adds minAllowed VPA ResourcePolicy values for Cluster Autoscaler based on observed and analyzed values on our landscapes.

We experienced cases in which the VPA recommender only recommended `10Mi` and this again led to OOM kills. `10Mi` with a request:limit ration of 1:10 is too low wrt our observation. Hence, we want to prevent such outliers. 

**Release note**:

```improvement operator
NONE
```


vpnachev (approved):
/lgtm

rfranzke (approved):
/lgtm

PR #3233, opened 2020-11-27
Add another special case to set_dependency_version script

AndreasBurger:

/kind bug
/priority normal

The script at `hack/.ci/set_dependency_version` is vendored in to and used by most of our components. There was a recent change to the etcd-druid component that requires another special case to be handled properly.

Also pinging @shreyas-s-rao since he'll need to revendor eventually.


rfranzke (approved):
/lgtm

shreyas-s-rao:
Thanks a lot @AndreasBurger!

PR #3235, opened 2020-11-27
Do not use FQDN for KUBERNETES_SERVICE_HOST injection

mvladev:

/area networking
/kind bug
/priority normal

Some Kubernetes clients are not respecting the [TLS extension specification][1] and are not removing the trailing dot. They request server name which is not recognized by envoy and TLS handshake fails.

[1]: https://tools.ietf.org/html/rfc6066#section-3

Fixes #

**Release note**:

```bugfix user
The `KUBERNETES_SERVICE_HOST` environment variable injected when `APIServerSNI` is enabled no longer includes a trailing dot (being a Fully Qualified Domain Name) due to several homebrew kubernetes clients not properly handling it and sending wrong server name when initiating a TLS conneciton.
```


rfranzke (approved):
/lgtm

Issue #3160, opened 2020-11-14
Support for cron-based maintenance time windows


vlerenc:
**What would you like to be added**:
Today we support only daily maintenance time windows (that we also use for reconciliation), but we see people asking for more flexibility here and have weekly or quarterly maintenance time windows. Also we would benefit from weekly maintenance time windows as we cannot migrate our seeds anymore during one night without running maintenance on multiple seeds at the same time. One can always opt-out and schedule maintenance manually or through a private controller, but that's inconvenient for our stakeholders and us.

**Why is this needed**:
There is customer demand (agreed maintenance tine windows on weekends or quarterly) and we have the need for our many seeds to avoid parallel maintenance and it could also help, for the time being until we have #2885 cc @stoyanr, to update our seeds not all in one go, but staged. Should we detect issues, we could then abort.

**Considerations**:
- Reconciliation should still happen once per 24h (unclear when during the day, in this case)
- OS updates can be skipped
- Kubernetes updates can only be executed consecutively, but @danielfoehrKn implemented that already in the context of our versioning policy where we potentially have to forcefully migrate clusters multiple times

PR #3242, opened 2020-11-28
Do not reconcile shoots multiple times during their maintenance time windows

rfranzke:

/area scalability
/kind enhancement
/priority critical

Today, the shoot controller has various problems that lead to scalability issues in very large landscapes:
1. It is reconciling shoots multiple times during their maintenance time windows (if the gardenlet's component config sets "reconcile in maintenance only" to `true`) based on the sync period. This also happens even if the reconciliation has already happened before for the current time window.
1. When the gardenlet restarts then all shoots whose maintenance time window is met will be enqueued and reconciled immediately. If many shoots have the same maintenance time window then this will produce a lot of load.

This PR adapts the behaviour in the following ways:
1. When the gardenlet has already reconciled a shoot in its maintenance time window then it doesn't do it again. Instead, it computes a random duration for the next time window and requeues the shoot. Already reconciled shoots are those whose last reconciliation was less then `24h` ago.
1. When the gardenlet is (re)started then it does no longer reconcile all shoots whose maintenance time windows are met immediately. Instead, it computes a random time for the current time window and requeues the shoot ("jittering"). This will have the effect that not all shoots are getting reconciled at the same time right after startup.

Both changes will lead to a lot of less load on the seeds, the garden cluster (apiservers), and to less network I/O.

/cc @vpnachev 

The implementation introduces a so called "reconciliationDueTracker" in the shoot controller. Its purpose is to distinguish the two cases:
1. gardenlet starts and checks if shoot whose maintenance time window is met must be reconciled
1. gardenlet runs and checks if a shoot whose maintenance time window is met must be reconciled but was already delayed/jittered before

A better implementation is probably to handle this situation in the `shootAdd()` function, i.e., right in the event handler for `CREATE SHOOT` events. However, this would require some duplication of the logic that is already part of `reconcileShoot()`, and it even might have other side effects. The logic is quite complex at the moment, so I was reluctant do follow this approach (and risking to break something). I guess the `reconciliationDueTracker` approach is easier to understand/review for now, it's more lightweight and has less duplication/complexity. Nevertheless, we should clean up this shoot control logic when possible so that the whole flow/logic and different cases become better understandable and readable.

**Release note**:

```feature operator
The shoot controller inside the gardenlet has been adapted to cater with large Gardener landscapes:
1. When the gardenlet has already reconciled a shoot cluster during its maintenance time window then it doesn't reconcile it again. Instead, it computes a random duration for the next time window and requeues the shoot. Already reconciled shoots are those whose last reconciliation was less then `24h` ago.
1. When the gardenlet is (re)started then it does no longer reconcile all shoots immediately whose maintenance time windows are met. Instead, it computes a random time for the current time window and requeues the shoot ("jittering", i.e., spreading the load). This will have the effect that not all shoots are getting reconciled at the same time right after startup.
```

### DIFF
--- a/pkg/apis/core/types_shoot.go
+++ b/pkg/apis/core/types_shoot.go
@@ -695,6 +695,13 @@ const (
 // Maintenance relevant types                                                                   //
 //////////////////////////////////////////////////////////////////////////////////////////////////
 
+const (
+	// MaintenanceTimeWindowDurationMinimum is the minimum duration for a maintenance time window.
+	MaintenanceTimeWindowDurationMinimum = 30 * time.Minute
+	// MaintenanceTimeWindowDurationMaximum is the maximum duration for a maintenance time window.
+	MaintenanceTimeWindowDurationMaximum = 6 * time.Hour
+)
+
 // Maintenance contains information about the time window for maintenance operations and which
 // operations should be performed.
 type Maintenance struct {
--- a/pkg/apis/core/v1alpha1/types_shoot.go
+++ b/pkg/apis/core/v1alpha1/types_shoot.go
@@ -870,6 +870,13 @@ const (
 // Maintenance relevant types                                                                   //
 //////////////////////////////////////////////////////////////////////////////////////////////////
 
+const (
+	// MaintenanceTimeWindowDurationMinimum is the minimum duration for a maintenance time window.
+	MaintenanceTimeWindowDurationMinimum = 30 * time.Minute
+	// MaintenanceTimeWindowDurationMaximum is the maximum duration for a maintenance time window.
+	MaintenanceTimeWindowDurationMaximum = 6 * time.Hour
+)
+
 // Maintenance contains information about the time window for maintenance operations and which
 // operations should be performed.
 type Maintenance struct {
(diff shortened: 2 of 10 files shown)


ialidzhikov (dismissed):
/lgtm
/needs second-opinion

vpnachev (dismissed):
/lgtm

Out of scope of this PR, I was thinking about to set the next reconciliation time in the shoot status after each successful reconciliation. Then the controller can use this as reference when it has to reconcile again. If it was down during this time, it should schedule new reconciliation during the maintenance time window, which can be the current or the next window depending what is the current time.

rfranzke:
> Out of scope of this PR, I was thinking about to set the next reconciliation time in the shoot status after each successful reconciliation. Then the controller can use this as reference when it has to reconcile again. If it was down during this time, it should schedule new reconciliation during the maintenance time window, which can be the current or the next window depending what is the current time.

Yes, good idea. Let's discuss those options when we tackle #3160

rfranzke:
I've pushed a small improvement based on a suggestion from @timuthy - thanks again for this nice catch. PTAL.

timuthy (approved):
Thanks @rfranzke 👍
/lgtm

PR #3243, opened 2020-11-30
Add custom kube-scheduler for control plane pods.

### LINKED ISSUE #2873: Bin-Packing for Control Planes
@mvladev once rolled out a change for shoot clusters that switched the scheduler to bin-packing, which significantly reduced our shooted seed control plane costs. Since the change was affecting all shoots and some workload fails to come with proper requests/limits (in contrast to Gardener shoot control planes), we saw issues with such workloads and had to revert the change again. @mvladev suggested to take another shot at the subject matter and find a solution that switches Gardener shooted seeds to bin-packing to drive down control plane costs.

**Why is this needed**:
As we have seen in the past when we rolled out this change for the first time, it helps driving costs down.

mvladev:

/area control-plane
/area cost
/area scalability
/kind enhancement
/priority normal

This PR introduces a new feature gate `SeedKubeScheduler` that adds a custom `kube-scheduler` in the seed clusters.

It weights `Nodes` with most resource usage higher, resulting in better packing of `Nodes`:

```yaml
apiVersion: kubescheduler.config.k8s.io/v1alpha2
kind: KubeSchedulerConfiguration
leaderElection:
  leaderElect: true
  leaseDuration: 15s
  renewDeadline: 10s
  resourceLock: leases
  resourceName: kube-scheduler
  resourceNamespace: gardener-kube-scheduler
  retryPeriod: 2s
profiles:
- schedulerName: gardener-shoot-controlplane-scheduler
  plugins:
    score:
      disabled:
      - name: NodeResourcesLeastAllocated
      - name: NodeResourcesBalancedAllocation
      enabled:
      - name: NodeResourcesMostAllocated
  ```

Fixes #2873 

The scheduler is with hard-coded configuration for now, but it can be made configurable in another PR.

**Release note**:

```feature operator
New `SeedKubeScheduler` feature gate is added to gardenlet. When enabled, it deploys a custom `kube-scheduler` in `gardener-kube-scheduler` namespace of Seed clusters with Kubernetes version `1.17` or greater. The scheduler assigns Shoot control plane Pods to Nodes with higher resource utilization, resulting in better bin-packing of control planes.
```


docs/deployment/feature_gates.md
```diff
@@ -69,5 +70,6 @@ A *General Availability* (GA) feature is also referred to as a *stable* feature.
...
 * `APIServerSNI` enables only one LoadBalancer to be used for every Shoot cluster API server in a Seed. Enable this feature when `ManagedIstio` is enabled or Istio is manually deployed in Seed cluster. See [GEP-8](../proposals/08-shoot-apiserver-via-sni.md) for more details.
-* `MountHostCADirectories` enables mounting common CA certificate directories in the Shoot API server pod that might be required for webhooks or OIDC. 
+* `MountHostCADirectories` enables mounting common CA certificate directories in the Shoot API server pod that might be required for webhooks or OIDC.
 * `SeedChange` enables updating the `spec.seedName` field during shoot validation from a non-empty value in order to trigger shoot control plane migration.
+* `SeedKubeScheduler` adds custom `kube-scheduler` in `gardener-kube-scheduler` namespace. It schedules [pods with scheduler name](../concepts/seed-admission-controller.md#mutating-webhooks) `gardener-kube-scheduler` on Nodes with higher resource utilization. It requires Seed cluster with kubernetes version `1.17` or higher.
```
rfranzke: Is there a good reason for not putting it into the `garden` namespace like all the other system components we deploy?

pkg/operation/seed/seed.go
```diff
@@ -673,6 +678,82 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+
+	if gardenletfeatures.FeatureGate.Enabled(features.SeedKubeScheduler) {
+		if err := scheduler.Deploy(ctx); err != nil {
+			return err
+		}
+	} else {
+		if err := scheduler.Destroy(ctx); err != nil {
+			return err
+		}
+	}
```
rfranzke: Can you move this into https://github.com/gardener/gardener/blob/master/pkg/operation/seed/seed.go#L801-L827?

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,438 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke, line 1: We already have a package for the kube-scheduler, https://github.com/gardener/gardener/tree/master/pkg/operation/botanist/controlplane/kubescheduler. Can't we rather extend it to work for both use-cases?

docs/deployment/feature_gates.md
mvladev: Putting everything in one namespace offers a very easy way to clean resources when disabling the feature. The extensions are already in a separate namespaces already.

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
mvladev: There are differences - RBAC, serviceaccounts etc. This was my first idea, but then the code became very hard to test - multiple branches and code paths.

docs/deployment/feature_gates.md
amshuman-kr: Why not use extensions to deploy it?

pkg/operation/seed/seed.go
mvladev: Done

docs/deployment/feature_gates.md
mvladev: I would need a new controller as the version of the `kube-scheduler` must match the Seed API server version.

pkg/features/features.go
```diff
@@ -83,4 +83,9 @@ const (
 	// owner: @stoyanr
 	// alpha: v1.12.0
 	SeedChange featuregate.Feature = "SeedChange"
+
+	// SeedKubeScheduler adds an additional kube-scheduler in seed clusters where the feature is enabled.
+	// owner: @mvladev
+	// alpha: v1.14.0
```
ialidzhikov:
```suggestion
	// alpha: v1.15.0
```

pkg/operation/seed/scheduler/gardener_kube_scheduler_test.go
```diff
@@ -0,0 +1,443 @@
+package scheduler_test
```
ialidzhikov, line 15: License header missing.

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,442 @@
...
+	return s, nil
+}
+
+type kubeScheduler struct {
+	client        client.Client
+	namespace     string
+	image         string
+	codec         serializer.CodecFactory
+	config        *kubeschedulerconfigv1alpha2.KubeSchedulerConfiguration
+	webhookClient *admissionregistrationv1beta1.WebhookClientConfig
```
ialidzhikov:
```suggestion
	webhookClientConfig *admissionregistrationv1beta1.WebhookClientConfig
```

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,442 @@
...
+		"/usr/local/bin/kube-scheduler",
+		fmt.Sprintf("--config=%s/%s", volumeMountPathConfig, dataKeyComponentConfig),
+		fmt.Sprintf("--secure-port=%d", port),
+		"--port=0",
+		"--v=2",
+	}
+}
+
+// WriteConfigFile writes the config into the given file name as YAML.
+func (k *kubeScheduler) componentConfig(cfg *kubeschedulerconfigv1alpha2.KubeSchedulerConfiguration) (string, string, error) {
```
ialidzhikov:
```suggestion
// componentConfig ...
```

pkg/operation/seed/scheduler/gardener_kube_scheduler_test.go
```diff
@@ -0,0 +1,443 @@
...
+		image    = "foo:v1"
+	)
+
+	var (
+		ctx                                      context.Context
+		c                                        client.Client
+		sched                                    component.DeployWaiter
+		version                                  *semver.Version
+		schedulerConfig, expectedSchedulerConfig *kubeschedulerconfigv1alpha2.KubeSchedulerConfiguration
+		webhookClient, expectedWebhookClient     *admissionregistrationv1beta1.WebhookClientConfig
```
ialidzhikov:
```suggestion
		webhookClientConfig, expectedWebhookClientConfig     *admissionregistrationv1beta1.WebhookClientConfig
```

ialidzhikov (dismissed):
I enabled the feature gate locally against v1.17.14 Seed but the schedulers were in `CrashLoopBackOff`:

```
$ k -n gardener-kube-scheduler get po
NAME                                      READY   STATUS             RESTARTS   AGE
gardener-kube-scheduler-f96bdd6f8-dspqq   0/1     CrashLoopBackOff   4          2m31s
gardener-kube-scheduler-f96bdd6f8-gdlhk   0/1     CrashLoopBackOff   4          2m31s

$ k -n gardener-kube-scheduler logs gardener-kube-scheduler-f96bdd6f8-dspqq --tail=1 -p
no kind "KubeSchedulerConfiguration" is registered for version "kubescheduler.config.k8s.io/v1alpha2" in scheme "k8s.io/kubernetes/pkg/scheduler/apis/config/scheme/scheme.go:30"
```

mvladev:
@ialidzhikov I've greatly redesigned the PR - now every supported Kubernetes version on the Seed (1.18 and 1.19 for now) have their scheduler API in that version. This was done because different scheduler versions support and have different APIs:

- `1.17` has `v1alpha1`
- `1.18` has `v1alpha1` and `v1alpha2`
- `1.19` has `v1beta1`

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,399 @@
...
+// New creates a new instance of DeployWaiter for the kube-scheduler.
+// It requires Seed cluster with version 1.18 or 1.19.
+func New(
+	client client.Client,
+	namespace string,
+	image string,
+	seedVersion *semver.Version,
+	config configurator.Configurator,
+	webhookClientConfig *admissionregistrationv1beta1.WebhookClientConfig,
+) (KubeScheduler, error) {
```
rfranzke:
```suggestion
) (component.DeployWaiter, error) {
```

There is no need for a wrapping interface

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,399 @@
...
+	volumeMountPathConfig                 = "/var/lib/kube-scheduler-config"
+)
+
+// KubeScheduler contains functions for a kube-scheduler deployer.
+type KubeScheduler interface {
+	component.DeployWaiter
+}
+
+// New creates a new instance of DeployWaiter for the kube-scheduler.
+// It requires Seed cluster with version 1.18 or 1.19.
```
rfranzke, line 63: Why? What about older versions? See https://github.com/gardener/gardener/blob/master/docs/usage/supported_k8s_versions.md#seed-cluster-versions

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,399 @@
...
+
+	if err := k.client.Delete(ctx, k.emptyVolumeSchedulerClusterRoleBinding()); client.IgnoreNotFound(err) != nil {
+		return err
+	}
+
+	if err := k.client.Delete(ctx, k.emptyRoleBinding()); client.IgnoreNotFound(err) != nil {
+		return err
+	}
+
+	return nil
```
rfranzke:
```suggestion
	return kutil.DeleteObjects(...)
```

pkg/operation/seed/seed.go
```diff
@@ -803,31 +804,28 @@ func bootstrapComponents(c kubernetes.Interface, namespace string, imageVector i
 		return nil, err
 	}
 
-	// cluster-autoscaler
-	components = append(components, clusterautoscaler.NewBootstrapper(c.Client(), namespace))
-
 	// etcd
 	var (
 		etcdImage                string
+		schedulerImage           string
```
rfranzke: This "block" was used for etcd only. Now we have scheduler and gsac artefacts here. Can you please separate this again for a better code overview?

pkg/operation/seed/seed.go
```diff
@@ -803,31 +804,28 @@ func bootstrapComponents(c kubernetes.Interface, namespace string, imageVector i
...
+		if err != nil {
+			return nil, err
 		}
-	}
-	components = append(components, etcd.NewBootstrapper(c.Client(), namespace, etcdImage, kubernetesVersion, etcdImageVectorOverwrite))
 
-	// gardener-seed-admission-controller
-	var gsacImage imagevector.Image
-	if imageVector != nil {
+		schedulerImage = image.String()
```
rfranzke: See comment above - please move this into a separate "block"

vendor/modules.txt
```diff
@@ -1274,6 +1277,7 @@ sigs.k8s.io/yaml
 # k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.18.10
 # k8s.io/code-generator => k8s.io/code-generator v0.18.10
 # k8s.io/component-base => k8s.io/component-base v0.18.10
+# k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.18.10
```
rfranzke: Why do we have this? You have copied all the config types anyway, right?

rfranzke (changes requested):
I've put a few comments inline. What I don't like too much is that the `pkg/operation/seed/scheduler` component is not using the `ManagedResource` approach like other refactored components in the seed. Hence, we don't have any health checks or any convenient reconciliation of resources when we change them (e.g., if we add a new `ConfigMap` tomorrow and delete an old `Secret` then we have to build this logic explicitly while the GRM does this automatically for us).
I would strongly vote for using the GRM approach so that we are consistent in the implementations of our refactored/new components.

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
mvladev: Some feature gates - e.g. Istio require that you have `1.16` or newer. The `kube-scheduler` prior to 1.18 had lots of the old legacy APIs in.

rfranzke: OK, agreed as it's an alpha feature gate anyways. Thanks.

mvladev:
@rfranzke I've reworked the PR to use the managed resources. Have another look.

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,368 @@
...
+	kubeSchedulerClusterRoleName          = "system:kube-scheduler"
+	volumeSchedulerClusterRoleName        = "system:volume-scheduler"
+	webhookName                           = "kube-scheduler.scheduling.gardener.cloud"
+	volumeMountPathConfig                 = "/var/lib/kube-scheduler-config"
+)
+
+// KubeScheduler contains functions for a kube-scheduler deployer.
+type KubeScheduler interface {
+	component.DeployWaiter
+}
```
rfranzke:
```suggestion
```

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,368 @@
...
+										corev1.ResourceMemory: resource.MustParse("64Mi"),
+									},
+									Limits: corev1.ResourceList{
+										corev1.ResourceCPU:    resource.MustParse("400m"),
+										corev1.ResourceMemory: resource.MustParse("512Mi"),
+									},
+								},
+								VolumeMounts: []corev1.VolumeMount{
+									{
+										Name:      "config",
```
rfranzke: You can have a variable for this and use it here and in line 242

pkg/operation/seed/scheduler/gardener_kube_scheduler_bootstrap.go
```diff
@@ -0,0 +1,137 @@
...
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	"k8s.io/utils/pointer"
+)
+
+// Bootstrap is used to bootstrap gardener-kube-scheduler in Seed clusters.
+func Bootstrap(
+	c kubernetes.Interface,
+	image string,
+	seedVersion *semver.Version,
+) (component.DeployWaiter, error) {
```
rfranzke:
```suggestion
) (
    component.DeployWaiter,
    error,
) {
```

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,368 @@
...
+// New creates a new instance of DeployWaiter for the kube-scheduler.
+// It requires Seed cluster with version 1.18 or 1.19.
+func New(
+	client client.Client,
+	namespace string,
+	image string,
+	seedVersion *semver.Version,
+	config configurator.Configurator,
+	webhookClientConfig *admissionregistrationv1beta1.WebhookClientConfig,
+) (component.DeployWaiter, error) {
```
rfranzke:
```suggestion
) (
    component.DeployWaiter,
    error,
) {
```

pkg/operation/seed/scheduler/gardener_kube_scheduler_bootstrap.go
```diff
@@ -0,0 +1,137 @@
...
+
+	scheduler, err := New(
+		c.DirectClient(),
+		namespace,
+		image,
+		seedVersion,
+		config,
+		&admissionregistrationv1beta1.WebhookClientConfig{
+			Service: &admissionregistrationv1beta1.ServiceReference{
+				Name:      "gardener-seed-admission-controller",
```
rfranzke: Export https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/seedsystemcomponents/seedadmission/seedadmission.go#L48 and use it here

pkg/operation/seed/scheduler/gardener_kube_scheduler_bootstrap.go
```diff
@@ -0,0 +1,137 @@
...
+				Namespace: v1beta1constants.GardenNamespace,
```
rfranzke: You should use the same namespace that you use here:
https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/seedsystemcomponents/seedadmission/seedadmission.go#L61

pkg/operation/seed/scheduler/gardener_kube_scheduler_bootstrap.go
```diff
@@ -0,0 +1,137 @@
...
+lDYl/nb1GtVvvkQD49VnV7mDnl6mG3LCMNCNH6WlZNv3AgMBAAGjQjBAMA4GA1Ud
+DwEB/wQEAwIBBjAPBgNVHRMBAf8EBTADAQH/MB0GA1UdDgQWBBSFA3LvJM21d8qs
+ZVVCe6RrTT9wiTANBgkqhkiG9w0BAQsFAAOCAQEAns/EJ3yKsjtISoteQ714r2Um
+BMPyUYTTdRHD8LZMd3RykvsacF2l2y88Nz6wJcAuoUj1h8aBDP5oXVtNmFT9jybS
+TXrRvWi+aeYdb556nEA5/a94e+cb+Ck3qy/1xgQoS457AZQODisDZNBYWkAFs2Lc
+ucpcAtXJtIthVm7FjoAHYcsrY04yAiYEJLD02TjUDXg4iGOGMkVHdmhawBDBF3Aj
+esfcqFwji6JyAKFRACPowykQONFwUSom89uYESSCJFvNCk9MJmjJ2PzDUt6CypR4
+epFdd1fXLwuwn7fvPMmJqD3HtLalX1AZmPk+BI8ezfAiVcVqnTJQMXlYPpYe9A==
+-----END CERTIFICATE-----`),
+		},
```
rfranzke: Export https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/seedsystemcomponents/seedadmission/seedadmission.go#L400 and use it here

pkg/operation/seed/scheduler/gardener_kube_scheduler_test.go
```diff
@@ -0,0 +1,440 @@
...
+				JustBeforeEach(func() {
+					s, err := New(c, deployNS, image, version, &dummyConfigurator{}, webhookClientConfig)
+					Expect(err).NotTo(HaveOccurred(), "New succeeds")
+
+					sched = s
+				})
+
+				Context("Deploy", func() {
+					JustBeforeEach(func() {
+						Expect(sched.Deploy(ctx)).NotTo(HaveOccurred(), "deploy succeeds")
```
rfranzke:
```suggestion
						Expect(sched.Deploy(ctx)).To(Suceed(), "deploy succeeds")
```

you could use the same for other occurrences

pkg/operation/seed/scheduler/gardener_kube_scheduler_bootstrap.go
```diff
@@ -0,0 +1,137 @@
...
+		err    error
+	)
+	const namespace = "gardener-kube-scheduler"
+
+	if seedVersion == nil {
+		return nil, errors.New("seedVersion is required")
+	}
+
+	switch seedVersion.Minor() {
+	case int64(18):
```
rfranzke: What about having some `github.com/Masterminds/semver` constraints and using them here (e.g., like in https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go#L465-L503)?

pkg/operation/seed/scheduler/gardener_kube_scheduler_test.go
```diff
@@ -0,0 +1,440 @@
...
+
+					It("namespace is created", func() {
+						actual := &corev1.Namespace{}
+						expected := &corev1.Namespace{
+							ObjectMeta: metav1.ObjectMeta{
+								Name:   deployNS,
+								Labels: expectedLabels,
+							},
+						}
+						Expect(c.Get(ctx, types.NamespacedName{Name: deployNS}, actual)).NotTo(HaveOccurred(), "can get namespace")
```
rfranzke: I like it, but, unfortunately, I don't get it: How does this test work? In `Deploy` you only create the `ManagedResource` with the fake client. Who creates the resources managed by the `ManagedResource` so that you can `Get()` them here with this fake client?

rfranzke (changes requested):
Very nice @mvladev, I like it a lot. I only have nit comments, please check :)

pkg/operation/seed/scheduler/gardener_kube_scheduler_test.go
mvladev: I don't get "real" resources just get the resources stored in the secret for that managed resource and then decode them into the correct K8S Object.

rfranzke: Ah, I see it now (further below), thanks, makes perfect sense and looks great.

mvladev: The namespace is created manually, because `ManagedResource` cannot be created in not-existing namespace.

pkg/operation/seed/scheduler/types.go
```diff
@@ -14,6 +14,31 @@
...
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+)
+
+const (
+	// GardenerShootControlPlaneSchedulerName is the name of the scheduler used for
+	// shoot control plane pods.
+	GardenerShootControlPlaneSchedulerName = "gardener-shoot-controlplane-scheduler"
+	// GardenerShootControlPlaneSchedulerWebhookPath is the path of the webhook server
+	// that sets the "gardener-shoot-controlplane-scheduler" schedulerName for Pods.
+	GardenerShootControlPlaneSchedulerWebhookPath = "/webhooks/default-pod-scheduler-name/" + GardenerShootControlPlaneSchedulerName
```
rfranzke: Can you have a constant with the path somwhere where in `pkg/seedadmission` and use it here and [here](https://github.com/gardener/gardener/blob/f8f07b321caa6c526d367947eaddf47ec5744c5f/cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go#L155)?

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,367 @@
...
+
+	if config == nil {
+		return nil, errors.New("config is required")
+	}
+
+	if webhookClientConfig == nil {
+		return nil, errors.New("webhookClient is required")
+	}
+
+	supportedVersions, err := semver.NewConstraint("1.18.x - 1.19.x")
```
rfranzke: Would 1.20 also work? If not (yet), can you add it (we already have added support for 1.20 in #3296)? If yes then please adapt the constraint(s) accordingly.

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,367 @@
...
+
+	resources, err := registry.AddAllAndSerialize(
+		kubeSchedulerClusterRoleBinding,
+		volumeSchedulerClusterRoleBinding,
+		roleBinding,
+		serviceAccount,
+		configMap,
+		deployment,
+		webhook,
+	)
```
rfranzke: Can we also have a VPA for the scheduler?

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,367 @@
...
+
+func getLabels() map[string]string {
+	return map[string]string{
+		"app":  "kubernetes",
+		"role": "scheduler",
+	}
+}
+
+func (k *kubeScheduler) Destroy(ctx context.Context) error {
+	return common.DeleteManagedResourceForSeed(ctx, k.client, "gardener-kube-scheduler", k.namespace)
```
rfranzke: You'd have to delete the namespace here if you created it manually yourself, right?

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,367 @@
...
+				TimeoutSeconds:          &timeout,
+				AdmissionReviewVersions: []string{admissionregistrationv1beta1.SchemeGroupVersion.Version},
+				ReinvocationPolicy:      &revocationPolicy,
+			}},
+		}
+		registry = managedresources.NewRegistry(kubernetes.SeedScheme, kubernetes.SeedCodec, kubernetes.SeedSerializer)
+	)
+
+	if _, err := controllerutil.CreateOrUpdate(ctx, k.client, namespace, func() error {
+		namespace.Labels = getLabels()
```
rfranzke: Hm, if this is an existing namespace then you'd overwrite all labels here - probably it's better to merge them instead?

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,367 @@
...
+	defer cancel()
+
+	return managedresources.WaitUntilManagedResourceHealthy(timeoutCtx, k.client, k.namespace, "gardener-kube-scheduler")
+}
+
+func (k *kubeScheduler) WaitCleanup(ctx context.Context) error {
+	timeoutCtx, cancel := context.WithTimeout(ctx, time.Minute*2)
+	defer cancel()
+
+	return managedresources.WaitUntilManagedResourceDeleted(timeoutCtx, k.client, k.namespace, "gardener-kube-scheduler")
```
rfranzke: Also wait for namespace being deleted?

rfranzke (changes requested):
Thanks @mvladev for addressing my suggestions.
I went over it again and found a few more minor nits 👀

pkg/operation/seed/scheduler/types.go
mvladev: I would rather keep the constant here as its about the scheduler.

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
mvladev: I'll just wait for the namespace to be deleted as the managed resource is in it.

pkg/operation/seed/scheduler/types.go
rfranzke: Hm, but the webhook path is defined by the webhook server which is in `cmd/gardener-seed-admission-controller/app`, so I the constant definitely belongs there (or rather in `pkg/seedadmission`). The gardener-kube-scheduler just uses it, but it does not define it.

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,395 @@
...
+
+func (k *kubeScheduler) WaitCleanup(ctx context.Context) error {
+	return retry.UntilTimeout(ctx, time.Second*2, time.Minute*2, func(ctx context.Context) (bool, error) {
+		if err := k.client.Get(ctx, client.ObjectKey{Name: k.namespace}, &corev1.Namespace{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				return retry.Ok()
+			}
+			return retry.MinorError(err)
+		}
+		return retry.MinorError(errors.Errorf("Namespace %s is not deleted yet", k.namespace))
```
rfranzke: You can use https://github.com/gardener/gardener/blob/4c6153dceae2e461607c9a38afca88277387dda0/pkg/utils/kubernetes/kubernetes.go#L128 instead :)

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,395 @@
...
+	})
+}
```
rfranzke: OK, deleting the namespace and waiting for it to be deleted is fine, but it's probably "dangerous" if somebody uses the `garden` namespace for the kube-scheduler. Does it make sense to have some validation for it?

mvladev:
I've updated the PR

rfranzke (commented):
Played with it, works perfectly, thanks!
As discussed via Slack: Let's change the replicas to either `1`, or (if you want to stay with `2`) then let's add a PDB+anti-affinity so that the desired effect is ensured.

mvladev:
Added PDB in 87a1ccb

rfranzke (dismissed):
/lgtm
/squash

Thanks @mvladev !

rfranzke:
@ialidzhikov your review from last month is blocking the merge, can you please either review again or dismiss?

ialidzhikov:
> @ialidzhikov your review from last month is blocking the merge, can you please either review again or dismiss?

I will check later today.

docs/deployment/feature_gates.md
timebertt: > Putting everything in one namespace offers a very easy way to clean resources when disabling the feature. The extensions are already in a separate namespaces already.

Isn't this also the case when using a ManagedResource for that?
TBH, I would prefer to have this consistent with other system components we deploy for example `etcd-druid` and `gardener-seed-admission-controller`. So deploying it via a MR into the `garden` namespace.
Sorry, if I missed something here. @rfranzke you resolved this, did you change your mind?

rfranzke: Yes, I don't care too much about it and I'm okay with the solution as it is now. We also already deploy the istio components and the extensions into separate namespaces.

pkg/operation/seed/scheduler/gardener_kube_scheduler_bootstrap.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/gardener/gardener/pkg/operation/botanist/seedsystemcomponents/seedadmission"
+	"github.com/gardener/gardener/pkg/operation/seed/scheduler/configurator"
+	schedulerconfigv18 "github.com/gardener/gardener/pkg/operation/seed/scheduler/v18"
+	schedulerconfigv19 "github.com/gardener/gardener/pkg/operation/seed/scheduler/v19"
+	schedulerconfigv20 "github.com/gardener/gardener/pkg/operation/seed/scheduler/v20"
+	seedadmissionpkg "github.com/gardener/gardener/pkg/seedadmission"
+	"github.com/gardener/gardener/pkg/utils/imagevector"
+	schedulerconfigv18v1alpha2 "k8s.io/kube-scheduler/v18/v1alpha2"
+	schedulerconfigv19v1beta1 "k8s.io/kube-scheduler/v19/v1beta1"
+	schedulerconfigv20v1beta1 "k8s.io/kube-scheduler/v20/v1beta1"
```
timuthy, line 30: Can you explain why you chose to copy `v1beta1`, as well? IIUC, we could directly refer to `https://github.com/kubernetes/kube-scheduler`, without copying the code.

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -0,0 +1,421 @@
...
+	if len(image) == 0 {
+		return nil, errors.New("image is required")
+	}
+
+	if config == nil {
+		return nil, errors.New("config is required")
+	}
+
+	if webhookClientConfig == nil {
+		return nil, errors.New("webhookClient is required")
```
ialidzhikov:
```suggestion
		return nil, errors.New("webhookClientConfig is required")
```

third_party/kube-scheduler/config/v1/types.go
```diff
@@ -0,0 +1,242 @@
+/*
```
timuthy, line 1: Where is this needed?

pkg/operation/seed/scheduler/gardener_kube_scheduler_bootstrap.go
```diff
@@ -0,0 +1,155 @@
...
+							{Name: "NodeResourcesBalancedAllocation"},
+						},
+						Enabled: []schedulerconfigv19v1beta1.Plugin{
+							{Name: "NodeResourcesMostAllocated"},
+						},
+					},
+				},
+			}},
+		})
+	case versionConstraintEqual1_20.Check(seedVersion):
```
timuthy: What's the reason not doing it for `>= 1.20`? This would potentially reduce maintenance effort in the future and if the scheduler didn't support the configuration version, we'd prematurely notice this failure during seed bootstrapping.

pkg/operation/seed/scheduler/gardener_kube_scheduler_bootstrap.go
```diff
@@ -0,0 +1,155 @@
...
+	if c == nil {
+		return nil, errors.New("client is required")
+	}
+
+	if image == nil {
+		return nil, errors.New("image is required")
+	}
+
+	if len(seedAdmissionControllerNamespace) == 0 {
+		return nil, errors.New("image is required")
```
ialidzhikov:
```suggestion
		return nil, errors.New("seedAdmissionControllerNamespace is required")
```

pkg/operation/seed/scheduler/types.go
```diff
@@ -14,6 +14,26 @@
...
-// GardenerShootControlPlaneSchedulerName is the name of the scheduler used for
-// shoot control plane pods.
-const GardenerShootControlPlaneSchedulerName = "gardener-shoot-controlplane-scheduler"
+import (
+	"github.com/Masterminds/semver"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+)
+
+var (
+	versionConstraintEqual1_18 *semver.Constraints
```
ialidzhikov:
```suggestion
	versionConstraintEqual118 *semver.Constraints
```

pkg/operation/seed/scheduler/v20/configurator.go
```diff
@@ -0,0 +1,79 @@
...
+	componentbaseconfigv1alpha1 "k8s.io/component-base/config/v1alpha1"
+	"k8s.io/utils/pointer"
+)
+
+type v20Configurator struct {
+	config *schedulerv20v1beta1.KubeSchedulerConfiguration
+	codec  serializer.CodecFactory
+}
+
+// NewConfigurator creates a Configurator for Kubernetes version 1.19.
```
timuthy:
```suggestion
// NewConfigurator creates a Configurator for Kubernetes version 1.20.
```

1.20 and above (see related comment)?

pkg/operation/seed/scheduler/v18/configurator.go
```diff
@@ -0,0 +1,79 @@
...
+	if err := schedulerv18v1alpha2.AddToScheme(scheme); err != nil {
+		return nil, err
+	}
+
+	config.LeaderElection.LeaderElectionConfiguration = componentbaseconfigv1alpha1.LeaderElectionConfiguration{
+		LeaseDuration:     metav1.Duration{Duration: 15 * time.Second},
+		RenewDeadline:     metav1.Duration{Duration: 10 * time.Second},
+		RetryPeriod:       metav1.Duration{Duration: 2 * time.Second},
+		ResourceLock:      "leases",
+		ResourceName:      "kube-scheduler",
```
timuthy:
```suggestion
		ResourceName:      "gardener-kube-scheduler",
```
Nit, to avoid potential name clashes. Please also adjust for other versions, if you follow this suggestion.

timuthy (changes requested):

pkg/operation/seed/scheduler/gardener_kube_scheduler_bootstrap.go
mvladev: - If the Seed cluster is v1.18 only `v1alpha1` and `v1alpha2` are available https://github.com/kubernetes/kube-scheduler/tree/v0.18.14/config
- If the Seed cluster is v1.19 or v1.20 only `v1beta1` is available https://github.com/kubernetes/kube-scheduler/tree/v0.19.0/config

we cannot simply upgrade the dependency to 1.19 or 1.20, because we would lose the `v1alpha1` and `v1alpha2` API for 1.18

pkg/operation/seed/scheduler/gardener_kube_scheduler_bootstrap.go
mvladev: I would probably do this when the API of the kube-scheduler is stable enough. Until `v1`version of the scheduler API is there I would rather test each version beforehand.

timuthy: The problem with the approach at the moment I see is that there is only a "soft" feedback if one misses to add a config for a new K8s version, e.g. `versionConstraintEqual1_21.Check(seedVersion)`.

You'll only notice that somethings wrong:
- during scheduling of CP pods of a shoot with this new version
- when the feature gate is turned on

No validation or integration test covers this at the moment.

mvladev: > The problem with the approach at the moment I see is that there is only a "soft" feedback if one misses to add a config for a new K8s version, e.g. versionConstraintEqual1_21.Check(seedVersion).

In such cases, the feature would be effectively disabled, without any side-effects - the `MutatingWebhookConfiguration` is removed and the control planes would be scheduled as if the scheduler is not there. The problem is that newer versions of the scheduler _might_ optionally require some additional flags or have some incompatible change in them. So it's better to have version for each one of them.

It's way safer to have this feature disabled than have it enabled and not working at all.

> during scheduling of CP pods of a shoot with this new version

The version which is checked is the Seed cluster and not the Shoot cluster.

> No validation or integration test covers this at the moment.

Such test would be added later down the line. It's still alpha state atm.

timuthy: Okay, got it. Thanks for explaining.

timuthy (approved):

go.mod
```diff
@@ -76,6 +79,9 @@ replace (
 	k8s.io/helm => k8s.io/helm v2.13.1+incompatible
 	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.18.10
 	k8s.io/kube-openapi => github.com/gardener/kube-openapi v0.0.0-20200831104310-b5db060350c7
+	k8s.io/kube-scheduler/v18 => ./third_party/kube-scheduler/v18 // update those modules as well when updating kubernetes version
+	k8s.io/kube-scheduler/v19 => ./third_party/kube-scheduler/v19 // update those modules as well when updating kubernetes version
+	k8s.io/kube-scheduler/v20 => ./third_party/kube-scheduler/v20 // update those modules as well when updating kubernetes version
```
vpnachev, line 84: I tried to vendor this change in provider-aws and I would say it is impossible. 
May be I am not educated enough, but I failed to vendor newer version of this repo trying with different `replace`s and `exclude`s - nothing works.

PR #3248, opened 2020-11-30
Garbage collect OutOf* Pods

timebertt:

/area ops-productivity
/kind enhancement
/priority normal

With this PR the shoot garbage collector also deletes failed Pods with the reason `OutOf*` in the Seed namespace and the `kube-system` namespace of the Shoot.

Fixes #

The kubelet sets the status reason in Pods to `OutOf{cpu,memory,...}`, if a Pod is assigned to a node, but the requested resources cannot be acquired / exceed the Node's capacity.
This can happen for example because of manual scheduling, but we have also observed such occurrences after a longer period of API server downtime.

**Release note**:

```noteworthy user
The Shoot garbage collector now also deletes failed Pods with the reason `OutOf*` in the Seed namespace and the `kube-system` namespace of the Shoot.
```

### DIFF
--- a/pkg/operation/botanist/garbage_collection.go
+++ b/pkg/operation/botanist/garbage_collection.go
@@ -67,7 +67,7 @@ func (b *Botanist) deleteStalePods(ctx context.Context, c client.Client, podList
 	var result error
 
 	for _, pod := range podList.Items {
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm

PR #3251, opened 2020-12-01
Re-evaluate shoot health status right after reconciliation

### LINKED ISSUE #3238: Re-evaluate shoot conditions immediately after reconciliation

/area monitoring
/kind enhancement
/priority normal

1. When the shoot reconciliation flow finishes successfully then the `.status.lastOperation.state` is set to `Succeeded`. The same update call should  set the conditions to `Progressing`.
1. The shoot care controller should enqueue shoots immediately after their `.status.lastOperation.state` was set to `Succeeded` (if the `.type` is not `Delete`) in order to re-perform the health checks.

**Why is this needed**:
Today, the shoot conditions are evaluated periodically (e..g, every minute). This means that after a shoot reconciliation it may take up to a minute until they reflect the actual health status. Let's show this health status faster by implementing above improvements.

rfranzke:

/area monitoring
/kind enhancement
/priority normal

The shoot reconciler sets the conditions to `Progressing` after it finished a successful reconciliation, and the care controller starts to re-evaluate the health status after this happened. This helps end-users to better understand whether their cluster is indeed healthy after a reconciliation. Earlier, it could take up to 30s/1m (based on the configured care controller sync period) until the actual status is reflected.

Fixes #3238

**Release note**:

```feature user
The shoot reconciler sets the conditions to `Progressing` after it finished a successful reconciliation, and the care controller starts to re-evaluate the health status after this happened. This helps end-users to better understand whether their cluster is indeed healthy after a reconciliation. Earlier, it could take up to `30s` / `1m` (based on the configured care controller sync period) until the actual status is reflected.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_care_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_care_control.go
@@ -67,6 +67,36 @@ func (c *Controller) shootCareAdd(obj interface{}) {
 	c.shootCareQueue.Add(key)
 }
 
+func (c *Controller) shootCareUpdate(oldObj, newObj interface{}) {
+	key, err := cache.MetaNamespaceKeyFunc(newObj)
+	if err != nil {
+		return
+	}
+
+	oldShoot, ok := oldObj.(*gardencorev1beta1.Shoot)
+	if !ok {
+		return
+	}
+	newShoot, ok := newObj.(*gardencorev1beta1.Shoot)
+	if !ok {
+		return
+	}
+
+	// re-evaluate shoot health status right after a reconciliation operation has finished
+	if shootReconciliationFinishedSuccessful(oldShoot, newShoot) {
+		c.shootCareQueue.Add(key)
+	}
+}
+
+func shootReconciliationFinishedSuccessful(oldShoot, newShoot *gardencorev1beta1.Shoot) bool {
+	return oldShoot.Status.LastOperation != nil &&
+		oldShoot.Status.LastOperation.Type != gardencorev1beta1.LastOperationTypeDelete &&
+		oldShoot.Status.LastOperation.State == gardencorev1beta1.LastOperationStateProcessing &&
+		newShoot.Status.LastOperation != nil &&
+		newShoot.Status.LastOperation.Type != gardencorev1beta1.LastOperationTypeDelete &&
+		newShoot.Status.LastOperation.State == gardencorev1beta1.LastOperationStateSucceeded
+}
+
 func (c *Controller) reconcileShootCareKey(key string) error {
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
--- a/pkg/gardenlet/controller/shoot/shoot_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control.go
@@ -587,18 +587,21 @@ func (c *Controller) updateShootStatusOperationStart(ctx context.Context, g gard
 
 func (c *Controller) updateShootStatusOperationSuccess(ctx context.Context, g gardencore.Interface, shoot *gardencorev1beta1.Shoot, shootSeedNamespace string, seedName *string, operationType gardencorev1beta1.LastOperationType) (*gardencorev1beta1.Shoot, error) {
 	var (
-		now          = metav1.NewTime(time.Now().UTC())
-		description  string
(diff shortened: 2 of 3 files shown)


pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -610,6 +613,25 @@ func (c *Controller) updateShootStatusOperationSuccess(ctx context.Context, g ga
...
+						shoot.Status.Conditions[i].Status = gardencorev1beta1.ConditionProgressing
+					}
+				}
+				for i, constr := range shoot.Status.Constraints {
+					switch constr.Type {
+					case gardencorev1beta1.ShootHibernationPossible,
+						gardencorev1beta1.ShootMaintenancePreconditionsSatisfied:
+						shoot.Status.Constraints[i].Status = gardencorev1beta1.ConditionProgressing
+					}
+				}
```
rfranzke: Should we also change reason/message? WDYT? I wouldn't change the times to keep the evaluation whether the status is `False` or still `Progressing` to the care controller and not duplicate it here, though.

pkg/gardenlet/controller/shoot/shoot_care_control.go
```diff
@@ -67,6 +67,32 @@ func (c *Controller) shootCareAdd(obj interface{}) {
...
+		return
+	}
+
+	// re-evaluate shoot health status right after a reconciliation operation has finished
+	if oldShoot.Status.LastOperation != nil &&
+		oldShoot.Status.LastOperation.Type != gardencorev1beta1.LastOperationTypeDelete &&
+		oldShoot.Status.LastOperation.State == gardencorev1beta1.LastOperationStateProcessing &&
+		newShoot.Status.LastOperation != nil &&
+		newShoot.Status.LastOperation.Type != gardencorev1beta1.LastOperationTypeDelete &&
+		newShoot.Status.LastOperation.State == gardencorev1beta1.LastOperationStateSucceeded {
```
timebertt: Can you extract this condition to a function like `shootOperationFinished` to make it more readable?

timebertt (commented):
Nit only

pkg/gardenlet/controller/shoot/shoot_control.go
timebertt: Hmm, also not too sure about this one. But I would propose to keep reason and message, set the status to `Progressing` and reset at least the `LastUpdateTime`.
This way, the thresholds for progressing conditions will count from the end of the reconciliation on. This makes sense, I guess, because we might change some stuff during reconciliation which might take a minute or so to be healthy again (e.g. loki/prometheus volumes).
WDYT?

timuthy: +1 for what @timebertt suggested

rfranzke: The problem (or rather consequence) with this would be that a condition that was `False` already before/during the reconciliation would now become `Progressing` for some time. Maybe the reconciliation fixes it, however, we don't know and would show a "wrong" status until it transitions to `False` again.

pkg/gardenlet/controller/shoot/shoot_care_control.go
rfranzke: done

pkg/gardenlet/controller/shoot/shoot_control.go
timuthy: Hm, you have a point there. However, isn't that a problem anyway, i.e. we shortly set it to `Progressing` even now?
What if we only changed it to `Progressing` + `LastUpdateTime` update when the status was != `False` before?

rfranzke: > i.e. we shortly set it to Progressing even now?

Hm, no, this is actually the purpose of `Progressing` and desired behaviour, right. But if something is already "broken" then we should probably not set it back to `Progressing`.

> What if we only changed it to Progressing + LastUpdateTime update when the status was != False before?

Nice, I like this. Will do!

timebertt: Agreed, sounds good 👍

timebertt (approved):
Thanks!
/lgtm now

timuthy (approved):
/lgtm

PR #3253, opened 2020-12-02
Add integration test testing the Loki multitenant ID usage

vlvasilev:

/area logging
/kind test
/priority normal

With this PR a integration test is added which test the Loki multitenancy.
The tenants are derived from the grafana-operatops and grafana-users ingresses.
The user logger application is deployed with name similar to the kube-apiserver in order to be processed from the proper fluent-bit rewrite tag filter and its logs to be send as user exposed logs.
The operator and the user logging application are deployed in the shoot(used for the test) control plane.
The logs are retrieved for examination from the Loki inside the shoot control plane as only those Loki instances support multitenancy.
The test should verify that operator can see their logs and the user ones, when users can see only theirs logs.

Fixes #

**Release note**:

```other operator
Add Loki multitenancy integration test. 
```

### DIFF
--- a/test/integration/shoots/logging/loki_tenant.go
+++ b/test/integration/shoots/logging/loki_tenant.go
@@ -1,0 +1,163 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logging
+
+import (
+	"context"
+	"time"
+
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+	"github.com/gardener/gardener/test/framework"
+	"github.com/gardener/gardener/test/framework/resources/templates"
+
+	"github.com/onsi/ginkgo"
+	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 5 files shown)


test/integration/shoots/logging/loki_tenant.go
```diff
@@ -0,0 +1,163 @@
...
+		ginkgo.By("Verify loki received all user logger application logs as an operator")
+		err = WaitUntilLokiReceivesLogs(ctx, 30*time.Second, f, operatorID, f.ShootSeedNamespace(), "pod_name", userLoggerRegex, expectedUserLogs, tenantDeltaLogsCount, f.SeedClient)
+		framework.ExpectNoError(err)
+
+		ginkgo.By("Verify loki received all operator logger application logs")
+		err = WaitUntilLokiReceivesLogs(ctx, 30*time.Second, f, operatorID, f.ShootSeedNamespace(), "pod_name", operatorLoggerRegex, expectedOperatorLogs, tenantDeltaLogsCount, f.SeedClient)
+		framework.ExpectNoError(err)
+
+		ginkgo.By("Verify that loki will not show operator logs to and user tenant")
+		err = WaitUntilLokiReceivesLogs(ctx, 30*time.Second, f, userID, f.ShootSeedNamespace(), "pod_name", operatorLoggerRegex, 0, tenantDeltaLogsCount, f.SeedClient)
```
schrodit: what does this assertion test?
The description is bit unclear

schrodit (commented):
just minor nit.
Rest is lgtm

test/integration/shoots/logging/loki_tenant.go
vlvasilev: Sorry, I don't know how I put `and` instead of `the`

vlvasilev: This will assert that user can't see the operator logs

dguendisch: Probably you wanted to write `end` instead of `and` :)

schrodit (dismissed):

dguendisch:
Works fine, my only question would be: aren't e.g. some 100 log lines sufficient? Currently the test uses 10k but we already have a "load" test with 200k logs.

vlvasilev:
> Works fine, my only question would be: aren't e.g. some 100 log lines sufficient? Currently the test uses 10k but we already have a "load" test with 200k logs.

even 1 log line would be sufficient, I will change it to 100

dguendisch (approved):
Thank you!
/lgtm

Kristian-ZH (approved):
/lgtm

vpnachev (approved):
/lgtm

PR #3254, opened 2020-12-02
[GEP-7] Adds check for seed provider type when changing the seed name of shoot

plkokanov:

/area control-plane-migration
/kind enhancement
/priority normal

This PR forbids changing the `Shoot.Spec.SeedName` field of a `Shoot` cluster if the new `Seed` has different cloud provider from the old one. Currently control planes cannot be migrated between `Seeds` with different providers, although we could implement this in the future.

Part of #1631 

**Release note**:

```noteworthy operator
Forbid control plane migration between `Seeds` with different cloud providers.
```

### DIFF
--- a/plugin/pkg/shoot/validator/admission_test.go
+++ b/plugin/pkg/shoot/validator/admission_test.go
@@ -1945,7 +1945,7 @@ var _ = Describe("validator", func() {
 		})
 
 		Context("backup configuration on seed", func() {
-			It("it should allow new Shoot creation when Seed doesn't have configuration for backup", func() {
+			It("should allow new Shoot creation when Seed doesn't have configuration for backup", func() {
 				oldShoot := shoot.DeepCopy()
 				oldShoot.Spec.SeedName = nil
 				seed.Spec.Backup = nil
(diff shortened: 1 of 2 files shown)


rfranzke (dismissed):
/lgtm

Thank you!

plugin/pkg/shoot/validator/admission_test.go
```diff
@@ -1951,6 +1953,27 @@ var _ = Describe("validator", func() {
 
 				Expect(err).To(HaveOccurred())
 			})
+
+			It("should fail to change Seed name, because cloud provider for new Seed is not equal to cloud provider for old Seed", func() {
+				oldSeedName := fmt.Sprintf("new(%s)", seedName)
```
vpnachev:
```suggestion
				oldSeedName := fmt.Sprintf("old-%s", seedName)
```

vpnachev (commented):
Just a minor suggestion for the name of the old seed to make it more clear. It is strange to me that the old seed is named "new..."
Also, names cannot contain brackets, ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names.

plugin/pkg/shoot/validator/admission_test.go
plkokanov: ops forgot to change this after I modified the variable name.

vpnachev (approved):
/lgtm

PR #3255, opened 2020-12-02
Improve infrastructure reconciliation

timuthy:

/area robustness
/kind enhancement
/priority normal

This PR improves the reconciliation cycles for infrastructure resources. Earlier, when a cluster was created or waking up from hibernation, the infrastructure was reconciled every time the shoot reconciliation flow stopped at a later point, e.g. timeout while waiting for workers to become ready. With this change, the infrastructure resource is not reconciled again if once done successfully.

Additionally, the infrastructure timeout has been increased from `5 minutes` -> `10 minutes`. The infrastructure extension already has proper error handling in place, i.e. it enqueues the resource automatically in case of an error. Re-triggering an infra reconciliation prematurely (because of low timeouts) isn't necessary and doesn't improve such situations. 

Even on the contrary:
1. Extension waits [15 minutes](https://github.com/gardener/gardener-extension-provider-gcp/blob/0706f81a5929278f63741420240c6d2f03fa087b/pkg/internal/terraform.go#L67) for the Terraformer Pod which is much longer than the entire infra timeout from Gardener (5 minutes)
2. Gardener times out and restarts shoot reconcile flow -> adds `gardener.cloud/operation: reconcile` to infra resource **BUT** cannot stop active reconciliation from step 1.
3. Extension enqueues infra while still in reconciliation (step 1.)
4. After successful reconciliation from step 1., the infra is reconciled again

--> 1 "unnecessary" reconciliation

Please note, a shorter timeout used to be reasonable because one might solved an occurring issue by changing the shoot spec accordingly. Since we have [early exists](https://github.com/gardener/gardener/blob/1de02bc33bdf924f7a7739fdd31fc565595a2651/pkg/operation/common/extensions.go#L170) in place for these scenarios, user experience shouldn't suffer from a higher timeout.
 
I decided to completely make use of the `deployInfrastructure` task, now even for creations and wake ups to keep it simple. Alternatively, we could have used the infrastructure `.status` sub-resource to deduce if we need to reconcile or not.

**Release note**:

```other operator
Gardener has improved infrastructure processing procedures in oder to avoid unnecessary reconciliation cycles.
```

### DIFF
--- a/pkg/operation/botanist/extensions/infrastructure/infrastructure.go
+++ b/pkg/operation/botanist/extensions/infrastructure/infrastructure.go
@@ -40,7 +40,7 @@ const (
 	DefaultSevereThreshold = 30 * time.Second
 	// DefaultTimeout is the default timeout and defines how long Gardener should wait
 	// for a successful reconciliation of an infrastructure resource.
-	DefaultTimeout = 5 * time.Minute
+	DefaultTimeout = 10 * time.Minute
 )
 
 // TimeNow returns the current time. Exposed for testing.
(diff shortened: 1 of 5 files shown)


rfranzke (approved):
/lgtm

Very nice PR and well explained description, thanks!

vpnachev (approved):
/lgtm

timebertt (approved):
Very nice!
Thanks for the investigation and fix :)
/lgtm

PR #3264, opened 2020-12-03
Clean up the usage of deprecated role label under charts/seed-bootstrap

ialidzhikov:
/kind cleanup

This PR cleans up most of the usage of the `garden.sapcloud.io/role` label key under `charts/seed-bootstrap`:
- for the monitoring components I switched `garden.sapcloud.io/role` usage to `role` to harmonise the labels in the monitoring chars. Almost all components in the garden ns use the `role` label instead of gardener.cloud/role (garden.sapcloud.io/role). I also believe that some of the usage of garden.sapcloud.io/role=controlplane|monitoring is because of legacy reason as some of the charts were moved from controlplane to the seed bootstrap - for example the dependency-watchdog, the kube-state-metrics
- the vpa components adaptation requires an additional step as the `.spec.selector` of a Deployment is immutable

Ref #1649

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/seed-bootstrap/charts/dependency-watchdog/templates/endpoint-deployment.yaml
+++ b/charts/seed-bootstrap/charts/dependency-watchdog/templates/endpoint-deployment.yaml
@@ -6,7 +6,6 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     role: dependency-watchdog-endpoint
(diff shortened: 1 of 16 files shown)


timebertt (approved):
Thanks!
/lgtm

vpnachev (approved):
/lgtm

PR #3265, opened 2020-12-04
Use vpa-admission-controller runtime namespace only in seed

rfranzke:

/area auto-scaling
/kind bug
/priority normal

This PR fixes a bug that causes the vpa-admission-controller not being able to update its status (inside `Lease` object) when its enabled for shoots.
Usually, the vpa-ac maintains a `Lease` object in the `kube-system` namespace of its cluster. When its ran in the seed then we override this namespace by setting the `NAMESPACE` environment variable to its runtime namespace (i.e., `garden`). This only works because the target cluster is the same like the runtime cluster.
In the shoot case it still runs in the seed, however, in a `shoot--foo--bar` namespace. In this case it cannot talk to the seed in anyway. Still, the `NAMESPACE` environment variable was set to a namespace in a seed which causes failures like this:

```
E1204 06:55:14.581193       1 status_updater.go:56] Status update by vpa-admission-controller-58bd747ff7-jrbs5 failed: namespaces "shoot--foo--bar" not found
```

With this fix we only override it when the vpa-ac is ran in the seed. Otherwise, if enabled for shoots, it will use the `kube-system` namespace for its `Lease` object.

**Release note**:

```bugfix operator
A bug has been fixed that caused the `vpa-admission-controller` to not being able to update its status (inside `Lease` object) when its enabled for shoot clusters.
```

### DIFF
--- a/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
+++ b/charts/seed-bootstrap/charts/vpa/charts/runtime/templates/deployment-admission-controller.yaml
@@ -53,15 +53,16 @@ spec:
         image: {{ index .Values.global.images "vpa-admission-controller" }}
         imagePullPolicy: IfNotPresent
         env:
-        - name: NAMESPACE
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.namespace
 {{- if not .Values.admissionController.enableServiceAccount }}
         - name: KUBERNETES_SERVICE_HOST
           value: kube-apiserver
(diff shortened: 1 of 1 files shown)


timuthy (approved):
/lgtm

timuthy:
Btw. what was the consequence of this bug? Any observed misbehavior?

rfranzke:
@timuthy I can't really tell, I don't know what this lease is used for. I have not seen any obvious misbehaviour, I just saw these erroneous logs by accident.

PR #3275, opened 2020-12-08
Add ownerRef to infra configmap and secret

vpnachev:

/area robustness
/kind enhancement
/priority normal

Ensures that the configmap and secrets, used by terraform, have owner reference to the Infrastructure resource.

Fixes #

**Release note**:

```feature dependency
The `ConfigMaps` and `Secrets` used to store the config and state of terraform now have owner reference to the Infrastructure resource.
```

### DIFF
--- a/extensions/pkg/terraformer/config.go
+++ b/extensions/pkg/terraformer/config.go
@@ -62,6 +62,12 @@ func (t *terraformer) SetDeadlinePod(d time.Duration) Terraformer {
 	return t
 }
 
+// SetOwnerRef configures the resource that will be used as owner of the secrets and configmaps
+func (t *terraformer) SetOwnerRef(owner *metav1.OwnerReference) Terraformer {
+	t.ownerRef = owner
+	return t
+}
+
 // UseV2 configures if it should use flags compatible with terraformer@v2.
 func (t *terraformer) UseV2(v2 bool) Terraformer {
 	t.useV2 = v2
@@ -104,7 +110,7 @@ func (t *terraformer) initializerConfig(ctx context.Context) *InitializerConfig 
 // Initializer to correctly create all the resources as specified in the given InitializerConfig.
 // A default implementation can be found in DefaultInitializer.
 func (t *terraformer) InitializeWith(ctx context.Context, initializer Initializer) Terraformer {
-	if err := initializer.Initialize(ctx, t.initializerConfig(ctx)); err != nil {
+	if err := initializer.Initialize(ctx, t.initializerConfig(ctx), t.ownerRef); err != nil {
 		t.logger.Error(err, "Could not create Terraformer ConfigMaps/Secrets")
 		return t
 	}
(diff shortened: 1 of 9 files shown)


extensions/pkg/terraformer/state.go
```diff
@@ -164,33 +164,62 @@ func sniffJSONStateVersion(stateConfigMap []byte) (uint64, error) {
...
+	if err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, configMap); err != nil {
 		return err
 	}
 
-	return nil
+	oldConfigMap := configMap.DeepCopy()
+	configMap.OwnerReferences = []metav1.OwnerReference{
+		*ownerRef,
+	}
+	return c.Patch(ctx, configMap, client.MergeFrom(oldConfigMap))
```
vpnachev: This code will never be executed for already existing shoots with non-empty state, because `config.InitializeState` is `false` here https://github.com/gardener/gardener/blob/ea37e4c21fae2d255928ce6dbf301e57e1292e68/extensions/pkg/terraformer/config.go#L171-L175 , thus the `Initialize` call of the `CreateState` StateConfigMapInitializer will be skipped.

The alternative options that I see are
- get it like it is now and without ownerRef for existing state configmaps
- move the owner reference ensurer to the https://github.com/gardener/terraformer, although it is not the controller responsible for the creation of the resources. 
- the other option is to add completely independent ownerRef ensurer somewhere here https://github.com/gardener/gardener/blob/ea37e4c21fae2d255928ce6dbf301e57e1292e68/extensions/pkg/controller/infrastructure/reconciler.go#L135-L145 

@gardener/gardener-maintainers any feedback is welcomed on how to proceed with this change.

extensions/pkg/terraformer/config.go
```diff
@@ -112,37 +112,64 @@ func (t *terraformer) InitializeWith(ctx context.Context, initializer Initialize
...
+	)
 }
 
 // CreateOrUpdateTFVarsSecret creates or updates the Terraformer variables Secret with the given tfvars.
-func CreateOrUpdateTFVarsSecret(ctx context.Context, c client.Client, namespace, name string, tfvars []byte) (*corev1.Secret, error) {
-	secret := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: name}}
+func CreateOrUpdateTFVarsSecret(ctx context.Context, c client.Client, namespace, name string, tfvars []byte, ownerRef *metav1.OwnerReference) (*corev1.Secret, error) {
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace, Name: name,
```
rfranzke:
```suggestion
			Namespace: namespace, 
			Name: name,
```

extensions/pkg/terraformer/state.go
rfranzke: Hm, what about doing it in the `execute` function  and remove it after several releases again? https://github.com/gardener/gardener/blob/master/extensions/pkg/terraformer/terraformer.go#L189

rfranzke (dismissed):
/lgtm
/invite @timebertt

extensions/pkg/terraformer/terraformer.go
```diff
@@ -297,6 +304,30 @@ func (t *terraformer) createOrUpdateTerraformerAuth(ctx context.Context) error {
 	return t.createOrUpdateRoleBinding(ctx)
 }
 
+func (t *terraformer) ensureStateHasOwnerRef(ctx context.Context) error {
+	infra := &extensionsv1alpha1.Infrastructure{}
+	if err := t.client.Get(ctx, kutils.Key(t.namespace, t.name), infra); err != nil {
+		return err
+	}
```
timebertt: I think, currently there the terraformer library doesn't know anything about the `Infrastructure` resource.
Instead it is just handed a namespace and name to use for deploying pods, but this doesn't have to be namespace/name of an infra resource. 
I'm wondering, if we should keep it like that.

One way to still add owner refs then, would be to add another function to set the owner reference on the terraformer object (extensions pass the infra object/metadata there) and propagate it from there to the initializers and so on.
If an extension doesn't call the func, no owner refs will be set.

WDYT?
I'm not sure, if there is a real usecase for the library independent from an infra reconciliation, so I don't know we need to consider this...

timebertt (commented):
One concern inline, rest looks good.

extensions/pkg/terraformer/terraformer.go
vpnachev: To be honest, I was looking how to deliver the `ownerRef` via the context `context.WithValue(...)` and then pull the value in the `execute` function. But this seems a bit ugly for something that will be removed after several releases, thus I decided not to do it. 

The other option that I was thinking is almost the same as your idea, to add 
```golang
ownerRef         *metav1.OwnerReference
```
field in https://github.com/gardener/gardener/blob/085d163e1cbd73397aa6b31197714729f8bf7c53/extensions/pkg/terraformer/types.go#L46-L69

and initialize it `New(ForConfig)`, but now I would prefer to use setter as you propose. 
I will try it out and will come back here with the results.

rfranzke: I was thinking about the same, but then refrained from mentioning it because of

> I'm not sure, if there is a real usecase for the library independent from an infra reconciliation

Probably there isn't. In a perfect world I would totally agree to not do it this way (keep the library independent of the `Infrastructure` object), but given that @vpnachev already implemented and verified, and, most importantly, this code is only temporary, I'd vote for keeping it as it is.

vpnachev: I have changed the implementation with ec3693b6c059628d57964f76de8fdb90f68dc48f. 
If you are fine with it, I will squash it.

rfranzke (approved):

timebertt (approved):
Nice, thank you!
I must admit, I like the usage of the library now better, ref https://github.com/gardener/gardener-extension-provider-aws/blob/0629fc51bac0d654011df5cbff40e32444473fb6/pkg/controller/infrastructure/actuator_reconcile.go#L92-L95.
It looks very clean now :)
/lgtm
/squash

PR #3277, opened 2020-12-08
Reduce Garden cluster connections for Gardenlet

timuthy:

/area scalability
/kind enhancement
/priority normal

This PR adds two improvements which reduces the amount of calls from Gardenlet -> Gardener/Kubernetes API Server:
- Retrieving shoot clients from a `clientMap` is now possible w/o requesting the `Shoot` resource every time. This becomes especially important if the Gardenlet cache is disabled (`CachedRuntimeClients: false`) in combination with regular shoot care operations.
- Garden cluster `Namespaces` are not list/wach-ed any more.

Caching the `seedName` and `shoot-namespace` should be safe, also for migration use-cases since the migration logic will [invalidate](https://github.com/gardener/gardener/blob/f120f2c74a503ab4a388aa055cd08d4081eaf5ed/pkg/operation/botanist/controlplane.go#L125) this cached information, once the API Server has removed from the old seed.

**Release note**:

```other operator
Required connections from Gardenlet to the Garden cluster has been reduced which will have positive effects on scalability and costs.
```

### DIFF
--- a/pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
+++ b/pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
@@ -39,6 +39,7 @@ type shootClientMap struct {
 
 // NewShootClientMap creates a new shootClientMap with the given factory and logger.
 func NewShootClientMap(factory *ShootClientSetFactory, logger logrus.FieldLogger) clientmap.ClientMap {
+	factory.clientKeyToSeedInfo = make(map[ShootClientSetKey]seedInfo)
 	return &shootClientMap{
 		ClientMap: NewGenericClientMap(factory, logger),
 	}
@@ -57,6 +58,13 @@ type ShootClientSetFactory struct {
 
 	// Log is a logger for logging entries related to creating Shoot ClientSets.
 	Log logrus.FieldLogger
+
+	clientKeyToSeedInfo map[ShootClientSetKey]seedInfo
+}
+
+type seedInfo struct {
+	namespace string
+	seedName  string
 }
 
 // CalculateClientSetHash calculates a SHA256 hash of the kubeconfig in the 'gardener' secret in the Shoot's Seed namespace.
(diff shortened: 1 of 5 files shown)


pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
```diff
@@ -123,7 +131,42 @@ func (f *ShootClientSetFactory) NewClientSet(ctx context.Context, k clientmap.Cl
 	return clientSet, err
 }
 
+var _ clientmap.Invalidate = &ShootClientSetFactory{}
+
+func (f *ShootClientSetFactory) InvalidateClient(k clientmap.ClientSetKey) error {
```
timebertt, line 137: Can you add a unit test, that ensures that the invalidation logic works (e.g. invalidating a shoot client, changing the cached information and then checking, that the updated information is used)?

timebertt (commented):
Very nice improvement, I like it! :)
One small suggestion from me.

pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
timuthy: I extended the already existing tests. I hope that's okay.

timebertt (dismissed):
Yes, thank you.
/lgtm apart from the linting error...

pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go
rfranzke, line 137: doc string :)

rfranzke (commented):
I'm not too deep into the details of the client map, just a high-level thought: Is this working well (if necessary at all) with control plane migration use-case?

timuthy:
> I'm not too deep into the details of the client map, just a high-level thought: Is this working well (if necessary at all) with control plane migration use-case?

I've gone through the code thoroughly and as the seed changes during migration, the migration logic invalidates the shoot client in the client map (see **Special notes for your reviewer** comment). So we should be safe for this use-case and atm I couldn't think of any other issues. /cc @plkokanov

rfranzke:
Thanks for double-checking @timuthy

plkokanov:
Took a look at the possible paths and I think I found a way that the client could not be invalidated:

During the prepare for migration procedure, the DeleteKubeAPIServer function is only called if there is a kube-apiserver deployment. However, in PrepareKubeAPIServerForMigration we first delete the deployment and after that call DeleteKubeAPIServer to invalidate the client. If the call to invalidate the client returns an error and it cannot be invalidated, the next time the migration procedure  is started the call to PrepareKubeAPIServerForMigration will be skipped and the client will never be invalidated. Check [here](https://github.com/gardener/gardener/blob/f120f2c74a503ab4a388aa055cd08d4081eaf5ed/pkg/operation/botanist/controlplane.go#L377)

Tbh, the way this looks right now seems weird to me. Not sure how it happen but [this line](https://github.com/gardener/gardener/blob/f120f2c74a503ab4a388aa055cd08d4081eaf5ed/pkg/operation/botanist/controlplane.go#L377) seems unnecessary.

plkokanov:
Seems to be ok from CP Migration PoV.

Only trouble would be if someone decided to manually delete the kube-apiserver deployment before triggering migration :)

timuthy:
> Seems to be ok from CP Migration PoV.
> 
> Only trouble would be if someone decided to manually delete the kube-apiserver deployment before triggering migration :)

Thanks for confirming so quickly @plkokanov.

timebertt (approved):

PR #3280, opened 2020-12-09
Delete disabled subcharts when applying a chart

### LINKED ISSUE #218: Unable to disable the remedy-controller for already existing Shoot

/kind bug
/priority normal
/platform azure

remedy-controller to be removed when the Shoot is annotated with `azure.provider.extensions.gardener.cloud/disable-remedy-controller=true` and it is reconciled afterwards

1. Create a Shoot

2. Ensure that the remedy-controller is present as expected

```
$ k -n shoot--dev--bar po -l app=remedy-controller-azure
NAME                                       READY   STATUS    RESTARTS   AGE
remedy-controller-azure-756f555b6b-g4h6s   1/1     Running   0          7h6m
```

3. Annotate the Shoot with `azure.provider.extensions.gardener.cloud/disable-remedy-controller=true` and then trigger a reconcile

```
$ k -n garden-dev annotate shoot bar azure.provider.extensions.gardener.cloud/disable-remedy-controller=true
$ k -n garden-dev annotate shoot bar gardener.cloud/operation=reconcile
```

4. Ensure that the remedy-controller is not removed

```
$ k -n shoot--dev--bar get po -l app=remedy-controller-azure
NAME                                       READY   STATUS    RESTARTS   AGE
remedy-controller-azure-756f555b6b-g4h6s   1/1     Running   0          7h11m
```

- Gardener version (if relevant):
- Extension version: v1.15.2
- Others:

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Enhances the `Chart` `Apply` method do delete disabled subcharts when applying a chart. Needed in order to fix https://github.com/gardener/gardener-extension-provider-azure/issues/218.

Fixes #

Tested this change successfully with https://github.com/gardener/gardener-extension-provider-azure/pull/223.

**Release note**:
```other developer
The `Apply` method of `chart.Interface` now includes one additional `client.Client` argument. This is needed in order to delete the disabled subcharts when applying a chart.
The type of the `SubCharts` field in the `chart.Chart` type has been changed to the new type `chart.SubChart` in order to add a `Condition` field that is only relevant for subcharts.
```

### DIFF
--- a/extensions/pkg/controller/controlplane/genericactuator/actuator.go
+++ b/extensions/pkg/controller/controlplane/genericactuator/actuator.go
@@ -204,7 +204,7 @@ func (a *actuator) reconcileControlPlaneExposure(
 	// Apply control plane exposure chart
 	a.logger.Info("Applying control plane exposure chart", "controlplaneexposure", kutil.ObjectName(cp), "values", values)
 	version := cluster.Shoot.Spec.Kubernetes.Version
-	if err := a.controlPlaneExposureChart.Apply(ctx, a.chartApplier, cp.Namespace, a.imageVector, a.gardenerClientset.Version(), version, values); err != nil {
+	if err := a.controlPlaneExposureChart.Apply(ctx, a.chartApplier, a.client, cp.Namespace, a.imageVector, a.gardenerClientset.Version(), version, values); err != nil {
 		return false, errors.Wrapf(err, "could not apply control plane exposure chart for controlplane '%s'", kutil.ObjectName(cp))
 	}
 
@@ -241,7 +241,7 @@ func (a *actuator) reconcileControlPlane(
 
 		// Apply config chart
 		a.logger.Info("Applying configuration chart", "controlplane", kutil.ObjectName(cp))
-		if err := a.configChart.Apply(ctx, a.chartApplier, cp.Namespace, nil, "", "", values); err != nil {
+		if err := a.configChart.Apply(ctx, a.chartApplier, a.client, cp.Namespace, nil, "", "", values); err != nil {
 			return false, errors.Wrapf(err, "could not apply configuration chart for controlplane '%s'", kutil.ObjectName(cp))
 		}
 	}
@@ -291,7 +291,7 @@ func (a *actuator) reconcileControlPlane(
 	// Apply control plane chart
 	version := cluster.Shoot.Spec.Kubernetes.Version
 	a.logger.Info("Applying control plane chart", "controlplane", kutil.ObjectName(cp))
-	if err := a.controlPlaneChart.Apply(ctx, a.chartApplier, cp.Namespace, a.imageVector, a.gardenerClientset.Version(), version, values); err != nil {
+	if err := a.controlPlaneChart.Apply(ctx, a.chartApplier, a.client, cp.Namespace, a.imageVector, a.gardenerClientset.Version(), version, values); err != nil {
 		return false, errors.Wrapf(err, "could not apply control plane chart for controlplane '%s'", kutil.ObjectName(cp))
 	}
 
--- a/pkg/utils/chart/chart.go
+++ b/pkg/utils/chart/chart.go
@@ -16,6 +16,7 @@ package chart
 
 import (
 	"context"
+	"strings"
 
 	"github.com/gardener/gardener/pkg/chartrenderer"
 	gardenerkubernetes "github.com/gardener/gardener/pkg/client/kubernetes"
@@ -32,7 +33,7 @@ import (
 type Interface interface {
 	// Apply applies this chart in the given namespace using the given ChartApplier. Before applying the chart,
 	// it collects its values, injecting images and merging the given values as needed.
-	Apply(context.Context, gardenerkubernetes.ChartApplier, string, imagevector.ImageVector, string, string, map[string]interface{}) error
+	Apply(context.Context, gardenerkubernetes.ChartApplier, client.Client, string, imagevector.ImageVector, string, string, map[string]interface{}) error
 	// Render renders this chart in the given namespace using the given chartRenderer. Before rendering the chart,
 	// it collects its values, injecting images and merging the given values as needed.
 	Render(chartrenderer.Interface, string, imagevector.ImageVector, string, string, map[string]interface{}) (string, []byte, error)
@@ -46,7 +47,13 @@ type Chart struct {
 	Path      string
 	Images    []string
 	Objects   []*Object
-	SubCharts []*Chart
+	SubCharts []*SubChart
+}
+
+// SubChart represents a Helm sub-chart. A sub-chart is a chart that also can be conditionally enabled or disabled.
+type SubChart struct {
+	Chart
+	Condition string
 }
 
 // Object represents an object deployed by a Chart.
@@ -60,6 +67,7 @@ type Object struct {
 func (c *Chart) Apply(
 	ctx context.Context,
 	chartApplier gardenerkubernetes.ChartApplier,
+	client client.Client,
 	namespace string,
 	imageVector imagevector.ImageVector,
 	runtimeVersion, targetVersion string,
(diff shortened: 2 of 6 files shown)


extensions/pkg/controller/controlplane/genericactuator/actuator.go
```diff
@@ -204,7 +204,7 @@ func (a *actuator) reconcileControlPlaneExposure(
 	// Apply control plane exposure chart
 	a.logger.Info("Applying control plane exposure chart", "controlplaneexposure", kutil.ObjectName(cp), "values", values)
 	version := cluster.Shoot.Spec.Kubernetes.Version
-	if err := a.controlPlaneExposureChart.Apply(ctx, a.chartApplier, cp.Namespace, a.imageVector, a.gardenerClientset.Version(), version, values); err != nil {
+	if err := a.controlPlaneExposureChart.Apply(ctx, a.chartApplier, a.client, cp.Namespace, a.imageVector, a.gardenerClientset.Version(), version, values); err != nil {
```
ialidzhikov, line 207: You need to add a release note that it is a breaking change.

pkg/utils/chart/chart.go
```diff
@@ -46,7 +47,12 @@ type Chart struct {
 	Path      string
 	Images    []string
 	Objects   []*Object
-	SubCharts []*Chart
+	SubCharts []*SubChart
+}
+
+type SubChart struct {
```
ialidzhikov, line 54: Missing doc string

pkg/utils/chart/chart.go
```diff
@@ -77,6 +84,20 @@ func (c *Chart) Apply(
 	if err != nil {
 		return errors.Wrapf(err, "could not apply chart '%s' in namespace '%s'", c.Name, namespace)
 	}
+
+	// Delete disabled subcharts
+	for _, sc := range c.SubCharts {
+		v, err := evaluateCondition(additionalValues, sc.Condition)
```
ialidzhikov, line 91: Can you add unit tests about the new handling that is being introduced with this PR?

pkg/utils/chart/chart.go
```diff
@@ -77,6 +84,20 @@ func (c *Chart) Apply(
...
+		if err != nil {
+			return errors.Wrapf(err, "could not evaluate condition %s for subchart %s", sc.Condition, sc.Name)
+		}
+		if !v {
+			if err := sc.Delete(ctx, client, namespace); err != nil {
```
ialidzhikov, line 96: How 'not found' and 'no match' errors are being handled?

ialidzhikov (changes requested):

pkg/utils/chart/chart.go
stoyanr: There are no unit tests at all for `chart.Apply` right now, but ok, I will add them.

pkg/utils/chart/chart.go
stoyanr: You can see the `Object.Delete` code a few lines below: https://github.com/gardener/gardener/blob/d5fd42721faa352ffbed071c0314d25f158b472a/pkg/utils/chart/chart.go#L159-L162, a `nil` is returned on `NotFound`. `Chart.Delete` resolves recursively to a series of `Object.Delete` invocations. No need to handle such errors here or elsewhere in this file where `Chart.Delete` is used.

pkg/utils/chart/chart.go
stoyanr: Will add it.

extensions/pkg/controller/controlplane/genericactuator/actuator.go
stoyanr: Done

pkg/utils/chart/chart.go
stoyanr: Done

pkg/utils/chart/chart.go
stoyanr: Done

pkg/utils/chart/chart.go
stoyanr: I noticed that `client.Delete` a few lines below the above doesn't handle `NotFound` and added it just in case.

timebertt (changes requested):
TBH, I see several problems with this change:
- we are duplicating the logic for evaluating subcharts conditions from the helm chart renderer
- we are introducing a breaking change in the Chart interface, though it's not completely obvious why a dedicated client is needed for applying a Chart, if a `ChartApplier` is already handed over (which normally carries out all the client operations like `Apply` and `Delete`)
- we are duplicating cleanup logic for charts, which we already have in place in the gardener-resource-manager

One of the motivations behind gardener-resource-manager is to tackle exactly those situations, where the set of resources rendered out by a chart changes over time / from reconciliation to reconciliation, and then grm takes care about everything, including merging new config with current config, cleaning up removed resources and so on.

So my proposal would be to not introduce this change in the Chart in g/g and rather solve https://github.com/gardener/gardener-extension-provider-azure/issues/218 by deploying the remedy controller via a `ManagedResource`, just like gardener does for the `refeference-resources` and also some extensions controllers do for other controllers running in the control plane.
WDYT @stoyanr?

Other opinions, @gardener/gardener-maintainers?

ialidzhikov:
> So my proposal would be to not introduce this change in the Chart in g/g and rather solve gardener/gardener-extension-provider-azure#218 by deploying the remedy controller via a ManagedResource, just like gardener does for the refeference-resources and also some extensions controllers do for other controllers running in the control plane.
WDYT @stoyanr?

Do you mean here to use a ManagedResource for applying the controlplane chart in the genericactuator - ref https://github.com/gardener/gardener/blob/v1.14.0/extensions/pkg/controller/controlplane/genericactuator/actuator.go#L285-L296?
-> If that is the case, I think @stoyanr had some concerns that it will be a PR that would affect all provider extensions. But yeah, I tend to agree that this is the cleaner approach. If @stoyanr does not have free capacity, then we can pick this. @rfranzke do you recall why the genericactuator.Actuator is using the chartApplier? I guess this is because of legacy reason and we can switch to a ManagedResource?

Or you have in mind a separate ManagedResource for the remedy-controller? I see that the existing ValuesProvider interface is not very suitable for it as it only returns chart values and delegates the chart creation to the generic actuator. However this is probably achievable with an Actuator that is deploying the ManagedResource and is calling the genericactuator.Actuator?

stoyanr:
Closed this PR and also https://github.com/gardener/gardener-extension-provider-azure/pull/223 that depends on it because I would not like to introduce fundamental, time consuming changes just to fix a minor issue.
My intention with the above change was to introduce a minor correction to an existing mechanism. There are good (albeit "non-obvious") reasons why this mechanism exists, I would not like to discuss them here. It's historically older than ManagedResources so at the time it was introduced there was no alternative. If the desire is to get rid of the mechanism altogether and replace it with something else, I am fine with this, however, I would not do this as part of this PR.
If for whatever reason you decide to merge this PR with its current design, let me know.

timebertt:
> Or you have in mind a separate ManagedResource for the remedy-controller?

I had this in mind actually, but your other idea with using MRs in general for the generic controlplane actuator might also be an option.

> Closed this PR and also gardener/gardener-extension-provider-azure#223 that depends on it because I would not like to introduce fundamental, time consuming changes just to fix a minor issue.

You can still choose to do the migration to a MR for the remedy-controller in provider-azure, if you want. That shouldn't be too time-consuming.
Nevertheless, I agree, that this is a minor issue, and it could also be tackled later on with some general clenaup if we agree on that (I see @ialidzhikov already agreed, that the MR is the cleaner approach here).

stoyanr:
> You can still choose to do the migration to a MR for the remedy-controller in provider-azure, if you want. That shouldn't be too time-consuming.

If you take a look at the code, you will see that the remedy-controller is deployed just as yet another controlplane chart, relying on existing mechanisms in the `genericactuator`. I don't think I would be able to deploy only the remedy controller as a MR without substantially reworking the `genericactuator`. Or even if I could, it would still be time-consuming enough while being much less "clean" from my perspective.

As I said, this was supposed to be a small change and this was the only reason I undertook doing it at this point. As it's not small anymore, but the issue it's fixing itself is relatively minor, I would rather focus on other, higher value tasks.

PR #3282, opened 2020-12-09
Use system-node-critical for apiserver-proxy Daemonset

mvladev:

/area scalability
/area networking
/kind bug
/priority normal

- Use `system-node-critical` for apiserver-proxy Daemonset
- Increase memory limits to avoid OOM killer
Fixes #

**Release note**:

```bugfix user
`apiserver-proxy` now uses `system-node-critical` priority class. Memory limit is also increased to avoid OOM killer.
```


rfranzke (approved):
/lgtm

PR #3285, opened 2020-12-09
Promote 2 nonserial beta tests to default

ialidzhikov:
/area testing
/kind cleanup

As nonserial and beta tests currently we have:
- ✅  `Shoot application metrics testing [BETA] [SHOOT] should read runtime metrics` - according the tm data from the last 2,053 runs 30 failed -> promotion from beta to default
- ✅  `Shoot network testing [BETA] [SHOOT] should reach all webservers on all nodes` - according the tm data from the last 2,053 run 13 failed ->  promotion from beta to default
- ❌  `Seed logging testing [BETA] [SHOOT] should get container logs from loki by tenant` - it was added recently -> we need to observe the success rate
- ❌  `Shoot worker operation testing [BETA] [SHOOT] Shoot node's operating systems should differ if the specified workers are different` - this test is skipped when it is called with shoot with only 1 worker pool, hence it is always skipped in our flavors -> we need to follow up how we proceed with this test (whether we adapt/remove the test, adapt the flavor, something else?). 

This PR is about the nonserial beta tests. The serial beta test - I will try to address them in a separate PR.

Ref https://github.com/gardener/gardener/issues/1564#issuecomment-682406933

**Release note**:

```other operator
NONE
```

### DIFF
--- a/test/framework/k8s_utils.go
+++ b/test/framework/k8s_utils.go
@@ -1,3 +1,17 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


vpnachev (dismissed):
/lgtm

schrodit (approved):

dguendisch (approved):

PR #3286, opened 2020-12-10
Only consider effective shoot for health checks

### LINKED ISSUE #2706: Invalid health checks for shoots w/ confineSpecUpdateRollout

/area control-plane
/area operations
/kind bug
/priority normal

**What happened**:
The `confineSpecUpdateRollout` (https://github.com/gardener/gardener/pull/2233) feature leads to shoot specs which will get active during the next maintenance window (or manual reconciliation) only. Gardener on the contrary determines the set of required health checks for an individual shoot based on exactly these yet inactive specs and this can lead to failed conditions until the spec gets active.

The set of executed health checks is based on the actual/active spec of the shoot cluster.

https://github.com/gardener/gardener/pull/2698/commits/7ddaf77a94f591e67077d4a41164f50e1cf3397c already changed this behavior for VPA, and cluster autoscaler deployment checks but other places might still be missing, e.g. Alertmanager.

**Environment**:

- Gardener version: v1.8.0

timuthy:

/area control-plane
/kind bug
/priority normal

This PR changes the care controller in a way that health checks to be executed now consider only the effective shoot spec, i.e. respecting the `.spec.maintenance.confineSpecRollout` option. More information can be found in the linked issue.

Fixes #2706

**Release note**:

```bugfix operator
Gardener health checks now take the effective Shoot specification into consideration if `.spec.maintenance.confineSpecRollout` is used. Earlier, `EveryNodeReady` or `ControlPlaneHealthy` conditions reported an invalid state if the specification was changed but not yet effective due to a rollout during shoot maintenance (`confineSpecRollout: true`).
```

### DIFF
--- a/pkg/operation/botanist/health_check.go
+++ b/pkg/operation/botanist/health_check.go
@@ -525,7 +525,7 @@ func (b *Botanist) checkControlPlane(
 		return exitCondition, err
 	}
 
-	wantsAlertmanager := !gardencorev1beta1helper.ShootIgnoresAlerts(effectiveShoot) && effectiveShoot.Spec.Monitoring != nil && effectiveShoot.Spec.Monitoring.Alerting != nil && len(effectiveShoot.Spec.Monitoring.Alerting.EmailReceivers) > 0
+	wantsAlertmanager := gardencorev1beta1helper.ShootWantsAlertManager(effectiveShoot)
 	if exitCondition, err := checker.CheckMonitoringControlPlane(b.Shoot.SeedNamespace, gardencorev1beta1helper.GetPurpose(effectiveShoot) == gardencorev1beta1.ShootPurposeTesting, wantsAlertmanager, condition, seedDeploymentLister, seedStatefulSetLister); err != nil || exitCondition != nil {
 		return exitCondition, err
 	}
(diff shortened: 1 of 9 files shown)


pkg/operation/botanist/health_check.go
```diff
@@ -522,30 +521,16 @@ func (b *Botanist) checkControlPlane(
...
-		return nil, errors.New("failed to execute control plane health checks because shoot is missing in cluster resource")
-	}
-
-	if exitCondition, err := checker.CheckControlPlane(shoot, b.Shoot.SeedNamespace, condition, seedDeploymentLister, seedEtcdLister, seedWorkerLister); err != nil || exitCondition != nil {
+	if exitCondition, err := checker.CheckControlPlane(effectiveShoot, b.Shoot.SeedNamespace, condition, seedDeploymentLister, seedEtcdLister, seedWorkerLister); err != nil || exitCondition != nil {
 		return exitCondition, err
 	}
-	if exitCondition, err := checker.CheckMonitoringControlPlane(b.Shoot.SeedNamespace, b.Shoot.GetPurpose() == gardencorev1beta1.ShootPurposeTesting, b.Shoot.WantsAlertmanager, condition, seedDeploymentLister, seedStatefulSetLister); err != nil || exitCondition != nil {
+
+	wantsAlertmanager := !gardencorev1beta1helper.ShootIgnoresAlerts(effectiveShoot) && effectiveShoot.Spec.Monitoring != nil && effectiveShoot.Spec.Monitoring.Alerting != nil && len(effectiveShoot.Spec.Monitoring.Alerting.EmailReceivers) > 0
```
rfranzke: Probably this could be a helper function as well that can be called here and [here](https://github.com/gardener/gardener/blob/master/pkg/operation/shoot/shoot.go#L163), wdyt?

rfranzke (approved):
/lgtm

PR #3289, opened 2020-12-10
Wait for apiserver Service to be deleted on hibernation

ialidzhikov:
/kind bug

We see often Shoots failing to wake up with err:
```yaml
  lastErrors:
  - codes:
    - ERR_CONFIGURATION_PROBLEM
    description: 'task "Waiting until Shoot control plane exposure has been reconciled"
      failed: Error while waiting for ControlPlane shoot--foo--bar/bar-exposure
      to become ready: extension encountered error during reconciliation: Error reconciling
      controlplane: could not get kube-apiserver service load balancer address: Service
      "kube-apiserver" not found'
    lastUpdateTime: "2020-12-10T11:59:47Z"
    taskID: Waiting until Shoot control plane exposure has been reconciled
```

The reason is that in `HibernateControlPlane` we delete the apiserver svc but we don't wait for it be completed. In some cases when the svc is not gone after hibernation (Seed cloud-controller-manager is unavailable for short period of time, or the LB deletion is taking time on cloud provider side) an immediate wake up can fail with this err.

Also the hibernation integration test is quite often failing with the reason. This PR should deflake it.

I tried to add a test but I think whole test suite should be reworked to use mock client instead of fake client. 

**Release note**:

```bugfix user
An race issue causing immediate wake up after hibernation to fail is now fixed. The hibernation is now waiting until the kube-apiserver Service is cleaned up.
```

### DIFF
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -305,6 +305,10 @@ func (b *Botanist) HibernateControlPlane(ctx context.Context) error {
 		if err := b.Shoot.Components.ControlPlane.KubeAPIServerService.Destroy(ctx); err != nil {
 			return err
 		}
+
+		if err := b.Shoot.Components.ControlPlane.KubeAPIServerService.WaitCleanup(ctx); err != nil {
+			return err
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


vpnachev (approved):
/lgtm

timebertt (approved):
Nice!
/lgtm

rfranzke (approved):

PR #3293, opened 2020-12-11
Add metrics to the admission controller

### LINKED ISSUE #3234: Add and Expose Admission Controller Metrics

/area monitoring
/kind enhancement
/priority normal

The Admission Controller should expose metrics about the resource creation it is blocking/could be blocking. Related to #2815

**Why is this needed**:

This would improve operations since an operator would not have to look at the logs, but instead on a dashboard to quickly see if there are any resources getting blocked by the admission controller.

wyb1:

/area monitoring
/kind enhancement
/priority normal

Expose metrics for the gardener admission controller. The admission controller will now expose the default go metrics in addition to 2 other metrics: 
`gardener_admission_controller_rejected_resources_total`
`gardener_admission_controller_invalid_webhook_requests_total`

These metrics track the number of resources that are rejected as well as any webhook requests that fail because they are invalid. 

Fixes #3234

/cc @timuthy @timebertt 

When updating a resource (that will be rejected) we noticed that the value of the metric got incremented by 2. After debugging this we noticed that the API Server was sending 2 requests to the webhook instead of 1 which is what we would have expected.

**Release note**:

```feature operator
The gardener admission controller now exposes metrics
```

### DIFF
--- a/pkg/admissioncontroller/server/metrics/metrics.go
+++ b/pkg/admissioncontroller/server/metrics/metrics.go
@@ -1,0 +1,57 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


pkg/admissioncontroller/server/metrics/metrics.go
```diff
@@ -0,0 +1,43 @@
+package metrics
```
rfranzke, line 15: license header

rfranzke (changes requested):
looks great, one nit

pkg/admissioncontroller/server/metrics/metrics.go
wyb1: Added license header :)

rfranzke (approved):
/lgtm
/invite @timuthy

timuthy (approved):
/lgtm

PR #3298, opened 2020-12-14
Correct order of admission plugins

timebertt:

/area usability
/kind bug
/priority normal

Before this PR, `gardener-apiserver` was calling the in-tree admission plugins after the webhook plugins. 
However, the correct behaviour would be to invoke all in-tree plugins first (for example including defaulting and so on) and then invoking the webhook plugins (that can further default/validate/...).

This PR corrects this bug.

xref https://github.com/gardener/gardener-extension-provider-alicloud/pull/203

/cc @minchaow @jia-jerry 

**Release note**:

```bugfix operator
By default, gardener-apiserver now invokes in-tree admission plugins before invoking the webhook plugins.
```

### DIFF
--- a/cmd/gardener-apiserver/app/gardener_apiserver.go
+++ b/cmd/gardener-apiserver/app/gardener_apiserver.go
@@ -38,22 +38,7 @@ import (
 	"github.com/gardener/gardener/pkg/openapi"
 	"github.com/gardener/gardener/pkg/version"
 	"github.com/gardener/gardener/pkg/version/verflag"
-	controllerregistrationresources "github.com/gardener/gardener/plugin/pkg/controllerregistration/resources"
-	"github.com/gardener/gardener/plugin/pkg/global/customverbauthorizer"
-	"github.com/gardener/gardener/plugin/pkg/global/deletionconfirmation"
-	"github.com/gardener/gardener/plugin/pkg/global/extensionvalidation"
-	"github.com/gardener/gardener/plugin/pkg/global/resourcereferencemanager"
-	plantvalidator "github.com/gardener/gardener/plugin/pkg/plant"
-	seedvalidator "github.com/gardener/gardener/plugin/pkg/seed/validator"
-	shootdns "github.com/gardener/gardener/plugin/pkg/shoot/dns"
-	"github.com/gardener/gardener/plugin/pkg/shoot/oidc/clusteropenidconnectpreset"
-	"github.com/gardener/gardener/plugin/pkg/shoot/oidc/openidconnectpreset"
-	shootquotavalidator "github.com/gardener/gardener/plugin/pkg/shoot/quotavalidator"
-	shoottolerationrestriction "github.com/gardener/gardener/plugin/pkg/shoot/tolerationrestriction"
-	shootvalidator "github.com/gardener/gardener/plugin/pkg/shoot/validator"
-	shootstatedeletionvalidator "github.com/gardener/gardener/plugin/pkg/shootstate/validator"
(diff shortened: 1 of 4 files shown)


rfranzke (dismissed):
/lgtm

timuthy (changes requested):
What about this one?
```
// This plugin must remain the last one in the list since it updates the quota usage
// which can only happen reliably if previous plugins permitted the request.
resourcequota.PluginName,
```

cmd/gardener-apiserver/app/gardener_apiserver.go
```diff
@@ -147,11 +129,14 @@ func NewOptions(out, errOut io.Writer) *Options {
 		schema.GroupKind{Group: gardencorev1alpha1.GroupName},
 		schema.GroupKind{Group: gardencorev1beta1.GroupName},
 	)
+	apiserver.RegisterAllAdmissionPlugins(o.Recommended.Admission.Plugins)
+	o.Recommended.Admission.RecommendedPluginOrder = apiserver.AllOrderedPlugins
```
timuthy, line 133: Before, we only appended to the `RecommendedPluginOrder` and now there is a hard coded list only. Can this be a problem if new admission plugins are added through the API Server lib?

timuthy (commented):
Please forget about my previous comment. Still have a question, though.

cmd/gardener-apiserver/app/gardener_apiserver.go
timebertt: Yes, we won't notice, when the default list of plugins in the API server library will change, that's correct.
Though, I would argue, that there is no way of correctly detecting automatically in which order we should insert our admission plugins between the defaults.

So, I added a small test, that will break, when the default set and/or order of admission plugins changes.
PTAL :)

timuthy (approved):
/lgtm

minchaow:
@timebertt: will these two commits be cherry-picked into hotfix branch? Tests of alicloud provider from TM is blocking, so we probably need it to unblock the validation process.

rfranzke:
@minchaow we can cherry-pick it and release it with the next hotfix, but a faster solution is to fix the test for the alicloud provider

minchaow:
> @minchaow we can cherry-pick it and release it with the next hotfix, but a faster solution is to fix the test for the alicloud provider

yes, it would be faster if we could fix the test for the alicloud, if so, we need modify the shoots' manifest from TM.

jia-jerry:
Hi @rfranzke Is it possible we could release a hotfix for Gardener this week or even by the end of today? Otherwise we need to change many places of TM configurations. And also, we could not create a shoot from dashboard as a smoke test.

rfranzke:
TBH, I would like to validate this change first for some time before hotfixing it immediately. It can have undesired side effects due to changed admission plugin order.

timebertt:
> TBH, I would like to validate this change first for some time before hotfixing it immediately. It can have undesired side effects due to changed admission plugin order.

Yes, I agree with that.

To sort things out in the short term, please fix the TM manifests for alicloud or roll back the changes on the effected landscape.

jia-jerry:
But even though, we can't create a shoot from the dashboard. It doesn't make sense to release alicloud extension and apply it to higher landscapes before we apply the current PR for g/g. Or shall we pause the release of AliCloud extension for now until g/g hotfix is released? cc@rfranzke

rfranzke:
From my chat with @minchaow:

> [...] However, if you need an URGENT solution, please go ahead and remove/disable the validation in Alicloud validation for now. It is MUCH faster and less “potentially disruptive”.
You could make the validation configurable via CLI flag for example and later remove this option again.

PR #3301, opened 2020-12-14
Check if shoot is no longer managed by this gardenlet before reconcilation

krgostev:
/area control-plane
/kind enhancement
/priority normal

This PR adds additional check before the start of the reconciliation process if the whether is managed by this gardenlet or not. If the shoot is no longer managed by this gardenlet (e.g., due to migration to another seed) it should no longer be requeued.  

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control.go
@@ -183,7 +183,6 @@ func (c *Controller) reconcileShootRequest(req reconcile.Request) (reconcile.Res
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-	key := shootKey(shoot)
 
 	// fetch related objects required for shoot operation
 	project, err := common.ProjectForNamespace(c.k8sGardenCoreInformers.Core().V1beta1().Projects().Lister(), shoot.Namespace)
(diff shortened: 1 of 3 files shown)


pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -183,8 +184,12 @@ func (c *Controller) reconcileShootRequest(req reconcile.Request) (reconcile.Res
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-	key := shootKey(shoot)
 
+	// if shoot is no longer managed by this gardenlet (e.g., due to migration to another seed) then don't requeue
+	if seedName := confighelper.SeedNameFromSeedConfig(c.config.SeedConfig); (len(seedName) > 0 && *shoot.Spec.SeedName != seedName) || (len(seedName) == 0 && !controllerutils.SeedLabelsMatch(c.seedLister, *shoot.Spec.SeedName, c.config.SeedSelector)) {
```
rfranzke: This is the same code like in https://github.com/gardener/gardener/blob/aec56a2c408c6cd67e8fe7aef3a613c06559dafe/pkg/gardenlet/controller/shoot/shoot_care_control.go#L124, right? Can you extract it into a function that can be used on both places?

rfranzke (changes requested):

rfranzke (approved):
/lgtm

PR #3306, opened 2020-12-15
[release-v1.14] Correct order of admission plugins 

jia-jerry:

/area usability
/kind bug
/priority normal

Before this PR, `gardener-apiserver` was calling the in-tree admission plugins after the webhook plugins.
However, the correct behaviour would be to invoke all in-tree plugins first (for example including defaulting and so on) and then invoking the webhook plugins (that can further default/validate/...).

This PR corrects this bug.

ref #3298

**Release note**:

```other operator
By default, gardener-apiserver now invokes in-tree admission plugins before invoking the webhook plugins.
```


timebertt (approved):
/lgtm

PR #3308, opened 2020-12-16
WorkerDelegate: factor out optional interface WorkerCredentialsDelegate

danielfoehrKn:

/area quality
/kind cleanup
/priority normal

Making the update of all machine class secrets in the generic worker actuator optional.
It is only performed if the actuator implements the deprecated WorkerCredentialsDelegate interface.

Alternatively, extensions  can now use the field in the machine class spec.CredentialsSecretRef so that [all refer to the same "cloudprovider" secret in the shoot namespace in the Seed](https://github.com/gardener/gardener-extension-provider-aws/pull/238).

Fixes #

**Release note**:

```other dependency
Making the implementation of the function `GetMachineControllerManagerCloudCredentials` in the `WorkerDelegate` optional. Alternatively, extensions  can now use the field in the machine class `spec.credentialsSecretRef` so that all machine classes refer to the same secret from the `Worker` field `spec.secretRef`. See [here for more details](https://github.com/gardener/machine-controller-manager/pull/578).
```

### DIFF
--- a/extensions/pkg/controller/worker/genericactuator/actuator_delete.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator_delete.go
@@ -64,13 +64,15 @@ func (a *genericActuator) Delete(ctx context.Context, worker *extensionsv1alpha1
 		return errors.Wrapf(err, "failed to deploy the machine classes")
 	}
 
-	// Update cloud credentials for all existing machine class secrets
-	cloudCredentials, err := workerDelegate.GetMachineControllerManagerCloudCredentials(ctx)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


rfranzke (changes requested):
If we now simply revert this then extension controllers that do not yet (want to) use the new `credentialsSecretRef` field are back to square one and have the bug where outdated machine class secrets are not updated.
As suggested in Slack, let's rather make this behaviour optional (move the function to a separate interface and check if the delegate implements it) but keep it for a while.

rfranzke (approved):
/lgtm

@danielfoehrKn if you confirm everything is OK then let's get this in so that you can proceed with https://github.com/gardener/gardener-extension-provider-aws/pull/238

PR #3222, opened 2020-11-25
Refactor ControlPlane extension deployment into component

rfranzke:
**How to categorize this PR?**

/area dev-productivity quality robustness
/kind technical-debt test
/priority normal
/topology seed
/platform all
/exp intermediate

**What this PR does / why we need it**:
This PR refactors the deployment for the `ControlPlane` extension resource and adopts the component concept. This improves several aspects and the overall code quality.

**Which issue(s) this PR fixes**:
Part of #2754

**Release note**:

```improvement operator
NONE
```

PR #3311, opened 2020-12-16
Refactor Worker deployment into component

rfranzke:

/kind technical-debt test
/priority normal
/topology seed
/platform all
/exp intermediate

This PR refactors the deployment for the `Worker` extension resource and adopts the component concept, similar to #3222. This improves several aspects and the overall code quality.

I've added some missing tests and fixed some small bugs along the road.

Part of #2754

**Release note**:

```improvement operator
NONE
```


plkokanov:
Looks good. Thanks for noticing and editing the migration.go file to properly destory/migrate resources.
I will ask @kris94 to run a test, docker update deleted my kind local setup yet again ...

rfranzke:
Sure. @plkokanov you might want to try out the nodeless setup - way easier :)

plkokanov:
Yep, switched to that, totally easier :D. Meanwhile, @kris94 tried running a migration with this PR but there's a problem with the Infrastructure state. Checking to see if it can be related to this PR (running migration with the master branch seems ok). 

Basically both the infra and worker states in the ShootState are empty strings.

rfranzke:
Hm, that sounds strange @plkokanov, this PR doesn't touch the `Infrastructure` or the `ShootState` handling, does it? 👀

plkokanov:
Indeed, that's why i'm puzzled as well. I'll run it locally now as well and see if I can reproduce it, but it's most likely caused by something else.

plkokanov:
Couldn't reproduce it so I'd say this is ok to be merged.

krgostev:
ok this was a bit strange to me as well, so I removed everything and started from scratch. After that everything that I tested is working as expected.

pkg/operation/botanist/extensions/controlplane/controlplane_test.go
```diff
@@ -425,12 +425,8 @@ var _ = Describe("ControlPlane", func() {
 			Expect(defaultDepWaiter.WaitMigrate(ctx)).To(HaveOccurred())
 		})
 
-		It("should not return error if resource gets migrated successfully (purpose != exposure)", func() {
-			values.Purpose = extensionsv1alpha1.Exposure
-			defaultDepWaiter = controlplane.New(log, c, values, time.Millisecond, 250*time.Millisecond, 500*time.Millisecond)
-
+		It("should not return error if resource gets migrated successfully (purpose == exposure)", func() {
```
ialidzhikov: Hm, there are now 2 similar tests that have the same IT name - `should not return error if resource gets migrated successfully (purpose == exposure)`. I guess one of them should be with `(purpose != exposure)`?

danielfoehrKn: It should be (purpose == normal) for this test.

danielfoehrKn (dismissed):
Tested & looks good to me

pkg/operation/botanist/extensions/controlplane/controlplane_test.go
ialidzhikov:
```suggestion
		It("should not return error if resource gets migrated successfully (purpose != exposure)", func() {
```

rfranzke: fixed

ialidzhikov (approved):
/lgtm

danielfoehrKn (approved):

PR #3315, opened 2020-12-17
Use nginx-ingress 0.41 version for Kubernetes 1.20 shoots

rfranzke:

/area security usability
/kind enhancement
/priority normal

We have not updated our nginx-ingress addon since very long time due to several issues that end-users faced after the upgrade. Instead, we have marked it as "not recommended for production" and simply kept it at `v0.22.0`.
Unfortunately, several vulnerabilities have been detected in `v0.22.0`. So, for Kubernetes 1.20 shoot clusters we now update the nginx-ingress addon version to `v0.41.2`. As end-users trigger the update to 1.20 usually explicitly they can adapt accordingly if necessary. Shoots with a lower Kubernetes version will remain with `v0.22.0` version.

/cc @ThormaehlenFred 

**Release note**:

```noteworthy user
The version for the `nginx-ingress` addon for shoots has been updated to `v0.41.2` **ONLY** for Kubernetes 1.20 shoot clusters. All shoot clusters with Kubernetes < 1.20 will remain with the current `v0.22.0` version. Please be reminded that the `nginx-ingress` addon is not recommended for production scenarios and that you should deploy (+ customize) your own ingress controller instead. Please use it only for development/evaluation purposes.
```


rfranzke:
/cc @vlerenc 

FYI: For shooted seeds with Kubernetes < 1.20 the new "managed ingress controller" feature that will be introduced with #3131 should be used. It also uses the new `v0.41.2` version of `nginx-ingress`.

vlerenc:
Thanks @rfranzke . If we update nginx for new Kubernetes versions, we somehow get out of the trap with the old nginx version. And yes, using the same throughout all our seeds is also great, I hope. :-)

charts/shoot-addons/charts/nginx-ingress/templates/clusterrole.yaml
```diff
@@ -12,24 +12,16 @@ rules:
   - apiGroups:
       - ""
     resources:
+{{- if semverCompare "< 1.20-0" .Capabilities.KubeVersion.GitVersion }}
```
timuthy, line 15: What do you think about passing and comparing the ingress controller version instead of comparing the K8s version? This is a little bit confusing, IMO.

timuthy (commented):
Looks good, just one suggestion.

charts/shoot-addons/charts/nginx-ingress/templates/clusterrole.yaml
rfranzke: Good idea, I thought about the same, but I don't know the ingress-controller version. Due to the image vector overwrite some sha can be injected, and then the comparison does not work anymore. :(

timuthy: Argh .... ok 😞

charts/shoot-addons/charts/nginx-ingress/templates/clusterrole.yaml
```diff
@@ -12,24 +12,16 @@ rules:
...
       - configmaps
+{{- end }}
       - endpoints
       - nodes
       - pods
       - secrets
     verbs:
       - list
       - watch
-{{- if and .Values.controller.scope.enabled .Values.controller.scope.namespace }}
```
timuthy, line 23: Do you want to completely remove this option from the chart, i.e, `values.yaml` and `controller-deployment.yaml`?

rfranzke: done

timuthy (approved):
/lgtm

timebertt (approved):

PR #3318, opened 2020-12-17
Refactor BackupEntry handling into components

rfranzke:

/kind technical-debt test
/priority normal
/topology seed
/platform all
/exp intermediate

This PR refactors the deployment for the `BackupEntry` extension resource as well as the `BackupEntry` core resource and adopts the component concept, similar to #3222. This improves several aspects and the overall code quality.

Please note that there are two `BackupEntry` resources (one in `core.gardener.cloud` and one in `extensions.gardener.cloud` API groups). This PR introduces the component concept for both.

/invite @plkokanov @swilen-iwanow @kris94 

Part of #2754

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/backupentry/backupentry.go
+++ b/pkg/operation/botanist/backupentry/backupentry.go
@@ -1,0 +1,165 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backupentry
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	"github.com/gardener/gardener/pkg/operation/botanist/component"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+	"github.com/gardener/gardener/pkg/utils/retry"
+
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	// DefaultInterval is the default interval for retry operations.
+	DefaultInterval = 5 * time.Second
+	// DefaultTimeout is the default timeout and defines how long Gardener should wait
+	// for a successful reconciliation of a BackupEntry resource.
+	DefaultTimeout = 10 * time.Minute
+)
+
+// TimeNow returns the current time. Exposed for testing.
+var TimeNow = time.Now
+
+// Values contains the values used to create a BackupEntry resource.
+type Values struct {
+	// Namespace is the namespace of the BackupEntry resource.
+	Namespace string
+	// Name is the name of the BackupEntry resource.
+	Name string
+	// OwnerReference is a reference to an owner for BackupEntry resource.
+	OwnerReference *metav1.OwnerReference
+	// SeedName is the name of the seed to which the BackupEntry shall be scheduled.
+	SeedName *string
+	// OverwriteSeedName indicates whether the provided SeedName in the values shall be used even if the BackupEntry
+	// already exists (if this is false then the existing `.spec.seedName` will be kept even if the SeedName in these
+	// values differs.
+	OverwriteSeedName bool
+	// BucketName is the name of the bucket in which the BackupEntry shall be reconciled. This value is only used if the
+	// BackupEntry does not exist yet. Otherwise, the existing `.spec.bucketName` will be kept even if the BucketName in
+	// these values differs.
+	BucketName string
+}
+
+// New creates a new instance of DeployWaiter for a BackupEntry.
+func New(
+	logger logrus.FieldLogger,
+	client client.Client,
+	values *Values,
+	waitInterval time.Duration,
+	waitTimeout time.Duration,
+) component.DeployWaiter {
+	return &backupEntry{
+		client:       client,
+		logger:       logger,
+		values:       values,
+		waitInterval: waitInterval,
+		waitTimeout:  waitTimeout,
+	}
+}
+
+type backupEntry struct {
+	values       *Values
+	logger       logrus.FieldLogger
+	client       client.Client
+	waitInterval time.Duration
+	waitTimeout  time.Duration
+}
+
+// Deploy uses the garden client to create or update the BackupEntry resource in the project namespace in the Garden.
+func (b *backupEntry) Deploy(ctx context.Context) error {
+	var (
+		backupEntry = &gardencorev1beta1.BackupEntry{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      b.values.Name,
+				Namespace: b.values.Namespace,
+			},
+		}
+		bucketName = b.values.BucketName
+		seedName   = b.values.SeedName
+	)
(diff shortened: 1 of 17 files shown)


pkg/operation/botanist/extensions/backupentry/backupentry.go
```diff
@@ -0,0 +1,218 @@
...
+	SetProviderConfig(*runtime.RawExtension)
+	SetRegion(string)
+	SetBackupBucketProviderStatus(*runtime.RawExtension)
+}
+
+// Values contains the values used to create a BackupEntry CRD
+type Values struct {
+	// Name is the name of the BackupEntry extension.
+	Name string
+	// Type is the type of BackupEntry plugin/extension (e.g calico)
```
ialidzhikov:
```suggestion
	// Type is the type of BackupEntry plugin/extension (e.g aws)
```

pkg/operation/botanist/backupentry/backupentry.go
```diff
@@ -0,0 +1,167 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	// DefaultInterval is the default interval for retry operations.
+	DefaultInterval = 5 * time.Second
+	// DefaultSevereThreshold is the default threshold until an error reported by another component is treated as 'severe'.
+	DefaultSevereThreshold = 30 * time.Second
```
ialidzhikov:
```suggestion
```

It seems to be unused.

pkg/operation/botanist/backupentry/backupentry.go
```diff
@@ -0,0 +1,167 @@
...
+	// for a successful reconciliation of a BackupEntry resource.
+	DefaultTimeout = 10 * time.Minute
+)
+
+// TimeNow returns the current time. Exposed for testing.
+var TimeNow = time.Now
+
+// Values contains the values used to create a BackupEntry resource.
+type Values struct {
+	// Namespace is the name of the BackupEntry resource.
```
ialidzhikov:
```suggestion
	// Namespace is the namespace of the BackupEntry resource.
```

pkg/operation/botanist/extensions/backupentry/backupentry.go
```diff
@@ -0,0 +1,218 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backupentry
+
+import (
+	"context"
+	"time"
+
+	"github.com/gardener/gardener/pkg/apis/core/v1alpha1"
```
ialidzhikov:
```suggestion
	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
```

pkg/operation/botanist/extensions/backupentry/backupentry.go
```diff
@@ -0,0 +1,218 @@
...
+type backupEntry struct {
+	values              *Values
+	logger              logrus.FieldLogger
+	client              client.Client
+	waitInterval        time.Duration
+	waitSevereThreshold time.Duration
+	waitTimeout         time.Duration
+}
+
+// Deploy uses the seed client to create or update the BackupEntry custom resource in the Shoot namespace in the Seed
```
ialidzhikov:
```suggestion
// Deploy uses the seed client to create or update the BackupEntry custom resource in the Seed
```

pkg/operation/botanist/extensions/backupentry/backupentry.go
```diff
@@ -0,0 +1,218 @@
...
+			BackupBucketProviderStatus: b.values.BackupBucketProviderStatus,
+		}
+
+		return nil
+	})
+
+	return err
+}
+
+// Restore uses the seed client and the ShootState to create the BackupEntry custom resource in the Shoot namespace in the Seed and restore its state
```
ialidzhikov:
```suggestion
// Restore uses the seed client and the ShootState to create the BackupEntry custom resource in the Seed and restore its state
```

pkg/operation/botanist/backupentry/backupentry_test.go
```diff
@@ -0,0 +1,230 @@
...
+			It("should create correct BackupEntry (newly created)", func() {
+				defer test.WithVars(&TimeNow, mockNow.Do)()
+				mockNow.EXPECT().Do().Return(now.UTC()).AnyTimes()
+
+				Expect(defaultDepWaiter.Deploy(ctx)).To(Succeed())
+
+				actual := &gardencorev1beta1.BackupEntry{}
+				Expect(c.Get(ctx, client.ObjectKey{Name: name, Namespace: namespace}, actual)).To(Succeed())
+
+				Expect(actual).To(DeepEqual(expected))
```
ialidzhikov, line 129: Why we use a custom `DeepEqual` matcher? I replaced the with the `DeepEqual` usage with gomega `Equal` matcher (which uses `reflect.DeepEqual` under the hood) and it worked well. And I was wondering why we don't simply use `Equal`.

```suggestion
				Expect(actual).To(Equal(expected))
```

rfranzke: The reason is to better show the mismatch/difference in case a test fails. While it's very hard to find it when using `Equal`, our custom `DeepEqual` matcher shows the YAMLs and makes it a lot easier to find the problem.

pkg/operation/botanist/backupentry/backupentry.go
rfranzke: It is used here, for example: https://github.com/gardener/gardener/pull/3318/files#diff-b637b9adf578441595351b86f64b2085fd8df9921a5f07e938930c1573c8becbR87

rfranzke:
Thanks for your review @ialidzhikov

ialidzhikov (changes requested):
I hit the following issue while creating a new Shoot locally

```
time="2021-01-05T10:27:15+02:00" level=error msg=Error backupentry=shoot--foo--bar-- error="retry failed with context canceled, last error: could not reconcile extension secret in seed: Secret \"entry-shoot--foo--bar--\" is invalid: metadata.name: Invalid value: \"entry-shoot--foo--bar--\": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')" flow="Backup Entry Reconciliation" task="Deploying backup entry secret to seed"
```

Is it also reproducible for you?

rfranzke:
@ialidzhikov No, I could not reproduce this. I again checked the code and I'm very much confused how this error shall be produced. Can you share more details? Any steps to reproduce it? I was trying with the following `Shoot` resource:

```yaml
kind: Shoot
metadata:
  name: rfranzke-aws
  namespace: garden-dev
spec:
  seedName: aws
  addons:
    kubernetesDashboard:
      enabled: true
    nginxIngress:
      enabled: true
  cloudProfileName: aws
  networking:
    nodes: 10.250.0.0/16
    type: calico
  provider:
    type: aws
    controlPlaneConfig:
      apiVersion: aws.provider.extensions.gardener.cloud/v1alpha1
      kind: ControlPlaneConfig
    infrastructureConfig:
      apiVersion: aws.provider.extensions.gardener.cloud/v1alpha1
      kind: InfrastructureConfig
      networks:
        vpc:
          cidr: 10.250.0.0/16
        zones:
        - internal: 10.250.112.0/22
          name: eu-west-1b
          public: 10.250.96.0/22
          workers: 10.250.0.0/19
    workers:
    - machine:
        type: m4.large
      maxSurge: 1
      maxUnavailable: 0
      maximum: 2
      minimum: 2
      name: cpu-worker
      volume:
        size: 20Gi
        type: gp2
      zones:
      - eu-west-1b
  region: eu-west-1
  secretBindingName: my-aws-account
  kubernetes:
    allowPrivilegedContainers: true
    kubeAPIServer:
      enableBasicAuthentication: false
    kubeControllerManager:
      nodeCIDRMaskSize: 24
    kubeProxy:
      mode: IPTables
    version: 1.20.0
  maintenance:
    autoUpdate:
      kubernetesVersion: true
      machineImageVersion: true
    timeWindow:
      begin: 220000+0100
      end: 230000+0100
```

ialidzhikov:
Well, I reproduced it again on new Shoot creation (I previously also reproduced with new Shoot creation).
I think it is related to `o.Shoot.Info.UID` that returns `""` when it is used for computing the backupentry name.
As far as I see we call `initializeOperation` and after that we call `TryUpdateShootStatus` that sets the shoot .status.uid but probably this UID is not set back to  `o.Shoot.Info`? Ref https://github.com/gardener/gardener/blob/v1.15.0/pkg/gardenlet/controller/shoot/shoot_control.go#L463-L479

ialidzhikov (approved):
/lgtm

/cc @plkokanov @kris94 @swilen-iwanow for second opinion

PR #3319, opened 2020-12-17
Extend error code mapping

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal

With this PR the following two errors are being handled as `ERR_INFRA_DEPENDENCIES` error:

```
* Error deleting Route Table "worker_route_table" (Resource Group "shoot--foo--bar"): network.RouteTablesClient#Delete: Failure sending request: StatusCode=400 -- Original Error: Code="InUseRouteTableCannotBeDeleted" Message="Route table worker_route_table is in use and cannot be deleted." Details=[]
```

```
* A resource with the ID "/subscriptions/<omitted>/resourceGroups/shoot--foo--bar" already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for "azurerm_resource_group" for more information.
```

**Release note**:

```other operator
NONE
```


timuthy (approved):
/lgtm

danielfoehrKn (approved):

PR #3320, opened 2020-12-17
Ensure stable order of extension webhooks when registering

timebertt:

/area quality
/kind bug
/priority normal

Before this PR, the extension webhook self-registration was indeterministic wrt the order of webhooks in a single `MutatingWebhookConfig`. Thus, it could happen that the webhooks were registered again with a different order, when an extension controller restarts (wich happens quite frequently with rollouts, VPA, etc.).
With this, mutations of extension webhooks (e.g. controlplane and controlplaneexposures) were sometimes applied in different order, resulting for example in changed order of command line flags in the kube-apiserver deployment, although there was no semantical change.
This behaviour often led to unnecessary rollouts of control plane components, which we generally want to avoid.

Fixes #

/cc @rfranzke 

This bug can be reproduced by starting e.g. provider-azure multiple times in a row and observing the order of the mutating webhook config.

**Release note**:

```bugfix dependency
Ensure a stable order of self-registered webhooks in extensions to avoid unnecessary rollouts of control plane components.
```


timuthy (approved):
/lgtm

rfranzke (approved):

PR #3326, opened 2020-12-18
Write updated kubeconfig to testmachinery output

schrodit:

/area testing
/kind bug
/priority normal

When the "kubeconfig rotate" tesst is executed, the kubeconfig that was downloaded during its creation is invalidated.
So subsequent steps that use that config are unable to connect to the shoot.
This PR updates the kubeconfig if it was rotated by the test.

The kubeconfig is only updated (written to the filesystem) if the testmachinery environment variable is set.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/test/framework/common.go
+++ b/test/framework/common.go
@@ -33,6 +33,10 @@ const (
 
 	// WorkerNamePrefix is the default prefix that will be used for Shoot workers
 	WorkerNamePrefix = "worker-"
+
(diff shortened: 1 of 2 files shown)


ialidzhikov (approved):
/lgtm

Thank you @dguendisch and @schrodit !

dguendisch (approved):
/lgtm

PR #3331, opened 2020-12-22
Fix ./hack/generate-controller-registration.sh

ialidzhikov:
/kind bug

We missed https://helm.sh/blog/new-location-stable-incubator-charts/. `https://kubernetes-charts.storage.googleapis.com` is replaced by `https://charts.helm.sh/stable`. helm < v2.17.0 still use the old location and `helm init` with such versions fail.

`helm init` fails with:

```
Adding stable repo with URL: https://kubernetes-charts.storage.googleapis.com 
Error: error initializing: Looks like "https://kubernetes-charts.storage.googleapis.com" is not a valid chart repository or cannot be reached: Failed to fetch https://kubernetes-charts.storage.googleapis.com/index.yaml : 403 Forbidden
```

Our `hack/generate-controller-registration.sh` is using `helm init` and this is causing `make generate` and builds for the extensions to fail.

Ref https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-extension-provider-azure-master/jobs/master-head-update-job/builds/30

**Release note**:

```other operator
NONE
```

### DIFF
--- a/hack/install-requirements.sh
+++ b/hack/install-requirements.sh
@@ -22,7 +22,7 @@ GO111MODULE=off go get golang.org/x/tools/cmd/goimports
 
 export GO111MODULE=on
 curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.27.0
-curl -s "https://raw.githubusercontent.com/helm/helm/v2.16.9/scripts/get" | bash -s -- --version 'v2.13.1'
+curl -s "https://raw.githubusercontent.com/helm/helm/v2.17.0/scripts/get" | bash -s -- --version 'v2.17.0'
 
 if [[ "$(uname -s)" == *"Darwin"* ]]; then
   cat <<EOM


timebertt (approved):
LGTM as such.

Should we additionally consider, adding `--skip-repos` to `helm init` ([docs](https://github.com/helm/helm/blob/a03d3e34ac4fc9f31d77647c6273f6b7da35d145/docs/helm/helm_init.md#options))? I guess, we don't need the repos for the ctrlreg gen, right?

ialidzhikov:
> Should we additionally consider, adding `--skip-repos` to `helm init` ([docs](https://github.com/helm/helm/blob/a03d3e34ac4fc9f31d77647c6273f6b7da35d145/docs/helm/helm_init.md#options))? I guess, we don't need the repos for the ctrlreg gen, right?

I tried to add the `--skip-repos` flag but `helm package` failed afterwards with:

```
Error: open /var/folders/p9/6dr9snnx6ws1dxkj4b66st6r0000gn/T/tmp.ALk2amzqRP/repository/local/index.yaml: no such file or directory
```

Looks like `helm package` needs the local repo.

timebertt:
Alright, thanks for checking, @ialidzhikov!

PR #3332, opened 2020-12-24
Enable autoscaler during rolling-update

### LINKED ISSUE #3193: Enable Cluster-Autoscaler during Rollouts.

/area auto-scaling
/area high-availability
/kind enhancement
/priority normal

Currently, we disable the cluster-autoscaler during rolling-updates of the cluster. With the growing size of the clusters, the roll-update could take a long time, sometimes even a couple of hours. In this case, the newly introduced workload would not be able to find space in the absence of autoscaling.

With this change https://github.com/gardener/machine-controller-manager/pull/496  at MCM, all the nodes recieve the annotation of [scale-down-disabled](https://github.com/gardener/autoscaler/blob/machine-controller-manager-provider/cluster-autoscaler/FAQ.md#how-can-i-prevent-cluster-autoscaler-from-scaling-down-a-particular-node), during the roll-out. This prevents autoscaler from scaling-down the worker-group, while still allows adding new nodes to settle the fresh workload.

Given the change above, gardener could keep autoscaler running even during the roll-outs, to allow scale-ups.

**Why is this needed**: For better availability of the workload during cluster rollouts.

hardikdr:

/area control-plane
/kind enhancement
/priority normal

**What this PR does / why we need it**: This PR enables the autoscaler during rolling-updates of the shoot. 
I couldn't find a place to add a test for this case, would be great to have quick guidance.

Fixes #3193 

**Release note**:

```feature user
The `cluster-autoscaler` is now activated even during rolling-update of the shoot clusters. The change in `machine-controller-manager` of adding the `cluster-autoscaler.kubernetes.io/scale-down-disabled` annotation during rolling-update is required, in order for autoscaler to not scale-down worker-pools (coming with machine-controller-manager `0.34.0`). 
```
```action operator
With the activated `cluster-autoscaler` during roll-outs, following are the minimum versions required for different provider-extensions: gardener-extension-provider-aws `v1.16.0`, gardener-extension-provider-openstack `v1.12.0`,  gardener-extension-provider-azure `v1.14.0`, gardener-extension-provider-gcp `v1.12.0`, gardener-extension-provider-alicloud `v1.18.0`, gardener-extension-provider-vsphere `v0.1.0`.
```

### DIFF
--- a/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
@@ -340,7 +340,6 @@ func (a *genericActuator) deployMachineDeployments(ctx context.Context, logger l
 // available by the machine-controller-manager. It polls the status every 5 seconds.
 func (a *genericActuator) waitUntilWantedMachineDeploymentsAvailable(ctx context.Context, logger logr.Logger, cluster *extensionscontroller.Cluster, worker *extensionsv1alpha1.Worker, alreadyExistingMachineDeploymentNames sets.String, alreadyExistingMachineClassNames sets.String, wantedMachineDeployments extensionsworker.MachineDeployments, clusterAutoscalerUsed bool) error {
 	logger.Info("Waiting until wanted machine deployments are available")
-	autoscalerIsScaledDown := false
 	workerStatusUpdatedForRollingUpdate := false
 
 	return retryutils.UntilTimeout(ctx, 5*time.Second, 5*time.Minute, func(ctx context.Context) (bool, error) {
@@ -425,14 +424,6 @@ func (a *genericActuator) waitUntilWantedMachineDeploymentsAvailable(ctx context
 			// numUnavailable == 0 makes sure that every machine joined the cluster (during creation & in the case of a rolling update with maxUnavailability > 0)
 			if numUnavailable == 0 && numUpdated == numberOfAwakeMachines && int(numHealthyDeployments) == len(wantedMachineDeployments) {
 				return retryutils.Ok()
-			}
-
-			// scale down cluster autoscaler during creation or rolling update
-			if clusterAutoscalerUsed && !autoscalerIsScaledDown {
-				if err := a.scaleClusterAutoscaler(ctx, logger, worker, 0); err != nil {
-					return retryutils.SevereError(err)
-				}
-				autoscalerIsScaledDown = true
 			}
 
 			// update worker status with condition that indicates an ongoing rolling update operation
--- a/pkg/operation/botanist/health_check.go
+++ b/pkg/operation/botanist/health_check.go
@@ -277,6 +277,8 @@ func computeRequiredControlPlaneDeployments(
 			return nil, err
 		}
 
+		// TODO: This check can be removed after few releases, as the cluster-autoscaler is now enabled even
+		// during the rolling-update. Related change: https://github.com/gardener/gardener/pull/3332
 		// if worker resource is processing (during maintenance), there might be a rolling update in progress
 		// during rolling updates, the autoscaler deployment is scaled down & therefore not required
 		rollingUpdateMightBeOngoing := false


rfranzke:
Thanks for the PR @hardikdr!

> I couldn't find a place to add a test for this case, would be great to have quick guidance.

We don't have unit tests for the generic actuator, it's still on our long to-do list

rfranzke:
@hardikdr as this change is dependent on a specific MCM version (which?) that has https://github.com/gardener/machine-controller-manager/pull/496, can you please clarify in the release note which minimum MCM version (controlled by the provider extensions) is required for everything to function properly?

hardikdr:
>can you please clarify in the release note which minimum MCM version (controlled by the provider extensions) is required for everything to function properly? 

Sure, that's a good point,  I adapted the release-note.

extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
```diff
@@ -427,8 +428,10 @@ func (a *genericActuator) waitUntilWantedMachineDeploymentsAvailable(ctx context
 				return retryutils.Ok()
 			}
 
-			// scale down cluster autoscaler during creation or rolling update
-			if clusterAutoscalerUsed && !autoscalerIsScaledDown {
+			operationType = gardencorev1beta1helper.ComputeOperationType(worker.ObjectMeta, worker.Status.LastOperation)
+
+			// scale down cluster autoscaler during creation.
+			if clusterAutoscalerUsed && !autoscalerIsScaledDown && operationType == gardencorev1beta1.LastOperationTypeCreate {
```
rfranzke: It is not necessary to scale down CA during creation because it is not even deployed yet (it will only be deployed after the machines came up). Hence, can we remove all the scale down code?

hardikdr: Hm sure. I removed that part of the code which brings it down. I'll check the failure of `verify` step in ci.

ialidzhikov: > Hm sure. I removed that part of the code which brings it down. I'll check the failure of verify step in ci.

You need to rebase to adopt https://github.com/gardener/gardener/pull/3347.

hardikdr: I see, thank you @ialidzhikov, I just rebased it.

rfranzke (dismissed):
/lgtm

ialidzhikov (changes requested):
Doesn't the health_check.go need an adaptation?

Currently there is a logic that does not health check the cluster-autoscaler if the status.lastOperation.state=Processing. This logic should be no longer needed with this change, right?

Ref https://github.com/gardener/gardener/blob/v1.15.0/pkg/operation/botanist/health_check.go#L280-L292

rfranzke:
@ialidzhikov Let's wait with this for a couple of releases to give all provider extensions time to adopt the change introduced in the generic worker actuator. Otherwise, we will have compatibility issues and false negative health checks. But it certainly makes sense to have a comment for such follow-up for the respective code in the health check logic, thanks for pointing this out.

ialidzhikov:
Sure, I missed that the change is in the generic actuator. Let's add a TODO for now that reminds us to cleanup the corresponding logic after several releases.

hardikdr:
Thanks, @ialidzhikov , that's a good catch. What's the convention for adding TODO, should I add it in the code you pointed to or? 
- https://github.com/gardener/gardener/blob/v1.15.0/pkg/operation/botanist/health_check.go#L280-L292

ialidzhikov:
> Thanks, @ialidzhikov , that's a good catch. What's the convention for adding TODO, should I add it in the code you pointed to or?
> 
> * https://github.com/gardener/gardener/blob/v1.15.0/pkg/operation/botanist/health_check.go#L280-L292

Yep, you can add a TODO above the corresponding code fragment stating that it can be removed after several releases and writing down a short explanation.

hardikdr:
Sure, I added a comment.

ialidzhikov (approved):
/lgtm

rfranzke (approved):

PR #3333, opened 2020-12-24
Reduce the initialDelaySeconds to 80 seconds for the Loki readinessProbe

vlvasilev:

/area logging
/kind enhancement
/priority normal

Whit this PR we reduce the Loki readinessProbe `initialDelaySeconds` from 120 to 80 seconds.
This is because the current control plane health controller check the pods every two minutes which is the initial delay of the Loki readiness probe. With the initial delay of 80 seconds the readiness probe is started before the health controller.
The `failureThreshold` is increased with 4 times to preserve the 2 minutes delay plus 30 second of retries.

Fixes #

**Release note**:

```other operator
The Loki `initialDelaySeconds` for the `readinessProbe` is reduces to 80 seconds.
```

### DIFF
--- a/charts/seed-bootstrap/charts/loki/values.yaml
+++ b/charts/seed-bootstrap/charts/loki/values.yaml
@@ -33,12 +33,14 @@ livenessProbe:
     path: /ready
     port: metrics
   initialDelaySeconds: 120
+  failureThreshold: 5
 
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm

PR #3340, opened 2020-12-28
Enhance remote garden setup

danielfoehrKn:

/area dev-productivity
/kind enhancement
/priority normal

Enables the remote garden setup to work
 - outside of Docker for Desktop (e.g Linux)
 - on Garden K8s clusters with DNS hostname for the LB (e.g on AWS)
 - adding annotation for NLB on  AWS (required because ELB classic cannot have UDP ports)
   - annotation is always added (should not make a problem on other clusters - will simply be ignored)

Also fixes a bug trying to remove a non-existing certificate file.

Fixes #

Using this for testing [managed seed PR](https://github.com/gardener/gardener/pull/3177)

**Release note**:

```other operator

```

### DIFF
--- a/hack/local-development/remote-garden/generate-certs
+++ b/hack/local-development/remote-garden/generate-certs
@@ -64,5 +64,8 @@ openssl x509 -req -in "$CERTS_DIR/$CLIENT_NAME.csr" -CA "$CERTS_DIR/$CA_NAME.crt
 
 # Clean up after we're done.
 rm "$CERTS_DIR/$SERVER_NAME.csr" "$CERTS_DIR/$CLIENT_NAME.csr"
-rm "$CERTS_DIR/$CA_NAME.srl"
 rm "$CERTS_DIR/$SERVER_NAME.conf" "$CERTS_DIR/$CLIENT_NAME.conf"
+
+if [ -f "$CERTS_DIR/$CA_NAME.srl" ]; then
+    rm "$CERTS_DIR/$CA_NAME.srl"
+fi
--- a/hack/local-development/remote-garden/open-gardener-tunnels
+++ b/hack/local-development/remote-garden/open-gardener-tunnels
@@ -53,6 +53,8 @@ kind: Service
 metadata:
   name: quic-lb
   namespace: garden
+  annotations:
+    service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
 spec:
   type: LoadBalancer
   selector:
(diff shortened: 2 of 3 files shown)


hack/local-development/remote-garden/generate-certs
```diff
@@ -64,5 +64,4 @@ openssl x509 -req -in "$CERTS_DIR/$CLIENT_NAME.csr" -CA "$CERTS_DIR/$CA_NAME.crt
 
 # Clean up after we're done.
 rm "$CERTS_DIR/$SERVER_NAME.csr" "$CERTS_DIR/$CLIENT_NAME.csr"
-rm "$CERTS_DIR/$CA_NAME.srl"
```
stoyanr, line 67: Why is this line removed? At least on my setup, such a file is created by `openssl` in one of the previous steps, and without this line it is still there in `$CERTS_DIR` after everything should have been cleaned up. If the behavior of `openssl` is different and this breaks on your side, please add a check if the file exists before removeing it.

hack/local-development/remote-garden/open-gardener-tunnels
```diff
@@ -263,11 +277,38 @@ echo "Applying gardener services..."
...
+  -v "$CERTS_DIR":/certs \
+  $QUIC_CLIENT_IMAGE \
   --server="$loadbalancerHostname:$APISERVER_LB_PORT" \
-  --upstream="host.docker.internal:$API_SERVER_SECURE_PORT" \
-  --ca-file=/certs/$CA_NAME.crt --cert-file=/certs/$CLIENT_NAME.crt --cert-key=/certs/$CLIENT_NAME.key
-docker run -d --name $ADMISSION_CONTROLLER_SERVICE_NAME-quic-client -l "$LABEL" --network host --rm -v "$CERTS_DIR":/certs $QUIC_CLIENT_IMAGE \
+  --upstream="$hostname:$API_SERVER_SECURE_PORT" \
+  --ca-file=/certs/$CA_NAME.crt \
+  --cert-file=/certs/$CLIENT_NAME.crt  \
+  --cert-key=/certs/$CLIENT_NAME.key
```
stoyanr, line 302: Is there any particular reason you removed the `--rm` flag? I think it should better stay to ensure we don't leave hanging containers, also it's used in similar `docker run` commands in `local-garden`

stoyanr (changes requested):
Nice PR, all improvements make a lot of sense. I tested on Mac and in general it works. I have only  minor suggestions, see the comments.

hack/local-development/remote-garden/open-gardener-tunnels
danielfoehrKn: Thanks - you are right. I removed the flag for testing to see container logs but forgot to add it again.

danielfoehrKn: added it again

hack/local-development/remote-garden/generate-certs
danielfoehrKn: thanks

stoyanr (approved):

rfranzke (approved):

PR #3341, opened 2020-12-29
Owner reference for extension webhooks

rfranzke:

/area quality
/kind technical-debt
/priority normal

Earlier, there were no means for auto-cleanup of extension webhooks. With this PR, when a namespace was provided via the `--webhook-config-namespace` flag, we add an owner reference for the webhook configuration (owner=namespace).
This will make the garbage collector to auto-delete the webhook config after the namespace is deleted, i.e., after the extension has been removed from a (seed) cluster.

Please note that this does usually not work for local development cases because here the "url" mode is used and no namespace is provided.

**Release note**:

```noteworthy dependency
During the extension webhook registration, when a namespace is provided via the `--webhook-config-namespace` flag, the webhook config is enhanced with an owner reference pointing to the provided namespace. This will lead to auto-cleanup of the webhook config when the extension is uninstalled from a seed (earlier, the webhook config was orphaned even after uninstallation from a seed).
```

### DIFF
--- a/extensions/pkg/webhook/registration.go
+++ b/extensions/pkg/webhook/registration.go
@@ -19,10 +19,15 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/gardener/gardener/pkg/utils/kubernetes"
+
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


danielfoehrKn (approved):
/lgtm

PR #3348, opened 2021-01-04
Webhook for Cloudprovider secret

kon-angelo:

/area control-plane
/kind enhancement
/priority normal

This PR adds a new mutating webhook in the extensions library that allows modification of the cloudprovider secret. Since extensions will use the cloudprovider secret reference directly in the machineclasses, extension providers can use this webhook to enhance the secret with additional data that will become available to MCM.

**Release note**:

```breaking developer
A new mutating webhook for the `cloudprovider` secret has been added in the extensions library. With this change, the `EnsurerContext` of the `genericmutator` package has been moved to a separate `context` package. Please adapt your usage of `genericmutator` accordingly.
```


extensions/pkg/webhook/cloudprovider/cloudprovider.go
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke, line 1: 2021 - nice :-D

extensions/pkg/webhook/cloudprovider/cloudprovider.go
```diff
@@ -0,0 +1,80 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	WebhookName = "cloudprovider"
```
rfranzke, line 32: doc string

extensions/pkg/webhook/cloudprovider/cloudprovider.go
```diff
@@ -0,0 +1,80 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cloudprovider
+
+import (
+	"github.com/gardener/gardener/extensions/pkg/webhook"
+	extensionswebhook "github.com/gardener/gardener/extensions/pkg/webhook"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	corev1 "k8s.io/api/core/v1"
```
rfranzke, line 22:
```suggestion

	corev1 "k8s.io/api/core/v1"
```

extensions/pkg/webhook/cloudprovider/mutator.go
```diff
@@ -0,0 +1,103 @@
...
+package cloudprovider
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/gardener/gardener/extensions/pkg/webhook"
+	"github.com/gardener/gardener/extensions/pkg/webhook/controlplane/genericmutator"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	"github.com/go-logr/logr"
```
rfranzke, line 25:
```suggestion

	"github.com/go-logr/logr"
```

extensions/pkg/webhook/cloudprovider/mutator.go
```diff
@@ -0,0 +1,103 @@
...
+		if !ok {
+			return fmt.Errorf("could not mutate: old object could not be casted to type %q", "Secret")
+		}
+	}
+
+	o, ok := new.(metav1.Object)
+	if !ok {
+		return fmt.Errorf("could not cast runtime object to %q", "metav1.Object")
+	}
+	etcx := genericmutator.NewEnsurerContext(m.client, o)
```
rfranzke: Can you move this `EnsurerContext` out from the `genericmutator` into some other package? It's not specific to the generic controlplane mutator anymore now.

extensions/pkg/webhook/cloudprovider/mutator_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/gardener/gardener/extensions/pkg/webhook"
+	"github.com/gardener/gardener/extensions/pkg/webhook/cloudprovider"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	mockcloudprovider "github.com/gardener/gardener/pkg/mock/gardener/extensions/webhook/cloudprovider"
+	"github.com/golang/mock/gomock"
```
rfranzke, line 26:
```suggestion

	"github.com/golang/mock/gomock"
```

cosmetics :) 💇🏼‍♀️ 💅🏼

rfranzke (changes requested):
Great PR, thank you! Just a few nits :)

extensions/pkg/webhook/cloudprovider/mutator.go
kon-angelo: Alternatively, maybe duplicate it locally in the package ? They are the same because I need to fetch the `Cluster` but this is a context specific to this admission webhook, so the similarity is "coincidental".

extensions/pkg/webhook/cloudprovider/cloudprovider.go
kon-angelo: I assume for this and the rest similar comments that the gardener packages should form a group before the rest of the imports as a convention ?

extensions/pkg/webhook/cloudprovider/mutator.go
rfranzke: Hm, I see what you mean, however, does this really bring any meaningful benefit? Both mutators need this context, and it seems unlikely (as of today, at least) that they will be different in the future. For now, I think duplicating it would not help in anyway. As long as we can share the code we should do it to reduce the maintenance efforts. If it becomes necessary to have dedicated contexts in the future then we are not blocked and can still do it.

extensions/pkg/webhook/cloudprovider/cloudprovider.go
rfranzke: Just cosmetics, I don't insist on this, still wanted to point it out because we often "follow this convention", but you can also keep it as is if you want. Some time back we agreed on accepting everything that is accepted by `goimports`, so please consider those comments optional.

rfranzke:
Can you check the failing `verify` step, please?

kon-angelo:
@rfranzke with this change (moving the `EnsurerContext`) the extensions will need to refactor some of their code for their webhooks to match the new package. I guess that means there should be some kind of release note to serve as a warning even though g/g itself is unaffected right  ?

extensions/pkg/webhook/cloudprovider/cloudprovider.go
```diff
@@ -0,0 +1,82 @@
...
+	types := []runtime.Object{&corev1.Secret{}}
+	handler, err := webhook.NewBuilder(mgr, logger).WithMutator(args.Mutator, types...).Build()
+	if err != nil {
+		return nil, err
+	}
+
+	namespaceSelector := buildSelector(args.Provider)
+	logger.Info("Creating webhook")
+
+	var (
```
dkistner: Nit. One line.

dkistner (commented):
Very nice PR. Thank you.
I would suggest to add a release note targeting developers to make them aware to implement this hook.

extensions/pkg/webhook/cloudprovider/cloudprovider.go
```diff
@@ -0,0 +1,82 @@
...
+		name = WebhookName
+	)
```
rfranzke:
```suggestion
```

Seems to be superfluous

rfranzke (approved):
/lgtm

Thank you!

PR #3349, opened 2021-01-04
Terraformer execution adopts still running pods

rfranzke:

/area ops-productivity
/kind enhancement technical-debt
/priority normal

This PR adapts the Terraformer library in a way that it does no longer delete still running Terraform pods from previous runs but it adopts them instead. This makes status/result propagation a lot faster and prevents unhelpful error message like "Terraform pod is still running".

Along the way, I cleaned up some technical debt that was left from the switch from a Terraform `Job` to a `Pod` some months back. Concretely, a few function signatures are adapted. Also, a few constants and function names were adapted to prevent stuttering according to the Golang conventions.

**Release note**:

```noteworthy dependency
The `terraformer` library in `extensions/pkg/terraformer` does now adopt still running pods instead of deleting them and waiting for their deletion. This enables a faster and more reliable status/result propagation and prevents unhelpful error messages. `Infrastructure` extension developers do not need to wait for a clean Terraform environment in their `Delete()` function anymore as the library is now handling this case out-of-the-box.
```

### DIFF
--- a/extensions/pkg/terraformer/terraformer.go
+++ b/extensions/pkg/terraformer/terraformer.go
@@ -35,15 +35,15 @@ import (
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
-	kutils "github.com/gardener/gardener/pkg/utils/kubernetes"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	"github.com/gardener/gardener/pkg/utils/retry"
 )
 
 const (
-	// TerraformerLabelKeyName is a key for label on a Terraformer Pod indicating the Terraformer name.
-	TerraformerLabelKeyName = "terraformer.gardener.cloud/name"
-	// TerraformerLabelKeyPurpose is a key for label on a Terraformer Pod indicating the Terraformer purpose.
-	TerraformerLabelKeyPurpose = "terraformer.gardener.cloud/purpose"
+	// LabelKeyName is a key for label on a Terraformer Pod indicating the Terraformer name.
+	LabelKeyName = "terraformer.gardener.cloud/name"
+	// LabelKeyPurpose is a key for label on a Terraformer Pod indicating the Terraformer purpose.
+	LabelKeyPurpose = "terraformer.gardener.cloud/purpose"
 )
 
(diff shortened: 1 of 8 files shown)


timuthy (commented):
One thought, LGTM otherwise:
Now that we adopt the Terraformer pod for a longer period, do you think it makes sense to also consider a change in TF config, i.e. deploy a new pod if TF config has changed?

rfranzke:
@timuthy This would happen automatically after the current TF execution has finished, or? Do you ask about aborting the current TF run instead, waiting, and then deploying a new one (instead of just waiting and then deploying a new one)?

timuthy:
> Do you ask about aborting the current TF run instead, waiting, and then deploying a new one (instead of just waiting and then deploying a new one)?

Yes, I was thinking about if we benefit from such a mechanism in case provider extensions provide a fix for TF config.

rfranzke:
Hm, probably not. My understanding is that we would start a new TF pod anyways if the current execution fails, and we can also not instantly abort a running TF pod (TF itself needs some time to gracefully terminate).

timuthy (dismissed):

timuthy:
@rfranzke can you resolve the conflict? We can then merge this PR from my point of view.
/lgtm

rfranzke:
@timuthy done

extensions/pkg/terraformer/terraformer.go
```diff
@@ -144,118 +147,141 @@ func (t *terraformer) Destroy(ctx context.Context) error {
...
+		pod, err = t.deployTerraformerPod(ctx, generateName, command)
 		if err != nil {
-			return fmt.Errorf("failed to deploy the Terraformer Pod with .meta.generateName %q: %w", generateName, err)
+			return fmt.Errorf("failed to deploy the Terraformer pod with .meta.generateName %q: %w", generateName, err)
 		}
+		logger.Info("Successfully created Terraformer pod", "pod", kutil.KeyFromObject(pod))
+	}
 
-		logger.Info("Successfully created Terraformer Pod", "pod", kutils.KeyFromObject(pod))
+	if pod != nil {
```
timuthy, line 233: Can this condition even be `false`?

rfranzke: It can be nil if there is no running pod and the if state is already empty.

timuthy (dismissed):
/lgtm

vpnachev (approved):
/lgtm

rfranzke:
/unhold
It seems there were no conflicts with #3393 - cc @timebertt 
cc @vpnachev @timuthy

PR #3231, opened 2020-11-26
Prepare for label key switch in the health check selector

ialidzhikov:
**How to categorize this PR?**

/kind cleanup
/priority normal

**What this PR does / why we need it**:
This PR adds the `gardener.cloud/role` label key to the controlplane components. This will allows us to switch the health check label selector to use `gardener.cloud/role` in a future version.

**Release note**:

```improvement operator
NONE
```

PR #3350, opened 2021-01-05
Switch the HealthChecker to use the 'gardener.cloud/role' label key

ialidzhikov:
/kind cleanup

This PR:
- adapts the HealthChecker to use the `gardener.cloud/role` label key to select controlplane components (Deployments, StatefulSets, Etcds) for health check. This is follow-up on https://github.com/gardener/gardener/pull/3231.
- removes the `garden.sapcloud.io/role` label key from most of the controlplane components as it is no longer used for health checks.
- contains also some minor cleanups

Part of #1649 

**Release note**:

```breaking operator
gardenlet's shoot-care-control is now using the `gardener.cloud/role` label key (until now it was `garden.sapcloud.io/role`) to perform health checks on controlplane components. Make sure you have first upgraded to at least Gardener v1.14 before you upgrade to this version of Gardener.
```

### DIFF
--- a/charts/seed-bootstrap/charts/loki/values.yaml
+++ b/charts/seed-bootstrap/charts/loki/values.yaml
@@ -12,7 +12,7 @@ authEnabled: true
 annotations: {}
 
 labels:
-  # deprecated label only for the health check
+  # deprecated label
   garden.sapcloud.io/role: logging
   gardener.cloud/role: logging
   app: loki
--- a/charts/seed-controlplane/charts/gardener-resource-manager/templates/deployment.yaml
+++ b/charts/seed-controlplane/charts/gardener-resource-manager/templates/deployment.yaml
@@ -6,7 +6,6 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     gardener.cloud/role: controlplane
-    garden.sapcloud.io/role: controlplane
     app: gardener-resource-manager
 spec:
   revisionHistoryLimit: 0
--- a/charts/seed-controlplane/charts/gardener-resource-manager/templates/rbac.yaml
+++ b/charts/seed-controlplane/charts/gardener-resource-manager/templates/rbac.yaml
@@ -5,7 +5,7 @@ metadata:
   name: gardener-resource-manager
   namespace: {{ .Release.Namespace }}
   labels:
-    garden.sapcloud.io/role: controlplane
+    gardener.cloud/role: controlplane
     app: gardener-resource-manager
 rules:
 - apiGroups:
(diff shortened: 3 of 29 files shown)


charts/seed-bootstrap/charts/loki/values.yaml
```diff
@@ -12,7 +12,7 @@ authEnabled: true
 annotations: {}
 
 labels:
-  # deprecated label only for the health check
+  # deprecated label
   garden.sapcloud.io/role: logging
```
rfranzke, line 16: Why is this not yet removed?

rfranzke (commented):
looks great, just a small question

charts/seed-bootstrap/charts/loki/values.yaml
ialidzhikov: Well, the `labels` are also used in the loki's Statefulset `spec.selector` which is immutable field. To workaround a similar issue with immutability of `spec.selector` for Deployment in https://github.com/gardener/gardener/pull/3264 I introduced a helper func that deletes the Deployment to allow new Deployment to be create right away with the desired spec. The same approach I think is not applicable for StatefulSets as they bring a PVC which we don't want to delete.

rfranzke: When a `StatefulSet` is deleted then the PVC remains in the system (at least not with our configuration), so I guess we can follow the same approach. We can do that separately, of course, so I'm going to approve the PR for now. Thanks!

rfranzke (dismissed):
/lgtm

charts/seed-bootstrap/charts/loki/values.yaml
ialidzhikov: Okay, I will give a try. Thank you!

ialidzhikov: I now updated the PR. Thank you once again!

rfranzke (dismissed):
/lgtm

pkg/operation/botanist/logging.go
```diff
@@ -74,5 +74,13 @@ func (b *Botanist) DeploySeedLogging(ctx context.Context) error {
 		}
 	}
 
+	// TODO (ialidzhikov): remove in a future version
+	stsKeys := []client.ObjectKey{
+		kutil.Key(b.Shoot.SeedNamespace, "loki"),
+	}
+	if err := common.DeleteStatefulSetsHavingDeprecatedRoleLabelKey(ctx, b.K8sSeedClient.Client(), stsKeys); err != nil {
```
timuthy: Aren't we required to wait for the STS to be gone before continuing?

ialidzhikov: I reverted this STS migration logic for now. I will tackle it in another PR as suggested initially.

rfranzke (approved):

timuthy (approved):

PR #3353, opened 2021-01-05
Evaluate and return kube-apiserver logs if deployment does not become ready

rfranzke:

/area ops-productivity usability
/kind enhancement
/priority normal

With this PR we are reading the most recent complete logs of the newest `kube-apiserver` pod of a shoot in case its owning deployment does not become ready.
This reduces ops effort because it allows a) classification of the error and b) the end-user to figure out the problem (faster) on his own.

**Special notes for your reviewer(s):**
I have split the PR into separate comments to make the review a little easier. Please
/squash
when merging.

**Release note**:

```noteworthy user
If the `kube-apiserver` deployment of a shoot cluster does not become ready then Gardener does now evaluate and return the most recent complete logs of the newest pod. It will be part of the `.status.last{Operation,Errors}`. As a consequence, start-up failures due to invalid feature gate or runtime config settings is being treated as configuration problem error.
```


pkg/operation/botanist/waiter.go
```diff
@@ -74,35 +76,58 @@ func (b *Botanist) WaitUntilKubeAPIServerIsDeleted(ctx context.Context) error {
...
+		}
+		if newestPod == nil {
+			return err
+		}
+
+		logs, err2 := kutil.MostRecentCompleteLogs(ctx, b.K8sSeedClient.Kubernetes().CoreV1().Pods(newestPod.Namespace), newestPod, "kube-apiserver", pointer.Int64Ptr(10))
+		if err2 != nil {
+			return err2
+		}
+
```
timuthy: Can we always return `err` and just log `err2`? Reading the logs should happen at best effort and hiding `err` by `err2` might be confusing for users/operators.

pkg/operation/botanist/waiter.go
```diff
@@ -74,35 +76,58 @@ func (b *Botanist) WaitUntilKubeAPIServerIsDeleted(ctx context.Context) error {
...
+		newestPod, err2 := kutil.NewestPodForDeployment(ctx, b.K8sSeedClient.DirectClient(), deployment)
+		if err2 != nil {
+			return err2
+		}
+		if newestPod == nil {
+			return err
+		}
+
+		logs, err2 := kutil.MostRecentCompleteLogs(ctx, b.K8sSeedClient.Kubernetes().CoreV1().Pods(newestPod.Namespace), newestPod, "kube-apiserver", pointer.Int64Ptr(10))
```
timuthy, line 122: I was going to write how you'd know that the `kube-apiserver` container is the villain here?! But I guess there is no reliant alternative to figure that out, right?

timuthy (commented):
Looks good to me, just two nits.

pkg/operation/botanist/waiter.go
rfranzke: Yeah, there could probably be a lot of other cases that can result in the kube-apiserver deployment not becoming ready (maybe the seed is full and the new KAPI pod is just pending, ...). Fetching and providing the logs is just "best-effort" but it's not guaranteed that you'll find the root cause of the not-ready problem here.

pkg/operation/botanist/waiter.go
rfranzke: Makes sense, I've wrapped `err` with the information in `err2` in f589ade

pkg/utils/kubernetes/kubernetes.go
```diff
@@ -383,3 +399,147 @@ func MergeOwnerReferences(references []metav1.OwnerReference, newReferences ...m
 
 	return references
 }
+
+// OwnedBy checks if the given object's owner reference contain an entry with the provided attributes.
```
timuthy:
```suggestion
// OwnedBy checks if the given object's owner reference contains an entry with the provided attributes.
```

pkg/utils/kubernetes/kubernetes.go
```diff
@@ -383,3 +399,147 @@ func MergeOwnerReferences(references []metav1.OwnerReference, newReferences ...m
...
+		if err := meta.EachListItem(listObj, func(obj runtime.Object) error {
+			if filterFn(obj) {
+				items = append(items, obj)
+			}
+			return nil
+		}); err != nil {
+			return nil, err
+		}
+
+		if err := meta.SetList(listObj, items); err != nil {
```
timuthy, line 445: Nit, deep copy list before mutating it further. Mutating the incoming object directly might be unexpected for the caller of this function.

rfranzke: Hm, why? It's already being "mutated" because we run `List()` and store the results in `listObj`. 
Now, if the user provides a `filterFn` then we filter the objects accordingly and update `listObj`. This is exactly what I would expect as a user when I provide a filter function, similarly to how I expect to find all objects in `listObj` when I call `List()` only.

timuthy: > similarly to how I expect to find all objects in listObj when I call List() only.

The `List()` function explicitly says that it takes the `listObj` to do that.

In this case, you have a `NewestObject` function that is devoted to find the newest object only, but as a byproduct it also mutates the passed `listObj`. This is unexpected from my point of view.

rfranzke: Let me add a doc string to clarify on this then. I explicitly want this behaviour so that the user can also run additional logic against the resulting items without the necessity of manually applying his filter function again to get rid of the undesired objects.

timuthy: Okay, thanks for clarifying. Can you also verify the `listObj` in the unit tests?

rfranzke: 👍🏻 done

timuthy (dismissed):
/lgtm

rfranzke:
Had to rebase to resolve some merge conflicts.

timuthy (approved):

Issue #3109, opened 2020-11-04
☂️-Issue for upgrading to k/*@v0.19 and controller-runtime@v0.7


timebertt:
**How to categorize this issue?**

/area quality robustness open-source
/kind enhancement
/priority normal

**What would you like to be added**:

[controller-runtime v0.7](https://github.com/kubernetes-sigs/controller-runtime/releases/v0.7.0) was released. We can now go ahead our dependencies to `k/*@v0.19` and `c-r@v0.7.0`.

This issue tracks all the tasks, that have to be done and should be followed up on, as the release is quite big.

**Why is this needed**:
We should adopt the newest improvements from upstream and shouldn't miss anything important during the migration 😄 

| | Status |
| --- | --- |
| ✅| completed |
| ☑️ | partly completed |
| 🚧| in progress |
| ❔| to be clarified |
| ℹ️| interesting, but not relevant (yet) |
| ❌| incomplete |

# To-Dos for g/g
- ✅ vendor `k/*@v1.19`, `c-r@v0.7.0`
- ✅ update our forks: `third_party/forked/kubernetes` (no action required) and `github.com/gardener/kube-openapi`
- ✅ remove `client.Client` implementation in `pkg/client/kubernetes/utils` in favor of `manager.Options.ClientDisableCacheFor` (ref https://github.com/kubernetes-sigs/controller-runtime/pull/1249)
- ✅ adapt all helper functions to use `client.Object` where necessary, switch from `controllerutils.Object` to `client.Object` (ref https://github.com/kubernetes-sigs/controller-runtime/pull/1195)
- ✅ adapt all our reconcilers to the new reconciler interface (ref https://github.com/kubernetes-sigs/controller-runtime/pull/1054)
- ✅ drop our c-r fork, which was vendored in order to include https://github.com/kubernetes-sigs/controller-runtime/pull/1151
- ✅ adapt `extensionshandler.EnqueueRequestsFromMapFunc` and all extensions mappers to upstream interface changes (ref https://github.com/kubernetes-sigs/controller-runtime/pull/1119)
- ✅ remove usage from deperacted/removed helper funcs in `pkg/log` https://github.com/kubernetes-sigs/controller-runtime/pull/1105
[...]

PR #3359, opened 2021-01-07
Enhance documentation for the Gardenlet's /healthz endpoint

danielfoehrKn:

/kind bug
/priority normal
/area documentation

Based on #2925 but uses direct clients instead of listers.

The health manager changes the gardenlets  /healthz endpoint to DOWN if it cannot renew the lease in the Garden cluster. 
However, this does not cover the case when the Gardener Extended API Server is down.

The problem is, that Seeds can be queried from the lister even though the Gardener Extended API Server is down. This can be prevented by using a direct client. 

Advantage: 
  - Gardenlet fails when Gardener Extended API Server is down. I think this is the correct behaviour, as the Gardenlet cannot work without it.

Disadvantage: 
  - more network traffic (one request per 10 s)
  - I propose to decrease the ping interval from 10s -> 20s to save network traffic.

UPDATE: THis PR now only includes enhanced documentation for the Gardenlet's /healthz endpoint.

Fixes #

**Release note**:

```doc operator
Enhance documentation for Gardenlet's /healthz endpoint.
```

### DIFF
--- a/docs/concepts/gardenlet.md
+++ b/docs/concepts/gardenlet.md
@@ -146,23 +146,24 @@ More information: [Example Gardenlet Component Configuration](../../example/20-c
 
 ## Heartbeats
 
-Similar to how Kubernetes is meanwhile using `Lease` objects for node heart beats 
+Similar to how Kubernetes uses `Lease` objects for node heart beats 
 (see [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/0009-node-heartbeat.md)), 
 the gardenlet is using `Lease` objects for heart beats of the seed cluster.
-Every two seconds, the gardenlet is checking its connectivity 
-to the seed cluster and then it renews its lease.
-The status is reported in the `GardenletReady` condition in its `Seed` object list.
+Every two seconds, the gardenlet checks that the seed cluster's `/healthz` 
+endpoint returns HTTP status code 200. 
+If that is the case, the gardenlet renews the lease in the Garden cluster in the `gardener-system-seed-lease` namespace and updates 
+the `GardenletReady` condition in the `status.conditions` field of the `Seed` resource(s).
+
 Similarly to the `node-lifecycle-controller` inside the `kube-controller-manager`, 
 the `gardener-controller-manager` features a `seed-lifecycle-controller` that sets
-the `GardenletReady` condition to `Unknown` in case the gardenlet stops 
-sending its heartbeat signals.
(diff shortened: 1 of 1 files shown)


timuthy:
> However, this does not cover the case when the Gardener Extended API Server is down.

I'm generally not objecting but curious if you experienced any particular problems leaving the Gardenlet up and running while the API server was down? So far I was quite confident that a component like Gardenlet can deal with such intermittent issues.

rfranzke:
I'm still struggling a bit with this with respect to the advantages we get compared with the network I/O?

timebertt:
If network I/O is a big concern, we might wait for https://github.com/gardener/gardener/issues/3109 and use the metadata only client. This should be enough for this scenario, as we don't need the spec and saves network traffic.

rfranzke:
Yeah, maybe. I'm still in the phase of understanding the motivation for this change and if we experienced any particular problem with it (or if we foresee problems if we don't introduce this now).

timebertt:
Alternatively, we could also keep the watch but use some event handlers for delete events instead of using the lister.
Though, this would make the logic event-driven instead of level-driven, which might lead to issues, when we miss events.

danielfoehrKn:
> I'm generally not objecting but curious if you experienced any particular problems leaving the Gardenlet up and running while the API server was down?

No, I did not. 
My main motivation was that I "expected" it to fail because I thought that the `/healthz`endpoint should always indicate whether the Gardenlet is healthy / can work. 

Thinking about it again,  we might be better off just documenting the behaviour - the `/healthz`endpoint only  states that Gardenlet has a lease with the Garden cluster.

rfranzke:
Yeah, @danielfoehrKn, maybe that's a good first step to clarify on what can be expected.

timebertt:
I guess, we should generally improve our health checks and do a clean separation between liveness and readiness (ref https://github.com/gardener/gardener-resource-manager/pull/102#discussion_r553790882).
Then we can report an unready state, when caches haven't been synced for a minute or so. This would also cover the stale lease/seed case addressed here.

danielfoehrKn:
> I guess, we should generally improve our health checks and do a clean separation between liveness and readiness (ref gardener/gardener-resource-manager#102 (comment)).
Then we can report an unready state, when caches haven't been synced for a minute or so. This would also cover the stale lease/seed case addressed here.

Maybe you can open an issue for this describing this in more detail?
For this PR, I would only include the documentation - if that is fine with you.

rfranzke (approved):
/lgtm

timuthy (approved):

PR #3363, opened 2021-01-08
Introduce per test suite timeout

timebertt:

/area testing dev-productivity
/kind bug
/priority normal

This PR introduces a general per test suite timeout for unit tests, when running via `make test-cover`, e.g. in the CI verify step.

Unit tests should be quick / fail early. Long running unit tests indicate either some issues in the tests or some performance problem in the underlying production code itself.

I had some tests, that took very long to fail, which was annoying to notice/debug.

I didn't add a timeout to `make test`, i.e. for manually/locally executing unit tests.
The reason for that is, that the go test `-timeout` flag is not "cacheable" and therefore will cause all test suites to be executed everytime, although the results may be cached ([ref](https://golang.org/cmd/go/#hdr-Testing_flags)).

**Release note**:

```other developer
Unit tests are now limited to a timeout of 2 minutes per test suite.
```

### DIFF
--- a/hack/test-cover.sh
+++ b/hack/test-cover.sh
@@ -19,7 +19,7 @@ source "$(dirname $0)/setup-envtest.sh"
 
 echo "> Test Cover"
 
-GO111MODULE=on ginkgo -cover -race -mod=vendor $@
+GO111MODULE=on ginkgo -cover -timeout=2m -race -mod=vendor $@
 
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 COVERPROFILE="$REPO_ROOT/test.coverprofile"


rfranzke (dismissed):
/lgtm

pkg/utils/secrets/basic_auth_test.go
```diff
@@ -81,7 +81,8 @@ var _ = Describe("Basic Auth Secrets", func() {
 				Expect(err).NotTo(HaveOccurred())
 				compareCurrentAndExpectedBasicAuth(obj, expectedBasicAuthObject, false, false)
 			})
-			It("should properly generate Basic Auth Object with Bcrypt hashed password if specified in config", func() {
+			// pending because test is consistently slow, see https://github.com/gardener/gardener/issues/3362
+			PIt("should properly generate Basic Auth Object with Bcrypt hashed password if specified in config", func() {
```
rfranzke: Now that #3365 was opened you can revert this change

timebertt: Sure, done.

ialidzhikov (approved):
/lgtm

rfranzke (approved):

PR #3367, opened 2021-01-08
Sets concurrency to envoy

mvladev:

/area scalability
/kind enhancement
/priority normal

By default, envoy proxy creates worker threads for every single of the CPUs. On machines with lots of CPUs this causes lots more CPU and memory usage that is needed.

Fixes #

**Release note**:

```other operator
NONE
```


rfranzke (approved):
/lgtm

PR #3369, opened 2021-01-08
Increase the Loki chunk idle period from 3m to 1h

vlvasilev:

/area logging
/kind bug
/priority critical

This PR increase the Loki chunk idle period from 3m to 1h.
This is required because when a chunk size is greatly lower than the default inode size we hit the maximum available files on the volume and Loki stops to work until old chunks are deleted. This issue was hit on Azure mainly.
 
Fixes #

**Release note**:


Kristian-ZH (approved):
/lgtm

vpnachev (approved):
/lgtm

PR #3374, opened 2021-01-11
Set revisionHistoryLimit to 1 for deployments

### LINKED ISSUE #3325: Increase RevisionHistoryLimit to 1

/area control-plane
/kind enhancement
/priority normal

Since a while we have set `revisionHistoryLimit: 0` to not spam seed clusters with obsolete `ReplicaSets`. However, it turns out useful to at least keep a short history i.e., `revisionHistoryLimit: 1`.

**Why is this needed**:
Sometimes we observe unexpected changes of components, usually due to restarts. An adequate analysis is often impossible or very though without having a diff between the current and last spec.

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal

This PR increases the `.spec.revisionHistoryLimit` for `Deployment`s to `1`.

Fixes #3325

This will lead to some DB size increase in etcd, though, probably not too much.

**Release note**:

```feature operator
The `.spec.revisionHistoryLimit` is now set to `1` for `Deployment`s.
```

### DIFF
--- a/charts/istio/istio-ingress/templates/deployment.yaml
+++ b/charts/istio/istio-ingress/templates/deployment.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
 {{ .Values.labels | toYaml | indent 4 }}
 spec:
-  revisionHistoryLimit: 0
+  revisionHistoryLimit: 1
   selector:
     matchLabels:
 {{ .Values.labels | toYaml | indent 6 }}
--- a/charts/istio/istio-istiod/templates/deployment.yaml
+++ b/charts/istio/istio-istiod/templates/deployment.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
 {{ .Values.labels | toYaml | indent 4 }}
 spec:
-  revisionHistoryLimit: 0
+  revisionHistoryLimit: 1
   strategy:
     rollingUpdate:
       maxSurge: 100%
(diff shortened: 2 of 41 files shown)


timebertt (commented):
I'm wondering: did you consider adding a mutating webhook to the seed-admission-controller for setting the `revisionHistoryLimit`?
Setting this by hand will lead to inconsistent settings, I guess. And the webhook approach would have the benefit, that it also applies to deployments not managed by gardener, e.g. `cloud-controller-manager`.

rfranzke:
I thought about it, but quickly discarded it again because there are a gazillion of other examples were extensions can theoretically deviate, maybe for good reasons. For example, we would like to have a `VPA` for all the components or checksum annotations for mounted volumes, etc. etc.
For now, to save some time and effort, I simply changed the values similar to how we early set them all to `0`.

vpnachev (changes requested):
Please, change it also
-  here https://github.com/gardener/gardener/blob/085d163e1cbd73397aa6b31197714729f8bf7c53/pkg/gardenlet/controller/shoot/seed_registration_control.go#L739
- here https://github.com/gardener/gardener/blob/085d163e1cbd73397aa6b31197714729f8bf7c53/pkg/operation/botanist/controlplane/clusterautoscaler/cluster_autoscaler.go#L172
- here https://github.com/gardener/gardener/blob/085d163e1cbd73397aa6b31197714729f8bf7c53/pkg/operation/botanist/controlplane/kubecontrollermanager/kube_controller_manager.go#L175
- and here https://github.com/gardener/gardener/blob/085d163e1cbd73397aa6b31197714729f8bf7c53/pkg/operation/botanist/controlplane/kubescheduler/kube_scheduler.go#L171

timebertt (approved):
/lgtm

rfranzke:
ping @vpnachev

vpnachev (approved):
/lgtm

PR #3379, opened 2021-01-12
Avoid Grafana restarts

timuthy:

/area monitoring
/kind bug
/priority normal

This PR fixes sporadic Grafana restarts during reconciliation. An unstable dashboard order caused different `sha256` checksum annotations and thus the creation of new Grafana pods.

/invite @danielfoehrKn 

**Release note**:

```bugfix operator
An issue has been fixed which caused unwanted restarts for Grafana instances.
```

### DIFF
--- a/pkg/operation/botanist/monitoring.go
+++ b/pkg/operation/botanist/monitoring.go
@@ -99,6 +99,9 @@ func (b *Botanist) DeploySeedMonitoring(ctx context.Context) error {
 		return err
 	}
 
+	// Need stable order before passing the dashboards to Grafana config to avoid unnecessary changes
+	kutil.ByName().Sort(existingConfigMaps)
+
 	// Read extension monitoring configurations
 	for _, cm := range existingConfigMaps.Items {
 		alertingRules.WriteString(fmt.Sprintln(cm.Data[v1beta1constants.PrometheusConfigMapAlertingRules]))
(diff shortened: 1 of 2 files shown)


pkg/operation/botanist/monitoring.go
```diff
@@ -99,6 +99,9 @@ func (b *Botanist) DeploySeedMonitoring(ctx context.Context) error {
 		return err
 	}
 
+	// Need stable order before passing the dashboards to Grafana config to avoid unnecessary changes
+	kutil.ByName().Sort(existingConfigMaps)
```
rfranzke, line 103: Awesome :-D

pkg/operation/seed/seed.go
```diff
@@ -75,12 +74,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-type byName []corev1.ConfigMap
```
rfranzke, line 78: Thanks for cleaning this up! Actually, I wanted to do this as well while introducing the generic sorter, but then forgot about it.

rfranzke (approved):
/lgtm

pkg/operation/botanist/monitoring.go
timuthy: The PR was a low-hanging fruit thanks to your groundwork :D

PR #3382, opened 2021-01-12
fix istiod rolebinding

danielfoehrKn:

/kind bug
/priority normal

I experienced the following problem when using the managed-istio feature gate from the current master branch.

Istiod deployment has insufficient RBAC rights
 - cannot write self-generated CA into secret `istio-ca-secret`
 - istio-ingressgateway mounts this secret -> fails to mount because secret non-existend
```
k logs deployment/istiod
2021-01-12T14:30:10.717108Z	error	pkica	Failed to write secret to CA (error: secrets is forbidden: User "system:serviceaccount:istio-system:istiod" cannot create resource "secrets" in API group "" in the namespace "istio-system"). Abort.
2021-01-12T14:30:10.717175Z	error	failed to create discovery service: failed to create CA: failed to create a self-signed istiod CA: failed to create CA due to secret write error
Error: failed to create discovery service: failed to create CA: failed to create a self-signed istiod CA: failed to create CA due to secret write error
```

The istiod rolebinding is created in the `default` namespace, I think it should be instead created in the` istio-system`.

Fixes #

I can see that in existing landscapes the missing secret `istio-ca-secret` does exist. 
Possibly I am missing something here, but I do not know who wrote that secret initially.

**Release note**:

```bugfix operator
Fixed a bug of the managed istio feature flag where the istio rolebinding was created in the wrong namespace.
```

### DIFF
--- a/charts/istio/istio-istiod/templates/rolebinding.yaml
+++ b/charts/istio/istio-istiod/templates/rolebinding.yaml
@@ -2,6 +2,7 @@ apiVersion: {{ include "rbacversion" . }}
 kind: RoleBinding
 metadata:
   name: istiod
+  namespace: {{ .Release.Namespace }}
   labels:
 {{ .Values.labels | toYaml | indent 4 }}
 roleRef:


rfranzke (approved):
/lgtm

ialidzhikov:
Cherry-picked into `release-v1.15` with 848a8b9d011daf257b618af1c9fe0b3044ce9f3b.

PR #3384, opened 2021-01-12
Add the 'gardener.cloud/role' label key to controlplane Pods

ialidzhikov:
/kind cleanup

This PR adds the `gardener.cloud/role` label key to the controlplane Pods. This is a preparation step that will allows us (after 
 several releases) to switch the `podSelector` of NetworkPolicies that currently select `garden.sapcloud.io/role` and the label selector of the dependency-watchdog component that is also using `garden.sapcloud.io/role` currently for selecting dependantPods that will be restarted after kube-apiserver recovers from crash-loops.
This PR adds the `gardener.cloud/role` label key to all contorlplane components except etcd-events and etcd-main. I didn't want to trigger a rolling update of our etcd for such small change as they are single node currently. So my suggestion is to add the `gardener.cloud/role` to the etcd Pods when we have in the pipeline a change that will roll out the etcds (for example the etcd Pod spec is updated or we roll out a new etcd image/version).

Part of #1649

**Release note**:

```other operator
NONE
```


charts/seed-controlplane/charts/gardener-resource-manager/templates/deployment.yaml
```diff
@@ -21,6 +20,7 @@ spec:
 {{ toYaml .Values.podAnnotations | indent 8 }}
 {{- end }}
       labels:
+        gardener.cloud/role: controlplane
```
rfranzke, line 23: cc @BeckerMax

rfranzke (dismissed):
/lgtm

rfranzke (approved):

rfranzke:
@timebertt I think it's fine to merge this PR even before https://github.com/gardener/gardener/pull/3393 because it does not seem to conflict. If you agree let's go ahead and merge it, otherwise we can /hold it, of course.

danielfoehrKn (approved):
/lgtm

PR #3385, opened 2021-01-12
Fix gardener-seed-scheduler permissions

mvladev:

/area scalability
/kind bug
/priority normal

Fixes missing lease permissions.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/operation/seed/scheduler/gardener_kube_scheduler.go
+++ b/pkg/operation/seed/scheduler/gardener_kube_scheduler.go
@@ -32,6 +32,7 @@ import (
 	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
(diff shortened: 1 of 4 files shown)


pkg/operation/seed/scheduler/gardener_kube_scheduler.go
```diff
@@ -65,6 +66,7 @@ func New(
 	client client.Client,
 	namespace string,
 	image string,
+	leaseResourceName string,
```
rfranzke: Is it really necessary that somebody has to specify this here? Can't it be "hard-coded" by the package? What's the use-case?

rfranzke (commented):
/assign

pkg/operation/seed/scheduler/gardener_kube_scheduler.go
mvladev: Reworked the PR.

vpnachev (approved):
/lgtm
/needs second-opinion

rfranzke (approved):

PR #3386, opened 2021-01-12
Fix debootstaping for gardener-kube-scheduler

### LINKED ISSUE #3378: Failed to unbootstrap Seed Cluster (image is required)
/kind bug
/kind regression
/priority normal

**What happened**:
Deletion of Seed cluster does not work.

Deletion of Seed cluster to work.

1. Register a v1.18.14 Seed cluster.

2. Make sure that it is successfully bootstrapped and registered.

3. Delete the Seed cluster

4. Make sure that the deboostrap fails with:
```
time="2021-01-12T11:38:54+02:00" level=error msg="Failed to unbootstrap Seed Cluster (image is required)." seed=seed-local
time="2021-01-12T11:39:10+02:00" level=error msg="Failed to unbootstrap Seed Cluster (image is required)." seed=seed-local
time="2021-01-12T11:39:25+02:00" level=error msg="Failed to unbootstrap Seed Cluster (image is required)." seed=seed-local
```

- Gardener version: master
- Others:

mvladev:

/area scalability
/kind bug
/priority normal

Fixes #3378

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/operation/seed/seed.go
+++ b/pkg/operation/seed/seed.go
@@ -883,8 +883,9 @@ func bootstrapComponents(c kubernetes.Interface, namespace string, imageVector i
 	}
 	components = append(components, seedadmission.New(c.Client(), namespace, gsacImage.String(), kubernetesVersion))
 
-	// gardener-seed-scheduler
-	schedulerImage := &imagevector.Image{}
+	// gardener-seed-scheduler.
+	var schedulerImage *imagevector.Image
+
 	if imageVector != nil {
 		schedulerImage, err = imageVector.FindImage(common.KubeSchedulerImageName, imagevector.TargetVersion(kubernetesVersion.String()))
 		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/operation/seed/seed.go
```diff
@@ -883,8 +884,8 @@ func bootstrapComponents(c kubernetes.Interface, namespace string, imageVector i
 	}
 	components = append(components, seedadmission.New(c.Client(), namespace, gsacImage.String(), kubernetesVersion))
 
-	// gardener-seed-scheduler
-	schedulerImage := &imagevector.Image{}
+	// gardener-seed-scheduler. Need dummy image in order to allow debootstrap to work
+	schedulerImage := &imagevector.Image{Name: "dummy", Repository: "dummy", Tag: pointer.StringPtr("dummy")}
```
rfranzke: Hm, TBH, I don't like it too much (rather not at all).  Can't we rather improve the validations to make instantiating a new object work with a `non-nil` image? Or, simply remove this entirely:

https://github.com/gardener/gardener/blob/4769e2b102502312967fc9d1c1ab8983eb767104/pkg/operation/seed/scheduler/gardener_kube_scheduler.go#L86-L88

This check isn't too meaningful anyways. If it's empty it doesn't work, yes, but for the validation to be really valuable you'd also have to check that a correct image is used. Hence, Maybe we should just get rid of it entirely instead of such workarounds?

mvladev:
I've reworked the PR. Those checks for images and so are only done during the `Deploy` call (where they are used).

rfranzke (approved):
/lgtm
Thanks @mvladev, this looks way cleaner!

timuthy (approved):

PR #3387, opened 2021-01-12
Fix 'make install'

ialidzhikov:
```
$ make install
> Install
# github.com/gardener/gardener/pkg/operation/botanist
pkg/operation/botanist/waiter.go:108:7: undefined: errors
```

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/waiter.go
+++ b/pkg/operation/botanist/waiter.go
@@ -16,6 +16,7 @@ package botanist
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"time"


rfranzke (approved):
/lgtm

Interesting, the first time since long that we run into such issue due to a non-up2date PR.
Any proposal how to prevent this from happening again in the future (other than requiring PRs to be up2date which is somewhat cumbersome)? Would it have helped if we re-run all CI/CD checks for already approved PRs if other PRs were merged in the meantime?

ialidzhikov:
In k/k, the PRs are retested after being lgtm-ed and approved-ed also right before merge. Probably one of the reason they do this double retesting is to catch this case with not up2date PRs. For now I wouldn't consider this issue as major one as we observe it rarely (I can recall only one other case 🤔 ) and we have the master-head-update jobs and some monitoring on them. I guess even the last committer/author should receive an email for a failed master-head-update job, right?

PR #3372, opened 2021-01-11
Add scheduler versions to third_party not a as module

mvladev:
**How to categorize this PR?**

/area dev-productivity
/kind cleanup
/priority normal

**What this PR does / why we need it**:

Having modules makes vendoring gardener repository hard.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator
NONE
```

Issue #2977, opened 2020-10-14
GAPI panics with "reflect: NumField of non-struct type *v1.LabelSelector"


timebertt:
**How to categorize this issue?**

/area quality
/kind bug
/priority critical

**What happened**:
When applying a `CloudProfile`, `gardener-apiserver` sometimes panics with the following stack trace:
```
E1014 10:44:19.444581       1 runtime.go:76] Observed a panic: reflect: NumField of non-struct type *v1.LabelSelector
goroutine 137801607 [running]:
k8s.io/apiserver/pkg/endpoints/handlers.finishRequest.func1.1(0xc01c8a52c0)
	/go/src/github.com/gardener/gardener/vendor/k8s.io/apiserver/pkg/endpoints/handlers/rest.go:233 +0xc9
panic(0x1fe4e40, 0xc004ed29a0)
	/usr/local/go/src/runtime/panic.go:969 +0x166
reflect.(*rtype).NumField(0x22e9180, 0x2c)
	/usr/local/go/src/reflect/type.go:973 +0xb7
sigs.k8s.io/structured-merge-diff/v3/value.buildStructCacheEntry(0x28d2500, 0x22e9180, 0xc01c2bb380, 0xc03f4c8a20, 0x1, 0x1)
	/go/src/github.com/gardener/gardener/vendor/sigs.k8s.io/structured-merge-diff/v3/value/reflectcache.go:146 +0x67
sigs.k8s.io/structured-merge-diff/v3/value.buildStructCacheEntry(0x28d2500, 0x218d920, 0xc01c2bb380, 0x0, 0x0, 0x0)
	/go/src/github.com/gardener/gardener/vendor/sigs.k8s.io/structured-merge-diff/v3/value/reflectcache.go:153 +0x1b7
sigs.k8s.io/structured-merge-diff/v3/value.typeReflectEntryOf(0xc02397e6f0, 0x28d2500, 0x218d920, 0xc019c8e410, 0xc00815de00)
	/go/src/github.com/gardener/gardener/vendor/sigs.k8s.io/structured-merge-diff/v3/value/reflectcache.go:119 +0x44e
sigs.k8s.io/structured-merge-diff/v3/value.TypeReflectEntryOf(0x28d2500, 0x218d920, 0x199)
	/go/src/github.com/gardener/gardener/vendor/sigs.k8s.io/structured-merge-diff/v3/value/reflectcache.go:94 +0x13d
[...]

PR #3393, opened 2021-01-13
Vendor k/*@v0.19 and controller-runtime@v0.7

timebertt:

/area open-source quality
/kind task
/priority normal

This PR vendors all the kubernetes dependencies in version `v0.19.6` and controller-runtime version `v0.7.0`.
See https://github.com/gardener/gardener/issues/3109 for more details.

This is only the strictly required part of https://github.com/gardener/gardener/issues/3109

I tried to work in different commits to ease the review of this monstrous PR, each one of them addresses *mainly* one breaking change in upstream (though, it's not 100% accurately separated, just to ease the review).
/squash

`gardener-seed-admission-controller` currently uses some custom decoding and encoding logic for admission request and review, but uses the controller-runtime types for passing them down to the handlers.
The upstream types were changed to `admission/v1` (ref https://github.com/kubernetes-sigs/controller-runtime/pull/1284).
I didn't want to go through the hassle of adapting our logic to handle both admission versions, I rather removed the custom logic and migrated the handlers to use the upstream implementation.
I haven't migrated to the controller-runtime manager/webhook server yet and also haven't done the same change in the `gardener-admission-controller`, as it wasn't strictly required and I wanted to keep this PR limited to the required changes in the upgrade process.

**Release note**:

```breaking dependency
⚠️ Go dependencies to `kubernetes/*` and `kubernetes-sigs/controller-runtime` were updated to `v0.19.6` and `v0.7.0` respectively. This imposes a lot of consequent breaking changes to go projects vendoring gardener/gardener. If your project/extension vendors gardener/gardener, please read the dedicated section in [this issue](https://github.com/gardener/gardener/issues/3109) carefully when upgrading your dependencies.
```


timebertt:
I had a quick chat with @mvladev about https://github.com/gardener/gardener/pull/3372#issuecomment-758745212.
The way I addressed it in this PR should be fine for now, as we don't use the `Extenders` field and its anyways only a renaming.
@mvladev wants to copy the `k8s.io/kube-scheduler/config/v1` as well and set things straight in this regard, but will do it after this PR.

pkg/gardenlet/bootstrap/bootstrap.go
```diff
@@ -21,52 +21,33 @@ import (
...
 	"strings"
 
-	"github.com/gardener/gardener/pkg/gardenlet/apis/config"
-	"github.com/gardener/gardener/pkg/gardenlet/bootstrap/certificate"
-	bootstraputil "github.com/gardener/gardener/pkg/gardenlet/bootstrap/util"
-	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
-
 	"github.com/sirupsen/logrus"
-	certificatesv1beta1 "k8s.io/api/certificates/v1beta1"
+	certificatesv1 "k8s.io/api/certificates/v1"
```
rfranzke, line 25: We support Garden clusters with at least Kubernetes v1.16. The `v1` version of the certificates API was only introduced with Kubernetes v1.19. Will all of this still work even if the garden cluster is < v1.19?

pkg/client/kubernetes/runtime_client.go
```diff
@@ -45,8 +48,8 @@ func NewDirectClient(config *rest.Config, options client.Options) (client.Client
 }
 
 // NewRuntimeClientWithCache creates a new client.client with the given config and options.
-// The client uses a new cache, which will be started immediately using the given stop channel.
-func NewRuntimeClientWithCache(config *rest.Config, options client.Options, stopCh <-chan struct{}) (client.Client, error) {
+// The client uses a new cache, which will be started immediately using the given context.
+func NewRuntimeClientWithCache(config *rest.Config, options client.Options, ctx context.Context, uncachedObjects ...client.Object) (client.Client, error) {
```
rfranzke: Can we make the context the first argument?

extensions/pkg/controller/utils.go
```diff
@@ -143,48 +137,16 @@ func HasFinalizer(obj runtime.Object, finalizerName string) (bool, error) {
...
-	if !finalizers.Has(finalizerName) {
-		return nil
-	}
-
-	finalizers.Delete(finalizerName)
-	accessor.SetFinalizers(finalizers.UnsortedList())
-
-	return client.Update(ctx, obj)
-}
+var DeleteFinalizer = controllerutils.RemoveFinalizer
```
rfranzke, line 144: Should we deprecate this like https://github.com/gardener/gardener/pull/3393/commits/65071d411feb2052610e1f3f93aef6714a1bebc7#diff-e9cde0128495c3c977af30cb41d328dd64f8baed137ef1f3ca1c67497ebcc7fdR100?

extensions/pkg/controller/utils.go
```diff
@@ -143,48 +137,16 @@ func HasFinalizer(obj runtime.Object, finalizerName string) (bool, error) {
...
-	if finalizers.Has(finalizerName) {
-		return nil
-	}
-
-	finalizers.Insert(finalizerName)
-	accessor.SetFinalizers(finalizers.UnsortedList())
-
-	return client.Update(ctx, obj)
-}
+var EnsureFinalizer = controllerutils.EnsureFinalizer
```
rfranzke, line 140: Should we deprecate this like https://github.com/gardener/gardener/pull/3393/commits/65071d411feb2052610e1f3f93aef6714a1bebc7#diff-e9cde0128495c3c977af30cb41d328dd64f8baed137ef1f3ca1c67497ebcc7fdR100?

pkg/utils/kubernetes/client/client.go
```diff
@@ -32,62 +32,47 @@ import (
...
-}
-
-// RemainingObjects retrieves the remaining objects of an `AreObjectsRemaining` error.
-//
-// If the error does not match `AreObjectsRemaining`, this returns nil.
-func RemainingObjects(err error) []runtime.Object {
-	if nEmpty, ok := err.(*objectsRemaining); ok {
-		return *nEmpty
+func NewObjectsRemaining(obj runtime.Object) error {
+	switch remaining := obj.(type) {
```
rfranzke, line 62: Nice! Much better to understand what's happening now.

pkg/utils/kubernetes/kubernetes.go
```diff
@@ -421,11 +418,7 @@ func OwnedBy(obj runtime.Object, apiVersion, kind, name string, uid types.UID) b
 // is provided then it will be applied for each object right after listing all objects. If no object remains then nil
 // is returned. The Items field in the list object will be populated with the result returned from the server after
 // applying the filter function (if provided).
-func NewestObject(ctx context.Context, c client.Client, listObj runtime.Object, filterFn func(runtime.Object) bool, listOpts ...client.ListOption) (runtime.Object, error) {
-	if !meta.IsListType(listObj) {
-		return nil, fmt.Errorf("provided <listObj> is not a List type")
-	}
-
+func NewestObject(ctx context.Context, c client.Client, listObj client.ObjectList, filterFn func(runtime.Object) bool, listOpts ...client.ListOption) (runtime.Object, error) {
```
rfranzke, line 421: Shouldn't the `filterFn` work with `client.Object` now?

extensions/pkg/controller/backupbucket/reconciler.go
```diff
@@ -73,25 +71,20 @@ func (r *reconciler) InjectClient(client client.Client) error {
 	return nil
 }
 
-func (r *reconciler) InjectStopChannel(stopCh <-chan struct{}) error {
-	r.ctx = contextutil.FromStopChannel(stopCh)
-	return nil
-}
-
-func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
```
rfranzke, line 74: Finally there is a context for `Reconcile`. Nice! :)

extensions/pkg/handler/enqueue_mapped.go
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: 2021 ;)

extensions/pkg/webhook/handler_shootclient.go
```diff
@@ -162,11 +175,11 @@ func (h *handlerShootClient) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	req := admission.Request{}
-	ar := admissionv1beta1.AdmissionReview{
+	ar := admissionv1.AdmissionReview{
```
rfranzke: We support shoot clusters with Kubernetes at least v1.10. The `v1` version of the admission API was only introduced with v1.16. Does all of this still work if we just use `v1` now?

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
```diff
@@ -183,7 +185,7 @@ func respond(w http.ResponseWriter, request *admission.Request, response admissi
 		}
 	}
 
-	jsonResponse, err := json.Marshal(admissionv1beta1.AdmissionReview{Response: &response.AdmissionResponse})
+	jsonResponse, err := json.Marshal(admissionv1.AdmissionReview{Response: &response.AdmissionResponse})
```
rfranzke: Same here - the seeds can be at least v1.11 but the `v1` only exists in `v1.16`. Does all of this still work with older versions?

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
```diff
@@ -143,18 +133,41 @@ func run(ctx context.Context, bindAddress string, port int, certPath, keyPath, k
...
+	extensionDeletionProtection := &webhook.Admission{Handler: seedadmission.NewExtensionDeletionProtection(logger)}
+	defaultSchedulerName := &webhook.Admission{Handler: admission.HandlerFunc(seedadmission.DefaultShootControlPlanePodsSchedulerName)}
+	if err := setFields(extensionDeletionProtection); err != nil {
+		panic(fmt.Errorf("error injecting dependencies into webhook handler: %w", err))
+	}
+	if err := setFields(defaultSchedulerName); err != nil {
+		panic(fmt.Errorf("error injecting dependencies into webhook handler: %w", err))
+	}
+
+	mux.Handle("/webhooks/validate-extension-crd-deletion", extensionDeletionProtection)
```
rfranzke, line 165: Maybe you can introduce a constant for this now that you are at it (similar to the `GardenerShootControlPlaneSchedulerWebhookPath` constant)

pkg/seedadmission/extension_crds.go
```diff
@@ -17,25 +17,63 @@ package seedadmission
...
+
+// InjectDecoder injects a decoder capable of decoding objects included in admission requests.
+func (h *extensionDeletionProtection) InjectDecoder(d *admission.Decoder) error {
+	h.decoder = d
+	return nil
+}
+
+// Handle implements the webhook handler for extension deletion protection.
+func (h *extensionDeletionProtection) Handle(ctx context.Context, request admission.Request) admission.Response {
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
rfranzke, line 69: Nice :)

pkg/seedadmission/pod_scheduler_name.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: 2021

pkg/seedadmission/pod_scheduler_name_test.go
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: 2021

pkg/seedadmission/pod_scheduler_name.go
```diff
@@ -0,0 +1,50 @@
...
+	if request.Kind != podGVK {
+		return admission.Allowed("resource is not corev1.Pod")
+	}
+
+	if request.SubResource != "" {
+		return admission.Allowed("subresources on pods are not supported")
+	}
+
+	return admission.Patched(
+		"schedulerName defaulted",
```
rfranzke: What about printing the `GardenerShootControlPlaneSchedulerName` here as well?

rfranzke (changes requested):
Well, first of all: Kudos and thanks a lot for this great PR. It's well structured and the dedicated commits helped tremendously to review it. It was a very smooth experience!

I've mainly found a few nits, not much. I've three "critical" questions when it comes to supporting old Kubernetes versions now that you upgraded certain APIs to new versions that are not available in older Kubernetes versions (which we still support, though).

pkg/gardenlet/bootstrap/bootstrap.go
timebertt: Valid point, I just noticed this as well when I wanted to start testing this.

So the CSR request itself should work, as the used util falls back to `certificates.k8s.io/v1beta1` if `v1` is not available, 
ref https://github.com/kubernetes/kubernetes/pull/91754.
The only thing, that still needs to be adapted is the cleanup in `DeleteBootstrapAuth`, but this should be easy.

Anyway, I will do this small adaption so we don't have to change garden cluster version requirement and ensure that everything still works with v1.16.

extensions/pkg/webhook/handler_shootclient.go
timebertt: Arg, I missed to change that part again similar to the seed-admission-controller. It should rather use the upstream c-r decoding logic instead of the custom implementation.

pkg/utils/kubernetes/kubernetes.go
timebertt: Hmm, `meta.EachListItem` only supports `runtime.Object` so we would need to type-assert there again before calling the filter.
So I think, this should be fine for now, as it's not strictly required to get the vendoring through.
I have an open task in the umbrella issue about getting rid of the remaining `meta.Accessor` usages and this would be covered by this.

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
timebertt: Let me do that later on, when I will migrate GSAMC to a pure c-r component.

pkg/seedadmission/pod_scheduler_name.go
timebertt: Actually I think, the reason field is supposed to hold machine-readable string, see https://github.com/kubernetes/apimachinery/blob/1d8c923392f087e29f1e5bb563e073b833adc454/pkg/apis/meta/v1/types.go#L644-L649
We anyway break this convention basically everywhere and controller-runtime also seems to break it with its helper funcs.
Also not to dramatic, but just a side-note, will add it.

extensions/pkg/controller/utils.go
timebertt: No. In addition to removing the finalizer from the in-memory object it also takes care about updating the object on the API server. And it actually already uses `controllerutil.RemoveFinalizer` for the in-memory part.

extensions/pkg/controller/utils.go
timebertt: Same here.

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
timebertt: Oh sorry, that diff is confusing. The logic is removed later on in a0ebc3d8e2d89cf0f931ae0f4e169af601333047
There the c-r logic is used for dealing with encoding and decoding and takes care about version compatibility.

pkg/client/kubernetes/options.go
```diff
@@ -94,29 +87,15 @@ func WithCacheResyncPeriod(resync time.Duration) ConfigFunc {
 // DirectClient().
 func WithDisabledCachedClient() ConfigFunc {
 	return func(config *Config) error {
-		config.disableCachedClient = true
+		config.disableCached = true
```
timuthy:
```suggestion
              config.disableCache = true
```

pkg/client/kubernetes/options.go
```diff
@@ -18,25 +18,18 @@ import (
...
-
-// cacheReaderOptions configures the specificallyCachedReader
-type cacheReaderOptions struct {
-	readSpecifiedFromCache    bool
-	specificallyCachedObjects []runtime.Object
+	clientOptions   client.Options
+	restConfig      *rest.Config
+	cacheResync     *time.Duration
+	disableCached   bool
+	disableCacheFor []client.Object
```
timuthy:
```suggestion
      uncachedObjects
```

pkg/utils/kubernetes/kubernetes.go
rfranzke: OK, makes sense, thanks.

pkg/seedadmission/pod_scheduler_name.go
rfranzke: Ah, I see, I didn't check it in detail what this argument is about. If it's a reason for the patch then "schedulerName defaulted" is wrong in the first place. It should be rather something like "shoot control plane pod created". Anyways, minor.

extensions/pkg/controller/utils.go
rfranzke: I see, thanks.

pkg/gardenlet/bootstrap/certificate/certificate.go
```diff
@@ -88,23 +88,25 @@ func requestCertificate(ctx context.Context, logger logrus.FieldLogger, certific
 
 	logger.Info("Creating certificate signing request...")
 
-	// LegacyUnkownSignerName based on https://github.com/kubernetes/kubernetes/blob/db4ca87d9d872b3c31df58860bac996c70df6b5b/pkg/apis/certificates/v1beta1/defaults.go#L52.
-	req, err := csr.RequestCertificate(certificateClient, csrData, name, certificatesv1beta1.LegacyUnknownSignerName, usages, privateKey)
+	// TODO (timebertt): figure out correct signerName for gardenlet CSRs
+	// kubernetes.io/legacy-unknown is disallowed when creating new CSR objects in certificates/v1,
+	// ref https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/1513-certificate-signing-request#api-changes-between-v1beta1-and-v1
+	reqName, reqUID, err := csr.RequestCertificate(client, csrData, name, certificatesv1.KubeAPIServerClientSignerName, usages, privateKey)
```
timuthy, line 91: As discussed, `certificatesv1.KubeAPIServerClientSignerName` should be fine when reading  https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#kubernetes-signers and comparing the usage field [here](https://github.com/gardener/gardener/blob/221e5d764c43e3ebde6c096d7dc2755521454d13/pkg/controllermanager/controller/certificatesigningrequest/csr_autoapprove_control.go#L195). GCM should also be allowed to approve the request since it has the `cluster-admin` role.

go.mod
```diff
@@ -28,39 +28,39 @@ require (
...
 	k8s.io/helm v2.16.1+incompatible
 	k8s.io/klog v1.0.0
-	k8s.io/kube-aggregator v0.18.10
-	k8s.io/kube-openapi v0.0.0-20200410145947-bcb3869e6f29 // release-1.18
-	k8s.io/kube-scheduler v0.18.10
-	k8s.io/kubelet v0.18.10
-	k8s.io/metrics v0.18.10
-	k8s.io/utils v0.0.0-20200619165400-6e3d28b6ed19
+	k8s.io/kube-aggregator v0.19.6
+	k8s.io/kube-openapi v0.0.0-20200805222855-6aeccd4b50c6 // release-1.18
```
timuthy:
```suggestion
      k8s.io/kube-openapi v0.0.0-20200805222855-6aeccd4b50c6 // keep this value in sync with k8s.io/apiserver
```

timuthy (commented):
Nice work!
I haven't test the changes yet but wanted to submit some minor suggestions, comments.

pkg/gardenlet/bootstrap/certificate/certificate.go
timebertt: Thanks for checking the details. I will remove the `TODO` once I've successfully tested the gardenlet bootstrapping cert rotation :)

timebertt:
I've added a new commit, which migrates the extensions' shoot webhook handler to upstream controller runtime handler.
The problem here was, that we use the remote IP to identify which API server is calling the webhook in order to construct a shoot client. 
I added a small `http.Handler` wrapper that injects the `RemoteAddr` into the context to retrieve it later on in the webhook handler.
This should theoretically work, though I don't have a quick and meaningful way to test this.
Let me know if you have concerns with getting it in like this or if you have some idea to properly test this.

extensions/pkg/webhook/handler_shootclient.go
timebertt: Addressed in the latest commit.

extensions/pkg/controller/healthcheck/reconciler.go
```diff
@@ -39,12 +30,19 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+
+	"github.com/gardener/gardener/extensions/pkg/controller"
```
ialidzhikov: Already imported in the line below.

```suggestion
```

extensions/pkg/controller/worker/reconciler_test.go
```diff
@@ -38,6 +31,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+
+	"github.com/gardener/gardener/extensions/pkg/controller"
```
ialidzhikov:
```suggestion
```

extensions/pkg/controller/worker/reconciler_test.go
```diff
@@ -57,7 +58,7 @@ var _ = Describe("Worker Reconcile", func() {
 		wantErr bool
 	}
 
-	//Ummutable throuth the function calls
+	// Ummutable throuth the function calls
```
ialidzhikov:
```suggestion
	// Immutable through the function calls
```

extensions/pkg/webhook/cloudprovider/cloudprovider.go
```diff
@@ -15,16 +15,16 @@
...
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/gardener/gardener/extensions/pkg/webhook"
```
ialidzhikov:
```suggestion
```

go.mod
```diff
@@ -14,74 +14,73 @@ require (
...
+	k8s.io/apiserver => k8s.io/apiserver v0.19.6
+	k8s.io/client-go => k8s.io/client-go v0.19.6
+	k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.19.6
+	k8s.io/code-generator => k8s.io/code-generator v0.19.6
+	k8s.io/component-base => k8s.io/component-base v0.19.6
 	k8s.io/helm => k8s.io/helm v2.13.1+incompatible
-	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.18.10
-	k8s.io/kube-openapi => github.com/gardener/kube-openapi v0.0.0-20200831104310-b5db060350c7
-	sigs.k8s.io/controller-runtime => github.com/gardener/controller-runtime v0.6.3-gardener.1
-	sigs.k8s.io/structured-merge-diff/v3 => sigs.k8s.io/structured-merge-diff/v3 v3.0.1-0.20201124164700-f5fd4ea1e4c9
```
ialidzhikov, line 86: I see that the `sigs.k8s.io/structured-merge-diff` replace is now removed but as far I see the upstream fix that we need https://github.com/kubernetes-sigs/structured-merge-diff/pull/173 is not included in the current version of `sigs.k8s.io/structured-merge-diff`.

```
$ go list -mod=mod -m all | grep sigs.k8s.io/structured-merge-diff | awk '{print $2}'
v4.0.1
```

As far as I see v4.0.1 does not contain https://github.com/kubernetes-sigs/structured-merge-diff/pull/173 and this can regress https://github.com/gardener/gardener/issues/2977. I can check whether I can file a cherry-pick PR about it.

timebertt: Yes, good point. I wasn't sure about this. It's still on my list for testing.

timebertt: Ok, just confirmed. This currently regresses #2977, so we probably need to update our fork again until we have an upstream cherry-pick.

ialidzhikov: Ref https://github.com/kubernetes-sigs/structured-merge-diff/pull/182

extensions/pkg/webhook/remote_addr.go
```diff
@@ -0,0 +1,45 @@
...
+
+// InjectFunc injects into the underlying handler.
+func (h remoteAddrInjectingHandler) InjectFunc(f inject.Func) error {
+	return f(h.Handler)
+}
+
+// ServerHTTP implements http.Handler by delegating to the underlying handler but injecting request.RemoteAddr into
+// the request's context.
+func (h remoteAddrInjectingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	h.Handler.ServeHTTP(w, r.Clone(context.WithValue(r.Context(), remoteAddrContextKey, r.RemoteAddr)))
```
timuthy, line 44: Any reason for cloning instead of using `r.WithContext()` ?

timebertt: `r.WithContext()` doesn't seem to clone headers and so on and also says the following in the doc string:

> // To create a new request with a context, use NewRequestWithContext.
// To change the context of a request, such as an incoming request you
// want to modify before sending back out, use Request.Clone. Between
// those two uses, it's rare to need WithContext.

I was a bit unsure, too. But this let me to use `Clone` instead.

timebertt:
Ok, guys. I've added another commit addressing the gardenlet bootstrapping process.
It is now falling back to `certificates/v1beta1` if `v1` is not available.
I successfully tested the gardenlet bootstrapping (with bootstrap token) and certificate rotation in my local setup on the following kubernetes versions:
- v1.16.15
- v1.17.9
- v1.18.15
- v1.19.7 (+ ServiceAccount bootstrapping)
- v1.20.2

I will now update our fork in order to already vendor kubernetes-sigs/structured-merge-diff#182.
I hope we should be good to go then.

extensions/pkg/webhook/handler_shootclient.go
rfranzke: Just to be sure: We don't have a problem here, right? It would still work with v1.10 clusters because the `admissionv1` objects are only used internally in the webhook code?

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
rfranzke: Same here (https://github.com/gardener/gardener/pull/3393#discussion_r557360687)?

pkg/utils/kubernetes/kubernetes.go
```diff
@@ -111,6 +111,7 @@ func Key(namespaceOrName string, nameOpt ...string) client.ObjectKey {
 }
 
 // KeyFromObject obtains the client.ObjectKey from the given metav1.Object.
+// Deprecated: use client.ObjectKeyFromObject instead.
```
ialidzhikov, line 114: As this func is now deprecated, shouldn't we replace all of its usage under `extensions/` to `client.ObjectKeyFromObject`?

extensions/pkg/webhook/remote_addr.go
timuthy: Okay, thanks. I was concerned about cloning because incoming requests might be big in size, but the implementation doesn't seem to copy the body. So we should be fine.

timebertt:
I've pushed a cherry-pick of https://github.com/kubernetes-sigs/structured-merge-diff/pull/182 to our `structured-merge-diff` fork and pinned the dependency to that again in order to not regress #2977.

I also tested all the metrics endpoints as I also removed the `replace` stmt for the prometheus lib, looks good to me so far.

With that, all remaining open ToDos should be resolved, I hope :)

extensions/pkg/webhook/handler_shootclient.go
timebertt: Yes, we use the c-r handlers now which take over decoding and caring about the right version of the `admission` API.

cmd/gardener-seed-admission-controller/app/gardener_seed_admission_controller.go
timebertt: Yes, same here :)

rfranzke (dismissed):
/lgtm

extensions/pkg/predicate/predicate_test.go
```diff
@@ -115,21 +117,24 @@ var _ = Describe("Predicate", func() {
 		)
 
 		BeforeEach(func() {
-			objectMeta := metav1.ObjectMeta{
-				Name: name,
+			configMpa := &corev1.ConfigMap{
```
ialidzhikov:
```suggestion
			configMap := &corev1.ConfigMap{
```

ialidzhikov (commented):
Minor nits inline.
My gardenlet logs are also spammed with
```
W0114 15:37:08.793640   71266 warnings.go:67] admissionregistration.k8s.io/v1beta1 ValidatingWebhookConfiguration is deprecated in v1.16+, unavailable in v1.22+; use admissionregistration.k8s.io/v1 ValidatingWebhookConfiguration
W0114 15:37:08.854582   71266 warnings.go:67] admissionregistration.k8s.io/v1beta1 MutatingWebhookConfiguration is deprecated in v1.16+, unavailable in v1.22+; use admissionregistration.k8s.io/v1 MutatingWebhookConfiguration
W0114 15:37:40.472323   71266 warnings.go:67] admissionregistration.k8s.io/v1beta1 ValidatingWebhookConfiguration is deprecated in v1.16+, unavailable in v1.22+; use admissionregistration.k8s.io/v1 ValidatingWebhookConfiguration
W0114 15:37:40.532682   71266 warnings.go:67] admissionregistration.k8s.io/v1beta1 MutatingWebhookConfiguration is deprecated in v1.16+, unavailable in v1.22+; use admissionregistration.k8s.io/v1 MutatingWebhookConfiguration
```

which might be a huge log pollution when we deploy this on a real landscape?

pkg/utils/kubernetes/kubernetes.go
timebertt: Yes, we should. Though, I wanted to do it separately from this PR as it's not strictly required for getting the vendor through and would increase the size of the PR again.
There are also a lot of other similar follow-up / cleanup tasks listed in https://github.com/gardener/gardener/issues/3109.

timebertt:
> My gardenlet logs are also spammed with ...

I haven't seen those logs in my test runs, as I haven't created `v1.19+` shoots, I guess.
It probably comes from the care controller, right? https://github.com/gardener/gardener/blob/0d884cc8291f2a6dafe977402ecc616a4c23042d/pkg/operation/botanist/constraints_check.go#L102

I guess, we can't just simply switch to `admissionregistration/v1` because it's only available from `v1.16` upwards, if I see it correctly.

timuthy:
> > My gardenlet logs are also spammed with ...
> 
> I haven't seen those logs in my test runs, as I haven't created `v1.19+` shoots, I guess.
> It probably comes from the care controller, right?
> 
> https://github.com/gardener/gardener/blob/0d884cc8291f2a6dafe977402ecc616a4c23042d/pkg/operation/botanist/constraints_check.go#L102
> 
> I guess, we can't just simply switch to `admissionregistration/v1` because it's only available from `v1.16` upwards, if I see it correctly.

Changing the default warning handler might be an alternative in this case: https://kubernetes.io/blog/2020/09/03/warnings/#customize-client-handling

timebertt:
> Changing the default warning handler might be an alternative in this case: https://kubernetes.io/blog/2020/09/03/warnings/#customize-client-handling

Yes, this seems like the better option for now.

I added a deduplicating warnings handler to all components.
@ialidzhikov @timuthy @rfranzke PTAL :)

rfranzke (approved):

ialidzhikov (approved):
/lgtm

timuthy (approved):
/lgtm

timebertt:
Nice! Thanks a lot guys for your thorough and fast reviews! ❤️ 
This means we are good to merge, once CI has passed, right?

rfranzke:
😱 Probably :)

PR #3394, opened 2021-01-13
Make ingress domain for Seeds immutable

### LINKED ISSUE #3281: Changing Seed IngressDomain will make Shoot monitoring unavailable until next Shoot Reconcile
/area networking
/kind bug

/priority normal

**What happened**:
After changing the `seed.spec.dns.ingressDomain` the monitoring Ingresses in the Shoot namespaces of the Seed will keep the old ingressDomain until the Shoots get [reconciled](https://github.com/gardener/gardener/blob/6a4526b135da570f567c2fb5bd597bd07e02b915/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go#L386).
If the Shoots are reconciled at a later point in time (for example in a later maintenance window) then Shoot monitoring will be unavailable until then.

Changing the ingressDomain in the Seed should not disable Shoot monitoring until the next Shoot reconcile.

Seeing two options currently:

A) Add migration logic to adapt Ingresses when the Seed is reconciled
B) Making ingressDomain immutable

BeckerMax:

  /area control-plane
  /area robustness
/kind bug
/priority normal

Make ingress domain for Seeds immutable

Changing the ingress domain on running Seed clusters can cause
a temporary unavailability of the Shoots monitoring.

We decided to not support the use case of changing the
ingress Domain for Seeds. This is in line with the
Shoot resource where the ingress domain is also immutable.

Fixes https://github.com/gardener/gardener/issues/3281

**Release note**:
```noteworthy operator
The ingress domain configuration for Seeds is now immutable.
```

### DIFF
--- a/hack/api-reference/core.md
+++ b/hack/api-reference/core.md
@@ -3082,7 +3082,7 @@ string
 <td>
 <em>(Optional)</em>
 <p>Domain is the external available domain of the Shoot cluster. This domain will be written into the
-kubeconfig that is handed out to end-users.</p>
+kubeconfig that is handed out to end-users. Once set it is immutable.</p>
 </td>
 </tr>
 <tr>
@@ -3771,7 +3771,7 @@ string
 </td>
 <td>
 <p>Domain specifies the IngressDomain of the Seed cluster pointing to the ingress controller endpoint. It will be used
-to construct ingress URLs for system applications running in Shoot clusters.</p>
+to construct ingress URLs for system applications running in Shoot clusters. Once set this field is immutable.</p>
 </td>
 </tr>
 <tr>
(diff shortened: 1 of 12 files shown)


BeckerMax:
Is there some expected documentation about which fields are immutable?

danielfoehrKn:
> Is there some expected documentation about which fields are immutable?

Would be nice to have, but I do not think it exists yet.

danielfoehrKn (dismissed):
/lgtm

rfranzke (dismissed):
/lgtm

timuthy:
> Is there some expected documentation about which fields are immutable?

We could add it to the doc string of the API field, so that it'll be mentioned in the API ref (https://gardener.cloud/documentation/references).

vpnachev (approved):
/lgtm

rfranzke (approved):

PR #3395, opened 2021-01-13
Add side-car container to kube-proxy to remove incorrect conntrack table entries

DockToFuture:

/area networking
/kind enhancement
/priority normal

We observed that from the shoot cluster the kubernetes `api-server` was sometimes unavailable.

This PR mitigates the issue by deleting the incorrect conntrack table entries, that might appear after restarting the `kube-proxy`. These incorrect conntrack table entries prevented the establishment of a tcp connection to the `api-server`.

After deploying this PR the `ApiServerUnreachableViaKubernetesService` alerts are expected to disappear.
**Release note**:

```bugfix operator
A side-car container is added to `kube-proxy` that deletes the incorrect conntrack table entries which sometime occur after restart of `kube-proxy` and prevent the establishment of a tcp connection to the `api-server`.
```

### DIFF
--- a/charts/shoot-core/components/charts/kube-proxy/templates/conntrack-fix-script.yaml
+++ b/charts/shoot-core/components/charts/kube-proxy/templates/conntrack-fix-script.yaml
@@ -1,0 +1,38 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: conntrack-fix-script
+  namespace: kube-system
+  labels:
+    app: kubernetes
+    gardener.cloud/role: system-component
+    origin: gardener
+    role: proxy
+data:
+  conntrack_fix.sh: |
+    #!/bin/sh -e
+    trap "kill -s INT 1" TERM
+    apk add conntrack-tools
+    sleep 120 & wait
+    date
+    # conntrack example:
+    # tcp      6 113 SYN_SENT src=21.73.193.93 dst=21.71.0.65 sport=1413 dport=443 \
+    #   [UNREPLIED] src=21.71.0.65 dst=21.73.193.93 sport=443 dport=1413 mark=0 use=1
+    eval "$(
+      conntrack -L -p tcp --state SYN_SENT \
+      | sed 's/=/ /g'                      \
+      | awk '$6 !~ /^10\./ &&
+             $8 !~ /^10\./ &&
+             $6  == $17    &&
+             $8  == $15    &&
+             $10 == $21    &&
+             $12 == $19 {
+               printf "conntrack -D -p tcp -s %s --sport %s -d %s --dport %s;\n",
+                                              $6,        $10,  $8,        $12}'
+    )"
+    while true; do
+      date
+      sleep 3600 & wait
+    done
+
--- a/charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
+++ b/charts/shoot-core/components/charts/kube-proxy/templates/kube-proxy-daemonset.yaml
@@ -21,6 +21,7 @@ spec:
         checksum/secret-kube-proxy: {{ include (print $.Template.BasePath "/kube-proxy-secret.yaml") . | sha256sum }}
         checksum/configmap-componentconfig: {{ include (print $.Template.BasePath "/componentconfig.yaml") . | sha256sum }}
         checksum/configmap-kube-proxy-cleanup-script: {{ include (print $.Template.BasePath "/kube-proxy-cleanup-script.yaml") . | sha256sum }}
+        checksum/configmap-conntrack-fix-script: {{ include (print $.Template.BasePath "/conntrack-fix-script.yaml") . | sha256sum }}
 {{- if .Values.podAnnotations }}
 {{ toYaml .Values.podAnnotations | indent 8 }}
 {{- end }}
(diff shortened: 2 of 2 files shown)


charts/shoot-core/components/charts/kube-proxy/templates/conntrack-fix-script.yaml
```diff
@@ -0,0 +1,38 @@
...
+    app: kubernetes
+    gardener.cloud/role: system-component
+    origin: gardener
+    role: proxy
+data:
+  conntrack_fix.sh: |
+    #!/bin/sh -e
+    trap "kill -s INT 1" TERM
+    apk add conntrack-tools
+    sleep 120 & wait
```
rfranzke, line 17: I guess you want to wait for kube-proxy process to start-up and do something meaningful here, right? I guess it's "ok" for such a mitigation for now, but generally it can take longer than 2m for the kube-proxy to startup while this container is already running (e.g., delayed image pull). It would be better to have a check if the kube-proxy process is running already. If you have an idea how to easily achieve that? As said, I don't insist on it because it's just a mitigation, but I wanted to point it out.

rfranzke (commented):
Looks good and works well locally. Just to confirm: The script is supposed to only cleanup `conntrack` entries 2 minutes after start-up but not during the lifetime of the kube-proxy itself, right?

DockToFuture:
> Looks good and works well locally. Just to confirm: The script is supposed to only cleanup `conntrack` entries 2 minutes after start-up but not during the lifetime of the kube-proxy itself, right?

Yes, it should only clean it up once after the restart of kube-proxy.

rfranzke (approved):
/lgtm

charts/shoot-core/components/charts/kube-proxy/templates/conntrack-fix-script.yaml
DockToFuture: I could replace it with 
``` bash
until pgrep kube-proxy; do
  sleep 2 & wait
done
```
and set `shareProcessNamespace: true` in the pod

I think other ways for ipc are much more complicated.

WDYT?

rfranzke: Yeah, I don't insist on this now, just wanted to point it out. If we find an easy solution then sure, otherwise I think we can still go ahead with such mitigation for now.

Your proposal looks good, but then you'd have to adapt the trap in line 15 as well, right (your script is not necessarily PID 1)? cc @istvanballok

wyb1 (approved):

charts/shoot-core/components/charts/kube-proxy/templates/conntrack-fix-script.yaml
DockToFuture: Sure, will do that. Should be sufficient to replace pid 1 with `$(pgrep "/bin/sh")`

ialidzhikov:
> This PR mitigates the issue by deleting the incorrect conntrack table entries, that might appear after restarting the kube-proxy. These incorrect conntrack table entries prevented the establishment of a tcp connection to the api-server.

Is it an upstream bug in kube-proxy? If yes, can you please link corresponding issues.

charts/shoot-core/components/charts/kube-proxy/templates/conntrack-fix-script.yaml
DockToFuture: I discussed with @wyb1 and @istvanballok that we will leave it like this for now. We will concentrate on fixing this in `kube-proxy`.

rfranzke: Sure, thanks. Merging then :)

rfranzke:
@ialidzhikov I guess there is no issue yet, @DockToFuture and colleagues, please cross-post the issue here after you have opened it in k/k.

DockToFuture:
> > This PR mitigates the issue by deleting the incorrect conntrack table entries, that might appear after restarting the kube-proxy. These incorrect conntrack table entries prevented the establishment of a tcp connection to the api-server.
> 
> Is it an upstream bug in kube-proxy? If yes, can you please link corresponding issues.

Yes we think it is an upstream bug but we will first try to find an easier way to make it reproducible and then create the corresponding upstream issue.

ialidzhikov:
> > > This PR mitigates the issue by deleting the incorrect conntrack table entries, that might appear after restarting the kube-proxy. These incorrect conntrack table entries prevented the establishment of a tcp connection to the api-server.
> > 
> > 
> > Is it an upstream bug in kube-proxy? If yes, can you please link corresponding issues.
> 
> Yes we think it is an upstream bug but we will first try to find an easier way to make it reproducible and then create the corresponding upstream issue.

Ok, I am looking forward to it.

ialidzhikov:
Cherry-picked into `release-v1.15` with 243cfebaa3a1d9edbb4e46349639dfdfb776f15f.

ialidzhikov:
> > > > This PR mitigates the issue by deleting the incorrect conntrack table entries, that might appear after restarting the kube-proxy. These incorrect conntrack table entries prevented the establishment of a tcp connection to the api-server.
> > > 
> > > 
> > > Is it an upstream bug in kube-proxy? If yes, can you please link corresponding issues.
> > 
> > 
> > Yes we think it is an upstream bug but we will first try to find an easier way to make it reproducible and then create the corresponding upstream issue.
> 
> Ok, I am looking forward to it.

ping @DockToFuture . I am still looking forward to it.

DockToFuture:
Progress is tracked in `gardener-monitoring` slack channel.

PR #3396, opened 2021-01-13
Improve cloud-config execution observability and enable systemd service restarts

### LINKED ISSUE #3237: Improved observability for cloud-config user-data applications on worker nodes

/area monitoring
/kind enhancement
/priority normal

1. The `Secret`s downloaded by the `cloud-config-downloader` systemd unit (running on all shoot worker nodes) should have a `checksum/data` annotation whose value is the checksum of the `Secret`s' `.data` map.
1. The `cloud-config-downloader` reads this annotation of the `Secret` and annotates the `Node` object corresponding to the VM it is running on with `checksum/cloud-config-data` whose value is the same like the `checksum/data`. It uses the kubelet's kubeconfig (with a dedicated user-agent header) so that it is limited to only the `Node` of its VM.
1. The shoot reconciliation flow can be enhanced with a wait step that compares the checksum annotation of the `Secret` with the checksum annotations of all nodes in order to identify nodes that still need to be updated with the most recent cloud-config user-data.
1. The shoot care controller flow can be enhanced with an additional check like in 3. It can report this as part of the `EveryNodeReady` condition.

⚠️ The cloud-config-downloader script is never updated for existing VMs, i.e., those improvements will only be possible for new clusters or for those whose machines have been rolled/replaced (all of them). Consequently, let's make all of this completely optional or only consider it for newly created shoots.

**Why is this needed**:
Improved `EveryNodeReady` condition with information whether it was updated with the most recent cloud-config user-data or whether this is still pending.

### LINKED ISSUE #3246: Annotation on Node for triggering kubelet restart

/area ops-productivity control-plane-migration disaster-recovery
/kind enhancement
/priority normal

Let's add an annotation, e.g. `node.gardener.cloud/operation=restart-kubelet` or something similar, that will, when added to a `Node` object, cause the `cloud-config` script on that node to restart the kubelet unit.

**Why is this needed**:

We have seen several occurrences in the past, where kubernetes components including the kubelet, were working on outdated data after restoring `etcd-main` and therefore receive error messages from the API server like this one:
```
Nov 25 08:46:15 shoot--garden--ccee-m3-eu1-cpu-worker-z1-85994-bhxxn kubelet[2003273]: E1125 08:46:15.079258 2003273 reflector.go:153] object-"shoot--d072924--hc-haas"/"gardener-resource-manager": Failed to list *v1.Secret: secrets "gardener-resource-manager" is forbidden: User "system:node:shoot--garden--ccee-m3-eu1-cpu-worker-z1-85994-bhxxn" cannot list resource "secrets" in API group "" in the namespace "shoot--d072924--hc-haas": no relationship found between node "shoot--garden--ccee-m3-eu1-cpu-worker-z1-85994-bhxxn" and this object
```

Restarting the kubelet helps in such circumstances, so let's add such an annotation in order to partly automate the procedure.
This will save operators quite some effort in such situations and might also be triggered by gardener itself in the future when it knows, that an etcd restoration happened, e.g. as part of a control plane migration.

rfranzke:

/area monitoring ops-productivity
/kind enhancement
/priority normal

This PR improves the following things:
1. Shoot reconciliation flow waits until all worker nodes have executed the most recent cloud-config user data.
2. Shoot care controller checks if the last successfully applied cloud-config user data on the nodes is outdated.
3. A new annotation on shoot `Node` objects allows to easily restart systemd services without the necessity to SSH into it.

Please note that 1. and 2. are optional features to ensure backwards-compatibility (they require a change in the `cloud-config-downloader` script which is only touched on rolling updates of nodes) i.e., existing clusters with long-living nodes will not get these features (only after the nodes were rolled).

Fixes #3237
Fixes #3246

The PR has 5 commits. The first 4 commits correspond to the 4 steps described in #3237. The fifth commit is about #3246 only. I combined both issues in one PR because the #3246 was a simple change of the same code. If you prefer to have it separately I can drop it and open a second PR, but the change is rather small and shouldn't distract from the rest.
/squash

**Release note**:

```feature user
The shoot reconciliation flow is now waiting until all worker nodes have executed the most recent cloud-config user data. Similarly, the shoot care controller checks if the last successfully applied cloud-config user data on all nodes is outdated or not (and reports this in the `EveryNodeReady` condition). Please note that both features are only available for new nodes.
```
``` feature operator
It is now possible to trigger a restart of systemd services on particular shoot worker nodes by annotating the corresponding `Node` object with `worker.gardener.cloud/restart-systemd-services=kubelet`, for example.
```


charts/seed-operatingsystemconfig/downloader/templates/scripts/_download-cloud-config.sh
```diff
@@ -3,24 +3,27 @@
...
 
   echo "Could not retrieve the cloud config script in secret with name $SECRET_NAME"
   exit 1
 fi
 
-echo "$SCRIPT" | base64 -d | bash
+SCRIPT="$(echo "$SECRET" | sed -rn 's/  script: (.*)/\1/p')"
+CHECKSUM="$(echo "$SECRET" | sed -rn 's/    checksum\/data-script: (.*)/\1/p' | sed -e 's/^"//' -e 's/"$//')"
+
+echo "$SCRIPT" | base64 -d | bash && echo "$CHECKSUM" > "$DIR_CLOUDCONFIG_DOWNLOADER/downloaded_checksum"
```
vpnachev:
```suggestion
echo "$CHECKSUM" > "$DIR_CLOUDCONFIG_DOWNLOADER/downloaded_checksum"
echo "$SCRIPT" | base64 -d | bash 
```

Otherwise the first execution of the new script will still use the old annotation.

charts/shoot-cloud-config/templates/scripts/_cloud-config-script.sh
```diff
@@ -104,4 +105,27 @@ if ! diff "$PATH_CLOUDCONFIG" "$PATH_CLOUDCONFIG_OLD" >/dev/null; then
 fi
 
 rm "$PATH_CLOUDCONFIG"
+
+ANNOTATION_RESTART_SYSTEMD_SERVICES="worker.gardener.cloud/restart-systemd-services"
+
+# Try to find Node object for this machine
+if [[ -f "$DIR_KUBELET/kubeconfig-real" ]]; then
+  {{`NODE="$(/opt/bin/kubectl --kubeconfig="$DIR_KUBELET/kubeconfig-real" get node -o go-template="{{ range \$i, \$item := .items }}{{ range \$j, \$address := \$item.status.addresses }}{{ if and (eq \$address.type \"Hostname\") (eq \$address.address \"$(hostname)\") }}{{ \$item.metadata.name }}{{ if (index \$item.metadata.annotations \"$ANNOTATION_RESTART_SYSTEMD_SERVICES\") }} {{ index \$item.metadata.annotations \"$ANNOTATION_RESTART_SYSTEMD_SERVICES\" }}{{ end }}{{ end }}{{ end }}{{ end }}")"`}}
```
vpnachev: I have no strong opinion, but I believe using the label [kubernetes.io/hostname](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#kubernetesiohostname) can simplify the request.

I mean
```bash
k get node -l kubernetes.io/hostname=$(hostname)
```

docs/usage/shoot_operations.md
```diff
@@ -40,3 +40,17 @@ Please note that only the token (and basic auth password, if enabled) are exchan
...
 ```
+
+## Restart systemd services on particular worker nodes
+
+It is possible to make Gardener restart particular systemd services on your shoot worker nodes if needed.
+The annotation is not set on the `Shoot` resource but directly on the `Node` object you want to target.
+For example, the following will restart the `kubelet`:
+
+```bash
+kubectl annotate node <node-name> worker.gardener.cloud/restart-systemd-services=kubelet
```
vpnachev: I believe it would be better if the documentation shows that multiple services can be restarted with single annotation, e.g.:
```suggestion
For example, the following will restart the `docker` and `kubelet`:

```bash
kubectl annotate node <node-name> worker.gardener.cloud/restart-systemd-services=docker,kubelet
```

vpnachev (commented):
I have some minor suggestions.

charts/shoot-cloud-config/templates/scripts/_cloud-config-script.sh
rfranzke: :) nice, didn't think about this, of course we can do it this way, thanks!

charts/shoot-cloud-config/templates/scripts/_cloud-config-script.sh
```diff
@@ -104,4 +105,27 @@ if ! diff "$PATH_CLOUDCONFIG" "$PATH_CLOUDCONFIG_OLD" >/dev/null; then
...
+
+  # Update checksum/cloud-config-data annotation on Node object if possible
+  if [[ ! -z "$NODENAME" ]] && [[ -f "$PATH_CHECKSUM" ]]; then
+    /opt/bin/kubectl --kubeconfig="$DIR_KUBELET/kubeconfig-real" annotate node "$NODENAME" "checksum/cloud-config-data=$(cat "$PATH_CHECKSUM")" --overwrite
+  fi
+
+  # Restart systemd services if requested
+  for service in $(echo "$SYSTEMD_SERVICES_TO_RESTART" | sed "s/,/ /g"); do
+    echo "Restarting systemd service $service due to $ANNOTATION_RESTART_SYSTEMD_SERVICES annotation"
+    systemctl restart "$service"
```
danielfoehrKn: To make it safer to use, we could check first if the systemd service actually exists. 
Otherwise, the script will silently fail and not remove the annotation. 
So every 30 seconds the services will be restarted. 
In the worst-case someone specified something like "kubelet, dockers" which will kill the node due to a spelling mistake.

charts/shoot-cloud-config/templates/scripts/_cloud-config-script.sh
danielfoehrKn: > eq \$address.type \"Hostname\") (eq \$address.address \"$(hostname)

What is the reason behind this check?
Thanks!

docs/usage/shoot_operations.md
```diff
@@ -40,3 +40,17 @@ Please note that only the token (and basic auth password, if enabled) are exchan
...
+```
+
+It may take up to a minute until the service is restarted.
+The annotation will be removed from the `Node` object afterwards.
```
danielfoehrKn:
```suggestion
The annotation will be removed from the `Node` object after every specified systemd service has been restarted.
```

docs/usage/shoot_operations.md
```diff
@@ -40,3 +40,17 @@ Please note that only the token (and basic auth password, if enabled) are exchan
...
+You could verify when/whether the kubelet restarted by using `kubectl describe node <node-name>` and looking for such a `Starting kubelet` event.
```
danielfoehrKn:
```suggestion
In the example mentioned above, you could additionally verify when/whether the kubelet restarted by using `kubectl describe node <node-name>` and looking for such a `Starting kubelet` event.
```

charts/shoot-cloud-config/templates/scripts/_cloud-config-script.sh
rfranzke: We have to find the name of the Kubernetes `Node` object to be able to perform `kubectl` commands with it.

charts/shoot-cloud-config/templates/scripts/_cloud-config-script.sh
rfranzke: Yeah, nice catch. What about something like 
```suggestion
    systemctl restart "$service" || true
```
to prevent that such mistake can lead to the annotation not being removed?

rfranzke:
Thanks for the reviews @vpnachev and @danielfoehrKn, I've pushed 53ad6df to address your suggestions.

charts/shoot-cloud-config/templates/scripts/_cloud-config-script.sh
danielfoehrKn: sure, good idea

danielfoehrKn (dismissed):

charts/shoot-cloud-config/templates/scripts/_cloud-config-script.sh
```diff
@@ -104,4 +105,28 @@ if ! diff "$PATH_CLOUDCONFIG" "$PATH_CLOUDCONFIG_OLD" >/dev/null; then
 fi
 
 rm "$PATH_CLOUDCONFIG"
+
+ANNOTATION_RESTART_SYSTEMD_SERVICES="worker.gardener.cloud/restart-systemd-services"
+
+# Try to find Node object for this machine
+if [[ -f "$DIR_KUBELET/kubeconfig-real" ]]; then
+  {{`NODE="$(/opt/bin/kubectl --kubeconfig="$DIR_KUBELET/kubeconfig-real" get node -l "kubernetes.io/hostname=$(hostname)" -o go-template="{{ if .items }}{{ (index .items 0).metadata.name }}{{ if (index (index .items 0).metadata.annotations \"$ANNOTATION_RESTART_SYSTEMD_SERVICES\") }} {{ index (index .items 0).metadata.annotations \"$ANNOTATION_RESTART_SYSTEMD_SERVICES\" }}{{ end }}")"`}}
```
vpnachev: I have some troubles with this command, logs from the node:
```
download-cloud-config.sh[10104]: error: error parsing template {{ if .items }}{{ (index .items 0).metadata.name }}{{ if (index (index .items 0).metadata.annotations "worker.gardener.cloud/restart-systemd-services") }} {{ index (index .items 0).metadata.annotations "worker.gardener.cloud/restart-systemd-services" }}{{ end }}, template: output:1: unexpected EOF
```

vpnachev: Some more detailed logs
```
root@ip-10-250-10-253:/var/lib/cloud-config-downloader# bash -eux tmp.sh
+ DIR_KUBELET=/var/lib/kubelet
+ DIR_CLOUDCONFIG_DOWNLOADER=/var/lib/cloud-config-downloader
+ DIR_CLOUDCONFIG=/var/lib/cloud-config-downloader/downloads
+ PATH_CLOUDCONFIG_DOWNLOADER_SERVER=/var/lib/cloud-config-downloader/credentials/server
+ PATH_CLOUDCONFIG_DOWNLOADER_CA_CERT=/var/lib/cloud-config-downloader/credentials/ca.crt
+ PATH_CLOUDCONFIG=/var/lib/cloud-config-downloader/downloads/cloud_config
+ PATH_CLOUDCONFIG_OLD=/var/lib/cloud-config-downloader/downloads/cloud_config.old
+ PATH_CHECKSUM=/var/lib/cloud-config-downloader/downloaded_checksum
+ mkdir -p /var/lib/cloud-config-downloader/downloads /var/lib/kubelet
+ docker-preload hyperkube k8s.gcr.io/hyperkube:v1.18.5
+ name=hyperkube
+ image=k8s.gcr.io/hyperkube:v1.18.5
+ echo 'Checking whether to preload hyperkube from k8s.gcr.io/hyperkube:v1.18.5'
Checking whether to preload hyperkube from k8s.gcr.io/hyperkube:v1.18.5
++ docker images -q k8s.gcr.io/hyperkube:v1.18.5
+ '[' -z 5220d2573479 ']'
+ echo 'No need to preload hyperkube from k8s.gcr.io/hyperkube:v1.18.5'
No need to preload hyperkube from k8s.gcr.io/hyperkube:v1.18.5
+ base64 -d
+ cat
+ '[' '!' -f /var/lib/cloud-config-downloader/downloads/cloud_config.old ']'
+ [[ ! -f /var/lib/kubelet/kubeconfig-real ]]
+ [[ ! -f /var/lib/kubelet/pki/kubelet-client-current.pem ]]
+ rm -f /var/lib/kubelet/kubeconfig-bootstrap
+ diff /var/lib/cloud-config-downloader/downloads/cloud_config /var/lib/cloud-config-downloader/downloads/cloud_config.old
+ rm /var/lib/cloud-config-downloader/downloads/cloud_config
+ ANNOTATION_RESTART_SYSTEMD_SERVICES=worker.gardener.cloud/restart-systemd-services
+ [[ -f /var/lib/kubelet/kubeconfig-real ]]
+++ hostname
++ /opt/bin/kubectl --kubeconfig=/var/lib/kubelet/kubeconfig-real get node -l kubernetes.io/hostname=ip-10-250-10-253.eu-west-1.compute.internal -o 'go-template={{ if .items }}{{ (index .items 0).metadata.name }}{{ if (index (index .items 0).metadata.annotations "worker.gardener.cloud/restart-systemd-services") }} {{ index (index .items 0).metadata.annotations "worker.gardener.cloud/restart-systemd-services" }}{{ end }}'
error: error parsing template {{ if .items }}{{ (index .items 0).metadata.name }}{{ if (index (index .items 0).metadata.annotations "worker.gardener.cloud/restart-systemd-services") }} {{ index (index .items 0).metadata.annotations "worker.gardener.cloud/restart-systemd-services" }}{{ end }}, template: output:1: unexpected EOF
+ NODE=

```

vpnachev (changes requested):

rfranzke:
Thanks & sorry @vpnachev, it should be fixed with 46ec06e

vpnachev (approved):
/lgtm

PR #3267, opened 2020-12-05
Upgrade konnectivity components to v0.0.14

mvladev:
**How to categorize this PR?**

/area control-plane
/area networking
/kind enhancement 
/priority normal

**What this PR does / why we need it**:

Use the new `replica-reloader` to handle dynamic scaling of the `kube-apiserver`.

**Which issue(s) this PR fixes**:

Part of #2956
Fixes #3208 

**Special notes for your reviewer**:
 
In another PR I'll move the `konnectivity-server` to a dedicated `Deployment` that would greatly reduce the need of restarting the `konnectivity-server`.

**Release note**:

```bugfix operator
`KonnectivityTunnel`'s stability is improved and now handles `kube-apiserver` autoscaling. It properly sets `--server-count` of `konnectivity-server` on such event.
```

/assign @DockToFuture

Issue #2956, opened 2020-10-09
☂️ -Issue for graduating KonnectivityTunnel feature gate to beta


einfachnuralex:
**How to categorize this issue?**

/area control-plane
/kind enhancement
/priority normal

**What would you like to be added:**
The Konnectivity feature is in alpha state at the moment and has serias stability issues.
It make sense to collect all problems here and track the progress to beta. 

Feel free to add your problems with Konnectivity feature here.

einfachnuralex:
Currently we run version 0.0.12 of the apiserver-network-proxy in a test environment and use the http-connect method of this proxy. 
Kubernetes version is 1.18.6/1.19.2 and we are working on a Openstack environment. Also we tried the GRPC connection method, but with the same results.
We implemented a liveness probe for the proxy which checks if the proxy is crashed, that was a problem in the past because of to low resource limits. 
Also we added the parameter `--delete-existing-uds-file=true` to the server, because a restart of the proxy server sometimes leed to a crash of the proxy because the existing socket file. 

But the connections over the apiserver-network-proxy is still unstable. Sometimes it runs for days without problems, sometimes there are connection issues frequently a day.

As we use shooted seeds the controlplane and extensions run in different clusters and communicate over the proxy. Then the webhook calls fails and the seed is not able to work properly.

Here are some sample errors:
```
I1002 07:04:00.489822       1 trace.go:116] Trace[34027178]: "Call mutating webhook" configuration:gardener-seed-admission-controller,webhook:pods.seed.admission.core.gardener.cloud,resource:/v1, Resource=pods,subresource:,operation:CREATE,UID:7191bc99-d482-4223-939c-9ba046b2304e (started: 2020-10-02 07:03:50.489203698 +0000 UTC m=+59064.563525480) (total time: 10.000557727s):             
W1002 07:04:00.489909       1 dispatcher.go:169] Failed calling webhook, failing open pods.seed.admission.core.gardener.cloud: failed calling webhook "pods.seed.admission.core.gardener.cloud": Post https://gardener-seed-admission-controller.garden.svc:443/webhooks/mutate-pods?timeout=10s: context deadline exceeded
E1002 07:04:00.489935       1 dispatcher.go:170] failed calling webhook "pods.seed.admission.core.gardener.cloud": Post https://gardener-seed-admission-controller.garden.svc:443/webhooks/mutate-pods?timeout=10s: context deadline exceeded
---
[...]

PR #3399, opened 2021-01-14
konnectivity-server is separate deployment when SNI is enabled

mvladev:

/area networking
/area control-plane
/kind enhancement
/priority normal

Shoot's `kube-apiserver` scales automatically with load. On scale-up events it could lead a situation where existing `konnectivity-agents` are not connected to the `konnectivity-server` sidecar for that API server.

This happens for some period of time until other `konnectivity-servers` are restarted by `replica-reloader`.

By moving `konnectivity-server` in a dedicated `Deployment`, `konnectivity-agents` connections are not closed when scale up/down of the `kube-apiserver` happens.

Continuation of #3267
Part of #2956

**Release note**:

```feature operator
`konnectivity-server` is no longer a sidecar to `kube-apiserver` when `APIServerSNI` and `KonnectivityTunnel` feature gates are enabled. This allows for existing opened tunnel connections from `konnectivity-agents` to not be interrupted when `kube-apiserver` is auto-scaled. 
```


rfranzke:
cc/FYI @danielfoehrKn

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -291,6 +291,16 @@ func (c *Controller) runReconcileShootFlow(o *operation.Operation) *gardencorev1
 			Fn:           flow.TaskFn(botanist.DeployKubeScheduler).RetryUntilTimeout(defaultInterval, defaultTimeout),
 			Dependencies: flow.NewTaskIDs(deploySecrets, waitUntilKubeAPIServerIsReady),
 		})
+		deployKonnectivityServer = g.Add(flow.Task{
+			Name:         "Deploying konnectivity-server",
+			Fn:           flow.TaskFn(botanist.DeployKonnectivityServer).RetryUntilTimeout(defaultInterval, defaultTimeout),
+			Dependencies: flow.NewTaskIDs(deploySecrets, waitUntilKubeAPIServerIsReady),
```
rfranzke: Any reason why we have to wait for KAPI to be ready? Can we deploy the k-server simultaneously to the KAPI deployment?

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,552 @@
...
+		name                          = ServerName
+		adminPort               int32 = 8133
+		healthPort              int32 = 8134
+		serverAgentPortUnsigned       = uint32(ServerAgentPort)
+	)
+
+	var (
+		minAvailable = intstr.FromInt(1)
+		tcpProto     = corev1.ProtocolTCP
+		serviceHost  = fmt.Sprintf("%s.%s.svc.cluster.local", name, k.Namespace)
```
rfranzke: You can use `gardencorev1beta1.DefaultDomain`

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,552 @@
...
+		}}
+		deployment = &appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: k.Namespace,
+				Labels:    getLabels(),
+			},
+			Spec: appsv1.DeploymentSpec{
+				Replicas:             pointer.Int32Ptr(k.Replicas),
+				RevisionHistoryLimit: pointer.Int32Ptr(0),
```
rfranzke:
```suggestion
				RevisionHistoryLimit: pointer.Int32Ptr(1),
```

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,552 @@
...
+								"--jitter-factor=5",
+								"--v=2",
+								"--",
+								"/proxy-server",
+								"--logtostderr=true",
+								"--cluster-cert=/certs/konnectivity-server/konnectivity-server.crt",
+								"--cluster-key=/certs/konnectivity-server/konnectivity-server.key",
+								"--server-cert=/certs/konnectivity-server/konnectivity-server.crt",
+								"--server-key=/certs/konnectivity-server/konnectivity-server.key",
+								"--server-ca-cert=/certs/client-ca/ca.crt",
```
rfranzke: Can you have some variables for these paths that can be used both here and below in the `volumeMount` section?

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,552 @@
...
+								"--agent-namespace=kube-system",
```
rfranzke: `metav1.NamespaceSystem`

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,552 @@
...
+								fmt.Sprintf("--agent-service-account=%s", AgentName),
+								"--kubeconfig=/auth/kubeconfig",
```
rfranzke: same as above (variables for path)

rfranzke (dismissed):
Awesome PR, love it :) Thanks!

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,552 @@
...
+								}},
+							},
+						},
+						ServiceAccountName: name,
+						Containers: []corev1.Container{{
+							Name:            name,
+							Image:           k.Image,
+							ImagePullPolicy: corev1.PullIfNotPresent,
+							Args: []string{
+								fmt.Sprintf("--namespace=%s", k.Namespace),
```
danielfoehrKn, line 233: where is the  `/replica-reloader` command?

pkg/operation/botanist/konnectivity_server.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		return konnectivity.OpDestroy(ks), nil
+	}
+
+	return konnectivity.NewServer(&konnectivity.ServerOptions{
+		Client:    b.K8sSeedClient.Client(),
+		Namespace: b.Shoot.SeedNamespace,
+		Image:     image.String(),
+		Replicas:  b.Shoot.GetReplicas(2),
```
danielfoehrKn, line 48: I assume this is done to keep the number of servers stable to avoid connection drops + no hvpa?
I think when we do performance tests, we should keep an eye on this.

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,552 @@
...
+
+const (
+	// ServerName is the name of kubernetes resources associated with konnectivity-server.
+	ServerName = "konnectivity-server"
+	// ServerKubeconfigSecretName is the name for the konnectivity-server's kubeconfig secret.
+	ServerKubeconfigSecretName = "konnectivity-server-kubeconfig"
+	// ServerTLSSecretName is the name of the konnectivity-server server certificate secret.
+	ServerTLSSecretName = ServerName
+	// ServerCASecretName is the name of the konnectivity-server server certificate authority secret.
+	ServerCASecretName = ServerName + "-ca"
```
danielfoehrKn:
```suggestion
	SecretNameServerCA = ServerName + "-ca"
```

pkg/operation/botanist/wanted_secrets.go
```diff
@@ -129,7 +142,7 @@ func (b *Botanist) generateWantedSecretConfigs(basicAuthAPIServer *secrets.Basic
 
 		konnectivityServerDNSNames = append([]string{
 			common.GetAPIServerDomain(b.Shoot.InternalClusterDomain),
-		}, kubernetes.DNSNamesForService(common.KonnectivityServerCertName, b.Shoot.SeedNamespace)...)
+		}, kubernetes.DNSNamesForService(konnectivity.ServerTLSSecretName, b.Shoot.SeedNamespace)...)
```
danielfoehrKn: The naming of the constant is a bit off.  Should be the name of the konnectivity service. I know the value is the same, but it makes it more clear for the reader.

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,552 @@
...
+						},
+						Labels: utils.MergeStringMaps(getLabels(), map[string]string{
+							v1beta1constants.LabelNetworkPolicyToDNS:              v1beta1constants.LabelNetworkPolicyAllowed,
+							v1beta1constants.LabelNetworkPolicyFromShootAPIServer: v1beta1constants.LabelNetworkPolicyAllowed,
+							v1beta1constants.LabelNetworkPolicyToShootAPIServer:   v1beta1constants.LabelNetworkPolicyAllowed,
+							v1beta1constants.LabelNetworkPolicyToSeedAPIServer:    v1beta1constants.LabelNetworkPolicyAllowed,
+						}),
+					},
+					Spec: corev1.PodSpec{
+						Affinity: &corev1.Affinity{
```
danielfoehrKn, line 216: What about adding an Affinity for kube-apiserver pods? 
Communication is now already not over UDS, but I think it would be good if they are located on the same host.

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,552 @@
...
+									corev1.ResourceCPU:    resource.MustParse("300m"),
+									corev1.ResourceMemory: resource.MustParse("400Mi"),
+								},
+							},
+							Ports: []corev1.ContainerPort{{
+								Name:          "server",
+								ContainerPort: ServerHTTPSPort,
+								HostPort:      ServerHTTPSPort,
+							}, {
+								Name:          "agent",
```
danielfoehrKn, line 286: this uses the adminPort for the "agent" labeled port? 
Missing one port? We need admin, agent, health and server right?

danielfoehrKn (changes requested):
Great idea.

Do I get this right, that this will mostly help with vertical scaling of the kube-apiserver pod? Then the konnectivity-server can keep running.
 If the apiserver scales the num of replicas, then the connectivity server process will be anyways restarted due to the replica reloader (this behaviour did not change). 
 
 In addition, why do we only want to have the apiserver-proxy as a separate deployment when SNI is enabled? Why does it not provide benefit without SNI? We could create pod affinity rules to co-locate kube-apiserver pods with the konnectivity-proxy to make up for not using the UDS. 
If feasible, I would vote for always deploying the connectivity server as a separate deployment.  Less code-duplication and better maintenance. The downside is that it will not use the UDS any more and I do not know what kind of effect that has on network performance. Nonetheless, I think the goal is to have SNI and Konnectivity enabled at the same time, so I would not make this code split unless its necessary.

I have not tested the PR yet, but will do when the issues are resolved.

mvladev:
> If the apiserver scales the num of replicas, then the connectivity server process will be anyways restarted due to the replica reloader (this behaviour did not change).

The konnectivity server would be only restarted if its deployment is scaled. It won't be restarted if the kube-apiserver is scaled.

> In addition, why do we only want to have the apiserver-proxy as a separate deployment when SNI is enabled? Why does it not provide benefit without SNI? We could create pod affinity rules to co-locate kube-apiserver pods with the konnectivity-proxy to make up for not using the UDS.

The problem is how to get traffic from the agents to the konnectivity server. For non-SNI enabled clusters it means that the Load balancer for the kube-apiserver is going to always send traffic from the agents to the API server pods. For konnectivity server to be a separate deployment in such situations, a tcp proxy would be needed to be added to the apiserver pod that would send traffic to the konnectivity server. E.g.

```
Agent -> kube apiserver LB -> Apiserver Pod (tcp proxy sidecar) -> konnectivity server
``` 

For scale-up of the apiserver this would have a benefit as no existing agent connections are going to be stopped. For scale-down this would kill some agent connections. 

I would open a separate PR with this feature.

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
mvladev: It's already build in the docker image https://github.com/gardener/replica-reloader/blob/e45a858c8d46db295d2ce1b86d13867bc5a6098f/Dockerfile#L35

danielfoehrKn:
> The konnectivity server would be only restarted if its deployment is scaled. It won't be restarted if the kube-apiserver is scaled.

True, I missed that the replica-reloader target the connectivity server deployment. I could not find hpa or hvpa for the Konnectivity server, how is it possible that it is being scaled?
Somehow I have a hard time wrapping my head around why the apiserver-proxy needs to be restarted with the right proxy-server amount in case its own deployment is scaled. I could not find detailed docs on the apiserver-proxy repo either to fully understand what  the flag does.
Can you point me to documentation or explain it again here? Thanks! 

> For non-SNI enabled clusters it means that the Load balancer for the kube-apiserver is going to always send traffic from the agents to the API server pods. For konnectivity server to be a separate deployment in such situations, a tcp proxy would be needed 

Makes sense. Would you leverage an existing managed istio deployment in the Seed to use an injected envoy for this? 

> For scale-down this would kill some agent connections.

Isn't that already the case today when the konnectivity server is a sidecar?

mvladev:
> Somehow I have a hard time wrapping my head around why the apiserver-proxy needs to be restarted with the right proxy-server amount in case its own deployment is scaled. I could not find detailed docs on the apiserver-proxy repo either to fully understand what the flag does.
Can you point me to documentation or explain it again here? Thanks!

There is no documentation - to put it simply - the agents (running on the shoot cluster) try connect to the amount of servers (running in the seed cluster). This allows every server to connect to workload running in the shoot cluster. e.g. 

if you have 2 replicas of konnectivity-server and only 1 agent (1 Node) then that agent would try connect to each replica of the server. 

The server count flag in the konnectivity-server is used to advertise the amount of available servers and their own ID when agents connect to it. So the agents can determine if they still need to connect to other servers.


> > For non-SNI enabled clusters it means that the Load balancer for the kube-apiserver is going to always send traffic from the agents to the API server pods. For konnectivity server to be a separate deployment in such situations, a tcp proxy would be needed
>
> Makes sense. Would you leverage an existing managed istio deployment in the Seed to use an injected envoy for this?

In this case we have non-SNI enabled cluster = no istio. I'll add the sidecar proxy manually.

> > For scale-down this would kill some agent connections.
>
> Isn't that already the case today when the konnectivity server is a sidecar?

Yes it is. Unfortunately, we cannot do anything for the scale down events. However, with a TCP proxy, scale up events would allow instant communication to Shoot workload.

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
mvladev: Even if they are on the same host, there is still a 50% chance that another replica would be used.

pkg/operation/botanist/konnectivity_server.go
mvladev: Yes, to keep connection drops to the minimum.

danielfoehrKn:
I'll try to test it today
EDIT: postpone to tomorrow

charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
```diff
@@ -224,10 +224,15 @@ spec:
         {{- if .Values.konnectivityTunnel.enabled }}
         - name: egress-selection-config
           mountPath: /etc/kubernetes/konnectivity
+        {{- if .Values.sni.enabled }}
```
danielfoehrKn, line 227: missing secret checksum on the API Server deployment

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,560 @@
...
+		serverAgentPortUnsigned       = uint32(ServerAgentPort)
+		serverMountPath               = "/certs/konnectivity-server"
+		clientCAMountPath             = "/certs/client-ca"
+		kubeconfigMountPath           = "/auth"
+	)
+
+	var (
+		minAvailable = intstr.FromInt(1)
+		tcpProto     = corev1.ProtocolTCP
+		serviceHost  = fmt.Sprintf("%s.%s.svc.%s", name, k.Namespace, gardenercorev1.DefaultDomain)
```
danielfoehrKn:
```suggestion
		konnectivityServerServiceHost  = fmt.Sprintf("%s.%s.svc.%s", name, k.Namespace, gardenercorev1.DefaultDomain)
```

Just makes it a bit easier to understand where it is routing to it in the VirtualService definition below.

pkg/operation/botanist/controlplane.go
```diff
@@ -603,7 +604,9 @@ func (b *Botanist) DeployKubeAPIServer(ctx context.Context) error {
 			"securePort":                443,
 			"podAnnotations":            podAnnotations,
 			"konnectivityTunnel": map[string]interface{}{
-				"enabled": b.Shoot.KonnectivityTunnelEnabled,
+				"enabled":    b.Shoot.KonnectivityTunnelEnabled,
+				"name":       konnectivity.ServerName,
+				"serverPort": konnectivity.ServerHTTPSPort,
```
danielfoehrKn, line 609: When only Konnectivity is enabled without SNI it fails with 

> Error: failed to validate server options with server port should be set to 0 not 9443 for UDS

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -290,6 +290,16 @@ func (c *Controller) runReconcileShootFlow(ctx context.Context, o *operation.Ope
 			Fn:           flow.TaskFn(botanist.DeployKubeScheduler).RetryUntilTimeout(defaultInterval, defaultTimeout),
 			Dependencies: flow.NewTaskIDs(deploySecrets, waitUntilKubeAPIServerIsReady),
 		})
+		deployKonnectivityServer = g.Add(flow.Task{
+			Name:         "Deploying konnectivity-server",
+			Fn:           flow.TaskFn(botanist.DeployKonnectivityServer).RetryUntilTimeout(defaultInterval, defaultTimeout),
```
danielfoehrKn, line 295:
```suggestion
			Fn:           flow.TaskFn(botanist.DeployKonnectivityServer).DoIf(o.Shoot.KonnectivityTunnelEnabled && useSNI).RetryUntilTimeout(defaultInterval, defaultTimeout),
```

Otherwise fails with

> "Deploying konnectivity-server\" failed: retry failed with context deadline exceeded, last error: missing client-ca secret information

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -290,6 +290,16 @@ func (c *Controller) runReconcileShootFlow(ctx context.Context, o *operation.Ope
...
+			Dependencies: flow.NewTaskIDs(deploySecrets),
+		})
+		waitUntilKonnectityServerReady = g.Add(flow.Task{
+			Name:         "Waiting until konnectivity-server is ready",
+			Fn:           flow.TaskFn(botanist.Shoot.Components.ControlPlane.KonnectivityServer.Wait),
```
danielfoehrKn, line 300:
```suggestion
			Fn:           flow.TaskFn(botanist.Shoot.Components.ControlPlane.KonnectivityServer.Wait).DoIf(o.Shoot.KonnectivityTunnelEnabled && useSNI),
```

pkg/operation/botanist/botanist.go
```diff
@@ -123,6 +123,10 @@ func New(ctx context.Context, o *operation.Operation) (*Botanist, error) {
 	if err != nil {
 		return nil, err
 	}
+	o.Shoot.Components.ControlPlane.KonnectivityServer, err = b.DefaultKonnectivityServer()
```
danielfoehrKn, line 130: Idk why we are instantiating the SNI and Konnectivity Server components without checking if the feature gates are enabled. 
Here is should only be created if (Konnectivity && !SNI)

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,560 @@
...
+								fmt.Sprintf("--authentication-audience=%s", ServerAudience),
+								"--keepalive-time=1h",
+								"--log-file-max-size=0",
+								"--delete-existing-uds-file=true",
+								"--mode=http-connect",
+								fmt.Sprintf("--server-port=%d", ServerHTTPSPort),
+								fmt.Sprintf("--agent-port=%d", ServerAgentPort),
+								fmt.Sprintf("--admin-port=%d", adminPort),
+								fmt.Sprintf("--health-port=%d", healthPort),
+								"--v=2",
```
danielfoehrKn, line 258:
```suggestion
								"--v=2",
								"--server-count",
```

Missing the last argument for replica-reloader. Will default to 1 instead of the correct number of replicas of the connectivity server deployment.

charts/seed-controlplane/charts/kube-apiserver-sni/templates/kube-apiserver-gateway.yaml
```diff
@@ -18,13 +18,3 @@ spec:
...
       mode: PASSTHROUGH
-{{- if .Values.enableKonnectivityTunnel }}
-  - hosts:
-{{ toYaml .Values.hosts | indent 4 }}
-    port:
-      number: 8132
-      name: tls-tunnel
-      protocol: TLS
-    tls:
-      mode: PASSTHROUGH
```
danielfoehrKn, line 29: CRD uses string as the TLS mode, however the API bindings use an int32.

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
```diff
@@ -0,0 +1,560 @@
...
+				Selector: k.IstioIngressLabels,
+				Servers: []*istionetworkingv1beta1.Server{{
+					Hosts: k.Hosts,
+					Port: &istionetworkingv1beta1.Port{
+						Number:   serverAgentPortUnsigned,
+						Protocol: "TLS",
+						Name:     "tls-tunnel",
+					},
+					Tls: &istionetworkingv1beta1.ServerTLSSettings{
+						Mode: istionetworkingv1beta1.ServerTLSSettings_PASSTHROUGH,
```
danielfoehrKn, line 359: The istio [Gateway go types](https://github.com/istio/api/blob/master/networking/v1beta1/gateway.pb.go) have the TLS mode as an int32 - you use the constant here.
However, the CRD for the Gateway we use [is a string](https://github.com/gardener/gardener/blob/e37bb3d5c8b4209702d90d71e59d172c55129d57/charts/istio/istio-crds/templates/crd-all.gen.yaml#L1546). 
I initially thought that this will be converted internally, however checking the Gateway resource in the cluster shows that the TLS mode PASSTHROUGH is missing on the Konnectivity server gateway.

>   
      name: tls-tunnel
      number: 8132
      protocol: TLS
    tls: {}
    
Vs. for the kube-apiserver
    
>
      name: tls
      number: 443
      protocol: TLS
    tls:
      mode: PASSTHROUGH


I cannot definitely say that this causes  TLS to be terminated as the istio-ingressgateway or if Passthrough is the default setting when nothing is specified.
However,  a tunnel connection cannot be established so this is a possible reason.

danielfoehrKn (changes requested):
**Only Konnectivity enabled** 

Needs some minor fixes to work properly. I added comments.

**Konnectivity  + SNI** 

Unfortunately, it is unable to establish a tunnel connection.
I added a comment pointing out a possible reason due to misconfigured TLS mode on the Gateway (unsure). However, I am not sure if that is the reason.

Logs on the connectivity server show HTTP connect requests from various components in the Shoot via the connectivity agent

> 2021/01/26 16:03:01 http: response.Write on hijacked connection from fmt.Fprintln (print.go:265)
I0126 16:03:01.412597      14 tunnel.go:38] "Received request for host" method="CONNECT" host="100.96.0.23:8443" userAgent=""
I0126 16:03:01.412621      14 tunnel.go:40] "TLS" commonName="kube-apiserver"
2021/01/26 16:03:01 http: response.WriteHeader on hijacked connection from sigs.k8s.io/apiserver-network-proxy/pkg/server.(*Tunnel).ServeHTTP (tunnel.go:74)
2021/01/26 16:03:01 http: response.Write on hijacked connection from fmt.Fprintln (print.go:265)

Logs of the API Server show connection resets and failure to get logs/forward to Shoot 

> E0126 16:08:55.266037       1 status.go:71] apiserver received an error that is not an metav1.Status: &errors.errorString{s:"error dialing backend: read tcp 10.223.130.23:33238->10.223.30.51:9443: read: connection reset by peer"}
E0126 16:09:04.680631       1 status.go:71] apiserver received an error that is not an metav1.Status: &fmt.wrapError{msg:"error trying to reach service: context canceled", err:(*errors.errorString)(0xc000104270)}
I0126 16:09:04.884657       1 client.go:360] parsed scheme: "passthrough"
I0126 16:09:04.884723       1 passthrough.go:48] ccResolverWrapper: sending update to cc: {[{https://etcd-main-client:2379  <nil> 0 <nil>}] <nil> <nil>}
I0126 16:09:04.884737       1 clientconn.go:948] ClientConn switching balancer to "pick_first"
I0126 16:09:04.884855       1 balancer_conn_wrappers.go:78] pickfirstBalancer: HandleSubConnStateChange: 0xc005c99680, {CONNECTING <nil>}
I0126 16:09:04.896339       1 balancer_conn_wrappers.go:78] pickfirstBalancer: HandleSubConnStateChange: 0xc005c99680, {READY <nil>}
I0126 16:09:04.904476       1 controlbuf.go:508] transport: loopyWriter.run returning. connection error: desc = "transport is closing"

pkg/operation/botanist/controlplane.go
danielfoehrKn:
```suggestion
```

Instead

```
if b.Shoot.KonnectivityTunnelEnabled {
		serverPort := int32(0)
		if b.APIServerSNIEnabled() {
			serverPort = konnectivity.ServerHTTPSPort
		}

		defaultValues["konnectivityTunnel"] = map[string]interface{}{
			"enabled":    b.Shoot.KonnectivityTunnelEnabled,
			"serverPort": serverPort,
		}
	}

```

pkg/operation/botanist/controlplane/konnectivity/konnectivity_server.go
mvladev: This doesn't matter - the default value for PASSTHROUGH is 0. The TLS object is empty (but not nil) and therefore this is a PASSTHROUGH cluster -  https://github.com/istio/istio/blob/65bd2ac632cbd6ec8e3fd85bb9874f685aea0cfd/pkg/config/gateway/gateway.go#L45-L57

charts/seed-controlplane/charts/kube-apiserver-sni/templates/kube-apiserver-gateway.yaml
mvladev: See bellow

mvladev:
I've reworked the PR.  Have another look @danielfoehrKn

pkg/operation/botanist/botanist.go
danielfoehrKn: Not resolved or adressed?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
danielfoehrKn: not resolved?

pkg/operation/botanist/controlplane.go
danielfoehrKn: Not resolved?

danielfoehrKn:
@mvladev maybe I a missing something but I do not see that you addressed the comments?

pkg/operation/botanist/controlplane.go
mvladev: It's always set. In the deployment is hard-coded to 0.

pkg/operation/botanist/botanist.go
mvladev: I've changed the implementation of `DefaultKonnectivityServer` now it does that check in it https://github.com/gardener/gardener/blob/178ee2799dd58f2e37d04ca387e22c6f30b105ff/pkg/operation/botanist/konnectivity_server.go#L25-L37

mvladev:
See my comments from above.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
mvladev: Not needed as now `DefaultKonnectivityServer` handles this.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
mvladev: Not needed as now `DefaultKonnectivityServer` handles this - e.g. if one of SNI or Konnectivity is not enabled, `Destroy` is called instead.

danielfoehrKn:
Ok, thanks, will test tomorrow

charts/seed-controlplane/charts/kube-apiserver/templates/deployment.yaml
```diff
@@ -346,7 +351,8 @@ spec:
         - --log-file-max-size=0
         - --delete-existing-uds-file=true
         - --mode=http-connect
-        - --server-port={{ .Values.konnectivityTunnel.serverPort }}
+        # the seerver port should always be 0 when using uds
```
danielfoehrKn:
```suggestion
        # the server port should always be 0 when using UDS
```

danielfoehrKn (changes requested):
**SNI + Konnectivity**: 
The reconciliation works and I am able to get logs via the connectivity server.

Independent of this PR, these logs make me suspicious and I wonder where they come from.
These are there with and without SNI.

Server
> I0202 13:23:28.916115      14 tunnel.go:118] "EOF from host" host="10.250.26.81:10250"
E0202 13:23:28.947704      14 server.go:559] "send to client stream failure" err="tls: use of closed connection"
E0202 13:23:28.947738      14 server.go:574] "CLOSE_RSP send to client stream error" err="tls: use of closed connection"
I0202 13:23:28.947776      14 server.go:166] "Remove frontend for agent" frontend=&server.ProxyClientConnection{Mode:"http-connect", Grpc:client.ProxyService_ProxyServer(nil), HTTP:(*tls.Conn)(0xc000060e00), connected:(chan struct {})(0xc000088360), connectID:17, agentID:"7e329af8-c65f-4a02-95e6-9fc74e346de0", start:time.Time{wall:0xbffe72742d79dcd6, ext:450946072508, loc:(*time.Location)(0x22800a0)}, backend:(*server.backend)(0xc0001c8380)} agentID="7e329af8-c65f-4a02-95e6-9fc74e346de0" connectionID=17

Agent

> I0202 13:24:35.329830       1 client.go:387] connection EOF
I0202 13:24:35.526739       1 client.go:387] connection EOF
I0202 13:25:08.779864       1 client.go:387] connection EOF
I0202 13:25:35.461489       1 client.go:387] connection EOF
I0202 13:25:42.987244       1 client.go:387] connection EOF

After rebasing, This PR is good to go from my PoV.

danielfoehrKn (approved):

PR #3406, opened 2021-01-14
kube-scheduler's v1 config api is copied for all versions

mvladev:

/area quality
/kind enhancement
/priority normal

Scheduler's API is changed between K8S versions. It's better to copy all references.

Fixes n/a

See #3393

**Release note**:

```other operator
NONE
```


rfranzke:
/rebase
Can you check the failing CI, please?

third_party/kube-scheduler/v18/v1alpha2/types.go
```diff
@@ -96,7 +96,7 @@ type KubeSchedulerConfiguration struct {
 	// Extenders are the list of scheduler extenders, each holding the values of how to communicate
 	// with the extender. These extenders are shared by all scheduler profiles.
 	// +listType=set
-	Extenders []v1.LegacyExtender `json:"extenders"`
+	Extenders []v1.Extender `json:"extenders"`
```
timebertt, line 99: Thanks 👍

timebertt (dismissed):
Thanks for taking care!
/lgtm

hack/test-cover.sh
```diff
@@ -15,8 +15,16 @@
 # limitations under the License.
 set -e
 
+TEST_BIN_DIR="$(dirname "${0}")/../dev/testbin"
```
timebertt, line 18: What's the motivation behind the test-related changes apart from bumping the k8s version?

mvladev: It was not up to date with https://github.com/kubernetes-sigs/controller-runtime/blob/master/hack/setup-envtest.sh 

Also, it's better to have it in a local directory than updating system binaries (as it was doing before)

timuthy: If you want to keep it in sync, what do you think about downloading it in the scope of `make revendor`?

rfranzke: @mvladev

mvladev:
I've updated the PR and now it automatically downloads the `setup-envtest.sh` based on the go module version of `controller-runtime`

timebertt (approved):
Thanks.
/lgtm now

timuthy (approved):

PR #3409, opened 2021-01-15
Allow kube-proxy PSP to use NET_ADMIN

rfranzke:

/area security
/kind regression
/priority critical

#3395 introduced a regression that prevents `kube-proxy` pods from being created in case `spec.kubernetes.allowPrivilegedContainers=false`:

```
 Warning  FailedCreate  18s (x6 over 59s)  daemonset-controller  Error creating: pods "kube-proxy-" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.volumes[2]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[3]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[4]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[6]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[7]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.initContainers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.containers[0].hostPort: Invalid value: 10249: Host port 10249 is not allowed to be used. Allowed ports: [] spec.containers[1].securityContext.capabilities.add: Invalid value: "NET_ADMIN": capability may not be added spec.containers[1].securityContext.capabilities.add: Invalid value: "NET_ADMIN": capability may not be added]
```

This PR adds `NET_ADMIN` to the allowed capabilities in kube-proxy's `PodSecurityPolicy`.

**Release note**:

```bugfix user
A bug has been fixed that prevented shoot clusters from coming up in case `.spec.kubernetes.allowPrivilegedContainers=false`.
```

### DIFF
--- a/charts/shoot-core/components/charts/kube-proxy/templates/psp/kube-proxy-psp.yaml
+++ b/charts/shoot-core/components/charts/kube-proxy/templates/psp/kube-proxy-psp.yaml
@@ -17,6 +17,8 @@ spec:
   - pathPrefix: /var/run/dbus/system_bus_socket
   - pathPrefix: /lib/modules
   - pathPrefix: /var/lib/kube-proxy
+  allowedCapabilities:
+  - NET_ADMIN
   runAsUser:
     rule: 'RunAsAny'
   seLinux:


ialidzhikov (approved):
/lgtm

PR #3411, opened 2021-01-15
Add disruptive testsdefinition and flag rotate test as disruptive

schrodit:

/area testing
/kind test
/priority normal

A new testsuite is introduced that contains disruptive tests.
Disruptive tests are special kind of test that may break something if the test is not successfully finished.
The TestDefinitions are labeled `disruptive` so that the testmachinery can act accordingly.

With that change, also the kubeconfig rotate test is updated to be of kind disruptive.
With that the previous issues with the test to update the kubeconfig in the testmachinery is fixed.

When the test works as expected we can also promote it again to the `default` test group.

Fixes #

**Release note**:

```other operator
TestDefinitions have been added that contains disruptive tests
```

### DIFF
--- a/.test-defs/TestSuiteShootBetaDisruptive.yaml
+++ b/.test-defs/TestSuiteShootBetaDisruptive.yaml
@@ -1,14 +1,14 @@
 kind: TestDefinition
 metadata:
-  name: shoot-release-serial-test-suite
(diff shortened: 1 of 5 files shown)


dguendisch (approved):
/lgtm

rfranzke (approved):

PR #3412, opened 2021-01-15
[GEP-13] Add ManagedSeed APIs

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Adds the new `ManagedSeed` resource as described in [GEP-13](docs/proposals/13-automated-seed-management.md#managedseeds).

* `ManagedSeed` resource APIs, including defaults, since they are used by the generated code.
* Small changes to the `GardenletConfiguration` types (for consistency between internal and external API versions).
* All generated code belonging to the above APIs, including generated clients.

The new APIs are not yet registered or used by the API server. This is on purpose, to make this PR easier to review and make sure it can be merged as it has no impact yet.

Fixes #

This is the first of several PRs that are split from #3177 since it became too big and hard to review. Before commenting, please check if the topic has been already discussed in the previous PR.

**Release note**:

```improvement operator
NONE
```


stoyanr:

This is the first of several PRs, as agreed. All comments on the old PR are addressed. I hope we could review and merge this one quickly.

stoyanr:
Currently unit tests are failing with:
```
  Expected success, but got an error:
      <*helper.ErrorWithCodes | 0xc0007f2810>: {
          message: "Error while waiting for BackupBucket /gextlib-backupbucket-test--b4if1 to become ready: extension operation was not updated yet",
          codes: nil,
      }
      Error while waiting for BackupBucket /gextlib-backupbucket-test--b4if1 to become ready: extension operation was not updated yet
```
I assume this is a flaky test, I can't reproduce the error while running `make test` locally and it doesn't seem to be related to my changes.

pkg/apis/seedmanagement/doc.go
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: Nit: Those copyrights for new files could be changed to 2021

pkg/apis/seedmanagement/helper/helper.go
```diff
@@ -0,0 +1,122 @@
...
+	return seedmanagement.BootstrapNone
+}
+
+// MergeWithParent returns the value of the MergeWithParent field of the given gardenlet, or false if the field is nil.
+func MergeWithParent(gardenlet *seedmanagement.Gardenlet) bool {
+	if gardenlet.MergeWithParent != nil {
+		return *gardenlet.MergeWithParent
+	}
+	return false
+}
```
rfranzke: Can this be replaced by using https://github.com/gardener/gardener/blob/master/pkg/utils/miscellaneous.go#L123-L126?

pkg/apis/seedmanagement/helper/helper_suite_test.go
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/apis/seedmanagement/helper/helper_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	Describe("#ConvertSeedExternal", func() {
+		It("should convert the internal Seed version to an external one", func() {
+			result, err := ConvertSeedExternal(seed)
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).To(Equal(seedx))
+		})
+	})
+})
```
rfranzke, line 150: Should we also have a unit test for `Bootstrap` (and if still applicable `MergeWithParent`)?

pkg/apis/seedmanagement/types_managedseed.go
```diff
@@ -0,0 +1,161 @@
...
+
+// ManagedSeedSpec is the specification of a ManagedSeed.
+type ManagedSeedSpec struct {
+	// Shoot references a Shoot that should be registered as Seed.
+	Shoot Shoot
+	// SeedTemplate is a template for a Seed object, that should be used to register a given cluster as a Seed.
+	// Either SeedTemplate or Gardenlet must be specified. When Seed is specified, the ManagedSeed controller will not deploy a gardenlet into the cluster
+	// and an existing gardenlet reconciling the new Seed is required.
+	SeedTemplate *SeedTemplate
+	// Gardenlet specifies, that the ManagedSeed controller should deploy a gardenlet into the cluster
```
rfranzke:
```suggestion
	// Gardenlet specifies that the ManagedSeed controller should deploy a gardenlet into the cluster
```

pkg/apis/seedmanagement/types_managedseed.go
```diff
@@ -0,0 +1,161 @@
...
+	Repository *string
+	// Tag is the image tag.
+	Tag *string
+	// PullPolicy is the image pull policy. One of Always, Never, IfNotPresent.
+	// Defaults to Always if latest tag is specified, or IfNotPresent otherwise.
+	PullPolicy *corev1.PullPolicy
+}
+
+// Bootstrap describes a mechanism for bootstrapping gardenlet connection to the Garden cluster.
+type Bootstrap string
```
rfranzke, line 132:
```suggestion
type BootstrapMode string
```

pkg/apis/seedmanagement/v1alpha1/types_managedseed.go
```diff
@@ -0,0 +1,190 @@
...
+	// +optional
+	Tag *string `json:"tag,omitempty" protobuf:"bytes,2,opt,name=tag"`
+	// PullPolicy is the image pull policy. One of Always, Never, IfNotPresent.
+	// Defaults to Always if latest tag is specified, or IfNotPresent otherwise.
+	// +optional
+	PullPolicy *corev1.PullPolicy `json:"pullPolicy,omitempty" protobuf:"bytes,3,opt,name=pullPolicy"`
+}
+
+// Bootstrap describes a mechanism for bootstrapping gardenlet connection to the Garden cluster.
+type Bootstrap string
```
rfranzke, line 158:
```suggestion
type BootstrapMode string
```

pkg/apis/seedmanagement/v1alpha1/v1alpha1_suite_test.go
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

rfranzke (commented):
/assign

pkg/apis/seedmanagement/types_managedseed.go
stoyanr: Is this really necessary? If I change the type name I should probably change the field name as well, and then adapt to this change in 5 other downstream PRs. The difference is very small and I personally prefer `Bootstrap` as it's shorter .

pkg/apis/seedmanagement/v1alpha1/types_managedseed.go
stoyanr: See above.

pkg/apis/seedmanagement/doc.go
stoyanr: Done

pkg/apis/seedmanagement/helper/helper.go
stoyanr: Yes, changed the `Bootstrap` function to have a similar interface for consistency.

pkg/apis/seedmanagement/helper/helper_suite_test.go
stoyanr: Done

pkg/apis/seedmanagement/helper/helper_test.go
stoyanr: Added

pkg/apis/seedmanagement/types_managedseed.go
stoyanr: Done

pkg/apis/seedmanagement/v1alpha1/v1alpha1_suite_test.go
stoyanr: Done

stoyanr:
@rfranzke Did all requested changes except the renaming from `Bootstrap` to `BootstrapMode`, see comments.

pkg/apis/seedmanagement/types_managedseed.go
rfranzke: I don't insist on it, but I prefer to have the "Mode" suffix because it's simply more clear what this is about. Is it really that much effort to change the other PRs accordingly? You have to rebase them anyways once this PR was merged.

stoyanr: From my perspective `BootstrapMode` is neither more nor less clear than `Bootstrap`, they are the same. There is a comment explaining the meaning and the usage, without it both would not be clear. It would take some effort to properly rename it everywhere and I question whether this is really needed. If you don't insist, could we keep it as it is?

rfranzke:
https://concourse.ci.gardener.cloud/builds/331527
```
>> make generate
 M pkg/apis/seedmanagement/v1alpha1/generated.proto
 M pkg/openapi/openapi_generated.go
```

stoyanr:
@rfranzke `>> make generate` is fixed now

rfranzke (dismissed):
/lgtm

@danielfoehrKn @timebertt do you want to take another look as well?

pkg/apis/seedmanagement/v1alpha1/defaults.go
```diff
@@ -0,0 +1,149 @@
...
+	if obj.Deployment == nil {
+		obj.Deployment = &GardenletDeployment{}
+	}
+
+	// Decode / initialize gardenlet config
+	var gardenletConfig *configv1alpha1.GardenletConfiguration
+	var err error
+	if obj.Config != nil {
+		// Decode gardenlet config to an external version
+		// Without defaults, since we don't want to set gardenlet config defaults in the resource at this point
```
danielfoehrKn, line 92: Because it is done in Line 108 instead of also covering the case when the config == nil?
I remember this has been a point of discussion.  I think the comment should either be removed or should explain why defaulting is not done at this point.

danielfoehrKn (commented):
Only a nit - otherwise looks great to me.

pkg/apis/seedmanagement/v1alpha1/defaults.go
danielfoehrKn: But this is only my opinion, if you wish we can go ahead and merge it from my PoV.

danielfoehrKn (dismissed):

timebertt:
> @timebertt do you want to take another look as well?

Ok, let me do a quick scan

pkg/apis/seedmanagement/types_managedseed.go
```diff
@@ -0,0 +1,161 @@
...
+	// Conditions represents the latest available observations of a ManagedSeed's current state.
+	Conditions []gardencore.Condition
+	// ObservedGeneration is the most recent generation observed for this ManagedSeed. It corresponds to the
+	// ManagedSeed's generation, which is updated on mutation by the API Server.
+	ObservedGeneration int64
+}
+
+const (
+	// ManagedSeedShootExists is a condition type for indicating whether the ManagedSeed's shoot exists.
+	ManagedSeedShootExists gardencore.ConditionType = "ShootExists"
```
timebertt, line 155: As said in the other PR, I think we shouldn't need this condition because a situation with a `False` value can (mostly) be avoided by admission.
If errors still arise during reconciliation, we should add them in the other conditions or add an event.
I'm fine with keeping it for now and removing it once we have discussed the controller implementation.
Also, it's still an alpha API, so it will be no harm to remove it later on.

pkg/apis/seedmanagement/v1alpha1/defaults.go
stoyanr: I am not quite sure what exactly is the issue that you see, and what do you propose? 

The comment explains that we don't want all the defaults of the `GardenletConfiguration` type itself at this point. As was discussed, this is because it would break merging with parent, also as @timebertt pointed out it's not idiomatic, other Kubernetes implementations (e.g. `Deployment`) don't apply defaults on embedded types as well. But I think a full comprehensive explanation would result in a rather biggish comment that is not appropriate here.

As for Line 108, what happens there is that particular `GardenletConfiguration` defaults, that are specific to the managed seed, are applied. This is quite independent of the way `GardenletConfiguration` is decoded.

pkg/apis/seedmanagement/v1alpha1/defaults.go
```diff
@@ -0,0 +1,149 @@
...
+			},
+		}
+	}
+
+	// Set gardenlet config defaults
+	setDefaultsGardenletConfiguration(gardenletConfig, name, namespace)
+
+	// Set gardenlet config back to obj.Config
+	gardenletConfigJSON, _ := json.Marshal(gardenletConfig)
+	obj.Config = &runtime.RawExtension{Raw: gardenletConfigJSON}
```
timebertt: It shouldn't be necessary to convert to JSON by hand. the `RawExtension` type takes care of correct marshalling, see https://github.com/kubernetes/apimachinery/blob/0e9f9cff521d6ebb5aaee87ae496b501bc27e974/pkg/runtime/extension.go#L37
```suggestion
	obj.Config = &runtime.RawExtension{Object: gardenletConfig}
```

pkg/apis/seedmanagement/types_managedseed.go
```diff
@@ -0,0 +1,161 @@
...
+	Spec gardencore.SeedSpec
+}
+
+// Gardenlet specifies gardenlet deployment parameters and the GardenletConfiguration used to configure gardenlet.
+type Gardenlet struct {
+	// Deployment specifies certain gardenlet deployment parameters, such as the number of replicas,
+	// the image, etc.
+	Deployment *GardenletDeployment
+	// Config is the GardenletConfiguration used to configure gardenlet.
+	Config *runtime.RawExtension
```
timebertt: `Config` should be of type `runtime.Object` in the internal API version, see https://github.com/kubernetes/apimachinery/blob/0e9f9cff521d6ebb5aaee87ae496b501bc27e974/pkg/runtime/types.go#L49-L94
```suggestion
	Config runtime.Object
```

timebertt (commented):
Thanks @stoyanr, the API-only PR really helps with a focused review.
I've done a rough scan over it again and only found a few more nits.
Other than that it looks really great now. I think, especially the `Gardenlet` type is much cleaner now :)

pkg/apis/seedmanagement/helper/helper.go
```diff
@@ -0,0 +1,114 @@
...
+func init() {
+	scheme = runtime.NewScheme()
+	utilruntime.Must(config.AddToScheme(scheme))
+	utilruntime.Must(configv1alpha1.AddToScheme(scheme))
+	utilruntime.Must(gardencore.AddToScheme(scheme))
+	utilruntime.Must(gardencorev1beta1.AddToScheme(scheme))
+}
+
+// DecodeGardenletConfig decodes the given raw extension into an internal GardenletConfiguration version.
+func DecodeGardenletConfig(rawConfig *runtime.RawExtension, withDefaults bool) (*config.GardenletConfiguration, error) {
```
timebertt, line 43: I think, the way `RawExtension` is supposed to work is to set decoded objects on the `Object` field to avoid double decoding efforts.
Can you check if `rawConfig.Object` is already set to a value of type `GardenletConfiguration` and only do the decoding work if that's not the case?

pkg/apis/seedmanagement/types_managedseed.go
stoyanr: I still believe this is needed. If for whatever reason the shoot is not found when reconciling the managed seed (e.g. admission plugins were inactive or buggy), shouldn't this situation be reflected in the status, similarly to all other errors? If yes, then none of the other conditions would do since this case is really special, also it would fail earlier. If you are fine with keeping it, let's keep it and decide later during the review of the controller implementation.

pkg/apis/seedmanagement/v1alpha1/defaults.go
stoyanr: OK

pkg/apis/seedmanagement/types_managedseed.go
stoyanr: I tried this earlier but couldn't make it work. Will give it another try. Perhaps you are aware that `RawExtension`, and not `Object` is used in internal versions of all other Gardener APIs that use this type.

pkg/apis/seedmanagement/helper/helper.go
stoyanr: This decoding function is really special since it has the `withDefaults` parameter. This is important in cases where we don't want defaults to be applied, e.g. when applying defaults on our type. I believe any default decoding would not do this properly (it would always apply defaults) and therefore decoding explicitly is needed, independently of whether `rawConfig.Object` is already set, for `withDefaults == false`.

pkg/apis/seedmanagement/types_managedseed.go
rfranzke: Hm, I'm not aware of the details of the controller implementation yet, but the argument ("If for whatever reason the shoot is not found when reconciling the managed seed (e.g. admission plugins were inactive or buggy)") sounds artificial/unlikely. If that's the only point then let's drop it and rather populate such issues with other conditions or events as @timebertt suggested. Personally, I'm fine with keeping it for now and revisiting it in the respective controller PR.

stoyanr: OK, let's keep it then, will remove it later if needed.

pkg/apis/seedmanagement/helper/helper.go
timebertt: ok

pkg/apis/seedmanagement/types_managedseed.go
timebertt: Yes, but let's do it the right way this time then, as it is supposed to be done by apimachinery.

stoyanr: OK, figured it out finally, will do as suggested. To make sure that it works, I had to rebase all downstream PRs up to the last one in order to test the actual controller reconciliation ... several times, until I got it right. This was **not** fun. I hope you are indeed having an easier time with the multiple PRs, because it's definitely not the case on my side.

pkg/apis/seedmanagement/helper/helper.go
stoyanr: In fact, testing showed that checking for `withDefaults` is not needed, because of the way defaults are currently implemented. So I'll simply add the check as suggested and adapt the tests accordingly.

stoyanr:
@timebertt I believe I addressed all your feedback.

timebertt:
```
>> make generate
 M pkg/apis/seedmanagement/v1alpha1/zz_generated.conversion.go
make: *** [Makefile:188: check-generate] Error 1
```

stoyanr:
@timebertt `>> make generate` should be fixed now

timebertt (approved):
Thanks @stoyanr for addressing my suggestions!
/lgtm now

PR #2370, opened 2020-05-22
Upgrade k8s dependencies to v1.17.6, controller-runtime to v0.5.3

timebertt:
**How to categorize this PR?**

/kind enhancement
/priority normal

**What this PR does / why we need it**:
This PR revendors k8s dependencies with version v0.17.6 (k8s version v1.17.6) and controller-runtime with version v0.5.3.
This contains bug fixes for the controller-runtime cache, which will be required for what I am currently working on.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement developer
The golang Kubernetes dependencies have been upgraded to `v1.17.6`.
```
```action operator
The minimum required version for the Garden Cluster was increased to `1.16`.
```

PR #3413, opened 2021-01-15
Enhance webhook constraints checks

timebertt:

/area quality ops-productivity
/kind enhancement
/priority normal

This PR enhances the webhook constraints checks in the following ways:
- webhooks with `FailurePolicy=Ignore` but `TimeoutSeconds>15` are now also marked as problematic (long timeouts are discouraged by the [k8s documentation](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts) and can cause problems on wakeup as well
- webhooks that act on `configmaps` in the `kube-system` namespace are now also marked as problematic (older versions of k8s control plane components used leader election via `configmaps`)
- the webhook matchers assumed a label `role=kube-system` on the `kube-system` namespace for shoots, although there is no such label, [ref](https://github.com/gardener/gardener/blob/3a403423130b8a52028b94ef756d97afa4963089/pkg/operation/botanist/systemcomponents/namespaces/namespaces.go#L82), that assumption is corrected

Fixes #

**Release note**:

```other user
Webhooks acting on `configmaps` in the `kube-system` namespace and webhooks with a `TimeoutSeconds>15` for problematic resources are now also blocking `Maintenance` and `Hibernation` operations. Please consult [this doc](https://github.com/gardener/gardener/blob/master/docs/usage/shoot_status.md#constraints) for more details.
```

### DIFF
--- a/docs/usage/shoot_status.md
+++ b/docs/usage/shoot_status.md
@@ -49,16 +49,45 @@ Let's check the following example to get better understanding. Let's say that th
 ### Constraints
 
 Constraints represent conditions of a Shoot’s current state that constraint some operations on it.
-
 Currently there are two constraints:
 
-- `HibernationPossible`
+**`HibernationPossible`**:
 
-  This constraint indicates whether a Shoot is allowed to be hibernated. The rationale behind this constraint is that a Shoot can have `ValidatingWebhookConfiguration`s or `MutatingWebhookConfiguration`s with rules for CREATE Pods or Nodes and `failurePolicy=Fail`. Such webhooks are preventing wake up for a Shoot cluster as new Nodes cannot be created or new system component Pods cannot be created because the webhook is not running. To prevent such deadlock situation, the gardener-apiserver does not allow Shoot to be hibernated when the `HibernationPossible` has status `False`.
+This constraint indicates whether a Shoot is allowed to be hibernated.
+The rationale behind this constraint is that a Shoot can have `ValidatingWebhookConfiguration`s or `MutatingWebhookConfiguration`s acting on resources that are critical for waking up a cluster.
+For example, if a webhook has rules for `CREATE/UPDATE` Pods or Nodes and `failurePolicy=Fail`, the webhook will block joining `Nodes` and creating critical system component Pods and thus block the entire wakeup
+operation, because the server backing the webhook is not running.
 
-- `MaintenancePreconditionsSatisfied`
+Even if the `failurePolicy` is set to `Ignore`, high timeouts (`>15s`) can lead to blocking requests of control plane components.
+That's because most control-plane API calls are made with a client-side timeout of `30s`, so if a webhook has `timeoutSeconds=30`
+the overall request might still fail as there is overhead in communication with the API server and potential other webhooks.
+Generally, it's [best pratice](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts)
+to specify low timeouts in WebhookConfigs.
+Also, it's [best practice](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#avoiding-operating-on-the-kube-system-namespace)
+to exclude the `kube-system` namespace from webhooks to avoid blocking critical operations on system components of the cluster.
+Shoot owners can do so by adding a `namespaceSelector` similar to this one to their webhook configurations:
+```yaml
+namespaceSelector:
+  matchExpressions:
+  - key: gardener.cloud/purpose
+    operator: NotIn
+    values:
+    - kube-system
+```
 
-  This constraint indicates whether all preconditions for a safe maintenance operation are satisfied (see also [this document](shoot_maintenance.md) for more information about what happens during a shoot maintenance). As of today, the same checks as in the `HibernationPossible` constraint are being performed (user-deployed webhooks that might interfere with potential rolling updates of shoot worker nodes). There is no further action being performed on this constraint's status (maintenance is still being performed). It is meant to make the user aware of potential problems that might occur due to his configurations. 
+If the Shoot still has webhooks with either `failurePolicy={Fail,nil}` or `failurePolicy=Ignore && timeoutSeconds>15` that
+act on [critical resources](https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/matchers/matcher.go#L60)
+in the `kube-system` namespace, Gardener will set the `HibernationPossible` to `False` indicating, that the Shoot can
+probably not be woken up again after hibernation without manual intervention of the Gardener Operator.
+`gardener-apiserver` will prevent any Shoot with the `HibernationPossible` constraint set to `False` from being hibernated,
+that is via manual hibernation as well as scheduled hibernation.
+
+**`MaintenancePreconditionsSatisfied`**:
+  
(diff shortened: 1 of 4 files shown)


pkg/operation/botanist/constraints_check.go
```diff
@@ -150,22 +138,51 @@ func (b *Botanist) CheckForProblematicWebhooks(ctx context.Context) (gardencorev
...
+) string {
+	failurePolicyString := "nil"
+	if failurePolicy != nil {
+		failurePolicyString = string(*failurePolicy)
+	}
+	timeoutString := "and unset timeoutSeconds "
+	if timeoutSeconds != nil {
+		timeoutString = fmt.Sprintf("and %ds timeout ", *timeoutSeconds)
+	}
+	return fmt.Sprintf("%s %q is problematic: webhook %q with failurePolicy %q %smight prevent worker nodes from properly joining the shoot cluster",
```
timuthy:
```suggestion
	return fmt.Sprintf("%s %q is problematic: webhook %q with failurePolicy %q %s might prevent worker nodes from properly joining the shoot cluster",
```

Nit, as it's easier to read.

pkg/operation/botanist/constraints_check.go
```diff
@@ -150,22 +138,51 @@ func (b *Botanist) CheckForProblematicWebhooks(ctx context.Context) (gardencorev
...
 // webhooks with rules for CREATE/UPDATE/* pods or nodes and failurePolicy=Fail/nil. If the Shoot contains such a
-// webhook, we can never wake up this shoot cluster again  as new nodes cannot get created/ready, or our system
+// webhook, we can never wake up this shoot cluster again as new nodes cannot get created/ready, or our system
 // component pods cannot get created/ready (because the webhook's backing pod is not yet running).
 func IsProblematicWebhook(
 	failurePolicy *admissionregistrationv1beta1.FailurePolicyType,
 	objSelector *metav1.LabelSelector,
 	nsSelector *metav1.LabelSelector,
 	rules []admissionregistrationv1beta1.RuleWithOperations,
+	timeoutSeconds *int32,
```
timuthy: How about passing the entire webhook instance?

pkg/operation/botanist/constraints_check.go
```diff
@@ -150,22 +138,51 @@ func (b *Botanist) CheckForProblematicWebhooks(ctx context.Context) (gardencorev
...
 ) bool {
 	if failurePolicy != nil && *failurePolicy != admissionregistrationv1beta1.Fail {
 		// in admissionregistration.k8s.io/v1 FailurePolicy is defaulted to `Fail`
 		// see https://github.com/kubernetes/api/blob/release-1.16/admissionregistration/v1/types.go#L195
 		// and https://github.com/kubernetes/api/blob/release-1.16/admissionregistration/v1/types.go#L324
 		// therefore, webhook with FailurePolicy==nil is also considered problematic
-		return false
+		if timeoutSeconds != nil && *timeoutSeconds <= 15 {
```
timuthy: Can you create a `const` with expressive naming?

pkg/operation/botanist/constraints_check.go
```diff
@@ -150,22 +138,51 @@ func (b *Botanist) CheckForProblematicWebhooks(ctx context.Context) (gardencorev
...
+			// most control-plane API calls are made with a client-side timeout of 30s, so if webhooks has
+			// timeoutSeconds==30 the overall request might still fail although failurePolicy==Ignore, as there
+			// is overhead in communication with the API server and possible other webhooks.
+			// in admissionregistration/v1 timeoutSeconds is defaulted to 10 while in v1beta1 it's defaulted to 30.
+			// be restrictive here and mark all webhooks without a timeout set or timeouts > 15s as problematic to
+			// avoid ops effort. It's clearly documented that users should specify low timeouts, see
+			// https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts
```
timuthy: Nice explanation. What do you think about rather adding it to `docs/usage/shoot_status.md` to explain in more detail to the user why this timeout is bad?

pkg/operation/botanist/constraints_check.go
```diff
@@ -150,22 +138,51 @@ func (b *Botanist) CheckForProblematicWebhooks(ctx context.Context) (gardencorev
...
+	configName string,
+	webhookName string,
+	failurePolicy *admissionregistrationv1beta1.FailurePolicyType,
+	timeoutSeconds *int32,
+) string {
+	failurePolicyString := "nil"
+	if failurePolicy != nil {
+		failurePolicyString = string(*failurePolicy)
+	}
+	timeoutString := "and unset timeoutSeconds "
```
timuthy:
```suggestion
	timeoutString := "and unset timeoutSeconds"
```

pkg/operation/botanist/constraints_check.go
```diff
@@ -150,22 +138,51 @@ func (b *Botanist) CheckForProblematicWebhooks(ctx context.Context) (gardencorev
...
+	if timeoutSeconds != nil {
+		timeoutString = fmt.Sprintf("and %ds timeout ", *timeoutSeconds)
```
timuthy:
```suggestion
		timeoutString = fmt.Sprintf("and %ds timeout", *timeoutSeconds)
```

timuthy (changes requested):
Some minor suggestion, looks good otherwise.

pkg/operation/botanist/constraints_check.go
timebertt: Unfortunately, that's not possible, as there are different types for `ValidatingWebhook` and `MutatingWebhook`.
There's `webhook.{NewValidatingWebhookAccessor,NewMutatingWebhookAccessor} ` for this purpose, though I dropped it as part of https://github.com/gardener/gardener/pull/2370/files#diff-a62189d77ddf4b67cfac7ebccc31708acf390d2826fbb07cd2c28d9c43898b94L69, because they only accept the `v1` types, which we can't use as of now.

pkg/operation/botanist/constraints_check.go
timebertt: Ah sure, it was a leftover.

timebertt:
Thanks @timuthy for your suggestions, I addressed them.
I took the chance to overhall the documentation about the hibernation constraint to better explain to users why we have it and how they should properly configure their webhooks.

timuthy (dismissed):
Thanks for revising the documentation as well.
/lgtm

timebertt:
Done, PTAL.

rfranzke (approved):
/lgtm

timuthy (approved):

ialidzhikov:
Hi, team. I wanted to share that this PR is kind of breaking my local setup with 

```
error: shoots.core.gardener.cloud "aws-local" could not be patched: shoots.core.gardener.cloud "aws-local" is forbidden: MutatingWebhookConfiguration "gardener-extension-provider-aws-shoot" is problematic: webhook "shoot.aws.extensions.gardener.cloud" with failurePolicy "Ignore" and 30s timeout might prevent worker nodes from properly joining the shoot cluster
```

Shouldn't we first fix our webhooks' timeouts before enabling such constraints?

timebertt:
Uhh, good catch. Thanks for reporting.

PR #3414, opened 2021-01-15
[GEP-13] Add ManagedSeed validation

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Adds validation for `ManagedSeed` resources.

* `ManagedSeed` validation code
* Changes to the `GardenletConfiguration` validation to ensure that a correct field path is used if it's invoked as part of `ManagedSeed` validation.
* Removed the check that `server` and `sni` (and their optional fields) are specified in `GardenletConfiguration` validation, as this is inconsistent (these fields are optional in the API) and would break the `ManagedSeed` validation unless the user specifies values for all these fields. This was already discussed and agreed upon during the review of #3177.

The new validation code is not testable (except by unit tests), since the `ManagedSeed` resource is not yet registered in the API server. This is on purpose, to make this PR easier to review.

Fixes #

Based on #3412. This is the second of several PRs that are split from #3177 since it became too big and hard to review. Before commenting, please check if the topic has been already discussed in the previous PR.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/apis/seedmanagement/validation/managedseed.go
+++ b/pkg/apis/seedmanagement/validation/managedseed.go
@@ -28,6 +28,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	metav1validation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
@@ -74,7 +75,7 @@ func ValidateManagedSeedSpec(spec *seedmanagement.ManagedSeedSpec, fldPath *fiel
 
 	switch {
 	case spec.SeedTemplate != nil:
-		allErrs = append(allErrs, validateSeedTemplate(spec.SeedTemplate, fldPath.Child("seedTemplate"))...)
+		allErrs = append(allErrs, validateSeedTemplate(&spec.SeedTemplate.ObjectMeta, &spec.SeedTemplate.Spec, fldPath.Child("seedTemplate"))...)
 	case spec.Gardenlet != nil:
 		allErrs = append(allErrs, validateGardenlet(spec.Gardenlet, fldPath.Child("gardenlet"))...)
 	}
@@ -96,7 +97,7 @@ func ValidateManagedSeedSpecUpdate(newSpec, oldSpec *seedmanagement.ManagedSeedS
 
 	switch {
 	case newSpec.SeedTemplate != nil && oldSpec.SeedTemplate != nil:
-		allErrs = append(allErrs, validateSeedTemplateUpdate(newSpec.SeedTemplate, oldSpec.SeedTemplate, fldPath.Child("seedTemplate"))...)
+		allErrs = append(allErrs, validateSeedTemplateUpdate(&newSpec.SeedTemplate.ObjectMeta, &oldSpec.SeedTemplate.ObjectMeta, &newSpec.SeedTemplate.Spec, &oldSpec.SeedTemplate.Spec, fldPath.Child("seedTemplate"))...)
 	case newSpec.Gardenlet != nil && oldSpec.Gardenlet != nil:
 		allErrs = append(allErrs, validateGardenletUpdate(newSpec.Gardenlet, oldSpec.Gardenlet, fldPath.Child("gardenlet"))...)
 	}
@@ -104,21 +105,26 @@ func ValidateManagedSeedSpecUpdate(newSpec, oldSpec *seedmanagement.ManagedSeedS
 	return allErrs
 }
 
-func validateSeedTemplate(seedTemplate *seedmanagement.SeedTemplate, fldPath *field.Path) field.ErrorList {
+func validateSeedTemplate(metadata *metav1.ObjectMeta, seedSpec *gardencore.SeedSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
+	// Ensure name is not specified since it will be set by the controller
+	if metadata.Name != "" {
+		allErrs = append(allErrs, field.Forbidden(fldPath.Child("metadata", "name"), "seed name is forbidden"))
+	}
+
 	// Validate labels, annotations, and spec
-	allErrs = append(allErrs, metav1validation.ValidateLabels(seedTemplate.Labels, fldPath.Child("metadata", "labels"))...)
-	allErrs = append(allErrs, apivalidation.ValidateAnnotations(seedTemplate.Annotations, fldPath.Child("metadata", "annotations"))...)
-	allErrs = append(allErrs, corevalidation.ValidateSeedSpec(&seedTemplate.Spec, fldPath.Child("spec"))...)
+	allErrs = append(allErrs, metav1validation.ValidateLabels(metadata.Labels, fldPath.Child("metadata", "labels"))...)
+	allErrs = append(allErrs, apivalidation.ValidateAnnotations(metadata.Annotations, fldPath.Child("metadata", "annotations"))...)
+	allErrs = append(allErrs, corevalidation.ValidateSeedSpec(seedSpec, fldPath.Child("spec"))...)
 
 	return allErrs
 }
 
-func validateSeedTemplateUpdate(newSeedTemplate, oldSeedTemplate *seedmanagement.SeedTemplate, fldPath *field.Path) field.ErrorList {
+func validateSeedTemplateUpdate(_, _ *metav1.ObjectMeta, newSeedSpec, oldSeedSpec *gardencore.SeedSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	allErrs = append(allErrs, corevalidation.ValidateSeedSpecUpdate(&newSeedTemplate.Spec, &oldSeedTemplate.Spec, fldPath.Child("spec"))...)
+	allErrs = append(allErrs, corevalidation.ValidateSeedSpecUpdate(newSeedSpec, oldSeedSpec, fldPath.Child("spec"))...)
 
 	return allErrs
 }
@@ -196,6 +202,11 @@ func validateGardenletDeployment(deployment *seedmanagement.GardenletDeployment,
 	if deployment.RevisionHistoryLimit != nil {
 		allErrs = append(allErrs, apivalidation.ValidateNonnegativeField(int64(*deployment.RevisionHistoryLimit), fldPath.Child("revisionHistoryLimit"))...)
 	}
+	if deployment.ServiceAccountName != nil {
+		for _, msg := range apivalidation.ValidateServiceAccountName(*deployment.ServiceAccountName, false) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("serviceAccountName"), *deployment.ServiceAccountName, msg))
+		}
+	}
 
 	if deployment.Image != nil {
 		allErrs = append(allErrs, validateImage(deployment.Image, fldPath.Child("image"))...)
@@ -210,6 +221,12 @@ func validateGardenletDeployment(deployment *seedmanagement.GardenletDeployment,
 func validateImage(image *seedmanagement.Image, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
+	if image.Repository != nil && *image.Repository == "" {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("repository"), *image.Repository, "repository must not be empty if specified"))
+	}
+	if image.Tag != nil && *image.Tag == "" {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("tag"), *image.Tag, "tag must not be empty if specified"))
+	}
 	if image.PullPolicy != nil {
 		validValues := []string{string(corev1.PullAlways), string(corev1.PullIfNotPresent), string(corev1.PullNever)}
 		if !utils.ValueExists(string(*image.PullPolicy), validValues) {
@@ -242,7 +259,7 @@ func validateGardenletConfiguration(gardenletConfig *config.GardenletConfigurati
 		}
 
 		// Validate seed
-		allErrs = append(allErrs, validateSeed(seed, seedConfigPath)...)
+		allErrs = append(allErrs, validateSeedTemplate(&seed.ObjectMeta, &seed.Spec, seedConfigPath)...)
 	}
 
 	if gardenletConfig.GardenClientConnection != nil {
@@ -273,32 +290,8 @@ func validateGardenletConfigurationUpdate(newGardenletConfig, oldGardenletConfig
 		}
 
 		// Validate seed
-		allErrs = append(allErrs, validateSeedUpdate(newSeed, oldSeed, seedConfigPath)...)
+		allErrs = append(allErrs, validateSeedTemplateUpdate(&newSeed.ObjectMeta, &oldSeed.ObjectMeta, &newSeed.Spec, &oldSeed.Spec, seedConfigPath)...)
 	}
-
-	return allErrs
-}
-
-func validateSeed(seed *gardencore.Seed, fldPath *field.Path) field.ErrorList {
-	allErrs := field.ErrorList{}
-
-	// Ensure name is not specified since it will be set by the controller
-	if seed.Name != "" {
-		allErrs = append(allErrs, field.Forbidden(fldPath.Child("metadata", "name"), "seed name is forbidden"))
-	}
-
-	// Validate labels, annotations, and spec
-	allErrs = append(allErrs, metav1validation.ValidateLabels(seed.Labels, fldPath.Child("metadata", "labels"))...)
-	allErrs = append(allErrs, apivalidation.ValidateAnnotations(seed.Annotations, fldPath.Child("metadata", "annotations"))...)
-	allErrs = append(allErrs, corevalidation.ValidateSeedSpec(&seed.Spec, fldPath.Child("spec"))...)
-
-	return allErrs
-}
-
-func validateSeedUpdate(newSeed, oldSeed *gardencore.Seed, fldPath *field.Path) field.ErrorList {
-	allErrs := field.ErrorList{}
-
-	allErrs = append(allErrs, corevalidation.ValidateSeedSpecUpdate(&newSeed.Spec, &oldSeed.Spec, fldPath.Child("spec"))...)
 
 	return allErrs
 }
--- a/pkg/apis/seedmanagement/validation/managedseed.go
+++ b/pkg/apis/seedmanagement/validation/managedseed.go
@@ -1,0 +1,327 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validation
+
+import (
+	"fmt"
+
+	gardencore "github.com/gardener/gardener/pkg/apis/core"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	corevalidation "github.com/gardener/gardener/pkg/apis/core/validation"
+	"github.com/gardener/gardener/pkg/apis/seedmanagement"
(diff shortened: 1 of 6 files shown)


stoyanr:

This is the second of several PRs, as agreed. All comments on the old PR are addressed. Marked as draft until #3412 is merged. Unit tests are right now missing but I will add them.

timebertt:
Can you
/rebase
now that #3412 is merged and make this one ready for review?

pkg/apis/seedmanagement/validation/managedseed.go
```diff
@@ -0,0 +1,317 @@
...
+	// Ensure bootstrap is not changed
+	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newGardenlet.Bootstrap, oldGardenlet.Bootstrap, fldPath.Child("bootstrap"))...)
+
+	// Ensure merge with parent is not changed
+	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newGardenlet.MergeWithParent, oldGardenlet.MergeWithParent, fldPath.Child("mergeWithParent"))...)
+
+	return allErrs
+}
+
+func validateGardenletDeployment(deployment *seedmanagement.GardenletDeployment, fldPath *field.Path) field.ErrorList {
```
danielfoehrKn, line 196: Currently, only the image is validated from the GardenletDeployment. What about the other config options like ReplicaCount, RevisionHistoryLimit,... ?

danielfoehrKn (changes requested):

pkg/apis/seedmanagement/validation/managedseed.go
stoyanr: Good point, do you have suggestions how they should be validated? I can think of:

* For `ReplicaCount` and `RevisionHistoryLimit` there should be a check for non-negative integer. 
* `PodLabels` and `PodAnnotations` could be validated using the standard functions for validating labels and annotations that are already used elsewhere.

We could also validate other fields such as `Resources`, `Env`, and `AdditionalVolumeMounts`, similarly to how this is done for corev1 containers, see https://github.com/kubernetes/kubernetes/blob/e1ba7546bfe591d4e4fc0ce08c00652b79703b00/pkg/apis/core/validation/validation.go#L2817. However, this code is in `k8s.io/kubernetes`, so my assumption is we don't want to directly depend upon it, so we would need to mostly copy / paste it. Do we really want to do this? It looks like quite a chunk.

stoyanr: @rfranzke @timebertt Do you have suggestions here?

danielfoehrKn: I would just validate the obvious like ReplicaCount and  RevisionHistoryLimit to not having to dependent or copy from k/k

rfranzke: I would probably only implement it for fields which can be validated quite easily (like replica count, revision history limit, etc.). We don't want to depend on `k8s.io/kubernetes` and we don't want to copy all this code. Those fields are anyways "advanced" use-cases, so by default I guess nobody would set them (actually, I don't even know of real-world examples where somebody would need this). Actually, we could also remove them entirely at the moment, nobody would probably set them (it wasn't possible earlier, or?). If you want to keep them then I would live with the fact that failures in configuring those fields properly will result in reconciliation errors for now.

stoyanr: OK, let's keep them for now, and I will add the validation for `ReplicaCount` , `RevisionHistoryLimit`, `PodLabels`, and `PodAnnotations`. I will also add a short comment explaining why the rest are not validated.

stoyanr: Done

stoyanr:
@timebertt @danielfoehrKn @rfranzke I addressed the single comment so far. Is there any more feedback?

pkg/apis/seedmanagement/validation/managedseed.go
```diff
@@ -0,0 +1,327 @@
...
+		if err != nil {
+			allErrs = append(allErrs, field.InternalError(seedConfigPath, fmt.Errorf("could not convert seed config: %v", err)))
+			return allErrs
+		}
+
+		// Validate seed
+		allErrs = append(allErrs, validateSeed(seed, seedConfigPath)...)
+	}
+
+	if gardenletConfig.GardenClientConnection != nil {
```
danielfoehrKn, line 265: I think it makes sense to validate the SeedClientCOnnection for the case when the SeedClientConnection.Kubeconfig is not set and merge with parent is true.
This is the case where I would think that the gardenlet inherits the kubeconfig to the wrong cluster.

Mhm, but how can a gardenlet specify that he wants to use the in-cluster service account token for the Seed connection? Leave the kubeconfig field empty and merge with parent == false? 
By the same token, is it possible to use merge with parent == true and still use the in-cluster service account token for the Seed connection?

stoyanr: What exactly do you suggest to validate? Having `seedClientConnection.kubeconfig` not set and `mergeWithParent` as true is perfectly valid, in fact this is what we expect to get most of the time. As we agreed previously, the controller unsets `seedClientConnection.kubeconfig` in the parent before merging (since we never expect it to be valid), so this case is not causing any issues. From my perspective, there is nothing to validate here.

I believe this also answers your other questions, please let me know if there is still something unclear.

danielfoehrKn: > unsets seedClientConnection.kubeconfig in the parent before merging (since we never expect it to be valid)

That is the part I was missing. Then it is not an issue. But should be documented, I already forgot it after our last discussions :)

rfranzke (dismissed):
/lgtm

danielfoehrKn (dismissed):

pkg/gardenlet/apis/config/validation/validation.go
```diff
@@ -24,44 +24,34 @@ import (
...
 
 // ValidateGardenletConfiguration validates a GardenletConfiguration object.
-func ValidateGardenletConfiguration(cfg *config.GardenletConfiguration) field.ErrorList {
+func ValidateGardenletConfiguration(cfg *config.GardenletConfiguration, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	if cfg.Controllers != nil {
 		if cfg.Controllers.Shoot != nil {
-			allErrs = append(allErrs, ValidateShootControllerConfiguration(cfg.Controllers.Shoot, field.NewPath("controllers", "shoot"))...)
+			allErrs = append(allErrs, ValidateShootControllerConfiguration(cfg.Controllers.Shoot, fieldPath(fldPath, "controllers", "shoot"))...)
```
timebertt: Optional: you can directly use `Child` if you want. It will result in the same behavior as `fieldPath()` if `fldPath == nil`.
```suggestion
			allErrs = append(allErrs, ValidateShootControllerConfiguration(cfg.Controllers.Shoot, fieldPath.Child("controllers", "shoot"))...)
```

pkg/gardenlet/apis/config/validation/validation_test.go
```diff
@@ -193,61 +182,6 @@ var _ = Describe("GardenletConfiguration", func() {
...
-		It("should forbid not specifying a sni configuration", func() {
-			cfg.SNI = nil
-
-			errorList := ValidateGardenletConfiguration(cfg)
-
-			Expect(errorList).To(ConsistOf(PointTo(MatchFields(IgnoreExtras, Fields{
-				"Type":  Equal(field.ErrorTypeRequired),
-				"Field": Equal("sni"),
-			}))))
-		})
```
timebertt, line 205: Can you rather change the test cases regarding the `server` and `sni` fields to check that not supplying them is also valid, instead of removing the test cases?
We should ensure, that those cases stay allowed, as forbidding it would conflict with the ManagedSeed validation, as you pointed out.

pkg/apis/seedmanagement/validation/managedseed.go
```diff
@@ -0,0 +1,327 @@
...
+		allErrs = append(allErrs, validateGardenletUpdate(newSpec.Gardenlet, oldSpec.Gardenlet, fldPath.Child("gardenlet"))...)
+	}
+
+	return allErrs
+}
+
+func validateSeedTemplate(seedTemplate *seedmanagement.SeedTemplate, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	// Validate labels, annotations, and spec
```
timebertt, line 116: I think, we should also enforce an empty `metadata.name` field in the seed template similar to the `gardenletConfig.seedConfig.metadata.name` check in the `validateSeed` function: https://github.com/gardener/gardener/blob/cde35895cd5fa3505e6970e41ba3ec2099e65d11/pkg/apis/seedmanagement/validation/managedseed.go#L282-L288

You can potentially even merge both funcs into one, which accepts a `metav1.ObjectMeta` and a `gardencore.SeedSpec`.

pkg/apis/seedmanagement/validation/managedseed.go
```diff
@@ -0,0 +1,327 @@
...
+	return allErrs
+}
+
+func validateGardenletDeployment(deployment *seedmanagement.GardenletDeployment, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if deployment.ReplicaCount != nil {
+		allErrs = append(allErrs, apivalidation.ValidateNonnegativeField(int64(*deployment.ReplicaCount), fldPath.Child("replicaCount"))...)
+	}
+	if deployment.RevisionHistoryLimit != nil {
```
timebertt, line 202: Can you add a similar check, that the `ServiceAccountName` field is not the empty string, if it is not `nil`?

pkg/apis/seedmanagement/validation/managedseed.go
```diff
@@ -0,0 +1,327 @@
...
+
+	// Ensure seed selector is not specified
+	if gardenletConfig.SeedSelector != nil {
+		allErrs = append(allErrs, field.Forbidden(fldPath.Child("seedSelector"), "seed selector is forbidden"))
+	}
+
+	// Validate gardenlet config
+	allErrs = append(allErrs, configvalidation.ValidateGardenletConfiguration(gardenletConfig, fldPath)...)
+
+	if gardenletConfig.SeedConfig != nil {
```
timebertt, line 251: I find it a bit confusing, that the `seedConfig`/`seedTemplate` of a `ManagedSeed` is validated, but the the `seedConfig` of the standalone `GardenletConfiguration` is not validated. Did I miss something here?

Should we move the basic validation of the `SeedTemplate` to `corevalidation` and then call it from the `seedmanagement` validation and `gardenletconfig` validation? Additional checks can then be performed in both packages like enforcing an empty `Name` or non-empty `Name`.

pkg/apis/seedmanagement/validation/managedseed.go
```diff
@@ -0,0 +1,327 @@
...
+	allErrs = append(allErrs, metav1validation.ValidateLabels(deployment.PodLabels, fldPath.Child("podLabels"))...)
+	allErrs = append(allErrs, apivalidation.ValidateAnnotations(deployment.PodAnnotations, fldPath.Child("podAnnotations"))...)
+
+	return allErrs
+}
+
+func validateImage(image *seedmanagement.Image, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if image.PullPolicy != nil {
```
timebertt, line 230: Can you also validate that `Repository` and `Tag` are not empty if set?

pkg/apis/seedmanagement/validation/managedseed.go
```diff
@@ -0,0 +1,327 @@
...
+	allErrs = append(allErrs, apivalidation.ValidateAnnotations(seed.Annotations, fldPath.Child("metadata", "annotations"))...)
+	allErrs = append(allErrs, corevalidation.ValidateSeedSpec(&seed.Spec, fldPath.Child("spec"))...)
+
+	return allErrs
+}
+
+func validateSeedUpdate(newSeed, oldSeed *gardencore.Seed, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	allErrs = append(allErrs, corevalidation.ValidateSeedSpecUpdate(&newSeed.Spec, &oldSeed.Spec, fldPath.Child("spec"))...)
```
timebertt: Should we also call `ValidateObjectMetaUpdate` here?

pkg/apis/seedmanagement/validation/managedseed_test.go
```diff
@@ -0,0 +1,477 @@
...
+			})
+
+			It("should allow valid resources", func() {
+				errorList := ValidateManagedSeed(managedSeed)
+
+				Expect(errorList).To(HaveLen(0))
+			})
+
+			It("should forbid empty or invalid fields in gardenlet", func() {
+				seedxCopy := seedx.DeepCopy()
```
timebertt: Just FYI: You don't need to `DeepCopy` here. `BeforeEach` is executed before each `It` and will reset the variables.

timebertt (changes requested):
Thanks, I mainly have minor remarks.

pkg/gardenlet/apis/config/validation/validation.go
stoyanr: I can't call `fieldPath.Child` (`fieldPath` is a function), I guess you mean `fldPath.Child`. I thought it would result in a nil pointer panic if `fldPath == nil` so I introduced the `fieldPath` function. However you are right - `Child` never uses its receiver to access fields, so there is no panic, and it actually behaves as if `NewPath` was called. 

I am still hesitant though - I find it unnatural and perhaps dangerous to call methods on nil receivers and depend on the implementation to not get a panic. Not sure if this behavior can be considered an implementation detail or not, the documentation also doesn't say much. Perhaps it's better to keep the function as a precaution?

pkg/gardenlet/apis/config/validation/validation_test.go
stoyanr: OK, I will change the existing test cases to ensure that not supplying `server` and `sni` is valid (by removing `Server` and `SNI` in `BeforeEach`). The test cases that I removed should stay removed I believe (there are no similar test cases for other fields that can be nil, so I don't see why `server` and `sni` should be an exception).

pkg/apis/seedmanagement/validation/managedseed.go
stoyanr: Good point, I thought about it, I didn't do it because:
* I could easily add a check for the name - but what about other metadata fields, e.g. `namespace`, etc.? They will be ignored as well, shouldn't we also validate them?
* The Kubernetes deployment (and other similar objects) also silently ignores metadata fields different from labels and annotations, the validation doesn't fail if e.g. a name is specified.

The graceful way to handle this would be to not use `ObjectMeta` at all, but have a special type e.g. `RestrictedObjectMeta` that only has labels and annotations, so specifying a `Name` is not even possible. Didn't do it because:
* Again Kubernetes deployment etc. doesn't do it this way (though I believe it should).
* It would be inconsistent between `SeedTemplate` and `GardenletConfiguration`, and I was wary of changing `GardenletConfiguration`, besides the `Name` there is actually used by the `gardenlet`.

The cleanest (but involving the most changes and so potentially most time-consuming and dangerous) way to handle this IMO:
* Introduce `RestrictedObjectMeta` with labels and annotations only (perhaps in `core`).
* `SeedTemplate` uses `RestrictedObjectMeta` (could also be moved to `core`).
*  Both `ManagedSeed` and `GardenletConfiguration` use `SeedTemplate`. `GardenletConfiguration` has an additional `SeedName` field in `SeedConfig`.

What do you think? I am bit reluctant to change `GardenletConfiguration` at this point, on the other hand perhaps now is a better time than later.

pkg/apis/seedmanagement/validation/managedseed.go
stoyanr: OK. I thought its' harmless but upon testing I found out that it actually leads to an error.

pkg/apis/seedmanagement/validation/managedseed.go
stoyanr: Yes, a `seedConfig` in a standalone `GardenletConfiguration` (outside of a `ManagedSeed`) is not validated. This is the status quo, I am not sure if we really want to change it at this point. If we go this way, then perhaps we should also use the `SeedTemplate` in both (see also https://github.com/gardener/gardener/pull/3414#discussion_r563792340) - if we are doing disruptive changes,  then let's at least change the design to a cleaner one. But as said above, I am not a big fan of such changes of this point.

pkg/apis/seedmanagement/validation/managedseed.go
stoyanr: On the other hand, if we want to avoid such disruptive changes now (which I would prefer), then we could simply add the name validation as you suggested and be done with it. I already did it already exactly as you suggested, by combining `validateSeedTemplate` and `validateSeed` into one (and the same for updates).

pkg/apis/seedmanagement/validation/managedseed.go
stoyanr: OK

pkg/apis/seedmanagement/validation/managedseed.go
stoyanr: I don't think so. Looking at the code at https://github.com/kubernetes/apimachinery/blob/ee96543cc6730a24af1c7323dac47e52c340ba39/pkg/api/validation/objectmeta.go#L233, it mostly validates updates to fields that are ignored (they shouldn't be there at all, see previous comments). The only exception are `Name` (immutable), `Labels`, and `Annotations`, but these are validated by `ValidateManagedSeed` that is called by `ValidateManagedSeedUpdate`. As `Name` is forbidden checking for immutable is meaningless.

pkg/apis/seedmanagement/validation/managedseed_test.go
stoyanr: OK, you are right, will remove it.

stoyanr:
@timebertt Thanks for your review, will do the changes I marked as resolved shortly. For the rest, see my comments. 

As said, I agree `GardenletConfiguration` can be designed in a cleaner way and outlined a proposal how it could be done. However, I am not sure we want to do such disruptive changes and I am a bit reluctant to introduce them now. So I would only do it if there is a strong consensus for it.

pkg/gardenlet/apis/config/validation/validation.go
stoyanr: Finally I removed the function in favor of `fldPath.Child`, should be ok.

pkg/gardenlet/apis/config/validation/validation_test.go
stoyanr: Done.

pkg/apis/seedmanagement/validation/managedseed.go
stoyanr: Another consideration: if we call `ValidateObjectMetaUpdate` here, we should also call `ValidateObjectMeta` in `validateSeed`, otherwise it would be strange - but this is not working, since it would treat e.g. `Name` as required (and in our case, it's forbidden).

pkg/apis/seedmanagement/validation/managedseed.go
stoyanr: Done

pkg/apis/seedmanagement/validation/managedseed.go
stoyanr: Done

pkg/apis/seedmanagement/validation/managedseed.go
timebertt: > On the other hand, if we want to avoid such disruptive changes now (which I would prefer), then we could simply add the name validation as you suggested and be done with it.

I would also keep the API types like they are (aligned with the k8s APIs) and only validate the `name` field.
I think, validating the other fields is not necessary, as they will be ignored. `name` will be overwritten by the controller and the reason for enforcing an unset `name` field, was to avoid confusion.
So, this is all good now from my perspective, thanks!

pkg/apis/seedmanagement/validation/managedseed.go
timebertt: > If we go this way, then perhaps we should also use the `SeedTemplate` in both (see also [#3414 (comment)](https://github.com/gardener/gardener/pull/3414#discussion_r563792340)) - if we are doing disruptive changes, then let's at least change the design to a cleaner one.

Agree, that would be also my preferred approach.

> But as said above, I am not a big fan of such changes of this point.

Also agree. I just wanted to double check, that I didn't miss anything here. 
Let's set this straight in a dedicated PR (is also a breaking change, so it definitely should be in its own PR). Can you open an issue for this or just open a PR afterwards?

pkg/apis/seedmanagement/validation/managedseed.go
timebertt: Ok, got the motivation now, makes sense. Thanks for clarifying.

timebertt (approved):
Thanks @stoyanr for considering my suggestions and clarifying my questions!
Looks great to me now overall :)
/lgtm

Also, thanks for adding a dedicated commit for the changes, this was quite helpful for me.
/squash

danielfoehrKn (approved):

pkg/apis/seedmanagement/validation/managedseed.go
stoyanr: OK, I will do it, I already have it on my list from the previous review.

PR #3415, opened 2021-01-15
[GEP-13] Add ManagedSeed apiserver storage

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Adds `gardener-apiserver` registration and storage for `ManagedSeed` resources.

Creating, updating, and deleting `ManagedSeed` resources can now be tested, but admission plugins are still missing. This is on purpose, to make this PR easier to review. Without admission plugins, the user has to specify a few mandatory values that would otherwise be filled by the plugins based on the shoot.

Fixes #

Based on #3414. This is the third of several PRs that are split from #3177 since it became too big and hard to review. Before commenting, please check if the topic has been already discussed in the previous PR.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/example/55-managedseed-gardenlet.yaml
+++ b/example/55-managedseed-gardenlet.yaml
@@ -1,0 +1,40 @@
+apiVersion: seedmanagement.gardener.cloud/v1alpha1
+kind: ManagedSeed
+metadata:
+  name: my-managed-seed
+  namespace: garden # Must be garden
+spec:
+  shoot:
+    name: crazy-botany
+  # gardenlet specifies that the ManagedSeed controller should deploy a gardenlet into the cluster
+  # with the given deployment parameters and GardenletConfiguration.
+  gardenlet:
+    bootstrap: ServiceAccount # Mechanism that should be used for bootstrapping gardenlet connection to the Garden cluster, one of ServiceAccount, BootstrapToken, None
+    mergeWithParent: true # If true, the deployment parameters and GardenletConfiguration of the parent gardenlet will be merged with the specified deployment parameters and GardenletConfiguration
+#   deployment: # gardenlet deployment parameters
+#     replicaCount: 1      
+#     revisionHistoryLimit: 1
+#     serviceAccountName: gardenlet
+#     image:
+#       repository: eu.gcr.io/gardener-project/gardener/gardenlet
+#       tag: latest
+#       pullPolicy: IfNotPresent
+#     resources:
+#       requests:
+#         cpu: 100m
+#         memory: 100Mi
+#       limits:
+#         cpu: 2000m
+#         memory: 512Mi
+#     podAnnotations:
+#       foo: bar
+#     podLabels:
+#       foo: bar
+#     additionalVolumes: []
+#     additionalVolumeMounts: []
+#     env: []
+#     vpa: true
+    config: # GardenletConfiguration resource
+      apiVersion: gardenlet.config.gardener.cloud/v1alpha1
+      kind: GardenletConfiguration
+#     <See `20-componentconfig-gardenlet.yaml` for more details>
--- a/example/55-managedseed-seedtemplate.yaml
+++ b/example/55-managedseed-seedtemplate.yaml
@@ -1,0 +1,17 @@
+apiVersion: seedmanagement.gardener.cloud/v1alpha1
+kind: ManagedSeed
+metadata:
+  name: my-managed-seed
+  namespace: garden # Must be garden
+spec:
+  shoot:
+    name: crazy-botany
+  # seedTemplate is a template for a Seed object, that should be used to register a given cluster as a Seed.
+  # When seedTemplate is specified, the ManagedSeed controller will not deploy a gardenlet into the cluster
+  # and an existing gardenlet reconciling the new Seed is required.    
+  seedTemplate:
+#   metadata:
+#     labels:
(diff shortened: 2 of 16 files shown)


stoyanr:

This is the third of several PRs, as agreed. All comments on the old PR are addressed. Marked as draft until #3414 is merged.

stoyanr:
@rfranzke @timebertt @danielfoehrKn Should ManagedSeeds support special "operation" annotations such as "reconcile" or "retry"? 

My assumption so far: 
* "retry" is not required since failures will be retried by the controller and we don't have `LastOperation` in the status.
* "reconcile" might be needed, since if the parent gardenlet settings changed and we want to redeploy the child with the changed merged settings, we would like to reconcile the `ManagedSeed` without otherwise changing its spec.

So far there is no support for "reconcile", please let me know what you think and I will add it if needed.

rfranzke:
> "reconcile" might be needed, since if the parent gardenlet settings changed and we want to redeploy the child with the changed merged settings, we would like to reconcile the ManagedSeed without otherwise changing its spec.

How does this work today? When the gardenlet settings/config change then it (re)starts then it enqueues all the shooted seeds and updates them eventually (this i randomized to prevent updating all child gardenlets at once). Wouldn't it work the same way with the `ManagedSeed` resource? Is there a scenario where something in the parent's gardenlet config/settings changes which does not require a restart (thus, mentioned mechanism would not apply)?

stoyanr:
@rfranzke In fact it already works exactly as you described with the controller implementation in #3418 (except that randomization is switched off for easier testing), I tested it to be sure. In this case, perhaps the "reconcile" behavior is not needed and so support for the "operation" annotation should not be added at all?

rfranzke (dismissed):
/lgtm

rfranzke:
Can you also enhance the `controlplane` Helm chart and add a file similar to https://github.com/gardener/gardener/blob/4c6153dceae2e461607c9a38afca88277387dda0/charts/gardener/controlplane/charts/application/templates/apiservice-v1beta1-core-gardener-cloud.yaml#L1-L24

timebertt (commented):
Thanks, all changes lgtm.
Would now be the time to add a `ManagedSeed` manifest to `example` to allow easy testing?

stoyanr:
@rfranzke @timebertt Added the requested apiservice chart template and examples (in a new commit).

example/55-managedseed-seedtemplate.yaml
```diff
@@ -0,0 +1,81 @@
...
+#     - key: seed.gardener.cloud/protected # protected
+#     - key: <some-key>
+#     volume:
+#       minimumSize: 20Gi
+#     ingress: # If set then `.spec.dns.ingressDomain` must not be set and `.spec.dns.provider` must be set.
+#       domain: ingress.dev.my-seed.example.com # Must be a subdomain of the shoot DNS domain, will be initialized automatically if empty
+#       controller:
+#         kind: nginx
+#         providerConfig:
+#           <some-optional-config-for-the-nginx-ingress-controller>
```
rfranzke: To prevent double maintenance, instead of duplicating all of this here, shall be simply "refer" to the `Seed` example?

example/55-managedseed-gardenlet.yaml
```diff
@@ -0,0 +1,113 @@
...
+    #     - key: seed.gardener.cloud/protected # protected
+    #     - key: <some-key>
+    #     volume:
+    #       minimumSize: 20Gi
+    #     ingress: # If set then `.spec.dns.ingressDomain` must not be set and `.spec.dns.provider` must be set.
+    #       domain: ingress.dev.my-seed.example.com # Must be a subdomain of the shoot DNS domain, will be initialized automatically if empty
+    #       controller:
+    #         kind: nginx
+    #         providerConfig:
+    #           <some-optional-config-for-the-nginx-ingress-controller>
```
rfranzke: To prevent double maintenance, instead of duplicating all of this here, shall be simply "refer" to the `GardenletConfiguration` example?

danielfoehrKn (dismissed):
/lgtm thanks!

example/55-managedseed-seedtemplate.yaml
stoyanr: What exactly to do you mean by "refer"? Just a comment or something else?

Note that I wanted to create examples that are actually usable for testing, as @timebertt suggested. I actually needed most of what you see to test my new code. If I remove the snippet that you pointed above, it would be harder for other people to come up with a usable managed seed.

rfranzke: Just a comment would be good enough I think, something like https://github.com/gardener/gardener/blob/06092f85bac3c36fc0e4bbf1adc962a2b928626d/example/20-componentconfig-gardenlet.yaml#L105. I get the point about the testing, but I guess it should be manageable. I just have the fear that this will become outdated otherwise, making the examples potentially similarly useless for testing.

stoyanr: OK, I still have my doubts but I did as requested.

example/55-managedseed-gardenlet.yaml
stoyanr: OK, I still have my doubts but I did as requested.

stoyanr:
@rfranzke @timebertt All review comments are now addressed.

rfranzke (approved):
/lgtm

Issue #2158, opened 2020-04-07
ShootValidator admission plugin is mutating


ialidzhikov:
**Why is this needed**:
Currently the `ShootValidator` admission plugin is doing quite a lot of things which are not simple validation- it is defaulting Shoot spec.networking; it is updating the `shoot.garden.sapcloud.io/tasks` annotation; it is defaulting the machine image version; it is translating spec.kubernetes.version with `<major>.<minor>` to `<major.minor.patch>` and probably other mutating operations.
If the name of the admission plugin is `ShootValidator`, I wouldn't expect a mutation to happen inside.
I guess we could split the mutation/defaulting logic in other admission plugins and allow fine-grained configuration by the Gardener admins.

**Tasks**:
- [x] Introduce the `ShootMutator` admission plugin: https://github.com/gardener/gardener/pull/13156
- [x] Move mutation of the Shoot metadata annotation from the `ShootValidator` to the `ShootMutator` admission plugin : https://github.com/gardener/gardener/pull/13171
- [x] Ensure `ShootMutator` is executed before the `ShootValidator` admission plugin: https://github.com/gardener/gardener/pull/13276
- [x] Move defaulting of the Shoot networks from the `ShootValidator` to the `ShootMutator` admission plugin: https://github.com/gardener/gardener/pull/13207
- [x] Fetch the referenced CloudProfile / NamespacedCloudProfile: https://github.com/gardener/gardener/pull/13222
- [x] Move defaulting of the Shoot Kubernetes versions from the `ShootValidator` to the `ShootMutator` admission plugin:  https://github.com/gardener/gardener/pull/13252
- [x] Move defaulting of the Shoot machine image version from the `ShootValidator` to the `ShootMutator` admission plugin: https://github.com/gardener/gardener/pull/13351
- [x] Make the `ShootValidator` admission plugin validating one: https://github.com/gardener/gardener/pull/13352
- [x] Make the `.spec.provider.workers[].machine.image` field required for the Shoot resource (part 8): https://github.com/gardener/gardener/pull/13399

rfranzke:
This issue is open since a long time without anyone picking it up or any attention at all. Do we think that this is important enough that we spend time with it within the next month (or even year)? It seems there is not too much benefit/value of splitting it up (what would be use cases for enabling just one of the plugins while disabling the other?). Maybe we should just rename it to prevent confusing w.r.t. the `Validator` suffix in its name? Any opinions?

timebertt:
[...]

PR #3416, opened 2021-01-15
[GEP-13] Add ManagedSeed admission plugins

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Adds `gardener-apiserver` admission plugins for `ManagedSeed` resources.

* Admission plugin for `ManagedSeed` resources that validates / mutates a managed seed against the shoot it references.
* Admission plugin for `Shoot` resources that prevents deleting a shoot if it's being referenced by a managed seed.

Creating, updating, and deleting `ManagedSeed` resources can now be tested with full admission, which means that the user has to supply fewer values. 

Fixes #

Based on #3415. This is the fourth of several PRs that are split from #3177 since it became too big and hard to review. Before commenting, please check if the topic has been already discussed in the previous PR.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/apiserver/plugins.go
+++ b/pkg/apiserver/plugins.go
@@ -49,6 +49,7 @@ var (
 		extensionvalidation.PluginName,             // ExtensionValidator
 		shoottolerationrestriction.PluginName,      // ShootTolerationRestriction
 		shootdns.PluginName,                        // ShootDNS
+		shootmanagedseed.PluginName,                // ShootManagedSeed
 		shootquotavalidator.PluginName,             // ShootQuotaValidator
 		shootvalidator.PluginName,                  // ShootValidator
 		seedvalidator.PluginName,                   // SeedValidator
@@ -80,6 +81,7 @@ var (
 		extensionvalidation.PluginName,             // ExtensionValidator
 		shoottolerationrestriction.PluginName,      // ShootTolerationRestriction
 		shootdns.PluginName,                        // ShootDNS
+		shootmanagedseed.PluginName,                // ShootManagedSeed
 		shootquotavalidator.PluginName,             // ShootQuotaValidator
 		shootvalidator.PluginName,                  // ShootValidator
 		seedvalidator.PluginName,                   // SeedValidator
@@ -90,6 +92,7 @@ var (
 		clusteropenidconnectpreset.PluginName,      // ClusterOpenIDConnectPreset
 		shootstatedeletionvalidator.PluginName,     // ShootStateDeletionValidator
 		customverbauthorizer.PluginName,            // CustomVerbAuthorizer
+		managedseedvalidator.PluginName,            // ManagedSeed
 		mutatingwebhook.PluginName,                 // MutatingAdmissionWebhook
 		validatingwebhook.PluginName,               // ValidatingAdmissionWebhook
 		resourcequota.PluginName,                   // ResourceQuota
--- a/pkg/apiserver/plugins.go
+++ b/pkg/apiserver/plugins.go
@@ -26,9 +26,11 @@ import (
 	"github.com/gardener/gardener/plugin/pkg/global/deletionconfirmation"
 	"github.com/gardener/gardener/plugin/pkg/global/extensionvalidation"
 	"github.com/gardener/gardener/plugin/pkg/global/resourcereferencemanager"
+	managedseedvalidator "github.com/gardener/gardener/plugin/pkg/managedseed/validator"
 	plantvalidator "github.com/gardener/gardener/plugin/pkg/plant"
 	seedvalidator "github.com/gardener/gardener/plugin/pkg/seed/validator"
 	shootdns "github.com/gardener/gardener/plugin/pkg/shoot/dns"
+	shootmanagedseed "github.com/gardener/gardener/plugin/pkg/shoot/managedseed"
 	"github.com/gardener/gardener/plugin/pkg/shoot/oidc/clusteropenidconnectpreset"
 	"github.com/gardener/gardener/plugin/pkg/shoot/oidc/openidconnectpreset"
 	shootquotavalidator "github.com/gardener/gardener/plugin/pkg/shoot/quotavalidator"
@@ -58,6 +60,7 @@ var (
 		shootstatedeletionvalidator.PluginName,     // ShootStateDeletionValidator
 		customverbauthorizer.PluginName,            // CustomVerbAuthorizer
 		shootvpa.PluginName,                        // ShootVPAEnabledByDefault
+		managedseedvalidator.PluginName,            // ManagedSeed
 
 		// new admission plugins should generally be inserted above here
 		// webhook, and resourcequota plugins must go at the end
@@ -104,6 +107,7 @@ func RegisterAllAdmissionPlugins(plugins *admission.Plugins) {
 	shoottolerationrestriction.Register(plugins)
 	shootquotavalidator.Register(plugins)
 	shootdns.Register(plugins)
+	shootmanagedseed.Register(plugins)
 	shootvalidator.Register(plugins)
 	seedvalidator.Register(plugins)
 	controllerregistrationresources.Register(plugins)
@@ -112,6 +116,7 @@ func RegisterAllAdmissionPlugins(plugins *admission.Plugins) {
 	clusteropenidconnectpreset.Register(plugins)
 	shootstatedeletionvalidator.Register(plugins)
 	customverbauthorizer.Register(plugins)
+	managedseedvalidator.Register(plugins)
 	resourcequota.Register(plugins)
 	shootvpa.Register(plugins)
 }
--- a/plugin/pkg/managedseed/validator/admission.go
+++ b/plugin/pkg/managedseed/validator/admission.go
@@ -1,0 +1,463 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validator
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+
+	gardencore "github.com/gardener/gardener/pkg/apis/core"
+	gardencorehelper "github.com/gardener/gardener/pkg/apis/core/helper"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	"github.com/gardener/gardener/pkg/apis/seedmanagement"
+	"github.com/gardener/gardener/pkg/apis/seedmanagement/helper"
+	admissioninitializer "github.com/gardener/gardener/pkg/apiserver/admission/initializer"
+	coreclientset "github.com/gardener/gardener/pkg/client/core/clientset/internalversion"
+	coreinformers "github.com/gardener/gardener/pkg/client/core/informers/internalversion"
+	corelisters "github.com/gardener/gardener/pkg/client/core/listers/core/internalversion"
+	clientkubernetes "github.com/gardener/gardener/pkg/client/kubernetes"
+	seedmanagementclientset "github.com/gardener/gardener/pkg/client/seedmanagement/clientset/versioned"
+	"github.com/gardener/gardener/pkg/operation/common"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/admission"
+	kubeinformers "k8s.io/client-go/informers"
+	kubecorev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+const (
+	// PluginName is the name of this admission plugin.
+	PluginName = "ManagedSeedValidator"
+)
+
+// Register registers a plugin.
+func Register(plugins *admission.Plugins) {
+	plugins.Register(PluginName, func(config io.Reader) (admission.Interface, error) {
+		return New()
+	})
+}
+
+// ValidateManagedSeed contains listers and and admission handler.
+type ValidateManagedSeed struct {
+	*admission.Handler
+	shootLister          corelisters.ShootLister
+	secretBindingLister  corelisters.SecretBindingLister
+	secretLister         kubecorev1listers.SecretLister
+	coreClient           coreclientset.Interface
+	seedManagementClient seedmanagementclientset.Interface
+	readyFunc            admission.ReadyFunc
+}
+
+var (
+	_ = admissioninitializer.WantsInternalCoreInformerFactory(&ValidateManagedSeed{})
+	_ = admissioninitializer.WantsInternalCoreClientset(&ValidateManagedSeed{})
+	_ = admissioninitializer.WantsSeedManagementClientset(&ValidateManagedSeed{})
+	_ = admissioninitializer.WantsKubeInformerFactory(&ValidateManagedSeed{})
+
+	readyFuncs = []admission.ReadyFunc{}
+)
+
+// New creates a new ValidateManagedSeed admission plugin.
+func New() (*ValidateManagedSeed, error) {
+	return &ValidateManagedSeed{
+		Handler: admission.NewHandler(admission.Create, admission.Update),
+	}, nil
+}
+
+// AssignReadyFunc assigns the ready function to the admission handler.
+func (v *ValidateManagedSeed) AssignReadyFunc(f admission.ReadyFunc) {
+	v.readyFunc = f
+	v.SetReadyFunc(f)
+}
+
+// SetInternalCoreInformerFactory gets Lister from SharedInformerFactory.
+func (v *ValidateManagedSeed) SetInternalCoreInformerFactory(f coreinformers.SharedInformerFactory) {
+	shootInformer := f.Core().InternalVersion().Shoots()
+	v.shootLister = shootInformer.Lister()
+
+	secretBindingInformer := f.Core().InternalVersion().SecretBindings()
+	v.secretBindingLister = secretBindingInformer.Lister()
+
+	readyFuncs = append(readyFuncs, shootInformer.Informer().HasSynced, secretBindingInformer.Informer().HasSynced)
+}
+
+// SetKubeInformerFactory gets Lister from SharedInformerFactory.
+func (v *ValidateManagedSeed) SetKubeInformerFactory(f kubeinformers.SharedInformerFactory) {
+	secretInformer := f.Core().V1().Secrets()
+	v.secretLister = secretInformer.Lister()
+
+	readyFuncs = append(readyFuncs, secretInformer.Informer().HasSynced)
+}
+
+// SetInternalCoreClientset sets the garden core clientset.
+func (v *ValidateManagedSeed) SetInternalCoreClientset(c coreclientset.Interface) {
+	v.coreClient = c
+}
+
+// SetSeedManagementClientset sets the garden seedmanagement clientset.
+func (v *ValidateManagedSeed) SetSeedManagementClientset(c seedmanagementclientset.Interface) {
+	v.seedManagementClient = c
+}
+
+// ValidateInitialization checks whether the plugin was correctly initialized.
+func (v *ValidateManagedSeed) ValidateInitialization() error {
+	if v.shootLister == nil {
+		return errors.New("missing shoot lister")
+	}
+	if v.secretBindingLister == nil {
+		return errors.New("missing secret binding lister")
+	}
+	if v.secretLister == nil {
+		return errors.New("missing secret lister")
+	}
+	if v.coreClient == nil {
+		return errors.New("missing garden core client")
+	}
+	if v.seedManagementClient == nil {
+		return errors.New("missing garden seedmanagement client")
+	}
+	return nil
+}
+
+var _ admission.MutationInterface = &ValidateManagedSeed{}
+
+// Admit validates and if appropriate mutates the given managed seed against the shoot that it references.
+func (v *ValidateManagedSeed) Admit(ctx context.Context, a admission.Attributes, o admission.ObjectInterfaces) error {
+	// Wait until the caches have been synced
+	if v.readyFunc == nil {
+		v.AssignReadyFunc(func() bool {
+			for _, readyFunc := range readyFuncs {
+				if !readyFunc() {
+					return false
+				}
+			}
+			return true
+		})
+	}
+	if !v.WaitForReady() {
+		return admission.NewForbidden(a, errors.New("not yet ready to handle request"))
+	}
+
+	// Ignore all kinds other than ManagedSeed
+	if a.GetKind().GroupKind() != seedmanagement.Kind("ManagedSeed") {
+		return nil
+	}
+
+	// Ignore updates to status or other subresources
+	if a.GetSubresource() != "" {
+		return nil
+	}
+
+	// Convert object to ManagedSeed
+	managedSeed, ok := a.GetObject().(*seedmanagement.ManagedSeed)
+	if !ok {
+		return apierrors.NewBadRequest("could not convert object to ManagedSeed")
+	}
+
+	var allErrs field.ErrorList
+	gk := schema.GroupKind{Group: seedmanagement.GroupName, Kind: "ManagedSeed"}
+
+	// Ensure shoot name is specified
+	shootNamePath := field.NewPath("spec", "shoot", "name")
+	if managedSeed.Spec.Shoot.Name == "" {
+		return apierrors.NewInvalid(gk, managedSeed.Name, append(allErrs, field.Required(shootNamePath, "shoot name is required")))
+	}
+
+	// Get shoot
+	shoot, err := v.getShoot(ctx, managedSeed.Namespace, managedSeed.Spec.Shoot.Name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return apierrors.NewInvalid(gk, managedSeed.Name, append(allErrs, field.Invalid(shootNamePath, managedSeed.Spec.Shoot.Name, fmt.Sprintf("shoot %s/%s not found", managedSeed.Namespace, managedSeed.Spec.Shoot.Name))))
+		}
+		return apierrors.NewInternalError(fmt.Errorf("could not get shoot %s/%s: %v", managedSeed.Namespace, managedSeed.Spec.Shoot.Name, err))
+	}
+
+	// Ensure shoot can be registered as seed (specifies a domain)
+	if shoot.Spec.DNS == nil || shoot.Spec.DNS.Domain == nil || *shoot.Spec.DNS.Domain == "" {
+		return apierrors.NewInvalid(gk, managedSeed.Name, append(allErrs, field.Invalid(shootNamePath, managedSeed.Spec.Shoot.Name, fmt.Sprintf("shoot %s does not specify a domain", kutil.ObjectName(shoot)))))
+	}
+
+	// Ensure shoot is not already registered as seed
+	ms, err := kutil.GetManagedSeed(ctx, v.seedManagementClient, managedSeed.Namespace, managedSeed.Spec.Shoot.Name)
+	if err != nil {
+		return apierrors.NewInternalError(fmt.Errorf("could not get managed seed for shoot %s/%s: %v", managedSeed.Namespace, managedSeed.Spec.Shoot.Name, err))
+	}
+	if ms != nil && ms.Name != managedSeed.Name {
+		return apierrors.NewInvalid(gk, managedSeed.Name, append(allErrs, field.Invalid(shootNamePath, managedSeed.Spec.Shoot.Name, fmt.Sprintf("shoot %s already registered as seed", kutil.ObjectName(shoot)))))
+	}
+
+	switch {
+	case managedSeed.Spec.SeedTemplate != nil:
+		// Admit seed spec against shoot
+		errs, err := v.admitSeedSpec(ctx, &managedSeed.Spec.SeedTemplate.Spec, shoot, field.NewPath("spec", "seedTemplate", "spec"))
+		if err != nil {
+			return err
+		}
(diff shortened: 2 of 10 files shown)


stoyanr:

This is the fourth of several PRs, as agreed. All comments on the old PR are addressed. Marked as draft until #3415 is merged. Unit tests are right now missing but I will add them.

plugin/pkg/shoot/managedseed/admission.go
```diff
@@ -0,0 +1,179 @@
...
+		return apierrors.NewInternalError(fmt.Errorf("could not get ManagedSeed for shoot '%s/%s': %v", a.GetNamespace(), a.GetName(), err))
+	}
+	if managedSeed == nil {
+		return nil
+	}
+
+	return admission.NewForbidden(a, fmt.Errorf("cannot delete shoot %s/%s since it is still referenced by a managed seed", a.GetNamespace(), a.GetName()))
+}
+
+func (d *ManagedSeed) newAttributesWithName(a admission.Attributes, name string) admission.Attributes {
```
timebertt: Receiver names are different
```suggestion
func (v *ManagedSeed) newAttributesWithName(a admission.Attributes, name string) admission.Attributes {
```

plugin/pkg/shoot/managedseed/admission.go
```diff
@@ -0,0 +1,179 @@
...
+	if v.seedManagementClient == nil {
+		return errors.New("missing garden seedmanagement client")
+	}
+	return nil
+}
+
+var _ admission.ValidationInterface = &ManagedSeed{}
+
+// Validate validates if the Shoot can be deleted. If the
+func (v *ManagedSeed) Validate(ctx context.Context, a admission.Attributes, _ admission.ObjectInterfaces) error {
```
timebertt, line 99: Hmm, do we really need an extra admission plugin for this?
What was your motivation not to add the logic to the `ShootValidator` plugin?

plugin/pkg/shoot/managedseed/admission_test.go
```diff
@@ -0,0 +1,207 @@
...
+					Namespace: namespace,
+				},
+				Spec: seedmanagementv1alpha1.ManagedSeedSpec{
+					Shoot: seedmanagementv1alpha1.Shoot{
+						Name: name,
+					},
+				},
+			}
+
+			admissionHandler, _ = New()
```
timebertt, line 73: Please don't ignore errors in tests

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/admission"
+	kubeinformers "k8s.io/client-go/informers"
+	kubecorev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+const (
+	// PluginName is the name of this admission plugin.
+	PluginName = "ManagedSeedValidator"
```
timebertt: Should we rename this plugin to just `ManagedSeed`? Ref https://github.com/gardener/gardener/issues/2158

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+
+	var allErrs field.ErrorList
+	gk := schema.GroupKind{Group: seedmanagement.GroupName, Kind: "ManagedSeed"}
+
+	// Ensure shoot name is specified
+	shootNamePath := field.NewPath("spec", "shoot", "name")
+	if managedSeed.Spec.Shoot.Name == "" {
+		return apierrors.NewInvalid(gk, managedSeed.Name, append(allErrs, field.Required(shootNamePath, "shoot name is required")))
+	}
```
timebertt, line 184: This is already validated by API validation (`seedmanagement/validation`), no need to duplicate it here.

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+		return apierrors.NewInvalid(gk, managedSeed.Name, append(allErrs, field.Invalid(shootNamePath, managedSeed.Spec.Shoot.Name, fmt.Sprintf("shoot %s does not specify a domain", kutil.ObjectName(shoot)))))
+	}
+
+	// Ensure shoot is not already registered as seed
+	ms, err := kutil.GetManagedSeed(ctx, v.seedManagementClient, managedSeed.Namespace, managedSeed.Spec.Shoot.Name)
+	if err != nil {
+		return apierrors.NewInternalError(fmt.Errorf("could not get managed seed for shoot %s/%s: %v", managedSeed.Namespace, managedSeed.Spec.Shoot.Name, err))
+	}
+	if ms != nil && ms.Name != managedSeed.Name {
+		return apierrors.NewInvalid(gk, managedSeed.Name, append(allErrs, field.Invalid(shootNamePath, managedSeed.Spec.Shoot.Name, fmt.Sprintf("shoot %s already registered as seed", kutil.ObjectName(shoot)))))
```
timebertt: Can you add the name of the other managed seed to the error message for better user experience?

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+	case managedSeed.Spec.SeedTemplate != nil:
+		// Admit seed spec against shoot
+		errs, err := v.admitSeedSpec(ctx, &managedSeed.Spec.SeedTemplate.Spec, shoot, field.NewPath("spec", "seedTemplate", "spec"))
+		if err != nil {
+			return err
+		}
+		allErrs = append(allErrs, errs...)
+
+	case managedSeed.Spec.Gardenlet != nil:
+		// Admit gardelnet against shoot
```
timebertt:
```suggestion
		// Admit gardenlet against shoot
```

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+
+	return allErrs, nil
+}
+
+func (v *ValidateManagedSeed) admitSeedSpec(ctx context.Context, spec *gardencore.SeedSpec, shoot *gardencore.Shoot, fldPath *field.Path) (field.ErrorList, error) {
+	var allErrs field.ErrorList
+
+	// Initialize backup provider
+	if spec.Backup != nil && spec.Backup.Provider == "" {
```
timebertt, line 282: For this to take effect, the operator has to specify at least one field in the `seed.spec.backup` field.
I'm wondering, if we should remove the `omitempty` tag from the `backup` field, so that one can specify `backup={}` and then provider, region and maybe even secretRef will be defaulted for `ManagedSeeds`.
WDYT?
Doesn't need to be tackled in this PR, though. Would be an additional convenience feature.

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+			return nil, err
+		}
+	}
+	if dnsProvider == nil {
+		return nil, fmt.Errorf("domain of shoot %s is neither a custom domain nor a default domain", kutil.ObjectName(shoot))
+	}
+	return dnsProvider, nil
+}
+
+func (v *ValidateManagedSeed) getSeedDNSProviderForCustomDomain(ctx context.Context, shoot *gardencore.Shoot) (*gardencore.SeedDNSProvider, error) {
```
timebertt: Unused parameter
```suggestion
func (v *ValidateManagedSeed) getSeedDNSProviderForCustomDomain(shoot *gardencore.Shoot) (*gardencore.SeedDNSProvider, error) {
```

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+			if err != nil {
+				if apierrors.IsInternalError(err) {
+					return allErrs, err
+				}
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("ingress"), spec.Ingress, err.Error()))
+			}
+			spec.DNS.Provider = dnsProvider
+		}
+		// TODO If spec.DNS.Provider is not nil should we check if the user specified exactly the DNS provider we found above?
+		// I assume not, since the shoot may have multiple DNS providers
```
timebertt: I assume the same

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+
+		if spec.Ingress.Domain == "" {
+			spec.Ingress.Domain = ingressDomain
+		} else if spec.Ingress.Domain != ingressDomain {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("ingress", "domain"), spec.Ingress.Domain, fmt.Sprintf("seed ingress domain must be equal to shoot DNS domain %s", ingressDomain)))
```
timebertt, line 307: I don't think that this is strictly required.
I think, it would be a valid setup to specify a different domain for the managed ingress than the shoot domain, right?

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+		}
+	} else {
+		if spec.DNS.IngressDomain == nil || *spec.DNS.IngressDomain == "" {
+			spec.DNS.IngressDomain = &ingressDomain
+		} else if *spec.DNS.IngressDomain != ingressDomain {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("dns", "ingressDomain"), spec.DNS.IngressDomain, fmt.Sprintf("seed ingress domain must be equal to shoot DNS domain %s", ingressDomain)))
```
timebertt: Same here.
If the operator deploys a custom ingress controller and a dns entry pointing to the ingress controller, they should be able to specify that custom domain in the `dns.ingressDomain` field.

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+		// I assume not, since the shoot may have multiple DNS providers
+
+		if spec.Ingress.Domain == "" {
+			spec.Ingress.Domain = ingressDomain
+		} else if spec.Ingress.Domain != ingressDomain {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("ingress", "domain"), spec.Ingress.Domain, fmt.Sprintf("seed ingress domain must be equal to shoot DNS domain %s", ingressDomain)))
+		}
+	} else {
+		if spec.DNS.IngressDomain == nil || *spec.DNS.IngressDomain == "" {
+			spec.DNS.IngressDomain = &ingressDomain
```
timebertt, line 311: Hmm, the `ingressDomain` variable holds the ingress domain according to the pattern used in the ingress addon.
I think, we should check if the addon is really enabled before we default it to this value.

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+		} else if *spec.DNS.IngressDomain != ingressDomain {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("dns", "ingressDomain"), spec.DNS.IngressDomain, fmt.Sprintf("seed ingress domain must be equal to shoot DNS domain %s", ingressDomain)))
+		}
+	}
```
timebertt, line 315: If the operator neither specified the managed ingress with a domain nor a different ingress domain and the shoot doesn't have the ingress addon enabled, this should be denied by this admission plugin, as we can't expose any ingresses on such a seed (or at least don't know where it can be reached).

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+	} else if spec.Provider.Type != shoot.Spec.Provider.Type {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("provider", "type"), spec.Provider.Type, fmt.Sprintf("seed provider type must be equal to shoot provider type %s", shoot.Spec.Provider.Type)))
+	}
+	if spec.Provider.Region == "" {
+		spec.Provider.Region = shoot.Spec.Region
+	} else if spec.Provider.Region != shoot.Spec.Region {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("provider", "region"), spec.Provider.Region, fmt.Sprintf("seed provider region must be equal to shoot region %s", shoot.Spec.Region)))
+	}
+
+	return allErrs, nil
```
timebertt, line 346: I think we should also cross check that exactly one of the seed VPA and shoot VPA is enabled, but not both and not none of both.

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,463 @@
...
+
+	return &gardencore.SeedDNSProvider{
+		Type:      *primaryProvider.Type,
+		SecretRef: secretRef,
+		Domains:   primaryProvider.Domains,
+		Zones:     primaryProvider.Zones,
+	}, nil
+}
+
+func (v *ValidateManagedSeed) getSeedDNSProviderForDefaultDomain(ctx context.Context, shoot *gardencore.Shoot) (*gardencore.SeedDNSProvider, error) {
```
timebertt: Unused parameter
```suggestion
func (v *ValidateManagedSeed) getSeedDNSProviderForDefaultDomain(shoot *gardencore.Shoot) (*gardencore.SeedDNSProvider, error) {
```

timebertt (changes requested):
Thanks! Please see feedback inline.
I will continue with the remaining unit tests and play around with it locally once it's addressed.

plugin/pkg/shoot/managedseed/admission.go
stoyanr: OK

plugin/pkg/shoot/managedseed/admission.go
stoyanr: * The `ShootValidator` plugin validates `Create` and `Update`, similarly to most other "validators". Fitting `Delete` would mean adding logic to differentiate both cases. It was simpler to add a new plugin than to modify the existing one.
* The `Delete` validation based on `ManagedSeed` non-existence is independent of anything in `ShootValidator`. It makes sense to have a separate plugin simply from "separation of concerns" (SOLID principles, etc.). This way it's much easier to understand, test, and maintain both plugins.
* There is already a number of different admission plugins for `Shoot` to check other conditions such as quotas, etc. From my perspective, adding another one fits well into this picture.

In general, unless there is an obvious reason to modify an existing type, I almost always create a new type to achieve separation of concerns and the other benefits mentioned above.

plugin/pkg/shoot/managedseed/admission_test.go
stoyanr: OK

plugin/pkg/shoot/managedseed/admission.go
timebertt: Thanks for clarifying, fine for me then.

plugin/pkg/managedseed/validator/admission.go
stoyanr: Plugins that validate (and sometimes mutate) `Create` / `Update` based mostly on field values are currently named `Validators` (Plant, Seed, Shoot, ShootQuota, etc.). I followed the same naming conventions for consistency. I agree with the point that `Validator` is wrong naming here, though I personally think the impact of this is low. I will rename it to `ManagedSeed` as suggested.

plugin/pkg/managedseed/validator/admission.go
stoyanr: I am afraid there is a need. You could try it yourself - remove the lines above and then create a `ManagedSeed` with an empty shoot name to see what happens.

Spoiler: Instead of a neat error message from validation, you would get:
```
Error from server (InternalError): error when creating "dev/managedseed-i500152-gcpx.yaml": Internal error occurred: could not get shoot garden/: resource name may not be empty
```
The reason is that the admission plugin kicks in before validation, and it needs to read the shoot.

In fact, the check in validation is the one that could be removed. However, I decided to keep it there as well because it's quite an obvious thing to do and it could disturb other people (including reviewers) that it't not there.

plugin/pkg/managedseed/validator/admission.go
stoyanr: OK

plugin/pkg/managedseed/validator/admission.go
stoyanr: OK

plugin/pkg/managedseed/validator/admission.go
stoyanr: What you described is mostly how it works today: `backup: {}` results in a fully and correctly initialized `backup` section in `seedConfig`, in `seedTemplate` you would also need to specify an empty `secretRef`, like this:
```yaml
backup:
  secretRef: {}
```
The difference comes from the fact that `gardenlet.config` is a raw extension and therefore not validated by the API server before the admission plugin kicks in, while `seedTemplate` is strongly typed and is validated.

In both cases, don't think `omitempty` plays any role.

plugin/pkg/managedseed/validator/admission.go
stoyanr: OK

plugin/pkg/managedseed/validator/admission.go
stoyanr: OK, I will remove this TODO then, it was mostly meant to draw reviewers attention.

plugin/pkg/managedseed/validator/admission.go
stoyanr: We are talking about the "seed ingress controller" here (as opposed to an ingress controller deployed to the shoot). Is it still valid to have a different domain? 

Looking at the current `use-as-seed` implementation, this doesn't seem to be the case, see https://github.com/gardener/gardener/blob/581c1663942f9af1c532656d86eb1e28322dd838/pkg/gardenlet/controller/shoot/seed_registration_control.go#L362-L385.

plugin/pkg/managedseed/validator/admission.go
stoyanr: This code was basically copied from the existing `use-as-seed` implementation. I had some doubts but decided to keep it as it is, I simply wasn't sure. See https://github.com/gardener/gardener/blob/581c1663942f9af1c532656d86eb1e28322dd838/pkg/gardenlet/controller/shoot/seed_registration_control.go#L362-L385.
Should I then check if the ingress addon is enabled, and only if it is, set the ingress domain, and otherwise leave it empty (which would fail seed validation)?

plugin/pkg/managedseed/validator/admission.go
stoyanr: OK

plugin/pkg/managedseed/validator/admission.go
stoyanr: OK, I agree in this case. Shouldn't we validate that it's at least a subdomain of the shoot domain? In the `use-as-seed` implementation (and also here, as I made sure to preserve it), there is a check if the shoot specifies a domain, and if it's not the case, the seed registration is rejected. What would be the point to do it if this domain is not used for anything?

plugin/pkg/managedseed/validator/admission.go
stoyanr: No need to do anything, if both are empty the Seed validation (that is part of the ManagedSeed validation) would deny the ManagedSeed with an appropriate message.
```
The ManagedSeed "i500152-gcpx" is invalid: ...: either specify spec.ingress or spec.dns.ingressDomain
```

plugin/pkg/managedseed/validator/admission.go
stoyanr: This is done by the controller. Unfortunately, we can't check it here because we don't have access to the seed. Back in the old PR, @rfranzke proposed some sophisticated mechanism based on new shoot conditions that would allow us to do it here, however I don't believe the outcome warrants the additional complexity.

Currently it works like this: the ManagedSeed would be accepted and stored, but the controller would recognize that both VPAs are enabled and report an error, the managed seed would not get reconciled until one of the VPAs is disabled. 

So I would propose to merge the admission plugin as it is and go back to the discussion during the review of the controller PR, if needed.

plugin/pkg/shoot/managedseed/admission.go
stoyanr: Done

plugin/pkg/shoot/managedseed/admission_test.go
stoyanr: Done

plugin/pkg/managedseed/validator/admission.go
stoyanr: Done

plugin/pkg/managedseed/validator/admission.go
stoyanr: Done

plugin/pkg/managedseed/validator/admission.go
stoyanr: Done

plugin/pkg/managedseed/validator/admission.go
stoyanr: Done

plugin/pkg/managedseed/validator/admission.go
stoyanr: Based on the above, I changed the check to `HasSuffix`. If you think this is still not what is needed here, let me know.

plugin/pkg/managedseed/validator/admission.go
stoyanr: Based on the above, I added a check for the ingress addon being enabled.

plugin/pkg/managedseed/validator/admission.go
stoyanr: Done

plugin/pkg/managedseed/validator/admission.go
stoyanr: I renamed the type to `ManagedSeed` but kept the package as `validator` (again, for consistency with other similar packages).

stoyanr:
@timebertt I addressed all comments marked as "resolved" (and some of the unresolved ones, but there I still have some doubts, see my comments), and replied to the rest.

pkg/apiserver/plugins.go
```diff
@@ -58,6 +60,7 @@ var (
 		shootstatedeletionvalidator.PluginName,     // ShootStateDeletionValidator
 		customverbauthorizer.PluginName,            // CustomVerbAuthorizer
 		shootvpa.PluginName,                        // ShootVPAEnabledByDefault
+		managedseedvalidator.PluginName,            // ManagedSeed
```
rfranzke, line 64: Shouldn't the `shootmanagedseed.PluginName` be in this list as well?

pkg/apiserver/plugins.go
```diff
@@ -112,6 +116,7 @@ func RegisterAllAdmissionPlugins(plugins *admission.Plugins) {
 	clusteropenidconnectpreset.Register(plugins)
 	shootstatedeletionvalidator.Register(plugins)
 	customverbauthorizer.Register(plugins)
+	managedseedvalidator.Register(plugins)
```
rfranzke, line 122: I guess both new plugins should be enabled by default, or? Can you add them to `DefaultOnPlugins`?

plugin/pkg/managedseed/validator/admission.go
rfranzke: Yeah, let's keep it like this for {shooted,managed} seeds. I cannot think of a use-case where one would want to have a different ingress domain (even if an own ingress controller is dpeloyed).

plugin/pkg/managedseed/validator/admission.go
rfranzke: Yeah, the problem with the VPA is the "migration phase". It's a bit special and cannot be tackled here already.

pkg/apiserver/plugins.go
stoyanr: Yes, it was "lost" during the rebase I did 2 hours ago, thanks for noticing.

pkg/apiserver/plugins.go
stoyanr: Done

pkg/apiserver/plugins.go
stoyanr: Done

rfranzke (approved):
/lgtm

plugin/pkg/managedseed/validator/admission.go
timebertt: > The reason is that the admission plugin kicks in before validation, and it needs to read the shoot.

Ok, I didn't have this in mind. But you're right of course.
Thanks for clarifying.

plugin/pkg/managedseed/validator/admission.go
timebertt: Yes, for the `seedConfig` case it works because of the raw extension, right.
If we would remove the `omitempty` tag for the `backup` field, one could also set `backup: {}` just like in the `seedConfig` case. 
I think, this would make it consistent and avoid confusion. WDYT?

plugin/pkg/managedseed/validator/admission.go
timebertt: Well, fine for me then. Can be changed later on, if we think it's relevant.

plugin/pkg/managedseed/validator/admission.go
timebertt: Ok, works for me for now 👍

plugin/pkg/managedseed/validator/admission.go
timebertt: I think, it's good now like you did it.

plugin/pkg/managedseed/validator/admission.go
timebertt: Ah, perfect. Sorry.

plugin/pkg/managedseed/validator/admission.go
timebertt: Ok, sgtm. Thanks for clarifying.

plugin/pkg/managedseed/validator/admission.go
```diff
@@ -0,0 +1,461 @@
...
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("networks", "pods"), spec.Networks.Pods, fmt.Sprintf("seed pods CIDR must be equal to shoot pods CIDR %s", *shoot.Spec.Networking.Pods)))
+	}
+	if spec.Networks.Services == "" && shoot.Spec.Networking.Services != nil {
+		spec.Networks.Services = *shoot.Spec.Networking.Services
+	} else if shoot.Spec.Networking.Services != nil && spec.Networks.Services != *shoot.Spec.Networking.Services {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("networks", "services"), spec.Networks.Pods, fmt.Sprintf("seed services CIDR must be equal to shoot services CIDR %s", *shoot.Spec.Networking.Services)))
+	}
+
+	// Initialize and validate provider
+	if spec.Provider.Type == "" {
```
timebertt, line 335: Unfortunately, if one creates/applies such a `ManagedSeed` manifest via kubectl
```yaml
spec:
  seedTemplate:
    spec:
      provider: {}
```
one has to use `--validate=false`:

```
$ k create -f local/managedseed/55-managedseed-seedtemplate.yaml
error: error validating "local/managedseed/55-managedseed-seedtemplate.yaml": error validating data: [ValidationError(ManagedSeed.spec.seedTemplate.spec.provider): missing required field "type" in com.github.gardener.gardener.pkg.apis.core.v1beta1.SeedProvider, ValidationError(ManagedSeed.spec.seedTemplate.spec.provider): missing required field "region" in com.github.gardener.gardener.pkg.apis.core.v1beta1.SeedProvider]; if you choose to ignore these errors, turn validation off with --validate=false
```

because the `provider.{type,region}` are not marked as optional.

I think, it would be nice to make them optional in the OpenAPI spec (by adding the `+optional` tag) to allow such `ManagedSeed` manifests and leverage all the defaulting by this admission plugin.
But, it doesn't need to be done in this PR. Just noting it down.

plugin/pkg/managedseed/validator/admission.go
stoyanr: I removed `omitempty`, it doesn't work as you suggest it would. With a resource that has `backup: {}` in `seedTemplate` I still get the same error as before:
```
$ k apply -f dev/managedseed-i500152-gcpx-ng.yaml
error: error validating "dev/managedseed-i500152-gcpx-ng.yaml": error validating data: [ValidationError(ManagedSeed.spec.seedTemplate.spec.backup): missing required field "provider" in com.github.gardener.gardener.pkg.apis.core.v1beta1.SeedBackup, ValidationError(ManagedSeed.spec.seedTemplate.spec.backup): missing required field "secretRef" in com.github.gardener.gardener.pkg.apis.core.v1beta1.SeedBackup]; if you choose to ignore these errors, turn validation off with --validate=false
```

In fact, I was also wrong, what you need for an empty backup in `seedTemplate` is:
```yaml
backup:
  provider: ""
  secretRef: {}
```

So you must specify these empty values, but they are overwritten by the admission / defaults mechanism to appropriate ones.

Why do you think `omitempty` should play any role? It only controls how values are serialized to JSON. How could it influence such things as you suggest?

As far as I can tell, for `backup: {}` to work we would need to make both `provider` and `secretRef` inside `backup` non-mandatory. Not sure if this is possible, but even if it is, the value of being able to specify just `backup: {}` instead of the above seems too small to warrant an incompatible API change.

timebertt (commented):
Played around with it locally and I really like it!
Everything looks good to me now, only have one new remark.

plugin/pkg/managedseed/validator/admission.go
timebertt: Ok, let's keep it like this for now, as it seems to be a bit more complicated.
The main usecase with the gardenletConfig works...

plugin/pkg/managedseed/validator/admission.go
stoyanr: You could still specify:

```yaml
spec:
  seedTemplate:
    spec:
      provider:
        type: ""
        region: ""
```

Are you sure you want to change the `Seed` spec just to avoid writing the above? There was a reason provider type and region were not initially optional, and I am not sure we want to change this at this point.

And BTW, even if we do it for `provider`, we still have the same issue with `dns`, and `networks`. The full "empty" `seedTemplate` looks like this:

```yaml
spec:
  seedTemplate:
    spec:
      dns:
        ingressDomain: ""
      networks:
        pods: ""
        services: ""
      provider:
        type: ""
        region: ""
```

All of the above will be correctly defaulted, but they have to be specified as empty values because they are mandatory in the API. Making `provider: {}` possible is on this background only a minor improvement. Making `seedTemplate: {}` possible would be the real improvement, but it would require making all fields above non-mandatory. Honestly, I wouldn't do this just for this reason. Even if there is a strong vote to change the `Seed` API in this direction, let's please do this in a separate PR, not in this one.

timebertt: > Even if there is a strong vote to change the `Seed` API in this direction, let's please do this in a separate PR, not in this one.

Yes, sure. Also written this in my comment above.

> Are you sure you want to change the `Seed` spec just to avoid writing the above? There was a reason provider type and region were not initially optional, and I am not sure we want to change this at this point.

Well, the implications of adding an `+optional` tag just makes the field optional in the sense of OpenAPI spec. But it can still be forbidden by validation when not set under circumstances (e.g. in a Seed, but in a SeedTemplate OTOH and might be optional).
Anyways, let's continue with this PR. This is only a minor remark in UX, but can still be improved later on.

timebertt (approved):
/lgtm
/squash

PR #3422, opened 2021-01-15
preserve unknown fields for status.state field of extension types.

MartinWeindel:

/area control-plane
/kind bug
/priority normal

The CRDs for the extension types (backupbucket, backupentry, containerruntime, controlplane, extension, infrastructure, network, operatingsystemconfig, and worker) have been fixed.
If the CRD spec field `preserveUnknownFields: true` is not set anymore, the `status.state` field can not store anything anymore.
Therefore the validation openAPI specification now has a flag `x-kubernetes-preserve-unknown-fields: true` for the status state field.

Additionally the kubebuilder annotations have been updated accordingly and completed with the additional printing columns, status subresource, and short names from the CRDs to generate more complete CRDs with `make generate-extensions-crds`.

Note that  on the canary landscape the spec field `preserveUnknownFields: true` is still set, but not in the dev landscape.

Fixes #

**Release note**:

```bugfix operator
fix CRD for extension types to allow storing anything in `status.state`.
```


timuthy:
For me it seems like we should adjust CRDs for other extensions as well, shouldn't we?

They all have `type: object` w/o further specifying any fields, i.e. `x-kubernetes-preserve-unknown-fields` must be `true`. WDYT @rfranzke ?


```
state:
              description: State can be filled by the operating controller with what ever data it needs.
              type: object
```

rfranzke:
Yes, I guess this makes sense @timuthy.

@MartinWeindel there is also https://github.com/gardener/gardener/blob/master/Makefile#L199-L201 - can you make it generating these `x-kubernetes-preserve-unknown-fields` fields as well?

MartinWeindel:
/title preserve unknown fields for status.state field of extension types.

rfranzke (approved):
/lgtm
Thanks @MartinWeindel, great improvement!

PR #3423, opened 2021-01-15
Enhahance Terraformer.IsStateEmpty to check for presence of terraformer finalizer

vpnachev:

/kind bug
/priority normal

After https://github.com/gardener/terraformer/pull/65 the terraform finalizers are removed at the end of a successful `destroy`. However, the finalizer deletion can be rejected for some reason by the API and the whole operation needs to executed again. At this point, most of the extensions skips to run a `destroy` pod because of empty state, ref: https://github.com/gardener/gardener-extension-provider-aws/blob/d4c1eaf8590296add7a4bca19903b6eee3cd293b/pkg/controller/infrastructure/actuator_delete.go#L62-L65, hence there is no  controller to remove the finalizer. 

_Update:_
Now the `IsStateEmpty` check is enhanced to also verify for presence of the terraformer finalizer.

Fixes #

Thanks @ialidzhikov for spotting it.

**Release note**:

```bugfix dependency
A bug in the extension library that was preventing the deletion of TF secret and configmaps with empty state is now fixed. 
```

### DIFF
--- a/extensions/pkg/terraformer/state.go
+++ b/extensions/pkg/terraformer/state.go
@@ -20,12 +20,18 @@ import (
 	"fmt"
 
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	// TerraformerFinalizer is the finalizer key set by the terraformer on the configmaps and secrets
+	TerraformerFinalizer = "gardener.cloud/terraformer"
 )
 
 type terraformStateV3 struct {
@@ -91,8 +97,25 @@ func (t *terraformer) GetStateOutputVariables(ctx context.Context, variables ...
 	return output, nil
 }
 
-// IsStateEmpty returns true if the Terraform state is empty, and false otherwise.
+// IsStateEmpty returns true if the Terraform state is empty and the terraformer finalizer
+// is not present on any of the used configmaps and secrets. Otherwise, it returns false.
 func (t *terraformer) IsStateEmpty(ctx context.Context) bool {
+	for _, obj := range []client.Object{
+		&corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.configName}},
+		&corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.stateName}},
+		&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}},
+	} {
+		resourceName := obj.GetName()
+		if err := t.client.Get(ctx, kutil.Key(t.namespace, resourceName), obj); client.IgnoreNotFound(err) != nil {
+			t.logger.Error(err, "failed to get resource", "name", resourceName)
+			return false
+		}
+
+		if controllerutil.ContainsFinalizer(obj, TerraformerFinalizer) {
+			return false
+		}
+	}
+
 	state, err := t.GetState(ctx)
 	if err != nil {
 		return apierrors.IsNotFound(err)
--- a/extensions/pkg/terraformer/state_test.go
+++ b/extensions/pkg/terraformer/state_test.go
@@ -1,0 +1,133 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package terraformer_test
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	logzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	. "github.com/gardener/gardener/extensions/pkg/terraformer"
+)
+
+var _ = Describe("terraformer", func() {
+	var (
+		c   client.Client
+		ctx context.Context
+		log logr.Logger
+	)
+
+	BeforeEach(func() {
+		ctx = context.Background()
+		log = logzap.New(logzap.WriteTo(GinkgoWriter))
+	})
+
+	Describe("#IsStateEmpty", func() {
+		var (
+			terraformer        Terraformer
+			state              *corev1.ConfigMap
+			config             *corev1.ConfigMap
+			variable           *corev1.Secret
+			emptyInfraState    = ""
+			nonEmptyInfraState = "Some non-empty infra state"
+		)
+		const (
+			purpose                = "purpose"
+			image                  = "image"
+			stateName              = name + "." + purpose + StateSuffix
+			configName             = name + "." + purpose + ConfigSuffix
+			variableName           = name + "." + purpose + VariablesSuffix
+			expectingEmptyState    = true
+			expectingNonEmptyState = false
+		)
+
+		BeforeEach(func() {
+			state = &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: stateName}}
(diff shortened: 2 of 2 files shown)


extensions/pkg/terraformer/config.go
```diff
@@ -245,17 +246,29 @@ func (t *terraformer) CleanupConfiguration(ctx context.Context) error {
 	t.logger.Info("Cleaning up all terraformer configuration")
 
 	t.logger.V(1).Info("Deleting Terraform variables Secret", "name", t.variablesName)
-	if err := t.client.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}); err != nil && !apierrors.IsNotFound(err) {
+	varObject := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}
+	if err := extensionscontroller.DeleteAllFinalizers(ctx, t.client, varObject); err != nil {
```
ialidzhikov: Shouldn't we remove only the `gardener.cloud/terraformer` finalizer?

extensions/pkg/terraformer/config.go
```diff
@@ -245,17 +246,29 @@ func (t *terraformer) CleanupConfiguration(ctx context.Context) error {
...
-	if err := t.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.configName}}); err != nil && !apierrors.IsNotFound(err) {
+	configObject := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.configName}}
+	if err := extensionscontroller.DeleteAllFinalizers(ctx, t.client, configObject); err != nil {
+		return err
+	}
+	if err := t.client.Delete(ctx, configObject); err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
 
+	stateObject := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.stateName}}
```
ialidzhikov: Nit: Can you exchange L266 with L267 to keep it similar with the above invocations?

extensions/pkg/terraformer/config.go
vpnachev: Good catch.

ialidzhikov (dismissed):
/lgtm

extensions/pkg/terraformer/config.go
```diff
@@ -245,17 +248,29 @@ func (t *terraformer) CleanupConfiguration(ctx context.Context) error {
 	t.logger.Info("Cleaning up all terraformer configuration")
 
 	t.logger.V(1).Info("Deleting Terraform variables Secret", "name", t.variablesName)
-	if err := t.client.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}); err != nil && !apierrors.IsNotFound(err) {
+	varObject := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}
+	if err := extensionscontroller.DeleteFinalizer(ctx, t.client, varObject, terraformerFinalizer); err != nil {
+		return err
+	}
+	if err := t.client.Delete(ctx, varObject); err != nil && !apierrors.IsNotFound(err) {
```
rfranzke:
```suggestion
	if err := t.client.Delete(ctx, varObject); client.IgnoreNotFound(err) != nil {
```

same for below

rfranzke (commented):
Wouldn't a better approach be to make the extensions not skip the destruction in case the state is already empty and rather let the Terraformer remove the finalizers that it previously added?

vpnachev:
> Wouldn't a better approach be to make the extensions not skip the destruction in case the state is already empty and rather let the Terraformer remove the finalizers that it previously added?

I believe It was introduced on purpose, for example [provider-aws](https://github.com/gardener/gardener-extension-provider-aws/pull/44/files#diff-6b24a57d44f2d6fcbeb1f6d0081c2f267acc12cc6aca6b1b5ed9672b32239c66R41-R48). Now the terraformer finalizer has broken this shortcut, and from my point of view the correct solution was to enhance the CleanupConfiguration.

rfranzke:
The Terraformer library is not dealing with the finalizer yet and should not start doing so, I believe. It should rather properly instrument the Terraformer - now that we have this new behaviour.
It might work at the moment as you propose, but it has some potential to hit us later again.
Instead of existing early in the extension it should probably only do the LB cleanup if the state is not empty.
WDYT?

danielfoehrKn:
I do not know all the details here, but why would the extensions remove a finalized from the terraformer? Wouldnt it be better to fix the terraformer to ensure the finalizer is removed from all the secrets (like @rfranzke suggests)? The extension would 
In case the terraformer fails to remove the secret and the pod is terminated, the terraformer pod should be re-spawned to try again, no?

>  Instead of existing early in the extension it should probably only do the LB cleanup if the state is not empty.

Is this to make sure that the terraformer pod is re-created and removes the finalized? How is this related to the LB cleanup?

vpnachev:
I have removed the finalizer clean up and added a check in the `IsStateEmpty` for the terraformer finalizer presence on any of the resources.

extensions/pkg/terraformer/state.go
```diff
@@ -93,11 +100,32 @@ func (t *terraformer) GetStateOutputVariables(ctx context.Context, variables ...
 
 // IsStateEmpty returns true if the Terraform state is empty, and false otherwise.
```
rfranzke: Can you explain your newly added behaviour in the doc string?

extensions/pkg/terraformer/state.go
```diff
@@ -93,11 +100,32 @@ func (t *terraformer) GetStateOutputVariables(ctx context.Context, variables ...
...
+			t.logger.Error(err, "failed to get resource", "name", resourceName)
+			return false
+		}
+
+		resourceHasFinalizer, err := extensionscontroller.HasFinalizer(obj, terraformerFinalizer)
+		if err != nil {
+			t.logger.Error(err, "failed to check finalizer presence for resource", "name", resourceName)
+			return false
+		}
+		needToFinalize = needToFinalize || resourceHasFinalizer
```
rfranzke:
```suggestion
		if resourceHasFinalizer {
			needToFinalize = true
			break
		}
```

rfranzke (commented):
/invite @timebertt

vpnachev:
I have updated the PR and added a small unit test.

rfranzke (dismissed):
/lgtm
/needs second-opinion

extensions/pkg/terraformer/state.go
```diff
@@ -91,13 +98,38 @@ func (t *terraformer) GetStateOutputVariables(ctx context.Context, variables ...
...
+		}
+
+		resourceHasFinalizer, err := extensionscontroller.HasFinalizer(obj, TerraformerFinalizer)
+		if err != nil {
+			t.logger.Error(err, "failed to check finalizer presence for resource", "name", resourceName)
+			return false
+		}
+		if resourceHasFinalizer {
+			needToFinalize = true
+			break
```
timebertt: What about directly return `false` here and skipping the rest?

timebertt (commented):
Nit only, rest LGTM

extensions/pkg/terraformer/state.go
```diff
@@ -91,13 +98,38 @@ func (t *terraformer) GetStateOutputVariables(ctx context.Context, variables ...
 	return output, nil
 }
 
-// IsStateEmpty returns true if the Terraform state is empty, and false otherwise.
+// IsStateEmpty returns true if the Terraform state is empty and the terraformer finalizer
+// is not present on any of the used configmpas and secrets. Otherwise, it returns false.
```
ialidzhikov:
```suggestion
// is not present on any of the used configmaps and secrets. Otherwise, it returns false.
```

extensions/pkg/terraformer/state.go
vpnachev: Done.

rfranzke (dismissed):

extensions/pkg/terraformer/state.go
```diff
@@ -91,8 +98,30 @@ func (t *terraformer) GetStateOutputVariables(ctx context.Context, variables ...
...
+		&corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.stateName}},
+		&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}},
+	} {
+		resourceName := obj.GetName()
+		if err := t.client.Get(ctx, kutil.Key(t.namespace, resourceName), obj); client.IgnoreNotFound(err) != nil {
+			t.logger.Error(err, "failed to get resource", "name", resourceName)
+			return false
+		}
+
+		resourceHasFinalizer, err := extensionscontroller.HasFinalizer(obj, TerraformerFinalizer)
```
ialidzhikov: I see that currently we have multiple HasFinalizer funcs. One of them is deprecated in favour of `sigs.k8s.io/controller-runtime/pkg/controller/controllerutil.ContainsFinalizer` after the controller-runtime@v0.7 revendoring. Should we use directly `controllerutil.ContainsFinalizer`?

/cc @timebertt 

```suggestion
		resourceHasFinalizer := controllerutil.ContainsFinalizer(obj, TerraformerFinalizer)
```

timebertt: Also noticed this. We can do it directly for this case, but I anyway have a general cleanup of this on my list as part of https://github.com/gardener/gardener/issues/3109.

timebertt (dismissed):
Thanks.
/lgtm

extensions/pkg/terraformer/state.go
vpnachev: I am having hard times when I have to chose between the multiple implementations of the same util functions. 
So I am open to this suggestion and have applied it.

extensions/pkg/terraformer/state.go
```diff
@@ -19,7 +19,9 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/gardener/gardener/pkg/controllerutils"
```
ialidzhikov: `github.com/gardener/gardener/pkg/controllerutils.HasFinalizer` is already deprecated in favour of `sigs.k8s.io/controller-runtime/pkg/controller/controllerutil.ContainsFinalizer`

```suggestion
	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
```

vpnachev: Done

ialidzhikov (approved):
/lgtm

timebertt (approved):

PR #3427, opened 2021-01-19
replace image vector component descriptor creation with component-cli

schrodit:

/area delivery
/kind task
/priority normal

The images defines in the `images.yaml` are now added to the component descriptor using the component-cli.
The new cli command understands the semantics of the images.yaml and adds the needed information to the componentdescriptor v2.

The new script basically does the same thing as the old bash script so all current automation will stay in place.
In addition, it adds new labels that are later used generate the image overwrites without the need of the gardener images.yaml during deployment.

For detailed information how the resources are translated see https://github.com/gardener/component-cli/blob/main/docs/reference/components-cli_image-vector_add.md .<br>
For the detailed implementation see https://github.com/gardener/component-cli/blob/main/pkg/imagevector/imagevector.go#L29

This is the first step towards testing release candidates.

cc @AndreasBurger @ccwienk 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/hack/.ci/component_descriptor
+++ b/hack/.ci/component_descriptor
@@ -8,20 +8,12 @@ descriptor_out_file="${COMPONENT_DESCRIPTOR_PATH}"
 
 echo "enriching creating component descriptor from ${BASE_DEFINITION_PATH}"
 
-if [[ -f "$repo_root_dir/charts/images.yaml" ]]; then
-  images="$(yaml2json < "$repo_root_dir/charts/images.yaml")"
-  eval "$(jq -r ".images |
-    map(select(.sourceRepository != \"$repo_name\") |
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/needs second-opinion

rfranzke:
FYI: There is more computation for the component descriptor in the following lines, I guess you are aware. What's the reason this was not adapted?

schrodit:
> FYI: There is more computation for the component descriptor in the following lines, I guess you are aware. What's the reason this was not adapted?

It was currently not adopted because the overwrite is not correctly used: see this issue https://github.com/gardener/gardener-extension-provider-aws/issues/234
Therefore we wanted to first see how we can fix it and then adapt the script.

ccwienk (approved):

ccwienk:
/unhold
after some clarifications, lgtm

I would however like to improve handling of "sourceRepository" attr (parse for repo-url, add rel-path, optionally add path to dockerfile)

PR #3429, opened 2021-01-19
Fix gardener-seed-admission controller, etcd backup-restore and extension parsers time format

vlvasilev:

/area logging
/kind enhancement
/priority normal

With this PR we change the time format of the gsac, etcd backup-restore and extension parsers.

Fixes #

**Release note**:

```bugfix operator
Fix gardener-seed-admission controller, etcd backup-restore and extension parsers time format
```

### DIFF
--- a/charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-configmap.yaml
+++ b/charts/seed-bootstrap/charts/fluent-bit/templates/fluent-bit-configmap.yaml
@@ -301,7 +301,7 @@ data:
         Name        extensionsParser
         Format      json
         Time_Key    ts
-        Time_Format %Y-%m-%dT%H:%M:%S.%Z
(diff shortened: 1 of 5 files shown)


Kristian-ZH:
Why the lower case `z` works on one place and does not work on another place?

vlvasilev:
> Why the lower case `z` works on one place and does not work on another place?

I am not so sure but `%Z   for the timezone name or abbreviation.` is when you have the exact timezone specified.
In our case, we have the only Z which means [Zulu time sone](https://www.timeanddate.com/time/zones/z) equal to the offset of 0 from the UTC  time zone. And `%z the hour and minute offset from UTC` parse only offset from UTC.

Kristian-ZH (approved):
/lgtm

rfranzke (approved):

PR #3434, opened 2021-01-20
Node exporter changes

wyb1:

/area monitoring
/kind enhancement
/priority normal

Node exporter collectors are now all disabled by default unless explicitly needed.

This PR changes the default requests and limits of the node-exporter and makes the node exporter of class `Guaranteed`. This should reduce problems with CPU throttling of the node-exporter. If there is an active VPA the node-exporter will no longer have very high limits, especially for memory. This is helpful because the node-exporter does not need so much memory (it needs more CPU) and will allow for better scheduling.

Fixes #

**Release note**:

```other operator
NONE
```


rfranzke (approved):
/lgtm

PR #3440, opened 2021-01-21
Adapt "problematic webhook timeout" check 

rfranzke:

/area usability
/kind enhancement
/priority normal

This PR is a builds on top of #3413 and adapts the behaviour as follows:

~1. "Problematic timeout check" is also performed for `Fail` failure policy (earlier, it was only done for `Ignore`)~
2. Extensions library registers webhooks (both seed and shoot) with a timeout of `10s`
3. For backwards-compatibility, we skip the entire "problematic webhook" check for shooted seeds to have enough time for extensions to revendor this change (and get 2.)

/assign @ialidzhikov @timebertt @timuthy 
/squash

See https://github.com/gardener/gardener/pull/3413#issuecomment-763974466 for more details - thanks @ialidzhikov for bringing this to attention.

**Release note**:

```other dependency
The extensions library is now registering webhooks for both seeds and shoots with a `10s` timeout.
```

### DIFF
--- a/pkg/operation/care/constraints.go
+++ b/pkg/operation/care/constraints.go
@@ -17,7 +17,9 @@ package care
 import (
 	"context"
 	"fmt"
+	"strings"
 
+	"github.com/gardener/gardener/extensions/pkg/webhook"
 	"github.com/gardener/gardener/pkg/operation"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -139,12 +141,33 @@ func (c *Constraint) constraintsChecks(
 // CheckForProblematicWebhooks checks the Shoot for problematic webhooks which could prevent shoot worker nodes from
 // joining the cluster.
 func (c *Constraint) CheckForProblematicWebhooks(ctx context.Context) (gardencorev1beta1.ConditionStatus, string, string, error) {
+	// Earlier, the Gardener extensions library was not setting the timeout for auto-registered webhooks, hence, they
+	// ended up with the default (30s) which we consider problematic in shoots. For backwards-compatibility, we skip the
+	// check for shooted seeds to prevent it from failures caused by extension webhooks having a too high timeout value.
+	// See for more details: https://github.com/gardener/gardener/pull/3413#issuecomment-763974466
+	// TODO: Remove this in a future version.
+	if shootedSeed, err := gardencorev1beta1helper.ReadShootedSeed(c.shoot.Info); err == nil && shootedSeed != nil {
+		return gardencorev1beta1.ConditionTrue,
+			"NoProblematicWebhooks",
+			"Check is not executed for shooted seeds.",
+			nil
+	}
+
 	validatingWebhookConfigs := &admissionregistrationv1beta1.ValidatingWebhookConfigurationList{}
 	if err := c.shootClient.List(ctx, validatingWebhookConfigs); err != nil {
 		return "", "", "", fmt.Errorf("could not get ValidatingWebhookConfigurations of Shoot cluster to check for problematic webhooks")
 	}
 
 	for _, webhookConfig := range validatingWebhookConfigs.Items {
+		// Earlier, the Gardener extensions library was not setting the timeout for auto-registered webhooks, hence, they
+		// ended up with the default (30s) which we consider problematic in shoots. For backwards-compatibility, we skip the
+		// check for shooted seeds to prevent it from failures caused by extension webhooks having a too high timeout value.
+		// See for more details: https://github.com/gardener/gardener/pull/3413#issuecomment-763974466
+		// TODO: Remove this in a future version.
+		if strings.HasPrefix(webhookConfig.Name, webhook.NamePrefix) && strings.HasSuffix(webhookConfig.Name, webhook.NameSuffixShoot) {
+			continue
+		}
+
 		for _, w := range webhookConfig.Webhooks {
 			if IsProblematicWebhook(w.FailurePolicy, w.ObjectSelector, w.NamespaceSelector, w.Rules, w.TimeoutSeconds) {
 				return gardencorev1beta1.ConditionFalse,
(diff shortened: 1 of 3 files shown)


rfranzke:
> 1. "Problematic timeout check" is also performed for `Fail` failure policy (earlier, it was only done for `Ignore`)

I stumbled across this when actually working on 2. and 3. - @timebertt is there something I'm missing, i.e., why was the check only performed for `Ignore` policy earlier?

pkg/operation/care/constraints.go
```diff
@@ -139,6 +139,18 @@ func (c *Constraint) constraintsChecks(
 // CheckForProblematicWebhooks checks the Shoot for problematic webhooks which could prevent shoot worker nodes from
 // joining the cluster.
 func (c *Constraint) CheckForProblematicWebhooks(ctx context.Context) (gardencorev1beta1.ConditionStatus, string, string, error) {
+	// Earlier, the Gardener extensions library was not setting the timeout for auto-registered webhooks, hence, they
+	// ended up with the default (30s) which we consider problematic in shoots. For backwards-compatibility, we skip the
+	// check for shooted seeds to prevent it from failures caused by extension webhooks having a too high timeout value.
+	// See for more details: https://github.com/gardener/gardener/pull/3413#issuecomment-763974466
+	// TODO: Remove this in a future version.
+	if shootedSeed, err := gardencorev1beta1helper.ReadShootedSeed(c.shoot.Info); err == nil && shootedSeed != nil {
```
ialidzhikov, line 149: Hm, I am confused here. Isn't `gardener-extension-provider-aws-shoot` installed in all Shoots? How the check for ShootedSeed helps here? I see this webhook created for all Shoots.

ialidzhikov:
```
error: shoots.core.gardener.cloud "aws-local" could not be patched: shoots.core.gardener.cloud "aws-local" is forbidden: MutatingWebhookConfiguration "gardener-extension-provider-aws-shoot" is problematic: webhook "shoot.aws.extensions.gardener.cloud" with failurePolicy "Ignore" and 30s timeout might prevent worker nodes from properly joining the shoot cluster
```

rfranzke: Hm yeah, you're right, of course, I missed that. Shall we then simply skip the check for webhooks with the `gardener-extension-*-shoot` pattern? Not nice, but at least backwards-compatible and easy.

ialidzhikov: Can't we set the timeout to `30s` for several releases to give time to extensions to adapt and after several releases we can set it back to 15s?

rfranzke: We can, of course. But we can also simply add the name check and still keep @timebertt's improvement for end-user deployed webhooks, i.e., we get improved ops experience right with the next Gardener release.
My personal opinion is that setting it to `30s` and waiting a few releases would better if the "workaround" would be complex, but it isn't complex and shouldn't harm in any way.

ialidzhikov: Okay, I am fine with both approaches then.

pkg/operation/care/constraints.go
```diff
@@ -207,21 +219,23 @@ func IsProblematicWebhook(
 	rules []admissionregistrationv1beta1.RuleWithOperations,
 	timeoutSeconds *int32,
 ) bool {
+	if timeoutSeconds != nil && *timeoutSeconds > WebhookMaximumTimeoutSecondsNotProblematic {
```
timebertt: This change will mark all webhooks with a high timeout as problematic and block hibernation, even if the webhooks are not even acting on critical resources / on the `kube-system` namespace.

pkg/operation/care/constraints_test.go
```diff
@@ -159,6 +159,7 @@ var _ = Describe("#IsProblematicWebhook", func() {
 					Expect(care.IsProblematicWebhook(testCase.build())).To(BeFalse(), "expected webhook not to be problematic")
 				},
 				append([]TableEntry{
+					Entry("failurePolicy 'Ignore' and nil timeoutSeconds", webhookTestCase{failurePolicy: &failurePolicyIgnore, operationType: &operationCreate}),
```
timebertt: The combination of `Ignore` and unset `timeoutSeconds` should be marked as problematic, as `timeoutSeconds` was defaulted to `30` in `v1beta1` (as noted in the comment).

timebertt (changes requested):
I'm a bit confused, TBH.


> > 1. "Problematic timeout check" is also performed for `Fail` failure policy (earlier, it was only done for `Ignore`)
> 
> I stumbled across this when actually working on 2. and 3. - @timebertt is there something I'm missing, i.e., why was the check only performed for `Ignore` policy earlier?

Hmm, the current logic with #3413 should be as follows:
- if a webhook has `failurePolicy=Fail`, check if it acts on problematic resources and mark it as problematics if so
- if a webhook has `failurePolicy=Ignore` and has a low timeout, it is not considered problematic, no matter what resources it acts on
- if a webhook has `failurePolicy=Ignore` and also has a high timeout, check if it acts on problematic resources and mark it as problematics if so (because it can still block wakeups, as explained in the comment)

> 2\. Extensions library registers webhooks (both seed and shoot) with a timeout of `10s`

That part makes perfect sense.

> 3\. For backwards-compatibility, we skip the entire "problematic webhook" check for shooted seeds to have enough time for extensions to revendor this change (and get 2.)

As @ialidzhikov already covered my confusion about this part.

pkg/operation/care/constraints.go
timebertt: I'm also fine with both approaches but slightly prefer the name check and exclusion of extension shoot webhooks to directly benefit from the enhanced checks with the next release.

rfranzke:
OK, I've dropped this commit for now @timebertt, I wasn't sure myself either. Maybe we can discuss it later in a quick chat.
For now, let's focus on the current problems we're seeing.

timebertt (approved):
Thanks. This looks better now :)
/lgtm

ialidzhikov (approved):
/lgtm

PR #3443, opened 2021-01-21
[release-v1.15] Calculate ingressDomain for ControllerInstallation deployment

timuthy:

/area robustness
/kind enhancement
/priority normal

#3131 implies an API change of a seed's ingress domain from `.spec.dns.ingressDomain` -> `.spec.ingress.domain`. For compatibility reasons we should consider this change when passing the ingress domain as a value to the chart of a `ControllerRegistration`.

/cc @rfranzke @BeckerMax

**Release note**:

```other operator
Gardener now considers the `seed.spec.ingress.domain` field when passing the value via `gradener.seed.ingressDomain` to `ControllerRegistration` charts.
```


vpnachev (approved):
/lgtm

PR #3444, opened 2021-01-21
Normalize versions when comparing against constraints

rfranzke:

/area usability
/kind bug
/priority normal

Currently, when run for example in a `GKE` cluster, the gardenlet fails bootstrapping the seed with the following error message:

```
time="2021-01-21T08:24:12Z" level=error msg="seed is not yet ready: condition \"Bootstrapped\" has invalid status False (expected True) due to BootstrappingFailed: could not find image \"kube-scheduler\" opts target version 1.16.15-gke.6000" operation=reconcile shoot=garden-foo/bar
```

As we are not interested in evaluating pre-release suffixes when we compare versions, we now normalize it (which removes the pre-release suffix) before comparing it against a constraint.

/cc @ialidzhikov @mvladev 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/operation/seed/seed.go
+++ b/pkg/operation/seed/seed.go
@@ -883,21 +883,18 @@ func bootstrapComponents(c kubernetes.Interface, namespace string, imageVector i
 	}
 	components = append(components, seedadmission.New(c.Client(), namespace, gsacImage.String(), kubernetesVersion))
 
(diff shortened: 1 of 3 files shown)


ialidzhikov (approved):
/lgtm

timebertt (approved):

PR #3445, opened 2021-01-21
Refactor gardener-resource-manager deployment

BeckerMax:

/kind enhancement
/priority normal
/topology seed
/platform all
/exp intermediate

**What this PR does / why we need it:**
This PR refactors the deployment for the gardener-resource manager and adopts the DeployWaiter interface(implicitly) for components. Consequently, it eliminates the Helm charts and adds full unit tests, instead. This improves several aspects and the overall code quality.
The grm is deployed in two configurations. A central one in the garden namespace and one in each shoot controlplane. This PR unifies the usage and provides one interface for both use cases.

Part of https://github.com/gardener/gardener/issues/2754

I on purpose did not declare the DeployWaiter interface on the producer side (in [resource-manager.go](https://github.com/gardener/gardener/compare/master...BeckerMax:refactor-resource-manager#diff-a66ff079fa067226cf9c172079a286e54f39e2a5ec6ad863136488759de35274R89)). I am aware that this differs to the other component implementations. As i do not need mocks I preferred to not declare the interface explicitly. But I am also open to include it if you think it adds value. Happy to hear your opinions.

```other operator
NONE
```


pkg/operation/botanist/controlplane/resourcemanager/logging.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

same for other new files, please

pkg/operation/botanist/resource_manager.go
```diff
@@ -0,0 +1,83 @@
...
+		b.K8sSeedClient.Client(),
+		b.Shoot.SeedNamespace,
+		image.String(),
+		b.Shoot.GetReplicas(1),
+		cfg,
+	), nil
+}
+
+// DeployGardenerResourceManager deploys the gardener-resource-manager
+func (b *Botanist) DeployGardenerResourceManager(ctx context.Context) error {
```
rfranzke, line 68: Can you write a unit test for this function, similar to https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/kubescheduler_test.go#L80-L122?

pkg/operation/seed/seed.go
```diff
@@ -839,7 +839,6 @@ func bootstrapComponents(c kubernetes.Interface, namespace string, imageVector i
 		return nil, err
 	}
 
-	// cluster-autoscaler
```
rfranzke, line 843: Why was this removed?

extensions/test/integration/operation/shoot_operation.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	"github.com/gardener/gardener/pkg/api/extensions"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	"github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```
rfranzke:
```suggestion
	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```

pkg/apis/core/v1beta1/constants/types_constants.go
```diff
@@ -239,6 +239,8 @@ const (
 	LabelScheduler = "scheduler"
 	// LabelExtensionProjectRole is a constant for a label value for extension project roles
 	LabelExtensionProjectRole = "extension-project-role"
+	// LabelResourceManager is a constant for a label value for the gardener-resource-manager.
+	LabelResourceManager = "gardener-resource-manager"
```
rfranzke: Why is this here? It should be better in the `resourcemanager` package, or?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	// SecretName is a constant for the secret name for the gardener resource manager's kubeconfig secret.
+	SecretName = "gardener-resource-manager"
+	// UserName is the name that should be used for the secret that the gardener resource manager uses to
+	// authenticate itself with the kube-apiserver (e.g., the common name in its client certificate).
+	UserName = "gardener-resource-manager"
```
rfranzke:
```suggestion
	UserName = "gardener.cloud:system:gardener-resource-manager"
```

I know, we spoke about it, but for now let's keep this old name as part of this refactoring PR. If needed, we can change it later/separately.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	return &ResourceManager{
+		client:    client,
+		image:     image,
+		namespace: namespace,
+		replicas:  replicas,
+		config:    config,
+	}
+}
+
+type ResourceManager struct {
```
rfranzke: doc string

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	HealthSyncPeriod           *string
+	KubeConfig                 *component.Secret
+	LeaseDuration              *string
+	MaxConcurrentHealthWorkers *int32
+	RenewDeadline              *string
+	ResourceClass              *string
+	RetryPeriod                *string
+	SyncPeriod                 *string
+	TargetDisableCache         *bool
+	WatchedNamespace           *string
```
rfranzke: Can you add some doc strings for these attributes?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	ConcurrentSyncs            *int32
+	ClusterRoleName            *string
+	DefaultLabels              *map[string]string
+	HealthSyncPeriod           *string
+	KubeConfig                 *component.Secret
+	LeaseDuration              *string
+	MaxConcurrentHealthWorkers *int32
+	RenewDeadline              *string
+	ResourceClass              *string
+	RetryPeriod                *string
```
rfranzke:
```suggestion
	RetryPeriod                *time.Duration
```

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	SyncPeriod                 *string
```
rfranzke:
```suggestion
	SyncPeriod                 *time.Duration
```

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+
+// Config holds the optional configuration options for the gardener resource manager
+type Config struct {
+	AlwaysUpdate               *bool
+	ConcurrentSyncs            *int32
+	ClusterRoleName            *string
+	DefaultLabels              *map[string]string
+	HealthSyncPeriod           *string
+	KubeConfig                 *component.Secret
+	LeaseDuration              *string
```
rfranzke:
```suggestion
	LeaseDuration              *time.Duration
```

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	config    Config
+}
+
+// Config holds the optional configuration options for the gardener resource manager
+type Config struct {
+	AlwaysUpdate               *bool
+	ConcurrentSyncs            *int32
+	ClusterRoleName            *string
+	DefaultLabels              *map[string]string
+	HealthSyncPeriod           *string
```
rfranzke:
```suggestion
	HealthSyncPeriod           *time.Duration
```

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	replicas  int32
+	config    Config
+}
+
+// Config holds the optional configuration options for the gardener resource manager
+type Config struct {
+	AlwaysUpdate               *bool
+	ConcurrentSyncs            *int32
+	ClusterRoleName            *string
+	DefaultLabels              *map[string]string
```
rfranzke:
```suggestion
	Labels              map[string]string
```

`map` is already a pointer type

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	HealthSyncPeriod           *string
+	KubeConfig                 *component.Secret
+	LeaseDuration              *string
+	MaxConcurrentHealthWorkers *int32
+	RenewDeadline              *string
```
rfranzke:
```suggestion
	RenewDeadline              *time.Duration
```

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+type ResourceManager struct {
+	client    client.Client
+	namespace string
+	image     string
+	replicas  int32
+	config    Config
+}
+
+// Config holds the optional configuration options for the gardener resource manager
+type Config struct {
```
rfranzke:
```suggestion
type Values struct {
```

that's how it's called in other packages

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+		return nil
+	})
+	return err
+}
+
+func (r *ResourceManager) emptyVPA() *autoscalingv1beta2.VerticalPodAutoscaler {
+	return &autoscalingv1beta2.VerticalPodAutoscaler{ObjectMeta: metav1.ObjectMeta{Name: "gardener-resource-manager", Namespace: r.namespace}}
+}
+
+func appLabel() *map[string]string {
```
rfranzke:
```suggestion
func appLabel() map[string]string {
```

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	return &map[string]string{
+		v1beta1constants.LabelApp: "gardener-resource-manager",
+	}
+}
+
+// SetKubeConfig enables the ResourceManager to deploy resources into a different
+// cluster than the one it is running in.
+func (r *ResourceManager) SetKubeConfig(k *component.Secret) { r.config.KubeConfig = k }
```
rfranzke: Can we call this `SetSecrets()` like in other components? Yes, today there is only the kubeconfig secret, but maybe this will change in the future. Though, main motivation here would be similarity to other components.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+
+func (r *ResourceManager) Wait(_ context.Context) error        { return nil }
```
rfranzke: doc strings

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+		clusterRole.Rules = policies
+		return nil
+	})
+	return err
+}
+
+func (r *ResourceManager) emptyClusterRole() (*rbacv1.ClusterRole, error) {
+	if r.config.ClusterRoleName == nil {
+		return nil, errors.New("creating Cluster Role failed - no name defined for the Cluster Role")
+	}
```
rfranzke: Can you remove the `ClusterRoleName` from the config and rather compute it here?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+			Name:      UserName,
+			Namespace: r.namespace,
+		}}
+		return nil
+	})
+	return err
+}
+
+func (r *ResourceManager) emptyClusterRoleBinding() (*rbacv1.ClusterRoleBinding, error) {
+	if r.config.ClusterRoleName == nil {
```
rfranzke: same here

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	_, err = controllerutil.CreateOrUpdate(ctx, r.client, clusterRoleBinding, func() error {
+		clusterRoleBinding.Labels = *r.config.DefaultLabels
+		clusterRoleBinding.RoleRef = rbacv1.RoleRef{
+			APIGroup: rbacv1.GroupName,
+			Kind:     "ClusterRole",
+			Name:     *r.config.ClusterRoleName,
+		}
+		clusterRoleBinding.Subjects = []rbacv1.Subject{{
+			Kind:      rbacv1.ServiceAccountKind,
+			Name:      UserName,
```
rfranzke: Given above proposed change of the `UserName`, this won't work anymore in the case that the source cluster = target cluster

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+		volumeName                = "gardener-resource-manager"
+	)
+
+	deployment := r.emptyDeployment()
+
+	_, err := controllerutil.CreateOrUpdate(ctx, r.client, deployment, func() error {
+		deployment.Labels = *r.config.DefaultLabels
+
+		deployment.Spec.Replicas = &r.replicas
+		deployment.Spec.RevisionHistoryLimit = pointer.Int32Ptr(0)
```
rfranzke:
```suggestion
		deployment.Spec.RevisionHistoryLimit = pointer.Int32Ptr(1)
```

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+func (r *ResourceManager) emptyService() *corev1.Service {
+	return &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: "gardener-resource-manager", Namespace: r.namespace}}
+}
+
+func (r *ResourceManager) ensureDeployment(ctx context.Context) error {
+	const (
+		limitCPU                  = "400m"
+		limitMemory               = "512Mi"
+		requestCPU                = "23m"
+		requestMemory             = "47Mi"
```
rfranzke: I guess we don't need these variables here

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+							FailureThreshold:    5,
+							PeriodSeconds:       10,
+							SuccessThreshold:    1,
+							TimeoutSeconds:      5,
+						},
+					},
+				},
+			},
+		}
+		if r.config.KubeConfig != nil {
```
rfranzke:
```suggestion

		if r.config.KubeConfig != nil {
```

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	_, err := controllerutil.CreateOrUpdate(ctx, r.client, deployment, func() error {
+		deployment.Labels = *r.config.DefaultLabels
+
+		deployment.Spec.Replicas = &r.replicas
+		deployment.Spec.RevisionHistoryLimit = pointer.Int32Ptr(0)
+		deployment.Spec.Selector = &metav1.LabelSelector{MatchLabels: *appLabel()}
+
+		deployment.Spec.Template = corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: utils.MergeStringMaps(*r.config.DefaultLabels, map[string]string{
```
rfranzke: Where is the app label?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	}
+	cmd = append(cmd, fmt.Sprintf("--leader-election-namespace=%s", r.namespace))
+	if r.config.RenewDeadline != nil {
+		cmd = append(cmd, fmt.Sprintf("--leader-election-renew-deadline=%s", *r.config.RenewDeadline))
+	}
+	if r.config.RetryPeriod != nil {
+		cmd = append(cmd, fmt.Sprintf("--leader-election-retry-period=%s", *r.config.RetryPeriod))
+	}
+	if r.config.ConcurrentSyncs != nil {
+		cmd = append(cmd, fmt.Sprintf("--max-concurrent-workers=%v", *r.config.ConcurrentSyncs))
```
rfranzke:
```suggestion
		cmd = append(cmd, fmt.Sprintf("--max-concurrent-workers=%d", *r.config.ConcurrentSyncs))
```

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+func (r *ResourceManager) printCommand(volumeMountPathKubeconfig string) []string {
+	cmd := []string{
+		"/gardener-resource-manager",
+	}
+	if r.config.AlwaysUpdate != nil {
+		cmd = append(cmd, fmt.Sprintf("--always-update=%v", *r.config.AlwaysUpdate))
+	}
+	cmd = append(cmd, fmt.Sprintf("--health-bind-address=:%v", healthPort))
+	if r.config.MaxConcurrentHealthWorkers != nil {
+		cmd = append(cmd, fmt.Sprintf("--health-max-concurrent-workers=%v", *r.config.MaxConcurrentHealthWorkers))
```
rfranzke:
```suggestion
		cmd = append(cmd, fmt.Sprintf("--health-max-concurrent-workers=%d", *r.config.MaxConcurrentHealthWorkers))
```

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+	if r.config.RenewDeadline != nil {
+		cmd = append(cmd, fmt.Sprintf("--leader-election-renew-deadline=%s", *r.config.RenewDeadline))
+	}
+	if r.config.RetryPeriod != nil {
+		cmd = append(cmd, fmt.Sprintf("--leader-election-retry-period=%s", *r.config.RetryPeriod))
+	}
+	if r.config.ConcurrentSyncs != nil {
+		cmd = append(cmd, fmt.Sprintf("--max-concurrent-workers=%v", *r.config.ConcurrentSyncs))
+	}
+	cmd = append(cmd, fmt.Sprintf("--metrics-bind-address=:%v", metricsPort))
```
rfranzke:
```suggestion
	cmd = append(cmd, fmt.Sprintf("--metrics-bind-address=:%d", metricsPort))
```

pkg/operation/seed/seed.go
```diff
@@ -861,6 +860,21 @@ func bootstrapComponents(c kubernetes.Interface, namespace string, imageVector i
 	}
 	components = append(components, etcd.NewBootstrapper(c.Client(), namespace, etcdImage, kubernetesVersion, etcdImageVectorOverwrite))
 
+	// gardener-resource-manager
```
rfranzke: Can you move this up, i.e,. GRM should be the very first component

rfranzke (changes requested):

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
vpnachev: Please, keep the name as it is before the PR. There are landscapes where these user names are used to define the audit policies. Just the leader election requests from the controller is generating too much unwanted audit events that might be already filtered out with a policy.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,553 @@
...
+				}),
+			},
+			Spec: corev1.PodSpec{
+				ServiceAccountName: serviceAccountName,
+				Containers: []corev1.Container{
+					{
+						Name:            containerName,
+						Image:           r.image,
+						ImagePullPolicy: corev1.PullIfNotPresent,
+						Command:         r.printCommand(volumeMountPathKubeconfig),
```
danielfoehrKn:
```suggestion
						Command:         r.printCommand(),
```

`printCommand` is only called once - we can just use the constant in the function?

rfranzke: +1, nice catch. Actually, `printCommand` is not the best name because nothing is printed. What about `computeCommand` like in other components?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: I am confused. I did not change the name in this PR as discussed. See old name: 
[charts/seed-bootstrap/templates/gardener-resource-manager/rbac.yaml](https://github.com/gardener/gardener/pull/3445/files#diff-756b94b9db03912f92e00c131c6e8632a4f8c2ed8f3a3b8f49445919c639ac2aL28)
Your `suggested change` includes a renaming.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: This would mean I would have to give it the name:
`gardener-resource-manager-seed` which I want to avoid as it ties the implementation to the current configuration of the central grm deployed in seeds.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: Yes, I assume that is why we do not change the UserName?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: DefaultLabels (now renamed to Labels) always includes the app label. This is ensured in the constructor : https://github.com/gardener/gardener/pull/3445/files#diff-a66ff079fa067226cf9c172079a286e54f39e2a5ec6ad863136488759de35274R90-R95

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: @danielfoehrKn can you elaborate again? the command is computed with several if clauses. How would you put that into a constant in the function?
I am very fine with renaming it to computeCommand.

pkg/apis/core/v1beta1/constants/types_constants.go
BeckerMax: I deleted it completely, I guess if something needs this label it can add it later.

BeckerMax:
@timuthy do you still plan to review the PR as you are assigned?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
rfranzke: @BeckerMax what you reference is the name of the service account. However, you use the `UserName` constant [here](https://github.com/gardener/gardener/pull/3445/files#diff-be4a114eed4ba1e885b1e46ba2489834b18b3e0471575f78948d40588ad8dceaL293) and this is clearly a change. Please revert it.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
rfranzke: Yes, I know and understand, I tried the same in other packages, but it didn't always work, and in the end, this whole package is part of g/g and you already use such terms (like "seed" or "shoot") in other places, e.g. https://github.com/gardener/gardener/pull/3445/files#diff-a66ff079fa067226cf9c172079a286e54f39e2a5ec6ad863136488759de35274R377.
Hence, it doesn't make it worse but rather easier if this `ClusterRoleName` disappears from the configuration. It's anyways somewhat hard to come up with a meaningful value here as a "user" of this package because it's an implementation detail (I would have to look up the code to check where this setting is used to come up with a "good" value).

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
rfranzke: OK, see comment above. I guess you are mixing up the service account name and the "user name" (which ends up in the `CommonName` field of a certificate).

pkg/operation/botanist/resource_manager.go
```diff
@@ -0,0 +1,88 @@
...
+func (b *Botanist) DefaultResourceManager() (*resourcemanager.ResourceManager, error) {
+	image, err := b.ImageVector.FindImage(common.GardenerResourceManagerImageName, imagevector.RuntimeVersion(b.SeedVersion()), imagevector.TargetVersion(b.ShootVersion()))
+	if err != nil {
+		return nil, err
+	}
+
+	oneMinute := time.Minute
+	fourtySeconds := time.Second * 40
+	fiveteenSeconds := time.Second * 15
+	tenSeconds := time.Second * 10
```
rfranzke: Can we get rid of these variables? They are not too helpful FMPOV, it's more straight-forward to write `15*time.Second` instead of `fiveteenSeconds`.

Maybe you want to have a helper function similar to `pointer.StringPtr()`.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
rfranzke: Hm, OK, I missed that, it's a little confusing. Can we remove the `Labels` attribute from the config values and rather compute it in the package itself, similar to how it's done in all other component packages?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
rfranzke: @danielfoehrKn questions why the (now renamed) `computeCommand()` function has the `volumeMountPathKubeconfig` parameter. You only call `computeCommand()` once and the value you pass for this parameter itself is a constant. Can you please move such constants out of the function and to the top of the package? This is a) how it's done in other packages, and b) allows you to simplify the signature of the `computeCommand()` function.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: I agree. Also made the tests easier. Thanks for your feedback!

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: Yes, that is true. Thanks for pointing this out.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,572 @@
...
+func (r *ResourceManager) ensureRBAC(ctx context.Context) error {
+	deployResourcesIntoAnotherCluster := r.values.KubeConfig != nil
+	if deployResourcesIntoAnotherCluster {
+		if r.values.WatchedNamespace == nil {
+			if err := r.ensureClusterRole(ctx, allowManagedResources); err != nil {
+				return err
+			}
+			if err := r.ensureClusterRoleBinding(ctx); err != nil {
+				return err
+			}
```
timuthy, line 201: I don't see this particular use case at the moment. Can you elaborate when this is needed?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,572 @@
...
+	return kutil.DeleteObjects(
+		ctx,
+		r.client,
+		objectsToDelete...,
+	)
+}
+
+func (r *ResourceManager) ensureRBAC(ctx context.Context) error {
+	deployResourcesIntoAnotherCluster := r.values.KubeConfig != nil
+	if deployResourcesIntoAnotherCluster {
```
timuthy:
```suggestion
	targetDiffersFromSourceCluster := r.values.KubeConfig != nil
	if targetDiffersFromSourceCluster {
```

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: OK, got it. That would mean I would parse if a target kubeconfig is configured and then in this case add the `v1beta1constants.GardenRoleControlPlane` label ? Something like this?
I feel this information should be passed via the constructor because the caller actually knows this and not the resource manager.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: I guess it is a matter of code style. I dislike global variabels/constants. The variable is used twice, once in ensureDeployment and once in computeCommand. My thought here is to limit the scope of the variable as much as possible so I declared it in the ensureDeployment function. I value that higher than having to pass a variable to the computeCommand function.
Would you still prefer to have it declared at the top? Then I will do so as I also see value in adhering to the existing code style.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: I also do not see a use case for it currently.
My motivation: The resource_manager package exposes the configuration `watchedNamespace` and `kubeconfig`. I thought how should it react if the kubeconfig is set but not the watchedNamespace.
An alternative solution would be to throw an error when calling deploy, like `If kubeconfig is set there always needs to be be a watchedNamespace configured`. Would you prefer that over the above implementation?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
rfranzke: Well, TBH: I don't care at all about this detail because the impact/value is very low/zero. I just mainly care about consistency and similarity to other "componentified" packages to make it easier to connect dots, especially for new or rare contributors. The impact/value of this is probably greater than zero.

With this argument, I'd personally prefer to do this like in other packages for now, i.e., have the constant at the top. It's still private to this package and not "global".

However, I won't insist on it because, as said, the impact is the "consequence" is very low. Also, it can still be improved (consistently for all packages) in the future if you see need, but probably our time is better invested elsewhere. :)

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
rfranzke: Well, no strong opinion. I can only refer to what I wrote in https://github.com/gardener/gardener/pull/3445#discussion_r566736098. With the "similarity" argument I'd prefer to remove it from here (because, as you can see, I was already confused because this package handles some things different than other packages), but it's a detail with low value, so I'm fine with whatever you decide.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: Ok. I think it still holds value to explain to you my thought process because I actually reasoned about this trade-off when developing and I wanted to hear your opinions.
I follow your point that consistency in the code base is more important than stylistic choices and that a refactoring can then be done separately in another PR touching all the different files/components.
Hope it was not too annoying to argue here and I will carry what what I learned to future PRs :).

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
timuthy: I see, thanks for explaining. I guess when you wanted to follow this approach consequently, you'd also need to support the 4th case i.e., `kubeConfig == nil &&  WatchedNamespace != nil`, which is not covered yet, right?

While I get your point, I also see the benefits to support solely the cases we need right now because when you read the code, it's not quite obvious which case is used where and less branches usually mean better readability. WDYT?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
rfranzke: No worries, many things are not obvious (especially not, if you are not very familiar with the code base or, particularly, those other packages), and I totally agree to your arguments. In general, it's a good guideline to not expose things too widely if not necessary. However, in those particular cases, it probably doesn't make much of a difference and has a low impact(, and probably a "larger value" when we strive for the consistency/similarity goal). Anyways, I'm still fine with whatever you prefer. And please continue to bring up similar thoughts in the future!

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: The case you point out above is covered. If `kubeconfig == nil` then it does not matter if `watchedNamespace` is set or not. The grm always needs a ClusterRole with full access to deploy into the whole cluster.

I agree with your observation that less code is always better and prematurely including code for other use cases which are not yet on the horizon is bad. I did a commit on another branch to see how it looks if I remove the above and add validation(test adaption missing). See https://github.com/gardener/gardener/commit/8ab692eef1f9807e3a69efd2c08d67e82b30b642

After looking at it I would personally leave it like it is right now as I do not see a major gain in readability with validation over the current implementation. For me the validation adds other questions, like why is this from the standpoint of the grm package a non supported usecase.
WDYT?

pkg/operation/botanist/resource_manager.go
BeckerMax: Done - though I think the test adds more of a burden than actual testing something meaningful.

BeckerMax:
@rfranzke I addressed all your remarks in the newest commit. Thanks for the comments so far and please take another look.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: I just realized I would need to validate in the deploy function because the kubeconfig can be set after the instance got created. This would need to be adapted in the above commit. Anyways, my point from above still stands.

timuthy: Okay, thanks for checking and explaining.

pkg/operation/botanist/resource_manager.go
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/operation/botanist/resource_manager.go
rfranzke: Maybe, yes. The function as of now is quite straight-forward. However, we have seen in the past that it might happen that such functions grow over time. If there was no initial test that can grow in parallel then it's likely that no tests are added at all or only very late (i.e., with additional effort/time consumption).

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,561 @@
...
+	})
+	return err
+}
+
+func (r *resourceManager) emptyRoleInWatchedNamespace() (*rbacv1.Role, error) {
+	if r.values.WatchedNamespace == nil {
+		return nil, errors.New("creating Role in watched namespace failed - no namespace defined")
+	}
+	return &rbacv1.Role{ObjectMeta: metav1.ObjectMeta{Name: roleName, Namespace: *r.values.WatchedNamespace}}, nil
+}
```
rfranzke, line 276: Hm, is it really necessary to return an error here? I don't see that this would a benefit, or? Can't we simply return `nil` and ease our life here?

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,561 @@
...
+	})
+	return err
+}
+
+func (r *resourceManager) emptyRoleBindingInWatchedNamespace() (*rbacv1.RoleBinding, error) {
+	if r.values.WatchedNamespace == nil {
+		return nil, errors.New("creating RoleBinding in watched namespace failed - no namespace defined")
+	}
+	return &rbacv1.RoleBinding{ObjectMeta: metav1.ObjectMeta{Name: "gardener-resource-manager", Namespace: *r.values.WatchedNamespace}}, nil
+}
```
rfranzke, line 305: same here?

rfranzke (commented):
Thanks @BeckerMax for addressing all the suggestions, it looks very good now!
I have a few minor remarks left, but generally I'm fine with how it is now. Thanks again!

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
BeckerMax: Mmh, isn't it closer to what you would expect in go to return an error instead of nil for the object? Even if I switch to nil I would still need to perform the checks when I call this function but I would have to check whether the returned role is `nil`. I feel we would not gain much from that.

BeckerMax: I see your point that this error feels kind of unnecessary but I am not sure how to improve it right now.

timuthy (commented):
No further comments from my side. From what I've seen, it looks good to me. Thanks for addressing the comments.

pkg/operation/botanist/controlplane/resourcemanager/resource_manager.go
rfranzke: Well, no strong opinion and no need to block anything here. Resolving the comment for now.

rfranzke (dismissed):
/lgtm

BeckerMax:
While testing I found that the labels on the deployment template were wrong for the central grm after the review corrections. Fixed with newest commit.
Now everything looks fine. Only intentional changes(for example livenessProbe) between the resources before/after refactoring.

rfranzke (approved):
/lgtm

PR #3377, opened 2021-01-12
Shoot care refactoring

timuthy:
**How to categorize this PR?**

/area quality
/kind cleanup
/priority normal

**What this PR does / why we need it**:
This PR submits a first batch of refactoring for shoot care operations:
- Extract health checks from `botanist` -> `/pkg/operation/care`
- Extract garbage collection from `botanist` -> `/pkg/operation/care`
- `Operation` is fed with shoot from `Cluster` object
- Adds and improves unit tests

**Special notes for your reviewer**:
Further refactoring is planned. Submitting these changes already should help to simplify reviews.

**Release note**:

```other operator
NONE
```

PR #3447, opened 2021-01-21
Prevent "lastTransitionTime" and "lastUpdateTime" of constraints to be updated continuously

ialidzhikov:
/kind bug

After #3377 the Shoot constraints  "lastTransitionTime" and "lastUpdateTime" are updated continuously no matter whether there is a change in the constraint or not. This is something we want to avoid.

The root cause for this regression seems to be that we try to lookup for the existing constraint not in `.status.constraints` but in `.status.conditions` which always returns a newly initialized constraint with updated  "lastTransitionTime" and "lastUpdateTime".

**Release note**:

```other operator
NONE
```


timebertt (approved):
/lgtm

timuthy (approved):
/lgtm
Thanks @ialidzhikov

PR #3457, opened 2021-01-26
Fix shoot conditions

timuthy:

/area quality
/kind bug
/priority normal

This PR fixes a bug in the shoot care controller where conditions were not merged correctly after executing health checks. Along the way I added unit tests for the `Care` function which was untested so far. Certain code dependencies needed to be isolated in order to inject dummy implementations during test execution.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_care_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_care_control.go
@@ -20,19 +20,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/gardener/gardener/pkg/client/kubernetes"
-	"github.com/gardener/gardener/pkg/operation/care"
-	"github.com/gardener/gardener/pkg/operation/common"
-	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/util/retry"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	gardencore "github.com/gardener/gardener/pkg/client/core/clientset/versioned"
 	gardencoreinformers "github.com/gardener/gardener/pkg/client/core/informers/externalversions/core/v1beta1"
+	"github.com/gardener/gardener/pkg/client/kubernetes"
 	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
 	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/keys"
 	"github.com/gardener/gardener/pkg/controllerutils"
@@ -40,11 +32,18 @@ import (
 	confighelper "github.com/gardener/gardener/pkg/gardenlet/apis/config/helper"
 	"github.com/gardener/gardener/pkg/logger"
 	"github.com/gardener/gardener/pkg/operation"
+	"github.com/gardener/gardener/pkg/operation/common"
 	shootpkg "github.com/gardener/gardener/pkg/operation/shoot"
 	"github.com/gardener/gardener/pkg/utils"
 	"github.com/gardener/gardener/pkg/utils/flow"
 	"github.com/gardener/gardener/pkg/utils/imagevector"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func (c *Controller) shootCareAdd(obj interface{}) {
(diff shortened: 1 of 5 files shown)


pkg/gardenlet/controller/shoot/shoot_care_control.go
```diff
@@ -143,10 +142,38 @@ type CareControlInterface interface {
...
 
+var (
+	// NewOperation is used to create a new `operation.Operation` instance.
+	NewOperation = defaultNewOperationFunc
+	// NewHealthCheck is used to create a new Health check instance.
+	NewHealthCheck = defaultNewHealthCheck
+	// NewConstraintCheck is used to create a new Constraint check instance.
+	NewConstraintCheck = defaultNewConstraintCheck
+	// NewGarbageCollector is used to create a new Constraint check instance.
+	NewGarbageCollector = defaultNewGarbageCollector
```
rfranzke, line 164: Are these `default...` variables needed? Can't you simply do something like this:

```suggestion
	// NewOperation is used to create a new `operation.Operation` instance.
	NewOperation = defaultNewOperationFunc
	// NewHealthCheck is used to create a new Health check instance.
	NewHealthCheck = care.NewHealth
	// NewConstraintCheck is used to create a new Constraint check instance.
	NewConstraintCheck = care.NewConstraint
	// NewGarbageCollector is used to create a new Constraint check instance.
	NewGarbageCollector = care.NewGarbageCollection
```

For the operation, I understand, here it's more complex. Did you decide to do it for the other functions then for similarity reasons? OK then... :)

pkg/gardenlet/controller/shoot/shoot_care_control.go
```diff
@@ -143,10 +142,38 @@ type CareControlInterface interface {
...
+)
+
 type defaultCareControl struct {
+	newOperation        NewOperationFunc
+	newHealthCheck      NewHealthCheckFunc
+	newConstraintCheck  NewConstraintCheckFunc
+	NewGarbageCollector NewGarbageCollectorFunc
+
```
rfranzke:
```suggestion
```

Is this on purpose or type that `NewGarbageCollector` starts with a `N` while all other start with `n`?
Generally, what's the reason that those are attributes on the `defaultCareControl` structure? You have exported variables for those functions already in lines 160-169, I guess that's enough for overriding them in the test? Something like:
https://github.com/gardener/gardener/blob/bc3c8c60e9a84d85816f5135b47815cf51cecae6/pkg/operation/botanist/backupentry/backupentry_test.go#L121

pkg/gardenlet/controller/shoot/shoot_care_control.go
```diff
@@ -143,10 +142,38 @@ type CareControlInterface interface {
 // implements the documented semantics for caring for Shoots. You should use an instance returned from NewDefaultCareControl()
 // for any scenario other than testing.
 func NewDefaultCareControl(clientMap clientmap.ClientMap, k8sGardenCoreInformers gardencoreinformers.Interface, secrets map[string]*corev1.Secret, imageVector imagevector.ImageVector, identity *gardencorev1beta1.Gardener, gardenClusterIdentity string, config *config.GardenletConfiguration) CareControlInterface {
-	return &defaultCareControl{clientMap, k8sGardenCoreInformers, secrets, imageVector, identity, gardenClusterIdentity, config}
+	return &defaultCareControl{
+		NewOperation,
+		NewHealthCheck,
+		NewConstraintCheck,
+		NewGarbageCollector,
```
rfranzke:
```suggestion
```

See questions/commit below

rfranzke (commented):
/assign

Thanks for fixing the bug and the added unit tests along the way!

pkg/gardenlet/controller/shoot/shoot_care_control.go
timuthy: The reason I had to introduce these default `var`s was because of the return types. For instance, `care.NewHealth` returns an instance of `care.Health` but the `NewHealthCheck` expects a `HealthCheck` interface type -> compile error. As an alternative I could have changed `care.NewHealth` to also return the interface type but this violates Golang [best practices](https://github.com/golang/go/wiki/CodeReviewComments#interfaces).

pkg/gardenlet/controller/shoot/shoot_care_control.go
timuthy: > Is this on purpose or type that NewGarbageCollector starts with a N while all other start with n?

Thanks, it wasn't on purpose. I'll change it to `newGarbageCollector `.

> Generally, what's the reason that those are attributes on the defaultCareControl structure?

The reason is that I only wanted to make these things injectable during object construction, i.e. once the `defaultCareControl` was created the behavior will never change. I can also make it using the exported vars every time if you think that's the cleaner way?

rfranzke: Yeah, at least that's what we do for other packages. If there is no other reason to keep these attributes in the structure then I'd vote for removing them and only use the function variables in the package and overwrite them in the test.

timuthy: Okay, no problem :)

pkg/gardenlet/controller/shoot/shoot_care_control.go
rfranzke: OK, yeah, let's keep it then. Thanks!

pkg/gardenlet/controller/shoot/shoot_care_control.go
rfranzke: Great, thanks!

rfranzke (approved):
/lgtm

ialidzhikov (approved):
/lgtm

Thank you for the unit tests!

PR #3459, opened 2021-01-26
Refactor gardener-seed-admission-controller

timebertt:

/area quality dev-productivity
/kind cleanup test
/priority normal

This PR refactors/enhances several things regarding the `seed-admission-controller`:

- migrate the component to a full controller-runtime component (only usage of manager was left)
- enable both admission review versions (including `admission/v1`) in the webhook configs, as it is now supported by the c-r webhook handlers
- decrease the timeouts of the extension deletion protection webhook to `10s` according to best pratices
- add an `envtest`-style integration test for the extension deletion protection webhook
- migrate the handcrafted integration test for the scheduler name webhook to an `envtest`-style test

The PR got a bit bigger than expected, but it's still split into several commits, so I hope this helps with the review.
/squash

**Release note**:

```other operator
`gardener-seed-admission-controller`'s webhooks now also accept reviews in version `admission/v1`. 
Also, webhook timeouts have been lowered to `10s` for the extension deletion protection webhooks.
```


pkg/operation/botanist/seedsystemcomponents/seedadmission/seedadmission.go
```diff
@@ -276,9 +278,8 @@ func (g *gardenerSeedAdmissionController) Deploy(ctx context.Context) error {
 		failurePolicy                  = admissionregistrationv1beta1.Fail
 		validatingWebhookConfiguration = &admissionregistrationv1beta1.ValidatingWebhookConfiguration{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      Name,
-				Namespace: g.namespace,
```
rfranzke, line 280: :)

pkg/seedadmission/extension_crds_test.go
```diff
@@ -529,46 +575,170 @@ var _ = Describe("Extension CRDs", func() {
...
+				&extensionsv1alpha1.ContainerRuntime{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: "foo"}},
+				&extensionsv1alpha1.ControlPlane{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: "foo"}},
+				&extensionsv1alpha1.Extension{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: "foo"}},
+				&extensionsv1alpha1.Infrastructure{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: "foo"}},
+				&extensionsv1alpha1.Network{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: "foo"}},
+				&extensionsv1alpha1.OperatingSystemConfig{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: "foo"}},
+				&extensionsv1alpha1.Worker{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: "foo"}},
+			}
+			crdObjects       = objects[0:9]
+			extensionObjects = objects[9:]
```
rfranzke: Hm, what about

```golang
crdObjects = []client.Object{
    &apiextensionsv1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: "backupbuckets.extensions.gardener.cloud"}},
    ...
}
extensionObjects = []client.Object{
    &extensionsv1alpha1.BackupBucket{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
    ...
}
objects = append(crdObjects, extensionObjects...)
```

instead?

rfranzke (dismissed):
/lgtm

Very nice, thanks for adding the integration test!

pkg/seedadmission/extension_crds_test.go
timebertt: Sure, done :)
/remove needs/rebase

rfranzke (dismissed):

stoyanr:
/lgtm

I have a general question: I used to write `envtest` while working on other projects in the past (SEN). I think it's the first such test added to g/g, possibly because it requires a controller runtime component (is this correct?). Are there plans to migrate other components (e.g. `gardenlet`) to controller runtime as well, so we could write add such tests for them as well?

timebertt:
> I think it's the first such test added to g/g, possibly because it requires a controller runtime component (is this correct?).

It's not the first one, but the second (https://github.com/gardener/gardener/pull/2830 was the first one) 😉 
You're right though, there are not many envtests in g/g currently.
I think, we should be able to leverage envtest for integration tests for some of our controllers as well, at least for the ones that already implement the `reconciler.Reconcile` interface. Though, we haven't done any of those yet.

> Are there plans to migrate other components (e.g. `gardenlet`) to controller runtime as well, so we could write add such tests for them as well?

I plan to do the refactoring for gardener-admission-controller this week (similar to this PR).
I think, the general goal is to do the same for the controller components as well, though it will be much more effort and might reveal some difficulties. As of now, there are no concrete plans for this, I guess.

timebertt:
Rebased to resolve conflicts.

rfranzke (approved):

PR #3461, opened 2021-01-27
Add `ShootVPAEnabledByDefault` API server feature gate

### LINKED ISSUE #3304: enable vpa by default for newly created shoots (configurable)

/area auto-scaling
/kind enhancement
/priority normal

The `node-exporter` should be under VPA.

**Why is this needed**:

Static requests and limits do not make sense for the node exporter. The resource consumption of the node exporter greatly depends on the size/activity of the node. If the nodes are too large the node exporter will have issues. Increasing the requests and limits across the board for all node exporters will mean wasted resources for smaller nodes. 

**Potential Issues**:

VPA is not active in all shoots by default so we cannot have VPA for the node exporter unless it is configured in the shoot.

rfranzke:

/area scalability
/kind enhancement
/priority normal

This PR adds a new `ShootVPAEnabledByDefault` feature gate (alpha, disabled by default) to the gardener-apiserver. If enabled, the `.spec.kubernetes.verticalPodAutoscaler.enabled` field for newly created `Shoot` resources is defaulted to `true`. Existing `Shoot`s are not modified, i.e., if VPA shall be enabled then it needs to be explicitly set.

Fixes #3304

**Release note**:

```feature operator
The new `ShootVPAEnabledByDefault` feature gate for the `gardener-apiserver` (alpha, disabled by default) controls whether the `.spec.kubernetes.verticalPodAutoscaler.enabled` field for newly created `Shoot` resources is defaulted to `true`. Existing `Shoot`s are not modified, i.e., if VPA shall be enabled then it needs to be explicitly set. In the future, the feature gate will be removed again and the API defaulting will be changed in general.
```

### DIFF
--- a/pkg/apis/core/v1alpha1/defaults.go
+++ b/pkg/apis/core/v1alpha1/defaults.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	"github.com/gardener/gardener/pkg/features"
 	"github.com/gardener/gardener/pkg/utils"
 	versionutils "github.com/gardener/gardener/pkg/utils/version"
 
@@ -28,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/utils/pointer"
 )
 
@@ -167,6 +169,13 @@ func SetDefaults_Shoot(obj *Shoot) {
 	if obj.Spec.Kubernetes.KubeProxy.Mode == nil {
 		defaultProxyMode := ProxyModeIPTables
 		obj.Spec.Kubernetes.KubeProxy.Mode = &defaultProxyMode
+	}
+
+	if obj.Spec.Kubernetes.VerticalPodAutoscaler == nil &&
+		len(obj.Status.TechnicalID) == 0 &&
+		features.IsFeatureGateKnown(utilfeature.DefaultFeatureGate, features.ShootVPAEnabledByDefault) &&
+		utilfeature.DefaultFeatureGate.Enabled(features.ShootVPAEnabledByDefault) {
+		obj.Spec.Kubernetes.VerticalPodAutoscaler = &VerticalPodAutoscaler{Enabled: true}
 	}
 
 	if obj.Spec.Addons == nil {
--- a/pkg/apis/core/v1beta1/defaults.go
+++ b/pkg/apis/core/v1beta1/defaults.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	"github.com/gardener/gardener/pkg/features"
 	"github.com/gardener/gardener/pkg/utils"
 	versionutils "github.com/gardener/gardener/pkg/utils/version"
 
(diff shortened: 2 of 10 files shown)


timebertt (commented):
Hmm, interesting way to address this via a feature gate.
Have you also considered an admission plugin?
I think, an admission plugin would be more fitting for this, as there is no "graduation process" associated with it and its merely a configuration option.
Somewhen in the feature we might consider enabling it by default, but maybe that's not even the goal. How would we reflect this decision then with the feature gate (it probably shouldn't stay forever)?
WDYT?

rfranzke:
Our goal is to change the API defaulting for `Shoot`s in general and always enable VPA. We couldn't do this in the past because it's an incompatible API change, and existing shoots might have already a VPA installed.
Now, with this option it's possible to prepare the end-users of a particular landscape for this upcoming change in the defaulting logic. Eventually, we want to remove the feature gate again. I also thought about the admission plugin, but given the plan I decided for a feature gate because it's simply easier to use and configure, hence, I only want to spend the minimal required effort for this temporary code.

pkg/apis/core/v1beta1/defaults.go
```diff
@@ -169,6 +171,12 @@ func SetDefaults_Shoot(obj *Shoot) {
 		obj.Spec.Kubernetes.KubeProxy.Mode = &defaultProxyMode
 	}
 
+	if features.IsFeatureGateKnown(utilfeature.DefaultFeatureGate, features.ShootVPAEnabledByDefault) &&
```
vpnachev:
```suggestion
	if obj.Spec.Kubernetes.VerticalPodAutoscaler == nil &&
	features.IsFeatureGateKnown(utilfeature.DefaultFeatureGate, features.ShootVPAEnabledByDefault) &&
```

pkg/apis/core/v1alpha1/defaults.go
```diff
@@ -169,6 +171,12 @@ func SetDefaults_Shoot(obj *Shoot) {
 		obj.Spec.Kubernetes.KubeProxy.Mode = &defaultProxyMode
 	}
 
+	if features.IsFeatureGateKnown(utilfeature.DefaultFeatureGate, features.ShootVPAEnabledByDefault) &&
```
vpnachev:
```suggestion
	if obj.Spec.Kubernetes.VerticalPodAutoscaler == nil && 
	features.IsFeatureGateKnown(utilfeature.DefaultFeatureGate, features.ShootVPAEnabledByDefault) &&
```

vpnachev (changes requested):
The VPA should be enabled by default only if the VPA is not already configured.

PS. Such change deserve new API version.

rfranzke:
> The VPA should be enabled by default only if the VPA is not already configured.

Nice catch, sure, thank you!

> PS. Such change deserve new API version.

Generally, I agree, but this has the downside that end-users have to pro-actively switch to this new API group and this can take forever. Yes, we can have deprecation periods for the old/existing API groups and eventually remove them, but all of this is again huge effort. We don't gain much out of this (other that we can say that we "follow the API conventions, etc." here), but the outcome would be the same.
I guess the way how it's proposed right now is a pragmatic alternative that doesn't cause too much "harm", and especially not too much effort.

mvladev (changes requested):
I'm against using feature gate in the `default.go` - why?

1. The API is used by external users and there defaulting happens when objects are deserialzed e.g. reading from yaml file. This feature gate is not something we can control or end-users are already aware of. Such feature gates should not be part of the API as the defaulting code should not rely on global state (e.g. the features) and should be idempotent. 
1. Exposing another package to potential end-users is bad.
1. See the K8S API - https://github.com/kubernetes/api/ while fields might be set / enabled with feature gates, it doesn't affect the official API (`k8s.io/kubernetes/pkg/features` is not imported anywhere). Such settings happen either via [registry at backend](https://github.com/kubernetes/kubernetes/blob/a0e613363ac0f3b7a797afe863c2c947d94a366d/pkg/registry/core/service/strategy.go#L185-L190), admission plugins or api version bump.

To solve this problem either 

1. bump the API version OR add admission plugin
2. add it to the registry logic - bad practice.

vpnachev (commented):
/lgtm codewise

But I also lean toward admission plugin to not violate the established API principles.

rfranzke:
OK, then let's 
/close
this PR if there are too many concerns with the proposal. As I'm busy with other things already and don't have the capacity to rework it entirely, I'll also unassign from the related issue for now.

PR #3466, opened 2021-01-28
Expose FetchEvents function

timuthy:

/area open-source
/kind enhancement
/priority normal

This PR exposes an already implemented functionality to fetch event messages for involved objects. The motivation for this change is to allow extensions / Gardener Resource Manager calling the functionality as well.

/cc @ialidzhikov 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/utils/kubernetes/kubernetes.go
+++ b/pkg/utils/kubernetes/kubernetes.go
@@ -184,9 +184,12 @@ func WaitUntilResourceDeletedWithDefaults(ctx context.Context, c client.Client, 
 // WaitUntilLoadBalancerIsReady waits until the given external load balancer has
 // been created (i.e., its ingress information has been updated in the service status).
 func WaitUntilLoadBalancerIsReady(ctx context.Context, kubeClient kubernetes.Interface, namespace, name string, timeout time.Duration, logger *logrus.Entry) (string, error) {
-	var loadBalancerIngress string
+	var (
+		loadBalancerIngress string
+		service             = &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace}}
+	)
 	if err := retry.UntilTimeout(ctx, 5*time.Second, timeout, func(ctx context.Context) (done bool, err error) {
-		loadBalancerIngress, err = GetLoadBalancerIngress(ctx, kubeClient.Client(), namespace, name)
+		loadBalancerIngress, err = GetLoadBalancerIngress(ctx, kubeClient.Client(), service)
 		if err != nil {
 			logger.Infof("Waiting until the %s service deployed is ready...", name)
 			// TODO(AC): This is a quite optimistic check / we should differentiate here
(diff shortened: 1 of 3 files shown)


pkg/utils/kubernetes/kubernetes.go
```diff
@@ -195,23 +194,26 @@ func WaitUntilLoadBalancerIsReady(ctx context.Context, kubeClient kubernetes.Int
...
-			"involvedObject.kind":      "Service",
-			"involvedObject.name":      name,
-			"involvedObject.namespace": namespace,
-			"type":                     corev1.EventTypeWarning,
+		const eventsLimit = 2
+
+		lbService := &corev1.Service{
+			TypeMeta: metav1.TypeMeta{
+				Kind: "Service",
+			},
```
rfranzke: Hmm, isn't there another way so that we don't need to explicitly set the `TypeMeta`? I don't have a concrete suggestion but maybe @timebertt has?

timebertt: One could use the client's Scheme to get the `GVK` of the object. E.g.:
```go
gvk, err := apiutil.GVKForObject(obj, c.Scheme())
```
Or directly use the scheme
```go
gvk, isUnversioned, err := c.Scheme().ObjectKinds(obj)
```

ialidzhikov:
/lgtm

Thank you for the tests. Shame that I didn't do it when I initially introduced this. :(

pkg/utils/kubernetes/kubernetes.go
timuthy: I followed this approach but ended up in more error handling etc. I chose an alternative and hope you're fine with it 👀 🙏

ialidzhikov: Just a minor concern. To my understanding now `FetchEventMessages` relies that `GetLoadBalancerIngress` func will refetch the svc and in this way populate its kind? It is also odd that `GetLoadBalancerIngress` accepts a `corev1.Service` and it does a refetch - with this signature a caller might fetch in advance the svc, call `GetLoadBalancerIngress` and expect that there will be no redundant get call. With the old signature that accepts only the namespace and the name, it is clear the a get request will be issued.

pkg/utils/kubernetes/kubernetes.go
```diff
@@ -317,8 +311,36 @@ func ReconcileServicePorts(existingPorts []corev1.ServicePort, desiredPorts []co
...
+func FetchEventMessages(ctx context.Context, c client.Client, obj client.Object, eventType string, eventsLimit int) (string, error) {
+	apiVersion, kind := obj.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
+	if apiVersion == "" {
+		return "", fmt.Errorf("apiVersion not specified for object %s/%s", obj.GetNamespace(), obj.GetName())
+	}
+	if kind == "" {
+		return "", fmt.Errorf("kind not specified for object %s/%s", obj.GetNamespace(), obj.GetName())
+	}
+	fieldSelector := client.MatchingFields{
+		"involvedObject.apiVersion": apiVersion,
```
timuthy, line 326: @ialidzhikov this was added after talking to @rfranzke and @timebertt. Can you confirm that this was not left out by intention earlier?

ialidzhikov: Yes, it makes sense to also include the apiVersion. 👍

pkg/utils/kubernetes/kubernetes.go
timuthy: Good point @ialidzhikov. I had a similar discussion with @rfranzke about this [here](https://github.com/gardener/gardener/pull/3353#discussion_r553443176).

timuthy: As an alternative I can revert the changes on `GetLoadBalancerIngress` and let it return the `Service` object i.e.:

```
GetLoadBalancerIngress(ctx context.Context, c client.Client, namespace, name string) (string, service *corev1.Service, error)
```

ialidzhikov (dismissed):
/lgtm

pkg/utils/kubernetes/kubernetes.go
timuthy: I updated the doc string of the function accordingly. Please let me know in case we need a different approach.

timebertt (approved):
/lgtm

rfranzke (approved):

PR #3467, opened 2021-01-28
Use protobuf in gardener components

### LINKED ISSUE #2836: Use protobuf in gardener components

/area performance cost
/kind enhancement
/priority normal

Currently all gardener components are configured to accept and send requests with content type `application/json`.
We should switch to protobuf (`application/vnd.kubernetes.protobuf`) by default, where possible.
As we heavily make use of CRDs in our Seed clusters it might not be possible to switch the seed client connection settings to protobuf at all, as CRDs don't support protobuf. But at least for the garden client connection it should be possible.
A fallback to json can be defined for `acceptContentTypes` with `application/vnd.kubernetes.protobuf;application/json`.

**Why is this needed**:
Using protobuf should be more efficient and might safe some data transfer costs.

timebertt:

/area performance cost networking
/kind enhancement
/priority normal

This PR configures clients in all gardener components to use the protobuf content type wherever possible.
Earlier we used `application/json` everywhere (set in gardener helm charts, defaulted in component configs).

Now, the content type fields are optional in the charts as well as the component configs and are not defaulted anymore.
If left empty, gardener will configure all clients that talk to k8s / gardener APIs using protobuf, for CRDs we still use `application/json` as they don't support protobuf.

Fixes https://github.com/gardener/gardener/issues/2836

I think, I cut down json requests to the minimum amount that is possible ATM. There are still json requests in the following cases:
- g-API posts `tokenreviews/subjectaccessreviews` to k-api via json as part of the delegating authn/authz mechanisms (not configurable in the API server library, IISIC)
- some kcm controllers (maybe gc controller?) seem to talk to g-api via json sometimes, also not configurable but not to frequent
- g-api and kcm watch the `kube-system/extension-apiserver-authentication` cm using json, also not configurable
- unstructured requests to c-r clients still use json (-> therefore also the chart applier) -> usage of these requests will decrease significantly with the ongoing refactorings

/squash

**Release note**:

```breaking operator
Gardener components now use the protobuf content type wherever possible when talking to the Gardener or Kubernetes APIs, if the content type fields are left empty in the respective component configs.
Operators can override this behavior by explicitly specifying `application/json` as the content type in the respective component configs.
```

### DIFF
--- a/charts/gardener/controlplane/charts/runtime/templates/admission-controller/configmap-componentconfig.yaml
+++ b/charts/gardener/controlplane/charts/runtime/templates/admission-controller/configmap-componentconfig.yaml
@@ -16,8 +16,12 @@ data:
     apiVersion: admissioncontroller.config.gardener.cloud/v1alpha1
     kind: AdmissionControllerConfiguration
     gardenClientConnection:
-      acceptContentTypes: {{ required ".Values.global.admission.config.gardenClientConnection.acceptContentTypes is required" .Values.global.admission.config.gardenClientConnection.acceptContentTypes }}
-      contentType: {{ required ".Values.global.admission.config.gardenClientConnection.contentType is required" .Values.global.admission.config.gardenClientConnection.contentType }}
+      {{- with .Values.global.admission.config.gardenClientConnection.acceptContentTypes }}
+      acceptContentTypes: {{ . | quote }}
+      {{- end }}
+      {{- with .Values.global.admission.config.gardenClientConnection.contentType }}
+      contentType: {{ . | quote }}
+      {{- end }}
       qps: {{ required ".Values.global.admission.config.gardenClientConnection.qps is required" .Values.global.admission.config.gardenClientConnection.qps }}
       burst: {{ required ".Values.global.admission.config.gardenClientConnection.burst is required" .Values.global.admission.config.gardenClientConnection.burst }}
       {{- if .Values.global.admission.config.gardenClientConnection.kubeconfig }}
--- a/cmd/gardener-apiserver/app/gardener_apiserver.go
+++ b/cmd/gardener-apiserver/app/gardener_apiserver.go
@@ -158,26 +158,31 @@ func (o *Options) config(kubeAPIServerConfig *rest.Config, kubeClient *kubernete
 
 	// Initialize admission plugins
 	o.Recommended.ExtraAdmissionInitializers = func(c *genericapiserver.RecommendedConfig) ([]admission.PluginInitializer, error) {
+		protobufLoopbackConfig := *gardenerAPIServerConfig.LoopbackClientConfig
+		if protobufLoopbackConfig.ContentType == "" {
+			protobufLoopbackConfig.ContentType = runtime.ContentTypeProtobuf
+		}
+
 		// core client
-		coreClient, err := gardencoreclientset.NewForConfig(gardenerAPIServerConfig.LoopbackClientConfig)
+		coreClient, err := gardencoreclientset.NewForConfig(&protobufLoopbackConfig)
 		if err != nil {
 			return nil, err
 		}
-		o.CoreInformerFactory = gardencoreinformers.NewSharedInformerFactory(coreClient, gardenerAPIServerConfig.LoopbackClientConfig.Timeout)
+		o.CoreInformerFactory = gardencoreinformers.NewSharedInformerFactory(coreClient, protobufLoopbackConfig.Timeout)
 
 		// versioned core client
-		versionedCoreClient, err := gardenversionedcoreclientset.NewForConfig(gardenerAPIServerConfig.LoopbackClientConfig)
+		versionedCoreClient, err := gardenversionedcoreclientset.NewForConfig(&protobufLoopbackConfig)
 		if err != nil {
 			return nil, err
 		}
-		o.ExternalCoreInformerFactory = gardenexternalcoreinformers.NewSharedInformerFactory(versionedCoreClient, gardenerAPIServerConfig.LoopbackClientConfig.Timeout)
+		o.ExternalCoreInformerFactory = gardenexternalcoreinformers.NewSharedInformerFactory(versionedCoreClient, protobufLoopbackConfig.Timeout)
 
 		// settings client
-		settingsClient, err := settingsclientset.NewForConfig(gardenerAPIServerConfig.LoopbackClientConfig)
+		settingsClient, err := settingsclientset.NewForConfig(&protobufLoopbackConfig)
 		if err != nil {
 			return nil, err
 		}
-		o.SettingsInformerFactory = settingsinformer.NewSharedInformerFactory(settingsClient, gardenerAPIServerConfig.LoopbackClientConfig.Timeout)
+		o.SettingsInformerFactory = settingsinformer.NewSharedInformerFactory(settingsClient, protobufLoopbackConfig.Timeout)
 
 		// dynamic client
 		dynamicClient, err := dynamic.NewForConfig(kubeAPIServerConfig)
@@ -223,8 +228,13 @@ func (o Options) run(stopCh <-chan struct{}) error {
 		return err
 	}
 
+	protobufConfig := *kubeAPIServerConfig
+	if protobufConfig.ContentType == "" {
+		protobufConfig.ContentType = runtime.ContentTypeProtobuf
+	}
+
 	// kube client
-	kubeClient, err := kubernetes.NewForConfig(kubeAPIServerConfig)
+	kubeClient, err := kubernetes.NewForConfig(&protobufConfig)
 	if err != nil {
 		return err
 	}
--- a/pkg/admissioncontroller/apis/config/v1alpha1/defaults_test.go
+++ b/pkg/admissioncontroller/apis/config/v1alpha1/defaults_test.go
@@ -15,15 +15,16 @@
 package v1alpha1_test
 
 import (
-	. "github.com/gardener/gardener/pkg/admissioncontroller/apis/config/v1alpha1"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	rbacv1 "k8s.io/api/rbac/v1"
+	componentbaseconfigv1alpha1 "k8s.io/component-base/config/v1alpha1"
+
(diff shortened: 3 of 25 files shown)


pkg/client/kubernetes/client.go
```diff
@@ -44,6 +49,16 @@ var (
 // KubeConfig is the key to the kubeconfig
 const KubeConfig = "kubeconfig"
 
+func init() {
+	// enable protobuf for Gardener API for controller-runtime clients
+	protobufSchemeBuilder := runtime.NewSchemeBuilder(
+		gardenercorescheme.AddToScheme,
+		settingsscheme.AddToScheme,
```
rfranzke, line 58: What about the new seedmanagement scheme?

rfranzke (changes requested):
Well done, only one comment

pkg/client/kubernetes/client.go
timebertt: Good point, thanks :)
Added.
/squash

timuthy (dismissed):
/lgtm

rfranzke (dismissed):
/lgtm

charts/gardener/controlplane/charts/runtime/templates/admission-controller/configmap-componentconfig.yaml
```diff
@@ -16,8 +16,12 @@ data:
     apiVersion: admissioncontroller.config.gardener.cloud/v1alpha1
     kind: AdmissionControllerConfiguration
     gardenClientConnection:
-      acceptContentTypes: {{ required ".Values.global.admission.config.gardenClientConnection.acceptContentTypes is required" .Values.global.admission.config.gardenClientConnection.acceptContentTypes }}
-      contentType: {{ required ".Values.global.admission.config.gardenClientConnection.contentType is required" .Values.global.admission.config.gardenClientConnection.contentType }}
+      {{- with .Values.global.admission.config.gardenClientConnection.acceptContentTypes }}
+      acceptContentTypes: {{ . | quote }}
+      {{- end }}
```
ialidzhikov, line 21: Nit: Is there a specific reason using `with` instead of `if`? I would vote to use `if` just for consistency reason as currently across the gardener chart `if` is being used in a consistent manner.

```suggestion
      {{- if .Values.global.admission.config.gardenClientConnection.acceptContentTypes }}
      acceptContentTypes: {{ .Values.global.admission.config.gardenClientConnection.acceptContentTypes }}
      {{- end }}
```

cmd/gardener-apiserver/app/gardener_apiserver.go
```diff
@@ -223,8 +228,13 @@ func (o Options) run(stopCh <-chan struct{}) error {
 		return err
 	}
 
+	protobufConfig := *kubeAPIServerConfig
+	if protobufConfig.ContentType == "" {
+		protobufConfig.ContentType = runtime.ContentTypeProtobuf
```
ialidzhikov, line 233: `ContentType` indicates only the content type of the request body that clients sends to the server (or it may also indicate the content type of the response from the server). What about the `AcceptContentTypes` (the content type of the response from the server that the client prefers) - what is its default value? `application/json`?

cmd/gardener-apiserver/app/gardener_apiserver.go
```diff
@@ -158,26 +158,31 @@ func (o *Options) config(kubeAPIServerConfig *rest.Config, kubeClient *kubernete
 
 	// Initialize admission plugins
 	o.Recommended.ExtraAdmissionInitializers = func(c *genericapiserver.RecommendedConfig) ([]admission.PluginInitializer, error) {
+		protobufLoopbackConfig := *gardenerAPIServerConfig.LoopbackClientConfig
+		if protobufLoopbackConfig.ContentType == "" {
+			protobufLoopbackConfig.ContentType = runtime.ContentTypeProtobuf
```
ialidzhikov, line 163: I don't have much details but to my knowledge the apiservers support content negotiation. Can't we just set our defaults `application/vnd.kubernetes.protobuf,application/json` and say - first try with `application/vnd.kubernetes.protobuf`, if the apiserver does not support it switch back to `application/json`? Ref https://github.com/kubernetes/kube-state-metrics/pull/264/files#r140285479

pkg/admissioncontroller/apis/config/v1alpha1/defaults_test.go
```diff
@@ -56,12 +57,31 @@ var _ = Describe("Defaults", func() {
...
 				Expect(obj.Server.ResourceAdmissionConfiguration.UnrestrictedSubjects[1].APIGroup).To(Equal(rbacv1.GroupName))
 				Expect(obj.Server.ResourceAdmissionConfiguration.UnrestrictedSubjects[2].APIGroup).To(Equal(""))
 			})
 		})
+
+		Describe("GardenClientConnection", func() {
+			It("should not default ContentType and AcceptContentTypes", func() {
+				SetObjectDefaults_AdmissionControllerConfiguration(obj)
+
+				// ContentType fields will be defaulted by client constructors / controller-runtime based on wether a
```
ialidzhikov:
```suggestion
				// ContentType fields will be defaulted by client constructors / controller-runtime based on whether a
```

pkg/gardenlet/apis/config/v1alpha1/defaults_test.go
```diff
@@ -62,6 +62,27 @@ var _ = Describe("Defaults", func() {
 			Expect(obj.SNI.Ingress.Namespace).To(PointTo(Equal("istio-ingress")))
 			Expect(obj.SNI.Ingress.ServiceName).To(PointTo(Equal("istio-ingressgateway")))
 		})
+
+		Describe("GardenClientConnection", func() {
+			It("should not default ContentType and AcceptContentTypes", func() {
+				SetObjectDefaults_GardenletConfiguration(obj)
+
+				// ContentType fields will be defaulted by client constructors / controller-runtime based on wether a
```
ialidzhikov:
```suggestion
				// ContentType fields will be defaulted by client constructors / controller-runtime based on whether a
```

pkg/scheduler/apis/config/v1alpha1/defaults_test.go
```diff
@@ -0,0 +1,80 @@
...
+					},
+				}))
+			})
+		})
+
+		Describe("ClientConnection", func() {
+			It("should not default ContentType and AcceptContentTypes", func() {
+				configv1alpha1.SetObjectDefaults_SchedulerConfiguration(obj)
+
+				// ContentType fields will be defaulted by client constructors / controller-runtime based on wether a
```
ialidzhikov:
```suggestion
				// ContentType fields will be defaulted by client constructors / controller-runtime based on whether a
```

ialidzhikov (changes requested):

charts/gardener/controlplane/charts/runtime/templates/admission-controller/configmap-componentconfig.yaml
timebertt: Hmm, I see a few usages of `with`.
Personally, I prefer `with` (especially for templating a single field), because it allows you to not duplicate the value path.
But I'm open to changing it, if you think consistency is a concern here.

cmd/gardener-apiserver/app/gardener_apiserver.go
timebertt: According to the [doc string](https://github.com/kubernetes/client-go/blob/ab82d40f6e857a3162e22ac8a5888b6314f9b0eb/rest/config.go#L282-L284) `AcceptContentTypes` is defaulted to `ContentType`.
My reasoning was, it should be enough to default `ContentType` because it will also be used for `AcceptContentTypes` if the operator left the field empty. If the operator specified a value, the operator probably knows what they are doing and the value should be respected.

timebertt: I checked the details now:
The `Accept` header is set to `AcceptContentTypes` if specified, otherwise set to `ContentType + ", */*"`, see [here](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L149-L154).

cmd/gardener-apiserver/app/gardener_apiserver.go
timebertt: The `Content-Type` header can't specify multiple content types according to my knowledge, as it has to specify the exact content type used in the request's body.
IIUC, the content type of the request body can't be negotiated.

The `AcceptContentType` OTOH supports content negotiation, as you correctly pointed out.
But, as it will be set to `ContentType + ", */*"` (see above), it will automatically be negotiated, if the operator didn't specify the setting explicitly. 
So I don't see the need to overwrite it. Do you agree?

cmd/gardener-apiserver/app/gardener_apiserver.go
ialidzhikov: Ok, thanks for explaining.

cmd/gardener-apiserver/app/gardener_apiserver.go
ialidzhikov: Ok, thanks for explaining.

ialidzhikov (dismissed):
/lgtm

pkg/gardenlet/apis/config/v1alpha1/defaults_test.go
```diff
@@ -62,6 +62,27 @@ var _ = Describe("Defaults", func() {
 			Expect(obj.SNI.Ingress.Namespace).To(PointTo(Equal("istio-ingress")))
 			Expect(obj.SNI.Ingress.ServiceName).To(PointTo(Equal("istio-ingressgateway")))
 		})
+
+		Describe("GardenClientConnection", func() {
```
stoyanr: Very minor nit: Similar test cases could be added for `SeedClientConnection` and `ShootClientConnection`.

stoyanr (dismissed):

pkg/gardenlet/apis/config/v1alpha1/defaults_test.go
timebertt: Yes, good point. Thanks!
Added checks for those two as well.

rfranzke (approved):

PR #3472, opened 2021-01-28
Use version and verflag pkgs from k8s.io/component-base

ialidzhikov:
/kind cleanup

Currently `github.com/gardener/gardener/pkg/version` and `github.com/gardener/gardener/pkg/version/verflag` are 90% similar to `k8s.io/component-base/version` and `k8s.io/component-base/version/verflag`. Until now we were not able to use the k8s.io/component-base packages as the programName was not configurable (hard-coded to `Kubernetes`). This limitation is fixed with https://github.com/kubernetes/kubernetes/pull/90139 and we can now reuse the pkgs from `k8s.io/component-base`.

**Release note**:

```other dependency
The packages `github.com/gardener/gardener/pkg/version` and `github.com/gardener/gardener/pkg/version/verflag` are now deprecated in favour of `k8s.io/component-base/version` and `k8s.io/component-base/version/verflag`.
```

### DIFF
--- a/cmd/gardener-admission-controller/app/gardener_admission_controller.go
+++ b/cmd/gardener-admission-controller/app/gardener_admission_controller.go
@@ -32,13 +32,13 @@ import (
 	"github.com/gardener/gardener/pkg/healthz"
 	"github.com/gardener/gardener/pkg/logger"
 	"github.com/gardener/gardener/pkg/server"
-	"github.com/gardener/gardener/pkg/version/verflag"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
 
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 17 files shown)


pkg/version/verflag/verflag.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Deprecated: use k8s.io/component-base/version/verflag instead.
```
timebertt, line 15: Can we maybe get rid of these packages directly?
Any reason for just deprecating it?

timebertt (approved):
Very nice, @ialidzhikov!
/lgtm

pkg/version/verflag/verflag.go
ialidzhikov: No concrete reason but I wanted to keep them as deprecated for 1-2 versions before we remove them. Just to make migration smoother in case a project vendors these packages.

vpnachev (approved):
/lgtm

pkg/version/verflag/verflag.go
timebertt: Ok, fine for me :)

PR #3476, opened 2021-01-29
Cleanup of runtime.Object and meta.Accessor

timebertt:

/kind cleanup
/priority normal

With `controller-runtime@v0.7` there are `client.Object` and `client.ObjectList` to distinguish between single API objects and object lists.
c-r has replaced their usage of `runtime.Object` (which is more generic) by those two interfaces to have a clearer func/interface contract of what implementations expect from passed objects.

I followed up on this and cleaned up our usage of `runtime.Object` as well and replaced it with `client.{Object,ObjectList}` wherever applicable.
Though, there are still a lot of `runtime.Object` usages left, e.g. in API and registry packages, which have to comply with other interfaces expecting `runtime.Object`. 
Also some of our helper funcs are generic and accept a single object as well as an object list, so for those cases `runtime.Object` is also still needed.
I think I cut it down to the bare minimum now.

I also cleaned up some unnecessary usages of `meta.Accessor` in favor of `client.Object` where it is helpful and makes code a bit slimmer.

/squash

**Release note**:

```breaking dependency
Some helper functions and interface types have been changed to accept `client.{Object,ObjectList}` instead of `runtime.Object` where applicable to have a clearer contract of what is expected from passed parameters.
```

### DIFF
--- a/pkg/utils/kubernetes/kubernetes.go
+++ b/pkg/utils/kubernetes/kubernetes.go
@@ -77,12 +77,8 @@ func HasMetaDataAnnotation(meta metav1.Object, key, value string) bool {
 }
 
 // HasDeletionTimestamp checks if an object has a deletion timestamp
-func HasDeletionTimestamp(obj runtime.Object) (bool, error) {
-	metadata, err := meta.Accessor(obj)
-	if err != nil {
-		return false, err
-	}
-	return metadata.GetDeletionTimestamp() != nil, nil
+func HasDeletionTimestamp(obj client.Object) bool {
+	return obj.GetDeletionTimestamp() != nil
 }
 
 func nameAndNamespace(namespaceOrName string, nameOpt ...string) (namespace, name string) {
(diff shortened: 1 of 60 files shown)


rfranzke:
❤️ all these cleanup PRs!

pkg/utils/kubernetes/kubernetes.go
```diff
@@ -77,12 +77,8 @@ func HasMetaDataAnnotation(meta metav1.Object, key, value string) bool {
 }
 
 // HasDeletionTimestamp checks if an object has a deletion timestamp
-func HasDeletionTimestamp(obj runtime.Object) (bool, error) {
-	metadata, err := meta.Accessor(obj)
-	if err != nil {
-		return false, err
-	}
-	return metadata.GetDeletionTimestamp() != nil, nil
+func HasDeletionTimestamp(obj client.Object) bool {
```
rfranzke: Do we still need this function now?

rfranzke (dismissed):
/lgtm

pkg/utils/kubernetes/kubernetes.go
timebertt: Hmm, probably not. Removed it.

rfranzke (approved):

PR #3477, opened 2021-01-29
Use SeedTemplate instead of Seed in GardenletConfiguration

### LINKED ISSUE #3469: Using SeedTemplate instead of Seed in GardenletConfiguration

/area control-plane
/kind enhancement
/priority normal

The design of the `GardenletConfiguration` resource should be improved:
* Currently, it embeds `Seed` in `SeedConfig`: https://github.com/gardener/gardener/blob/a1b0891be6061c9511b7d6038e381c4a1a6c0f04/pkg/gardenlet/apis/config/types.go#L288-L291. Instead, it should embed `SeedTemplate`: https://github.com/gardener/gardener/blob/a1b0891be6061c9511b7d6038e381c4a1a6c0f04/pkg/apis/seedmanagement/v1alpha1/types_managedseed.go#L75-L83 using the internal version in `config` and the external one in `v1alpha1`. 
* Currently, it does not validate the `Seed` embedded via `SeedConfig`. The `SeedTemplate` should be validated by calling a function similar to e.g. https://github.com/gardener/gardener/blob/a1b0891be6061c9511b7d6038e381c4a1a6c0f04/pkg/apis/seedmanagement/validation/managedseed.go#L114-L128. 
* The explicit validation of `SeedConfig` should be removed from `seedmanagement/validation`.
* Potentially, `SeedTemplate` should be also moved to `core` (and its validation to `core/validation`) to avoid the extra dependency to `seedmanagement`. This dependency would not lead to a circular dependency because `GardenletConfiguration` is not used explicitly in `ManagedSeed` (only via a `RawExtension`), but it's perhaps still a good idea to avoid it.

**Why is this needed**:
As discussed in https://github.com/gardener/gardener/pull/3414#discussion_r563806533, it is inconsistent and confusing that `seedConfig` / `seedTemplate` of a ManagedSeed is validated, but the the `seedConfig` of the standalone `GardenletConfiguration` is not validated. Since this is potentially a breaking change, we agreed to do it in a separate PR.

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Changes from `Seed` to `SeedTemplate` in `GardenletConfiguration`.

Fixes #3469

**Release note**:
```other operator
NONE
```


rfranzke (dismissed):
/lgtm

stoyanr:
There is a bug I just found that needs to be fixed, please don't merge. I converted the PR to draft, will reopen it back when I am done.

stoyanr:
The issue is fixed now, the PR can be merged if there are no more comments.

pkg/gardenlet/apis/config/v1alpha1/types.go
```diff
@@ -347,7 +347,7 @@ type LeaderElectionConfiguration struct {
 
 // SeedConfig contains configuration for the seed cluster.
 type SeedConfig struct {
-	gardencorev1beta1.Seed `json:",inline"`
+	gardencorev1beta1.SeedTemplate `json:",inline"`
```
timebertt, line 350: WDYT about removing the `SeedConfig` type in favor of the new `SeedTemplate`?

We might even consider renaming the `seedConfig` field to `seedTemplate` to make it consistent with the `ManagedSeed`.
Yes, I know, this is a breaking change. But the gardenlet config API is only in an `alpha` version, which means we are able to introduce breaking changes by API conventions.
Though, this is probably too much effort for such a cosmetic change and is not worth it.
WDYT?

pkg/gardenlet/controller/seed/seed.go
```diff
@@ -187,7 +187,14 @@ func (c *Controller) Run(ctx context.Context, workers int) {
 			seed.Labels = utils.MergeStringMaps(map[string]string{
 				v1beta1constants.GardenRole: v1beta1constants.GardenRoleSeed,
 			}, c.config.SeedConfig.Labels)
-			seed.Spec = c.config.SeedConfig.Seed.Spec
+
+			// Convert gardenlet config to an external version
+			cfg, err := confighelper.ConvertGardenletConfigurationExternal(c.config)
```
timebertt: Nit: Can you move this up, so it's outside of `CreateOrUpdate`?
It doesn't depend on the object fetched from the API server.

timebertt (commented):
Thanks a lot for cleaning this up! I think, it's much cleaner now.
Only one question and one minor nit from my side, rest looks good.

pkg/gardenlet/apis/config/v1alpha1/types.go
stoyanr: Yes, I think it's not worth it. As it is an existing YAML gardenlet configuration is basically compatible, even if a different type is being used. To introduce this we should either create a new version (`v1alpha2`) and take care of conversions, or we should be prepared to break all existing gardenlet configurations ... I don't think we want to spend the time and effort to fix them all.

pkg/gardenlet/controller/seed/seed.go
stoyanr: Done

stoyanr:
Rebased on top of latest master, fixed managed seed admission, and added a new commit with code review changes.
/squash

rfranzke (approved):

pkg/gardenlet/apis/config/v1alpha1/types.go
timebertt: Yes, agreed.

What about the first question: should we remove the `SeedConfig` type and change the `seedConfig` field to be of type `SeedTemplate` (no indirection)?

stoyanr: I don't like `seedConfig` of type `SeedTemplate`, and I don't think the indirection hurts in any way. I would vote for no more changes.

Could we merge this PR already?

timebertt: Ok, fine for me.

timebertt (approved):

PR #3479, opened 2021-01-29
Add options for configuring leader election resource lock

timebertt:

/area scalability
/kind enhancement
/priority normal

This PR adds a flag / option to the extensions lib to specify the leader election resource lock (currently defaulted to `configmapsleases` just like in `c-r@v0.7.0`).

From the doc string:

> When changing the default resource lock, please make sure to migrate via multilocks to
avoid situations where multiple running instances of your controller have each acquired leadership
through different resource locks (e.g. during upgrades) and thus act on the same resources concurrently.
For example, if you want to migrate to the "leases" resource lock, you might do so by migrating
to the respective multilock first ("configmapsleases" or "endpointsleases"), which will acquire
a leader lock on both resources. After one release with the multilock as a default, you can
go ahead and migrate to "leases". Please also keep in mind, that users might skip versions
of your controller, so at least add a flashy release note when changing the default lock.
>
> Note: before controller-runtime version v0.7, the resource lock was set to "configmaps".
Please keep this in mind, when planning a proper migration path for your controller.

Fixes #

As you can see in https://github.com/gardener/gardener-extension-provider-aws/pull/263, there isn't really anything to do for extensions to leverage the new option (at least for now). Extensions might decide to change the default resource lock later on after a couple of releases.

**Release note**:

```feature dependency
The extensions library now allows to specify the leader election resource lock (currently defaulted to `configmapsleases`). Please read through the doc string of the respective field (`ManagerOptions.LeaderElectionResourceLock`) carefully before changing the default resource lock.
```

### DIFF
--- a/extensions/pkg/controller/cmd/options.go
+++ b/extensions/pkg/controller/cmd/options.go
@@ -18,19 +18,23 @@ import (
 	"fmt"
 	"os"
 
-	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
-
 	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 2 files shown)


timuthy (approved):
/lgtm

PR #3482, opened 2021-02-01
Add MaintenanceWindow to the Loki hvpa

Kristian-ZH:

/area logging
/kind enhancement
/priority normal

Add MaintenanceTimeWindow for scale down to Loki's hvpa in `garden` namespace.
This will prevent Loki from frequently scale downs.

Fixes #

@amshuman-kr @vlvasilev 

**Release note**:

```other operator
MaintenanceTimeWindow for scale-down is added to Loki's hvpa in garden namespace
```

### DIFF
--- a/pkg/operation/seed/seed.go
+++ b/pkg/operation/seed/seed.go
@@ -383,9 +383,6 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
 
 	if loggingEnabled {
 		lokiValues["authEnabled"] = false
-		lokiValues["hvpa"] = map[string]interface{}{
-			"enabled": hvpaEnabled,
-		}
 
 		lokiVpa := &autoscalingv1beta2.VerticalPodAutoscaler{ObjectMeta: metav1.ObjectMeta{Name: "loki-vpa", Namespace: v1beta1constants.GardenNamespace}}
 		if err := k8sSeedClient.Client().Delete(ctx, lokiVpa); client.IgnoreNotFound(err) != nil && !meta.IsNoMatchError(err) {
@@ -393,6 +390,26 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
 		}
 
 		if hvpaEnabled {
+			shootInfo := &corev1.ConfigMap{}
+			maintenanceBegin := "220000-0000"
+			maintenanceEnd := "230000-0000"
+			if err := k8sSeedClient.Client().Get(ctx, kutil.Key(metav1.NamespaceSystem, "shoot-info"), shootInfo); err != nil {
+				if !apierrors.IsNotFound(err) {
+					return err
+				}
+			} else {
+				maintenanceBegin = shootInfo.Data["maintenanceBegin"]
+				maintenanceEnd = shootInfo.Data["maintenanceEnd"]
+			}
+
+			lokiValues["hvpa"] = map[string]interface{}{
+				"enabled": true,
+				"maintenanceTimeWindow": map[string]interface{}{
+					"begin": maintenanceBegin,
+					"end":   maintenanceEnd,
+				},
+			}
+
 			currentResources, err := common.GetContainerResourcesInStatefulSet(ctx, k8sSeedClient.Client(), kutil.Key(v1beta1constants.GardenNamespace, "loki"))
 			if err != nil {
 				return err
--- a/charts/seed-bootstrap/charts/loki/templates/hvpa.yaml
+++ b/charts/seed-bootstrap/charts/loki/templates/hvpa.yaml
@@ -10,6 +10,10 @@ metadata:
 {{ toYaml .Values.labels | indent 4 }}
 spec:
   replicas: 1
+{{- if .Values.hvpa.maintenanceTimeWindow }}
+  maintenanceTimeWindow:
+{{ toYaml .Values.hvpa.maintenanceTimeWindow | indent 4 }}
+{{- end }}
   hpa:
     selector:
       matchLabels:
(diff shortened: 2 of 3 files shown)


pkg/operation/seed/seed.go
```diff
@@ -385,6 +385,10 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
 		lokiValues["authEnabled"] = false
 		lokiValues["hvpa"] = map[string]interface{}{
 			"enabled": hvpaEnabled,
+			"maintenanceTimeWindow": map[string]interface{}{
+				"begin": "220000-0000",
+				"end":   "230000-0000",
```
amshuman-kr: Does it make sense to synchronize this with the shooted seed's maintenance window? Is it even possible?

amshuman-kr (commented):
LGTM but for the minor question below.

pkg/operation/seed/seed.go
rfranzke: It makes sense if it's a shooted seed, yes. You can try reading the `kube-system/shoot-info` config map. If it exists then you'll find the maintenance time window in its data:

```yaml
apiVersion: v1
data:
  maintenanceBegin: 120000+0000
  maintenanceEnd: 130000+0000
  ...
kind: ConfigMap
metadata:
  annotations:
    resources.gardener.cloud/description: |-
      DO NOT EDIT - This resource is managed by gardener-resource-manager.
  creationTimestamp: "2019-12-10T12:10:25Z"
  labels:
    shoot.gardener.cloud/no-cleanup: "true"
  name: shoot-info
  namespace: kube-system
  resourceVersion: "307696014"
  selfLink: /api/v1/namespaces/kube-system/configmaps/shoot-info
  uid: 0c3404b0-1b46-11ea-894d-5e2e6521d6ea
```

, so you can take these values. If it does not exist then what about computing a random time window here?

timuthy:
Can you share more information about this change?

For instance:
> This will prevent Loki from frequently scale downs.

Now scale downs are disabled from 22:00 - 23:00 UTC. Is that exactly the critical time range when Loki was restarted frequently? If yes, why and not outside of this window? Is that true for every seed?

rfranzke:
> Now scale downs are disabled from

IIUIC, it's the other way round: Scale downs are only enabled/allowed during the maintenance time window. As for the concrete value (begin/end) I suggested https://github.com/gardener/gardener/pull/3482#discussion_r567797833. WDYT?

timuthy:
> > Now scale downs are disabled from
> 
> IIUIC, it's the other way round: Scale downs are only enabled/allowed during the maintenance time window. As for the concrete value (begin/end) I suggested [#3482 (comment)](https://github.com/gardener/gardener/pull/3482#discussion_r567797833). WDYT?

Okay, thanks :)

Kristian-ZH:
/hold

Let's hold this PR for now, because I took over it from @vlvasilev (He is DoD this week) and would like to spend some time to investigate the HVPA's behaviour.
What I was monitoring yesterday is that the Loki works with ~500-600mb memory, but the pod requested memory was 2gb.
I will continue to monitor it today and will try to find why the requested memory is much bigger than the used memory

rfranzke:
I've marked the PR as "draft" for now. Kindly mark it as "ready" again when you have done the changes.

Kristian-ZH:
I have added the maintenance time windows in the seed clusters based on the one in `kube-system/shoot-info` CM.

We decided to not add maintenanceTimeWindow in the shoots for now, because:
  1) This will make all Loki instances to scale down only once in 24h and they will consume much more resources.
  2) With this PR we increase MinChange (+ other stabilisation changes) values for all Loki instances which could be enough to 
       make them more stable

pkg/operation/seed/seed.go
```diff
@@ -383,16 +383,33 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
 		lokiVpa := &autoscalingv1beta2.VerticalPodAutoscaler{ObjectMeta: metav1.ObjectMeta{Name: "loki-vpa", Namespace: v1beta1constants.GardenNamespace}}
 		if err := k8sSeedClient.Client().Delete(ctx, lokiVpa); client.IgnoreNotFound(err) != nil && !meta.IsNoMatchError(err) {
 			return err
 		}
 
 		if hvpaEnabled {
+			shootInfo := &corev1.ConfigMap{}
+			maintenanceBegin := "220000-0000"
+			maintenanceEnd := "230000-0000"
+			if err := k8sSeedClient.Client().Get(ctx, kutil.Key(metav1.NamespaceSystem, "shoot-info"), shootInfo); err != nil {
```
amshuman-kr, line 396: If this is for the shooted seed case, shouldn't we use the shoot client instead of the seed client?

amshuman-kr (commented):
I might be misreading the code, but it looks like the hard-coded maintenance window is used for normal shoots. Is this intentional?

pkg/operation/seed/seed.go
amshuman-kr: I am totally confused 😅  This is the bootstrapping case, so `k8sSeedClient` seems right. I take back the comment (but I do not get the option to resolve it)

Kristian-ZH: The hardcoded maintenance window is when the `shoot-info` CM is not found in the seed cluster. In this case I think that it does not matter if the maintenance window is hardcoded or random-generated.

The maintenance window is added only for Loki in `garden` namespace in the seed cluster, e.g. `ShootedSeed` cluster.
We also have Lokis in the shoot-namespaces but for now I think that it will be better to not add the maintenance window on them.

timuthy (approved):
/lgtm

rfranzke (approved):

PR #3486, opened 2021-02-02
Use autoscaling/v2beta2 for seeds >= 1.12

timebertt:

/kind cleanup
/priority normal

`autoscaling/v2beta1` is deprecated starting with k8s v1.19: https://github.com/kubernetes/kubernetes/pull/90463
Use `autoscaling/v2beta2` when deleting HPAs in the Seed, where available (k8s >= v1.12).

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/operation/botanist/controlplane.go
+++ b/pkg/operation/botanist/controlplane.go
@@ -21,7 +21,26 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/gardener/gardener/pkg/utils/kubernetes/health"
+	hvpav1alpha1 "github.com/gardener/hvpa-controller/api/v1alpha1"
+	"github.com/sirupsen/logrus"
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv2beta1 "k8s.io/api/autoscaling/v2beta1"
+	autoscalingv2beta2 "k8s.io/api/autoscaling/v2beta2"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm

timebertt:
cc @danielfoehrKn

pkg/operation/botanist/controlplane.go
```diff
@@ -857,6 +850,21 @@ func (b *Botanist) DeployKubeAPIServer(ctx context.Context) error {
 			},
 		}
 
+		seedVersionGE112, err := version.CompareVersions(b.K8sSeedClient.Version(), ">=", "1.12")
```
timuthy, line 853:
```suggestion
		seedVersionGE112, err := version.CompareVersions(b.K8sSeedClient.Version(), ">=", "1.12-0")
```
IIRC, this is necessary to also support versions used by other providers e.g., `v1.18.14-gke.1200`.

timuthy (changes requested):

pkg/operation/botanist/controlplane.go
rfranzke: With https://github.com/gardener/gardener/pull/3444 this will be handled transparently so there is no need to have the `-0` suffix.

timuthy (approved):

pkg/operation/botanist/controlplane.go
timebertt: Thanks for checking the details :)

PR #3487, opened 2021-02-02
Add BackupBucketReady seed condition

### LINKED ISSUE #3428: Propagate BackupBucket creation errors to the Seed resource status

/area backup
/area storage
/area ops-productivity
/kind enhancement
/priority normal

If BuckupBucket creation fails (e.g. because of https://github.com/gardener/gardener-extension-provider-azure/issues/237), the error is reported in the `BackupBucket` status but the Seed itself is shown as ready. 

```sh
$ g k -- get backupbuckets XXX -ojson | jq '{ "name": .metadata.name, "spec": { "seedName": .spec.seedName }, "lastError": .status.lastError }'
{
  "name": "XXX",
  "spec": {
    "seedName": "SSS"
  },
  "lastError": {
    "description": "flow \"Backup Bucket Reconciliation\" encountered task errors: [task \"Waiting until backup bucket is reconciled\" failed: Error while waiting for BackupBucket /XXX to become ready: extension encountered error during reconciliation: Error reconciling backupbucket: storage.AccountsClient#Create: Failure sending request: StatusCode=403 -- Original Error: Code=\"RequestDisallowedByPolicy\" Message=\"Resource '' was disallowed by policy. Policy identifiers: '[{\\\"policyAssignment\\\":{\\\"name\\\":\\\"Storage accounts must use TLS Version 1.2 and above\\\"\"}]]"
  }
}

$ g k -- get seed SSS -o json | jq '{ "name": .metadata.name, "conditions": .status.conditions }'
{
  "name": "SSS",
  "conditions": [
    {
      "message": "All extensions installed into the seed cluster are ready and healthy.",
      "reason": "AllExtensionsReady",
      "status": "True",
      "type": "ExtensionsReady"
    },
    {
      "message": "Gardenlet is posting ready status.",
      "reason": "GardenletReady",
      "status": "True",
      "type": "GardenletReady"
    },
    {
      "message": "Seed cluster has been bootstrapped successfully.",
      "reason": "BootstrappingSucceeded",
      "status": "True",
      "type": "Bootstrapped"
    }
  ]
}
```

We should propagate the error to the Seed status that the seed is actually not healthy/ready.

**Why is this needed**:
If BackupBucket creation fails, shoot creation if scheduled on that seed will fail. So, might be useful to indicate this already in the Seed resource status.

timuthy:

/area ops-productivity
/kind enhancement
/priority normal

This PR adds a new condition type `BackupBucketReady` to `seed.status.conditions` in case the seed specifies a backup configuration. Please refer to the linked issue for more details.

Fixes #3428

/cc @amshuman-kr 

**Release note**:

```feature operator
`Seed` resources now have a new condition type `BackupBucketReady` that is added when the corresponding seed has a backup configuration. `Seeds` whose `BackupBucketReady` condition is `status: "False"` are considered `NotReady` and thus are excluded from scheduling during that time.
```

### DIFF
--- a/pkg/apis/core/v1beta1/helper/helper.go
+++ b/pkg/apis/core/v1beta1/helper/helper.go
@@ -20,22 +20,23 @@ import (
 	"strings"
 	"time"
 
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/json"
-	"k8s.io/utils/pointer"
-
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	versionutils "github.com/gardener/gardener/pkg/utils/version"
 
 	"github.com/Masterminds/semver"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // Now determines the current metav1.Time.
@@ -1195,4 +1196,16 @@ func KubernetesDashboardEnabled(addons *gardencorev1beta1.Addons) bool {
 // NginxIngressEnabled returns true if the nginx-ingress addon is enabled in the Shoot manifest.
 func NginxIngressEnabled(addons *gardencorev1beta1.Addons) bool {
 	return addons != nil && addons.NginxIngress != nil && addons.NginxIngress.Enabled
+}
+
+// ConstantConditionPatch creates a strategic merge patch with the the given conditions.
+func ConstantConditionPatch(conditions ...gardencorev1beta1.Condition) (client.Patch, error) {
+	patchData, err := json.Marshal(conditions)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal status for patch: %s", err)
+	}
+
+	data := fmt.Sprintf("{\"status\":{\"conditions\":%s}}", patchData)
+
+	return client.RawPatch(types.StrategicMergePatchType, []byte(data)), nil
 }
--- a/pkg/apis/core/v1beta1/types_seed.go
+++ b/pkg/apis/core/v1beta1/types_seed.go
@@ -311,6 +311,8 @@ const (
 	SeedExtensionsReady ConditionType = "ExtensionsReady"
 	// SeedGardenletReady is a constant for a condition type indicating that the Gardenlet is ready.
 	SeedGardenletReady ConditionType = "GardenletReady"
+	// SeedBackupBucketReady is a constant for a condition type indicating that the associated BackupBucket is ready.
+	SeedBackupBucketReady ConditionType = "BackupBucketReady"
 )
 
 // Resource constants for Gardener object types
--- a/pkg/gardenlet/controller/backupbucket/backup_bucket_reconciler.go
+++ b/pkg/gardenlet/controller/backupbucket/backup_bucket_reconciler.go
@@ -116,10 +116,18 @@ func (r *reconciler) reconcileBackupBucket(ctx context.Context, gardenClient kub
 		return reconcile.Result{}, err
 	}
 
-	seedClient, err := r.clientMap.GetClient(ctx, keys.ForSeedWithName(*backupBucket.Spec.SeedName))
+	seed := &gardencorev1beta1.Seed{}
+	if err := gardenClient.Client().Get(ctx, kutil.Key(*backupBucket.Spec.SeedName), seed); err != nil {
+		backupBucketLogger.Errorf("Failed to get seed: %+v", err)
+		return reconcile.Result{}, err
+	}
+
+	seedClient, err := r.clientMap.GetClient(ctx, keys.ForSeedWithName(seed.Name))
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("failed to get seed client: %w", err)
 	}
+
+	conditionBackupBucketReady := gardencorev1beta1helper.GetOrInitCondition(seed.Status.Conditions, gardencorev1beta1.SeedBackupBucketReady)
 
 	a := newActuator(gardenClient, seedClient, backupBucket, r.logger)
 	if err := a.Reconcile(ctx); err != nil {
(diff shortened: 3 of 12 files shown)


pkg/scheduler/controller/common/seed.go
```diff
@@ -27,5 +27,12 @@ func VerifySeedReadiness(seed *gardencorev1beta1.Seed) bool {
 	if cond := gardencorev1beta1helper.GetCondition(seed.Status.Conditions, gardencorev1beta1.SeedGardenletReady); cond == nil || cond.Status != gardencorev1beta1.ConditionTrue {
 		return false
 	}
+
+	// Only consider condition if it's found because we haven't maintained it in previous releases.
+	// TODO: Handle this condition more conservatively in the future.
+	if cond := gardencorev1beta1helper.GetCondition(seed.Status.Conditions, gardencorev1beta1.SeedBackupBucketReady); cond != nil && cond.Status != gardencorev1beta1.ConditionTrue {
```
vpnachev: Actually this condition is optional and should not be checked if `seed.spec.backup==nil`, right?

timuthy:
/hold
I just noticed that we rather need a strategic merge patch. I'll work on that.

pkg/scheduler/controller/common/seed.go
timuthy: Yes, that is also one of the reasons the condition is `cond != nil && cond.Status != gardencorev1beta1.ConditionTrue`.

timuthy: I added it in addition to make it clear.

pkg/gardenlet/controller/backupbucket/backup_bucket_reconciler.go
```diff
@@ -132,20 +140,40 @@ func (r *reconciler) reconcileBackupBucket(ctx context.Context, gardenClient kub
...
 
+	if updateErr := patchSeedCondition(ctx, gardenClient.Client(), seed, gardencorev1beta1helper.UpdatedCondition(conditionBackupBucketReady,
+		gardencorev1beta1.ConditionTrue, "BackupBucketAvailable", msg)); updateErr != nil {
+		return reconcile.Result{}, updateErr
+	}
+
 	return reconcile.Result{}, nil
 }
 
+func patchSeedCondition(ctx context.Context, c client.Client, seed *gardencorev1beta1.Seed, condition gardencorev1beta1.Condition) error {
```
amshuman-kr, line 168: Though, at present, there is one backup bucket per seed, technically, there could be more than one backup bucket per seed. Does it make sense to handle this case already?

amshuman-kr (commented):
Thanks for the PR @timuthy! Overall LGTM. Just one question below.

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -1196,3 +1197,15 @@ func KubernetesDashboardEnabled(addons *gardencorev1beta1.Addons) bool {
 func NginxIngressEnabled(addons *gardencorev1beta1.Addons) bool {
 	return addons != nil && addons.NginxIngress != nil && addons.NginxIngress.Enabled
 }
+
+// ConstantConditionPatch creates a strategic merge patch with the the given conditions.
```
vpnachev, line 1201:
```suggestion
// ConstantConditionPatch creates a strategic merge patch with the given conditions.
```

pkg/apis/core/v1beta1/types_seed.go
```diff
@@ -311,6 +311,8 @@ const (
 	SeedExtensionsReady ConditionType = "ExtensionsReady"
 	// SeedGardenletReady is a constant for a condition type indicating that the Gardenlet is ready.
 	SeedGardenletReady ConditionType = "GardenletReady"
+	// SeedBackupBucketReady is a constant for a condition type indicating that the associated BackupBucket is ready.
+	SeedBackupBucketReady ConditionType = "BackupBucketReady"
```
vpnachev, line 315: I am not completely sure, but it looks like these constants are alphabetically sorted. Could move this one to the top.

vpnachev (commented):
Just some minor nits.
Otherwise lgtm.

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -1196,3 +1197,15 @@ func KubernetesDashboardEnabled(addons *gardencorev1beta1.Addons) bool {
...
+func ConstantConditionPatch(conditions ...gardencorev1beta1.Condition) (client.Patch, error) {
+	patchData, err := json.Marshal(conditions)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal status for patch: %s", err)
+	}
+
+	data := fmt.Sprintf("{\"status\":{\"conditions\":%s}}", patchData)
```
rfranzke, line 1208:
```suggestion
	data := fmt.Sprintf(`{"status":{"conditions":%s}}`, patchData)
```

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -1196,3 +1197,15 @@ func KubernetesDashboardEnabled(addons *gardencorev1beta1.Addons) bool {
 func NginxIngressEnabled(addons *gardencorev1beta1.Addons) bool {
 	return addons != nil && addons.NginxIngress != nil && addons.NginxIngress.Enabled
 }
+
+// ConstantConditionPatch creates a strategic merge patch with the the given conditions.
+func ConstantConditionPatch(conditions ...gardencorev1beta1.Condition) (client.Patch, error) {
```
rfranzke, line 1202: Does it make sense to have the patch type as a parameter instead of hard-coding it to strategic-merge-patch?

```suggestion
func ConstantConditionPatch(patchType types.PatchType, conditions ...gardencorev1beta1.Condition) (client.Patch, error) {
```

pkg/scheduler/controller/common/common_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestCommon(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Common Suite")
```
rfranzke, line 26:
```suggestion
	RunSpecs(t, "Scheduler Controller Common Suite")
```

pkg/scheduler/controller/common/seed.go
```diff
@@ -27,5 +27,14 @@ func VerifySeedReadiness(seed *gardencorev1beta1.Seed) bool {
 	if cond := gardencorev1beta1helper.GetCondition(seed.Status.Conditions, gardencorev1beta1.SeedGardenletReady); cond == nil || cond.Status != gardencorev1beta1.ConditionTrue {
 		return false
 	}
+
+	if seed.Spec.Backup != nil {
+		// Only consider condition if it's found because we haven't maintained it in previous releases.
+		// TODO: Handle this condition more conservatively in the future
```
rfranzke, line 33: What does "more conservatively" mean? 👀

pkg/gardenlet/controller/backupbucket/backup_bucket_reconciler.go
rfranzke: Good point @amshuman-kr, I guess it makes sense. A `BackupBucket` resource can (not today, but will hopefully be implemented in the future as part of GEP-7?, cc @plkokanov) be moved to a new seed, and this new seed will then have > 1 associated `BackupBucket`s.

rfranzke (commented):
I have some doubts about the implementation. Isn't it "wrong" that the main controller reconciling `BackupBucket`s is now suddenly writing a condition into a different resource (`Seed`)?

I would rather propose to introduce a new "secondary" controller in the `pkg/gardenlet/controller/seed` package which watches `BackupBucket`s and enqueues the referenced `Seed` when an event is triggered. Then, this `Seed` is reconciled (via a new, smaller reconciliation function (not the big one which also bootstraps)) in the sense that all `BackupBucket`s referenced it are checked for readiness. Then it can maintain your new condition in the `Seed`'s `status`. This will also solve @amshuman-kr's comment https://github.com/gardener/gardener/pull/3487/files#r568752939.

timuthy:
Thanks for your reviews. I'll incorporate your suggestions this week.

timuthy:
> I would rather propose to introduce a new "secondary" controller in the `pkg/gardenlet/controller/seed` package which watches `BackupBucket`s and enqueues the referenced `Seed` when an event is triggered. Then, this `Seed` is reconciled (via a new, smaller reconciliation function (not the big one which also bootstraps)) in the sense that all `BackupBucket`s referenced it are checked for readiness. Then it can maintain your new condition in the `Seed`'s `status`. This will also solve @amshuman-kr's comment https://github.com/gardener/gardener/pull/3487/files#r568752939.

I was wondering if we created such a second controller, why we'd want to have it in the Gardenlet instead of the Gardener-Controller-Manager? The controller in GCM can watch all BBs in the Garden cluster and then let an additional, small reconcile loop checking if BBs related to a Seed are healthy. Can you think of any technical disadvantage @rfranzke @amshuman-kr?

timuthy:
I'm closing this PR in favor of #3531 which implements the controller approach suggested by @rfranzke.

Issue #2143, opened 2019-06-11
Run all `make` commands in container.


mvladev:
All `make` commands require various dev resources to be available:

- `golang`
- `helm`
- various `gnu` utils  such as `tar`, `gzip`, `tr`, `base64`, `cut`, `find`, `grep`, `xargs`.
- `gingo`
- linters.

I think that the most reasonable way to get reproducible builds and things such as `make generate` is to run those commands in a pre-build container image.

poelzi:
I highly suggest using nix and a proper shell/default.nix for the build process.
https://nixos.org/explore.html
[niv](https://github.com/nmattia/niv) allows you to have proper dependencies across different operating systems.

voelzmo:
Closing in favor of #4519 to keep the conversation in a single place. Thanks for raising this issue, @mvladev!

PR #3489, opened 2021-02-02
Speed-up make generate

### LINKED ISSUE #3388: make generate takes very long

/area dev-productivity
/kind enhancement
/priority normal

Make `make generate` faster.

**Ideas**
I think, [controller-tools](https://github.com/kubernetes-sigs/controller-tools) might be a bit faster than [code-generator](https://github.com/kubernetes/code-generator).
We could try, switching to controller-tools for the generations it supports.

Another idea is to parallelize different code generations. 
On my machine, `make generate` consistently utilizes about 50% cpu.
So at least for local runs, the time can maybe be split by half by parallelization.

**Why is this needed**:

`make generate` takes forever.
This is a very bad developer experience and also costs countless hours of CI time.

stoyanr:

/area dev-productivity
/kind enhancement
/priority normal

Adds `generate-parallel` make target and corresponding scripts to run code generation in parallel in order to speed it up. The original `generate` target is for the moment preserved for easy comparison. Right now this should be considered a draft to allow testing and feedback, eventually the old sequential target and scripts might be removed in favor of the parallel ones.

Fixes #3388

* The new scripts are based on [GNU Parallel](https://www.gnu.org/software/parallel/), run `brew install parallel` to install on Mac
* Run `make generate-parallel` for the new parallel code generation, `make generate` for the sequential one.
* GNU parallel automatically limits the number of parallel jobs to the number of cores on your laptop. Right now this is not configurable, not sure if it should be.
* GNU parallel ensures that the output of parallel jobs is not garbled. This means that the output of the entire job is presented after the job has finished. 

On my laptop (2,7 GHz Quad-Core Intel Core i7, 16 GB RAM), `make generate` takes about 16:50 min, while `make generate-parallel` takes about 11:10 min, an improvement of about 35%. Further gains by additional parallelisation might be possible but would not be substantial (in the range of 10-20 seconds) and would require significant changes to the current code generation. All my 8 cores are already at 100% for about 90% of the above 11:10 min.

Reviewers, could you please post execution times on your laptops?

**Release note**:

```other operator
NONE
```


stoyanr:
Further info if you are interested:

* Protobuf generation should happen first and should not run in parallel to anything else, or else there are errors. Also, generating protobuf for the 4 API packages in parallel is both slower and leads to incorrectly generated code. Therefore `hack/update-protobuf.sh` should not be touched, and is already rather fast (~30 sec on my laptop).
* Generation of API groups within a package can't be parallelized, e.g. attempting to run `deepcopy,defaulters` in parallel to `conversion` leads to errors. But since there are 9 packages + openapi, running these in parallel is possible. So the new `hack/update-codegen-parallel.sh` scripts does exactly this.
* Running all other `go generate` tasks in parallel is possible. So the new `hack/generate-parallel.sh` script discovers all packages that have `//go:generate` and runs `go generate` on them in parallel. Note that this does not include `hack/update-protobuf.sh` and `hack/update-codegen.sh` - the `go:generate` directives for these 2 scripts have been removed and they are now invoked by the `Makefile` directly instead.

hack/generate-parallel.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
vpnachev:
```suggestion
# Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

hack/update-codegen-parallel.sh
```diff
@@ -0,0 +1,248 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
vpnachev:
```suggestion
# Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

hack/generate-parallel.sh
```diff
@@ -0,0 +1,37 @@
...
+ROOTS=${ROOTS:-$(git grep --files-with-matches -e '//go:generate' cmd extensions pkg plugin test | \
+	xargs -n 1 dirname | \
+	sed 's,^,github.com/gardener/gardener/,;' | \
+	sort | uniq
+)}
+popd > /dev/null
+
+read -ra PACKAGES <<< $(echo ${ROOTS})
+
+GO111MODULE=on parallel go generate -mod=vendor {} ::: ${PACKAGES[@]}
```
vpnachev: Add some info how to install this tool in https://github.com/gardener/gardener/blob/master/docs/development/local_setup.md. You can also add it to https://github.com/gardener/gardener/blob/master/hack/install-requirements.sh.

If you enable the CI/CD to benefit from the parallel execution, you also have to double check if it is required here https://github.com/gardener/gardener/blob/master/.ci/verify#L13

vpnachev (commented):
I took a quick look on the changes.

For me it failed at first after around 4 minutes with
```bash
$ make generate-parallel
...
I0202 20:40:08.705714    8245 main.go:229] using package=github.com/gardener/gardener/pkg/apis/seedmanagement/v1alpha1
W0202 20:40:08.724796    8245 main.go:423] not found external link source for type k8s.io/apimachinery/pkg/runtime.RawExtension
W0202 20:40:08.726634    8245 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.Condition
W0202 20:40:08.726943    8245 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedSpec
W0202 20:40:08.727006    8245 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedBackup
W0202 20:40:08.727070    8245 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedDNS
W0202 20:40:08.727130    8245 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedNetworks
W0202 20:40:08.727190    8245 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedProvider
W0202 20:40:08.727408    8245 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedTaint
W0202 20:40:08.727471    8245 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedVolume
W0202 20:40:08.727539    8245 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedSettings
W0202 20:40:08.727594    8245 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.Ingress
I0202 20:40:08.728977    8245 main.go:165] written to ../../../../hack/api-reference/seedmanagement.md
# github.com/gardener/gardener/pkg/mock/go/context
../../../../go/context/funcs.go:1:1: syntax error: package statement must be first
```

The next attempt was successful and  took 4 minutes and 17 seconds.
The old `make generate` took me around 8 minutes and 29 seconds.

hack/generate-parallel.sh
```diff
@@ -0,0 +1,37 @@
...
+CURRENT_DIR="$(dirname $0)"
+PROJECT_ROOT="${CURRENT_DIR}"/..
+if [ "${PROJECT_ROOT#/}" == "${PROJECT_ROOT}" ]; then
+  PROJECT_ROOT="./$PROJECT_ROOT"
+fi
+
+pushd "$PROJECT_ROOT" > /dev/null
+ROOTS=${ROOTS:-$(git grep --files-with-matches -e '//go:generate' cmd extensions pkg plugin test | \
+	xargs -n 1 dirname | \
+	sed 's,^,github.com/gardener/gardener/,;' | \
```
rfranzke, line 30: These scripts in the `hack` dir are also used by extensions, so this here should probably be configurable.

rfranzke (commented):
I executed `make generate-parallel`, but it failed after `18m22s` with

```
../../../../go/context/funcs.go:1:1: expected 'package', found 'EOF'
../../../mock/go/context/funcs.go:1:1: expected 'package', found 'EOF'
prog.go:14:2: cannot find module providing package github.com/gardener/gardener/pkg/client/kubernetes: working directory is not part of a module
prog.go:12:2: cannot find module providing package github.com/golang/mock/mockgen/model: working directory is not part of a module
2021/02/03 08:27:33 Loading input failed: exit status 1
pkg/mock/gardener/client/kubernetes/doc.go:18: running "mockgen": exit status 1
make: *** [generate-parallel] Error 1
```

There were also intermediate "issues" earlier, but I *think* I've seen them also earlier and they are not causing harm (?):

```
2021/02/03 08:12:30 gopathwalk: scanning directory /Users/d065849/go/src/github.com/gardener/gardener: readdirent: interrupted system call
2021/02/03 08:12:30 gopathwalk: scanning directory /Users/d065849/go/src/github.com/gardener/gardener/vendor: readdirent: interrupted system call
2021/02/03 08:14:42 gopathwalk: scanning directory /Users/d065849/go/src/github.com/gardener/gardener: readdirent: interrupted system call
2021/02/03 08:14:42 gopathwalk: scanning directory /Users/d065849/go/src/github.com/gardener/gardener/vendor: readdirent: interrupted system call
```

hack/generate-parallel.sh
stoyanr: OK

hack/update-codegen-parallel.sh
stoyanr: OK

hack/generate-parallel.sh
stoyanr: I will do this, I intentionally didn't bother at this stage since I wanted to collect feedback first.

hack/generate-parallel.sh
stoyanr: Yes, in fact I thought we should perhaps preserve the original `generate` script as it is for use by extensions, and use `generate-parallel.sh` only here. For extensions code generation is fast enough from my perspective and there is little to be gained by parallelisation, further more as pointed out above it's by no means guaranteed that whatever you put in `//go:generate` directives can be executed in parallel. I had to replace some `go:generate` by explicit invocations in the `Makefile` to make it work.

rfranzke: OK, true, makes sense. There is not much to gain and it's not likely that this will change in the near future.

stoyanr:
@vpnachev 

> ```shell
> W0202 20:40:08.724796    8245 main.go:423] not found external link source for type k8s.io/apimachinery/pkg/runtime.RawExtension
> ```

The "not found external link" errors are present also with the old code generation, I guess it's a flaw in the documentation generation of the new `seedmanagement` package, will take a look at this but will not address it in this PR.

> ```shell
> ../../../../go/context/funcs.go:1:1: syntax error: package statement must be first
> ```

Haven't seen such errors, even though I ran `make generate-parallel` multiple times yesterday. Do you have an idea what might be causing them? How often do you see them, you mentioned that it worked for you on the second run?

stoyanr:
@rfranzke

> ```shell
> ../../../../go/context/funcs.go:1:1: expected 'package', found 'EOF'
> ```

Haven't seen anything like this either. Is it every run or it occurs sporadically like the error @vpnachev reported above?

I will continue running `make generate-parallel` today to see if I could reproduce any of these issues.

stoyanr:
@vpnachev @rfranzke The errors that you are reporting are `mockgen` errors, to check how often you are able to reproduce them you could run just `hack/generate-parallel.sh` instead of `make generate-parallel`, it's significantly faster (2:20 min vs. 11:10 min on my machine). Or if you are not able to reproduce them via just ` hack/generate-parallel.sh` that would also be an interesting finding.

Before running the code generation, make sure you don't have any intermediate output from old failed or forcefully interrupted runs, by running `git status` and deleting any garbage that might be there. Such output may cause `mockgen` to fail, happened a few times on my side but I am not sure if the errors were the same as you reported.

vpnachev:
> The "not found external link" errors are present also with the old code generation ..

If this is true, we should see it often in the CI/CD that runs the code generation as part of each PR verification, but I am note aware of any occurrence of the problem there.

> Do you have an idea what might be causing them?

No.

> How often do you see them, you mentioned that it worked for you on the second run?

I have run it twice, as I said, the first run failed and the second one succeeded as I have not changed anything.

stoyanr:
> If this is true, we should see it often in the CI/CD that runs the code generation as part of each PR verification, but I am note aware of any occurrence of the problem there.

The `seedmanagement` package was only recently introduced. I can definitely see these errors when I run `make generate`, you could try it out (or you could run just `hack/generate.sh` while on this PR for a much faster try). 

> I have run it twice, as I said, the first run failed and the second one succeeded as I have not changed anything.

I got this, I was hoping you could run `hack/generate-parallel.sh` a few more time (should take < 1 min on your much faster laptop) and tell me if you were able to reproduce them.

stoyanr:
@vpnachev @rfranzke 

I ran `hack/generate-parallel.sh` 5 times in a row now without seeing any errors (and yesterday I ran it at least 10 times, perhaps more). I am afraid I won't be able to solve the issues that you see without further input from your side.

rfranzke:
Interesting, I just ran it again and it failed during the first try with the same error. 😕 Here is the full log:

```
$ ./hack/generate-parallel.sh
Academic tradition requires you to cite works you base your article on.
If you use programs that use GNU Parallel to process data for an article in a
scientific publication, please cite:

  Tange, O. (2020, November 22). GNU Parallel 20201122 ('Biden').
  Zenodo. https://doi.org/10.5281/zenodo.4284075

This helps funding further development; AND IT WON'T COST YOU A CENT.
If you pay 10000 EUR you should feel free to use GNU Parallel without citing.

More about funding GNU Parallel and the citation notice:
https://www.gnu.org/software/parallel/parallel_design.html#Citation-notice

To silence this citation notice: run 'parallel --citation' once.

I0203 09:15:05.852136   26870 main.go:127] parsing go packages in directory .
I0203 09:15:31.337514   26870 main.go:229] using package=github.com/gardener/gardener/pkg/apis/settings/v1alpha1
I0203 09:15:31.419848   26870 main.go:165] written to ../../../../hack/api-reference/settings.md
I0203 09:15:05.174287   26860 main.go:127] parsing go packages in directory .
I0203 09:16:39.277234   26860 main.go:229] using package=github.com/gardener/gardener/pkg/apis/core/v1beta1
I0203 09:16:39.406902   26860 main.go:165] written to ../../../../hack/api-reference/core.md
I0203 09:15:05.310606   26861 main.go:127] parsing go packages in directory .
I0203 09:16:39.615061   26861 main.go:229] using package=github.com/gardener/gardener/pkg/apis/extensions/v1alpha1
I0203 09:16:39.728224   26861 main.go:165] written to ../../../../hack/api-reference/extensions.md
I0203 09:15:05.417525   26862 main.go:127] parsing go packages in directory .
I0203 09:16:44.377585   26862 main.go:229] using package=github.com/gardener/gardener/pkg/apis/seedmanagement/v1alpha1
W0203 09:16:44.420875   26862 main.go:423] not found external link source for type k8s.io/apimachinery/pkg/runtime.RawExtension
W0203 09:16:44.422506   26862 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.Condition
W0203 09:16:44.422793   26862 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedSpec
W0203 09:16:44.422861   26862 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedBackup
W0203 09:16:44.422932   26862 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedDNS
W0203 09:16:44.423014   26862 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedNetworks
W0203 09:16:44.423074   26862 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedProvider
W0203 09:16:44.423271   26862 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedTaint
W0203 09:16:44.423358   26862 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedVolume
W0203 09:16:44.423422   26862 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.SeedSettings
W0203 09:16:44.423481   26862 main.go:423] not found external link source for type github.com/gardener/gardener/pkg/apis/core/v1beta1.Ingress
I0203 09:16:44.424680   26862 main.go:165] written to ../../../../hack/api-reference/seedmanagement.md
../../../../go/context/mocks.go:1:1: expected 'package', found 'EOF'
../../../mock/go/context/mocks.go:1:1: expected 'package', found 'EOF'
prog.go:14:2: cannot find module providing package github.com/gardener/gardener/pkg/client/kubernetes: working directory is not part of a module
prog.go:12:2: cannot find module providing package github.com/golang/mock/mockgen/model: working directory is not part of a module
2021/02/03 09:19:39 Loading input failed: exit status 1
pkg/mock/gardener/client/kubernetes/doc.go:18: running "mockgen": exit status 1
```

stoyanr:
@rfranzke I am rather clueless, so shooting in the dark ... Are you on Linux or Mac? Could you please post the output of `go env`? Could you also please try `hack/generate-parallel.sh` instead of `./hack/generate-parallel.sh` (though `./hack/generate-parallel.sh` works for me as well)? Could you also put `set -x` a the beginning of the `hack/generate-parallel.sh` script, execute it, and send me the output?

hack/update-codegen-parallel.sh
```diff
@@ -0,0 +1,248 @@
...
+    --input-dirs=k8s.io/apimachinery/pkg/version \
+    --input-dirs=k8s.io/apimachinery/pkg/runtime \
+    --input-dirs=k8s.io/apimachinery/pkg/util/intstr \
+    --report-filename=${PROJECT_ROOT}/pkg/openapi/api_violations.report \
+    --output-package=github.com/gardener/gardener/pkg/openapi \
+    -h "${PROJECT_ROOT}/hack/LICENSE_BOILERPLATE.txt"
+}
+export -f openapi_definitions
+
+parallel ::: \
```
timebertt: On my machine, this seems to do nothing at all, while substituting `:::` with `--` works perfectly fine. 🤔

timebertt (commented):
On my machine, I'm getting an even more obscure behavior: `update-codegen-parallel.sh` and `generate-parallel.sh` don't seem to be generating anything at all...

hack/update-codegen-parallel.sh
rfranzke: Did you run `brew install parallel`?

timebertt: Yes, but I already had `parallel` installed as part of the `moreutils` package

stoyanr: Perhaps you have a different version of `parallel`? Though according to the documentation, `:::` is what should work, not `--`.

My version is:
```
$ parallel --version
GNU parallel 20201122
...
```

timebertt: Ok, just confirmed that those are two different versions, that are not compatible with regards to argument handling...
```
$ parallel --version
/usr/local/Cellar/moreutils/0.63/libexec/bin/parallel: illegal option -- -
parallel [OPTIONS] command -- arguments
	for each argument, run command with argument, in parallel
parallel [OPTIONS] -- commands
	run specified commands in parallel
```
😄 

When unlinking `moreutils` and linking `parallel` instead, it works...
This cries out for https://github.com/gardener/gardener/issues/2143

stoyanr:
Together with @rfranzke we found that `github.com/gardener/gardener/pkg/mock/go/context` was erroneously used in some production code (instead of `context`). We believe this was causing all the issues. I found 2 such usages and fixed them, now it should work without any errors.

@vpnachev Could you try on your side?

Also, I added some debugging output to the `hack/generate-parallel.sh` script, should be useful if there are further errors, will remove it before declaring the PR ready to be merged.

rfranzke:
I tried it again (`make generate-parallel`) and can confirm that it's working now for me. Thanks!

timebertt:
Now, that I use the proper `parallel` version, `update-codegen-parallel.sh` works. 
But `gen-crd-api-reference-docs` fails for me. Not sure what the exact problem is here: 
```
$ make generate-parallel
...
Generate github.com/gardener/gardener/pkg/apis/extensions/v1alpha1
I0203 12:07:36.821747    3265 main.go:127] parsing go packages in directory .
W0203 12:07:37.103013    3265 parse.go:247] Ignoring child directory github.com/gardener/gardener/pkg/apis/extensions/v1alpha1/helper: unable to import "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1/helper": go/build: go list github.com/gardener/gardener/pkg/apis/extensions/v1alpha1/helper: exit status 2
go: cannot find GOROOT directory: /usr/local/Cellar/go/1.15.5/libexec

I0203 12:07:37.106456    3265 main.go:229] using package=github.com/gardener/gardener/pkg/apis/extensions/v1alpha1
F0203 12:07:37.135378    3265 main.go:457] type invalid type has kind=Unsupported which is unhandled
pkg/apis/extensions/v1alpha1/doc.go:17: running "gen-crd-api-reference-docs": exit status 255
Generate github.com/gardener/gardener/pkg/apis/settings/v1alpha1
I0203 12:07:36.821728    3267 main.go:127] parsing go packages in directory .
I0203 12:07:37.267630    3267 main.go:229] using package=github.com/gardener/gardener/pkg/apis/settings/v1alpha1
F0203 12:07:37.310036    3267 main.go:457] type invalid type has kind=Unsupported which is unhandled
pkg/apis/settings/v1alpha1/doc.go:22: running "gen-crd-api-reference-docs": exit status 255
Generate github.com/gardener/gardener/pkg/apis/seedmanagement/v1alpha1
I0203 12:07:36.822022    3274 main.go:127] parsing go packages in directory .
I0203 12:07:37.641611    3274 main.go:229] using package=github.com/gardener/gardener/pkg/apis/seedmanagement/v1alpha1
F0203 12:07:37.666994    3274 main.go:457] type invalid type has kind=Unsupported which is unhandled
pkg/apis/seedmanagement/v1alpha1/doc.go:22: running "gen-crd-api-reference-docs": exit status 255
Generate github.com/gardener/gardener/pkg/apis/core/v1beta1
I0203 12:07:36.822146    3266 main.go:127] parsing go packages in directory .
W0203 12:07:38.755144    3266 parse.go:247] Ignoring child directory github.com/gardener/gardener/pkg/apis/core/v1beta1/constants: unable to import "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants": go/build: go list github.com/gardener/gardener/pkg/apis/core/v1beta1/constants: exit status 2
go: cannot find GOROOT directory: /usr/local/Cellar/go/1.15.5/libexec

W0203 12:07:38.782015    3266 parse.go:247] Ignoring child directory github.com/gardener/gardener/pkg/apis/core/v1beta1/helper: unable to import "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper": go/build: go list github.com/gardener/gardener/pkg/apis/core/v1beta1/helper: exit status 2
go: cannot find GOROOT directory: /usr/local/Cellar/go/1.15.5/libexec

I0203 12:07:38.820961    3266 main.go:229] using package=github.com/gardener/gardener/pkg/apis/core/v1beta1
F0203 12:07:38.852202    3266 main.go:457] type invalid type has kind=Unsupported which is unhandled
pkg/apis/core/v1beta1/doc.go:22: running "gen-crd-api-reference-docs": exit status 255
```

stoyanr:
@timebertt 

> W0203 12:07:37.103013    3265 parse.go:247] Ignoring child directory github.com/gardener/gardener/pkg/apis/extensions/v1alpha1/helper: unable to import "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1/helper": go/build: go list github.com/gardener/gardener/pkg/apis/extensions/v1alpha1/helper: exit status 2
go: cannot find GOROOT directory: /usr/local/Cellar/go/1.15.5/libexec

Seems like an issue with your Go setup? What does `go env` say? Do you indeed have a `/usr/local/Cellar/go/1.15.5/libexec` directory? Do you see such issues only with `make generate-parallel` (or `hack/generate-parallel.sh` which should be the same in your case)? You could use `hack/generate.sh ./cmd/... ./extensions/... ./pkg/... ./plugin/... ./test/...` to check the old generation while on the PR without `protobuf` and `codegen`, which is much faster.

timebertt (commented):
Indeed, my problem was not related to the changes in this PR.
It seems like, one hast to reinstall `gen-crd-api-reference-docs` when updating the go version to get rid of the error message above.
We can remove the tool before invoking it similar to the k8s generators [here](https://github.com/gardener/gardener/blob/81105aed79ce2107e4c37dc169e70921409510af/hack/update-codegen.sh#L21). Will do it in a separate PR.

With `make generate-parallel`, I now had a successful run within `8.53m` and was able to achieve a consistently high CPU usage.
The old `make generate` took about `22.30m`.
(running macOS 10.15 with an 2,3 GHz 8-Core i9 and 32GB).

> Also, I added some debugging output to the `hack/generate-parallel.sh` script, should be useful if there are further errors, will remove it before declaring the PR ready to be merged.

I would actually vote for keeping the `Generate <pkg>` output. I think, it's useful to see what's going on.

vpnachev:
Could you enhance the `make clean` so that we can easily find files that seems to be no longer generated?
You can apply this patch (also double check if I missed some other generated file).
```patch
--- a/hack/clean.sh
+++ b/hack/clean.sh
@@ -20,5 +20,14 @@ echo "> Clean"

 for source_tree in $@; do
   find "$(dirname "$source_tree")" -type f -name "zz_*.go" -exec rm '{}' \;
-  grep -lr --include="*.go" "//go:generate packr2" . | xargs -I {} packr2 clean "{}/.."
+  find "$(dirname "$source_tree")" -type f -name "generated.proto" -exec rm '{}' \;
+  find "$(dirname "$source_tree")" -type f -name "generated.pb.go" -exec rm '{}' \;
+  find "$(dirname "$source_tree")" -type f -name "openapi_generated.go" -exec rm '{}' \;
+  grep -lr '// Code generated by MockGen. DO NOT EDIT' "$(dirname "$source_tree")" | xargs rm -f
+  grep -lr '// Code generated by client-gen. DO NOT EDIT' "$(dirname "$source_tree")" | xargs rm -f
+  grep -lr '// Code generated by informer-gen. DO NOT EDIT' "$(dirname "$source_tree")" | xargs rm -f
+  grep -lr '// Code generated by lister-gen. DO NOT EDIT' "$(dirname "$source_tree")" | xargs rm -f
+  grep -lr --include="*.go" "//go:generate packr2" "$(dirname "$source_tree")" | xargs -I {} packr2 clean "{}/.."
 done
+
+find ./hack/api-reference/ -type f -name "*.md" -exec rm '{}' \;
```

Based on the above, I found that https://github.com/gardener/gardener/blob/master/pkg/mock/gardener/utils/errors/funcs.go is no longer generated with both parallel and sequential generation, but would be helpful to find whether the parallel generation is properly generating everything that is required.

stoyanr:
I tried if it could be possible to generate mocks in parallel with API groups, in an attempt to further increase the parallelism that could yield further speed gains if the machine has more than 8 cores, since `hack/update-codegen-parallel.sh` starts only 10 parallel processes. I found a way to do it with only very small changes to the current scripts.

Unfortunately, this doesn't work. Since mock generation tries to load all the dependencies of the mocked packages, they inevitably load also some API packages, and if they are currently being generated, this leads to random errors of the sort:
```
# github.com/gardener/gardener/pkg/gardenlet/apis/config/v1alpha1
../../../../../gardenlet/apis/config/v1alpha1/zz_generated.conversion.go:1:1: syntax error: package statement must be first
...
Generate github.com/gardener/gardener/pkg/mock/gardener/client/kubernetes
../../../../../apis/settings/v1alpha1/zz_generated.conversion.go:26:2: cannot find package
../../../../../gardenlet/apis/config/v1alpha1/zz_generated.conversion.go:1:1: expected 'package', found 'EOF'
```

So finally, the code generation must happen exactly in the order currently proposed:
* First protobuf without any parallelism, `hack/update-protobuf.sh`
* Then API groups and OpenAPI definitions, max 10 jobs since tasks within a package can't be parallelized, `hack/update-codegen-parallel.sh`
* Then `go generate` for all remaining packages, mostly mocks and documentation, `hack/generate-parallel.sh`

stoyanr:
Added a new commit that:
* Addresses code review feedback, including changes to `hack/clean.sh`
* Removes the `generate-parallel` make target in favor of a parallel `generate`
* Adds generation for `pkg/mock/gardener/utils/errors/funcs.go`, this mock is still used but for some reason was no longer generated.

Note that `hack/generate.sh` is preserved unchanged since it's used by extensions, see https://github.com/gardener/gardener/pull/3489#discussion_r569191267.

I consider this PR ready to be merged now, please let me know if there is further feedback.

/cc @vpnachev @timebertt @rfranzke

Makefile
```diff
@@ -161,7 +161,9 @@ check:
 
 .PHONY: generate
 generate:
-	@hack/generate.sh ./cmd/... ./extensions/... ./pkg/... ./plugin/... ./test/...
+	@hack/update-protobuf.sh
+	@hack/update-codegen.sh
+	@hack/generate-parallel.sh
```
timebertt: Nit: can you move the package roots list from `hack/generate-parallel.sh` to here and pass it as an argument similar to the other make targets?
This way, when adding a new package root, one directly knows they should add it here, too.
```suggestion
	@hack/generate-parallel.sh cmd extensions pkg plugin test
```

timebertt (dismissed):
/lgtm
One really minor nit.

Makefile
stoyanr: Sure, I'll do it.

ialidzhikov:
I see some reports that code-generator is extremely slow with `GO111MODULE=on` - see https://github.com/kubernetes/code-generator/issues/69. Which is also the case for us: https://github.com/gardener/gardener/blob/3def7440925abab5f8c5b46d0d6fef51bcbe1c26/hack/generate.sh#L21

In this PR https://github.com/jenkins-x/lighthouse/pull/944 someone reported that disabling `GO111MODULE` dropped from 5m37s to 7s. 

Before trying out parallel, did you tried to set `GO111MODULE=off`?

stoyanr:
> Before trying out parallel, did you tried to set GO111MODULE=off?

I don't see how `GO111MODULE` could have any major impact. About 80% of the time, before and after parallel, is spent in `hack/update-codegen.sh`, and there is no difference if it's invoked via `go generate` (before my changes) or directly from the `Makefile` (after my changes). Perhaps it could have some impact on the speed of the multiple `mockgen` invocations, will try it out, but it's bound to be limited to <20%.

ialidzhikov:
With 

```diff
diff --git a/hack/generate.sh b/hack/generate.sh
index 0b3338a1c..d7695690b 100755
--- a/hack/generate.sh
+++ b/hack/generate.sh
@@ -18,4 +18,4 @@ set -e

 echo "> Generate"

-GO111MODULE=on go generate -mod=vendor $@
+GO111MODULE=off go generate $@
```

Result:

```
$ time make generate

real	9m12.326s
user	8m6.663s
sys	11m38.209s
```

And the code-gen was quite lightweight, I believe most of the time was spent in gomock.

@stoyanr , can you try out the above patch **on the master branch** and share if it fixes the issue for you?

stoyanr:
@ialidzhikov Is this on `master` or on this PR?

ialidzhikov:
> @ialidzhikov Is this on `master` or on this PR?

master

stoyanr:
@ialidzhikov And what were your times before the change?

stoyanr:
@ialidzhikov You are right, I can't believe my eyes:
```
make generate  332.32s user 322.72s system 240% cpu 4:32.02 total
```

It seems `GO111MODULE` is somehow implicitly set to `on`, because even if I execute `hack/update-codegen.sh` directly (something I did many times already), I get the slow times (>12 min on my laptop), while if I execute it like this:
```
time GO111MODULE=off hack/update-codegen.sh
```
I get:
```
GO111MODULE=off hack/update-codegen.sh  124.69s user 211.08s system 361% cpu 1:32.96 total
```

This also means with parallel execution the times are likely to drop to 1-2 min on my laptop. Let me try it out, and let me also confirm everything is ok by running the improved `make clean` in this PR.

ialidzhikov:
> This also means with parallel execution the times are likely to drop to 1-2 min on my laptop. Let me try it out, and let me also confirm everything is ok by running the improved make clean in this PR.

I don't think we need parallel anymore. I haven't seen it on any upstream project and does not bring much sense to me.

stoyanr:
> I don't think we need parallel anymore. I haven't seen it on any upstream project and does not bring much sense to me.

Well, I disagree. I would say we need at least a `generate-parallel` so people could use it if they want. I would still wait 2 mins rather than 4. Also, the change is really small and I don't see any harm in having it.

ialidzhikov:
> @ialidzhikov And what were your times before the change?

Here are my times on my >= 4 years old macbook with i7 and 16GB RAM.

On current master:

```
$ time make generate

real	26m51.982s
user	29m0.168s
sys	46m7.292s
```

On master with the patch from https://github.com/gardener/gardener/pull/3489#issuecomment-773863621

```
$ time make generate

real	9m12.326s
user	8m6.663s
sys	11m38.209s
```

stoyanr:
OK, with parallel execution, the times about halved as I suspected:
```
make generate  514.90s user 377.43s system 634% cpu 2:20.70 total
```
I also checked with `make clean`, everything seems to be properly generated. There is a small difference in one of the mocks, but I don't think this has any impact.

Now, obviously `GO111MODULE=off` is the huge improvement, but parallel execution brings the times down with another ~50%. We need to decide how to go on in this PR:

* Keep `make generate` sequential as before and introduce a new `generate-parallel` target.
* Use the parallel `make generate` (and optionally introduce a `generate-sequential` target).

The only advantage of sequential generation is that it doesn't burn your CPU as much while it's generating, so you could do other things more comfortably during this time. But with times in the range of 2 mins (faster on newer laptops), this hardly seem to matter, I would just wait rather than trying to do something else while my CPU is burning.

Also, we need to consider the impact on CI. I think there the principle is "the faster the better", but I could be missing something.

/cc @rfranzke @timebertt @vpnachev What do you think?

rfranzke:
I think offering parallel execution is still good, and in the CI system we should definitely use it. Locally, it can stay an individual decision whether to generate sequentially or in parallel.

stoyanr:
@rfranzke Do you think I should introduce a `generate-sequential` make target then for developers to use it if they want? Or perhaps do something more sophisticated, like an env variable controlling the parallelism level? I will personally never use this, so it's hard for me to envision what people might prefer.

timebertt:
Amazing! With the newest changes, `make generate` is under 2m on my machine: 🚀 
```
$ time make generate
...
make generate  680.01s user 537.14s system 1140% cpu 1:46.76 total
```

rfranzke:
> @rfranzke Do you think I should introduce a `generate-sequential` make target then for developers to use it if they want? Or perhaps do something more sophisticated, like an env variable controlling the parallelism level? I will personally never use this, so it's hard for me to envision what people might prefer.

Hm, not sure either, let's start simple, so `generate-sequential` and `generate` (parallel execution) sounds good to me. I'm also fine if you have `generate` (sequential) and `generate-parallel`, doesn't matter.

stoyanr:
OK, did all the requested changes and added a `generate-sequential` target. Measurement on my laptop are similar to:

```
make generate  522.58s user 384.88s system 610% cpu 2:28.72 total
...
make generate-sequential  327.99s user 319.66s system 233% cpu 4:37.95 total
```

stoyanr:
Also time measurement for `make check-generate` that is used in every CI run (this uses parallel execution since it calls `make generate`):
```
make check-generate  524.23s user 390.81s system 588% cpu 2:35.42 total
```

stoyanr:
In https://concourse.ci.gardener.cloud/builds/436582, you can see `make generate` on this PR takes around 2 min:
```
11:43:09 >> make generate
11:45:19 >> make revendor
```

For comparison, on other PRs it takes more than 12, e.g. https://concourse.ci.gardener.cloud/builds/425938:
```
16:22:06 >> make generate
16:34:24 >> make revendor
```

Makefile
stoyanr: Done

timebertt (dismissed):
/lgtm

vpnachev (dismissed):
With the latest changes I am down to 1m40s from the 8m30s previously.

/lgtm

rfranzke (dismissed):
/lgtm

timuthy (dismissed):
It also works perfectly for me. Thanks for the improvement.

hack/update-codegen.sh
```diff
@@ -22,187 +22,259 @@ rm -f ${GOPATH}/bin/*-gen
...
+    --input-dirs=k8s.io/apimachinery/pkg/util/intstr \
+    --report-filename=${PROJECT_ROOT}/pkg/openapi/api_violations.report \
+    --output-package=github.com/gardener/gardener/pkg/openapi \
+    -h "${PROJECT_ROOT}/hack/LICENSE_BOILERPLATE.txt"
+}
+export -f openapi_definitions
+
+if [[ $# -gt 0 && "$1" == "--parallel" ]]; then
+  shift 1
+  parallel ::: \
```
ialidzhikov:
```suggestion
  parallel --will-cite ::: \
```

hack/generate-parallel.sh
```diff
@@ -0,0 +1,37 @@
...
+ROOTS=${ROOTS:-$(git grep --files-with-matches -e '//go:generate' "$@" | \
+	xargs -n 1 dirname | \
+	sed 's,^,github.com/gardener/gardener/,;' | \
+	sort | uniq
+)}
+popd > /dev/null
+
+read -ra PACKAGES <<< $(echo ${ROOTS})
+
+GO111MODULE=off parallel echo Generate {}';' go generate -mod=vendor {} ::: ${PACKAGES[@]}
```
ialidzhikov:
```suggestion
GO111MODULE=off parallel --will-cite echo Generate {}';' go generate -mod=vendor {} ::: ${PACKAGES[@]}
```

Makefile
```diff
@@ -161,6 +161,14 @@ check:
 
 .PHONY: generate
 generate:
+	@GO111MODULE=off hack/update-protobuf.sh
+	@GO111MODULE=off hack/update-codegen.sh --parallel
+	@hack/generate-parallel.sh cmd extensions pkg plugin test
+
+.PHONY: generate-sequential
```
ialidzhikov, line 170: To prevent someone to remove the `GO111MODULE=off` part someday when we forget about this detail.

```suggestion
# We need to explicitly pass GO111MODULE=off to k8s.io/code-generator as it is significantly slower
# when it is executed without GO111MODULE=off - ref https://github.com/kubernetes/code-generator/issues/100.
.PHONY: generate-sequential
```

Makefile
```diff
@@ -161,6 +161,14 @@ check:
 
 .PHONY: generate
```
ialidzhikov, line 164: To prevent someone to remove the `GO111MODULE=off` part someday when we forget about this detail. 

```suggestion
# We need to explicitly pass GO111MODULE=off to k8s.io/code-generator as it is significantly slower
# when it is executed without GO111MODULE=off - ref https://github.com/kubernetes/code-generator/issues/100.
.PHONY: generate
```

hack/generate.sh
```diff
@@ -18,4 +18,4 @@ set -e
 
 echo "> Generate"
 
-GO111MODULE=on go generate -mod=vendor $@
+GO111MODULE=off go generate -mod=vendor $@
```
ialidzhikov, line 23: Hmm, I guess you idea with the explicit `GO111MODULE=off` is for the usage of this script in the extensions repositories? If yes, does it make sense again to add more details about why it is explicitly set to `GO111MODULE=off`?

```suggestion
# We need to explicitly pass GO111MODULE=off to k8s.io/code-generator as it is significantly slower
# when it is executed without GO111MODULE=off - ref https://github.com/kubernetes/code-generator/issues/100.
GO111MODULE=off go generate -mod=vendor $@
```

ialidzhikov (commented):
Minor nits. Thank you for looking into this issue.

hack/update-codegen.sh
stoyanr: Done

hack/generate-parallel.sh
stoyanr: Done

Makefile
stoyanr: As the `generate` and `generate-sequential` targets are adjacent in the Makefile, only left the comment on the `generate` target, I don't think we need 2 identical comments 5 lines across each other.

Makefile
stoyanr: Done

hack/generate.sh
stoyanr: Done, and added the same comment to `generate-parallel.sh` as well. Now I think every usage of `GO111MODULE=off` is commented.

stoyanr:
@ialidzhikov Addressed your feedback.

Thanks to everyone for your feedback and help with this issue!

timebertt (approved):

ialidzhikov (approved):
/lgtm

PR #3494, opened 2021-02-03
Streamline extension components

rfranzke:

/area dev-productivity
/kind cleanup
/priority normal

This PR streamlines the extension components under `./pkg/operation/botanist/extensions` in the following ways:
* Interfaces which were not yet part of the individual extension packages (but rather part of `./pkg/operation/shoot/types.go`) were moved to the individual extension packages
* Interfaces were renamed to `Interface`
* Mock packages earlier maintained in `./pkg/mock/gardener` are now moved into `mock` packages in the individual extension package directories

**Release note**:

```other operator
NONE
```


timuthy (commented):
I was thinking back and forth when reviewing this PR because technically it looks good but then OTHO the interfaces are now exactly used in an unrecommended way, like @BeckerMax once mentioned [ref](https://github.com/golang/go/wiki/CodeReviewComments#interfaces). Nevertheless, I don't see any clear disadvantage following your approach because the component related logic is Gardener internal, i.e. we don't expose a service here. But maybe there are other opinions from @gardener/gardener-maintainers.

danielfoehrKn:
I do not have a strong opinion on go interface best practices. 
However, I think it decreases code cluttering, and as @timuthy said, it's only used Gardener internally anyways.

kon-angelo:
I generally agree that methods should return the struct instead of the interface. You can use somethings like:
```go
var _ myStruct = myInterface
```
to supplement the `New` methods and get some nice compile time errors. 

The "client-side" interface definition is more debatable. If the implementor is used in many places in the code for a "pure" approach you would need define an interface every time on client side (and also redefine mocks for that for each place used). That takes considerably more engineering discipline to keep clean. In practice, there is usually a single IF implementation (+mocks) used in a single package and that can work with the "client-side" approach.

timuthy:
Any further opinions, maybe @BeckerMax I see that you assigned yourself?

rfranzke:
Do you see any drawbacks or problems with the current approach or is it just about pleasing the conventions? I'm not against them, but if there is no clear benefit I would rather prefer to proceed with the PR instead of investing more time now in another refactoring, especially as the proposed approach is used in several other places already, both in g/g and k/k. I'm not aware of any problematic experiences with it in the past years.

Actually, we were partly following those conventions and had defined the interfaces in the `shoot` package, but I didn't like it too much to smear the interface somewhere else (at some very unexpected place) while the implementation is in yet another place. I think following the convention makes sense when we can expect more than implementation of the interface (e.g., for admission plugins or authorizer plugins, etc.). This isn't true for our components. Actually, we only have the interface for mock generation.

So, please, let's either discuss concrete issues or go ahead and work on something with more impact.

timuthy:
Thanks @rfranzke. As I said, no blocker from my point of view, as we don't benefit from the convention here.

BeckerMax:
> Interfaces which were not yet part of the individual extension packages (but rather part of ./pkg/operation/shoot/types.go) were moved to the individual extension packages

+1 looks better this way

> Interfaces were renamed to Interface

Out of curiosity what was your motivation for this change? Was it readability? Did you see this somewhere else?

_Regarding the interface discussion:_
I will thrive for implementing the interfaces on the Client Side in Go and I think this has many advantages (not only esoterical ones). Also generally reducing mocks and testing with the real [implementation](https://github.com/golang/go/wiki/CodeReviewComments#interfaces) sounds like a good idea (no mock ceremony in the tests). But I also have no answer how to efficiently write tests for our code without having the Iface and the mocks on the Producer side. Out of practicality our current approach works and we do not face any concrete drawbacks.

/lgtm

timuthy (approved):

rfranzke:
> Out of curiosity what was your motivation for this change? Was it readability? Did you see this somewhere else?

Hm, no strong reasoning behind it. I think we saw this in some other packages and external libraries. When a package was called `foo` then often the Interface was called `foo.Interface` instead of `foo.Foo`. It's not important, I just wanted to "harmonize" it because in some packages we were using `foo.Interface` already while we were using `foo.Foo`.

PR #3425, opened 2021-01-18
Wait until the finalizer of the credentials secret is removed

ialidzhikov:
/kind bug

Ref https://github.com/gardener/gardener-extension-provider-aws/pull/238#issuecomment-761986865

**Release note**:

```bugfix operator
The generic Worker actuator does now wait until the machine-controller-manager finalizer is removed from the credentials secret that is referenced from the machine classes.
```

PR #3497, opened 2021-02-04
Wait for the out-of-tree mcm provider finalizer as well

ialidzhikov:
/kind bug

In https://github.com/gardener/gardener/pull/3425 we introduced a check that wait until the mcm finalizer is released from the cloudprovider Secret.
It seems that the finalizers used by in-tree MCM and out-of-tree MCM providers are different and we have to respect both of them.

**Release note**:

```bugfix operator
An issue causing the generic Worker actuator to not wait until the finalizer of the out-of-tree machine controller provider is removed from the credentials secret is now fixed.
```


prashanth26 (approved):
Thanks for this fix. 

/lgtm

rfranzke (approved):
/lgtm

PR #3506, opened 2021-02-05
Use `patch` to update Seed.Status

BeckerMax:
  /area quality
/kind enhancement
/priority normal

We were seeing log output when updating seed conditions like 
```
Operation cannot be fulfilled on seeds.<..> : the object has been modified; please apply your changes to the latest version and try again
```
which sometimes delayed the update or even needed a gardenlet restart to recover in some cases.

Before this change the seed status was updated using optimistic locking which could result in ongoing retries if different clients were updating the Seed object.

As an optimization we are now switching to a `patch` call which will send only the diff to the apiserver. We ensure that other conditions are not changed by:
a. only sending the actual diff (for example only the specific condition type that changed)
b. Requesting the strategy `StrategicMergePatchType` from the server.

The downside of this change is that if two clients try to update the same Condition type (or any other field in the status) it could happen due to no guaranteed ordering that the outdated update wins (as it arrives at server at a later point in time). We currently do not see this as an issue as one condition type is only updated by one controller and not multiple ones in parallel.

**Release note**:
```other operator
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/seed/seed_control.go
+++ b/pkg/gardenlet/controller/seed/seed_control.go
@@ -382,14 +382,6 @@ func (c *defaultControl) updateSeedStatus(
 ) error {
 	seedCopy := seed.DeepCopy()
 
-	// remove "available condition"
-	for i, c := range seed.Status.Conditions {
-		if c.Type == "Available" {
-			seed.Status.Conditions = append(seed.Status.Conditions[:i], seed.Status.Conditions[i+1:]...)
-			break
-		}
-	}
-
 	seed.Status.Conditions = gardencorev1beta1helper.MergeConditions(seed.Status.Conditions, updateConditions...)
 	seed.Status.ObservedGeneration = seed.Generation
 	seed.Status.Gardener = c.identity
--- a/pkg/gardenlet/controller/seed/seed_control.go
+++ b/pkg/gardenlet/controller/seed/seed_control.go
@@ -24,7 +24,6 @@ import (
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
-	gardencore "github.com/gardener/gardener/pkg/client/core/clientset/versioned"
 	gardencoreinformers "github.com/gardener/gardener/pkg/client/core/informers/externalversions"
 	gardencorelisters "github.com/gardener/gardener/pkg/client/core/listers/core/v1beta1"
 	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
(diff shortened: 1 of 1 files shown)


pkg/gardenlet/controller/seed/seed_control.go
```diff
@@ -374,32 +374,38 @@ func (c *defaultControl) ReconcileSeed(obj *gardencorev1beta1.Seed, key string)
...
-			seed.Status.Gardener = c.identity
-			seed.Status.KubernetesVersion = &k8sVersion
-			seed.Status.Capacity = capacity
-			seed.Status.Allocatable = allocatable
-			return seed, nil
-		},
-	); err != nil {
+	// remove "available condition"
+	for i, c := range seed.Status.Conditions {
+		if c.Type == "Available" {
```
timuthy: Do we still need that @rfranzke? Or does this even work with the `merge` strategy @BeckerMax?

timuthy (changes requested):
Looks good to me besides one concern.

pkg/gardenlet/controller/seed/seed_control.go
BeckerMax: TLDR: The logic still works even with this merge strategy

I just tested it:
For testing I added a custom Type named `mytype`  to the conditions.  Then I changed the logic in line 387 to remove `mytype`.
The patchBytes then included the following
```
time="2021-02-05T20:59:35+01:00" level=info msg="{\"status\":{\"conditions\":[{\"$patch\":\"delete\",\"type\":\"mytype\"}]}}"
```
And the condition with type `mytype` indeed got removed from the conditions. So even with types.StrategicMergePatchType the explicit delete in the patch is respected.

mvladev:
In general you should not do that - if there is a conflict - something in the resource was changed when you reach the reconcilation point - the spec, status or other field. In most cases it's a safe to overwrite it, but in some cases the spec was changed and it could lead into situations where you are creating a status which is calculated wrong.

This a big problem in the Gardener today. The golden rule is - you start to reconcile and if you update the status or anything in the resource you are processing you should stop the current reconcilation and start over.

E.g. in your reconcilation you want to add a finalizer if it's not present, do some changes and update the status. What you need to do is check if the finalizer needs update, add it, send the update request and exit. The resource is requeued, then you don't have to add finalizer and only update the status. The resource is re-queued again and this time you do nothing. Fast reconcilation is the best in such cases.

BeckerMax:
@mvladev So basically you are saying that the computed status update could be wrong if the controller operates on an outdated resourceVersion and therefore the update call should fail? And the root for our problems here is that the reconciliation cycle takes too long for Seeds?

Would you propose then to instead work on speeding up the reconciliation time?

rfranzke:
Generally, I agree to @mvladev, but this is not always applicable to Gardener (at least not with the current state of the code/implementation). Some of Gardener's controllers have long reconciliation loops, and this is something that will only improve over time.
For example, as part of #2754 we are making the seed reconciliation more lightweight (by using `ManagedResource`s) and we can, once all components are refactored, simply wait for them to be ready (similar to what we do in the `ControllerInstallation` controller (returning an error and leveraging the exponential backoff)).
However, this is something that will not change overnight, so let's please continue with the PR as planned for now to get rid of the issues that we occasionally see today.

timuthy (dismissed):
/lgtm

timebertt (dismissed):
/lgtm

BeckerMax:
As discussed I deleted the removal of the `Available` condition type as it was a temporary migration logic no longer needed.

timebertt (approved):
/lgtm

timuthy (approved):

Issue #3495, opened 2021-02-04
Hanging ControllerInstallations block Seed deletion


timebertt:
**How to categorize this issue?**

/area usability
/kind bug
/priority normal

**What happened**:

When deleting a Seed, we currently see some problems with the cleanup of controller installations:

- the federated seed controller stops to properly report the `ControllerInstallationRequired` condition ([this part](https://github.com/gardener/gardener/blob/03794c4fab012f16625540cd2989ab8f328be93d/pkg/gardenlet/controller/federatedseed/extensions/controllerinstallation_control.go#L196)) when the gardenlet is restarted and the seed is already marked for deletion, because of this early exit: https://github.com/gardener/gardener/blob/03794c4fab012f16625540cd2989ab8f328be93d/pkg/gardenlet/controller/federatedseed/seed_control.go#L179-L182
Therefore, the gardener controller manager won't remove the `ControllerInstallation` in case the last reported condition status is `Required` (see [here](https://github.com/gardener/gardener/blob/9e188335bf4c1a3cc29f75f7a5ecadc24b0fb250/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go#L363-L365)), and the seed deletion will be block forever.
- When the Seed uses the gardener managed dns for the seed ingress feature (i.e. `seed.spec.dns.provider` is set), a ControllerInstallation for types `DNSProvider/<seed-dns-provider-type>` will hang forever, as it is always computed as a wanted ControllerInstallation and there is no logic for the seed deletion case: https://github.com/gardener/gardener/blob/9e188335bf4c1a3cc29f75f7a5ecadc24b0fb250/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go#L289-L291

**What you expected to happen**:

The Seed deletion should succeed and should not require any manual cleanup.

**How to reproduce it (as minimally and precisely as possible)**:

For the first bug: delete a Seed and start the gardenlet *afterwards*. See, that the reported `ControllerInstallationRequired` will not be updated anymore.

For the second bug: delete a Seed, that uses the managed ingress dns feature. See, that the `ControllerInstallation` for the dns provider will not be cleaned up.

In both cases the Seed will be stuck in deletion forever.

**Environment**:

- Gardener version: `v1.16.0`

timebertt:
I assume, these are the issues, that you, @stoyanr and @invidian, also encountered. Can you confirm or are there additional bugs?

/assign @timebertt 
for the first bug
and 
[...]

PR #3511, opened 2021-02-08
Clean-up ingress DNS Resources on Seed deletion

BeckerMax:
  /area control-plane
  /kind bug
/priority normal

Without this change Seeds which have turned on the managed Ingress controller feature will not properly be cleaned up on deletion.
With this change all created DNS ingress resources will be deleted on Seed deletion. This includes:
* DNSProvider
* ProviderSecret
* DNSEntry
* DNS ControllerInstallation

The Controllerinstallation might still not be cleaned up until https://github.com/gardener/gardener/issues/3495 is resolved.

**Release note**:

```bugfix operator
Fixes a bug on seed deletion which caused dns related resources not being cleaned-up if `seed.spec.ingress.kind` was set.
```

### DIFF
--- a/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
+++ b/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
@@ -347,27 +347,43 @@ func computeWantedControllerRegistrationNames(
 		}
 	}
 
-	for _, requiredExtension := range wantedKindTypeCombinations.UnsortedList() {
-		names, ok := kindTypeToControllerRegistrationNames[requiredExtension]
-		if !ok {
-			return nil, fmt.Errorf("need to install an extension controller for %q but no appropriate ControllerRegistration found", requiredExtension)
-		}
-
-		wantedControllerRegistrationNames.Insert(names...)
+	installedAndRequired := installedAndRequiredRegistrationNames(controllerInstallationList, seedObjectMeta.Name)
+	if seedObjectMeta.DeletionTimestamp == nil {
+		wantedControllerRegistrationNames.Insert(installedAndRequired.List()...)
 	}
 
+	for _, wantedExtension := range wantedKindTypeCombinations.UnsortedList() {
+		names, ok := kindTypeToControllerRegistrationNames[wantedExtension]
+		if !ok {
+			return nil, fmt.Errorf("need to install an extension controller for %q but no appropriate ControllerRegistration found", wantedExtension)
+		}
+		if seedObjectMeta.DeletionTimestamp != nil {
+			for _, name := range names {
+				if installedAndRequired.Has(name) {
+					wantedControllerRegistrationNames.Insert(name)
+				}
+			}
+		} else {
+			wantedControllerRegistrationNames.Insert(names...)
+		}
+	}
+
+	// filter controller registrations with non-matching seed selector
+	return controllerRegistrationNamesWithMatchingSeedLabelSelector(wantedControllerRegistrationNames.UnsortedList(), controllerRegistrations, seedObjectMeta.Labels)
+}
+
+func installedAndRequiredRegistrationNames(controllerInstallationList *gardencorev1beta1.ControllerInstallationList, seedName string) sets.String {
+	requiredControllerRegistrationNames := sets.NewString()
 	for _, controllerInstallation := range controllerInstallationList.Items {
-		if controllerInstallation.Spec.SeedRef.Name != seedObjectMeta.Name {
+		if controllerInstallation.Spec.SeedRef.Name != seedName {
 			continue
 		}
 		if !gardencorev1beta1helper.IsControllerInstallationRequired(controllerInstallation) {
 			continue
 		}
-		wantedControllerRegistrationNames.Insert(controllerInstallation.Spec.RegistrationRef.Name)
+		requiredControllerRegistrationNames.Insert(controllerInstallation.Spec.RegistrationRef.Name)
 	}
-
-	// filter controller registrations with non-matching seed selector
-	return controllerRegistrationNamesWithMatchingSeedLabelSelector(wantedControllerRegistrationNames.UnsortedList(), controllerRegistrations, seedObjectMeta.Labels)
+	return requiredControllerRegistrationNames
 }
 
 // computeRegistrationNameToInstallationNameMap computes a map that maps the name of a ControllerRegistration to the name of an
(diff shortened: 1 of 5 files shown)


pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -283,9 +283,16 @@ func computeKindTypesForShoots(
 // Seed configuration
 func computeKindTypesForSeed(
 	seed *gardencorev1beta1.Seed,
+	numberOfShoots int,
+	numberOfBuckets int,
 ) sets.String {
 	var wantedKindTypeCombinations = sets.NewString()
 
+	seedInDeletionAndEmpty := seed.ObjectMeta.DeletionTimestamp != nil && numberOfBuckets == 0 && numberOfShoots == 0
```
rfranzke: Below logic only evaluates the `.spec.dns` field today, so why is the `numberOfBuckets` playing a role here?

pkg/operation/seed/seed.go
```diff
@@ -1107,13 +1102,50 @@ func handleDNSProvider(ctx context.Context, gardenClient, seedClient client.Clie
...
+func emptyDNSProvider() *dnsv1alpha1.DNSProvider {
+	return &dnsv1alpha1.DNSProvider{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: v1beta1constants.GardenNamespace,
+			Name:      "seed",
+		},
+	}
+}
+
+func emptyProviderSecret() *corev1.Secret {
```
```suggestion
func emptyDNSProviderSecret() *corev1.Secret {
```

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
BeckerMax: We want to keep access to logging and monitoring until everything is cleaned up. So if deletion of Shoots or Backupbuckets fail you can still debug it.

rfranzke: Hm, not sure how this answer is related, but isn't it conceptually wrong to check for buckets/shoots here? I mean, this is maybe a "good?" approximation, but not the "correct" solution, right? The `DNSProvider`/`DNSEntry` resources in the `Seed` could still be present even if there are no `Shoot`s/`BackupBucket`s anymore (maybe the DNS secret is wrong and they cannot be deleted - then an operator needs to change it first).

The `Seed` deletion is a bit special, I agree. I would propose to rather adapt https://github.com/gardener/gardener/blob/0a5df55ecf3a3824662450a7c8d6498d55605215/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go#L319-L371. I think we can skip computing the names of the required `ControllerRegistration`s if the `Seed` has a `deletionTimestamp` and the `Required` condition of the backing `ControllerInstallation` is `False`.

As concrete implementation proposal:

1. Move https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go#L359-L367 to https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go#L349
2. Before https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go#L366, store the name of the `ControllerRegistration` in a new set `requiredControllerRegistrationNames`. At this point we know that there is at least one `ControllerInstallation` with `Required=True` referencing the `ControllerRegistration`.
3. In https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go#L355, if the `Seed`'s `deletionTimestamp` is set then iterate over `names` and only insert the name to `wantedControllerRegistrationNames` in case it exists in `requiredControllerRegistrationNames`.

rfranzke:
> (maybe the DNS secret is wrong and they cannot be deleted - then an operator needs to change it first).

Btw, I guess this scenario is not supported by your added logic in the seed deletion flow.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
BeckerMax: Done

BeckerMax:
> maybe the DNS secret is wrong and they cannot be deleted - then an operator needs to change it first

I added logic to copy the dns secret again upon seed deletion.

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
BeckerMax: As I addressed you remarks please take another look @rfranzke

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -347,27 +347,43 @@ func computeWantedControllerRegistrationNames(
...
+			return nil, fmt.Errorf("need to install an extension controller for %q but no appropriate ControllerRegistration found", wantedExtension)
+		}
+		if seedObjectMeta.DeletionTimestamp != nil {
+			for _, name := range names {
+				if installedAndRequired.Has(name) {
+					wantedControllerRegistrationNames.Insert(name)
+				}
+			}
+		} else {
+			wantedControllerRegistrationNames.Insert(names...)
```
rfranzke, line 367: Suggestion to reduce the nesting and improve readability:

```suggestion

		// <- Maybe you can add a comment here and explain what is going on and why we need this special behaviour
		if seedObjectMeta.DeletionTimestamp == nil {
			wantedControllerRegistrationNames.Insert(names...)
			continue
		}
		
		for _, name := range names {
			if installedAndRequired.Has(name) {
				wantedControllerRegistrationNames.Insert(name)
			}
		}
```

pkg/gardenlet/controller/seed/seed_control.go
```diff
@@ -253,6 +253,19 @@ func (c *defaultControl) ReconcileSeed(obj *gardencorev1beta1.Seed, key string)
...
+					return err
+				}
+			}
+			if err = seedpkg.DeleteIngressDNSEntry(ctx, seedClient, seedObj); err != nil {
+				return err
+			}
+			if err = seedpkg.DeleteDNSProvider(ctx, seedClient.Client()); err != nil {
+				return err
+			}
+
```
rfranzke, line 268: Hm, what's this? Shouldn't this be part of the `DebootstrapCluster` function?

rfranzke: Also, doesn't this rather belong to #3575 (it conflicts with it anyways, or?)

BeckerMax: Sure, should I just merge the PRs into one and open it with 2 commits?

rfranzke: Sounds like a good proposal.

timebertt:
@BeckerMax what's the status on this PR?

BeckerMax:
I am blocked this week and last week with other priorities. I am going to come back to it beginning of next week.

BeckerMax:
Closed as now included in: https://github.com/gardener/gardener/pull/3575

PR #3518, opened 2021-02-08
Fix bug when debootstrapping Seed cluster

### LINKED ISSUE #3513: Seed unbootstrap fails with "could not find image gardener-resource-manager"

/area control-plane
/kind bug
/priority critical

**What happened**:
Seed unbootstrap fails with:
```
Failed to unbootstrap Seed Cluster (could not find image "gardener-resource-manager" opts runtime version v1.18.14 target version v1.18.14).
```

Seed unbootstrap succeeds.

Register a seed, allow it to bootstrap successfully, then delete it to trigger the unbootstrap.

Caused by #3445 that was merged recently. The code at https://github.com/gardener/gardener/blob/0557bd7414facc5627b0f97e3175abb72ac3a11b/pkg/operation/seed/seed.go#L861-L864 tries to find the "gardener-resource-manager" image in the image vector. This is ok during bootstrap, but during unbootstrap the passed image vector is `nil` (see https://github.com/gardener/gardener/blob/0557bd7414facc5627b0f97e3175abb72ac3a11b/pkg/operation/seed/seed.go#L835), which causes the above error.

**Environment**:

- Gardener version:
- Others:

BeckerMax:
/kind bug
/priority critical

Without this change debootstraping a Seed cluster will fail. Introduced with https://github.com/gardener/gardener/pull/3445

Fixes https://github.com/gardener/gardener/issues/3513

```other operator
NONE
```

### DIFF
--- a/pkg/operation/seed/seed.go
+++ b/pkg/operation/seed/seed.go
@@ -858,9 +858,13 @@ func bootstrapComponents(c kubernetes.Interface, namespace string, imageVector i
 	}
 
 	// gardener-resource-manager
-	image, err := imageVector.FindImage(common.GardenerResourceManagerImageName, imagevector.RuntimeVersion(c.Version()), imagevector.TargetVersion(c.Version()))
-	if err != nil {
-		return nil, err
+	var grmImage string
+	if imageVector != nil {
+		image, err := imageVector.FindImage(common.GardenerResourceManagerImageName, imagevector.RuntimeVersion(c.Version()), imagevector.TargetVersion(c.Version()))
+		if err != nil {
+			return nil, err
+		}
+		grmImage = image.String()
 	}
 	cfg := resourcemanager.Values{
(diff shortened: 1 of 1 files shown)


BeckerMax:
The initially reported error is gone
/hold until I tested the complete Seed deletion flow.

stoyanr (approved):

BeckerMax:
Closed because we are going to refactor (de-)bootstrapping instead. See:
https://github.com/gardener/gardener/issues/3513#issuecomment-775195115

BeckerMax:
Can be merged as a first mitigation. Resources managed by the grm (etcd-druid, seed-admission controller) need to be cleaned up manually upon Seed deletion until https://github.com/gardener/gardener/issues/3513#issuecomment-775195115 is done

rfranzke (approved):

PR #3520, opened 2021-02-08
Scale down MCM deployment instead of deleting it during restore

krgostev:

/area control-plane
/kind enhancement
/priority normal

After the worker restore operation if the reconcile fails for some reason the restore will start again. The MCM then will exists and all resources will be deleted and recreated. This behaviour is not expected during migration as the machines should be preserved.  
Scaling down the MCM to '0' instead of deleting all resources should be sufficient enough to ensure that the restore is successful.
Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/extensions/pkg/controller/worker/genericactuator/machine_controller_manager.go
+++ b/extensions/pkg/controller/worker/genericactuator/machine_controller_manager.go
@@ -125,7 +125,7 @@ func (a *genericActuator) waitUntilMachineControllerManagerIsDeleted(ctx context
 
 func (a *genericActuator) scaleMachineControllerManager(ctx context.Context, logger logr.Logger, worker *extensionsv1alpha1.Worker, replicas int32) error {
 	logger.Info("Scaling machine-controller-manager", "replicas", replicas)
-	return kubernetes.ScaleDeployment(ctx, a.client, kutil.Key(worker.Namespace, McmDeploymentName), replicas)
+	return client.IgnoreNotFound(kubernetes.ScaleDeployment(ctx, a.client, kutil.Key(worker.Namespace, McmDeploymentName), replicas))
 }
 
 func (a *genericActuator) applyMachineControllerManagerShootChart(ctx context.Context, workerDelegate WorkerDelegate, workerObj *extensionsv1alpha1.Worker, cluster *controller.Cluster) error {
--- a/pkg/client/kubernetes/scaling.go
+++ b/pkg/client/kubernetes/scaling.go
@@ -17,8 +17,10 @@ package kubernetes
 import (
 	"context"
 	"fmt"
+	"time"
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+	"github.com/gardener/gardener/pkg/utils/retry"
 
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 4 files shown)


plkokanov:
Thanks for the PR.

We still should wait for the MCM to be scaled down before proceeding with the addition of the machine objects.

Just a note from my side: this is necessary as we've ran into problems when the worker managedresource gets deleted and then recreated as this leads to deletions/recreations of RBAC resources in the shoot and it could take the MCM quite a long time to have the correct access for all necessary resources in the shoots.

extensions/pkg/controller/worker/genericactuator/machine_controller_manager.go
```diff
@@ -124,8 +125,58 @@ func (a *genericActuator) waitUntilMachineControllerManagerIsDeleted(ctx context
...
+			return nil
+		}
+		return err
+	}
 	logger.Info("Scaling machine-controller-manager", "replicas", replicas)
-	return kubernetes.ScaleDeployment(ctx, a.client, kutil.Key(worker.Namespace, McmDeploymentName), replicas)
+	if err := kubernetes.ScaleDeployment(ctx, a.client, kutil.Key(worker.Namespace, McmDeploymentName), replicas); err != nil {
+		return err
+	}
+	return a.waitUntilMachineControllerManagerIsScaled(ctx, worker)
```
rfranzke: It would be more clear to move this out of the `scaleMachineControllerManager` function and rather call it in `Restore`.

extensions/pkg/controller/worker/genericactuator/machine_controller_manager.go
```diff
@@ -124,8 +125,58 @@ func (a *genericActuator) waitUntilMachineControllerManagerIsDeleted(ctx context
...
+			logger.Info("Machine-controller-manager is not deployed yet")
+			return nil
+		}
+		return err
+	}
 	logger.Info("Scaling machine-controller-manager", "replicas", replicas)
-	return kubernetes.ScaleDeployment(ctx, a.client, kutil.Key(worker.Namespace, McmDeploymentName), replicas)
+	if err := kubernetes.ScaleDeployment(ctx, a.client, kutil.Key(worker.Namespace, McmDeploymentName), replicas); err != nil {
+		return err
+	}
```
rfranzke: Could you use https://github.com/gardener/gardener/blob/03794c4fab012f16625540cd2989ab8f328be93d/pkg/client/kubernetes/scaling.go#L54-L63?

extensions/pkg/controller/worker/genericactuator/machine_controller_manager.go
```diff
@@ -124,8 +125,58 @@ func (a *genericActuator) waitUntilMachineControllerManagerIsDeleted(ctx context
...
+
+		return retry.Ok()
+	}); err != nil {
+		var retryError *retry.Error
+		if !errors.As(err, &retryError) {
+			return err
+		}
+	}
+
+	return nil
```
rfranzke: Can you move https://github.com/gardener/gardener/blob/03794c4fab012f16625540cd2989ab8f328be93d/pkg/operation/botanist/waiter.go#L253-L275 to `pkg/utils/kubernetes` and then call it here?

extensions/pkg/controller/worker/genericactuator/machine_controller_manager.go
krgostev: I think this is the same method that it is used

extensions/pkg/controller/worker/genericactuator/machine_controller_manager.go
krgostev: Thanks for pointing this out, it looks much better

extensions/pkg/controller/worker/genericactuator/machine_controller_manager.go
```diff
@@ -125,7 +125,12 @@ func (a *genericActuator) waitUntilMachineControllerManagerIsDeleted(ctx context
 
 func (a *genericActuator) scaleMachineControllerManager(ctx context.Context, logger logr.Logger, worker *extensionsv1alpha1.Worker, replicas int32) error {
 	logger.Info("Scaling machine-controller-manager", "replicas", replicas)
-	return kubernetes.ScaleDeployment(ctx, a.client, kutil.Key(worker.Namespace, McmDeploymentName), replicas)
+	err := kubernetes.ScaleDeployment(ctx, a.client, kutil.Key(worker.Namespace, McmDeploymentName), replicas)
+	if apierrors.IsNotFound(err) {
+		logger.Info("Machine-controller-manager is not deployed yet")
+		return nil
+	}
+	return err
```
rfranzke: Hm, do we actually care about this? I'm asking because you handle this "not found" case explicitly and add a log message, but is it really relevant/does it help? Or can we simply go with

```suggestion
	return client.IgnoreNotFound(kubernetes.ScaleDeployment(ctx, a.client, kutil.Key(worker.Namespace, McmDeploymentName), replicas))
```

rfranzke (commented):
Looks good, thanks for adapting. One last minor question.

extensions/pkg/controller/worker/genericactuator/machine_controller_manager.go
plkokanov: I also do not think this log is necessary. Every time restore is called for the first (right after migration) time the MCM will not be available.

pkg/client/kubernetes/scaling.go
```diff
@@ -70,3 +72,28 @@ func scaleResource(ctx context.Context, c client.Client, obj client.Object, repl
 	// see: https://github.com/kubernetes-sigs/controller-runtime/issues/172
 	return c.Patch(ctx, obj, client.RawPatch(types.MergePatchType, patch))
 }
+
+// WaitUntilDeploymentScaledToDesiredReplicas waits for the number of available replicas to be equal to the deployment's desired replicas count.
+func WaitUntilDeploymentScaledToDesiredReplicas(ctx context.Context, client client.Client, key types.NamespacedName, desiredReplicas int32) error {
+	return retry.UntilTimeout(ctx, 5*time.Second, 300*time.Second, func(ctx context.Context) (done bool, err error) {
+		deployment := &appsv1.Deployment{}
+		if err := client.Get(ctx, key, deployment); err != nil {
```
plkokanov, line 80: If the deployment is not found we will get an error, but I think we should just return retry.Ok() in that case, similar to client.IgnoreNotFound() in scaleMachineControllerManager.

plkokanov (changes requested):

pkg/client/kubernetes/scaling.go
rfranzke: Hm, but why in this function? Isn't it odd that we return "OK" if we want to scale a deployment to 5 replicas and then it suddenly disappears?
I think this behaviour is highly specific to the restoration flow, so let's move this graceful "not found" handling out from here to the `Restore` function.
Sorry @kris94 for the back-and-forth :(

plkokanov: Yes, moving it out makes sense.

rfranzke (approved):
/lgtm

PR #3524, opened 2021-02-09
Remove the Affinity from the Loki StatefulSet only for int. tests

vlvasilev:

/area logging
/kind test
/priority normal

By moving the garden/loki-0 pod into the etcd worker pool the  [Loki StatefulSet](https://github.com/gardener/gardener/blob/master/charts/seed-bootstrap/charts/loki/templates/statefulset.yaml) received a affinity configuration by the [kupid](https://github.com/gardener/kupid). When logging integration test is run a seed is simulate on a shoot cluster, but the shoots don't have etcd worker pool and Loki can't be scale up. With this PR the affinity section is remove only for the integration tests.
 
Fixes #

**Release note**:

``` bugfix operator
The affinity section is removed from the Loki StatefulSet for the integration tests
```

### DIFF
--- a/test/integration/shoots/logging/seed_logging_stack.go
+++ b/test/integration/shoots/logging/seed_logging_stack.go
@@ -139,6 +139,7 @@ var _ = ginkgo.Describe("Seed logging testing", func() {
 				lokiSts.Spec.Template.Spec.Containers[index].Resources = r
 			}
 		}
+		lokiSts.Spec.Template.Spec.Affinity = nil
 		framework.ExpectNoError(create(ctx, f.ShootClient.Client(), lokiSts))
 
 		ginkgo.By("Wait until Loki StatefulSet is ready")


test/integration/shoots/logging/seed_logging_stack.go
```diff
@@ -139,6 +139,8 @@ var _ = ginkgo.Describe("Seed logging testing", func() {
 				lokiSts.Spec.Template.Spec.Containers[index].Resources = r
 			}
 		}
+		// Remove the affinity because shoots don't have etcd worker pool
```
vpnachev: Seeds also do not have etcd worker pool by default. This is a Gardener landscape installation detail.
```suggestion
```

vpnachev (changes requested):

vpnachev (approved):
/lgtm

PR #3525, opened 2021-02-09
Fix conflicts when bootstrapping new seeds

timebertt:

/kind bug
/priority normal

Even after #3506, I still see some conflicts, when boostrapping new seeds.
This PR fixes two occurrences of those conflicts by replacing an `Update` call with `Patch`.

```text
time="2021-02-09T10:46:43+01:00" level=error msg="Seed bootstrapping failed: Operation cannot be fulfilled on seeds.core.gardener.cloud \"seed-gcp\": the object has been modified; please apply your changes to the latest version and try again" seed=seed-gcp
```

Should probably be cherry-picked to `v1.15` and `v1.16`.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/operation/seed/seed.go
+++ b/pkg/operation/seed/seed.go
@@ -17,7 +17,6 @@ package seed
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"path/filepath"
 	"strings"
@@ -693,8 +692,10 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
 			return err
 		}
 
+		seedCopy := seed.Info.DeepCopy()
 		seed.Info.Status.ClusterIdentity = &seedClusterIdentity
-		if err := k8sGardenClient.Client().Status().Update(ctx, seed.Info); err != nil {
+
+		if err := k8sGardenClient.Client().Status().Patch(ctx, seed.Info, client.MergeFrom(seedCopy)); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/operation/seed/seed.go
```diff
@@ -693,8 +692,10 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
 			return err
 		}
 
+		seedCopy := seed.Info.DeepCopy()
 		seed.Info.Status.ClusterIdentity = &seedClusterIdentity
-		if err := k8sGardenClient.Client().Status().Update(ctx, seed.Info); err != nil {
+
+		if err := k8sGardenClient.Client().Status().Patch(ctx, seed.Info, client.MergeFrom(seedCopy)); err != nil {
```
timuthy, line 698: Is it really safe to use a json merge patch here? Imagine there is another controller which updates the status in the meantime, like this controller which maintains the backup bucket condition in the future (#3487). I think this intermediate update will get lost, right?

timebertt: No, it shouldn't get lost, as the patch that the seed controller is sending will only contain the diff (i.e. `.status.clusterIdentity`) and will not overwrite other fields, if they have been updated concurrently.
Please correct me, if I'm wrong.

timuthy: I see, it won't get lost because the conditions aren't modified.

timuthy (approved):
/lgtm

ialidzhikov (approved):
/lgtm

ialidzhikov:
> Should probably be cherry-picked to `v1.15` and `v1.16`.

@timebertt , can you file the cherry-pick PRs?

timebertt:
> > Should probably be cherry-picked to `v1.15` and `v1.16`.
> 
> @timebertt , can you file the cherry-pick PRs?

Done.

PR #3531, opened 2021-02-09
Add BackupBucketsReady seed condition

### LINKED ISSUE #3428: Propagate BackupBucket creation errors to the Seed resource status

timuthy:

/area ops-productivity
/kind enhancement
/priority normal

This PR adds a new condition type `BackupBucketsReady` to `seed.status.conditions` in case the seed specifies a backup configuration. Please refer to the linked issue for more details.

Fixes #3428

/cc @amshuman-kr 

**Release note**:

```feature operator
`Seed` resources now have a new condition type `BackupBucketsReady` that is added when the corresponding seed has a backup configuration or related `BackupBuckets`. `Seeds` whose `BackupBucketsReady` condition is `status: "False"` are considered `NotReady` and thus are excluded from scheduling during that time.
```


timuthy:
PR remains in draft mode because some unit tests for the new controller are missing. Nevertheless, you can already have a look @amshuman-kr @rfranzke @vpnachev.

pkg/controllermanager/controller/seed/seed_backup_bucket_control.go
```diff
@@ -0,0 +1,194 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func (c *Controller) backupBucketEnqueue(obj client.Object) {
+	key := client.ObjectKeyFromObject(obj)
+	c.seedBackupBucketQueue.Add(key.String())
```
rfranzke: Why are you enqueueing the `BackupBucket` instead of the referenced `Seed` here? I guess this would save some code here.

rfranzke (changes requested):
Overall, looks quite good.

1. What about the migration case, i.e., when one seed is migrated to another? I guess your new controller does not handle it yet properly (or?), i.e., the `BackupBucketReady` condition for the old seed might stay invalid (maybe the old seed does not even have associated buckets anymore now, so the condition must be removed)?
1. Should we use the opportunity to enhance  https://github.com/gardener/gardener/blob/master/docs/concepts/controller-manager.md a little bit?

pkg/controllermanager/controller/seed/seed_backup_bucket_control.go
timuthy: I also thought about that but was then just following this example https://github.com/gardener/gardener/blob/277374cfffbd034648d228b1344f22782904cf06/pkg/gardenlet/controller/seed/seed_extension_control.go#L68. Is there an essential difference I don't see?

rfranzke: You are not really following this example. This "seed extension" check is exactly implementing what I'm suggesting: It watches `ControllerInstallation`s and enqueues the referenced `Seed`: https://github.com/gardener/gardener/blob/277374cfffbd034648d228b1344f22782904cf06/pkg/gardenlet/controller/seed/seed_extension_control.go#L36-L42
The reconciler then reconciles the `Seed` and does not know about the `ControllerInstallation` anymore.

You are doing it differently: You are watching `BackupBucket`s and then enqueue them here. Your reconciles then "reconciles" the `BackupBucket`, but the only thing it does it to call another reconciler for reconciling the referenced `Seed`.

I am suggesting to remove this intermediate step because it seems to be not necessary. You can directly enqueue the referenced `Seed` and have only one reconciler for the `Seed` directly, like in the example you sent.

timuthy: Okay sure, I'll change it. I still don't understand the difference to gardener/pkg/gardenlet/controller/seed/seed_extension_control.go. For me it's a mapping in both cases, from Resource A -> B. What the "2nd" controller (reconciling B) knows or doesn't know, shouldn't be relevant.

rfranzke: Yes, it's a mapping in both cases, but you currently get an event with the object, then enqueue the object's name, then you read the object by name again, and then you call your actual business logic with some information out of the object.

The "seed extension" check gets an event with the object and immediately extracts the relevant information so that the business logic can be called right away.

Obviously, the result/behaviour of the controller is the same. The proposed approach is just simpler and results in less code.

timuthy:
> What about the migration case, i.e., when one seed is migrated to another? I guess your new controller does not handle it yet properly (or?), i.e., the BackupBucketReady condition for the old seed might stay invalid (maybe the old seed does not even have associated buckets anymore now, so the condition must be removed)?

This case it not completely clarified, IIUC. Why should a seed not have any buckets at all if it specifies a backup configuration in `seed.spec.backup`? This can/should actually only happen if the seed is getting deleted. I decided to maintain the condition as `UNKNOWN` since this is some kind of dangling state. Nevertheless, I revised the update handler slightly to also consider the old seed object (0ade4e608baf8d30b9ca876bd39a3d98b25e5735). Am I missing something?

timuthy:
Any feedback @rfranzke ?

docs/concepts/controller-manager.md
```diff
@@ -119,3 +119,23 @@ The Shoot Reference Controller can inspect the following references:
...
+
+#### "Backup Bucket" Reconciler
+Every time a `BackupBucket` object is created or updated, the referenced `Seed` object is enqueued for reconciliation.
+It's the reconciler's task to check the `status` subresource of all existing `BackupBuckets` that belong to this seed.
+If at least one `BackupBucket` has `.status.lastError`, the seed condition `BackupBucketsReady` will turn `false` and
+consequently the seed is considered as `NotReady`. Once the `BackupBucket` is healthy again, the seed will be re-queued
+and the condition will turn `true`.
+
+#### "Lifecycle" Reconciler
+The "Lifecycle" reconciler processes seed objects which are enqueued every 10 seconds in order to check if the responsible
```
```suggestion
The "Lifecycle" reconciler processes `Seed` objects which are enqueued every 10 seconds in order to check if the responsible 
```

docs/concepts/controller-manager.md
```diff
@@ -119,3 +119,23 @@ The Shoot Reference Controller can inspect the following references:
...
+It's the reconciler's task to check the `status` subresource of all existing `BackupBuckets` that belong to this seed.
+If at least one `BackupBucket` has `.status.lastError`, the seed condition `BackupBucketsReady` will turn `false` and
+consequently the seed is considered as `NotReady`. Once the `BackupBucket` is healthy again, the seed will be re-queued
+and the condition will turn `true`.
+
+#### "Lifecycle" Reconciler
+The "Lifecycle" reconciler processes seed objects which are enqueued every 10 seconds in order to check if the responsible 
+Gardenlet is still responding and operable. Therefore, it checks renewals via `Lease` objects of the seed in the garden cluster
+which are renewed regularly by the Gardenlet.
+In case a `Lease` is not renewed for the configured amount in `config.Controllers.Seed.MonitorPeriod.Duration`, the reconciler
```
```suggestion
In case a `Lease` is not renewed for the configured amount in `config.controllers.seed.monitorPeriod.duration`, the reconciler
```

docs/concepts/controller-manager.md
```diff
@@ -119,3 +119,23 @@ The Shoot Reference Controller can inspect the following references:
...
+assumes that the Gardenlet stopped operating and updates the `GardenletReady` condition to `Unknown`.
+Additionally, conditions and constraints of all shoot clusters running on the affected seed are set to `Unknown` as well
```
```suggestion
Additionally, conditions and constraints of all `Shoot` resources scheduled on the affected seed are set to `Unknown` as well
```

pkg/controllermanager/controller/seed/seed_backup_bucket_reconcile.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (c *Controller) backupBucketUpdate(oldObj, newObj interface{}) {
+	oldBackupBucket, ok1 := oldObj.(*gardencorev1beta1.BackupBucket)
+	newBackupBucket, ok2 := newObj.(*gardencorev1beta1.BackupBucket)
+
+	if !ok1 || !ok2 {
+		return
+	}
+
+	if apiequality.Semantic.DeepEqual(oldBackupBucket.Spec, newBackupBucket.Spec) || apiequality.Semantic.DeepEqual(oldBackupBucket.Status, newBackupBucket.Status) {
```
rfranzke: Hm, should this be 
```suggestion
	if !apiequality.Semantic.DeepEqual(oldBackupBucket.Spec, newBackupBucket.Spec) || !apiequality.Semantic.DeepEqual(oldBackupBucket.Status, newBackupBucket.Status) {
```

? Maybe I'm missing what this is supposed to do 👀

pkg/controllermanager/controller/seed/seed_backup_bucket_reconcile.go
```diff
@@ -0,0 +1,186 @@
...
+		c.backupBucketEnqueue(newBackupBucket)
+	}
+
+	var (
+		oldSeedName = oldBackupBucket.Spec.SeedName
+		newSeedName = newBackupBucket.Spec.SeedName
+	)
+
+	if oldSeedName != nil && (newSeedName == nil || *oldSeedName != *newSeedName) {
```
rfranzke:
```suggestion
	if !apiequality.Semantic.DeepEqual(oldBackupBucket.Spec.SeedName, newBackupBucket.Spec.SeedName) {
```

should have the same effect, or?

pkg/controllermanager/controller/seed/seed_backup_bucket_reconcile.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Logger.Debugf("[BACKUPBUCKET SEED RECONCILE] %s - skipping because Seed has been deleted", req.NamespacedName)
+			return reconcile.Result{}, nil
+		}
+		logger.Logger.Infof("[BACKUPBUCKET SEED RECONCILE] %s - unable to retrieve seed object from store: %v", req.NamespacedName, err)
+		return reconcile.Result{}, err
+	}
+
+	bbs, err := b.backupBucketLister.List(labels.Everything())
```
rfranzke, line 96: I'm looking forward to the day when we have switched the controllers to controller-runtime so that we can use field selectors here :)

rfranzke (commented):
Sorry for the delay @timuthy.
Mainly nits to be clarified, overall I'm very happy with the PR now, thanks for applying the suggestions 🤝

timuthy:
Thanks for your input and nice findings @rfranzke. PTAL.

rfranzke (approved):
/lgtm

PR #3538, opened 2021-02-10
Fix kubeconfig rotation for konnectivity-servers

mvladev:

/area control-plane
/kind bug
/priority normal

During kubeconfig rotation, the static tokens for konnectivity-server are rotated agents cannot authenticate.

Fixes #

**Release note**:

```bugfix operator
Fixed a bug causing `KonnectivityTunnel` to break when kubeconfig rotation occurs.
```


vpnachev (commented):
Can you add it to the list only if the feature gate is enabled?
This way we will save one `delete` request that either will fail.

rfranzke (dismissed):
/lgtm

mvladev:
@vpnachev have another look.

vpnachev (approved):
/lgtm

PR #3540, opened 2021-02-11
Fix name of grm vpa after refactoring

BeckerMax:
/area control-plane
/kind bug
/priority normal

After https://github.com/gardener/gardener/pull/3445 the grm vpa was wrongly named. This PR fixes that.

**Release note**:

```other operator
NONE
```


danielfoehrKn (approved):

timebertt (commented):
Am I assuming correctly, that after the refactoring before this PR, the different naming caused two VPAs to be deployed?
That's the problem you want to fix with this PR, right?
If so, lgtm

rfranzke (approved):

rfranzke:
Yes @timebertt, I don't know why, but it seems we suffix the names of VPA objects with "-vpa" for most of our resources. @BeckerMax discovered that he missed this during the refactoring, so this PR now corrects it.

PR #3541, opened 2021-02-11
Fix apiVersion error when fetching events

timuthy:

/area ops-productivity
/kind bug
/priority normal

This PR fixes a bug introduced with https://github.com/gardener/gardener/pull/3466. We cannot assume that `TypeMeta` is set after retrieving an object from the API server, even though the object has a `TypeMeta` in the storage. [WithoutVersionDecoder](https://github.com/kubernetes/kubernetes/blob/63f89493ec3c03c3cc9681e86d13da31cb25b2dd/staging/src/k8s.io/apimachinery/pkg/runtime/helper.go#L255) removes the GVK information after decoding.

/cc @rfranzke @mvladev

**Release note**:

```other operator
NONE
```


timebertt (dismissed):
/lgtm

rfranzke (dismissed):

timebertt (approved):
/lgtm

PR #3543, opened 2021-02-11
Fix gardener-resource-manager deletion order

### LINKED ISSUE #3513: Seed unbootstrap fails with "could not find image gardener-resource-manager"

BeckerMax:

/area control-plane
/kind bug
/priority normal

Currently upon Seed deletion the gardener-resource-manager could
be cleaned up before all managed resources are deleted, leaving them
hanging.

With this change the grm is only deleted after all managed resources
are cleaned up.

In the long run we want to refactor the seed creation/deletion flow using a flow graph. This is an intermediate fix until the refactoring is merged.

Fixes #3513 

**Release note**:
```other operator
NONE
```

### DIFF
--- a/pkg/operation/seed/seed.go
+++ b/pkg/operation/seed/seed.go
@@ -17,11 +17,13 @@ package seed
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"strings"
 	"time"
 
+	"github.com/gardener/gardener-resource-manager/pkg/apis/resources/v1alpha1"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorelisters "github.com/gardener/gardener/pkg/client/core/listers/core/v1beta1"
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
Thanks!
/lgtm for now, especially as we plan to refactor this with the `flow` pkg soon

timuthy (approved):

PR #3550, opened 2021-02-12
Use patch to update the extension data in the ShootState

plkokanov:

/area control-plane-migration
/kind enhancement
/priority normal

This PR changes the way the extension states and their related resources (if any) are added to the ShootState. `Patch` is used to persist the changes and it is only called if there are differences between the extensions' states in the `Shoot`'s control plane and the corresponding states in the `ShootState`

Part of #1631 #2689 

**Release note**:

```other operator
Use PATCH to update the extensions' state and relevant resource data to the ShootState.
```

### DIFF
--- a/pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
+++ b/pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
@@ -21,11 +21,10 @@ import (
 	"github.com/sirupsen/logrus"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	apiextensions "github.com/gardener/gardener/pkg/api/extensions"
@@ -39,9 +38,11 @@ import (
 	"github.com/gardener/gardener/pkg/operation/common"
 	"github.com/gardener/gardener/pkg/utils"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 )
 
-type shootStateControl struct {
+// ShootStateControl is used to update data about extensions and any resources required by them in the ShootState.
+type ShootStateControl struct {
 	k8sGardenClient kubernetes.Interface
 	seedClient      kubernetes.Interface
 	log             *logrus.Entry
@@ -49,10 +50,21 @@ type shootStateControl struct {
 	shootRetriever  *ShootRetriever
 }
 
-func (s *shootStateControl) createShootStateSyncReconcileFunc(kind string, objectCreator func() client.Object) reconcile.Func {
+// NewShootStateControl creates a new instance of ShootStateControl
+func NewShootStateControl(k8sGardenClient, seedClient kubernetes.Interface, log *logrus.Entry, recorder record.EventRecorder, shootRetriever *ShootRetriever) *ShootStateControl {
+	return &ShootStateControl{
+		k8sGardenClient: k8sGardenClient,
+		seedClient:      seedClient,
+		log:             log,
+		recorder:        recorder,
+		shootRetriever:  shootRetriever,
+	}
+}
+
+// CreateShootStateSyncReconcileFunc creates a function which can be used by the reconciliation loop to sync the extension state and its resources to the ShootState
+func (s *ShootStateControl) CreateShootStateSyncReconcileFunc(kind string, objectCreator func() client.Object) reconcile.Func {
 	return func(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
-		obj := objectCreator()
-		err := s.seedClient.Client().Get(ctx, req.NamespacedName, obj)
+		extensionObj, err := s.getExtensionObject(ctx, req.NamespacedName, objectCreator)
 		if apierrors.IsNotFound(err) {
 			s.log.Debugf("Skipping ShootState sync because resource with kind %s is missing in namespace %s", kind, req.NamespacedName)
 			return reconcile.Result{}, nil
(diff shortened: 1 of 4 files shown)


pkg/apis/core/v1alpha1/types_shootstate.go
```diff
@@ -56,11 +56,22 @@ type ShootStateSpec struct {
...
 	// Extensions holds the state of custom resources reconciled by extension controllers in the seed
+	// +patchMergeKey=kind
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=kind
+	// +listMapKey=name
+	// +listMapKey=purpose
 	// +optional
-	Extensions []ExtensionResourceState `json:"extensions,omitempty" protobuf:"bytes,2,rep,name=extensions"`
+	Extensions []ExtensionResourceState `json:"extensions,omitempty" patchStrategy:"merge" patchMergeKey:"kind" protobuf:"bytes,2,rep,name=extensions"`
```
rfranzke: Hm, there can be multiple entries with the same kind. Is this problematic?

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
```diff
@@ -61,75 +73,87 @@ func (s *shootStateControl) createShootStateSyncReconcileFunc(kind string, objec
...
-				}
-				if obj == nil {
-					return fmt.Errorf("object not found %v", resource.ResourceRef)
-				}
-				raw := &runtime.RawExtension{}
-				if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj, raw); err != nil {
-					return err
-				}
-				updateShootStateResourceData(shootState, &resource.ResourceRef, raw)
+		if shouldUpdateExtensionData || len(resourcesToUpdate) > 0 {
```
timebertt:
```suggestion
		if apiequality.Semantic.DeepEqual(shootState, shootStateCopy) {
```
WDYT?

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
```diff
@@ -61,75 +73,87 @@ func (s *shootStateControl) createShootStateSyncReconcileFunc(kind string, objec
...
+			if err := s.k8sGardenClient.Client().Patch(ctx, shootState, client.MergeFrom(shootStateCopy)); err != nil {
```
timebertt: This patch call uses a `json-merge` patch. So the `mergeKeys` you added to the API, won't be leveraged.
Instead, the entire list of resources will be replaced.
We either need a `strategic-merge` patch or a `json-merge` patch with optimistic locking here.

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
```diff
@@ -190,12 +214,40 @@ func purposeToString(purpose *string) string {
...
+		return nil, fmt.Errorf("could not get cluster with name %s : %v", clusterName, err)
+	}
+
+	shoot, err := s.shootRetriever.FromCluster(cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	shootState := &gardencorev1alpha1.ShootState{}
+	if err := s.k8sGardenClient.DirectClient().Get(ctx, kutil.Key(shoot.Namespace, shoot.Name), shootState); err != nil {
```
timebertt: This will retrieve the entire `ShootState` every time an extension object is changed (which can happen quite frequently).
```suggestion
	if err := s.k8sGardenClient.Client().Get(ctx, kutil.Key(shoot.Namespace, shoot.Name), shootState); err != nil {
```

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control_test.go
```diff
@@ -0,0 +1,299 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var _ = Describe("ShootState Control", func() {
```
timebertt, line 45: cc @timuthy

timebertt (changes requested):
Thanks for adding the test!

pkg/apis/core/v1alpha1/types_shootstate.go
plkokanov: the following entries above the field should take care of that:
```
// +listType=map
// +listMapKey=kind
// +listMapKey=name
// +listMapKey=purpose
```

I tested it by editing the `status.state` of the controlplane and controlplane exposure extension resources and got both of the entries in the shootstate.

Probably makes sense to add it in the unit tests although I use the fake client there.

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
plkokanov: Was contemplating a lot about that. Generally didn't like comparing the entire shootState object.

timebertt: Ok, fine for me then

pkg/apis/core/v1alpha1/types_shootstate.go
rfranzke: Hm, can you explain more about these comments/settings? I haven't seen them yet, I think 👀 . Where do their effects manifest?

plkokanov: I read a bit more and i've missed a pretty important part about the documentation on those comments/settings. They are only used with "server side apply". So unless I use that in the Patch call there is no reason for them to be in the shootstate types def.

Also the reason I got both entries was that i wasn't using the correct patch strategy as Tim pointed in his [comment](https://github.com/gardener/gardener/pull/3550#discussion_r577363249). 

Anyway after playing some more here's what I found out that there is a problem if I want to generate the patch data for a strategic merge patch with kutil.CreateTwoWayMergePatch(...)  and the extension's state already exists in the `ShootState` and only a part of the new state is changed. As an example if we have the following in the `ShootState (this only happens if the RawExtension holds an array):
```
- kind: worker
  state:
    machineDeployments:
    - entry1
    - entry2
```
And we try to create patch data with from a `ShootState` with modified extension state looking like this:
```
- kind: worker
  state:
    machineDeployments:
    - entry1
    - entry3
```
Generating the patch data will fail with: 
```
Unable to find api field in struct RawExtension for the json field \"machineDeployments\" 
```
There are other subtle errors - if in the example above the new state contains a new struct entry that should replace `machineDeployments` it will instead be added to the state.

The internals of CreateTwoWayMergePatch() is quite complex though and I'm not sure why that's happening yet (would need more time to understand it)

A json-merge patch with optimistic locking seems to do the job, and we won't have to worry about extensions with the same kind and name. @rfranzke @timebertt should we just go for that? Today I'll try out server side apply as well.

timebertt: > A json-merge patch with optimistic locking seems to do the job, and we won't have to worry about extensions with the same kind and name.

Yeah, probably that's the easiest way to go for now and should already be sufficient, I think.
If it still makes some troubles, we can think about `strategic-merge-patch` or `server-side-apply` again.

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
timebertt: Ping, I think this one is not resolved.

timebertt (commented):
lgtm apart from one open suggestion

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
plkokanov: Sorry, must've reverted it when I removed the merge annotations. Modified it to use `Client()` now.

timebertt (approved):
Thanks!
/lgtm

PR #3556, opened 2021-02-12
Removing the finalizers from the Terraform config resources during cleanup

krgostev:

/area control-plane-migration
/area control-plane
/kind bug
/priority normal

This PR fixes an issue that can be observed during CP migration. When `func CleanupConfiguration(...)` the finalizers of the ConfigMaps and Secrets are not removed and they can't be deleted.
Fixes #

**Release note**:

```noteworthy dependency
The `Terraformer` interface has now a new function `RemoveTerraformerFinalizerFromConfig` which will remove the "terraformer" finalizer from the `Secret`/`ConfigMap` resources.
```

### DIFF
--- a/extensions/pkg/terraformer/config.go
+++ b/extensions/pkg/terraformer/config.go
@@ -24,6 +24,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	"github.com/gardener/gardener/pkg/controllerutils"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 )
 
@@ -245,20 +246,42 @@ func (t *terraformer) CleanupConfiguration(ctx context.Context) error {
 	t.logger.Info("Cleaning up all terraformer configuration")
 
 	t.logger.V(1).Info("Deleting Terraform variables Secret", "name", t.variablesName)
-	if err := t.client.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}); err != nil && !apierrors.IsNotFound(err) {
+	if err := t.client.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}); client.IgnoreNotFound(err) != nil {
 		return err
 	}
 
 	t.logger.V(1).Info("Deleting Terraform configuration ConfigMap", "name", t.configName)
-	if err := t.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.configName}}); err != nil && !apierrors.IsNotFound(err) {
+	if err := t.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.configName}}); client.IgnoreNotFound(err) != nil {
 		return err
 	}
 
 	t.logger.V(1).Info("Deleting Terraform state ConfigMap", "name", t.stateName)
-	if err := t.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.stateName}}); err != nil && !apierrors.IsNotFound(err) {
+	if err := t.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.stateName}}); client.IgnoreNotFound(err) != nil {
 		return err
 	}
 
+	return nil
+}
+
+// RemoveTerraformerFinalizerFromConfig deletes the terraformer finalizer from the two ConfigMaps and the Secret which store the Terraform configuration and state.
+func (t *terraformer) RemoveTerraformerFinalizerFromConfig(ctx context.Context) error {
+	t.logger.Info("Cleaning up all terraformer configuration finalizers")
+
+	for _, obj := range []client.Object{
+		&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}},
+		&corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.stateName}},
+		&corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.configName}},
+	} {
+		if err := t.client.Get(ctx, kutil.Key(t.namespace, obj.GetName()), obj); client.IgnoreNotFound(err) != nil {
+			return err
+		}
+
+		if controllerutil.ContainsFinalizer(obj, TerraformerFinalizer) {
+			if err := controllerutils.PatchRemoveFinalizers(ctx, t.client, obj, TerraformerFinalizer); client.IgnoreNotFound(err) != nil {
+				return err
+			}
+		}
+	}
 	return nil
 }
 
--- a/extensions/pkg/terraformer/terraform_test.go
+++ b/extensions/pkg/terraformer/terraform_test.go
@@ -49,6 +49,8 @@ const (
 	variablesName     = "variables"
 	stateName         = "state"
 	infraUID          = "2a540a5c-1b8c-11e8-b291-0a580a2c025a"
+	purpose           = "purpose"
+	image             = "image"
 )
 
 var _ = Describe("terraformer", func() {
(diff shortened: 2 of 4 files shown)


extensions/pkg/terraformer/config.go
```diff
@@ -245,17 +246,29 @@ func (t *terraformer) CleanupConfiguration(ctx context.Context) error {
 	t.logger.Info("Cleaning up all terraformer configuration")
 
 	t.logger.V(1).Info("Deleting Terraform variables Secret", "name", t.variablesName)
-	if err := t.client.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}); err != nil && !apierrors.IsNotFound(err) {
+	secret := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}
+	if err := t.client.Delete(ctx, secret); err != nil && !apierrors.IsNotFound(err) {
```
rfranzke: FYI: You can use a helper method for this:

```suggestion
	if err := t.client.Delete(ctx, secret); client.IgnoreNotFound(err) {
```

rfranzke (commented):
Looks good, thanks. Does it make sense to use the opportunity to provide some unit tests for this function?

/invite @timebertt @vpnachev

extensions/pkg/terraformer/config.go
```diff
@@ -245,17 +246,29 @@ func (t *terraformer) CleanupConfiguration(ctx context.Context) error {
...
+		return err
+	}
+	if err := extensionscontroller.DeleteAllFinalizers(ctx, t.client, secret); err != nil {
```
plkokanov: Just saw this, sorry for not mentioning sooner... If the secret is not found and you call `DeleteAllFinalizers` the inner call to tryUpdate will fail as the secret does not exist. Same for the configMap and stateConfigMap. You can reuse client.IgnoreNotFound() here as well I think.

plkokanov (changes requested):

extensions/pkg/terraformer/config.go
krgostev: Yes, I should have seen that

extensions/pkg/terraformer/config.go
```diff
@@ -245,17 +246,29 @@ func (t *terraformer) CleanupConfiguration(ctx context.Context) error {
 	t.logger.Info("Cleaning up all terraformer configuration")
 
 	t.logger.V(1).Info("Deleting Terraform variables Secret", "name", t.variablesName)
-	if err := t.client.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}); err != nil && !apierrors.IsNotFound(err) {
+	secret := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}
+	if err := t.client.Delete(ctx, secret); client.IgnoreNotFound(err) != nil {
+		return err
+	}
+	if err := extensionscontroller.DeleteAllFinalizers(ctx, t.client, secret); client.IgnoreNotFound(err) != nil {
```
vpnachev: Why all finalizers?
You can remove just `TerraformerFinalizer` that we are aware is set. Removing any other finalizer can brake other controller that gardener is not aware of.

plkokanov: Yep, should be similar to what we're doing for the worker controller and the mcm finalizers there.

vpnachev:
Hm, this is currently affecting the general `tf.Destroy` also and I have proposed the same thing some time ago in https://github.com/gardener/gardener/pull/3423  with https://github.com/gardener/gardener/commit/6deb49f629590eb0cd4f93658d8426e1500795aa, but we decided to not do it. 

If we do it now, then I can close https://github.com/gardener/gardener/pull/3561 as it will be obsolete.

plkokanov:
If we leave the removal here do we risk deleting resources before the terraform pod has finished destroying the infrastructure and handling the finalizers itself? I'm guessing that's why you initially added the retry logic in #3561
If that's the case, then I guess we should move the cleanup outside of this function and only have it in the Migrate function of the infra controller.

vpnachev:
Actually, we have observed successful infrastructure destruction without the finalizers to be removed from the configmaps/secret which led to seed namespace stuck in deletion. 

> If we leave the removal here do we risk deleting resources before the terraform pod has finished destroying the infrastructure

No, it waits the terraform pod to finish the execution of the `destroy` command and just after that invoke the `CleanupConfiguration`. https://github.com/gardener/gardener/blob/06d4e71f7e8e3391beb41e46db15cf416bb3e90f/extensions/pkg/terraformer/terraformer.go#L139-L145

vpnachev (dismissed):
/lgtm
/needs second-opinion

extensions/pkg/terraformer/config.go
```diff
@@ -245,17 +247,29 @@ func (t *terraformer) CleanupConfiguration(ctx context.Context) error {
 	t.logger.Info("Cleaning up all terraformer configuration")
 
 	t.logger.V(1).Info("Deleting Terraform variables Secret", "name", t.variablesName)
-	if err := t.client.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}); err != nil && !apierrors.IsNotFound(err) {
+	secret := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}
+	if err := t.client.Delete(ctx, secret); client.IgnoreNotFound(err) != nil {
+		return err
+	}
+	if err := extensionscontroller.DeleteFinalizer(ctx, t.client, secret, TerraformerFinalizer); client.IgnoreNotFound(err) != nil {
```
plkokanov:
```suggestion
	if err := extensionscontroller.PatchRemoveFinalizers(ctx, t.client, secret, TerraformerFinalizer); client.IgnoreNotFound(err) != nil {
```

plkokanov (changes requested):

plkokanov (dismissed):
lgtm

vpnachev (dismissed):
/lgtm
/needs second-opinion

krgostev:
/hold

After some more discussions we decided to separate the cleanup of the finalizers in another function and call it in the migration flow

extensions/pkg/terraformer/config.go
```diff
@@ -245,23 +246,45 @@ func (t *terraformer) CleanupConfiguration(ctx context.Context) error {
...
-	if err := t.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.stateName}}); err != nil && !apierrors.IsNotFound(err) {
+	if err := t.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.stateName}}); client.IgnoreNotFound(err) != nil {
 		return err
 	}
 
 	return nil
 }
 
+// RemoveTerraformerFinalizerFromConfig deletes the terraformer finalizer from the two ConfigMaps which store the Terraform configuration and state.
+func (t *terraformer) RemoveTerraformerFinalizerFromConfig(ctx context.Context) error {
```
rfranzke, line 267: Can you run `make generate` and add a unit test for this function?

vpnachev:
> After some more discussions we decided to separate the cleanup of the finalizers in another function and call it in the migration flow

It looks it is still not doing this :sunglasses:

plkokanov:
> > After some more discussions we decided to separate the cleanup of the finalizers in another function and call it in the migration flow
> 
> It looks it is still not doing this 😎

Yeah, the actual calls to the new function will have to be done in each cloud provider extension since that's where the Migrate implementations are.

extensions/pkg/terraformer/config.go
```diff
@@ -245,23 +246,45 @@ func (t *terraformer) CleanupConfiguration(ctx context.Context) error {
...
 	t.logger.V(1).Info("Deleting Terraform state ConfigMap", "name", t.stateName)
-	if err := t.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.stateName}}); err != nil && !apierrors.IsNotFound(err) {
+	if err := t.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.stateName}}); client.IgnoreNotFound(err) != nil {
 		return err
 	}
 
 	return nil
 }
 
+// RemoveTerraformerFinalizerFromConfig deletes the terraformer finalizer from the two ConfigMaps which store the Terraform configuration and state.
```
vpnachev: Mention also the secret.

vpnachev (commented):
Just a docu improvement from me apart from what Rafael already requests.

rfranzke (approved):
/lgtm
/unhold

vpnachev (approved):
/lgtm

PR #3560, opened 2021-02-12
Remove finalizers from the worker secret in the migration flow

krgostev:

/area control-plane-migration
/kind bug
/priority normal

In the migration flow the MCM finalizer of the secret that is referenced by the worker is not removed. This PR adds functionality that checks that secret and removes only the MCM finalizers if necessary. 
Fixes #

**Release note**:

```noteworthy dependency
In the generic worker actuator's migration flow, the MCM finalizer of the secret that is referenced by the `Worker` is not removed. We have now added functionality that checks that secret and removes only the MCM finalizers if necessary. 
```

### DIFF
--- a/extensions/pkg/controller/worker/genericactuator/actuator.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator.go
@@ -24,6 +24,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
@@ -41,8 +42,10 @@ import (
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	gardenerkubernetes "github.com/gardener/gardener/pkg/client/kubernetes"
+	"github.com/gardener/gardener/pkg/controllerutils"
 	"github.com/gardener/gardener/pkg/utils/chart"
 	"github.com/gardener/gardener/pkg/utils/imagevector"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 )
 
 // GardenPurposeMachineClass is a constant for the 'machineclass' value in a label.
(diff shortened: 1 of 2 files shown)


extensions/pkg/controller/worker/genericactuator/actuator.go
```diff
@@ -252,6 +254,26 @@ func (a *genericActuator) shallowDeleteMachineClassSecrets(ctx context.Context,
...
+		return err
+	}
+	if controllerutil.ContainsFinalizer(secret, mcmFinalizer) {
+		controllerutil.RemoveFinalizer(secret, mcmFinalizer)
+	}
+	if controllerutil.ContainsFinalizer(secret, mcmProviderFinalizer) {
+		controllerutil.RemoveFinalizer(secret, mcmProviderFinalizer)
+	}
+
+	return nil
```
vpnachev: You need to update/patch the secret back to the API.

vpnachev (changes requested):
Just out of curiosity, is this reproducible after https://github.com/gardener/gardener/pull/3497?

krgostev:
It is still reproducible, because in the [migration flow](https://github.com/gardener/gardener/blob/1b9a0e1379dff432e65368c5237c837826d9b7ac/extensions/pkg/controller/worker/genericactuator/actuator_migrate.go#L47) we have to delete the MCM first that is responsible for removing the finalizer.

extensions/pkg/controller/utils.go
```diff
@@ -119,6 +119,9 @@ var EnsureFinalizer = controllerutils.EnsureFinalizer
 // If it is set, it removes it and issues an update.
 var DeleteFinalizer = controllerutils.RemoveFinalizer
 
+// PatchDeleteFinalizer ensures that the given finalizer is not present anymore in the given object via a patch request.
+var PatchDeleteFinalizer = controllerutils.PatchRemoveFinalizers
```
vpnachev: Can you avoid this alias?
Last time when I was looking how to remove a finalizer, I found 2 different implementations in g/g and one alias which which is misleading to me.

extensions/pkg/controller/worker/genericactuator/actuator.go
rfranzke: This is not yet resolved, right?

extensions/pkg/controller/worker/genericactuator/actuator.go
```diff
@@ -252,6 +255,30 @@ func (a *genericActuator) shallowDeleteMachineClassSecrets(ctx context.Context,
...
+			return err
+		}
+	}
+	if controllerutil.ContainsFinalizer(secret, mcmProviderFinalizer) {
+		if err := controllerutils.PatchRemoveFinalizers(ctx, a.client, secret, mcmProviderFinalizer); err != nil {
+			return err
+		}
+	}
+
+	return nil
```
vpnachev: Some improvement suggestions
```suggestion
	finalizersToRemove := []string{}
	if controllerutil.ContainsFinalizer(secret, mcmFinalizer) {
		finalizersToRemove = append(finalizersToRemove, mcmFinalizer)
	}
	if controllerutil.ContainsFinalizer(secret, mcmProviderFinalizer) {
		finalizersToRemove = append(finalizersToRemove, mcmProviderFinalizer)
	}
	if len(finalizersToRemove) == 0 {
		return nil
	}

	return controllerutils.PatchRemoveFinalizers(ctx, a.client, secret, finalizersToRemove...)
```

Some improvement suggestions.

extensions/pkg/controller/worker/genericactuator/actuator.go
vpnachev: It is, he has switched from `controllerutil` to `controllerutilS`.

rfranzke: 😕 👀  Is this better? Now we are making two separate calls instead of collecting all changes and doing doing only one call. Maybe this situation cannot occur in reality, but still, seems a little odd to me.

vpnachev: My concern was that there was no `patch` request previously with https://github.com/gardener/gardener/commit/51b71c68f19a38f2c96f4caaed33c7450eea95fe.

rfranzke (approved):
/lgtm
/remove status/author-action
/second-opinion

vpnachev (approved):
/lgtm

plkokanov (approved):
/lgtm

PR #3561, opened 2021-02-13
[terraformer] Ensure terraformer finalizer is removed on Destroy.

vpnachev:
If the terraformer finalizer is not removed from the resource, the
Destroy will fail.

/area ops-productivity
/kind enhancement
/priority normal

Introduces additional check that ensures the configmap and secrets used by terraformer are properly finalized on destroy.

Fixes #

**Release note**:

```other dependency
The `terraformer.Destroy` now ensures that the finalizers from the configmaps and secrets are properly removed. 
```

### DIFF
--- a/extensions/pkg/terraformer/terraformer.go
+++ b/extensions/pkg/terraformer/terraformer.go
@@ -141,7 +141,16 @@ func (t *terraformer) Destroy(ctx context.Context) error {
 	if err := t.execute(ctx, "destroy"); err != nil {
 		return err
 	}
-	return t.CleanupConfiguration(ctx)
+
+	if err := t.CleanupConfiguration(ctx); err != nil {
+		return err
+	}
+
+	if !t.resourcesAreFinalized(ctx) {
+		return fmt.Errorf("terraformer finalizers are still present, needs to retry")
+	}
+
+	return nil
 }
 
 // execute creates a Terraform Pod which runs the provided scriptName (apply or destroy), waits for the Pod to be completed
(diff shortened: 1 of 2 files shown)


rfranzke (approved):
/lgtm

timuthy (commented):
I wondered if you can add unit test(s) for this logic? Apart from that, it looks good to me.

extensions/pkg/terraformer/terraformer.go
```diff
@@ -141,7 +141,16 @@ func (t *terraformer) Destroy(ctx context.Context) error {
...
 		return err
 	}
-	return t.CleanupConfiguration(ctx)
+
+	if err := t.CleanupConfiguration(ctx); err != nil {
+		return err
+	}
+
+	if !t.resourcesAreFinalized(ctx) {
+		return fmt.Errorf("terraformer finalizers are still present, needs to retry")
```
timebertt, line 150: Not sure, whether a retry will always help, as terraformer refuses to execute when one of the resources is missing, e.g. the `tf-config`.
Though, in some cases it might help, so no objection against proceeding like this.

timebertt (approved):
/lgtm
Small side-comment inline.

extensions/pkg/terraformer/terraformer.go
vpnachev: Valid point. It can be in deadlock here. Cannot delete `Infrastructure` because of the still existing configmap/secret and cannot delete the configmaps/secrets because not all of them are present at the same time.

I am starting to think it was a mistake to let the terraformer manage the finalizers on these resources, mainly because it is not an active controller watching the resources, but just a one-time job per Infrastructure reconciliation. 

What do you think about managing the finalizer from within the Infrastructure controller?

vpnachev:
/hold
until we see how https://github.com/gardener/gardener/pull/3556#issuecomment-779179833 will evolve because it can make this PR obsolete.

extensions/pkg/terraformer/terraformer.go
timebertt: Oh yeah, true. Didn't notice that it will even be a deadlock. That's not good.



> I am starting to think it was a mistake to let the terraformer manage the finalizers on these resources, mainly because it is not an active controller watching the resources, but just a one-time job per Infrastructure reconciliation.
> 
> What do you think about managing the finalizer from within the Infrastructure controller?

Hmm, why do you think so? Terraformer is actually creating external infrastructure on behalf of the k8s resources, so it would be the correct component to also protect the resources with finalizers, don't you think?

vpnachev: > Hmm, why do you think so? Terraformer is actually creating external infrastructure on behalf of the k8s resources, so it would be the correct component to also protect the resources with finalizers, don't you think?

I completely agree, and this was the reason why I voted to implement it this way back then.

vpnachev:
Closing in favor of https://github.com/gardener/terraformer/pull/71 which resolves the real issue.

Issue #2227, opened 2020-04-23
Shoot deletion flow fails to delete VolumeSnapshot CRDs


ialidzhikov:
**What happened**:
Currently during Shoot deletion we delete resources such as Services, Deployments, PVCs, etc. The user may also have `VolumeSnapshot`s as we already have multiple provider extensions supporting CSI and `VolumeSnapshot`s.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a Shoot.

2. Create a `VolumeSnapshot`.

3. Delete the Shoot.

4. As the `volumesnapshots.snapshot.storage.k8s.io` CRD is triggered for deletion, the `VolumeSnapshot` CR deletion fails with:
```yaml
  status:
    boundVolumeSnapshotContentName: snapcontent-e12bf34c-a377-44ce-9873-0b0b74670252
    creationTime: "2020-04-23T19:33:59Z"
    error:
      message: Failed to get snapshot class with error volumesnapshotclass.snapshot.storage.k8s.io
        "default-snapshot-class" not found
      time: "2020-04-23T19:52:31Z"
    readyToUse: false
    restoreSize: 6Gi
```

```
$ k get crds
NAME                                             CREATED AT
volumesnapshotcontents.snapshot.storage.k8s.io   2020-04-23T18:38:11Z
volumesnapshots.snapshot.storage.k8s.io          2020-04-23T18:38:11Z
```

As you can see, the `volumesnapshotclass.snapshot.storage.k8s.io` CRD is already deleted.

5. Ensure that the `volumesnapshots.snapshot.storage.k8s.io` CRD will be finalized after 1h and the respective VolumeSnapshot will leak.

**Anything else we need to know?**:

**Environment**:

- Gardener version: master
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

rfranzke:
The underlying problem seems to be that there are CRDs which depend on each other. The `VolumeSnapShot*` CRDs are applied by the extension's `ControlPlane` controller. But this controller does not have a hook to intervene during cluster cleanup at beginning of shoot deletion flow. Generally, there might be also user-defined CRDs having the same problem.

For the deletion we already have the `shoot.gardener.cloud/no-cleanup: "true"` label contract which ensures that Gardener will not delete the resource.
[...]

PR #3562, opened 2021-02-14
Use separate ManagedResource for ControlPlane CRDs that are installed in the Shoot cluster

ialidzhikov:
/area ops-productivity
/kind bug

As described in https://github.com/gardener/gardener/issues/2227#issuecomment-682468386, currently `extensions/pkg/controller/controlplane/genericactuator.Actuator` uses one ManagedResource (named `extension-controlplane-shoot`) for the resources that are installed in the Shoot cluster for the given ControlPlane. With the CSI enablement in the provider extensions this ManagedResource holds CRDs and the RBAC for controller leader election. The deletion of such ManagedResource is likely to fail (hang forever) as it can first delete the RBAC for controller leader election and the corresponding controller (`external-snapshotter`) is no longer able to acquire leadership to proceed with CR (`VolumeSnapshot`) deletions.

Part of https://github.com/gardener/gardener/issues/2227

**Release note**:

```bugfix operator
`extensions/pkg/controller/controlplane/genericactuator.Actuator` can now use a separate ManagedResource for ControlPlane CRDs that are installed in the Shoot cluster to separate the deletion of CRDs from the deletion of the RBAC for controller leader election.
```

### DIFF
--- a/extensions/pkg/controller/controlplane/genericactuator/actuator.go
+++ b/extensions/pkg/controller/controlplane/genericactuator/actuator.go
@@ -57,6 +57,8 @@ type ValuesProvider interface {
 	GetControlPlaneChartValues(ctx context.Context, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster, checksums map[string]string, scaledDown bool) (map[string]interface{}, error)
 	// GetControlPlaneShootChartValues returns the values for the control plane shoot chart applied by this actuator.
 	GetControlPlaneShootChartValues(ctx context.Context, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster, checksums map[string]string) (map[string]interface{}, error)
+	// GetControlPlaneShootCRDsChartValues returns the values for the control plane shoot CRDs chart applied by this actuator.
+	GetControlPlaneShootCRDsChartValues(ctx context.Context, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster) (map[string]interface{}, error)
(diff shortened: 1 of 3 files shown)


ialidzhikov:
~~The corresponding PR on provider extension side - https://github.com/gardener/gardener-extension-provider-aws/pull/285. It should give a better overview how the changes in the extension library from this PR are consumed by the provider extension.~~

ialidzhikov:
I now see yet another flaw in the logic that decides whether it is possible and needed to migrate the Shoot CRDs to separate ManagedResource. I already reworked this part multiple times but it still seems that it is not handling all cases as expected...

ialidzhikov:
So my proposal to go forward with this issue/PR would be:
- Remove the incomplete Shoot CRDs migration logic from this PR. In this way this PR will contain only the generic actuator enhancement that is optional. And in this way providers that still didn't migrated to CSI (like provider-azure) can benefit from this separate ManagedResource enhancement and directly adapt its charts without considering backwards compatibility.
- For the providers that already adopted CSI and require migration of resources from one ManagedResource to another one - I would vote for the `Ignore` mode described in https://github.com/gardener/gardener-resource-manager/issues/82. In this way each provider-extension will be able to mark the CRDs in the old ManagedResource with `Ignore` mode and this will allow graceful migration.

I haven't discussed this with anyone so feel free to comment if you have objections or better ideas. For now I will start with the first point and will clean up the migration logic from this PR and will leave only the genericactuator enhancement for the separate ManagedResource.

rfranzke (approved):
/lgtm

Sorry for the delay.

PR #3563, opened 2021-02-15
Update "Features, Hotfixes, and Releases" doc

rfranzke:

/area documentation dev-productivity open-source
/kind enhancement
/priority normal

This PR updates the https://github.com/gardener/gardener/blob/master/docs/development/process.md document with more information as decided in the last retrospective meeting.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/docs/development/process.md
+++ b/docs/development/process.md
@@ -1,18 +1,38 @@
-# Creating a new Feature
+# Features, Hotfixes, and Releases
 
-If you want to contribute to the Gardener, please do that always on a dedicated branch on your own fork named after the purpose of the code changes, for example `feature/helm-integration`.
-Please do not forget to rebase your branch **regularly**.
+This document describes how to contribute features or hotfixes, and how new Gardener releases are usually scheduled, validated, etc.
 
-If you have finished your work, please create a pull request **based on `master`**. It will be reviewed and merged if no further changes are requested from you.
+## Contributing new Features or Fixes
 
-:warning: Please ensure that your modifications pass the lint checks, formatting checks, static code checks, and unit tests by executing
+Please refer to the [Gardener contributor guide](https://github.com/gardener/documentation/blob/master/CONTRIBUTING.md).
+Besides a lot of a general information, it also provides a checklist for newly created pull requests that may help you to prepare your changes for an efficient review process.
+
+:warning: Please ensure that your modifications pass the verification checks (linting, formatting, static code checks, tests, etc.) by executing
 
 ```bash
 make verify
 ```
 
-Please do not file your pull request unless you receive a successful response from here!
+before filing your pull request.
 
-## Creating a new Release
+The guide applies for both changes to the `master` and to any `release-*` branch.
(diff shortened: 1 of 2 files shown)


timuthy (approved):
/lgtm

docs/development/process.md
```diff
@@ -1,18 +1,38 @@
...
+The guide applies for both changes to the `master` and to any `release-*` branch.
+All changes must be submitted via a pull request and be reviewed and approved by at least one code owner.
 
-Please refer to the [Gardener contributor guide](https://github.com/gardener/documentation/blob/master/CONTRIBUTING.md).
+## Releases
+
+There is no fixed schedule for new releases of the `gardener/gardener` component.
+The [@gardener-maintainers](https://github.com/orgs/gardener/teams/gardener-maintainers) are trying to provide a new release roughly every other week (depending on their capacity and the stability/robustness of the `master` branch).
+
+Hotfixes are usually only maintained for the latest minor release as well as the minor release before that.
```
timebertt, line 26: How do we want to handle older release branches / patch releases?
We've had a few occurrences, where users from the community wanted to upgrade their gardener versions but hit a bug that was only fixed in newer gardener versions and it couldn't be cherry-picked/released to the release branch, they were trying to upgrade to.
Should we keep older release branches and make patch releases for such cases?
Still, we wouldn't be able to properly validate such patch releases...

timebertt (commented):
Thanks for writing this up!
One additional thought/question inline.

docs/development/process.md
vpnachev: k8s has such [case](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-release/cherry-picks.md#cherry-picks-for-unsupported-releases) in their documentation, but we are leaning toward faster release cycle and it will be hard to maintain too old versions. If we do it, it can be interpreted as extending the support time-window for this old release and more cherry-pick request can be opened. 

I believe this is happening rarely, so these users can cherry-pick the fixes and build custom images locally with `make docker-images`  to unblock their landscape upgrades.

vpnachev (approved):
/lgtm

docs/development/process.md
rfranzke: Yes, I agree with @vpnachev. At the moment, we neither have the capacity nor the infrastructure to maintain more than two minor releases (the current one + the one before that). I intentionally formulated it a little "fuzzy" because we _sometimes_ also cherry-pick fixes to older release branches, but it happens very rarely and we don't "officially" support that. I think the sentence is "precise" enough for now while we are still flexible enough due to the usage of _usually_ 😃.

timebertt: > I believe this is happening rarely, so these users can cherry-pick the fixes and build custom images locally with `make docker-images` to unblock their landscape upgrades.

That's true, of course 👍 

> I think the sentence is "precise" enough for now while we are still flexible enough due to the usage of _usually_ 😃.

Ha, ok 😄 
I agree, that its' precise enough for now.

timebertt (approved):
/lgtm

Issue #1723, opened 2019-12-13
SeedAuthorizer restricting gardenlet permission in garden cluster


rfranzke:
With #1601 we have introduced the gardenlet, a component comparable with the kubelet in Kubernetes.

Kubernetes has a dedicated [`NodeAuthorizer` admission plugin](https://github.com/kubernetes/kubernetes/tree/master/plugin/pkg/auth/authorizer/node) which ensures that a kubelet responsible for node `X` may only touch resources (like `Pod`s, `Secret`s, etc.) that are related to node `X`.

In our first introduction of the Gardenlet we haven't implemented such mechanism yet.
Instead, similar to Gardener v0 (where the gardener-controller-manager was bound to the `cluster-admin` role), the gardenlet gets full administrator privileges in the garden cluster.

Let's follow Kubernetes' approach and implement a `SeedAuthorizer` admission plugin which ensures that gardenlets only touch resources related to the `Seed`s they are responsible for.

| | Status |
| --- | --- |
| ✅| completed |
| 🚧| in progress |
| ❔| to be clarified |
| ℹ️| interesting, but not relevant (yet) |
| ❌| incomplete |

Prerequisites:
- ✅ Copy secrets into separate and seed dedicated namespace (see https://github.com/gardener/gardener/issues/1725)
- ✅ Transfer Gardener-Admission-Controller into Controller-Runtime handler (see https://github.com/gardener/gardener/issues/3109)

To-Dos:
- ✅ Skeleton for "SeedAuthorizer"  [authorization webhook](https://kubernetes.io/docs/reference/access-authn-authz/webhook/) ➡️ #3571
- ✅ Skeleton for "SeedRestriction" [admission webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) ➡️ #3571
- ✅ Resource Dependency Graph ➡️ #3631 
- ✅ Functionality for identifying seed identity from a user info ➡️ #3671
- ✅ Adapt gardenlet to properly read global secrets from its seed namespace (see #1725)
- ✅ Implementation for various and involved kinds (`CloudProfiles`, `Shoots`, `Secrets`, etc.).
- ✅ Prevent that a gardenlet can issue a certificate for a seed it is not responsible for

rfranzke:
Due to the fact that the `ControllerRegistration` resource may contain sensitive data in its `.spec.deployment.providerConfig` we want to restrict the gardenlet to only access those that are relevant for its own `Seed`.
[...]

PR #3571, opened 2021-02-15
Boilerplate for `SeedAuthorizer` authorization plugin and `SeedRestriction` admission plugin

rfranzke:

/area security
/kind enhancement
/priority normal

With this PR a new HTTP handler is added to the `gardener-admission-controller` (GAC). The handler is meant to be configured as authorization webhook for the garden cluster. At the moment, the backing authorizer never has an opinion on the request. In the future, it will be enhanced with fine-granular checks for requests made by gardenlets to provide them only the bare minimum permission and restrict their requests to resources related to their respective seeds.

Part of #1723

/invite @timuthy 
/assign @timuthy 
/squash

In the local development scenario, the auth webhook is disabled by default to prevent forcing developers to start the GAC.

**Release note**:

```other operator
NONE
```


hack/local-development/local-garden/run-kube-apiserver
```diff
@@ -10,19 +10,48 @@ if ! ( [[ "$(uname -s)" == *"Darwin"* ]] || [[ "$(uname -s)" == "Linux" && "$(un
...
+MOUNTS="-v $SCRIPTPATH/certificates/certs:/certs -v $SCRIPTPATH/certificates/keys:/keys -v $SCRIPTPATH/configs:/configs"
 ETCD_PORT=12379
 LISTEN_PORT="2443"
 
+ADMISSION_CONTROLLER_EXTERNAL_NAME=gardener.localhost
+if [[ "$(uname -s)" == *"Darwin"* ]] || [[ "$(uname -s)" == "Linux" && "$(uname -r)" =~ "microsoft-standard" ]] ; then
+  ADMISSION_CONTROLLER_EXTERNAL_NAME=host.docker.internal
+fi
+
+cat<< EOF > "$SCRIPTPATH/configs/authwebhook.yaml"
```
timuthy, line 26: Can we create this in the `dev` dir?

hack/local-development/local-garden/start.sh
```diff
@@ -46,4 +48,4 @@ $(dirname $0)/apply-rbac-garden-ns
 
 echo "# Now you can start using the cluster at with \`export KUBECONFIG=hack/local-development/local-garden/kubeconfigs/default-admin.conf\`"
 echo "# Then you need to run \`make dev-setup\` to setup config and certificates files for gardener's components and to register the gardener-apiserver."
-echo "# Finally, run \`make start-apiserver,start-controller-manager,start-scheduler,start-gardenlet\` to start the gardener components as usual."
\ No newline at end of file
+echo "# Finally, run \`make start-apiserver,start-controller-manager,start-scheduler,start-gardenlet\` to start the gardener components as usual."
```
timuthy, line 51: How about the following or something similar?
```
if $ACTIVATE_SEEDAUTHORIZER; then
  echo '# Additionally, make sure to run make start-admission-controller'
fi
```

pkg/admissioncontroller/webhooks/auth/seed/handler.go
```diff
@@ -0,0 +1,175 @@
...
+		h.log.Error(err, "unable to process a request with an unknown content type", "content type", contentType)
+		h.writeResponse(w, nil, Errored(http.StatusBadRequest, err))
+		return
+	}
+
+	// Both v1 and v1beta1 SubjectAccessReview types are exactly the same, so the v1beta1 type can be decoded into the
+	// v1 type. However the runtime codec's decoder guesses which type to decode into by type name if an Object's
+	// TypeMeta isn't set. By setting TypeMeta of an unregistered type to the v1 GVK, the decoder will coerce a v1beta1
+	// SubjectAccessReview to v1. The actual SubjectAccessReview GVK will be used to write a typed response in case the
+	// webhook config permits multiple versions, otherwise this response will fail.
```
timuthy: I noticed a small difference in the `SubjectAccessReviewSpec` between `v1` and `v1beta1`:

[v1beta1](https://github.com/kubernetes/api/blob/48bd8381a38a486f8b3cdf28cf7334a45b182f2e/authorization/v1beta1/types.go#L149):
```Groups []string `json:"group```

[v1](https://github.com/kubernetes/api/blob/48bd8381a38a486f8b3cdf28cf7334a45b182f2e/authorization/v1/types.go#L140):
```Groups []string `json:"groups```

So depending on what version is sent by the API server, I suspect that we lose the group information during decode,  right?

**Review by timuthy — CHANGES_REQUESTED** [review]
Nice feature to kick this topic off! A few comments from my side, otherwise it looks good to me.

File: hack/local-development/local-garden/start.sh (hunk 2 above)
Comment (rfranzke) [reply hack/local-development/local-garden/start.sh:51]: Will do

File: pkg/admissioncontroller/webhooks/auth/seed/handler.go (hunk 3 above)
Comment (rfranzke) [reply pkg/admissioncontroller/webhooks/auth/seed/handler.go]: 👀 nice catch, I will check it

File: hack/local-development/local-garden/run-kube-apiserver (hunk 1 above)
Comment (rfranzke) [reply hack/local-development/local-garden/run-kube-apiserver:26]: We can, but I avoided this to not "smear" the different configuration for the local kube-apiserver in too many places. I think it's better to have everything in one place.

Comment (timuthy) [reply hack/local-development/local-garden/run-kube-apiserver:26]: Hm, to me the `dev` dir was always such a place. Everything that is generated, configuration related or data for development lands there and is ignored by Git i.e., no additional `.gitignore` and once place to go to change configuration. The etcd data dir for our local K8s is also already stored in  `dev`.

I think the certs and kubeconfigs are only inside the `local-garden` dir to check them in which we ought to get rid of in the future.

WDYT?

Comment (rfranzke) [reply hack/local-development/local-garden/run-kube-apiserver:26]: I'm fine to either have everything in `hack` or everything in `dev`, but not scattered around both places (OK, it's partly already like this, because the etcd data dir is in `dev`, however, I don't like this and don't want to make it worse). So not sure, I'm still not convinced. Even if the target picture is to have everything (including the certs, etc.) in `dev` then I would still vote to keep it as it is for now and rather refactor it consistently with this subsequent endeavour.

Also, (at least that's my understanding) `dev` was more meant to contain files that the developer might want to tweak (e.g., component configuration or shoot examples). This auth configuration is not of this kind, it's static and it's highly unlikely that a developer ever needs to touch or even care about it. It's just that there is one setting that is depending on the operating system, so it can't be "hard-coded" like the other files for the local garden. Anyways, that's just a minor detail and not too relevant - my main motivation is consistency.

Comment (timuthy) [reply hack/local-development/local-garden/run-kube-apiserver:26]: > my main motivation is consistency

Mine is too and since we've never defined or discussed the structure, the way I described it was the way I understood it (`dev` == generated config files for dev, ignored by Git, `hack` == static and checked-in assets). 

> Also, (at least that's my understanding) dev was more meant to contain files that the developer might want to tweak

Okay fine, let's follow your definition then.

Comment (rfranzke) [reply hack/local-development/local-garden/run-kube-apiserver:26]: Yeah, let's continue for now as this is minor with little impact anyways. If you already have ideas how to improve it then it can be done consistently in the future. 👍🏻

**Comment by rfranzke** [comment]
I have worked on the boilerplate for the `SeedRestriction` admission plugin which will also be part of GAC. It's not much code and closely related, so I plan to also push it as separate commit to this PR. I hope that's fine. If you prefer to rather have it separately then I can also extract it and open a dedicated PR. I'll ping you once it's ready for review again.

File: pkg/admissioncontroller/webhooks/auth/seed/handler.go (hunk 3 above)
Comment (rfranzke) [reply pkg/admissioncontroller/webhooks/auth/seed/handler.go]: Hi @timuthy, I've addressed this in 031f21a, please check it out. I'm not too happy with it, but I didn't have better ideas. Maybe you have one?

**Comment by rfranzke** [comment]
I've extended the PR as planned and added the `SeedRestriction` admission plugin boilerplate with c191d8a.

Also, I have added a dedicated commit (70d8bbc) with documentation including the following flow diagram: 

image

File: hack/local-development/dev-setup-register-gardener [hunk 4]
```diff
@@ -352,5 +352,13 @@ $ADMISSION_CONTROLLER_PORT_STRING
...
+  admissionReviewVersions: ["v1beta1"]
```
timuthy: Let's change it to `v1` when https://github.com/gardener/gardener/pull/3577 is merged.
```suggestion
  admissionReviewVersions: ["v1"]
```

pkg/admissioncontroller/webhooks/auth/seed/handler.go
```diff
@@ -0,0 +1,232 @@
...
+			"allowed", status.Allowed,
+			"denied", status.Denied,
+			"reason", status.Reason,
+			"error", status.EvaluationError,
+		)
+		log.V(1).Info("wrote response")
+	}
+}
+
+func (h *handler) decodeRequestBody(body []byte) (*authorizationv1beta1.SubjectAccessReviewSpec, *schema.GroupVersionKind, error) {
```
timuthy: Why not `authorizationv1`? 👀

pkg/admissioncontroller/webhooks/auth/seed/handler.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+		return &tmp.Spec, gvk, nil
+
+	case authorizationv1.SchemeGroupVersion.WithKind("SubjectAccessReview"):
+		var tmp authorizationv1.SubjectAccessReview
+		if err := scheme.Convert(&obj, &tmp, nil); err != nil {
+			return nil, nil, err
+		}
+
+		// "convert" authorizationv1.SubjectAccessReviewSpec to authorizationv1beta1.SubjectAccessReviewSpec
```
timuthy: How about if we leave it like before i.e., always decode to `authorizationv1` and if the API server sent `authorizationv1beta1`, decode it to `authorizationv1beta1` + only feed the `sarSpec.Groups` portion (because that's the difference between the two API). Maybe I'm on the wrong track but I think this would spare us some boilerplate?! Anyways, please consider it as a nit. I'm also fine with the current handling.

timuthy (changes requested):

pkg/admissioncontroller/webhooks/auth/seed/handler.go
rfranzke: According to https://kubernetes.io/docs/reference/access-authn-authz/webhook/#request-payloads:

> When faced with an authorization decision, the API Server POSTs a JSON- serialized authorization.k8s.io/v1beta1 SubjectAccessReview object describing the action.

timuthy: Okay, thanks. It's also [configureable](https://github.com/kubernetes/kubernetes/pull/84768) since `v1.17`, so we can keep it in mind when we drop `v1.16` support for Garden clusters.

timuthy (approved):
Awesome! Thanks a lot for incorporating the feedback 👍 
/lgtm

timuthy:
@BeckerMax are you still reviewing / want to provide feedback?

BeckerMax:
Thanks for asking. No further comments from my side.

PR #3573, opened 2021-02-16
[GEP-13] Add managed seed integration test

stoyanr:

/area control-plane
/kind test
/priority normal

Adds integration tests for `ManagedSeed` creation and deletion

Fixes #

* Intended to eventually replace `seed_registration` and `seed_deletion` tests. Currently these tests are only executed as part of the CP migration integration test flow. Once they are not needed anymore, they will be removed in a separate PR.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.test-defs/DeleteManagedSeed.yaml
+++ b/.test-defs/DeleteManagedSeed.yaml
@@ -1,0 +1,16 @@
+kind: TestDefinition
+metadata:
+  name: delete-managed-seed
+spec:
+  owner: gardener-oq@listserv.sap.com
+  description: Tests the deletion of a managed seed.
+  activeDeadlineSeconds: 4200
+
+  command: [bash, -c]
+  args:
+  - >-
+    go test -timeout=0 -mod=vendor ./test/system/managed_seed_deletion
+    --v -ginkgo.v -ginkgo.progress -ginkgo.noColor
+    -kubecfg="$TM_KUBECONFIG_PATH/gardener.config"
+    -managed-seed-name=$MANAGED_SEED_NAME
+  image: golang:1.15.3
--- a/test/framework/gardener_utils.go
+++ b/test/framework/gardener_utils.go
@@ -22,6 +22,7 @@ import (
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	gardencorev1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	seedmanagementv1alpha1 "github.com/gardener/gardener/pkg/apis/seedmanagement/v1alpha1"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
 	"github.com/gardener/gardener/pkg/operation/common"
 	"github.com/gardener/gardener/pkg/scheduler/apis/config"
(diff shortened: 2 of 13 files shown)


stoyanr:
/cc @timebertt @kris94 

Currently `verify` fails with 
```
/go/src/github.com/gardener/gardener/extensions/pkg/controller/backupbucket/backupbucket_test.go:64

  Expected
      <v1beta1.LastOperationState>: Processing
  to equal
      <v1beta1.LastOperationState>: Succeeded
```
Not sure what's wrong, but it's for sure not caused by changes in this PR. I also can't reproduce this error locally.

test/framework/managedseedframework.go
```diff
@@ -0,0 +1,267 @@
...
+
+// CreateManagedSeed creates a new managed seed according to the managed seed framework configuration.
+func (f *ManagedSeedFramework) CreateManagedSeed(ctx context.Context) error {
+	if f.GardenClient == nil {
+		return errors.New("no gardener client is defined")
+	}
+
+	// Get shoot
+	shoot := &gardencorev1beta1.Shoot{}
+	if err := f.GardenClient.DirectClient().Get(ctx, kutil.Key(v1beta1constants.GardenNamespace, f.Config.ShootName), shoot); err != nil {
```
schrodit, line 156: does the shoot have to reside in the garden namespace?

.test-defs/DeleteManagedSeed.yaml
```diff
@@ -0,0 +1,16 @@
+kind: TestDefinition
+metadata:
+  name: delete-managed-seed
+spec:
+  owner: gardener-oq@listserv.sap.com
+  description: Tests the deletion of a managed seed.
+  activeDeadlineSeconds: 3600
```
schrodit: the graceful test delete timeout is also 1 hours so the testmachinery could kill the test before the test runs into its own timeout.
Lets increase it by 10 minutes to be save.

schrodit (changes requested):

test/framework/managedseedframework.go
stoyanr: Yes. It has to be in the same namespace as the `ManagedSeed` resource, and `ManagedSeed` resources are currently restricted to the `garden` namespace. In the future this might change if something like "private seeds" is introduced but there are no concrete plans for this.

.test-defs/DeleteManagedSeed.yaml
stoyanr: I am not sure I understand this. If this is the case, shouldn't we actually decrease the timeout by 10 minutes rather than increasing it? Also, is this only applicable to the `DeleteManagedSeed` definition but not to the `CreateManagedSeed`?

schrodit: so currently there are 2 timeouts, one that is used by the testmachinery and one that is used by the go test.

When the go test timeout is reached, the test get into fail state and the cleanup handlers + dump of the state executed.
When the testmachinery timeout is reached the pod, that runs the go test, will get deleted. This means that if the testmachinery is reached before the go test timout no celanup handlers or dumps are executed.

So it is better to have the testmachinery timeout only as safety mechanism if the test timout does not work.

> Also, is this only applicable to the DeleteManagedSeed definition but not to the CreateManagedSeed?
Its also applicable in the create managed seed definition

stoyanr: OK, I think I got it now. Increased this timeout to 4200 as suggested, the `CreateManagedSeed` timeout was already 7200 so no need to touch it I guess.

dguendisch (commented):
/lgtm

schrodit (dismissed):

test/framework/gardener_utils.go
```diff
@@ -666,3 +667,75 @@ func (f *GardenerFramework) WaitForSeedToBeDeleted(ctx context.Context, seed *ga
 		return retry.MinorError(fmt.Errorf("seed %q still exists", seed.Name))
 	})
 }
+
+// CreateManagedSeed creates a new managed seed and waits for it to be created and successfully reconciled.
+func (f *GardenerFramework) CreateManagedSeed(ctx context.Context, managedSeed *seedmanagementv1alpha1.ManagedSeed) error {
+	// Create the managed seed
+	err := retry.UntilTimeout(ctx, 20*time.Second, 5*time.Minute, func(ctx context.Context) (done bool, err error) {
+		err = f.GardenClient.DirectClient().Create(ctx, managedSeed)
```
timebertt: `Create/Update/Delete/Patch` always go directly to the API server, so no need to use the `DirectClient` for those calls.
```suggestion
		err = f.GardenClient.Client().Create(ctx, managedSeed)
```
Note to self: I really need to clean this up and make it a `APIReader` instead. Will hopefully start with this next week as part of https://github.com/gardener/gardener/issues/2822.

test/framework/resources/templates/default-shoot.yaml
```diff
@@ -28,4 +28,6 @@ spec:
     kubernetesDashboard:
       enabled: true
       authenticationMode: token
+  tolerations:
+  - key: seed.gardener.cloud/protected
```
timebertt: @schrodit Is this template used in the usual create shoot tests?
If so, should we avoid to tolerate protected seeds?
There might be protected seeds in the system for development or testing purposes and we don't necessarily want to schedule TM shoots on them.

What about using a taint specific to the current test run to protect test Seeds, so that only shoots are scheduled on them if really intended?

timebertt (commented):
Thanks!
Two minor comments from my side. Rest looks good.

test/framework/resources/templates/default-shoot.yaml
dguendisch: Good catch! Yes that template is used for the test shoot creations and we certainly don't want test shoots generally to be scheduled on to some private test seeds on dev or staging for example.
So you'd suggest to taint a test seed with `testrun-123` and then adapt the shoot creation to tolerate `testrun-123`? Sounds like a good idea to me.

timebertt: Yep, something like this, is what I had in mind.
Probably [`TM_TESTRUN_ID`](https://github.com/gardener/test-infra/blob/f11e8da17be3d35447b4498d7763ac61056bcfaf/pkg/testmachinery/config.go#L72) can be used for that purpose by the managed seed creation and shoot creation to add taints and tolerations.

stoyanr: This means I should change the `create-shoot` integration test, adding an additional parameter to pass custom tolerations, correct? And then I should pass a toleration based on `$TM_TESTRUN_ID` in `CreateShoot.yaml`? And similarly for the `create-managed-seed` test?

dguendisch: Yes, sounds a bit like overhead, but we really need to avoid scheduling test shoots onto some probably broken test seeds. Unless @schrodit has a better idea?

timebertt: > adding an additional parameter to pass custom tolerations, correct?

I don't think, we need an extra option for passing custom tolerations for now.
We can probably just add a toleration `test.gardener.cloud/test-run=$TM_TESTRUN_ID` (or a similar key) to every shoot created by any integration test, as it doesn't hurt.
In the managed seed test we could then add a corresponding taint (also no custom one needed for now) in any case.
This would basically make the managed seed a "private seed" for just this TestRun.

If we create a `ManagedSeed` and some shoots in the same TestRun they would be able to be scheduled there.
Though, if we want to use the `ManagedSeed` in some other TestRun, we would need to think of a different/shared taint. But I assume, this wouldn't be required for now.

WDYT?

stoyanr: Yes, makes sense. Will do as suggested.

stoyanr: Nice idea, and I even implemented it, but when I tried to actually execute the integration test I got the following error:
```
time="2021-02-23T15:58:59Z" level=fatal msg="Cannot create shoot tm-x93-ruf-seed: shoots.core.gardener.cloud \"tm-x93-ruf-seed\" is forbidden: error while validating tolerations against allowlist: [spec.tolerations[0]: Forbidden: only the following tolerations are allowed: [seed.gardener.cloud/protected]]"
```
I figured out that this whitelist is per project, so we would need to ensure that the above toleration is properly added to the 2 projects we are using for integration tests (`garden` and `it`). @timebertt @dguendisch I will contact you offline for the proper way to do this.

test/framework/shoot_utils.go
```diff
@@ -295,6 +298,16 @@ func setShootNetworkingSettings(shoot *gardencorev1beta1.Shoot, cfg *ShootCreati
 	}
 }
 
+// setShootNetworkingSettings sets the Shoot's networking settings from the given config
```
timebertt: nit: probably a copy+paste mistake

test/framework/resources/templates/default-shoot.yaml
timebertt: I would even say, we could always allow this taint. It's a well-known taint key which is reserved for Integration Tests, so no one should use it for anything other than that, which makes it harmless to add / always allow.

test/framework/shoot_utils.go
stoyanr: Fixed, thanks

test/framework/gardener_utils.go
stoyanr: Done

test/framework/resources/templates/default-shoot.yaml
stoyanr: @dguendisch Will take care, so I consider this done as far as this PR is concerned.

stoyanr:
@timebertt @dguendisch I think I addressed all feedback, and there are now green playground integration tests with the latest changes. From my perspective this PR is ready to be merged.

timebertt (approved):
Very nice, looks great, thank you!
/lgtm

PR #3575, opened 2021-02-17
Refactor Seed (de)bootstrap to flow graph

BeckerMax:
  /area control-plane
/kind technical-debt

This PR addresses different issues all around seed de(bootstrap):
- Clean-up of managed Ingress DNS-related resources (DNSEntry, ControllerInstallation, DNSProvider)  when deleting a Seed
- Ensuring that the gardener resource manager gets deployed before/deleted after all components with managed resources.

We implemented a quick fix with #3543 which is reverted with this change and properly implemented using a flow graph.

**Special notes for your reviewer**

I decided to not use the DNSProvider component (used in the Shoot bootstrap) because the use case differs and forces me to  to implement around it so that the benefits of using it are minimal. The problems of the dnsProvider component were:
- it uses a fixed provider secret name which differs from the secret name currently deployed (we would need some extra clean-up logic).
- on destroy() the secret will not be updated (but it is needed in the seed destroy case)
- the secret will not be deleted upon deletion of the dnsProvider

I introduced a change to how wantedControllerRegistrationNames are calculated, see [wantedControllerRegistrationNames](https://github.com/gardener/gardener/pull/3575/files#diff-0d8996ff3546653f546482d70a198c7c4da27e0102f684a87d5e982df34896d4R350).
Except from refactoring naming I changed that In case of deletion only already installed and required installations are added. That leads to the dns-controllerinstallation being removed when it is no longer required. We can also check in  [computeKindTypesForSeed](https://github.com/gardener/gardener/pull/3575/files#diff-0d8996ff3546653f546482d70a198c7c4da27e0102f684a87d5e982df34896d4L289) and not add the kind/type if the seed is in deletion.

I moved checking associatedControllerInstallations from [from seed_control.go](https://github.com/gardener/gardener/pull/3575/files#diff-628c8891682f4a4b0f5c49cb8051057ed8a5a3bf338438659ef68ccfb66385e5R243) [to the flow graph](https://github.com/gardener/gardener/pull/3575/files#diff-8495b57b25d39744fe1b20b1e1ce695d130f730142af86ff9c5929afcdbabeb4R1001) originally this was introduced in https://github.com/gardener/gardener/pull/3653 
cc @timebertt in . I did this because the dns-controllerinstallation is needed for cleaning up the dnsentry (which was not properly cleaned up before)

**Release note**:

```other operator
On Seed deletion resources related to the managed Ingress are now properly cleaned up.
```

### DIFF
--- a/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
+++ b/pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
@@ -297,6 +297,11 @@ func computeKindTypesForSeed(
 ) sets.String {
 	var wantedKindTypeCombinations = sets.NewString()
 
+	// enable clean up of controller installations in case of seed deletion
+	if seed.DeletionTimestamp != nil {
+		return sets.NewString()
+	}
+
 	if seed.Spec.DNS.Provider != nil {
 		wantedKindTypeCombinations.Insert(common.ExtensionID(dnsv1alpha1.DNSProviderKind, seed.Spec.DNS.Provider.Type))
 	}
@@ -330,7 +335,7 @@ func computeControllerRegistrationMaps(
 // computeWantedControllerRegistrationNames computes the list of names of ControllerRegistration objects that are desired
 // to be installed. The computation is performed based on a list of required kind/type combinations and the proper mapping
 // to existing ControllerRegistration objects. Additionally, all names in the alwaysPolicyControllerRegistrationNames list
-// will be returned.
+// will be returned and all currently installed and required installations.
 func computeWantedControllerRegistrationNames(
 	wantedKindTypeCombinations sets.String,
 	controllerInstallationList *gardencorev1beta1.ControllerInstallationList,
@@ -358,27 +363,32 @@ func computeWantedControllerRegistrationNames(
 		}
 	}
 
-	for _, requiredExtension := range wantedKindTypeCombinations.UnsortedList() {
-		names, ok := kindTypeToControllerRegistrationNames[requiredExtension]
+	for _, wantedExtension := range wantedKindTypeCombinations.UnsortedList() {
+		names, ok := kindTypeToControllerRegistrationNames[wantedExtension]
 		if !ok {
-			return nil, fmt.Errorf("need to install an extension controller for %q but no appropriate ControllerRegistration found", requiredExtension)
+			return nil, fmt.Errorf("need to install an extension controller for %q but no appropriate ControllerRegistration found", wantedExtension)
 		}
-
 		wantedControllerRegistrationNames.Insert(names...)
 	}
 
+	wantedControllerRegistrationNames.Insert(installedAndRequiredRegistrationNames(controllerInstallationList, seedObjectMeta.Name).List()...)
+
+	// filter controller registrations with non-matching seed selector
+	return controllerRegistrationNamesWithMatchingSeedLabelSelector(wantedControllerRegistrationNames.UnsortedList(), controllerRegistrations, seedObjectMeta.Labels)
+}
+
+func installedAndRequiredRegistrationNames(controllerInstallationList *gardencorev1beta1.ControllerInstallationList, seedName string) sets.String {
+	requiredControllerRegistrationNames := sets.NewString()
 	for _, controllerInstallation := range controllerInstallationList.Items {
-		if controllerInstallation.Spec.SeedRef.Name != seedObjectMeta.Name {
+		if controllerInstallation.Spec.SeedRef.Name != seedName {
 			continue
 		}
 		if !gardencorev1beta1helper.IsControllerInstallationRequired(controllerInstallation) {
 			continue
 		}
-		wantedControllerRegistrationNames.Insert(controllerInstallation.Spec.RegistrationRef.Name)
+		requiredControllerRegistrationNames.Insert(controllerInstallation.Spec.RegistrationRef.Name)
 	}
-
-	// filter controller registrations with non-matching seed selector
-	return controllerRegistrationNamesWithMatchingSeedLabelSelector(wantedControllerRegistrationNames.UnsortedList(), controllerRegistrations, seedObjectMeta.Labels)
+	return requiredControllerRegistrationNames
 }
 
 // computeRegistrationNameToInstallationNameMap computes a map that maps the name of a ControllerRegistration to the name of an
--- a/pkg/gardenlet/controller/seed/seed_control.go
+++ b/pkg/gardenlet/controller/seed/seed_control.go
@@ -247,24 +247,17 @@ func (c *defaultControl) ReconcileSeed(obj *gardencorev1beta1.Seed, key string) 
 			return err
 		}
 
-		associatedControllerInstallations, err := controllerutils.DetermineControllerInstallationAssociations(ctx, gardenClient.Client(), seed.Name)
-		if err != nil {
-			seedLogger.Error(err.Error())
-			return err
-		}
-
 		associatedBackupBuckets, err := controllerutils.DetermineBackupBucketAssociations(ctx, gardenClient.Client(), seed.Name)
 		if err != nil {
 			seedLogger.Error(err.Error())
 			return err
 		}
 
-		if len(associatedShoots) == 0 && len(associatedControllerInstallations) == 0 && len(associatedBackupBuckets) == 0 {
+		if len(associatedShoots) == 0 && len(associatedBackupBuckets) == 0 {
 			seedLogger.Info("No Shoots, ControllerInstallations or BackupBuckets are referencing the Seed. Deletion accepted.")
 
-			// Debootstrap the Seed cluster.
-			if err := seedpkg.DebootstrapCluster(ctx, seedClient); err != nil {
-				message := fmt.Sprintf("Failed to unbootstrap Seed Cluster (%s).", err.Error())
+			if err := seedpkg.RunDeleteSeedFlow(ctx, seedClient, gardenClient, seedObj, seedLogger); err != nil {
+				message := fmt.Sprintf("Failed to delete Seed Cluster (%s).", err.Error())
 				conditionSeedBootstrapped = gardencorev1beta1helper.UpdatedCondition(conditionSeedBootstrapped, gardencorev1beta1.ConditionFalse, "DebootstrapFailed", message)
 				seedLogger.Error(message)
 				_ = c.updateSeedStatus(ctx, gardenClient.Client(), seed, "<unknown>", capacity, allocatable, conditionSeedBootstrapped)
@@ -305,11 +298,6 @@ func (c *defaultControl) ReconcileSeed(obj *gardencorev1beta1.Seed, key string) 
 		parentLogMessage := "Can't delete Seed, because the following objects are still referencing it:"
 		if len(associatedShoots) != 0 {
 			message := fmt.Sprintf("%s Shoots=%v", parentLogMessage, associatedShoots)
-			seedLogger.Info(message)
-			c.recorder.Event(seed, corev1.EventTypeNormal, v1beta1constants.EventResourceReferenced, message)
-		}
-		if len(associatedControllerInstallations) != 0 {
-			message := fmt.Sprintf("%s ControllerInstallations=%v", parentLogMessage, associatedControllerInstallations)
 			seedLogger.Info(message)
 			c.recorder.Event(seed, corev1.EventTypeNormal, v1beta1constants.EventResourceReferenced, message)
 		}
@@ -381,7 +369,7 @@ func (c *defaultControl) ReconcileSeed(obj *gardencorev1beta1.Seed, key string) 
 	}
 
 	// Bootstrap the Seed cluster.
-	if err := seedpkg.BootstrapCluster(ctx, gardenClient, seedClient, seedObj, gardenSecrets, c.imageVector, c.componentImageVectors, c.config.DeepCopy()); err != nil {
+	if err := seedpkg.RunReconcileSeedFlow(ctx, gardenClient, seedClient, seedObj, gardenSecrets, c.imageVector, c.componentImageVectors, c.config.DeepCopy(), seedLogger); err != nil {
 		conditionSeedBootstrapped = gardencorev1beta1helper.UpdatedCondition(conditionSeedBootstrapped, gardencorev1beta1.ConditionFalse, "BootstrappingFailed", err.Error())
 		_ = c.updateSeedStatus(ctx, gardenClient.Client(), seed, seedKubernetesVersion, capacity, allocatable, conditionSeedBootstrapped)
 		seedLogger.Errorf("Seed bootstrapping failed: %+v", err)
--- a/pkg/operation/seed/seed.go
+++ b/pkg/operation/seed/seed.go
@@ -17,7 +17,6 @@ package seed
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"path/filepath"
 	"strings"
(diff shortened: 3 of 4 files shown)


pkg/operation/seed/seed.go
```diff
@@ -239,7 +237,14 @@ func deployCertificates(ctx context.Context, seed *Seed, k8sSeedClient kubernete
 }
 
 // BootstrapCluster bootstraps a Seed cluster and deploys various required manifests.
-func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubernetes.Interface, seed *Seed, secrets map[string]*corev1.Secret, imageVector imagevector.ImageVector, componentImageVectors imagevector.ComponentImageVectors, conf *config.GardenletConfiguration) error {
+func BootstrapCluster(ctx context.Context,
+	k8sGardenClient, k8sSeedClient kubernetes.Interface,
+	seed *Seed, secrets map[string]*corev1.Secret,
```
```suggestion
func BootstrapCluster(
	ctx context.Context,
	k8sGardenClient kubernetes.Interface,
	k8sSeedClient kubernetes.Interface,
	seed *Seed,
	secrets map[string]*corev1.Secret,
```

If you use multiple lines then let's do it consistently :)

pkg/gardenlet/controller/seed/seed_control.go
```diff
@@ -253,9 +253,8 @@ func (c *defaultControl) ReconcileSeed(obj *gardencorev1beta1.Seed, key string)
 		if len(associatedShoots) == 0 && len(associatedBackupBuckets) == 0 {
 			seedLogger.Info("No Shoots or BackupBuckets are referencing the Seed. Deletion accepted.")
 
-			// Debootstrap the Seed cluster.
-			if err := seedpkg.DebootstrapCluster(ctx, seedClient); err != nil {
-				message := fmt.Sprintf("Failed to unbootstrap Seed Cluster (%s).", err.Error())
+			if err := seedpkg.RunDeleteSeedFlow(seedClient, seedLogger); err != nil {
```
rfranzke: Hm, why was this renamed? I would expect to either call this `DebootstrapCluster` or rename `BootstrapCluster` to `RunReconcileSeedFlow`.

pkg/operation/seed/seed.go
```diff
@@ -802,7 +807,7 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
 	}
 
 	// managed nginx-ingress handling
-	if err := handleIngressDNSEntry(ctx, k8sSeedClient, chartApplier, seed); err != nil {
+	if err := handleIngressDNSEntry(ctx, k8sSeedClient, chartApplier, seed, seedLogger); err != nil {
```
rfranzke: Why is this no regular task function?

rfranzke (commented):
I don't like that the `bootstrapComponents` function was deleted and that you now have "duplicated" the code to create the components. Can you revert this?

BeckerMax:
> I don't like that the bootstrapComponents function was deleted and that you now have "duplicated" the code to create the components. Can you revert this?

You mean I have two functions [runCreateSeedFlow](https://github.com/gardener/gardener/pull/3575/files#diff-8495b57b25d39744fe1b20b1e1ce695d130f730142af86ff9c5929afcdbabeb4R830) and [RunDeleteSeedFlow](https://github.com/gardener/gardener/pull/3575/files#diff-628c8891682f4a4b0f5c49cb8051057ed8a5a3bf338438659ef68ccfb66385e5R256) now that both call the constructors of the components?
I do not see a way around this. Before you would just add all components and run deploy/destroy in parallel. Now we need an explicit ordering which differs in creation and deletion. So I can not just add all components to an array and use that in both scenarios (how it was before). As a plus I also could get rid of the nil checks and implicit meaning that `ImageVector == nil` means deletionFlow which I found confusing.

rfranzke:
> You mean I have two functions runCreateSeedFlow and RunDeleteSeedFlow now that both call the constructors of the components?

Yes, exactly.

> Before you would just add all components and run deploy/destroy in parallel. Now we need an explicit ordering which differs in creation and deletion. So I can not just add all components to an array and use that in both scenarios (how it was before). As a plus I also could get rid of the nil checks and implicit meaning that ImageVector == nil means deletionFlow which I found confusing.

Hm, OK... true. 😢 Fine then, let's proceed for now as you suggest.

pkg/gardenlet/controller/seed/seed_control.go
BeckerMax: I would suggest to follow the naming of the ShootFlow as you proposed and go with BootstrapCluster -> RunReconcileSeedFlow

BeckerMax:
ready for review again

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -347,27 +347,35 @@ func computeWantedControllerRegistrationNames(
 		}
 	}
 
-	for _, requiredExtension := range wantedKindTypeCombinations.UnsortedList() {
-		names, ok := kindTypeToControllerRegistrationNames[requiredExtension]
-		if !ok {
-			return nil, fmt.Errorf("need to install an extension controller for %q but no appropriate ControllerRegistration found", requiredExtension)
+	// In case of deletion only add already installed and required installations
```
rfranzke: This might work for the extensions of a `Seed`, but I don't think that we forbid creating `Shoot`s on `Seed`s already having a deletion timestamp. Anyhow, maybe it's indeed a good idea to move this handling to the function specific for seeds.

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
-		return err
-	}
+	return runCreateSeedFlow(ctx, k8sSeedClient, k8sGardenClient, imageVector, imageVectorOverwrites, seed, seedLogger)
+}
 
-	// Deploy component specific resources
-	bootstrapComponents, err := bootstrapComponents(k8sSeedClient, v1beta1constants.GardenNamespace, imageVector, imageVectorOverwrites)
+func runCreateSeedFlow(ctx context.Context, sc, gc kubernetes.Interface, imageVector imagevector.ImageVector, imageVectorOverwrites map[string]string, seed *Seed, seedLogger *logrus.Entry) error {
+	var (
+		g         = flow.NewGraph("Seed cluster creation")
```
rfranzke: Can you move this to line 910 where you also add the tasks?

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+		namespace = v1beta1constants.GardenNamespace
```
rfranzke: Personally, I don't find variables like this helpful.

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
-		})
+	if err := updateDNSProviderSecret(ctx, sc, gc, seed); err != nil {
+		return err
 	}
 
-	return flow.Parallel(bootstrapFunctions...)(ctx)
-}
+	// setup for flow graph
+	var dnsEntry component.DeployWaiter
+	dnsConfig := seed.Info.Spec.DNS
```
rfranzke: same here

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+	if managedIngress(seed) {
+		ingressLoadBalancerAddress, err := getManagedIngressLoadBalancerAddress(ctx, sc, seedLogger)
+		if err != nil {
+			return err
+		}
+		dnsEntry = getManagedIngressDNSEntry(sc, seed.GetIngressFQDN("*"), ingressLoadBalancerAddress, seedLogger)
+	} else {
+		dnsEntry = getManagedIngressDNSEntry(sc, seed.GetIngressFQDN("*"), "", seedLogger)
+	}
```
```suggestion
	var ingressLoadBalancerAddress string
	if managedIngress(seed) {
		ingressLoadBalancerAddress, err = getManagedIngressLoadBalancerAddress(ctx, sc, seedLogger)
		if err != nil {
			return err
		}
	}

	dnsEntry := getManagedIngressDNSEntry(sc, seed.GetIngressFQDN("*"), ingressLoadBalancerAddress, seedLogger)
```

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
-		debootstrapFunctions = append(debootstrapFunctions, func(ctx context.Context) error {
-			return component.OpDestroyAndWait(fn).Destroy(ctx)
-		})
+	cfg := resourcemanager.Values{
+		ConcurrentSyncs:  pointer.Int32Ptr(20),
+		HealthSyncPeriod: utils.DurationPtr(time.Minute),
+		ResourceClass:    pointer.StringPtr(v1beta1constants.SeedResourceManagerClass),
+		SyncPeriod:       utils.DurationPtr(time.Hour),
 	}
+	rm := resourcemanager.New(sc.Client(), namespace, grmImage.String(), 1, cfg)
```
rfranzke: Personal preference:

```suggestion
	resourceManager := resourcemanager.New(sc.Client(), namespace, grmImage.String(), 1, resourcemanager.Values{
		ConcurrentSyncs:  pointer.Int32Ptr(20),
		HealthSyncPeriod: utils.DurationPtr(time.Minute),
		ResourceClass:    pointer.StringPtr(v1beta1constants.SeedResourceManagerClass),
		SyncPeriod:       utils.DurationPtr(time.Hour),
	})
```

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+		repository = gardenerSeedAdmissionControllerImage.Repository
+		tag = *gardenerSeedAdmissionControllerImage.Tag
 	}
-	return errors.New("cannot delete gardener-resource-manager there are managed resources in the garden namespace left")
-}
+	gsacImage := imagevector.Image{
+		Repository: repository,
+		Tag:        &tag,
+	}
+	seedAdmission := seedadmissioncontroller.New(sc.Client(), namespace, gsacImage.String(), kubernetesVersion)
```
rfranzke: You use three names to describe the "same" thing: `gardenerSeedAdmissionController`, `gsac`, `seedAdmission`. I suggest to streamline this.

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
 
-func deployGardenerResourceManager(ctx context.Context, c kubernetes.Interface, namespace string, imageVector imagevector.ImageVector) error {
-	image, err := imageVector.FindImage(charts.ImageNameGardenerResourceManager, imagevector.RuntimeVersion(c.Version()), imagevector.TargetVersion(c.Version()))
+	schedulerImage, err := imageVector.FindImage(charts.ImageNameKubeScheduler, imagevector.TargetVersion(kubernetesVersion.String()))
+	if err != nil {
+		return err
+	}
+	sched, err := gardenerkubescheduler.Bootstrap(sc.DirectClient(), namespace, schedulerImage, kubernetesVersion)
```
```suggestion
	scheduler, err := gardenerkubescheduler.Bootstrap(sc.DirectClient(), namespace, schedulerImage, kubernetesVersion)
```

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+			Name:         "Deploying kube-scheduler for shoot control plane pods",
+			Fn:           sched.Deploy,
+			Dependencies: flow.NewTaskIDs(deployResourceManager),
+		})
+	)
+
+	f := g.Compile()
+	if err := f.Run(flow.Opts{
+		Logger: seedLogger,
+	}); err != nil {
```
```suggestion
	if err := g.Compile().Run(flow.Opts{Logger: seedLogger}); err != nil {
```

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
 	}
-	rm := resourcemanager.New(c.Client(), namespace, image.String(), 1, cfg)
-	return rm.Deploy(ctx)
+
+	return nil
 }
 
-func bootstrapComponents(c kubernetes.Interface, namespace string, imageVector imagevector.ImageVector, imageVectorOverwrites map[string]string) ([]component.DeployWaiter, error) {
-	var components []component.DeployWaiter
+// Deletes certain resources from the seed cluster.
```
rfranzke: doc string should start with function name

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+func RunDeleteSeedFlow(ctx context.Context, sc, gc kubernetes.Interface, seed *Seed, seedLogger *logrus.Entry) error {
+	var (
+		namespace = v1beta1constants.GardenNamespace
+		g         = flow.NewGraph("Seed cluster deletion")
```
rfranzke: see above

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
 	}
 
-	// cluster-autoscaler
-	components = append(components, clusterautoscaler.NewBootstrapper(c.Client(), namespace))
+	//setup for flow graph
+	dnsEntry := getManagedIngressDNSEntry(sc, seed.GetIngressFQDN("*"), "", seedLogger)
+	kubernetesVersion, err := semver.NewVersion(sc.Version())
+	if err != nil {
+		return err
+	}
```
rfranzke: can you move this up (before you initiate the components)?

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+	autoscaler := clusterautoscaler.NewBootstrapper(sc.Client(), namespace)
+	seedAdmission := seedadmissioncontroller.New(sc.Client(), namespace, "", kubernetesVersion)
+	rm := resourcemanager.New(sc.Client(), namespace, "", 0, resourcemanager.Values{})
+	etcdDruid := etcd.NewBootstrapper(sc.Client(), namespace, "", kubernetesVersion, nil)
+	kubeScheduler, err := gardenerkubescheduler.Bootstrap(sc.DirectClient(), namespace, nil, kubernetesVersion)
+	if err != nil {
+		return err
+	}
```
rfranzke: See above w.r.t. streamlined variable names.

rfranzke: What about using a `var ()` block?

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
-		if err != nil {
-			return nil, err
-		}
-		etcdImage = image.String()
+
+	f := g.Compile()
+	if err := f.Run(flow.Opts{
+		Logger: seedLogger,
+	}); err != nil {
+		return flow.Errors(err)
```
rfranzke: see above

pkg/operation/seed/seed.go
```diff
@@ -828,151 +828,351 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+		k8sSeedClient,
+		v1beta1constants.GardenNamespace,
+		"nginx-ingress-controller",
+		time.Minute,
+		seedLogger,
+	)
+	if err != nil {
+		return "", err
+	}
+	return loadBalancerAddress, nil
```
```suggestion
	return kutil.WaitUntilLoadBalancerIsReady(
		ctx,
		k8sSeedClient,
		v1beta1constants.GardenNamespace,
		"nginx-ingress-controller",
		time.Minute,
		seedLogger,
	)
```

Generally, I'm wondering if this function is helpful or not (it's not encapsulating too much).

rfranzke (changes requested):

rfranzke:
I have mostly styling suggestions to improve readability. Generally, I wonder why you merged both your PRs into a single one (the changes in the `ControllerRegistration` controller seem unrelated to the changes in the seed controller).

pkg/operation/seed/seed.go
BeckerMax: Sometimes I use functions so I can give an operation a name in order to avoid writing a comment. In this case the function does not add any value - thanks.

BeckerMax:
I agreed with all the suggestions and adapted it. Can you PTAL?
Once you are done I will squash the review commit and create 2 final commits for this PR (as suggested in the last comment). Also a I would like to test it once more when all review adaptions are done.

rfranzke (commented):
@BeckerMax thanks, it looks good now! Please go ahead with your tests, lgtm from my side.

BeckerMax:
@rfranzke I split it in two commits and finished testing. Ready to be merged from my side.

rfranzke (approved):
/lgtm

PR #3576, opened 2021-02-17
Delay the deletion of the external DNS

vpnachev:

/area ops-productivity
/kind enhancement
/priority normal

Delay the deletion of the external DNS record to allow shoot owners to clean-up resources in case they cannot be deleted by Gardener or a Gardener extension.

Fixes #

**Release note**:

```other user
The external DNS record for the kubernetes API server is now deleted after the kubernetes API server. This is useful for shoot cluster owners that need to clean some kubernetes resources that can cause the shoot cluster deletion to stuck.
```


rfranzke (approved):
/lgtm

PR #3582, opened 2021-02-18
Add seed reconciler for secret replication

timuthy:

/area security
/kind enhancement
/priority normal

This PR adds a new seed reconciler which:
- creates a dedicated namespace per seed `seed-<seed-name>`
- replicates secrets with `gardener.cloud/role` label from the `garden` namespace to `seed-<seed-name>` namespaces

In the future, this dedicated namespace will allow us to restrict Gardenlets to only access these namespaces (see #1723).

Fixes parts of #1723

The seed lifecycle controller was refactored along the way and now implements the `reconcile.Reconciler` interface.

/cc @rfranzke

**Release note**:

```feature operator
A new `Seed` reconciler was added to the Gardener-Controller-Manager. It creates a dedicated namespace per seed in the Garden cluster `seed-<seed-name>` and copies common secrets from the `garden` Namespace (labelled with `gardener.cloud/role`) to the seed namespace. Gardenlets are supposed to read secrets (or namespaced objects in general)  from seed dedicated namespaces only in the future.
```


pkg/controllermanager/controller/seed/seed.go
```diff
@@ -111,19 +139,39 @@ func NewSeedController(
 		AddFunc: seedController.seedLifecycleAdd,
 	})
 
+	seedInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: seedController.seedAdd,
```
rfranzke, line 120: Probably we should also react on update, right? I'm imaging a world in which the default domain secrets can have "seed selectors", i.e., they would only apply for seeds having specific labels, so we should reconcile this when the labels change. OTOH, if it's too much effort for now, we can also postpone it to the time when we introduce such seed selectors.

pkg/controllermanager/controller/seed/seed_garden_secret_mapper.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+func filterGardenSecret(obj interface{}) bool {
+	secret, ok := obj.(*corev1.Secret)
+	if !ok {
+		return false
+	}
+	if secret.Namespace != v1beta1constants.GardenNamespace {
+		return false
+	}
```
rfranzke, line 34: Can't we already restrict the shared informer factory to only watch the `garden` namespace?

pkg/controllermanager/controller/seed/seed_reconcile.go
```diff
@@ -0,0 +1,204 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("seed-%s", seedObj.Name),
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(seedObj, gardencorev1beta1.SchemeGroupVersion.WithKind("Seed")),
+			},
+		},
+	}
+
+	// TODO: use controllerutil.CreateOrUpdate once cache is graduated to beta https://github.com/gardener/gardener/issues/2822
+	if err := gardenClient.Client().Create(ctx, namespace); err != nil {
```
rfranzke, line 95: Wouldn't you use the DirectClient() here then?

pkg/controllermanager/controller/seed/seed_reconcile.go
```diff
@@ -0,0 +1,204 @@
...
+	for _, s := range secrets {
+		secret := s
+		if exclude.Has(secret.Name) {
+			continue
+		}
+		fns = append(fns, func(ctx context.Context) error {
+			if err := gardenClient.Client().Delete(ctx, secret); client.IgnoreNotFound(err) != nil {
+				return err
+			}
+			return nil
```
rfranzke:
```suggestion
			return client.IgnoreNotFound(gardenClient.Client().Delete(ctx, secret))
```

pkg/controllermanager/controller/seed/seed_reconcile.go
```diff
@@ -0,0 +1,204 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: secret.Annotations,
+					Labels:      secret.Labels,
+					Name:        secret.Name,
+					Namespace:   namespace.Name,
+				},
+				Data: secret.Data,
+			}
+			// TODO: use controllerutil.CreateOrUpdate once cache is graduated to beta https://github.com/gardener/gardener/issues/2822
+			if err := gardenClient.Client().Create(ctx, seedSecret); err != nil {
```
rfranzke, line 175: same question like above

pkg/controllermanager/controller/seed/seed_reconcile.go
```diff
@@ -0,0 +1,204 @@
...
+		})
+	}
+
+	for _, f := range fns {
+		if err := f(ctx); err != nil {
+			return nil, err
+		}
+	}
+
+	//return secretNames, flow.Parallel(fns...)(ctx)
```
rfranzke: ?

pkg/controllermanager/controller/seed/seed.go
timuthy: > I'm imaging a world in which the default domain secrets can have "seed selectors", i.e., they would only apply for seeds having specific labels

Wouldn't it be the other way around i.e., the seed have a secret selector / secret refs for default domains? I think there won't be seed selectors for secrets, as this is not supported by the API, right? 👀

pkg/controllermanager/controller/seed/seed_garden_secret_mapper.go
timuthy: The [shared informer](https://github.com/gardener/gardener/blob/d8eea89323d880440d9f1034e4b8f2403f72d947/pkg/controllermanager/controller/seed/seed.go#L109) also includes `gardener.cloud/role` secrets from other namespaces as we cache them for [update calls](https://github.com/gardener/gardener/blob/d8eea89323d880440d9f1034e4b8f2403f72d947/pkg/controllermanager/controller/seed/seed_reconcile.go#L169).

pkg/controllermanager/controller/seed/seed.go
rfranzke: > Wouldn't it be the other way around i.e., the seed have a secret selector / secret refs for default domains? 

Hm, the problem with this would be that a `Seed` can "fish" secrets from other seeds (that it's not supposed to see) by simply changing this selector.

> I think there won't be seed selectors for secrets, as this is not supported by the API, right? 👀

Yeah, we would probably have to continue the path that we use today to make the secret carrying additional information about the DNS (e.g., domain, provider, etc.): Labels/Annotations. Such "seed selector" would probably be an annotation on the secret?

pkg/controllermanager/controller/seed/seed_garden_secret_mapper.go
rfranzke: I didn't get it 😢 Can you explain more?

pkg/controllermanager/controller/seed/seed_reconcile.go
timuthy: There is no cache interaction for client write operations. @timebertt are you aware of the opposite?

timebertt: > @timebertt are you aware of the opposite?

Nope. I think so, too.

pkg/controllermanager/controller/seed/seed_garden_secret_mapper.go
timuthy: Sorry, my bad. I meant that we also need to list secrets from other namespaces for cleaning them up, see [here](https://github.com/gardener/gardener/blob/d8eea89323d880440d9f1034e4b8f2403f72d947/pkg/controllermanager/controller/seed/seed_reconcile.go#L120). But you got me thinking: maybe we can use a [metadata-only watch](https://github.com/kubernetes-sigs/controller-runtime/pull/1174) for this so save memory.

timuthy: I tried out the metadata only watch in ce0203cdd49eb6f455084d87f487cf09cf526145 which worked quite well. Nevertheless, I'd like analyse the situation with the help of profiling and monitoring to see if it's really worth. I already pushed it so you can have a look and provide feedback if you want (tests haven't been adjusted yet).

pkg/controllermanager/controller/seed/seed.go
timuthy: To me it sounds like we should rather make `SecretBindings` fit for such use-cases instead of working with selector-annotations. Anyways, like you suggest, let's discuss this topic when we take care about it. We can still adjust the enqueuing logic later.

pkg/controllermanager/controller/seed/seed_garden_secret_mapper.go
timuthy: After some debugging I compiled some advantages/disadvantages:
\+ Saves bandwidth as only the metadata is transferred
\+ Less memory consumption for GCM as the client cache only stores the metadata
\- Additional API server watch for resource --> additional watch cache for resource

Looking at this new reconciler, I'm not sure if the memory and bandwidth are really an issue since the amount of `gardener.cloud/role` secrets in the seed namespaces is rather small. Hence, I'd rather use a secret informer for `gardener.cloud/role` for ALL namespaces instead of haven one secret informer for the `garden` namespace and a metadata informer for other namespaces.
Any opinions?

timuthy: I reverted it back to use the secret informer and as discussed moved it to the GCM factory, so that the ref controller can also make use of the same informer (0124670).

pkg/controllermanager/controller/seed/seed_reconcile.go
```diff
@@ -0,0 +1,204 @@
...
+			}
+			return nil
+		})
+	}
+
+	return flow.Parallel(fns...)(ctx)
+}
+
+func (r *reconciler) syncGardenSecrets(ctx context.Context, gardenClient kubernetes.Interface, namespace *corev1.Namespace) ([]string, error) {
+	secretsGardenRole, err := r.secretLister.Secrets(v1beta1constants.GardenNamespace).List(gardenRoleSelector)
```
timebertt, line 152: Just to verify my understanding: we are preventing to sync the shoot secrets (`kubeconfig`, `ssh-keypair` and so) of shoots in the `garden` namespace by creating a special informer in the `factory.go` [here](https://github.com/gardener/gardener/blob/01246700287e2fdf37c3cea83883351913322ae2/pkg/controllermanager/controller/factory.go#L103-L108), right?
So, if we wouldn't do it there, then the `secretLister` here would also return the shoot secrets, correct?

timebertt (commented):
One question from my side.
Looks good overall :)

pkg/controllermanager/controller/seed/seed_reconcile.go
timuthy: > Just to verify my understanding: we are preventing to sync the shoot secrets (kubeconfig, ssh-keypair and so) of shoots in the garden namespace

Actually, we do this for all namespaces.

> So, if we wouldn't do it there, then the secretLister here would also return the shoot secrets, correct?

Yes, that's correct. Do you think that's too implicit?

timebertt: > Do you think that's too implicit?

Hmm, maybe it would be better to have an explicit selector here.
I'm just a bit afraid, that we might miss the implications of changing the special informer creation and end up with sensitive secrets synced across all seed namespaces...

WDYT about using a label selector for the domain secrets here explicitly? It shouldn't hurt, right?

rfranzke: There are also other secrets that need to be synced (alerting smtp, openvpn dh key, ...)

timebertt: Ok, then what about using the same selector like the special informer (i.e. unselecting all shoot secrets)?

timuthy: Okay, good idea. I've added it to the PR.

timebertt (approved):
Thanks for addressing my suggestion.
Very nice!
/lgtm

PR #3583, opened 2021-02-18
Execute cloud-config if change in cloud config downloader is detected.

### LINKED ISSUE #3580: [OSC] Ensure cloud config downloader scripts are always updated on the node

/area ops-productivity
/kind enhancement
/priority normal

We have observed a Node on GCP was rebooted and then the user data was re-applied. As the user data contains the downloader OSC, the reboot caused the files part of the dowloader to be reverted to their content at the time when the VM was created. The node was created with a gardener version without the features https://github.com/gardener/gardener/pull/3449 and https://github.com/gardener/gardener/pull/3396, but they were applied before the reboot, i.e. the node has the checksum annotation.

After the reboot, the downloader script is the older one that is not aware about the checksum file, i.e. these lines are missing https://github.com/gardener/gardener/blob/b87014c435217716a61f229f9d12fa4f715d0b03/pkg/operation/botanist/extensions/operatingsystemconfig/downloader/templates/scripts/download-cloud-config.tpl.sh#L23-L24
thus the script cannot fetch the latest checksum and apply it on the node. It also properly does not detect difference in the cloud-config script here https://github.com/gardener/gardener/blob/b87014c435217716a61f229f9d12fa4f715d0b03/pkg/operation/botanist/extensions/operatingsystemconfig/executor/templates/scripts/execute-cloud-config.tpl.sh#L87 and does not apply it, this way the downloader script will be outdated till the next update of the osc original.

To avoid this situation, a md/sha checksum of `download-cloud-config.sh` can be maintained in a file on the node via https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/extensions/operatingsystemconfig/executor/templates/scripts/execute-cloud-config.tpl.sh and if change is detected in the sum, then we must run the script, too.

In the usual flow this improvement is obsolete because it is already implemented by the current diff check.
In the case with rebooted node, the checksum will differ because the file was changed by something external to the osc-original, e.g. the agent on the node responsible for the user data, and checksum mismatch will be detected. 

As the checksum file is not part of the userdata, its content will not be affected by the restart.

**Why is this needed**:
To avoid false positive errors on the nodes status like
```
task \"Waiting until all shoot worker nodes have updated the cloud config\
  \ user data\" failed: retry failed with context deadline exceeded, last error: 1\
  \ error occurred:\n\t* the last successfully applied cloud config on node \"shoot--foo--bar-worker-abcd-z1-7d975-g8cfg\"\
  \ is outdated (current: e94238ac0f827f152c3b4baad8c7b9be0345c6714a5bdc6211b7e78793dc1075,\
  \ desired: 4b50559b8e6d39628e3abc6ffcac7704c97b04c9a38220b177a782f2321b547c)\n\n
``

### PR DESCRIPTION
**Pull request by vpnachev**

/area ops-productivity
/kind enhancement
/priority normal

Add additional change detection for the `download-cloud-config.sh` file in case it is changed by external agent, e.g. VM reboot.

Fixes #3580 

**Release note**:

```feature operator
An additional change detection mechanism for the file `download-cloud-config.sh` is now used to ensure the file is up-to-date even after VM reboot. 
```


rfranzke (commented):
Looks great, but I guess you have to adapt the executor unit test?

vpnachev:
> Looks great, but I guess you have to adapt the executor unit test?

Done.

rfranzke (approved):
/lgtm

timuthy (approved):
Nice observation!
/lgtm

PR #3591, opened 2021-02-19
Avoid excessive execution of cloud-config

vpnachev:

/area ops-productivity
/kind enhancement
/priority normal

Update the checksum of the `download-cloud-config.sh` right after it is applied. Otherwise the update will be detected twice and this will restart the systemd services twice. 

Follow-up on #3583 

**Release note**:

```other operator
NONE
```


rfranzke (dismissed):
/lgtm

pkg/operation/botanist/extensions/operatingsystemconfig/executor/templates/scripts/execute-cloud-config.tpl.sh
```diff
@@ -104,7 +104,7 @@ if ! diff "$PATH_CLOUDCONFIG" "$PATH_CLOUDCONFIG_OLD" >/dev/null || ! diff "$PAT
 {{- end }}
     echo "Successfully restarted all units referenced in the cloud config."
     cp "$PATH_CLOUDCONFIG" "$PATH_CLOUDCONFIG_OLD"
-    cp "$PATH_CCD_SCRIPT_CHECKSUM" "$PATH_CCD_SCRIPT_CHECKSUM_OLD"
+    md5sum ${PATH_CCD_SCRIPT} > "$PATH_CCD_SCRIPT_CHECKSUM_OLD"
```
timuthy: Shouldn't it already be enough to move the line up before the daemons are restarted? This change is somehow hard to understand and brittle for maintenance. WDYT?

vpnachev: I prefer to keep the update of the `*_OLD` files together.
Also, after the restarts it is guaranteed that all required changes are applied without failures.

timuthy: Okay, can you then put a small comment inline why copying is not enough like done a line above?

vpnachev: I was thinking the same, but also considered that this file is downloaded every 30 seconds, thus I decided to not add comment, but seems it is required.

timuthy (approved):
Thanks @vpnachev 
/lgtm

rfranzke:
ℹ️ I will open a cherry-pick PR for `release-v1.17` once #3603 is merged.

PR #3596, opened 2021-02-19
Restart CoreDNS on each maintenance

vpnachev:
/area networking
/kind discussion
/priority normal

Restart CoreDNS on each maintenance.
The reason for this change is that we have seen several times shoot cluster with broken networking because all of the dns traffic is going only to one of the all CoreDNS replicas, even though the HPA has scaled it up to the maximum 5 replicas. In these cases just a plain restart of the CoreDNS pods resolved the networking issues in the cluster.

#### Session Affinity
I tried to set `sessionAffinity` https://github.com/gardener/gardener/commit/46afca976623aae4b8784e189da3cbea4438dc01 to the `kube-dns` service and used the command `watch -n 0.1 dig google.com` from a pod running in the cluster to verify whether there is an improvement or not. It got worse :open_mouth:. I found that the client was always stuck to the same backend pod. Reducing the timeout to 5 seconds had no effect.
coredns-session-affinity-with-description

Based on the above, I think the session affinity will affect negatively all clusters in a landscape, so I prefer to not implement it.

#### Restarting the CoreDNS
It is influenced by the control plane restarted. It is optional feature of the maintenance controller disabled by default.
Firstly, I implemented direct coredns deployment patching https://github.com/gardener/gardener/commit/0132b981f32122a2de1de4ced7873849203ae9e2, but It has the disadvantage that the timestamp is not preserved in the managed resource and GRM revert the deployment spec almost immediately. Thus it cannot guarantee that all pods will be restarted.
Then I switched to injecting the timestamp in the chart values https://github.com/gardener/gardener/commit/f24c5a7ee7a1f89bb0bf868d0e12f8629031e059 which is working exactly as it is required. I just not like so much how the task is removed from the shoot.

#### General improvements
- added initial delay to the readiness probe of CoreDNS.
- decreased `maxUnavailable` for the CoreDNS deployment from 1 to 0.

Fixes #

I am opening this PR as draft and to collect some initial feedback. 
I will squash/remove commits before to make it `ready`.

**Release note**:

```feature operator
CoreDNS deployment of shoot clusters can now be automatically restarted during the shoot's maintenance time window. This is used to solve problems with clients stuck to single replica of the deployment and thus overloading it. The feature can be enabled via the `ControllerManagerConfiguration` under `.controllers.shootMaintecance.enableShootCoreAddonRestarter` (see `example/20-componentconfig-gardener-controller-manager.yaml`).
```

### DIFF
--- a/pkg/controllermanager/apis/config/types.go
+++ b/pkg/controllermanager/apis/config/types.go
@@ -180,6 +180,8 @@ type ShootMaintenanceControllerConfiguration struct {
 	ConcurrentSyncs int
 	// EnableShootControlPlaneRestarter configures whether adequate pods of the shoot control plane are restarted during maintenance.
 	EnableShootControlPlaneRestarter *bool
+	// EnableShootCoreAddonRestarter configures whether some core addons to be restarted during maintenance.
+	EnableShootCoreAddonRestarter *bool
 }
 
 // ShootQuotaControllerConfiguration defines the configuration of the
--- a/pkg/operation/botanist/addons.go
+++ b/pkg/operation/botanist/addons.go
@@ -19,11 +19,13 @@ import (
 	"fmt"
 	"path/filepath"
 	"strings"
+	"time"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	"github.com/gardener/gardener/pkg/chartrenderer"
+	"github.com/gardener/gardener/pkg/controllerutils"
 	netpol "github.com/gardener/gardener/pkg/operation/botanist/addons/networkpolicy"
 	"github.com/gardener/gardener/pkg/operation/botanist/component"
 	"github.com/gardener/gardener/pkg/operation/botanist/controlplane/konnectivity"
@@ -32,14 +34,19 @@ import (
 	"github.com/gardener/gardener/pkg/utils/secrets"
 	versionutils "github.com/gardener/gardener/pkg/utils/version"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// SecretLabelKeyManagedResource is a key for a label on a secret with the value 'managed-resource'.
-const SecretLabelKeyManagedResource = "managed-resource"
+const (
+	// SecretLabelKeyManagedResource is a key for a label on a secret with the value 'managed-resource'.
+	SecretLabelKeyManagedResource = "managed-resource"
+	// gardenerRestartedAtKey is annotation key for timestamp used to restart components.
+	gardenerRestartedAtKey = "gardener.cloud/restarted-at"
+)
 
 // GenerateKubernetesDashboardConfig generates the values which are required to render the chart of
 // the kubernetes-dashboard properly.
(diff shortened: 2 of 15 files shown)


pkg/operation/botanist/addons.go
```diff
@@ -504,3 +525,26 @@ func (b *Botanist) generateOptionalAddonsChart(_ context.Context) (*chartrendere
 func (b *Botanist) outOfClusterAPIServerFQDN() string {
 	return fmt.Sprintf("%s.", b.Shoot.ComputeOutOfClusterAPIServerAddress(b.APIServerAddress, true))
 }
+
+// getCoreDNSRestartTimestamp returns a timestamp that can potentially restart the CoreDNS deployment.
+func (b *Botanist) getCoreDNSRestartTimestamp(ctx context.Context) (string, error) {
```
vpnachev, line 530: I have the idea to restart only when the deployment is running on the maximum allowed replicas. Any opinions?

vlerenc:
Thanks @vpnachev. I hope as well that it will help.

timuthy:
@vpnachev I assume your results are based on `iptables`. Do you think we have the same issues with `IPVS` or need to handle it differently? Maybe it works better with RR as the [default](https://kubernetes.io/blog/2018/07/09/ipvs-based-in-cluster-load-balancing-deep-dive/#parameter-changes) and we don't need a restart at all?

vpnachev:
> Do you think we have the same issues with IPVS or need to handle it differently?

To be honest, the cluster that was recently affected by this issue was using Cilium as network extension which can be also a factor. My observation was for cluster with IPTables+Calico and will spend some time to evaluate Cilium and IPVS. 

>  Maybe it works better with RR as the default and we don't need a restart at all?

I do not understand this, could you elaborate it?

timuthy:
> I do not understand this, could you elaborate it?

I was wondering if it was an `iptables` only problem and if `IPVS` worked perfectly with regards to load balancing, then we'd probably not need the restart mechanism in case IPVS is enabled.

vpnachev:
I have executed the same test with IPVS and cilium shoot clusters - the result is absolutely the same. Therefore, I am proposing to restart the CoreDNS just like we are doing with control plane pods in the maintenance time window.

pkg/operation/botanist/addons.go
```diff
@@ -32,14 +34,19 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// SecretLabelKeyManagedResource is a key for a label on a secret with the value 'managed-resource'.
-const SecretLabelKeyManagedResource = "managed-resource"
+const (
+	// SecretLabelKeyManagedResource is a key for a label on a secret with the value 'managed-resource'.
+	SecretLabelKeyManagedResource = "managed-resource"
+	// gardenerRestartAtKey is annotation key for timestamp used to restart components.
+	gardenerRestartAtKey = "gardener.cloud/restartedAt"
```
```suggestion
	gardenerRestartedAtKey = "gardener.cloud/restarted-at"
```

pkg/operation/botanist/addons.go
```diff
@@ -311,6 +318,20 @@ func (b *Botanist) generateCoreAddonsChart(ctx context.Context) (*chartrenderer.
 	}
 	shootInfo["extensions"] = strings.Join(extensions, ",")
 
+	coreDNSRestartTimestamp, err := b.getCoreDNSRestartTimestamp(ctx)
+	if err != nil {
+		return nil, err
```
rfranzke, line 323: Should `not found` errors be ignored here?

timuthy (commented):
Thanks for the analysis @vpnachev.

pkg/controllermanager/apis/config/types.go
```diff
@@ -180,6 +180,8 @@ type ShootMaintenanceControllerConfiguration struct {
 	ConcurrentSyncs int
 	// EnableShootControlPlaneRestarter configures whether adequate pods of the shoot control plane are restarted during maintenance.
 	EnableShootControlPlaneRestarter *bool
+	// EnableShootCoreDNSRestarter configures whether CoreDNS deployment to be  restarted during maintenance.
+	EnableShootCoreDNSRestarter *bool
```
timuthy: How about naming this `EnableShootCoreAddonRestarter` which leaves more room for future use-cases?

rfranzke (approved):
/lgtm

timuthy (approved):

PR #3607, opened 2021-02-22
[release-v1.17] Remove finalizers from the worker secret in the migration flow

rfranzke:
Cherry-pick of #3560.

**Release note**:

```noteworthy dependency
In the generic worker actuator's migration flow, the MCM finalizer of the secret that is referenced by the `Worker` is not removed. We have now added functionality that checks that secret and removes only the MCM finalizers if necessary. 
```


ialidzhikov (approved):
/lgtm

PR #3584, opened 2021-02-18
Do not try to remove the finalizer of Plant Secret when it is not found

ialidzhikov:
/kind bug

Currently gardener-controller-manager is not able to delete a Plant when its Secret is not found:

```
time="2021-02-18T06:13:04Z" level=info msg="Error syncing plant garden-it/e2e-9n26q: failed to remove finalizer from plant secret 'garden-it/test-secret-plant-r9xtc': could not remove \"core.gardener.cloud/plant\" finalizer: resource name may not be empty"
time="2021-02-18T06:13:04Z" level=info msg="Error syncing plant garden-it/e2e-9n26q: failed to remove finalizer from plant secret 'garden-it/test-secret-plant-r9xtc': could not remove \"core.gardener.cloud/plant\" finalizer: resource name may not be empty"
time="2021-02-18T06:13:04Z" level=info msg="Error syncing plant garden-it/e2e-9n26q: failed to remove finalizer from plant secret 'garden-it/test-secret-plant-r9xtc': could not remove \"core.gardener.cloud/plant\" finalizer: resource name may not be empty"
time="2021-02-18T06:13:04Z" level=info msg="Error syncing plant garden-it/e2e-9n26q: failed to remove finalizer from plant secret 'garden-it/test-secret-plant-r9xtc': could not remove \"core.gardener.cloud/plant\" finalizer: resource name may not be empty"
time="2021-02-18T06:13:04Z" level=info msg="Error syncing plant garden-it/e2e-9n26q: failed to remove finalizer from plant secret 'garden-it/test-secret-plant-r9xtc': could not remove \"core.gardener.cloud/plant\" finalizer: resource name may not be empty"
time="2021-02-18T06:13:04Z" level=info msg="Error syncing plant garden-it/e2e-9n26q: failed to remove finalizer from plant secret 'garden-it/test-secret-plant-r9xtc': could not remove \"core.gardener.cloud/plant\" finalizer: resource name may not be empty"
```

**Release note**:

```bugfix operator
An issue causing gardener-controller-manager to not be able to delete a Plant when the Plant Secret is not found is now fixed.
```

PR #3615, opened 2021-02-23
[release-v1.17] Do not try to remove the finalizer of Plant Secret when it is not found

rfranzke:
Cherry-pick of #3584.

**Release note**:

```other operator
An issue causing gardener-controller-manager to not be able to delete a Plant when the Plant Secret is not found is now fixed.
```


ialidzhikov (approved):
/lgtm

PR #3617, opened 2021-02-23
Ensure ManagedSeed VPA is not enabled if shoot VPA is enabled

stoyanr:

/area control-plane
/kind bug
/priority normal

Ensures ManagedSeed VPA is not enabled if shoot VPA is enabled.

Fixes #

I chose to fix the issue in the admission plugin, rather than in the (deprecated) seed registration controller (handling the `use-as-seed` annotation).
* If the seed VPA is not explicitly enabled or disabled - set it to the opposite of the shoot VPA setting
* If the seed VPA is explicitly enabled and the shoot VPA is also enabled - error
* If the seed VPA is explicitly disabled and the shoot VPA is also disabled - ok, the user might want to use their own VPA

In addition, the safety check in the managed seed controller is also fixed to properly take into account the fact that the seed VPA is enabled by default, similarly to all other seed settings.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/plugin/pkg/managedseed/validator/admission.go
+++ b/plugin/pkg/managedseed/validator/admission.go
@@ -332,6 +332,19 @@ func (v *ManagedSeed) admitSeedSpec(spec *gardencore.SeedSpec, shoot *gardencore
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("provider", "region"), spec.Provider.Region, fmt.Sprintf("seed provider region must be equal to shoot region %s", shoot.Spec.Region)))
 	}
 
+	// Initialize and validate settings
+	shootVPAEnabled := gardencorehelper.ShootWantsVerticalPodAutoscaler(shoot)
+	if spec.Settings == nil || spec.Settings.VerticalPodAutoscaler == nil {
+		if spec.Settings == nil {
+			spec.Settings = &gardencore.SeedSettings{}
+		}
+		spec.Settings.VerticalPodAutoscaler = &gardencore.SeedSettingVerticalPodAutoscaler{
+			Enabled: !shootVPAEnabled,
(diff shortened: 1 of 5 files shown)


plugin/pkg/managedseed/validator/admission.go
```diff
@@ -332,6 +332,19 @@ func (v *ManagedSeed) admitSeedSpec(spec *gardencore.SeedSpec, shoot *gardencore
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("provider", "region"), spec.Provider.Region, fmt.Sprintf("seed provider region must be equal to shoot region %s", shoot.Spec.Region)))
 	}
 
+	// Initialize and validate settings
+	shootVPAEnabled := gardencorehelper.ShootWantsVerticalPodAutoscaler(shoot)
+	if spec.Settings == nil || spec.Settings.VerticalPodAutoscaler == nil {
+		if spec.Settings == nil {
+			spec.Settings = &gardencore.SeedSettings{}
+		}
```
timebertt, line 340: Can you move it up before the first `if`?

timebertt (commented):
One nit, rest lgtm

plugin/pkg/managedseed/validator/admission.go
stoyanr: I probably could, but should I? I find the code as it's written easier to understand - we only change something in spec if we are sure we want to apply a default, not otherwise.
I am anxious to get this reviewed and merged to fix the broken dev - could we focus on the significant parts please?

ialidzhikov (approved):
/lgtm
/needs second-opinion

timebertt (approved):
/lgtm

PR #3618, opened 2021-02-23
Scale down Loki only once per week

Kristian-ZH:

/area logging
/kind enhancement
/priority normal

1. Configure HVPA to scale down Loki only once per week to reduce the frequent restarts.

2. Change the ScaleDown Maintenance time window to be 1 hour after the shoot's one. This will prevent Loki to be scaled down and to lose logs during the shoot update.

Fixes #

@dguendisch 

**Release note**:

```other operator
None
```

### DIFF
--- a/pkg/operation/seed/seed.go
+++ b/pkg/operation/seed/seed.go
@@ -401,8 +401,17 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
 					return err
 				}
 			} else {
-				maintenanceBegin = shootInfo.Data["maintenanceBegin"]
-				maintenanceEnd = shootInfo.Data["maintenanceEnd"]
+				shootMaintenanceBegin, err := utils.ParseMaintenanceTime(shootInfo.Data["maintenanceBegin"])
+				if err != nil {
+					return err
+				}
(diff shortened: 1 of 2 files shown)


dguendisch:
/lgtm
thanks!

pkg/operation/seed/seed.go
```diff
@@ -401,8 +401,21 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+					return err
+				}
+				maintenanceBegin = beginTime.Add(1 * time.Hour).Format(maintenanceTimeLayout)
+
+				shootMaintenanceEnd := shootInfo.Data["maintenanceEnd"]
+				endTime, err := time.Parse(maintenanceTimeLayout, shootMaintenanceEnd)
+				if err != nil {
+					return err
+				}
+				maintenanceEnd = endTime.Add(1 * time.Hour).Format(maintenanceTimeLayout)
```
rfranzke: Use https://github.com/gardener/gardener/blob/b79f180e762a79bee67f13fc1cac66ab5f0212e2/pkg/utils/timewindow.go#L47-L55?

rfranzke (approved):
/lgtm

PR #3628, opened 2021-02-25
Update GEP-13, esp. ManagedSeed and ManagedSeedSet sections

stoyanr:

/area documentation
/kind enhancement
/priority normal

Updates GEP-13:
* Update the "ManagedSeed Resources" section to reflect the way ManagedSeeds were eventually implemented.
* Update the "ManagedSeedSet" section to elaborate on new ideas especially for managing ManagedSeed updates.

Fixes #

**Release note**:

```other operator
NONE
```


vlerenc:
@stoyanr What was the motivation to use one resource type that can now do two different things?

Referring to:
> Either the seedTemplate or the gardenlet section must be specified, but not both

I am curious, because someone’s/my naive thoughts when reading this were: why then not two different resource types / resources in practice where e.g. the gardenlet would be another resource (request for installation) and the controller would install it once the seed (as referred to by it) becomes available?

docs/proposals/13-automated-seed-management.md
```diff
@@ -263,18 +283,66 @@ ManagedSeedSets are reconciled by a new *managed seed set controller* in `garden
...
+
+* Ordered, graceful deployment and scaling.
+* Ordered, automated rolling updates. Such rolling updates can be *partitioned* (limited to replicas with ordinal numbers greater than or equal to the "partition") to achieve *phased rollouts*.
+
+A ManagedSeedSet, unlike a StatefulSet, does not need to maintain a stable identity for its ManagedSeeds. Furthermore, it would not be practical to always remove the replicas with the largest ordinal numbers when scaling in, since the corresponding seeds may have shoots scheduled onto them, while other seeds, with lower ordinals, may have fewer shoots (or none), and therefore be much better candidates for being removed.
+
+On the other hand, it would be beneficial if a ManagedSeedSet, like a StatefulSet, provides ordered deployment and scaling, ordered rolling updates, and phased rollouts. The main advantage of these features is that a deployment or update failure would affect fewer replicas (ideally just one), containing any potential damage and making the situation easier to handle, thus achieving some of the goals stated in [Issue #87](https://github.com/gardener/gardener/issues/87).
```
danielfoehrKn: I can also imagine that a phased rollout would help to contain Seed rolling updates outside business hours to avoid Seed being rolled when most stakeholders are wanting to use their clusters.

docs/proposals/13-automated-seed-management.md
```diff
@@ -263,18 +283,66 @@ ManagedSeedSets are reconciled by a new *managed seed set controller* in `garden
...
+
+Although such ordinal numbers cannot provide stable identity, they can provide a predictable ordering for deployments and updates, and can also be used to partition rolling updates similarly to StatefulSet ordinal numbers.
+
+#### Update Strategies
+
+The ManagedSeedSet's `.spec.updateStrategy` field allows configuring automated rolling updates for the ManagedSeeds and Shoots in a ManagedSeedSet.
+
+**Rolling Updates**
+
+The `RollingUpdate` update strategy implements automated, rolling update for the ManagedSeeds and Shoots in a ManagedSeedSet. With this strategy, the ManagedSeedSet controller will update each ManagedSeed and Shoot in the ManagedSeedSet. It will proceed from the largest number to the smallest, updating each ManagedSeed and its corresponding Shoot one at a time. It will wait until both the Shoot and the Seed of an updated ManagedSeed are Ready prior to updating its predecessor.
```
danielfoehrKn, line 324: What about the Shoots that are deployed on those ManagedSeeds? 
In a bad case, a ManagedSeed. (the Shoot and Seed) are READY, but the control plane pods have troubles. E.g there was an update to the kubelet configuration for a certain worker pool of the Seed. cluster.

The rollout would then only continue if not more than X percent of the Shoot clusters on the Seed have Reconcile or Readiness (only control plane) problems. 

WDYT

stoyanr:
> @stoyanr What was the motivation to use one resource type that can now do two different things?

Because a single resource type is more suitable, easier to handle, and easier to use. Also I don't see this as "two different things" - in both cases, it's still a "managed seed", that is a cluster (currently a shoot, in the future perhaps also something else) registered as a seed. There is some shared configuration (e.g. the name of the shoot that should be registered) and a lot (I mean, really a lot) of shared logic between both cases. Having two resource types would make both the implementation and the user experience more complex and yield no benefits, at least I can't think of some.

The `gardenlet` case is much more than a "request for installation". Especially if you consider the deletion ("uninstallation") scenario, besides just undeploying gardenlet it goes through additional phases to ensure that the Seed registered by gardenlet and its secrets have been properly deleted (gardenlet itself doesn't do anything of this), making it even more similar to the `seedTemplate` case. 

Also, consider the case of a ManagedSeedSet. Should we have then two different Set types, or different types of templates in the same Set type? Both approaches would further increase the complexity compared to what's currently proposed, and again yield no benefits.

docs/proposals/13-automated-seed-management.md
stoyanr: OK, I will mention this as well, thanks.

docs/proposals/13-automated-seed-management.md
stoyanr: OK, makes sense in general, not sure how complex that would be to implement. Perhaps before implementing the ManagedSeedSet controller, we should consider introducing a "managed seed care" controller that would update the ManagedSeed resource itself with the health of its Seed and Shoots deployed on top, since keeping track of all these might be too much for the ManagedSeedSet controller to handle.

danielfoehrKn: Ok, I like that idea. Also makes the initial implementation easier

docs/proposals/13-automated-seed-management.md
```diff
@@ -263,18 +283,66 @@ ManagedSeedSets are reconciled by a new *managed seed set controller* in `garden
...
+
+On the other hand, it would be beneficial if a ManagedSeedSet, like a StatefulSet, provides ordered deployment and scaling, ordered rolling updates, and phased rollouts. The main advantage of these features is that a deployment or update failure would affect fewer replicas (ideally just one), containing any potential damage and making the situation easier to handle, thus achieving some of the goals stated in [Issue #87](https://github.com/gardener/gardener/issues/87).
+
+Based on the above considerations, we propose the following mechanism for handling ManagedSeed identity and order:
+
+* A ManagedSeedSet uses *ordinal numbers generated by an increasing sequence* to identify ManagedSeeds and Shoots it creates and manages. These numbers always starts from 0 and are incremented by 1 for each newly added replica.
```
rfranzke:
```suggestion
* A ManagedSeedSet uses *ordinal numbers generated by an increasing sequence* to identify ManagedSeeds and Shoots it creates and manages. These numbers always start from 0 and are incremented by 1 for each newly added replica. 
```

docs/proposals/13-automated-seed-management.md
```diff
@@ -263,18 +283,66 @@ ManagedSeedSets are reconciled by a new *managed seed set controller* in `garden
...
 
-There is an important difference between seeds and pods or nodes in that seeds are more "heavyweight" and therefore updating a set of seeds by introducing new seeds and moving shoots to them tends to be much more complex, time-consuming, and prone to failures compared to updating the seeds "in place". Furthermore, updating seeds in this way depends on a mature implementation of [GEP-7: Shoot Control Plane Migration](07-shoot-control-plane-migration.md), which is not available right now. Therefore, ManagedSeedDeployments are not described in detail here, as properly dealing with the challenges mentioned requires a separate GEP.
+* During an update, determine which replicas are still not on the latest revision and therefore should be updated.
+* Display the revision history of a ManagedSedSet via `kubectl rollout history`.
+* Roll back all ManagedSedSet replicas to a specific revision via `kubectl rollout undo`
```
rfranzke:
```suggestion
* Roll back all ManagedSeedSet replicas to a specific revision via `kubectl rollout undo`
```

docs/proposals/13-automated-seed-management.md
```diff
@@ -263,18 +283,66 @@ ManagedSeedSets are reconciled by a new *managed seed set controller* in `garden
...
 
-There is an important difference between seeds and pods or nodes in that seeds are more "heavyweight" and therefore updating a set of seeds by introducing new seeds and moving shoots to them tends to be much more complex, time-consuming, and prone to failures compared to updating the seeds "in place". Furthermore, updating seeds in this way depends on a mature implementation of [GEP-7: Shoot Control Plane Migration](07-shoot-control-plane-migration.md), which is not available right now. Therefore, ManagedSeedDeployments are not described in detail here, as properly dealing with the challenges mentioned requires a separate GEP.
+* During an update, determine which replicas are still not on the latest revision and therefore should be updated.
+* Display the revision history of a ManagedSedSet via `kubectl rollout history`.
```
rfranzke:
```suggestion
* Display the revision history of a ManagedSeedSet via `kubectl rollout history`.
```

rfranzke (commented):
I'm very much looking forward to having this in place, thanks for updating the document!

docs/proposals/13-automated-seed-management.md
stoyanr: Hm, I don't see any difference?

docs/proposals/13-automated-seed-management.md
stoyanr: OK

docs/proposals/13-automated-seed-management.md
stoyanr: OK

docs/proposals/13-automated-seed-management.md
rfranzke: Sorry, fixed...

docs/proposals/13-automated-seed-management.md
```diff
@@ -224,17 +240,21 @@ metadata:
   name: crazy-botany
   namespace: garden
 spec:
-  replicas: 1
+  replicas: 3
   selector:
     matchLabels:
       foo: bar
+  updateStrategy:
```
stoyanr, line 247: Questions to reviewers:
* Since only a single update strategy (`RollingUpdate`) is planned, do we need the entire field as specified here? I tend to prefer it this way, since it's more familiar and open for extensions. Any other opinions?
* Do we need a field similar to `podManagementPolicy` in `StatefulSet`? Or to put it in a different way, do you think "parallel" management policy could also be useful in some cases in addition to the "orderedReady" policy described in the document?

docs/proposals/13-automated-seed-management.md
stoyanr: OK

docs/proposals/13-automated-seed-management.md
stoyanr: OK, I will mention this in the document.

docs/proposals/13-automated-seed-management.md
```diff
@@ -263,18 +283,66 @@ ManagedSeedSets are reconciled by a new *managed seed set controller* in `garden
...
+
+Based on the above considerations, we propose the following mechanism for handling ManagedSeed identity and order:
+
+* A ManagedSeedSet uses *ordinal numbers generated by an increasing sequence* to identify ManagedSeeds and Shoots it creates and manages. These numbers always starts from 0 and are incremented by 1 for each newly added replica. 
+* Replicas (both ManagedSeeds and Shoots) are named after the ManagedSeedSet with the ordinal number appended. For example, for a ManagedSeedSet named `test` its replicas are named `test-0`, `test-1`, etc.
+* Gaps in the sequence created by removing replicas with ordinal numbers in the middle of the range are never filled in. A newly added replica always receives a number that is not only free, but also unique to itself. For example, if there are 2 replicas named `test-0` and `test-1` and any one of them is removed, a newly added replica will still be named `test-2`.   
+
+Although such ordinal numbers cannot provide stable identity, they can provide a predictable ordering for deployments and updates, and can also be used to partition rolling updates similarly to StatefulSet ordinal numbers.
```
timebertt: > Although such ordinal numbers cannot provide stable identity

Hmmm, why do you think so? The ManagedSeeds' identities are stable, aren't they?
Basically, every Seed (Shoot/ManagedSeed-tuple) has a stable identity and it is never changed. We go even further, as we say, identities that were once used should not be reused.

docs/proposals/13-automated-seed-management.md
timebertt: Have you checked, if `kubectl rollout undo` will directly work with custom resources like this?
Or is there something required additionally to make this work?

docs/proposals/13-automated-seed-management.md
timebertt: > * Since only a single update strategy (`RollingUpdate`) is planned, do we need the entire field as specified here? I tend to prefer it this way, since it's more familiar and open for extensions. Any other opinions?

+1, I would also keep it this way.

> * Do we need a field similar to `podManagementPolicy` in `StatefulSet`? Or to put it in a different way, do you think "parallel" management policy could also be useful in some cases in addition to the "orderedReady" policy described in the document?

I don't see any necessity for adding a different management policy as of now, as I also don't see any use case currently.

timebertt (commented):
Thank you! The update is well-described und quite clear from my perspective.

I don't have any additional concerns / thoughts since the last time we talked about it.
So I'm curious about opinions from other reviewers.
/lgtm

docs/proposals/13-automated-seed-management.md
stoyanr: Yes, you are right. I will rephrase this, what should be said is that "although such ordinal numbers can also provide stable identity, in this case it is much more important that they can provide a predictable ordering ...".

docs/proposals/13-automated-seed-management.md
stoyanr: No, I don't think it will work directly. [Documentation](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#rollout) says explicitly that valid resource types are deployments, daemonsets, and statefulsets, and I don't see how they could implement it to support custom types out of the box. But I think it should be possible to support rollout commands for our types via a [kubectl plugin](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/). I will update the document to properly mention this.

stoyanr:
@danielfoehrKn @rfranzke @timebertt Updated the document to reflect your feedback, in a new commit.
/squash

timebertt (approved):
Looks good, thank you!
Approving now, but asking @rfranzke @danielfoehrKn for
/second-opinion
/lgtm

rfranzke (approved):
/lgtm

danielfoehrKn (approved):

PR #3633, opened 2021-02-26
Promote `ManagedIstio` and `APIServerSNI` feature gates

### LINKED ISSUE #2942: Graduate APIServerSNI (+ ManagedIstio) feature gate

rfranzke:

/area open-source
/kind enhancement
/priority normal

The `ManagedIstio` and `APIServerSNI` feature gates in the `gardenlet` have been promoted to beta.
Along the way, I updated some feature gate owners.

Fixes #2942

/squash

**Release note**:

```action operator
The `ManagedIstio` and `APIServerSNI` feature gates in the `gardenlet` have been promoted to beta and are now enabled by default. If you run your own istio installation then you have to disable the `ManagedIstio` feature gate (and probably also the `APIServerSNI`) in your gardenlet configurations.
```


pkg/gardenlet/features/features.go
```diff
@@ -28,9 +28,9 @@ var (
 		features.Logging:                {Default: false, PreRelease: featuregate.Alpha},
 		features.HVPA:                   {Default: false, PreRelease: featuregate.Alpha},
 		features.HVPAForShootedSeed:     {Default: false, PreRelease: featuregate.Alpha},
-		features.ManagedIstio:           {Default: false, PreRelease: featuregate.Alpha},
+		features.ManagedIstio:           {Default: false, PreRelease: featuregate.Beta},
 		features.KonnectivityTunnel:     {Default: false, PreRelease: featuregate.Alpha},
-		features.APIServerSNI:           {Default: false, PreRelease: featuregate.Alpha},
+		features.APIServerSNI:           {Default: false, PreRelease: featuregate.Beta},
```
rfranzke: I did not enable them by default yet. Should we do it?

vpnachev: Yes.

vpnachev: Also, enable them by default here https://github.com/gardener/gardener/blob/5eefe2db2e64db1d7947c3bdf605ac7e7879346b/example/20-componentconfig-gardenlet.yaml#L88-L90

vpnachev (changes requested):
Please, update also https://github.com/gardener/gardener/blob/master/docs/deployment/feature_gates.md#feature-gates-for-alpha-or-beta-features.

mvladev:
Hold on - end-users might already have Istio installed on their clusters, enabling it by default might  cause problems for them.

vpnachev:
What is the plan then? I am thinking about deprecating the `ManagedIstio` and moving it to an extension, wdyt?

rfranzke:
I agree with @mvladev. My gud feeling is that it's not a good idea to enable it by default yet. 

> What is the plan then? I am thinking about deprecating the ManagedIstio and moving it to an extension, wdyt?

😕 Why? At least at this point, what is the benefit? I vote for being pragmatic now: Promote the feature gates to beta to indicate that we consider it as "mature enough to be used", but do not enable it by default to not interfere with existing installations.

mvladev:
> I agree with @mvladev. My gud feeling is that it's not a good idea to enable it by default yet.
> 
> > What is the plan then? I am thinking about deprecating the ManagedIstio and moving it to an extension, wdyt?
> 
> 😕 Why? At least at this point, what is the benefit? I vote for being pragmatic now: Promote the feature gates to beta to indicate that we consider it as "mature enough to be used", but do not enable it by default to not interfere with existing installations.

I would agree with @vpnachev . I would rather have the isito installation of of the gardener. Now that I think about it - probably would do the same for the `GardenerSeedScheduler` feature gate as well. Or we keep it beta, but do not enable it.

rfranzke:
Yes, generally, I agree as well, but I don't see how this is related now and why we would need to hold the beta promotion now. Actually, our (or at least mine) goal with this PR is only to indicate that the feature is mature enough and "battle-tested" so that others can feel comfortable enough to try it out as well. IMO this does not rule out the opportunity to move everything out of g/g in the future when we have the time and necessity to do so?
OTOH, I am also totally fine with keeping everything as it is (no beta promotion) because we anyways never followed this process, so it probably doesn't matter too much to consumers of Gardener (I assume, many are using "alpha" gates like `Logging`, `HVPA`, and so on for a long time already).
What do you prefer?

vpnachev:
I would prefer just to try to establish some policy like k8s. Alpha feature are disabled by default and needs to be explicitly enabled. Beta and GA features are enabled by default and needs to be explicitly disabled.

Thinking again about https://github.com/gardener/gardener/pull/3633#issuecomment-787428033, I actually do not see it as a blocker. We just need to add an `action required` release note and ask the users that already have Istio in their seed clusters to explicitly disable the `ManagedIstio` feature gate.

rfranzke:
OK, I've now enabled them by default and added a corresponding release note. Please check.

timebertt (dismissed):
/lgtm
I think, this is the way to go.

If we want to allow users to disable the gardener managed istio installation after graduating the feature gate to GA, we can still add a configuration option, which is then more fitting for having optional components than a feature gate in my opinion...

vpnachev (dismissed):
/lgtm

We could enable them by default here https://github.com/gardener/gardener/blob/ef495eb662556cdb7dff4401eaa32ddc539e604b/example/20-componentconfig-gardenlet.yaml#L88-L90 but it can also be done with some other PR.

rfranzke:
@vpnachev done

timebertt (dismissed):
/lgtm

vpnachev (dismissed):
Thanks!
/lgtm

rfranzke:
The `release-v1.18` branch is already created, so this PR won't make it into the v1.18 release. I've pushed 31656b1 which changes the docs to `v1.19`.

vpnachev (approved):
/lgtm

timebertt (approved):
/lgtm

PR #3638, opened 2021-03-01
[GEP-13] Add ManagedSeedSet APIs

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Adds the new `ManagedSeedSet` resource as described in [GEP-13](docs/proposals/13-automated-seed-management.md#managedseeds), with the updates proposed in #3628 .

* `ManagedSeedSet` resource APIs, including defaults, since they are used by the generated code.
* All generated code belonging to the above APIs, including generated clients.

The new APIs are not yet registered or used by the API server. This is on purpose, to make this PR easier to review and make sure it can be merged as it has no impact yet.

Fixes #

This PR makes the bold assumptions that most of the proposals from #3628 will be accepted. Consequently, it adopts most (but not all) fields from `StatefulSet`, including the `UpdateStrategy` spec field and status fields that are only meaningful for managing updates using the "ordered ready" strategy and `ControllerRevision` objects for keeping revision history.

**Release note**:

```other operator
NONE
```


timebertt:
I will wait with the review until https://github.com/gardener/gardener/pull/3628 has more feedback / is merged.

pkg/apis/seedmanagement/types_managedseedset.go
```diff
@@ -0,0 +1,128 @@
...
+	// NextReplicaNumber is the ordinal number that will be assigned to the next replica of the ManagedSeedSet.
+	NextReplicaNumber int32
+	// CurrentReplicas is the number of ManagedSeeds created by the ManagedSeedSet controller from the ManagedSeedSet version
+	// indicated by CurrentRevision.
+	CurrentReplicas int32
+	// UpdatedReplicas is the number of ManagedSeeds created by the ManagedSeedSet controller from the ManagedSeedSet version
+	// indicated by UpdateRevision.
+	UpdatedReplicas int32
+	// CurrentRevision, if not empty, indicates the version of the ManagedSeedSet used to generate ManagedSeeds with smaller
+	// ordinal numbers during updates.
```
danielfoehrKn, line 115: "with smaller ordinal numbers during updates" --> smaller than what?

The field description in the stateful set status is similar so I think I do not have the full picture how this field is going to be used. 
Do you mind explaining it? Thanks

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
```diff
@@ -0,0 +1,143 @@
...
+	// Selector is a label query over ManagedSeeds and Shoots that should match the replica count.
+	// It must match the ManagedSeeds and Shoots template's labels.
+	// +optional
+	Selector *metav1.LabelSelector `json:"selector" protobuf:"bytes,2,opt,name=selector"`
+	// Template describes the ManagedSeed that will be created if insufficient replicas are detected.
+	// Each ManagedSeed created / updated by the ManagedSeedSet will fulfill this template.
+	Template ManagedSeedTemplate `json:"template" protobuf:"bytes,3,opt,name=template"`
+	// ShootTemplate describes the Shoot that will be created if insufficient replicas are detected for hosting the corresponding ManagedSeed.
+	// Each Shoot created / updated by the ManagedSeedSet will fulfill this template.
+	ShootTemplate gardencorev1beta1.ShootTemplate `json:"shootTemplate,omitempty" protobuf:"bytes,4,rep,name=shootTemplate"`
```
rfranzke:
```suggestion
	ShootTemplate gardencorev1beta1.ShootTemplate `json:"shootTemplate" protobuf:"bytes,4,rep,name=shootTemplate"`
```

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
```diff
@@ -0,0 +1,143 @@
...
+
+// ManagedSeedSetSpec is the specification of a ManagedSeedSet.
+type ManagedSeedSetSpec struct {
+	// Replicas is the desired number of replicas of the given Template. Defaults to 1.
+	// +optional
+	Replicas *int32 `json:"replicas,omitempty" protobuf:"varint,1,opt,name=replicas"`
+	// Selector is a label query over ManagedSeeds and Shoots that should match the replica count.
+	// It must match the ManagedSeeds and Shoots template's labels.
+	// +optional
+	Selector *metav1.LabelSelector `json:"selector" protobuf:"bytes,2,opt,name=selector"`
```
rfranzke:
```suggestion
	Selector *metav1.LabelSelector `json:"selector,omitempty" protobuf:"bytes,2,opt,name=selector"`
```

pkg/apis/seedmanagement/types_managedseedset.go
```diff
@@ -0,0 +1,128 @@
...
+type ManagedSeedSetStatus struct {
+	// ObservedGeneration is the most recent generation observed for this ManagedSeedSet. It corresponds to the
+	// ManagedSeedSet's generation, which is updated on mutation by the API Server.
+	ObservedGeneration int64
+	// Replicas is the number of replicas (ManagedSeeds and their corresponding Shoots) created by the ManagedSeedSet controller.
+	Replicas int32
+	// ReadyReplicas is the number of ManagedSeeds created by the ManagedSeedSet controller that have a Ready Condition.
+	ReadyReplicas int32
+	// NextReplicaNumber is the ordinal number that will be assigned to the next replica of the ManagedSeedSet.
+	NextReplicaNumber int32
```
danielfoehrKn, line 107: I assume that is just a helper field as it is at least not in the stateful set status?

pkg/apis/seedmanagement/types_managedseedset.go
stoyanr: The description in the StatefulSet doesn't not fit unfortunately. The StatefulSet always has replicas with ordinals from 0 to `replicas` - 1, so it's possible to specify the 2 ranges precisely. In our case, the ordinal numbers are ever increasing, so it's not possible to specify these ranges precisely. There will be `CurrentReplicas` replicas on `CurrentRevision` (with smaller ordinal numbers), `UpdatedReplicas` on `UpdatedRevision` (with larger ordinal numbers), and a "cutoff" number (or range) somewhere in the middle separating both ranges, but it's impossible to precisely tell the cutoff value (the controller will calculate it each time after enumerating all replicas). I chose to write a shorter comment rather than explaining all this. If you have an idea how to put it more concisely, please share.

pkg/apis/seedmanagement/types_managedseedset.go
stoyanr: This is needed for the "ever increasing sequence" of replica ordinal numbers to work. We need to store somewhere the next replica ordinal number (starting with 0), and the status of the resource is the most obvious choice. When a new replica is created, it will get this number as ordinal, and the value in the status will be incremented by 1. The StatefulSet doesn't need such a field since its ordinals are calculated differently.
Come think of it, because of this difference, there might be other fields needed in the status, but it's rather hard to foresee all of them now. I will add them when the need for them becomes more clear.

timebertt:
/assign 
Plan on reviewing this afternoon.

pkg/apis/seedmanagement/types_managedseedset.go
danielfoehrKn: okay thanks for the explanation.

pkg/apis/seedmanagement/types_managedseedset.go
danielfoehrKn: thanks. No, it is fine with me how it is currently.

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
stoyanr: OK

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
stoyanr: OK

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
```diff
@@ -0,0 +1,143 @@
...
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ManagedSeedSetList is a list of ManagedSeed objects.
+type ManagedSeedSetList struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard list object metadata.
+	// +optional
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+	// Items is the list of ManagedSeeds.
```
timebertt:
```suggestion
	// Items is the list of ManagedSeedSets.
```

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
```diff
@@ -0,0 +1,143 @@
...
+	// +optional
+	Spec ManagedSeedSetSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+	// Status is the current status of ManagedSeeds and Shoots in this ManagedSeedSet.
+	// +optional
+	Status ManagedSeedSetStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ManagedSeedSetList is a list of ManagedSeed objects.
```
timebertt, line 42:
```suggestion
// ManagedSeedSetList is a list of ManagedSeedSet objects.
```

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
```diff
@@ -0,0 +1,143 @@
...
+	// be maintained in the ManagedSeedSet's revision history. Defaults to 10.
+	// +optional
+	RevisionHistoryLimit *int32 `json:"revisionHistoryLimit,omitempty" protobuf:"varint,6,opt,name=revisionHistoryLimit"`
+}
+
+// ManagedSeedSetUpdateStrategy specifies the strategy that the ManagedSeedSet
+// controller will use to perform updates. It includes any additional parameters
+// necessary to perform the update for the indicated strategy.
+type ManagedSeedSetUpdateStrategy struct {
+	// Type indicates the type of the ManagedSeedSetUpdateStrategy. Defaults to ManagedSeedSetUpdateStrategyType.
```
timebertt:
```suggestion
	// Type indicates the type of the ManagedSeedSetUpdateStrategy. Defaults to RollingUpdate.
```

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
```diff
@@ -0,0 +1,143 @@
...
+	// +optional
+	Type *ManagedSeedSetUpdateStrategyType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type,casttype=ManagedSeedSetUpdateStrategyType"`
+	// RollingUpdate is used to communicate parameters when Type is ManagedSeedSetUpdateStrategyType.
```
timebertt:
```suggestion
	// RollingUpdate is used to communicate parameters when Type is RollingUpdate.
```

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
```diff
@@ -0,0 +1,143 @@
...
+	// RevisionHistoryLimit is the maximum number of revisions that will
+	// be maintained in the ManagedSeedSet's revision history. Defaults to 10.
+	// +optional
+	RevisionHistoryLimit *int32 `json:"revisionHistoryLimit,omitempty" protobuf:"varint,6,opt,name=revisionHistoryLimit"`
+}
+
+// ManagedSeedSetUpdateStrategy specifies the strategy that the ManagedSeedSet
+// controller will use to perform updates. It includes any additional parameters
+// necessary to perform the update for the indicated strategy.
+type ManagedSeedSetUpdateStrategy struct {
```
timebertt: Regarding the naming of all the types and consts related to the update strategy:
I think it would be fine to remove `ManagedSeedSet` from their names.
For example: `RollingUpdateManagedSeedSetUpdateStrategyType` is quite lengthy/bumpy and will be harder comprehend in code. We could just name it `RollingUpdateStrategyType` instead, as I aspect there won't be any other `Set` type/controller in this API group anytime soon and it's rather clear that the update strategy types/consts belong to the `ManagedSeedSet`.
WDYT?

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
```diff
@@ -0,0 +1,143 @@
...
+	// +optional
+	CollisionCount *int32 `json:"collisionCount,omitempty" protobuf:"varint,9,opt,name=collisionCount"`
+	// Conditions represents the latest available observations of a ManagedSeedSet's current state.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +optional
+	Conditions []gardencorev1beta1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,10,rep,name=conditions"`
+}
+
+// TODO Condition constants
```
timebertt, line 142: Do you have any condition types in mind already?

pkg/apis/seedmanagement/v1alpha1/defaults_managedseedset.go
```diff
@@ -0,0 +1,61 @@
...
+func SetDefaults_ManagedSeedSetUpdateStrategy(obj *ManagedSeedSetUpdateStrategy) {
+	// Set default type
+	if obj.Type == nil {
+		t := RollingUpdateManagedSeedSetUpdateStrategyType
+		obj.Type = &t
+	}
+
+	if *obj.Type == RollingUpdateManagedSeedSetUpdateStrategyType {
+		// Set rolling update defaults
+		if obj.RollingUpdate == nil {
```
timebertt: I don't think we need to default this. If it's not specified, we can just keep it empty.
I think, this is also what the `StatefulSet` does.

timebertt (changes requested):
Very nice work, @stoyanr!
I like that we follow a very similar approach to the `StatefulSet`, which makes it easier to comprehend and we can also take some advantage of the maturity of the API design and controller implementation.

I only found a few nits in docstrings and a few small suggestions.

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
stoyanr: OK

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
stoyanr: OK

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
stoyanr: OK

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
stoyanr: OK

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
stoyanr: Yes, I agree, will shorten the names.

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
stoyanr: No, I haven't really thought much about that ... I put this TODO as a placeholder since I suspect there will be some conditions to be added later. OK?

pkg/apis/seedmanagement/v1alpha1/defaults_managedseedset.go
stoyanr: The defaults of a `StatefulSet` are a mess :-) I deviated from them in a few other places as well. But OK, I agree it's better not to default this particular field, will remove it.

stoyanr:
@rfranzke @timebertt Added a new commit to address your feedback. Also note that I made the `selector` field mandatory, this would save me some clumsy extra validation in the next PR. It's not optional (although a pointer) also in the `StatefulSet`.

timebertt (approved):
Very nice, looks all good to me!
/lgtm

Thanks for adding a separate commit with your iterative changes. This made the re-review very efficient.
/squash

rfranzke (approved):

PR #3640, opened 2021-03-01
Move mocks into folders in package directories

rfranzke:

/area dev-productivity
/kind cleanup
/priority normal

Instead of having mocks in `./pkg/mock/...` representing the directory structure we are now generating the mocks in dedicated folders in the package directories themselves (i.e., if package `foo` is in `./pkg/path/to/foo` then the mock would be in `pkg/path/to/foo/mock` instead of `./pkg/mock/gardener/path/to/foo`.
Hence, the `./pkg/mock/gardener` folder is eliminated completely. Only the mocks for third-party/vendored packages remain in `./pkg/mock` now.

/cc @BeckerMax 

**Release note**:

```action dependency
The mocks for Gardener packages were moved to dedicated folders in the respective package directories, i.e., if there is package `foo` in `./pkg/path/to/foo` then the mock would be in `pkg/path/to/foo/mock` instead of `./pkg/mock/gardener/path/to/foo`. Only the mocks for third-party/vendored packages remain in `./pkg/mock`.
```


timebertt (approved):
Thanks for cleaning up.
/lgtm

timebertt:
https://github.com/gardener/gardener/pull/3641 will conflict with this.
Can we merge this PR, so that I can rebase https://github.com/gardener/gardener/pull/3641?

rfranzke:
I think we can, it's only a little restructuring of the mocks. So I'll merge it, please rebase your PR :) Thanks!

PR #3641, opened 2021-03-01
Use patches for working on finalizers

timebertt:

/area dev-productivity scalability
/kind cleanup technical-debt
/priority normal

This is the first PR of a few PRs working towards getting rid of the `DirectClient` in our clientset (see https://github.com/gardener/gardener/issues/2822#issuecomment-702132185 for the detailed reasoning).
In order to do that, it adds an `APIReader` to our clientset (similar to controller-runtime concepts) and deprecates `DirectClient` in favor of it.

This PR eliminates the first category of `DirectClient` usages: working on finalizers. We now use patches for that without retry-on-conflict semantics in most cases (see below for details). This way we avoid extra reads from the API server.

Part of #2822

We had a lot of `DirectClient` usages for working on finalizers, because we saw a lot of conflicts with a cached client in this area.
In general, controllers should just return such errors encountered during reconciliation and let the object be requeued with exponential backoff instead of handling retries on their own.
Though, not all of our controllers can tolerate stale cache reads (leading to conflict errors) because they have long-running reconciliation loops, where a requeue is undesirable.
In extension controllers for example at the end of a deletion run, we should retry on conflict, because we don't know what the actuator is doing and if it would be able to shortcut another deletion request or just lead to unnecessary Cloudprovider API calls contributing to rate limits and so on.

**Release note**:

```breaking dependency
- Semantics of `controllerutils.{EnsureFinalizer,RemoveFinalizer}` were changed. Both funcs now use `PATCH` requests instead of `UPDATE` and `RemoveFinalizer` expects an additional `client.Reader` for reading from the API server.
- `extensioncontroller.{EnsureFinalizer,DeleteFinalizer}` have been removed in favor of the funcs in `controllerutils`.
- `controllerutils.PatchFinalizers` was renamed to `PatchAddFinalizers`.
```
```breaking developer
Semantics of `controllerutils.{EnsureFinalizer,RemoveFinalizer}` were changed. Both funcs now use `PATCH` requests instead of `UPDATE` and `RemoveFinalizer` expects an additional `client.Reader` for reading from the API server.
Please use `controllerutils.{PatchFinalizers,PatchRemoveFinalizers}` preferably were applicable, if your controller is able to tolerate conflict errors tolerated by stale reads.
```


timebertt:
Rebased to resolve conflicts.

pkg/controllerutils/finalizers.go
```diff
@@ -16,82 +16,57 @@ package controllerutils
...
 // PatchFinalizers adds the given finalizers to the object via a patch request.
-func PatchFinalizers(ctx context.Context, c client.Client, obj client.Object, finalizers ...string) error {
-	beforePatch := obj.DeepCopyObject()
-
-	for _, finalizer := range finalizers {
-		controllerutil.AddFinalizer(obj, finalizer)
-	}
-
-	return c.Patch(ctx, obj, client.MergeFromWithOptions(beforePatch, client.MergeFromWithOptimisticLock{}))
+func PatchFinalizers(ctx context.Context, writer client.Writer, obj client.Object, finalizers ...string) error {
```
rfranzke:
```suggestion
func PatchAddFinalizers(ctx context.Context, writer client.Writer, obj client.Object, finalizers ...string) error {
```

extensions/pkg/controller/backupbucket/reconciler.go
```diff
@@ -156,12 +163,12 @@ func (r *reconciler) delete(ctx context.Context, bb *extensionsv1alpha1.BackupBu
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("failed to get backup bucket secret: %+v", err)
 	}
-	if err := extensionscontroller.DeleteFinalizer(ctx, r.client, secret, FinalizerName); err != nil {
+	if err := controllerutils.RemoveFinalizer(ctx, r.reader, r.client, secret, FinalizerName); err != nil {
```
rfranzke, line 166: Why do you use the `RemoveFinalizer` function here which will retry internally while you use the `PatchFinalizers` function during `reconciliation` which won't retry?

timebertt:
Pushed two more commits to fix failing tests and add the missing inject funcs, I missed earlier.

extensions/pkg/controller/backupbucket/reconciler.go
timebertt: This is the case I tried to describe with:

> In extension controllers for example at the end of a deletion run, we should retry on conflict, because we don't know what the actuator is doing and if it would be able to shortcut another deletion request or just lead to unnecessary Cloudprovider API calls contributing to rate limits and so on.

Basically, we don't know what the actuator is doing and an entire retry of a potentially long-running reconciliation just because of the finalizer removal seems undesirable to me.

At the beginning of the reconciliation loop where the finalizer is added, the controller/actuator can definitely handle a requeue/retry, as it's more or less the first action.

Do you have any concerns with this approach or suggest to do it differently?

pkg/controllerutils/finalizers.go
timebertt: Done, PTAL.

extensions/pkg/controller/backupbucket/reconciler.go
rfranzke: OK, I get it now, but I wonder whether we could simply then also use the "retry" method at the beginning of the reconciliation. What you describe is reasonable, but I think the current code disguises these considerations and it's non-obvious if you are not into these details.

timebertt: Would some additional comments here and there also help?
Otherwise, I can also just use the retry-on-conflict variant for adding finalizers.

rfranzke: Probably it would, but I think it would end up unnecessarily verbose, so just using the other function would make the code easier read- and understandable (at least for me).

timebertt: Ok, done. PTAL :)

rfranzke (approved):
Thanks!

/lgtm

PR #3642, opened 2021-03-02
clarify how worker group kubelet configuration behaves

BeckerMax:

/area quality
/priority normal

Clarify how worker group kubelet setting behaves with toplevel kubelet settings.

I looked for other places in the documentation I should adapt but found only this place.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/hack/api-reference/core.md
+++ b/hack/api-reference/core.md
@@ -9266,7 +9266,8 @@ KubeletConfig
 </td>
 <td>
 <em>(Optional)</em>
-<p>Kubelet contains configuration settings for all kubelets of this worker pool.</p>
+<p>Kubelet contains configuration settings for all kubelets of this worker pool.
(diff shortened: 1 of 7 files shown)


example/90-shoot.yaml
```diff
@@ -67,7 +67,7 @@ spec:
     #   effect: NoSchedule
     # caBundle: <some-ca-bundle-to-be-installed-to-all-nodes-in-this-pool>
     # kubernetes:
-    #   kubelet:
+    #   kubelet: # if configured all `spec.kubernetes.kubelet` settings will be overwritten for this worker group (no merge of settings).
```
ialidzhikov: I believe the most appropriate place for such details is the doc string of the corresponding field -> it will then part of the API reference. We don't explain each and every field in the example files.

rfranzke: Probably both would be fine.

ialidzhikov:
@BeckerMax , you need to run `make generate`.

danielfoehrKn (approved):

PR #3643, opened 2021-03-02
[GEP-13] Add ManagedSeedSet validation

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Adds validation for `ManagedSeedSet` resources.

* `ManagedSeedSet` validation code
* Changes to `Shoot`, `Seed`, and `ManagedSeed` validation code to relax validation checks for their spec when they are specified in a template. This is needed to avoid having to specify fields in `ManagedSeedSet`  templates that are otherwise optional and are defaulted when the corresponding resources are actually created.

The new validation code is not testable (except by unit tests), since the `ManagedSeedSet` resource is not yet registered in the API server. This is on purpose, to make this PR easier to review.

Fixes #

Based on #3638, marked as "Draft" until that one is merged.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/apis/core/validation/shoot.go
+++ b/pkg/apis/core/validation/shoot.go
@@ -93,6 +93,26 @@ func ValidateShootUpdate(newShoot, oldShoot *core.Shoot) field.ErrorList {
 	return allErrs
 }
 
+// ValidateShootTemplate validates a ShootTemplate.
+func ValidateShootTemplate(shootTemplate *core.ShootTemplate, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	allErrs = append(allErrs, metav1validation.ValidateLabels(shootTemplate.Labels, fldPath.Child("metadata", "labels"))...)
+	allErrs = append(allErrs, apivalidation.ValidateAnnotations(shootTemplate.Annotations, fldPath.Child("metadata", "annotations"))...)
+	allErrs = append(allErrs, ValidateShootSpec(shootTemplate.ObjectMeta, &shootTemplate.Spec, fldPath.Child("spec"))...)
+
+	return allErrs
+}
+
+// ValidateShootTemplateUpdate validates a ShootTemplate before an update.
+func ValidateShootTemplateUpdate(newShootTemplate, oldShootTemplate *core.ShootTemplate, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	allErrs = append(allErrs, ValidateShootSpecUpdate(&newShootTemplate.Spec, &oldShootTemplate.Spec, newShootTemplate.ObjectMeta, fldPath.Child("spec"))...)
+
+	return allErrs
+}
+
 // ValidateShootSpec validates the specification of a Shoot object.
 func ValidateShootSpec(meta metav1.ObjectMeta, spec *core.ShootSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
--- a/pkg/apis/core/validation/shoot.go
+++ b/pkg/apis/core/validation/shoot.go
@@ -837,23 +837,14 @@ func validateMaintenance(maintenance *core.Maintenance, fldPath *field.Path) fie
 	allErrs := field.ErrorList{}
 
 	if maintenance == nil {
-		allErrs = append(allErrs, field.Required(fldPath, "maintenance information is required"))
 		return allErrs
 	}
 
-	if maintenance.AutoUpdate == nil {
-		allErrs = append(allErrs, field.Required(fldPath.Child("autoUpdate"), "auto update information is required"))
-	}
-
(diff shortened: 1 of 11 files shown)


pkg/apis/core/validation/shoot.go
```diff
@@ -93,6 +93,26 @@ func ValidateShootUpdate(newShoot, oldShoot *core.Shoot) field.ErrorList {
...
+func ValidateShootTemplateUpdate(newShootTemplate, oldShootTemplate *core.ShootTemplate, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	allErrs = append(allErrs, ValidateShootSpecUpdate(&newShootTemplate.Spec, &oldShootTemplate.Spec, newShootTemplate.ObjectMeta, fldPath.Child("spec"))...)
+
+	return allErrs
+}
+
 // ValidateShootSpec validates the specification of a Shoot object.
 func ValidateShootSpec(meta metav1.ObjectMeta, spec *core.ShootSpec, fldPath *field.Path) field.ErrorList {
```
rfranzke, line 117: Unlike other functions similar to this one (`ValidateSeedSpec`, `ValidateManagedSeedSpec`), this function does not take an `inTemplate bool` parameter. Instead, I see that some validation code was entirely removed. Shouldn't it be similar for all the three cases?

rfranzke:
```
> Check
Executing golangci-lint
WARN [runner] Can't run linter goanalysis_metalinter: S1020: failed prerequisites: [(inspect@github.com/gardener/gardener/pkg/gardenlet/apis/config/validation_test [github.com/gardener/gardener/pkg/gardenlet/apis/config/validation.test], isgenerated@github.com/gardener/gardener/pkg/gardenlet/apis/config/validation_test [github.com/gardener/gardener/pkg/gardenlet/apis/config/validation.test]): analysis skipped: errors in package: [/go/src/github.com/gardener/gardener/pkg/gardenlet/apis/config/validation/validation_test.go:182:51: too few arguments in call to ValidateGardenletConfiguration /go/src/github.com/gardener/gardener/pkg/gardenlet/apis/config/validation/validation_test.go:199:51: too few arguments in call to ValidateGardenletConfiguration /go/src/github.com/gardener/gardener/pkg/gardenlet/apis/config/validation/validation_test.go:205:51: too few arguments in call to ValidateGardenletConfiguration]] 
ERRO Running error: S1020: failed prerequisites: [(inspect@github.com/gardener/gardener/pkg/gardenlet/apis/config/validation_test [github.com/gardener/gardener/pkg/gardenlet/apis/config/validation.test], isgenerated@github.com/gardener/gardener/pkg/gardenlet/apis/config/validation_test [github.com/gardener/gardener/pkg/gardenlet/apis/config/validation.test]): analysis skipped: errors in package: [/go/src/github.com/gardener/gardener/pkg/gardenlet/apis/config/validation/validation_test.go:182:51: too few arguments in call to ValidateGardenletConfiguration /go/src/github.com/gardener/gardener/pkg/gardenlet/apis/config/validation/validation_test.go:199:51: too few arguments in call to ValidateGardenletConfiguration /go/src/github.com/gardener/gardener/pkg/gardenlet/apis/config/validation/validation_test.go:205:51: too few arguments in call to ValidateGardenletConfiguration]] 
make: *** [Makefile:170: check] Error 3
```

pkg/apis/core/validation/shoot.go
stoyanr: This is exactly the discussion we had with @timebertt . Initially, it had the `inTemplate` parameter. Tim however argued, and I agreed with him, that the `inTemplate` parameter should be avoided if possible since it makes the validation code harder to maintain. To achieve this, we should distinguish clearly between the following two cases:

* Validation that is in contradiction with the `+optional` on a field and this is not justified by any special condition. In such cases, there is always defaulting code (or an admission plugin) to set the field to a valid value, however it only kicks in if an actual resource is created. Such validation should simply be removed. It should be removed carefully however to allow empty non-mandatory fields, but still reject non-empty fields if other conditions are violated. 
* Validation that really must be different between "in-template" and otherwise. We looked hard for ways to avoid the `inTemplate` parameter also for such cases but couldn't find a better approach.

So for `Shoot`, it turned out all validation code that failed the template validation fell in the first category, so an `inTemplate` parameter was not needed there. For `Seed` and `ManagedSeed` however we had cases in the second category, so the `inTemplate` parameter was kept.

rfranzke: I agree that this parameter makes the whole validation code hard to maintain and that it should be avoided if possible. Great that you already discussed it and I think the outcome makes sense. Thanks for clarifying.

rfranzke (approved):
/lgtm

@danielfoehrKn do you plan to take a look here? If yes, please /assign yourself until EOB so that we get to know. Otherwise, as @stoyanr and @timebertt already discussed this PR out-of-band and @timebertt is out of office this week, I would go ahead with merging then.

PR #3577, opened 2021-02-17
Refactor gardener-admission-controller

timebertt:
**How to categorize this PR?**

/area quality dev-productivity
/kind cleanup
/priority normal

**What this PR does / why we need it**:

This PR refactors/enhances GAC similar to GSAC in https://github.com/gardener/gardener/pull/3459.

- migrate the component to a full controller-runtime component (use webhook handlers/decoders, metrics, health endpoints and so on) instead of custom coding for those mechanisms
- enable both admission review versions (including `admission/v1`) in the webhook configs, as it is now supported by the c-r webhook handlers
- decrease the timeouts of the extension deletion protection webhook to `10s` according to best pratice

**Which issue(s) this PR fixes**:
Part of #3109

**Special notes for your reviewer**:
/squash

**Release note**:

```breaking operator
The `gardener-admission-controller` configuration API and http endpoints were changed in several aspects:
- the fields `server.https.tls.server{Cert,Key}Path` have been removed in favor of `server.https.tls.serverCertDir` (the cert directory is expected to contain a `tls.crt` and `tls.key` file)
- metrics and health endpoints are now exposed as plain HTTP endpoints on dedicated ports (configurable via `server.{healthProbes,metrics}.port`
- the `gardener-admission-controller` service included in Gardener's helm chart has a new named port (`metrics`) for exposing the metrics endpoint
If you deploy this component/configuration manually, please adapt your usage accordingly. Gardener's helm charts were adapted to the changes.
```
```other operator
`gardener-admission-controller`'s webhooks now also accept reviews in version `admission/v1`. Also, webhook timeouts have been lowered to `10s`.
```
```feature operator
`gardener-admission-controller` now exposes several metrics about its webhooks (e.g. `controller_runtime_webhook_latency_seconds_bucket`, `controller_runtime_webhook_requests_in_flight` and `controller_runtime_webhook_requests_total`)
The metric `gardener_admission_controller_invalid_webhook_requests_total` was removed in favor of the newly added metrics.
```

PR #3644, opened 2021-03-02
Dedicated packages for webhook handlers in GAC

rfranzke:

/area dev-productivity
/kind cleanup
/priority normal

For better isolation we should have dedicated package for the webhook handlers in the GAC (gardener-admission-controller).

Follow-up after #3577 and #3571

I'll work on a similar PR for the GSAC (gardener-seed-admission-controller).
PR has several commits, so
/squash

**Release note**:

```noteworthy operator
NONE
```


pkg/admissioncontroller/webhooks/admission/namespacedeletion/admission.go
```diff
@@ -36,109 +36,107 @@ import (
...
 		cacheReader: cache,
+		logger:      logger,
 	}, nil
 }
 
-// InjectLogger injects a logger into the handler.
-func (h *namespaceDeletionHandler) InjectLogger(l logr.Logger) error {
-	h.logger = l.WithName(namespaceValidatorName)
-	return nil
+type plugin struct {
```
timebertt: Naming nit: `plugin` might be confusing, as this is not an admission plugin. Maybe `handler` fits better?

pkg/admissioncontroller/webhooks/admission/namespacedeletion/admission.go
```diff
@@ -36,109 +36,107 @@ import (
...
 // Handle implements the webhook handler for namespace deletion validation.
-func (h *namespaceDeletionHandler) Handle(ctx context.Context, request admission.Request) admission.Response {
+func (p *plugin) Handle(ctx context.Context, request admission.Request) admission.Response {
 	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
 	defer cancel()
 
 	// If the request does not indicate the correct operation (DELETE) we allow the review without further doing.
 	if request.Operation != admissionv1.Delete {
-		return admissionAllowed("operation is not DELETE")
+		return admission.Allowed("operation is not DELETE")
```
timebertt: FYI: I chose to have a custom `admissionAllowed` func instead of using the upstream funcs, as they put the string into the `Reason` field, which is not 100% correct if we use such description like this one. According to the [API docs](https://github.com/gardener/gardener/blob/03794c4fab012f16625540cd2989ab8f328be93d/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L644-L649), `Reason` is supposed to be a machine-readable description, while `Message` is the human-readable counterpart.
The webhook plugins and kubectl also build on that and transport error messages / reason accordingly, which results in weird error outputs for denied webhook responses.

timebertt (dismissed):
/lgtm

pkg/admissioncontroller/webhooks/admission/namespacedeletion/admission.go
rfranzke: Hm, OK, I saw this as well but was wondering why the controller-runtime doesn't support this then? Shall we switch back to our "custom" function? We don't use this in the GSAC, however, GSAC runs in the seed while GAC is in the garden cluster and its responses could be visible to end-users.

timebertt: Yeah, I was asking myself the same question. But I haven't found the time to ask/discuss/contribute upstream yet.
I would probably keep the custom funcs for now in GAC because of the enduser visibility, however I'm not insisting on it.

rfranzke: OK, makes sense, will do.

pkg/admissioncontroller/webhooks/admission/namespacedeletion/admission.go
rfranzke: Yes, makes sense, done in d80c0c5

pkg/admissioncontroller/webhooks/admission/namespacedeletion/admission.go
rfranzke: Done in af6adcd

timebertt (dismissed):
Thank you!
/lgtm

timebertt (approved):
/lgtm

PR #3645, opened 2021-03-02
Add error code for retryable configuration problems

### LINKED ISSUE #3020: Error code for misconfigured `PodDisruptionBudget`s blocking machine drain processes

/area ops-productivity
/kind enhancement
/priority normal

We can reuse the `ERR_CONFIGURATION_PROBLEM` error code to classify shoot errors in which a misconfigured `PodDisruptionBudget` (min available=1), e.g. for istio, blocks the machine drain process.
This is to hide it from the erroneous shoot list in the Gardener Dashboard, and to quickly categorize what the problem is.

Similar to how the shoot care controller detects problematic webhooks, it could detect problematic pod disruption budgets, and write a new constraint into the `Shoot`'s `.status.constraints[]` list.

**Why is this needed**:
In some cases, such misconfigured `PodDisruptionBudget`s created by end-users, e.g. for istio, which leads to a long-visible error message:

  ```text
  task "Waiting until shoot worker nodes have been reconciled" failed: Error while waiting for Worker shoot--foo--bar/bar to become ready: extension encountered error during reconciliation: Error reconciling worker: Failed while waiting for all machine deployments to be ready: 'retry failed with context deadline exceeded, last error: Waiting until all old machines are drained and terminated. Waiting for 2 machine(s)  ...'
  ```

ialidzhikov:
/area ops-productivity
/kind enhancement
/priority normal

Fix #3020

**Release note**:

```other operator
A new error code for retryable configuration problems (for example misconfigured PodDisruptoinBudget that does not allow voluntary Pod evictions) is now added.
```

### DIFF
--- a/docs/usage/shoot_status.md
+++ b/docs/usage/shoot_status.md
@@ -55,14 +55,12 @@ Currently there are two constraints:
 
 This constraint indicates whether a Shoot is allowed to be hibernated.
 The rationale behind this constraint is that a Shoot can have `ValidatingWebhookConfiguration`s or `MutatingWebhookConfiguration`s acting on resources that are critical for waking up a cluster.
-For example, if a webhook has rules for `CREATE/UPDATE` Pods or Nodes and `failurePolicy=Fail`, the webhook will block joining `Nodes` and creating critical system component Pods and thus block the entire wakeup
-operation, because the server backing the webhook is not running.
+For example, if a webhook has rules for `CREATE/UPDATE` Pods or Nodes and `failurePolicy=Fail`, the webhook will block joining `Nodes` and creating critical system component Pods and thus block the entire wakeup operation, because the server backing the webhook is not running.
 
 Even if the `failurePolicy` is set to `Ignore`, high timeouts (`>15s`) can lead to blocking requests of control plane components.
 That's because most control-plane API calls are made with a client-side timeout of `30s`, so if a webhook has `timeoutSeconds=30`
 the overall request might still fail as there is overhead in communication with the API server and potential other webhooks.
-Generally, it's [best pratice](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts)
-to specify low timeouts in WebhookConfigs.
+Generally, it's [best pratice](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts) to specify low timeouts in WebhookConfigs.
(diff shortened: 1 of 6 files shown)


rfranzke (approved):
/lgtm

FYI @gardener/dashboard-maintainers

petersutter:
Thanks for the notification. Do you have a suggestion which help text that we could provide to users in order to lower ops effort.
e.g. for `ERR_CONFIGURATION_PROBLEM ` we have these texts:
```js
shortDescription: 'Configuration Problem',
description: 'There is a configuration problem that is most likely caused by your Shoot specification. Please double-check the error message and resolve the problem.'
```
https://github.com/gardener/dashboard/blob/d5a8cba24e88ef800da520a8fbcc21093bef6645/frontend/src/utils/errorCodes.js#L69-L73

ialidzhikov:
For now I would suggest
```
shortDescription: 'Retryable Configuration Problem',
description: 'There is a configuration problem. Please double-check the error message and resolve the problem.'
```

It is too generic but on the other side (retryable) configuration problem can be anything. Also suggesting that something might be wrong with the Shoot spec is not valid (for now) for this error code. But in any case for the PDB case (the only case for now behind this new error code) the Shoot lastError will be good enough to display what is the configuration problem.

petersutter:
@ialidzhikov thanks for the suggestion, however, what does `Retryable Configuration Problem` mean, especially what do you mean with `Retryable`? What can the user retry in this case?

ialidzhikov:
> @ialidzhikov thanks for the suggestion, however, what does `Retryable Configuration Problem` mean, especially what do you mean with `Retryable`? What can the user retry in this case?

The current `ERR_CONFIGURATION_PROBLEM` is interpreted as non-retryable error code. Which means that if such error occurs, gardenlet will mark the Shoot as failed and won't retry the last operation. I introduced `ERR_RETRYABLE_CONFIGURATION_PROBLEM` as we now have a configuration problem error that is likely to be resolved in a ungraceful manner after given period of time. And we don't want to mark the corresponding Shoots as failed when such error occurs (and is resolved in ungraceful manner after 2h for example).
So this error code is not for end user to indicate them that they need to retry something, it is rather for gardenlet that it can retry the last operation with backoff. From end user perspective, they should known that they have a configuration issue that they need to resolve. So probably let's use the same shortDescription for both error codes to prevent any misunderstanding that someone needs to retry something manually:

```
shortDescription: 'Configuration Problem',
```

danielfoehrKn (approved):
/lgtm

PR #3646, opened 2021-03-02
Dedicated packages for webhook handlers in GSAC

rfranzke:

/area dev-productivity
/kind cleanup
/priority normal

Similar to #3644, for better isolation we should have dedicated package for the webhook handlers in the GSAC (gardener-seed-admission-controller).

Follow-up after #3459

PR has several commits, so
/squash

**Release note**:

```noteworthy operator
NONE
```


pkg/seedadmission/webhooks/admission/extensioncrds/admission_test.go
```diff
@@ -575,170 +576,21 @@ var _ = Describe("Extension Deletion Protection", func() {
 			})
 		})
 	})
-
-	Describe("Integration Test", func() {
```
timebertt: Why did you move this to the `suite_test.go` file?

timebertt (dismissed):
/lgtm

pkg/seedadmission/webhooks/admission/extensioncrds/admission_test.go
rfranzke: Hm, I would need the `restConfig` of the testEnv here in this files and I don't have access to it. Do you suggest to start dedicated test envs in the new admission packages instead of the "central" test env in `./pkg/seedadmissioncontroller`?

timebertt: Ah, of course. Makes perfect sense. I didn't notice the package move in the diff...
No, I would not start dedicated test envs for this.

What do you think about having separate files (like `extension_crds_test.go` and `podschedulername_test.go`) containing the different integration tests in the `seedadmissioncontroller` package then instead of having both tests in the `suite_test.go`?

rfranzke: Sounds like a good idea, will do!

rfranzke: Done in 295ba93 and df9953f.

I've also renamed `plugin` to `handler` in 11354c0.

timebertt (approved):
Nice, thank you!
/lgtm

PR #3648, opened 2021-03-02
Remove DirectClient usage in integration tests

timebertt:

/kind cleanup
/priority normal

This PR removes all `DirectClient` usages under `test/`. The clients constructed there are anyways uncached and read directly from the API server.

Part of https://github.com/gardener/gardener/issues/2822

No other change, just a good old "find+replace all".

**Release note**:

```other operator
NONE
```

### DIFF
--- a/test/framework/applications/guestbooktest.go
+++ b/test/framework/applications/guestbooktest.go
@@ -231,7 +231,7 @@ func (t *GuestBookTest) Cleanup(ctx context.Context) {
 	// Clean up shoot
 	ginkgo.By("Cleaning up guestbook app resources")
 	deleteResource := func(ctx context.Context, resource client.Object) error {
-		err := t.framework.ShootClient.DirectClient().Delete(ctx, resource)
+		err := t.framework.ShootClient.Client().Delete(ctx, resource)
 		if apierrors.IsNotFound(err) {
 			return nil
 		}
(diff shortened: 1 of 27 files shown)


dguendisch (approved):
/lgtm

rfranzke (approved):
/lgtm

I was wondering what was the initial rationale behind using `DirectClient()` in the integration tests (instead of `Client()`)? Will this have an impact on potential test flakes due to maybe missing retry logic?

schrodit:
> I was wondering what was the initial rationale behind using DirectClient() in the integration tests (instead of Client())?

When the new clients were introduced, the cached clients were not initialized in the tests as far as I remember. 
@timebertt suggest to only use the direct client when he introduced it.

timebertt:
> > I was wondering what was the initial rationale behind using DirectClient() in the integration tests (instead of Client())?
> 
> When the new clients were introduced, the cached clients were not initialized in the tests as far as I remember.
> @timebertt suggest to only use the direct client when he introduced it.

Hmm, I think I switched to the `DirectClient` for integration tests before I introduced the `kubernetes.UseCachedRuntimeClients` switch (off by default). Now that it's there, `Client()` and `DirectClient()` will anyway return the same client instance.



> Will this have an impact on potential test flakes due to maybe missing retry logic?

Therefore, nope. This is basically no difference.

schrodit (approved):

PR #3651, opened 2021-03-03
Vendor k/*@v0.20 and controller-runtime@v0.8

rfranzke:

/area open-source quality
/kind task
/priority normal

This PR vendors all the Kubernetes dependencies in version `v0.20.2` and `controller-runtime` version `v0.8.3`.

/invite @timebertt @ialidzhikov 
/squash

I got rid of our fork of `k8s.io/kube-openapi` ~~and our pinning of `sigs.k8s.io/structured-merge-diff/v4`~~ - do you see issues with this?

**Release note**:

```noteworthy dependency
⚠️ Go dependencies to `kubernetes/*` and `kubernetes-sigs/controller-runtime` were updated to `v0.20.2` and `v0.8.3` respectively.
```


pkg/apis/seedmanagement/v1alpha1/zz_generated.conversion.go
```diff
@@ -286,8 +296,7 @@ func autoConvert_v1alpha1_ManagedSeedSpec_To_seedmanagement_ManagedSeedSpec(in *
 	if in.SeedTemplate != nil {
 		in, out := &in.SeedTemplate, &out.SeedTemplate
 		*out = new(core.SeedTemplate)
-		// TODO: Inefficient conversion - can we improve it?
-		if err := s.Convert(*in, *out, 0); err != nil {
```
rfranzke, line 512: Due to this, I had to introduce https://github.com/gardener/gardener/pull/3651/files#diff-c7b64578d2ac95623dfd44edeb4a58700b755fa9bda6b8267965ec156363a6e6R76-R83. I'm not sure why the generation does not find the functions here: https://github.com/gardener/gardener/blob/6f6cf8a41b46e01bac9761c01d0f8bc0e22e8c61/pkg/apis/core/v1beta1/zz_generated.conversion.go#L4366-L4369

I played with `--extra-peer-dirs` but without success. Any pointers are appreciated.

/cc @stoyanr @ialidzhikov

pkg/gardenlet/apis/config/v1alpha1/zz_generated.conversion.go
```diff
@@ -891,8 +901,7 @@ func Convert_config_SeedClientConnection_To_v1alpha1_SeedClientConnection(in *co
 }
 
 func autoConvert_v1alpha1_SeedConfig_To_config_SeedConfig(in *SeedConfig, out *config.SeedConfig, s conversion.Scope) error {
-	// TODO: Inefficient conversion - can we improve it?
-	if err := s.Convert(&in.SeedTemplate, &out.SeedTemplate, 0); err != nil {
```
rfranzke, line 895: same here

ialidzhikov:
> I got rid of our fork of k8s.io/kube-openapi and our pinning of sigs.k8s.io/structured-merge-diff/v4 - do you see issues with this?

Yes. I guess we will be able to get rid of the replace for `sigs.k8s.io/structured-merge-diff/v4` only after https://github.com/kubernetes/kubernetes/pull/99038.

rfranzke:
OK, thanks @ialidzhikov. Also, @timuthy mentioned that we now should be able to get rid of https://github.com/gardener/gardener/tree/master/third_party/forked/kubernetes, so I will look into that.
/status author-action

pkg/apis/seedmanagement/v1alpha1/zz_generated.conversion.go
schrodit: We also ran into a similar problem with k8s 1.20.
It looks like a regression in the code generator for k8s 1.20: https://github.com/kubernetes/kubernetes/issues/98380

rfranzke:
OK, thanks to @ialidzhikov and @timuthy I've updated the PR now (and also use the latest, just released `v0.8.3` version of the controller-runtime library).

pkg/apis/seedmanagement/v1alpha1/zz_generated.conversion.go
rfranzke: Thanks @schrodit for pointing this out. Then I guess the workaround I applied is fine for now until a fix was released, right?

extensions/pkg/controller/utils_test.go
```diff
@@ -163,9 +166,11 @@ var _ = Describe("Utils", func() {
 				},
 			}
 			workerWithAnnotation := worker.DeepCopyObject()
-			ctx := context.TODO()
+			expectedWorker := worker.DeepCopy()
+			delete(expectedWorker.Annotations, annotation)
 
-			c.EXPECT().Patch(ctx, worker, client.MergeFrom(workerWithAnnotation))
+			ctx := context.TODO()
+			test.EXPECTPatch(ctx, c, expectedWorker, workerWithAnnotation)
```
ialidzhikov, line 173: Hmm, this is odd. I really liked the way it was before this PR (without any helper func for expecting patch)
```
			c.EXPECT().Patch(ctx, worker, client.MergeFrom(workerWithAnnotation))
```

I see that the old way fails with error like:

```
    Unexpected call to *client.MockClient.Patch([context.TODO 0xc00025ad80 0xc000899da0]) at /go/src/github.com/gardener/gardener/extensions/pkg/controller/utils.go:193 because:
    expected call at /go/src/github.com/gardener/gardener/extensions/pkg/controller/utils_test.go:171 doesn't match the argument at index 2.
    Got: &{application/merge-patch+json 0x1e25600 0xc00025b200 {false}}
    Want: is equal to &{application/merge-patch+json 0x1e25600 0xc00025afc0 {false}}
```

Do you know why with the new version of controller-runtime it now fails with similar error ? I see https://github.com/kubernetes-sigs/controller-runtime/pull/1406 but I cannot correlate it to the failure.

The reason I am asking is that currently the helper func `test.EXPECTPatch` is always expecting a `MergeFrom` PatchOption and I assume this func won't be useful if there is usage of `StrategicMergeFrom` or anything other PatchOption.

rfranzke: > Do you know why with the new version of controller-runtime it now fails with similar error ? I see kubernetes-sigs/controller-runtime#1406 but I cannot correlate it to the failure.

This is explained in the commit message of 0fdab6e6b1d1f858e4bf6e363b66a0dd957b51f1:

> Due to kubernetes-sigs/controller-runtime#1413,
naive gomock expectations don't work anymore. The reason is that the
`mergeFromPatch` struct has now a function attribute (`createPatch`),
and as `gomega` is using `reflect.DeepEqual` under the hood, the
comparison now fails:
>
> Func values are deeply equal if both are nil; otherwise they are not deeply equal.
(from https://golang.org/pkg/reflect/#DeepEqual)
>
> Hence, we introduce our own `test.EXPECTPatch()` function.
>
> Thanks to @timuthy for pointing this out.

I'm also not too happy with it and am very much open for better ideas.

> The reason I am asking is that currently the helper func test.EXPECTPatch is always expecting a MergeFrom PatchOption and I assume this func won't be useful if there is usage of StrategicMergeFrom or anything other PatchOption.

Maybe, but we don't use `StrategicMergeFrom` or any `PatchOption` yet, so the function can be extended as soon as we need it.

ialidzhikov: Ok, thanks for explaining. I guess the proper way to fix this is to adapt the `createPatch` field from func to an interface in the upstream. I can have a look into this one with some lower prio. But yeah, it seems that it is out of the scope of this PR.

ialidzhikov (dismissed):
/lgtm
/needs second-opinion

extensions/pkg/controller/utils_test.go
rfranzke: Yeah, probably @timebertt can help here as the initial author of this change in k-sigs/controller-runtime :)

rfranzke:
Rebased to resolve the conflicts.

ialidzhikov (approved):
/lgtm

timuthy (approved):
/lgtm
Thanks for cleaning up the forked code as well 👍

extensions/pkg/controller/utils_test.go
ialidzhikov: I now opened https://github.com/kubernetes-sigs/controller-runtime/pull/1439. Let me know if you have more elegant solution to the problem in mind.

PR #3653, opened 2021-03-03
Fix issues with hanging Seed deletion

timebertt:

/area usability
/kind bug
/priority normal

This PR fixes two issues with hanging Seed deletions:

- the federated seed controller now doesn't skip the controller setup for Seeds that are already marked for deletion (see code comments and https://github.com/gardener/gardener/issues/3495#issue-801096042 for detailed rationale)
- the gardenlet's seed controller now doesn't release a terminating seed until all controllerinstallations referencing the seed are cleaned up. This is needed, because the seed debootstrap would otherwise delete the gardener-resource-manager which is required to cleanup the controllerinstallations and also it deletes the seed's kubeconfig secret before other controllers can finish their work (e.g. cleaning up controllerinstallations)

Both issues contributed to races, that made automatic seed deletion deadlock in some cases and thus required manual intervention.

Part of #3495

**Release note**:

```bugfix operator
Some issues with hanging `ControllerInstallations` have been resolved, that caused the `Seed` deletion to deadlock and required manual cleanup.
```

### DIFF
--- a/pkg/gardenlet/controller/federatedseed/seed_control.go
+++ b/pkg/gardenlet/controller/federatedseed/seed_control.go
@@ -95,10 +95,18 @@ func NewFederatedSeedController(ctx context.Context, clientMap clientmap.ClientM
 
 func (c *Controller) seedAdd(obj interface{}) {
 	seedObj := obj.(*gardencorev1beta1.Seed)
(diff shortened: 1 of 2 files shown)


timuthy (approved):
/lgtm

rfranzke (approved):
/lgtm

PR #3654, opened 2021-03-03
Cleanup usage of PartialObjectMetadataList

timebertt:

/kind cleanup
/priority normal

I just noticed, the gvk of `PartialObjectMetadataList` is supposed to be set to `*List` instead.
Otherwise, if the reader is cached, it returns an error like this:
```
non-list type *v1.PartialObjectMetadataList (kind \"core.gardener.cloud/v1beta1, Kind=Shoot\") passed as output
```

Consequently, this PR cleans up our usages of `PartialObjectMetadataList`.

Actually, we don't have any usages of `PartialObjectMetadataList` in conjunction with cached readers.
Nevertheless, correcting this now, before someone runs into the trap...


timuthy (approved):
/lgtm

ialidzhikov (approved):
/lgtm

PR #3656, opened 2021-03-03
[GEP-13] Add typed image vectors to ManagedSeed gardenlet deployment parameters

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Introduces image vector API types and uses them to specify `imageVectorOverwrite` and `componentImageVectorOverwrites` in `seedmanagement.GardenletDeployment`. These new types are currently used only by the managed seed controller to properly produce the gardenlet Helm chart values. The existing `imagevector` package and its types are still used internally as before.

Fixes #

See https://github.com/gardener/gardener/pull/3627#discussion_r584520850.

**Release note**:

```other operator
It is now possible to specify `imageVectorOverwrite` and `componentImageVectorOverwrites` in the `spec.gardenlet.deployment` of a `ManagedSeed`.
```


stoyanr:
/cc @timebertt @danielfoehrKn
Still in draft because I need to test it a bit more, but otherwise ready for review.

stoyanr:
I tested it a bit more, all seems fine. I consider it ready for review / merge now.

pkg/apis/core/v1beta1/types_imagevector.go
```diff
@@ -0,0 +1,53 @@
...
+
+// ComponentImageVectors is a list of components and their images.
+type ComponentImageVectors []ComponentImageVector
+
+// ComponentImageVector specifies the name and the list of images for a component.
+type ComponentImageVector struct {
+	// Name is the component name, e.g. "etcd-druid"
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+	// ImageVector is the list of images for the component.
+	ImageVector ImageVector `json:"imageVector" protobuf:"bytes,2,opt,name=imageVector"`
```
timebertt, line 52: Just to be on the safe side: In `pkg/utils/imagevector` this field is named `imageVectorOverwrite`: https://github.com/gardener/gardener/blob/f27309dfc975088aa747dc897fe3864f9764c68d/pkg/utils/imagevector/types.go#L46

Was this a conscious decision? Should we use the same field names to avoid confusion?

pkg/apis/core/v1beta1/types_imagevector.go
```diff
@@ -0,0 +1,53 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1beta1
+
+// ImageVector is a list of images.
+type ImageVector []ImageSource
```
timebertt, line 18: I'm asking myself, if we should reserve the type name `ImageVector` for a type, that we can use to read the image vector from a file (like in `charts/images.yaml`) later on, once we eventually get rid of the duplicated code in `pkg/utils/imagevector`.

I.e. something like this:
```go
type ImageVector struct {
    Images []ImageSource `json:images`
}
```

We could use `[]ImageSource` directly in the `seedmanagement` API types.
What is your opinion?

pkg/apis/core/v1beta1/types_imagevector.go
```diff
@@ -0,0 +1,53 @@
...
+	RuntimeVersion *string `json:"runtimeVersion,omitempty" protobuf:"bytes,5,opt,name=runtimeVersion"`
+	// TargetVersion is the Kubernetes version that the image can target (operate on).
+	// It should be specified if the image can target only specific Kubernetes version(s).
+	// For supported syntax, see https://github.com/Masterminds/semver#hyphen-range-comparisons
+	// +optional
+	TargetVersion *string `json:"targetVersion,omitempty" protobuf:"bytes,6,opt,name=targetVersion"`
+}
+
+// ComponentImageVectors is a list of components and their images.
+type ComponentImageVectors []ComponentImageVector
```
timebertt, line 45: Same here question about the type for reading from a file applies here as well.

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -1247,3 +1250,55 @@ func BackupBucketIsErroneous(bb *gardencorev1beta1.BackupBucket) (bool, string)
 	}
 	return true, lastErr.Description
 }
+
+func ReadImageVector(r io.Reader) (gardencorev1beta1.ImageVector, error) {
```
timebertt, line 1254: doc string missing for exported symbols

pkg/apis/core/v1beta1/helper/helper.go
```diff
@@ -1247,3 +1250,55 @@ func BackupBucketIsErroneous(bb *gardencorev1beta1.BackupBucket) (bool, string)
...
+	iv, err := imagevector.Read(r)
+	if err != nil {
+		return nil, err
+	}
+	var images []gardencorev1beta1.ImageSource
+	for _, is := range iv {
+		images = append(images, gardencorev1beta1.ImageSource(*is))
+	}
+	return images, nil
```
timebertt, line 1263: WDYT about ditching the `imagevector` package and the custom conversion logic in favor of proper decoders?

pkg/gardenlet/controller/managedseed/managedseed_valueshelper.go
```diff
@@ -299,38 +307,48 @@ func getParentGardenletDeployment(imageVector imagevector.ImageVector, shoot *ga
...
-	var imageVectorOverwrite string
+func getParentImageVectorOverwrite() (gardencorev1beta1.ImageVector, error) {
 	if overWritePath := os.Getenv(imagevector.OverrideEnv); len(overWritePath) > 0 {
 		data, err := ioutil.ReadFile(overWritePath)
 		if err != nil {
-			return "", err
+			return nil, err
 		}
-		imageVectorOverwrite = string(data)
+		return gardencorev1beta1helper.ReadImageVector(strings.NewReader(string(data)))
```
timebertt, line 338: Can you use `os.Open` and use the returned file as a reader instead of reading all bytes and copying them into another reader?

pkg/gardenlet/controller/managedseed/managedseed_valueshelper.go
```diff
@@ -299,38 +307,48 @@ func getParentGardenletDeployment(imageVector imagevector.ImageVector, shoot *ga
...
-	var componentImageVectorOverwrites string
+func getParentComponentImageVectorOverwrites() (gardencorev1beta1.ComponentImageVectors, error) {
 	if overWritePath := os.Getenv(imagevector.ComponentOverrideEnv); len(overWritePath) > 0 {
 		data, err := ioutil.ReadFile(overWritePath)
 		if err != nil {
-			return "", err
+			return nil, err
 		}
-		componentImageVectorOverwrites = string(data)
+		return gardencorev1beta1helper.ReadComponentImageVectors(strings.NewReader(string(data)))
```
timebertt, line 349: Same here.

timebertt (commented):
Thanks a lot for adding proper API types for this!
I hope we can get rid of this `pkg/utils/imagevector` package soon and only use the proper API types from there on.

pkg/apis/core/v1beta1/types_imagevector.go
stoyanr: Yes, absolutely it was conscious and long thought over. I started with the same names but soon realized something is wrong. In a type named `ComponentImageVector` a field should not be named `ImageVectorOverwrite`, it makes no sense. If this field name is kept, the type itself should be named `ComponentImageVectorOverwrite`. But then, there is an inconsistency between the names of the 2 types - `ImageVector` vs. `ComponentImageVectorOverwrite`, so also the first type should be named `ImageVectorOverwrite` instead of `ImageVector`.

So I really see only 2 sensible naming alternatives:
* `ImageVector`, `ComponentImageVector`, and field names avoiding "overwrite", suggesting that the types are used for all kind of image vectors, including overwrites.
* `ImageVectorOverwrite`, `ComponentImageVectorOverwrite`, and field names using "overwrite", suggesting that the types are used only for overwriting image vectors.

Although currently the 2 new types are used only for overwrites, I tend to strongly prefer the first alternative since it's both shorter and doesn't imply any restrictions on the types usage.

Compatibility is not a concern from my perspective - the existing types are anyway not suitable to become API types without changes, so I chose the most sensible design for the new types and made sure they are loaded / saved from / to the same stream representation as the old.

pkg/apis/core/v1beta1/types_imagevector.go
stoyanr: I started like this (and similarly for `ComponentImageVector`), but abolished the structs at some point. It's just more verbose without any additional value. From my perspective, whatever you can do with a struct type you can also do with the slice alias, including "reading the image vector from a file" - there is even already helper code for this, which could be used for what you suggest without any changes.

pkg/apis/core/v1beta1/types_imagevector.go
stoyanr: See the answer to the other question.

pkg/apis/core/v1beta1/helper/helper.go
stoyanr: Will add

pkg/gardenlet/controller/managedseed/managedseed_valueshelper.go
stoyanr: There is no copying, `strings.NewReader` just creates a wrapper over the string to allow it to function as a `Reader`. I could do as you suggest, but the only thing that this would save would be some part of the memory to hold `data` (precisely, the difference between the length of `data` and 512). Since our image vectors are rather small, this is negligible. I will also have to introduce `defer os.Close` and proper error handling for everything, so the amount of code will be the same or more. Honestly I prefer the current implementation, but I could of course change it if you insist.

pkg/apis/core/v1beta1/helper/helper.go
stoyanr: We don't want to do this right now, isn't this what we agreed? Otherwise we would have to change a lot more code, deal with backward compatibility, etc. Perhaps it makes sense, but let's do it in a different PR. I regard this PR as the minimal change required to introduce strongly typed image vectors and use them in the API. My goal was to touch the existing `imagevector` as little as possible, and do it in a compatible way.

Under these assumptions, concretely the method above has to stay as it is, since we are constrained by the representation already used in existing configuration files / configmaps.

pkg/gardenlet/controller/managedseed/managedseed_valueshelper.go
stoyanr: See above.

example/55-managedseed-gardenlet.yaml
```diff
@@ -34,6 +34,17 @@ spec:
...
+#     - name: pause-container
+#       repository: my-custom-image-registry/pause-amd64
+#       tag: "3.1"
+#       targetVersion: ">= 1.12"
+#     componentImageVectorOverwrites:
+#     - name: etcd-druid
+#       imageVector:
+#       - name: etcd
+#         repository: my-custom-image-registry/etcd
+#         tag: latest
```
rfranzke, line 47: Hm, what's your immediate motivation to add this feature? IIUC, the gardenlet deployed for `ManagedSeed`s inherit the same image vector overwrites like the parent gardenlet, right? So, overriding this here would in the majority of the cases not be necessary, is it? Do you have a concrete use case in mind or did you just add it because you had time and because it's in general a nice feature? Are there any plans to implement the same for the extensions? AFAIK they don't support per-seed overriding of this data.

Second question would be whether this content shouldn't rather be put into a `ConfigMap` and referenced from here. Typically, this can grow quite fast (many entries), and putting all of this here would blow up the `ManagedSeed` resource.

stoyanr: See https://github.com/gardener/gardener/pull/3627#discussion_r584520850 for the motivation. Even though this is not expected to be used (much) in managed seeds (and this is why it wasn't implemented originally), we want to reuse the `GardenletDeployment` type (and its defaults, validation, etc.) for the gardenlet landscaper component, and there it's a must. So after discussing this with @timebertt and @danielfoehrKn , we decided to add the 2 fields to this shared type. I think it doesn't hurt for managed seeds as well, and it even makes the current "merge with parent" implementation more straightforward and consistent.

Regarding "blowing-up" resources - I see your point, but referencing `ConfigMap` would contradict the entire idea of having strongly typed image vectors in the first place. Instead, if separate resources are to be used, they should rather be dedicated image vector resources, e.g. `ImageVector` and `ComponentImageVectors`. Perhaps this makes sense, we should discuss this idea further.

rfranzke: > Regarding "blowing-up" resources - I see your point, but referencing ConfigMap would contradict the entire idea of having strongly typed image vectors in the first place. Instead, if separate resources are to be used, they should rather be dedicated image vector resources, e.g. ImageVector and ComponentImageVectors. Perhaps this makes sense, we should discuss this idea further.

I don't know, it feels odd to introduce such custom resources. This is clearly configuration data for the gardenlet, so having a `ConfigMap` seems way more appropriate.
I can imagine that inlining this in the `Imports` resource for the gardenlet landscaper component might also become a problem for the Landscaper and its deployers in the long run.
 
What was again the motivation for those strongly typed fields? Was it just to have early validation? If yes, why isn't it "good enough" if the gardenlet fails after start-up if its configuration is broken? This is what is happening for most components, including the kubelet. Yes, we could have argued similarly for the inlined `GardenletConfiguration` in `ManagedSeed`, but it's still different because it grow arbitrarily in size like the image vectors overrides.

pkg/apis/core/v1beta1/types_imagevector.go
timebertt: Ok, thanks for the clarification. I agree with you and prefer the new naming (alternative 1) 👍

pkg/apis/core/v1beta1/helper/helper.go
timebertt: Ok for me. Let's do it later on.

pkg/gardenlet/controller/managedseed/managedseed_valueshelper.go
timebertt: I don't insist. But you're right, it shouldn't matter too much.
Resolving this one.

example/55-managedseed-gardenlet.yaml
timebertt: Ok, agree with the point about avoiding too large resources because of an inlined ImageVector.

> What was again the motivation for those strongly typed fields? Was it just to have early validation?

The main motivation for having strongly typed fields was to have a clean API for values of which we already know the concrete structure, i.e. not representing image vectors as strings, because there is a well-known structure for them.
This would in turn then allow early validation of the structural integrity of the image vecotr, which is desirable FMPOV.

> If yes, why isn't it "good enough" if the gardenlet fails after start-up if its configuration is broken? This is what is happening for most components, including the kubelet.

There is also other configuration data we try to validate as early on as possible, e.g. Shoot audit policy config maps. We do this, because it will take quite some time until the error surfaces if we don't fail early on. Same applies to the image vector IMO. If we create a new shoot and deploy a gardenlet into it after ~8m and fail only because there is a space missing in the YAML, it would be quite the bad UX.
Another scenario where it would be even more problematic is when rolling existing gardenlets of a `ManagedSeedSet`. If there is some YAML-format issue and new gardenlet instances won't be able to start because of that, the existing Seeds won't have any functional gardenlet, although the error could have been caught very early on.

timebertt: Anyway, @stoyanr and I will have a chat about this later on and discuss how to proceed here.

pkg/apis/core/v1beta1/types_imagevector.go
timebertt: Let's discuss this in person later on, as it might depend on the outcome of our decision regarding https://github.com/gardener/gardener/pull/3656#discussion_r590509303.

stoyanr:
Closing this PR. According to our discussion with @timebertt, @rfranzke, and @danielfoehrKn we don't want to use embedded fields with image vectors in other resources because they tend to be large. @danielfoehrKn later determined that image vectors are also not needed in `GardenletDeployment` by the gardenlet landscaper component. Therefore, most of this PR is not needed anymore, the only valuable part would be the validation of image vectors. This will be added in a new PR on the original internal types.

Issue #1648, opened 2019-11-22
Purge context.TODO()


ialidzhikov:
There are multiple usages of `context.TODO()` in g/gardener and g/gardener-extensions in non-test code. We could revise, context-ify our packages and purge the usage of context.TODO().

https://golang.org/pkg/context/

rfranzke:
Note: Will improve (but not fully be fixed) with #2489

PR #1094, opened 2019-06-12
Reconcile shoots only during maintenance time

adracus:
**What this PR does / why we need it**:
* Restructure & refactor `reconcileShootKey` to make requeues / failures more obvious
* Change behavior so reconciles only happen during maintenance times
* Always try to sync cluster resource to seed if possible
* Create test cases

**Which issue(s) this PR fixes**:
Fixes #251 

**Special notes for your reviewer**:
cc @rfranzke @vlerenc 

**Release note**:

```noteworthy operator
It is now possible to configure Gardener to only reconcile shoots during their respective maintenance time windows by setting `.controllers.shoot.reconcileInMaintenanceOnly=true` in the controller-manager's component configuration. If the shoot's `spec` changes outside of their maintenance time windows then they will be reconciled immediately. Also, they will be reconciled immediately if they didn't complete their last operation successfully. With this in place users are now able to have a HA setup of shoots if their maintenance time windows do not overlap.
```

PR #3658, opened 2021-03-04
Cleanup context.TODO(), migrate from DeprecatedCreateWorker to CreateWorker (1)

rfranzke:

/area dev-productivity
/kind cleanup
/priority normal

This PR starts with migration a couple of our controllers from `controllerutils.DeprecatedCreateWorker` (deprecated since #1094) to `controllerutils.CreateWorker`. This helps to get rid of some remaining usages of `context.TODO()`.

Part of #1648 

/squash

**Release note**:

```other operator
NONE
```

### DIFF
--- a/cmd/gardener-scheduler/app/gardener_scheduler.go
+++ b/cmd/gardener-scheduler/app/gardener_scheduler.go
@@ -24,6 +24,9 @@ import (
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	gardencoreinformers "github.com/gardener/gardener/pkg/client/core/informers/externalversions"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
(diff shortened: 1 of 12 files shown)


timuthy (approved):
/lgtm
One comment: In some places we're now using a client instead of a lister ([here](https://github.com/gardener/gardener/pull/3658/commits/6decbe9aa94308a6b2b8ba3666bf70399a075a74#diff-fad5b5d66446acece1d81a3b7f5f8604cafad2fa05b22b7250bc7a4846abb7a1R121) for example), so that we'll have double maintained caches/list-watches in case the Controller-Runtime client cache is enabled. I would have rather waited for the beta graduation of the client cache and changed to client usage in a dedicated refactoring PR. 
Nevertheless, I also don't see it as critical because it's only for the GCM and thus in-cluster traffic. However, you may want to consider this when you continue your refactoring for Gardenlet components.

rfranzke:
Thanks @timuthy, I will keep an 👁️

PR #3660, opened 2021-03-04
[GEP-13] Make ManagedSeed spec.shoot field optional

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Makes the `ManagedSeed` `spec.shoot` field optional to avoid having to specify it in a `ManagedSeedSet` template.

Fixes #

**Release note**:

```improvement operator
NONE
```


pkg/registry/seedmanagement/managedseed/strategy.go
```diff
@@ -165,5 +165,12 @@ func ShootNameTriggerFunc(obj runtime.Object) string {
 		return ""
 	}
 
+	return GetShootName(managedSeed)
+}
+
+func GetShootName(managedSeed *seedmanagement.ManagedSeed) string {
```
rfranzke, line 172:
```suggestion
func getShootName(managedSeed *seedmanagement.ManagedSeed) string {
```

stoyanr: It's used outside of this package, see https://github.com/stoyanr/gardener/blob/34bef2b51fa1c8083aa01ab22227353ada296259/pkg/registry/seedmanagement/managedseed/storage/tableconvertor.go#L87

rfranzke: OK, missed that. Then I would only suggest to add a doc string as it's exported. Rest lgtm.

stoyanr: Thanks, done.

rfranzke (approved):
/lgtm

PR #3662, opened 2021-03-05
Extend list of non-retryable error codes

rfranzke:

/area ops-productivity scalability
/kind enhancement
/priority normal

To prevent "spamming" an infrastructure provider's API, we are now extending list of non-retryable error codes with `ERR_INFRA_INSUFFICIENT_PRIVILEGES`, `ERR_INFRA_QUOTA_EXCEEDED`, and `ERR_INFRA_DEPENDENCIES`.

`ERR_INFRA_RESOURCES_DEPLETED` was not added because it usually resolves automatically after some time.

**Release note**:

```noteworthy user
When a shoot is erroring with `ERR_INFRA_INSUFFICIENT_PRIVILEGES`, `ERR_INFRA_QUOTA_EXCEEDED` or `ERR_INFRA_DEPENDENCIES` then it is now immediately set to the `Failed` status (this already happens also for `ERR_INFRA_UNAUTHORIZED` or `ERR_CONFIGURATION_PROBLEM`). This prevents Gardener from automatically retrying the operation. If you are hit by it, please manually retry the operation once you have resolved the issue.
```

### DIFF
--- a/pkg/apis/core/v1beta1/helper/errors.go
+++ b/pkg/apis/core/v1beta1/helper/errors.go
@@ -203,8 +203,9 @@ func HasNonRetryableErrorCode(lastErrors ...gardencorev1beta1.LastError) bool {
 		for _, code := range lastError.Codes {
 			if code == gardencorev1beta1.ErrorInfraUnauthorized ||
 				code == gardencorev1beta1.ErrorInfraInsufficientPrivileges ||
-				code == gardencorev1beta1.ErrorConfigurationProblem ||
-				code == gardencorev1beta1.ErrorInfraDependencies {
+				code == gardencorev1beta1.ErrorInfraDependencies ||
+				code == gardencorev1beta1.ErrorInfraQuotaExceeded ||
+				code == gardencorev1beta1.ErrorConfigurationProblem {
 				return true
 			}
 		}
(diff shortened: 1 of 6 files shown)


pkg/apis/core/v1beta1/helper/errors.go
```diff
@@ -201,7 +201,10 @@ func LastErrorWithTaskID(description string, taskID string, codes ...gardencorev
 func HasNonRetryableErrorCode(lastErrors ...gardencorev1beta1.LastError) bool {
 	for _, lastError := range lastErrors {
 		for _, code := range lastError.Codes {
-			if code == gardencorev1beta1.ErrorInfraUnauthorized || code == gardencorev1beta1.ErrorConfigurationProblem {
+			if code == gardencorev1beta1.ErrorInfraUnauthorized ||
```
ialidzhikov, line 204: You can add this detail in  `/docs/usage/shoot_status.md#error-codes` - some of the error codes there have the description `It is classified as a non-retryable error code.`. You can also add this to the doc string of the field.

rfranzke: done, thank you!

timuthy (dismissed):
/lgtm

ialidzhikov (dismissed):
/lgtm

vlerenc:
I am wondering, maybe we should also treat `ErrorInfraQuotaExceeded` as non-retryable?

Most cases where we actually understand the error will probably be non-retryable, unless it's a fluke in the infrastructure like e.g. `ErrorInfraResourcesDepleted` (it may resolve without end user intervention).

Isn't that then the rule? If something requires end user intervention like not authorised, missing permissions, infra dependencies, quota depleted, etc. we say "please fix it" and in cases where the problem is with the infrastructure itself, we retry? WDYT?

rfranzke:
@vlerenc Yes, thanks, makes sense. I pushed 9dee074 which adds this behaviour.
/squash

ialidzhikov (approved):
/lgtm

PR #3673, opened 2021-03-07
Fix update of generatedSecretRef in status of backupBucket

MartinWeindel:

/area control-plane
/kind bug
/priority normal

On creating a new Azure seed, the status of backupbucket in the virtual garden cluster was not updated with the `generatedSecretRef`. As the status of the backupbucket is an own object, using `CreateOrUpdate` on the backupbucket object itself ignores changes to status fields. Instead the status must be updated explicitly.

Fixes #

**Release note**:

```bugfix operator
An issue preventing the status of the BackupBucket to be properly updated is now fixed.
```

### DIFF
--- a/pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go
+++ b/pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go
@@ -269,12 +269,11 @@ func (a *actuator) waitUntilBackupBucketExtensionReconciled(ctx context.Context)
 			}
 
 			if generatedSecretRef != nil || backupBucket.Status.ProviderStatus != nil {
-				_, err := controllerutil.CreateOrUpdate(ctx, a.gardenClient.Client(), a.backupBucket, func() error {
+				return kutil.TryUpdateStatus(ctx, kretry.DefaultBackoff, a.gardenClient.Client(), a.backupBucket, func() error {
 					a.backupBucket.Status.GeneratedSecretRef = generatedSecretRef
 					a.backupBucket.Status.ProviderStatus = backupBucket.Status.ProviderStatus
 					return nil
 				})
-				return err
 			}
 
 			return nil
--- a/pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go
+++ b/pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go
@@ -269,7 +269,7 @@ func (a *actuator) waitUntilBackupBucketExtensionReconciled(ctx context.Context)
 			}
 
 			if generatedSecretRef != nil || backupBucket.Status.ProviderStatus != nil {
-				return kutil.TryUpdateStatus(ctx, kretry.DefaultBackoff, a.gardenClient.Client(), a.backupBucket, func() error {
+				return kutil.TryUpdateStatus(ctx, kretry.DefaultBackoff, a.gardenClient.DirectClient(), a.backupBucket, func() error {
 					a.backupBucket.Status.GeneratedSecretRef = generatedSecretRef
 					a.backupBucket.Status.ProviderStatus = backupBucket.Status.ProviderStatus
 					return nil


MartinWeindel:
Thanks to @dkistner for helping on analysing the issue.

rfranzke:
Thanks for the PR, nice catch. Can you explain why we didn't recognize it earlier? The code seems to be like this for almost a year. Why does it only happen to fail now?

ialidzhikov:
I checked now locally with a sample `main.go` and it seems that as expected `controllerutil.CreateOrUpdate` does not update the status. Switching the same code to `kutil.TryUpdateStatus` worked well.

ialidzhikov (dismissed):
/lgtm

MartinWeindel:
> Thanks for the PR, nice catch. Can you explain why we didn't recognize it earlier? The code seems to be like this for almost a year. Why does it only happen to fail now?

I asked the question myself, but I cannot explain it.

rfranzke:
@dkistner maybe? It would be good to understand this.

dkistner:
No unfortunately I did also not really understand why this happened.

I just had the suspicion that it is might only an issue for Azure Seeds as they are using different credentials to access the backup object store (storage account). I guess on other providers the generated secret in the status of the BackupBucket contains more or less the same fields as the secret in the spec of the BackupBucket and if the generated secret would be not present the one from the spec is used. Therefore we might not noticed it? I'm just guessing and did not verify that yet. One argument against my suspicion is that we have Azure Seeds which are younger than a year and had at least to my knowledge not the same issue.

pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go
```diff
@@ -269,12 +269,11 @@ func (a *actuator) waitUntilBackupBucketExtensionReconciled(ctx context.Context)
 			}
 
 			if generatedSecretRef != nil || backupBucket.Status.ProviderStatus != nil {
-				_, err := controllerutil.CreateOrUpdate(ctx, a.gardenClient.Client(), a.backupBucket, func() error {
+				return kutil.TryUpdateStatus(ctx, kretry.DefaultBackoff, a.gardenClient.Client(), a.backupBucket, func() error {
```
ialidzhikov: Can you adapt to use the `DirectClient`? I see that some other parties (the progressReporter for example) that update the status use the `DirectClient`.

```suggestion
				return kutil.TryUpdateStatus(ctx, kretry.DefaultBackoff, a.gardenClient.DirectClient(), a.backupBucket, func() error {
```

ialidzhikov (commented):
I tried to reproduce locally the issue that this PR is targeting to fix but I had no success.
I see that there is one `backupBucket` variable in the actuator and actually all updates refer to this shared `a.backupBucket` variable.

https://github.com/gardener/gardener/blob/8480f604ea285ede4d0dd7cc1959826ec9ca2a94/pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go#L59

So I think I am not able to reproduce this issue with the master branch as in 

https://github.com/gardener/gardener/blob/8480f604ea285ede4d0dd7cc1959826ec9ca2a94/pkg/gardenlet/controller/backupbucket/backup_bucket_actuator.go#L272-L276

we again mutate the shared variable `a.backupBucket`.

Although `controllerutil.CreateOrUpdate` does not update the status, we mutate the shared variable and it is being updated multiple times afterwards - by the progressReporter and by the reconciler afterwards. This could explain why this issue is not easily reproducible locally and why we haven't hit it so far.

So as conclusion I think we are good to go with this PR. As shared above, I tried out locally and `controllerutil.CreateOrUpdate` does not update the obj status, we should use `kutil.TryUpdateStatus` as done is this PR. 

I am leaving only one minor comment inline.

ialidzhikov (approved):
/lgtm
/squash

PR #3686, opened 2021-03-10
Fix etcd-druid seed unbootstrap

### LINKED ISSUE #3685: De-bootstrapping a seed sometimes fails with "no matches for kind Etcd"

/area control-plane
/kind bug
/priority normal

**What happened**:

De-bootstrapping a seed sometimes fails with:
```
Failed to unbootstrap Seed Cluster (1 error occurred:\n\t* no matches for kind \"Etcd\" in version \"druid.gardener.cloud/v1alpha1\"\n\n)
```

No such errors during seed de-bootstrapping.

* With `master`: delete a Seed resource, wait until the `etcd-druid` managed resource is deleted from the seed cluster, and immediately restart the `gardenlet` responsible for the seed manually.
* With `v1.18.0-dev-7f0ae078d09fe5927d01a4df0821c2bf0d98332a` the issue is reproducible without having to restart gardenlet.

I believe the issue is caused by the fact that the code in https://github.com/gardener/gardener/blob/e23f221891e9ef45d09e57f553a02e7191f210ed/pkg/operation/botanist/controlplane/etcd/bootstrap.go#L295 is not re-entrant - once the managed resource is deleted, it can no longer succeed because listing the `Etcd` resources requires the CRD, but the CRD had already been deleted.

**Environment**:

- Gardener version: `v1.18.0-dev-7f0ae078d09fe5927d01a4df0821c2bf0d98332a`
- Others:

stoyanr:

/area control-plane
/kind bug
/priority normal

Fixes #3685

**Release note**:

```bugfix operator
A bug preventing seed deletion to hang due to already deleted CRD `etcds.druid.gardener.cloud` is now fixed.
```

### DIFF
--- a/pkg/operation/botanist/controlplane/etcd/bootstrap.go
+++ b/pkg/operation/botanist/controlplane/etcd/bootstrap.go
@@ -36,6 +36,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm

PR #3687, opened 2021-03-10
Allow ingress traffic to coredns from a pod running with hostNetwork

DockToFuture:

/area networking
/kind bug
/priority normal

Allow ingress traffic to coredns from a pod running with `hostNetwork: true` and `dnsPolicy: ClusterFirstWithHostNet`.
This was denied in the past.

**Release note**:

```other operator
Allow ingress traffic to coredns from a pod running with `hostNetwork: true` and `dnsPolicy: ClusterFirstWithHostNet`
```

/cc: @mandelsoft @vlerenc

### DIFF
--- a/pkg/operation/botanist/addons.go
+++ b/pkg/operation/botanist/addons.go
@@ -226,6 +226,7 @@ func (b *Botanist) generateCoreAddonsChart(ctx context.Context) (*chartrenderer.
 			"vpaEnabled":        b.Shoot.WantsVerticalPodAutoscaler,
 		}
 		coreDNSConfig = map[string]interface{}{
+			"nodeNetwork": b.Shoot.GetNodeNetwork(),
 			"service": map[string]interface{}{
 				"clusterDNS": b.Shoot.Networks.CoreDNS.String(),
 				// TODO: resolve conformance test issue before changing:
--- a/charts/shoot-core/components/charts/coredns/templates/coredns-networkpolicy.yaml
+++ b/charts/shoot-core/components/charts/coredns/templates/coredns-networkpolicy.yaml
@@ -36,6 +36,8 @@ spec:
       podSelector: {}
     - ipBlock:
         cidr: {{ .Values.global.podNetwork }}
+    - ipBlock:
+        cidr: {{ .Values.nodeNetwork }}
   policyTypes:
   - Egress
   - Ingress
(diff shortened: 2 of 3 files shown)


vlerenc:
Thank you very much @DockToFuture, that was very fast.

pkg/operation/botanist/addons.go
```diff
@@ -226,6 +226,9 @@ func (b *Botanist) generateCoreAddonsChart(ctx context.Context) (*chartrenderer.
 			"vpaEnabled":        b.Shoot.WantsVerticalPodAutoscaler,
 		}
 		coreDNSConfig = map[string]interface{}{
+			"global": map[string]interface{}{
+				"nodeNetwork": b.Shoot.GetNodeNetwork(),
+			},
```
rfranzke: Shouldn't this be here:

https://github.com/gardener/gardener/blob/fb1715be4030b69279ff75e2337a417d4e7e66ee/pkg/operation/botanist/addons.go#L223-L227

rfranzke (changes requested):
/assign

pkg/operation/botanist/addons.go
DockToFuture: Why? I wanna have it in the `coredns` chart and not in the `components` chart.

DockToFuture: Why? I wanna have it in the `coredns` chart and not in the `components` chart.

rfranzke: You have added it to the `global` field in the `values.yaml`, see https://github.com/gardener/gardener/pull/3687/files#diff-e0e1ddf2d0b93e57689704cca6bffbba3f2022e115ea56defdea31d239aff3a1R6.

DockToFuture: Yes, for the `coredns` chart but not for the `components` chart, which would be here https://github.com/gardener/gardener/blob/master/charts/shoot-core/components/values.yaml#L1-L4

rfranzke: OK, then it might work by accident. Still, I suggest to have it not below `global` if you only want it for CoreDNS (even if you can make it work like this). The `global` should only be used for values shared across multiple charts.

DockToFuture: I've removed it from the global section.

rfranzke (approved):
/lgtm

PR #3700, opened 2021-03-12
Dynamically refresh configuration secrets in the `garden` namespace

### LINKED ISSUE #1725: Dynamically refresh configuration secrets in the `garden` namespace
Today, the gardenlet does read secrets for default domains, internal domain, alerting, etc. only once during start-up, and then never again.
This forces administrators to manually restart the gardenlet in all seed clusters in case he/she changes such a secret.
Although this scenario happens very rarely, we should enable the gardenlet to dynamically refresh the secrets such that a restart is not required.

timuthy:

/area security
/area robustness
/kind enhancement
/priority normal

With this PR, Gardenlet and Gardener-Controller-Manager read "garden" configuration secrets dynamically during reconciliation instead of only once during start-up (see #1725 for more details). Caching is automatically used via the Controller-Runtime cache (in Gardenlet) or a SecretLister (in Gardener-Controller-Manager).

Fixes #1725

Additional wrappers for the Controller-Runtime cache were required to circumvent that a secret watch is automatically started for **all namespaces**. We only want to start watches to seed namespaces in the Garden cluster (#3582) and for other namespaces e.g. `garden` we might fall back to a get request via the API reader. This is necessary because due to security concerns we'll restrict any list-watches to non-seed related namespaces in the Garden cluster (#1723).

**Release note**:

```feature operator
Gardener-Controller-Manager now reads Gardener configuration secrets like `internal-domain`, `default-domain`, etc. dynamically during reconciliation. Earlier the secrets were only read and stored in memory during start-up, so that any changes to those secrets were not reflected until the next restart.
```
```feature operator
Gardenlet now reads Gardener configuration secrets like `internal-domain`, `default-domain`, etc. dynamically during shoot reconciliation. Earlier the secrets were only read and stored in memory during start-up, so that any changes to those secrets were not reflected until the next restart.
```


pkg/client/kubernetes/cache/cache.go
```diff
@@ -0,0 +1,126 @@
...
+var _ cache.Cache = &Aggregator{}
+
+// Aggregator is a cache that can hold different cache implementations depending on the objects' GVKs.
+type Aggregator struct {
+	fallback   cache.Cache
+	gvkToCache map[schema.GroupVersionKind]cache.Cache
+	scheme     *runtime.Scheme
+}
+
+func (c *Aggregator) cacheForObject(obj runtime.Object) cache.Cache {
```
rfranzke: Should `client.Object` be used here?

pkg/client/kubernetes/cache/cache.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	cache, ok := c.gvkToCache[gvk]
+	if !ok {
+		cache = c.fallback
+	}
+	return cache
+}
+
+func (c *Aggregator) Get(ctx context.Context, key client.ObjectKey, obj client.Object) error {
```
rfranzke: doc string (same for all below)

pkg/client/kubernetes/cache/cache.go
```diff
@@ -0,0 +1,126 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	runtimecache "sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ cache.Cache = &Aggregator{}
+
+// Aggregator is a cache that can hold different cache implementations depending on the objects' GVKs.
+type Aggregator struct {
+	fallback   cache.Cache
```
rfranzke: WDYT about

```suggestion
	fallbackCache   cache.Cache
```

pkg/client/kubernetes/cache/cache.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (c *Aggregator) Start(ctx context.Context) error {
+	for gvk, cache := range c.gvkToCache {
+		go func(gvk schema.GroupVersionKind, cache runtimecache.Cache) {
+			err := cache.Start(ctx)
+			if err != nil {
+				logger.Logger.Errorf("cache failed to start for %q: %v", gvk.String(), err)
+			}
+		}(gvk, cache)
```
rfranzke, line 104: Are these function parameters needed?

pkg/client/kubernetes/cache/cache.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (c *Aggregator) WaitForCacheSync(ctx context.Context) bool {
+	synced := true
+	for _, cache := range c.gvkToCache {
+		if s := cache.WaitForCacheSync(ctx); !s {
+			synced = s
+		}
+	}
+	synced = synced && c.fallback.WaitForCacheSync(ctx)
+	return synced
```
rfranzke:
```suggestion
	return synced && c.fallback.WaitForCacheSync(ctx)
```

pkg/client/kubernetes/cache/cache.go
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke, line 1: /invite @timebertt 
/needs second-opinion

pkg/client/kubernetes/cache/cache_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestCache(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Cache Suite")
```
rfranzke:
```suggestion
	RunSpecs(t, "Client Kubernetes Cache Suite")
```

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
```diff
@@ -19,11 +19,16 @@ import (
 	"fmt"
 
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	corev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
```
rfranzke:
```suggestion
	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
```

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
```diff
@@ -52,18 +65,64 @@ func (f *GardenClientSetFactory) CalculateClientSetHash(context.Context, clientm
...
+	}
+	if ls := f.SeedSelector.Selector; ls != nil {
+		directClient, err := client.New(f.RESTConfig, client.Options{
+			Scheme: kubernetes.GardenScheme,
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		seeds := &corev1beta1.SeedList{}
```
rfranzke: Can you move this to line 100?

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
```diff
@@ -52,18 +65,64 @@ func (f *GardenClientSetFactory) CalculateClientSetHash(context.Context, clientm
...
+		if err != nil {
+			return nil, err
+		}
+
+		if err := directClient.List(ctx, seeds, client.MatchingLabelsSelector{Selector: seedSelector}); err != nil {
+			return nil, err
+		}
+
+		for _, seed := range seeds.Items {
+			seedNamespaces = append(seedNamespaces, seedpkg.ComputeGardenNamespace(seed.Name))
```
rfranzke: What happens when a new seed that would match would be added? I guess this might need to be refreshed when the client is retrieved from the client map, or?

Edit: Ah, or is this done here (https://github.com/gardener/gardener/pull/3700/commits/007c16d467a077aea8976410e7c5426388185e6d#diff-628c8891682f4a4b0f5c49cb8051057ed8a5a3bf338438659ef68ccfb66385e5R99-R103)? But then it must be done for all the places where the garden client is used and interacts with only matching seeds, or?

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
```diff
@@ -52,18 +65,64 @@ func (f *GardenClientSetFactory) CalculateClientSetHash(context.Context, clientm
...
+		}
+	}
+
+	// Do not use multi-namespace informers if `CachedRuntimeClients` feature is enabled because today we still need
```
rfranzke: Can you remove or adapt this comment?

pkg/gardenlet/apis/config/helper/helpers.go
```diff
@@ -36,6 +38,30 @@ func SeedNameFromSeedConfig(seedConfig *config.SeedConfig) string {
 	return seedConfig.SeedTemplate.Name
 }
 
+// SeedNames returns all seed names matching the given config or LabelSelector if no config is given.
+func SeedNames(seedConfig *config.SeedConfig, seedLister gardencorelisters.SeedLister, labelSelector *metav1.LabelSelector) []string {
```
rfranzke: I don't think we should have a function in the `helper` package that itself depends on the lister. Maybe you can change the signature to already provide the `[]*gardencorev1beta1.Seed`?

pkg/gardenlet/controller/factory.go
```diff
@@ -128,15 +131,20 @@ func (f *GardenletControllerFactory) Run(ctx context.Context) error {
...
+	if len(seedNames) < 1 {
+		return fmt.Errorf("no seed selected by Gardenlet")
 	}
 
+	// Read Garden secrets from any seed namespace as we assume they are synced accordingly by the Gardener Controller Manager
+	secrets, err := garden.ReadGardenSecrets(
+		ctx,
+		k8sGardenClient.Cache(),
+		f.k8sGardenCoreInformers.Core().V1beta1().Seeds().Lister(),
+		seedpkg.ComputeGardenNamespace(seedNames[0]),
```
rfranzke: Why only `seedNames[0]` is taken?

pkg/gardenlet/controller/shoot/shoot_care_control.go
```diff
@@ -277,13 +279,22 @@ func (c *defaultCareControl) Care(shootObj *gardencorev1beta1.Shoot, key string)
...
 
+	// Only read Garden secrets once because we don't rely on up-to-date secrets for health checks.
+	if c.gardenSecrets == nil {
+		secrets, err := garden.ReadGardenSecrets(ctx, gardenClient.Cache(), c.k8sGardenCoreInformers.Seeds().Lister(), seedpkg.ComputeGardenNamespace(*shoot.Spec.SeedName))
+		if err != nil {
+			return fmt.Errorf("error reading Garden secrets: %w", err)
+		}
+		c.gardenSecrets = secrets
+	}
+
```
rfranzke, line 290: Out of curiosity, why are the secrets required for the health check? Can we potentially simply initialize the `Operation` with `nil`?

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -244,11 +245,31 @@ func (c *Controller) initializeOperation(ctx context.Context, logger *logrus.Ent
...
+
+	shootCount := len(shootList)
+	if len(shoot.Status.UID) == 0 {
+		// Do not consider shoot that is just being created.
+		shootCount -= 1
+	}
+
+	if err := garden.VerifyInternalDomainSecret(ctx, gardenClient, shootCount, gardenSecrets[common.GardenRoleInternalDomain]); err != nil {
+		return nil, err
+	}
```
rfranzke: I guess this might be quite expensive and we should avoid it. I think we should rather remove this "verification" here in the code to the GAC and a validating webhook for this internal domain secret. For now, looking at the implementation of `VerifyInternalDomainSecret`, I guess it's sufficient to set `1` instead of the real shoot count (or `0` if the UID is empty).

pkg/operation/garden/garden.go
```diff
@@ -165,27 +167,60 @@ func DomainIsDefaultDomain(domain string, defaultDomains []*Domain) *Domain {
...
+				secretsCopy = append(secretsCopy, *secret.DeepCopy())
+			}
+			return secretsCopy, nil
+		},
+		seedLister,
+		namespace,
+	)
+}
+
+type secretLister func(ctx context.Context, namespace string, selector labels.Selector) ([]corev1.Secret, error)
```
rfranzke:
```suggestion
type listSecretsFunc func(ctx context.Context, namespace string, selector labels.Selector) ([]corev1.Secret, error)
```

pkg/operation/garden/garden.go
```diff
@@ -243,13 +278,14 @@ func ReadGardenSecrets(k8sInformers kubeinformers.SharedInformerFactory, k8sGard
...
-			secretsMap[common.GardenRoleGlobalMonitoring] = monitoringSecret
+			secretsMap[common.GardenRoleGlobalMonitoring] = &monitoringSecret
 			logger.Logger.Infof("Found monitoring basic auth secret %s.", secret.Name)
 		}
 	}
 
 	// Check if an internal domain secret is required
-	seeds, err := k8sGardenCoreInformers.Core().V1beta1().Seeds().Lister().List(labels.Everything())
+	// TODO: Only seeds reconciled by Gardenlet
+	seeds, err := seedLister.List(labels.Everything())
```
rfranzke: ?

rfranzke (changes requested):

pkg/client/kubernetes/cache/cache.go
timuthy: Yes, it avoids that you refer to `for` variables within a go routine. It's the same like:

```
	for gvk, cache := range c.gvkToCache {
                gvkCopy := gvk
                cacheCopy := cache
		go func() {
			err := cacheCopy.Start(ctx)
			if err != nil {
				logger.Logger.Errorf("cache failed to start for %q: %v", gvkCopy.String(), err)
			}
		}()
```

pkg/gardenlet/controller/factory.go
timuthy: It's explained in the comment above:
`// Read Garden secrets from any seed namespace as we assume they are synced accordingly by the Gardener Controller Manager`

pkg/client/kubernetes/cache/cache.go
rfranzke: Sure, cool, makes sense. Thanks!

pkg/gardenlet/controller/factory.go
rfranzke: 👍🏻 Makes sense. Maybe you can add another sentence that this might need adaption when it can be better controlled which secrets are replicated for which seeds :)

pkg/client/kubernetes/cache/cache.go
timuthy: Unfortunately, this doesn't work for the `List` function because there we have `client.ObjectList` which is not compatible with `client.Client`. So I chose `runtime.Object` wich works for `client.Object` (from `func Get`) and `client.ObjectList` (from `func Get`)

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
timuthy: Yes, the seed controller (pkg/gardenlet/controller/seed/seed_control.go) will invalidate the Garden client when (a) a new seed is added or (b) a seed is removed because then we expect new/removed namespaces in the Garden cluster. 

> But then it must be done for all the places where the garden client is used and interacts with only matching seeds, or?

As soon as the Garden client is requested again  in one of the many places (after it was invalidated), the updated seed list (either with the new or removed seed) is considered. Not sure if you have another use-case in mind?

pkg/gardenlet/apis/config/helper/helpers.go
timuthy: > Maybe you can change the signature to already provide the []*gardencorev1beta1.Seed?

That was actually the motivation to create a new function in the first place, because if you already provided a list with seeds, then the function itself is kind of useless. I can move the whole logic to the [factory](https://github.com/gardener/gardener/blob/f95a15cbe53ea68492fa4da20998828695122bcb/pkg/gardenlet/controller/factory.go#L94) and remove this function or move it elsewhere (e.g. private function in factory)?

pkg/gardenlet/controller/shoot/shoot_control.go
timuthy: Makes sense, thank you 👍 
Edit:
> For now, looking at the implementation of VerifyInternalDomainSecret, I guess it's sufficient to set 1 instead of the real shoot count (or 0 if the UID is empty).

This would mean that the internal domain secret can be changed whenever a new Shoot is created, IIUC?

pkg/gardenlet/controller/shoot/shoot_care_control.go
timuthy: I think the secrets themselves are not required but we need it to construct the operation. Otherwise, we'll have the following error:
```
time="2021-03-12T14:56:01+01:00" level=error msg="could not initialize a new operation: unable to figure out which secret should be used for dns"
```

I'm also not too happy with it but I think using the whole operation object to perform health checks comes at a price which we can see here. It was btw. one of the reasons I wanted to invest more time to refactor our health checks but unfortunately I haven't had enough capacity yet.

pkg/client/kubernetes/cache/cache.go
rfranzke: OK, makes sense, thanks!

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
rfranzke: I think @timebertt had implemented a mechanism that allows to invalidate/refresh the client automatically so that you don't have to handle it explicitly in the "business logic" code. For example, for the `Shoot` client a hash is computed and if it changes then it's auto-refreshed, see https://github.com/gardener/gardener/blob/4c6153dceae2e461607c9a38afca88277387dda0/pkg/client/kubernetes/clientmap/internal/shoot_clientmap.go#L62-L80

The garden clientmap doesn't implement this yet, see https://github.com/gardener/gardener/blob/4c6153dceae2e461607c9a38afca88277387dda0/pkg/client/kubernetes/clientmap/internal/garden_clientmap.go#L49-L52

Maybe we can use this? Probably @timebertt can help and guide us here a little :)

pkg/gardenlet/apis/config/helper/helpers.go
rfranzke: Sure, that's also fine. I'm only worried about the Go package dependency graph, so moving this to somewhere else is totally fine.

pkg/gardenlet/controller/shoot/shoot_care_control.go
rfranzke: OK, I see... 😢 We should definitely optimize this to make it less heavy-weight and to save ourselves coding like this.
Thanks for explaining!

pkg/client/kubernetes/cache/errors.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+)
+
+// APIError checks if the given error is API related.
+func APIError(err error) bool {
```
timebertt: WDYT?
```suggestion
func IsAPIError(err error) bool {
```

pkg/gardenlet/controller/shoot/shoot_control.go
timuthy: I had a second thought about tweaking this calculation but eventually noticed that this call shouldn't be expensive at all.
We use a lister, so the objects are anyway cached, and the lister even returns the original objects from the cache (no deep copy).

rfranzke: Yeah, but we can probably still simplify this and save some code here.

pkg/client/kubernetes/cache/cache.go
```diff
@@ -0,0 +1,126 @@
...
+	synced = synced && c.fallback.WaitForCacheSync(ctx)
+	return synced
+}
+
+func (c *Aggregator) IndexField(ctx context.Context, obj client.Object, field string, extractValue client.IndexerFunc) error {
+	return c.cacheForObject(obj).IndexField(ctx, obj, field, extractValue)
+}
+
+// NewAggregator creates a new instance of an aggregated cache.
+func NewAggregator(fallback cache.Cache, gvkToCache map[schema.GroupVersionKind]cache.Cache, scheme *runtime.Scheme) *Aggregator {
```
timebertt: Nit: What's the value of exposing `Aggregator`, if it's not supposed to be instantiated or manipulated directly (unexposed members)?
Should we simply return `cache.Cache` here and make `Aggregator` unexported?

pkg/client/kubernetes/cache/cache.go
```diff
@@ -0,0 +1,126 @@
...
+		cache = c.fallback
+	}
+	return cache
+}
+
+func (c *Aggregator) Get(ctx context.Context, key client.ObjectKey, obj client.Object) error {
+	err := c.cacheForObject(obj).Get(ctx, key, obj)
+	if err != nil && !APIError(err) {
+		return NewCacheError(err)
+	}
```
timebertt: Can you elaborate why we need to distinguish between api errors and "cache errors"? 
What are the "cache errors" that could arise and that you want to handle differently?
And why don't we need such special handling in `IndexField`?

pkg/client/kubernetes/cache/cache.go
```diff
@@ -0,0 +1,126 @@
...
+	<-ctx.Done()
+
+	return nil
+}
+
+func (c *Aggregator) WaitForCacheSync(ctx context.Context) bool {
+	synced := true
+	for _, cache := range c.gvkToCache {
+		if s := cache.WaitForCacheSync(ctx); !s {
+			synced = s
```
timebertt: Can we do an early exit here?
Future `WaitForCacheSync` calls will probably also fail, if one starts to fail (e.g. because of a timeout).

pkg/client/kubernetes/cache/cache_test.go
```diff
@@ -0,0 +1,237 @@
...
+			})
+
+			go func() {
+				for count := 0; count < 2; count++ {
+					<-testChan
+				}
+				cancel()
+			}()
+
+			Expect(aggregator.Start(informerCtx)).To(Succeed())
```
timebertt: Hmm, I think this test would not terminate, if some errors occur under the hood and `Start` is not called on either of the caches. Can you improve the test to make use of ginkgo's build-in [async patterns](https://onsi.github.io/ginkgo/#asynchronous-tests)?
E.g. like this (applies a default timeout of `1s` by adding a `Done` parameter)
```suggestion
			go func() {
				defer GinkgoRecover()
				Expect(aggregator.Start(informerCtx)).To(Succeed())
			}()

			Eventually(testChan).Should(Receive())
			Eventually(testChan).Should(Receive())
			cancel()
			close(done)
```

pkg/client/kubernetes/client.go
```diff
@@ -359,3 +372,34 @@ func defaultContentTypeProtobuf(c *rest.Config) *rest.Config {
 	}
 	return &config
 }
+
+var _ client.Client = &fallbackClient{}
+
+type fallbackClient struct {
```
timebertt, line 380: Basically part 2 of the question in `cache.go`:
For which "cache errors" do we need special handling / fallback logic?

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
timebertt: > I think @timebertt had implemented a mechanism that allows to invalidate/refresh the client automatically so that you don't have to handle it explicitly in the "business logic" code. For example, for the Shoot client a hash is computed and if it changes then it's auto-refreshed, see

Yep, this is exactly how `CalculateClientSetHash` is supposed to be used: to implicitly refresh clientsets if the underlying configuration changed.

Though, I'm asking myself if this special handling for the `seedSelector` use case (i.e. gardenlet running outside of the seed cluster) is really worth the effort and maintenance complexity. This will likely cause us some headache in the future. 
WDYT not doing any special cache handling for a gardenlet responsible for multiple seeds (i.e. configuring a `seedSelector`). This use case is anyways not supposed to be used in production and has probably also other security flaws. If we just don't do those cache optimizations for the "dev scenario" it will save us from refreshing the clientsets.

timebertt (commented):
Took a look at the first commit about the caching part.
While I generally like the implementation/mechanism, I left you some questions to clarify some details.

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
rfranzke: Hm, how much code would it be? I'm not too happy with intentionally not fully supporting the configuration options we offer.

Generally, maybe we should then rather go into the direction of dropping support for this `seedSelector` feature if we anyways don't recommend it for productive setups and don't know of use-cases (other than dev). Assuming it's dropped, we can make the dev environment work by simply not providing any `seedConfig` which would mean "target all seeds".

timebertt: > Generally, maybe we should then rather go into the direction of dropping support for this seedSelector feature if we anyways don't recommend it for productive setups and don't know of use-cases (other than dev). Assuming it's dropped, we can make the dev environment work by simply not providing any seedConfig which would mean "target all seeds".

That sounds like a good idea to me.

pkg/gardenlet/controller/shoot/shoot_control.go
timuthy: Do you have an idea how? As explained in my first comment, setting the count to `1` doesn't help.

pkg/client/kubernetes/cache/cache.go
timuthy: The Multinamespace Cache returns an untyped error ([ref](https://github.com/kubernetes-sigs/controller-runtime/blob/4462fd4ebb88171af826622c1d6b73383aaf2cdf/pkg/cache/multi_namespace_cache.go#L132)) in case of a Get/List call for unknown namespaces. As this error is untyped, the idea was to ask for API errors and declare the rest as cache errors.

pkg/client/kubernetes/client.go
timuthy: Fallbacks are necessary so that you can use a cached client with a [Multinamespace Cache](https://github.com/kubernetes-sigs/controller-runtime/blob/4462fd4ebb88171af826622c1d6b73383aaf2cdf/pkg/cache/multi_namespace_cache.go#L65) but only for namespaces that are allowed for Gardenlet. There are other namespaces in the Garden cluster e.g., `garden`, which the Gardenlet is permitted to only GET specific resources, but is never allowed to LIST/WATCH the entire namespace. Hence, we need to prevent that the cached Runtime Client automatically starts a watch for these "restricted" namespaces.

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
timuthy: > Assuming it's dropped, we can make the dev environment work by simply not providing any seedConfig which would mean "target all seeds".

Hm, especially for dev uses cases one adds/removes seeds and therefore you need to consider newly added or removed seed namespaces in the Garden cluster. Here it doesn't make a difference if you target a subset of seeds or all seeds.

pkg/gardenlet/controller/shoot/shoot_control.go
rfranzke: Sorry, I missed that. Yeah, you're right... OK, then let's keep it like this and rather move this code to GAC as planned. 👍🏻

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
rfranzke: Hm, that's true. @timebertt would you still vote for not supporting it? Personally, I don't feel too comfortable with this approach...

pkg/client/kubernetes/cache/cache.go
timebertt: Makes sense, thanks for clarifying.
It would nice to add an error type upstream, but this is obviously out of scope for this PR.

Edit:
Can you add a short comment with this explanation to the `CacheError` type so that it's obvious what the type is supposed to be used for?

pkg/client/kubernetes/client.go
timebertt: Yep, makes sense.
Now I fully understood the rationale :)

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
timebertt: > Here it doesn't make a difference if you target a subset of seeds or all seeds.

That's true.

What I actually meant with 

> WDYT not doing any special cache handling for a gardenlet responsible for multiple seeds (i.e. configuring a seedSelector). This use case is anyways not supposed to be used in production and has probably also other security flaws. If we just don't do those cache optimizations for the "dev scenario" it will save us from refreshing the clientsets.

is to basically don't use any caching at all for secrets if the gardenlet is responsible for mulitple seeds. We could simply always read directly from the garden cluster and thereby avoid clientset/cache invalidating/rebuilding.

I'm not insisting on it, just trying to figure out, which way we are the least likely to run into bugs in the future because of added complexity.
WDYT?

rfranzke: Ah, interesting, I didn't get it that way. But yeah, I think that would be totally fine. However, I cannot judge what this means with respect to the code complexity. @timuthy :)

timebertt:
/hold

pkg/client/kubernetes/clientmap/internal/garden_clientmap.go
timebertt: I recognized another problem with the approach to dynamically invalidate the garden client during runtime.
Currently all of our controllers assume, that the garden clientset is never invalidated (the watches will not be reestablished when they are stopped). Meaning, if we would either need to adapt all controllers to reestablish watches or restart the gardenlet if the garden client should be invalidated...
Neither of this seems like a good option to me.

timuthy: Valid points @timebertt and @rfranzke. I originally wanted to fully support cached secrets for dev/multiple seed use cases, because in production with a Gardenlet per seed, caching will become the default option and if we now workaround it during dev, bugs for common setups might remain unnoticed for a longer time. OTOH, if we don't support invalidating the Gardenclient, we should probably go with @timebertt's suggestion and skip coaching, which is a tradeoff but ATM I don't have any further ideas either.

timebertt: > because in production with a Gardenlet per seed, caching will become the default option and if we now workaround it during dev, bugs for common setups might remain unnoticed for a longer time

Yeah agree, I was also worried about this.
Unfortunately I don't have a better idea on how to deal with this 😕

pkg/gardenlet/controller/shoot/shoot_control.go
rfranzke: FYI: I have opened https://github.com/gardener/gardener/pull/3756 which implements the internal domain validation in the GAC as new webhook handler.

pkg/client/kubernetes/cache/cache.go
timuthy: I added a comment to `pkg/client/kubernetes/client.go`

timuthy:
Thanks for your reviews @rfranzke and @timebertt. I addressed all your comments in c56a187a8e780a15265bac0c251b7fca13f37560.

rfranzke:
```
>> make revendor
 M go.sum
```

pkg/client/kubernetes/runtime_client.go
```diff
@@ -98,6 +101,37 @@ func setClientOptionsDefaults(config *rest.Config, options *client.Options) erro
...
+
+		fallbackCache, err := newCache(config, options)
+		if err != nil {
+			return nil, err
+		}
+
+		gvkToCache := make(map[schema.GroupVersionKind]cache.Cache)
+		for object, fn := range typeToNewCache {
+			gvks, _, err := scheme.ObjectKinds(object)
+			if err != nil || len(gvks) != 1 {
```
timebertt: Nit: if `len(gvks) > 1` `err` is probably `nil`, so we're returning `nil` here in that case.
Can you use `apiutil.GVKForObject` (which handles this case already) instead of `scheme.ObjectKinds`?

pkg/gardenlet/controller/factory.go
```diff
@@ -156,7 +165,7 @@ func (f *GardenletControllerFactory) Run(ctx context.Context) error {
 	var (
 		controllerInstallationController = controllerinstallationcontroller.NewController(f.clientMap, f.k8sGardenCoreInformers, f.cfg, f.recorder, gardenNamespace, f.gardenClusterIdentity)
 		seedController                   = seedcontroller.NewSeedController(f.clientMap, f.k8sGardenCoreInformers, f.k8sInformers, f.healthManager, secrets, imageVector, componentImageVectors, f.identity, f.cfg, f.recorder)
```
timebertt: Here we still use the statically fetched garden secrets, mainly for the monitoring configuration IISIC.
Are there plans to also dynamically fetch those or is this not needed?

timebertt (commented):
Very nicely done!
One nit and one question from my side left.

pkg/gardenlet/controller/factory.go
timuthy: Nice catch! I also adapted this handling in 7572fc06111be0e13a9f9912692d69c35a73ccff and improved logging slightly when reading the Garden secrets.

timebertt (approved):
Thank you very much, love it!
/lgtm

rfranzke (approved):
/lgtm

PR #3707, opened 2021-03-12
Move constants away from common package

rfranzke:

/area dev-productivity
/kind cleanup
/priority normal

This PR is a good step towards eliminating the common package (or, at least, to better declare its purpose). It moves some constants to better places.
Also, constants for the image names in the image vector are now auto-generated to prevent double maintenance and ambiguous naming.

/squash

**Release note**:

```other operator
NONE
```


hack/generate-imagename-constants.sh
```diff
@@ -0,0 +1,51 @@
...
+
+const ("
+
+for image_name in $(yaml2json < "$PATH_CHARTS/images.yaml" | jq -r '[.images[].name] | unique | .[]'); do
+  variable_name="$(camelCase "$image_name")"
+
+  out="
+$out
+	// ImageName$variable_name is a constant for an image in the image vector with name '$image_name'.
+	ImageName$variable_name = \"$image_name\""
```
timebertt, line 42: Very nice, I like it 🚀

hack/generate-imagename-constants.sh
```diff
@@ -0,0 +1,51 @@
...
+done
+
+out="
+$out
+)
+"
+
+echo "$out" > "$PATH_CHARTS/images.go"
+go fmt "$PATH_CHARTS/images.go"
```
timebertt: Use `goimports` instead to have it consistent with `make format`?

timebertt (dismissed):
Very nice PR, thanks for cleaning up!
/lgtm

hack/generate-imagename-constants.sh
rfranzke: Good point, done

timebertt (approved):
Thanks!
/lgtm

PR #3711, opened 2021-03-15
remove workaround for konnectivity image and introduce new labels

schrodit:

/area delivery
/kind cleanup
/priority 3

Removes the workaround for the konnectivity-server image and replaces the implicit image tagging with explicit labels.

The `component-cli iv add` flags are still kept to ensure backwards compatibility with the extensions that also use the ci script.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -21,14 +21,29 @@ images:
   sourceRepository: github.com/gardener/etcd-druid
   repository: eu.gcr.io/gardener-project/gardener/etcd-druid
   tag: "v0.4.1"
+  labels:
+    # get the version from the gardener/etcd-druid component
+  - name: imagevector.gardener.cloud/component-reference
+    value:
+      componentName: github.com/gardener/etcd-druid
 - name: gardener-resource-manager
   sourceRepository: github.com/gardener/gardener-resource-manager
   repository: eu.gcr.io/gardener-project/gardener/gardener-resource-manager
   tag: "v0.22.0"
+  labels:
+    # get the version from the gardener/gardener-resource-manager component
+  - name: imagevector.gardener.cloud/component-reference
+    value:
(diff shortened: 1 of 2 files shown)


rfranzke (commented):
Why is all of this needed? Can't it be inferred automatically (like it was done before)? I really don't want to be bothered too much with all these component descriptor details, so please hide as much as possible to reduce the cognitive overhead for everybody.

schrodit:
> Why is all of this needed? Can't it be inferred automatically (like it was done before)? I really don't want to be bothered too much with all these component descriptor details, so please hide as much as possible to reduce the cognitive overhead for everybody.

sure we can leave it as it was before and only keep the specific handling for the konnectivity-server.
My initial idea was that issues like we saw with the konnectivity-server image will not occur anymore in the future.
We could also try to enhance the validation in the PR so that such issues are discovered before it is merged into master?

cc @AndreasBurger @ccwienk

rfranzke:
Yes, the `charts/images.yaml` should not contain anything related to the component descriptor or CI/CD artefacts in general. The purpose of the `.ci/component_descriptor` script is to translate this into whatever the CI/CD needs.

> We could also try to enhance the validation in the PR so that such issues are discovered before it is merged into master?

I'm not aware of the details but, generally, this would be highly appreciated.

schrodit:
ok then I will close this PR in favor of a better validation.

> The purpose of the .ci/component_descriptor script is to translate this into whatever the CI/CD needs.

All exceptions should then also reside in the component-descriptor script and will be maintained there?

rfranzke:
Okay, thank you very much.

> All exceptions should then also reside in the component-descriptor script and will be maintained there?

For now, I would think yes, that's how we did it in the past and it worked quite okay.

PR #3713, opened 2021-03-15
[GEP-13] Add ManagedSeedSet controller

stoyanr:

/area control-plane
/kind enhancement
/priority normal

Adds a controller for `ManagedSeedSet` resources in `gardener-controller-manager`.

Only creation and deletion of replica shoots and managed seeds is performed currently by this controller. Updates, revision history, etc. are not yet covered.

Fixes #

**Release note**:

```improvement operator
NONE
```


rfranzke:
/assign
@stoyanr OK, sure. I won't make it today (most likely), though.

timebertt:
/assign
Is this `/ready`?

Edit:
Ah, I see there are still some `TODO`s in code.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
```diff
@@ -0,0 +1,666 @@
...
+		seedsByName[seed.Name] = &seeds.Items[i]
+	}
+
+	// Initialize replicas
+	var replicas []*replica
+	for i, shoot := range shoots.Items {
+		// Get shoots scheduled onto this seed
+		var scheduledShoots []*gardencorev1beta1.Shoot
+		if seedsByName[shoot.Name] != nil {
+			shoots, err := a.gardenClient.GardenCore().CoreV1beta1().Shoots(metav1.NamespaceAll).List(ctx, metav1.ListOptions{
```
timebertt: I think using a meta-data only list would be better here, as you don't need the spec/status of shoots.

stoyanr:
Here is a list of things I still have in mind for this PR (perhaps some of them could be done also in a different PR):

1. Functionality:
* Detect cases where a replica status / health hasn't changed for a certain (configurable) period of time (e.g. 1 hour), and return w/o requeue, to avoid waiting on a replica that is never going to be "healthy" due to external issues. This would require adding new fields in status to keep track of such replicas.
* Introduce a condition for aborted reconciliation, due to one of the reasons mentioned above, e.g. failed shoot, and over max number of retries, or replica status / health hasn't changed for a certain period of time
* Instead of status fields such as `replicas`, `readyReplicas`, etc. , introduce a `replicas` field that contains a list of all replicas with properties such as name, status, seed readiness, shoot health, etc. The reason is that there is no simple way to express "readiness" for our replicas and so a fuller picture makes more sense.
* Controller configuration (max number of retries, etc.)

2. Refactoring for testability:
* `Replica` as an interface implemented by the current internal `replica` type
* `ReplicaGetter` as an interface and a standalone internal type implementing it instead of the current `getReplicas` method of the actuator.

These 2 internal types could then be tested independently of the actuator and also mocked to make the actuator testable. As it stands, it's already too complex to be meaningfully tested by unit tests.

3. Unit tests themselves

If you have comments / feedback on some of the above, please let me know.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
```diff
@@ -0,0 +1,666 @@
...
+	return fmt.Sprintf("%s-%d", set.Name, ordinal)
+}
+
+// parentNameAndOrdinalRegex is a regular expression that extracts the parent name and ordinal from the name of a replica object
+var parentNameAndOrdinalRegex = regexp.MustCompile("(.*)-([0-9]+)$")
+
+// getParentNameAndOrdinal gets the parent name and ordinal from the given replica object name.
+// If the object was not created by a ManagedSeedSet, its parent name is considered to be empty string,
+// and its ordinal is considered to be -1.
+func getParentNameAndOrdinal(name string) (string, int) {
```
timebertt: I don't see any usage of the first return parameter, yet.
For what do you think, you will need it?
In general, the "parent name" should be determined by looking at the `ownerReferences`.

stoyanr:
@timebertt No it's not ready so I keep the "Draft", but I would appreciate comments on the overall approach as mentioned, especially if you have strong concerns about some parts. It's also already testable if you want to give it a try.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
```diff
@@ -0,0 +1,666 @@
...
+	return shootHealthStatus(r.shoot)
+}
+
+func (r *replica) isReady() bool {
+	return r.getStatus() == statusManagedSeedRegistered && r.isSeedReady() && r.getShootHealthStatus() == operationshoot.StatusHealthy
+}
+
+func (r *replica) isDeletable() bool {
+	return len(r.scheduledShoots) == 0 &&
+		!kutil.HasMetaDataAnnotation(r.shoot, seedmanagementv1alpha1constants.AnnotationProtectFromDeletion, "true")
```
timebertt: Why did you chose to put the protection annotation on the shoot?
Without much thought, it seems to make more sense to have it on the `ManagedSeed`, as the `ManagedSeedSet` manages individual `ManagedSeed`s.

timebertt (commented):
Ok, I started to look into the PR, but I agree it's quite complex already.
I started to add a few inline comments so far.

TBH though, I can't judge the overall approach, yet.
What would help me is to have a "pre-review" session in which you can give a rough overview of the code, the most important lines of code and the general mechanism of the new controller.
We've done this before and I think, it helped the reviewers to faster get up to speed in the review process.
I see you also have some open questions about some aspects of the approach, we could also discuss them in person instead of here in GH.

Would you mind setting up a session with me and @rfranzke (and maybe @danielfoehrKn if he also wants to join)?

stoyanr:
> Would you mind setting up a session with me and @rfranzke (and maybe @danielfoehrKn if he also wants to join)?

Yes, makes sense, let's do this. You are right it's quite complex and hard to understand the overall approach without a proper explanation.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: Because the shoot is the only replica object guaranteed to exist at all times. If I want to protect a half-created replica from deletion, the only object that is guaranteed to be there is the shoot.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: No, I don't think this is needed, so I removed it.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: Right now, we are only interested if there are scheduled shoots or not, so this could be simplified even further. However, perhaps we would like at some point to include the health of the scheduled shoots in the calculation whether to advance to the next replica, as also mentioned in the GEP. Right now this is not done because it's only interesting during updates (a newly created replica has no scheduled shoots). Having this in mind, perhaps it should stay as it is?

stoyanr: As we agreed that we won't take into account scheduled shoots when determining replica readiness for now, I will do as suggested.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: We agreed to support this annotation also on managed seeds (in addition to shoots), I will add this.

stoyanr:
With @timebertt, we had a lengthy discussion and agreed on the following further improvements to be done in this PR:

* Instead of constant requeueing at a constant interval, enqueue sets if one of their controlled objects (shoots, managed seeds, seeds) changes. To avoid code duplication, a "mini-framework" for this should be introduced. To avoid too many reconciles, only relevant changes to the currently "pending replica" should lead to requeues. To be able to recognize the currently pending replica and the reason why it's pending a new `PendingReplica` field should be introduced in `status`.
* Keep the existing mechanism for retrying failed shoots with configurable max retries. In addition, introduce a condition to be set when max retries have been reached and the set is "stuck" on a failed shoot. **Note:** After introducing the new `PendingReplica` field (with `Name`, `Reason`, and `Since`), I realized that such a condition is no longer needed, since it would contain mostly the same information, so I won't introduce it for now.
* List scheduled shoots as a meta-data only list (see comment above).
* Support `protect-from-deletion` annotation on managed seeds as well as shoots (see comment above).
* Record events in addition to logging messages for important actions on replicas.
* Controller configuration (max number of retries, etc.)
* Unit tests

In addition, the following improvements should come in future PRs:

* Enhance `ManagedSeed` controller to avoid constant requeueing during deletion, taking advantage of the "mini-framework" introduced with this PR. **Note:** We didn't discuss this, but there are strong reasons for such a change, since the current deletion mechanism relies on returning errors (retried with backoff) until certain conditions are fulfilled. This is ugly for multiple reasons: it leads to too many reconciles, (bogus) errors and log messages about them, and since the deletion of the `garden` namespace takes some time, it's often the case that due to the backoff, the controller is able to finish the deletion only a few minutes after that.
* Add `ManagedSeedSet` admission plugin for update / deletion for checking whether the replicas field is valid against the number of deletable replicas.
* Enhance ManagedSeedSet controller with support for updates and rollbacks

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: @timebertt, sorry for the stupid question, but how exactly can I do a "meta-data only list"? I can't find this in APIs or generated code, and google is not helpful since I can't compose a query that doesn't return a ton of irrelevant results.

rfranzke: https://github.com/gardener/gardener/blob/0f459405b52a4f94ac2960f9f46098aa33d73dd1/pkg/controllermanager/controller/project/project_control_delete.go#L68-L72

stoyanr: Thanks. I tried this approach but there is an issue: reading via the `client.Client` using a field selector fails with an error `Index with name field:spec.seedName does not exist`. Reading via the `gardencoreclientset.Interface` works, but there it's not possible (or is quite unobvious) to get a `PartialObjectMetadataList` instead of a `ShootList`. Do you have any ideas?
**Update:** Using `APIReader()` instead of `Client()` does work, is this the right way to go?

timebertt: > reading via the client.Client using a field selector fails with an error Index with name field:spec.seedName does not exist. Reading via the

This error indicates, that the client's cache is not indexed. When using cached clients, one has to explicitly add an indexer to the cache for the specific field that shall be used in a field selector. (I currently can't find any reference usage)

But using the API reader is simpler in this case and shouldn't harm here, as you will anyways only need to read at most one shoot (`client.Limit(1)`).

stoyanr: OK, great, it's done then.

stoyanr:
/cc @rfranzke @timebertt @danielfoehrKn I believe I did all improvements listed above, except the unit tests (and perhaps updates to documentation / examples). So except unit tests, this PR is ready for review now. It also seems to work quite well, so you could give it a try. Here is a sample resource for testing:

```yaml
apiVersion: seedmanagement.gardener.cloud/v1alpha1
kind: ManagedSeedSet
metadata:
  name: <name>
  namespace: garden
spec:
  replicas: 1
  selector:
    matchLabels:
      name: <name>
  template:
    metadata:
      labels:
        name: <name>
    spec:
      gardenlet:
        config:
          apiVersion: gardenlet.config.gardener.cloud/v1alpha1
          kind: GardenletConfiguration
          seedConfig:
            metadata:
              labels:
                name: <name>
  shootTemplate:
    metadata:
      labels:
        name: <name>
    spec:
      cloudProfileName: gcp
      dns:
        # "replica-name" will be replaced by the actual replica name
        domain: replica-name.garden.shoot.dev.k8s-hana.ondemand.com
      kubernetes:
        version: 1.18.14
      networking:
        type: calico
        pods: 10.225.128.0/17
        nodes: 10.250.0.0/16
        services: 10.225.0.0/17
      provider:
        type: gcp
        controlPlaneConfig:
          apiVersion: gcp.provider.extensions.gardener.cloud/v1alpha1
          kind: ControlPlaneConfig
          zone: europe-west1-c
        infrastructureConfig:
          apiVersion: gcp.provider.extensions.gardener.cloud/v1alpha1
          kind: InfrastructureConfig
          networks:
            workers: 10.250.0.0/16
        workers:
          - name: worker-test
            machine:
              type: n1-standard-4
            minimum: 1
            maximum: 2
            volume:
              type: pd-standard
              size: 50Gi
            zones:
              - europe-west1-c
      region: europe-west1
      secretBindingName: shoot-operator-gcp
      seedName: gcp4
```

stoyanr:
Started working on unit tests, added some in a new commit to allow for more focused reviews.

timebertt:
Sorry for the delay @stoyanr. Haven't made it to a complete review yet.
Have blocked some time for it tomorrow.

stoyanr:
I finished working on unit tests, so this PR is now no longer Draft.

pkg/controllermanager/controller/managedseedset/managedseedset_replica.go
```diff
@@ -0,0 +1,473 @@
...
+		if err != nil {
+			return nil, err
+		}
+		replacePlaceholdersInSeedSpec(&gardenletConfig.SeedConfig.Spec, name)
+	}
+
+	return managedSeed, nil
+}
+
+const placeholder = "replica-name"
```
timebertt, line 401: Bringing this up again here for more opinions:
I would not use such a placeholder in the `ManagedSeedTemplate` and rather let the controller prepend the shoot name to the domain when creating the shoot from the template. This is similar to what the `StatefulSet` controller does for pvc templates: it takes the name specified in the template and appends the pod name to it.

This approach would work at least for the shoot domain, but wouldn't work for the ingress domain (it typically has the pattern `ingress.<seed-name>.seed.<gardener-installation-domain>`). Instead of replacing the placeholder in the template, we could deny specifying the ingress domain and let the controller always assign it according to this (or a different) pattern.

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
```diff
@@ -137,6 +137,42 @@ type ManagedSeedSetStatus struct {
...
+	// ShootReconcilingReason indicates that the replica's shoot is reconciling.
+	ShootReconcilingReason PendingReplicaReason = "ShootReconciling"
+	// ShootDeletingReason indicates that the replica's shoot is deleting.
+	ShootDeletingReason PendingReplicaReason = "ShootDeleting"
+	// ShootReconcileFailedReason indicates that the reconciliation of this replica's shoot has failed.
+	ShootReconcileFailedReason PendingReplicaReason = "ShootReconcileFailed"
+	// ShootDeleteFailedReason indicates that the deletion of tis replica's shoot has failed.
+	ShootDeleteFailedReason PendingReplicaReason = "ShootDeleteFailed"
+	// ManagedSeedPreparingReason indicates that the replica's managed seed is preparing.
+	ManagedSeedPreparingReason PendingReplicaReason = "ManagedSeedPreparing"
```
timebertt, line 159: IISIC, `ManagedSeedPreparing` is used, when the shoot has been successfully reconciled and the controller has just created the `ManagedSeed`. 
To me, `ManagedSeedReconciling` would be a bit more intuitive. WDYT?

pkg/apis/seedmanagement/v1alpha1/constants/types_constants.go
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package constants
+
+const (
+	// AnnotationProtectFromDeletion is a constant for an annotation on a managed seed replica shoot to protect it from deletion.
```
timebertt:
```suggestion
	// AnnotationProtectFromDeletion is a constant for an annotation on a replica of a ManagedSeedSet (either ManagedSeed or Shoot) to protect it from deletion.
```

pkg/apis/seedmanagement/v1alpha1/constants/types_constants.go
```diff
@@ -0,0 +1,23 @@
...
+	AnnotationProtectFromDeletion = "seedmanagement.gardener.cloud/protect-from-deletion"
+
+	// AnnotationRetries is a constant for an annotation on a managed seed replica shoot to count the number of retries.
+	AnnotationRetries = "seedmanagement.gardener.cloud/retries"
```
timebertt: I'm not convinced yet, that the controller should store this information in an annotation. I think, generally annotations are a good mechanism for users to give hints to controllers, for example like the `protect-from-deletion` annotation. 
But I think, as long as its pure controller state and no user interaction is intended, the controller should store the state in some proper status field of the resource it controls.

WDYT about storing it in the `PendingReplica` struct?
Do you have a strong reason for storing this in an annotation, that I'm missing?

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -239,6 +242,19 @@ type ShootReferenceControllerConfiguration struct {
...
 
+// ManagedSeedSetControllerConfiguration defines the configuration of the
+// ManagedSeedSet controller.
+type ManagedSeedSetControllerConfiguration struct {
+	// ConcurrentSyncs is the number of workers used for the controller to work on
+	// events.
+	ConcurrentSyncs int `json:"concurrentSyncs"`
+	// MaxShootRetries is the maximum number of times to retry failed shoots before giving up. Defaults to 3.
+	// +optional
+	MaxShootRetries *int `json:"maxShootRetries,omitempty"`
```
timebertt, line 253: Can you make this field configurable in `charts/gardener/controlplane/charts/runtime/templates/controller-manager/configmap-componentconfig.yaml` and add it to the example in `example/20-componentconfig-gardener-controller-manager.yaml`?

pkg/utils/kubernetes/kubernetes.go
```diff
@@ -77,6 +78,67 @@ func HasMetaDataAnnotation(meta metav1.Object, key, value string) bool {
...
 
+// SetAnnotationsAndUpdate sets the given annotations on the given object and updates it, taking care of conflicts.
+func SetAnnotationsAndUpdate(ctx context.Context, c client.Client, obj client.Object, keyValues ...string) error {
+	annotations := keyValuesToAnnotations(keyValues...)
+	if !hasAnnotations(obj, annotations) {
+		return clientretry.RetryOnConflict(clientretry.DefaultBackoff, func() error {
+			setAnnotations(obj, annotations)
+			return c.Update(ctx, obj)
+		})
+	}
```
timebertt: WDYT about rather using patches instead of updates in order to prevent conflicts in the first place?

pkg/utils/kubernetes/kubernetes.go
```diff
@@ -77,6 +78,67 @@ func HasMetaDataAnnotation(meta metav1.Object, key, value string) bool {
...
+	if !doesNotHaveAnnotationKeys(obj, keys...) {
+		return clientretry.RetryOnConflict(clientretry.DefaultBackoff, func() error {
+			deleteAnnotations(obj, keys...)
+			return c.Update(ctx, obj)
+		})
+	}
+	return nil
+}
+
+func keyValuesToAnnotations(keyValues ...string) map[string]string {
```
timebertt: This func will silently ignore the last key, if the value is missing (i.e. odd number of args).
I generally don't like such functions, that will silently swallow programming level mistakes, that can be avoided by using proper parameter types, like a `map` in this case.

I propose to either use a `key string` and a `value string` param or `map`, instead.

pkg/controllermanager/controller/managedseedset/managedseedset.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/gardener/gardener/pkg/controllermanager/apis/config"
+	"github.com/gardener/gardener/pkg/controllerutils"
+	"github.com/gardener/gardener/pkg/gardenlet"
+	kutils "github.com/gardener/gardener/pkg/utils/kubernetes"
+)
+
+// Controller controls ManagedSeedSets.
+type Controller struct {
+	gardenClient client.Client
+	config       *config.ControllerManagerConfiguration
```
timebertt: Nit: unused member.

pkg/controllermanager/controller/managedseedset/managedseedset.go
```diff
@@ -0,0 +1,224 @@
...
+	c.managedSeedSetInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.managedSeedSetAdd,
+		UpdateFunc: c.managedSeedSetUpdate,
+		DeleteFunc: c.managedSeedSetDelete,
+	})
+
+	// Initialize scheme
+	scheme := runtime.NewScheme()
+	utilruntime.Must(gardencorev1beta1.AddToScheme(scheme))
+	utilruntime.Must(seedmanagementv1alpha1.AddToScheme(scheme))
```
timebertt: You can ruse `kubernetes.GardenScheme` here if you want.

pkg/utils/kubernetes/eventhandler.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+	newObj, ok := new.(client.Object)
+	if !ok {
+		return
+	}
+
+	// Check resource versions, as we may receive update events for all object of certain kind
+	if newObj.GetResourceVersion() == oldObj.GetResourceVersion() {
+		return
+	}
```
timebertt, line 147: Can you elaborate? I don't understand this check.

pkg/utils/kubernetes/eventhandler.go
```diff
@@ -0,0 +1,328 @@
...
+
+	// If this is the final controller in the chain, return it, otherwise move up the chain
+	if index == len(h.ControllerTypes)-1 {
+		return controller
+	}
+	return h.getControllerOf(controller, index+1)
+}
+
+func (h *ControlledResourceEventHandler) logEnqueue(controller, obj client.Object, eventType string) {
+	h.Logger.Infof("Enqueueing %s %s due to %s event on %s %s", h.getLastControllerKind(), ObjectName(controller),
```
timebertt:
```suggestion
	h.Logger.Debugf("Enqueueing %s %s due to %s event on %s %s", h.getLastControllerKind(), ObjectName(controller),
```

pkg/utils/kubernetes/eventhandler.go
```diff
@@ -0,0 +1,328 @@
...
+// ControlledResourceEventHandler is an implementation of cache.ResourceEventHandler that enqueues controllers upon
+// controlled resource events, if a predicate check is successful.
+type ControlledResourceEventHandler struct {
+	// ControllerTypes is a list of controller types. If multiple controller types are specified,
+	// there is a chain of controllers. Only the top-most controller (with the last listed type) is eventually enqueued.
+	ControllerTypes []ControllerType
+	// Ctx is the context used to get controller objects.
+	Ctx context.Context
+	// Client is the client used to get controller objects.
+	Client client.Client
```
timebertt: We should use a `client.Reader` here, as it's generally desirable to pass a `cache.Cache` as the used reader to ensure we're not making too many API calls.

pkg/controllermanager/controller/managedseedset/managedseedset.go
```diff
@@ -0,0 +1,224 @@
...
+
+	// Add event handler for controlled shoots
+	c.shootInformer.AddEventHandler(&kutils.ControlledResourceEventHandler{
+		ControllerTypes: []kutils.ControllerType{
+			{Type: &seedmanagementv1alpha1.ManagedSeedSet{}},
+		},
+		Ctx:                        ctx,
+		Client:                     c.gardenClient,
```
timebertt: Please pass a the garden client's cache here to avoid making too many API calls in the predicates.

pkg/controllermanager/controller/managedseedset/managedseedset_predicates.go
```diff
@@ -0,0 +1,96 @@
...
+	shoot, ok := obj.(*gardencorev1beta1.Shoot)
+	if !ok {
+		return false
+	}
+	set, ok := controller.(*seedmanagementv1alpha1.ManagedSeedSet)
+	if !ok {
+		return false
+	}
+
+	if set.Status.PendingReplica == nil || set.Status.PendingReplica.Name != shoot.Name {
```
timebertt, line 52: What if a Shoot's health status is changing, but the Set is not waiting for this particular Shoot?
I think, we should still enqueue and update the Set's status (in particular the `readReplicas` field).

pkg/controllermanager/controller/managedseedset/managedseedset_controller.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	key, err := cache.MetaNamespaceKeyFunc(newObj)
+	if err != nil {
+		return
+	}
+
+	c.managedSeedSetQueue.Add(key)
+}
+
+func (c *Controller) managedSeedSetDelete(obj interface{}) {
```
timebertt, line 52: Why do we need to enqueue the set when it already has been deleted?

pkg/controllermanager/controller/managedseedset/managedseedset_reconciler.go
```diff
@@ -0,0 +1,160 @@
...
+		return reconcile.Result{}, fmt.Errorf("could not reconcile ManagedSeedSet %s deletion: %w", kutil.ObjectName(set), err)
+	}
+	r.getLogger(set).Debug("Deletion reconciled")
+
+	// If there is no pending replica, remove gardener finalizer
+	if status.PendingReplica == nil {
+		r.getLogger(set).Debug("Removing finalizer")
+		if err := controllerutils.PatchRemoveFinalizers(ctx, r.gardenClient.Client(), set, gardencorev1beta1.GardenerName); err != nil {
+			return reconcile.Result{}, fmt.Errorf("could not remove gardener finalizer: %w", err)
+		}
```
timebertt, line 137: We should return without requeueing after successfully removing the finalizer
```suggestion
		}
		return reconcile.Result{}, nil
```

pkg/controllermanager/controller/managedseedset/managedseedset_replicagetter.go
```diff
@@ -0,0 +1,108 @@
...
+	shoots := &gardencorev1beta1.ShootList{}
+	if err := rg.client.Client().List(ctx, shoots, client.InNamespace(set.Namespace), client.MatchingLabelsSelector{Selector: selector}); err != nil {
+		return nil, err
+	}
+	managedSeeds := &seedmanagementv1alpha1.ManagedSeedList{}
+	if err := rg.client.Client().List(ctx, managedSeeds, client.InNamespace(set.Namespace), client.MatchingLabelsSelector{Selector: selector}); err != nil {
+		return nil, err
+	}
+	seeds := &gardencorev1beta1.SeedList{}
+	if err := rg.client.Client().List(ctx, seeds, client.MatchingLabelsSelector{Selector: selector}); err != nil {
```
timebertt, line 70: IISIC, we don't enforce by validate, that the set's selector also matches the `SeedConfig`/`SeedTemplate`.
Is this correct? If so, should we enforce it?

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
```diff
@@ -0,0 +1,345 @@
...
+func replicaIsReady(r Replica) bool {
+	return r.GetStatus() == StatusManagedSeedRegistered && r.IsSeedReady() && r.GetShootHealthStatus() == operationshoot.StatusHealthy
+}
+
+func replicaDebugString(r Replica) string {
+	return fmt.Sprintf("%s:%s,%t,%s", r.GetFullName(), r.GetStatus().String(), r.IsSeedReady(), r.GetShootHealthStatus())
+}
+
+func replicaManagedSeedExists(status ReplicaStatus) bool {
+	return status == StatusManagedSeedPreparing || status == StatusManagedSeedDeleting || status == StatusManagedSeedRegistered
```
timebertt: If `StatusUnknown` would be the lowest status (`==0`), you could just do `status >= StatusManagedSeedPreparing` here.

timebertt (changes requested):
Thanks a lot @stoyanr for this great PR, well done!

I've made it through all of the production code and left you some suggestions. I think, most of them are rather minor, in general it already looks quite good.
As always, if things are unclear or need additional discussion, let's not hesitate to discuss in person.

Will play around with it and look at the tests in the next few days.

pkg/controllermanager/controller/managedseedset/managedseedset_replica.go
stoyanr: Honestly, I don't see why we should introduce such an artificial limitation when we could just use a placeholder for the replica name as already implemented. Not to mention that this solution is also easier to implement and maintain. What's actually wrong with it so you insist on changing it?

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
stoyanr: I started with such naming but then reconsidered. Unlike shoots, the reconciliation of managed seeds is rather fast and they may be reconciled multiple times before they reach the `Registered` state. I felt a different verb from "reconcile" should be used since at any given moment, the managed seed is not really reconciling but rather waiting for a condition to be fulfilled.

pkg/apis/seedmanagement/v1alpha1/constants/types_constants.go
stoyanr: OK

pkg/apis/seedmanagement/v1alpha1/constants/types_constants.go
stoyanr: It was introduced before `PendingReplica` and so there was hardly an alternative. I agree it should rather be moved to `PendingReplica`.

pkg/controllermanager/apis/config/v1alpha1/types.go
stoyanr: OK

pkg/utils/kubernetes/kubernetes.go
stoyanr: Hm, ok, I implemented it like this because I was looking to be consistent with already existing, similar functions. I will give it a try.

pkg/utils/kubernetes/kubernetes.go
stoyanr: The API is like this on purpose to allow for convenient passing of multiple key / value parameters, something not possible with any of the solutions you propose. 

However, after I remove the `retries` annotation the need for setting multiple annotations in this case goes away, so I will simply remove this function.

pkg/controllermanager/controller/managedseedset/managedseedset.go
stoyanr: Will remove.

pkg/controllermanager/controller/managedseedset/managedseedset.go
stoyanr: Great, thanks for the tip. I looked for an existing scheme I could use but didn't find this one.

pkg/utils/kubernetes/eventhandler.go
stoyanr: I am also not really sure, but: https://github.com/kubernetes/kubernetes/blob/54449be03127f883e52df48b95adef36925ae2cf/pkg/controller/statefulset/stateful_set.go#L198-L202

pkg/utils/kubernetes/eventhandler.go
stoyanr: OK

pkg/utils/kubernetes/eventhandler.go
stoyanr: Wouldn't `client.Client` use a cache as well (if got via `Client()`) ? What if the object we are interested in is not already in the cache (or out of date), wouldn't then a cache-based reader miss the object or fetch an old state?

pkg/controllermanager/controller/managedseedset/managedseedset.go
stoyanr: See my question above, I agree using a cache is essential, but perhaps I don't quite get the difference between the current implementation and what you propose.

pkg/controllermanager/controller/managedseedset/managedseedset_controller.go
stoyanr: Hm, just in case, perhaps, to make sure the final state of the resource is properly taken care of? 

The same question would be relevant to all other controllers. All (or most, if there are exception I couldn't find them) currently existing controllers have such a function that similarly enqueues the resource.

pkg/controllermanager/controller/managedseedset/managedseedset_predicates.go
stoyanr: From the perspective of reconciling scaling in or out (and I suspect updates also, but I can't be sure before I implement it), we should not enqueue such a shoot. Let's consider the different cases:

* If there are no pending replicas, the scaling in / out has finished, and until the next scaling in / out there is no reason for the set to be interested in any of the shoot's health.
* If there is a different pending replica with a higher ordinal, the replica with changing health has already been healthy during this scaling in / out, so no reason to be interested in it any more.
* If there is a different pending replica with a lower ordinal, rescheduling the set would lead to a wasted reconcile, since it would most likely simply reach this lower ordinal replica and stop there. The replica with changing health would not be reached.

Regarding the `readyReplicas` field - yes it may be "out of date" during periods of time while some shoots are unhealthy. Is this really so important?

pkg/controllermanager/controller/managedseedset/managedseedset_reconciler.go
stoyanr: OK, yes, good point.

pkg/controllermanager/controller/managedseedset/managedseedset_replicagetter.go
stoyanr: We should enforce it. I thought about it, but then decided to introduce it, together with any other changes to validation (e.g. for `PendingReplica`) in a different PR. Is this ok with you? I could add it here as well if you think it's better.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: OK

pkg/apis/seedmanagement/v1alpha1/types_managedseedset.go
timebertt: Ok, fine for me then.

pkg/utils/kubernetes/eventhandler.go
timebertt: Ah ok, it's for the case of a relist. I wasn't aware, that the event handler would then get an update event. 
Ok, it surely doesn't hurt.

pkg/utils/kubernetes/eventhandler.go
timebertt: > Wouldn't client.Client use a cache as well (if got via Client()) ?

Hopefully we will get there soon. But ATM, if the `CachedRuntimeClients` feature gate is disabled, `Client()` will be uncached.

> What if the object we are interested in is not already in the cache (or out of date)

The changed object itself will always be in the cache already, as it's based on the same watch as the event handler (at least that's true for your controller, which only relies on c-r informers).

The owning object will almost certainly also be in the cache already. Because if it wasn't in the cache already, no controller would have acted on it yet and no ownees would've been created to trigger any events.

pkg/controllermanager/controller/managedseedset/managedseedset_predicates.go
timebertt: Hmm, ok. My understanding of the `readyReplicas` field was, that it would also take the shoots' health status into account. Meaning, it wouldn't just simply show how many replicas it has successfully created (or rolled out) but rather how many of them are actually currently ready (in the sense of healthy).
But it seems you had something else in mind (basically what was my understanding of the `currentReplicas` field).

The question is, what the meaning of the status field should be now. (Can discuss this in our sync tomorrow).

pkg/controllermanager/controller/managedseedset/managedseedset_controller.go
timebertt: Hmm, ok. I think, this is conceptually not quite correct (also the current implementations you referenced) and will just lead to some error message like `stopping reconciliation because something has been deleted`.
Anyways, it's minor. You can choose.

pkg/controllermanager/controller/managedseedset/managedseedset_replicagetter.go
timebertt: Ok, then we're on the same page.
Both is fine for me. You can choose.

pkg/utils/kubernetes/eventhandler.go
stoyanr: OK, I trust you on that, I will change it as suggested.

pkg/controllermanager/controller/managedseedset/managedseedset_replicagetter.go
stoyanr: OK, let's do it in a separate PR then.

pkg/controllermanager/controller/managedseedset/managedseedset_replica.go
timebertt: Replacing some magical string in a resource template just feels wrong to me and I can't think of an example where kubernetes does this.
I wanted to raise this here to get some more opinions (maybe I'm alone with my opinion, I don't know...)
@gardener/gardener-maintainers, any preferences?

pkg/controllermanager/controller/managedseedset/managedseedset_controller.go
stoyanr: I don't see such error messages, and I am rather inclined to leave things as they are just in case.

pkg/controllermanager/controller/managedseedset/managedseedset_predicates.go
stoyanr: Yes, let's discuss the meaning of the field in details. In general, I am rather inclined to avoid triggering additional reconciles just for this one field, but I am not 100% sure.

pkg/controllermanager/controller/managedseedset/managedseedset_replica.go
stoyanr: Perhaps kubernetes doesn't do it because they simply don't have resources similar to this one, where the generated replica name could be used in some non-trivial way inside the template. We however do have this case here. A "magical string" would be less magical if properly documented.

rfranzke: I understand both arguments, though, I would also rather tend to simplify the UX here. I see that using the placeholder has the benefit that the domain can be chosen arbitrarily, but I'm not sure if this is so essential that it justifies having such "magic strings". OTOH, @stoyanr is right that we would impose artificial limitations otherwise.

Generally, I don't like that this is not "optional", i.e., if I want to specify the domain then I have to know about this "magic string" and what happens when I use it. This is not very complex (though, avoidable "cognitive overhead"), but still, if I would have implemented it then I would probably forbid specifying both `.spec.dns.domain` in the `Shoot` and the `.spec.{dns.ingressDomain,ingress.domain}` fields in the `Seed` and add a separate field `.spec.baseDomain` in the `ManagedSeedSet`. Then the `Shoot` domain can be prepended with the `<replica-name>` and the `.spec.ingress.domain` can be prepended with `ingress.<replica-name>` as suggested by @timebertt.
This has the benefit that the operator is not bothered with these placeholder details, but obviously, it limits him using domains with other prefixes. For me personally, the improved UX prevails here.

Btw, I would always use the managed ingress for `ManagedSeedSet`s and forbid specifying `.spec.dns.ingressDomain` in general (independent of above question).

Btw, can you rather use `example.com` instead of `ondemand.com` in your examples?

pkg/controllermanager/controller/managedseedset/managedseedset_predicates.go
stoyanr: We agreed to change this as suggested, and also to change the actuator so that it only acts on the pending replica, if there is one, instead of iterating over all replicas.

pkg/controllermanager/controller/managedseedset/managedseedset_replica.go
stoyanr: @rfranzke We took another look at the implications together with @timebertt and jointly decided in favor of the placeholder approach. The main reason for this is that if we try to allow empty DNS domain in either seed or shoot templates, in some cases (e.g. custom providers), this would also fail the validation, which would complicate things even further - we would either need to introduce further limitations, or to find some way around these validation failures. 

By comparison, the placeholder approach requires no limitations and can be implemented with the least amount of code / changes to existing code. If we have a strong reason we could still change it in the future I believe.

pkg/controllermanager/controller/managedseedset/managedseedset_predicates.go
stoyanr: I realized we missed something rather important so far in this discussion. What happens if I e.g. delete one of the managed seeds manually (or change it, or change its shoot)?

I think the expectation is that the `ManagedSeedSet` controller will eventually restore the set to the state in the resource, including recreating manually deleted objects. This cannot happen if the controller only acts on the pending replica, if there is one. Therefore, IMO the controller must, as originally implemented, iterate over all replicas to find the first one that it needs to act upon. In addition, the filter functions should be extended to make sure such events enqueue the set.

For now I will implement it like this and test it a bit, we should discuss this further.

stoyanr:
@timebertt I added a new commit with the changes resulting from the feedback above. With the exception of the one unresolved comment still remaining, I think it addresses everything as agreed.

pkg/controllermanager/controller/managedseedset/managedseedset_predicates.go
stoyanr: I added a new commit that improves upon the above by ensuring that updates to non-pending replicas don't interrupt scaleouts / scaleins, while at the same time making sure that any non-ready non-pending replicas are eventually taken care of. Please take a look.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
```diff
@@ -0,0 +1,414 @@
...
+		retries := getPendingReplicaRetries(status, r.GetName(), seedmanagementv1alpha1.ShootReconcilingReason)
+		if int(retries) < *a.cfg.MaxShootRetries {
+			a.infoEventf(set, "Retrying shoot %s reconciliation", r.GetFullName())
+			if err := r.RetryShoot(ctx, a.gardenClient.Client()); err != nil {
+				return false, err
+			}
+			updatePendingReplica(status, r.GetName(), seedmanagementv1alpha1.ShootReconcilingReason, pointer.Int32Ptr(retries+1))
+		} else {
+			a.infoEventf(set, "Not retrying shoot %s reconciliation since max retries has been reached", r.GetFullName())
+			updatePendingReplica(status, r.GetName(), seedmanagementv1alpha1.ShootReconcileFailedReason, nil)
```
timebertt: IISIC, this call will clear `status.pendingReplia.retries` (set to `nil`), which in turn makes the controller retry the shoot reconciliation again, as soon as the set is enqueued again. 
Is this correct or am I missing something?
```suggestion
			updatePendingReplica(status, r.GetName(), seedmanagementv1alpha1.ShootReconcileFailedReason, &retries)
```
Same would apply to the failed deletion case below.

timebertt (commented):
Thanks for addressing all of my concerns and suggestions.
I did another round of review and the code looks really good now. I'm very happy with the latest changes.
Only have one minor question, see inline.

I will continue with some manual tests and give some final feedback accordingly (probably tomorrow).

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
```diff
@@ -0,0 +1,414 @@
...
+	set *seedmanagementv1alpha1.ManagedSeedSet,
+	status *seedmanagementv1alpha1.ManagedSeedSetStatus,
+	ordinal int,
+) error {
+	r := a.replicaFactory.NewReplica(set, nil, nil, nil, false)
+	a.infoEventf(set, "Creating shoot %s", getFullName(set, ordinal))
+	if err := r.CreateShoot(ctx, a.gardenClient.Client(), ordinal); err != nil {
+		return err
+	}
+	updatePendingReplica(status, r.GetName(), seedmanagementv1alpha1.ShootReconcilingReason, nil)
```
timebertt, line 288: We should set the pending replica before creating a new replica (same applies to the deletion below).
If a Shoot fails to be created because of some reason, the controller will post an event, but the pending replica is not set (which is important for further reconciliations), e.g.:
```
3m18s       Warning   ReconcileError   managedseedset/gcp-europe-west1   Could not reconcile creation or update: the length of the shoot name and the project name must not exceed 21 characters (project: garden; shoot: gcp-europe-west1-0)
```

pkg/controllermanager/controller/managedseedset/managedseedset_controller.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	if set.Generation == set.Status.ObservedGeneration {
+		return
+	}
+	key, err := cache.MetaNamespaceKeyFunc(newObj)
+	if err != nil {
+		return
+	}
+
+	c.managedSeedSetQueue.Add(key)
```
timebertt, line 49: I observed multiple cases (with the `CachedRuntimeClients` feature gate enabled), where the finalizer addition led to a immediate requeue and second reconciliation of the set.
In such cases the shoot cache wasn't synced fast enough so that the controller didn't recognize that it already created some replica and thus immediately created another one (although `spec.replicas` was set to `1`).

There are probably multiple ways to solve this. Some ideas would be:
- don't enqueue if we detect that the finalizer was added (not really correct)
- update the observed generation even before adding the finalizer (would also not solve the root cause with a stale cache)
- fetch replicas using an API reader (also not quite nice, because of the network traffic)
- fetch a metadata only list of replicas using an API reader and cross check the count with the list from the cache

Anyways, I'm not quite sure yet, what's the proper way to solve this.
Maybe you have a good idea? Maybe we can chat about it or even look at some k8s code, how they do it.

pkg/controllermanager/controller/managedseedset/managedseedset_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+	return logger.NewFieldLogger(r.logger, "managedSeedSet", kutil.ObjectName(set))
+}
+
+func (r *reconciler) updateStatus(ctx context.Context, set *seedmanagementv1alpha1.ManagedSeedSet, status *seedmanagementv1alpha1.ManagedSeedSetStatus) error {
+	return kutil.TryPatchStatus(ctx, retry.DefaultBackoff, r.gardenClient.Client(), set, func() error {
+		if status != nil {
+			set.Status = *status
+		}
+		return nil
+	})
```
timebertt: `TryPatchStatus` patches the status with optimistic locking, so it can likely lead to conflicts (especially with cached clients).
We should instead use a strategic merge patch without optimistic locking here:
```suggestion
if status == nil {
		return nil
	}
	patch := client.StrategicMergeFrom(set.DeepCopy())
	set.Status = *status
	return r.gardenClient.Client().Status().Patch(ctx, set, patch)
```

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
```diff
@@ -0,0 +1,414 @@
...
+	switch {
+	case scalingOut:
+		// Initialize a new replica and create its shoot
+		if err := a.createReplica(ctx, set, status, int(status.NextReplicaNumber)); err != nil {
+			return status, err
+		}
+
+		// Increment Replicas and NextReplicaNumber in status
+		status.Replicas++
+		status.NextReplicaNumber++
```
timebertt: If the status update at the end of the reconciliation fails, e.g. because of a conflict error or transient network issues, we cannot recover the next replica number, it simply gets lost.
I observed multiple cases, where this lead to the following error, from which the controller can't recover without a human manually patching the status:
```
time="2021-04-01T10:57:11+02:00" level=info msg="Error syncing ManagedSeedSet garden/gcp-eu: could not update status: ManagedSeedSet.seedmanagement.gardener.cloud \"gcp-eu\" is invalid: status.nextReplicaNumber: Invalid value: 0: cannot be less than status.replicas"
```

We should find some mechanism to detect which replica number we should use next even if we lost some status update.
One idea would be:
Before creating a new replica, check the highest ordinal of existing replicas and use at least one number higher than that if `NextReplicaNumber` lags behind.
This can still lead to reusing an ordinal in case the replica with the highest ordinal has been deleted.
Maybe you have a better idea?

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: Yes, you are right, fixed.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
```diff
@@ -0,0 +1,414 @@
...
+	case !r.IsSeedReady() && !scalingIn:
+		// This replica's seed is not ready, wait for it to be ready before moving to the next replica
+		a.infoEventf(set, "Waiting for seed %s to be ready", r.GetName())
+		updatePendingReplica(status, r.GetName(), seedmanagementv1alpha1.SeedNotReadyReason, nil)
+		return true, nil
+
+	case r.GetShootHealthStatus() != operationshoot.StatusHealthy && !scalingIn:
+		// This replica's shoot is not healthy, wait for it to be healthy before moving to the next replica
+		a.infoEventf(set, "Waiting for shoot %s to be healthy", r.GetFullName())
+		updatePendingReplica(status, r.GetName(), seedmanagementv1alpha1.ShootNotHealthyReason, nil)
```
timebertt, line 270: If a set is currently ready (all replicas healthy and no pending replica) and one of the replicas becomes unhealthy, we make it the pending replica and wait for it to be created.
Now that we ensure, that we don't block ongoing scale operations when previous replicas become unhealthy, I think we should similarly ensure, that a transient/flaky replica health status doesn't block future scale operations. I.e. we should probably not set the pending replica if a shoot's health status changes and only update the `readyReplicas` field.

timebertt (commented):
So I finished my test spin with `ManagedSeedSet` and it looks very promising already, I like it very much!

I have found some issues / hickups, that we probably want to fix right away, before merging this PR.

I also found some other issues mainly wrt UX and validation, which I will leave out here to not block the PR any further than necessary and in order to get the first working version in. We can discuss them afterwards in person, I can open issues for them, whatever you prefer.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: I had similar thoughts, but there is a problem: it's from my perspective impossible to properly distinguish between 1) a shoot that was never healthy after it was created and 2) a shoot that just became unhealthy. Consider for example the last replica during a scale-out: since the target replica count has been reached, if this replica is unhealthy it's impossible to tell in which of the 2 situations we are in.

I think in the first case, we definitely want to stop and wait until it becomes healthy, since the reason might well be in our MSS resource.

My personal opinion is that we shouldn't try to make this too perfect. I would err on the "safe side", and let future scale-out operations be blocked in cases like this. If they are blocked for too long, likely the issue is going to prevent the proper scale-out anyway. Note that only scale-outs will be blocked, scale-ins will still be performed because there the shoot health is not taken into account.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: How did you manage to reproduce this, just by using cached runtime clients, or by also doing something else?

I don't have a better idea than what you propose. I will change the code so that `ordinal` is the greater of `NextReplicaNumber` and the highest existing replica ordinal + 1, and then set `NextReplicaNumber` to `ordinal+1` instead of simply incrementing it. I don't think reusing an ordinal is such a big issue.

pkg/controllermanager/controller/managedseedset/managedseedset_reconciler.go
stoyanr: OK

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: If the shoot fails to be created, this will result in an error, which will be automatically requeued with backoff. The same applies to all errors elsewhere in this method. It doesn't matter that there is no pending replica, the reconcile will be retried until the error is gone, and at that point the pending replica will be set appropriately. 

I did not quite get what was the behavior that you observed and didn't like? Let me try to test it myself, I think I see how it could be reproduced.

pkg/controllermanager/controller/managedseedset/managedseedset_controller.go
stoyanr: It seems to me only your fourth proposal is kind of good enough. I don't have a better idea.

I wonder though, what actually happened next? OK, an extra replica was created, but at some point it was deleted, to reach the target count, correct? Perhaps we could live with this, at least for now?

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: I tested this, it behaves exactly as I expected it should. I got multiple identical errors from the multiple failed attempts to reconcile the MSS:

```
Warning  ReconcileError  84s (x12 over 106s)  gardener-controller-manager  Could not reconcile creation or update: the length of the shoot name and the project name must not exceed 21 characters (project: garden; shoot: i500152-gcpx-long-name-should-fail-0)
```

I don't think we should do anything more here. I don't really like the idea of setting the pending replica to a replica that may not even exist, and I don't think we would be gaining anything.

pkg/controllermanager/controller/managedseedset/managedseedset_controller.go
stoyanr: The interesting part is that I am also testing with `CachedRuntimeClients` feature gate enabled (and have been doing that all the time), but never observed anything like you describe. Did you do something in addition, or is there a way to reliably reproduce the above issue?

stoyanr: I was not able to reproduce this, but implemented the "fourth proposal" anyway.

pkg/controllermanager/controller/managedseedset/managedseedset_actuator.go
stoyanr: Implemented as outlined above.

pkg/controllermanager/controller/managedseedset/managedseedset_reconciler.go
stoyanr: Done

stoyanr:
Added a new commit with some enhancements to validation:
* Ensure that the seed template labels match the MSS selector
* Validate `PendingReplica` field in status

timebertt (approved):
Thanks for updating the PR again and addressing my concerns.
I tested it again and it works very well!
/lgtm now

I assume, `/squash` would be fine for you, right?

/hold until v1.20 is released.

PR #3715, opened 2021-03-16
Random cloud-config user data execution delay per shoot worker node

rfranzke:

/area scalability robustness
/kind enhancement
/priority normal

We have seen upstream issues with readiness probes (e.g., https://github.com/kubernetes/kubernetes/issues/78733) when too many kubelets are restarting at the ~same time on too many nodes. With this PR, every shoot worker node computes a random delay in `[30s,5m]`. If the time difference of the respective last execution (including restarts of systemd units) hasn't exceeded this delay then nothing is done (the script exits early).

I have moved the user-triggered systemd restart via the annotation on the node to the top so that this is always done (independent of the delay).
I suggest to use the separate commits for the review.
/squash

/invite @vpnachev 
/cc @dguendisch @vlerenc 

**Release note**:

```noteworthy user
Every shoot worker node now randomly delays the execution of the cloud-config user data by up to `5m` (earlier, the maximum delay was ~`30s`). This is to prevent too many systemd unit restarts (e.g., kubelet restarts) at the ~same time when there is a change (e.g., a Kubernetes patch version update).
```


pkg/operation/botanist/component/extensions/operatingsystemconfig/executor/templates/scripts/execute-cloud-config.tpl.sh
```diff
@@ -120,6 +122,24 @@ if [[ -f "{{ .pathKubeletKubeconfigReal }}" ]]; then
     echo "Restarting systemd service {{ .unitNameCloudConfigDownloader }} due to $ANNOTATION_RESTART_SYSTEMD_SERVICES annotation"
     systemctl restart "{{ .unitNameCloudConfigDownloader }}" || true
   fi
+
+  # If the time difference from the last execution till now is smaller than the node-specific delay then we exit early
+  # and don't apply the (potentially updated) cloud-config user data. This is to spread the restarts of the systemd
+  # units and to prevent too many restarts happening on the nodes at roughly the same time.
```
vpnachev, line 128: I really find it helpful to have these comments, but this is download by each node every 30 seconds. On large landscapes this can cost considerable amount of money.

vpnachev (approved):
Thanks!
/lgtm

I have a general concern about the comments in the osc original as it affect the network traffic, but this is something that is not in the scope of this PR.

pkg/operation/botanist/component/extensions/operatingsystemconfig/executor/templates/scripts/execute-cloud-config.tpl.sh
timebertt: If the content size of the cloud-config secret is a concern, we can think about doing pre-flight request with meta-data only and compare checksums and only download the content if the checksum changed.

vpnachev: Thanks! This sounds very good.

rfranzke: Hm, what about changing this from `30s` to `60s`? Now it doesn't matter too much anymore. I had actually done it, but reverted it again.

vpnachev: Also sounds to good to me. I just do not want to request such here changes becayse they are more or less out of the scope of this PR.

timebertt:
Let's get this into the upcoming release.
/milestone v1.19

rfranzke:
@timebertt sure, shall we merge it than? Or will you/somebody still take a look?

timebertt:
Yes, merging this now, I don't see anyone assigned and @vpnachev already reviewed.

vpnachev:
I fine to be merged as it is now.

rfranzke:
Thanks guys :)

PR #3708, opened 2021-03-13
Add CloudProfile handling to SeedAuthorizer

timuthy:
**How to categorize this PR?**

/area security
/kind enhancement
/priority normal

**What this PR does / why we need it**:
After #3571 and #3631 this PR adds the first check for `CloudProfile`s to the `SeedAuthorizer`. Please take a look at the enclosed documentation for more information.

Co-authored-by: Rafael Franzke <rafael.franzke@sap.com>

**Which issue(s) this PR fixes**:
Part of #1723

**Release note**:

```feature operator
NONE
```

PR #3716, opened 2021-03-16
Fix shoot care check

rfranzke:

/area usability
/kind bug
/priority normal

After #3708, the gardenlet does no longer WATCH `CloudProfile`s. Hence, trying to fetch the `CloudProfile` via a lister when constructing the `Operation` in the shoot care controller does not work anymore.

/cc @timuthy
/squash

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/operation/shoot/shoot.go
+++ b/pkg/operation/shoot/shoot.go
@@ -63,7 +63,7 @@ func NewBuilder() *Builder {
 		shootObjectFunc: func(context.Context) (*gardencorev1beta1.Shoot, error) {
 			return nil, fmt.Errorf("shoot object is required but not set")
 		},
-		cloudProfileFunc: func(string) (*gardencorev1beta1.CloudProfile, error) {
+		cloudProfileFunc: func(context.Context, string) (*gardencorev1beta1.CloudProfile, error) {
 			return nil, fmt.Errorf("cloudprofile object is required but not set")
 		},
 		shootSecretFunc: func(context.Context, client.Client, string, string) (*corev1.Secret, error) {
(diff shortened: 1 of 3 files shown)


pkg/operation/shoot/shoot.go
```diff
@@ -100,13 +100,29 @@ func (b *Builder) WithShootObjectFromLister(shootLister gardencorelisters.ShootL
...
+	b.cloudProfileFunc = func(context.Context, string) (*gardencorev1beta1.CloudProfile, error) { return cloudProfileObject, nil }
 	return b
 }
 
-// WithCloudProfileObjectFromLister sets the cloudProfileFunc attribute at the Builder after fetching it from the given lister.
-func (b *Builder) WithCloudProfileObjectFromLister(cloudProfileLister gardencorelisters.CloudProfileLister) *Builder {
-	b.cloudProfileFunc = func(name string) (*gardencorev1beta1.CloudProfile, error) { return cloudProfileLister.Get(name) }
+// WithCloudProfileObjectFromAPIReader sets the cloudProfileFunc attribute at the Builder after fetching it from the
+// given API reader.
+func (b *Builder) WithCloudProfileObjectFromAPIReader(reader client.Reader) *Builder {
```
timebertt: Doesn't have to be an **API**Reader, right? You can also pass a cache, right?
```suggestion
func (b *Builder) WithCloudProfileObjectFromReader(reader client.Reader) *Builder {
```

timebertt (commented):
One nit

pkg/operation/shoot/shoot.go
rfranzke: Yep, done

timebertt (approved):
Thanks.
/lgtm

PR #3718, opened 2021-03-16
Update RBAC for GAC

rfranzke:

/area security
/kind bug
/priority normal

With #3631 the GAC watches many more resources but we didn't update its RBAC role accordingly.

**Release note**:

```other operator
NONE
```


timebertt (approved):
Thanks for the quick fix!
/lgtm

PR #3720, opened 2021-03-16
 Validate audit log policies via webhooks

BeckerMax:
/area control-plane
/area audit-logging
/kind enhancement
/priority normal

Before this change a malformed auditpolicy referenced by a Shoot will cause the Shoot to fail to deploy (or to delete) as deploying the Shoot API server fails.

Now, configmaps containing audit policies as well as Shoot resources are validated once created/updated whether the referenced audit policy is valid.

Also includes:

remove audit policy validation from botanist

instead it is now validated inside the gardener-admission-controller in a webhook

I tried to limit the calls to the gardener API server by this web hook as much as possible. Especially, listing and iterating over all Shoots in a ns to validate an audit policy. I tried to do this validation only as the very last step if everything else was already valid.

```other operator
Audit policies referenced by shoots are now validated upon updating shoot or configmap resources.
```


pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+	if err := h.getOldObject(request, oldCm); err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+
+	if request.Operation == admissionv1.Delete {
+		if controllerutil.ContainsFinalizer(oldCm, shoot.FinalizerName) {
+			return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("deletion of audit-policy is forbidden as it is still referenced by a shoot (cm has finalizer)"))
+		}
+		return acadmission.Allowed("deletion of audit-policy accepted")
+	}
```
vpnachev: Hm, do we need this? A finalizer is already set and it will prevent the resource to be removed from the store as long as it is used by any shoot.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+
+		auditPolicy, err := getAuditPolicy(cm)
+		if err != nil {
+			return admission.Errored(http.StatusUnprocessableEntity, err)
+		}
+
+		oldAuditPolicy, ok := oldCm.Data[auditPolicyConfigMapDataKey]
+		if ok {
+			if oldAuditPolicy == auditPolicy {
+				return acadmission.Allowed("audit policy did not change")
```
vpnachev:
```suggestion
				return acadmission.Allowed("audit policy not changed")
```

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+			return acadmission.Allowed("configmap is not referenced by a Shoot")
+		}
+
+		auditPolicy, err := getAuditPolicy(cm)
+		if err != nil {
+			return admission.Errored(http.StatusUnprocessableEntity, err)
+		}
+
+		oldAuditPolicy, ok := oldCm.Data[auditPolicyConfigMapDataKey]
+		if ok {
```
vpnachev: what if `ok == false` ?

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+		if err := h.apiReader.List(ctx, existingShoots, client.InNamespace(request.Namespace)); err != nil {
+			return admission.Errored(http.StatusInternalServerError, err)
+		}
+		if err := validatePolicyAgainstShoots(request, schemaVersion, existingShoots); err != nil {
+			return admission.Errored(http.StatusUnprocessableEntity, err)
+		}
+
+		return acadmission.Allowed("configmap change is valid")
+	}
+	return acadmission.Allowed("operation is not update or delete")
```
vpnachev: can you move this before the if block.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+
+// IsValidAuditPolicyVersion checks whether the api server support the provided audit policy apiVersion
+func IsValidAuditPolicyVersion(shootVersion string, schemaVersion *schema.GroupVersionKind) (bool, error) {
+	auditGroupVersion := schemaVersion.GroupVersion().String()
+
+	if auditGroupVersion == "audit.k8s.io/v1" {
+		return version.CheckVersionMeetsConstraint(shootVersion, ">= v1.12")
+	}
+	return true, nil
+}
```
vpnachev: You can add a TODO to update this check when https://github.com/kubernetes/kubernetes/issues/98035 is resolved.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+	return nil
+}
+
+func getAuditPolicy(cm *corev1.ConfigMap) (auditPolicy string, err error) {
+	auditPolicy, ok := cm.Data[auditPolicyConfigMapDataKey]
+	if !ok {
+		return "", fmt.Errorf("missing '.data.policy' in audit policy configmap")
+	}
+	if len(auditPolicy) == 0 {
+		return "", fmt.Errorf("empty audit policy. Provide non-empty audit policy")
```
vpnachev, line 261: Empty audit policy is valid audit policy. We are currently using it, ref:
- https://github.com/gardener/gardener/blob/b84868a76c38343ec3add63055320e12878418e7/charts/seed-controlplane/charts/kube-apiserver/values.yaml#L112
- https://github.com/gardener/gardener/blob/b84868a76c38343ec3add63055320e12878418e7/charts/seed-controlplane/charts/kube-apiserver/templates/configmap-audit-policy.yaml#L9

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+func validatePolicyAgainstShoots(request admission.Request, schemaVersion *schema.GroupVersionKind, shoots *gardencorev1beta1.ShootList) error {
+	for _, s := range shoots.Items {
+		if hasAuditPolicy(&s) {
+			cmRef := s.Spec.Kubernetes.KubeAPIServer.AuditConfig.AuditPolicy.ConfigMapRef
+			if cmRef.Name == request.Name {
+				if isValidVersion, err := IsValidAuditPolicyVersion(s.Spec.Kubernetes.Version, schemaVersion); err != nil {
+					return err
+				} else if !isValidVersion {
+					err := fmt.Errorf("your shoot cluster version %q is not compatible with audit policy version %q", s.Spec.Kubernetes.Version, schemaVersion.GroupVersion().String())
+					return err
```
vpnachev:
```suggestion
					return fmt.Errorf("your shoot cluster version %q is not compatible with audit policy version %q", s.Spec.Kubernetes.Version, schemaVersion.GroupVersion().String())
```

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission_test.go
```diff
@@ -0,0 +1,536 @@
...
+---
+apiVersion: audit.k8s.io/v1beta1
+kind: Policy
+rules:
+  - level: RequestResponse
+    resources:
+    - group: "
+      resources: ["pods"]
+`
+		inValidAuditPolicy = `
```
vpnachev:
```suggestion
		invalidAuditPolicy = `
```

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission_test.go
```diff
@@ -0,0 +1,536 @@
...
+				for _, gvk := range gvks {
+					ok, err := auditpolicy.IsValidAuditPolicyVersion(shootVersion, &gvk)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(ok).To(BeTrue())
+				}
+			}
+		})
+
+		It("should return false with error because of not valid semver version", func() {
+			shootVersion := "1.ab.0"
```
vpnachev, line 244: Could you also try with something like `1.12.3-dev`.
We had some troubles with such versions (with the OCI images and GKE versions) and it will be nice to have a test for it.

vpnachev (commented):
Thanks.
Overall looks good to me. I have some minor comments and change requests.

hack/local-development/dev-setup-register-gardener
```diff
@@ -354,5 +354,51 @@ $ADMISSION_CONTROLLER_PORT_STRING
...
+  namespaceSelector:
+    matchLabels:
+      gardener.cloud/role: project
+  admissionReviewVersions: ["v1", "v1beta1"]
+  timeoutSeconds: 10
+  clientConfig:
+    service:
+      namespace: garden
+      name: gardener-admission-controller
+      path: /webhooks/validate-audit-policies
```
rfranzke: Hm, I don't really get why we need the two rules for the local development case. Can you explain?

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	switch request.Kind {
+	case shootGVK:
+		return h.admitShoot(ctx, request)
+	case configmapGVK:
+		return h.admitConfigMap(ctx, request)
+	}
+	return acadmission.Allowed("resource is not core.gardener.cloud.v1beta1.shoot or corev1.configmap")
```
rfranzke:
```suggestion
	return acadmission.Allowed("resource is not core.gardener.cloud/v1beta1.Shoot or v1.ConfigMap")
```

pkg/admissioncontroller/webhooks/admission/auditpolicy/auditpolicy_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+package auditpolicy_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestNamespaceDeletion(t *testing.T) {
```
rfranzke:
```suggestion
func TestAuditPolicy(t *testing.T) {
```

pkg/admissioncontroller/webhooks/admission/auditpolicy/auditpolicy_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "AdmissionController Admission Auditpolicy Suite")
```
rfranzke:
```suggestion
	RunSpecs(t, "AdmissionController Admission AuditPolicy Suite")
```

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+
+func (h *handler) getOldObject(request admission.Request, oldObj runtime.Object) error {
+	if len(request.OldObject.Raw) != 0 {
+		if err := h.decoder.DecodeRaw(request.OldObject, oldObj); err != nil {
+			return err
+		}
+	} else {
+		return fmt.Errorf("could not find old object")
+	}
+	return nil
```
rfranzke:
```suggestion
	if len(request.OldObject.Raw) != 0 {
		return h.decoder.DecodeRaw(request.OldObject, oldObj)
	}
	return fmt.Errorf("could not find old object")
```

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	auditPolicyCm := &corev1.ConfigMap{}
+	if err := h.apiReader.Get(ctx, kutil.Key(shoot.Namespace, cmRef.Name), auditPolicyCm); err != nil {
+		if apierrors.IsNotFound(err) {
+			return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("referenced audit policy does not exist: namespace: %s, name: %s", cmRef.Namespace, cmRef.Name))
+		}
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("could not retrieve config map: %s", err))
+	}
+	auditPolicy, err := getAuditPolicy(auditPolicyCm)
```
rfranzke, line 149:
```suggestion

	auditPolicy, err := getAuditPolicy(auditPolicyCm)
```

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+
+	return acadmission.Allowed("referenced audit policy is valid")
+}
+
+func (h *handler) admitConfigMap(ctx context.Context, request admission.Request) admission.Response {
+	var (
+		oldCm = &corev1.ConfigMap{}
+		cm    = &corev1.ConfigMap{}
+	)
+	if err := h.getOldObject(request, oldCm); err != nil {
```
rfranzke:
```suggestion

	if err := h.getOldObject(request, oldCm); err != nil {
```

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
rfranzke: Yeah, I just wanted to write a similar comment. Is there a reason why you added this?

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+			return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("failed to decode the provided audit policy err=%v", err))
+		}
+		auditPolicyInternal, ok := auditPolicyObj.(*audit_internal.Policy)
+		if !ok {
+			return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failure to cast to audit Policy type: %v", schemaVersion))
+		}
+		errList := auditvalidation.ValidatePolicy(auditPolicyInternal)
+		if len(errList) != 0 {
+			return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("provided invalid audit policy err=%v", errList))
+		}
```
rfranzke: I haven't checked in detail, but this is the same code like above, or? Maybe you can move it inside a function and call it from both places?

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+
+		// Validate schema version against Shoot Kubernetes version
+		existingShoots := &gardencorev1beta1.ShootList{}
```
rfranzke:
```suggestion
		shootList := &gardencorev1beta1.ShootList{}
```

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+	if len(auditPolicy) == 0 {
+		return "", fmt.Errorf("empty audit policy. Provide non-empty audit policy")
+	}
+	return auditPolicy, nil
+}
+
+func validatePolicyAgainstShoots(request admission.Request, schemaVersion *schema.GroupVersionKind, shoots *gardencorev1beta1.ShootList) error {
+	for _, s := range shoots.Items {
+		if hasAuditPolicy(&s) {
+			cmRef := s.Spec.Kubernetes.KubeAPIServer.AuditConfig.AuditPolicy.ConfigMapRef
```
rfranzke: Is this variable helpful?

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,294 @@
...
+			if cmRef.Name == request.Name {
+				if isValidVersion, err := IsValidAuditPolicyVersion(s.Spec.Kubernetes.Version, schemaVersion); err != nil {
+					return err
+				} else if !isValidVersion {
```
rfranzke: I guess you don't need `else if` here, or? Just another `if` case would do the same and reduces nesting (= improves readability)?

rfranzke (commented):
Very nice PR, I like it! I mainly have styling suggestions to improve readability.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
vpnachev: The only reason that come to my mind is if the configmap become immutable after the deletionTimestamp is set, but this is not the case.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: There is nothing to do if the old configmap does not have a Data field. Then we are going to proceed validating the new one (which has a [data field](https://github.com/BeckerMax/gardener/blob/038f014492000e079c733e1e627cfa297dfd5832/pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go#L198)).

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: I agree with you.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: There was the same check before:
https://github.com/gardener/gardener/pull/3720/files#diff-201c3f0c956474991f7e0ebb09d5cb2c23b1b8dbfdad19c36c81c188ce1f13a5L862
How did we deploy an empty policy?

hack/local-development/dev-setup-register-gardener
BeckerMax: The only difference is the clientConfig
The gardener extension API server uses the following URL to contact the gardener-admission-controller:
```
  clientConfig:
    url: https://127.0.0.1:$ADMISSION_CONTROLLER_SECURE_PORT/webhooks/validate-audit-policies
```
The kubernetes cluster runs in a docker environment and uses a service pointing to `host.docker.internal` to connect to it:
```
    service:
      namespace: garden
```
You can see the same pattern in the other webhook configurations.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: I tried to do that but gave up after about 20minutes. The problem is that we need the `schemaVersion` later and then we need at least 3 return values (schema version, error Code and error message) for a function. Then this function is kind of unintuitive validating but also returning a version -  so I left it like it is.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: nice :)

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: It is a bit less succinct as I need to do the `err!= nil` check in a separate line. But I agree it is more easy to read.

pkg/admissioncontroller/webhooks/admission/auditpolicy/auditpolicy_suite_test.go
BeckerMax: Thanks for caching that :)

hack/local-development/dev-setup-register-gardener
rfranzke: Makes sense. Thanks!

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
rfranzke: Hm, I see, it wouldn't improve it too much. OK then.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission_test.go
BeckerMax: I added a test - can you check if I understood you correctly?

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
vpnachev: Excuse me for the inconvenience, it is my mistake.
It is not empty policy, but a policy with a single rule disabling audit logging https://github.com/gardener/gardener/blob/4c6153dceae2e461607c9a38afca88277387dda0/charts/seed-controlplane/charts/kube-apiserver/templates/_helpers.tpl#L93-L96.

timebertt:
/hold

rfranzke (dismissed):
/lgtm

charts/gardener/controlplane/charts/application/templates/validatingwebhook-admission-controller.yaml
```diff
@@ -108,5 +108,41 @@ webhooks:
     {{- end }}
     caBundle: {{ required ".Values.global.admission.config.server.https.tls.caBundle is required" (b64enc .Values.global.admission.config.server.https.tls.caBundle) }}
   sideEffects: None
+- name: validate-audit-policies.gardener.cloud
```
timebertt: Why is this inside `{{- if .Values.global.admission.seedRestriction.enabled }}`?

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,285 @@
...
+)
+
+var (
+	runtimeScheme = runtime.NewScheme()
+	codecs        = serializer.NewCodecFactory(runtimeScheme)
+	decoder       = codecs.UniversalDecoder()
+	_             = auditv1alpha1.AddToScheme(runtimeScheme)
+	_             = auditv1beta1.AddToScheme(runtimeScheme)
+	_             = auditv1.AddToScheme(runtimeScheme)
+	_             = audit_internal.AddToScheme(runtimeScheme)
```
timebertt: Please don't ignore errors in initializations.
Please use an `init` func instead and `utilruntime.Must`, e.g like this:
https://github.com/gardener/gardener/blob/bcdf52d75e2df05c6ff2ee49524964a8c2b05302/cmd/gardener-admission-controller/app/gardener_admission_controller.go#L63-L71

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,285 @@
...
+
+func (h *handler) admitShoot(ctx context.Context, request admission.Request) admission.Response {
+	var (
+		shoot    = &gardencorev1beta1.Shoot{}
+		oldShoot = &gardencorev1beta1.Shoot{}
+	)
+	if request.Operation != admissionv1.Create && request.Operation != admissionv1.Update {
+		return acadmission.Allowed("operation is not Create or Update")
+	}
+
```
timebertt, line 123: WDYT about an early-exit for subresource calls?

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,285 @@
...
+	if err := h.apiReader.Get(ctx, kutil.Key(shoot.Namespace, cmRef.Name), auditPolicyCm); err != nil {
+		if apierrors.IsNotFound(err) {
+			return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("referenced audit policy does not exist: namespace: %s, name: %s", cmRef.Namespace, cmRef.Name))
+		}
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("could not retrieve config map: %s", err))
+	}
+
+	auditPolicy, err := getAuditPolicy(auditPolicyCm)
+	if err != nil {
+		return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("Error: %s in %v/%v", err, cmRef.Namespace, cmRef.Name))
```
timebertt: `%v` / `%w` should be used for formatting errors.
Nit: usually the error is also appended and not prepended, otherwise further wrapping will look weird.
```suggestion
		return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("error getting auditlog policy from ConfigMap %s/%s: %w", cmRef.Namespace, cmRef.Name, err))
```

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,285 @@
...
+		return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("failed to decode the provided audit policy err=%v", err))
+	}
+	auditPolicyInternal, ok := auditPolicyObj.(*audit_internal.Policy)
+	if !ok {
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failure to cast to audit Policy type: %v", schemaVersion))
+	}
+	errList := auditvalidation.ValidatePolicy(auditPolicyInternal)
+	if len(errList) != 0 {
+		return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("provided invalid audit policy err=%v", errList))
+	}
```
timebertt: This code is duplicated, right?

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,285 @@
...
+
+	// Validate schema version against Shoot Kubernetes version
+	shootList := &gardencorev1beta1.ShootList{}
+	if err := h.apiReader.List(ctx, shootList, client.InNamespace(request.Namespace)); err != nil {
```
timebertt, line 211: TBH, listing all shoots of a (potentially large project) every time a configmap gets updated doesn't seem like a good idea to me.
We should do some optimization here:
- either use a cache
- only do the check, if the audit policy version changed
- or something similar

WDYT?

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,285 @@
...
+	return nil
+}
+
+// Handle implements the webhook handler for audit log policy validation
+func (h *handler) Handle(ctx context.Context, request admission.Request) admission.Response {
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	switch request.Kind {
+	case shootGVK:
```
timebertt: This should also catch requests to `shoots.core.gardener.cloud/v1alpha1` and basically all possible future Gardener API versions.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,285 @@
...
+func validatePolicyAgainstShoots(request admission.Request, schemaVersion *schema.GroupVersionKind, shoots *gardencorev1beta1.ShootList) error {
+	for _, s := range shoots.Items {
+		if hasAuditPolicy(&s) {
+			if s.Spec.Kubernetes.KubeAPIServer.AuditConfig.AuditPolicy.ConfigMapRef.Name == request.Name {
+				isValidVersion, err := IsValidAuditPolicyVersion(s.Spec.Kubernetes.Version, schemaVersion)
+				if err != nil {
+					return err
+				}
+				if !isValidVersion {
+					return fmt.Errorf("your shoot cluster version %q is not compatible with audit policy version %q", s.Spec.Kubernetes.Version, schemaVersion.GroupVersion().String())
```
timebertt: It would be helpful to include the shoot name with the incompatible K8s version in the error message.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,285 @@
...
+
+		if hasAuditPolicy(oldShoot) && reflect.DeepEqual(cmRef, oldShoot.Spec.Kubernetes.KubeAPIServer.AuditConfig.AuditPolicy.ConfigMapRef) {
+			return acadmission.Allowed("audit policy reference did not change in Shoot resource")
+		}
+	}
+
+	auditPolicyCm := &corev1.ConfigMap{}
+	if err := h.apiReader.Get(ctx, kutil.Key(shoot.Namespace, cmRef.Name), auditPolicyCm); err != nil {
+		if apierrors.IsNotFound(err) {
+			return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("referenced audit policy does not exist: namespace: %s, name: %s", cmRef.Namespace, cmRef.Name))
```
timebertt: I guess, this replaces the reference check in the admission plugin, right?
So can this be removed now:
https://github.com/gardener/gardener/blob/bcdf52d75e2df05c6ff2ee49524964a8c2b05302/plugin/pkg/global/resourcereferencemanager/admission.go#L547-L554
?

timebertt (changes requested):
Overall, well done!
I have some concerns, see inline.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: Generally yes. It also sets the resourceVersion - should we do this then in the auditpolicy admission controller? https://github.com/gardener/gardener/blob/c40a8fd73a0dc871d692850411084e8a4dc454fc/plugin/pkg/global/resourcereferencemanager/admission.go#L553

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: check is gone now in new version

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: Thanks very helpful!

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: 👍🏼

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: created a function to increase maintainability eventhough the function is a bit unintuitive in the return values

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: We are checking for a resource version change before this list. Therefore, listing all Shoots will only happen if an already existing cm with an auditpolicy is changed. Which I guess is not happening that often. 
On a side note: As we are now checking for config map equality via the resource Version instead of the content of data field: Is there any mechanism that changes labels/annotations,..  anything that could change the resource Version of the ConfigMap frequently? Because if yes, I would check the data field instead of the resource Version to prevent unnecessary listing of all shoots in the ns.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
BeckerMax: Done also for configmap

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -124,41 +134,26 @@ func (h *handler) admitShoot(ctx context.Context, request admission.Request) adm
...
 	}
 
+	if shoot.Spec.Kubernetes.KubeAPIServer.AuditConfig.AuditPolicy.ConfigMapRef.ResourceVersion == auditPolicyCm.ResourceVersion {
+		return acadmission.Allowed("no change detected in referenced configmap holding audit policy")
+	}
+
 	auditPolicy, err := getAuditPolicy(auditPolicyCm)
 	if err != nil {
-		return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("Error: %s in %v/%v", err, cmRef.Namespace, cmRef.Name))
+		return admission.Errored(http.StatusUnprocessableEntity, fmt.Errorf("error getting auditlog policy from ConfigMap %s/%s: %w", cmRef.Namespace, cmRef.Name, err))
```
timebertt: `cmRef.Namespace` is unset, right. Should be `shoot.Namespace`.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -229,6 +217,22 @@ func (h *handler) admitConfigMap(ctx context.Context, request admission.Request)
 	return acadmission.Allowed("configmap change is valid")
 }
 
+func validateAuditPolicySemantics(auditPolicy string) (schemaVersion *schema.GroupVersionKind, errCode int32, err error) {
+	auditPolicyObj, schemaVersion, err := policyDecoder.Decode([]byte(auditPolicy), nil, nil)
+	if err != nil {
+		return nil, http.StatusUnprocessableEntity, fmt.Errorf("failed to decode the provided audit policy err=%v", err)
```
timebertt: Nit: can you make the returned errors consistent?
```suggestion
return nil, http.StatusUnprocessableEntity, fmt.Errorf("failed to decode the provided audit policy: %v", err)
```

timebertt (commented):
Well done!
Only two nits.

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
timebertt: Yeah, you're right. Checking the ConfigMap's content (i.e. `cm.Data["policy"]`) is better. We should do that.

timebertt (dismissed):
Thanks for addressing my suggestions!
/lgtm

timebertt:
```
>> make revendor
 M go.sum
make: *** [Makefile:166: check-generate] Error 1
```

timebertt (dismissed):
/lgtm

BeckerMax:
Hey @vpnachev @rfranzke @timebertt I wanted to say thank you for your thorough reviews. The code is much better now and I learned  a lot from your comments 👍🏼

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -0,0 +1,290 @@
...
+		cm    = &corev1.ConfigMap{}
+	)
+
+	if err := h.getOldObject(request, oldCm); err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+
+	if request.Operation != admissionv1.Update {
+		return acadmission.Allowed("operation is not update")
+	}
```
rfranzke, line 182: Shouldn't this be moved above before trying to get the old object?

BeckerMax: definitely!

rfranzke (approved):
/lgtm

rfranzke:
@vpnachev do you also plan to take another 👀 ?

vpnachev:
> @vpnachev do you also plan to take another eyes ?

I have it in my list, but if I do not come with feedback by EoB tomorrow, please proceed.

vpnachev (approved):
/lgtm

PR #3721, opened 2021-03-17
Increase initialDelaySeconds to 40 for nginx-ingress probes

rfranzke:

/area scalability
/kind enhancement
/priority normal

In larger seed clusters (>100 shoot control planes) the nginx-ingress-controller is sometimes crash-looping because it's unable to start. The VPA scales down its resources, however, during start-up time the nginx-ingress-controller takes a lot of CPU because it syncs its caches (afterwards, its CPU consumption goes down again by a lot). Due to the fact that the `initialDelaySeconds` are set quite low, if the VPA was too optimistic, the nginx-ingress-controller is no longer able to start up. The VPA does not understand that the following crash-loop is due to CPU-throttling (from its perspective, the pods get killed by kubelet because they are not ready/healthy for whatever reason).
Let's increase this delay to give the `nginx-ingress-controller` more time to start-up until we get to a more sophisticated VPA recommender.

**Release note**:

```bugfix operator
NONE
```

### DIFF
--- a/charts/seed-bootstrap/charts/nginx-ingress/templates/controller-deployment.yaml
+++ b/charts/seed-bootstrap/charts/nginx-ingress/templates/controller-deployment.yaml
@@ -81,7 +81,7 @@ spec:
               path: /healthz
               port: 10254
               scheme: HTTP
-            initialDelaySeconds: 10
+            initialDelaySeconds: 40
             periodSeconds: 10
             successThreshold: 1
             timeoutSeconds: 1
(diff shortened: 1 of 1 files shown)


timebertt (approved):
/lgtm in general
One question though, have you considered adding a `startupProbe` instead?

mvladev:
> /lgtm in general
> One question though, have you considered adding a `startupProbe` instead?

The startup probe is only enabled by default since `1.18.x`.

PR #3722, opened 2021-03-17
Make garden/loki-0 PriorityClass value configurable via Gardenlet configuration.

vlvasilev:

/area Logging
/kind enhancement
/priority 3

With this PR we make garden/loki-0 PriorityClass value configurable via Gardenlet configuration.
When it is configured the garden/loki-0 uses different PriorityClass from the others Loki pods.
When the PriorityClass value is not configured or removed the garden/loki-0 uses the same PriorityClass as the others Loki pods.
 
Fixes #

**Release note**:

```other operator
Make garden/loki-0 PriorityClass value configurable via Gardenlet configuration.
```


amshuman-kr (dismissed):
LGTM

timebertt:
> With this PR we increase only the garden/loki-0 priority to 100 in order to not be preempted by etcd pods in the etcd worker pool.

Hmm, I don't understand this sentence. Normally, no `loki` pods should run on a dedicated worker pool.
So is this related to the change?

vlvasilev:
> Hmm, I don't understand this sentence. Normally, no `loki` pods should run on a dedicated worker pool.
> So is this related to the change?

Only the garden/loki-0 pod run on the etcd worker pool.

Kristian-ZH:
@timebertt garden/loki-0 pod runs on the ETCD pool because we need a more stable infrastructure, especially since the lock has disks and there are serious downtimes in the attachment process

Kristian-ZH (approved):

pkg/operation/seed/seed.go
```diff
@@ -393,6 +393,11 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
 			return err
 		}
 
+		lokiValues["priorityClass"] = map[string]interface{}{
+			"value": 100,
```
timebertt: This should be higher than the priority of `gardener-shoot-controlplane` (currently set to `100`), right?

timebertt (dismissed):
General question: shouldn't we rather have a general priority class like `gardener-seed-components` instead to raise priority for all pods in the `garden` namespace?

pkg/operation/seed/seed.go
Kristian-ZH: The mess up here is coming from the fact that there are two different pools: The etcd and the general one.
Both pools share the same priority classes. Here for example: gardener-shoot-controlplane is used from kube-apiserver and etcd pod. but in the same time both pods do not share the same nodes.

The Idea of Vladimir is to make the priorityclass of the Loki to be equal to the one of the ETCD (e.g. 100). In this way Loki pods will not be preempted by the etcd pods

vpnachev:
> garden/loki-0 priority to 100 in order to not be preempted by etcd pods 

So we prefer to have a downtime for the API of a shoot cluster instead of loosing some application logs?

pkg/operation/seed/seed.go
vlvasilev: There is no need loki to preempt or be placed ahead of the etcd in the scheduling queue. Once loki find its place on a node it will not be preempted by other etcd pod. So 100 is enough.

vlvasilev:
> So we prefer to have a downtime for the API of a shoot cluster instead of loosing some application logs?

Unfortunately we loose a lot of logs especially on live landscape when a lot of shoot are created in a short period of time and the reserve excess capacity is not enough.

timebertt:
/hold
Can be cherry-picked later on as soon as we reached an agreement.

Kristian-ZH:
> > garden/loki-0 priority to 100 in order to not be preempted by etcd pods
> 
> So we prefer to have a downtime for the API of a shoot cluster instead of loosing some application logs?

API-servers of the shoot clusters run on different worker pool.

Loki will not replace any of the etcd's because they will have the same priority class.
The only possible downtime is when a big amount of shoot clusters are created in the same time. 
But if this happen, I do not think that Loki will be the problem, because it consumes less resources than the etcd

rfranzke:
The gardener/gardener repository does not make any assumptions about worker pools of the seed...

amshuman-kr:
> API-servers of the shoot clusters run on different worker pool

But the etcds backing the apiserver are on the dedicated worker pool and hence API for the shoot clusters are indirectly affected if an etcd pod is pending because it couldn't pre-empt something else to make space.

Considering the number of shoot clusters (20) that got created in the same seed which caused the pre-emption of `garden/loki` in the first place, majority of the new shoots would anyway have faced downtime as `garden/loki` would have made room for only one or two etcds. 

I think we should revisit how much excess capacity we reserve on the dedicated worker nodes. Currently we reserve `cpu: 600m, memory: 2G` (2 replicas of `cpu: 300m, memory: 1G`. This is nowhere nearly enough to accommodate `20` new etcds in one go. On the other hand, reserving too much excess capacity would be wasteful if those many new etcds do not get created regularly.

Another thing to keep in mind is that the large number of new etcds might be related to the recent (last week) activation of dedicated workers in the live landscape which prompted all etcds to be recreated/migrated to the dedicated nodes (in the shoot maintenance windows). Maybe the number of new etcds might be lower now that all the etcds have migrated.

vlvasilev:
> The gardener/gardener repository does not make any assumptions about worker pools of the seed...

So we either attach the garden/loki-0 to the controlplane priorityclass or make it configurable by landscape setup?

vpnachev (dismissed):
/lgtm

Kristian-ZH (dismissed):

rfranzke:
Hello colleagues, I see you approved this PR now, but what about the concerns from @timebertt and especially the "conceptual" issue that we are doing things here with assumptions of how specific landscapes were setup (dedicated worker pool)?

vlvasilev:
I can make it configurable via the landscape setup. It will be a relatively easy solution. What do you think?

rfranzke:
> I can make it configurable via the landscape setup. It will be a relatively easy solution. What do you think?

I guess it makes more sense, yes. Other opinions? @timebertt, @vpnachev?

vpnachev:
> > I can make it configurable via the landscape setup. It will be a relatively easy solution. What do you think?
> 
> I guess it makes more sense, yes. Other opinions? @timebertt, @vpnachev?

Sounds good to me. I guess, this configuration should be here https://github.com/gardener/gardener/blob/77e09dd522d7bfb682cc35d0be437201b606061b/pkg/gardenlet/apis/config/v1alpha1/types.go#L377-L382

pkg/gardenlet/apis/config/types.go
```diff
@@ -309,10 +309,23 @@ type FluentBit struct {
...
 
+// Loki contains configuration for the Loki.
+type Loki struct {
+	// GardenLoki contains configuration for the Loki in garden namespace.
+	GardenLoki *GardenLoki
+}
+
+// GardenLoki contains configuration for the Loki in garden namespace.
+type GardenLoki struct {
+	PriorityClassValue *int
```
vpnachev: Need doc string.
```suggestion
	// PriorityClassValue ....
	PriorityClassValue *int
```

pkg/gardenlet/apis/config/v1alpha1/types.go
```diff
@@ -374,11 +374,25 @@ type FluentBit struct {
...
 
+// Loki contains configuration for the Loki.
+type Loki struct {
+	// GardenLoki contains configuration for the Loki in garden namespace.
+	GardenLoki *GardenLoki `json:"gardenLoki,omitempty" yaml:"gardenLoki,omitempty"`
+}
+
+// GardenLoki contains configuration for the Loki in garden namespace.
+type GardenLoki struct {
+	PriorityClassValue *int `json:"priorityClassValue,omitempty" yaml:"priorityClassValue,omitempty"`
```
vpnachev: Also needs docu string.

pkg/operation/common/types.go
```diff
@@ -283,4 +283,7 @@ const (
 
 	// ShootDNSIngressName is a constant for the DNS resources used for the shoot ingress addon.
 	ShootDNSIngressName = "ingress"
+
+	// GardenPriorityClassName is the name of the PriorityClass for the Loki in the garden namespace
+	GardenPriorityClassName = "garden-loki"
```
vpnachev:
```suggestion
	// GardenLokiPriorityClassName is the name of the PriorityClass for the Loki in the garden namespace
	GardenLokiPriorityClassName = "garden-loki"
```

pkg/operation/seed/seed.go
```diff
@@ -1293,3 +1311,26 @@ func deleteIngressController(ctx context.Context, c client.Client) error {
...
+func deletePriorityClassIfValueNotTheSame(ctx context.Context, k8sClient client.Client, priorityClassName string, valueToCompare int) error {
+	result, err := comparePriorityClassValue(ctx, k8sClient, priorityClassName, valueToCompare)
+	if err != nil {
+		return err
+	}
+	if result == 0 {
+		return nil
+	}
+	pc := &schedulingv1.PriorityClass{ObjectMeta: metav1.ObjectMeta{Name: priorityClassName}}
+	if err := k8sClient.Delete(ctx, pc); client.IgnoreNotFound(err) != nil && !meta.IsNoMatchError(err) {
```
vpnachev: Why are you checking `!meta.IsNoMatchError(err)` for a standard resource?

pkg/operation/seed/seed.go
```diff
@@ -1293,3 +1311,26 @@ func deleteIngressController(ctx context.Context, c client.Client) error {
 		&corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: "nginx-ingress-k8s-backend", Namespace: v1beta1constants.GardenNamespace}},
 	)
 }
+
+func comparePriorityClassValue(ctx context.Context, k8sClient client.Client, priorityClassName string, valueToCompare int) (int, error) {
+	pc := &schedulingv1.PriorityClass{}
+	if err := k8sClient.Get(ctx, kutil.Key(priorityClassName), pc); client.IgnoreNotFound(err) != nil && !meta.IsNoMatchError(err) {
+		return 0, err
+	}
+	return valueToCompare - int(pc.Value), nil
```
vpnachev: if the resource does not exist, this `valueToCompare` what value will have?

pkg/operation/seed/seed.go
```diff
@@ -1293,3 +1311,26 @@ func deleteIngressController(ctx context.Context, c client.Client) error {
...
+}
+
+func deletePriorityClassIfValueNotTheSame(ctx context.Context, k8sClient client.Client, priorityClassName string, valueToCompare int) error {
+	result, err := comparePriorityClassValue(ctx, k8sClient, priorityClassName, valueToCompare)
+	if err != nil {
+		return err
+	}
+	if result == 0 {
+		return nil
+	}
```
vpnachev: Could you simplify this to just 
1. get the resource
1. If there is error
   1. If it does not exist, return without errors
   1. return the error
1. if the value is the same, return without errors
1. delete the resource

What do you think?

pkg/gardenlet/apis/config/types.go
```diff
@@ -309,10 +309,23 @@ type FluentBit struct {
 	OutputSection *string
 }
 
+// Loki contains configuration for the Loki.
+type Loki struct {
+	// GardenLoki contains configuration for the Loki in garden namespace.
+	GardenLoki *GardenLoki
```
vpnachev: While you are touching these lines, what do you think to add the same configuration options but for the shoot lokis?

vpnachev (changes requested):
Thanks!
Just some minor change requests and several questions.

/assign

pkg/operation/seed/seed.go
```diff
@@ -393,6 +394,23 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+				return err
+			}
+			lokiValues["priorityClass"] = map[string]interface{}{
+				"value": priority,
+				"name":  common.GardenPriorityClassName,
+			}
+		} else {
+			pc := &schedulingv1.PriorityClass{ObjectMeta: metav1.ObjectMeta{Name: common.GardenPriorityClassName}}
+			if err := k8sSeedClient.Client().Delete(ctx, pc); client.IgnoreNotFound(err) != nil && !meta.IsNoMatchError(err) {
+				return err
```
vpnachev: This will delete it on each seed bootstrap reconciliation if the loki config is not specified. Why?

pkg/gardenlet/apis/config/types.go
vlvasilev: Not exactly the same but, am planning to make the Loki ConfigMap adjustable via the gardelet configuration.
And because the central Loki is in much higher pressure the configuration could be different.

pkg/operation/seed/seed.go
vlvasilev: There is a case when after removing the configuration, there is no one to remove this PriorityClass and it will be removed only when the Logging stack is deleted.

pkg/operation/seed/seed.go
vlvasilev: `valueToCompare ` is passed as argument it does not depends on any resource

pkg/operation/seed/seed.go
vlvasilev: ok

pkg/operation/seed/seed.go
vpnachev: Sorry. I mean `pc.Value`. It should be zero?

pkg/operation/seed/seed.go
vpnachev: I got it. When the config is not set, there is nothing to use the `garden-loki` priority class, therefore it is not needed.

pkg/operation/seed/seed.go
```diff
@@ -393,6 +394,23 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+			if err := deletePriorityClassIfValueNotTheSame(ctx, k8sSeedClient.Client(), common.GardenLokiPriorityClassName, priority); err != nil {
+				return err
+			}
+			lokiValues["priorityClass"] = map[string]interface{}{
+				"value": priority,
+				"name":  common.GardenLokiPriorityClassName,
+			}
+		} else {
+			pc := &schedulingv1.PriorityClass{ObjectMeta: metav1.ObjectMeta{Name: common.GardenLokiPriorityClassName}}
+			if err := k8sSeedClient.Client().Delete(ctx, pc); client.IgnoreNotFound(err) != nil && !meta.IsNoMatchError(err) {
```
vpnachev:
```suggestion
			if err := k8sSeedClient.Client().Delete(ctx, pc); client.IgnoreNotFound(err) != nil {
```

vpnachev (changes requested):
Thanks!

Just one more change and rest is looking good to me.

vpnachev (dismissed):
/lgtm

pkg/gardenlet/apis/config/types.go
```diff
@@ -309,10 +309,24 @@ type FluentBit struct {
...
```
rfranzke:
```suggestion
	Garden *GardenLoki
```

pkg/gardenlet/apis/config/types.go
```diff
@@ -309,10 +309,24 @@ type FluentBit struct {
...
+}
+
+// GardenLoki contains configuration for the Loki in garden namespace.
+type GardenLoki struct {
+	// PriorityClassValue is the priority value for the Loki
+	PriorityClassValue *int
```
rfranzke: What about

```suggestion
	Priority *int32
```

pkg/gardenlet/apis/config/v1alpha1/types.go
```diff
@@ -374,11 +374,25 @@ type FluentBit struct {
 	OutputSection *string `json:"output,omitempty" yaml:"output,omitempty"`
 }
 
+// Loki contains configuration for the Loki.
+type Loki struct {
+	// GardenLoki contains configuration for the Loki in garden namespace.
+	GardenLoki *GardenLoki `json:"gardenLoki,omitempty" yaml:"gardenLoki,omitempty"`
```
rfranzke:
```suggestion
	Garden *GardenLoki `json:"garden,omitempty" yaml:"garden,omitempty"`
```

pkg/gardenlet/apis/config/v1alpha1/types.go
rfranzke: This is not yet resolved.

rfranzke:
```suggestion
	Priority *int `json:"priority,omitempty" yaml:"priority,omitempty"`
```

pkg/operation/seed/seed.go
```diff
@@ -393,6 +394,23 @@ func BootstrapCluster(ctx context.Context, k8sGardenClient, k8sSeedClient kubern
...
+			}
+			lokiValues["priorityClass"] = map[string]interface{}{
+				"value": priority,
+				"name":  common.GardenLokiPriorityClassName,
+			}
+		} else {
+			pc := &schedulingv1.PriorityClass{ObjectMeta: metav1.ObjectMeta{Name: common.GardenLokiPriorityClassName}}
+			if err := k8sSeedClient.Client().Delete(ctx, pc); client.IgnoreNotFound(err) != nil {
+				return err
+			}
```
rfranzke, line 411: What about calling `deletePriorityClassIfValueNotTheSame` with `valueToCompare=-1`?

rfranzke (changes requested):
Please adapt https://github.com/gardener/gardener/blob/master/example/20-componentconfig-gardenlet.yaml

pkg/operation/seed/seed.go
vlvasilev: What do you mean. I am not using uint. If the current value is not -1 the PC will be deleted.

rfranzke: Yes, in this case you always want to delete the `PriorityClass`, and you already have a function for deleting it, so I suggest the following:

```suggestion
			if err := deletePriorityClassIfValueNotTheSame(ctx, k8sSeedClient.Client(), common.GardenLokiPriorityClassName, -1) != nil {
				return err
			}
```

The current value can never be `-1`, so it will always be deleted as desired.

Anyways, it doesn't make a big difference, so I'm approving the PR (thanks for addressing the other suggestions), this was just something that caught my eye :)

rfranzke (approved):
/lgtm

dev/20-componentconfig-gardenlet.yaml
```diff
@@ -0,0 +1,116 @@
+---
+apiVersion: gardenlet.config.gardener.cloud/v1alpha1
```
vpnachev, line 2: How does this file was included in this PR when dev directory is excluded here ? https://github.com/gardener/gardener/blob/1ea177fa7ead2be9122368affbac2e634f08a175/.gitignore#L6

vpnachev: Removed with https://github.com/gardener/gardener/pull/3789.

PR #3725, opened 2021-03-17
Preparing the CP Integration tests to use the Managed Seed implementation

krgostev:

/area control-plane-migration
/kind test

This PR removes the deprecated test definitions for registering and deleting seed. With the new Managed Seed implementation they are no longer needed. Also, this PR adjust the Shoot toleration in the migration test to the toleration that is used in the Managed Seed IT
Fixes #

**Release note**:

```other operator
NONE
```


timebertt:
/hold

rfranzke (commented):
/invite @dguendisch @schrodit

test/framework/shootmigrationtest.go
```diff
@@ -99,13 +99,14 @@ func (t *ShootMigrationTest) MigrateShoot(ctx context.Context) error {
 	t.MigrationTime = metav1.Now()
 	return t.GardenerFramework.MigrateShoot(ctx, &t.Shoot, t.TargetSeed, func(shoot *gardencorev1beta1.Shoot) error {
 		t := gardencorev1beta1.Toleration{}
-		t.Key = core.SeedTaintProtected
+		t.Key = SeedTaintTestRun
+		t.Value = pointer.StringPtr("12345") //GetTestRunID())
```
schrodit:
```suggestion
		t.Value = pointer.StringPtr("12345") //GetTestRunID())
```
what does the `GettestrunID` should the value be the id of the current testrun?
Can you add a comment to explain the value.

schrodit (changes requested):

test/framework/shootmigrationtest.go
krgostev: `GetTestRunID` returns the current testmachinery testrun ID.
The value should match the taint that is set in the seed management test [here](https://github.com/gardener/gardener/blob/bcdf52d75e2df05c6ff2ee49524964a8c2b05302/test/framework/managedseedframework.go#L262)
And yes, it should be `GetTestRunID()`. I have forgot to revert it after my local tests

schrodit (approved):

PR #3728, opened 2021-03-17
Eliminate seed-dns Helm chart

rfranzke:

/area dev-productivity
/kind enhancement
/priority normal

The DNS handling was already "componentified", but the Helm charts were still used to create `DNSOwner`, `DNSProvider`, and `DNSEntry`.
With this PR, the `seed-dns` chart is eliminated.

/squash

Part of #2754

**Release note**:

```other operator
NONE
```


timebertt:
/hold
until `v1.19.0` is released.

pkg/operation/botanist/component/extensions/dns/dnsentry.go
```diff
@@ -30,64 +28,78 @@ import (
...
-// NewDNSEntry creates a new instance of DeployWaiter for a specific DNS entry.
+// NewEntry creates a new instance of DeployWaiter for a specific DNS emptyEntry.
 // <waiter> is optional and it's defaulted to github.com/gardener/gardener/pkg/utils/retry.DefaultOps()
-func NewDNSEntry(
-	values *EntryValues,
-	shootNamespace string,
-	applier kubernetes.ChartApplier,
-	chartsRootPath string,
+func NewEntry(
 	logger logrus.FieldLogger,
```
BeckerMax, line 48: nit: if you reorder the params anyway do you want to do it alphabetically?

pkg/operation/botanist/component/extensions/dns/dnsprovider.go
```diff
@@ -17,83 +17,115 @@ package dns
...
-	return &dnsProvider{
-		values:         values,
-		shootNamespace: shootNamespace,
-		ChartApplier:   applier,
-		chartPath:      filepath.Join(chartsRootPath, "seed-dns", "provider"),
-		logger:         logger,
-		client:         client,
-		waiter:         waiter,
+	return &provider{
+		logger:    logger,
```
BeckerMax, line 66: ordering

pkg/operation/botanist/dns.go
```diff
@@ -126,63 +129,55 @@ func (b *Botanist) DefaultExternalDNSProvider(seedClient client.Client) componen
...
 }
 
 // DefaultExternalDNSOwner returns DeployWaiter which removes the external DNSOwner.
 func (b *Botanist) DefaultExternalDNSOwner(seedClient client.Client) component.DeployWaiter {
-	return component.OpDestroy(dns.NewDNSOwner(
+	return component.OpDestroy(dns.NewOwner(
+		seedClient,
+		b.Shoot.SeedNamespace,
 		&dns.OwnerValues{
 			Name: DNSExternalName,
```
BeckerMax, line 168: Active used to be defaulted to `true` now it is defaulted to `false` this could introduce a change here, right?

charts/seed-dns/entry/values.yaml
```diff
@@ -1,6 +0,0 @@
-name: internal
```
BeckerMax, line 1: Name does not get defaulted to `internal` anymore but that seems to be fine as it is set explicitly when created.

BeckerMax (commented):
thanks, some nits - only 1 real question. Looks way better after the refactoring!

pkg/operation/botanist/component/extensions/dns/dnsentry.go
rfranzke: No, we often have logger first, client next, etc. There is no strict guideline (and I don't think we need one), I just aligned it with the majority of the other code.

pkg/operation/botanist/component/extensions/dns/dnsprovider.go
rfranzke: see above

charts/seed-dns/entry/values.yaml
rfranzke: Yes

pkg/operation/botanist/dns.go
rfranzke: Very nice catch, thanks a lot. I will check and adapt!

rfranzke: Done with 115ec34, have a look

BeckerMax (approved):

PR #3729, opened 2021-03-17
Update github templates with priority label changes

ialidzhikov:
/priority 4

**Release note**:

```other operator
NONE
```


rfranzke (dismissed):
/lgtm

timebertt:
Alternative consideration: we could also drop the default priority label in the templates (or at least for PRs).
I never found them useful because rarely no one changes them and so they just contribute to the large noise of our labels.
If someone wants to express an elevated priority they can explicitly label the issue/PR accordingly.
WDYT?

ialidzhikov:
> Alternative consideration: we could also drop the default priority label in the templates (or at least for PRs).
> I never found them useful because rarely no one changes them and so they just contribute to the large noise of our labels.
> If someone wants to express an elevated priority they can explicitly label the issue/PR accordingly.
> WDYT?

I now removed the priority label from the PR template. @gardener/gardener-maintainers , let me know if there is a strong opinion that the priority label should be present by default in the PR template.

vpnachev (approved):
/lgtm

rfranzke:
Should we then also remove the priority label from the issue templates?

timebertt:
Merging now, to include it in the release. 

> Should we then also remove the priority label from the issue templates?

Can still be done later :)

PR #3730, opened 2021-03-18
Fix client.Reader injection

ialidzhikov:
/kind bug
/kind regression

To my knowledge we should use `InjectAPIReader` (not `InjectReader`) to inject a client.Reader.

See https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.3/pkg/runtime/inject/inject.go#L47-L49

Ref https://github.com/gardener/gardener-extension-provider-alicloud/issues/252

**Release note**:

```bugfix dependency
An issue causing nil pointer dereference in the extension library is now fixed.
```


timebertt (approved):
Thanks!
/lgtm

PR #3738, opened 2021-03-18
Prevent deadlock when Terraform configuration cleanup fails 

rfranzke:

/area dev-productivity
/kind enhancement

In rare cases we see `Shoot`s stuck in deletion because of the following error:

```
Error deleting infrastructure: 1 error occurred: * 2/3 Terraform resources are missing
```

The reason is that the `CleanupConfiguration()` function deletes the two `ConfigMap`s and the one `Secret` one-by-one. Hence, when it fails in between then the "all TF config resources exist"-check in the next `Destroy()` execution fails forever.
This deadlock is now prevented as follows:

1. The `CleanupConfiguration()` function first deletes the state `ConfigMap` (this only happens after a successful `terraform destroy` run), i.e., it's safe to delete the state now.
2. This enables us to improve the "all TF config resources exist"-check in a way that tolerations missing resources when the command is `destroy` (because we know that this can only mean that the state (and maybe more) is missing, but this also means that `terraform destroy` succeeded already in a previous run).

Also, I got rid of the retry behaviour. I *think* the rationale behind it was to prevent transient errors when a new `Infrastructure` is created. In this case, it could have happened that the TF config resources are created, but when the `execute()` function tries to read them again then, due to stale caches, this could lead to cache misses and undesired error messages that potentially would be populated to the end-user. Now, I don't even execute this check when both config + state were freshly initialized (this is also the only situation in which such cache misses are undesired/unexpected). Otherwise, it is only checked once whether all resources exist, and if not, then the error is directly returned.

/cc @timebertt 
/invite @timebertt 

Along the road, I introduced some constants for the Terraform commands.
/squash

**Release note**:

```bugfix user
A rare issue with the `Infrastructure` destruction that may result in `Shoot` resources stuck in deletion has been fixed.
```

### DIFF
--- a/extensions/pkg/terraformer/config.go
+++ b/extensions/pkg/terraformer/config.go
@@ -111,11 +111,16 @@ func (t *terraformer) initializerConfig(ctx context.Context) *InitializerConfig 
 // Initializer to correctly create all the resources as specified in the given InitializerConfig.
 // A default implementation can be found in DefaultInitializer.
 func (t *terraformer) InitializeWith(ctx context.Context, initializer Initializer) Terraformer {
-	if err := initializer.Initialize(ctx, t.initializerConfig(ctx), t.ownerRef); err != nil {
+	config := t.initializerConfig(ctx)
+
+	if err := initializer.Initialize(ctx, config, t.ownerRef); err != nil {
 		t.logger.Error(err, "Could not create Terraformer ConfigMaps/Secrets")
 		return t
 	}
-	t.configurationDefined = true
+
+	t.configurationInitialized = true
+	t.stateInitialized = config.InitializeState
+
 	return t
 }
 
--- a/extensions/pkg/terraformer/config.go
+++ b/extensions/pkg/terraformer/config.go
@@ -250,6 +250,11 @@ func (t *terraformer) ConfigExists(ctx context.Context) (bool, error) {
 func (t *terraformer) CleanupConfiguration(ctx context.Context) error {
 	t.logger.Info("Cleaning up all terraformer configuration")
 
+	t.logger.V(1).Info("Deleting Terraform state ConfigMap", "name", t.stateName)
+	if err := t.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.stateName}}); client.IgnoreNotFound(err) != nil {
+		return err
+	}
+
 	t.logger.V(1).Info("Deleting Terraform variables Secret", "name", t.variablesName)
 	if err := t.client.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: t.namespace, Name: t.variablesName}}); client.IgnoreNotFound(err) != nil {
 		return err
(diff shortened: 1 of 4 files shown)


timebertt (approved):
/lgtm
Though, still don't like the confusing coding in the terraformer library at all.
It's quite brittle and very likely to cause us more headaches.

/second-opinion

rfranzke:
Hm, what do you mean/how is it related to the PR? I had hoped that it became less confusing now (at least the part I touched)?

timebertt:
Yes, the part you touched became clearer now.
It was a general statement about the library. Sorry for the confusion.

extensions/pkg/terraformer/config.go
```diff
@@ -111,11 +111,16 @@ func (t *terraformer) initializerConfig(ctx context.Context) *InitializerConfig
...
+	config := t.initializerConfig(ctx)
+
+	if err := initializer.Initialize(ctx, config, t.ownerRef); err != nil {
 		t.logger.Error(err, "Could not create Terraformer ConfigMaps/Secrets")
 		return t
 	}
-	t.configurationDefined = true
+
+	t.configurationInitialized = true
+	t.stateInitialized = config.InitializeState
```
ialidzhikov, line 122: I find confusing having `configurationInitialized` and `stateInitialized` in the `InitializeWith` func. Currently on infrastructure deletion in the provider extensions, in most of the cases we don't run `InitializeWith`. For example in provider-aws we only call `newTerraformer` - https://github.com/gardener/gardener-extension-provider-aws/blob/4b402bd01068b6bbd0a31b188bc2d16cd2b91d19/pkg/controller/infrastructure/actuator_delete.go#L49 , there is no call to `InitializeWith` func before we call `tf.Destroy`.
With this PR it currently works somehow, but I think that these `configurationInitialized` and `stateInitialized` fields are always set to `false` when `InitializeWith` is not called on deletion (most of the cases for provider extensions). Then these fields will be `false` when the actual state in the system is not `false`.

ialidzhikov (commented):
Thanks for this PR! I played with it locally and it generally prevents the above-mentioned deadlock case.
I have 1 inline comment.

extensions/pkg/terraformer/config.go
rfranzke: Do you have a suggestion? E.g., where to move it?

ialidzhikov: I think it should be in the `execute` func. Do we need actually this check for `!(t.configurationInitialized && t.stateInitialized)`? Can't we always check for the existing resources (as we do currently in the master) and simply tolerate if not all resources exists on destroy? Or better - can we simply skip the check for existing resources on destroy?

rfranzke: > but I think that these configurationInitialized and stateInitialized fields are always set to false when InitializeWith is not called on deletion (most of the cases for provider extensions).
> [...]
> I think it should be in the execute func.

I think that setting these attributes in the `InitializeWith` function is correct. From the doc string:

```golang
// InitializeWith initializes the Terraformer with the given Initializer. It is expected from the
// Initializer to correctly create all the resources as specified in the given InitializerConfig.
// [...]
```

So, only when the initializer is called then all configuration resources are created/"initialized". It's not related to the `execute` function.

> Then these fields will be false when the actual state in the system is not false.

Yes, but this is correct, isn't it? These fields do not indicate whether the configuration resources in the system are initialized or not. They only indicate that the current instance of `Terraformer` has initialized them.

> Do we need actually this check for !(t.configurationInitialized && t.stateInitialized)? Can't we always check for the existing resources (as we do currently in the master) and simply tolerate if not all resources exists on destroy? Or better - can we simply skip the check for existing resources on destroy?

Yes, while working on this PR I was also thinking into this direction (completely removing this check - not only for the 'destroy' case). However, I decided to keep it because it's actually helpful and prevents starting a TF pod if the prerequisites are not fulfilled. It would also be okay without the check (we would maybe start a TF pod and it would fail, so the error message wouldn't be too helpful).
Anyways, I don't have a very strong opinion here. I slightly prefer to keep it as it is now. Alternatively, the entire check should be removed (not only for the 'destroy' case).

/cc @timebertt

timebertt: I have no strong opinion here.

ialidzhikov (approved):
/lgtm

PR #3681, opened 2021-03-09
Cleanup context.TODO(), migrate from DeprecatedCreateWorker to CreateWorker (2)

rfranzke:
**How to categorize this PR?**

/area dev-productivity
/kind cleanup
/priority normal

**What this PR does / why we need it**:
Similar to #3658, this PR continues with migration a couple of our controllers from `controllerutils.DeprecatedCreateWorker` (deprecated since #1094) to `controllerutils.CreateWorker`. This helps to get rid of some remaining usages of `context.TODO()`.

**Which issue(s) this PR fixes**:
Part of #1648 

**Special notes for your reviewer**:
ℹ️ Once this PR is merged, I'll follow-up with more PRs that also cover the remaining/not-yet migrated controllers.
/squash

**Release note**:

```other operator
NONE
```

PR #3746, opened 2021-03-22
Use controller-runtime informers for refactored controllers (1)

rfranzke:

/area scalability networking
/kind enhancement

Follow-Up of #3658. After this PR, multiple watches are created via controller-runtime and via the "native" Kubernetes informers. This not only leads to increased network I/O and memory usage but also to race conditions as the informer backing the event handler is different from the informer backing the (cached) client used to retrieve an object after an event. Hence, if not both caches are synced fast enough the reconciliation cannot be performed.

Part of #2822 

/assign @timebertt 
/assign @timebertt 

After this PR is merged I will open a similar PR for the refactored controllers in #3681.

**Release note**:

```other operator
NONE
```


pkg/scheduler/controller/shoot/scheduler_control.go
```diff
@@ -172,13 +170,23 @@ func (r *reconciler) reportEvent(project *gardencorev1beta1.Shoot, eventType str
 }
 
 // determineSeed returns an appropriate Seed cluster (or nil).
-func determineSeed(shoot *gardencorev1beta1.Shoot, seedLister gardencorelisters.SeedLister, shootLister gardencorelisters.ShootLister, cloudProfileLister gardencorelisters.CloudProfileLister, strategy config.CandidateDeterminationStrategy) (*gardencorev1beta1.Seed, error) {
+func determineSeed(
+	ctx context.Context,
+	gardenClient client.Client,
+	shoot *gardencorev1beta1.Shoot,
+	seedLister gardencorelisters.SeedLister,
+	cloudProfileLister gardencorelisters.CloudProfileLister,
```
timebertt: Can we also switch to a c-r clients here?
It seems like this should be easily doable.
Thereby we could avoid confusion and get rid of the remaining `Lister`/`SharedInformerFactory` usage.

pkg/controllermanager/controller/shoot/shoot_quota_control.go
```diff
@@ -120,7 +112,7 @@ func (r *shootQuotaReconciler) Reconcile(ctx context.Context, request reconcile.
 		expirationTime = shoot.CreationTimestamp.Add(time.Duration(*clusterLifeTime*24) * time.Hour).Format(time.RFC3339)
 		metav1.SetMetaDataAnnotation(&shoot.ObjectMeta, common.ShootExpirationTimestamp, expirationTime)
 
-		shootUpdated, err := gardenClient.GardenCore().CoreV1beta1().Shoots(shoot.Namespace).Update(ctx, shoot, kubernetes.DefaultUpdateOptions())
+		shootUpdated, err := r.gardenClient.GardenCore().CoreV1beta1().Shoots(shoot.Namespace).Update(ctx, shoot, kubernetes.DefaultUpdateOptions())
```
timebertt: Switch to c-r here?

pkg/controllermanager/controller/shoot/shoot_quota_control.go
```diff
@@ -136,12 +128,12 @@ func (r *shootQuotaReconciler) Reconcile(ctx context.Context, request reconcile.
...
 
 		// We have to annotate the Shoot to confirm the deletion.
-		if err := common.ConfirmDeletion(ctx, gardenClient.Client(), shoot); err != nil {
+		if err := common.ConfirmDeletion(ctx, r.gardenClient.Client(), shoot); err != nil {
 			return reconcile.Result{}, err
 		}
 
 		// Now we are allowed to delete the Shoot (to set the deletionTimestamp).
-		if err := gardenClient.GardenCore().CoreV1beta1().Shoots(shoot.Namespace).Delete(ctx, shoot.Name, metav1.DeleteOptions{}); err != nil {
+		if err := r.gardenClient.GardenCore().CoreV1beta1().Shoots(shoot.Namespace).Delete(ctx, shoot.Name, metav1.DeleteOptions{}); err != nil {
```
timebertt: And also here?
Then we could change the member
`gardenClient kubernetes.Interface`
to `gardenClient client.Client`.

timebertt (commented):
Very nice, thanks a lot for taking care!
Just minor suggestions.

pkg/scheduler/controller/shoot/scheduler_control.go
rfranzke: Yeah, I didn't do this because the `CachedRuntimeClient` feature gate is alpha and disabled by default, i.e., we would never have watches and do direct calls here.
OTOH, we could either

1. remove the listers and otherwise do nothing for now and live with the direct calls (the GSCHED usually runs next to GAPI, hence, the calls should at least not contribute to network I/O, and also GSCHED is not too busy because usually not a lot of a shoots get created), or
1. explicitly start the C-R informers to have WATCHes for `CloudProfile`s and `Seed`s.

What do you prefer?

timebertt: > Yeah, I didn't do this because the `CachedRuntimeClient` feature gate is alpha and disabled by default, i.e., we would never have watches and do direct calls here.

True of course. We could still explicitly get/list from the cache which is started in both cases.

It will be more tricky for gcm and gardenlet as we will then cause duplicated watches, but for the scheduler (where its an isolated controller) we should be able to easily tackle the problem this way.

I'm basically fine with either way, but slightly prefer explicitly listing from the cache.

rfranzke: OK, I've pushed 4480443 now, I hope this fits to your expectations :)

pkg/scheduler/controller/shoot/scheduler.go
```diff
@@ -70,21 +70,11 @@ func NewGardenerScheduler(
...
-
-		seedInformer         = coreV1beta1Informer.Seeds()
-		seedLister           = seedInformer.Lister()
-		cloudProfileInformer = coreV1beta1Informer.CloudProfiles()
-		cloudProfileLister   = cloudProfileInformer.Lister()
-		shootQueue           = workqueue.NewNamedRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(config.Schedulers.Shoot.RetrySyncPeriod.Duration, 12*time.Hour), "gardener-shoot-scheduler")
-	)
-
 	schedulerController := &SchedulerController{
 		k8sGardenCoreInformers: gardenCoreInformerFactory,
```
timebertt: Can be removed now, I guess.

timebertt (commented):
Very nice, love it!
Only one more nit.

pkg/scheduler/controller/shoot/scheduler.go
rfranzke: Nice catch, done! Please check d8b3f17 and b76327c.

cmd/gardener-scheduler/app/gardener_scheduler.go
```diff
@@ -267,19 +267,15 @@ func (g *GardenerScheduler) startScheduler(ctx context.Context) {
 	if err != nil {
 		panic(fmt.Errorf("failed to create shoot scheduler controller: %+v", err))
 	}
-	//backupBucketScheduler := backupbucketcontroller.NewGardenerScheduler(ctx, g.K8sGardenClient, g.K8sGardenCoreInformers, g.Config, g.Recorder)
```
timebertt, line 270: Very nice, also wanted to drop that code 🚀

timebertt (approved):
/lgtm

rfranzke:
You didn't squash-merge... :)

timebertt:
Yeah, wasn't sure what you preferred, so I acted like a robot and checked whether the label was present and used the default merge mechanism accordingly 🤖 😉

rfranzke:
Touché :)

PR #3753, opened 2021-03-22
Add ConfigMap handling to SeedAuthorizer

rfranzke:

/area security
/kind enhancement

This PR adds handling for `ConfigMap` resource to the `SeedAuthorizer`.

Part of #1723 

Similar to #3708

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
+++ b/pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
@@ -22,6 +22,7 @@ import (
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 
 	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	auth "k8s.io/apiserver/pkg/authorization/authorizer"
 )
@@ -47,6 +48,8 @@ var _ = auth.Authorizer(&authorizer{})
 var (
 	// Only take v1beta1 because the Authorize function only checks the resource group and the resource.
 	cloudProfileResource = gardencorev1beta1.Resource("cloudprofiles")
+
+	configMapResource = corev1.Resource("configmaps")
 )
 
 // TODO: Revisit all `DecisionNoOpinion` later. Today we cannot deny the request for backwards compatibility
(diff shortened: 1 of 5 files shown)


pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
```diff
@@ -64,6 +67,8 @@ func (a *authorizer) Authorize(_ context.Context, attrs auth.Attributes) (auth.D
 		switch requestResource {
 		case cloudProfileResource:
```
timebertt, line 68: BTW: what about other API versions of the `core.gardener.cloud` API?

pkg/admissioncontroller/webhooks/auth/seed/authorizer_test.go
```diff
@@ -196,5 +195,85 @@ var _ = Describe("Seed", func() {
 				Expect(reason).To(BeEmpty())
 			})
 		})
+
+		Context("when requested for ConfigMaps", func() {
```
timebertt, line 199: Side-note (not strictly related to this PR): it would be nice to write an integration test for this and try out something with gardener resources in envtest as well.
WDYT? Should we give it a try?

timebertt (approved):
/lgtm
Love to see the documentation evolving with the latest changes!

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
rfranzke: https://github.com/gardener/gardener/blob/8a51383c03fc647c8231c737c734a9751c3c23cd/pkg/admissioncontroller/webhooks/auth/seed/authorizer.go#L48

pkg/admissioncontroller/webhooks/auth/seed/authorizer_test.go
rfranzke: Yes, I would love to do so. I talked about it with @timuthy already. We would need to have a similar `envtest` package that also sets up `gardener-apiserver` and then we would be able to do so. Do you have something like this in mind?

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
timebertt: Ah right, thanks.

pkg/admissioncontroller/webhooks/auth/seed/authorizer_test.go
timebertt: Yes something like this. Basically compile `gardener-apiserver` in `setup-envtest.sh` and then wrap `envtest` to connect it to the temp etcd.
Alternatively we could generate CRDs from our API types and use them for imitating g-api.
Anyhow, this should be doable, I guess.

PR #3761, opened 2021-03-23
Add CP dependency to OSConfig

kon-angelo:

/area control-plane
/kind bug

Fixes a race condition encountered for an Azure cluster where the `OperatingSystemConfig` would be created before the `ControlPlane`. 

This PR aims to delay the deployment of `OperatingSystemConfig` until after the deployment of CP. In certain extensions (Azure)  

Fixes #

Additional details for the original issue can be seen here: https://github.com/gardener/gardener-extension-provider-azure/issues/267. 

This PR fixes one of the triggering causes (fixing failure during the first reconciliation) for this bug and should be taken into consideration along with https://github.com/gardener/gardener/pull/3762 (which prevents retrying reconciliations).

**Release note**:

```other developer
The `OperatingSystemConfig` is now created after `ControlPlane` has been deployed
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
@@ -313,7 +313,7 @@ func (c *Controller) runReconcileShootFlow(ctx context.Context, o *operation.Ope
 		deployOperatingSystemConfig = g.Add(flow.Task{
 			Name:         "Deploying operating system specific configuration for shoot workers",
 			Fn:           flow.TaskFn(botanist.DeployOperatingSystemConfig).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(deployReferencedResources, waitUntilInfrastructureReady),
+			Dependencies: flow.NewTaskIDs(deployReferencedResources, waitUntilInfrastructureReady, waitUntilControlPlaneReady),
 		})
 		waitUntilOperatingSystemConfigReady = g.Add(flow.Task{
 			Name:         "Waiting until operating system configurations for worker nodes have been reconciled",


vpnachev:
Could give an example for such race condition or actual problem?

vpnachev:
In chat with @kon-angelo it turns out that `EnsureKubeletCloudProviderConfig` has some difficulties at the moment.
My opinion is that the kubelet and OSC related mutators https://github.com/gardener/gardener/blob/fb1715be4030b69279ff75e2337a417d4e7e66ee/extensions/pkg/webhook/controlplane/genericmutator/mutator.go#L42-L76 should be moved out of the `ControlPlane` handler and rather have their own that mutates the `OperatingSystemConfig` resources on `create`/`update`.

rfranzke:
@vpnachev I don't understand your proposal/concerns. The issue @kon-angelo tries to fix here is not related to the fact that kubelet and OSC related mutators are part of the `controlplane` webhook, is it? The `ControlPlane` extension resource typically generates the cloud-provider configuration. The kubelet sometimes needs this configuration, hence, the task which "deploys" the kubelet must be dependent on a successful `ControlPlane` reconciliation to ensure the needed cloud-provider configuration exists in the system.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -313,7 +313,7 @@ func (c *Controller) runReconcileShootFlow(ctx context.Context, o *operation.Ope
 		deployOperatingSystemConfig = g.Add(flow.Task{
 			Name:         "Deploying operating system specific configuration for shoot workers",
 			Fn:           flow.TaskFn(botanist.DeployOperatingSystemConfig).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(deployReferencedResources, waitUntilInfrastructureReady),
+			Dependencies: flow.NewTaskIDs(deployReferencedResources, waitUntilInfrastructureReady, deployControlPlane),
```
```suggestion
			Dependencies: flow.NewTaskIDs(deployReferencedResources, waitUntilInfrastructureReady, waitUntilControlPlaneReady),
```

rfranzke (changes requested):

rfranzke (approved):
/lgtm
/invite @vpnachev

vpnachev (approved):
/lgtm

PR #3762, opened 2021-03-23
RetryableError should not block reconciliation attempts

kon-angelo:

/area control-plane
/kind bug

This PR will allow retrying the reconciliation, after an unsuccessful attempt would return an error code of `RetryableError` from the infrastructure. Previously the `RetryableError` marked as infrastructure dependency error would not be reconciled.

Fixes #

**Release note**:

```other operator
Infrastructure dependency errors containing the `RetryableError` will not stop automatic reconciliation attempts.
```

### DIFF
--- a/docs/usage/shoot_status.md
+++ b/docs/usage/shoot_status.md
@@ -99,6 +99,7 @@ Known error codes are:
 - `ERR_INFRA_INSUFFICIENT_PRIVILEGES` - indicates that the last error occurred due to insufficient infrastructure privileges. It is classified as a non-retryable error code.
 - `ERR_INFRA_QUOTA_EXCEEDED` - indicates that the last error occurred due to infrastructure quota limits. It is classified as a non-retryable error code.
 - `ERR_INFRA_DEPENDENCIES` - indicates that the last error occurred due to dependent objects on the infrastructure level. It is classified as a non-retryable error code.
+- `ERR_RETRYABLE_INFRA_DEPENDENCIES` - indicates that the last error occurred due to dependent objects on the infrastructure level, but the operation should be retried.
 - `ERR_INFRA_RESOURCES_DEPLETED` - indicates that the last error occurred due to depleted resource in the infrastructure.
 - `ERR_CLEANUP_CLUSTER_RESOURCES` - indicates that the last error occurred due to resources in the cluster that are stuck in deletion.
 - `ERR_CONFIGURATION_PROBLEM` - indicates that the last error occurred due to a configuration problem. It is classified as a non-retryable error code.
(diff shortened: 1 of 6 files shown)


rfranzke (dismissed):
/lgtm
/assign @ialidzhikov

pkg/apis/core/v1beta1/types_common.go
```diff
@@ -32,6 +32,8 @@ const (
 	// ErrorInfraDependencies indicates that the last error occurred due to dependent objects on the infrastructure level.
 	// It is classified as a non-retryable error code.
 	ErrorInfraDependencies ErrorCode = "ERR_INFRA_DEPENDENCIES"
+	// ErrorRetryableInfraDependencies indicates that the last error occurred due to dependent objects on the infrastructure level, but operation should be retried.
+	ErrorRetryableInfraDependencies ErrorCode = "ERR_RETRYABLE_INFRA_DEPENDENCIES"
```
ialidzhikov, line 36: Can you update https://github.com/gardener/gardener/blob/master/docs/usage/shoot_status.md#error-codes?
Can you add the error code also for the internal type and v1alpha1 - https://github.com/gardener/gardener/blob/f86f6e4ce3ff57545551e0331411e654aa7da33f/pkg/apis/core/types_common.go#L22 and https://github.com/gardener/gardener/blob/f86f6e4ce3ff57545551e0331411e654aa7da33f/pkg/apis/core/v1alpha1/types_common.go#L20?

ialidzhikov (changes requested):

docs/usage/shoot_status.md
```diff
@@ -99,6 +99,7 @@ Known error codes are:
 - `ERR_INFRA_INSUFFICIENT_PRIVILEGES` - indicates that the last error occurred due to insufficient infrastructure privileges. It is classified as a non-retryable error code.
 - `ERR_INFRA_QUOTA_EXCEEDED` - indicates that the last error occurred due to infrastructure quota limits. It is classified as a non-retryable error code.
 - `ERR_INFRA_DEPENDENCIES` - indicates that the last error occurred due to dependent objects on the infrastructure level. It is classified as a non-retryable error code.
+- `ERR_RETRYABLE_INFRA_DEPENDENCIES` - indicates that the last error occurred due to dependent objects on  on the infrastructure level. "Retryable" means that the occurred error is likely to be resolved in a ungraceful manner after given period of time.
```
ialidzhikov: This statement about "retryable" with `is likely to be resolved in a ungraceful manner after given period of time` is applicable for `ERR_RETRYABLE_CONFIGURATION_PROBLEM`. I guess there are no `ungraceful manners` in the `ERR_RETRYABLE_INFRA_DEPENDENCIES`, so le't reword this to avoid confusion.

```suggestion
- `ERR_RETRYABLE_INFRA_DEPENDENCIES` - indicates that the last error occurred due to dependent objects on the infrastructure level, but the operation should be retried.
```

ialidzhikov (approved):
Thanks!

/lgtm

PR #3768, opened 2021-03-24
Use controller-runtime informers for refactored controllers (2)

rfranzke:

/area scalability networking
/kind enhancement

Follow-Up of #3681 and similar to #3746. After this PR, multiple watches are created via controller-runtime and via the "native" Kubernetes informers. This not only leads to increased network I/O and memory usage but also to race conditions as the informer backing the event handler is different from the informer backing the (cached) client used to retrieve an object after an event. Hence, if not both caches are synced fast enough the reconciliation cannot be performed.

Part of #2822 

/assign @timebertt 
/assign @timebertt 
/squash

**Release note**:

```other operator
NONE
```


pkg/controllermanager/controller/shoot/configmap_control.go
```diff
@@ -106,8 +96,8 @@ func (r *configMapReconciler) Reconcile(ctx context.Context, request reconcile.R
 
 			if shoot.Spec.Kubernetes.KubeAPIServer.AuditConfig.AuditPolicy.ConfigMapRef.ResourceVersion != configMap.ResourceVersion {
 				logger.Logger.Infof("[SHOOT CONFIGMAP controller] schedule for reconciliation shoot %v ", shootKey)
-				// send empty patch to let the admission plugin add the config map resource version
-				if err := kubernetes.SubmitEmptyPatch(context.TODO(), gardenClient.Client(), shoot); err != nil {
+				// send empty patch to let the admission webhook in GAC add or update the config map resource version
+				if err := kubernetes.SubmitEmptyPatch(context.TODO(), r.gardenClient, &shoot); err != nil {
```
timebertt:
```suggestion
				if err := kubernetes.SubmitEmptyPatch(ctx, r.gardenClient, &shoot); err != nil {
```

pkg/controllermanager/controller/project/project_control_delete.go
```diff
@@ -31,36 +32,46 @@ import (
...
 			return reconcile.Result{}, fmt.Errorf("failed to check if namespace is empty: %w", err)
 		}
 
 		if !isEmpty {
 			r.reportEvent(project, true, gardencorev1beta1.ProjectEventNamespaceNotEmpty, "Cannot release namespace %q because it still contains Shoots.", *namespace)
-			_, _ = updateProjectStatus(ctx, gardenClient.GardenCore(), project.ObjectMeta, setProjectPhase(gardencorev1beta1.ProjectTerminating))
+			_ = kutil.TryUpdateStatus(ctx, retry.DefaultBackoff, gardenClient, project, func() error {
+				project.Status.Phase = gardencorev1beta1.ProjectTerminating
+				return nil
+			})
```
timebertt: We not directly switch to a patch? :)

timebertt (commented):
Thanks a lot!
Two minor suggestions.

pkg/controllermanager/controller/project/project_control_delete.go
rfranzke: OK, I've switched to `TryPatchStatus`, I hope this is what you had in mind.

pkg/controllermanager/controller/project/project_control_delete.go
```diff
@@ -44,7 +44,8 @@ func (r *projectReconciler) delete(ctx context.Context, project *gardencorev1bet
 
 		if !isEmpty {
 			r.reportEvent(project, true, gardencorev1beta1.ProjectEventNamespaceNotEmpty, "Cannot release namespace %q because it still contains Shoots.", *namespace)
-			_ = kutil.TryUpdateStatus(ctx, retry.DefaultBackoff, gardenClient, project, func() error {
+
+			_ = kutil.TryPatchStatus(ctx, retry.DefaultBackoff, gardenClient, project, func() error {
```
timebertt: Sorry, `TryPatchStatus` is not what I had in mind. I was rather interested in getting rid of retries, but `TryPatchStatus` also does optimistic locking, so it can also cause conflicts: https://github.com/gardener/gardener/blob/03794c4fab012f16625540cd2989ab8f328be93d/pkg/utils/kubernetes/patch.go#L61

Can you rather do something like this?
https://github.com/gardener/gardener/blob/6ca719c367521bf7e2c1d69f0720af6923a11616/pkg/controllermanager/controller/plant/plant_control.go#L210-L220

rfranzke: Okay, thanks, maybe b9789cc is better :)

timebertt (approved):
Yes, this looks great now.
Thanks for addressing my suggestions!
/lgtm

PR #3780, opened 2021-03-26
Consolidate utilities for working with ManagedResources

rfranzke:

/area dev-productivity
/kind cleanup technical-debt

We had multiple places with utility functions for working with `ManagedResource`s.
With this PR, those are now consolidated and moved into a single place, `pkg/utils/managedresources`.

/squash

**Release note**:

```breaking dependency
⚠️ The utility functions for working with `ManagedResource`s have been mostly moved from `pkg/operation/common` to `pkg/utils/managedresources`. Please note that the signature of the functions might have changed. Especially, the order of the `name, namespace string` parameters is now `namespace, name string`.
```


pkg/utils/managedresources/managedresources.go
```diff
@@ -133,8 +133,8 @@ func WaitUntilManagedResourceDeleted(ctx context.Context, client client.Client,
 	return kutil.WaitUntilResourceDeleted(ctx, client, mr, IntervalWait)
 }
 
-// KeepManagedResourceObjects updates the keepObjects field of the managed resource with the given name in the given namespace.
-func KeepManagedResourceObjects(ctx context.Context, c client.Client, namespace, name string, keepObjects bool) error {
+// UpdateKeepObjects updates the keepObjects field of the managed resource with the given name in the given namespace.
+func UpdateKeepObjects(ctx context.Context, c client.Client, namespace, name string, keepObjects bool) error {
```
BeckerMax: I would expect it to be named `SetKeepObjects` as it sets the value of a field. UpdateKeepObjects was not clear to me just from the naming (it only became clear when reading the comment) because I thought it "updates objects that are kept somewhere".

vpnachev (dismissed):
/lgtm

pkg/utils/managedresources/managedresources.go
```diff
@@ -43,39 +109,55 @@ func CreateManagedResourceFromUnstructured(ctx context.Context, client client.Cl
...
-		Reconcile(ctx); err != nil {
-		return errors.Wrapf(err, "could not create or update secret '%s/%s' of managed resources", namespace, name)
+	return deployManagedResource(ctx, secret, managedResource)
+}
+
+// CreateForSeed deploys a ManagedResource CR for the seed's gardener-resource-manager.
+func CreateForSeed(ctx context.Context, client client.Client, namespace, name string, keepObjects bool, data map[string][]byte) error {
+	var (
+		secretName, secret = NewSecret(client, namespace, name, data, true)
+		managedResource    = NewForSeed(client, namespace, name, keepObjects).WithSecretRef(secretName)
```
BeckerMax, line 129: Why don't we call New() directly here. I feel `NewForSeed` and `NewForShoot` does not help readability.

pkg/utils/managedresources/managedresources.go
```diff
@@ -85,19 +167,30 @@ func DeleteManagedResource(ctx context.Context, client client.Client, namespace
...
+	// DeleteForSeed is a function alias for deleteWithSecretNamePrefix.
+	DeleteForSeed = deleteWithSecretNamePrefix
+	// DeleteForShoot is a function alias for deleteWithSecretNamePrefix.
+	DeleteForShoot = deleteWithSecretNamePrefix
+)
+
+func deleteWithSecretNamePrefix(ctx context.Context, client client.Client, namespace string, name string) error {
+	return Delete(ctx, client, namespace, name, true)
+}
+
```
BeckerMax, line 188: Why do we need this? 
Can't the calling functions just directly call:
`Delete(ctx, client, namespace, name, true)` ?
And we delete the function `DeleteForSeed` and `DeleteForShoot`?

BeckerMax (commented):
I think what I learned from this review is that the more parameters a function has the harder it is understand like `RenderChartAndCreateManagedResource` (especially when reading the code for the first time). Anyways, definitely looks better after the refactoring I think. Just some suggestions/ ask for clarifications below.

pkg/utils/managedresources/managedresources.go
rfranzke: OK, done

pkg/utils/managedresources/managedresources.go
rfranzke: We could replace `NewForSeed` with a direct `New` call and eliminate `NewForSeed` entirely (it's only used here). However, there is also `NewForShoot` which is called also at some other place, hence, it cannot be eliminated. In order to not duplicate things, I would rather keep it as is, i.e., calling `NewForShoot` here. For similarity reasons, I would also keep `NewForSeed` (even if it's not strictly needed today, but it also doesn't harm//not sure why this should impact readability?).

pkg/utils/managedresources/managedresources.go
rfranzke: Yes, you're right, we could do this of course. I've done it like this for two reasons:
1. To not touch too much of the other code (already using `DeleteForS{hoot,eed}` because the PR was already somewhat large.
1. This code already uses `CreateForS{hoot,eed}` which encapsulates the fact that the secret name should prefixed. For similarity reasons, there are also `DeleteForS{hoot,eed}` which encapsulates the same. If we now replace these functions by direct calls to `Delete` then the `secretNameWithPrefix` argument would be needed to be set to `true` explicitly - for all the places (while it's not needed when calling `CreateForS{hoot,eed}`. Honestly, I think this might lead to inconsistencies and would make the functions harder to use.

pkg/utils/managedresources/managedresources.go
BeckerMax: Ok, let's leave it then.

pkg/utils/managedresources/managedresources.go
BeckerMax: ok, I get you logic. We can keep it then.

BeckerMax (approved):

Issue #1278, opened 2019-08-02
Shoots not woken up during hibernation schedule if GCM down


rfranzke:
**What happened**:
If a shoot has a hibernation schedule but the gardener-controller-manager is down during that time for whatever reason the shoot is not woken up.

**What you expected to happen**:
The shoot user expects its shoots to get woken up when configured, or at least as soon as possible after that (if GCM comes up again later).

**How to reproduce it (as minimally and precisely as possible)**:
Configure a hibernation schedule in a shoot, terminate GCM, see the shoot not getting woken up.

**Anything else we need to know?**:
Currently, we don't know whether the hibernation schedule has already been executed. It might be the case that a shoot is woken up automatically at 6 a.m., but the end-user decides to hibernate it again at 6:30 a.m. In this case GCM should not wake it up again. Probably we have to distinguish these two cases.

rfranzke:
It seems that this is a corner case not getting much attention, and given that hibernation is mostly a development feature anyways (especially the schedules), it's unlikely that anybody will pick this up anytime soon.
/close

PR #3782, opened 2021-03-27
Introduce new error code for cloud provider request throttling

### LINKED ISSUE #3775: Rate limit exceeded errors are wrongly detected as non-retriable
**What happened**:
Currently rate limit issues are detected as non-retriable errors. It means that if a cloud provider rate limit exceeded error occurs, the Shoot will be set to Failed and will be no longer retried.

Rate limit exceeded errors should be retied.

ialidzhikov:
/kind bug

This PR moves the cloud provider request rate limit exceeded error code handling from the `ERR_INFRA_QUOTA_EXCEEDED` error code (non-retyable) to a new error code `ERR_INFRA_REQUEST_THROTTLING` (retryable). So now all kind of rate limit errors will be automatically retried. This should fix issues like #3775 where the Shoot is marked as Failed right away (hence manual retry is needed) when a transient rate limit error occurs.

Fixes #3775

**Release note**:

```other operator
Shoot operations that error due to cloud provider rate limit exceeded errors are now classified with the new `ERR_INFRA_REQUEST_THROTTLING` error code. Previously these errors were classified as `ERR_INFRA_QUOTA_EXCEEDED` and they were no longer retried. The new error code `ERR_INFRA_REQUEST_THROTTLING` is classified as retryable and such errors will be now automatically retried.
```

### DIFF
--- a/docs/usage/shoot_status.md
+++ b/docs/usage/shoot_status.md
@@ -98,6 +98,7 @@ Known error codes are:
 - `ERR_INFRA_UNAUTHORIZED` - indicates that the last error occurred due to invalid infrastructure credentials. It is classified as a non-retryable error code.
 - `ERR_INFRA_INSUFFICIENT_PRIVILEGES` - indicates that the last error occurred due to insufficient infrastructure privileges. It is classified as a non-retryable error code.
 - `ERR_INFRA_QUOTA_EXCEEDED` - indicates that the last error occurred due to infrastructure quota limits. It is classified as a non-retryable error code.
+- `ERR_INFRA_REQUEST_THROTTLING` - indicates that the last error occurred due to exceeded infrastructure API throttling limit. It is classified as a retryable error code.
 - `ERR_INFRA_DEPENDENCIES` - indicates that the last error occurred due to dependent objects on the infrastructure level. It is classified as a non-retryable error code.
 - `ERR_RETRYABLE_INFRA_DEPENDENCIES` - indicates that the last error occurred due to dependent objects on the infrastructure level, but the operation should be retried.
 - `ERR_INFRA_RESOURCES_DEPLETED` - indicates that the last error occurred due to depleted resource in the infrastructure.
--- a/extensions/pkg/controller/infrastructure/reconciler.go
+++ b/extensions/pkg/controller/infrastructure/reconciler.go
@@ -17,6 +17,7 @@ package infrastructure
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -48,6 +49,9 @@ const (
 	EventInfrastructureMigration string = "InfrastructureMigration"
 	// EventInfrastructureRestoration an event reason to describe infrastructure restoration.
 	EventInfrastructureRestoration string = "InfrastructureRestoration"
+	// backoffOnInfrastructureRequestThrottling is the backoff used by the infrastructure reconciler
+	// to retry infrastructure request throttling errors.
+	backoffOnInfrastructureRequestThrottling = 10 * time.Minute
 )
 
 type reconciler struct {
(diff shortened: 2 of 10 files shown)


MartinWeindel (dismissed):
lgtm

rfranzke (dismissed):
As discussed, having a dedicated error code for API rate limits/throttling makes sense.
However, we should change the "retryable" part so that we allow the limits to recover in case they exceed.

ialidzhikov:
> However, we should change the "retryable" part so that we allow the limits to recover in case they exceed.

I don't have a good implementation idea for now.
Even if we manage to tweak the retry mechanism in the shoot controller, I think that this won't influence the retry mechanism in the provider extensions for example. If I am not wrong , `DefaultControllerRateLimiter` is used currently by the provider extensions. So, if for example an `Infrastructure` fails to be deleted because of rate limits exceeded error, then the corresponding resource will be requeued and retried according to the `DefaultControllerRateLimiter` -> which more or less means to use the `ItemExponentialFailureRateLimiter` -> aggressive backoffs on rate limit errors. So even if we tweak the shoot controller, the provider extension can retry with aggressive backoffs and exhaust the cloud provider rate limits.

rfranzke:
@ialidzhikov What about the approach of setting the shoot to `Failed`, but have a new controller in GCM which auto-retries shoots failed due to rate limits after some time? This is not relying on any internal workqueue, plays well together with the extensions, and is even resistent against controller restarts (with a new queue and backoffs).

ialidzhikov:
> @ialidzhikov What about the approach of setting the shoot to Failed, but have a new controller in GCM which auto-retries shoots failed due to rate limits after some time? This is not relying on any internal workqueue, plays well together with the extensions, and is even resistent against controller restarts (with a new queue and backoffs).

Thanks for your proposal. ` new controller in GCM which auto-retries shoots [...]` can do the trick of course, but I don't think it is a sustainable solution. It increases the coupling between gardenlet and gardener-controller-manager and the overall complexity.
So, if currently we have concerns about aggressive backoffs for the `Infrastructure` controllers that are likely to cause rate limit exceeded errors for accounts that host huge amount of Shoots, then probably we can try to adapt the RateLimiter for the infrastructure controller. https://github.com/kubernetes-sigs/controller-runtime/pull/731 makes possible to supply a custom `RateLimiter`. I think that the default [`DefaultControllerRateLimiter`](https://github.com/kubernetes/kubernetes/blob/b0abe89ae259d5e891887414cb0e5f81c969c697/staging/src/k8s.io/client-go/util/workqueue/default_rate_limiters.go#L39-L45) might be too aggressive especially for Infrastructure reconciliations. In most of the cases `NewItemExponentialFailureRateLimiter(5*time.Millisecond, 1000*time.Second)` will be the one that will give the higher backoff. Having in mind the exponential backoff definition `baseDelay*2^<num-failures>`, this RateLimiter will give the following backoffs for the first 8 failures:
```
1 - 10ms
2 - 20ms
3 - 40ms
4 - 80ms
5 - 160ms
6 - 320ms
7 - 640ms
8 - 1280ms
...
```

You can see that the first 7 backoffs are less than 1s which seems to be aggressive enough. The maxDelay of 1000s (16.67m) also seems aggressive enough.

If we tweak this for example with `NewItemExponentialFailureRateLimiter(50*time.Millisecond, 2000*time.Second)`, we the following backoffs
```
1 - 100ms
2 - 200ms
3 - 400ms
4 - 800ms
5 - 1600ms
6 - 3200ms
7 - 6400ms
8 - 12800ms
...
```

and a maxDelay of 2000s (33.33m). Which is much less aggressive for an Infrastructure reconciliation -> less load on cloud provider APIs.
Also in the gardenlet ComponentConfig we have `retryDuration` that defaults to 12h. If the Shoot errors for 12h, then it will be set to Failed and no longer retried.

On the other hand we have the shoot controller that "requests" infrastructure reconciliations with the `gardener.cloud/operation=reconcile` annotation and I guess this ignores any kind of backoff in the infrastructure controller. So the custom RateLimiter would only help for having less aggressive backoffs on Infrastructure deletion? We would also need to tweak the Shoot controller RateLimiter to be less aggressive?

ialidzhikov:
I now added 6957749070be880d854b450b60022cc4ef269349 . It adapts the shoot, infrastructure and worker controllers to use `RequeueAfter` with a constant backoff when there is a rate limit error.
I didn't take the approach with `new controller in GCM which auto-retries shoots [...]` because even today we have the issue that hibernation with schedule and maintenance activities do not happen if the gardener-controller-manager is down - see https://github.com/gardener/gardener/issues/1278 for more details. So I still don't think that it is a good approach to take Shoot reconciliation decisions outside of gardenlet.

timuthy (commented):
I wonder if the extension controllers (infra, worker) should rather exit w/o re-enqueuing and we leave it up to the Gardenlet to reconcile the shoot after a while (like here 10 Mins). Then the `Infrastructure` or `Worker` resource will be reconciled automatically, right?

ialidzhikov:
> I wonder if the extension controllers (infra, worker) should rather exit w/o re-enqueuing and we leave it up to the Gardenlet to reconcile the shoot after a while (like here 10 Mins). Then the Infrastructure or Worker resource will be reconciled automatically, right?

Hm, what about the retries in the deletion flow? As far as I see currently gardenlet only deletes the Infrastructure/Worker and waits until the resource is removed from the store. I think in this case we cannot leave it up to gardenlet and in the deletion flow we rather rely on the backoff in the extension controller.

timuthy:
> > I wonder if the extension controllers (infra, worker) should rather exit w/o re-enqueuing and we leave it up to the Gardenlet to reconcile the shoot after a while (like here 10 Mins). Then the Infrastructure or Worker resource will be reconciled automatically, right?
> 
> Hm, what about the retries in the deletion flow? As far as I see currently gardenlet only deletes the Infrastructure/Worker and waits until the resource is removed from the store. I think in this case we cannot leave it up to gardenlet and in the deletion flow we rather rely on the backoff in the extension controller.

I see, you're right. The only point I see here is, that we probably reconcile `Infrastructure` or `Worker` twice.

My theory:
1. Gardenlet triggers `Infrastructure` reconciliation in reconcile flow
2. Extension controller reconciles `Infrastructure` and hits rate limit error -> enqueue after `10 min`
3. Gardenlet exists flow and enqueues shoot after `10 min`
4. Extension controller reconciles `Infrastructure` after enqueuing from (2.)
5. Gardenlet starts shoot flow after enqueuing from (3.)
6. Gradenlet triggers another reconciliation for `Infrastructure`
7. Extension controller reconciles `Infrastructure` again

ialidzhikov:
> I see, you're right. The only point I see here is, that we probably reconcile `Infrastructure` or `Worker` twice.
> 
> My theory:
> 
> 1. Gardenlet triggers `Infrastructure` reconciliation in reconcile flow
> 2. Extension controller reconciles `Infrastructure` and hits rate limit error -> enqueue after `10 min`
> 3. Gardenlet exists flow and enqueues shoot after `10 min`
> 4. Extension controller reconciles `Infrastructure` after enqueuing from (2.)
> 5. Gardenlet starts shoot flow after enqueuing from (3.)
> 6. Gradenlet triggers another reconciliation for `Infrastructure`
> 7. Extension controller reconciles `Infrastructure` again

You are right. I can reproduce this. I was assuming that requeues that occur while an item is being processing are ignored but it is not the case. :(

ialidzhikov:
We had a discussion with @rfranzke and @timuthy on this topic. We decided to proceed with the `new controller in GCM which auto-retries shoots failed due to rate limits after some time` approach. This approach has the advantage that the Shoot will be in `Failed` state until upcoming `retry`, hence according to the extensibility contract the extension controllers won't retry the failed operation in the meantime. It should be also resilient to Pod restarts to make sure that we don't retry abnormally when GCM/gardenlet is in `CrashLoopBackOff`.

I will rework the change and probably open new PR (as this one is already long one to follow).

PR #3785, opened 2021-03-29
test resources: replace docker.io images

MartinWeindel:
…/3rd images

/area testing
/kind technical-debt

Some deployment templates of the test framework still use images from docker.io which can lead to test failures because of pull rate limits. These images are replaced with copies from `eu.gcr.io/gardener-project/3rd`

Example:
```
event for os-loadtest-7f4d6df99b-bjtqp: {kubelet shoot--it--tmpsj-50g-worker-1-z1-7c4d9-t8lf4} Failed: Failed to pull image \"alpine:3.11\": rpc error: code = Unknown desc = Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"
``` 

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/test/framework/resources/templates/logger-app.yaml.tpl
+++ b/test/framework/resources/templates/logger-app.yaml.tpl
@@ -18,7 +18,7 @@ spec:
     spec:
       containers:
       - name: logger
-        image: k8s.gcr.io/logs-generator:v0.1.1
+        image: eu.gcr.io/gardener-project/3rd/k8s_gcr_io/logs-generator:v0.1.1
         args:
           - /bin/sh
           - -c
--- a/test/framework/resources/templates/network-nginx-deamonset.yaml.tpl
+++ b/test/framework/resources/templates/network-nginx-deamonset.yaml.tpl
@@ -14,11 +14,11 @@ spec:
         app: net-nginx
     spec:
       containers:
-      - image: nginx:1.17.5
+      - image: eu.gcr.io/gardener-project/3rd/nginx:1.17.5
         name: net-nginx
         ports:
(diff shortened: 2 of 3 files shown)


dguendisch (approved):
/lgtm
thanks alot!

test/framework/resources/templates/logger-app.yaml.tpl
```diff
@@ -18,7 +18,7 @@ spec:
     spec:
       containers:
       - name: logger
-        image: k8s.gcr.io/logs-generator:v0.1.1
+        image: eu.gcr.io/gardener-project/3rd/k8s_gcr_io/logs-generator:v0.1.1
```
timebertt, line 21: Hmm, `k8s.gcr.io/logs-generator:v0.1.1` is already on GCR. Do we need to copy it?

timebertt (commented):
Thanks!

test/framework/resources/templates/logger-app.yaml.tpl
MartinWeindel: Good question. I only did it because there are copies of other images of k8s.gcr.io in `eu.gcr.io/gardener-project/3rd/k8s_gcr_io`. What's the reason for these copies then?

timebertt: > What's the reason for these copies then?

I don't know. But I guess, it doesn't matter anyways...

timebertt (approved):
/lgtm

Issue #3765, opened 2021-03-24
Support envtests with gardener resources


timebertt:
**How to categorize this issue?**

/area dev-productivity testing
/kind enhancement

**What would you like to be added**:

Support for writing integration tests with controller-runtime's `envtest` package that can use gardener API resources.

A promising approach is to add a wrapper for `envtest`, that is spinning up and registering g-api in the test environment.

**Why is this needed**:

`envtest` is great for writing integration tests that can be executed without any infra against a temporary control plane.
Though, we currently can't use gardener resources, as `envtest` primarily only supports `CRDs` but not extension API servers.

### Steps

- [x] add gardener envtest package ➡️ https://github.com/gardener/gardener/pull/3796
- [x] adding an example integration test suite for a controller ➡️ https://github.com/gardener/gardener/pull/3925, see https://github.com/gardener/gardener/blob/278e9b4e584404e14ec1b0d214987a19f4352eb0/pkg/controllermanager/controller/shoot/shoot_retry_control_test.go#L44 also see https://github.com/gardener/gardener/issues/4655
- [x] adding an example integration test suite for a webhook (e.g. GAC audit policy webhook)
- [x] ~contributing upstream and get rid of our fork~ not required anymore with https://github.com/kubernetes-sigs/controller-runtime/pull/1449#issuecomment-863049761
- [x] disable race detector for envtests (see https://github.com/gardener/gardener/issues/4036) ➡️ https://github.com/gardener/gardener/pull/4265
- [x] ~~retry starting g-api on failure to further improve stability during resource contention in pipelines~~ ➡️ not observed in prow anymore, postponed until observed again

timebertt:
Minimum required changes for this scenario in upstream `envtest` package are proposed in https://github.com/kubernetes-sigs/controller-runtime/pull/1449.

rfranzke:
@timebertt can you outline the next steps here so that everybody is aware? Is it only about contributing the required changes upstream and getting rid of our fork again, or is there more?

timebertt:
I updated the issue description with the next steps. However, I'm not working on any of those tasks currently.
/unassign
Others are welcome to contribute to this :)

timebertt:
All tasks are done now!
/close

PR #3796, opened 2021-03-30
Add Gardener Envtest

timebertt:

/area testing dev-productivity
/kind enhancement

This PR adds `pkg/envtest` which has a wrapper around controller-runtime's `envtest` package.
This wrapper can be used analogously to the upstream package to start a temporary control plane for integration tests (e.g. controllers, webhooks, etc.), but additionally starts, configures and registers `gardener-apiserver`, so that tests and the tested code can work on gardener resources.
The tests in `pkg/envtest` demonstrate how the test environment can be started und used.

Part of https://github.com/gardener/gardener/issues/3765

There are several prefactoring/cleanup commits before the actual implementation to make reviewing faster.
/squash

This PR consciously does not address:
- adding unit tests for the `envtest` package itself
- adding some developer documentation on how to use the package

The PR is mainly meant to give a first glimpse of how such a test environment could look like and to give a starting point to get some first experiences with it.

⏳ It depends on https://github.com/kubernetes-sigs/controller-runtime/pull/1449, which contains required changes in the upstream envtest package.
Depending on the feedback there and in this PR, I can either drop vendoring my fork or replace it via some other workaround.

**Release note**:

```feature developer
A new package `pkg/envtest` has been added, which can be used to bootstrap a temporary Kubernetes control plane including `gardener-apiserver` in integration tests. With this, developers can start writing integration tests for controllers, webhooks and so on that work on Gardener API resources.
```


timebertt:
cc @rfranzke

rfranzke:
```
> Generate / Vendor Check
>> make generate
>> make revendor
 M go.mod
 M vendor/modules.txt
make: *** [Makefile:165: check-generate] Error 1
```

go.mod
```diff
@@ -80,5 +80,6 @@ replace (
 	k8s.io/component-base => k8s.io/component-base v0.20.2
 	k8s.io/helm => k8s.io/helm v2.13.1+incompatible
 	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.20.2
+	sigs.k8s.io/controller-runtime => github.com/timebertt/controller-runtime v0.6.1-0.20210324082610-55b965eb86b1
```
rfranzke: Do you want to switch to https://github.com/gardener/controller-runtime?

pkg/envtest/apiserver.go
```diff
@@ -0,0 +1,443 @@
...
+	Err io.Writer
+	// HealthCheckEndpoint is the path of the healthcheck endpoint (defaults to "/healthz").
+	// It will be polled until receiving http.StatusOK (or StartTimeout occurs), before
+	// returning from Start.
+	HealthCheckEndpoint string
+
+	// terminateFunc holds a func that will terminate this GardenerAPIServer.
+	terminateFunc func()
+	// exited is a channel that will be closed, when this GardenerAPIServer exits.
+	exited chan struct{}
```
rfranzke: Cosmetics:

Exported and unexported attributes are mixed here, can you rather create dedicated sections (first all exported, then all unexported)?

pkg/envtest/apiserver.go
```diff
@@ -0,0 +1,443 @@
...
+}
+
+// Start brings up the GardenerAPIServer, waits for it to be healthy and registers Gardener's APIs.
+func (a *GardenerAPIServer) Start() error {
```
rfranzke: Cosmetics: Shouldn't this rather be

```suggestion
func (g *GardenerAPIServer) Start() error {
```

timebertt:
Thanks, done.

pkg/envtest/apiserver.go
```diff
@@ -0,0 +1,443 @@
...
+	} else {
+		if err := a.runAPIServerInProcess(); err != nil {
+			return err
+		}
+	}
+
+	startCtx, cancel := context.WithTimeout(context.Background(), a.StartTimeout)
+	defer cancel()
+
+	// TODO: retry starting GardenerAPIServer on failure
```
rfranzke, line 125: Hm, why? Is the c-r envtest doing this as well? Why should it fail (I assume you want to tackle sporadic failures here)? 👀

pkg/envtest/apiserver.go
```diff
@@ -0,0 +1,443 @@
...
+	}
+	return err
+}
+
+var allGardenerAPIGroupVersions = []schema.GroupVersion{
+	gardencorev1beta1.SchemeGroupVersion,
+	gardencorev1alpha1.SchemeGroupVersion,
+	settingsv1alpha1.SchemeGroupVersion,
+	seedmanagementv1alpha1.SchemeGroupVersion,
+}
```
rfranzke: Minor: Does it make sense to maintain this here: https://github.com/gardener/gardener/tree/master/pkg/api?

pkg/envtest/apiserver.go
```diff
@@ -0,0 +1,443 @@
...
+		if err := c.Create(ctx, apiService); err != nil {
+			return err
+		}
+	}
+
+	// wait for all the APIServices to be available
+	if err := retry.Until(ctx, waitPollInterval, func(ctx context.Context) (bool, error) {
+		for _, apiService := range allAPIServices {
+			if err := c.Get(ctx, client.ObjectKeyFromObject(apiService), apiService); err != nil {
+				return retry.MinorError(err)
```
rfranzke, line 346: Shouldn't this be a SevereError?

pkg/envtest/environment_test.go
```diff
@@ -0,0 +1,61 @@
...
+	})
+
+	It("should be able to manipulate resource from core.gardener.cloud/v1beta1", func() {
+		project := &gardencorev1beta1.Project{ObjectMeta: metav1.ObjectMeta{Name: "foo"}}
+		Expect(testClient.Create(ctx, project)).To(Succeed())
+		Expect(testClient.Get(ctx, client.ObjectKeyFromObject(project), project)).To(Succeed())
+		Expect(common.ConfirmDeletion(ctx, testClient, project)).To(Succeed())
+		Expect(testClient.Delete(ctx, project)).To(Succeed())
+	})
+	It("should be able to manipulate resource from core.gardener.cloud/v1alpha1", func() {
```
rfranzke, line 47: Cosmetics/readability:

```suggestion

	It("should be able to manipulate resource from core.gardener.cloud/v1alpha1", func() {
```

Same below

pkg/envtest/port.go
```diff
@@ -0,0 +1,96 @@
...
+	}()
+	resolvedHost = addr.IP.String()
+	return
+}
+
+// SuggestPort suggests an address a process can listen on. It returns
+// a tuple consisting of a free port and the hostname resolved to its IP.
+// It makes sure that new port allocated does not conflict with old ports
+// allocated within 1 minute.
+func SuggestPort(listenHost string) (port int, resolvedHost string, err error) {
```
rfranzke: Any reason for this to be exported?

pkg/envtest/apiserver.go
timebertt: Yes, c-r is also doing this (see [here](https://github.com/kubernetes-sigs/controller-runtime/blob/10ae090c1d3ac0c560dfa1a29b2517eb8d74442b/pkg/envtest/server.go#L285)). That's why I added a `TODO` here.
It might be good to tackle some test flakes. But as we don't have any experience with running these tests in our pipelines yet, I just left itup to future PRs :)
Let me know, if you think, we should have retries right away.

rfranzke: Interesting, thanks for sharing the link. I think it's fine for now, let's keep it and get some experience 👍🏻

pkg/envtest/apiserver.go
timebertt: I think, it's better to give it another try when we receive some error here. If we exit early in that case, test suites might fail sporadically under cpu pressure or whatever, although the situation might recover with a bit more patience :)
Also, a few retries shouldn't hurt when executing locally, I guess, as the default startup timeout is `20s` (`1m` if executed via `hack/test[-cover].sh`).

pkg/envtest/port.go
timebertt: Nope.

pkg/envtest/apiserver.go
rfranzke: OK, makes sense as it's for unit tests anyways 👍🏻 Thanks!

pkg/envtest/apiserver.go
timebertt: I was also thinking about putting this list in a central place.
Though, I found `pkg/api` and `pkg/apis` to be odd, as they also host types/helpers for the extensions API group, which is not hosted by g-api.
I moved it to `pkg/apiserver` now, hope this is a good fit.

go.mod
timebertt: Yes, done with 0461714099032d15fae4572065b412eb7109d60a.
We can drop it again or move to a different workaround, once we got feedback on the upstream PR.

timebertt:
@rfranzke thanks for your feedback, I addressed your suggestions.
PTAL :)

rfranzke (approved):
Love it, really nice PR ❤️

/lgtm

danielfoehrKn:
I'll have another quick look today, already scanned it.
/assign

danielfoehrKn (approved):
Thanks, looks good. I guess I'll have to start writing tests with it to gain more experience :)

PR #3799, opened 2021-03-30
Changing the image version of the test definitions

krgostev:

/area testing
/kind bug

This PR is fixing the indentation of the `image` property of the CreateManagedSeed test definition.
Also the image of the test definitions is changed to `eu.gcr.io/gardener-project/3rd/golang:1.16.2`
Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.test-defs/CreateManagedSeed.yaml
+++ b/.test-defs/CreateManagedSeed.yaml
@@ -16,4 +16,4 @@ spec:
     -shoot-name=$SHOOT_NAME
     -deploy-gardenlet=$DEPLOY_GARDENLET
     -backup-provider=$BACKUP_PROVIDER
-  image: golang:1.16.2
+  image: eu.gcr.io/gardener-project/3rd/golang:1.16.2
--- a/.test-defs/CreateManagedSeed.yaml
+++ b/.test-defs/CreateManagedSeed.yaml
@@ -16,4 +16,4 @@ spec:
     -shoot-name=$SHOOT_NAME
     -deploy-gardenlet=$DEPLOY_GARDENLET
     -backup-provider=$BACKUP_PROVIDER
-image: golang:1.16.2
+  image: golang:1.16.2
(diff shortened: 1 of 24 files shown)


.test-defs/CreateManagedSeed.yaml
```diff
@@ -16,4 +16,4 @@ spec:
     -shoot-name=$SHOOT_NAME
     -deploy-gardenlet=$DEPLOY_GARDENLET
     -backup-provider=$BACKUP_PROVIDER
-image: golang:1.16.2
+  image: golang:1.16.2
```
vpnachev: In some environments we are hit by the docker hub rate limits.
It would be better to use an image from the GCR.
```suggestion
  image: eu.gcr.io/gardener-project/3rd/golang:1.16.2
```

krgostev: Sure, then I suppose it would be good to do the same for all test definitions?

vpnachev: I do not know, it was already done for some of them with https://github.com/gardener/gardener/pull/3785.
It would be nice somebody from @gardener/test-infra-maintainers to suggest how to proceed.

schrodit: Until now we did not see any issues in the testmachinery with the docker.io images.
I think mostly because the node's image cache work great.

But to be safe for the future lets use our mirrored image for all tests.
@dguendisch any objections?

dguendisch: We do run occasionally into docker hub rate limits, therefore Martin recently switched the images used by most/all(?) tests to gcr: https://github.com/gardener/gardener/pull/3785
 
> Sure, then I suppose it would be good to do the same for all test definitions?

Yes please.

schrodit (dismissed):
/lgtm

stoyanr (approved):
/lgtm

vpnachev (approved):
/lgtm
/reviewed ok-to-test

PR #3774, opened 2021-03-25
Add Project handling to SeedAuthorizer

rfranzke:
**How to categorize this PR?**

/area security
/kind enhancement

**What this PR does / why we need it**:
This PR adds handling for `Project` resource to the `SeedAuthorizer`.

**Which issue(s) this PR fixes**:
Part of #1723 

**Special notes for your reviewer**:
Similar to #3708
/squash

On the way (not directly required for the PR), I added a field selector for `.spec.namespace`.

**Release note**:

```other operator
NONE
```

PR #3803, opened 2021-03-31
Add ShootState handling to SeedAuthorizer

rfranzke:

/area security
/kind enhancement

This PR adds handling for `ShootState` resource to the `SeedAuthorizer`.

Part of #1723 

Similar to #3708
/squash

**Release note**:

```other operator
NONE
```


pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
```diff
@@ -259,7 +259,7 @@ func (s *ShootStateControl) getShootStateFromCluster(ctx context.Context, cluste
 	}
 
 	shootState := &gardencorev1alpha1.ShootState{}
-	if err := s.k8sGardenClient.Client().Get(ctx, kutil.Key(shoot.Namespace, shoot.Name), shootState); err != nil {
+	if err := s.k8sGardenClient.APIReader().Get(ctx, kutil.Key(shoot.Namespace, shoot.Name), shootState); err != nil {
```
rfranzke: Well, honestly, I don't like this change too much. Theoretically, it can mean a lot of API requests to get the most recent version of the `ShootState`, depending on how often the extension resources in the seed cluster are changed.

Ideally, I would want to avoid `get` requests at all (and only use `patch`), but this isn't possible because the `.spec.extensions[]` field in the `ShootState` would require multiple patch-merge keys so that I can directly construct a strategic-merge patch without the risk to overwrite other data.

OTOH, the `CachedRuntimeClient` feature gate is still alpha and disabled by default, so exchanging the `Client()` with the `APIReader()` doesn't make a difference for now, so it's at least not getting worse.

Still, any ideas are welcome :)

rfranzke:
/hold
/author-action
Together with @timuthy, we decided to handle the "create" case in the SeedRestriction admission plugin.

rfranzke:
/unhold
/remove status/author-action

@timuthy, I've updated the PR according to our discussions. The `SeedRestriction` admission handler now takes care about `create` requests while the rest is handled by the `SeedAuthorizer`.

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
plkokanov: Just a note - when I was changing the Update() which was used to sync extension resource to use Patch() instead, I thought about adding a single key for the .spec.extensions[] entries that would be constructed by concatenating their name, kind and purpose, e.g.:
```
extensions:
- key: my-shoot-controlplane-exposure
  data: ...
```

But I'm not sure if a patch can be generated properly when the `data` field contains embedded arrays. Might be worthwhile to try it out.

pkg/admissioncontroller/webhooks/admission/seedrestriction/admission_test.go
```diff
@@ -16,44 +16,196 @@ package seedrestriction_test
...
+				request.Name = name
+				request.Namespace = namespace
+				request.UserInfo = seedUser
+				request.Resource = metav1.GroupVersionResource{
+					Group:    gardencorev1alpha1.SchemeGroupVersion.Group,
+					Resource: "shootstates",
+				}
+			})
+
+			DescribeTable("should have no opinion because no allowed verb",
```
timuthy: Shouldn't it be "should forbid because no allowed verb"?

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
timuthy: I'm not sure about this either. There is a [multi-fields merge key proposal](https://github.com/kubernetes/community/blob/4026287dc3a2d16762353b62ca2fe4b80682960a/contributors/design-proposals/cli/multi-fields-merge-key.md) but I don't see any progress there. Also Server-Side-Apply might help in the future.
Anyways and for the time being, what do you thing about excluding `ShootStates` from caching and using `Get()` calls, like [done for Projects](https://github.com/gardener/gardener/pull/3774/files#diff-024c58793bc7f54362711176c0772fed7d1f0c84ce61ac09c0997b2b2f325a13R298)?

timuthy (commented):
Very nice PR 🎉 I didn't find anything critical, just two comments.

pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go
rfranzke: @timuthy Done. Actually, it's not worse compared to today/the `master` branch because the `CachedRuntimeClient` is disabled by default, so there wasn't a cache before. Let's see how this goes.

timuthy (approved):
/lgtm

PR #3804, opened 2021-03-31
Deserialize custom resources with EnableStrict

### LINKED ISSUE #270: Deserialize extension resources (Infrastructure, etc.) using strict deseriazliation, including during validation

/area control-plane
/kind enhancement
/priority 3
/platform azure

Extension resources such as `Infrastructure` should be deserialized using the so-called "strict mode" that was introduced recently. In this mode, deserialization fails if there is any data that is not part of the API definition.

**Why is this needed**:

We recently had the case where the infrastructure creation failed because the user did not indent `natGateway` properly in an infrastructure resource :

```yaml
kind: Shoot
spec:
  provider:
    type: azure
    infrastructureConfig:
      kind: InfrastructureConfig
      apiVersion: azure.provider.extensions.gardener.cloud/v1alpha1
      networks:
        ...
      natGateway: # <--- nat gateway should be under networks
        ...
```

Changing the deserialization as described would enable us to detect and report such mistakes earlier, e.g. when the Shoot resource is created / updated, rather than much later when the infrastructure creation fails.

We would like to introduce this first for azure, later for other providers as well.

stoyanr:

/area control-plane
/kind enhancement
/priority 3

Adds the `EnableStrict` option to the codec factory of the client context used by extension controllers, to ensure that extension resource configs such as `ControlPlaneConfig` and `WorkerConfig` are always deserialized in "strict" mode. 

Fixes https://github.com/gardener/gardener-extension-provider-azure/pull/270 for `ControlPlaneConfig` and `WorkerConfig` resources, see also https://github.com/gardener/gardener-extension-provider-azure/pull/271

**Release note**:

```breaking user
Extension resources configs, namely `ControlPlaneConfig` and `WorkerConfig`, are now deserialized in "strict" mode. This means that deserializing resources with fields that are not allowed by the API schema will result in errors. Shoots containing such resources will fail with an appropriate error until you manually update the shoot to make sure any extension resources contained in it are valid. Note that due to other changes will not be able to create new shoots containing such resources, since they will be rejected by validation.
```

### DIFF
--- a/extensions/pkg/controller/common/context.go
+++ b/extensions/pkg/controller/common/context.go
@@ -37,7 +37,7 @@ type ClientContext struct {
 // NewClientConntext offers the possibility to create a ClientContext without injection.
 func NewClientContext(client client.Client, scheme *runtime.Scheme, decoder runtime.Decoder) ClientContext {
 	if decoder == nil && scheme != nil {
-		decoder = serializer.NewCodecFactory(scheme).UniversalDecoder()
+		decoder = serializer.NewCodecFactory(scheme, serializer.EnableStrict).UniversalDecoder()
 	}
 	return ClientContext{client: client, scheme: scheme, decoder: decoder}
 }
@@ -46,7 +46,7 @@ func NewClientContext(client client.Client, scheme *runtime.Scheme, decoder runt
 func (cc *ClientContext) InjectScheme(scheme *runtime.Scheme) error {
 	cc.scheme = scheme
 	if scheme != nil {
-		cc.decoder = serializer.NewCodecFactory(scheme).UniversalDecoder()
+		cc.decoder = serializer.NewCodecFactory(scheme, serializer.EnableStrict).UniversalDecoder()
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


mvladev:
Wouldn't this break existing resources which were created with such errors?

stoyanr:
@mvladev See the long discussion in https://github.com/gardener/gardener-extension-provider-azure/pull/271. In short, yes, existing resources will be broken, but this should be ok.

stoyanr:
I tested this change with revendored `provider-gcp`, `networking-calico`, and `shoot-dns-service` extensions and it seems to work as expected. Unknown fields in all extension resource configs (`InfrastructureConfig`, `ControlPlaneConfig`, `WorkerConfig`, `NetworkConfig`, `CertConfig`) cause errors during reconciliation (assuming a shoot containing such invalid resources was existing or somehow created), and all such errors are correctly categorized as `ERR_CONFIGURATION_PROBLEM`, so the reconciliation fails without retry. Removing the unknown fields and reapplying the shoot leads to successful reconciliation (perhaps after a retry).

rfranzke (approved):
/lgtm

PR #3806, opened 2021-03-31
Isto is upgraded to 1.9.2

mvladev:
**How to categorize this PR?**

/area networking
/kind enhancement

**What this PR does / why we need it**:

There are multiple CVEs (not affecting us) that were fixed in this version. Envoy is also updated.

**Which issue(s) this PR fixes**:
Fixes # n/a

**Special notes for your reviewer**:

**Release note**:

```other operator
NONE
```

PR #3810, opened 2021-04-01
Revert "Isto is upgraded to 1.9.2"

mvladev:
Reverts gardener/gardener#3806

Envoy filters have to be updated to use `v3` of the API before we upgrade.

```
2021-04-01T08:36:00.570017Z     warning envoy config    gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) 0.0.0.0_8443: The v2 xDS major version is deprecated and disabled by default. Support for v2 will be removed from Envoy at the start of Q1 2021. You may make use of v2 in Q4 2020 by following the advice in https://www.envoyproxy.io/docs/envoy/latest/faq/api/transition. (envoy.config.filter.network.tcp_proxy.v2.TcpProxy)
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -252,11 +252,11 @@ images:
 - name: istio-proxy
   sourceRepository: github.com/istio/istio
   repository: docker.io/istio/proxyv2
-  tag: "1.9.2"
+  tag: "1.8.0"
 - name: istio-istiod
   sourceRepository: github.com/istio/istio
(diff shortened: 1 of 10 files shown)


vpnachev (approved):
/lgtm

ref: https://istio.io/latest/news/releases/1.9.x/announcing-1.9/upgrade-notes/#envoyfilter-xds-v2-removal

PR #3808, opened 2021-04-01
Support single node seed clusters by adding seed-admission-controller deployment configuration

Gerrit91:
**How to categorize this PR?**

/area control-plane
/kind enhancement

**What this PR does / why we need it**:
In our test setup we can only have a single node seed cluster. However, the seed-admission-controller is deployed with strong pod anti affinity into the seed's garden namespace with a static amount of 3 replicas. This causes the managed resource for this controller to go unhealthy (the desired replica count cannot be reached).

We deal with this problem by manually adding a HPA that reduces the replica count to 1 after every seed reconciliation. This PR aims for a cleaner solution by offering a seed configuration setting to reduce the amount of replicas of the seed-admission-controller deployment.

**Which issue(s) this PR fixes**:
~

**Special notes for your reviewer**:
~

**Release note**:

```feature operator
Seed clusters with less than three nodes are now supported. In order for this to work, it is required to reduce the number of seed-admission-controller replicas through the new seed configuration for seed-admission-controller.
```

PR #3811, opened 2021-04-01
Dynamically set replicas for seed-admission-controller deployment.

Gerrit91:

/area control-plane
/kind enhancement

In our test setup we can only have a single node seed cluster. However, the seed-admission-controller is deployed with strong pod anti affinity into the seed's garden namespace with a static amount of 3 replicas. This causes the managed resource for this controller to go unhealthy (the desired replica count cannot be reached).

We deal with this problem by manually adding a HPA that reduces the replica count to 1 after every seed reconciliation. This PR aims for a cleaner solution by dynamically detecting the amount of nodes in the seed cluster, reducing the number of replicas for seed clusters with less than 3 nodes.

~

This is the follow-up of https://github.com/gardener/gardener/pull/3808.

**Release note**:

```feature operator
Seed clusters with less than three nodes are now supported. In earlier versions of Gardener the seed-admission-controller deployment was causing unhealthy managed resources for small seed clusters, preventing seed bootstrapping from succeeding.
```

### DIFF
--- a/pkg/operation/botanist/component/seedadmissioncontroller/seedadmissioncontroller.go
+++ b/pkg/operation/botanist/component/seedadmissioncontroller/seedadmissioncontroller.go
@@ -258,7 +258,6 @@ func (g *gardenerSeedAdmissionController) Deploy(ctx context.Context) error {
 			},
 		}
 
-		minAvailable        = intstr.FromInt(1)
 		podDisruptionBudget = &policyv1beta1.PodDisruptionBudget{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      Name,
@@ -266,7 +265,7 @@ func (g *gardenerSeedAdmissionController) Deploy(ctx context.Context) error {
 				Labels:    getLabels(),
 			},
 			Spec: policyv1beta1.PodDisruptionBudgetSpec{
-				MinAvailable: &minAvailable,
+				MaxUnavailable: &maxUnavailable,
 				Selector: &metav1.LabelSelector{
 					MatchLabels: getLabels(),
 				},
--- a/pkg/operation/botanist/component/seedadmissioncontroller/seedadmissioncontroller.go
+++ b/pkg/operation/botanist/component/seedadmissioncontroller/seedadmissioncontroller.go
@@ -60,6 +60,8 @@ const (
 	port            = 10250
 	volumeName      = Name + "-tls"
 	volumeMountPath = "/srv/gardener-seed-admission-controller"
+
+	defaultReplicas = int32(3)
 )
 
 // New creates a new instance of DeployWaiter for the gardener-seed-admission-controller.
@@ -85,6 +87,11 @@ type gardenerSeedAdmissionController struct {
 }
 
 func (g *gardenerSeedAdmissionController) Deploy(ctx context.Context) error {
+	replicas, err := g.getReplicas(ctx)
+	if err != nil {
+		return err
+	}
+
 	var (
 		registry = managedresources.NewRegistry(kubernetes.SeedScheme, kubernetes.SeedCodec, kubernetes.SeedSerializer)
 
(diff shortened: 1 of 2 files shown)


gardener-robot-ci-1:
Thank you @Gerrit91 for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/operation/botanist/component/seedadmissioncontroller/seedadmissioncontroller.go
```diff
@@ -174,16 +181,25 @@ func (g *gardenerSeedAdmissionController) Deploy(ctx context.Context) error {
 			},
 		}
 
-		deployment = &appsv1.Deployment{
+		// if maxUnavailable would not be set, new pods don't come up in small seed clusters
+		// (due to the pod anti affinity new pods are stuck in pending state)
+		maxUnavailable = intstr.FromInt(1)
```
vpnachev, line 186: Could this be `0` if the number of nodes is grater than the `replicas`?

Gerrit91: I think so, yes, but it also enforces a re-distribution of the pods onto other nodes. I guess, when rolling the pods one by one the `maxUnvailable: 1` configuration should not be too big of a problem regarding time of deployment. What do you think? Do you want default settings back for seed clusters > 3 nodes?

vpnachev: Thanks! I am fine with the current proposal. The other 2 replicas could keep serving the requests.

vpnachev (commented):
By the way, there is a PDB that will not work very well when the admission controller is deployed with single replica https://github.com/gardener/gardener/blob/d9ffd58cf1b557926b43a788e6cbfaf2491cad3d/pkg/operation/botanist/component/seedadmissioncontroller/seedadmissioncontroller.go#L262-L274

pkg/operation/botanist/component/seedadmissioncontroller/seedadmissioncontroller.go
```diff
@@ -174,16 +181,25 @@ func (g *gardenerSeedAdmissionController) Deploy(ctx context.Context) error {
...
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      deploymentName,
 				Namespace: g.namespace,
 				Labels:    getLabels(),
 			},
 			Spec: appsv1.DeploymentSpec{
 				RevisionHistoryLimit: pointer.Int32Ptr(1),
-				Replicas:             pointer.Int32Ptr(3),
-				Selector:             &metav1.LabelSelector{MatchLabels: getLabels()},
+				Replicas:             pointer.Int32Ptr(replicas),
```
rfranzke:
```suggestion
				Replicas:             &replicas,
```

? :)

pkg/operation/botanist/component/seedadmissioncontroller/seedadmissioncontroller.go
```diff
@@ -405,6 +421,21 @@ func init() {
 	utilruntime.Must(err)
 }
 
+func (g *gardenerSeedAdmissionController) getReplicas(ctx context.Context) (int32, error) {
+	nodeList := &corev1.NodeList{}
+	err := g.client.List(ctx, nodeList)
+	if err != nil {
+		return 0, err
+	}
```
rfranzke: As you are only interested in the number of items in the list, can you use a `PartialObjectMetadataList` like here:

https://github.com/gardener/gardener/blob/5988d5ea7bda62f66253ad473f154be814a8e5b3/pkg/admissioncontroller/webhooks/admission/internaldomainsecret/admission.go#L154-L169

rfranzke (commented):
Thanks @Gerrit91 :) For me it looks good, only https://github.com/gardener/gardener/pull/3811#pullrequestreview-626146023 is open.

vpnachev:
> Thanks @Gerrit91 :) For me it looks good, only [#3811 (review)](https://github.com/gardener/gardener/pull/3811#pullrequestreview-626146023) is open.

I think it is safe to switch from `pdb.spec.minAvailable=1` to `pdb.spec.maxUnavailable=1`.

Gerrit91:
Thanks to both of you for your reviews. I am not so experienced with PDBs and therefore I am not really sure if this configuration is perfect. Switched to `pdb.spec.maxUnavailable=1` as proposed by vpnachev.

vpnachev (approved):
/lgtm
/reviewed ok-to-test
/merge squash

Gerrit91:
Thanks a lot for merging! Very much appreciated. :)

With metal-stack we're a bit behind (currently on 1.14 release), so I hope we will soon come to the point where I can see this working in our real environments. :)

Issue #3816, opened 2021-04-02
Gardenlet panics when deployed to a managed seed


stoyanr:
**How to categorize this issue?**

/area control-plane
/kind bug
/priority 3

**What happened**:

When deployed to a managed seed via a `ManagedSeed` resource, gardenlet panics:

```
panic: need an internal domain secret but found none

goroutine 1 [running]:
k8s.io/apimachinery/pkg/util/runtime.Must(...)
	/go/src/github.com/gardener/gardener/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:171
github.com/gardener/gardener/pkg/gardenlet/controller.(*GardenletControllerFactory).Run(0xc000f079b8, 0x256c860, 0xc0006f1280, 0xc000254300, 0xc00005e800)
	/go/src/github.com/gardener/gardener/pkg/gardenlet/controller/factory.go:145 +0x1ce5
github.com/gardener/gardener/cmd/gardenlet/app.(*Gardenlet).startControllers(0xc000991720, 0x256c860, 0xc0006f1280, 0x252d101, 0x251a740)
	/go/src/github.com/gardener/gardener/cmd/gardenlet/app/gardenlet.go:493 +0x152
github.com/gardener/gardener/cmd/gardenlet/app.(*Gardenlet).Run.func1(0x256c860, 0xc0006f1280, 0x256c860, 0xc0006f07c0)
	/go/src/github.com/gardener/gardener/cmd/gardenlet/app/gardenlet.go:445 +0x65
github.com/gardener/gardener/cmd/gardenlet/app.(*Gardenlet).Run(0xc000991720, 0x256c860, 0xc00052ecc0, 0x0, 0x0)
	/go/src/github.com/gardener/gardener/cmd/gardenlet/app/gardenlet.go:475 +0x790
github.com/gardener/gardener/cmd/gardenlet/app.run(0x256c860, 0xc00052ecc0, 0xc0003c08a0, 0x1db7840, 0xc0007d5d60)
	/go/src/github.com/gardener/gardener/cmd/gardenlet/app/gardenlet.go:171 +0x10a
github.com/gardener/gardener/cmd/gardenlet/app.NewCommandStartGardenlet.func1(0xc0007ec840, 0xc00028e6a0, 0x0, 0x1, 0x0, 0x0)
	/go/src/github.com/gardener/gardener/cmd/gardenlet/app/gardenlet.go:201 +0x99
[...]

PR #3756, opened 2021-03-22
Move internal domain secret validation to gardener-admission-controller

rfranzke:
**How to categorize this PR?**

/area dev-productivity security
/kind cleanup

**What this PR does / why we need it**:
After the introduction of gardenlet, the internal domain secret validation handling was still left to be handled in a better way. With #3753, the gardenlet is restricted to only read `ConfigMap`s related to `Shoot`s for the specific `Seed`s. This doesn't play together nicely with today's `VerifyInternalDomainSecret` function.
Also, #3700 has challenges as well to properly use the `VerifyInternalDomainSecret` function, and we already discussed that it's better to move this into a specific handler in GAC (https://github.com/gardener/gardener/pull/3700#discussion_r593078316).

**Special notes for your reviewer**:
In a future release we have to delete the `garden/gardener-controller-manager-internal-config` `ConfigMap` (it's not used anymore now) - I'll take care to add it for the 1.21 release.

/squash

**Release note**:

```noteworthy operator
The `gardener-admission-controller` does now have a new handler for validating the internal domain `Secret` (earlier, there was no API validation at all).
```

PR #3819, opened 2021-04-02
Fix bug in Internal domain admission that does not honor replication of the secret

vpnachev:

/area security
/kind bug regression
/priority 2
/merge keep-commits

Fix bug in Internal domain admission that does not honor replication of the secret after https://github.com/gardener/gardener/pull/3756.
With the current master, the GCM is full with logs like 
```
time="2021-04-02T15:45:51Z" level=info msg="Error syncing Seed seed-1: failed to sync garden secrets: 1 error occurred:\n\t* admission webhook \"internal-domain-secret.gardener.cloud\" denied the request: cannot create internal domain secret because there can be only one secret with the 'internal-domain' secret role\n\n"
time="2021-04-02T15:45:58Z" level=info msg="Error syncing Seed /seed-1: failed to sync garden secrets: 1 error occurred:\n\t* admission webhook \"internal-domain-secret.gardener.cloud\" denied the request: cannot create internal domain secret because there can be only one secret with the 'internal-domain' secret role\n\n
```

There are 2 roblems:
1. the admission controller was checking for existing internal domain secret in garden namespace, while the secret can be in other namespace, e.g. seed namespace
1. in case of update or deletion of internal domain secret in a seed namespace, it should list only the shoots scheduled on this seed.

Fixes partially https://github.com/gardener/gardener/issues/3816#issuecomment-812460023

**Release note**:

```bugfix operator
A bug in the internal domain secret admission controller preventing the replication of the internal domain secrets in seed namespaces is now fixed. 
```

```bugfix operator
A bug in the internal domain secret admission controller preventing deletion of the internal domain secret replica in the seed namespace when there is no shoot scheduled on the seed is now fixed.
```

/invite @stoyanr


pkg/operation/seed/utils/utils_test.go
```diff
@@ -0,0 +1,57 @@
...
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("utils", func() {
+
+	DescribeTable("#ComputeGardenNamespace",
+		func(name, expected string) {
+			calculated := ComputeGardenNamespace(name)
+			Expect(calculated).To(Equal(expected))
+
```
ialidzhikov: Nit

```suggestion
```

pkg/operation/seed/utils/utils_test.go
```diff
@@ -0,0 +1,57 @@
...
+		},
+		Entry("empty name", "", "seed-"),
+		Entry("garden", "garden", "seed-garden"),
+		Entry("dash", "-", "seed--"),
+		Entry("garden prefixed with dash", "-garden", "seed--garden"),
+	)
+
+	DescribeTable("#ComputeSeedName",
+		func(name, expected string, expectErr bool) {
+			calculated, err := ComputeSeedName(name)
```
ialidzhikov:
```suggestion
			actual, err := ComputeSeedName(name)
```

ialidzhikov (commented):
I was able to reproduce the issue that this PR is fixing. With the current master I was not able to create a Seed that gets Ready because of the deadlock issue mentioned above. With this PR, I was able to create a Seed that gets Ready. 
Thanks for the PR!
Minor nits inline.

pkg/operation/seed/utils/utils_test.go
```diff
@@ -0,0 +1,57 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("utils", func() {
+
+	DescribeTable("#ComputeGardenNamespace",
+		func(name, expected string) {
+			calculated := ComputeGardenNamespace(name)
```
ialidzhikov:
```suggestion
			actual := ComputeGardenNamespace(name)
```

ialidzhikov (dismissed):
/lgtm

pkg/operation/seed/utils/utils.go
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: Can you rather move this to `pkg/utils/gardener/seed.go`?

pkg/admissioncontroller/webhooks/admission/internaldomainsecret/admission.go
```diff
@@ -74,15 +76,19 @@ func (h *handler) Handle(ctx context.Context, request admission.Request) admissi
 	if request.SubResource != "" {
 		return acadmission.Allowed("subresources on Secrets are not handled")
 	}
+	seedName, err := seedutils.ComputeSeedName(request.Namespace)
+	if request.Namespace != v1beta1constants.GardenNamespace && err != nil {
```
rfranzke: Personally, I would have preferred to work with an empty string if the seed name cannot be computed from the namespace instead of now having to handle an error.

pkg/admissioncontroller/webhooks/admission/internaldomainsecret/admission.go
```diff
@@ -27,9 +27,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	acadmission "github.com/gardener/gardener/pkg/admissioncontroller/webhooks/admission"
+	gardenercore "github.com/gardener/gardener/pkg/apis/core"
```
rfranzke:
```suggestion
	gardencore "github.com/gardener/gardener/pkg/apis/core"
```

pkg/admissioncontroller/webhooks/admission/internaldomainsecret/admission.go
```diff
@@ -167,3 +181,10 @@ func (h *handler) internalDomainSecretExists(ctx context.Context) (bool, error)
 
 	return len(secrets.Items) > 0, nil
 }
+
+// withSeedNameSelector extends the given selector with spec.seedName=<seedName>
+func withSeedNameSelector(listOptions []client.ListOption, seedName string) []client.ListOption {
```
rfranzke: I don't think this function is needed/it can be done inline to prevent unnecessary hopping.

rfranzke (changes requested):
/assign

Can you explain why the last commit is needed?

Also, while the separate commits are very helpful for the review, for me all of them belong together and should rather be squashed on merge, or?

vpnachev:
> Can you explain why the last commit is needed?

The problem is that the admission controller is rejecting the create request in a wrong way so that, for example,  the client here: https://github.com/gardener/gardener/blob/4e2e9d8d8cd9ec97a87f0a868090d507fa39ec01/pkg/controllermanager/controller/seed/seed_reconcile.go#L168-L171
cannot understand that the resource already exists. Well, #3821 is about to fix the client, but there might be other clients that are using `apierrors.IsAlreadyExists` to check whether the resource exists or not. Therefore, the admission can simply allow create requests for existing resources and let the apiserver reject properly.

> Also, while the separate commits are very helpful for the review, for me all of them belong together and should rather be squashed on merge, or?

No hard opinion. 
/remove merge/keep-commits

rfranzke (approved):
/lgtm

PR #3821, opened 2021-04-03
Use controllerutil.CreateOrUpdate to replicate garden secrets in seed namespace

vpnachev:

/kind enhancement

Use `controllerutil.CreateOrUpdate` to replicate garden secrets in seed namespace.
This also removes an unwanted `Create` request for already existing resources.

Fixes #

~~I need more time to fix the test~~ Done.

**Release note**:

```other operator

```


pkg/controllermanager/controller/seed/seed_reconcile.go
```diff
@@ -164,18 +162,11 @@ func (r *reconciler) syncGardenSecrets(ctx context.Context, gardenClient kuberne
...
-				_, err2 := kutil.TryUpdateSecret(ctx, gardenClient.Kubernetes(), retry.DefaultRetry, seedSecret.ObjectMeta, func(obj *corev1.Secret) (*corev1.Secret, error) {
-					obj.Annotations = seedSecret.Annotations
-					obj.Labels = seedSecret.Labels
-					obj.Data = seedSecret.Data
-					return obj, nil
-				})
-				return err2
+
+			if _, err := controllerutil.CreateOrUpdate(ctx, gardenClient.Client(), seedSecret, func() error {
+				return nil
```
rfranzke, line 167:
```suggestion
				seedSecret.Annotations = secret.Annotations
				seedSecret.Labels = secret.Labels
				seedSecret.Data = secret.Data
				return nil
```

and remove those from above `seedSecret`

rfranzke (changes requested):
/assign

pkg/controllermanager/controller/seed/seed_reconcile.go
vpnachev: Done.

vpnachev:
/status author-action

The test will most probably fail :face_with_thermometer:

pkg/controllermanager/controller/seed/seed_reconcile.go
```diff
@@ -157,25 +155,18 @@ func (r *reconciler) syncGardenSecrets(ctx context.Context, gardenClient kuberne
...
+					Namespace: namespace.Name,
 				},
-				Data: secret.Data,
 			}
-			// TODO: use controllerutil.CreateOrUpdate once cache is graduated to beta https://github.com/gardener/gardener/issues/2822
-			if err := gardenClient.Client().Create(ctx, seedSecret); err != nil {
-				if !apierrors.IsAlreadyExists(err) {
-					return err
-				}
-				_, err2 := kutil.TryUpdateSecret(ctx, gardenClient.Kubernetes(), retry.DefaultRetry, seedSecret.ObjectMeta, func(obj *corev1.Secret) (*corev1.Secret, error) {
```
rfranzke, line 172: Can this function be removed now?

vpnachev: Yes, but let's do it with some other PR.

rfranzke: Hm, why? You are already touching the code, so can't you just remove it now that nobody uses it anymore?

rfranzke (approved):
/lgtm

pkg/controllermanager/controller/seed/seed_reconcile.go
vpnachev: Because it now needs some more 30 minutes just to re-build and verify the PR. And this is just in case the CI does not fail.

Anyway, I have applied the change.

timuthy:
Actually, I deliberately chose to not use `controllerutil.CreateOrUpdate` because when the Controller-Runtime cache is enabled, all `Secrets` in the Garden cluster are cached automatically as soon as `CreateOrUpdate` is called. This especially becomes critical for large landscapes were we count plenty of `Secret` objects. Even worse, we use `secretLister`s in some places as well, so that we maintain two caches to `Secret`s.

```
// TODO: use controllerutil.CreateOrUpdate once cache is graduated to beta https://github.com/gardener/gardener/issues/2822
```

I had added the comment above to the code because once the cache was graduated, we can completely move away form the direct use of K8s informers and instead only use Controller-Runtime caches (maybe with proper label selectors) which at least eliminates duplicate caches.

Can you elaborate why we still moved to `controllerutil.CreateOrUpdate` without a doubt @vpnachev @rfranzke?

vpnachev:
I admit, I have seen the TODO, but there was the same TODO and it was migrated already to `CreateOrUpdate` [here](https://github.com/vpnachev/gardener/commit/4e11b9cec086176cf774ceb4e1f7f9aee0b4fff8#diff-32a41fa0e4b023df45cab2332e0719a7e3ed951613760df60d16d973174edac8R93-R97), therefore I took the decision to do the same for the secret as it was done for the namespace.

timuthy:
For the mentioned reasons, I prefer to include this change at a later point in time unless we can be sure that we don't run into any performance issues. WDYT?

vpnachev:
I have no hard opinion about it. If it is done, this `create` request https://github.com/gardener/gardener/blob/53170fea340bbf31012a5df9963e5493158be2c9/pkg/controllermanager/controller/seed/seed_reconcile.go#L168-L171
should be exchanged with `get` request and create the resource only if does not exist.

rfranzke:
In https://github.com/gardener/gardener/pull/3803/commits/70f42e6ea00ef2ca0a873038c4927c426d60b5a7 (#3803), I'm adding a `PatchOrCreate` function which avoids the `Get()` call. It's not yet merged, but would this be an option for this PR?

timuthy:
> should be exchanged with get request and create the resource only if does not exist.

If we did this, we could also stick to your implementation but pass the `DirectClient` for the time being.

> In [70f42e6](https://github.com/gardener/gardener/commit/70f42e6ea00ef2ca0a873038c4927c426d60b5a7) (#3803), I'm adding a `PatchOrCreate` function which avoids the `Get()` call. It's not yet merged, but would this be an option for this PR?

Hm, I'm afraid that the underlying client implementation of the Controller-Runtime still invokes a [Get call](https://github.com/kubernetes-sigs/controller-runtime/blob/11dfabf5e2e83cdb8cba28fedcda0ae5e10474f0/pkg/client/typed_client.go#L113).

timebertt:
> Hm, I'm afraid that the underlying client implementation of the Controller-Runtime still invokes a Get call.

The underlying client implementation doesn't invoke a `GET` call against the API server or cache on `Patch()`. The `cache.getObjMeta` you're referencing is retrieving some object meta (RESTMapping and so on) plus the `RESTClient` from the cache of `RESTClient`s, which doesn't involve any actual API calls (expect discovery calls depending on the RESTMapper).

PR #3802, opened 2021-03-31
[GEP-15] Bastion Management and SSH Key Pair Rotation

petersutter:
**How to categorize this PR?**

/area documentation
/kind enhancement

**What this PR does / why we need it**:
Adds a GEP for managing bastions and SSH key pair rotation.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement operator
NONE
```

PR #3771, opened 2021-03-24
[GEP-14] New cluster VPN proposal

marwinski:
**How to categorize this PR?**

/area networking
/kind enhancement

**What this PR does / why we need it**:

This pull request contains the GEP for an updated cluster VPN implementation.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```doc developer
This pull request contains the GEP for an updated cluster VPN implementation.
```

Issue #3419, opened 2021-01-15
New Gardener project role for accessing, downloading and rotating the super admin kubeconfig/secret


donistz:
**How to categorize this issue?**
/area security
/kind enhancement
/priority critical

**What would you like to be added**:
A new Gardener role that will segregate the permissions to access, download and rotate the kubeconfig/secret of the super admin user of the shoot cluster
**Why is this needed**:
There are two APIs that can be used to manage shoot cluster resources: The Gardener API and Kubernetes API.
When a shoot cluster is created all users with Gardener project role "Admin" get access to see, download and rotate the super admin kubeconfig and all other admin activities for the cluster.
To improve the default security for Kubernetes clusters management via Gardener, it is necessary to segregate in a separate role the access to the super admin kubeconfig/secret that grants full anonymous access to the shoot cluster. This way there could be gardener project members with admin access who can perform admin activities via the Gardener API but will have to be explicitly granted access to the Kubernetes API for the shoot clusters. At the moment every project admin has access to the super admin secret. The alternative to the admin role is a role called "viewer" and grants only display access via the Gardener API. At the moment the project owners cannot regulate easily the access to the super admin secret without implementing custom permissions in their project namespace and assigning them manually via the Gardener API to individual project members.
When a new project is created, this new role must be assigned always only to the Owner of the project. The Owner or a user with UAM permissions will be able to assign the new role to other members of the project. For backwards compatibility the role should be assigned also to all project members and service accounts who will have "admin" role at the moment of the implementation of this improvement. The project owner will have to decide if the role has to be remove from some of the members/service accounts later.

timuthy:
Did I get it correctly that you suggest to split the `admin` role into the following two roles?
- Admin (e.g. `ProjectOperator`) who only manages assets in the project namespace of the Garden cluster
- Admin (e.g. `ShootAdmin`) who has full access to the shoot cluster

donistz:
> Did I get it correctly that you suggest to split the `admin` role into the following two roles?
> 
> * Admin (e.g. `ProjectOperator`) who only manages assets in the project namespace of the Garden cluster
> * Admin (e.g. `ShootAdmin`) who has full access to the shoot cluster

Yes, this is absolutely what I mean.

vlerenc:
Link to current roles: https://github.com/gardener/gardener/blob/master/docs/usage/projects.md

[...]

Issue #3138, opened 2020-11-09
Deactivate the kubeconfig provided in Gardener dashboard


dimakow:
/area security
/kind enhancement
/priority normal

**What would you like to be added**:
Hello gardener team,
please provide the possibility to deactivate the admin-kubeconfig accessible in Gardener dashboard.

**Why is this needed**:
We are already creating & rotating a kubeconfig each 24h with the help of service accounts. Since we do not need the gardener provided kubeconfig, we would like to disable it completely.

rfranzke:
/touch
Now that #5220 was promoted we could look into this issue.

rfranzke:
There is a PR (https://github.com/gardener/gardener/pull/5524) refactoring the entire logic of the end-user kubeconfig handling - it makes sense to base the work on top of this PR and not on top of `master`.

PR #3823, opened 2021-04-05
[GEP-16] Dynamic kubeconfig generation for Shoot clusters GEP

mvladev:

/area security
/kind enhancement

This  is a `GEP` that introduces new  `Shoot` subresource called `AdminKubeConfigRequest` allowing for users to dynamically generate a short-lived `kubeconfigs` that can be used to access the `Shoot` cluster as `cluster-admin`.

Fixes # n/a

**Release note**:

```other operator
NONE
```

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -66,6 +66,7 @@
 * [GEP-11: Utilize API Server Network Proxy to Invert Seed-to-Shoot Connectivity](proposals/11-apiserver-network-proxy.md)
 * [GEP-12: OIDC Webhook Authenticator](proposals/12-oidc-webhook-authenticator.md)
 * [GEP-13: Automated Seed Management](proposals/13-automated-seed-management.md)
+* [GEP-16: Dynamic kubeconfig generation for Shoot clusters](proposals/16-adminkubeconfig-subresource.md)
 
 ## Development
 
--- a/docs/proposals/16-adminkubeconfig-subresource.md
+++ b/docs/proposals/16-adminkubeconfig-subresource.md
@@ -1,0 +1,126 @@
+---
+title: Dynamic kubeconfig generation for Shoot clusters
+gep-number: 16
+creation-date: 2021-04-02
+status: implementable
+authors:
+- "@mvladev"
+reviewers:
+- "@rfranzke"
+---
+
+# GEP-16: Dynamic kubeconfig generation for Shoot clusters
+
+## Table of Contents
+
+- [GEP-16: Dynamic kubeconfig generation for Shoot clusters](#gep-16-dynamic-kubeconfig-generation-for-shoot-clusters)
+  - [Table of Contents](#table-of-contents)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [Proposal](#proposal)
+  - [Alternatives](#alternatives)
+
+## Summary
+
+This `GEP` introduces new  `Shoot` subresource called `AdminKubeConfigRequest` allowing for users to dynamically generate a short-lived `kubeconfig` that can be used to access the `Shoot` cluster as `cluster-admin`.
+
+## Motivation
+
+Today, when access to the created `Shoot` clusters is needed, a `kubeconfig` with static token credentials is used. This static token is in the `system:masters` group, granting it `cluster-admin` privileges. The `kubeconfig` is generated when the cluster is reconciled, stored in `ShootState` and replicated in the `Project`'s namespace in a `Secret`. End-users can fetch the secret and use the `kubeconfig` inside it.
+
+There are several problems with this approach:
+
+- The token in the `kubeconfig` does not have any expiration, so end-users have to request a `kubeconfig` credential rotation if they want revoke the token.
+- There is no user identity in the token. e.g. if user `Joe` gets the `kubeconfig` from the `Secret`, user in that token would be `system:cluster-admin` and not `Joe` when accessing the `Shoot` cluster with it. This makes auditing events in the cluster almost impossible.
+
+### Goals
+
+- Add a `Shoot` subresource called `adminkubeconfig` that would produce a `kubeconfig` used to access that `Shoot` cluster.
+- The `kubeconfig` is not stored in the API Server, but generated for each request.
+- In the `AdminKubeConfigRequest` send to that subresource, end-users can specify the expiration time of the credential.
+- The identity (user) in the Gardener cluster would be part of the identity (x509 client certificate). E.g if `Joe` authenticates against the Gardener API server, the generated certificate for `Shoot` authentication would have the following subject:
+
+  - Common Name: `Joe`
+  - Organisation: `system:masters`
+
+- The maximum validity of the certificate can be enforced by setting a flag on the `gardener-apiserver`.
+- Deprecate and remove the old `{shoot-name}.kubeconfig` secrets in each `Project` namespace.
+
+### Non-Goals
+
+- Generate `OpenID Connect` kubeconfigs
+
+## Proposal
+
+The `gardener-apiserver` would serve a new `shoots/adminkubeconfig` resource. It can only accept `CREATE` calls and accept `AdminKubeConfigRequest`. A `AdminKubeConfigRequest` would have the following structure:
+
+```yaml
+apiVersion: authentication.gardener.cloud/v1alpha1
+kind: AdminKubeConfigRequest
+spec:
+  expirationSeconds: 3600
+```
+
+Where `expirationSeconds` is the validity of the certificate in seconds. It this case it would be `1 hour`. The maximum validity of a `AdminKubeConfigRequest` is configured by `--shoot-admin-kubeconfig-max-expiration` flag in the `gardener-apiserver`.
+
+When such request is received, the API server would find the `ShootState` associated with that cluster and generate a `kubeconfig`. The x509 client certificate would be signed by the `Shoot` cluster's CA and the user used in the subject's common name would be from the `User.Info` used to make the request.
+
+```yaml
+apiVersion: authentication.gardener.cloud/v1alpha1
+kind: AdminKubeConfigRequest
+spec:
+  expirationSeconds: 3600
+status:
+  expirationTimestamp: "2021-02-22T09:06:51Z"
+  kubeConfig: # this is normally base64-encoded, but decoded for the example
+    apiVersion: v1
+    clusters:
+    - cluster:
+        certificate-authority-data: LS0tLS1....
+        server: https://api.shoot-cluster
+      name: shoot-cluster-a
+    contexts:
+    - context:
+        cluster: shoot-cluster-a
+        user: shoot-cluster-a
+      name: shoot-cluster-a
+    current-context: shoot-cluster-a
+    kind: Config
+    preferences: {}
+    users:
+    - name: shoot-cluster-a
+      user:
+        client-certificate-data: LS0tLS1CRUd...
+        client-key-data: LS0tLS1CRUd...
+```
+
+New feature gate called `AdminKubeConfigRequest` would enable this API in the `gardener-apiserver`. The old `{shoot-name}.kubeconfig` would be kept, but deprecated and removed in the future.
+
+In order to get the server's address used in the `kubeconfig`, the Shoot's `status` should be updated with new entries:
+
+```yaml
+apiVersion: core.gardener.cloud/v1beta1
+kind: Shoot
+metadata:
+  name: crazy-botany
+  namespace: garden-dev
+spec: {}
+status:
+  advertisedAddresses:
+  - name: external
+    url: https://api.shoot-cluster.external.foo
+  - name: internal
+    url: https://api.shoot-cluster.internal.foo
+  - name: ip
+    url: https://1.2.3.4
+```
+
+This is needed, because the Gardener API server might not know on which IP address the API server is advertised on (e.g. DNS is disabled).
+
+If there are multiple entries, each would be added in a separate `cluster` in the `kubeconfig` and a `context` with the same name would be added added as well. The current context would be selected as the first entry in the `advertisedAddresses` list (`.status.advertisedAddresses[0]`).
+
+## Alternatives
+
+- [Dynamic OpenID Connect Webhook Authenticator](https://github.com/gardener/oidc-webhook-authenticator) can be used instead. Ideally cluster admins can enable either or both.


docs/README.md
```diff
@@ -66,6 +66,7 @@
 * [GEP-11: Utilize API Server Network Proxy to Invert Seed-to-Shoot Connectivity](proposals/11-apiserver-network-proxy.md)
 * [GEP-12: OIDC Webhook Authenticator](proposals/12-oidc-webhook-authenticator.md)
 * [GEP-13: Automated Seed Management](proposals/13-automated-seed-management.md)
+* [GEP-14: Dynamic kubeconfig generation for Shoot clusters](proposals/14-adminkubeconfig-subresource.md)
```
vpnachev:
```suggestion
* [GEP-16: Dynamic kubeconfig generation for Shoot clusters](proposals/16-adminkubeconfig-subresource.md)
```

There is already opened PRs for GEP-14 #3771 and GEP-15 #3802

docs/proposals/14-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,103 @@
+---
+title: Dynamic kubeconfig generation for Shoot clusters
+gep-number: 14
```
vpnachev:
```suggestion
gep-number: 16
```

docs/proposals/14-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,103 @@
...
+creation-date: 2021-04-02
+status: implementable
+authors:
+- "@mvladev"
+reviewers:
+- "@rfranzke"
+---
+
+# GEP-14: Dynamic kubeconfig generation for Shoot clusters
```
vpnachev:
```suggestion
# GEP-16: Dynamic kubeconfig generation for Shoot clusters
```

docs/proposals/14-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,103 @@
...
+The `gardener-apiserver` would serve a new `shoots/adminkubeconfig` resource. It can only accept `CREATE` calls and accept `AdminKubeConfigRequest`. A `AdminKubeConfigRequest` would have the following structure:
+
+```yaml
+apiVersion: authentication.gardener.cloud/v1alpha1
+kind: AdminKubeConfigRequest
+spec:
+  expirationSeconds: 3600
+```
+
+Where `expirationSeconds` is the validity of the certificate in seconds. It this case it would be `1 hour`. The maximum validity of a `AdminKubeConfigRequest` is configured by `--admin-kubeconfig-max-token-expiration` flag in the `gardener-apiserver`.
```
vpnachev:
```suggestion
Where `expirationSeconds` is the validity of the certificate in seconds. It this case it would be `1 hour`. The maximum validity of a `AdminKubeConfigRequest` is configured by `--shoot-admin-kubeconfig-max-token-expiration` flag in the `gardener-apiserver`.
```

docs/proposals/14-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Proposal
+
+The `gardener-apiserver` would serve a new `shoots/adminkubeconfig` resource. It can only accept `CREATE` calls and accept `AdminKubeConfigRequest`. A `AdminKubeConfigRequest` would have the following structure:
+
+```yaml
+apiVersion: authentication.gardener.cloud/v1alpha1
+kind: AdminKubeConfigRequest
+spec:
+  expirationSeconds: 3600
```
vpnachev: I am fine that OpenID is not a goal for this GEP, but it would be nice to implement the API in a way that it can easily be extended to support this case in the future.

```suggestion
spec:
  type: client-ca # In the future, a new type `oidc` can be supported
  expirationSeconds: 3600
```

vpnachev (commented):
Thanks!
Overall looks good to me.

Could you link https://github.com/gardener/gardener/issues/1433 in the GEP. There we have started a discussion about this problem. 

/title "[GEP-16] Dynamic kubeconfig generation for Shoot clusters GEP"

docs/proposals/14-adminkubeconfig-subresource.md
mvladev: I would have a different subresource for OIDC later, because the semantics there is completely different.

docs/proposals/14-adminkubeconfig-subresource.md
rfranzke: Does it make sense to differentiate between end-users and Gardener operators here?

docs/proposals/14-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,103 @@
...
+    kind: Config
+    preferences: {}
+    users:
+    - name: shoot-cluster-a
+      user:
+        client-certificate-data: LS0tLS1CRUd...
+        client-key-data: LS0tLS1CRUd...
+```
+
+New feature gate called `AdminKubeConfigRequest` would enable this API in the `gardener-apiserver`. The old `{shoot-name}.kubeconfig` would be kept, but deprecated and removed in the future.
```
rfranzke: Why is the feature gate needed? If we have one for GAPI, we probably also need one for gardenlet to control whether the old `{shoot-name}.kubeconfig` secret is synced to the project namespace.

rfranzke (commented):
/assign

Very nice, I like it a lot, thanks!
Can you also talk about whether this subresource is allowed to be used by `Project` members, or whether we need a new member role/how you envision this to be implemented?

mvladev:
> Can you also talk about whether this subresource is allowed to be used by Project members, or whether we need a new member role/how you envision this to be implemented?

Access to the subresource can be configured by the Gardener administrators via RBAC. IMO it's out of scope.

docs/proposals/14-adminkubeconfig-subresource.md
mvladev: As I mentioned - any RBAC is out of scope. This API should not be aware of those personas.

docs/proposals/14-adminkubeconfig-subresource.md
mvladev: This is a new API that and administrators could want to not enable this functionality. If there is a security issue - it's a quick way to disable it.

rfranzke: OK, fair enough.

docs/proposals/14-adminkubeconfig-subresource.md
rfranzke: > > Can you also talk about whether this subresource is allowed to be used by Project members, or whether we need a new member role/how you envision this to be implemented?
> 
> Access to the subresource can be configured by the Gardener administrators via RBAC. IMO it's out of scope.

I don't think it's out of scope. For example, there is https://github.com/gardener/gardener/issues/3138, and when end-users enable this behaviour then they need a way to get access to their clusters. The `adminkubeconfig` subresource is this way, however, we would also need means to control this, i.e., who of the project members is able to issue such admin kubeconfig. I think a new role for this makes sense. It's similar to the `uam` role that we had introduced. Without this I don't see a way to deprecate and remove the today's behaviour (always replicating the cluster-admin shoot kubeconfig in the project namespace).

rfranzke:
> > Can you also talk about whether this subresource is allowed to be used by Project members, or whether we need a new member role/how you envision this to be implemented?
> 
> Access to the subresource can be configured by the Gardener administrators via RBAC. IMO it's out of scope.

I don't think it's out of scope. For example, there is https://github.com/gardener/gardener/issues/3138, and when end-users enable this behaviour then they need a way to get access to their clusters. The `adminkubeconfig` subresource is this way, however, we would also need means to control this, i.e., who of the project members is able to issue such admin kubeconfig. I think a new role for this makes sense. It's similar to the `uam` role that we had introduced. Without this I don't see a way to deprecate and remove the today's behaviour (always replicating the cluster-admin shoot kubeconfig in the project namespace).

mvladev:
The `admin` role should be sufficient.

rfranzke:
I have some doubts about this, I would rather expect that end-users want to better control who gets access to this endpoint and can arbitrarily issue admin kubeconfigs. cc @donistz

mvladev:
> I have some doubts about this, I would rather expect that end-users want to better control who gets access to this endpoint and can arbitrarily issue admin kubeconfigs.

This is abuse of the `cluster-admin` kubeconfig - the `cluster-admin` kubeconfig should only be used by project administrators to bootstrap `RBAC`  in their cluster. End-users should authenticate via `OIDC`.

mvladev:
Using a `cluster-admin` for daily operations is really bad security practice.

rfranzke:
> > I have some doubts about this, I would rather expect that end-users want to better control who gets access to this endpoint and can arbitrarily issue admin kubeconfigs.
> 
> This is abuse of the `cluster-admin` kubeconfig - the `cluster-admin` kubeconfig should only be used by project administrators to bootstrap `RBAC` in their cluster. End-users should authenticate via `OIDC`.

Exactly, this is what I'm referring to. Concretely, some project owners want to restrict who of the members can issue such cluster-admin kubeconfig in the first place - in order to "force" them to use the OIDC. However, most project members are admins (so that they can create clusters, etc.). If you now allow all admins to access this subresource then we haven't gained too much. This is why I was proposing a new role, similar to `uam`, which can be used to control this access.

donistz:
> I don't think it's out of scope. For example, there is #3138, and when end-users enable this behaviour then they need a way to get access to their clusters. The `adminkubeconfig` subresource is this way, however, we would also need means to control this, i.e., who of the project members is able to issue such admin kubeconfig. I think a new role for this makes sense. It's similar to the `uam` role that we had introduced. Without this I don't see a way to deprecate and remove the today's behaviour (always replicating the cluster-admin shoot kubeconfig in the project namespace).

I think the "uam" role could be used for this purpose. In general,  "uam" + "admin" = "super admin" = "owner". 
The purpose of the permissions behind these short valid kubeconfigs will be to configure OIDC and initial clusterrolebindings (UAM) and when this is done such short lived super admin kubeconfigs will not be necessary.

rfranzke:
No, please let's not start mixing things which do not belong together. I just said it's similar to `uam`, but it should be a completely separate role to control this.

> and when this is done such short lived super admin kubeconfigs will not be necessary.

gardenctl v2 and the operators will use it as well, not only end-users.

donistz:
> it should be a completely separate role to control this.

The role is in fact the "Owner" role that grants full access. If we introduce another one it will be the same with a different name. I don't think this is necessary. We simply can say that who ever is the "Owner" or is granted in the Project with "Admin" + "UAM" roles will be able to generate such short lived kubeconfigs that will grant anonymous full access to ALL clusters inside the Gardener project.

rfranzke:
No, again, those are different things and I am against mixing things up here. There might be project admins which shall be allowed to create those short-lived admin kubeconfigs but not to manage the project members (or vice versa). Consequently, a dedicated project role would be necessary to reflect this.

donistz:
> project admins which shall be allowed to create those short-lived admin kubeconfigs but not to manage the project members (or vice versa)

A Project member with "uam" role can grant a person "X" (who is not a member of the project yet) the "Admin" role to the project - the effect is:  the person "X" gets "Admin access" to all clusters in the project
A Project member with "new_role_for_short_lived_cert_gen" - creates short_lived_certs for all clusters in the project and gives them to the person "X" - the effect is:  the person "X" gets "Admin access" to all clusters in the project
@rfranzke  Where is the segregation of duties?

rfranzke:
It seems that my point didn't become clear. Let me explain again.

Today, when a project member is added as `admin` then he can not only create/update/delete clusters but he can also access them by downloading the shoot-specific admin kubeconfig from the project namespace.
Tomorrow, with @mvladev's proposal, the plan is to deprecate and remove this shoot-specific admin kubeconfig from the project namespace entirely. This is to foster security by "motivating" end-users to use OIDC. 
If we allowed every project admin to create short-lived cluster-admin kubeconfigs then we wouldn't improve security much (compared to today, nothing would change).

I am thinking about projects where only a subset of the members shall be allowed to create short-lived cluster-admin kubeconfigs while the others shall still be allowed to create/update/delete clusters but are forced to use OIDC when interacting with it. This is also what #3138 is about. In order to control exactly this, we need a separate member role.

It can also be tackled when we implement #3138, for example, no problem, it's not strictly bound to this GEP. For backwards-compatibility reasons we would need to allow all project admins to access this subresource anyways, similar to how we did it with the `uam` role. I just wanted to raise some awareness about these considerations.

Do you still think it's not needed?

mvladev:
Access to that subresource be given to people who are able to `create` / `update` / `patch` a Shoot cluster. Why? Because those people can already specify an OpenID Connect provider. They can simply update the `Shoot` cluster with malicious OIDC provider, where they'll be `cluster-admins` regardless.

donistz:
> Access to that subresource be given to people who are able to `create` / `update` / `patch` a Shoot cluster. Why? Because those people can already specify an OpenID Connect provider. They can simply update the `Shoot` cluster with malicious OIDC provider, where they'll be `cluster-admins` regardless.

This is getting too complicated .. but you are right. 
So, I think that **these short-lived kubeconfigs have to grant minimum permissions (RBAC Only),** for example, because the person should ONLY create several things via the RBAC in order to allow the admins logging via OIDC to start managing their clusters.

docs/proposals/14-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,103 @@
...
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [Proposal](#proposal)
+  - [Alternatives](#alternatives)
+
+## Summary
+
+This `GEP` introduces new  `Shoot` subresource called `AdminKubeConfigRequest` allowing for users to dynamically generate a short-lived `kubeconfigs` that can be used to access the `Shoot` cluster as `cluster-admin`.
```
vlerenc: s/kubeconfigs/kubeconfig

docs/proposals/14-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Motivation
+
+Today, when access to the created `Shoot` clusters is needed, a `kubeconfig` with static token credentials is used. This static token is in the `system:masters` group, granting it `cluster-admin` privileges. The `kubeconfig` is generated when the cluster is reconciled, stored in `ShootState` and replicated in the `Project`'s namespace in a `Secret`. End-users can fetch the secret and use the `kubeconfig` inside it.
+
+There are several problems with this approach:
+
+- The token in the `kubeconfig` does not have any expiration, so end-users have to issue request a `kubeconfig` credential rotation call manually.
```
vlerenc: s/end-users have to issue request a `kubeconfig` credential rotation call manually/end-users have to request a `kubeconfig` credential rotation if they want revoke the token

vlerenc (dismissed):
Really a nice idea to solve the problem of getting temporary access. Maybe not your primary pick @mvladev (I don't know) as it's just not as slick as Google's approach (with deep IAM integration), but then again, Google is a hyperscaler, Gardener is a tool/project that might grab too much responsibility/raise the requirements bar too high with an always-on oidc-webhook-authenticator.

rfranzke:
@donistz I just stumbled across #3419 - IIUC, here you suggested exactly the same thing like me in https://github.com/gardener/gardener/pull/3823#issuecomment-814115746. Can you explain how this now relates to the GEP fo @mvladev?

donistz:
> @donistz I just stumbled across #3419 - IIUC, here you suggested exactly the same thing like me in [#3823 (comment)](https://github.com/gardener/gardener/pull/3823#issuecomment-814115746). Can you explain how this now relates to the GEP fo @mvladev?

@rfranzke Yes, my proposal was to implement two levels of administrative access (admin and operator or super admin (owner) and admin). This proposal is so far not accepted because it is difficult to "draw the line" between the responsibilities and if we want to make it customizable for the stakeholders, it will become too complicated for implementation.
Here for this GEP we are talking only about permissions to generate short-lived kubeconfig files and I don't like the idea to introduce a role only for the purpose of this resource and to keep the admin role permissions generic.
It will be good if we decide to introduce 2 levels of admin privileges and we can discuss the implementation options for enabling such feature in a phased approach.

rfranzke:
OK, as written in https://github.com/gardener/gardener/pull/3823#issuecomment-814115746, these considerations are not directly related to the GEP and it seems there is some more potential/requirement for discussing how this will look like in the end. Let's tackle this separately/as a follow-up.

vlerenc:
> OK, as written in #3823 (comment), these considerations are not directly related to the GEP and it seems there is some more potential/requirement for discussing how this will look like in the end. Let's tackle this separately/as a follow-up.

Agreed, but just for later, so that the information is not lost:

E.g. Google (see https://cloud.google.com/iam/docs/understanding-roles?hl=de#kubernetes-engine-roles) supports:

- `admin`: CRUD for GKE, cluster-admin for the clusters as such
- `viewer`: Read-Only for GKE, Read-only for the clusters as such
- `developer`: Read-Only for GKE, all but role bindings, web hooks, etc. for the clusters as such

- `clusterAdmin`: CRUD for GKE, nothing for the clusters as such
- `clusterViewer`: Read-Only for GKE, nothing for the clusters as such

Gardener is no hyperscaler and has no own "IAM sibling service", so what users can do in the clusters is something they'll have to tweak using Kubernetes native RBAC primitives in our case. So from Gardener PoV, the above GKE roles map to ours like this (https://github.com/gardener/gardener/blob/master/docs/usage/projects.md):

| GKE | Gardener |
|---|---|
| `admin` | `admin` |
| `viewer` | `viewer` |
| `developer` | `viewer` |
| `clusterAdmin` | `admin` |
| `clusterViewer` | `viewer` |

Still, there is a key difference: Obtaining access via `gcloud` or whatever doesn't give a user automatically permissions in the cluster per se, so there is no point in having additional roles or permissions in GCP limiting whether you can "download a kubeconfig" or not (it's also not offered directly). What you can do in the cluster (authorization), even if you have access (authentication), depends on the roles/permissions you have in GCP IAM. Note here especially the GCP `developer` role that gives you far-reaching permissions, but not `cluster-admin`.

Furthermore, just to complete this view into the world of the hyperscalers, while GKE supports basically only one permission per each aspect of C-R-U-D, especially only one that entitles someone to update a cluster (whatever aspect of it), AWS breaks it down into 6 list, 9 read, 2 tag, and 16 write permissions, which allows for a lot more/finer control. Then again, such a model is also a lot more effort to implement would we consider it.

> If we allowed every project admin to create short-lived cluster-admin kubeconfigs then we wouldn't improve security much (compared to today, nothing would change).

I think, we would. The problem is that the kubeconfig you can download today, may leak. The short-lived cert on the other hand is useless once it expires (pretty soon). That's why we actually want to get rid of basic auth and static tokens (and what probably is behind #3138, but of course they could have also restricted further access).

That said, I am not certain we absolutely must introduce a new role. If we stick to `admin`, these administrator can issue themselves short-lived cluster-admin kubeconfigs. UAM and viewers cannot. Side remark: This is orthogonal whether they have configured static tokens or not or OIDC or not.

However, if we want to offer a dedicated role, we would need to do somewhat more then allow or disallow requesting short-lived cluster-admin kubeconfigs, because there may be other ways to escalate privileges, e.g. modifying the OIDC configuration and everything else, now or in the future, that could be used "to get into the cluster".

And then there is the mixed approach (having a new dedicated role, but still allowing to get a short-lived kubeconfigs), similar to GKE's `developer` role, where we generally allow to get a short-lived kubeconfigs, just not cluster-admin for the new role, but specifically bound like GKE's `developer` role, i.e. all but role bindings, web hooks, etc. Then again, I am not sure I like going into that business at all - it feels messy to decide what's in and what's out and I see already some Gardener adopters call for this or that to be included or excluded. I'd rather keep it simple and if asked straight on, delay the proliferation of new roles, if possible.

mvladev:
I've updated the PR with new `status.advertisedAdddresses` field. 
/cc @rfranzke

docs/proposals/16-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,126 @@
...
+The `gardener-apiserver` would serve a new `shoots/adminkubeconfig` resource. It can only accept `CREATE` calls and accept `AdminKubeConfigRequest`. A `AdminKubeConfigRequest` would have the following structure:
+
+```yaml
+apiVersion: authentication.gardener.cloud/v1alpha1
+kind: AdminKubeConfigRequest
+spec:
+  expirationSeconds: 3600
+```
+
+Where `expirationSeconds` is the validity of the certificate in seconds. It this case it would be `1 hour`. The maximum validity of a `AdminKubeConfigRequest` is configured by `--shoot-admin-kubeconfig-max-token-expiration` flag in the `gardener-apiserver`.
```
rfranzke:
```suggestion
Where `expirationSeconds` is the validity of the certificate in seconds. It this case it would be `1 hour`. The maximum validity of a `AdminKubeConfigRequest` is configured by `--shoot-admin-kubeconfig-max-certificate-duration` flag in the `gardener-apiserver`.
```

docs/proposals/16-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,126 @@
...
+
+When such request is received, the API sever would find the `ShootState` associated with that cluster and generate a `kubeconfig`. The x509 client certificate would be signed by the `Shoot` cluster's CA and the user used in the subject's common name would be from the `User.Info` used to make the request.
```
rfranzke:
```suggestion
When such request is received, the API server would find the `ShootState` associated with that cluster and generate a `kubeconfig`. The x509 client certificate would be signed by the `Shoot` cluster's CA and the user used in the subject's common name would be from the `User.Info` used to make the request.
```

docs/proposals/16-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Proposal
+
+The `gardener-apiserver` would serve a new `shoots/adminkubeconfig` resource. It can only accept `CREATE` calls and accept `AdminKubeConfigRequest`. A `AdminKubeConfigRequest` would have the following structure:
+
+```yaml
+apiVersion: authentication.gardener.cloud/v1alpha1
+kind: AdminKubeConfigRequest
+spec:
+  expirationSeconds: 3600
```
rfranzke, line 63: What about 
```suggestion
  durationSeconds: 3600
```

docs/proposals/16-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,126 @@
...
+When such request is received, the API sever would find the `ShootState` associated with that cluster and generate a `kubeconfig`. The x509 client certificate would be signed by the `Shoot` cluster's CA and the user used in the subject's common name would be from the `User.Info` used to make the request.
+
+```yaml
+apiVersion: authentication.gardener.cloud/v1alpha1
+kind: AdminKubeConfigRequest
+spec:
+  expirationSeconds: 3600
+status:
+  expirationTimestamp: "2021-02-22T09:06:51Z"
+  kubeConfig: # this is normally base64-encoded, but decoded for the example
```
rfranzke, line 77:
```suggestion
  kubeconfig: # this is normally base64-encoded, but decoded for the example
```

(This is also how the data keys in secrets are named)

rfranzke (commented):
Thanks @mvladev, looks good, minor comments.

docs/proposals/16-adminkubeconfig-subresource.md
mvladev: I'm using the same field name as the [TokenRequest](https://github.com/kubernetes/api/blob/6eed676bc189447648fd568ad1429fa8154a8ea0/authentication/v1/types.go#L151-L155) API in Kubernetes

rfranzke: OK, fine then

docs/proposals/16-adminkubeconfig-subresource.md
mvladev: I'll rather keep it `--shoot-admin-kubeconfig-max-expiration`

In the future we might change the implementation to not use `certificates`, but some other authentication methods.

rfranzke: OK, makes sense.

docs/proposals/16-adminkubeconfig-subresource.md
mvladev: The resource is called `AdminKubeConfigRequest` - there we already have the `Config` with capital `C`.

rfranzke: Still, I would prefer to keep it lowercase in a consistent way. But we can also discuss this in the implementation PR later.

rfranzke (approved):
/lgtm

docs/proposals/16-adminkubeconfig-subresource.md
```diff
@@ -0,0 +1,126 @@
...
+```yaml
+apiVersion: authentication.gardener.cloud/v1alpha1
+kind: AdminKubeConfigRequest
+spec:
+  expirationSeconds: 3600
+```
+
+Where `expirationSeconds` is the validity of the certificate in seconds. It this case it would be `1 hour`. The maximum validity of a `AdminKubeConfigRequest` is configured by `--shoot-admin-kubeconfig-max-expiration` flag in the `gardener-apiserver`.
+
+When such request is received, the API server would find the `ShootState` associated with that cluster and generate a `kubeconfig`. The x509 client certificate would be signed by the `Shoot` cluster's CA and the user used in the subject's common name would be from the `User.Info` used to make the request.
```
timebertt, line 68: > The x509 client certificate would be signed by the `Shoot` cluster's CA

Would this be the client CA? As far as I see, we don't have a dedicated client CA yet. But would this be the right time to introduce it?

timebertt, line 68: Do you have any performance concerns for generating x509 client certs in the API server? I can imagine, that this will become quite the heavy load on the API server once it's exposed to endusers and might become a scalability limitation.
I assume, you still chose the synchronous design here in order to not persist the client creds anywhere, right?
Still, I'm asking myself if we could make this work asynchronously to not run into any performance issues in the future.

timebertt (commented):
Very nice work @mvladev, I like the GEP a lot!
Although it's already merged, I have two questions regarding the proposal (haven't had the time to check over the last 2 weeks).
Can you maybe elaborate a bit on those?

docs/proposals/16-adminkubeconfig-subresource.md
mvladev: > Would this be the client CA?

Yes

> As far as I see, we don't have a dedicated client CA yet. But would this be the right time to introduce it?

Yes we can - what we can do is create a new CA, append it to the existing one (so kubelet and other components are able be authenticated) and use it to as `--client-ca-file`.

This however is out of scope of this GEP.

mvladev: > Do you have any performance concerns for generating x509 client certs in the API server? 

I'm more concerned with the size of the `ShootState` that is all stored in the `kube-apiserver` in-memory. However, this is a separate concern.

> Still, I'm asking myself if we could make this work asynchronously to not run into any performance issues in the future.

The idea is to not store it anywhere. The API is very similar to the `TokenRequest` API in Kubernetes where also `JWT` tokens are generated on the flight. Also the fact that is created per user makes async creation impossible.

We should probably introcude proper [flow controll and API fairness](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/), but this should be done in a separate PR.

timebertt: Ok, thanks for clarifying.

> This however is out of scope of this GEP.

Should we open an issue for using dedicated CAs for client certs, then?

timebertt: Makes sense, thanks for clarifying!

PR #3832, opened 2021-04-07
Fix Seed deletion error due to uncaught NotFound

### LINKED ISSUE #3830: ResourceNotFound error on Seed deletion
Deleting a Seed can fail with the following error:
```
level=error msg="Failed to unbootstrap Seed Cluster (1 error occurred:\n\t* the server could not find the requested resource (get etcds.druid.gardener.cloud)\n\n)." seed=cluster01
```

/kind bug
/priority 3

I was able to consistently reproduce it:

1. Disable `CachedRuntimeClients` feature gate
2. Delete Seed and let the deletion artificially fail after the `etcds.druid.gardener.cloud` crd was cleaned up in the deletion flow.
3. On the next reconcilation when retrying the deletion the above error occurs (in some cases it won't, see explanation below).

I talked with @timebertt why a `NotFound` error is thrown upon deletion and not the expected `IsNoMatchError` which we already catch [here](https://github.com/gardener/gardener/blob/53170fea340bbf31012a5df9963e5493158be2c9/pkg/operation/botanist/component/etcd/bootstrap.go#L298):

Without the `CachedRuntimeClient` feature gate the typed Client is used for the [List call](https://github.com/gardener/gardener/blob/261076fc89705bb02dcc92e3595d43e56e544be1/vendor/sigs.k8s.io/controller-runtime/pkg/client/typed_client.go#L154). It relies on the [dynamicRestMapper to discover apiResources](https://github.com/gardener/gardener/blob/261076fc89705bb02dcc92e3595d43e56e544be1/vendor/sigs.k8s.io/controller-runtime/pkg/client/apiutil/dynamicrestmapper.go#L77).
The dynamicRestMapper has a cache. When the `etcds.druid.gardener.cloud` crd gets deleted the dynamicRestMapper still has the mapping to the no longer existing resource. The TypedClient performs a [get request](https://github.com/gardener/gardener/blob/261076fc89705bb02dcc92e3595d43e56e544be1/vendor/sigs.k8s.io/controller-runtime/pkg/client/typed_client.go#L154) as part of the List call which then will return a `notFound` error as the crd is no longer existing. This error can be transient because if in the mean time the DynamicRestMapper cache is updated then again an `IsNoMatchError` will be thrown.

_What happens if the CachedRuntimeClient feature gate is activated?_
Then a List call is executed using an informerCache here:
https://github.com/gardener/gardener/blob/261076fc89705bb02dcc92e3595d43e56e544be1/vendor/sigs.k8s.io/controller-runtime/pkg/cache/informer_cache.go#L72
Which will cache the result(the acutal objects) and does not rely on a DynamicRestMapper. Therefore, no NotFound error is returned in this case.

/assign

BeckerMax:
/area quality
/kind bug

see https://github.com/gardener/gardener/issues/3830

Fixes #3830 

**Release note**:

```other operator
Fix a bug which could block seed deletion due to an already deleted etcd crd
```

### DIFF
--- a/pkg/operation/botanist/component/etcd/bootstrap.go
+++ b/pkg/operation/botanist/component/etcd/bootstrap.go
@@ -36,6 +36,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


timebertt (approved):
Thanks!
/lgtm

Please open cherry-pick PRs for the last releases where applicable once this one is merged.

rfranzke (approved):

PR #3835, opened 2021-04-07
Disable etcd scale down.

amshuman-kr:

/area control-plane auto-scaling high-availability cost
/kind post-mortem

Disable etcd scale down.

Fixes https://github.tools.sap/kubernetes-live/issues-live/issues/250

There is a risk of longer downtime (also, possibly corrupted WAL files) during maintenance window in bigger clusters (if they have large fluctuations in usage pattern during `24h` period) which is triggered by too drastic a scale down during maintenance window followed by sub-optimal subsequent scale ups.

Please see here for more details -> https://github.tools.sap/kubernetes-live/issues-live/issues/250#issuecomment-202646

cc @vlerenc @istvanballok @mliepold @ialidzhikov @rfranzke @dguendisch 

**Release note**:

```other operator
Disable etcd scale down to avoid potential long down time during maintenance window.
```

### DIFF
--- a/pkg/operation/botanist/component/etcd/etcd.go
+++ b/pkg/operation/botanist/component/etcd/etcd.go
@@ -352,11 +352,11 @@ func (e *etcd) Deploy(ctx context.Context) error {
 
 	if e.hvpaConfig != nil && e.hvpaConfig.Enabled {
 		var (
-			hpaLabels                   = map[string]string{v1beta1constants.LabelRole: "etcd-hpa-" + e.role}
-			vpaLabels                   = map[string]string{v1beta1constants.LabelRole: "etcd-vpa-" + e.role}
-			updateModeAuto              = hvpav1alpha1.UpdateModeAuto
-			updateModeMaintenanceWindow = hvpav1alpha1.UpdateModeMaintenanceWindow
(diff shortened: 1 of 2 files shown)


rfranzke:
The changes lgtm, but I'm curious about the next steps now that we disable scale-down of etcd entirely. Doesn't it mean it can now grow arbitrarily large (e.g., if there is a short load spike) and nobody will scale it down?

amshuman-kr:
> but I'm curious about the next steps now that we disable scale-down of etcd entirely.

I have an idea to handle this with the new recommender by configuring relatively large scale down moving window for etcd which would safely include any normal load cycle including during shoot maintenance window (say, `24h` or even `48h`) so that the recommendations will settle to something around the peak usage over that period. If the load pattern changes over time and the peaks reduce then the scale down also will happen eventually once the scale down window moves.

But as a first step, we have made the scale down window a CLI flag (which means global configuration). This will not be sufficient for the purpose mentioned above.

Making it configurable per component will require either changes to VPA spec or we use annotations on the VPA resource. I think it is worth doing it with annotations first before taking up the VPA CRD changes with the community.

amshuman-kr:
cc @kallurbsk Something to be pursued after the first cut -> https://github.com/gardener/gardener/pull/3835#issuecomment-815058372

PR #3839, opened 2021-04-08
[GEP-13] Reconcile managed seeds on external changes to seeds and seed secrets

stoyanr:

/area control-plane
/kind enhancement

Introduces a mechanism for enqueueing managed seeds for reconciliation whenever external changes to controlled seeds and seed secrets are detected.

Fixes #

* A checksum on the so called "model" (includes only the parts of the resource that might be changed externally) is calculated and added it as an annotation when creating / updating that resource. Upon an event, the model is checked against the checksum and if there is a difference, the MS is enqueued for reconciliation.
* At the moment this has no effect on seeds created by the `gardenlet`. The `gardenlet` code adds the checksum and the MS is enqueued correctly, the MS controller redeploys `gardenlet` but the seed is not updated since `gardenlet` is not restarted, because the deployment generated by the chart is exactly the same as before. I have some ideas how I could deal with that but I would like to discuss them first.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/apis/core/v1beta1/constants/types_constants.go
+++ b/pkg/apis/core/v1beta1/constants/types_constants.go
@@ -338,6 +338,12 @@ const (
 	// `alpha.featuregates.shoot.gardener.cloud/apiserver-sni-pod-injector` annotation that disables the pod injection.
 	AnnotationShootAPIServerSNIPodInjectorDisableValue = "disable"
 
+	// AnnotationSeedIngressClass is a constant for an annotation on a Seed to specify the ingress class.
+	AnnotationSeedIngressClass = "seed.gardener.cloud/ingress-class"
+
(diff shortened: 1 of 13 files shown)


stoyanr:
After a discussion with @timebertt we decided to not merge this PR. The reasons:

* It introduces some (albeit modest) additional complexity for calculating models and model checksums. It's not clear whether the value added is really justified. Seeds / seed secrets are not in danger to be modified by end users, so inadvertent changes by operators or other controllers is the most this PR could protect against.
* We couldn't think of a good mechanism to properly update back seeds created by gardenlet (see the second bullet in the PR description) without introducing not one, but 2 gardenlet restarts for each external change, which seems excessive.
* There is now a sync period for managed seeds (defaults to 1 hour), so at most after this sync period the seed / seed secrets will be reconciled to their desired state even without this mechanism.

PR #3840, opened 2021-04-08
Let gardenlet run even without any seed resource to exist

### LINKED ISSUE #3814: Gardenlet cannot start if there is no Seed resource

/area dev-productivity 
/kind bug
/priority 3

**What happened**:
```shell
$ make start-gardenlet
time="2021-04-01T21:23:22+03:00" level=info msg="Starting Gardenlet..."
time="2021-04-01T21:23:22+03:00" level=info msg="Feature Gates: APIServerSNI=true,CachedRuntimeClients=true,HVPA=true,HVPAForShootedSeed=true,KonnectivityTunnel=false,Logging=true,ManagedIstio=true,MountHostCADirectories=false,NodeLocalDNS=true,SeedKubeScheduler=false"
time="2021-04-01T21:23:22+03:00" level=info msg="No kubeconfig secret given in the configuration under `.gardenClientConnection.kubeconfigSecret`. Skipping the kubeconfig bootstrap process and certificate rotation."
time="2021-04-01T21:23:22+03:00" level=info msg="Falling back to the kubeconfig specified in the configuration under `.gardenClientConnection.kubeconfig`"
time="2021-04-01T21:23:22+03:00" level=info msg="Creating new ClientSet for key \"garden\"" ClientMap=GardenClientMap
time="2021-04-01T21:23:22+03:00" level=info msg="No TLS server certificates provided... self-generating them now..."
time="2021-04-01T21:23:22+03:00" level=info msg="TLS server certificates successfully self-generated."
I0401 21:23:22.364548  347410 leaderelection.go:243] attempting to acquire leader lease garden/gardenlet-leader-election...
time="2021-04-01T21:23:22+03:00" level=info msg="Starting new HTTPS server on 0.0.0.0:2720"
I0401 21:23:22.368162  347410 leaderelection.go:253] successfully acquired lease garden/gardenlet-leader-election
time="2021-04-01T21:23:22+03:00" level=info msg="Acquired leadership, starting controllers."
time="2021-04-01T21:23:22+03:00" level=error msg="failed to run gardenlet controllers: no seed selected by Gardenlet
```

The gardenlet to start and wait in background till a seed resource is created.

Follow the local setup guide [here](https://github.com/gardener/gardener/blob/master/docs/development/local_setup.md#run-the-gardener) with the sample gardenlet config from [here](https://github.com/gardener/gardener/blob/b55bc408d3fe4281407986e985355e35debe9d67/example/20-componentconfig-gardenlet.yaml).

Seems to be introduced with https://github.com/gardener/gardener/pull/3700/files#diff-23f364a91783c492dd8e0ea3fc0bbea4b94272d23325336664cd92b71cf3344fR133-R135

**Environment**:

- Gardener version: master/ b55bc408d3fe4281407986e985355e35debe9d67
- Others:

vpnachev:

/area dev-productivity
/kind bug

Let gardenlet run even without any seed resource to exist

Fixes #3814

**Release note**:

```bugfix developer
A bug that prevented gardenlet to start-up when there is no seed in the garden cluster is now fixed.
```

### DIFF
--- a/pkg/gardenlet/controller/factory.go
+++ b/pkg/gardenlet/controller/factory.go
@@ -135,9 +135,9 @@ func (f *GardenletControllerFactory) Run(ctx context.Context) error {
 		return fmt.Errorf("timed out waiting for Kube caches to sync")
 	}
 
-	seeds := seedNames(f.cfg.SeedConfig, f.k8sGardenCoreInformers.Core().V1beta1().Seeds().Lister(), f.cfg.SeedSelector)
-	if len(seeds) < 1 {
-		return fmt.Errorf("no seed selected by Gardenlet")
+	gardenSecretsNamespace := v1beta1constants.GardenNamespace
+	if seeds := seedNames(f.cfg.SeedConfig, f.k8sGardenCoreInformers.Core().V1beta1().Seeds().Lister(), f.cfg.SeedSelector); len(seeds) > 0 {
+		gardenSecretsNamespace = gardenerutils.ComputeGardenNamespace(seeds[0])
 	}
 
 	// Register Seed object if desired
@@ -154,7 +154,7 @@ func (f *GardenletControllerFactory) Run(ctx context.Context) error {
 		ctx,
 		k8sGardenClient.Cache(),
 		f.k8sGardenCoreInformers.Core().V1beta1().Seeds().Lister(),
-		gardenerutils.ComputeGardenNamespace(seeds[0]),
+		gardenSecretsNamespace,
 	)
 	runtime.Must(err)
 
--- a/pkg/gardenlet/controller/factory.go
+++ b/pkg/gardenlet/controller/factory.go
@@ -25,7 +25,6 @@ import (
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencoreinformers "github.com/gardener/gardener/pkg/client/core/informers/externalversions"
-	gardencorelisters "github.com/gardener/gardener/pkg/client/core/listers/core/v1beta1"
 	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap"
 	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/keys"
 	gardenmetrics "github.com/gardener/gardener/pkg/controllerutils/metrics"
@@ -41,7 +40,6 @@ import (
 	shootcontroller "github.com/gardener/gardener/pkg/gardenlet/controller/shoot"
 	"github.com/gardener/gardener/pkg/healthz"
 	"github.com/gardener/gardener/pkg/logger"
-	"github.com/gardener/gardener/pkg/operation/garden"
 	"github.com/gardener/gardener/pkg/utils"
 	gardenerutils "github.com/gardener/gardener/pkg/utils/gardener"
 	"github.com/gardener/gardener/pkg/utils/imagevector"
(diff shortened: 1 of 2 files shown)


pkg/gardenlet/controller/factory.go
```diff
@@ -135,9 +135,9 @@ func (f *GardenletControllerFactory) Run(ctx context.Context) error {
 		return fmt.Errorf("timed out waiting for Kube caches to sync")
 	}
 
-	seeds := seedNames(f.cfg.SeedConfig, f.k8sGardenCoreInformers.Core().V1beta1().Seeds().Lister(), f.cfg.SeedSelector)
-	if len(seeds) < 1 {
-		return fmt.Errorf("no seed selected by Gardenlet")
+	gardenSecretsNamespace := v1beta1constants.GardenNamespace
```
timuthy: I'm afraid that this won't work in the future as Gardenlet will not be permitted to read secrets from the `garden` namespace.

I actually wonder if we can skip the `garden.ReadGardenSecrets` check in line 153 and handle non existing secrets gracefully (/cc @rfranzke)

vpnachev: I also have some doubts about this, but if gardenlet is not allowed to read secrets from `garden` namespace then it will simply fail with authorization error. I think this should be fine, wdyt?

timuthy: Hm, I would rather favor skipping the check. Reading `Secrets` from the `garden` namespace is something we don't expect to work, so we probably shouldn't try to do it in the first place (also for sake of code maintenance).

vpnachev: Thanks for the feedback! I have removed this secret sync now.

pkg/controllermanager/controller/seed/seed_reconcile.go
```diff
@@ -88,7 +88,7 @@ func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		},
 	}
 
-	if _, err := controllerutil.CreateOrUpdate(ctx, gardenClient.Client(), namespace, func() error {
+	if _, err := controllerutil.CreateOrUpdate(ctx, gardenClient.DirectClient(), namespace, func() error {
```
rfranzke: Shouldn't it be

```suggestion
	if _, err := controllerutil.CreateOrUpdate(ctx, gardenClient.APIReader(), namespace, func() error {
```

given that `DirectClient()` is already deprecated?

pkg/controllermanager/controller/seed/seed_reconcile.go
```diff
@@ -160,7 +160,7 @@ func (r *reconciler) syncGardenSecrets(ctx context.Context, gardenClient kuberne
 				},
 			}
 
-			if _, err := controllerutil.CreateOrUpdate(ctx, gardenClient.Client(), seedSecret, func() error {
+			if _, err := controllerutil.CreateOrUpdate(ctx, gardenClient.DirectClient(), seedSecret, func() error {
```
rfranzke: Same here

pkg/gardenlet/controller/factory.go
timuthy: Thanks for changing it.

pkg/controllermanager/controller/seed/seed_reconcile.go
vpnachev: I do not get it. `CreateOrUpdate` needs a client supporting `Update` request, but `APIReader()` supports only `Get` and `List`.

vpnachev: Looking into https://github.com/gardener/gardener/pull/3641 I think we will have to re-implement the `controllerutil.CreateOrUpdate` to make it work with reader and client. As this is out of the scope of this PR, I removed this commit.

timuthy (approved):
/lgtm

PR #3844, opened 2021-04-08
Prevent overlaps in seed/shoot networks with vpn default network range.

ScheererJ:

/area networking
/area usability
/kind technical-debt

Gardener uses a solution based on openvpn for the communication from the shoot control plane in the seed cluster to the shoot cluster (nodes, pods, services). The openvpn network uses the network range 192.168.123.0/24. To prevent hard to debug situations when either seed or shoot clusters have networks overlapping the vpn range, this pull request prevents any seed/shoot overlap in the relevant network specifications.

No bug is fixed, but it addresses a recommendation from the review of #3771 .

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/apis/core/validation/seed.go
+++ b/pkg/apis/core/validation/seed.go
@@ -114,6 +114,10 @@ func ValidateSeedSpec(seedSpec *core.SeedSpec, fldPath *field.Path, inTemplate b
 	allErrs = append(allErrs, cidrvalidation.ValidateCIDRParse(networks...)...)
 	allErrs = append(allErrs, cidrvalidation.ValidateCIDROverlap(networks, networks, false)...)
 
+	vpnDefaultRanges := []cidrvalidation.CIDR{cidrvalidation.NewCIDR(v1beta1constants.DefaultVpnRange, field.NewPath(""))}
+	allErrs = append(allErrs, cidrvalidation.ValidateCIDROverlap(vpnDefaultRanges, networks, false)...)
+	allErrs = append(allErrs, cidrvalidation.ValidateCIDROverlap(networks, vpnDefaultRanges, false)...)
+
 	if seedSpec.Backup != nil {
 		if len(seedSpec.Backup.Provider) == 0 {
 			allErrs = append(allErrs, field.Required(fldPath.Child("backup", "provider"), "must provide a backup cloud provider name"))
--- a/pkg/utils/validation/cidr/disjoint.go
+++ b/pkg/utils/validation/cidr/disjoint.go
@@ -15,8 +15,10 @@
 package cidr
 
 import (
+	"fmt"
 	"net"
 
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
@@ -33,6 +35,9 @@ func ValidateNetworkDisjointedness(fldPath *field.Path, shootNodes, shootPods, s
 	if shootNodes != nil && seedNodes != nil && NetworksIntersect(*shootNodes, *seedNodes) {
 		allErrs = append(allErrs, field.Invalid(pathNodes, *shootNodes, "shoot node network intersects with seed node network"))
 	}
+	if shootNodes != nil && NetworksIntersect(*shootNodes, v1beta1constants.DefaultVpnRange) {
+		allErrs = append(allErrs, field.Invalid(pathNodes, *shootNodes, fmt.Sprintf("shoot node network intersects with default vpn network (%s)", v1beta1constants.DefaultVpnRange)))
+	}
 
 	if shootServices != nil {
 		if NetworksIntersect(seedServices, *shootServices) {
@@ -40,6 +45,9 @@ func ValidateNetworkDisjointedness(fldPath *field.Path, shootNodes, shootPods, s
 		}
 		if NetworksIntersect(seedPods, *shootServices) {
 			allErrs = append(allErrs, field.Invalid(pathServices, *shootServices, "shoot service network intersects with seed pod network"))
+		}
+		if NetworksIntersect(v1beta1constants.DefaultVpnRange, *shootServices) {
+			allErrs = append(allErrs, field.Invalid(pathServices, *shootServices, fmt.Sprintf("shoot service network intersects with default vpn network (%s)", v1beta1constants.DefaultVpnRange)))
 		}
 	} else {
 		allErrs = append(allErrs, field.Required(pathServices, "services is required"))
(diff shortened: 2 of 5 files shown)


rfranzke:
Nice, thanks for the PR!
/assign

timuthy (commented):
What do you think about adding another intersection check for the shoot network defaults in the seed ([ref](https://github.com/gardener/gardener/blob/453b70ed9a40c0eb72694452d55d54b9d074ea29/pkg/apis/core/v1beta1/types_seed.go#L201))?

ScheererJ:
> What do you think about adding another intersection check for the shoot network defaults in the seed ([ref](https://github.com/gardener/gardener/blob/453b70ed9a40c0eb72694452d55d54b9d074ea29/pkg/apis/core/v1beta1/types_seed.go#L201))?

The checks in seed.go (https://github.com/gardener/gardener/pull/3844/files#diff-cdac4a2d9dc31c55c3905961bad1207395c8cc22394be66464ec2ac09e3a80a4) should actually also consider the shoot defaults of the seed as the defaults, if present, are added to the networks slice.

Do you think another intersection is required?

timuthy:
> Do you think another intersection is required?

No, sorry I missed that in my review. Thanks for pointing it out.

pkg/utils/validation/cidr/disjoint.go
```diff
@@ -20,6 +20,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
+const defaultVpnRange = "192.168.123.0/24"
```
rfranzke: Maybe this constant should be moved to the `pkg/apis/core/v1beta1/constants` package

ScheererJ: Should it then be reused in both files?

rfranzke: Yep :)

ScheererJ: Moved constant to constants package as requested.

pkg/apis/core/validation/seed.go
```diff
@@ -114,6 +114,10 @@ func ValidateSeedSpec(seedSpec *core.SeedSpec, fldPath *field.Path, inTemplate b
 	allErrs = append(allErrs, cidrvalidation.ValidateCIDRParse(networks...)...)
 	allErrs = append(allErrs, cidrvalidation.ValidateCIDROverlap(networks, networks, false)...)
 
+	vpnDefaultRanges := []cidrvalidation.CIDR{cidrvalidation.NewCIDR(v1beta1constants.DefaultVpnRange, field.NewPath("vpn-default"))}
```
rfranzke:
```suggestion
	vpnDefaultRanges := []cidrvalidation.CIDR{cidrvalidation.NewCIDR(v1beta1constants.DefaultVpnRange, field.NewPath(""))}
```

There is no `vpn-default` field in the API, so better leave it empty to prevent confusion.

pkg/utils/validation/cidr/disjoint.go
```diff
@@ -17,6 +17,7 @@ package cidr
 import (
 	"net"
 
+	"github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```
rfranzke:
```suggestion
	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```

pkg/utils/validation/cidr/disjoint.go
```diff
@@ -33,6 +34,9 @@ func ValidateNetworkDisjointedness(fldPath *field.Path, shootNodes, shootPods, s
 	if shootNodes != nil && seedNodes != nil && NetworksIntersect(*shootNodes, *seedNodes) {
 		allErrs = append(allErrs, field.Invalid(pathNodes, *shootNodes, "shoot node network intersects with seed node network"))
 	}
+	if shootNodes != nil && seedNodes != nil && NetworksIntersect(*shootNodes, constants.DefaultVpnRange) {
```
rfranzke:
```suggestion
	if shootNodes != nil && NetworksIntersect(*shootNodes, constants.DefaultVpnRange) {
```

pkg/utils/validation/cidr/disjoint.go
```diff
@@ -33,6 +34,9 @@ func ValidateNetworkDisjointedness(fldPath *field.Path, shootNodes, shootPods, s
...
+		allErrs = append(allErrs, field.Invalid(pathNodes, *shootNodes, "shoot node network intersects with default vpn network ("+constants.DefaultVpnRange+")"))
```
rfranzke: Use `fmt.Sprintf`?

pkg/utils/validation/cidr/disjoint.go
```diff
@@ -41,6 +45,9 @@ func ValidateNetworkDisjointedness(fldPath *field.Path, shootNodes, shootPods, s
 		if NetworksIntersect(seedPods, *shootServices) {
 			allErrs = append(allErrs, field.Invalid(pathServices, *shootServices, "shoot service network intersects with seed pod network"))
 		}
+		if NetworksIntersect(constants.DefaultVpnRange, *shootServices) {
+			allErrs = append(allErrs, field.Invalid(pathServices, *shootServices, "shoot service network intersects with default vpn network ("+constants.DefaultVpnRange+")"))
```
rfranzke: Use `fmt.Sprintf`?

pkg/utils/validation/cidr/disjoint.go
```diff
@@ -52,6 +59,9 @@ func ValidateNetworkDisjointedness(fldPath *field.Path, shootNodes, shootPods, s
 		if NetworksIntersect(seedServices, *shootPods) {
 			allErrs = append(allErrs, field.Invalid(pathPods, *shootPods, "shoot pod network intersects with seed service network"))
 		}
+		if NetworksIntersect(constants.DefaultVpnRange, *shootPods) {
+			allErrs = append(allErrs, field.Invalid(pathPods, *shootPods, "shoot pod network intersects with default vpn network ("+constants.DefaultVpnRange+")"))
```
rfranzke: Use `fmt.Sprintf`?

rfranzke (changes requested):
Can you please add some unit test cases for both packages?

pkg/utils/validation/cidr/disjoint.go
ScheererJ: Sure, I can change it accordingly. Why is there a preference for sprintf compared to "ordinary" string concatenation?

rfranzke: readability

pkg/apis/core/validation/seed.go
ScheererJ: Done.

pkg/utils/validation/cidr/disjoint.go
ScheererJ: Done.

pkg/utils/validation/cidr/disjoint.go
ScheererJ: Good find. Adapted.

pkg/utils/validation/cidr/disjoint.go
ScheererJ: Done.

pkg/utils/validation/cidr/disjoint.go
ScheererJ: Done.

pkg/utils/validation/cidr/disjoint.go
ScheererJ: Done.

rfranzke (commented):
Looks better now, thanks. Just unit tests are missing.

ScheererJ:
> Looks better now, thanks. Just unit tests are missing.

Added tests.

rfranzke (approved):
/lgtm

PR #3845, opened 2021-04-08
Do not retry shoot status updates on conflict when starting a reconcile, restore or migrate operation

plkokanov:

/area control-plane-migration
/kind bug

There is a possibility that the Shoot can be reconciled/restored while it's lastOperation is still `Migrate` which then causes the extension resources to not be annotated with `gardener.cloud/operation=restore` as that actually requires the lastOperation to be `Restore`. Instead they will be annotated with `gardener.cloud/operation=reconcile`. 

Part of #1631 

Generally this is the expected flow:
1. The migrate flow is started (this removes the control plane of the Shoot in the __Source Seed__)
2. At the end of the migrate flow the Shoot.Status is updated to contain:
```
lastOperation:
  type: Restore
  status: Pending
seedName: nil
```
3. This update causes the generation of the Shoot to increase and it is requeued for reconciliation
4. The shoot controller should then properly read the new status and Restore the Shoot's control plane in the __Destination Seed__

However during the migration flow the Shoot can be added to the reconciliation queue. Once the migration flow finishes and the Shoot is popped from the queue to be reconciled again, outdated data for the Shoot resource can be retrieved (perhaps due to caching) which leads to the following differences between what the client sees and what is actually on the server

1. The shoot controller starts the migration flow and updates the Shoot with
```
lastOperation
  type: Migrate
  status: Processing
seedName: oldSeed
```
2. The shoot controller finishes the migration flow and updates the Shoot with
```
lastOperation
  type: Restore
  status: Pending
seedName: nil
```
3. The shoot is popped for reconciliation but now the controller sees the following old value
```
lastOperation
   type: Migrate
   status: Processing
seedName: oldSeed
```
4. This causes the migration flow to be executed again
During this when the `TryUpdateStatus` at the start of the migration flow tries to update the `Shoot.Status` it first fetches the Shoot and actually gets the most recent one with the `Restore Pending` lastOperation. However, that is overwritten with the following (basically [the code after this line](https://github.com/gardener/gardener/blob/453b70ed9a40c0eb72694452d55d54b9d074ea29/pkg/gardenlet/controller/shoot/shoot_control.go#L587)):
```
lastOperation:
  type: Migrate
  status: Processing
seedName: nil
```
Notice that during this update the seedName is kept to nil as we do not change it when a Migration operation starts.
5. At the end of the migration the Status is again updated with
```
lastOperation:
   type: Restore
   status: Pending
seedName: nil
```
6. During this time, It can happen that the Shoot is again requeued for reconciliation and the controller fetches old information again where the status is:
```
lastOperation:
  type: Migrate
  status: Processing
seedName: nil
```
Due to the `Status.seedName`=nil, [this check](https://github.com/gardener/gardener/blob/453b70ed9a40c0eb72694452d55d54b9d074ea29/pkg/gardenlet/controller/shoot/shoot_control.go#L218) is skipped and the shoot is Reconciled instead of being restored.

By returning an error we force the shoot to be reconciled again until it fetches the correct data from about the Shoot's status.

**Release note**:

```bugfix operator
Fixes a possible caching issue by directly returning an error when updating the `Shoot.Status` to reflect the start of a reconcile, restore or migrate operation, instead of retrying the update on conflict.
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control.go
@@ -344,8 +344,7 @@ func (c *Controller) deleteShoot(ctx context.Context, logger *logrus.Entry, gard
 	// Trigger regular shoot deletion flow.
 	c.recorder.Event(shoot, corev1.EventTypeNormal, gardencorev1beta1.EventDeleting, "Deleting Shoot cluster")
 	shootNamespace := shootpkg.ComputeTechnicalID(project.Name, shoot)
-	shoot, err = c.updateShootStatusOperationStart(ctx, gardenClient.GardenCore(), shoot, shootNamespace, operationType)
-	if err != nil {
+	if err = c.updateShootStatusOperationStart(ctx, gardenClient.Client(), shoot, shootNamespace, operationType); err != nil {
 		return reconcile.Result{}, err
 	}
 
@@ -404,6 +403,7 @@ func (c *Controller) reconcileShoot(ctx context.Context, logger *logrus.Entry, g
 		alreadyReconciledDuringThisTimeWindow      = gutil.LastReconciliationDuringThisTimeWindow(shoot)
 		regularReconciliationIsDue                 = isUpToDate && isNowInEffectiveShootMaintenanceTimeWindow && !alreadyReconciledDuringThisTimeWindow
 		reconcileAllowed                           = !failedOrIgnored && ((!reconcileInMaintenanceOnly && !confineSpecUpdateRollout(shoot.Spec.Maintenance)) || !isUpToDate || (isNowInEffectiveShootMaintenanceTimeWindow && !alreadyReconciledDuringThisTimeWindow))
+		err                                        error
 	)
 
 	if !controllerutil.ContainsFinalizer(shoot, gardencorev1beta1.GardenerName) {
@@ -452,8 +452,7 @@ func (c *Controller) reconcileShoot(ctx context.Context, logger *logrus.Entry, g
 
 	c.recorder.Event(shoot, corev1.EventTypeNormal, gardencorev1beta1.EventReconciling, "Reconciling Shoot cluster state")
 	shootNamespace := shootpkg.ComputeTechnicalID(project.Name, shoot)
-	shoot, err := c.updateShootStatusOperationStart(ctx, gardenClient.GardenCore(), shoot, shootNamespace, operationType)
-	if err != nil {
+	if err := c.updateShootStatusOperationStart(ctx, gardenClient.Client(), shoot, shootNamespace, operationType); err != nil {
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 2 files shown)


stoyanr (dismissed):
/lgtm

timuthy (dismissed):

rfranzke:
@timebertt wanted to take a look at the flow as well - should we wait for his feedback before merging this PR?

timebertt:
Now taking a look 👁️

timebertt:
/hold
Have some feedback/questions. Let's discuss later.

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -405,6 +405,10 @@ func (c *Controller) reconcileShoot(ctx context.Context, logger *logrus.Entry, g
 		reconcileAllowed                           = !failedOrIgnored && ((!reconcileInMaintenanceOnly && !confineSpecUpdateRollout(shoot.Spec.Maintenance)) || !isUpToDate || (isNowInEffectiveShootMaintenanceTimeWindow && !alreadyReconciledDuringThisTimeWindow))
 	)
 
+	if operationType == gardencorev1beta1.LastOperationTypeMigrate {
```
timebertt: As discussed in person: let's not rely on this check here and rather solve the general problem by the following approach:
in `updateShootStatusOperationStart` let's not retry the status update on conflict and instead directly return any error to requeue the shoot (with exponential backoff as usual) to make sure, the controller always starts with recent data and doesn't start any operation based on stale data.

/author-action

timebertt (changes requested):

rfranzke:
ping @plkokanov

plkokanov:
@timebertt @rfranzke I updated the PR, name and release note to better reflect the actual changes after the discussion we had.

timuthy (approved):
/lgtm

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -404,6 +403,7 @@ func (c *Controller) reconcileShoot(ctx context.Context, logger *logrus.Entry, g
 		alreadyReconciledDuringThisTimeWindow      = gutil.LastReconciliationDuringThisTimeWindow(shoot)
 		regularReconciliationIsDue                 = isUpToDate && isNowInEffectiveShootMaintenanceTimeWindow && !alreadyReconciledDuringThisTimeWindow
 		reconcileAllowed                           = !failedOrIgnored && ((!reconcileInMaintenanceOnly && !confineSpecUpdateRollout(shoot.Spec.Maintenance)) || !isUpToDate || (isNowInEffectiveShootMaintenanceTimeWindow && !alreadyReconciledDuringThisTimeWindow))
+		err                                        error
```
rfranzke, line 406: Is this needed?

rfranzke (approved):
/lgtm

pkg/gardenlet/controller/shoot/shoot_control.go
plkokanov: it' because of changing
```
shoot, err := c.updateShootStatusOperationStart(ctx, gardenClient.GardenCore(), shoot, shootNamespace, operationType)
...
```
to
```
if err := c.updateShootStatusOperationStart(ctx, gardenClient.Client(), shoot, shootNamespace, operationType); err != nil {
...
```
and [this line](https://github.com/gardener/gardener/blob/38dce592e5b2aeafda39e3ccdc928a9579190ca6/pkg/gardenlet/controller/shoot/shoot_control.go#L494):
```
o.Shoot.Info, err = c.updateShootStatusOperationSuccess(ctx, gardenClient.GardenCore(), o, o.Shoot.Info, o.Shoot.SeedNamespace, &o.Seed.Info.Name, operationType)
```

timebertt (approved):
Thanks!
/lgtm

PR #3847, opened 2021-04-08
Reconcile managed seeds via the operation annotation

### LINKED ISSUE #3724: Trigger ManagedSeed gardenlet rollout

/kind enhancement
/priority normal

Currently, there is no way to manually trigger an immediate rollout of a ManagedSeed's gardenlet.
ManagedSeed rollouts are jittered with `gardenletConfiguration.controllers.managedSeed.syncJitterPeriod`. 
But sometimes an operator wants to shortcut this delay and immediately rollout changes to a gardenlet's configuration.

The trigger could be the same like the one we already have for shoots, extensions resources and managed resources: an annotation `gardener.cloud/operation=reconcile`.

stoyanr:

/area control-plane
/kind enhancement

Adds support for reconciling `ManagedSeeds` via the `operation` annotation

Fixes #3724

**Release note**:

```other operator
It is now possible to trigger an immediate reconciliation of a `ManagedSeed` (and therefore a rollout of its `gardenlet`) by adding the annotation `gardener.cloud/operation=reconcile`.
```

### DIFF
--- a/pkg/registry/seedmanagement/managedseed/strategy.go
+++ b/pkg/registry/seedmanagement/managedseed/strategy.go
@@ -19,8 +19,10 @@ import (
 	"fmt"
 
 	"github.com/gardener/gardener/pkg/api"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	"github.com/gardener/gardener/pkg/apis/seedmanagement"
 	"github.com/gardener/gardener/pkg/apis/seedmanagement/validation"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/fields"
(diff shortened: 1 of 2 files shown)


rfranzke (approved):
/lgtm

pkg/registry/seedmanagement/managedseed/strategy.go
```diff
@@ -65,12 +67,31 @@ func (s Strategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Object)
...
+	if !apiequality.Semantic.DeepEqual(oldManagedSeed.Spec, newManagedSeed.Spec) {
+		return true
+	}
+
+	// The deletion timestamp was set
+	if oldManagedSeed.DeletionTimestamp == nil && newManagedSeed.DeletionTimestamp != nil {
+		return true
+	}
+
+	// The operation annotation was added with value "reconcile"
```
BeckerMax, line 86: Discussion regarding style:
For me comments are always a burden as they are in danger to be outdated when the code below changes but the comment was forgotten to be changed. In this case I think the function names already provide enough meaning so the comments are unnecessary. WDYT?
(this comment should not hold merging the PR back)

BeckerMax, line 86: Discussion regarding style:
For me comments are always a burden as they are in danger to be outdated when the code below changes but the comment was forgotten to be changed. In this case I think the function names already provide enough meaning so the comments are unnecessary. WDYT?
(this comment should not hold merging the PR back)

stoyanr: > comments are always a burden as they are in danger to be outdated when the code below changes but the comment was forgotten to be changed

I don't do this. If there is a comment, I update it also (perhaps I do forget sometimes, but very rarely). If there is not, I often add one. I am a "comment addict" since in more than 20 years of practice I've found comments help me much more often than mislead me. I believe the same is true for most people and the danger of comments becoming outdated and misleading is greatly overrated.

This particular piece of code is quite similar to other `mustIncreaseGeneration` functions, e.g. https://github.com/gardener/gardener/blob/ff684eab504da137bd4b8ed5670f6108f3c400d9/pkg/registry/core/shoot/strategy.go#L71-L112. They are also full of comments. So also for the sake of consistency with the rest of the code, I think the comments should be kept.

PR #3116, opened 2020-11-05
Allow updates to Shoots if spec is unchanged

timebertt:
**How to categorize this PR?**

/area quality
/kind bug
/priority critical

**What this PR does / why we need it**:
This PR adapts the `resourcereferencemanager` admission plugin to allow any updates to Shoot's if the spec wasn't changed.
This allows for example to confirm the shoot deletion, even if some referenced object (e.g. audit policy) has already been deleted.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement user
Gardener now allows to edit a Shoot's metadata (including confirming the deletion), even if some referenced object (e.g. audit policy) has already been deleted.
```

PR #3855, opened 2021-04-09
Fix audit policy webhook

timebertt:

/kind regression
/priority 1

This PR fixes several regressions in the AuditPolicy webhook in GAC (ref #3720) introduced in `g/g@v1.20.0`:

- shoot's were still mutated, even if they were already marked for deletion. When the referenced audit policy was changed during ongoing deletion, this caused the gardenlet to fail removing its finalizer, which in turn caused Shoots to hang in a `Deleted Succeeded` state forever.
- shoot's were validated/mutated, even if the spec didn't change. If for some reason, the referenced audit policy was already gone or change in an invalid manner, this prevents any updates to shoot metadata (e.g. adding the deletion confirmation annotation or removing the finalizer).
- create/update requests for shoot's in `v1alpha1` fail, because the admission webhook always tries to decode them in version `v1beta1`
I added some missing unit test cases along the way.

It also fixes one bug in GCM:

- when a referenced auditpolicy was updated, GCM is supposed to immediately start a reconciliation for the referencing shoots. It does so by sending an empty patch, relying on the auditpolicy webhook (or `resourcereference` admission plugin in older versions) to update the `resourceVersion` field in the `ConfigMapRef`. However, since https://github.com/gardener/gardener/pull/3116 updating the `resourceVersion` was only done if the shoot spec changed. This is fixed in GCM now, by explicitly updating the spec with the updated resourceVersion of the audit policy configmap.

Thanks @BeckerMax for your help and 👀 and 🧠 on this.

**Release note**:

```bugfix user
Several regressions related to the `AuditPolicy` validation are fixed.
```

### DIFF
--- a/pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
+++ b/pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
@@ -30,6 +30,7 @@ import (
 	"gomodules.xyz/jsonpatch/v2"
 	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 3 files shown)


rfranzke (approved):
/lgtm
/second-opinion
/invite @BeckerMax

timuthy (approved):
/lgtm

timebertt:
> /squash

Let's please keep the commits, they are logically self-contained.
/keep-commits

pkg/admissioncontroller/webhooks/admission/auditpolicy/admission.go
```diff
@@ -288,3 +317,12 @@ func hasAuditPolicy(shoot *gardencorev1beta1.Shoot) bool {
 		apiServerConfig.AuditConfig.AuditPolicy.ConfigMapRef != nil &&
 		len(apiServerConfig.AuditConfig.AuditPolicy.ConfigMapRef.Name) != 0
 }
+
+func hasAuditPolicyInternal(shoot *gardencore.Shoot) bool {
```
vpnachev, line 321: I would prefer to have this method only once. You can use `gardencorev1beta1.Convert_v1beta1_ShootList_To_core_ShootList` in the `admitConfigMap` method to convert the v1beta1 shoot list to the internal version. But this is only personal opinion.

timebertt: Yeah, I was thinking about doing some conversion.
Though, I chose not to do so, because having a func for both API versions seems much more efficient. Converting all shoots of a namespace everytime one of the auditpolicy configmaps is changed seemed like quite the overhead to me.

vpnachev: Thanks!

vpnachev (approved):
/lgtm

BeckerMax:
/lgtm 
discussed concerns personally ready to merge from my side

timebertt:
Thanks to all reviewers!

PR #3862, opened 2021-04-09
Drop support for Kubernetes < 1.15

rfranzke:
**How to categorize this PR?**

/area open-source dev-productivity
/kind cleanup

**What this PR does / why we need it**:
This PR drops all code related to Kubernetes versions < `1.15` as suggested in #3170. Since quite a long time we don't test or maintain these versions anymore, and given that they are quite old and also no longer maintained by the Kubernetes community, it doesn't make sense to carry to code along any longer.

This means that the minimum Kubernetes version for seed clusters is also raised to `1.15` (from `1.11`).

The `.spec.kubernetes.kubeControllerManager.horizontalPodAutoscaler.{up,down}scaleDelay` fields have been dropped because they are no longer meaningful.

**Which issue(s) this PR fixes**:
Fixes #3170

**Release note**:

```action operator
⚠️ Gardener does no longer support shoot clusters with Kubernetes versions < 1.15. With this change, the `.spec.kubernetes.kubeControllerManager.horizontalPodAutoscaler.{up,down}scaleDelay` fields have been dropped because they are no longer meaningful. Make sure to upgrade all existing clusters before upgrading to this Gardener version.
```
```action operator
⚠️ The minimum Kubernetes version for seed clusters has been raised from `v1.11` to `v1.15`. Make sure that all your registered seed clusters meet this requirement before upgrading to this Gardener version.
```

PR #3865, opened 2021-04-12
[GEP-15] add bastions.extensions.gardener.cloud CRD

xrstf:
/area ops-productivity
/kind enhancement

This PR adds the base requirements for adding the implementation for GEP-15 (Bastion Management and SSH Key Pair Rotation). The proof of concept will be on AWS, and this PR is needed to form the contract to then build the actuator in the https://github.com/gardener/gardener-extension-provider-aws

Fixes #

**Release note**:

```other operator
NONE
```


gardener-robot-ci-2:
Thank you @xrstf for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/apis/extensions/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke:
```suggestion
// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/apis/extensions/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,119 @@
...
+// BastionResource is a constant for the name of the Bastion resource.
+const BastionResource = "Bastion"
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:scope=Namespaced,path=bastions,singular=bastion
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name=IP,JSONPath=".spec.bastionIP",type=string,description="The public IP address of the temporary bastion host"
+// +kubebuilder:printcolumn:name=State,JSONPath=".status.state",type=string,description="The bastion's status."
+// +kubebuilder:printcolumn:name=Age,JSONPath=".metadata.creationTimestamp",type=date,description="creation timestamp"
```
rfranzke: I think that's not needed?

pkg/apis/extensions/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Bastion is a bastion or jump host that is dynamically created
+// to provide SSH access to shoot nodes.
+type Bastion struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object metadata.
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
```
rfranzke: I think you can remove all the `protobuf` tags, we don't generate this for the CRDs, only for the API groups served by GAPI (gardener-apiserver).

pkg/apis/extensions/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,119 @@
...
+	// Spec is the specification of this OpenIDConnect preset.
```
rfranzke:
```suggestion
	// Spec is the specification of this Bastion.
```

pkg/apis/extensions/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,119 @@
...
+	return &b.Status
+}
+
+// BastionSpec contains the specification for an SSH bastion host.
+type BastionSpec struct {
+	// DefaultSpec is a structure containing common fields used by all extension resources.
+	DefaultSpec `json:",inline"`
+	// ShootRef is the target cluster to which a tunnel should be
+	// established.
+	ShootRef corev1.LocalObjectReference `json:"shootRef" protobuf:"bytes,1,name=shootRef"`
```
rfranzke: OK, now I'm confused, I was assuming this is the `Bastion` resource in the `extensions.gardener.cloud` API group "living" in the seed cluster. The `Shoot` resources are only created in the garden cluster, so a `shootRef` doesn't make sense here?

pkg/apis/extensions/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,119 @@
...
+	// ClientIP is the external IP that needs to be whitelisted
+	// on the cloud provider side to allow SSH ingress.
+	ClientIP BastionClientIP `json:"clientIP" protobuf:"bytes,3,name=clientIP"`
+	// SeedName is the name where the referenced shoot is running.
+	// This must not be set by the user.
+	// +optional
+	SeedName string `json:"seedName,omitempty" protobuf:"bytes,4,opt,name=seedName"`
+	// ProviderType is the cloud provider used for the shoot.
+	// +optional
+	ProviderType string `json:"providerType,omitempty" protobuf:"bytes,5,opt,name=providerType"`
```
rfranzke: You have inlined `DefaultSpec` which already includes a `type` field
```suggestion
```

pkg/apis/extensions/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,119 @@
...
+	// PublicKey is the base64-encoded PGP key of the user used to
+	// encrypt the automatically created SSH keypair.
+	PublicKey string `json:"publicKey" protobuf:"bytes,2,name=publicKey"`
+	// ClientIP is the external IP that needs to be whitelisted
+	// on the cloud provider side to allow SSH ingress.
+	ClientIP BastionClientIP `json:"clientIP" protobuf:"bytes,3,name=clientIP"`
+	// SeedName is the name where the referenced shoot is running.
+	// This must not be set by the user.
+	// +optional
+	SeedName string `json:"seedName,omitempty" protobuf:"bytes,4,opt,name=seedName"`
```
rfranzke: See above - not needed as the resource already is in the seed

pkg/apis/extensions/v1alpha1/types_bastion.go
xrstf: All of them or just the Age column? With Age I was also unsure, but I copied what we had for other CRDs ^^

pkg/apis/extensions/v1alpha1/types_bastion.go
xrstf: `make generate-extensions-crds` exists for what reason then?

And yes, I was wondering how the extension CRDs move from this repo to the extensions-* repos. The AWS one does have example CRDs, but they are not spec'ed out and seem handwritten.

pkg/apis/extensions/v1alpha1/types_bastion.go
xrstf: Correct, I started with the operations API group and then switched to the extensions (i.e. buiilding bottom-up instead of top-down), so I had some fields mixed up.

pkg/apis/extensions/v1alpha1/types_bastion.go
rfranzke: Hm, I missed that we are indeed using these `kubebuilder` tags also for other resources. It seems fine for now then, resolving the comment.

pkg/apis/extensions/v1alpha1/types_bastion.go
rfranzke: `make generate-extensions-crds` generates the CRD definitions that are placed in https://github.com/gardener/gardener/tree/master/charts/seed-bootstrap/charts/extensions/templates.

> And yes, I was wondering how the extension CRDs move from this repo to the extensions-* repos. The AWS one does have example CRDs, but they are not spec'ed out and seem handwritten.

I think that's true for the example CRDs as they are only used for the stand-alone local setup of the extension (w/o Gardener). When an extension is used with Gardener then the gardenlet installs them into the seed by applying above linked chart.

xrstf: Do you want me to put the new CRD YAML file into the chart or is this done at a later point in time?

rfranzke: I think it makes sense to add it as part of this PR because it adds the APIs as well 👍🏻

xrstf: Hmm when generating the CRDs, I get them as `apiextensions/v1` already, whereas all other CRDs in that chart are still v1beta1. I therefore tried to just regenerate all CRDs and copy them over, but a few, like that for ManagedResources, was not included in the extensions group. So I'm not sure if we want to have one oddball CRD or if I messed up somehow.

Also, I manually added the `gardener.cloud/deletion-protected: "true"` label to the CRD, assuming all extensions should have it set.

xrstf: @rfranzke How do we want to solve this? :)

rfranzke: I fear that we still need the CRDs in `v1beta1` because we still support Kubernetes versions that don't have the `v1` API of the CRDs yet. 

> Also, I manually added the `gardener.cloud/deletion-protected: "true"` label to the CRD, assuming all extensions should have it set.

Sounds good 👍🏻

xrstf:
Just for adding the CRD, this should do the trick now, unless I overlooked something more ^^

rfranzke:
Looks good now, let's now wait for the merge of the proposal document in https://github.com/gardener/gardener/pull/3802.
/merge squash

xrstf:
/hold

I overlooked the BastionIP => ipAddress change when doing the CRDs.

charts/seed-bootstrap/charts/extensions/templates/crd-bastion.yaml
```diff
@@ -0,0 +1,237 @@
...
+    type: string
+  - description: The bastion's status.
+    JSONPath: .status.state
+    name: State
+    type: string
+  subresources:
+    status: {}
+{{- if semverCompare ">= 1.15-0" .Capabilities.KubeVersion.GitVersion }}
+  preserveUnknownFields: false
+{{- end }}
```
rfranzke: After #3862, the condition is no longer needed

pkg/apis/extensions/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,96 @@
...
+// to provide SSH access to shoot nodes.
+type Bastion struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object metadata.
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Spec is the specification of this Bastion.
+	Spec BastionSpec `json:"spec"`
+	// Status is the bastion's status.
+	// +optional
+	Status BastionStatus `json:"status"`
```
rfranzke:
```suggestion
	Status BastionStatus `json:"status,omitempty"`
```

pkg/apis/extensions/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// BastionIngressPolicy represents an ingress policy for SSH bastion hosts.
+type BastionIngressPolicy struct {
+	// From defines a list of sources from which ingress should be allowed.
+	From []BastionIngressPolicyFrom `json:"from"`
+}
+
+// BastionIngressPolicy represents an ingress source for a bastion host.
+type BastionIngressPolicyFrom struct {
```
rfranzke:
```suggestion
```

As discussed in Slack

pkg/apis/extensions/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,96 @@
...
+	// IPBlock defines an IP block that is allowed to access the bastion.
+	IPBlock networkingv1.IPBlock `json:"ipBlock"`
+}
+
+// BastionStatus holds the most recently observed status of the Bastion.
+type BastionStatus struct {
+	// DefaultStatus is a structure containing common fields used by all extension resources.
+	DefaultStatus `json:",inline"`
+	// IPAddress is the external IP address of the bastion host.
+	IPAddress string `json:"ipAddress"`
```
rfranzke: Needs to be changed as discussed in Slack

rfranzke (changes requested):
Had another quick check - we plan to merge #3802 tomorrow and there are probably some final adjustments that need to be reflected in the API

rfranzke:
#3802 has been merged now, this PR can be rebased and adapted according to the latest proposal

xrstf:
@rfranzke @petersutter This PR has been updated to the final state of GEP-15. The sister PR in https://github.com/gardener/gardener-extension-provider-aws/pull/318 is aligned to this PR as well.

pkg/apis/extensions/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,93 @@
...
+	return &b.Status
+}
+
+// BastionSpec contains the specification for an SSH bastion host.
+type BastionSpec struct {
+	// DefaultSpec is a structure containing common fields used by all extension resources.
+	DefaultSpec `json:",inline"`
+	// UserData is the base64-encoded user data for the bastion instance. This should
+	// contain code to provision the SSH key on the bastion instance.
+	UserData string `json:"userData"`
```
rfranzke: > UserData is the base64-encoded user data

Shouldn't this be then 
```suggestion
	UserData []byte `json:"userData"`
```

rfranzke (changes requested):
- Please also extend https://github.com/gardener/gardener/tree/master/docs#extensions with a description of the extension contract for this new resource.
- For the deletion protection, can you extend https://github.com/gardener/gardener/blob/87fb662cf339f0cd815c5eb39d223fa571a010fa/pkg/operation/botanist/component/seedadmissioncontroller/seedadmissioncontroller.go#L118-L128, https://github.com/gardener/gardener/blob/87fb662cf339f0cd815c5eb39d223fa571a010fa/pkg/operation/botanist/component/seedadmissioncontroller/seedadmissioncontroller.go#L332-L342, https://github.com/gardener/gardener/blob/87fb662cf339f0cd815c5eb39d223fa571a010fa/pkg/seedadmissioncontroller/extensioncrds_suite_test.go#L56-L64, https://github.com/gardener/gardener/blob/87fb662cf339f0cd815c5eb39d223fa571a010fa/pkg/seedadmissioncontroller/webhooks/admission/extensioncrds/admission.go#L86-L94, https://github.com/gardener/gardener/blob/87fb662cf339f0cd815c5eb39d223fa571a010fa/pkg/seedadmissioncontroller/webhooks/admission/extensioncrds/admission_test.go#L91-L99, and  https://github.com/gardener/gardener/blob/77e09dd522d7bfb682cc35d0be437201b606061b/pkg/seedadmissioncontroller/seedadmission_suite_test.go#L143-L152?

rfranzke:
Ah, and one more place: https://github.com/gardener/gardener/blob/master/example/seed-admission-controller/10-validatingwebhookconfiguration.yaml

pkg/apis/extensions/v1alpha1/types_bastion.go
xrstf: Why though? I never understood that logic. base64 is often used to make something ASCII-only that could contain special characters. So it literally "makes binary into a string", but in Go/Kubernetes, base64 is often stored as a byte array.

I mean I can of course change it, but why is []byte better than a string here? ^^

rfranzke: I guess, we initially adapted this from k/k, mainly "inspired" by `Secret`s. Indeed, there is a difference because secrets can contain arbitrary (possible non-string) data which probably does not always apply to all our base64-encoded fields.
Anyways, if it doesn't make a huge difference to you/the implementation, then let's just do it for the sake of similarity to already existing APIs such as https://github.com/gardener/gardener/blob/a22eae846187cae0d64e34cceb47d22fa9600209/pkg/apis/extensions/v1alpha1/types_worker.go#L129-L131.

xrstf: Done

docs/extensions/bastion.md
```diff
@@ -0,0 +1,42 @@
+# Contract: `Bastion` resource
+
+The Gardener project allows users to connect to Shoot worker nodes via SSH. As nodes are usually firewalled and not directly accessible from the public internet, [GEP-15](../proposals/15-manage-bastions-and-ssh-key-pair-rotation.md) introduced the concept of "Bastions". A bastion is a dedicated server that only serves to allow SSH ingress to the worker nodes.
+
+`Bastion` resources are usually created by `gardenctl` when a user wants to establish an SSH connection. They contain the user's public SSH key and IP address, in order to provision the server accordingly: The public key is put onto the Bastion and SSH ingress is only authorized for the given IP address (in fact, it's not a single IP address, but a set of IP ranges, however for most purposes a single IP is be used).
```
rfranzke: As of the latest update of the GEP, it's not the user's public SSH key but a randomly computed key. Anyways, that's an implementation detail of `gardenctl`, so for this document I would just write

> contain a public SSH key

docs/extensions/bastion.md
```diff
@@ -0,0 +1,42 @@
+# Contract: `Bastion` resource
```
rfranzke, line 1: Can you add this doc to the index (https://github.com/gardener/gardener/blob/master/docs/README.md#extensions)?

docs/extensions/bastion.md
rfranzke: Generally, as this doc just describes the "API"/extension contract from g/g's point of view, maybe let's not talk about `gardenctl` at all here?

extensions/pkg/controller/bastion/controller.go
```diff
@@ -0,0 +1,83 @@
...
+	if ignoreOperationAnnotation {
+		return []predicate.Predicate{
+			predicate.GenerationChangedPredicate{},
+		}
+	}
+
+	return []predicate.Predicate{
+		predicate.Or(
+			extensionspredicate.HasOperationAnnotation(),
+			extensionspredicate.LastOperationNotSuccessful(),
```
rfranzke, line 61: This predicate is probably not needed because the `Bastion` is not created as part of a shoot reconciliation but completely separate, so the last operation of the `Shoot` should not play a role here, or?

extensions/pkg/controller/bastion/reconciler.go
```diff
@@ -0,0 +1,195 @@
...
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	"github.com/gardener/gardener/pkg/controllerutils"
+)
+
+const (
+	// EventBastionReconciliation an event reason to describe bastion reconciliation.
+	EventBastionReconciliation string = "BastionReconciliation"
+	// EventBastionDeletion an event reason to describe bastion deletion.
+	EventBastionDeletion string = "BastionDeletion"
+)
```
rfranzke: Since https://github.com/gardener/gardener/pull/3869/commits/844d0be41f519196e58c66b438ea0a7ddb05e735 (already merged), the extension controllers no longer emit events, so let's remove them here as well.

extensions/pkg/controller/bastion/reconciler.go
```diff
@@ -0,0 +1,195 @@
...
+	})
+}
+
+func (r *reconciler) updateStatusSuccess(ctx context.Context, bastion *extensionsv1alpha1.Bastion, lastOperationType gardencorev1beta1.LastOperationType, description string) error {
+	return extensionscontroller.TryUpdateStatus(ctx, retry.DefaultBackoff, r.client, bastion, func() error {
+		bastion.Status.ObservedGeneration = bastion.Generation
+		bastion.Status.LastOperation, bastion.Status.LastError = extensionscontroller.ReconcileSucceeded(lastOperationType, description)
+		return nil
+	})
+}
```
rfranzke: See https://github.com/gardener/gardener/pull/3869 - maybe you can adapt this accordingly.

rfranzke (changes requested):
Nice @xrstf, a few follow-ups after the "rebase" need to be addressed, overall looks great!

docs/extensions/bastion.md
xrstf: I am unsure where this documentation should end and the proposal begins. I didn't want to repeat things from the proposal, otherwise 2 places might go out of sync with reality over time. ^^

rfranzke: How is this related to the suggestion not talking about one special user of this API (`gardenctl`) but only about the API in general? 👀

extensions/pkg/controller/bastion/controller.go
xrstf: Will removing this cause the requeue to not happen anymore? At least the AWS provider's controller relies on requeues to wait for the bastion to come up, for example.

extensions/pkg/controller/bastion/reconciler.go
```diff
@@ -0,0 +1,154 @@
...
+
+	cluster, err := extensionscontroller.GetCluster(ctx, r.client, bastion.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if extensionscontroller.IsFailed(cluster) {
+		r.logger.Info("Stop reconciling Bastion of failed Shoot.", "name", bastion.Name)
+		return reconcile.Result{}, nil
+	}
```
rfranzke: I guess this is not needed for the same reason why the `LastOperationFailed` predicate was removed - the `Bastion` is not "tight" to the shoot cluster reconciliation, as opposed to most other resources.

rfranzke (commented):
Very nice, one minor nit in the reconciler, other than that lgtm
/reviewed ok-to-test

extensions/pkg/controller/bastion/controller.go
rfranzke: Sorry, I made a mistake. I was wrongly thinking that the `LastOperationNotSuccessful` predicate is about the last operation of the related `Shoot` (for whatever reason, I was mixing it up with https://github.com/gardener/gardener/blob/77e09dd522d7bfb682cc35d0be437201b606061b/extensions/pkg/predicate/predicate.go#L82-L85). Please add it again.

rfranzke:
CI pipeline failed with:

```
> Generate / Vendor Check
>> make generate
 M hack/api-reference/extensions.md
```

xrstf:
I rebased and re-ran `make install-requirements`, but `make generate` consistently ends up with

```
$ make generate
topological order bytes
topological order encoding/base64
topological order encoding/json
topological order errors
topological order fmt
topological order github.com/gardener/gardener/pkg/apis/core
topological order github.com/gardener/gardener/pkg/apis/core/v1beta1/constants
topological order github.com/gardener/gardener/pkg/apis/seedmanagement
topological order github.com/gardener/gardener/pkg/apis/seedmanagement/helper
topological order github.com/gardener/gardener/pkg/apis/settings
....
topological order github.com/gardener/gardener/pkg/apis/core/v1alpha1
topological order github.com/gardener/gardener/pkg/apis/seedmanagement/v1alpha1
2021/04/28 10:52:33 google/protobuf/descriptor.proto: File not found.
github.com/gogo/protobuf/gogoproto/gogo.proto: Import "google/protobuf/descriptor.proto" was not found or had errors.
github.com/gogo/protobuf/gogoproto/gogo.proto:38:8: "google.protobuf.EnumOptions" is not defined.
github.com/gogo/protobuf/gogoproto/gogo.proto: "google.protobuf.EnumOptions" is not defined.
...
github.com/gogo/protobuf/gogoproto/gogo.proto: "google.protobuf.FieldOptions" is not defined.
github.com/gogo/protobuf/gogoproto/gogo.proto: "google.protobuf.FieldOptions" is not defined.
github.com/gogo/protobuf/gogoproto/gogo.proto: "google.protobuf.FieldOptions" is not defined.
github.com/gogo/protobuf/gogoproto/gogo.proto: "google.protobuf.FieldOptions" is not defined.
github.com/gardener/gardener/pkg/apis/core/v1beta1/generated.proto: Import "github.com/gogo/protobuf/gogoproto/gogo.proto" was not found or had errors.
github.com/gardener/gardener/pkg/apis/core/v1alpha1/generated.proto: Import "github.com/gardener/gardener/pkg/apis/core/v1beta1/generated.proto" was not found or had errors.
github.com/gardener/gardener/pkg/apis/core/v1alpha1/generated.proto: Import "github.com/gogo/protobuf/gogoproto/gogo.proto" was not found or had errors.
github.com/gardener/gardener/pkg/apis/core/v1alpha1/generated.proto:642:12: "github.com.gardener.gardener.pkg.apis.core.v1beta1.NamedResourceReference" is not defined.
github.com/gardener/gardener/pkg/apis/core/v1alpha1/generated.proto: warning: Import k8s.io/apimachinery/pkg/runtime/schema/generated.proto but not used.
2021/04/28 10:52:33 protoc -I . -I /home/xrstf/gospace/src -I ./hack/../vendor --gogo_out=/home/xrstf/gospace/src /home/xrstf/gospace/src/github.com/gardener/gardener/pkg/apis/core/v1alpha1/generated.proto
2021/04/28 10:52:33 Unable to generate protoc on github.com.gardener.gardener.pkg.apis.core.v1alpha1: exit status 1
```

@ialidzhikov Rafael mentioned that you might be able to help? I must have overlooked something obvious :-D The same errors happen if I run `make generate` on the master branch.

ialidzhikov:
I think I faced the same issue in my local setup. Back then I simply started a local container with `docker run --rm -v ${GOPATH}/src:/go/src -it golang:1.16.3 bash` and ran `make generate-sequantial` there to workaround this issue.
I am not sure what exactly causes the issue.

xrstf:
You're right,

```bash
docker run --rm -v ${GOPATH}/src:/go/src -it golang:1.16.3 \
  bash -c 'apt update && apt install unzip && cd /go/src/github.com/gardener/gardener && make install-requirements generate-sequential'
```

did the trick. :)

rfranzke (approved):
/lgtm

PR #3867, opened 2021-04-12
Clean up the workaround for extension library webhooks

ialidzhikov:
/kind cleanup

This PR removes the temporary workaround that was skipping the `ProblematicWebhooks` check for shoot webhooks (because they were previously registered with `timeout: 30s` - something that leads the webhook to be classified as problematic) and shooted seeds. See https://github.com/gardener/gardener/pull/3440 for more details.

**Release note**:

```breaking operator
The temporary workaround in the `ProblematicWebhooks` check that was skipping Shoot webhooks is now removed. Before updating to this version of Gardener, please make sure that the provider extensions in the system vendor at least `github.com/gardener/gardener@v1.16.0`.
```

### DIFF
--- a/pkg/operation/care/constraints.go
+++ b/pkg/operation/care/constraints.go
@@ -17,9 +17,7 @@ package care
 import (
 	"context"
 	"fmt"
-	"strings"
 
-	"github.com/gardener/gardener/extensions/pkg/webhook"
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm

PR #3869, opened 2021-04-13
First steps towards removal of duplication in extension controllers

rfranzke:

/area dev-productivity
/kind cleanup

There is a lot of duplication in the `extensions/controllers/*` package, and this PR is the first step to eliminate it. 
The goal of the PR is the removal of the duplicated `updateStatus{Processing,Error,Success}` functions in the `reconciler`s in of the various controllers. They were all kind of similar, but still different and inconsistent (which changes with this PR).

Along the way, I removed all events emitted from the extension controllers. They were never evaluated and not used at all. In large seed clusters they might even be harmful because they unnecessarily pollute the seed cluster's etcd. They were also not consistently used across the various extension controllers.

/squash

**Release note**:

```other operator
NONE
```

### DIFF
--- a/extensions/pkg/controller/status.go
+++ b/extensions/pkg/controller/status.go
@@ -89,6 +89,10 @@ func (s *statusUpdater) InjectClient(c client.Client) {
 }
 
 func (s *statusUpdater) Processing(ctx context.Context, obj extensionsv1alpha1.Object, lastOperationType gardencorev1beta1.LastOperationType, description string) error {
+	if s.client == nil {
+		return fmt.Errorf("client is not set. Call InjectClient() first")
+	}
+
 	s.logger.Info(description, s.logKeysAndValues(obj)...)
 
 	return TryUpdateStatus(ctx, retry.DefaultBackoff, s.client, obj, func() error {
@@ -100,6 +104,10 @@ func (s *statusUpdater) Processing(ctx context.Context, obj extensionsv1alpha1.O
 }
 
 func (s *statusUpdater) Error(ctx context.Context, obj extensionsv1alpha1.Object, err error, lastOperationType gardencorev1beta1.LastOperationType, description string) error {
+	if s.client == nil {
+		return fmt.Errorf("client is not set. Call InjectClient() first")
+	}
+
 	errDescription := gardencorev1beta1helper.FormatLastErrDescription(fmt.Errorf("%s: %v", description, err))
 	s.logger.Error(fmt.Errorf(errDescription), "error", s.logKeysAndValues(obj)...)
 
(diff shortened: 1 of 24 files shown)


pkg/api/extensions/accessor_test.go
```diff
@@ -194,6 +212,23 @@ var _ = Describe("Accessor", func() {
...
+				It("should get the observed generation", func() {
+					var generation int64 = 1337
+					acc := mkUnstructuredAccessorWithStatus(extensionsv1alpha1.DefaultStatus{ObservedGeneration: generation})
+					Expect(acc.GetObservedGeneration()).To(Equal(generation))
+				})
+			})
+
+			Describe("#SetObservedGeneration", func() {
+				It("should set the observed generation", func() {
+					var generation int64 = 1337
```
BeckerMax, line 225: nice one 😎

pkg/api/extensions/accessor.go
```diff
@@ -147,6 +147,18 @@ func (u unstructuredStatusAccessor) GetLastOperation() *gardencorev1beta1.LastOp
 	return lastOperation
 }
 
+// SetLastOperation implements Status.
+func (u unstructuredStatusAccessor) SetLastOperation(lastOp *gardencorev1beta1.LastOperation) {
```
BeckerMax, line 151: Is it critical/ why are we ignoring all errors here? Meaning we do not know wether setting the Operation was successful.

extensions/pkg/controller/status.go
```diff
@@ -51,3 +59,78 @@ func ReconcileSucceeded(t gardencorev1beta1.LastOperationType, description strin
...
+var _ = StatusUpdater(&statusUpdater{})
+
+func (s *statusUpdater) InjectClient(c client.Client) {
+	s.client = c
+}
+
+func (s *statusUpdater) Processing(ctx context.Context, obj extensionsv1alpha1.Object, lastOperationType gardencorev1beta1.LastOperationType, description string) error {
+	s.logger.Info(description, s.logKeysAndValues(obj)...)
+
+	return TryUpdateStatus(ctx, retry.DefaultBackoff, s.client, obj, func() error {
```
BeckerMax, line 98: Can `s.client` be unset causing a null pointer exception here? Or how do we make sure that InjectClient was always called before?

BeckerMax (dismissed):
Generally there still is a lot of duplication between the extension controllers and this becomes even more obvious after you refactoring. I assume that is what you mean with `first steps towards removal of duplication` in the title.  
My questions below are mostly to understand the code better. 
/lgtm

pkg/api/extensions/accessor.go
rfranzke: I don't have a good answer because this code was contributed by somebody who left the team, but I guess it's not too critical because I think this entire code might not be even used at all. From https://github.com/gardener/gardener/blob/77e09dd522d7bfb682cc35d0be437201b606061b/pkg/api/extensions/accessor.go#L33-L42, I understand that we always work with the dedicated Golang type instead of `unstructured` because all our extension objects implement the `extensionsv1alpha1.Object`.
Most likely, we can remove this unstructured accessor in a subsequent PR.

extensions/pkg/controller/status.go
rfranzke: Yeah, it's better to check this explicitly which I have done now, thanks.

pkg/api/extensions/accessor.go
BeckerMax: thanks, makes sense.

BeckerMax (approved):

rfranzke:
@timuthy Do you want to merge this one as part of v1.21 or wait?

PR #3874, opened 2021-04-13
[ci:component:github.com/gardener/gardener-resource-manager:v0.22.0->v0.23.0]

ghost:
**Release Notes*:
``` other developer github.com/gardener/gardener-resource-manager #118 @ialidzhikov
gardener-resource-manager now supports a `Ignore` mode for resources managed by a ManagedResource. The primary use case for this mode is a migration of resource from one ManagedResource to another one.
```

``` feature operator github.com/gardener/gardener-resource-manager #117 @timebertt
It is now possible to specify the leader election resource lock via the command line flag `--leader-election-resource-lock` (defaults to `configmapsleases`) and the chart value `leaderElection.resourceLock`. Please be careful when changing the resource lock and always migrate via multilocks in order to prevent situations where multiple instances of the controller are running with leader election and thus acting on the same resources.
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -24,7 +24,7 @@ images:
 - name: gardener-resource-manager
   sourceRepository: github.com/gardener/gardener-resource-manager
   repository: eu.gcr.io/gardener-project/gardener/gardener-resource-manager
-  tag: "v0.22.0"
+  tag: "v0.23.0"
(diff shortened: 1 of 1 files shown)


ialidzhikov (approved):
/lgtm

timebertt (approved):
/lgtm

PR #3876, opened 2021-04-13
Increase limits for etcd to avoid potential long down time during maintenance window.

amshuman-kr:

/area control-plane auto-scaling high-availability cost
/kind post-mortem

Increase limits for etcd.

Fixes https://github.tools.sap/kubernetes-live/issues-live/issues/250

Replacement for #3835 after out-of-band discussion.

There is a risk of longer downtime (also, possibly corrupted WAL files) during maintenance window in bigger clusters (if they have large fluctuations in usage pattern during 24h period) which is triggered by too drastic a scale down during maintenance window followed by sub-optimal subsequent scale ups.

The increased limit values are based on the spikes in CPU and memory usage in some of the largest shoot clusters in the landscapes including the one mentioned in the linked issue.

Please see here for more details -> https://github.tools.sap/kubernetes-live/issues-live/issues/250#issuecomment-202646

cc @mvladev  @vlerenc @istvanballok @mliepold @ialidzhikov @rfranzke @dguendisch

**Release note**:

```other operator
Increase limits for etcd to avoid potential long down time during maintenance window.
```

### DIFF
--- a/pkg/operation/botanist/component/etcd/etcd.go
+++ b/pkg/operation/botanist/component/etcd/etcd.go
@@ -434,11 +434,11 @@ func (e *etcd) Deploy(ctx context.Context) error {
 				},
 				LimitsRequestsGapScaleParams: hvpav1alpha1.ScaleParams{
 					CPU: hvpav1alpha1.ChangeParams{
-						Value:      pointer.StringPtr("1"),
+						Value:      pointer.StringPtr("2"),
 						Percentage: pointer.Int32Ptr(40),
 					},
 					Memory: hvpav1alpha1.ChangeParams{
-						Value:      pointer.StringPtr("1G"),
+						Value:      pointer.StringPtr("5G"),
 						Percentage: pointer.Int32Ptr(40),
 					},
 				},
@@ -599,8 +599,8 @@ func (e *etcd) computeContainerResources(foundSts bool, existingSts *appsv1.Stat
 				corev1.ResourceMemory: resource.MustParse("1G"),
 			},
 			Limits: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("900m"),
-				corev1.ResourceMemory: resource.MustParse("3G"),
+				corev1.ResourceCPU:    resource.MustParse("2300m"),
+				corev1.ResourceMemory: resource.MustParse("6G"),
 			},
 		}
 		resourcesBackupRestore = &corev1.ResourceRequirements{
--- a/pkg/operation/botanist/component/etcd/etcd_test.go
+++ b/pkg/operation/botanist/component/etcd/etcd_test.go
@@ -183,8 +183,8 @@ var _ = Describe("Etcd", func() {
 					corev1.ResourceMemory: resource.MustParse("1G"),
 				},
 				Limits: corev1.ResourceList{
-					corev1.ResourceCPU:    resource.MustParse("900m"),
-					corev1.ResourceMemory: resource.MustParse("3G"),
+					corev1.ResourceCPU:    resource.MustParse("2300m"),
+					corev1.ResourceMemory: resource.MustParse("6G"),
 				},
 			}
 			if existingResourcesContainerEtcd != nil {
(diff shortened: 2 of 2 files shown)


rfranzke (approved):
/lgtm
/invite @timuthy

pkg/operation/botanist/component/etcd/etcd.go
```diff
@@ -434,11 +434,11 @@ func (e *etcd) Deploy(ctx context.Context) error {
 				},
 				LimitsRequestsGapScaleParams: hvpav1alpha1.ScaleParams{
 					CPU: hvpav1alpha1.ChangeParams{
-						Value:      pointer.StringPtr("1"),
+						Value:      pointer.StringPtr("2"),
 						Percentage: pointer.Int32Ptr(40),
 					},
 					Memory: hvpav1alpha1.ChangeParams{
-						Value:      pointer.StringPtr("1G"),
+						Value:      pointer.StringPtr("5G"),
```
timuthy, line 441: Why is it desired to increase by a ration of **5** (`1G` -> `5G`) while we increased the limits below only by **2** (`3G`- > `6G`)?
Frankly, I can't quite interpret the `LimitsRequestsGapScaleParams` feature (also with the help of HVPA documentation). So it'd help if you could elaborate a bit, also for the record :)

timuthy (commented):
I honestly can't judge this change very well, esp. when it comes to the HVPA change and maybe I've missed some information from the out-of-band discussion that you mentioned in the PR description. 
Still, I found one irregularity FMPOV which I raised with a comment.

vlerenc:
@amshuman-kr Following up on the discussion out-of-band: @mvladev suggested higher limits, but in the end, if we do not disable scale-down, how is that helping again in avoiding as many involuntary ETCD "reschedulings" as possible?

Maybe the pod isn't shot down if the limits are sufficiently high and the node has capacity (so nothing against this PR), but shouldn't we also prevent any ETCD scale-downs until we have clustered ETCD? If the load increases (not only in spikes, but sufficiently enough for VPA to scale up) and then decreases again (e.g. during the day or week), won't VPA scale down the ETCD and by that cause an ETCD "rescheduling" with the known and undesired consequences?

I.e. shouldn't we do this, but also prevent the scale-down in a separate/another PR?

pkg/operation/botanist/component/etcd/etcd.go
amshuman-kr: This change was triggered because of https://github.tools.sap/kubernetes-live/issues-live/issues/250. Because of this I checked some of the largest of etcds in canary and live. I found the memory spikes of `~5G` the largest among them. Hence, the gap configuration absolute value of `5G`. And because the default requests are `1G`, increased the limits to `6G` (1+5). The original limits of `3G` seems to be arbitrary/historical from before the HVPA support for `LimitsRequestsGapScaleParams`.

The `LimitsRequestsGapScaleParams` were introduced to improve the VPA's [proportional scaling of limits](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler#limits-control) because proportional scaling doesn't work well at both min and max ends of the scale because the absolute gap between requests and limits shrinks considerably at min and max ends of the scale. The `LimitsRequestsGapScaleParams` allows for both an absolute value as well as a percentage value to be configured for the gap between requests and limits. Whatever happens to be larger is taken. Typically, the absolute value applies at the lower end of the scale and the percentage applies at the higher end. In this case, the gap is guaranteed to be at least `5G`. It could be larger at higher ends of the scale if percentage quantity exceeds `5G`.

amshuman-kr:
> I.e. shouldn't we do this, but also prevent the scale-down in a separate/another PR?

I have the local branch for the PR that I closed still. So, can open a new PR to disable ETCD scale down if that is agreed upon.

pkg/operation/botanist/component/etcd/etcd.go
timuthy: Thanks for the explanation @amshuman-kr.

timuthy (approved):

Issue #3873, opened 2021-04-13
Proposal: Expose Extension Provider Status in Gardener cluster 


danielfoehrKn:
**How to categorize this issue?**

/area quality
/kind enhancement
/priority 3

**What would you like to be added**:

I'd like to propose a way to expose the **ProviderStatus**  of an extension resource ("lives" in the Seed) to a resource in the Garden cluster.

Generally, there are three ways how to accomplish this

1. Put the information into the Shoot status 
2. Use config map or secret to store the information
3. Create a dedicated resource in the project accessible by shoot owners (raw.Config type for the state, can be used for any extension)

The downside of 1) is that it further bloats the Shoot resource, we want to avoid this especially because a provider status can be large.

For 2) 
 - advantage: less code needed because we reuse an existing type. We only need to store information, not actively reconcile it.
 - disadvantage:   
      - Gardener internal consumers (admission webhooks etc.) need to deal with an awkward format parsed from a config map.
      - provider status itself is generic, but there is general information that is not (extension kind (Infrastructure), type (aws), associated shoot that is not. This information would have to be either included in the naming of the config map resource, put into untyped fields in the cm or as annotations.

I propose to use 3) and create a new resource `ExtensionProviderStatus`.  
This has the advantage of an API with well-defined format  (besides the generic provider status) and easier handling for consumers (possibly e.g use field-selector).

**Concretely**

The Gardenlet creates & updates the resource **ExtensionProviderStatus** (namespace: project-ns, name <shoot-name>-<extension-kind>) in the garden cluster with the  **ProviderStatus** field of the particular extension resources in the Seed.
This can be done similar to the `ShootState` resource in the federated-seed-controller.

```
apiVersion: core.gardener.cloud/v1beta1
kind: ExtensionProviderStatus
metadata:
  # e.g <shoot>-infrastructure  | <shoot>-worker
  name: <shoot-name>-<extension-kind>
  namespace: garden--myproject
spec:
  kind: Infrastructure
  type: aws
  shootName: my-shoot
status:
  # provider status type raw.Extension
  apiVersion: aws.provider.extensions.gardener.cloud/v1alpha1
  ec2:
    keyName: shoot--xx-publickey
  iam:
    instanceProfiles:
      - name: shoot----seed-aws-boot-nodes
        purpose: nodes
    roles:
      - arn: arn:aws:iam::802691470131:role/shoot----seed-aws-boot-nodes
        purpose: nodes
  kind: InfrastructureStatus
[...]

PR #3878, opened 2021-04-13
Add Resource "ShootExtensionStatus" in  core.gardener.cloud

danielfoehrKn:
_**How_ to categorize this PR?**

/area quality
/kind enhancement

Adds a new resource in the `core.gardener.cloud` resource group with the name `ShootExtensionStatus`.
This resource is intended to hold the information of the `ProviderStatus` field from multiple Gardener Extension resources.

It is modeled after the `ShootState` resource. Hence, there is only a `Spec` (not Status), as no spec is actually reconciled.

Part of #3873

This PR only contains the type definitions & client generations.

**Release note**:

```feature user
Add the resource `ShootExtensionStatus` to the resource group `core.gardener.cloud` which is intended to hold the information of the `ProviderStatus` field from multiple Gardener Extension resources from the Seed cluster. 
```


pkg/apis/core/v1beta1/types_shootproviderstate.go
```diff
@@ -0,0 +1,71 @@
...
+// of a Gardener extension resource from a Shoot's namespace in the Seed cluster.
+type ExtensionProviderState struct {
+	// Kind of the extension resource
+	Kind string `json:"kind" protobuf:"bytes,1,opt,name=kind"`
+	// Type of the extension resource
+	// +optional
+	Type *string `json:"type" protobuf:"bytes,2,opt,name=type"`
+	// Status is the provider status of the extension resource
+	// +optional
+	Status *runtime.RawExtension `protobuf:"bytes,3,opt,name=status"`
```
timuthy: Why is this optional / a pointer?

pkg/apis/core/types_shootproviderstate.go
```diff
@@ -0,0 +1,62 @@
...
+
+// ExtensionProviderState contains the kind, the type and the last observed provider status
+// of a Gardener extension resource from a Shoot's namespace in the Seed cluster.
+type ExtensionProviderState struct {
+	// Kind of the extension resource
+	Kind string
+	// Type of the extension resource
+	Type *string
+	// Status is the provider status of the extension resource
+	Status *runtime.RawExtension
```
timuthy:
```suggestion
	Status runtime.Object
```

Let's follow the recommendation from the doc string of `RawExtension` (see [here](https://github.com/kubernetes/apimachinery/blob/d23dd757f6df1aa5f84b372fefbe4249d3e59943/pkg/runtime/types.go#L49)).

timuthy (changes requested):
Do you also want to include necessary storage changes for `pkg/registry/core`?

rfranzke:
What about calling the resource `ShootExtensionStatus` instead? Two rationales, first is to not "restrict" it to provider extensions (there might be other extensions like networking or what not that might produce some status), second is to not confuse it with the "state" which is used for control plane migration (also, in the `extensions.gardener.cloud` APIs the fields are called `.status.state` vs. `status.providerStatus`).

pkg/apis/core/types_shootproviderstate.go
```diff
@@ -0,0 +1,62 @@
...
+	metav1.ListMeta
+	// Items is the list of ShootProviderStates.
+	Items []ShootProviderState
+}
+
+// ShootProviderStateSpec is the specification of the ShootProviderState.
+type ShootProviderStateSpec struct {
+	// ProviderStatus holds a list containing the information of the providerStatus field of Extension resources
+	// in the Seed cluster
+	ProviderStatus []ExtensionProviderState
```
rfranzke:
```suggestion
	ExtensionStatus []ExtensionStatus
```

pkg/apis/core/types_shootproviderstate.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// ExtensionProviderState contains the kind, the type and the last observed provider status
+// of a Gardener extension resource from a Shoot's namespace in the Seed cluster.
+type ExtensionProviderState struct {
```
rfranzke:
```suggestion
type ExtensionStatus struct {
```

danielfoehrKn:
I added validation and storage implementation in a second commit.

I am a bit torn between naming the plural of the resource "shootextensionstatus" or "shootextensionstatus" - see [here](https://github.com/gardener/gardener/blob/540e1cefbaddec1172fc043c5b2d9a1d67dc47c3/pkg/registry/core/shootextensionstatus/storage/storage.go#L51). On the one hand, `statuses` seems to be the right plural for `status`.
Suggestions welcome.

EDIT: plural being used is "shootextensionstatuses"

pkg/apis/core/v1beta1/types_shootproviderstate.go
danielfoehrKn: It was modeled after the `ShootState` resource. But I think you are right. If there is no Status, we can remove the entry from the list.

danielfoehrKn:
Looping in @plkokanov to have a look at the patch strategy for the `ExtensionStatus` list as this should be similar like in the `ShootState` and I remember you tweaking something there in the past. Thanks!

danielfoehrKn:
I played around with the strategic-merge-patch on the `extensionStatus` list. 
I encountered problems due to the `rawExtension` field similar to:
`unable to find api field in struct RawExtension for the json field “ec2”`. 

I want to go forward with a `json-merge-patch` with optimistic-locking like it is already implemented for the ShootState resource [in the federated-seed-controller](https://github.com/gardener/gardener/blob/3d5f3f3fd3d65b57f775cb14fb251463ace8b731/pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go#L128).
I removed the strategic-merge specific API comments for the list `ExtensionStatus` .

@rfranzke @timuthy

pkg/apis/core/types_shootextensionstatus.go
```diff
@@ -0,0 +1,62 @@
...
+	// Standard list object metadata.
+	metav1.ListMeta
+	// Items is a list of ShootExtensionStatus.
+	Items []ShootExtensionStatus
+}
+
+// ShootExtensionStatusSpec is the specification of the ShootExtensionStatus.
+type ShootExtensionStatusSpec struct {
+	// ExtensionStatus holds a list of statuses of Extension resources
+	// in the Seed cluster
```
rfranzke:
```suggestion
	// ExtensionStatus holds a list of statuses of extension controllers.
```

(same for other occurrences)

pkg/apis/core/types_shootextensionstatus.go
```diff
@@ -0,0 +1,62 @@
...
+// ExtensionStatus contains the kind, the type and the last observed status
+// of a Gardener extension resource from a Shoot's namespace in the Seed cluster.
+type ExtensionStatus struct {
+	// Kind of the extension resource
+	Kind string
+	// Type of the extension resource
+	Type string
+	// Status contains the status of the extension resource
+	Status runtime.Object
+}
```
rfranzke: You might also need an optional `Purpose` field

pkg/apis/core/v1beta1/register.go
```diff
@@ -69,6 +69,8 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 		&SeedList{},
 		&Shoot{},
 		&ShootList{},
+		&ShootExtensionStatus{},
+		&ShootExtensionStatusList{},
```
rfranzke: What about waiting with `v1beta1` promotion for a while and only add it to `v1alpha1` in the beginning (like done for the `ShootState`)?

pkg/apis/core/v1beta1/types_shootextensionstatus.go
```diff
@@ -0,0 +1,70 @@
...
+
+// ShootExtensionStatusSpec is the specification of the ShootExtensionStatus.
+type ShootExtensionStatusSpec struct {
+	// ExtensionStatus holds a list of statuses of Extension resources
+	// in the Seed cluster
+	// No strategic merge patch on this list as it contains a rawExtension field
+	// which causes problems during the patch computation
+	// Clients should use a json-merge-patch instead handing over the up-to-date list
+	// +optional
+	ExtensionStatus []ExtensionStatus `json:"extensionStatus,omitempty" protobuf:"bytes,1,rep,name=extensionStatus"`
```
rfranzke:
```suggestion
	Statuses []ExtensionStatus `json:"statuses,omitempty" protobuf:"bytes,1,rep,name=statuses"`
```

(same elsewhere)

pkg/apis/core/validation/shootextensionstatus.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return allErrs
+}
+
+// ValidateShootExtensionStatusSpecUpdate validates the update to the specification of a ShootExtensionStatus
+func ValidateShootExtensionStatusSpecUpdate(_, _ *core.ShootExtensionStatusSpec) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	return allErrs
+}
```
rfranzke: Does it make sense to prevent kind/type from being changed?

rfranzke (commented):
Can you also add an example YAML file to the `example` folder?

pkg/apis/core/v1beta1/register.go
danielfoehrKn: I thought the goal is to phase out the v1alpha1 and not add any more resources to it? Did I misunderstand that?

pkg/apis/core/validation/shootextensionstatus.go
danielfoehrKn: Type possibly yes, but not kind. Kind is the key in the list, so changing it is like adding a new entry and removing the old one.

pkg/apis/core/v1beta1/register.go
danielfoehrKn: Anyways, I am fine with changing it to `v1alpha1` if that is the process

danielfoehrKn:
> Can you also add an example YAML file to the example folder?

Shouldn't we maintain only resources in the `example` folder that shall be manually applied / are necessary to set up Gardener?
For instance, the `ShootState` resource is also not in the example folder.

rfranzke (dismissed):
/lgtm
/hold until v1.21 was released

rfranzke (dismissed):

danielfoehrKn:
@timuthy is there something missing from your side?

pkg/apis/core/types_shootextensionstatus.go
```diff
@@ -0,0 +1,63 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ShootExtensionStatus holds the status information of extensions of a Shoot cluster
+type ShootExtensionStatus struct {
+	metav1.TypeMeta
+	// Standard object metadata.
+	// Designed to have an owner reference to the associated Shoot resource
+	metav1.ObjectMeta
+	// Specification of the ShootExtensionStatus.
+	Spec ShootExtensionStatusSpec
```
timuthy: WDYT about adding not using `Spec` but `ExtensionStatuses` directly here? I think it's somehow confusing to store a state (`ProviderStatus`) in a `Spec` field. In addition, there is no controller which would use this `Spec` and transfer it to an actual state, i.e., there is no `Spec <-> Status` pair. (/cc @rfranzke)

danielfoehrKn: Ok I see what you mean.
I modeled it after the ShootState resource as it is very similar mechanics.
So the argument to model it like this is only for consistency.
WDYT @timuthy

timuthy: Yes, I kind of wondered why we did it for the `ShootState` in the first place and maybe should change it there after some time.

I get the consistency argument but rather vote for following the K8s convention (see [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status)) if we now have the chance to do so:

> All objects that represent a physical resource whose state may vary from the user's desired intent SHOULD have a "spec" and a "status". Objects whose state cannot vary from the user's desired intent MAY have only "spec", and MAY rename "spec" to a more appropriate name.

rfranzke: I'd also vote for @timuthy's proposal. When we promote the `ShootState` to `v1beta1` we should do it similarly 👍🏻

danielfoehrKn: thanks for your input. Will adapt it

rfranzke (dismissed):
/lgtm

rfranzke:
Needs another
/rebase
and some unit tests are failing, please check

/author-action

rfranzke (approved):
/lgtm

@timuthy - any final opinions?

timuthy (approved):
/lgtm
Thank you

Issue #3672, opened 2021-03-05
core.BackupEntry is reconciled multiple times in a row


timebertt:
**How to categorize this issue?**

/area robustness quality
/kind bug
/priority normal

**What happened**:

In some gardenlet logs I observed, that a `core.BackupEntry` is always reconciled multiple times in a row.

```
time="2021-03-05T06:49:06Z" level=info msg=Succeeded flow="Shoot cluster reconciliation" operation=reconcile shoot=garden-project/some-shoot task="Deploying backup entry"
time="2021-03-05T06:49:06Z" level=info msg="Backup entry has been successfully reconciled." operation=reconcile shoot=garden-project/some-shoot
time="2021-03-05T06:49:06Z" level=info msg=Succeeded flow="Shoot cluster reconciliation" operation=reconcile shoot=garden-project/some-shoot task="Waiting until the backup entry has been reconciled"
time="2021-03-05T06:49:06Z" level=info msg=Starting backupentry=shoot--it--some-shoot--backup-bucket-uid flow="Backup Entry Reconciliation"
time="2021-03-05T06:49:06Z" level=info msg=Succeeded backupentry=shoot--it--some-shoot--backup-bucket-uid flow="Backup Entry Reconciliation" task="Waiting until the backup bucket is reconciled"
time="2021-03-05T06:49:06Z" level=info msg=Succeeded backupentry=shoot--it--some-shoot--backup-bucket-uid flow="Backup Entry Reconciliation" task="Deploying backup entry secret to seed"
time="2021-03-05T06:49:06Z" level=info msg=Succeeded backupentry=shoot--it--some-shoot--backup-bucket-uid flow="Backup Entry Reconciliation" task="Deploying backup entry extension resource"
time="2021-03-05T06:49:06Z" level=info msg=Succeeded backupentry=shoot--it--some-shoot--backup-bucket-uid flow="Backup Entry Reconciliation" task="Waiting until backup entry is reconciled"
time="2021-03-05T06:49:06Z" level=info msg=Finished backupentry=shoot--it--some-shoot--backup-bucket-uid flow="Backup Entry Reconciliation"
time="2021-03-05T06:49:06Z" level=error msg="Failed to ensure gardener finalizer on backupentry: Operation cannot be fulfilled on backupentries.core.gardener.cloud \"shoot--it--some-shoot--backup-bucket-uid\": the object has been modified; please apply your changes to the latest version and try again" backupentry=shoot--it--some-shoot--backup-bucket-uid
[...]

PR #3880, opened 2021-04-14
[GEP-7] Properly handle migration and restoration of BackupEntry

### LINKED ISSUE #3838: BackupEntries remain in the Source Seed after Control Plane Migration

/area control-plane-migration
/kind bug
/priority 2

**What happened**:
Sometimes after executing Control Plane Migration, the migrated `Shoot`'s `extensions.gardener.cloud/v1alpha1.BackupEntry` is still present in the __Source Seed__.

There should be no `BackupEntry` for the migrated `Shoot` in the __Source Seed__

Create a `Shoot` cluster and migrate it. The issue dues not occur with every migration.

After closer investigation of the issue here is what happens:
During the `Migrate` operation where we remove the `Shoot`'s control plane from the __Source Seed__, the BackupEntry in the Seed is annotated with `gardener.cloud/operation=migrate` which removes its finalizers. Afterwards the resource is deleted. At the end of the `Migrate` operation the `Spec.SeedName` of the `BackupEntry` in the garden cluster is switched to the __Destination Seed__. However, it seems that the `BackupEntry` controller in the `gardenlet` reconciles the resource either after the `BackupEntry` has been removed from the __Source Seed__ but before the `Spec.SeedName` has been changed or reconciles it with outdated data. This basically redeploys the `BackupEntry` on the __Source Seed__

**Environment**:

- Gardener version: 1.20.0
- Others:

plkokanov:

/area control-plane-migraton
/kind bug

This PR makes it so that the handling of migration/restoration for the `extensions.BackupEntry` resources is handled by the BackupEntry controller in the `gardenlet`.  Additionally, the `core.BackupEntry` resource is no longer reconciled multiple times in a row.

Fixes #3838 and  #3672
Part of #1631 

As part of #3672:
1. The `BackupEntry.Status.ObservedGeneration` is updated to equal `BackupEntry.Metadata.Generation` near the start of the operation instead of at its end.
2. The reconciliation logic checks if the `gardener` finalizer already exists in the `BackupEntry` resource before attempting to add it.
3. If the finalizer does not exist, the reconciliation will add it and then return without error. The addition of the finalizer will automatically requeue the `BackupEntry` for reconciliation (similar to how it is done for the `Shoot`).
4. I have kept the logic that removes the `gardener.cloud/operation` annotations at the end of the reconcile and migrate operations as their presence does not cause multiple reconciliations in a row since we do not update the `BackupEntry.Metadata` field during those operations (apart from the finalizers, but that is already covered in 3.).

**Release note**:

```bugfix operator
Migration and restoration of `extensions.gardener.cloud.BackupEntry` resources is now handled by the BackupEntry controller in the `gardenlet`.
```
```bugfix operator
The `core.gardener.cloud.BackupEntry` resource is no longer reconciled multiple times in a row.
```


rfranzke:
Does it make sense to split the changes into multiple commits to ease and speed-up the review process? Ref https://github.com/gardener/documentation/blob/master/CONTRIBUTING.md#contributing-bigger-changes

plkokanov:
@rfranzke yep I will split them up.

plkokanov:
@rfranzke I rebased and split the changes into multiple commits.

rfranzke:
Thank you @plkokanov, will try to check asap.
/assign

pkg/controllerutils/seedfilter.go
```diff
@@ -149,10 +149,26 @@ func BackupEntryFilterFunc(ctx context.Context, c client.Client, seedName string
...
+		return seedLabelsMatchWithClient(ctx, c, *backupEntry.Status.SeedName, labelSelector)
+	}
+}
+
+// BackupEntryIsManagedByThisGardenlet checks if the given BackupEntry is managed by this gardenlet by comparing it with the seed name from the GardenletConfiguration
+// or by checking whether the seed labels match the seed selector from the GardenletConfiguration.
+func BackupEntryIsManagedByThisGardenlet(ctx context.Context, c client.Client, backupEntry *gardencorev1beta1.BackupEntry, gc *config.GardenletConfiguration) bool {
+	seedName := confighelper.SeedNameFromSeedConfig(gc.SeedConfig)
+	if len(seedName) > 0 {
+		return *backupEntry.Spec.SeedName == seedName
```
rfranzke:
```suggestion
		return backupEntry.Spec.SeedName != nil && *backupEntry.Spec.SeedName == seedName
```

pkg/gardenlet/controller/backupentry/backup_entry_reconciler.go
```diff
@@ -81,6 +81,16 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	}
+
+	if shouldMigrateBackupEntry(be) {
+		return r.migrateBackupEntry(ctx, gardenClient, be)
+	}
+
+	if !controllerutils.BackupEntryIsManagedByThisGardenlet(ctx, gardenClient.Client(), be, r.config) {
+		r.logger.Debugf("Skipping because BackupEntry is not managed by this gardenlet in seed %s", *be.Spec.SeedName)
+		return reconcile.Result{}, nil
+	}
```
rfranzke, line 93: Shouldn't this check be earlier/before calling the actuator?

pkg/operation/botanist/backupentry.go
```diff
@@ -62,3 +63,10 @@ func (b *Botanist) DefaultExtensionsBackupEntry(seedClient client.Client) extens
 		extensionsbackupentry.DefaultTimeout,
 	)
 }
+
+func (b *Botanist) DeployBackupEntry(ctx context.Context) error {
```
rfranzke, line 69: doc string

pkg/operation/botanist/component/backupentry/backupentry.go
```diff
@@ -163,6 +135,77 @@ func (b *backupEntry) Wait(ctx context.Context) error {
...
+func (b *backupEntry) Migrate(ctx context.Context) error {
+	backupEntry := &gardencorev1beta1.BackupEntry{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      b.values.Name,
+			Namespace: b.values.Namespace,
+		},
+	}
+
+	if err := b.client.Get(ctx, kutil.Key(b.values.Namespace, b.values.Name), backupEntry); err != nil {
+		return client.IgnoreNotFound(err)
```
rfranzke, line 148: Just to confirm: This returns `nil` when the `BackupEntry` was not found. Is this the expected behaviour?

pkg/operation/botanist/component/backupentry/backupentry.go
```diff
@@ -163,6 +135,77 @@ func (b *backupEntry) Wait(ctx context.Context) error {
...
+	}
+
+	metav1.SetMetaDataAnnotation(&backupEntry.ObjectMeta, v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationMigrate)
+	metav1.SetMetaDataAnnotation(&backupEntry.ObjectMeta, v1beta1constants.GardenerTimestamp, TimeNow().UTC().String())
+	backupEntry.Spec.SeedName = b.values.SeedName
```
rfranzke: Wouldn't changing the seed name be good enough?

pkg/operation/common/utils.go
```diff
@@ -121,6 +121,16 @@ func ExtractShootDetailsFromBackupEntryName(backupEntryName string) (shootTechni
 	return shootTechnicalID, shootUID
 }
 
+// GetShootNameFromOwnerReferences
```
rfranzke: ????

pkg/operation/common/utils.go
```diff
@@ -121,6 +121,16 @@ func ExtractShootDetailsFromBackupEntryName(backupEntryName string) (shootTechni
...
+func GetShootNameFromOwnerReferences(ownerReferences []metav1.OwnerReference) string {
+	for _, ownerRef := range ownerReferences {
+		if ownerRef.Kind == "Shoot" {
+			return ownerRef.Name
+		}
+	}
+	return ""
+}
```
rfranzke: Move this function to `pkg/utils/gardener` and add a unit test

pkg/operation/shoot/shoot.go
```diff
@@ -254,10 +254,10 @@ func (b *Builder) Build(ctx context.Context, c client.Client) (*Shoot, error) {
 	return shoot, nil
 }
 
-// GetExtensionComponents returns a list of component.DeployMigrateWaiters of all extension components.
-func (s *Shoot) GetExtensionComponents() []component.DeployMigrateWaiter {
+// GetExtensionComponentsToMigrate returns a list of component.DeployMigrateWaiters of extension components that
+// should be migrated by the shoot controller.
+func (s *Shoot) GetExtensionComponentsToMigrate() []component.DeployMigrateWaiter {
```
rfranzke:
```suggestion
func (s *Shoot) GetMigrateableExtensionComponents() []component.DeployMigrateWaiter {
```

or

```suggestion
func (s *Shoot) GetExtensionComponentsForMigration() []component.DeployMigrateWaiter {
```

pkg/registry/core/backupentry/strategy.go
```diff
@@ -81,8 +80,12 @@ func mustIncreaseGeneration(oldBackupEntry, newBackupEntry *core.BackupEntry) bo
...
 	}
 
-	if kutil.HasMetaDataAnnotation(&newBackupEntry.ObjectMeta, v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationReconcile) {
-		return true
+	if val, ok := newBackupEntry.ObjectMeta.Annotations[v1beta1constants.GardenerOperation]; ok {
+		if val == v1beta1constants.GardenerOperationReconcile ||
+			val == v1beta1constants.GardenerOperationMigrate ||
+			val == v1beta1constants.GardenerOperationRestore {
+			return true
+		}
```
rfranzke: This will always increase the generation for every change as long as the above conditions are met. Instead, it should only be increased once when the annotation is set, correct?

rfranzke (changes requested):
The dedicated commits helped A LOT, thank you! Kindly do the adaptations in a separate commit as well, please.

/merge squash

pkg/gardenlet/controller/backupentry/backup_entry_reconciler.go
plkokanov: I added this check just in case (exactly as we do it in the Shoot), because of a problem we encountered after the `Shoot` migration as the shoot could be requeued with the old `spec.SeedName` - https://github.com/gardener/gardener/pull/3301

pkg/operation/botanist/component/backupentry/backupentry.go
plkokanov: Yes, I assumed that if the `core.BackupEntry` was not found it might have been deleted and therefore the `extensions.BackupEntry` in the Seed cluster should also be deleted, so there's nothing to be migrated.

pkg/operation/botanist/component/backupentry/backupentry.go
plkokanov: Added this for consistency reasons as we add the `Restore` and `Reconcile` operation annotations in the respective functions.

rfranzke: Hm, I can't follow - why is this an argument for putting some annotations while there's already a better way to trigger the operation (changing the seed name)? We don't do the same for `Shoot`s either, do we?

pkg/registry/core/backupentry/strategy.go
plkokanov: Yep, it would increase the generation for metadata changes (status updates won't do anything). 

However, I was worried about the following scenario:
1. Resource is annotated/updated and generation is increased
2. The resource is picked for reconciliation
3. The status is updated and the ObservedGeneration is set to equal the Generation
4. The controller crashes before being able to remove the operation annotation.

When the controller restarts it should requeue the BackupEntry via the `Add` event, but I was afraid if for some reason the `Add` event never gets fired. Then, when we actually want to reconcile the BackupEntry resource we won't be able to do that even by changing its Timestap annotation `metav1.SetMetaDataAnnotation(&backupEntry.ObjectMeta, v1beta1constants.GardenerTimestamp, TimeNow().UTC().String())` as this would not increase the the generation.
I'm really not sure if there's a possibility for `Add` events to be missed, as I haven't encountered such a case before, so I could improve the check to compare the old vs the new annotations, wdyt @rfranzke .

rfranzke: Why should the resource be picked for reconciliation (step 2) after it was annotated with the "migrate" or "restore" operation? And even if, then there would be another attempt to set the same operation annotation by gardenlet, right?
So, what you write sounds a little artificial/spiritual to me 😄 I would vote to implement it "the proper way" and follow up if we see concrete issues with it. This current behaviour (and also what is happening in the `shoot/strategy.go`) might have undesired side effects.

plkokanov: Spiritual describes it pretty nice 😄. I'll adapt it to use the comparison of the old and new annotations.

plkokanov:
Rebased and applied review comments.

stoyanr (approved):
/lgtm

rfranzke (approved):
/lgtm

PR #3883, opened 2021-04-15
[GEP-16] Add advertise addresses in Shoot status

mvladev:

/area control-plane
/kind  api-change

`advertiseAddresses` is added to the `Shoot` status. It contains a list of URLs on which the `kube-apiserver` can be reached.

Part of #3823 

**Release note**:

```feature user
New `.status.advertisedAddresses` field in the `Shoot` resource now provides a list of advertised URLs of the Kubernetes API Server.
```


pkg/operation/operation.go
```diff
@@ -599,3 +600,35 @@ func (o *Operation) ComputeIngressHost(prefix string) string {
 	shortID := strings.Replace(o.Shoot.Info.Status.TechnicalID, shoot.TechnicalIDPrefix, "", 1)
 	return fmt.Sprintf("%s-%s.%s", prefix, shortID, o.Seed.IngressDomain())
 }
+
+// ToAdvertisedAddresses returns list of advertised addresses on a Shoot cluster.
+func (o *Operation) ToAdvertisedAddresses() []gardencorev1beta1.ShootAdvertisedAddress {
+	addresses := []gardencorev1beta1.ShootAdvertisedAddress{}
```
rfranzke: What about

```suggestion
	var addresses []gardencorev1beta1.ShootAdvertisedAddress
```

to return `nil` when there is no address instead of an empty list?

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -602,7 +602,14 @@ func (c *Controller) updateShootStatusOperationStart(ctx context.Context, g gard
...
 
-func (c *Controller) updateShootStatusOperationSuccess(ctx context.Context, g gardencore.Interface, shoot *gardencorev1beta1.Shoot, shootSeedNamespace string, seedName *string, operationType gardencorev1beta1.LastOperationType) (*gardencorev1beta1.Shoot, error) {
+func (c *Controller) updateShootStatusOperationSuccess(
+	ctx context.Context,
+	g gardencore.Interface,
+	op *operation.Operation,
+	shoot *gardencorev1beta1.Shoot,
+	shootSeedNamespace string,
+	seedName *string,
+	operationType gardencorev1beta1.LastOperationType) (*gardencorev1beta1.Shoot, error) {
```
rfranzke: For readability:

```suggestion
func (c *Controller) updateShootStatusOperationSuccess(
	ctx context.Context,
	g gardencore.Interface,
	op *operation.Operation,
	shoot *gardencorev1beta1.Shoot,
	shootSeedNamespace string,
	seedName *string,
	operationType gardencorev1beta1.LastOperationType,
) (
	*gardencorev1beta1.Shoot, 
	error,
) {
```

However, you can now probably remove the 

```
	shoot *gardencorev1beta1.Shoot,
	shootSeedNamespace string,
```

arguments as you pass the `*Operation`, so you can access them via `op.Shoot` or `op.Shoot.SeedNamespace`

Also, `	g gardencore.Interface,` is probably not needed as well now that you can access it via `op.K8sGardenClient.GardenCore()`?

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -673,12 +680,24 @@ func (c *Controller) updateShootStatusOperationSuccess(ctx context.Context, g ga
...
 
-func (c *Controller) updateShootStatusOperationError(ctx context.Context, gardenClient kubernetes.Interface, shoot *gardencorev1beta1.Shoot, description string, operationType gardencorev1beta1.LastOperationType, lastErrors ...gardencorev1beta1.LastError) (*gardencorev1beta1.Shoot, error) {
+func (c *Controller) updateShootStatusOperationError(
+	ctx context.Context,
+	gardenClient kubernetes.Interface,
+	op *operation.Operation,
+	shoot *gardencorev1beta1.Shoot,
+	description string,
+	operationType gardencorev1beta1.LastOperationType,
+	lastErrors ...gardencorev1beta1.LastError) (*gardencorev1beta1.Shoot, error) {
```
rfranzke: same here

rfranzke (commented):
nits

pkg/operation/operation.go
mvladev: It doesn't matter for the API at the end.

rfranzke: True, but I think in most other places we prefer returning `nil` in such cases. If it doesn't make a difference can you adapt?

pkg/gardenlet/controller/shoot/shoot_control.go
mvladev: In some cases the operation would be `nil`

rfranzke (approved):
/lgtm

Issue #3875, opened 2021-04-13
[GEP-7] Control Plane Migration: Etcd Backup Migration / Leader Election PoC


stoyanr:
**How to categorize this issue?**

/area control-plane-migration
/kind enhancement
/priority 3

**What would you like to be added**:

We would like to implement a PoC for the etcd backup migration / leader election mechanism as currently proposed with #3066, in order to evaluate the feasibility of the proposed approach and try to address the remaining open issues. We don't aim yet for production quality code and won't be opening PRs to gardener components while still at the PoC stage.

Below is a rough outline of what we would like to implement as part of the PoC. Based on what we learned or ongoing discussions, we might change direction and implement things differently. In such cases, we will update this issue accordingly.

1. `gardenlet` restoration flow
  * Creates one additional BE for source seed, changes the seed name to the dest seed in the existing BE
  * Annotates the `etcd` resource with `operation=restore`
  * Waits for the status of the `etcd` resource to indicate that the migration is finished
  * Annotates the `etcd` resource with `operation=reconcile`
  * Deletes the source seed BE

2. `etcd-druid`
  * Reconciles an `etcd` resource with `operation=restore` and creates `etcd-backup-restore migrate` job
  * Waits for the `etcd-backup-restore migrate` job to complete and updates the status of the `etcd` resource

3. BE controller
  * Handles 2 BEs correctly to avoid name collisions in secrets

4. `etcd-backup-restore` support for `migrate`
  * Writes header files in source and dest BBs for the migrate operation
  * Waits for a snapshot newer than the header timestamp to be created (final snapshot) with a timeout (e.g. 10 min)
  * Migrates all snapshots from source to dest BB (final snapshot first, rest after that)
  * Updates header files that the migration was performed

5. `etcd-backup-restore` normal operation
  * Checks header files before taking a snapshot
  * If source detected
    * If final snapshot not taken (how to detect?), cuts access to main etcd (how?), takes final snapshot, and exits
    * Otherwise, just exits
  * If destination detected
    * If a snapshot exists, restores it
    * Otherwise, waits with a timeout, then exits
  * Otherwise, takes a normal snapshot as before

vlerenc:
@stoyanr I think I understand 2, 4, and 5, but can you elaborate a bit more on 1 and 3?

stoyanr:
> I think I understand 2, 4, and 5, but can you elaborate a bit more on 1 and 3?

[...]

PR #3884, opened 2021-04-16
Add Backup{Bucket,Entry}, ManagedSeed, ControllerInstallation handling to Seed{Authorizer,Restriction}

rfranzke:

/area security
/kind enhancement

This PR adds handling for `BackupBucket` resource to the `SeedAuthorizer` and `SeedRestriction` handlers.

Part of #1723 

Similar to #3708

**Release note**:

```other operator
NONE
```


rfranzke:
I've added support for the `BackupEntry` resource to this PR as well because it's very similar to what needs to be done for `BackupBucket`s. 

/squash
/invite @timuthy

rfranzke:
There's now another commit that adds support for `ManagedSeed`s as it's again very similar to the other changes in this PR.

rfranzke:
There's now another commit that adds support for `ControllerInstallation`s as it's again very similar to the other changes in this PR.

docs/deployment/gardenlet_api_access.md
```diff
@@ -31,6 +31,7 @@ Today, the following rules are implemented:
 | Resource        | Verbs                                                         | Path                                          | Description |
 | --------------- | ------------------------------------------------------------- | ----------------------------------------------| ----------- |
 | `BackupBucket`  | `get`, `list`, `watch`, `create`, `update`, `patch`, `delete` | `BackupBucket` -> `Seed`                      | Allow `get`, `list`, `watch` requests for all `BackupBucket`s. Allow only `create`, `update`, `patch`, `delete` requests for `BackupBucket`s assigned to the `gardenlet`'s `Seed`. |
+| `BackupEntry`   | `get`, `list`, `watch`, `create`, `update`, `patch`           | `BackupEntry` -> `Seed`                       | Allow `get`, `list`, `watch` requests for all `BackupEntry`s. Allow only `create`, `update`, `patch` requests for `BackupEntry`s assigned to the `gardenlet`'s `Seed` and referencing `BackupBucket`s assigned to the `gardenlet`'s `Seed`. |
```
stoyanr: This has consequences for CP Migration if what is proposed with #3066 (see also the PoC outlined in #3875) is implemented. At some point in the destination seed, there will be a BackupEntry referencing the old ("source") BackupBucket. This entry will only exist for a limited time while the copying of backups is performed from the source to the destination (usually a few minutes to a few hours, perhaps more if there are errors), but it has to be possible to create it, otherwise I don't really see how copying backups could work at all. What would be the mitigation for this special case?

stoyanr: Also, why is `delete` not allowed at all? Perhaps it's because `gardenlet` is not expected to delete BackupEntries, but again this might change due to the above proposal (the temporary BackupEntry would need to be first created and then deleted).

stoyanr (dismissed):
/lgtm, only some questions regarding the `BackupEntry` authorization related to the CP Migration PoC we are currently working on

docs/deployment/gardenlet_api_access.md
rfranzke: No problem, if you plan to change the handling in the future then just keep in mind to also adapt this seed authorization/restriction coding. For now, I want to reflect the current state.

> Also, why is delete not allowed at all? Perhaps it's because gardenlet is not expected to delete BackupEntries

No strong reason, today, the gardenlet does not explicitly delete `BackupEntry` resources but rather relies on garbage collection after shoot deletion via owner references. Again, if this changes in the future, feel free to adapt if needed.

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
```diff
@@ -73,75 +73,68 @@ func (a *authorizer) Authorize(_ context.Context, attrs auth.Attributes) (auth.D
...
 
-	// When a new ShootState is created then it doesn't yet exist in the graph, so we only handle update operations
-	// here. The create case is handled in the SeedRestriction admission handler.
-	if attrs.GetVerb() == "create" {
+	// When a new object is created then it doesn't yet exist in the graph, so usually such requests are always allowed
+	// as the 'create case' is typically handled in the SeedRestriction admission handler. Similarly, resources for
+	// which the gardenlet has a controller need to be listed/watched, so those verbs would also be allowed here.
+	if utils.ValueExists(attrs.GetVerb(), alwaysAllowedVerbs) {
 		return auth.DecisionAllow, "", nil
 	}
```
timuthy: I wonder if we should move this check to the top so that `allowedVerbs` doesn't have to contain all elements of  `alwaysAllowedVerbs`. WDYT?

Concretely we can rearrange the order to:
- `checkSubresource`
- `utils.ValueExists(attrs.GetVerb(), alwaysAllowedVerbs)`
- `a.checkVerb(seedName, attrs, allowedVerbs...)`

timuthy (commented):
Just one comment from my side, looks good otherwise. Thanks for splitting the commits for an easier review 👍

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
rfranzke: Yes, good idea, thanks! I've done it with 8d87acc.

PS: When calling the `checkVerb` function I merge `allowedVerbs` and `alwaysAllowedVerbs` to provide a more meaningful error message in case this function returns an error.

timuthy (approved):
/lgtm

PR #3885, opened 2021-04-16
Aggregate container label for seed:images:count

wyb1:

/area monitoring
/kind enhancement

The container label should also be part of the recording rule for the metric `seed:images:count`.

Often `image` and `image_id` are not human readable, so to identify what the image is it can be helpful to have the container name included in the metric.

Fixes #

**Release note**:

```other operator
NONE
```


timebertt (approved):
/lgtm

PR #3893, opened 2021-04-19
Add ShootExtensionStatus controllers

danielfoehrKn:

/area quality
/kind enhancement

Contains 3 commits
- Add ShootExtensionStatus to SeedRestriction Admission Webhook
- Add ShootExtensionStatus to SeedAuthorizer Authorization Webhook
- Add federated seed controllers for the ShootExtensionStatus resource.

Part of #3873 

 - Creation of the `ShootExtensionStatus` is done via ShootReconciliation flow. Deletion is implicit via owner reference on the particular Shoot resource.
- Adds the reconciliation of the `ShootExtensionStatus` to the federated seed controller in the Gardenlet

Only enables the sync for the `Infrastructure` and `Worker` extension resources for now.
 - This is to avoid unnecessary network traffic. The mentioned resources are the only ones that currently write a `providerStatus`  in their status. 
 - Also, the Gardenlet is unable to detect and therefore `not` sync an "empty" `providerStatus` that only contains `apiVersion` and `kind` such as:
 ```
  providerStatus:
      apiVersion: calico.networking.extensions.gardener.cloud/v1alpha1
      kind: NetworkStatus
```

Fixes #

**Release note**:

```other operator
THe Gardenlet exposes the provider status of extension resources from the Seed cluster in the Garden cluster via the "ShootExtensionStatus" resource.
```

### DIFF
--- a/pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_shootextensionstatus.go
+++ b/pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_shootextensionstatus.go
@@ -1,0 +1,73 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package graph
+
+import (
+	"time"
+
+	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
+	toolscache "k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+)
+
+func (g *graph) setupShootExtensionStatusWatch(informer cache.Informer) {
+	informer.AddEventHandler(toolscache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			status, ok := obj.(*gardencorev1alpha1.ShootExtensionStatus)
+			if !ok {
+				return
+			}
+			g.handleShootExtensionStatusCreate(status)
+		},
+
+		DeleteFunc: func(obj interface{}) {
+			if tombstone, ok := obj.(toolscache.DeletedFinalStateUnknown); ok {
+				obj = tombstone.Obj
+			}
+			status, ok := obj.(*gardencorev1alpha1.ShootExtensionStatus)
+			if !ok {
+				return
+			}
+			g.handleShootExtensionStatusDelete(status)
+		},
+	})
+}
+
+func (g *graph) handleShootExtensionStatusCreate(status *gardencorev1alpha1.ShootExtensionStatus) {
+	start := time.Now()
+	defer func() {
+		metricUpdateDuration.WithLabelValues("ShootExtensionStatus", "CreateOrUpdate").Observe(time.Since(start).Seconds())
+	}()
+	g.lock.Lock()
+	defer g.lock.Unlock()
+
+	var (
+		shootExtensionStatusVertex = g.getOrCreateVertex(VertexTypeShootExtensionStatus, status.Namespace, status.Name)
+		shootVertex                = g.getOrCreateVertex(VertexTypeShoot, status.Namespace, status.Name)
+	)
+
+	g.addEdge(shootExtensionStatusVertex, shootVertex)
+}
+
+func (g *graph) handleShootExtensionStatusDelete(status *gardencorev1alpha1.ShootExtensionStatus) {
+	start := time.Now()
+	defer func() {
+		metricUpdateDuration.WithLabelValues("ShootExtensionStatus", "Delete").Observe(time.Since(start).Seconds())
+	}()
+	g.lock.Lock()
+	defer g.lock.Unlock()
+
+	g.deleteVertex(VertexTypeShootExtensionStatus, status.Namespace, status.Name)
+}
--- a/pkg/extensions/cluster.go
+++ b/pkg/extensions/cluster.go
@@ -22,7 +22,9 @@ import (
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"github.com/gardener/gardener/pkg/controllerutils"
+	"github.com/gardener/gardener/pkg/operation/common"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -211,4 +213,22 @@ func GetShoot(ctx context.Context, c client.Client, namespace string) (*gardenco
 // NewGardenDecoder returns a new Garden API decoder.
 func NewGardenDecoder() runtime.Decoder {
 	return serializer.NewCodecFactory(gardenScheme).UniversalDecoder()
+}
+
+// ClusterFromRequest returns the cluster resource for a given reconcile.Request
+func ClusterFromRequest(ctx context.Context, seedClient client.Client, req reconcile.Request) (*extensionsv1alpha1.Cluster, error) {
+	var clusterName string
+	if req.Namespace == "" {
+		// Handling for cluster-scoped backupentry extension resources.
+		clusterName, _ = common.ExtractShootDetailsFromBackupEntryName(req.Name)
+	} else {
+		clusterName = req.Namespace
+	}
+
+	cluster := &extensionsv1alpha1.Cluster{}
+	if err := seedClient.Get(ctx, kutil.Key(clusterName), cluster); err != nil {
+		return nil, err
+	}
+
+	return cluster, nil
 }
--- a/pkg/gardenlet/controller/federatedseed/extensions/artifact.go
+++ b/pkg/gardenlet/controller/federatedseed/extensions/artifact.go
@@ -34,6 +34,7 @@ import (
 type controllerArtifacts struct {
 	stateArtifacts                  map[string]*artifact
 	controllerInstallationArtifacts map[string]*artifact
+	extensionStatusArtifacts        map[string]*artifact
 	hasSyncedFuncs                  []cache.InformerSynced
 	shutDownFuncs                   []func()
 }
@@ -58,6 +59,7 @@ func newControllerArtifacts() controllerArtifacts {
 	a := controllerArtifacts{
 		controllerInstallationArtifacts: make(map[string]*artifact),
 		stateArtifacts:                  make(map[string]*artifact),
+		extensionStatusArtifacts:        make(map[string]*artifact),
 	}
 
 	gvk := dnsv1alpha1.SchemeGroupVersion.WithKind(dnsv1alpha1.DNSProviderKind)
@@ -70,11 +72,13 @@ func newControllerArtifacts() controllerArtifacts {
 			return ok1 && ok2 && oldExtensionObj.Spec.Type != newExtensionObj.Spec.Type
 		}),
 		disabledArtifact(),
+		disabledArtifact(),
 	)
 
 	gvk = extensionsv1alpha1.SchemeGroupVersion.WithKind(extensionsv1alpha1.BackupBucketResource)
 	a.registerExtensionControllerArtifacts(
 		newControllerInstallationArtifact(gvk, func() client.ObjectList { return &extensionsv1alpha1.BackupBucketList{} }, extensionTypeChanged),
+		disabledArtifact(),
 		disabledArtifact(),
 	)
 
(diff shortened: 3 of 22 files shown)


pkg/gardenlet/controller/federatedseed/extensions/shootextensionstatus_control.go
```diff
@@ -0,0 +1,234 @@
...
+
+func (s *ShootExtensionStatusControl) patchShootExtensionStatus(ctx context.Context, shootExtensionStatus *gardencorev1alpha1.ShootExtensionStatus, shootExtensionStatusCopy *gardencorev1alpha1.ShootExtensionStatus, kind string) (reconcile.Result, error) {
+	if err := s.gardenClient.Patch(ctx, shootExtensionStatus, client.MergeFromWithOptions(shootExtensionStatusCopy, client.MergeFromWithOptimisticLock{})); err != nil {
+		message := fmt.Sprintf("The %q extension status for Shoot %q was NOT successfully synced: %v", shootExtensionStatus.Name, kind, err)
+		s.log.Error(message)
+		return reconcile.Result{}, err
+	}
+
+	message := fmt.Sprintf("The provider status of the %q extension of Shoot %q was successfully synced to the Garden cluster", kind, shootExtensionStatus.Name)
+	s.log.Info(message)
```
timebertt:
```suggestion
	s.log.Debug(message)
```

pkg/gardenlet/controller/federatedseed/extensions/shootextensionstatus_control.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	if shoot == nil {
+		return nil, fmt.Errorf("cluster resource %s doesn't contain shoot resource", cluster.Name)
+	}
+	return shoot, nil
+}
+
+func (s *ShootExtensionStatusControl) patchShootExtensionStatus(ctx context.Context, shootExtensionStatus *gardencorev1alpha1.ShootExtensionStatus, shootExtensionStatusCopy *gardencorev1alpha1.ShootExtensionStatus, kind string) (reconcile.Result, error) {
+	if err := s.gardenClient.Patch(ctx, shootExtensionStatus, client.MergeFromWithOptions(shootExtensionStatusCopy, client.MergeFromWithOptimisticLock{})); err != nil {
```
timebertt, line 181: This has a high chance of conflicting with other current updates to the `ShootExtensionStatus`.
Have you considered a strategic merge patch without optimistic locking?
This will need some `patchMergeKey` here: https://github.com/gardener/gardener/blob/abfb062f5121cf53b7e52ddf65022c3f6b867bb4/pkg/apis/core/v1alpha1/types_shootextensionstatus.go#L34

pkg/gardenlet/controller/federatedseed/extensions/shootextensionstatus_control.go
```diff
@@ -0,0 +1,234 @@
...
+			shoot, err := s.getShootForRequest(ctx, req)
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to delete the provider status of the %q extension (%s/%s) from the ShootExtensionStatus: %w",
+					kind,
+					req.NamespacedName.Namespace,
+					req.NamespacedName.Name,
+					err)
+			}
+
+			shootExtensionStatus := &gardencorev1alpha1.ShootExtensionStatus{}
```
timebertt, line 78: Can you check how this will play with the `SeedAuthorizer`? 
I guess we can't/shouldn't simply watch all `ShootExtensionStatus`es.

timebertt (commented):
I have added some general concerns, haven't checked everything in detail yet.

pkg/gardenlet/controller/federatedseed/extensions/shootextensionstatus_control.go
danielfoehrKn: Yeah, I tried using the `strategicMergePatch` but there is an issue with using `rawExtension` field. That is also the reason why in the ShootState a json-merge-patch is used @plkokanov .

It seems like when using `strategicMergePatch`  it also tries to merge the `rawExtension` field - and it does not know how to do that for lists inside the field (instead of simply replacing the field content).

This can be observed consequently only when there is a conflict with an existing list entry.

>  unable to find api field in struct RawExtension for the json field “ec2”

If you know how to use the `strategicMergePatch` safely with `rawExtension` fields, I would of course prefer that.

pkg/gardenlet/controller/federatedseed/extensions/shootextensionstatus_control.go
danielfoehrKn: Good point regarding the SeedAuthorizer, though I would propose to take care of the seed restriction admission plugin & authorizer webhook in a follow-up PR.

However, we do not watch `ShootExtensionStatuses` here as far as I can tell (only get request on a `ShootExtensionStatus` in a certain namespace). Or is the controller-runtime opening informer-based watches on all `ShootExtensionStatuses` under the hood?. 
If yes, what can be done about it?

rfranzke: > though I would propose to take care of the seed restriction admission plugin & authorizer webhook in a follow-up PR.

At this point, we think only `Secret` handling is missing in the seed authorizer. I plan to open a PR for it next week. After it's open/merge merge, I would like to go ahead and play with Gardener while activating the seed authorizer.
Hence, if we now NOT merge this PR and don't have the necessary code for `ShootExtensionStatus`es in the seed authorizer, I'm blocked. So, please add it to this PR instead of postponing it (it shouldn't be too much of necessary code).

> If yes, what can be done about it?

You might want to look into https://github.com/gardener/gardener/pull/3803.

pkg/gardenlet/controller/federatedseed/extensions/shootextensionstatus_control.go
danielfoehrKn: The ShootState controller also logs with [Info level on success](https://github.com/gardener/gardener/blob/3d5f3f3fd3d65b57f775cb14fb251463ace8b731/pkg/gardenlet/controller/federatedseed/extensions/shootstate_control.go#L135).  Is there a reason you would like to change that?

danielfoehrKn:
Added
 - client does not list & watch anymore for ShootExtensionStatus resources
 - Added handling in Admission & Authorization webhooks in separate PRs

PTAL @rfranzke @timebertt

pkg/gardenlet/controller/federatedseed/extensions/shootextensionstatus_control.go
timebertt: Well, it's just annoying to log such messages on `Info` and unnecessarily spams the logs 😄

pkg/gardenlet/controller/federatedseed/extensions/shootextensionstatus_control.go
timebertt: Yeah, checked this in detail now and it seems to be very tricky indeed.

I see several problems:
- we would need a `UID` field in `ExtensionStatus` which can be used as a `patchMergeKey` for `ShootExtensionStatus.Statuses` (strategic merge can't handle multiple merge keys, I think)
- we would need to use the [`replace` directive](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md#replace-directive) to tell strategic merge patch to not merge but replace the `RawExtension` field. The problem here is, that adding it to the `Status` field (`patchStrategy:"replace"`) is not sufficient, as it will send all statuses on every patch. The directive has to be added inside `RawExtension`'s json. However, then the patch computation will still fail, as the computation doesn't take the directives into account but only the patch strategy.

TL;DR: this is just to confusing to figure out, so let's use the merge patch with optimistic locking as you proposed 👍

danielfoehrKn:
Do you still want to take a look ? @timebertt & @rfranzke

timebertt:
Sorry for the delay, it's quite full here. Will try to recheck today.

pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_shootextensionstatus.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"time"
+
+	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
+	toolscache "k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+)
+
+func (g *graph) setupShootExtensionStatusWatch(informer cache.Informer) {
```
rfranzke, line 25: I suggest to use a metadata-only watch to reduce the memory footprint of GAC (you seem to be only interested in namespace/name of the object).

pkg/gardenlet/controller/federatedseed/extensions/artifact.go
```diff
@@ -209,6 +230,25 @@ func newStateArtifact(gvk schema.GroupVersionKind, newObjFunc func() client.Obje
 	return a
 }
 
+func newExtensionStatusArtifact(gvk schema.GroupVersionKind, newObjFunc func() client.Object, fn predicateFn) *artifact {
```
rfranzke, line 233: It's probably not needed to let gardenlet watch all these objects and return the status "asap" similar to the `ShootState`. The `.status.providerStatus` should only change after reconciliation, hence, it would be sufficient if you add another task to the end of the reconciliation flow which syncs the provider statuses, for example.

pkg/gardenlet/controller/federatedseed/extensions/common.go
```diff
@@ -0,0 +1,33 @@
...
+	"context"
+
+	apiextensions "github.com/gardener/gardener/pkg/api/extensions"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetExtensionObject is a helper to get the extension object for a key and objectCreator function
+func GetExtensionObject(ctx context.Context, seedClient client.Client, key types.NamespacedName, objectCreator func() client.Object) (extensionsv1alpha1.Object, error) {
```
rfranzke, line 27: unit test?

pkg/extensions/cluster.go
```diff
@@ -212,3 +214,21 @@ func GetShoot(ctx context.Context, c client.Client, namespace string) (*gardenco
 func NewGardenDecoder() runtime.Decoder {
 	return serializer.NewCodecFactory(gardenScheme).UniversalDecoder()
 }
+
+// ClusterFromRequest returns the cluster resource for a given reconcile.Request
+func ClusterFromRequest(ctx context.Context, seedClient client.Client, req reconcile.Request) (*extensionsv1alpha1.Cluster, error) {
```
rfranzke, line 219: unit test?

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
```diff
@@ -165,7 +169,7 @@ func (c *Controller) runReconcileShootFlow(ctx context.Context, o *operation.Ope
 		deployInfrastructure = g.Add(flow.Task{
 			Name:         "Deploying Shoot infrastructure",
 			Fn:           flow.TaskFn(botanist.DeployInfrastructure).RetryUntilTimeout(defaultInterval, defaultTimeout),
-			Dependencies: flow.NewTaskIDs(deploySecrets, deployCloudProviderSecret, deployReferencedResources),
+			Dependencies: flow.NewTaskIDs(deploySecrets, deployCloudProviderSecret, deployReferencedResources, ensureShootExtensionStatus),
```
rfranzke, line 172: Why is "only" this task depending on `ensureShootExtensionStatus`? I suggest to remove it here because, consequently, you would have to add it to all other tasks creating extension resources as well.

pkg/operation/botanist/component/shootextensionstatus/shoot_extension_status.go
```diff
@@ -0,0 +1,78 @@
...
+		shoot:        shoot,
+	}
+}
+
+type shootExtensionStatus struct {
+	gardenClient client.Client
+	shoot        *gardencorev1beta1.Shoot
+}
+
+func (m *shootExtensionStatus) Deploy(ctx context.Context) error {
```
rfranzke, line 50:
```suggestion
func (s *shootExtensionStatus) Deploy(ctx context.Context) error {
```

pkg/operation/botanist/component/shootextensionstatus/shoot_extension_status.go
```diff
@@ -0,0 +1,78 @@
...
+) ShootExtensionStatus {
+	return &shootExtensionStatus{
+		gardenClient: client,
+		shoot:        shoot,
+	}
+}
+
+type shootExtensionStatus struct {
+	gardenClient client.Client
+	shoot        *gardencorev1beta1.Shoot
```
rfranzke, line 47: You only seem to require namespace/name

pkg/operation/botanist/component/shootextensionstatus/shoot_extension_status.go
```diff
@@ -0,0 +1,78 @@
...
+	ownerReference.BlockOwnerDeletion = &blockOwnerDeletion
+
+	_, err := controllerutils.StrategicMergePatchOrCreate(ctx, m.gardenClient, status, func() error {
+		status.OwnerReferences = []metav1.OwnerReference{*ownerReference}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	return nil
```
rfranzke, line 72:
```suggestion
	return err
```

pkg/operation/botanist/component/shootextensionstatus/shoot_extension_status.go
rfranzke, line 50: Unit test?

rfranzke, line 50: Why is this implemented different than `ShootState`? To me, implementing the component interface seems unnecessary overhead and can probably be avoided.

timebertt:
Not much to add to @rfranzke's comment from my side.

pkg/gardenlet/controller/shoot/shoot_control_reconcile.go
danielfoehrKn: > I suggest to remove it here  

where would you put it instead?

pkg/gardenlet/controller/federatedseed/extensions/artifact.go
danielfoehrKn: I cannot follow. This is generic functionality that should be used by other components. How do you know that it is not need to "watch all these objects and return the status "asap"? 
Just to name a few possibilities 
 - admission plugins forbid altering of infrastructure resource during Shoot creation,....
 - Third-party that wants to act upon created infrastructure resources. 
   - Why Let them wait until Shoot is reconciled?
   - What happens if the Infrastructure is already created, but the initial creation of the Shoot fails? 
 - During Shoot deletion: Fails before Shoot can be deleted. Infra resources are already gone but still in the `ShootExtensionStatus`.

Anyways, i mentioned a couple of times in our internal syncs that I will implement it this way.
Plus this PR is now a couple of weeks old and this comment would mean to completely change the implementation + tests.
As this functionality has low priority there are much more important things to be tackled, I consider dropping this PR and revert the previous commit with the unused API.
If this becomes a higher priority in the future we can revisit it.

My lesson for this is, that I ll not accept tasks that are supposed to be done as part of a 5-day quality week that will foreseeably take much longer. This is especially true when adding new APIs that always spark discussions. 
In addition, I would appreciate if these kinds of ideas that have a huge impact on the implementation are communicated if possible in advance.

rfranzke: IIUC, the implementation leverages the already established watches for syncing the `ShootState` in the `federatedseed` controller, correct? In this case (no new watches are created), it seems to be fine to proceed with the implementation as is. You have enlisted a few cases in which immediate-syncing seems beneficial (assuming the existing watches are used).

In retrospect, even if the change seemed to be small enough, a GEP would've been the proper way for proposing this enhancement. #3873 didn't have much activity (rather none) and there was no "main reviewer" assigned to it. Also, I don't think there was a dedicated meeting about this topic and its implementation details. Probably, it cannot be assumed that there is enough bandwidth to provide such "implementation ideas" ad-hoc just because it was mentioned once or twice in a sync while everybody else is busy with other topics (unless there is a "main reviewer", of course).

Anyways, let's proceed now and finish this up. It's not necessary to revert everything. Significant time was already invested and we still would like to have the feature.

timuthy:
@danielfoehrKn as you mentioned the intention to rewrite this functionality, can we close this PR?

timuthy:
I suggest to close this PR if there is no feedback by the end of the week.

PR #3900, opened 2021-04-20
Stop running logging integration test on chost nodes

Kristian-ZH:

/area logging
/kind bug

Fluent-bit reads from `/var/log/journal` the systemd logs.
Chost nodes are using different directory by default: `/run/log/journal` so the logging stack does not work on `chost` nodes and this test is not relevant for them.

Fixes #

@vlvasilev 

**Release note**:

```other operator
None
```

### DIFF
--- a/test/integration/shoots/logging/seed_logging_stack.go
+++ b/test/integration/shoots/logging/seed_logging_stack.go
@@ -17,6 +17,7 @@ package logging
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


dguendisch:
Maybe stupid question but doesn't it make more sense to make the logging stack work on chost nodes as well? Or is that not easily possible?

Kristian-ZH:
> Maybe stupid question but doesn't it make more sense to make the logging stack work on chost nodes as well? Or is that not easily possible?

We have tried different approach here:
The Journald Path can be dynamically configured with environment variable, but the fluent-bit image is really lightweight image which does not have shell, bash, ... . So we can not even run such a code in it: 
```
JOURNALD_PATH=`/var/log/journal`
if [ ! -d ${JOURNALD_PATH} ] ; then JOURNALD_PATH=`/run/log/journal`; fi
``` 
The only option is to wrap the image into bigger one with more tools `alpine` for example, but we are not sure if it is worth, because we will have to support this image after that.

There is no way with k8s also to read this variable from file for example (If init-container puts it to a file before that).
The only supported ways are from Secret, Configmap and Container spec. 

And also we can not make the property configurable by the Gardenlet ( with the values file), because a WorkerPool can have multiple workers with different images.

dguendisch (commented):
/lgtm

dguendisch (dismissed):

rfranzke (commented):
/invite @vpnachev

test/integration/shoots/logging/seed_logging_stack.go
```diff
@@ -80,6 +81,12 @@ var _ = ginkgo.Describe("Seed logging testing", func() {
 	lokiPriorityClass := &schedulingv1.PriorityClass{}
 
 	framework.CBeforeEach(func(ctx context.Context) {
+		for _, worker := range f.Shoot.Spec.Provider.Workers {
+			if strings.Contains(worker.Machine.Image.Name, "chost") {
+				ginkgo.Skip("Skipping logging intergation test for chost nodes")
```
schrodit: nit
```suggestion
				ginkgo.Skip("Skipping logging integration test for chost nodes")
```

vpnachev:
What do you think about using the `X.Y.Z-debug` version of the fluent bit image? It is more or less the same image but with `busybox` and this will provide you with shell to evaluate properly the value of `JOURNALD_PATH`.

Kristian-ZH:
Yes, this is possible. Then we will discuss this with @vlvasilev  and will open another PR with the new change.

/close

Kristian-ZH:
/open

As the debug image also works in debug mode, which is not the behaviour we expect, we came back to this PR

dguendisch (approved):

schrodit (approved):

Issue #1724, opened 2019-12-13
Harden Gardenlet RBAC privileges in the seed cluster


rfranzke:
With #1601 we have introduced the gardenlet, a component comparable with the kubelet in Kubernetes.

However, in this first introduction we have not hardened the `ClusterRole`s we bind the gardenlet to in the seed ([ref](https://github.com/gardener/gardener/blob/1e591cc942e3c968407ace20f3e9cc0f183f7438/charts/gardener/gardenlet/charts/runtime/templates/rbac.yaml#L5-L22)).

This is true for both the garden and the seed cluster.
For the garden cluster, we plan to address this with a special authorization plugin, see #1724.
For the seed cluster, we should harden these RBAC roles in order to only allow access to those resources it needs to access.

rfranzke:
/assign @vpnachev @ialidzhikov 

First step would be to record the gardenlet's activities on the seed and improve its RBAC role (https://github.com/gardener/gardener/blob/master/charts/gardener/gardenlet/charts/runtime/templates/rbac.yaml#L16-L22).
@vpnachev suggested to use https://github.com/liggitt/audit2rbac to convert the recorded audit logs into RBAC rules.

/security
/normal
/in-progress

timebertt:
FYI: with https://github.com/gardener/gardener/pull/3535 the gardenlet also needs permission for `leases` in the seed.

PR #3763, opened 2021-03-23
Add fine-grained RBACs for gardenlet in the seed cluster

vpnachev:
**How to categorize this PR?**

/area security
/kind enhancement

**What this PR does / why we need it**:
Configures `gardenlet` with more fine-grained permissions in the seed cluster.

**Which issue(s) this PR fixes**:
Part of #1724 

**Special notes for your reviewer**:

**Release note**:

```other operator
The RBACs for the `gardenlet` in the seed cluster are now fine-grained defined and limited as much as possible.
```

PR #3901, opened 2021-04-20
Gardener components no longer use admin kubeconfig for local garden dev setup

rfranzke:

/area dev-productivity
/kind enhancement

Today, all Gardener components are using the admin kubeconfig for the local garden development setup. This has the downside that the RBAC rules which we maintain in 

Softly related to #1723 and #1724.

/squash

**Release note**:

```noteworthy developer
When using the local garden development environment, the Gardener components do now use dedicated kubeconfigs constrained by RBAC rules (earlier, they were always using the admin kubeconfig).
```


rfranzke:
@vpnachev - an idea as follow-up of this PR to also make the gardenlet use the RBAC privileges in the seed that you plan to apply with #3763:

In https://github.com/gardener/gardener/blob/master/hack/local-development/start-gardenlet, 

1. create a `ServiceAccount` in the `garden` namespace of all the seed clusters (by reading all `Seed`s and using the kubeconfig referenced by `.spec.secretRef`)
2. apply the RBAC resources for the gardenlet in the seeds
3. create a new, temporary kubeconfig which uses the token of the `ServiceAccount` created in 1.
4. update the kubeconfig in the `.spec.secretRef` of the various seeds

It might have drawbacks, just brain-storming here. WDYT?

vpnachev:
> @vpnachev - an idea as follow-up of this PR to also make the gardenlet use the RBAC privileges in the seed that you plan to apply with #3763:
> 
> In https://github.com/gardener/gardener/blob/master/hack/local-development/start-gardenlet,
> 
>     1. create a `ServiceAccount` in the `garden` namespace of all the seed clusters (by reading all `Seed`s and using the kubeconfig referenced by `.spec.secretRef`)
> 
>     2. apply the RBAC resources for the gardenlet in the seeds
> 
>     3. create a new, temporary kubeconfig which uses the token of the `ServiceAccount` created in 1.
> 
>     4. update the kubeconfig in the `.spec.secretRef` of the various seeds
> 
> 
> It might have drawbacks, just brain-storming here. WDYT?

Thanks for the suggestion! However, I am still thinking that it should be as it is today. Why
1. A new seed can be added at any time and gardenlet will use the configured kubeconfig. On restart of gardenlet, it will create the service account and bound the (cluster)roles to it. This can lead to authorization issues if new permissions are required, or if the seed is deleted with no restart of gardenlet such missing permissions will not be detected.
1. It still possible to run gardenlet remotely in productive landscapes with the whatever kubeconfig is configured by the landscape administrator. Should we implement such automation in this case - I think no. It would be a bit weird gardenlet to amend externally configured kubeconfig.

For now I will not add anything like this in #3763 as it has been prolonged too much and a lot of conflicts needs to be resolved. Such automation can be contributed later, if we find a suitable solution.

rfranzke:
@vpnachev My recommendation was only for the local development environment. Of course, in production code we should not have anything like this. I just think we should find a way to make sure to detect issues in the gardenlet's RBAC rules already when developing locally. We don't have to follow above approach (it was just some short brain-storming/food for thought), and we don't have to do anything now with #3763, but I think we should eventually have some solution for it.

charts/gardener/controlplane/charts/application/templates/clusterrole-apiserver.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.global.controller.enabled }}
+---
+apiVersion: {{ include "rbacversion" . }}
+kind: ClusterRole
+metadata:
+  name: gardener.cloud:system:gardener-controller-manager
```
stoyanr:
```suggestion
  name: gardener.cloud:system:gardener-apiserver
```

charts/gardener/controlplane/charts/application/templates/clusterrole-controller-manager.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.global.apiserver.enabled }}
+---
+apiVersion: {{ include "rbacversion" . }}
+kind: ClusterRole
+metadata:
+  name: gardener.cloud:system:gardener-apiserver
```
stoyanr:
```suggestion
  name: gardener.cloud:system:gardener-controller-manager
```

charts/gardener/controlplane/charts/application/templates/clusterrole-apiserver.yaml
```diff
@@ -0,0 +1,19 @@
...
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+rules:
+- apiGroups:
+  - "*"
+  resources:
+  - "*"
+  verbs:
+  - "*"
```
stoyanr, line 18: Are these permissions (and similarly for other cluster roles introduced) going to be narrowed in a later PR? If not, how exactly having a dedicated role for e.g. apiserver helps with security hardening?

hack/local-development/local-garden/start.sh
```diff
@@ -42,6 +42,9 @@ for i in 1..10; do
   echo "# Waiting until Kube-Apiserver is available"
 done
 
+echo "# Configuring RBAC resources for Gardener components"
+$(dirname $0)/configure-rbac
+
```
stoyanr, line 47: This change is only done for the `local-garden` setup now. How about `remote-garden`, don't we need something similar there?

hack/local-development/start-admission-controller
```diff
@@ -16,9 +16,13 @@
 
 source $(dirname "${0}")/common/helpers
 
-mktemp_kubeconfig
-kubeconfig="$__tmp_kubeconfig"
-trap cleanup_kubeconfig EXIT
+if [[ $(k8s_env) == "$NODELESS" ]]; then
```
stoyanr, line 19: Why is this change only done for `$NODELESS` and not for the other environment types, e.g. `$REMOTE`? Wouldn't it have some consequences on testing locally using these environment types, e.g. perhaps seed authorizer not working?

stoyanr (changes requested):
Few minor nits and some questions regarding why is this enabled only for `local-garden / $NODELESS` as opposed to other local environment types, e.g. `remote-garden / $REMOTE`.

charts/gardener/controlplane/charts/application/templates/clusterrole-apiserver.yaml
rfranzke: Yes, I think we should eventually narrow them down as well for GAPI/GCM, for sure. Today, both components were bound to the `cluster-admin` `ClusterRole`. The newly introduced roles are equivalent for now, but they can be used in the future to harden also the rules.

hack/local-development/local-garden/start.sh
rfranzke: I don't know how to achieve this for the `remote-garden`, unfortunately. In this case, we don't control the Kubernetes cluster and do not have access to the CA certificate in order to issue new client certs for the Gardener components. While we could use e.g. `ServiceAccount`s for GAPI,GCM,GSCHED,GAC, we have to use a client cert for gardenlet because of #1723 and how the seed identification works (not achievable with `ServiceAccount`s).

hack/local-development/start-admission-controller
rfranzke: > Why is this change only done for $NODELESS and not for the other environment types, e.g. $REMOTE?

Mainly due to https://github.com/gardener/gardener/pull/3901#discussion_r618116532.

> Wouldn't it have some consequences on testing locally using these environment types, e.g. perhaps seed authorizer not working?

WDYM? Seed Authorizer does only work in local-garden because only here we can configure the kube-apiserver with the auth-webhook flags.

hack/local-development/local-garden/start.sh
stoyanr: A `remote-garden` is usually just a shoot in our `dev` landscape, or in the general case a cluster that the developer owns. In this case, the developer also can get hold of the CA certificate and e.g. pass it as a parameter to a script, or am I missing something?

hack/local-development/start-admission-controller
stoyanr: A `remote-garden` is usually just a shoot in our `dev` landscape, or in the general case a cluster that the developer owns. In this case, why wouldn't it be possible to configure `kube-apiserver` with the right flags? OK, it's probably not something we support right now for shoots, but is it just functionality that is missing or a more general limitation?

hack/local-development/local-garden/start.sh
stoyanr: We discussed this and it became clear that even though it would be possible, it won't be used much and it would be inconsistent for `gardenlet` due to the requirement to have the `Webhook` authorization flag. So let's not do it here, it could be added later if there is a strong need.

rfranzke: As discussed via phone, let's not invest into the remote garden setup for now because the way forward is not exactly clear, especially whether the effort we'd put is worth the benefits (this is also with the assumption that the local-garden is the default/most common development environment).
I added a disclaimer for the remote garden with 81041e3.

hack/local-development/start-admission-controller
rfranzke: As discussed via phone, it's probably rather a general limitation (of managed Kubernetes services in general), i.e., we would not want to enable end-users to inject auth webhooks, especially because this would mean that we needed to run untrusted code in the seed (next to the shoot's control plane) for low latency. There might be many other drawbacks and limitations, so we don't plan to invest into it which makes the remote garden development as outlined above quite complicated (assuming that such remote clusters are shoots).

stoyanr: We discussed this. Even though it could be possible to add such functionality for shoots we will probably never do it due to the security and other implications. So even if it's enabled for `$REMOTE`, there would be no easy way for us to take advantage of it.

We also thought it would be a good idea to add a sentence to the documentation stating that the remote garden setup doesn't support the seed authorizer and therefore should be used with care.

stoyanr (approved):
/lgtm

Issue #3898, opened 2021-04-20
☂️ issue: Use dedicated bootstrap-tokens per shoot worker machine (instead of a long-valid shared token)


BeckerMax:
## Use dedicated bootstrap-tokens per shoot worker machine (instead of a long-valid shared token)

_Background:_
When a new node of a shoot cluster is created the kubelet needs to register with the shoot api server. The kubelet uses a bootstrap-token for authentication.

_New:_
In the proposed new flow the worker controller (here MCM)  creates a temporary bootstrap-token for each created vm. 
If the featureFlag `BootstrapTokenForVMs` is enabled a file with the content "<<BOOTSTRAP_TOKEN>>" is added to the operatingsystem-config. It is passed to the worker controller. The worker controller generates a temporary token for every new worker instance(node). It replaces the  "<<BOOTSTRAP_TOKEN>>" string with the created token and adds it to the user-data placed on the vm on startup.
The cloud-config-downloader(original user-data) will then refer to the new temporary bootstrap-token in the kubelet-bootstrap script.

See the depicted flow in the following graph:

![final_secure_bootstrap_token](https://user-images.githubusercontent.com/7465180/115367041-1ef68100-a1c6-11eb-82bf-242930d9c0a5.png)

For old flow see also https://github.com/gardener/gardener/blob/a62e5622facd20d70c069c17bf42947dbea427ef/docs/extensions/operatingsystemconfig.md#how-does-gardener-bootstrap-the-machines

**Steps:**

Adapt worker controller to add bootstrap-token on vm creation:
- [x]  MCM -> merged: https://github.com/gardener/machine-controller-manager/pull/351
- [x] Verify if there are other worker controllers needed to be adapted 
- [x] Adapt worker controller contract to include handling bootstrap-token

Test and adapt os-extensions to support passing the <<BOOTSTRAP_TOKEN>> unencoded
- [x]   gardenlinux -> https://github.com/gardener/gardener-extension-os-gardenlinux/pull/29
- [x]   suse-chost -> https://github.com/gardener/gardener-extension-os-suse-chost/pull/41
- [x] ubuntu -> https://github.com/gardener/gardener-extension-os-ubuntu/pull/42
- [x] flatcar -> https://github.com/gardener/gardener-extension-os-coreos/pull/24

Adapt g/g code to support new bootstrap-token:
- [x] PR opened: https://github.com/gardener/gardener/pull/3902

Next steps:
- [x] release all os-extensions.
[...]

PR #3902, opened 2021-04-20
Enh/bootstrap tokens for machines

BeckerMax:
Add support for short-lived bootstrap tokens for nodes in g/g. 
For more information see https://github.com/gardener/gardener/issues/3898

/area security
/kind enhancement

**Release note**:
```other operator
Gardener now supports using worker-controller generated bootstrap-tokens for machines, see new flow [here|https://github.com/gardener/gardener/docs/extensions/operatingsystemconfig.md#detailed-bootstrap-flow-with-a-worker-generated-bootstrap-token]. ⚠️ If you maintain an infrastructure extension make sure to use a worker controller that supports generating a bootstrap token, see [here|https://github.com/gardener/machine-controller-manager/pull/351] and if you maintain an os-extension make sure to support the `transmitUnencoded` flag similar to [os-gardenlinx|https://github.com/gardener/gardener-extension-os-gardenlinux/pull/29]. Currently, the old flow is still supported but we plan to deprecate it in the future.
```

This PR lets the kubelet use the worker generated token on startup if it exists in the userData.

** Notes to the Reviewer **
The `transmitUnencoded` flag is needed as the MCM expects the string "<<BOOTSTRAP_TOKEN>>" in clear text in the user data. The contract of the os-extension is that all content is sent base64 encoded to the machine. The flag `transmitUnencoded` allows sending clear text(unencoded) content to the machine making it easier to be manipulated on the way. This also requires an adaptation in the extension-controller. See: https://github.com/gardener/gardener-extension-os-gardenlinux/pull/29

/squash


docs/deployment/feature_gates.md
```diff
@@ -73,3 +74,4 @@ A *General Availability* (GA) feature is also referred to as a *stable* feature.
 * `MountHostCADirectories` enables mounting common CA certificate directories in the Shoot API server pod that might be required for webhooks or OIDC.
 * `SeedChange` enables updating the `spec.seedName` field during shoot validation from a non-empty value in order to trigger shoot control plane migration.
 * `SeedKubeScheduler` adds custom `kube-scheduler` in `gardener-kube-scheduler` namespace. It schedules [pods with scheduler name](../concepts/seed-admission-controller.md#mutating-webhooks) `gardener-kube-scheduler` on Nodes with higher resource utilization. It requires Seed cluster with kubernetes version `1.18` or higher.
+* `BootstrapTokenForVMs` enables a security enhancement for the bootstrap flow of the kubelet. The bootstrap-token is now uniquely generated for each VM and invalidated after successful creation. This requires support in the worker controller (i.e. machine-controller-manager >`v0.26.0`) and in the os-extension (i.e. gardener-extension-os-gardenlinux >`v.0.9.0`)
```
rfranzke: What about rather calling this feature gate `BootstrapTokenProvidedByWorker` or something similar to make it more clear?

docs/extensions/operatingsystemconfig.md
```diff
@@ -42,6 +42,18 @@ The high-level flow is as follows:
 
 1. After the VM has been provisioned the `downloader` script starts and fetches the appropriate `Secret` for its worker pool (containing the "original" user-data) and applies it.
 
+### Detailed bootstrap flow with the feature gate `BootstrapTokenForVMs` enabled
+
+If the featureFlag `BootstrapTokenForVMs` is enabled a file with the content "<<BOOTSTRAP_TOKEN>>" is added to the operatingsystem-config.
```
rfranzke:
```suggestion
If the feature gate `BootstrapTokenForVMs` is enabled a file with the content `<<BOOTSTRAP_TOKEN>>` is added to the `OperatingSystemConfig`.
```

docs/extensions/operatingsystemconfig.md
```diff
@@ -42,6 +42,18 @@ The high-level flow is as follows:
...
+It is passed to the worker controller (here MCM).
```
rfranzke: MCM is not the worker controller.

docs/extensions/operatingsystemconfig.md
```diff
@@ -42,6 +42,18 @@ The high-level flow is as follows:
...
+
+The worker controller generates a temporary token for every new worker instance(node).
```
rfranzke:
```suggestion
The worker controller has to generate a temporary bootstrap token for every new worker instance(node).
```

rfranzke: Actually, from the extension API itself, this description it not quite accurate. Of course, this is the recommended implementation, however, contract-wise the `Worker` controller could also do it like Gardener does it today, i.e., generating one bootstrap token for all VMs if appropriate.

docs/extensions/operatingsystemconfig.md
```diff
@@ -42,6 +42,18 @@ The high-level flow is as follows:
...
+It replaces the "<<BOOTSTRAP_TOKEN>>" string with the created token and adds it to the user-data placed on the vm on startup.
```
rfranzke:
```suggestion
It replaces the `<<BOOTSTRAP_TOKEN>>` string with the created bootstrap token and adds it to the user-data placed on the vm on startup.
```

pkg/operation/botanist/component/extensions/operatingsystemconfig/downloader/downloader.go
```diff
@@ -94,6 +96,10 @@ const (
 	// PathCredentialsClientKey is a constant for a path containing the 'client private key' credentials part for the
 	// download.
 	PathCredentialsClientKey = PathCredentialsDirectory + "/client.key"
+	// PathBootstrapToken Path to machine bootstraptoken
+	PathBootstrapToken = PathCredentialsDirectory + "/bootstrap-token"
+	// BootstrapTokenString is the token that is expected to be replaced by the worker controller with the actual token
+	BootstrapTokenString = "<<BOOTSTRAP_TOKEN>>"
```
rfranzke:
```suggestion
	BootstrapTokenPlaceholder = "<<BOOTSTRAP_TOKEN>>"
```

pkg/operation/botanist/component/extensions/operatingsystemconfig/downloader/downloader_test.go
```diff
@@ -87,7 +93,33 @@ var _ = Describe("Downloader", func() {
...
+					Content: extensionsv1alpha1.FileContent{
+						Inline: &extensionsv1alpha1.FileContentInline{
+							Encoding: "b64",
+							Data:     utils.EncodeBase64([]byte("<<BOOTSTRAP_TOKEN>>")),
+						},
+						TransmitUnencoded: pointer.BoolPtr(true),
+					},
+				}))
+		})
+		It("should not include the bootstrap_token file (w/o BootstrapTokenForVMs)", func() {
```
rfranzke:
```suggestion

		It("should not include the bootstrap_token file (w/o BootstrapTokenForVMs)", func() {
```

pkg/operation/botanist/component/extensions/operatingsystemconfig/downloader/downloader.go
```diff
@@ -94,6 +96,10 @@ const (
 	// PathCredentialsClientKey is a constant for a path containing the 'client private key' credentials part for the
 	// download.
 	PathCredentialsClientKey = PathCredentialsDirectory + "/client.key"
+	// PathBootstrapToken Path to machine bootstraptoken
```
rfranzke: this is not a sentence

pkg/operation/botanist/component/extensions/operatingsystemconfig/executor/templates/scripts/execute-cloud-config.tpl.sh
```diff
@@ -88,11 +94,14 @@ users:
 - name: kubelet-bootstrap
   user:
     as-user-extra: {}
-    token: {{ .bootstrapToken }}
+    token: "$BOOTSTRAP_TOKEN"
 EOF
 
 else
   rm -f "{{ .pathKubeletKubeconfigBootstrap }}"
+  if [[ -f "{{ .pathBootstrapToken }}" ]]; then
```
rfranzke: The `if` should not be needed here, `rm -f` handles this automatically

pkg/operation/botanist/component/extensions/operatingsystemconfig/executor/templates/scripts/execute-cloud-config.tpl.sh
```diff
@@ -68,6 +68,12 @@ fi
 
 md5sum ${PATH_CCD_SCRIPT} > ${PATH_CCD_SCRIPT_CHECKSUM}
 
+BOOTSTRAP_TOKEN="{{ .bootstrapToken }}"
+# If a worker controller created bootstrapToken exists then use it
```
rfranzke:
```suggestion
# If a bootstrap token created by the Worker extension exists then use it
```

pkg/operation/botanist/component/extensions/operatingsystemconfig/executor/templates/scripts/execute-cloud-config.tpl.sh
```diff
@@ -68,6 +68,12 @@ fi
...
+if [[ -f "{{ .pathBootstrapToken }}" ]]; then
+  BOOTSTRAP_TOKEN=$(cat "{{ .pathBootstrapToken }}")
```
rfranzke:
```suggestion
  BOOTSTRAP_TOKEN="$(cat "{{ .pathBootstrapToken }}")"
```

rfranzke (dismissed):
/assign

Great PR, looks good overall, I mainly have nit comments. However, I'm missing the code which would make Gardener no longer compute a bootstrap token if the feature gate is enabled.

pkg/operation/botanist/component/extensions/operatingsystemconfig/downloader/downloader.go
```diff
@@ -197,6 +203,21 @@ WantedBy=multi-user.target`),
...
 
+	if gardenletfeatures.FeatureGate.Enabled(features.BootstrapTokenForVMs) {
+		files = append(files,
+			extensionsv1alpha1.File{
+				Path:        PathBootstrapToken,
+				Permissions: pointer.Int32Ptr(0644),
+				Content: extensionsv1alpha1.FileContent{
+					Inline: &extensionsv1alpha1.FileContentInline{
+						Encoding: "b64",
+						Data:     utils.EncodeBase64([]byte(BootstrapTokenString)),
```
danielfoehrKn: You can remove the base64 encoding to make it more clear.

danielfoehrKn (dismissed):
Looks good.
Can you later add an entry in the compatibility matrix of the os-extension providers? Because you need a certain version of the os-extension to be able to work with `TransmitUnencoded`

Similar to [here](https://github.com/gardener/gardener-extension-provider-aws).

docs/extensions/operatingsystemconfig.md
BeckerMax: Got it. The MCM is generated by our worker controllers.

pkg/operation/botanist/component/extensions/operatingsystemconfig/executor/templates/scripts/execute-cloud-config.tpl.sh
BeckerMax: you are right. It exits with 0. Thanks, didn't know that.

BeckerMax:
@rfranzke Thanks for your review! I addressed your remarks. Regarding:

> However, I'm missing the code which would make Gardener no longer compute a bootstrap token if the feature gate is enabled.

If I adapt the [logic](https://github.com/gardener/gardener/blob/ce40aae8c31535e17772f65298af7c1aacf65b8a/pkg/operation/botanist/operatingsystemconfig.go#L133) this would lead to something I wanted to avoid: Some fresh VMs during migration might time out. Scenario:
1. FeatureGate enabled
2. VM gets created with the old user-data (shoot has not yet been reconciled). Meaning it does **not** have a file with the worker generated BOOTSTRAP_TOKEN.
3. shoot gets reconciled
4. cloud-config secret gets updated and does not include the bootstrap-token anymore.
5. VM downloads cloud-config with updated secret
5. VM creation times out after 20mins

I prefer to not create the chance of one-time VM timeouts.

My way forward would be instead:
1. Still enable the old token. Allow a smooth migration.
2. Wait some time
3. Deprecate featureGate and make it true by default
4. Remove logic of generating the old token

I am also open for discussion.

BeckerMax:
@danielfoehrKn Yes, I guess I need to add it to all the infrastructure extension-providers as well as to the os-extension providers.

pkg/operation/botanist/component/extensions/operatingsystemconfig/downloader/downloader.go
```diff
@@ -197,6 +204,20 @@ WantedBy=multi-user.target`),
 		},
 	}
 
+	if gardenletfeatures.FeatureGate.Enabled(features.BootstrapTokenProvidedByWorker) {
```
rfranzke: Actually, now having a second look, I'm wondering why we need this feature gate here in the first place. Can't we always create this file? You could adapt the executor script in the way that it does only consider the bootstrap token file if its content is not equal to the placeholder `<<BOOTSTRAP_TOKEN>>`. WDYT?

BeckerMax: As we want the worker bootstrap-token to be the default I agree with your proposal.
But: What if an os-extension version is used that does not know the `TransmitUnencoded` flag? I need to test whether it is just ignored, or whether this would lead to any problems (also: if an outdated os-extension accepts it then it would place it encoded on the VM).

rfranzke: Good point, indeed. It seems the executor scripts need to be slightly "more intelligent", i.e., maybe it should check if the content of the file matches the regexp of a valid bootstrap token. Neither `<<BOOTSTRAP_TOKEN>>` nor `base64(<<BOOTSTRAP_TOKEN>>)` would pass.

BeckerMax: Looking again at this (always adding the file and removing the feature gate check there) is not solving any problems, or? It just pushes validation logic from here into the shell script. Why not leave it explicit in the gardener code. If the feature gate is `false` this file is never used - so we can also not add it. It is also clearer to the person reading the code that this file is just relevant if you enable the feature gate.

rfranzke: IMO, this feature gate adds unnecessary complexity and avoidable cognitive overhead for operators during the deployment/configuration of gardenlet and the extensions. We are perfectly able to deliver this feature in a backwards-compatible way by enhancing the executor script, preventing mentioned downsides.
I guess the changes in the script would be minimal/just a few lines, so I'm not sure if any arguments w.r.t. code readability/maintainability apply/outweigh the mentioned complexity.

BeckerMax:
We(cc @rfranzke)  discussed the topic and decided to remove the featureGate. Reason is that setting the feature gate to `false` will result in unresponsive VMs until the Shoots are reconciled. Failing flow:
- FeatureGate set to `true`
- After some time tokens of 'old flow' with the validity of 24hs will be removed.
- FeatureGate set to `false`
- g/g will no longer add a file with string "<<BOOTSTRAP_TOKEN>>" to the os-config userdata
- worker controller will no longer place generated 'bootstrap_token' into the userdata sent to the VM
- *Until* the shoot gets reconciled no token using the 'old-way' will be generated 
- Therefore, all VMs will time out on creation until the shoots get reconciled.

How we move forward:
1. Remove featureGate
2. Gardenlet will always try to generate a file with string "<<BOOTSTRAP_TOKEN>>"
3. If available the kubelet uses the worker generated bootstrap-token on start-up
4. After some releases remove support for creating the token the old way and inform the users that with this g/g version a compatible version of the os-extension-provider and the infrastructure-extension is required.

BeckerMax:
PTAL @rfranzke @danielfoehrKn 
With the reasoning mentioned in the comment above I removed the feature flag with the recent commit.

I tested it with an os-provider-extension version which does not support the `transmit-unencoded` flag. Then the kubelet bootstrap will use the old shared token.
In case the os-config extension supports the flag then the worker generated token is used.

An outdated MCM is not really an issue as the feature of creating a bootstrap_token has been included in the provider extension since 1.5 years. It should work with an outdated version anyways though.

PR would be ready to merge from my side, except I wanna streamline the commit messages after you reviewed it.

rfranzke:
@BeckerMax Can you check the failing PR check?

timebertt:
> @BeckerMax This PR contains commits including `@mentions` in the commit message: 3d2131c5b2780fc9d233b0205778212b42224721, 04f0607ea34c447da2a3367d760596e6f3c9ee84
Please remove the mentions from the commit messages to prevent notifying the mentioned users every time the commit is pushed to some fork of this repository.

docs/extensions/operatingsystemconfig.md
```diff
@@ -42,6 +42,23 @@ The high-level flow is as follows:
 
 1. After the VM has been provisioned the `downloader` script starts and fetches the appropriate `Secret` for its worker pool (containing the "original" user-data) and applies it.
 
+### Detailed bootstrap flow with a worker generated bootstrap-token
+
+With gardener v1.23 a file with the content "<<BOOTSTRAP_TOKEN>>" is added to the operatingsystem-config.
```
rfranzke:
```suggestion
With gardener v1.23 a file with the content `<<BOOTSTRAP_TOKEN>>` is added to the `OperatingSystemConfig`.
```

Which? There are two of such resources for a shoot.

docs/extensions/operatingsystemconfig.md
```diff
@@ -42,6 +42,23 @@ The high-level flow is as follows:
...
+It is passed to the worker controller.
```
rfranzke:
```suggestion
It is passed to the `Worker` controller.
```

docs/extensions/operatingsystemconfig.md
```diff
@@ -42,6 +42,23 @@ The high-level flow is as follows:
...
+
+The worker controller has to guarantee that:
+- a temporary bootstrap token is created.
+- the `"<<BOOTSTRAP_TOKEN>>"` in the user data is replaced by the generated token.
```
rfranzke:
```suggestion
- the `<<BOOTSTRAP_TOKEN>>` in the user data is replaced by the generated token.
```

docs/extensions/operatingsystemconfig.md
```diff
@@ -42,6 +42,23 @@ The high-level flow is as follows:
 
 1. After the VM has been provisioned the `downloader` script starts and fetches the appropriate `Secret` for its worker pool (containing the "original" user-data) and applies it.
 
+### Detailed bootstrap flow with a worker generated bootstrap-token
+
+With gardener v1.23 a file with the content "<<BOOTSTRAP_TOKEN>>" is added to the operatingsystem-config.
+It is passed to the worker controller.
+
+The worker controller has to guarantee that:
```
rfranzke:
```suggestion
The `Worker` controller has to guarantee that:
```

docs/extensions/operatingsystemconfig.md
```diff
@@ -42,6 +42,23 @@ The high-level flow is as follows:
...
+- a temporary bootstrap token is created.
+- the `"<<BOOTSTRAP_TOKEN>>"` in the user data is replaced by the generated token.
+One implementation of that is depicted in the picture where the MCM creates a temporary token and replaces the placeholder.
```
rfranzke:
```suggestion
One implementation of that is depicted in the picture where the machine-controller-manager creates a temporary token and replaces the placeholder.
```

docs/extensions/operatingsystemconfig.md
```diff
@@ -42,6 +42,23 @@ The high-level flow is as follows:
...
+
+As part of the user-data the bootstrap-token is placed on the newly created VM under a defined path.
+The cloud-config-script will then refer to the file path of the added bootstrap token in the kubelet-bootstrap script.
+
+If either the worker controller does not replace the `<<BOOTSTRAP_TOKEN>>` or the os-provider-extension does not support the `transmit-unencoded` flag then the bootstrap flow will fall back to the old flow of using a shared bootstrap_token between workers of a Shoot.
```
rfranzke:
```suggestion
If either the `Worker` controller does not replace the `<<BOOTSTRAP_TOKEN>>` or the OS extension does not support the `transmitUnencoded` flag then the bootstrap flow will fall back to the old flow of using a shared bootstrap token between the worker nodes of a shoot cluster.
```

pkg/operation/botanist/component/extensions/operatingsystemconfig/downloader/downloader.go
```diff
@@ -94,6 +94,11 @@ const (
 	// PathCredentialsClientKey is a constant for a path containing the 'client private key' credentials part for the
 	// download.
 	PathCredentialsClientKey = PathCredentialsDirectory + "/client.key"
+	// PathBootstrapToken is the path on the shoot worker nodes at which the the bootstrap token for the kubelet
```
rfranzke:
```suggestion
	// PathBootstrapToken is the path of a file on the shoot worker nodes in which the the bootstrap token for the kubelet
```

pkg/operation/botanist/component/extensions/operatingsystemconfig/downloader/downloader.go
```diff
@@ -94,6 +94,11 @@ const (
...
+	// bootstrap are stored
```
rfranzke:
```suggestion
	// bootstrap is stored.
```

rfranzke (commented):
Very nice @BeckerMax, I've only suggestions regarding styling in the docs.
Also, please provide a meaningful release note for extension developers. :)

rfranzke (approved):
/lgtm

@danielfoehrKn please re-review or dismiss

Issue #3889, opened 2021-04-19
Replace Prometheus VPA with HVPA


wyb1:
**How to categorize this issue?**

/area monitoring
/kind enhancement
/priority 3

**What would you like to be added**:

The Prometheus VPA should be replaced with HVPA similar to how [loki uses HVPA](https://github.com/gardener/gardener/pull/3025) instead of VPA.

**Why is this needed**:

Better control of Prometheus scaling

rfranzke:
@wyb1 are you still planning to follow-up on HVPA for Prometheus for shoot control planes?

wyb1:
Yes, I am planning on adding HVPA for the shoot Prometheus as a feature gate. Will have to implement this and do some testing to make sure there is no operational effort needed for the migration.

rfranzke:
Thanks! What's the reason for the feature gate?

wyb1:
To allow easy rollback if we find out Prometheus gets stuck in a `CrashLoopBackOff` when migrating (we experienced this when we added hvpa to the seed-bootstrap). This way we will not break most shoot prometheus instances.

rfranzke:
@wyb1 What about this one? There was no activity since long, do you still have plans for introducing the HVPA?

PR #3903, opened 2021-04-20
Add hvpa for Prometheus deployed by seed-bootstrap

wyb1:

/area monitoring
/kind enhancement

Add HVPA for all prometheus instances managed by `seed-bootstrap`

Partial fix for #3889

Once hvpa is rolled out for the Prometheus in the shoot control plane the above issue will be fixed.

I kept the original vpa objects in case hvpa is disabled.

**Release note**:

```other operator
Add HVPA for all prometheus instances managed by `seed-bootstrap`
```

### DIFF
--- a/pkg/operation/seed/seed.go
+++ b/pkg/operation/seed/seed.go
@@ -350,6 +350,16 @@ func RunReconcileSeedFlow(
 		if err := common.DeleteHvpa(ctx, k8sSeedClient, v1beta1constants.GardenNamespace); client.IgnoreNotFound(err) != nil {
 			return err
 		}
+	} else {
+		// Clean up stale vpa objects
+		resources := []client.Object{
+			&autoscalingv1beta2.VerticalPodAutoscaler{ObjectMeta: metav1.ObjectMeta{Name: "prometheus-vpa", Namespace: v1beta1constants.GardenNamespace}},
+			&autoscalingv1beta2.VerticalPodAutoscaler{ObjectMeta: metav1.ObjectMeta{Name: "aggregate-prometheus-vpa", Namespace: v1beta1constants.GardenNamespace}},
+		}
+
+		if err := kutil.DeleteObjects(ctx, k8sSeedClient.Client(), resources...); err != nil {
+			return err
+		}
 	}
 
 	// Fetch component-specific dependency-watchdog configuration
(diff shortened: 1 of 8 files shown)


pkg/operation/seed/seed.go
```diff
@@ -350,6 +350,18 @@ func RunReconcileSeedFlow(
...
+		resources := []client.Object{
+			&autoscalingv1beta2.VerticalPodAutoscaler{ObjectMeta: metav1.ObjectMeta{Name: "prometheus-vpa", Namespace: v1beta1constants.GardenNamespace}},
+			&autoscalingv1beta2.VerticalPodAutoscaler{ObjectMeta: metav1.ObjectMeta{Name: "aggregate-prometheus-vpa", Namespace: v1beta1constants.GardenNamespace}},
+		}
+
+		for _, resource := range resources {
+			if err := k8sSeedClient.Client().Delete(ctx, resource); client.IgnoreNotFound(err) != nil {
+				return err
+			}
+		}
```
rfranzke: You could use `kutil.DeleteObjects(...)`

rfranzke (dismissed):
/lgtm

pkg/operation/seed/seed.go
```diff
@@ -350,6 +350,16 @@ func RunReconcileSeedFlow(
...
 			return err
 		}
+	} else {
+		// Clean up stale vpa objects
+		resources := []client.Object{
+			&autoscalingv1beta2.VerticalPodAutoscaler{ObjectMeta: metav1.ObjectMeta{Name: "prometheus-vpa", Namespace: v1beta1constants.GardenNamespace}},
+			&autoscalingv1beta2.VerticalPodAutoscaler{ObjectMeta: metav1.ObjectMeta{Name: "aggregate-prometheus-vpa", Namespace: v1beta1constants.GardenNamespace}},
+		}
+
+		if err := kutil.DeleteObjects(ctx, k8sSeedClient.Client(), resources...); client.IgnoreNotFound(err) != nil {
```
ialidzhikov: There is no need for `client.IgnoreNotFound(err) != nil` as this check is already done in `DeleteObject` func. See https://github.com/gardener/gardener/blob/779eb21a6118f81d1e0cb2fd92463158cdc643f6/pkg/utils/kubernetes/object.go#L47

```suggestion
		if err := kutil.DeleteObjects(ctx, k8sSeedClient.Client(), resources...); err != nil {
```

ialidzhikov (commented):
/lgtm
/needs second-opinion

Let's hold this one until v1.21.0 is released
/hold

Kristian-ZH:
I am not sure if this is still relevant, but when we added HVPA for Loki, we should also set dynamically the resources recommended by the VPA into the Sts's chart: 

https://github.com/gardener/gardener/blob/master/pkg/operation/seed/seed.go#L461 // for garden Loki
https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/logging.go#L66 // for shoot Lokis

Without this change, the Gardenlet did not know about the updates and was reverting them.

rfranzke:
Good point @Kristian-ZH, thank you!
/author-action

wyb1:
Thanks for the feedback! PTAL
Note: the templating doesn't like it when there is a `-` in the Values so I used a workaround with `Index`
```
{{- index .Values.prometheus.resources "prometheus-config-reloader" | toYaml | nindent 10 }}
```

rfranzke (approved):
/lgtm
/invite @Kristian-ZH

Kristian-ZH (approved):
/lgtm

PR #3904, opened 2021-04-20
Add envtest suite for Seed

danielfoehrKn:

/area quality
/kind enhancement

Add an envtest setup for the Seed cluster, similar to the existing Garden cluster envtest suite.

Adds CRD's typically deployed during the SeedBootstrap
 - Gardener Extension CRDs (Worker, Infrastructure, ...)
 - Istio CRDs
 - VPA CRDs
 - HVPA CRD

Fixes #

Includes copied of the HVPA & VPA CRDs in the `resources` sub-directory as they contain helm-specific templating instructions in the `chart` folder.
Alternatively, these resources could also be rendered to a `temp` directory and then applied.

I wanted to use this functionality with PR #3893 to test the federated seed controller.

**Release note**:

```other operator

```

### DIFF
--- a/pkg/envtest/seed/environment.go
+++ b/pkg/envtest/seed/environment.go
@@ -1,0 +1,72 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package envtestseed
+
+import (
+	"path/filepath"
+
+	"github.com/gardener/gardener/charts"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var log = logf.Log.WithName("seed.test-env")
+
+// SeedTestEnvironment wraps envtest.Environment and registers CRDs typically deployed during
+// Seed bootstrap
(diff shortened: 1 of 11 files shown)


rfranzke:
Hm, why do we need this dedicated testenv for the seed? It's not quite similar to the garden - here, we need to run the `gardener-apiserver` in order to be able to create Gardener-managed resources.

For your use-case the standard testenv provided by ctrl-runtime can be used, or? It seems that you only want some helper functions that properly prepare the environment by installing the CRDs, etc.?

> Alternatively, these resources could also be rendered to a temp directory and then applied.

+1, let's try to not duplicate them if not really necessary. Also, you shouldn't need to render them to a temporary directory - couldn't you directly apply them?

danielfoehrKn:
> Hm, why do we need this dedicated testenv for the seed?

For reuse in tests that need a Seed cluster e.g for the federated seed controllers.

> For your use-case the standard testenv provided by ctrl-runtime can be used, or? It seems that you only want some helper functions that properly prepare the environment by installing the CRDs, etc.?

Yes, it is only a wrapper around ctrl-runtime envtest with installed CRDs. 

>  It seems that you only want some helper functions that properly prepare the environment by installing the CRDs, etc.?

Is this not a valid reason?

rfranzke:
OK, I was just wondering about the general approach to wrap the envtest entirely. Maybe it's simpler/good enough/less code to just define some helper functions in the `pkg/utils/test` package? I don't have a strong opinion, just wanted to chat about it, cc @timebertt who is much more into this.

timebertt:
/assign
Will take a look

pkg/envtest/seed/environment.go
```diff
@@ -0,0 +1,72 @@
...
+	pathIstio := filepath.Join(e.chartPath, "istio", "istio-crds", "templates", "crd-all.gen.yaml")
+
+	e.Environment.CRDDirectoryPaths = []string{
+		pathExtensionCRDs,
+		pathVPA,
+		pathVPACheckpoints,
+		pathHVPA,
+		pathIstio,
+	}
+	e.Environment.ErrorIfCRDPathMissing = true
```
timebertt, line 58: I totally agree with @rfranzke. I don't think the wrapper brings any real value in comparison to some helper funcs.
The only "new" code in this wrapper is the few lines which determine the paths to the relevant CRDs.

Please extract this part into a helper func which can be called by test suites that want to test against some resources on a seed. The func can render CRDs if necessary (and not done upfront already) and return some paths or objects to apply.

timebertt (changes requested):

PR #3906, opened 2021-04-21
Only validate shoot resources when changed

timuthy:

/area quality
/kind regression

This PR fixes an issue when shoots have `shoot.spec.resources` specified. Earlier, Gardener checked their access permission every time the shoot spec was changed. This unfortunately also involves updates from components like the Gardener-Resource-Manager or Gardener-Scheduler which never have access to resources defined in `shoot.spec.resources`, so the spec change was denied by the admission plugin.

/cc @rfranzke

I might plan to come up with a more elegant solution as soon as time allows.

**Release note**:

```other operator
A bug has been fixed which prevented shoot from being scheduled, hibernated or maintained when they hold references to external resources via `shoot.spec.resources`.
```

### DIFF
--- a/plugin/pkg/global/resourcereferencemanager/admission.go
+++ b/plugin/pkg/global/resourcereferencemanager/admission.go
@@ -274,6 +274,8 @@ func (r *ReferenceManager) Admit(ctx context.Context, a admission.Attributes, o 
 			}
 			annotations[v1beta1constants.GardenCreatedBy] = a.GetUserInfo().GetName()
 			shoot.Annotations = annotations
+
+			oldShoot = &core.Shoot{}
 		case admission.Update:
 			// skip verification if spec wasn't changed
 			// this way we make sure, that users can always annotate/label the shoot if the spec doesn't change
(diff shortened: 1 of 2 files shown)


ialidzhikov (dismissed):
/lgtm

plugin/pkg/global/resourcereferencemanager/admission.go
```diff
@@ -529,64 +535,75 @@ func (r *ReferenceManager) ensureSeedReferences(ctx context.Context, seed *core.
 	return r.lookupSecret(ctx, seed.Spec.SecretRef.Namespace, seed.Spec.SecretRef.Name)
 }
 
-func (r *ReferenceManager) ensureShootReferences(ctx context.Context, attributes admission.Attributes, shoot *core.Shoot) error {
-	if _, err := r.cloudProfileLister.Get(shoot.Spec.CloudProfileName); err != nil {
-		return err
-	}
+func (r *ReferenceManager) ensureShootReferences(ctx context.Context, attributes admission.Attributes, oldShoot, shoot *core.Shoot) error {
+	shootCreation := oldShoot == nil
```
rfranzke: In the ShootValidator for example, we initialize the `oldShoot` with an empty `&gardencorev1beta1.Shoot{}` structure so that we don't have to have such `shootCreation` variable. WDYT about doing the same here?

Also, what about the other resources in this admission plugin? I haven't checked them, but I assume we should behave similarly for them?

timuthy: As discussed, I addressed your feedback about the empty shoot assignment and will adjust checks for other resources in another PR in the future due to time restrictions.

rfranzke (approved):
/lgtm

PR #3911, opened 2021-04-21
Add scale subresource for ManagedSeedSets

### LINKED ISSUE #3907: Implement scale subresource for ManagedSeedSets

/area ops-productivity
/kind enhancement

We could implement the [scale subresource](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/autoscaling/horizontal-pod-autoscaler.md#scale-subresource) on `ManagedSeedSet`s similar to `Deployment`/`StatefulSet`/etc.
We can check https://github.com/kubernetes/kubernetes/blob/v1.21.0/pkg/registry/apps/deployment/storage/storage.go for reference.

**Why is this needed**:

This will allow operators to scale ManagedSeedSets via a simple `kubectl` command:
```bash
kubectl scale mss my-seeds --replicas 3
```

stoyanr:

/area ops-productivity
/kind enhancement

Adds a scale subresource for ManagedSeedSets in order to allow operators to scale them via a simple `kubectl` command, e.g. `kubectl scale mss/my-seeds --replicas 3`

Fixes #3907

**Release note**:

```other operator
ManagedSeedSets can now be scaled via the `scale` command, e.g. `kubectl scale mss/my-seeds --replicas 3`
```

### DIFF
--- a/pkg/api/register.go
+++ b/pkg/api/register.go
@@ -19,10 +19,12 @@ import (
 	seedmanagementinstall "github.com/gardener/gardener/pkg/apis/seedmanagement/install"
 	settingsinstall "github.com/gardener/gardener/pkg/apis/settings/install"
 
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 )
 
 var (
@@ -36,6 +38,8 @@ func init() {
 	coreinstall.Install(Scheme)
 	seedmanagementinstall.Install(Scheme)
 	settingsinstall.Install(Scheme)
+
+	utilruntime.Must(autoscalingv1.AddToScheme(Scheme))
 
 	metav1.AddToGroupVersion(Scheme, schema.GroupVersion{Version: "v1"})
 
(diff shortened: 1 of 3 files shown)


timebertt:
Wow, that was fast 🚀 😄 
/assign

BeckerMax:
Did you consider adding unit tests for some of the methods?
One option would be to use our new env-tests for testing this functionality.

Rest of the PR looks fine.

Paired review with @timebertt

stoyanr:
> Did you consider adding unit tests for some of the methods?
> One option would be to use our new env-tests for testing this functionality.

I am a big fan of unit tests and in general put a lot of effort to add them properly. But in this case, I think it's excessive:

* The logic is rather simple, independent of other pieces, and unlikely to change much.
* The `scale` field is impossible to mock since `genericregistry.Store` is a struct, not an interface. Of course I could introduce my own interface for the struct methods used and use it instead of the struct, then mock it in order to write unit tests. Or I could write a "fake" store (they have something similar in k/k), although this is even more work. This doesn't pay off well if it's only going to be used for this single case.

There are right now 0 (zero) unit tests for `storage.go` files in the entire g/g codebase, probably because of the reasons above. I am rather reluctant to write the first one (and likely the only one for the foreseeble future, knowing that such were never written / requested so far).

For similar reasons, I don't think `envtest` is a good option either. They are great for testing controllers end-to-end, but have their own complexity and other drawbacks. This functionality here is not even a controller, although it may be possible to use envtest to test it it's hardly worth the trouble.

timebertt (approved):
Ok, thanks for reasoning about it.
We were mostly thinking about funcs like `scaleFromManagedSeedSet` that could be unit tested or the entire thing e2e.
But you're probably right, the code probably won't change soon / at all, so it might not be worth the extra effort.

/lgtm

rfranzke:
I guess we can then go ahead and merge this PR @timebertt @BeckerMax? :)

BeckerMax (approved):

PR #3925, opened 2021-04-22
Introduce new error code for cloud provider rate limit exceeded errors

### LINKED ISSUE #3775: Rate limit exceeded errors are wrongly detected as non-retriable

ialidzhikov:
/kind bug
/keep-commits

Rework of https://github.com/gardener/gardener/pull/3782

Fixes #3775

**Release note**:

```noteworthy user
Shoot operations that error due to cloud provider rate limit exceeded errors are now classified with the new `ERR_INFRA_REQUEST_THROTTLING` error code. Previously these errors were classified as `ERR_INFRA_QUOTA_EXCEEDED` and they were no longer retried. There is now a new control loop in GCM that is responsible for retrying such failed Shoots due to rate limit exceeded errors.
```


pkg/controllermanager/controller/shoot/shoot_retry_control.go
```diff
@@ -0,0 +1,140 @@
...
+	shootLogger.Info("[SHOOT RETRY] Shoot was successfully retried")
+
+	return reconcile.Result{}, nil
+}
+
+func shootFailedDueToRateLimits(shoot *gardencorev1beta1.Shoot) bool {
+	return isShootFailed(shoot) && gardencorev1beta1helper.HasErrorCode(shoot.Status.LastErrors, gardencorev1beta1.ErrorInfraRateLimitsExceeded)
+}
+
+func isShootFailed(shoot *gardencorev1beta1.Shoot) bool {
```
rfranzke, line 117: Could this be replaced with https://github.com/gardener/gardener/blob/a4a55c852b1486a4aaa3372771ab7b365737d3a3/pkg/utils/gardener/shoot.go#L52-L57?

pkg/controllermanager/controller/shoot/shoot.go
```diff
@@ -143,6 +147,12 @@ func NewShootController(
 		UpdateFunc: shootController.shootReferenceUpdate,
 	})
 
+	shootInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    shootController.shootRetryAdd,
+		UpdateFunc: shootController.shootRetryUpdate,
+		// TODO: check that we handle also the deletion case
```
rfranzke: WDYM - what needs to be handled in this case?

rfranzke (commented):
Nice PR @ialidzhikov, thank you! Do you want to take the opportunity to write a small integration test using the new Gardener envtest package for this controller?

pkg/controllermanager/controller/shoot/shoot.go
ialidzhikov: Nice catch, it was a leftover. I removed it now.

pkg/controllermanager/controller/shoot/shoot_retry_control.go
ialidzhikov: The reason I introduced a new func is that the existing one check also the `.status.gardener.version` field with `shoot.Status.Gardener.Version == version.Get().GitVersion`. This version check is valid when it is performed by gardenlet. On the other side GCM can run another version (custom build) and I think other components (except gardenlet) should not check the `.status.gardener.version` when checking if a Shoot is failed or not (currently the extension library also uses a func that does not check the `.status.gardener.version`). That's why I didn't reuse the existing `IsShootFailed` func.

ialidzhikov:
> Do you want to take the opportunity to write a small integration test using the new Gardener envtest package for this controller?

I will have a look 👀

ialidzhikov:
I want to add one small thing - I already shared with you that we have cloud provider that has too aggressive rate limits on invalid credentials and returns `Too many requests [...]` error. So right now with this new controller in GCM such Shoots with invalid credentials will be retried always (at the given retryPeriod, defaults to 10m). @timuthy @rfranzke , I want to let you know about this corner case one more time.

rfranzke (approved):
/lgtm

Nice test!

timuthy (approved):
/lgtm

timebertt:
@ialidzhikov Very nice test indeed, love it! 👏 
We should definitely write more tests of this kind :)

PR #3926, opened 2021-04-22
Ensure grm is present during hibernation

timebertt:

/area robustness ops-productivity
/kind enhancement

This PR ensures, that grm is present during hibernation (i.e. not scaled down prematurely).
It will be scaled down as part of the `HibernateControlPlane` step.

When new extension version are deployed and a shoot is hibernated before being fully reconciled with the new extension version, it might happen that changes to MRs are not reconciled, as grm is scaled down too early in the hibernation flow, which might in turn block the hibernation flow.

For example, we observed, that after https://github.com/gardener/machine-controller-manager/pull/591 some clusters couldn't be hibernated anymore, because the needed changes in RBAC rules were not reconciled by grm yet.

**Release note**:

```other operator
Ensure `gardener-resource-manager` is present during hibernation.
```

### DIFF
--- a/pkg/operation/botanist/resource_manager.go
+++ b/pkg/operation/botanist/resource_manager.go
@@ -58,11 +58,20 @@ func (b *Botanist) DefaultResourceManager() (resourcemanager.ResourceManager, er
 		RetryPeriod:   utils.DurationPtr(time.Second * 10),
 	}
 
+	// ensure grm is present during hibernation (if the cluster is not hibernated yet) to reconcile any changes to
+	// MRs (e.g. caused by extension upgrades) that are necessary for completing the hibernation flow.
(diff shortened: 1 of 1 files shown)


rfranzke (commented):
Did you look into whether the `b.Shoot.GetReplicas` function needs to be changed in general? Or is the function fine and we just have a special case for GRM here?

timebertt:
> Did you look into whether the `b.Shoot.GetReplicas` function needs to be changed in general? Or is the function fine and we just have a special case for GRM here?

Yes, I checked this. GRM is indeed a special case here, because we want to have it present during the hibernation flow and it will be scaled down later on in `HibernateControlPlane`, whereas in other components we use the `GetReplicas` func to actually scale them down in the flow (in the same func that also deploys them during normal reconciliation).

rfranzke (approved):
OK, thanks for the explanation @timebertt!

/lgtm

PR #3933, opened 2021-04-23
Fixed vpn-seed-server envoy filter configuration not being deleted when feature is disabled.

ScheererJ:

/area networking
/kind bug

In case the reverse vpn is not enabled the envoy filter is still applied causing the istio configuration to become inconsistent. This change ensures that the envoy filter for the reverse vpn is only created in case it is enabled.

Fixes the istio configuration in case reverse vpn is not enabled.

**Release note**:

```other operator
NONE
```


rfranzke (approved):
/lgtm

PR #3936, opened 2021-04-23
Add default shoots capacity for ManagedSeeds

stoyanr:

/area ops-productivity
/kind enhancement

Adds a default shoot capacity of 250 for ManagedSeeds. This is the same default added when using the `use-as-seed` annotation. It is needed in order to have an equivalent behavior to the annotation if a ManagedSeed with no resources specified is created.

Fixes #

This is needed in order to properly get rid of the `use-as-seed` annotation in our landscape setup without adding an arbitrary default in the landscape setup scripts.

**Release note**:

```other operator
NONE
```


rfranzke (approved):
/lgtm

timebertt (approved):
/lgtm

PR #3938, opened 2021-04-23
Ensure that reversed vpn and konnectivity tunnel are not set at the same time.

ScheererJ:

/area networking
/kind bug

Ensure that at most one of reversed vpn and konnectivity tunnel is enabled per shoot.

Fixes rare conditions were reverse vpn and konnectivity tunnel could be enabled at the same time leading to nil dereferences crashing the gardenlet.

**Release note**:

```other operator
NONE
```


rfranzke (approved):
/lgtm

PR #3939, opened 2021-04-23
Remove watch on secrets from ManagedSeed controller

stoyanr:

/area control-plane
/kind bug

Removes watching of secrets in the garden cluster from the ManagedSeed controller. During deletion, the controller will requeue with `WaitSyncPeriod` when waiting for the seed secrets to be deleted instead of relying on events, similarly to the way it does it for resources that are not in the garden cluster.

Fixes #

In the future, `gardenlet `may no longer be allowed to list/watch secrets - it will only be allowed to get secrets that somehow belong to its seed. Therefore, the watching of secrets by ManagedSeed controller in the garden cluster should be removed.

**Release note**:

```bugfix operator
The ManagedSeed controller no longer watches secrets in the garden cluster.
```

### DIFF
--- a/pkg/gardenlet/controller/managedseed/managedseed.go
+++ b/pkg/gardenlet/controller/managedseed/managedseed.go
@@ -35,7 +35,6 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
-	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


rfranzke (approved):
Thank you for the fast adaptation!
/lgtm
/second-opinion
/invite @timebertt

timebertt (approved):
/lgtm

PR #94, opened 2018-02-23
Make `.spec.dns.domain` required, set default

mvladev:
- defaulting `.spec.dns.domain` to cluster.local only if `.spec.dns.provider==umanaged`.
- validation: require `.spec.dns.domain` to be set in any case.
- validation: require `.spec.dns.domain` to be a valid domain name.

PR #3943, opened 2021-04-23
Add FluentBitIdleInputPlugins alert

vlvasilev:

/area logging
/kind enhancement

Recently we found that sometimes flued-bit pods stuck. They stops to prosses logs.
It is difficult to catch when this happens because the pods are still using CPU and Memory and looks like other working fluent-bit pods. The only difference is in the metric `fluentbit_input_bytes_total` which doesn't increase.
The new alert `FluentBitIdleInputPlugins` warn us when this is happening.
 
This new alert will help debugging  [#94](https://github.com/gardener/logging/issues/94)

**Release note**:

```other operator
New alert `FluentBitIdleInputPlugins` for idle fluent-bit pods.
```

### DIFF
--- a/charts/seed-bootstrap/aggregate-prometheus-rules/fluent-bit.rules.yaml
+++ b/charts/seed-bootstrap/aggregate-prometheus-rules/fluent-bit.rules.yaml
@@ -19,7 +19,25 @@ groups:
         description: >
           There are no fluent-bit pods running on seed: {{$externalLabels.seed}}.
           No logs will be collected.
-  
+
+    - alert: FluentBitIdleInputPlugins
+      expr: |
+        sum by (pod) (
+          increase(
+            fluentbit_input_bytes_total{pod=~"fluent-bit.*"}[4m]
+          )
+        ) == 0
+      for: 6h
+      labels:
+        severity: warning
+        service: logging
+        type: seed
+        visibility: operator
+      annotations:
+        summary: Fluent-bit input plugins haven't process any data for the past 6 hours
+        description: >
+          The input plugins of Fluent-bit pod {{$labels.pod}} running on seed {{$externalLabels.seed}} haven't collected any logs for the last 6 hours.
+
     - alert: FluentBitReceivesLogsWithoutMetadata
       expr: |
         sum by (pod) (
@@ -37,7 +55,7 @@ groups:
         description: >
           {{$labels.pod}} receives logs without metadata on seed:
           {{$externalLabels.seed}}. These logs will be dropped.
-  
+
     - alert: FluentBitSendsOoOLogs
       expr: |
         sum by (pod) (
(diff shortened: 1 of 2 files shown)


charts/seed-bootstrap/aggregate-prometheus-rules/fluent-bit.rules.yaml
```diff
@@ -19,7 +19,25 @@ groups:
...
-  
+
+    - alert: FluentBitIdleInputPlugins
+      expr: |
+        sum by (pod) (
+          increase(
+            fluentbit_input_bytes_total{pod=~"fluent-bit.*"}[4m]
+          )
+        ) == 0
+      for: 6h
```
vpnachev, line 30: Hm, isn't 6 hours too long?

vpnachev (commented):
/invite @wyb1 @istvanballok

charts/seed-bootstrap/aggregate-prometheus-rules/fluent-bit.rules.yaml
vlvasilev: I don't know. There is two cases when this alert can be triggered without hitting this bug.
1. Non of the monitored component is logging a single line for a period of time.
    There was a similar alert which was triggered on dev and staging when there was a seed with one or few clusters(which were in hibernated state) and there was nothing to be logged on one of its node.
    The systemd input plugin should produce some logs after all but the question is for how long we have to wait.
 2. The node is under  high pressure and the input buffer is full. When the output plugging is processing the logs too slow the inputs are blocked until there is some free space in the buffer. In this case we have either a structural defect in the fluent-bit flow or we face network outage or massive Loki down event.
 
 So I don't have any evidence what will be the proper time.
 I am open for proposals.

vpnachev: My point is that for 6 hours anything can happen and something else can restart stuck pod and you will not find it. Of course, the disadvantage of shorter firing times is that more false positive alerts might be generated.

Personally, I have no hard opinion.

vpnachev (approved):
/lgtm
/needs second-opinion

rfranzke (approved):
Personally, I'm fine with the alert for now (it seems there is no clear argument for or against the `6h`, so I guess we can start with it and change/adapt later if we see need).
/lgtm

However, I'm somewhat worried about

> Recently we found that sometimes flued-bit pods stuck. They stops to prosses logs.
It is difficult to catch when this happens because the pods are still using CPU and Memory and looks like other working fluent-bit pods. The only difference is in the metric fluentbit_input_bytes_total which doesn't increase.

Is there an upstream issue about it or are the fluent-bit maintainers aware of it?

vlvasilev:
> Is there an upstream issue about it or are the fluent-bit maintainers aware of it?

I have no evidence whether the cause for he issue is in the fluent-bit itself or in our plugin.
I am making a new plugin release which is using [pprof](https://golang.org/pkg/net/http/pprof/) HTTP server to give us information about goroutines, mutexes and so on.

wyb1 (approved):

PR #3954, opened 2021-04-26
Cleanup context.TODO(), migrate from DeprecatedCreateWorker to CreateWorker (3)

rfranzke:

/area dev-productivity
/kind cleanup

Similar to #3658 and #3681, this PR continues with migration a couple of our controllers from `controllerutils.DeprecatedCreateWorker` (deprecated since #1094) to `controllerutils.CreateWorker`. This helps to get rid of some remaining usages of `context.TODO()`. Also, it replace client-go informers/listers with informers from the controller-runtime.

Part of #1648 

/squash

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/controllermanager/controller/controllerregistration/controllerregistration_control.go
+++ b/pkg/controllermanager/controller/controllerregistration/controllerregistration_control.go
@@ -23,32 +23,31 @@ import (
 
 	"github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-func (c *Controller) controllerRegistrationAdd(obj interface{}) {
+func (c *Controller) controllerRegistrationAdd(ctx context.Context, obj interface{}) {
 	key, err := cache.MetaNamespaceKeyFunc(obj)
 	if err != nil {
 		return
 	}
 	c.controllerRegistrationQueue.Add(key)
 
-	seedList, err := c.seedLister.List(labels.Everything())
-	if err != nil {
+	seedList := &gardencorev1beta1.SeedList{}
+	if err := c.gardenClient.List(ctx, seedList); err != nil {
 		return
 	}
 
-	for _, seed := range seedList {
+	for _, seed := range seedList.Items {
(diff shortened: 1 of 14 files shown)


stoyanr (dismissed):
/lgtm

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -42,129 +39,92 @@ import (
...
+		if apierrors.IsNotFound(err) {
+			r.logger.Infof("Object %q is gone, stop reconciling: %v", request.Name, err)
+			return reconcile.Result{}, nil
+		}
+		r.logger.Infof("Unable to retrieve object %q from store: %v", request.Name, err)
+		return reconcile.Result{}, err
+	}
 
+	logger := logger.NewFieldLogger(r.logger, "controllerregistration-seed", seed.Name)
 	logger.Infof("[CONTROLLERINSTALLATION SEED] Reconciling %s", seed.Name)
```
ialidzhikov: Nit: you could omit the seed name when logging as the `logger` is now a field one.

```suggestion
	logger.Info("[CONTROLLERINSTALLATION SEED] Reconciling")
```

ialidzhikov (commented):
I spotted only 1 nit.

ialidzhikov (dismissed):
/lgtm

pkg/controllermanager/controller/controllerregistration/controllerregistration_control.go
```diff
@@ -23,32 +23,31 @@ import (
...
+func (c *Controller) controllerRegistrationAdd(ctx context.Context, obj interface{}) {
 	key, err := cache.MetaNamespaceKeyFunc(obj)
 	if err != nil {
 		return
 	}
 	c.controllerRegistrationQueue.Add(key)
 
-	seedList, err := c.seedLister.List(labels.Everything())
-	if err != nil {
+	seedList := &gardencorev1beta1.SeedList{}
```
timebertt: Should we use a metadata-only list here?

timebertt (commented):
One suggestion.

General note: with this PR, we are doing a lot of direct get/list calls that were previously read from the cache. Though, I assume that this is not a problem, as gcm can communicate directly in-cluster with the garden API server.
When we perform similar refactorings in the gardenlet, we might have to be a bit careful.

pkg/controllermanager/controller/controllerregistration/controllerregistration_control.go
rfranzke: For sure, nice catch. Done with 14d2747

timebertt (approved):
Thanks for addressing my feedback and thanks for the PR in general!
/lgtm

PR #3955, opened 2021-04-26
Cleanup context.TODO(), migrate from DeprecatedCreateWorker to CreateWorker (4)

rfranzke:
**How to categorize this PR?**

/area dev-productivity
/kind cleanup

**What this PR does / why we need it**:
Similar to #3658, #3681, #3954, this PR continues with migration a couple of our controllers from `controllerutils.DeprecatedCreateWorker` (deprecated since #1094) to `controllerutils.CreateWorker`. This helps to get rid of some remaining usages of `context.TODO()`. Also, it replace client-go informers/listers with informers from the controller-runtime.

**Which issue(s) this PR fixes**:
Part of #1648 

**Special notes for your reviewer**:
ℹ️ Once this PR is merged, I'll follow-up with more PRs that also cover the remaining/not-yet migrated controllers.
/squash

**Release note**:

```other operator
NONE
```

PR #3956, opened 2021-04-26
Cleanup context.TODO(), migrate from DeprecatedCreateWorker to CreateWorker (5)

rfranzke:

/area dev-productivity
/kind cleanup

Similar to #3658, #3681, #3954, #3955, this PR continues with migration a couple of our controllers from `controllerutils.DeprecatedCreateWorker` (deprecated since #1094) to `controllerutils.CreateWorker`. This helps to get rid of some remaining usages of `context.TODO()`. Also, it replace client-go informers/listers with informers from the controller-runtime.

Part of #1648 

/squash

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/controllermanager/controller/project/project_stale_control.go
+++ b/pkg/controllermanager/controller/project/project_stale_control.go
@@ -29,9 +29,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
-	kubecorev1listers "k8s.io/client-go/listers/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
@@ -41,13 +39,11 @@ func NewProjectStaleReconciler(
 	l logrus.FieldLogger,
 	config *config.ProjectControllerConfiguration,
 	gardenClient client.Client,
-	secretLister kubecorev1listers.SecretLister,
 ) reconcile.Reconciler {
 	return &projectStaleReconciler{
 		logger:       l,
 		config:       config,
 		gardenClient: gardenClient,
-		secretLister: secretLister,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


pkg/controllermanager/controller/project/project_stale_control.go
```diff
@@ -172,64 +149,73 @@ func (r *projectStaleReconciler) reconcile(ctx context.Context, project *gardenc
 	return r.gardenClient.Delete(ctx, project)
 }
 
-func (r *projectStaleReconciler) projectInUseDueToShoots(namespace string) (bool, error) {
-	shootList, err := r.shootLister.Shoots(namespace).List(labels.Everything())
-	return len(shootList) > 0, err
+func (r *projectStaleReconciler) projectInUseDueToShoots(ctx context.Context, namespace string) (bool, error) {
+	shootList := &gardencorev1beta1.ShootList{}
```
timebertt: Please use metadata-only lists with `client.Limit(1)` in the funcs that only check for existence of any respective object.
Otherwise we will stress client and server quite heavily with large list calls.

timebertt (changes requested):
One request from my side.

pkg/controllermanager/controller/project/project_stale_control.go
rfranzke: Yes, makes sense, please check it out: ac864f8

pkg/controllermanager/controller/project/project_stale_control.go
```diff
@@ -172,64 +149,63 @@ func (r *projectStaleReconciler) reconcile(ctx context.Context, project *gardenc
...
+func (r *projectStaleReconciler) projectInUseDueToBackupEntries(ctx context.Context, namespace string) (bool, error) {
+	return kutil.IsNamespaceInUse(ctx, r.gardenClient, namespace, gardencorev1beta1.SchemeGroupVersion.WithKind("BackupEntryList"))
 }
 
-func (r *projectStaleReconciler) projectInUseDueToSecrets(namespace string) (bool, error) {
-	secretList, err := r.secretLister.Secrets(namespace).List(labels.Everything())
-	if err != nil {
+func (r *projectStaleReconciler) projectInUseDueToSecrets(ctx context.Context, namespace string) (bool, error) {
+	secretList := &corev1.SecretList{}
+	if err := r.gardenClient.List(ctx, secretList, client.InNamespace(namespace)); err != nil {
```
timebertt: Before, the secret lister was restricted to secrets not controller by gardener using [this labelselector](https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/factory.go#L105).
I think we should restrict listing the secrets with `client.MatchingLabels` similarly here, as listing all secrets in a given namespace can be quite the expensive operation for both client and server.

timebertt (commented):
Thanks for adopting my suggestion.
One more ask from my side.

pkg/controllermanager/controller/project/project_stale_control.go
rfranzke: Nice finding, thanks, that wasn't too obvious. PTAL :)

timebertt (approved):
Very nice, thank you!
/lgtm

PR #3957, opened 2021-04-26
Add Seed, Shoot, ControllerRegistration handling to Seed{Authorizer,Restriction}

rfranzke:

/area security
/kind enhancement

This PR adds handling for `Shoot`s, `Seed`s, and `ControllerRegistration`s to the `SeedAuthorizer` and `SeedRestriction` handlers.

Part of #1723 

Similar to #3884
/squash

**Release note**:

```other operator
NONE
```


pkg/admissioncontroller/webhooks/admission/seedrestriction/admission.go
```diff
@@ -131,6 +134,19 @@ func (h *handler) admitBackupEntry(ctx context.Context, seedName string, request
...
+	if request.Operation != admissionv1.Create {
+		return admission.Errored(http.StatusBadRequest, fmt.Errorf("unexpected operation: %q", request.Operation))
+	}
+
+	seed := &gardencorev1beta1.Seed{}
+	if err := h.decoder.Decode(request, seed); err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	return h.admit(seedName, &seed.Name)
```
stoyanr: I think this won't work when the ManagedSeed controller tries to create a managed seed that uses `seedTemplate` rather than `gardenlet`. In this scenario, the `gardenlet` that creates the seed might not (in most case, will not) be responsible for it and therefore will not identify itself using the correct seed identity, so the check will fail. 

I assume this is intentional, correct? There is actually no way to properly cover the `seedTemplate` case since that usually involves a `gardenlet` that is responsible for multiple seeds using `seedSelector`? It might be a good idea to add a sentence somewhere in the documentation stating that this is the case, so people are not surprised if they try to test the seed authorizer / restriction with such gardenlets / managed seeds.

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
```diff
@@ -110,6 +119,18 @@ func (a *authorizer) Authorize(_ context.Context, attrs auth.Attributes) (auth.D
 			return a.authorizeRead(seedName, graph.VertexTypeProject, attrs)
 		case secretBindingResource:
 			return a.authorizeRead(seedName, graph.VertexTypeSecretBinding, attrs)
+		case seedResource:
+			return a.authorize(seedName, graph.VertexTypeSeed, attrs,
+				[]string{"update", "patch", "delete"},
```
stoyanr: I wonder if this would really correctly authorize "delete" here. To my knowledge, the only controller that deletes seeds is the ManagedSeed controller. However, it will never delete the seed it's gardenlet is responsible for (the "parent" seed), it will always try to delete "child" seeds. Even though such child seeds are always created by the "child" gardenlet, they are deleted by the "parent" instead as part of the reconciling the deletion of a ManagedSeed resource. Have you tested this?

I am not sure how to deal with this situation. Reworking gardenlet to delete its own seed seems pretty complicated, this is why it wasn't done like that in the first place. I am afraid we might need, at least temporarily, either allow "delete" of all seeds, or think about some more sophisticated "graph" logic to allow parent gardenlets to delete their children.

stoyanr (dismissed):
Looks good except for an open question regarding seed deletion, I think we would block the deletion in the ManagedSeed controller if we merge it like this, unless I am missing something.

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
stoyanr: Note: I just realized this can't be really tested, at least not with the current Nodeless `local-garden`. It requires a `remote-garden`, but with this setup `seed-authorizer` can't be properly enabled, see https://github.com/gardener/gardener/pull/3901#discussion_r617431845.

pkg/admissioncontroller/webhooks/admission/seedrestriction/admission.go
rfranzke: Again, thanks for the very valuable feedback! As discussed via phone, with b1d139e I am now properly handling the cases where a "parent gardenlet" performs some requests for `Seed`s which it is not primarily responsible for, but which resulted out of a `ManagedSeed` reconciliation.

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
rfranzke: Again, thanks for the very valuable feedback! As discussed via phone, with b1d139e I am now properly handling the cases where a "parent gardenlet" performs some requests for `Seed`s which it is not primarily responsible for, but which resulted out of a `ManagedSeed` reconciliation.

I will try to look into how to enable the `Seed{Authorizer,Restriction}` for the `remote-garden` setup when I'm done with the major implementation tasks.

rfranzke:
Sorry @stoyanr, i couldn't make it earlier, but I've worked on your feedback in b1d139e.

/invite @timuthy 
/assign @timuthy 

Tim, could you make take a look as Stoyan will be out of office next week?

timuthy:
> Tim, could you make take a look as Stoyan will be out of office next week?

Sure, will do 👍

timuthy (approved):
/lgtm

PR #3958, opened 2021-04-26
Add Event, Lease handling to Seed{Authorizer,Restriction}

rfranzke:

/area security
/kind enhancement

This PR adds handling for `Event`s, `Lease`s to the `SeedAuthorizer` and `SeedRestriction` handlers. Additionally, the gardenlet will be allowed to read the `kube-system/cluster-identity` `ConfigMap`.

Part of #1723 

Similar to #3884 and #3957
/squash

**Release note**:

```other operator
NONE
```


pkg/admissioncontroller/webhooks/auth/seed/authorizer_test.go
```diff
@@ -982,5 +993,159 @@ var _ = Describe("Seed", func() {
...
+					Name:            name,
+					Namespace:       namespace,
+					APIGroup:        corev1.SchemeGroupVersion.Group,
+					Resource:        "events",
+					ResourceRequest: true,
+					Verb:            "create",
+				}
+			})
+
+			DescribeTable("should allow without consulting the graph because verb is get, list, or watch",
```
stoyanr:
```suggestion
			DescribeTable("should allow without consulting the graph because verb is create",
```

pkg/admissioncontroller/webhooks/auth/seed/authorizer_test.go
```diff
@@ -982,5 +993,159 @@ var _ = Describe("Seed", func() {
...
+				Entry("get", "get"),
+				Entry("list", "list"),
+				Entry("watch", "watch"),
+				Entry("update", "update"),
+				Entry("patch", "patch"),
+				Entry("delete", "delete"),
+				Entry("deletecollection", "deletecollection"),
+			)
+
+			It("should have no opinion because request is for a subresource", func() {
```
stoyanr, line 1113:
```suggestion
			It("should deny because request is for a subresource", func() {
```
"no opinion" == "deny", correct? I find it confusing that the descriptions say "deny" in some places and "have no opinion" in others, perhaps the same wording could be used everywhere?

pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_lease.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (g *graph) handleLeaseCreate(lease *coordinationv1.Lease) {
+	start := time.Now()
+	defer func() {
+		metricUpdateDuration.WithLabelValues("Lease", "Create").Observe(time.Since(start).Seconds())
+	}()
+	g.lock.Lock()
+	defer g.lock.Unlock()
+
+	g.deleteVertex(VertexTypeLease, lease.Namespace, lease.Name)
```
stoyanr: Why deleting the vertex before adding it back?

stoyanr (dismissed):
/lgtm, some minor nits / clarifications

pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_lease.go
rfranzke: I've copied this part from the handling of another resource, but you're right, for `Lease`s it doesn't make too much sense.
Generally, the rationale is partly described [here](https://github.com/gardener/gardener/blob/master/docs/deployment/gardenlet_api_access.md#pitfalls): 

> When there is a relevant update to an existing resource, i.e., when a reference to another resource is changed, then the corresponding vertex (along with all associated edges) is first deleted from the graph before it gets added again with the up-to-date edges. [...]

I've removed it now, thanks for spotting!

stoyanr (dismissed):
/lgtm, thanks!

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
```diff
@@ -59,6 +59,7 @@ var (
 	cloudProfileResource           = gardencorev1beta1.Resource("cloudprofiles")
 	configMapResource              = corev1.Resource("configmaps")
 	controllerInstallationResource = gardencorev1beta1.Resource("controllerinstallations")
+	eventResource                  = corev1.Resource("events")
```
timuthy: Should we also add `eventsv1` (`k8s.io/api/events/v1/`)?

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
```diff
@@ -150,6 +164,20 @@ func (a *authorizer) authorizeEvents(seedName string, attrs auth.Attributes) (au
 	return auth.DecisionAllow, "", nil
 }
 
+func (a *authorizer) authorizeLease(seedName string, attrs auth.Attributes) (auth.Decision, string, error) {
+	if attrs.GetName() == "gardenlet-leader-election" &&
+		utils.ValueExists(attrs.GetVerb(), []string{"create", "get", "watch", "update"}) {
+
+		return auth.DecisionAllow, "", nil
+	}
```
timuthy, line 174: Do I get it correctly that the use case for this is when you don't have a virtual Garden, i.e., seed == garden? Won't this request subsequently be denied by the SeedRestriction admission?

timuthy (changes requested):
Just a question and a suggestion from my side, looks good otherwise 👍

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
rfranzke: Yeah, probably, I don't think it's in use yet, but I've added it anyways so that we don't have to care about it in the future in case the client-go internally switches to `events.k8s.io`. See 9c74d19fd

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
rfranzke: Indeed, thank you very much for catching this! Fixed with 9c74d19fd

timuthy (approved):
/lgtm

PR #3959, opened 2021-04-27
Parse the webhook URL to detect and remove port given for SAN

NotTheEvilOne:
/area control-plane
/kind bug

In a local test environment running an extension provider created a certificate with SAN composed of server name and port. Reason was the `webhook-url=localhost:8444` being used in the code as the SAN server name without any further parsing. This PR ensures that the URL given is not a server name and port or if it is the case only considers the server name part for certificate creation.

### DIFF
--- a/extensions/pkg/webhook/certificates.go
+++ b/extensions/pkg/webhook/certificates.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"strings"
 
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	"github.com/gardener/gardener/pkg/utils/secrets"
@@ -124,6 +125,13 @@ func generateNewCAAndServerCert(mode, namespace, name, url string) (*secrets.Cer
 		ipAddresses []net.IP
 	)
 
+	serverName := url
+	serverNameData := strings.SplitN(url, ":", 3)
+
+	if len(serverNameData) == 2 {
+		serverName = serverNameData[0]
+	}
+
 	switch mode {
 	case ModeURL:
 		if addr := net.ParseIP(url); addr != nil {
@@ -132,7 +140,7 @@ func generateNewCAAndServerCert(mode, namespace, name, url string) (*secrets.Cer
 			}
 		} else {
 			dnsNames = []string{
-				url,
+				serverName,
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


gardener-robot-ci-2:
Thank you @NotTheEvilOne for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

rfranzke (dismissed):
/lgtm
/cla

vpnachev:
Can we have a unit test about this change?

NotTheEvilOne:
I'll have a look at it.

extensions/pkg/webhook/certificates.go
```diff
@@ -124,6 +125,13 @@ func generateNewCAAndServerCert(mode, namespace, name, url string) (*secrets.Cer
 		ipAddresses []net.IP
 	)
 
+	serverName := url
+	serverNameData := strings.SplitN(url, ":", 3)
```
timebertt, line 129: Why not use `url.Parse` instead?

NotTheEvilOne: Short answer: Because the string in question is not an URL.

Long answer: I haven't considered the `url.Parse` method before because even that the command line argument contains "url" in the name it does not contain a scheme or slashes, only the host and port portion of it. Looking at the [documentation](https://golang.org/pkg/net/url/#URL) the scheme is clearly required for `Parse()` to interpret the string given. I've now tested `url.Parse` without success, `urlData.Host` is always an empty string. Even worse `urlData.Host` may contain the port in it again. It should be named authority therefore.

Citing the documentation for type URL:
> Host        string    // host or host:port

Testing `scheme://host:port` results in `urlData.Host` being empty, parsing `scheme://host:port/` results in `urlData.Host` containing `host:port` as documented. So the parsing algorithm of Go 1.16.3 is additionally broken in my eyes as it does not handle valid forms of the URL defined in RFC 3986.

timebertt: Ok, thanks for checking 👍

timebertt (dismissed):
/lgtm
/ok-to-test

rfranzke:
The `verify` step failed:

```
>> make generate
>> make revendor
> Check
Executing golangci-lint
Executing go vet
Executing gofmt/goimports
Unformatted files detected:
extensions/pkg/webhook/certificates_test.go
```

timebertt (approved):
/lgtm
/squash
/ok-to-test

PR #3960, opened 2021-04-28
Don't wait for terminating ControlPlane to be reconciled

timebertt:

/area ops-productivity
/kind enhancement

With this PR, the shoot deletion flow, doesn't redeploy/wait for the `ControlPlane` resource, if it is already marked for deletion.
This should help in solving a corner case issue for shoot deletion, in which the `ControlPlane` resource is already deleted and kcm is not present (either because of an unsuccessful creation or hibernated shoot) and the flow was interrupted somewhere in the middle.

Fixes #

I don't have clear steps to reproduce and I'm also a bit unsure, if this is the best way to solve the issue, but it's the best idea I had so far.
As said, it is a real corner case, so I would also be fine with not doing the change for now and reconsider something like this in the future, if it causes some more ops effort. Then we might also be able to better investigate and understand the issue.

So opening up for discussion here :)

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/gardenlet/controller/shoot/shoot_control_delete.go
+++ b/pkg/gardenlet/controller/shoot/shoot_control_delete.go
@@ -611,11 +611,14 @@ func needsControlPlaneDeployment(ctx context.Context, o *operation.Operation, ku
 	// If the `ControlPlane` resource and the kube-apiserver deployment do no longer exist then we don't want to re-deploy it.
 	// The reason for the second condition is that some providers inject a cloud-provider-config into the kube-apiserver deployment
 	// which is needed for it to run.
-	exists, err := extensionResourceStillExists(ctx, o.K8sSeedClient.APIReader(), &extensionsv1alpha1.ControlPlane{}, namespace, name)
+	exists, markedForDeletion, err := extensionResourceStillExists(ctx, o.K8sSeedClient.APIReader(), &extensionsv1alpha1.ControlPlane{}, namespace, name)
 	if err != nil {
 		return false, err
 	}
-	if !exists && !kubeAPIServerDeploymentFound {
+	// treat `ControlPlane` in deletion as if it is already gone. If it is marked for deletion, we also shouldn't wait
+	// for it to be reconciled, as it can potentially block the whole deletion flow (deletion depends on other control
+	// plane components like kcm and grm) which are scaled up later in the flow
+	if (!exists || markedForDeletion) && !kubeAPIServerDeploymentFound {
 		return false, nil
 	}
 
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm

ialidzhikov:
Hmm, I am not super convinced that this is the fix we are looking for. I see that KCM is deployed in the deletion flow only when  the KCM deployment is found. See

https://github.com/gardener/gardener/blob/bc79dee6b7a2a1e4d8deef2a8ddad75b871f456f/pkg/gardenlet/controller/shoot/shoot_control_delete.go#L308

For https://github.com/orgs/gardener/projects/6#card-58434015, when we checked similar clusters with @vpnachev , we saw that KCM deployment was missing. And our conclusion was that the Shoot was not created successfully and KCM was never deployed as part of the reconciliation flow.

So the question that I have in mind is whether we can always deploy KCM on Shoot deletion? Do we know why we only deploy it when the KCM deployment exists?

timebertt:
> So the question that I have in mind is whether we can always deploy KCM on Shoot deletion? Do we know why we only deploy it when the KCM deployment exists?

Well, that's a valid point. I think, this can also be improved.

Nevertheless, do you have any objections about the change in this PR in particular? I think it might still make sense in general.
Otherwise we can drop the PR.

/hold

ialidzhikov:
Sorry for the delay from my side.

> Nevertheless, do you have any objections about the change in this PR in particular? I think it might still make sense in general.
Otherwise we can drop the PR.

~~I now had a more detailed look and played with the PR locally. I found a case that this PR will regress. Currently the ControlPlane is also responsible for deploying cloud-specific Secrets and Configmaps that are used by kube-apiserver and other controlplane components (for example the `cloud-provider-config` ConfigMap on AWS). With this change, if a ControlPlane fails to be destroyed, in the retry we won't redeployed the ControlPlane -> the Secrets/ConfigMaps won't be redeployed -> kube-apiserver and other system components potentially will fail to start. You can easily reproduce this case by simply setting dummy finalizers to the ControlPlanes.~~


```
$ k -n shoot--foo--bar describe po kube-apiserver-5876c6d887-nhw6l

Events:
  Type     Reason       Age               From               Message
  ----     ------       ----              ----               -------
  Normal   Scheduled    13s               default-scheduler  Successfully assigned shoot--foo--bar/kube-apiserver-5876c6d887-nhw6l to cpu-worker
  Warning  FailedMount  6s (x5 over 13s)  kubelet            MountVolume.SetUp failed for volume "cloud-provider-config" : configmap "cloud-provider-config" not found
```

~~The shoot reconciliation also fails with:~~

```
  lastErrors:
  - description: 'task "Deploying Kubernetes API server" failed: retry failed with
      context deadline exceeded, last error: failed to apply manifests: 1 error occurred:
      could not apply object of kind "Deployment" "shoot--foo--bar/kube-apiserver":
      admission webhook "controlplane.aws.extensions.gardener.cloud" denied the request:
      could not get configmap ''shoot--foo--bar/cloud-provider-config'':
      ConfigMap "cloud-provider-config" not found'
    lastUpdateTime: "2021-05-10T14:33:45Z"
    taskID: Deploying Kubernetes API server
```

~~That's why I would vote to drop this PR. WDYT?~~

ialidzhikov:
Upps, I think I was wrong about https://github.com/gardener/gardener/pull/3960#issuecomment-836778896. The ControlPlane is actually being redeployed in the case that I described.

ialidzhikov (approved):
/lgtm

PR #3963, opened 2021-04-28
Use cache in .ci/verify

timebertt:

/area dev-productivity
/kind enhancement

This PR uses concourse's cache for keeping the go build/test cache and golangci-lint cache between verify runs.
It relies on the experimental `cache_paths` attribute to be setup for `.cache_concourse` in the pipeline definition (see [cc-utils docs](https://gardener.github.io/cc-utils/pipeline_step.html#pipeline-step)).
For now, the pipeline definition only specifies the cache for the PR `verify` step.

In order for the go test results to be cached, it uses `make test` instead of `make test-cov` during PR verfication, as the `-cover` flag is not "cacheable" (see [go docs](https://golang.org/pkg/cmd/go/internal/test/)):

> The rule for a match in the cache is that the run involves the same
test binary and the flags on the command line come entirely from a
restricted set of 'cacheable' test flags, defined as -cpu, -list,
-parallel, -run, -short, and -v. If a run of go test has any test
or non-test flags outside this set, the result is not cached.

Although concourse's caching mechanism is quite primitive and worker-local, it should still help in speeding up PR verification times significantly (also, unit tests without coverage enabled run a lot faster already).

Fixes #

### DIFF
--- a/.ci/verify
+++ b/.ci/verify
@@ -1,14 +1,19 @@
-#!/bin/bash
+#!/bin/bash -x
 
 set -o errexit
 set -o nounset
 set -o pipefail
+
+concourse_cache_dir="$PWD/.cache_concourse"
+use_cache=no
+[ -d "$concourse_cache_dir" ] && use_cache=yes
 
 cd "$(dirname $0)/.."
 
 git config --global user.email "gardener@sap.com"
 git config --global user.name "Gardener CI/CD"
 
+# TODO: use pre-built image instead of installing in every pipeline run
 apt-get update
 apt-get install -y unzip jq parallel
 
@@ -16,4 +21,33 @@ mkdir -p /go/src/github.com/gardener/gardener
 cp -r . /go/src/github.com/gardener/gardener
 cd /go/src/github.com/gardener/gardener
 
-make verify-extended
+if [ $use_cache = yes ] ; then
+  make install-requirements
+
+  # list of all to-be-cached directories
+  cache_dirs=(
+    "$(go env GOCACHE)"
+    "$(golangci-lint cache status | grep -i 'dir:' | awk '{print $2}')"
+    # TODO: cache tools (golangci-lint, mockgen, envtest bins, promtool, protoc, ...)
+  )
+
+  echo "> retrieving all cached dirs from concourse cache dir"
+  for cache in "${cache_dirs[@]}" ; do
+    mkdir -p "$cache"
+    cache_name=$(echo "$cache" | tr ' /' '_')
+    if [ -d "$concourse_cache_dir/$cache_name" ] ; then
+      cp -TR "$concourse_cache_dir/$cache_name" "$cache"
+    fi
+  done
+
+  # run test instead of test-cov to speed-up PR verification as `-cover` flag is not "cacheable"
+  make check-generate verify
+
+  echo "> moving all cached dirs back to concourse cache dir"
+  for cache in "${cache_dirs[@]}" ; do
+    cache_name=$(echo "$cache" | tr ' /' '_')
+    cp -TR "$cache" "$concourse_cache_dir/$cache_name"
+  done
+else
+  make verify-extended
+fi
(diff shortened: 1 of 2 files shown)


timebertt:
The first test run with cached test results was very fast, see https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-master/jobs/master-pull-request-job/builds/595#L607c07c7:375
The `verify` step (including `check-generate`) took only about 5 mins and the overall pipeline run about 10 mins 🎉

timebertt:
I cleaned up the PR and made it ready for review. Basic functionality is verified and works as expected, so we are ready to gain some first experience with this. There's still room for further improvement, but I would like to get this in for some first real-life results.
/ready

.ci/verify
```diff
@@ -1,19 +1,53 @@
...
+    fi
+  done
+
+  # run test instead of test-cov to speed-up PR verification as `-cover` flag is not "cacheable"
+  make check-generate verify
+
+  echo "> moving all cached dirs back to concourse cache dir"
+  for cache in "${cache_dirs[@]}" ; do
+    cache_name=$(echo "$cache" | tr ' /' '_')
+    cp -TR "$cache" "$concourse_cache_dir/$cache_name"
```
timebertt: Unfortunately, `mv -T` failed during my tests with the following error:
```
mv: inter-device move failed: '/tmp/build/86a4261d/.cache_concourse/_root_.cache_go-build' to '/root/.cache/go-build'; unable to remove target: Directory not empty
```

So I switched to `cp -RT` instead. Though, this might grow the cache size infinitely over time, I think.

@ccwienk should we use a different mechanism instead?
You mentioned `rsync`, but I think we might also be able to achieve the same results with a `rm "$concourse_cache_dir/$cache_name"` before doing a `mv`, right?

timebertt: Ok, changed to `rm -rf` + `mv` now.
Have to give it another test spin in concourse for verifying that it works.
/hold

.ci/verify
```diff
@@ -1,19 +1,55 @@
...
+    mkdir -p "$cache"
+    cache_name=$(echo "$cache" | tr ' /' '_' | cut -c 2-)
+    if [ -d "$concourse_cache_dir/$cache_name" ] ; then
+      rm -rf "$cache"
+      mv "$concourse_cache_dir/$cache_name" "$cache"
+    fi
+  done
+
+  # run test instead of test-cov to speed-up PR verification as `-cover` flag is not "cacheable"
+  make check-generate verify
```
rfranzke, line 45: IIUC, this will not run the `test-cov` rule even for the release job. I guess, at least for releases, we should provide coverage results. Can you change this?

.ci/verify
timebertt: Test run worked well: https://concourse.ci.gardener.cloud/teams/gardener/pipelines/gardener-master/jobs/master-pull-request-job/builds/613#L6072d300:622
So we can keep this approach :)
/unhold

.ci/verify
timebertt: Nope, currently the cache is only configured for the PR job. All other jobs will still run the uncached version (including `test-cov`)

rfranzke: OK, is this some configuration in the meta/ci ref? What about the head-update job of the master branch?

rfranzke (approved):
/lgtm

.ci/verify
timebertt: Yes, exactly. It is configured in `meta/ci` for the PR job (so not only for the master PR job, but for all PRs I guess).
The head-update job doesn't use the cache so far.

timebertt: ref https://github.com/gardener/gardener/commit/d15f35e9ef2775f715a24bd6651f42397988f5d0

rfranzke: Thanks, I guess you plan to backport this PR to prevent complexities in the branch.cfg?

timebertt: No, I haven't thought about it so far. I think we don't need extra configuration, as the cache will simply be ignored on older `release-*` branches, which shouldn't hurt.
The cache is pipeline-task-specific, IIUC, so it will stay empty for those old branches.

PR #3972, opened 2021-04-29
Add tests for the vpn tunnel

DockToFuture:
Co-authored-by: Johannes Scheerer <johannes.scheerer@sap.com>

/area networking
/kind enhancement

This PR adds two tests to the shoot test suite to test the vpn tunnel connection.
The first test tries to fetch the logs from a pod and the second one tries to copy data from one container to another via `kubectl cp`.

Fixes #

**Release note**:

```other operator
none
```

### DIFF
--- a/test/framework/resources/templates/vpntunnel-copy.yaml.tpl
+++ b/test/framework/resources/templates/vpntunnel-copy.yaml.tpl
@@ -1,0 +1,64 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Name }}
+  namespace: {{ .Namespace }}
+  labels:
+    app: {{ .AppLabel }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: {{ .AppLabel }}
+  template:
+    metadata:
+      labels:
+        app: {{ .AppLabel }}
+    spec:
+      initContainers:
+      - image: eu.gcr.io/gardener-project/3rd/alpine:3.13.5
+        name: data-generator
+        command:
+        - dd
+        - if=/dev/urandom
+        - of=/data/data
+        - bs=1M
+        - count={{ .SizeInMB }}
+        volumeMounts:
+        - name: source-data
+          mountPath: /data
+      containers:
+      - image: bitnami/kubectl:{{ .KubeVersion }}
+        name: hyperkube
+        command:
+        - sleep
+        - "3600"
+        env:
+        - name: KUBECONFIG
+          value: /secret/kubeconfig
+        volumeMounts:
+        - name: source-data
+          mountPath: /data
+        - name: kubecfg
+          mountPath: /secret
+      - image: bitnami/kubectl:{{ .KubeVersion }}
+        name: target-container
+        command:
+        - sleep
+        - "3600"
+        volumeMounts:
+        - name: target-data
+          mountPath: /data
+      securityContext:
+        fsGroup: 65532
+        runAsUser: 65532
+        runAsNonRoot: true
+      volumes:
+      - name: target-data
+        emptyDir: {}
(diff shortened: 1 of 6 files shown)


rfranzke (dismissed):
/invite @schrodit @dguendisch

test/integration/shoots/vpntunnel/vpntunnel.go
```diff
@@ -0,0 +1,212 @@
...
+
+		err = f.WaitUntilDeploymentsWithLabelsIsReady(ctx, loggerLabels, namespace, f.ShootClient)
+		framework.ExpectNoError(err)
+
+		ginkgo.By("Get kubeconfig and extract token")
+		err = framework.DownloadKubeconfig(ctx, f.SeedClient, f.ShootSeedNamespace(), "kubecfg", kubecfgPath)
+		defer os.Remove(kubecfgPath)
+		framework.ExpectNoError(err)
+		data, err := ioutil.ReadFile(kubecfgPath)
+		framework.ExpectNoError(err)
```
schrodit: I think it's easier to not write the kubeconfig to the filesystem.
```suggestion
		data, err := GetObjectFromSecret(ctx, client, namespace, name, KubeconfigSecretKeyName)
	         if err != nil {
		  return err
	         }
```

schrodit: alternativly you can also get the token from the provided shoot client:
```
f.ShootClient.RESTConfig().BearerToken
```

test/framework/resources/templates/vpntunnel-copy.yaml.tpl
```diff
@@ -0,0 +1,64 @@
...
+  selector:
+    matchLabels:
+      app: {{ .AppLabel }}
+  template:
+    metadata:
+      labels:
+        app: {{ .AppLabel }}
+    spec:
+      initContainers:
+      - image: alpine:3.10.3
```
dguendisch:
```suggestion
      - image: eu.gcr.io/gardener-project/3rd/alpine:3.13.5
```
You could use the latest alpine from our gcr (instead of dockerhub) (but we also have `3.10.3` should you need that one specifically)

dguendisch (changes requested):
only a nit, rest lgtm

test/integration/shoots/vpntunnel/vpntunnel.go
DockToFuture: Where does the ShootClient get initialized, we have an empty restConfig here. Do we need to pass a Config to `NewShootFramework`?

test/framework/resources/templates/vpntunnel-copy.yaml.tpl
DockToFuture: updated it

test/integration/shoots/vpntunnel/vpntunnel.go
DockToFuture: Then I think we don't gain anything.

dguendisch (dismissed):
/lgtm

test/integration/shoots/vpntunnel/vpntunnel.go
schrodit: The shootclient is initialized [here](https://github.com/gardener/gardener/blob/master/test/framework/shootframework.go#L227) where the default `kubernetes.NewClientFromSecret` is used so the `Config should be always set`

DockToFuture: Kubeconfig is now stored in memory instead of downloaded.
Unfortunately the RestConfig does not contain the bearer token therefore we will leave it as it is.

schrodit (dismissed):

rfranzke (approved):

PR #3973, opened 2021-04-29
Restricted reversed vpn to only kubernetes 1.18 and higher.

ScheererJ:
Unfortunately, the api aggregation in kubernetes 1.16/1.17 does not respect the egress configuration of the kube-apiserver. Therefore, kube-apiserver calls to aggregated apis will fail.
In our scenario, the metrics-server runs in the  shoot cluster data plane. Its readiness/health check require direct connectivity from the kube-apiserver to the metrics-server in kubernetes 1.16/1.17.
This is why this change disables the reversed vpn for kubernetes 1.16/1.17.

/area networking
/kind bug

Unfortunately, the api aggregation in kubernetes 1.16/1.17 does not respect the egress configuration of the kube-apiserver. Therefore, kube-apiserver calls to aggregated apis will fail.
In our scenario, the metrics-server runs in the  shoot cluster data plane. Its readiness/health check require direct connectivity from the kube-apiserver to the metrics-server in kubernetes 1.16/1.17.
This is why this change disables the reversed vpn for kubernetes 1.16/1.17.

Fixes failing integration tests with kubernetes 1.16/1.17 and reversed vpn enabled.

**Release note**:

```other operator
NONE
```


rfranzke (approved):
/lgtm

PR #3974, opened 2021-04-29
[GEP-15] Add operations.gardener.cloud API group, Bastion resource

xrstf:

/area ops-productivity
/kind enhancement

This adds a new API group, `operations.gardener.cloud`, with `v1alpha1` being its first and only version for now.

This adds the new `Bastion` type to that group, including some basic validation and storage logic. After this is merged, it should be possible to create/update/delete Bastion resources in the garden cluster, but no syncing to seed clusters is implemented yet (i.e. the gardenlet part).

Fixes #

**Release note**:

```other operator
NONE
```

(I would define one single release-note in the final PR that makes all of this actually usable to endusers, and keep all other PR's as `NONE`)


gardener-robot-ci-2:
Thank you @xrstf for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

test/framework/dump.go
```diff
@@ -190,6 +190,17 @@ func (f *GardenerFramework) dumpGardenerExtensionsInNamespace(ctx context.Contex
...
+	bastions := &v1alpha1.BastionList{}
+	err = k8sClient.Client().List(ctx, bastions, client.InNamespace(namespace))
+	result = multierror.Append(result, err)
+	if err == nil {
+		for _, entry := range bastions.Items {
+			f.dumpGardenerExtension(&entry)
+		}
+		f.Logger.Println()
+	}
+
```
xrstf, line 203: This is unrelated to this PR, but kind of snuck into it. Should I remove it?

pkg/apis/core/types_bastion.go
```diff
@@ -0,0 +1,88 @@
...
+	Items []Bastion
+}
+
+// BastionSpec is the specification of a Bastion.
+type BastionSpec struct {
+	// ShootRef defines the target shoot for a Bastion.
+	ShootRef BastionShootReference
+	// SSHPublicKey is the user's public key.
+	SSHPublicKey string
+	// Ingress controls from where the creation bastion host should be reachable.
```
petersutter:
```suggestion
	// Ingress controls from where the created bastion host should be reachable.
```

pkg/apis/core/types_bastion.go
```diff
@@ -0,0 +1,88 @@
...
+	Ingress *corev1.LoadBalancerIngress
+	// Conditions represents the latest available observations of a Bastion's current state.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +optional
+	Conditions []Condition
+	// Last time the condition transitioned from one status to another.
+	LastHeartbeatTimestamp metav1.Time
+	// Last time the condition was updated.
+	ExpirationTimestamp metav1.Time
```
petersutter:
```go
	// Last time the condition transitioned from one status to another.
[...]
	// Last time the condition was updated.
```
the comments are wrong. The fields have nothing to do with the condition

pkg/apis/core/types_bastion.go
```diff
@@ -0,0 +1,88 @@
...
+	// +optional
+	metav1.ListMeta
+	// Items is the list of Bastion.
+	Items []Bastion
+}
+
+// BastionSpec is the specification of a Bastion.
+type BastionSpec struct {
+	// ShootRef defines the target shoot for a Bastion.
+	ShootRef BastionShootReference
```
petersutter: should we reuse the corev1.ObjectReference type?

pkg/registry/core/bastion/backupbucket_suite_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
petersutter: `/bastion/backupbucket_suite_test.go`

you missed to rename the file

pkg/apis/core/types_bastion.go
xrstf: Yeah I wasn't 100% sure what exactly these types mean, so I just copied the documentation from another type. GEP-15 didn't make it clear to me yet how and when exactly these fields will be set. Thanks for reminding me. :)

pkg/apis/core/types_bastion.go
xrstf: The `namespace cannot be set` line in GEP-15 made my write this dedicated type. But there is `LocalObjectReference`, that could be perfect. WDYT?

petersutter: yes you are right, this is a better fit

pkg/apis/core/types_bastion.go
petersutter: we can also have a short call on this..

xrstf: The new comments should be better. Not referring to gardenctl made them sound a bit awkward.

pkg/admissioncontroller/webhooks/admission/seedrestriction/admission.go
```diff
@@ -89,6 +91,8 @@ func (h *handler) Handle(ctx context.Context, request admission.Request) admissi
 		return h.admitBackupBucket(seedName, request)
 	case backupEntryResource:
 		return h.admitBackupEntry(ctx, seedName, request)
+	case bastionResource:
```
rfranzke, line 107: Please extend https://github.com/gardener/gardener/blob/c7a65696fa66b7f39bd88698c7e84fcb2331fd5b/charts/gardener/controlplane/charts/application/templates/validatingwebhook-admission-controller.yaml#L99-L107 (for the other resources it'll be done as part of #3957. Similarly, please extend https://github.com/gardener/gardener/blob/c7a65696fa66b7f39bd88698c7e84fcb2331fd5b/hack/local-development/dev-setup-register-gardener#L347-L356.

pkg/admissioncontroller/webhooks/admission/seedrestriction/admission.go
```diff
@@ -131,6 +135,19 @@ func (h *handler) admitBackupEntry(ctx context.Context, seedName string, request
 	return h.admit(seedName, backupBucket.Spec.SeedName)
 }
 
+func (h *handler) admitBastion(seedName string, request admission.Request) admission.Response {
```
rfranzke, line 157: Please add a unit test similar to the tests for the other resources

pkg/admissioncontroller/webhooks/auth/seed/authorizer_test.go
```diff
@@ -808,6 +808,100 @@ var _ = Describe("Seed", func() {
...
+				attrs           *auth.AttributesRecord
+			)
+
+			BeforeEach(func() {
+				name, namespace = "foo", "bar"
+				attrs = &auth.AttributesRecord{
+					User:            seedUser,
+					Name:            name,
+					Namespace:       namespace,
+					APIGroup:        gardencorev1beta1.SchemeGroupVersion.Group,
```
rfranzke:
```suggestion
					APIGroup:        gardencorev1alpha1.SchemeGroupVersion.Group,
```

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
```diff
@@ -58,6 +58,7 @@ var (
 	// group and the resource (but it ignores the version).
 	backupBucketResource           = gardencorev1beta1.Resource("backupbuckets")
 	backupEntryResource            = gardencorev1beta1.Resource("backupentries")
+	bastionResource                = gardencorev1beta1.Resource("bastions")
```
rfranzke:
```suggestion
	bastionResource                = gardencorev1alpha1.Resource("bastions")
```

pkg/apis/core/types_bastion.go
```diff
@@ -0,0 +1,86 @@
...
+	IPBlock networkingv1.IPBlock
+}
+
+// BastionStatus holds the most recently observed status of the Bastion.
+type BastionStatus struct {
+	// SeedName is the name of the seed to which this Bastion is currently scheduled. This field is populated
+	// at the beginning of a create/reconcile operation.
+	SeedName string
+	// ProviderType is cloud provider used by the referenced Shoot.
+	ProviderType string
```
rfranzke: OK, I see that the merged GEP states that both fields should be in the `status`, but according to the discussion in https://github.com/gardener/gardener/pull/3802#discussion_r614626150 we decided to move them to `spec`. I don't know why this is not reflected in the document, cc @petersutter @timebertt.

pkg/apis/core/types_bastion.go
```diff
@@ -0,0 +1,86 @@
...
+	// Ingress holds the public IP and/or hostname of the bastion instance.
+	Ingress *corev1.LoadBalancerIngress
+	// Conditions represents the latest available observations of a Bastion's current state.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +optional
```
rfranzke: Such `// +optional` tags are not needed in the internal API

pkg/apis/core/field_constants.go
```diff
@@ -24,6 +24,10 @@ const (
 	// the Seed cluster of a core.gardener.cloud/v1beta1 BackupEntry.
 	BackupEntrySeedName = "spec.seedName"
 
+	// BastionSeedName is the field selector path for finding
+	// the Seed cluster of a core.gardener.cloud/v1alpha1 Bastion.
+	BastionSeedName = "status.seedName"
```
rfranzke:
```suggestion
	BastionSeedName = "spec.seedName"
```

See comment below

pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_bastion.go
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke, line 1: Please extend https://github.com/gardener/gardener/blob/master/docs/deployment/gardenlet_api_access.md

pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_bastion.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (g *graph) handleBastionCreateOrUpdate(bastion *gardencorev1alpha1.Bastion) {
+	start := time.Now()
+	defer func() {
+		metricUpdateDuration.WithLabelValues("Bastion", "CreateOrUpdate").Observe(time.Since(start).Seconds())
+	}()
+	g.lock.Lock()
+	defer g.lock.Unlock()
+
+	g.deleteAllOutgoingEdges(VertexTypeBastion, bastion.Namespace, bastion.Name, VertexTypeSeed)
```
rfranzke: According to https://github.com/gardener/gardener/blob/master/docs/deployment/gardenlet_api_access.md#pitfalls, I guess this line is not needed.

pkg/apis/core/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,84 @@
...
+
+// BastionStatus holds the most recently observed status of the Bastion.
+type BastionStatus struct {
+	// SeedName is the name of the seed to which this Bastion is currently scheduled. This field is populated
+	// at the beginning of a create/reconcile operation.
+	SeedName string `json:"seedName,omitempty" protobuf:"bytes,1,opt,name=seedName"`
+	// ProviderType is cloud provider used by the referenced Shoot.
+	ProviderType string `json:"providerType,omitempty" protobuf:"bytes,2,name=providerType"`
+	// Ingress holds the public IP and/or hostname of the bastion instance.
+	Ingress *corev1.LoadBalancerIngress `json:"ingress,omitempty" protobuf:"bytes,3,opt,name=ingress"`
```
rfranzke: Remove `omitempty` as those fields are not optional?

pkg/apis/core/validation/bastion.go
```diff
@@ -0,0 +1,88 @@
...
+func ValidateBastionSpec(spec *core.BastionSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if len(spec.ShootRef.Name) == 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("shootRef.name"), spec.ShootRef.Name, "shoot reference must not be empty"))
+	}
+	if len(spec.SSHPublicKey) == 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("sshPublicKey"), spec.SSHPublicKey, "sshPublicKey must not be empty"))
+	}
+	// TODO: validate SSH key thoroughly?
```
rfranzke: Hm, maybe yes, depends on how much effort this is. However, as `Bastion` is an end-user facing resource, it probably makes sense to validate this.

pkg/registry/core/bastion/bastion_suite_test.go
```diff
@@ -12,6 +12,16 @@
...
 
-package mock
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestBackupbucket(t *testing.T) {
```
rfranzke:
```suggestion
func TestBastion(t *testing.T) {
```

pkg/registry/core/bastion/bastion_suite_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: Why was the `mock/doc.go` file removed? 👀

pkg/registry/core/bastion/storage/tableconvertor.go
```diff
@@ -0,0 +1,87 @@
...
+type convertor struct {
+	headers []metav1beta1.TableColumnDefinition
+}
+
+func newTableConvertor() rest.TableConvertor {
+	return &convertor{
+		headers: []metav1beta1.TableColumnDefinition{
+			{Name: "Name", Type: "string", Format: "name", Description: swaggerMetadataDescriptions["name"]},
+			{Name: "IP", Type: "string", Format: "name", Description: "The bastion host's IP and/or hostname."},
+			{Name: "Age", Type: "date", Description: swaggerMetadataDescriptions["creationTimestamp"]},
```
rfranzke: What about also having columns for the shoot name, maybe the seed name, and the expiration timestamp/last heartbeat?

test/framework/dump.go
rfranzke: Hm, not sure, @schrodit can you advise, please?

rfranzke (changes requested):

pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_bastion.go
xrstf: Without this line, the graph unit tests didn't pass for some reason. I was also expecting that removing the vertex would remove all edges to it, but that was not the case. I can remove it to demo the error, if you like :)

pkg/apis/core/v1alpha1/types_bastion.go
xrstf: Why is the Ingress inside the status not optional? It only exists after an extension has reconciled and provided a status in the seed, which would then be copied upstream to the garden / core.

pkg/registry/core/bastion/bastion_suite_test.go
xrstf: That's an excellent question, I certainly did not do this on purpose. Let me check.

pkg/apis/core/v1alpha1/types_bastion.go
rfranzke: OK, if it's optional then please add the `// +optional` tag here

pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_bastion.go
rfranzke: Removing the vertex calls https://github.com/gardener/gardener/blob/dd4154eefb3f0d842f01b49594e9008b55c1974b/pkg/admissioncontroller/webhooks/auth/seed/graph/graph.go#L189 which states:

> RemoveNode removes the node with the given ID from the graph, as well as any edges attached to it. If the node is not in the graph it is a no-op.

So, I'm not sure why the tests didn't pass anymore, maybe you can check again and if you still don't succeed, I'll try to also take a detailed look beginning of next week.

pkg/apis/core/types_bastion.go
petersutter: yes seems I missed it. @xrstf can you change it accordingly?

pkg/registry/core/bastion/bastion_suite_test.go
xrstf: I can only explain it a momentary "surely everything in pkg/client is generated, let's just rm -rf it and regenerate it" lapse in conscience. I've restored the files now.

pkg/apis/core/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,86 @@
...
+
+// BastionIngressPolicy represents an ingress policy for SSH bastion hosts.
+type BastionIngressPolicy struct {
+	// IPBlock defines an IP block that is allowed to access the bastion.
+	IPBlock networkingv1.IPBlock `json:"ipBlock" protobuf:"bytes,1,opt,name=ipBlock"`
+}
+
+// BastionStatus holds the most recently observed status of the Bastion.
+type BastionStatus struct {
+	// Ingress holds the public IP and/or hostname of the bastion instance.
```
rfranzke:
```suggestion
	// Ingress holds the public IP and/or hostname of the bastion instance.
	// +optional
```

pkg/registry/core/bastion/bastion_suite_test.go
rfranzke: Hm, I still don't see them 👀

rfranzke (commented):
Can you also add an example YAML file to https://github.com/gardener/gardener/tree/master/example? :)

pkg/apis/core/types_bastion.go
petersutter: @xrstf could you also update https://github.com/gardener/gardener/blob/master/docs/proposals/15-manage-bastions-and-ssh-key-pair-rotation.md?

pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_bastion.go
xrstf: You were right, the line was not needed.

pkg/registry/core/bastion/storage/tableconvertor.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func newTableConvertor() rest.TableConvertor {
+	return &convertor{
+		headers: []metav1beta1.TableColumnDefinition{
+			{Name: "Name", Type: "string", Format: "name", Description: swaggerMetadataDescriptions["name"]},
+			{Name: "Shoot", Type: "string", Format: "name", Description: "The Shoot this bastion belongs to."},
+			{Name: "Seed", Type: "string", Format: "name", Description: "The Seed cluster on which the Shoot is scheduled."},
+			{Name: "IP", Type: "string", Format: "name", Description: "The bastion host's IP and/or hostname."},
+			{Name: "Heartbeat", Type: "string", Format: "name", Description: "The time and date when the last heartbeat occured."},
```
petersutter:
```suggestion
			{Name: "Heartbeat", Type: "string", Format: "name", Description: "The time and date when the last heartbeat occurred."},
```

pkg/registry/core/bastion/storage/tableconvertor.go
```diff
@@ -0,0 +1,116 @@
...
+		lastHeartbeat := "<never>"
+		if !bastion.Status.LastHeartbeatTimestamp.IsZero() {
+			lastHeartbeat = fmt.Sprintf("%s ago", duration.HumanDuration(time.Since(bastion.Status.LastHeartbeatTimestamp.Time)))
+		}
+		cells = append(cells, lastHeartbeat)
+
+		expires := "<pending>"
+		if !bastion.Status.ExpirationTimestamp.IsZero() {
+			expires = metatable.ConvertToHumanReadableDateType(*bastion.Status.ExpirationTimestamp)
+		}
```
petersutter:
```go
	LastHeartbeatTimestamp *metav1.Time
	ExpirationTimestamp *metav1.Time
```

possible nil pointer dereference?
```go
		if !bastion.Status.LastHeartbeatTimestamp.IsZero() {
 // ..
		if !bastion.Status.ExpirationTimestamp.IsZero() {
```

xrstf: `IsZero` is safe for nil values: https://github.com/kubernetes/apimachinery/blob/v0.21.0/pkg/apis/meta/v1/time.go#L60-L62

pkg/apis/core/v1alpha1/types_bastion.go
```diff
@@ -0,0 +1,89 @@
...
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +optional
+	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,2,rep,name=conditions"`
+	// Last time the condition transitioned from one status to another.
+	// +optional
+	LastHeartbeatTimestamp *metav1.Time `json:"lastHeartbeatTimestamp,omitempty" protobuf:"bytes,3,opt,name=lastHeartbeatTimestamp"`
+	// Last time the condition was updated.
+	// +optional
+	ExpirationTimestamp *metav1.Time `json:"expirationTimestamp,omitempty" protobuf:"bytes,4,opt,name=expirationTimestamp"`
```
petersutter: fix comments. See https://github.com/gardener/gardener/blob/af7709ad8055ea7f6a351ebf32dc126e16e44f97/pkg/apis/core/types_bastion.go#L75-L81

plugin/pkg/bastion/validator/admission_test.go
```diff
@@ -0,0 +1,208 @@
...
+			Expect(err).To(BeInvalidError())
+			Expect(getErrorList(err)).To(ConsistOf(
+				PointTo(MatchFields(IgnoreExtras, Fields{
+					"Type":  Equal(field.ErrorTypeRequired),
+					"Field": Equal("spec.shootRef.name"),
+				})),
+			))
+		})
+
+		It("should forbid the Bastion creation if the Shoot does not exist", func() {
```
petersutter, line 161: I'm missing a test that allows the Bastion creation

xrstf: I've added one now :)

pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_bastion.go
```diff
@@ -0,0 +1,93 @@
...
+	start := time.Now()
+	defer func() {
+		metricUpdateDuration.WithLabelValues("Bastion", "CreateOrUpdate").Observe(time.Since(start).Seconds())
+	}()
+	g.lock.Lock()
+	defer g.lock.Unlock()
+
+	g.deleteVertex(VertexTypeBastion, bastion.Namespace, bastion.Name)
+
+	bastionVertex := g.getOrCreateVertex(VertexTypeBastion, bastion.Namespace, bastion.Name)
```
rfranzke: Nit: This could be moved into the `if {}` block

pkg/admissioncontroller/webhooks/auth/seed/graph/graph_test.go
```diff
@@ -868,6 +921,16 @@ var _ = Describe("graph", func() {
...
 		}()
 		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			bastion1Copy := bastion1.DeepCopy()
+			bastion1.Spec.SSHPublicKey = "new-key"
+			fakeInformerBackupEntry.Update(bastion1Copy, bastion1)
+			lock.Lock()
+			defer lock.Unlock()
+			paths[VertexTypeBackupEntry] = append(paths[VertexTypeBastion], pathExpectation{VertexTypeBastion, bastion1.Namespace, bastion1.Name, VertexTypeSeed, "", seed1.Name, BeTrue()})
```
rfranzke:
```suggestion
			paths[VertexTypeBastion] = append(paths[VertexTypeBastion], pathExpectation{VertexTypeBastion, bastion1.Namespace, bastion1.Name, VertexTypeSeed, "", seed1.Name, BeTrue()})
```

(I guess this is the reason for the flakes you observe)

pkg/admissioncontroller/webhooks/auth/seed/graph/graph_test.go
```diff
@@ -992,6 +1055,16 @@ var _ = Describe("graph", func() {
...
 		}()
 		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			bastion1Copy := bastion1.DeepCopy()
+			bastion1.Spec.SSHPublicKey = "another-new-key"
+			fakeInformerBackupEntry.Update(bastion1Copy, bastion1)
+			lock.Lock()
+			defer lock.Unlock()
+			paths[VertexTypeBackupEntry] = append(paths[VertexTypeBastion], pathExpectation{VertexTypeBastion, bastion1.Namespace, bastion1.Name, VertexTypeSeed, "", seed1.Name, BeTrue()})
```
rfranzke:
```suggestion
			paths[VertexTypeBastion] = append(paths[VertexTypeBastion], pathExpectation{VertexTypeBastion, bastion1.Namespace, bastion1.Name, VertexTypeSeed, "", seed1.Name, BeTrue()})
```

plugin/pkg/bastion/validator/admission.go
```diff
@@ -0,0 +1,152 @@
...
+		fieldErr := field.Invalid(shootPath, shootName, "shoot is not yet assigned to a seed")
+		return apierrors.NewInvalid(gk, bastion.Name, field.ErrorList{fieldErr})
+	}
+
+	return nil
+}
+
+func (v *Bastion) getShoot(ctx context.Context, namespace, name string) (*gardencore.Shoot, error) {
+	return v.coreClient.Core().Shoots(namespace).Get(ctx, name, v1.GetOptions{})
+}
```
rfranzke: Minor Nit: This function doesn't seem to be too helpful FMPOV.

plugin/pkg/bastion/validator/admission.go
```diff
@@ -0,0 +1,152 @@
...
+
+		return apierrors.NewInternalError(fmt.Errorf("could not get shoot %s/%s: %v", bastion.Namespace, shootName, err))
+	}
+
+	// ensure shoot is already assigned to a seed
+	if shoot.Spec.SeedName == nil || len(*shoot.Spec.SeedName) == 0 {
+		fieldErr := field.Invalid(shootPath, shootName, "shoot is not yet assigned to a seed")
+		return apierrors.NewInvalid(gk, bastion.Name, field.ErrorList{fieldErr})
+	}
+
```
rfranzke, line 146: I miss the population of the `.spec.providerType` and `.spec.seedName` fields?

rfranzke (changes requested):

pkg/admissioncontroller/webhooks/auth/seed/graph/eventhandler_bastion.go
xrstf: Technically, now that we forbid creating Bastions for Shoots that have no Seed yet, we could remove the entire `if` statement? WDYT?

rfranzke: Hmm, not sure, we have the control plane migration use-case where a `Shoot`'s `.spec.seedName` can change. It's not clear to me what would happen to `Bastion`s in this case. Probably it should be simply terminated immediately instead of "moving" it as well. So yes, we could remove this statement most likely. OTOH, it doesn't harm or produce maintenance effort. The benefit of keeping it would be to not include implementation details of the API validation, for example, so it's easier to "not miss" it in case above description changes. Anyways, please do as you like - I don't have a strong opinion on this.

xrstf:
> why 12 hours? In GEP-15 the default was suggested to be 1 hour. The client will heartbeat the bastion and thus keep it alive as long as the session is open, so 1 hour should be sufficient.

That commit was just a WIP to demonstrate something in Slack, this was not the final value.

plugin/pkg/bastion/validator/admission.go
```diff
@@ -49,6 +49,7 @@ type Bastion struct {
 	*admission.Handler
 	coreClient coreclientset.Interface
 	readyFunc  admission.ReadyFunc
+	timeToLive time.Duration
```
petersutter: remove

test/framework/dump.go
schrodit: Can you give a little more context why do you want to remove it?

In general, the idea of that dump is that in case of an error in an integration test, all available information in the system is printed to the logs.
So that afterwards the failed test can be debugged.

In this case, I guess that the bastion list is mostly empty and will not show up in the logs.
I would rather keep it so that in case there is a bastion resource we are be able to debug it (or at least have the information that there was a bastion resource)

petersutter (dismissed):
/lgtm

xrstf:
@petersutter I think the test failure is unrelated.

pkg/registry/core/bastion/strategy.go
```diff
@@ -0,0 +1,189 @@
...
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/registry/generic"
+	"k8s.io/apiserver/pkg/storage"
+	"k8s.io/apiserver/pkg/storage/names"
+)
+
+const (
+	// BastionTimeToLive is the duration the ExpirationTimestamp of a bastion will be extended
+	// by on every heartbeat.
+	BastionTimeToLive = 1 * time.Hour
```
rfranzke: Shouldn't this be configurable? Maybe via a CLI flag on GAPI?

pkg/registry/core/bastion/strategy.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (bastionStrategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Object) {
+	newBastion := obj.(*core.Bastion)
+	oldBastion := old.(*core.Bastion)
+	newBastion.Status = oldBastion.Status
+
+	if mustIncreaseGeneration(oldBastion, newBastion) {
+		newBastion.Generation = oldBastion.Generation + 1
+	}
+}
```
rfranzke: What about the following part outlined in the [GEP](https://github.com/gardener/gardener/blob/master/docs/proposals/15-manage-bastions-and-ssh-key-pair-rotation.md):

> when gardener.cloud/operation: keepalive is set it will be removed by GAPI from the annotations and status.lastHeartbeatTimestamp will be set with the current timestamp. The status.expirationTimestamp will be calculated by taking the last heartbeat timestamp and adding x minutes (configurable, default 60 Minutes).

This needs to be added to the `PrepareForUpdate` function.

plugin/pkg/bastion/validator/admission.go
```diff
@@ -0,0 +1,151 @@
...
+	// ensure shoot is already assigned to a seed
+	if shoot.Spec.SeedName == nil || len(*shoot.Spec.SeedName) == 0 {
+		fieldErr := field.Invalid(shootPath, shootName, "shoot is not yet assigned to a seed")
+		return apierrors.NewInvalid(gk, bastion.Name, field.ErrorList{fieldErr})
+	}
+
+	// update bastion
+	bastion.Spec.SeedName = shoot.Spec.SeedName
+	bastion.Spec.ProviderType = &shoot.Spec.Provider.Type
+
```
rfranzke, line 150: What about the following part outlined in the [GEP](https://github.com/gardener/gardener/blob/master/docs/proposals/15-manage-bastions-and-ssh-key-pair-rotation.md):

> on creation, sets metadata.annotations["gardener.cloud/created-by"] according to the user that created the resource

This needs to be added here as well.

rfranzke (changes requested):

test/framework/dump.go
rfranzke: OK, thanks @schrodit, then let's keep it.

pkg/registry/core/bastion/strategy.go
xrstf: There was no method in place yet to inject configuration into the strategy and I wanted to avoid putting one in for this PR.

rfranzke: OK, I'm fine to postpone it to a later PR, but we should not forget about it :)

pkg/registry/core/bastion/strategy.go
```diff
@@ -36,44 +36,57 @@ import (
...
 
-	bastion.Generation = 1
-	bastion.Status = core.BastionStatus{
-		LastHeartbeatTimestamp: &now,
-		ExpirationTimestamp:    &expires,
+	bastion.Status.LastHeartbeatTimestamp = &now
+	bastion.Status.ExpirationTimestamp = &expires
+
+	if bastion.Annotations != nil {
+		delete(bastion.Annotations, v1alpha1constants.GardenerOperation)
```
petersutter: check if the operation is really `keepalive` before deleting it

pkg/registry/core/bastion/storage/tableconvertor.go
```diff
@@ -103,7 +103,13 @@ func (c *convertor) ConvertToTable(ctx context.Context, obj runtime.Object, tabl
 
 		expires := "<pending>"
 		if !bastion.Status.ExpirationTimestamp.IsZero() {
-			expires = metatable.ConvertToHumanReadableDateType(*bastion.Status.ExpirationTimestamp)
+			// cells = append(cells, *bastion.Status.ExpirationTimestamp)
```
rfranzke: Should this line be removed?

pkg/registry/core/bastion/strategy.go
```diff
@@ -58,26 +58,35 @@ func (bastionStrategy) NamespaceScoped() bool {
...
 	expires := metav1.NewTime(now.Add(s.timeToLive))
 
-	bastion.Generation = 1
-	bastion.Status = core.BastionStatus{
-		LastHeartbeatTimestamp: &now,
-		ExpirationTimestamp:    &expires,
+	bastion.Status.LastHeartbeatTimestamp = &now
+	bastion.Status.ExpirationTimestamp = &expires
+
+	if bastion.Annotations != nil {
```
rfranzke: This condition is not needed

plugin/pkg/bastion/validator/admission.go
```diff
@@ -147,5 +148,13 @@ func (v *Bastion) Admit(ctx context.Context, a admission.Attributes, o admission
...
 	bastion.Spec.ProviderType = &shoot.Spec.Provider.Type
 
+	if bastion.Annotations == nil {
+		bastion.Annotations = map[string]string{}
+	}
+
+	if userInfo := a.GetUserInfo(); userInfo != nil {
+		bastion.Annotations[v1alpha1constants.GardenerCreatedBy] = userInfo.GetName()
+	}
+
```
rfranzke: Nit: You could use `metav1.SetMetaDataAnnotation` :)

rfranzke: Though, this annotation should only be set for the first creation and should not be changeable (this could go into the `pkg/apis/core/validation` package), or?

rfranzke (changes requested):

petersutter (dismissed):

rfranzke:
Thanks, no further comments for now, please rebase. I will play with it locally and come back.

xrstf:
@rfranzke What about now? :)

xrstf:
I have sqaushed the PR up to `d79471e` and then added 2 more commits to move everything over to the new `operations.gardener.cloud` API group. `make clean generate test` succeeded, I used `gimps` to reformat/group all import statements, checked the copyright years in the license headers, I was able to `make dev-setup dev-setup-register-gardener` and then `make start-apiserver` and create a Bastion:

```
$ k -n garden-dev get bastions.operations.gardener.cloud
NAME               SHOOT          SEED       IP          HEARTBEAT   EXPIRES   AGE
my-first-bastion   crazy-botany   aws-seed   <pending>   12s ago     59m       12s
```

There are probably a few spots I missed while creating the new API group. Let me know :)

pkg/apis/operations/types.go
```diff
@@ -0,0 +1,24 @@
...
+package operations
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// Object is an operations object resource.
+type Object interface {
+	metav1.Object
+	// GetProviderType gets the type of the provider.
+	GetProviderType() string
+}
```
rfranzke: Why is this interface needed?

pkg/apis/operations/types_utils.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ErrorCode is a string alias.
+type ErrorCode string
+
+// Condition holds the information about the state of a resource.
+type Condition struct {
```
rfranzke: Why not reusing the `Condition`s type from one of our existing API groups?

pkg/apis/operations/v1alpha1/constants/types_constants.go
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: Let's keep our constants in `apis/core/v1beta1/constants`. Actually, the location is not quite clear and I plan to relocate it soon so that it doesn't make the impression that it's bound to the `core/v1beta1` package.

pkg/apis/operations/v1alpha1/helper/condition_builder.go
```diff
@@ -0,0 +1,155 @@
...
+	"fmt"
+
+	gardenoperationsv1alpha1 "github.com/gardener/gardener/pkg/apis/operations/v1alpha1"
+
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ConditionBuilder build a Condition.
+type ConditionBuilder interface {
```
rfranzke: Same like above - I guess it's not needed and we can re-use what we already have for other API groups.

pkg/apis/operations/v1alpha1/helper/helper.go
```diff
@@ -0,0 +1,133 @@
...
+
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Now determines the current metav1.Time.
+var Now = metav1.Now
+
+// InitCondition initializes a new Condition with an Unknown status.
+func InitCondition(conditionType gardenoperationsv1alpha1.ConditionType) gardenoperationsv1alpha1.Condition {
```
rfranzke: same

pkg/apis/operations/v1alpha1/types.go
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+const (
+	// GardenerSeedLeaseNamespace is the namespace in which Gardenlet will report Seeds'
+	// status using Lease resources for each Seed
+	GardenerSeedLeaseNamespace = "gardener-system-seed-lease"
```
rfranzke: use the existing constant instead of duplciating it here

rfranzke: btw, why do you need it in the first place?

pkg/apis/operations/v1alpha1/types_common.go
```diff
@@ -0,0 +1,86 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+// ErrorCode is a string alias.
+type ErrorCode string
```
rfranzke: same

pkg/apis/operations/v1alpha1/types_utils.go
```diff
@@ -0,0 +1,65 @@
...
+// limitations under the License.
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ConditionStatus is the status of a condition.
+type ConditionStatus string
```
rfranzke: same

pkg/registry/operations/rest/storage_core.go
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: Rename file to `storage_operations.go`

rfranzke (changes requested):
Please add a similar file like https://github.com/gardener/gardener/blob/master/charts/gardener/controlplane/charts/application/templates/apiservice-v1alpha1-seedmanagement-gardener-cloud.yaml to the `gardener` Helm chart

pkg/apis/operations/types.go
xrstf: I had to make many decisions as to what would probably be "boilerplate" for any new API  group in Gardener. Copying this interface, as well as all the Condition stuff, constants etc. was done because I did not want to introduce cross-group references (i.e. the operations group using core stuff). Coin toss went wrong, apparently.

rfranzke: OK, I see. Generally, I guess there's no reason why cross-group references should be avoided (we have this already in a lot of places and the majority of native Kubernetes types is similar). As for this interface - it's not needed at this point (you probably copied from the `extensions` group, I assume, though, this one is different compared to `core`/`settings`/`seedmanagement` because it's not served by GAPI but CRDs and some other conditions apply).
So, if you can reuse what we already have, then let's do this for now :)

plugin/pkg/bastion/validator/admission.go
```diff
@@ -0,0 +1,156 @@
...
+		return apierrors.NewInvalid(gk, bastion.Name, field.ErrorList{fieldErr})
+	}
+
+	// update bastion
+	bastion.Spec.SeedName = shoot.Spec.SeedName
+	bastion.Spec.ProviderType = &shoot.Spec.Provider.Type
+
+	if userInfo := a.GetUserInfo(); userInfo != nil {
+		metav1.SetMetaDataAnnotation(&bastion.ObjectMeta, v1beta1constants.GardenCreatedBy, userInfo.GetName())
+	}
```
rfranzke: This should only be done once for the initial creation - otherwise, nobody except the initial creator of the resource (including the to-be-introduced gardenlet controller) won't be able to change the resource:

```
k annotate bastion example-bastion gardener.cloud/operation=keepalive --overwrite --as=foo --as-group=system:masters
The Bastion "example-bastion" is invalid: metadata.annotations: Invalid value: "foo": field is immutable
```

rfranzke (changes requested):

plugin/pkg/bastion/validator/admission.go
xrstf: Fixed.

petersutter: good catch, however how is this code related to the annotations being immutable?

rfranzke: @petersutter This code always sets the `created-by` annotation value to the user making the request. In `pkg/apis/operations/validation`, it is checked that this annotation never changes.

rfranzke (approved):
/lgtm

plugin/pkg/bastion/validator/admission.go
petersutter: ah ok got it

petersutter (approved):
/lgtm

PR #3977, opened 2021-04-29
Switch back to the dockerhub base images

ialidzhikov:
/kind enhancement
/priority 5

As the switch to kaniko builder already happened with https://github.com/gardener/cc-utils/commit/117d7da3ce974d472862dcf7bcd86b15f2b5dad0, I think we can now switch back to the dockerhub base images and not face any rate limit issues.
The gcr copies that we now use are not multi-arch.

**Release note**:

```other operator
NONE
```


timebertt (approved):
/lgtm

PR #3975, opened 2021-04-29
Fixes for release validation v1.22

rfranzke:
**How to categorize this PR?**
Including fixes:

* Remove validating the removal of the use-as-seed annotation (see https://github.com/gardener/gardener/pull/3978)
* Adapt resource names for Gardener component's RBAC

/area open-source
/kind regression
/priority 2

**What this PR does / why we need it**:
With #3901, we have introduced dedicated `ClusterRole`s for the `gardener-apiserver` and `gardener-controller-manager` components. However, we missed that the `roleRef` is immutable in the respective `ClusterRoleBinding`s:

```
ClusterRoleBinding.rbac.authorization.k8s.io "gardener.cloud:controller-manager:admin" is invalid: roleRef: Invalid value: rbac.RoleRef{APIGroup:"rbac.authorization.k8s.io", Kind:"ClusterRole", Name:"gardener.cloud:system:gardener-controller-manager"}: cannot change roleRef
```

As we want to keep the dedicated `ClusterRole`s (to potentially further restrict the permissions in the future), let's now use the opportunity to properly adapt and streamline the resource names. This will result in new `ClusterRoleBinding` objects, i.e., they won't conflict with the existing ones in the system, however, operators have to make sure to cleanup the old roles after applying the Helm chart (see release note).

**Special notes for your reviewer**:
/invite @BeckerMax 

**Release note**:

```action operator
There are now dedicated `ClusterRole`s for the `gardener-apiserver` and `gardener-controller-manager`. As the `roleRef` fields of the binding `ClusterRoleBinding`s are immutable, also the names of the `ClusterRoleBinding`s have been changed. When you apply this version of Gardener then you have to make sure to cleanup the old resources that will be still left in your system be running the following command:
`kubectl delete clusterrolebinding/gardener.cloud:apiserver:admin clusterrolebinding/gardener.cloud:controller-manager:admin clusterrole/gardener.cloud:system:gardener-scheduler clusterrolebinding/gardener.cloud:system:gardener-scheduler --ignore-not-found --wait=false`
```

PR #3978, opened 2021-04-29
Remove validating the removal of the use-as-seed annotation

stoyanr:

/area control-plane
/kind enhancement

Removes validating the removal of the `use-as-seed` annotation from the shoot validator, in order to be able to remove the `use-as-seed` annotation as part of the migration to `ManagedSeeds`.

Fixes #

**Release note**:

```other operator
It is now possible to remove the `use-as-seed` annotation even if the seed is in use by shoots. This is part of deprecating the `use-as-seed` annotation in favor of using ManagedSeeds directly.
```


timebertt (commented):
Am I right to assume, that if you now remove the `use-as-seed` annotation without removing the `ownerReference` on the `ManagedSeed` first, the `ManagedSeed` will also be deleted?

In that case, should we rather enhance the validation to allow removing the annotation if there is a corresponding `ManagedSeed` without an `ownerReference` instead of completely removing the validation?

stoyanr:
> In that case, should we rather enhance the validation to allow removing the annotation if there is a corresponding `ManagedSeed` without an `ownerReference` instead of completely removing the validation?

I would say no, since we want to deprecate the `use-as-seed` annotation, not just keep it as a "proxy" for managed seeds as today. If we should ever have such a check, it should be on a `ManagedSeed`, not `use-as-seed`.

Right now there is no such check, so it's possible to delete such a managed seed. The deletion will fail though since it will try to delete the seed, which in turn will fail until all shoots are removed. I could perhaps add an admission plugin that forbids the deletion of a used ManagedSeed it you think it makes sense (in a different PR).

timebertt:
Good point, @stoyanr. You're right.

> I could perhaps add an admission plugin that forbids the deletion of a used ManagedSeed it you think it makes sense (in a different PR).

I actually thought, we had this in place already. Can you add it then?

timebertt (approved):
/lgtm

BeckerMax:
closing this one in favor of https://github.com/gardener/gardener/pull/3975

PR #3981, opened 2021-04-30
Landscaper control plane component - add import configuration and validation

danielfoehrKn:

/area control-plane
/kind enhancement

Adds the
- import configuration (with each control plane component having its own configuration in a dedicated file & commit)
   - the Gardener API server
   - the GCM
   - the Gardener Admission Controller
   - the Gardener Scheduler
- conversion coding
- validation coding
- README explaining the configuration values (will be expanded in the future)

of the Landscaper control-plane component.
**The goal is to create the component incrementally to expedite and ease the review process.** 

Essentially the component will apply the Gardener control plane chart (charts/gardener/controlplane).
The config file of the Landscaper control-plane component contains (with minor differences) the same configuration.
Adds type-safety and validation.
 
 In a subsequent PR,  I will add the actual application code.

Fixes #

**Release note**:

```other operator
Add API types for the landscaper control plane component.
```


landscaper/pkg/controlplane/apis/imports/types_apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+
+// APIServerDeploymentConfiguration contains certain configurations for the deployment
+// of the Gardener Extension API server
+type APIServerDeploymentConfiguration struct {
+	// CommonDeploymentConfiguration contains common deployment configurations
+	// Defaults:
+	//  Resources: Requests (CPU: 100m, memory 100Mi), Limits (CPU: 300m, memory: 256Mi)
+	CommonDeploymentConfiguration
+	// LivenessProbe allows to overwrite the default liveness probe
+	// Defaults:
```
schrodit, line 45: I do not see any defaulting here. Is the defaulting added in the external type or do you plan to add it in the helm chart?

danielfoehrKn: I will add the defaulting in the external type. 
However, I am planning not to duplicate defaulting, which is already done in the helm chart.
At least that is currently the plan, but not 100% sure if this is how it should be done.

WDYT also @rfranzke

danielfoehrKn:
@rfranzke @schrodit or anyone else wants to be the main contact person for the control-plane component story? 
Will be a sequence of PRs where it would help to have an involved reviewer.

rfranzke:
I will definitely look into it, though, this week is KubeCon and a few other large PRs are ongoing in parallel.

schrodit:
> @rfranzke @schrodit or anyone else wants to be the main contact person for the control-plane component story?
> Will be a sequence of PRs where it would help to have an involved reviewer.

I will also have a look. Specifically on the Landscaper part.

landscaper/pkg/controlplane/apis/imports/types_apiserver.go
schrodit: > However, I am planning not to duplicate defaulting

Yes that was the background of my question. I think defaulting in different places will result in confusion and could be hard to debug.

charts/gardener/controlplane/local-values.yaml
```diff
@@ -3,6 +3,7 @@
 global:
   # Gardener API server configuration values
   apiserver:
+    clusterIdentity: garden-cluster-identity
```
rfranzke: See below - for the `local-values.yaml` it probably doesn't matter too much. Actually, what is the `local-values.yaml` used for? Can it be deleted?

charts/gardener/controlplane/values.yaml
```diff
@@ -1,7 +1,10 @@
+# NOTE: When making changes to default values & validation, please adjust the API comments and validation
+# in package landscaper/pkg/controlplane/apis/imports
 global:
   # Gardener API server configuration values
   apiserver:
     enabled: true
+    clusterIdentity: garden-cluster-identity
```
rfranzke: This value was consciously not added to the `values.yaml` to force operators to provide a dedicated name. Otherwise, it can easily lead to conflicts and strange behaviour. Please remove it again.

charts/gardener/controlplane/values.yaml
```diff
@@ -405,10 +409,6 @@ global:
           bindAddress: 0.0.0.0
           port: 2719
           tls:
-            caBundle: |
-              -----BEGIN CERTIFICATE-----
-              ...
-              -----END CERTIFICATE-----
```
rfranzke, line 415: Why was this removed?

landscaper/pkg/controlplane/apis/imports/types_imports.go
```diff
@@ -0,0 +1,162 @@
...
+// Prerequisites:
+//  - existing etcd cluster that is accessible from the runtime cluster
+//  - there must be a Kubernetes API server that is properly set up for API aggregation
+//  - The kubeconfig for this API server has to be configured either in the field RuntimeCluster or
+//    in the field VirtualGarden in the import configuration
+// This component extends the given Kubernetes API server with the Gardener API server that serves
+// all of the Gardener resource groups (e.g for Shoots)
+// To setup API server aggregation, the Kubernetes API server has to be properly configured with configuration flags.
+// For more information how to configure the aggregation layer see:
+// https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer
```
rfranzke: Hm, why is this here? To me it sounds to better fit into a document and not into a comment inside the code?

landscaper/pkg/controlplane/apis/imports/types_admission_controller.go
```diff
@@ -0,0 +1,64 @@
...
+	// can be configured here
+	ComponentConfiguration runtime.Object
+}
+
+// SeedRestriction configures the SeedRestriction admission plugin
+type SeedRestriction struct {
+	// Enabled configures whether the SeedRestriction admission plugin should be enabled.
+	// Sets up the ValidatingWebhookConfiguration pointing to the webhook server in the Gardener Admission Controller
+	// serving the SeedRestriction webhook
+	// Default: true
```
rfranzke: Shouldn't it be a `*bool` then? Otherwise, how do you know when to apply the default value?

landscaper/pkg/controlplane/apis/imports/types_apiserver.go
rfranzke: Good point, how was it done for the gardenlet landscaper component?

rfranzke (commented):
What about unit tests for the validation logic?

charts/gardener/controlplane/values.yaml
danielfoehrKn: I checked and did not find a usage for the GCM for the CA bundle. 
Am I missing something?

charts/gardener/controlplane/values.yaml
danielfoehrKn: okay I see, will do.

charts/gardener/controlplane/local-values.yaml
danielfoehrKn: I am not sure. I think the idea was to provide a minimal viable values.yaml with some standard configuration for local testing. 
But I never used it or introduced it.
I would say it can be deleted.

landscaper/pkg/controlplane/apis/imports/types_imports.go
danielfoehrKn: I will also add a more detailed version to the documentation for the component itself. 
I think that is helpful information to understand the configuration options. I would like to keep it.

danielfoehrKn:
> What about unit tests for the validation logic?

I first wanted to verify the API itself to avoid double maintenance --> having to adjust both the validation and its unit tests. 
I will add it in the next PR together with the external API version & defaulting to complete the `api` package. 
Does that sound ok?

landscaper/pkg/controlplane/apis/imports/types_apiserver.go
danielfoehrKn: defaulting was not duplicated.

In general, I wanted to make the deployment as easy as possible.
If there is something reasonable to be defaulted / generated which is currently not (e.g the TLS certificates and CA Bundles) then I'll add that. 
- given that there is no existing CA bundle deployed yet

Also for hvpa, if the user does not explicitly configure it,  I want to only enable it if the hvpa CRD exists. Otherwise hvpa will not be enabled per default.

danielfoehrKn:
Picking up this PR again.

charts/gardener/controlplane/values.yaml
danielfoehrKn: I would propose to generate a random gardener identity in the controlplane component if not explicitly provided (making deployment easy).
I will remove the value from the helm chart though as suggested.

charts/gardener/controlplane/local-values.yaml
danielfoehrKn: I would delete it in a separate PR. This PR does not touch the existing loca-values.yaml

landscaper/pkg/controlplane/apis/imports/types_imports.go
danielfoehrKn: I added a dedicated README file with more detailed information and removed this comment.

charts/gardener/controlplane/values.yaml
```diff
@@ -253,8 +255,6 @@ global:
  #      truncateMaxBatchSize: 10485760
  #      truncateMaxEventSize: 102400
  #      version: audit.k8s.io/v1
- #  clusterIdentity: garden-cluster-identity
-
```
rfranzke, line 257: Why was this removed?

pkg/scheduler/apis/config/encoding/encoding.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+	return &runtime.RawExtension{
+		Raw:    raw,
+		Object: cfg,
+	}, nil
+}
+
+// EncodeschedulerConfigurationToBytes encodes the given external schedulerConfiguration version into a byte slice.
+func EncodeschedulerConfigurationToBytes(cfg *schedulerconfigv1alpha1.SchedulerConfiguration) ([]byte, error) {
```
rfranzke: use camel case

pkg/scheduler/apis/config/encoding/encoding.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+	}
+	if _, _, err := getDecoder(withDefaults).Decode(bytes, nil, cfg); err != nil {
+		return nil, err
+	}
+	return cfg, nil
+}
+
+// EncodeschedulerConfiguration encodes the given external GardenletConfiguration version into a raw extension.
+func EncodeschedulerConfiguration(cfg *schedulerconfigv1alpha1.SchedulerConfiguration) (*runtime.RawExtension, error) {
```
rfranzke: use camel case

pkg/scheduler/apis/config/helper/helper_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestHelper(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Gardener Scheduler API Helper Suite")
```
rfranzke:
```suggestion
	RunSpecs(t, "Gardener Scheduler APIs Config Helper Suite")
```

pkg/admissioncontroller/apis/config/encoding/encoding_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestEncoding(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "ControllerManager API Encoding Suite")
```
rfranzke:
```suggestion
	RunSpecs(t, "AdmissonController API Encoding Suite")
```

charts/gardener/controlplane/values.yaml
```diff
@@ -5,7 +5,7 @@ global:
   apiserver:
     enabled: true
     serviceEnabled: true
-    replicaCount: 1
+    replicaCount: 3
```
rfranzke: Why is this changed?

landscaper/pkg/gardenlet/controller/gardenlet_delete.go
```diff
@@ -49,11 +49,11 @@ func (g Landscaper) Delete(ctx context.Context) error {
...
 			return fmt.Errorf("cannot delete seed '%s' which is still used by at least one shoot", g.gardenletConfiguration.SeedConfig.Name)
 		}
+	}
 
-		err := g.deleteSeedAndWait(ctx, seed)
-		if err != nil {
-			return err
-		}
+	err = g.deleteSeedAndWait(ctx, seed)
+	if err != nil {
```
rfranzke: Why this change?/why not this:

```suggestion
	if err := g.deleteSeedAndWait(ctx, seed); err != nil {
```

landscaper/pkg/controlplane/apis/imports/types_imports.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// VirtualGarden contains configuration for the "Virtual Garden" setup option of Gardener
+type VirtualGarden struct {
+	// Enabled configures whether to setup Gardener with the "Virtual Garden" setup option of Gardener
+	// Please note that as a prerequisite, the API server pods of the "Virtual Garden" already need to be deployed to
+	// the runtime cluster (this should be done automatically by a preceding component when using the standard installation via the landscaper)
+	// and must be able to communicate with the Gardener Extension API server pod that will
+	// be deployed to the Garden namespace
+	Enabled *bool
```
rfranzke: Why this is a pointer?

landscaper/pkg/controlplane/apis/imports/v1alpha1/types_admission_controller.go
```diff
@@ -0,0 +1,64 @@
...
+	// +optional
+	*Configuration `json:"inline,omitempty"`
+}
+
+// SeedRestriction configures the SeedRestriction admission plugin
+type SeedRestriction struct {
+	// Enabled configures whether the SeedRestriction admission plugin should be enabled.
+	// Sets up the ValidatingWebhookConfiguration pointing to the webhook server in the Gardener Admission Controller
+	// serving the SeedRestriction webhook
+	// Default: true
```
rfranzke: The default should be false, I guess?

landscaper/pkg/controlplane/README.md
```diff
@@ -0,0 +1,441 @@
...
+- you can directly **configure and upgrade the API server** itself (e.g. to set up API aggregation) - **this is typically not the case on managed Kubernetes offerings**.
+- there are no performance concerns considering the `runtime` cluster's API server will serve both the Gardener API plus hosting other workload. 
+
+### 2) Virtual Garden
+
+**Prerequisites**
+- An Kubernetes API server deployment configured for API aggregation in the `runtime` cluster.
+  This component does not deploy the such API server. It must already exist in the runtime cluster
+  and setup for [API aggregation](https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer).
+  This is the case when the [virtual garden component](https://github.com/gardener/virtual-garden) of the lanscaper has already been deployed to the `runtime` cluster.
```
rfranzke:
```suggestion
  This is the case when the [virtual garden component](https://github.com/gardener/virtual-garden) of the landscaper has already been deployed to the `runtime` cluster.
```

landscaper/pkg/controlplane/README.md
```diff
@@ -0,0 +1,441 @@
...
+
+In this setup,
+- the `runtime` cluster hosts the Gardener control Plane pods
```
rfranzke: s/Plane/plane

landscaper/pkg/controlplane/README.md
```diff
@@ -0,0 +1,441 @@
...
+- a dedicated Kubernetes API server deployed in the `runtime` cluster is extendet by the Gardener Extension API server (`virtual-garden` API server).
```
rfranzke: s/extendet/extended

rfranzke (commented):
Generally it looks OK, I only have minor comments. What about unit tests for the validation functions?

landscaper/pkg/controlplane/apis/imports/v1alpha1/types_admission_controller.go
danielfoehrKn: Secure by default? :)

rfranzke: While the intention is great, there are still some manual steps that need to be performed before this can be set to `true`, see https://github.com/gardener/gardener/blob/master/docs/deployment/gardenlet_api_access.md#seedrestriction-admission-webhook-enablement.

charts/gardener/controlplane/values.yaml
danielfoehrKn: The Admission Controller gets [3 replicas per default](https://github.com/gardener/gardener/blob/ed03f939949dfd80ec7317ab72f361dfff05dfc0/charts/gardener/controlplane/values.yaml#L261), this does not match the one replica for the API server getting most of the load. I thought it is a sensible default. But I can surely remove it again.

charts/gardener/controlplane/values.yaml
danielfoehrKn: It seems like I added that in an earlier commit in the series and then removed it again based on your previous suggestion.

landscaper/pkg/controlplane/apis/imports/v1alpha1/types_admission_controller.go
danielfoehrKn: okay, so only if the Seed Authorizer is enabled, this can be true. Shall it be defaulted to true in the case the Seed Authorizer is enabled?

charts/gardener/controlplane/values.yaml
rfranzke: Yeah, let's keep it `1` for now. Probably we could make the default for GAC `1` as well to not waste too many resources in small dev/experimental landscapes.

landscaper/pkg/controlplane/apis/imports/v1alpha1/types_admission_controller.go
rfranzke: Yes, good idea, I think this makes sense!

landscaper/pkg/gardenlet/controller/gardenlet_delete.go
danielfoehrKn: Reverted the change in error. thanks for spotting.

landscaper/pkg/controlplane/apis/imports/v1alpha1/types_admission_controller.go
danielfoehrKn: Added the defaulting.

rfranzke (approved):
/lgtm

PR #3982, opened 2021-04-30
Cleanup context.TODO(), migrate from DeprecatedCreateWorker to CreateWorker (6)

rfranzke:

/area dev-productivity
/kind cleanup

Similar to #3658, #3681, #3954, #3955, #3956 this is my final PR (for now) continuing with the migration of a couple of our controllers from `controllerutils.DeprecatedCreateWorker` (deprecated since #1094) to `controllerutils.CreateWorker`. This helps to get rid of some remaining usages of `context.TODO()`. Also, it replaces client-go informers/listers with informers from the controller-runtime.

Part of #1648 

/squash

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/controllermanager/controller/event/event.go
+++ b/pkg/controllermanager/controller/event/event.go
@@ -16,6 +16,7 @@ package event
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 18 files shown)


pkg/controllermanager/controller/event/event.go
```diff
@@ -30,54 +31,70 @@ import (
...
+) (
+	*Controller,
+	error,
+) {
+	gardenClient, err := clientMap.GetClient(ctx, keys.ForGarden())
+	if err != nil {
+		return nil, err
+	}
+
+	eventInformer, err := gardenClient.Cache().GetInformer(ctx, &corev1.Event{})
```
timebertt: As you are already on it: should we switch to a metadata-only watch for events in order to not cache all the events?

pkg/controllermanager/controller/seed/seed_lifecycle_reconcile.go
```diff
@@ -54,55 +52,48 @@ func (c *Controller) seedLifecycleAdd(obj interface{}) {
...
+	gardenClient kubernetes.Interface
+	leaseLister  coordinationlister.LeaseLister
+	shootLister  gardencorelisters.ShootLister
+	config       *config.ControllerManagerConfiguration
 }
 
 func (c *livecycleReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
-	seedObj, err := c.seedLister.Get(req.Name)
+	seed := &gardencorev1beta1.Seed{}
+	err := c.gardenClient.Client().Get(ctx, req.NamespacedName, seed)
```
timebertt: Can you read from the cache here? This is executed quite frequently, I think we have to be a bit careful here.

pkg/controllermanager/controller/seed/seed_lifecycle_reconcile.go
```diff
@@ -147,14 +142,14 @@ func (c *livecycleReconciler) Reconcile(ctx context.Context, req reconcile.Reque
 
 	seedLogger.Debugf("Gardenlet didn't send a heartbeat for at least %s - setting the shoot conditions/constraints to 'unknown' for all shoots on this seed", c.config.Controllers.Seed.ShootMonitorPeriod.Duration)
 
-	shootList, err := c.shootLister.List(labels.Everything())
-	if err != nil {
+	shootList := &gardencorev1beta1.ShootList{}
+	if err := c.gardenClient.Client().List(ctx, shootList); err != nil {
```
timebertt: Uff, this one is quite heavy for server and client. Can we read from the cache here? Maybe also using a field selector for `spec.seedName`?

pkg/controllermanager/controller/seed/seed_reconcile.go
```diff
@@ -125,35 +122,36 @@ func (r *reconciler) cleanupStaleSecrets(ctx context.Context, gardenClient kuber
 	var fns []flow.TaskFn
 	exclude := sets.NewString(existingSecrets...)
 
-	secrets, err := r.secretLister.Secrets(namespace).List(gardenRoleSelector)
-	if err != nil {
+	secretList := &corev1.SecretList{}
```
timebertt, line 119: Can you replace `gardenClient kubernetes.Interface` with `gardenClient client.Client` in this file?

pkg/controllermanager/controller/factory.go
```diff
@@ -62,21 +61,19 @@ import (
...
+func NewGardenControllerFactory(clientMap clientmap.ClientMap, kubeInformerFactory kubeinformers.SharedInformerFactory, cfg *config.ControllerManagerConfiguration, recorder record.EventRecorder) *GardenControllerFactory {
 	return &GardenControllerFactory{
-		cfg:                    cfg,
-		clientMap:              clientMap,
-		k8sGardenCoreInformers: gardenCoreInformerFactory,
-		k8sInformers:           kubeInformerFactory,
-		recorder:               recorder,
+		cfg:          cfg,
+		clientMap:    clientMap,
+		k8sInformers: kubeInformerFactory,
```
timebertt: We still have the `kubeInformerFactory` because of the specialized secret informer, correct?
I noticed, that we now have two secret informers: the specialized one from the `kubeInformerFactory` and another one for all secrets created [here](https://github.com/gardener/gardener/blob/77717eeb78fc2b98334fce0836dcadfe994aad36/pkg/controllermanager/controller/seed/seed.go#L77) / [here](https://github.com/gardener/gardener/blob/852e4ad58abb3523542f0059c5258926d02a8857/pkg/controllermanager/controller/plant/plant.go#L72).
We should probably either remove one of the informers or replace the c-r informer by a metada-only informer.

pkg/controllermanager/controller/factory.go
```diff
@@ -106,14 +93,14 @@ func (f *GardenControllerFactory) Run(ctx context.Context) error {
...
-			opts.LabelSelector = uncontrolledSecretSelector.String()
+			opts.LabelSelector = gutil.UncontrolledSecretSelector.String()
 		})
 		return secretInformer
 	})
 
 	// Delete legacy (and meanwhile unused) ConfigMap after https://github.com/gardener/gardener/pull/3756.
 	// TODO: This code can be removed in a future release.
-	if err := kutil.DeleteObject(ctx, k8sGardenClient.Client(), &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "gardener-controller-manager-internal-config", Namespace: v1beta1constants.GardenNamespace}}); err != nil {
+	if err := kutil.DeleteObject(ctx, k8sGardenClient.Client(), &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "gutil-controller-manager-internal-config", Namespace: v1beta1constants.GardenNamespace}}); err != nil {
```
timebertt: Ups. I guess, you might wanna revert his line 😄

timebertt (changes requested):
Well done, thank you!
Looks very good overall, only a few suggestions from my side.

pkg/controllermanager/controller/factory.go
rfranzke: > We should probably either remove one of the informers or replace the c-r informer by a metada-only informer.

What do you suggest concretely? Shall we simply get rid of the `k8sInformers` and only use the cr-informers now?

timebertt: That would be one option. I would just go for it and try it.
I checked the amount of secrets in one of our larger gardener installation and saw, that we have about 50MiB secrets there. So from the memory usage perspective it shouldn't be an issue to just cache all secrets. Not so sure about the implications wrt CPU usage on server/gcm side, but I would think that it's also negligible. 
Once, we vendored c-r@v0.9 we'll be able to use filtered watches for those informers again (ref https://github.com/kubernetes-sigs/controller-runtime/pull/1435)

rfranzke: OK, done with ec61603b6e1d432d0ddef503c50b81e191eb58e3

timebertt (approved):
Very nice, thanks for addressing my suggestions!
/lgtm
/second-opinion

timuthy (approved):

PR #3983, opened 2021-04-30
Refactor kube-apiserver-{service,sni} Helm charts into component

rfranzke:

/kind technical-debt test
/priority 3
/topology seed
/platform all
/exp intermediate
/merge squash

This PR refactors the deployment for the exposure-related resources of the kube-apiservers of shoot clusters. It adopts the component concept. Consequently, it eliminates the Helm chart and adds unit tests, instead. This improves several aspects and the overall code quality.

- Part of #2754
- The `EnvoyFilter` is not handled as Golang structure because there's no meaningful way to do so. Also, it's way more convenient to keep it as a template that gets applied using the `kubernetes.Applier`. 

**Release note**:

```improvement operator
NONE
```


pkg/operation/botanist/component/kubeapiserverexposure/kube_apiserver_sni.go
```diff
@@ -0,0 +1,256 @@
...
+func (s *sni) Deploy(ctx context.Context) error {
+	var (
+		destinationRule = s.emptyDestinationRule()
+		envoyFilter     = s.emptyEnvoyFilter()
+		gateway         = s.emptyGateway()
+		virtualService  = s.emptyVirtualService()
+
+		hostName                  = fmt.Sprintf("%s.%s.svc.%s", v1beta1constants.DeploymentNameKubeAPIServer, s.namespace, gardencorev1beta1.DefaultDomain)
+		envoyFilterSpec           bytes.Buffer
+		envoyFilterTemplateValues = map[string]interface{}{
```
mvladev: Can you add struct for the values instead of free-style `map[string]interface`?

rfranzke: done with c128c00

rfranzke:
Does somebody still plan to review this PR? :)
/remove lifecycle/stale

pkg/operation/botanist/component/kubeapiserverexposure/kube_apiserver_service.go
```diff
@@ -0,0 +1,228 @@
...
+// from the outside.
+type serviceValues struct {
+	annotations              map[string]string
+	serviceType              corev1.ServiceType
+	enableSNI                bool
+	enableKonnectivityTunnel bool
+	gardenerManaged          bool
+}
+
+// NewService creates a new instance of DeployWaiter for a specific DNS entry.
```
ScheererJ: "for a specific DNS entry"
I do not get the correlation to the DNS entry. Of course the service will be exposed by a DNS entry, but I do not see the DNS entry here. Right?

pkg/operation/botanist/component/kubeapiserver/kube_apiserver.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kubeapiserver
+
+const (
+	// Port is the port exposed by the kube-apiserver.
+	Port = 443
```
ScheererJ, line 19: If the port is exposed anyway, why not use this variable for example also in kube_apiserver_service.go?

ScheererJ (dismissed):
Looks good overall, just two minor questions.

timuthy:
> Does somebody still plan to review this PR? :)
> /remove lifecycle/stale

Sorry for the delay @rfranzke. I'll block some time later today to have a second look.

pkg/operation/botanist/component/kubeapiserverexposure/kube_apiserver_service.go
rfranzke: Yeah, you're right, it was a copy-paste issue, thanks for catching!

pkg/operation/botanist/component/kubeapiserver/kube_apiserver.go
rfranzke: Hm, I have used it in `kube_apiserver_service.go`. Am I missing something? Can you point to the concrete places where you suggest to use it?

pkg/operation/botanist/component/kubeapiserverexposure/kube_apiserver_service.go
```diff
@@ -0,0 +1,228 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	servicePortName = "kube-apiserver"
+	servicePort     = 443
```
ScheererJ, line 38: I was referring to this which seems to be the same as `kubepapiserver.Port`. You can argue that the `port` and `targetPort` could differ, but it does not seem to be the case here. Therefore, I would simply reuse `kubeapiserver.Port` here. Anyway, this is just nitpicking. No hard feelings if you think it is better otherwise.

ScheererJ (commented):
> Hm, I have used it in kube_apiserver_service.go. Am I missing something? Can you point to the concrete places where you suggest to use it?

I was referring to this usage.

pkg/operation/botanist/component/kubeapiserver/kube_apiserver.go
ScheererJ: See the reference below (in kube_apiserver_service.go).

pkg/operation/botanist/component/kubeapiserverexposure/kube_apiserver_service.go
rfranzke: > You can argue that the port and targetPort could differ

Exactly, that was exactly the reason why I didn't do it :) But you're right, most likely their values will always be the same.
Still, let's keep it for now 👍🏻 I might clean it up when it comes to the refactoring of the `kube-apiserver` Helm chart.

ScheererJ (approved):
/lgtm

timuthy (approved):
I didn't have time to check everything in detail but overall it looks good to me 👍

PR #3992, opened 2021-05-04
Add CertificateSigningRequest handling to Seed{Authorizer,Restriction}

rfranzke:

/area security
/kind enhancement

This PR adds handling for `CertificateSigningRequest`s to the `SeedAuthorizer` and `SeedRestriction` handlers. This is needed for the case where a gardenlet wants to rotate its own client certificate. It does not apply for the initial bootstrapping phase (here, a gardenlet uses a bootstrap token which is authorized via RBAC rules ([ref](https://github.com/gardener/gardener/blob/64cbedca119a4aefbaafdcb42471f4bd66c38127/charts/gardener/controlplane/charts/application/templates/clusterrole-seed-bootstrapper.yaml) and [ref](https://github.com/gardener/gardener/blob/64cbedca119a4aefbaafdcb42471f4bd66c38127/charts/gardener/controlplane/charts/application/templates/clusterrolebinding-seed-bootstrapper.yaml)).

A gardenlet is only allowed to create a CSR for its own seed, i.e., it's won't be possible to get access to other seed clusters by compromising their identities.

Part of #1723 

/squash

**Release note**:

```other operator
NONE
```


pkg/admissioncontroller/seedidentity/identity.go
```diff
@@ -15,6 +15,7 @@
 package seedidentity
```
timuthy, line 15: Should we move this package to `/pkg/utils/gardener` or to some other central place?

pkg/gardenlet/bootstrap/certificate/certificate.go
```diff
@@ -54,6 +58,10 @@ func RequestCertificate(ctx context.Context, logger logrus.FieldLogger, client k
 	return certData, privateKeyData, csrName, nil
 }
 
+// DigestedName is an alias for bootstraputil.DigestedName.
```
timuthy, line 61: Why the dot (.) at the end?

pkg/gardenlet/bootstrap/certificate/certificate.go
```diff
@@ -107,3 +115,80 @@ func requestCertificate(ctx context.Context, logger logrus.FieldLogger, client k
 
 	return certData, reqName, nil
 }
+
+// waitForCertificate is heavily inspired from k8s.io/client-go/util/certificate/csr.WaitForCertificate. We don't call
+// this function directly because it performs LIST/WATCH requests while waiting for the certificate. However, gardenlet
+// is only allowed to GET CSR resources related to its seed.
+func waitForCertificate(ctx context.Context, client kubernetesclientset.Interface, reqName string, reqUID types.UID) ([]byte, error) {
```
timuthy, line 122: I'm also wondering if it makes sense to make this function public (public utility function) and write dedicates tests for it? I know that the logic behind the fallback to `v1beta1` is pretty much the same but this branch is untested at the moment. WDYT?

charts/gardener/controlplane/charts/application/templates/clusterrole-seed-bootstrapper.yaml
```diff
@@ -11,11 +11,3 @@ rules:
   verbs:
   - create
   - get
-  - list
-  - watch
-- apiGroups:
-  - certificates.k8s.io
-  resources:
-  - certificatesigningrequests/seedclient
```
timuthy, line 19: Do we still need "seedclient" here? https://github.com/gardener/gardener/blob/c44b1ac4f536b5d11aba3b55a2baccd1db3345da/pkg/controllermanager/controller/certificatesigningrequest/csr_autoapprove_control.go#L107

timuthy (changes requested):
Nice, looks good. Just minor questions, suggestions.

pkg/admissioncontroller/webhooks/admission/seedrestriction/admission.go
```diff
@@ -144,6 +149,29 @@ func (h *handler) admitBackupEntry(ctx context.Context, seedName string, request
...
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	x509cr, err := utils.DecodeCertificateRequest(csr.Spec.Request)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	if !gutil.IsSeedClientCert(x509cr, csr.Spec.Usages) {
+		return admission.Errored(http.StatusForbidden, fmt.Errorf("can only create CSRs for seed clusters"))
```
danielfoehrKn, line 168: For my understanding: doesn't that mean that this admission webhook will deny ALL CSR CREATE requests that are not created for Gardenlets?
Does that block usage of CSRs globally e.g manually to get an x509 for another user & group from the API server?
--> is that desired?

danielfoehrKn (commented):
Looks good, but have not tested the rotation itself yet.

Questions:
- The seed restriction forbids the Gardenlet to create a CSR for another seed name.
   How can anyone get into possession of the <ambiguous> client certificate?
   I am asking because this one can create itself certs for any seed name if I see that correctly.

In addition: Can you also add one sentence to [the documentation ](https://github.com/gardener/gardener/blob/4490c11d423b03c4e009eba8c1ef197488aacad4/docs/deployment/deploy_gardenlet_manually.md#L173) clarifying that in addition to RBAC, there is also an optional Authorization & Admission webhook available to restrict Gardenlets access (with a link to the respective docs)?

pkg/admissioncontroller/webhooks/admission/seedrestriction/admission.go
rfranzke: > doesn't that mean that this admission webhook will deny ALL CSR CREATE requests that are not created for Gardenlets?
Does that block usage of CSRs globally e.g manually to get an x509 for another user & group from the API server?

No, the `SeedRestriction` handler only handles requests from gardenlets and ignores everything else.

charts/gardener/controlplane/charts/application/templates/clusterrole-seed-bootstrapper.yaml
danielfoehrKn: Would someone be so kind to describe what that "seedclient" subresource in the CSR is for? Thanks!

At least I can see that the GCM creates a SubjectAccessReview for this subresource [here](https://github.com/gardener/gardener/blob/6fb1a16640e298eb22aed196d790ed842da7bbfb/pkg/controllermanager/controller/certificatesigningrequest/csr_autoapprove_control.go#L103)

rfranzke:
> How can anyone get into possession of the client certificate?

Well, nobody knows ;-) That's exactly the situation that we want to improve with the seed authorizer/restriction: If the gardenlet's client certificate for the garden cluster leaks then the "damage" should be restricted to the seed it is responsible for. It shall not be possible that other seeds can get affected as well.

> I am asking because this one can create itself certs for any seed name if I see that correctly.

WDYM, can you clarify, please? Of course, if there is a user in the garden cluster with access to create CSRs then he can issue itself a client cert for a seed. But this is out of scope of the entire seed authorizer/restriction topic.

> In addition: Can you also add one sentence to the documentation clarifying that in addition to RBAC, there is also an optional Authorization & Admission webhook available to restrict Gardenlets access (with a link to the respective docs)?

OK, ~will do~ done.

charts/gardener/controlplane/charts/application/templates/clusterrole-seed-bootstrapper.yaml
rfranzke: This is used to control whether GCM shall auto-approve CSRs created by gardenlets. Today, GCM has this privilege because it's cluster-admin in the whole garden (https://github.com/gardener/gardener/blob/2b8530028eee6425136887e7578bfac6c5f533a5/charts/gardener/controlplane/charts/application/templates/clusterrole-controller-manager.yaml#L12-L18).
In the future, when we further restrict these rules we might want to leave this decision up to the operators of Gardener - similar to how Kubernetes leaves it up to the operators whether KCM shall auto-approve CSRs by kubelets.

pkg/gardenlet/bootstrap/certificate/certificate.go
rfranzke: There are already some tests for `rotateCertificate` which implicitly tests this function as well. If it's only about the `v1beta1` fallback: I would not invest anything given that it'll be removed with v1.22 and is pretty straight-forward.

pkg/gardenlet/bootstrap/certificate/certificate.go
rfranzke: 😃 Well, I'm doing this for all doc strings :)

pkg/admissioncontroller/seedidentity/identity.go
rfranzke: No strong opinion. If I see it correctly this package is only used from within `pkg/admissioncontroller` today. Do you have an use-case in mind motivating the move proposal?

pkg/admissioncontroller/webhooks/admission/seedrestriction/admission.go
danielfoehrKn: ok, gotcha. Found the code responsible. Thanks

pkg/gardenlet/bootstrap/certificate/certificate.go
timuthy: Sure, I was wondering if the doc string continues in the next line 😄

danielfoehrKn:
> WDYM, can you clarify, please? Of course, if there is a user in the garden cluster with access to create CSRs then he can issue itself a client cert for a seed. But this is out of scope of the entire seed authorizer/restriction topic.

Having a client certificate in the group `gardener.cloud:system:seeds` and user `gardener.cloud:system:seed:<ambiguous>`
allows the creation of a CSR not only for the CN `gardener.cloud:system:seed:<ambiguous>` [but also for any CN](https://github.com/gardener/gardener/blob/998aeb7d45317f7b4d784fda237ddef873ec318a/pkg/admissioncontroller/webhooks/admission/seedrestriction/admission.go#L259) `gardener.cloud:system:seed:seed-1..N`.

I have a bad gut feeling about Gardenlets with effectively a wildcard to create certificates for any Seed cluster + being able to rotate that wildcard cert, too.
Do we break users by disallowing rotation and issuing of x509 certs via CSR for certs with CN `gardener.cloud:system:seed:<ambiguous>`?
This would at least limit such powerful certs to its cert lifetime so the installation is safe after that time passed.
This would impact Gardenlets reconciling multiple seeds - but this setup is only recommended for development use-cases anyways (where cert rotation does not matter).

pkg/admissioncontroller/seedidentity/identity.go
timuthy: > If I see it correctly this package is only used from within pkg/admissioncontroller today.

Makes sense, thank you 👍

rfranzke:
> Having a client certificate in the group gardener.cloud:system:seeds and user gardener.cloud:system:seed:<ambiguous>
allows the creation of a CSR not only for the CN gardener.cloud:system:seed:<ambiguous> but also for any CN gardener.cloud:system:seed:seed-1..N.

Yep, correct.

> I have a bad gut feeling about Gardenlets with effectively a wildcard to create certificates for any Seed cluster + being able to rotate that wildcard cert, too.

Me too. Though, this is a non-production use-case. If security should be improved then a dedicated gardenlet per seed is required - this is what we recommend also for other reasons for productive usage of Gardener.

> Do we break users by disallowing rotation and issuing of x509 certs via CSR for certs with CN gardener.cloud:system:seed:<ambiguous>?
This would at least limit such powerful certs to its cert lifetime so the installation is safe after that time passed.
This would impact Gardenlets reconciling multiple seeds - but this setup is only recommended for development use-cases anyways (where cert rotation does not matter).

I would not invest anything/care much about it due to the reason stated above: This scenario (one gardenlet responsible for multiple seeds and hence having a client cert with "ambiguous") is a non-production use-case. Also, it's not only about CSRs. For all the other resource handlings, such a gardenlet is allowed to interact with resource of other seeds (even if it's not responsible for it) because the seed authorizer cannot determine this.

pkg/gardenlet/bootstrap/certificate/certificate.go
rfranzke: 😂 Fixed with ee585e09714d1a182f8276c04d8e0ff7e8d7c783

timuthy (dismissed):
/lgtm

danielfoehrKn:
@rfranzke okay makes sense to not invest in the <ambiguous> topic. 

/lgtm

timuthy (approved):

rfranzke:
Thanks for your reviews folks! :)

Issue #3961, opened 2021-04-28
Adapt to helm v3


timebertt:
**How to categorize this issue?**

/area dev-productivity
/kind enhancement

**What would you like to be added**:

We should finally adapt all our hack scripts and so on to helm v3 and install v3 in `install-requirements.sh`.

**Why is this needed**:

Most developers will have migrated to helm v3 by now. When doing some development for gardener that uses helm for templating stuff (e.g. `hack/local-development/local-garden/configure-rbac` or `hack/generate-controller-registration.sh`), they will have to switch back to helm v2 again, as scripts break with v3 or result in other generation results.
This is quite annoying and I've been bitten by this quite often now.

rfranzke:
@timebertt can this be closed?

timebertt:
Yep 👍 
/close

PR #3994, opened 2021-05-04
Support helm3 in configure-rbac

poelzi:
helm3 has different calling syntax then helm2 for rendering templates
only. Detect if running under helm2 or 3 and use the proper syntax

/area dev-productivity
/kind enhancement

support for helm3

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/hack/generate-controller-registration.sh
+++ b/hack/generate-controller-registration.sh
@@ -45,9 +45,9 @@ VERSION_FILE="$3"
 DEST="$4"
 KIND_AND_TYPE="$5"
 
-VERSION="$(cat "$VERSION_FILE")"
-
 ( [[ -z "$NAME" ]] || [[ -z "$CHART_DIR" ]] || [[ -z "$DEST" ]] || [[ -z "$KIND_AND_TYPE" ]]) && usage
+
+VERSION="$(cat "$VERSION_FILE")"
 
 KINDS_AND_TYPES=("$KIND_AND_TYPE" "${@:6}")
 
--- a/hack/install-requirements.sh
+++ b/hack/install-requirements.sh
@@ -22,7 +22,7 @@ GO111MODULE=off go get -u golang.org/x/tools/cmd/goimports
 
 export GO111MODULE=on
 curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.27.0
-curl -s "https://raw.githubusercontent.com/helm/helm/v2.17.0/scripts/get" | bash -s -- --version 'v2.17.0'
+curl -s "https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3" | bash -s -- --version 'v3.5.4'
 
 platform=$(uname -s)
 if [[ ${platform} == "Linux" ]]; then
--- a/hack/local-development/local-garden/configure-rbac
+++ b/hack/local-development/local-garden/configure-rbac
@@ -20,26 +20,19 @@ KUBECONFIGPATH=${1:-"$(dirname $0)/kubeconfigs/default-admin.conf"}
 
 gardenlet_admin_templates=""
 
-HELM2=$(helm version 2>&1 | grep 'SemVer:"v2' || true)
-if [ -n "$HELM2" ]; then
-  HT="-x"
-else
-  HT="-s"
-fi
-
 if is_seedauthorizer_enabled; then
   KUBECONFIG=$KUBECONFIGPATH kubectl delete clusterrole/gardener.cloud:system:seeds        --ignore-not-found
   KUBECONFIG=$KUBECONFIGPATH kubectl delete clusterrolebinding/gardener.cloud:system:seeds --ignore-not-found
 else
-  gardenlet_admin_templates="$HT templates/clusterrole-seeds.yaml $HT templates/clusterrolebinding-seeds.yaml"
+  gardenlet_admin_templates="-s templates/clusterrole-seeds.yaml -s templates/clusterrolebinding-seeds.yaml"
 fi
 
 helm template \
   "$(dirname "$0")/../../../charts/gardener/controlplane/charts/application" \
-  $HT templates/clusterrole-admission-controller.yaml \
-  $HT templates/clusterrole-apiserver.yaml \
-  $HT templates/clusterrole-controller-manager.yaml \
-  $HT templates/clusterrole-scheduler.yaml $gardenlet_admin_templates \
+  -s templates/clusterrole-admission-controller.yaml \
+  -s templates/clusterrole-apiserver.yaml \
+  -s templates/clusterrole-controller-manager.yaml \
+  -s templates/clusterrole-scheduler.yaml $gardenlet_admin_templates \
   --set global.admission.enabled=true \
   --set global.apiserver.enabled=true \
   --set global.controller.enabled=true \
--- a/hack/local-development/local-garden/configure-rbac
+++ b/hack/local-development/local-garden/configure-rbac
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/usr/bin/env bash
 #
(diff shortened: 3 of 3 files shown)


gardener-robot-ci-1:
Thank you @poelzi for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

timebertt:
Thanks @poelzi. I get the motivation of the change, as I also ran into this, so I opened https://github.com/gardener/gardener/issues/3961.
I'm a bit reluctant to introduce a switch here and there for helm 2 vs 3, deal with this in all places and have inconsistent handling all over the place. I would rather like to see all usages of `helm` in our hack scripts adapted all at once, so we can completely switch to helm 3 everywhere.
So far, I haven't found the time to do it. Would you like to tackle https://github.com/gardener/gardener/issues/3961?

poelzi:
@timebertt I updated the install scripts to install version 3.5.4 and tested what is located in the hack folder. I removed the version check

hack/generate-controller-registration.sh
```diff
@@ -45,10 +45,10 @@ VERSION_FILE="$3"
 DEST="$4"
 KIND_AND_TYPE="$5"
 
-VERSION="$(cat "$VERSION_FILE")"
-
 ( [[ -z "$NAME" ]] || [[ -z "$CHART_DIR" ]] || [[ -z "$DEST" ]] || [[ -z "$KIND_AND_TYPE" ]]) && usage
 
+VERSION="$(cat "$VERSION_FILE")"
```
timebertt, line 50: I think we can also remove line 69 now:
https://github.com/gardener/gardener/blob/14b4c3730c492fbd86e8841cbe40f9ac82b8e4b4/hack/generate-controller-registration.sh#L69

timebertt (commented):
@poelzi thanks for updating the PR.
I now cross-checked the change with the aws extension and it seems to work mostly fine (`generate-controller-registration.sh` works and `ControllerRegistration` is installable/upgradable in a gardener installation).

However there is one issue when running `make check` (the `check-charts` part):
```
> Check Helm charts
Checking for chart symlink errors
Checking whether all charts can be rendered
walk.go:74: found symbolic link in path: /Users/d067603/go/src/github.com/gardener/gardener-extension-provider-aws/charts/gardener-extension-admission-aws/charts/application/values.yaml resolves to /Users/d067603/go/src/github.com/gardener/gardener-extension-provider-aws/charts/gardener-extension-admission-aws/values.yaml
walk.go:74: found symbolic link in path: /Users/d067603/go/src/github.com/gardener/gardener-extension-provider-aws/charts/gardener-extension-admission-aws/charts/runtime/values.yaml resolves to /Users/d067603/go/src/github.com/gardener/gardener-extension-provider-aws/charts/gardener-extension-admission-aws/values.yaml
Error: YAML parse error on aws-infra/templates/main.tf: error converting YAML to JSON: yaml: line 427: mapping values are not allowed in this context

Use --debug flag to render out invalid YAML
```
It complains when checking the `terraform` chart https://github.com/gardener/gardener-extension-provider-aws/tree/master/charts/internal/aws-infra, which we use to template the terraform config, as the resources there are not valid yaml.
I see two options for dealing with this:
- either exclude the terraform config chart from being checked in `check-charts`
- or completely remove the chart (it's a bit hacky anyway) and replace it with some plain go template (should be sufficient for the use case)

(similar things would need to be done in the other provider extensions using terraform as well)

@gardener/gardener-maintainers opinions on this one?

Side-note: I also checked the [helm v3 migration guide](https://helm.sh/docs/topics/v2_v3_migration) and it seems like there are no breaking changings wrt the templating engine itself (which is mainly the part of helm we actually use), so I assume, proceeding like this with the upgrade should be fine.

rfranzke:
@timebertt I don't get what the issue with the TF chart of AWS is - I don't even find this line:

> Error: YAML parse error on aws-infra/templates/main.tf: error converting YAML to JSON: yaml: line 427: mapping values are not allowed in this context

Can you explain?

timebertt:
helm 3 complains, if the rendered document is not valid YAML (e.g. when using `helm template`). Then you get the above error.

rfranzke:
Aha, interesting. Then I guess we should go for using native Go templates and `embed` instead of the chart (this also fits to what we do with #2754 in g/g). However, that's quite some effort as it needs to be done for most provider extensions...

timebertt:
> However, that's quite some effort as it needs to be done for most provider extensions...

True. That's why I'm asking for more opinions :)
Should we still go ahead with the helm upgrade now and add the adaption part in the extensions to the backlog (there's some time left until we'll vendor the next gardener release)?

rfranzke:
I would rather vote for reversing this, i.e., first adapt the extensions, and when we are happy then we can go ahead and update Helm to v3. This way there is no "pressure" to adapt.

timebertt:
Ok, @poelzi would you be fine with holding your PR until then?

poelzi:
The first version of my patch was fine with just checking which version of helm was running and using the appropriate switch depending on the helm version. I'm not using the install requirements script because I'm running nixos and will not clutter my system with a script downloading executables somewhere.

rfranzke:
Yeah, I agree, the demand is out of scope of @poelzi's PR. Hence, to proceed, I'd propose to go back to the initial version of this PR and have the switch in the `configure-rbac` script so that it works with both Helm 2/3. In parallel (but independently), we should try to prepare the extensions for the Helm 3 update.

rfranzke:
FYI: I've opened a PR for the AWS extension which replaces the "infra" Helm chart with native Golang templates: https://github.com/gardener/gardener-extension-provider-aws/pull/332

timebertt:
> FYI: I've opened a PR for the AWS extension which replaces the "infra" Helm chart with native Golang templates: [gardener/gardener-extension-provider-aws#332](https://github.com/gardener/gardener-extension-provider-aws/pull/332)

Now that we have this as an example, shouldn't we just go ahead with this PR as is? It shouldn't be too much effort to adapt the other extensions similarly, right?
(we also still have ~2 weeks left before revendoring this g/g version)

rfranzke:
Depends on whether the extensions maintainers have time and capacity to adapt... Generally, I think it produces avoidable "pressure", but let's ask them: @kon-angelo @dkistner @prashanth26 @minchaow Do you think you can do something similar like https://github.com/gardener/gardener-extension-provider-aws/pull/332 for the other provider extensions within the next week?

Let's
/hold
until this is clarified.

prashanth26:
> Depends on whether the extensions maintainers have time and capacity to adapt... Generally, I think it produces avoidable "pressure", but let's ask them: @kon-angelo @dkistner @prashanth26 @minchaow Do you think you can do something similar like [gardener/gardener-extension-provider-aws#332](https://github.com/gardener/gardener-extension-provider-aws/pull/332) for the other provider extensions within the next week?
> 
> Let's
> /hold
> until this is clarified.

Hi Guys,

I have some priority feature releases on other components (like a serialized reattachment of PVs and more) to make this week which will probably eat up most of my time. I know this change is probably not too difficult to make, however, I would like to avoid committing to this change by EO this week. For me, it would be great if I could get a couple of weeks from today to incorporate and test this change. However, if others are okay with the above deadline and if this is something needed on priority, I can try (no promises) to make the change later this week.

Thanks & Regards,
Prashanth

dkistner:
Same for me.
I have some priority dev task to continue in the next ~1-2 weeks and I will most likely not have time to look into it this week.
Same is probably true for @kon-angelo, but he wanted to explore a little bit what would be required to do, but he has also some major development activities in front.

kon-angelo:
My current task involves quite a bit of refactoring for the terraform files in Azure and I wanted to see what will be missing when we swap to the go templates. I ended up making the PR for Azure, but as the other colleagues said, there other high priority tasks in the works and I would hope that we get a bit more time.

rfranzke:
@poelzi @timebertt WDYT about rolling back the PR to the initial version (switch between v2/v3) for now? After the colleagues have adapted the extensions we can delete the v2 parts.

timebertt (approved):
Let's get this in :) 
/lgtm

rfranzke (approved):
/lgtm
/unhold

Thanks for your help, colleagues!

PR #3995, opened 2021-05-04
Add support for ControllerDeployments

timuthy:

/area security
/kind api-change
/kind enhancement

This PR adds a new resource `ControllerDeployment` to the `core.gardener.cloud` API group with related controllers. A `ControllerDeployment` is supposed to contain definitions about how extension controllers are deployed to seed clusters. Earlier, these deployment instruction were part of `ControllerRegistration`s but due to security reasons it's now possible to outsource this information (usually Helm chart + values) to a dedicated `ControllerDeployment` which is again referenced by a `ControllerRegistration`.
With the help of the `SeedAuthorizer`, the Gardenlet is only permitted to read `ControllerDeployment`s (may contain sensitive information) which correspond to the seed, the Gardenlet is responsible for. This restriction was not possible with `ControllerRegistration`s only (see linked issue).

Fixes parts of #1723

**Release note**:

```feature operator
A new resource `ControllerDeployment` was added to the `core.gardener.cloud` API group. A `ControllerDeployment` is supposed to contain specifications about how extension controllers are deployed to seed clusters. Specifying deployments via `ControllerRegistration`s directly is therefore deprecated. Instead, we recommend to use a dedicated `ControllerDeployment` resource and reference it in a `ControllerRegistration` via `spec.deployment.DeploymentRefs[].Name`. Please consult the [documentation](/docs/extensions/controllerregistration.md) for more information.
As `ControllerDeployment`s may contain sensitive information, we recommend to add this type to the Gardener API server's `EncryptionConfiguration`. This happens automatically if you deploy Gardener via our pre-configured Helm charts (`/charts/gardener`). 
```
```breaking developer
Deployment specific information from `ControllerRegistration`s was outsourced to a new resource type `ControllerDeployment`. If you use Gardener's `/hack` scripts and utilities, please make sure to run `make generate` after you have updated your Gardener dependencies to this version, as it updates the example registration in your extension. Otherwise, please make sure to adjust your example `ControllerRegistration` manifest to the new structure.
```


pkg/apis/core/types_controllerregistration.go
```diff
@@ -69,18 +69,27 @@ type ControllerResource struct {
...
+type DeploymentRef struct {
+	// Name is the name of the `ControllerDeployment` that is being referred to.
+	Name string
+}
+
+// ControllerRegistrationDeployment contains information for how this controller is deployed.
+type ControllerRegistrationDeployment struct {
 	// Type is the deployment type.
-	Type string
+	Type *string
```
rfranzke, line 82: This field should also have a deprecation note, or?

pkg/apis/core/v1alpha1/types_controllerdeployment.go
```diff
@@ -0,0 +1,53 @@
...
+	// Items is the list of ControllerDeployments.
+	Items []ControllerDeployment `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+
+// ControllerDeploymentSpec is the specification of a ControllerDeployment.
+type ControllerDeploymentSpec struct {
+	// Type is the deployment type.
+	Type string `json:"type" protobuf:"bytes,1,opt,name=type"`
+	// ProviderConfig contains type-specific configuration. It contains assets that deploy the controller.
+	ProviderConfig runtime.RawExtension `json:"providerConfig,omitempty" protobuf:"bytes,2,opt,name=providerConfig"`
```
rfranzke: Why `omitempty`?

pkg/apis/core/v1beta1/types_controllerdeployment.go
```diff
@@ -0,0 +1,53 @@
...
+	// Items is the list of ControllerDeployments.
+	Items []ControllerDeployment `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+
+// ControllerDeploymentSpec is the specification of a ControllerDeployment.
+type ControllerDeploymentSpec struct {
+	// Type is the deployment type.
+	Type string `json:"type" protobuf:"bytes,1,opt,name=type"`
+	// ProviderConfig contains type-specific configuration. It contains assets that deploy the controller.
+	ProviderConfig runtime.RawExtension `json:"providerConfig,omitempty" protobuf:"bytes,2,opt,name=providerConfig"`
```
rfranzke: Why `omitempty`?

pkg/apis/core/v1beta1/types_controllerregistration.go
```diff
@@ -75,11 +75,20 @@ type ControllerResource struct {
 	Primary *bool `json:"primary,omitempty" protobuf:"varint,5,opt,name=primary"`
 }
 
-// ControllerDeployment contains information for how this controller is deployed.
-type ControllerDeployment struct {
+// DeploymentRef contains information about `ControllerDeployment` references.
+type DeploymentRef struct {
+	// Name is the name of the `ControllerDeployment` that is being referred to.
+	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
```
rfranzke: Why `omitempty`?

pkg/apis/core/v1alpha1/types_controllerregistration.go
```diff
@@ -75,11 +75,20 @@ type ControllerResource struct {
 	Primary *bool `json:"primary,omitempty" protobuf:"varint,5,opt,name=primary"`
 }
 
-// ControllerDeployment contains information for how this controller is deployed.
-type ControllerDeployment struct {
+// DeploymentRef contains information about `ControllerDeployment` references.
+type DeploymentRef struct {
+	// Name is the name of the `ControllerDeployment` that is being referred to.
+	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
```
rfranzke: Why `omitempty`?

pkg/apis/core/validation/controllerdeployment.go
```diff
@@ -0,0 +1,47 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validation
+
+import (
+	"github.com/gardener/gardener/pkg/apis/core"
```
rfranzke, line 18:
```suggestion
	"github.com/gardener/gardener/pkg/apis/core"

```

pkg/apis/core/validation/controllerregistration.go
```diff
@@ -88,17 +88,38 @@ func ValidateControllerRegistrationSpec(spec *core.ControllerRegistrationSpec, f
...
-			allErrs = append(allErrs, metav1validation.ValidateLabelSelector(spec.Deployment.SeedSelector, deploymentPath.Child("seedSelector"))...)
+			allErrs = append(allErrs, metav1validation.ValidateLabelSelector(deployment.SeedSelector, deploymentPath.Child("seedSelector"))...)
+		}
+
+		deploymentRefsCount := len(deployment.DeploymentRefs)
+		if deploymentRefsCount > 0 {
+			if deployment.ProviderConfig != nil {
+				allErrs = append(allErrs, field.Forbidden(deploymentPath.Child("providerConfig"), "specifying a providerConfig is not allowed while also specifying deploymentRefs"))
+			}
+			if deployment.Type != nil && len(*deployment.Type) > 0 {
```
rfranzke:
```suggestion
			if deployment.Type != nil {
```

would be good enough and disallow `type: ""`

pkg/apis/core/types_controllerdeployment.go
```diff
@@ -0,0 +1,52 @@
...
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ControllerDeployment contains information about how this controller is deployed.
+type ControllerDeployment struct {
+	metav1.TypeMeta
+	// Standard object metadata.
+	metav1.ObjectMeta
+	// Spec contains the specification of this deployment.
+	Spec ControllerDeploymentSpec
```
rfranzke: I'm wondering whether we need `Spec` in this resource - WDYT?

pkg/apis/core/v1alpha1/types_controllerinstallation.go
```diff
@@ -49,10 +49,13 @@ type ControllerInstallationList struct {
...
-	RegistrationRef corev1.ObjectReference `json:"registrationRef" protobuf:"bytes,1,opt,name=registrationRef"`
-	// SeedRef is used to reference a Seed resources.
-	SeedRef corev1.ObjectReference `json:"seedRef" protobuf:"bytes,2,opt,name=seedRef"`
+	// DeploymentRef is used to reference a ControllerDeployment resource.
+	// +optional
+	DeploymentRef *corev1.ObjectReference `json:"deploymentRef" protobuf:"bytes,1,opt,name=deploymentRef"`
+	// RegistrationRef is used to reference a ControllerRegistration resource.
+	RegistrationRef corev1.ObjectReference `json:"registrationRef" protobuf:"bytes,2,opt,name=registrationRef"`
+	// SeedRef is used to reference a Seed resource.
+	SeedRef corev1.ObjectReference `json:"seedRef" protobuf:"bytes,3,opt,name=seedRef"`
```
rfranzke:
```suggestion
	// RegistrationRef is used to reference a ControllerRegistration resource.
	RegistrationRef corev1.ObjectReference `json:"registrationRef" protobuf:"bytes,1,opt,name=registrationRef"`
	// SeedRef is used to reference a Seed resource.
	SeedRef corev1.ObjectReference `json:"seedRef" protobuf:"bytes,2,opt,name=seedRef"`
	// DeploymentRef is used to reference a ControllerDeployment resource.
	// +optional
	DeploymentRef *corev1.ObjectReference `json:"deploymentRef,omitempty" protobuf:"bytes,3,opt,name=deploymentRef"`
```

- Changing the order in protobuf is an incompatible change
- Also `omitempty` was missing

pkg/apis/core/v1beta1/types_controllerinstallation.go
```diff
@@ -49,10 +49,13 @@ type ControllerInstallationList struct {
...
-	RegistrationRef corev1.ObjectReference `json:"registrationRef" protobuf:"bytes,1,opt,name=registrationRef"`
-	// SeedRef is used to reference a Seed resources.
-	SeedRef corev1.ObjectReference `json:"seedRef" protobuf:"bytes,2,opt,name=seedRef"`
+	// DeploymentRef is used to reference a ControllerDeployment resource.
+	// +optional
+	DeploymentRef *corev1.ObjectReference `json:"deploymentRef" protobuf:"bytes,1,opt,name=deploymentRef"`
+	// RegistrationRef is used to reference a ControllerRegistration resource.
+	RegistrationRef corev1.ObjectReference `json:"registrationRef" protobuf:"bytes,2,opt,name=registrationRef"`
+	// SeedRef is used to reference a Seed resource.
+	SeedRef corev1.ObjectReference `json:"seedRef" protobuf:"bytes,3,opt,name=seedRef"`
```
rfranzke:
```suggestion
	// RegistrationRef is used to reference a ControllerRegistration resource.
	RegistrationRef corev1.ObjectReference `json:"registrationRef" protobuf:"bytes,1,opt,name=registrationRef"`
	// SeedRef is used to reference a Seed resource.
	SeedRef corev1.ObjectReference `json:"seedRef" protobuf:"bytes,2,opt,name=seedRef"`
	// DeploymentRef is used to reference a ControllerDeployment resource.
	// +optional
	DeploymentRef *corev1.ObjectReference `json:"deploymentRef,omitempty" protobuf:"bytes,3,opt,name=deploymentRef"`
```

- Changing the order in protobuf is an incompatible change
- Also `omitempty` was missing

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -414,7 +414,20 @@ func deployNeededInstallations(
 
 		logger.Infof("Deploying wanted ControllerInstallation for %q", registrationName)
 
-		if err := deployNeededInstallation(ctx, c, seed, controllerRegistrations[registrationName].obj, registrationNameToInstallationName[registrationName]); err != nil {
+		var (
+			controllerDeployment   *gardencorev1beta1.ControllerDeployment
+			controllerRegistration = controllerRegistrations[registrationName].obj
+		)
```
rfranzke, line 420:
```suggestion
		)

```

pkg/controllermanager/controller/controllerdeployment/controllerdeployment.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/gardener/gardener/pkg/controllerutils"
+	"github.com/gardener/gardener/pkg/gardenlet"
+)
+
+// FinalizerName is the finalizer used by this controller.
+const FinalizerName = "core.gardener.cloud/controllerdeployment"
+
+// Controller controls ManagedSeedSets.
+type Controller struct {
+	clientMap clientmap.ClientMap
```
rfranzke: It seems unnecessary to store this on the struct?

pkg/controllermanager/controller/controllerdeployment/controllerdeployment.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// Run runs the Controller until the given context is cancelled.
+func (c *Controller) Run(ctx context.Context, workers int) {
+	var waitGroup sync.WaitGroup
+
+	c.controllerDeploymentInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.controllerDeploymentAdd,
+		UpdateFunc: c.controllerDeploymentUpdate,
+	})
```
rfranzke: As opposed to most other controllers, why are the event handlers only added here?

pkg/controllermanager/controller/controllerdeployment/controllerdeployment.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if !cache.WaitForCacheSync(ctx.Done(), c.controllerDeploymentInformer.HasSynced, c.controllerRegistrationInformer.HasSynced) {
+		c.logger.Error("Timed out waiting for caches to sync")
+		return
+	}
```
rfranzke: Also, as opposed to most other controllers, why don't you use the `hasSynced` attribute? Then you can remove the informer from the controller struct?

pkg/controllermanager/controller/controllerdeployment/controllerdeployment_control.go
```diff
@@ -0,0 +1,105 @@
...
+func NewReconciler(clientMap clientmap.ClientMap, l logrus.FieldLogger) reconcile.Reconciler {
+	return &controllerDeploymentReconciler{
+		logger:    l,
+		clientMap: clientMap,
+	}
+}
+
+type controllerDeploymentReconciler struct {
+	logger    logrus.FieldLogger
+	clientMap clientmap.ClientMap
```
rfranzke:
```suggestion
	gardenClient client.Client
```

pkg/controllermanager/controller/controllerdeployment/controllerdeployment_control.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	c.controllerDeploymentQueue.Add(key)
+}
+
+func (c *Controller) controllerDeploymentUpdate(_, newObj interface{}) {
+	c.controllerDeploymentAdd(newObj)
+}
+
+// NewReconciler creates a new instance of a reconciler which reconciles ControllerDeployments.
+func NewReconciler(clientMap clientmap.ClientMap, l logrus.FieldLogger) reconcile.Reconciler {
```
rfranzke: Can you pass the gardenClient directly? Also, can you make the logger the first argument (that's how most other controllers are built)?

pkg/controllermanager/controller/controllerdeployment/controllerdeployment_control.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		for _, controllerRegistration := range controllerRegistrationList.Items {
+			deployment := controllerRegistration.Spec.Deployment
+			if deployment == nil {
+				continue
+			}
+			if len(deployment.DeploymentRefs) < 1 {
+				continue
+			}
```
rfranzke: This is not needed

pkg/controllermanager/controller/controllerdeployment/controllerdeployment_control_test.go
```diff
@@ -0,0 +1,214 @@
...
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	fakeclientmap "github.com/gardener/gardener/pkg/client/kubernetes/clientmap/fake"
+	"github.com/gardener/gardener/pkg/client/kubernetes/clientmap/keys"
+	. "github.com/gardener/gardener/pkg/controllermanager/controller/controllerdeployment"
+	"github.com/gardener/gardener/pkg/logger"
+	mockclient "github.com/gardener/gardener/pkg/mock/controller-runtime/client"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+)
+
+var _ = Describe("Controller", func() {
```
rfranzke, line 44: Should we rather have an envtest for this reconciler (similar to https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/shoot/shoot_retry_control_test.go)?

pkg/controllermanager/controller/controllerdeployment/controllerdeployment_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+package controllerdeployment_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestProject(t *testing.T) {
```
rfranzke:
```suggestion
func TestControllerDeployment(t *testing.T) {
```

pkg/controllermanager/controller/controllerregistration/controllerdeployment_control.go
```diff
@@ -0,0 +1,52 @@
...
+package controllerregistration
+
+import (
+	"context"
+
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	"github.com/gardener/gardener/pkg/logger"
+)
+
+func (c *Controller) controllerDeploymentAdd(ctx context.Context, obj interface{}) {
```
rfranzke, line 24: What about some small unit tests similar to https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/controllerregistration/controllerregistration_control_test.go?

pkg/gardenlet/controller/controllerinstallation/controllerinstallation_control.go
```diff
@@ -353,13 +370,20 @@ func updateConditions(ctx context.Context, gardenClient kubernetes.Interface, co
...
 	}
 
 	if deployment := controllerRegistration.Spec.Deployment; deployment != nil {
+		if deploymentName := controllerInstallation.Spec.DeploymentRef; deploymentName != nil {
+			controllerDeployment := &gardencorev1beta1.ControllerDeployment{}
+			if err := cl.Get(ctx, kutil.Key(deploymentName.Name), controllerDeployment); err != nil {
+				return false, err
+			}
+			return controllerDeployment.Spec.Type == installationTypeHelm, nil
+		}
```
rfranzke: Why is this nested here?
Generally, it was a little confusing to read this because you are checking the `ControllerInstallation`'s `.spec.deploymentRef` here while you are checking the `ControllerRegistration`'s `.spec.deployment.type` below.

As you already read the `ControllerRegistration` with the client, what about only considering its spec and iterating over `.spec.deployment.deploymentRefs` and read the respective `ControllerDeployment`s?

pkg/admissioncontroller/webhooks/auth/seed/authorizer.go
```diff
@@ -103,6 +104,12 @@ func (a *authorizer) Authorize(_ context.Context, attrs auth.Attributes) (auth.D
 			return a.authorizeRead(seedName, graph.VertexTypeCloudProfile, attrs)
 		case configMapResource:
 			return a.authorizeConfigMap(seedName, attrs)
+		case controllerDeploymentResource:
+			return a.authorize(seedName, graph.VertexTypeControllerDeployment, attrs,
+				[]string{"get"},
+				nil,
+				nil,
+			)
```
rfranzke: use `authorizeRead`?

rfranzke (changes requested):
Very nice!

Apart from the inline comments:

- Can you extend the graph picture on top of https://github.com/gardener/gardener/blob/998aeb7d45317f7b4d784fda237ddef873ec318a/docs/deployment/gardenlet_api_access_graph.png, please?
- Can you extend https://github.com/gardener/gardener/blob/master/charts/gardener/controlplane/charts/runtime/templates/controller-manager/configmap-componentconfig.yaml?

pkg/controllermanager/controller/controllerdeployment/controllerdeployment_control_test.go
timuthy: As discussed with @timebertt, @stoyanr and @ialidzhikov last week, adding envtests comes with a price, especially if we setup and tear down the environment in every suite and our pipeline resources aren't sufficient enough (sometimes) which has ended in flaky tests so far. The CC cache can improve the situation but we are not sure, yet.
Bottom line, I think the controller is so simple and very similar to https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/controllerregistration/controllerregistration_control_test.go, so that these unit tests cover the given cases sufficiently and an envtest doesn't seem too beneficial to me (except for consistency reasons). WDYT?

pkg/gardenlet/controller/controllerinstallation/controllerinstallation_control.go
timuthy: > As you already read the ControllerRegistration with the client, what about only considering its spec and iterating over .spec.deployment.deploymentRefs and read the respective ControllerDeployments?

This meant that we'd considered every `ControllerDeployment` in the list. We may neglect this today since only one is allowed but in the future, if more refs will be allowed, we are only interested in the ref that comes from the `ControllerInstallation`, right? Because then we know for sure, it's the deployment that affects the seed the Gardenlet is responsible for.

rfranzke: > we are only interested in the ref that comes from the ControllerInstallation, right?

Right, of course. Makes sense, then let's keep it as is. However, the nesting is not needed, or?

pkg/controllermanager/controller/controllerdeployment/controllerdeployment_control_test.go
rfranzke: Yeah, I'm a bit unsure about this problem domain. While doing my recent refactorings in the controllers, I learned that it was very tedious to adapt the unit tests because too many implementation details were tested (all these mocks with many different clients, etc.). My impression is that at least the tests for the reconcilers/controllers would be easier to maintain and way clearer if we used envtests.
I don't insist on changing it now for this PR (everything is implemented already), but we should have a talk about it and clarify how to proceed in the future. If the only argument for not doing it is a potentially unstable CI/CD environment then let's try to work on this with the colleagues or even switch to a more stable env. But maybe you guys also discussed other reasons.

timuthy: > If the only argument for not doing it is a potentially unstable CI/CD environment

That's actually not my main concern but that we should evaluate if an integration test over unit test would be beneficial here. In this controller, we have only a few dependencies, so I wonder why we shouldn't prefer small and fast unit tests. But yes, let's have a discussion about this.

timebertt: > While doing my recent refactorings in the controllers, I learned that it was very tedious to adapt the unit tests because too many implementation details were tested (all these mocks with many different clients, etc.). My impression is that at least the tests for the reconcilers/controllers would be easier to maintain and way clearer if we used envtests.

+100, that's also the reason why I'm pushing envtests for our controllers. I learned to hate mocks for controller tests during my recent refactorings / cleanups as well. They force you to adapt soo many unit tests that rely on mocks when changing a simple helper func from `TryUpdate` to `Patch`. 

Yes, let's have a chat about this :)

pkg/apis/core/types_controllerdeployment.go
timuthy: I was also pondering about that and eventually decided for the `Spec` because then we are open to add a `Status` in the future. Otherwise, I don't have a strong opinion. WDYT?

rfranzke: Not sure, what would we need a status for? Isn't this resource similar to `ConfigMap`s/`ShootState`s/...?

timuthy: You're right, we will probably always maintain the status of an extension controller "deployment" in the `ControllerInstallation`. I removed the spec in my latest commit.

charts/gardener/controlplane/values.yaml
```diff
@@ -40,6 +40,7 @@ global:
         kind: EncryptionConfiguration
         resources:
           - resources:
+            - controllerdeployments.core.gardener.cloud
```
rfranzke, line 43: Nice catch, I think this deserves a special mention in the release note.

pkg/apis/core/validation/controllerregistration_test.go
```diff
@@ -277,6 +277,17 @@ var _ = Describe("validation", func() {
 			}))))
 		})
 
+		It("should forbid to set an empty deploymentpkg/controllermanager/controller/controllerdeployment/controllerdeployment_control_test.go type", func() {
```
rfranzke: fix description string :)

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
```diff
@@ -478,12 +479,12 @@ func deployNeededInstallation(
 		kutil.SetMetaDataLabel(&controllerInstallation.ObjectMeta, common.RegistrationSpecHash, registrationSpecHash)
 
 		if controllerDeployment != nil {
-			deploymentSpecMap, err := convertObjToMap(controllerDeployment.Spec)
+			deploymentMap, err := convertObjToMap(controllerDeployment)
```
rfranzke: Hmm, interesting.. this now in corporates also the metadata etc. of the `ControllerDeployment`, right? Not sure if that's a good idea, maybe we should still compute a checksum only based on `type` and `providerConfig`?

rfranzke (changes requested):
Thanks for addressing the suggestions @timuthy! Looks very good now, found a few more things.

Once you updated the PR I still need to play with it locally, but I guess we can work towards getting it merged by end of this week :)

pkg/controllermanager/controller/controllerregistration/controllerregistration_seed_control.go
timuthy: That was actually one advantage of having a `spec` container, also when you look at the [mustIncreaseGeneration](https://github.com/gardener/gardener/pull/3995/files#diff-cdb4693983f8eb335e83f9782b23f9f76ab81ce8929e5fcc4fed6e08f7ff5ee8R59) function. I decided to take the entire `ControllerDeployment` object for the sake of simplicity and because we don't anticipate frequent metadata changes for this resource type. I almost vote for getting back the `spec`, so we don't have to explicitly list every field in such functions which is always bad for maintenance. WDYT?

rfranzke: I think the impact is very little if you simply use `type` and `providerConfig` explicitly here. No need to have `spec` again just to be able to call to compute the hash sum more "conveniently". Currently, whenever the `.metadata.resourceVersion` changes (for whatever reason) the hash changes which is undesirable, I assume.

timuthy: It was just that one misses to adjust these places when new fields are added to the API (not because of implementation effort) which wouldn't be required if we had one spec.

rfranzke: OK, sure, though, it's unlikely/not planned/didn't even happen since introduction of the `ControllerRegistration` (for the `.spec.deployment` field) that we add some new field that needs to be considered here.
You could still make sure that this place is not forgotten when the API is extended by adding a comment to the API definition of the `ControllerDeployment`.
Again, given that the frequency of code changes in this area are to be expected very minimal/quasi zero and that the impact of adapting these lines are quite minimal, I still suggest to only consider the relevant fields for the hash sum.

rfranzke (approved):
/lgtm

Thanks a lot for the PR, played with it locally - works like charm 🦄 :)

vpnachev:
> With the help of the SeedAuthorizer, the Gardenlet is only permitted to read ControllerDeployments (may contain sensitive information) which correspond to the seed, the Gardenlet is responsible for. This restriction was not possible with ControllerRegistrations only (see linked issue).

What was the reason to implement new resource instead of using the existing ones? I mean,  wasn't it possible to copy the deployment spec into the `ControllerInstallation.Status`, let gardenlet access only the controller installations for the seed that it is responsible for (`spec.seedRef.name=<seed-managed-by-this-gardenlet>`) and forbid gardenlet to read ControllerRegistrations?

PR #3998, opened 2021-05-05
Fix handling of aggregate-prometheus resources with hvpa

wyb1:

/area monitoring
/kind bug

Fix an issue where the gardenlet overwrites changes from `hvpa` for the `aggregate-prometheus`.

Fixes #

/cc @amshuman-kr 
**Release note**:

```bugfix operator
Fix an issue where the gardenlet overwrites changes from `hvpa` for the `aggregate-prometheus`.
```

### DIFF
--- a/pkg/operation/seed/seed.go
+++ b/pkg/operation/seed/seed.go
@@ -563,9 +563,9 @@ func RunReconcileSeedFlow(
 			if err != nil {
 				return err
 			}
-			if len(currentResources) != 0 && currentResources[resource] != nil {
+			if len(currentResources) != 0 && currentResources["prometheus"] != nil {
 				monitoringResources[resource] = map[string]interface{}{
-					resource: currentResources[resource],
+					"prometheus": currentResources["prometheus"],
 				}
 			}
 		}
--- a/charts/seed-bootstrap/templates/aggregate-prometheus/statefulset.yaml
+++ b/charts/seed-bootstrap/templates/aggregate-prometheus/statefulset.yaml
@@ -58,7 +58,7 @@ spec:
           successThreshold: 1
           timeoutSeconds: 3
         resources:
-          {{- toYaml .Values.prometheus.resources.prometheus | nindent 10 }}
+          {{- toYaml .Values.aggregatePrometheus.resources.prometheus | nindent 10 }}
         volumeMounts:
         - mountPath: /etc/prometheus/config
           name: config


pkg/operation/seed/seed.go
```diff
@@ -563,7 +563,7 @@ func RunReconcileSeedFlow(
 			if err != nil {
 				return err
 			}
-			if len(currentResources) != 0 && currentResources[resource] != nil {
+			if len(currentResources) != 0 && currentResources["prometheus"] != nil {
 				monitoringResources[resource] = map[string]interface{}{
 					resource: currentResources[resource],
```
timuthy:
```suggestion
					"prometheus": currentResources[resource],
```
Right?

wyb1: I fixed it now. `resource` should not be changed because it is the key for the values https://github.com/gardener/gardener/blob/master/pkg/operation/seed/seed.go#L781
`currentResources[resource]` should always be `currentResources["prometheus"]` because the container name is always `prometheus`.

timuthy:
@rfranzke I talked to @wyb1 in person. There are some changes missing.
/status author-action

pkg/operation/seed/seed.go
```diff
@@ -563,9 +563,9 @@ func RunReconcileSeedFlow(
 			if err != nil {
 				return err
 			}
-			if len(currentResources) != 0 && currentResources[resource] != nil {
+			if len(currentResources) != 0 && currentResources["prometheus"] != nil {
 				monitoringResources[resource] = map[string]interface{}{
-					resource: currentResources[resource],
+					resource: currentResources["prometheus"],
```
amshuman-kr: If I understood the code correctly, there is a mismatch among the three things that should ideally be the same.

1. The values rendering in this function along with [here](https://github.com/gardener/gardener/pull/3998/files#diff-8495b57b25d39744fe1b20b1e1ce695d130f730142af86ff9c5929afcdbabeb4L780-L781) is effectively passing `.Values.aggregatePrometheus.resources.aggregate-prometheus`.
2. The [value.yaml](https://github.com/gardener/gardener/blob/1c3fed84dd2c50aef861b7e0e57f2931c0cbe9fd/charts/seed-bootstrap/values.yaml#L56-L71) of the chart is expecting the structure to be  effectively `.Values.aggregatePrometheus.resources.prometheus`.
3. The [`StatefulSet` YAML](https://github.com/gardener/gardener/blob/1c3fed84dd2c50aef861b7e0e57f2931c0cbe9fd/charts/seed-bootstrap/templates/aggregate-prometheus/statefulset.yaml#L61) is using `.Values.prometheus.resources.prometheus`.

amshuman-kr (changes requested):
@wyb1 I think there is still a problem with the resources rendered rightly for aggregated-prometheus. Please see my comments below.

timuthy (approved):
/lgtm

amshuman-kr (approved):
LGTM

PR #4000, opened 2021-05-05
[GEP-17] Rate limiting DNS entries created by shoot-dns-service

MartinWeindel:

/area control-plane
/area robustness
/area security
/kind enhancement

This `GEP` introduces rate limits for `DNSEntries` created with the help of the [shoot-dns-service](https://github.com/gardener/gardener-extension-shoot-dns-service) to protect the [dns-controller-manager](https://github.com/gardener/external-dns-management) from being throttled, i.e to protect the DNS management of Gardener from intentional or unintentional *DoS* attacks.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/docs/proposals/17-ratelimits-for-shoot-dns-entries.md
+++ b/docs/proposals/17-ratelimits-for-shoot-dns-entries.md
@@ -35,7 +35,7 @@ The `dns-controller-manager` runs update requests for DNS record on the configur
 ### Goals
 
 - The rate limit is set and applied separatedly for each `DNSProvider`
-- All update requests caused by `DNSEntries` created or updated from the `shoot-dns-service` should only be forwarded to the backend DNS service, if the update requests are within rate limit of the `DNSProvider` the `DNSEntry` is assigned to.
+- All update requests caused by `DNSEntries` created or updated from the `shoot-dns-service` should only be processed by the backend DNS service, if the update requests are within rate limit of the `DNSProvider` the `DNSEntry` is assigned to.
 - Deletion of `DNSEntries` is not rate limited to avoid freezing on shoot cluster deletion.
 - Updates of `DNSEntries` created by Gardener (e.g. `internal`, `external`, and `ingress`) should not be blocked by the rate limits.
 - If creation/updates of `DNSEntries` of the `shoot-dns-service` are pending because of a rate limit, this should also made be visible in the shoot cluster (either by events or status update)
--- a/docs/proposals/17-ratelimits-for-shoot-dns-entries.md
+++ b/docs/proposals/17-ratelimits-for-shoot-dns-entries.md
@@ -101,6 +101,24 @@ On replicating a `DNSEntry` from the shoot cluster nothing must be changed, as a
 ### Gardener
 
 The rate limit for the default external `DNSProvider` is configured for the default domains of the garden.
+As there are already other annotations in the [default domain secret](https://github.com/gardener/gardener/blob/7616e05b6dedd94b4c6e48c26ac026da4d177191/example/10-secret-default-domain.yaml#L13), an annotation `dns.gardener.cloud/rate-limit` should contain values for both requestsPerDay and burst parameters.
+
+Example:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: default-domain
+  namespace: garden
+  annotations:
+    dns.gardener.cloud/provider: aws-route53
+    dns.gardener.cloud/domain: example.com
+    dns.gardener.cloud/rate-limit: "requestsPerDay=120,burst=20"
+  # dns.gardener.cloud/include-zones: "first-zone-id,second-zone-id"
+  # dns.gardener.cloud/exclude-zones: "first-zone-id,second-zone-id"
+```
+
 It must also be possible to set a rate-limit for an additional `DNSProvider` specified in the shoot manifest optionally.
 
 On creating a new shoot (with enabled DNS), the external `DNSProvider` is created with the rate limit for the default domain or as specified in the shoot manifest for the primary `DNSProvider`.
--- a/docs/proposals/17-ratelimits-for-shoot-dns-entries.md
+++ b/docs/proposals/17-ratelimits-for-shoot-dns-entries.md
@@ -1,0 +1,112 @@
+---
+title: Rate limits for DNS entries created by shoot-dns-service
+gep-number: 17
+creation-date: 2021-05-03
+status: implementable
+authors:
+- "@MartinWeindel"
+reviewers:
+- "@timuthy"
+---
+
+# GEP-17: Rate limits for DNS entries created by shoot-dns-service
+
+## Table of Contents
+
+- [GEP-17: Rate limits for DNS entries created by shoot-dns-service](#gep-17-rate-limits-for-dns-entries-created-by-shoot-dns-service)
+  - [Table of Contents](#table-of-contents)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [Proposal](#proposal)
+    - [DNS controller manager](#dns-controller-manager)
+    - [Gardener](#gardener)
+  - [Alternatives](#alternatives)
+
+## Summary
+
+This `GEP` introduces rate limits for `DNSEntries` created with the help of the [shoot-dns-service](https://github.com/gardener/gardener-extension-shoot-dns-service) to protect the [dns-controller-manager](https://github.com/gardener/external-dns-management) from being throttled, i.e to protect the DNS management of Gardener from intentional or unintentional *DoS* attacks.
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


docs/proposals/17-ratelimits-for-shoot-dns-entries.md
```diff
@@ -0,0 +1,112 @@
...
+Here the state `Updating` is a new, additional state to be implemented to make transparent that spec and backend are not yet synchronized.
+
+Deletion of `DNSEntry` is always applied immediately to avoid possible extreme delays on shoot deletion.
+
+As the status of `DNSEntries` created by the `shoot-dns-service` in the CP is already forwarded to the source objects in the shoot cluster in form
+of events, nothing needs to be added here. Additionally, the status for shoot-side `DNSEntries` is updated accordingly.
+
+On replicating a `DNSEntry` from the shoot cluster nothing must be changed, as annotations are not copied. This means a user cannot escape the rate limiting by adding an annotation like `dns.gardener.cloud/not-rate-limited: "true"`
+
+### Gardener
```
danielfoehrKn, line 101: How is this rate limit for the default provider determined? 
Should that be defined by an operator during the Gardener installation e.g together with the default domain secret?

Just an idea: what about setting the maximum daily requests possible for a certain provider (and region?) and Gardener determines a "fair share" of requests per Shoot cluster. 
This limit could grow and shrink based on the number of Shoot clusters.
However, it is possibly too complex to implement & no benefit besides having too low/high rate limits.

rfranzke:
Please add the document to the [index](https://github.com/gardener/gardener/blob/master/docs/README.md#proposals).

docs/proposals/17-ratelimits-for-shoot-dns-entries.md
MartinWeindel: We have collected some statistics from samples over one week on the canary and live landscapes. The maximum seen number of DNS entries created by the shoot-dns-service within one hour was 10. This number only contains DNS entries in the default domain.
At least with the current usage pattern of our users, this should be a good starting point (i.e. a rate of 240 creations/updates per day).

MartinWeindel:
> Please add the document to the [index](https://github.com/gardener/gardener/blob/master/docs/README.md#proposals).

done

docs/proposals/17-ratelimits-for-shoot-dns-entries.md
```diff
@@ -0,0 +1,112 @@
...
+  rateLimit:
+    requestsPerDay: 120
+    burst: 20
+```
+
+Each valid `DNSEntry` is already automatically assigned to an `DNSProvider` by matching the domain name and the realm.
+This assignment now also includes the regime of the rate limit of this `DNSProvider`.
+The creation or update of DNS records is postponed until the rate limit restrictions are fulfilled.
+The entry stays in the state `Pending` or `Updating` until the request is applied on the DNS cloud service.
+Here the state `Updating` is a new, additional state to be implemented to make transparent that spec and backend are not yet synchronized.
```
timuthy, line 92: Can you elaborate why we need `Updating`, i.e., `Pending` is not sufficient?

docs/proposals/17-ratelimits-for-shoot-dns-entries.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Goals
+
+- The rate limit is set and applied separatedly for each `DNSProvider`
+- All update requests caused by `DNSEntries` created or updated from the `shoot-dns-service` should only be forwarded to the backend DNS service, if the update requests are within rate limit of the `DNSProvider` the `DNSEntry` is assigned to.
```
timuthy:
```suggestion
- All update requests caused by `DNSEntries` created or updated from the `shoot-dns-service` should only be processed by the backend DNS service, if the update requests are within rate limit of the `DNSProvider` the `DNSEntry` is assigned to.
```

Because forwarding happens since the `shoot-dns-service` doesn't know anything about rate limits, right?

docs/proposals/17-ratelimits-for-shoot-dns-entries.md
```diff
@@ -0,0 +1,112 @@
...
+
+This proposal involves changes in two components:
+
+- [dns-controller-manager](https://github.com/gardener/external-dns-management) for implementation of rate limiting
+
+- [gardener](https://github.com/gardener/gardener) for configuration of rate limits, applying them to the external `DNSProvider` and annotating `DNSEntries` to exclude then from rate limiting
+
+### DNS controller manager
+
+The spec of a `DNSProvider` is extended by an optional `rateLimit` section.
```
timuthy, line 57: Will the `status` also be enhanced to track the daily limits?

docs/proposals/17-ratelimits-for-shoot-dns-entries.md
```diff
@@ -0,0 +1,112 @@
...
+Deletion of `DNSEntry` is always applied immediately to avoid possible extreme delays on shoot deletion.
+
+As the status of `DNSEntries` created by the `shoot-dns-service` in the CP is already forwarded to the source objects in the shoot cluster in form
+of events, nothing needs to be added here. Additionally, the status for shoot-side `DNSEntries` is updated accordingly.
+
+On replicating a `DNSEntry` from the shoot cluster nothing must be changed, as annotations are not copied. This means a user cannot escape the rate limiting by adding an annotation like `dns.gardener.cloud/not-rate-limited: "true"`
+
+### Gardener
+
+The rate limit for the default external `DNSProvider` is configured for the default domains of the garden.
```
timuthy, line 103: Which will probably be another "well known" annotation (e.g. `dns.gardener.cloud/rate-limit`) in the [default domain secret](https://github.com/gardener/gardener/blob/7616e05b6dedd94b4c6e48c26ac026da4d177191/example/10-secret-default-domain.yaml#L13). I thought about alternatives like configuring default values via `ControllerRegistration` or adding support to specify an entire `DNSProvider` in Gardener (via `Unstructured`), but yet another annotation might be the best option to choose for now. WDYT?

timuthy (changes requested):
Thanks for your GEP @MartinWeindel. I left a view comments in the review, but in general it sounds good to me.

docs/proposals/17-ratelimits-for-shoot-dns-entries.md
MartinWeindel: The `Pending` state is only used for newly created `DNSEntries`, .i.e the entry is requested but not yet realised as a DNS record. If a `DNSEntry` is updated, there is currently no state signalling that there is difference between specified and active configuration. Typically the synchronisation happens within seconds, but with the rate limit this could take much longer now. The state `Updating` would signal that a DNS record is already realised, but that it may not reflect the latest changes in the spec.

docs/proposals/17-ratelimits-for-shoot-dns-entries.md
MartinWeindel: If the standard K8s client rate limiter is used (`NewTokenBucketRateLimiter`), you only know if a request would be allowed or not at the moment. There is no value representing "how much" of the daily limit is used. Anyway it is used in a way that there is no difference between requests/day or requests/second. I.e. the allowed requests are distributed equally over the day. And the rate limiter would be reset on restart, i.e. it is not persisted as this would much more laborious.
Instead I would rather see metrics for counting the requests per provider and for yes/no if the request limit is reached at at given point of time.

docs/proposals/17-ratelimits-for-shoot-dns-entries.md
MartinWeindel: Looks ok for me.

timuthy: Can you add this explanation to the GEP?

MartinWeindel: done, see last commit https://github.com/gardener/gardener/pull/4000/commits/9f61c6d558f328bf0cf96fbacb3fd7fa62916db4

timuthy (approved):
/lgtm
@mandelsoft do you want to take another look?

rfranzke (changes requested):
Can this proposal be adapted based on https://github.com/gardener/gardener/pull/4231? Ideally, g/g would not even need to be aware of any specifics for the `shoot-dns-service` extension.

MartinWeindel:
@rfranzke 
> Ideally, g/g would not even need to be aware of any specifics for the shoot-dns-service extension.

Are you thinking in the direction, that the shoot-dns-service should be responsible for creating the `DNSProvider` as defined at `spec.dns.providers` of the shoot manifest?

rfranzke:
@MartinWeindel Definitely, yes. If the approach discussed last week and proposed in #4231 will be pursued then g/g won't know anything about the `dns.gardener.cloud/v1alpha1` API anymore, I assume. In this case, the `shoot-dns-service` will be responsible for the `DNSProvider` and `DNSOwner` objects for shoot clusters.

rfranzke:
Should this PR be closed due to the recent architectural changes? I'd hope that there wouldn't be any adaptations required in g/g?

MartinWeindel:
Rate limiting makes still sense to avoid high request numbers from a misbehaving shoot cluster.
As far as understand the recent architectural changes, at least the `external` DNSProvider must still be created by Gardener.
Is this correct? @stoyanr

rfranzke:
Gardener won't know any `dns.gardener.cloud/v1alpha1` APIs anymore. Instead, the `shoot-dns-service` extension should take over reconciliation of the `DNSProvider` objects in the future.

MartinWeindel:
Closed as rate limits should be realized without adaptions in g/g because of recent decision to deal Gardener DNS records without `DNSProviders`.
Default rate limits will be configured in the `ControllerDeployment` of the `dns-external` extension. For custom providers there will be some way to overwrite the defaults in the provider config of the shoot-dns-service. 
See https://github.com/gardener/gardener-extension-shoot-dns-service/issues/70

PR #4001, opened 2021-05-06
Fix SeedSettingExcessCapacityReservation in seed example

voelzmo:
/area documentation
/kind bug

SeedSettings doesn't have a key excessCapacity:
https://gardener.cloud/documentation/references/core/#core.gardener.cloud/v1beta1.SeedSettings

Therefore, using the example currently fails with 
```
error validating data: ValidationError(Seed.spec.settings): unknown field "excessCapacity" in com.github.gardener.gardener.pkg.apis.core.v1beta1.SeedSettings; if you choose to ignore these errors, turn validation off with --validate=false
```

**Release note**:
```other operator
NONE
```

### DIFF
--- a/example/50-seed.yaml
+++ b/example/50-seed.yaml
@@ -58,7 +58,7 @@ spec:
     blockCIDRs:
     - 169.254.169.254/32
   settings:
-    excessCapacity:
+    excessCapacityReservation:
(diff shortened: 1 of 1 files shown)


timuthy (approved):
/lgtm

PR #4005, opened 2021-05-07
[GEP-15] Add Bastion controller

xrstf:

/area ops-productivity
/kind enhancement

This implements the "expiration controller", responsible for cleaning up expired/rotten Bastions in the garden cluster.

**Release note**:

```other operator
NONE
```


pkg/controllermanager/controller/bastion/bastion_control_test.go
```diff
@@ -0,0 +1,116 @@
...
+					},
+				}
+
+				return nil
+			})
+
+			deleted := false
+			mockClient.EXPECT().Delete(ctx, gomock.AssignableToTypeOf(&operationsv1alpha1.Bastion{})).Do(func(_ context.Context, _ *operationsv1alpha1.Bastion) {
+				deleted = true
+			})
```
xrstf: I don't think this is the best way to assert that the `Delete` endpoint has been called exactly once. Please advise on how to do this better. ^^

xrstf:
@rfranzke The verify job seems to have flaked, but I think it's unrelated. Other than that, I just need to test this PR a bit.

xrstf:
Tested it locally, seems to work well, producing logs like

> time="2021-05-07T16:44:21+02:00" level=info msg="Deleting expired bastion" bastion=garden-dev/my-first-bastion controller=bastion-controller expired="2021-05-07 01:18:54 +0200 CEST"
> 
> time="2021-05-07T16:43:33+02:00" level=info msg="Deleting bastion because it reached its maximum lifetime" bastion=garden-dev/my-first-bastion controller=bastion-controller created="2021-05-07 00:18:54 +0200 CEST"

pkg/controllermanager/apis/config/types.go
```diff
@@ -80,6 +84,17 @@ type ControllerManagerControllerConfiguration struct {
...
 
+// BastionControllerConfiguration defines the configuration of the Bastion
+// controller.
+type BastionControllerConfiguration struct {
+	// ConcurrentSyncs is the number of workers used for the controller to work on
+	// events.
+	ConcurrentSyncs int
+	// MaxLifetime is the maximum time a Bastion resource can exist before it is
+	// forcefully deleted.
+	MaxLifetime time.Duration
```
rfranzke: Most other fields use `metav1.Duration` instead of `time.Duration`:

```suggestion
	MaxLifetime metav1.Duration
```

(same for versioned type)

charts/gardener/controlplane/charts/runtime/templates/controller-manager/configmap-componentconfig.yaml
```diff
@@ -31,6 +31,11 @@ data:
       kubeconfig: /etc/gardener-controller-manager/kubeconfig/kubeconfig
       {{- end }}
     controllers:
+      {{- if .Values.global.controller.config.controllers.bastion }}
```
rfranzke, line 34: Maybe you can also extend https://github.com/gardener/gardener/blob/master/docs/concepts/controller-manager.md#control-loops with information to this controller, and https://github.com/gardener/gardener/blob/master/example/20-componentconfig-gardener-controller-manager.yaml with example configuration

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -96,6 +98,17 @@ type ControllerManagerControllerConfiguration struct {
...
 
+// BastionControllerConfiguration defines the configuration of the Bastion
+// controller.
+type BastionControllerConfiguration struct {
+	// ConcurrentSyncs is the number of workers used for the controller to work on
+	// events.
+	ConcurrentSyncs int `json:"concurrentSyncs"`
+	// MaxLifetime is the maximum time a Bastion resource can exist before it is
+	// forcefully deleted.
+	MaxLifetime time.Duration `json:"maxLifetime"`
```
rfranzke: Should this be optional and defaulted?

pkg/controllermanager/controller/bastion/bastion.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/gardener/gardener/pkg/logger"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	ControllerName = "bastion-controller"
```
rfranzke, line 40: doc string for exported variable

pkg/controllermanager/controller/bastion/bastion_control.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	// delete the bastion once it has reached its maximum lifetime
+	if time.Since(bastion.CreationTimestamp.Time) > r.maxLifetime {
+		logger.WithField("created", bastion.CreationTimestamp.Time).Info("Deleting bastion because it reached its maximum lifetime")
+		err := r.gardenClient.Delete(ctx, bastion)
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, nil
```
rfranzke: Shouldn't we auto-requeue here with the duration of the time until `min(.status.expirationTimestamp, .metadata.creationTimestamp+r.maxLifetime)` so that the `Bastion` gets deleted as soon as one of the deletion condition is met?

pkg/controllermanager/controller/bastion/bastion_control_test.go
rfranzke: I guess this is being made sure by gomock out-of-the-box. There is https://pkg.go.dev/github.com/golang/mock/gomock#Call.Times which defaults to `1`, i.e., if there was another `Delete` call by the client the test will complain. Similarly, it will complain if there is no `Delete` call by the client. Hence, you can remove the `deleted` variable entirely.

rfranzke (changes requested):

pkg/controllermanager/controller/bastion/bastion_control.go
xrstf: Can we add like 3 seconds on top of that, to prevent any timing skews? I don't feel too comfortable re-queueing with millisecond precision. ^^

pkg/controllermanager/controller/bastion/bastion_control_test.go
xrstf: That is what I tried, but even with

```
mockClient.EXPECT().Delete(ctx, gomock.AssignableToTypeOf(&operationsv1alpha1.Bastion{})).Times(2)
```

the tests still succeed. Removing the EXPECT line correcly fails the test with

> Unexpected call to *client.MockClient.Delete([context.TODO &Bastion{ObjectMeta:{bastion      0 2021-05-08 07:31:53.71348648 +0200 CEST m=-21599.974411513 <nil> <nil> map[] map[] [] []  []},Spec:BastionSpec{ShootRef:{},SeedName:nil,ProviderType:nil,SSHPublicKey:,Ingress:[]BastionIngressPolicy{},},Status:BastionStatus{Ingress:nil,Conditions:[]Condition{},LastHeartbeatTimestamp:<nil>,ExpirationTimestamp:2021-05-08 13:31:48.713486958 +0200 CEST m=-4.974411038,},}]) at /home/xrstf/gospace/src/github.com/gardener/gardener/pkg/controllermanager/controller/bastion/bastion_control.go:95 because: there are no expected calls of the method "Delete" for that receiver

xrstf: I forgot to call `mockCtrl.Finish()`.

pkg/controllermanager/controller/bastion/bastion_control.go
xrstf: Nevermind, it should be fine.

pkg/controllermanager/apis/config/v1alpha1/types.go
xrstf: I made it optional now; I assume it's already defaulted by the `defaults.go`, or is there more defaulting action somewhere else?

pkg/controllermanager/apis/config/v1alpha1/defaults_test.go
```diff
@@ -17,32 +17,36 @@ package v1alpha1_test
 import (
 	"time"
 
+	configv1alpha1 "github.com/gardener/gardener/pkg/controllermanager/apis/config/v1alpha1"
```
rfranzke: Nit: Usually, we dot-imported the tested package 
```suggestion
	. "github.com/gardener/gardener/pkg/controllermanager/apis/config/v1alpha1"
```

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -96,6 +98,17 @@ type ControllerManagerControllerConfiguration struct {
...
 
+// BastionControllerConfiguration defines the configuration of the Bastion
+// controller.
+type BastionControllerConfiguration struct {
+	// ConcurrentSyncs is the number of workers used for the controller to work on
+	// events.
+	ConcurrentSyncs int `json:"concurrentSyncs"`
+	// MaxLifetime is the maximum time a Bastion resource can exist before it is
+	// forcefully deleted (defaults to '24h').
+	MaxLifetime *metav1.Duration `json:"maxLifetime"`
```
```suggestion
	// forcefully deleted (defaults to '24h').
	// +optional
	MaxLifetime *metav1.Duration `json:"maxLifetime,omitempty"`
```

pkg/controllermanager/controller/factory.go
```diff
@@ -20,11 +20,12 @@ import (
 
 	gardencore "github.com/gardener/gardener/pkg/apis/core"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
-	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+	gardencorev1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```
rfranzke: Why was this renamed? :) I think `v1beta1constants` is usually used.

rfranzke (commented):
Very nice, looks great, a few minor nits before lgtm :)

pkg/controllermanager/apis/config/v1alpha1/types.go
rfranzke: Nope, that's it - looks good, thanks!

pkg/controllermanager/controller/factory.go
xrstf: I ran my formatter to ensure consistent imports, but I was already wary in this and the changed dot-import for the tests. Guess I just revert the last commit.

pkg/controllermanager/controller/bastion/bastion_control.go
```diff
@@ -0,0 +1,120 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	now := time.Now()
+
+	// delete the bastion once it has expired
+	if bastion.Status.ExpirationTimestamp != nil && now.After(bastion.Status.ExpirationTimestamp.Time) {
+		logger.WithField("expired", bastion.Status.ExpirationTimestamp.Time).Info("Deleting expired bastion")
+		err := r.gardenClient.Delete(ctx, bastion)
+		return reconcile.Result{}, err
```
timebertt:
```suggestion
		return reconcile.Result{}, client.IgnoreNotFound(r.gardenClient.Delete(ctx, bastion))
```

pkg/controllermanager/controller/bastion/bastion_control.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	// delete the bastion once it has reached its maximum lifetime
+	if time.Since(bastion.CreationTimestamp.Time) > r.maxLifetime {
+		logger.WithField("created", bastion.CreationTimestamp.Time).Info("Deleting bastion because it reached its maximum lifetime")
+		err := r.gardenClient.Delete(ctx, bastion)
+		return reconcile.Result{}, err
```
timebertt: same here

pkg/controllermanager/apis/config/v1alpha1/types.go
```diff
@@ -96,6 +98,18 @@ type ControllerManagerControllerConfiguration struct {
...
+// BastionControllerConfiguration defines the configuration of the Bastion
+// controller.
+type BastionControllerConfiguration struct {
+	// ConcurrentSyncs is the number of workers used for the controller to work on
+	// events.
+	ConcurrentSyncs int `json:"concurrentSyncs"`
+	// MaxLifetime is the maximum time a Bastion resource can exist before it is
+	// forcefully deleted (defaults to '24h').
+	// +optional
+	MaxLifetime *metav1.Duration `json:"maxLifetime"`
```
```suggestion
	MaxLifetime *metav1.Duration `json:"maxLifetime,omitempty"`
```

rfranzke (commented):
You forgot 1 of the suggestions from a previous feedback :)

pkg/controllermanager/apis/config/v1alpha1/types.go
xrstf: Dammit. :D

xrstf: Amended the previous commit.

xrstf: No wait, maybe I need to regen code first..

xrstf: No codegen needed this time.

rfranzke: Seems to not be needed :)

rfranzke (dismissed):
The changes proposed in this PR
/lgtm

Though, what about https://github.com/gardener/gardener/pull/4006#discussion_r628727626? It seems you plan to implement this separately/in another PR, is it? Even if, fine for me, but maybe we should then start collecting the open TODOs in an issue to not loose track. :)

xrstf:
I am not planning on implementing any migration for the bastion as of now. Or at least migrations would be by far the least important thing on the todo list, unless shoots are constantly moved from seed to seed (if seed migration happens once in a blue moon and users can simply create another Bastion if they need to, then I don't see the urgency in migration).

rfranzke:
Yes, that's all agreed, I think. We don't need migration for `Bastion`s. However, what we probably need is to cater with the situation that a `Shoot` is being migrated from seed A to B while there are still related `Bastion`s.

In https://github.com/gardener/gardener/pull/4006#discussion_r628727626, the suggestion was:

> to enhance the `Bastion` controller in GCM (added via #4005) to auto-delete the `Bastion` resources referencing a shoot whose `.spec.seedName` changes.

Do you think this makes sense and if yes, do you want to add it to this PR or implement it separately (in which case we should open/enhance some issue to track the open points)? :)

xrstf:
> Do you think this makes sense

totally!

> do you want to add it to this PR

Seems simple enough, I'll try to add it to this PR.

rfranzke:
Okay, great, thank you :)
Then let's
/hold
until the new changes arrive.

Also,
/invite @petersutter
for a review.

petersutter:
looks good from my side 👍 . Awaiting final changes

xrstf:
@petersutter @rfranzke implemented the deletion if the seed doesn't match anymore

pkg/controllermanager/controller/bastion/bastion_control.go
```diff
@@ -89,6 +93,23 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, nil
 	}
 
+	// fetch associated Shoot
+	shoot := gardencorev1beta1.Shoot{}
+	shootKey := kutil.Key(bastion.Namespace, bastion.Spec.ShootRef.Name)
+	if err := r.gardenClient.Get(ctx, kutil.Key(bastion.Namespace, bastion.Spec.ShootRef.Name), &shoot); err != nil {
+		return reconcile.Result{}, fmt.Errorf("could not get shoot %v: %v", shootKey, err)
```
petersutter, line 147: if the shoot is gone, we should probably also delete the bastion, or not?

rfranzke: Good point. Probably we should delete all `Bastion`s as soon as a shoot gets a deletion timestamp.

petersutter: Should you be able to create a bastion for a shoot with a deletion timestamp?

xrstf: How? :) Manually? Set an owner ref? Does the graph tooling in Gardener already take care of removing nodes that have edges to a Shoot?

rfranzke: > Should you be able to create a bastion for a shoot with a deletion timestamp?

I cannot think about a scenario in which is would be required, can you? If a shoot has a deletion timestamp you want to get rid of it, and we don't have any special cleanup logic on the worker nodes themselves, so I doubt SSH-ing will be needed?

> Does the graph tooling in Gardener already take care of removing nodes that have edges to a Shoot?

@xrstf Do you mean the resource dependency graph in GAC and the situation that a `Bastion` for a shoot gets deleted? In this case: Yes, the bastion vertex as well as related edges to the shoot vertex get deleted from the graph. Not sure how it is related in this thread, though.

petersutter: > I cannot think about a scenario in which is would be required, can you?

@rfranzke me neither, just wanted to make sure that nothing is overlooked. Fine then

xrstf: > Yes, the bastion vertex as well as related edges to the shoot vertex get deleted from the graph. Not sure how it is related in this thread, though.

I was under the impression that the graph is used to perform cleanup operations in the cluster, to model the dependencies independently of Kubernetes ownerships/finalizers. There is probably some documentation, but that was just my first assumption seeing a graph in the codebase. ^^

rfranzke: :) No, the graph is used to restrict which resources in the garden cluster a gardenlet is allowed to read (similar to how the `NodeAuthorizer` is implemented in Kubernetes).

xrstf: I've added code and tests to handle both missing shoots and shoots that are currently in deletion.

petersutter: However this only cleanses up the bastion on the next reconcile loop of the bastion, which could be the expiration date of the bastion - at the latest.
To clean it up faster I guess we should (rather?/in addition?) set an owner ref so that the bastion is deleted once the shoot is deleted

rfranzke: Looking at the implementation of the AWS extension, this does not work. It depends on the `extensions.gardener.cloud/v1alpha1.Cluster` resource which is already gone after the `Shoot` is deleted.

So, what about the following:
1. Either prevent `Shoot` deletion entirely if there are still related `Bastion`s in the system (less preferred)
2. At the beginning of both the shoot migration and deletion flows, check if there are still related `Bastion`s in the system, and only continue if the result is `0`. In parallel, this controller here in GCM should watch `Shoot`s and requeue related `Bastion`s if the `.metadata.deletionTimestamp` of the `Shoot` was set to ensure the related `Bastion`s are being cleaned up asap (preferred).

WDYT?

petersutter: +1 for the second option

plus, we should also validate that the shoot is not in deletion when creating a new bastion. Currently we only check if the shoot exists https://github.com/gardener/gardener/pull/3974/files#diff-a56db20e6a9b4aad3584c6d1250433921d7cdf4def9f1272d3de53ba21f3dd3fR129-R139

xrstf: > In parallel, this controller here in GCM should watch Shoots and requeue related Bastions if the .metadata.deletionTimestamp of the Shoot was set to ensure the related Bastions are being cleaned up asap (preferred).

Wouldn't this be solved by the owner ref?

rfranzke: No, the owner reference will make the garbage collector deleting the `Bastion` only after the `Shoot` disappeared from the store. If the shoot has only a non-`nil` `.metadata.deletionTimestamp` then nothing happens.

xrstf: > At the beginning of both the shoot migration and deletion flows, check if there are still related Bastions in the system, and only continue if the result is 0. In parallel, this controller here in GCM should watch Shoots and requeue related Bastions if the .metadata.deletionTimestamp of the Shoot was set to ensure the related Bastions are being cleaned up asap (preferred).

@rfranzke If we only trigger a reconcile for the bastions when a Shoot was deleted, what happens when a Shoot is being migrated? If I understood you correctly, the Shoot controller in the GCM would notice the existing Bastions and refuse the migration. But then nothing further happens, because the Bastion controller in the GCM does nothing. Shouldn't the Bastion controller also react to "shoots that are scheduled for migration", so that it can cleanup existing Bastions to allow the shoot migration to start? Or do we really want to just wait until all Bastions reach their natural end-of-life before migrating the shoot?

pkg/controllermanager/controller/bastion/bastion_control.go
```diff
@@ -0,0 +1,193 @@
...
+	// only shoot deletions should trigger this, so we can cleanup Bastions
+	if shoot.DeletionTimestamp == nil {
+		return
+	}
+
+	// list all bastions that reference this shoot
+	bastionList := operationsv1alpha1.BastionList{}
+	listOptions := client.ListOptions{
+		Namespace: shoot.Namespace,
+		// FieldSelector: fields.SelectorFromSet(fields.Set{operations.BastionShootName: shoot.Name}),
```
xrstf: @petersutter This should use the field selector, but I was not able to get it to work, hence why I went the slow road and filter in the loop below.

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -384,6 +401,28 @@ func (c *Controller) isSeedReadyForMigration(seed *gardencorev1beta1.Seed) error
 	return health.CheckSeedForMigration(seed, c.identity)
 }
 
+func (c *Controller) countShootBastions(ctx context.Context, shoot *gardencorev1beta1.Shoot, gardenClient kubernetes.Interface) (int, error) {
+	// list all bastions that reference this shoot
+	bastionList := operationsv1alpha1.BastionList{}
+	listOptions := client.ListOptions{
+		Namespace: shoot.Namespace,
+		// FieldSelector: fields.SelectorFromSet(fields.Set{operations.BastionShootName: shoot.Name}),
```
xrstf: Same here, this should also work, but doesn't :-D

pkg/controllermanager/controller/bastion/bastion_control.go
petersutter: I guess it's because only `BastionSeedName` is indexed
https://github.com/xrstf/gardener/blob/bastion-controller/pkg/registry/operations/bastion/storage/storage.go#L69

xrstf: I tried that, but since https://github.com/gardener/gardener/blob/master/vendor/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L313-L315 doesn't allow multiple TriggerFuncs, I thought I had to revert. Not sure if "find bastion by seed" or "find bastion by shoot" is more important.

petersutter: I would say that "find bastion by seed" is more important. Ok then leave it as is with the loop and just remove the fieldSelector comment

xrstf: Fixed.

petersutter (dismissed):
/lgtm

pkg/controllermanager/controller/bastion/bastion_control.go
```diff
@@ -0,0 +1,190 @@
...
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		logger.Logger.Errorf("Couldn't get key for object %+v: %v", obj, err)
+		return
+	}
+	c.bastionQueue.Add(key)
+}
+
+func (c *Controller) bastionUpdate(_, newObj interface{}) {
+	c.bastionAdd(newObj)
```
rfranzke, line 45: Shouldn't we only enqueue if there is a relevant change (i.e., only if the `.metadata.generation` does not match the `.status.observedGeneration`)?

pkg/controllermanager/controller/bastion/bastion_control.go
```diff
@@ -0,0 +1,190 @@
...
+	// only shoot deletions should trigger this, so we can cleanup Bastions
+	if shoot.DeletionTimestamp == nil {
+		return
+	}
+
+	// list all bastions that reference this shoot
+	bastionList := operationsv1alpha1.BastionList{}
+	listOptions := client.ListOptions{Namespace: shoot.Namespace}
+
+	if err := c.gardenClient.List(context.TODO(), &bastionList, &listOptions); err != nil {
```
rfranzke: Can you pass a proper context here, please (similar to https://github.com/gardener/gardener/blob/90f60bb2c5cf4c167505016f3ff1c93a5a070966/pkg/controllermanager/controller/controllerregistration/controllerregistration.go#L119)?

rfranzke: Does it make sense to introduce a field selector for `.spec.shootRef.name` and use it here?

pkg/controllermanager/controller/bastion/bastion_control.go
rfranzke: > Shouldn't the Bastion controller also react to "shoots that are scheduled for migration", so that it can cleanup existing Bastions to allow the shoot migration to start? Or do we really want to just wait until all Bastions reach their natural end-of-life before migrating the shoot?

Yes, of course, we should! I haven't written it in my description above, sorry. We should delete all `Bastion`s if a shoot's deletion timestamp gets set or if migration is triggered (by changing its seed name).

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -384,6 +401,25 @@ func (c *Controller) isSeedReadyForMigration(seed *gardencorev1beta1.Seed) error
 	return health.CheckSeedForMigration(seed, c.identity)
 }
 
+func (c *Controller) countShootBastions(ctx context.Context, shoot *gardencorev1beta1.Shoot, gardenClient kubernetes.Interface) (int, error) {
+	// list all bastions that reference this shoot
+	bastionList := operationsv1alpha1.BastionList{}
+	listOptions := client.ListOptions{Namespace: shoot.Namespace}
+
+	if err := gardenClient.Client().List(context.TODO(), &bastionList, &listOptions); err != nil {
```
rfranzke:
```suggestion
	if err := gardenClient.Client().List(ctx, &bastionList, &listOptions); err != nil {
```

rfranzke: You could use the same field selector as described above.

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -384,6 +401,25 @@ func (c *Controller) isSeedReadyForMigration(seed *gardencorev1beta1.Seed) error
 	return health.CheckSeedForMigration(seed, c.identity)
 }
 
+func (c *Controller) countShootBastions(ctx context.Context, shoot *gardencorev1beta1.Shoot, gardenClient kubernetes.Interface) (int, error) {
+	// list all bastions that reference this shoot
+	bastionList := operationsv1alpha1.BastionList{}
+	listOptions := client.ListOptions{Namespace: shoot.Namespace}
```
rfranzke: To prevent overloading client/server, you can add `client.Limit(1)` to the list options. You are only interested whether there is at least one bastion, you are not interested in the concrete amount.

pkg/registry/operations/bastion/strategy_test.go
```diff
@@ -23,14 +23,15 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/core/v1"
```
rfranzke:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

pkg/registry/operations/bastion/strategy.go
```diff
@@ -180,6 +180,16 @@ func MatchBastion(label labels.Selector, field fields.Selector) storage.Selectio
 	}
 }
 
+// ShootNameTriggerFunc returns spec.shootRef.name of given Bastion.
+func ShootNameTriggerFunc(obj runtime.Object) string {
```
rfranzke: Is this used somewhere? 👀

pkg/gardenlet/controller/shoot/shoot_control.go
rfranzke: In order to make it work, you probably need something similar to https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/factory.go#L201-L231

xrstf: Do you insist on using a Field selector? Peter and me had settled on just fetching all Bastions, considering how there will be probably very few Bastions at any given time in a project.

pkg/controllermanager/controller/bastion/bastion_control.go
xrstf: Yeah sure i guess

xrstf: Or not. `operationsv1alpha1.BastionStatus` does not have `ObservedGeneration`.

rfranzke: Interesting 👀 Shouldn't it get it then?

pkg/controllermanager/controller/bastion/bastion_control.go
xrstf: Had that. Didn't work, because K8s only supprts a single IndexerFunc: https://github.com/gardener/gardener/pull/4005/commits/a00cb3fe4bc73cbebfc7d8f2fd33d35b72be5c03

pkg/gardenlet/controller/shoot/shoot_control.go
rfranzke: I don't insist, probably your assumption is fair.

pkg/controllermanager/controller/bastion/bastion_control.go
xrstf: Does it really save anything? Reconciling is quick and painless and happens mostly based on cached data.

rfranzke: > Reconciling is quick and painless and happens mostly based on cached data.

OK, then let's keep it as is.

pkg/controllermanager/controller/bastion/bastion_control.go
xrstf: Looks like I've already implemented it that way ^^ https://github.com/gardener/gardener/pull/4005/files#diff-756a27d8ca9fb61d2ebea7d5b726400f0c0d4edff1965c0de26e65cd9c23c4bcR150-R166

pkg/controllermanager/controller/bastion/bastion_control.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// the Shoot for this bastion has been migrated to another Seed, we have to garbage-collect
+	// the old bastion (bastions are not migrated, users are required to create new bastions);
+	// equality is the correct check here, as the admission plugin already prevents Bastions
+	// from existing without a spec.SeedName being set. So it cannot happen that we accidentally
+	// delete a Bastion without seed (i.e. an unreconciled, new Bastion);
+	// under normal operations, shoots cannot be migrated to another seed while there are still
+	// bastions for it, so this check here is just a safety measure.
+	if !equality.Semantic.DeepEqual(shoot.Status.SeedName, bastion.Spec.SeedName) {
```
rfranzke: This should be 
```suggestion
	if !equality.Semantic.DeepEqual(shoot.Spec.SeedName, bastion.Spec.SeedName) {
```

Otherwise, the `Bastion` will only be deleted when the shoot has already been migrated (which will prevent a graceful deletion of the bastion).

pkg/gardenlet/controller/shoot/shoot_control.go
rfranzke: This comment was not resolved

pkg/gardenlet/controller/shoot/shoot_control.go
rfranzke: This comment was not resolved

pkg/gardenlet/controller/shoot/shoot_control.go
```diff
@@ -23,8 +23,9 @@ import (
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
-	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
+	v1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
```
rfranzke, line 26: Why was this changed? 👀

xrstf: gimps. Why have the import above be `v1beta1constants` and not `gardencorev1beta1constants`?

rfranzke: I don't know, there is no policy I think, just "soft conventions". The constants pkg was always called `v1beta1constants` and the helper pkg was always called `gardencorev1beta1helper`. Anyways, I was just wondering about this (apparently unrelated) change.

rfranzke (commented):
looks good now, no further comments except that the `verify` step needs to be fixed :)

pkg/gardenlet/controller/shoot/shoot_control.go
xrstf: I run the formatter on all files I touch in a PR, and I thought that this change was surely uncontroversial, as it just applies the convention consistently. ;)

rfranzke (approved):
/lgtm
/hold until v1.23 is released

PR #4006, opened 2021-05-07
[GEP-15] Add Bastion handling to gardenlet

xrstf:

/area ops-productivity
/kind enhancement

This PR adds the necessary logic to the gardenlet to sync Bastion resources from the Garden into each relevant Seed cluster.

Fixes #

**Release note**:

```other operator
NONE
```


pkg/gardenlet/apis/config/v1alpha1/types.go
```diff
@@ -129,6 +129,9 @@ type GardenletControllerConfiguration struct {
 	// BackupEntry defines the configuration of the BackupEntry controller.
 	// +optional
 	BackupEntry *BackupEntryControllerConfiguration `json:"backupEntry"`
+	// Bastion defines the configuration of the Bastion controller.
+	// +optional
+	Bastion *BastionControllerConfiguration `json:"bastion"`
```
rfranzke:
```suggestion
	Bastion *BastionControllerConfiguration `json:"bastion,omitempty"`
```

(please also add it to `BackupBucket` and `BackupEntry`, it probably was forgotten

rfranzke: Please extend https://github.com/gardener/gardener/blob/master/example/20-componentconfig-gardenlet.yaml and https://github.com/gardener/gardener/blob/master/charts/gardener/gardenlet/charts/runtime/templates/configmap-componentconfig.yaml

pkg/gardenlet/controller/bastion/bastion.go
```diff
@@ -0,0 +1,164 @@
...
+// NewBastionController takes a Kubernetes client for the Garden clusters <k8sGardenClient>, a struct
+// holding information about the acting Gardener, a <bastionInformer>, and a <recorder> for
+// event recording. It creates a new Gardener controller.
+func NewBastionController(ctx context.Context, clientMap clientmap.ClientMap, config *config.GardenletConfiguration, recorder record.EventRecorder) (*Controller, error) {
+	gardenClient, err := clientMap.GetClient(ctx, keys.ForGarden())
+	if err != nil {
+		return nil, fmt.Errorf("failed to get garden client: %w", err)
+	}
+
+	bastionInformer, err := gardenClient.Cache().GetInformer(ctx, &gardencorev1alpha1.Bastion{})
```
rfranzke: This should be `operationsv1alpha1` now :)

pkg/gardenlet/controller/bastion/bastion.go
```diff
@@ -0,0 +1,164 @@
...
+	var waitGroup sync.WaitGroup
+
+	c.bastionInformer.AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: controllerutils.BastionFilterFunc(ctx, c.gardenClient, confighelper.SeedNameFromSeedConfig(c.config.SeedConfig), c.config.SeedSelector),
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc:    c.bastionAdd,
+			UpdateFunc: c.bastionUpdate,
+			DeleteFunc: c.bastionDelete,
+		},
+	})
```
rfranzke: Can you move this up into the `NewBastionController` function? I see that you were inspired from the `BackupEntry` controllers and I have no explanation why it was done like this, but I guess most controllers register the event handlers already in the `New()` functions. Maybe you can also move it up in both the `BackupBucket`/`BackupEntry` controllers

pkg/gardenlet/controller/bastion/bastion.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		logger.Logger.Errorf("Couldn't get key for object %+v: %v", obj, err)
+		return
+	}
+
+	c.bastionQueue.Add(key)
+}
+
+func (c *Controller) bastionUpdate(_, newObj interface{}) {
+	c.bastionAdd(newObj)
```
rfranzke: Should we only call `bastionAdd` if the `.metadata.generation` changes?

pkg/gardenlet/controller/factory.go
```diff
@@ -101,6 +102,7 @@ func (f *GardenletControllerFactory) Run(ctx context.Context) error {
 		// Garden core informers
 		backupBucketInformer           = f.k8sGardenCoreInformers.Core().V1beta1().BackupBuckets().Informer()
 		backupEntryInformer            = f.k8sGardenCoreInformers.Core().V1beta1().BackupEntries().Informer()
+		bastionInformer                = f.k8sGardenCoreInformers.Core().V1alpha1().Bastions().Informer()
```
rfranzke:
```suggestion
		bastionInformer                = f.k8sGardenCoreInformers.Operations().V1alpha1().Bastions().Informer()
```

pkg/operation/botanist/component/extensions/bastion/bastion.go
```diff
@@ -0,0 +1,190 @@
...
+// Deploy uses the seed client to create or update the Bastion custom resource in the Seed.
+func (b *bastion) Deploy(ctx context.Context) error {
+	_, err := b.deploy(ctx)
+	return err
+}
+
+func (b *bastion) deploy(ctx context.Context) (extensionsv1alpha1.Object, error) {
+	bastion := &extensionsv1alpha1.Bastion{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: b.values.Name,
```
rfranzke: Should we create them in the corresponding namespace of the related `Shoot`?

pkg/operation/botanist/component/extensions/bastion/bastion.go
```diff
@@ -0,0 +1,190 @@
...
+	b.values.ProviderConfig = providerConfig
+}
+
+func (b *bastion) SetUserData(userData []byte) {
+	b.values.UserData = userData
+}
+
+func (b *bastion) SetIngress(ingress []gardencorev1alpha1.BastionIngressPolicy) {
+	b.values.Ingress = ingress
+}
```
rfranzke: Why do you need those functions? I guess all of these values are already known by the time the `Interface` is instantiated?

pkg/gardenlet/controller/bastion/bastion_actuator.go
```diff
@@ -0,0 +1,316 @@
...
+		return errors.Wrapf(err, "could not reconcile extension secret in seed")
+	}
+
+	return nil
+}
+
+// deleteBastionExtensionSecret deletes secret referred by Bastion extension resource in seed.
+func (a *actuator) deleteBastionExtensionSecret(ctx context.Context) error {
+	return client.IgnoreNotFound(a.seedClient.Client().Delete(ctx, emptyExtensionSecret(a.bastion)))
+}
```
rfranzke: remove this

pkg/gardenlet/controller/bastion/bastion_actuator.go
```diff
@@ -0,0 +1,316 @@
...
+)
+
+// Actuator acts upon Bastion resources.
+type Actuator interface {
+	// Reconcile reconciles the Bastion.
+	Reconcile(context.Context) error
+	// Delete deletes the Bastion.
+	Delete(context.Context) error
+	// Migrate migrates the Bastion.
+	Migrate(context.Context) error
```
rfranzke: I'm not too sure if should implement this for the `Bastion` resource... probably we shouldn't. I would suggest to enhance the `Bastion` controller in GCM (added via #4005) to auto-delete the `Bastion` resources referencing a shoot whose `.spec.seedName` changes. OK? 

cc @plkokanov @stoyanr

rfranzke (changes requested):

xrstf:
@rfranzke please note that this is a draft and there are many pieces of dead code in here. Maybe hold back on reviews for now :)

pkg/gardenlet/controller/bastion/bastion_actuator.go
xrstf: I had settled on not implementing migration and also not really having an actuator abstraction at all for this controller. It's IMHO not worth it to handle seed migration for bastions if bastions are at most alive for 1 day and one can easily create new bastions once the seed has been changed for a shoot. If I understand migration correctly.

plkokanov: Basically, if you do not want to keep the bastion instances alive during migration, you could do what @rfranzke suggested or delete the `operations.gardener.cloud.Bastion` resource during the shoot control plane migration flow - somewhere [here]( https://github.com/gardener/gardener/blob/master/pkg/gardenlet/controller/shoot/shoot_control_migrate.go#L153) so that the shoot namespace can be properly cleaned up.

xrstf:
The basic workflow should be working fine now. I tested creating/deleting bastions in the garden while observing the extension in the seed and comparing the status.

Right after creation, this is the state (left: seed, right: garden):

screenshot-2021-05-10-222643

After reconciliation:

screenshot-2021-05-10-222750

@petersutter @rfranzke If you want, I would apreciate a quick review if the basics are correct. I can then add some more tests and probably some documentation.

pkg/gardenlet/controller/bastion/bastion.go
```diff
@@ -0,0 +1,177 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	// ControllerName is the name of this controller.
+	ControllerName = "bastion-controller"
+)
+
+// Controller controls BackupEntries.
```
petersutter:
```suggestion
// Controller controls Bastions.
```

pkg/gardenlet/controller/bastion/bastion_reconciler.go
```diff
@@ -0,0 +1,256 @@
...
+
+// reconciler implements the reconcile.Reconcile interface for bastion reconciliation.
+type reconciler struct {
+	clientMap clientmap.ClientMap
+	recorder  record.EventRecorder
+	logger    logrus.FieldLogger
+	config    *config.GardenletConfiguration
+}
+
+// newReconciler returns the new backupBucker reconciler.
```
petersutter:
```suggestion
// newReconciler returns the new bastion reconciler.
```
regarding `backupBucker` could you also fix the typo from where you have copied it? :)

pkg/gardenlet/controller/bastion/bastion.go
```diff
@@ -0,0 +1,177 @@
...
+	bastionInformer runtimecache.Informer
+	bastionQueue    workqueue.RateLimitingInterface
+
+	workerCh               chan int
+	numberOfRunningWorkers int
+}
+
+// NewBastionController takes a Kubernetes client for the Garden clusters <k8sGardenClient>, a struct
+// holding information about the acting Gardener, a <bastionInformer>, and a <recorder> for
+// event recording. It creates a new Gardener controller.
```
petersutter: please update the doc string, it does not seem to be aligned with the parameters

pkg/gardenlet/controller/bastion/bastion_reconciler.go
```diff
@@ -0,0 +1,256 @@
...
+	// retrieve Kubernetes client for Seed cluster
+	seedClient, err := r.clientMap.GetClient(ctx, keys.ForSeedWithName(*bastion.Spec.SeedName))
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get seed client: %w", err)
+	}
+
+	// get Shoot for the bastion
+	shoot := gardencorev1beta1.Shoot{}
+	shootKey := kutil.Key(bastion.Namespace, bastion.Spec.ShootRef.Name)
+	if err := gardenClient.Client().Get(ctx, kutil.Key(bastion.Namespace, bastion.Spec.ShootRef.Name), &shoot); err != nil {
```
petersutter:
```suggestion
	if err := gardenClient.Client().Get(ctx, shootKey, &shoot); err != nil {
```

pkg/gardenlet/controller/bastion/bastion_reconciler.go
```diff
@@ -0,0 +1,256 @@
...
+		return fmt.Errorf("failed ensure bastion extension resource: %v", err)
+	}
+
+	// CreateOrUpdate left the extBastion variable updated, so we can conveniently
+	// check the bastion's status in the seed without having to fetch it again.
+	if extBastion.Status.LastOperation == nil || extBastion.Status.LastOperation.State != gardencorev1beta1.LastOperationStateSucceeded {
+		patchReadyCondition(ctx, gardenClient, bastion, gardencorev1alpha1.ConditionFalse, "Reconciling", "Reconciling has not finished yet.")
+
+		return &ctrlerror.RequeueAfterError{
+			RequeueAfter: 1 * time.Second,
```
petersutter: Could this cause high load for the seed? E.g. I saw some code where it says:
```go
	return reconcile.Result{
		RequeueAfter: 15 * time.Second, // prevent ddos-ing the seed
	}, nil
```

pkg/gardenlet/controller/bastion/bastion_reconciler.go
petersutter: what if the shoot is in deletion?

petersutter (changes requested):

pkg/gardenlet/controller/bastion/bastion_reconciler.go
xrstf: It could. My thought was that during this reconciliation, some users sits in front of their machine and is impatiently waiting for the SSH connection to come up, and I wanted to serve them ASAP.

pkg/gardenlet/controller/bastion/bastion_reconciler.go
xrstf: If we go with https://github.com/gardener/gardener/pull/4005#discussion_r629378055 (i.e. delete the bastion via the GCM), do we need additional handling in the gardenlet?

petersutter: maybe the question is rather: what if the shoot is deleted before the bastion was cleaned up. In this case you just exit the reconcile loop without having the chance to cleanup the bastion, or not?

xrstf: This would also be solved by adding your proposed owner ref, right? Unless someone manually removes the owner ref, so my follow up question would be: would the admission plugin for bastions _always_ (re)set the owner ref, or only once during creation? If it set it always, it might make manual cleanup impossible?

* So do we want to do `bastion.OwnerReferences = []...` or `if admissionOp == Create { bastion.OwnerReferences = []... }`?
* Do we want to overwrite possible other owner refs (which would basically mean not allowing other owner refs? :thinking:) or do we want to `pkg/utils/kubernetes.MergeOwnerReferences()` ?

petersutter:
@xrstf this PR is still marked as `Draft`, do you want to mark it as `Ready for review`?

xrstf:
@petersutter I rebased it now and marked it as ready-for-review. I ran the tests locally and they passed, but on Concourse the backupbucket tests failed. Not sure if that's a flake.

petersutter (dismissed):
/lgtm

except maybe https://github.com/gardener/gardener/pull/4006#discussion_r630106493 @rfranzke do you think this is an issue?

charts/gardener/gardenlet/charts/runtime/templates/configmap-componentconfig.yaml
```diff
@@ -77,6 +77,8 @@ data:
 {{ toYaml .Values.global.gardenlet.config.controllers.backupEntry.deletionGracePeriodShootPurposes | indent 8 }}
         {{- end }}
       {{- if .Values.global.gardenlet.config.controllers.controllerInstallation }}
+      bastion:
+        concurrentSyncs: {{ required ".Values.global.gardenlet.config.controllers.bastion.concurrentSyncs is required" .Values.global.gardenlet.config.controllers.bastion.concurrentSyncs }}
```
rfranzke, line 80: This should not be dependent on the `controllerInstallation` configuration

cmd/gardenlet/app/gardenlet.go
```diff
@@ -399,6 +401,7 @@ func NewGardenlet(ctx context.Context, cfg *config.GardenletConfiguration) (*Gar
 		Recorder:               recorder,
 		ClientMap:              clientMap,
 		K8sGardenCoreInformers: gardencoreinformers.NewSharedInformerFactory(k8sGardenClient.GardenCore(), 0),
+		K8sOperationsInformers: operationsinformers.NewSharedInformerFactory(k8sGardenClient.GardenOperations(), 0),
```
rfranzke: If I see it correctly, this is not used anywhere in the gardenlet controller. Actually, we strive to get rid of the client-go informer factories in favor of the controller-runtime constructs. Please remove it if it's not used.

pkg/gardenlet/controller/factory.go
```diff
@@ -98,6 +103,7 @@ func NewGardenletControllerFactory(
 // Run starts all the controllers for the Garden API group. It also performs bootstrapping tasks.
 func (f *GardenletControllerFactory) Run(ctx context.Context) error {
 	var (
+		bastionInformer                = f.k8sOperationsInformers.Operations().V1alpha1().Bastions().Informer()
```
rfranzke: Please remove this to not start duplicated watches. In the `Bastion` controller you already start a watch with controller-runtime

```golang
	bastionInformer, err := gardenClient.Cache().GetInformer(ctx, &operationsv1alpha1.Bastion{})
	if err != nil {
		return nil, fmt.Errorf("failed to get Bastion Informer: %w", err)
	}
```

See also comment above: We want to get rid of the client-go informers.

pkg/seedadmissioncontroller/webhooks/admission/extensioncrds/admission.go
```diff
@@ -85,7 +85,6 @@ func (h *handler) Handle(ctx context.Context, request admission.Request) admissi
 
 		metav1.GroupVersionResource{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "backupbuckets"},
 		metav1.GroupVersionResource{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "backupentries"},
-		metav1.GroupVersionResource{Group: extensionsv1alpha1.SchemeGroupVersion.Group, Version: extensionsv1alpha1.SchemeGroupVersion.Version, Resource: "bastions"},
```
rfranzke, line 88: If you want to remove `Bastion`s from the deletion protection then there are more places which you'd have to touch, e.g. https://github.com/gardener/gardener/blob/07a56785cb69aff4a1529943dfc8f5d6ec97f015/pkg/operation/botanist/component/seedadmissioncontroller/seedadmissioncontroller.go#L121. Please check the initial PR to find all other places.

OTOH, does it really hurt? I have no strong opinion, but I struggle a bit with the heterogeneity. `Bastion`s would be "special" (not protected) while all other `extensions.gardener.cloud` resources require the confirmation.

pkg/gardenlet/controller/bastion/bastion.go
```diff
@@ -0,0 +1,177 @@
...
+	// ControllerName is the name of this controller.
+	ControllerName = "bastion-controller"
+)
+
+// Controller controls Bastions.
+type Controller struct {
+	gardenClient client.Client
+	config       *config.GardenletConfiguration
+	reconciler   reconcile.Reconciler
+	recorder     record.EventRecorder
```
rfranzke: This seems to be only used in the reconciler, no need to store it here

pkg/gardenlet/controller/bastion/bastion.go
```diff
@@ -0,0 +1,177 @@
...
+)
+
+const (
+	// ControllerName is the name of this controller.
+	ControllerName = "bastion-controller"
+)
+
+// Controller controls Bastions.
+type Controller struct {
+	gardenClient client.Client
```
rfranzke: This seems to be only used in the reconciler, no need to store it here

pkg/gardenlet/controller/bastion/bastion.go
```diff
@@ -0,0 +1,177 @@
...
+	config       *config.GardenletConfiguration
```
rfranzke: This seems to be only used in the `NewBastionController` function and in the reconciler, no need to store it here

pkg/gardenlet/controller/bastion/bastion.go
```diff
@@ -0,0 +1,177 @@
...
+	reconciler   reconcile.Reconciler
+	recorder     record.EventRecorder
+
+	bastionInformer runtimecache.Informer
```
rfranzke: Can you rather work with `hasSyncedFuncs` like in most other controllers and not store the informer here?

pkg/gardenlet/controller/bastion/bastion_reconciler.go
```diff
@@ -0,0 +1,256 @@
...
+const (
+	// finalizerName is the Kubernetes finalizerName that is used to control the cleanup of
+	// Bastion resources in the seed cluster.
+	finalizerName = gardencorev1alpha1.GardenerName
+)
+
+// reconciler implements the reconcile.Reconcile interface for bastion reconciliation.
+type reconciler struct {
+	clientMap clientmap.ClientMap
+	recorder  record.EventRecorder
```
rfranzke: The recorder seems to be unused. Remove it entirely from the controller?

pkg/gardenlet/controller/bastion/bastion_reconciler.go
```diff
@@ -0,0 +1,256 @@
...
+		return nil
+	}); err != nil {
+		patchReadyCondition(ctx, gardenClient, bastion, gardencorev1alpha1.ConditionFalse, "FailedReconciling", err.Error())
+
+		return fmt.Errorf("failed ensure bastion extension resource: %v", err)
+	}
+
+	// CreateOrUpdate left the extBastion variable updated, so we can conveniently
+	// check the bastion's status in the seed without having to fetch it again.
+	if extBastion.Status.LastOperation == nil || extBastion.Status.LastOperation.State != gardencorev1beta1.LastOperationStateSucceeded {
```
rfranzke: What if the `ingress` has changed, for example? The `lastOperation` might still be `Succeeded` from a previous reconciliation, however, the new changed `ingress` values were not yet applied. The gardenlet would wrongly end its reconciliation too fast.
Also, I'm missing the `operation` and `timestamp` annotations (like done for all other extension resources). Maybe https://github.com/gardener/gardener/blob/07a56785cb69aff4a1529943dfc8f5d6ec97f015/pkg/extensions/customresources.go#L49-L76 could be used to wait for the healthiness?

pkg/gardenlet/controller/bastion/bastion_reconciler.go
rfranzke: The reconciliation is quite fast, though, the `CachedRuntimeClient` feature gate is still alpha and not enabled by default, i.e., there is no caching and always direct calls (at least for now/a few more weeks). Maybe something like `5s`, `10s` or `15s` is indeed still good enough for now? OTOH, I don't expect too many `Bastion`s being created and the creation process itself is probably rather fast, so the impact might be negligible.

Personally, I would probably change this to at least `5s` or `10s`, but I don't insist for now due to the probably low frequency of created `Bastion` resources.

rfranzke (changes requested):

charts/gardener/gardenlet/charts/runtime/templates/configmap-componentconfig.yaml
xrstf: Wow, that was eagle eyed :)

cmd/gardenlet/app/gardenlet.go
xrstf: This `K8sOperationsInformers` is handed to `NewGardenletControllerFactory()`, which then uses it to create a Bastion informer.

pkg/seedadmissioncontroller/webhooks/admission/extensioncrds/admission.go
xrstf: I can live with either, but I can undestand why Bastions would be treated more lightly (i.e. no deletion protection), as they are not needed at all for a cluster to function / be operated.

rfranzke: 👍🏻 That's true and an argument for not protecting it. Please proceed as you prefer.

pkg/gardenlet/controller/bastion/bastion_reconciler.go
xrstf: I'm not even sure `WaitUntilExtensionCRReady` would help here. If just the `spec.ingress` changed and nothing else, we cannot easily determine if the changes were reconciled.

WDYT about simply 

```go
// create or update the bastion in the seed cluster
	if _, err := controllerutil.CreateOrUpdate(ctx, seedClient, extBastion, func() error {
		extBastion.Spec.UserData = createUserData(bastion)
		extBastion.Spec.Ingress = extensionsIngress
		extBastion.Spec.Type = *bastion.Spec.ProviderType

		// invalidate previous status, in case the spec'ed ingress has changed and reconciling is needed
		extBastion.Status.Ingress = nil

		return nil
	}); err != nil {
		patchReadyCondition(ctx, gardenClient, bastion, gardencorev1alpha1.ConditionFalse, "FailedReconciling", err.Error())

		return fmt.Errorf("failed ensure bastion extension resource: %v", err)
	}
```

rfranzke: Usually, the gardenlet sets the `gardener.cloud/operation=reconcile` annotation to extension resources. The `WaitUntilExtensionCRReady` function waits until this annotation was removed by the extension controller and until the `status` indicates that the last observed state was successfully reconciled. Why do you think it wouldn't help?

Resetting the status won't work as you cannot patch the `status` with the regular `seedClient`. You'd have to make two separate calls, first resetting the status and then updating the spec (or vice versa), though, I don't see why we should deviate from what is usually done when gardenlet interacts with `extensions.gardener.cloud` resources.

petersutter:
@xrstf not related to this PR, but would you mind to add the small detail to the GEP-15 proposal, that no egress traffic should be allowed from the bastion?
See also the discussion here https://github.com/gardener/gardener-extension-provider-aws/pull/318#issuecomment-824119158

> @xrstf can you restrict the security group of the bastion to the bare minimum? E.g. no egress traffic should be allowed and is not needed. Only the connection from the bastion to the shoot nodes should be required I presume.
> Maybe we can add that also to the GEP-15
> 
> cc @ThormaehlenFred

> @petersutter I have implemented the additional egress restriction for the bastion's security group. The bastion security group can now only establish SSH connections to the worker security group.

pkg/gardenlet/controller/bastion/bastion_reconciler.go
xrstf: Okay, I am reverting the controller to actively wait for reconciliation instead of requeueing.

xrstf:
> but would you mind to add the small detail to the GEP-15 proposal, that no egress traffic should be allowed from the bastion?

Added a sentence.

pkg/gardenlet/controller/bastion/bastion_reconciler.go
```diff
@@ -0,0 +1,266 @@
...
+
+	// wait for the extension controller to reconcile possible changes
+	if err := extensions.WaitUntilExtensionCRReady(
+		ctx,
+		seedClient,
+		logger,
+		func() client.Object { return &extensionsv1alpha1.Bastion{} },
+		extensionsv1alpha1.BastionResource,
+		shoot.Status.TechnicalID,
+		bastion.Name,
```
rfranzke: Nit:
```suggestion
		extBastion.Namespace,
		extBastion.Name,
```

pkg/gardenlet/controller/bastion/bastion_reconciler.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	// create or update the bastion in the seed cluster
+	if _, err := controllerutil.CreateOrUpdate(ctx, seedClient, extBastion, func() error {
+		extBastion.Spec.UserData = createUserData(bastion)
+		extBastion.Spec.Ingress = extensionsIngress
+		extBastion.Spec.Type = *bastion.Spec.ProviderType
+
+		return nil
+	}); err != nil {
```
rfranzke: I think we still need to set the operation and timestamp annotations here, similar to https://github.com/gardener/gardener/blob/7c4388ee8bb0eaecfbae451772dc2577df45cc3f/pkg/operation/botanist/component/extensions/network/network.go#L188-L189

rfranzke (commented):
Thanks for addressing the suggestions! :)

xrstf:
@petersutter I think the failure above is a flake. Can you trigger a retest?

petersutter:
> @petersutter I think the failure above is a flake. Can you trigger a retest?

only with the second testrun it succeeded :D :(

rfranzke (approved):
/lgtm

